PR #6816, opened 2025-07-14
fix BranchDelete

### LINKED ISSUE #6797: bug: branch delete can crash
### Component

API Server / GraphQL

### Infrahub version

1.3.2

### Current Behavior

deleting a very large branch can crash and fail, causing the Branch to appear as though it has been deleted, but leaving all of the edges on this branch intact on the database

### Expected Behavior

deleting a very large branch should be possible

### Steps to Reproduce

1. make a branch with many 1000s of changes
2. try to delete the branch via the UI
3. most likely, it will silently fail. the branch will no longer show up in the Branch menu of the UI, but all of the edges associated with this branch will remain in the database

### Additional Information

_No response_

ajtmccarty:
fixes #6797 

existing logic to delete a branch could run of out memory and crash during the delete query, leaving all of the branch's edges in place on the database, but removing the `Branch` object successfully, so there would be no indication to an end user that the branch was not successfully deleted

this PR does the following
1. move the query to delete the Branch object _after_ the query to delete all of the branch's edges, so if the edge delete query fails, the Branch will remain in the database (although it will have a status of DELETING and will be invisible to the end user, but will be helpful for troubleshooting).
2. update the query to delete the branch's edges (`DeleteBranchRelationshipsQuery`) to handle very large branches. this involves using subqueries with `IN TRANSACTIONS`, so the deletes are batched into smaller transactions and executed instead of trying to do 1 million deletes in 1 transaction. I've tested this locally with a 6 million edge branch and it works.
3. new migration to find any partially deleted Branches and run the `DeleteBranchRelationshipsQuery` on them
4. a new `DELETING` status for a branch
  a. a new `BranchStatus` enum with a `DELETING` value and support for Enum values in `StandardNode` database logic
  b. set the Branch to status=DELETING when starting the delete. deleting a very large branch can take minutes and we should not allow any changes on a branch being deleted while the delete is running.
  c. a few small changes to the logic for retrieving a branch from the database to ignore those with status=DELETING


backend/infrahub/core/branch/enums.py
```diff
@@ -0,0 +1,8 @@
+from infrahub.utils import InfrahubStringEnum
+
+
+class BranchStatus(InfrahubStringEnum):
+    OPEN = "OPEN"
+    NEED_REBASE = "NEED_REBASE"
+    CLOSED = "CLOSED"
+    DELETING = "DELETING"
```
ajtmccarty, line 8: @LucasG0 added this enum in this PR b/c I also need to add another status for a Branches

backend/infrahub/core/branch/models.py
```diff
@@ -147,6 +152,20 @@ async def get_by_name(cls, name: str, db: InfrahubDatabase) -> Branch:
...
+        limit: int = 1000,
+        ids: list[str] | None = None,
+        name: str | None = None,
+        **kwargs: dict[str, Any],
+    ) -> list[Self]:
+        branches = await super().get_list(db=db, limit=limit, ids=ids, name=name, **kwargs)
+        branches = [branch for branch in branches if branch.status != BranchStatus.DELETING]
+
+        return branches
+
```
ajtmccarty, line 168: to make sure DELETING branches are not retrieved from the database

backend/infrahub/core/branch/models.py
```diff
@@ -131,14 +133,17 @@ def update_schema_hash(self, at: Timestamp | str | None = None) -> bool:
...
         MATCH (n:Branch)
         WHERE n.name = $name
+        AND NOT n.status IN $ignore_statuses
         RETURN n
         """
 
-        params = {"name": name}
+        params: dict[str, Any] = {"name": name}
+        if ignore_deleting:
+            params["ignore_statuses"] = [BranchStatus.DELETING.value]
```
ajtmccarty, line 146: to make sure DELETING branches are not retrieved from the database

backend/infrahub/core/node/standard.py
```diff
@@ -191,6 +192,9 @@ def to_db(self) -> dict[str, Any]:
                 continue
 
             attr_value = getattr(self, attr_name)
+            if isinstance(attr_value, Enum):
+                attr_value = attr_value.value
+
```
ajtmccarty, line 197: to handle `Enum` values on `StandardNode`s

backend/infrahub/core/query/branch.py
```diff
@@ -3,43 +3,12 @@
...
-
-        self.params["node_id"] = db.to_database_id(self.node_id)
-        self.params["now"] = self.at.to_string()
-        self.params["branch"] = self.branch.name
-        self.params["branch_level"] = self.branch.hierarchy_level
-        self.params["status"] = RelationshipStatus.ACTIVE.value
-
-        self.add_to_query(query)
-
-
```
ajtmccarty, line 42: dead code

backend/infrahub/core/query/branch.py
```diff
@@ -52,31 +21,31 @@ def __init__(self, branch_name: str, **kwargs: Any):
...
+
+// reduce results to a single row
+WITH 1 AS one LIMIT 1
+
+// find any orphaned vertices and delete them
+MATCH (n)
+WHERE NOT exists((n)--())
+CALL (n) {
+  DELETE n
+} IN TRANSACTIONS
```
ajtmccarty, line 48: these `IN TRANSACTIONS` subqueries make sure the DB doesn't run out of memory when deleting a LOT of edges (say, 6 million)

backend/infrahub/core/branch/models.py
```diff
@@ -248,9 +267,13 @@ async def delete(self, db: InfrahubDatabase) -> None:
             raise ValidationError(f"Unable to delete {self.name} it is the default branch.")
         if self.is_global:
             raise ValidationError(f"Unable to delete {self.name} this is an internal branch.")
-        await super().delete(db=db)
+
+        self.status = BranchStatus.DELETING
+        await self.save(db=db)
```
ogenstad, line 272: Do we need to terminate a transaction here?

backend/infrahub/core/migrations/graph/m032_cleanup_orphaned_branch_relationships.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Delete edges for branches that were not completely deleted
+    """
+
+    name: str = "032_cleanup_deleted_branches"
+    minimum_version: int = 31
+
+    async def validate_migration(self, db: InfrahubDatabase) -> MigrationResult:  # noqa: ARG002
+        result = MigrationResult()
+        return result
```
ogenstad: How about just `return MigrationResult()`

backend/infrahub/core/migrations/graph/m032_cleanup_orphaned_branch_relationships.py
```diff
@@ -0,0 +1,82 @@
...
+            for branch_name in orphaned_branch_names:
+                log.info(f"Cleaning up branch '{branch_name}'...")
+                delete_query = await DeleteBranchRelationshipsQuery.init(db=db, branch_name=branch_name)
+                await delete_query.execute(db=db)
+                log.info(f"Branch '{branch_name}' cleaned up.")
+
+        except Exception as exc:
+            migration_result.errors.append(str(exc))
+            log.error(f"Error during branch cleanup: {exc}")
+            return migration_result
```
ogenstad: This return seems redundant

ogenstad (approved):

backend/infrahub/core/branch/enums.py
LucasG0: Alright, you can remove `NEED_REBASE` then

LucasG0 (approved):

backend/infrahub/core/branch/models.py
ajtmccarty: we do not
the `IN TRANSACTIONS` subqueries in the cypher query that runs next can be run inside of a transaction

backend/infrahub/core/migrations/graph/m032_cleanup_orphaned_branch_relationships.py
ajtmccarty: updated

backend/infrahub/core/migrations/graph/m032_cleanup_orphaned_branch_relationships.py
ajtmccarty: updated

PR #6840, opened 2025-07-16
Add is_draft proposed change attribute

LucasG0:


backend/infrahub/core/schema/definitions/core/propose_change.py
```diff
@@ -38,6 +38,7 @@
             default_value=ProposedChangeState.OPEN.value,
             optional=True,
         ),
+        Attr(name="is_draft", kind="Boolean", optional=True, default_value=False),
```
ajtmccarty: does it need to be optional if it has a default value?

LucasG0: No right

ajtmccarty (approved):

backend/infrahub/core/protocols.py
```diff
@@ -480,6 +480,7 @@ class CoreProposedChange(CoreTaskTarget):
     source_branch: String
     destination_branch: String
     state: Enum
+    is_draft: Boolean
```
ogenstad, line 483: I've been going back on forth with this. Wondering if we should have a boolean attribute, if `draft` should just be a new entry in the "state" enum?

The benefit of having an "is_draft" is that you can still query for all open proposed changes in one query instead of having to specify multiple states.

LucasG0: We rediscussed it and as a PC can be both draft AND open/merged/closed/... , having a separated boolean looks cleaner. It's also how it's done in Github

ogenstad (approved):

Issue #6845, opened 2025-07-17
bug: frontend error when searching "temp" in search anywhere


BeArchiTek:
### Component

Frontend UI

### Infrahub version

1.3.3

### Current Behavior

When searching temp in the search anywhere , I am getting :

<img width="1095" height="723" alt="Image" src="https://github.com/user-attachments/assets/a5b4635b-2983-4f10-9092-ca0a819edf90" />

### Expected Behavior

no crash :) 

### Steps to Reproduce

load a schema with at least one `genrate_template: true` for a node.

PR #6847, opened 2025-07-17
Fix search anywhere with template schema

pa-lem:
Issue: https://github.com/opsmill/infrahub/issues/6845


frontend/app/src/shared/components/search/search-nodes.tsx
```diff
@@ -128,9 +128,9 @@ const NodesOptions = ({ node }: NodesOptionsProps) => {
 
           {columns
             .filter(({ name }) => !["name", "label"].includes(name))
-            .map((column) => (
+            .map((column, index) => (
               <NodeAttribute
-                key={column.id}
+                key={`${column.label}_${index}`}
```
bilalabbad: how about `column.name`?

bilalabbad (approved):

PR #6857, opened 2025-07-18
feat: prepare for task manager scale out

fatih-acar:
- Few task worker fixes
- Add scale out to testcontainers:

   - It's disabled by default (to have the docker integration tests within this repo run without it)
   - It can be enabled with the `INFRAHUB_TESTING_TASKMGR_SCALEOUT` env var
   - For the bootstrap/init problem: Prefect recommends using an init container, as seen in the docs here: https://docs.prefect.io/v3/advanced/self-hosted#database-migrations https://docs.prefect.io/v3/advanced/self-hosted#3-servers-1-background-service-postgres-redis 
    But adding an init container to our workflows requires change almost everywhere: CI, Helm charts, docker compose, etc
   A workaround is to leverage our internal locking stack and our custom Prefect wrapper to have a "synchronized" init phase.

## Summary by CodeRabbit

* **New Features**
  * Added support for scalable task manager and background services in test environments, including new Docker Compose services and environment variables for distributed setups.
  * Introduced configurable pagination size for server and worker services, with a default value of 50.

* **Bug Fixes**
  * Improved initialization logic for Prefect services to better handle distributed and webserver-only modes.

* **Chores**
  * Enhanced test container setup with additional environment variables and configuration options for task manager scaling and background services.


backend/infrahub/prefect_server/bootstrap.py
```diff
@@ -0,0 +1,18 @@
...
+
+    await db.create_db()
+    session = await db.session()
+
+    async with session:
+        await run_block_auto_registration(session=session)
+
+
+if __name__ == "__main__":
+    asyncio.run(init_prefect())
```
ogenstad, line 18: I'm not sure we'll ever run the Python script in a way that would require this if statement for `__main__`?

python_testcontainers/infrahub_testcontainers/docker-compose-cluster.test.yml
```diff
@@ -166,13 +166,32 @@ services:
 
   task-manager:
     image: "${INFRAHUB_TESTING_DOCKER_IMAGE}:${INFRAHUB_TESTING_IMAGE_VERSION}"
-    command: uvicorn --host 0.0.0.0 --port 4200 --factory infrahub.prefect_server.app:create_infrahub_prefect
+    command: "${PREFECT_SERVER_COMMAND:-gunicorn -k uvicorn.workers.UvicornWorker -b 0.0.0.0:4200 'infrahub.prefect_server.app:create_infrahub_prefect()'}"
```
ogenstad, line 169: You've added a `()` prefix to create_infrahub_prefect` is that actually correct?

backend/infrahub/prefect_server/app.py
```diff
@@ -1,16 +1,47 @@
...
+        await init_prefect()
+
+
 def create_infrahub_prefect() -> FastAPI:
+    if (
+        os.getenv("PREFECT_API_BLOCKS_REGISTER_ON_START") == "false"
+        and os.getenv("PREFECT_API_DATABASE_MIGRATE_ON_START") == "false"
+    ):
+        # We are probably running distributed mode
+        from infrahub import config
```
ogenstad, line 40: As we need to do this inline import perhaps we should move it so that all of the inline imports are done in the same location.

Alternatively we could have something like:

```python
infrahub.prefect_server.prod:create_infrahub_prod_prefect
```

Then we could just import everything at the top. But I don't know if it causes other problems.

Either or is fine with me, just adding some random thoughts.
Then this function just imports the other one.

python_testcontainers/infrahub_testcontainers/docker-compose.test.yml
```diff
@@ -77,13 +77,32 @@ services:
 
   task-manager:
     image: "${INFRAHUB_TESTING_DOCKER_IMAGE}:${INFRAHUB_TESTING_IMAGE_VERSION}"
-    command: uvicorn --host 0.0.0.0 --port 4200 --factory infrahub.prefect_server.app:create_infrahub_prefect
+    command: "${PREFECT_SERVER_COMMAND:-gunicorn -k uvicorn.workers.UvicornWorker -b 0.0.0.0:4200 'infrahub.prefect_server.app:create_infrahub_prefect()'}"
```
ogenstad, line 80: Same comment as previous, could just be that it's a gunicorn thing with (), but I don't think we have it in other places..

ogenstad (approved):

backend/infrahub/prefect_server/bootstrap.py
fatih-acar: I used that bootstrap module as an init container at some point (before changing my mind due to all the constraints I described in the PR body). - running `python -m infrahub.prefect_server.bootstrap`

I thought it can still be handy in some cases, so I left this if statement...

python_testcontainers/infrahub_testcontainers/docker-compose-cluster.test.yml
fatih-acar: Yes, this is gunicorn specific syntax when using a "factory"

backend/infrahub/prefect_server/app.py
fatih-acar: I think this is mainly cosmetic... I could also have created a dedicated python module for the distributed mode (with locking included)... I think we can come back to this later if it causes real issues...

ajtmccarty (approved):

PR #6860, opened 2025-07-19
Change main license to Apache 2.0

dgarros:
This PR changes the license of Infrahub to Apache 2.0

Both the Apache 2.0 and AGPL licenses are recognized as open-source licenses. However, the Apache 2.0 license is more permissive as it does not obligate users to share their modified versions of the code with the community.
As a result of this change, we will no longer need each contributor to sign a Contributor License Agreement, I will deactivate the bot once the PR has been merged

We decided to make this change because we learned that the AGPL license could pose challenges or even prevent some companies from deploying and utilizing the software. A month ago, I initiated a poll on LinkedIn to gather additional insights.

https://www.linkedin.com/posts/damiengarros_autocon3-activity-7337795650471796736-8uBl?utm_source=share&utm_medium=member_desktop&rcm=ACoAAACsliABzsE5LEGYGjCNNNC0SwMsl9H4AUQ

The results were telling: nearly 50% of respondents expressed concerns about the AGPL license, and 30% mentioned they often avoid it altogether.


fatih-acar (approved):
Let it be more permissive :+1:

PR #6866, opened 2025-07-20
Refactor NumberPoolGetUsed query to account for deleted node and branch

### LINKED ISSUE #6865: bug: NumberPool ids previously used in a branch are not available after the branch has been deleted
### Component

API Server / GraphQL

### Infrahub version

1.3.3

### Current Behavior

When there are multiple nodes in the schema, with a number attribute of the same name, for example :
- Node A - `internal_id`
- Node B - `internal_id`

If a Resource Pool or a NumberPool is used on these attributes, it's possible that numbers from a pool allocated in a branch will not be available again after the branch has been deleted

Similar situation can also happen with a node that has been deleted

### Expected Behavior

Numbers from a pool allocated in a branch should be available when the branch has been deleted
Same with Numbers from a pool previously allocated to a node that has been deleted.

### Steps to Reproduce

- Load a schema with 2 nodes (NodeA and NodeB), each with an attribute named `number` of kind `NumberPool`
- Create 3 instances of each in the main branch
- Create a new branch
- Create 3 more instances of NodeA in the branch and 3 instances of NodeB in main
- Delete the branch
- Create 1 instance of NodeA in main

The last node created should have the value 7 allocated to its attribute number instead of 4

### Additional Information

_No response_

dgarros:
Fixes #6865 

This PR mainly refactor the query `NumberPoolGetUsed` and adds more tests around the query and around the NumberPool in general.

The main issues were that the query was matching only on the attribute name and not on the node id, also the query was not accounting properly for nodes that have been deleted. 

I've also added a query and some tests to ensure that all reservations will be updated when a node is being converted (and has a new id)

Side note, there is more work needed around the relationship `IS_RESERVED`, currently this relationship is not being deleted when a node or a branch is being deleted. It's working for now because we are using the UUID of the node as the identifier but this is currently preventing us to use Number Pool with Upsert.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved number pool allocation so that values no longer in use are now correctly returned to the pool for reuse.

* **New Features**
  * Enhanced support for updating and tracking reservations in number pools, including the ability to change reservation identifiers.
  * Added support for filtering and retrieving richer reservation data in number pools.

* **Tests**
  * Added comprehensive tests to verify number pool behavior across branches, object type conversions, and reservation updates, ensuring correct allocation, isolation, and reuse of numbers.

### DIFF
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, AsyncGenerator
+from typing import TYPE_CHECKING, Any, Generator
 
 from pydantic import BaseModel, ConfigDict
 
@@ -11,6 +11,13 @@ from infrahub.core.query import Query, QueryType
 if TYPE_CHECKING:
     from infrahub.core.protocols import CoreNumberPool
     from infrahub.database import InfrahubDatabase
+
+
+class NumberPoolIdentifierData(BaseModel):
+    model_config = ConfigDict(frozen=True)
+
+    value: int
+    identifier: str
 
 
 class IPAddressPoolGetIdentifiers(Query):
@@ -160,7 +167,7 @@ class NumberPoolGetReserved(Query):
     def __init__(
         self,
         pool_id: str,
-        identifier: str,
+        identifier: str | None = None,
         **kwargs: dict[str, Any],
     ) -> None:
         self.pool_id = pool_id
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/query/resource_manager.py
```diff
@@ -193,9 +195,26 @@ def get_reservation(self) -> int | None:
...
+It works for now because the query has been updated to match the identifier in IS_RESERVED with the UUID of the related node
+But in the future, if we need to use an identifier that is not the UUID, we will need to clean up the relationships
+This will be especially important as we want to support upsert with NumberPool
+"""
+
+
 class NumberPoolGetUsed(Query):
     name = "number_pool_get_used"
     type = QueryType.READ
+    return_model = NumberPoolGetUsedData
```
dgarros: This is an experiment to see if we could associate a model for each query to avoid needing too much information about the structure of the query within the code.

Here I'm using a Pydantic Model to define the expected format of the response, the data is being returned by `iter_results` with a generator
```python
    async def iter_results(self) -> AsyncGenerator[NumberPoolGetUsedData]:
        for result in self.results:
            yield self.return_model.model_construct(
                value=result.get_as_type("value", return_type=int),
                identifier=result.get_as_type("identifier", return_type=str),
            )
```

backend/infrahub/core/query/resource_manager.py
```diff
@@ -176,26 +185,104 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
+        self.return_labels = ["pool.uuid AS pool_id", "r", "new_rel"]
+
+
+"""
+Important!: The relationship IS_RESERVED for Number is not being cleaned up when the node or the branch is deleted
+I think this is something we should address in the future.
+It works for now because the query has been updated to match the identifier in IS_RESERVED with the UUID of the related node
+But in the future, if we need to use an identifier that is not the UUID, we will need to clean up the relationships
+This will be especially important as we want to support upsert with NumberPool
+"""
```
gmazoyer, line 279: Do we have an issue opened to log this?

gmazoyer (approved):

backend/infrahub/core/query/resource_manager.py
```diff
@@ -218,31 +305,38 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
         self.params["attribute_name"] = self.pool.node_attribute.value
 
-        # NOTE, there is a bug with IS_RESERVED that is affecting this query
-        # Currently the workaround is to use DISTINCT to avoid duplicates but overtime the query could become slower than expected
-        # This should be fixed in the future
         query = """
-        MATCH (pool:%(number_pool)s { uuid: $pool_id })
-        CALL (pool) {
-            MATCH (pool)-[res:IS_RESERVED]->(av:AttributeValue)<-[hv:HAS_VALUE]-(attr:Attribute)
+        MATCH (pool:%(number_pool)s { uuid: $pool_id })-[res:IS_RESERVED]->(av:AttributeValue)
```
fatih-acar, line 309: This rewrite should solve a scaling issue I saw at some point.
Will run tests against this when I can get back to work on this task...

backend/infrahub/core/query/resource_manager.py
dgarros: Not yet I will open one

PR #6868, opened 2025-07-21
IFC-1637 Add permission to review proposed changes

gmazoyer:
## Summary by CodeRabbit

* **New Features**
  * Introduced a new permission allowing users to review proposed changes.
  * Added a reviewer role with the ability to approve or reject proposed changes.

* **Improvements**
  * Streamlined and unified the creation of account roles and groups during initialization.
  * Enhanced permission checks for reviewing proposed changes to ensure only authorized users can perform this action.

* **Bug Fixes**
  * Added tests to verify that users without the required permission cannot review proposed changes.

### DIFF
--- a/backend/infrahub/core/initialization.py
+++ b/backend/infrahub/core/initialization.py
@@ -427,19 +427,19 @@ async def create_default_role(db: InfrahubDatabase) -> CoreAccountRole:
     return role
 
 
-async def create_proposed_change_approver_role(db: InfrahubDatabase) -> CoreAccountRole:
-    approver_permission = await Node.init(db=db, schema=InfrahubKind.GLOBALPERMISSION)
-    await approver_permission.new(
+async def create_proposed_change_reviewer_role(db: InfrahubDatabase) -> CoreAccountRole:
+    reviewer_permission = await Node.init(db=db, schema=InfrahubKind.GLOBALPERMISSION)
+    await reviewer_permission.new(
         db=db,
         action=GlobalPermissions.REVIEW_PROPOSED_CHANGE.value,
         decision=PermissionDecision.ALLOW_ALL.value,
         description="Allow a user to approve or revoke proposed changes",
     )
-    await approver_permission.save(db=db)
+    await reviewer_permission.save(db=db)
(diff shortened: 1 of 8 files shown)


changelog/+ifc1637.added.md
```diff
@@ -0,0 +1 @@
+Add a permission to allow users to review proposed changes (identifier `global:review_proposed_change:allow_all`). Users with existing Infrahub instances may need to create this permission to use it.
```
ajtmccarty, line 1: is it possible to add a migration to add this permission?

gmazoyer: I'm unsure if this is the job of a migration or the one of the `upgrade` command (which setup the permissions if none of them exist).

backend/infrahub/core/initialization.py
```diff
@@ -423,16 +424,29 @@ async def create_default_roles(db: InfrahubDatabase) -> Node:
...
 
-    group_name = "Infrahub Users"
-    group = await Node.init(db=db, schema=InfrahubKind.ACCOUNTGROUP)
-    await group.new(db=db, name=group_name, roles=[role])
-    await group.save(db=db)
-    log.info(f"Created account group: {group_name}")
+    return role
+
+
+async def create_proposed_change_reviewer_role(db: InfrahubDatabase) -> CoreAccountRole:
```
ajtmccarty, line 430: do we need a followup issue to make sure that this is executed on an existing system b/c `first_time_initialization` won't run?

gmazoyer: We have an internal issue about this, attached to the permission system in general.

changelog/+ifc1637.added.md
dgarros: The function `upgrade_permissions` should take care of it automatically when executing the command `infrahub upgrade`, would be good to validate that

gmazoyer: It does setup the permissions and roles but only if none of them exist.

ogenstad (approved):

PR #6874, opened 2025-07-22
IFC-1690 Fix available prefixes when parent prefix is empty

gmazoyer:
This change makes sure that we return subnets of the container prefix
for which we asked the available space. When a container prefix has no
space reserved, the next 2 smaller subnets that fit inside the container
must be returned. For instance, if 192.0.2.0/24 has no subnets, the
available space query must report 192.0.2.0/25 and 192.0.2.128/25.

### DIFF
--- a/backend/infrahub/graphql/resolvers/ipam.py
+++ b/backend/infrahub/graphql/resolvers/ipam.py
@@ -12,6 +12,7 @@ from infrahub.core.constants import InfrahubKind
 from infrahub.core.ipam.constants import PrefixMemberType
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
+from infrahub.core.protocols import BuiltinIPNamespace, BuiltinIPPrefix
 from infrahub.graphql.parser import extract_selection
 from infrahub.graphql.permissions import get_permissions
 
@@ -47,8 +48,8 @@ async def _build_ip_range_node(
     schema: NodeSchema,
     address: IPvAnyAddress,
     last_address: IPvAnyAddress,
-    ip_namespace: Node,
-    ip_prefix: Node,
+    ip_namespace: BuiltinIPNamespace,
+    ip_prefix: BuiltinIPPrefix,
 ) -> Node:
     n = await Node.init(schema=schema, db=db, branch=branch)
     await n.new(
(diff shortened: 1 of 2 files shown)


backend/infrahub/graphql/resolvers/ipam.py
```diff
@@ -205,15 +206,31 @@ async def _resolve_available_prefix_nodes(
         available_nodes.append(node)
 
     # Properly sort existing nodes with available prefixes
-    return sorted(existing_nodes + available_nodes, key=lambda n: n.prefix.obj)
+    with_available_prefixes = sorted(existing_nodes + available_nodes, key=lambda n: n.prefix.obj)
```
ajtmccarty, line 209: I think this would be easier to read if you returned `with_available_prefixes` here
something like 
```
if len(with_available_prefixes) > 1 or next(iter(with_available_prefixes)).prefix.obj != prefix.prefix.obj:
    return with_available_prefixes

available_nodes.clear()
...
```

backend/infrahub/graphql/resolvers/ipam.py
```diff
@@ -205,15 +206,31 @@ async def _resolve_available_prefix_nodes(
...
+
+    # If the only available prefix is the same as the container prefix, this means the container prefix is empty and we should therefore at least
+    # offer two smaller prefixes allocatable within it
+    if len(with_available_prefixes) == 1 and next(iter(with_available_prefixes)).prefix.obj == prefix.prefix.obj:
```
ajtmccarty: can this be `with_available_prefixes[0]` instead of `next(iter(...))`?

ajtmccarty (approved):

backend/infrahub/graphql/resolvers/ipam.py
gmazoyer: Ruff has [a rule](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) that prefers the `next(iter(...))` over the single element fetch, hence why I use that construct. But can move it back to `with_available_prefixes[0]` if prefered.

ajtmccarty: it looks like that rule is more for when you create a list and only care about the first element, but do whichever you prefer

PR #6875, opened 2025-07-22
IFC-1633 Add events related to proposed change actions

gmazoyer:
This change introduce events related to actions taken on a proposed change. For now, several new events are part of the PR:

- `infrahub.proposed_change.merged`
- `infrahub.proposed_change.approved`
- `infrahub.proposed_change.rejected`
- `infrahub.proposed_change.approval_revoked`
- `infrahub.proposed_change.rejection_revoked`

## Summary by CodeRabbit

* **New Features**
  * Introduced new event types for proposed change reviews and merges, including approval, rejection, revocation, review requests, and merging actions.
  * Added corresponding GraphQL event types and extended event mapping for these new events.
  * Enabled asynchronous event dispatching after proposed change review decisions and merges.

* **Documentation**
  * Expanded event documentation to include detailed descriptions and payloads for all new proposed change events.

* **Bug Fixes**
  * Enhanced test coverage to verify that proposed change review and merge actions correctly emit the new events.

* **Chores**
  * Updated changelog to reflect the addition of proposed change review and merge events.

### DIFF
--- a/backend/infrahub/events/proposed_change_action.py
+++ b/backend/infrahub/events/proposed_change_action.py
@@ -17,6 +17,7 @@ class ProposedChangeEvent(InfrahubEvent):
             "infrahub.proposed_change.id": self.proposed_change_id,
             "infrahub.proposed_change.name": self.proposed_change_name,
             "infrahub.proposed_change.state": self.proposed_change_state,
+            "infrahub.branch.name": self.meta.context.branch.name,
         }
 
 
@@ -31,7 +32,6 @@ class ProposedChangeReviewEvent(ProposedChangeEvent):
             "infrahub.proposed_change.reviewer_account_id": self.reviewer_account_id,
             "infrahub.proposed_change.reviewer_account_name": self.reviewer_account_name,
             "infrahub.proposed_change.reviewer_decision": self.reviewer_decision,
-            "infrahub.branch.name": self.meta.context.branch.name,
         }
 
 
@@ -46,7 +46,6 @@ class ProposedChangeReviewRevokedEvent(ProposedChangeEvent):
             "infrahub.proposed_change.reviewer_account_id": self.reviewer_account_id,
             "infrahub.proposed_change.reviewer_account_name": self.reviewer_account_name,
             "infrahub.proposed_change.reviewer_former_decision": self.reviewer_former_decision,
-            "infrahub.branch.name": self.meta.context.branch.name,
         }
 
 
@@ -63,7 +62,6 @@ class ProposedChangeMergedEvent(ProposedChangeEvent):
             **super().get_resource(),
             "infrahub.proposed_change.merged_by_account_id": self.merged_by_account_id,
             "infrahub.proposed_change.merged_by_account_name": self.merged_by_account_name,
-            "infrahub.branch.name": self.meta.context.branch.name,
         }
 
 
@@ -84,7 +82,6 @@ class ProposedChangeReviewRequestedEvent(ProposedChangeEvent):
             **super().get_resource(),
             "infrahub.proposed_change.review_requested_by_account_id": self.requested_by_account_id,
             "infrahub.proposed_change.review_requested_by_account_name": self.requested_by_account_name,
-            "infrahub.branch.name": self.meta.context.branch.name,
         }
 
 
--- a/backend/infrahub/events/proposed_change_action.py
+++ b/backend/infrahub/events/proposed_change_action.py
@@ -20,6 +20,23 @@ class ProposedChangeEvent(InfrahubEvent):
         }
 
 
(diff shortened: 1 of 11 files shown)


ogenstad (commented):
LGTM overall, though it would be nice with some tests where we send and query for the events afterwards.

backend/infrahub/events/proposed_change_action.py
```diff
@@ -0,0 +1,112 @@
...
+    reviewer_account_name: str = Field(..., description="The name of the user who reviewed the proposed change")
+    reviewer_decision: str = Field(..., description="The decision made by the reviewer")
+
+    def get_resource(self) -> dict[str, str]:
+        return {
+            **super().get_resource(),
+            "infrahub.proposed_change.reviewer_account_id": self.reviewer_account_id,
+            "infrahub.proposed_change.reviewer_account_name": self.reviewer_account_name,
+            "infrahub.proposed_change.reviewer_decision": self.reviewer_decision,
+            "infrahub.branch.name": self.meta.context.branch.name,
```
ajtmccarty: looks like this can be part of `ProposedChangeEvent` instead of included in the `get_resorce` method of the sub-classes

backend/infrahub/graphql/types/event.py
```diff
@@ -163,5 +186,9 @@ class Meta:
     events.BranchDeletedEvent.event_name: BranchDeletedEvent,
     events.GroupMemberAddedEvent.event_name: GroupEvent,
     events.GroupMemberRemovedEvent.event_name: GroupEvent,
+    events.ProposedChangeApprovedEvent.event_name: ProposedChangeReviewEvent,
+    events.ProposedChangeApprovalRevokedEvent.event_name: ProposedChangeReviewRevokedEvent,
+    events.ProposedChangeRejectedEvent.event_name: ProposedChangeReviewEvent,
+    events.ProposedChangeRejectionRevokedEvent.event_name: ProposedChangeReviewRevokedEvent,
```
ajtmccarty, line 214: are the PC merged and PC review requested events not required here?

backend/infrahub/task_manager/event.py
```diff
@@ -192,6 +204,16 @@ def _return_event_specifics(self) -> dict[str, Any]:
...
+            case "infrahub.proposed_change.approved" | "infrahub.proposed_change.rejected":
+                event_specifics = {
+                    **self._return_proposed_change_reviewer(),
+                    **self._return_proposed_change_reviewer_decision(),
+                }
+            case "infrahub.proposed_change.approval_revoked" | "infrahub.proposed_change.rejection_revoked":
+                event_specifics = {
+                    **self._return_proposed_change_reviewer(),
+                    **self._return_proposed_change_reviewer_former_decision(),
+                }
```
ajtmccarty, line 244: are the other event types required here?

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -284,30 +301,74 @@ async def _handle_decision(
...
+                    reviewer_former_decision=ProposedChangeApprovalDecision.REJECT.value,
+                    meta=event_meta,
+                )
+
             case _:
                 raise ValidationError(input_value=f"Invalid decision {decision}")
 
+        if event:
+            event_service = await get_event_service()
+            await event_service.send(event=event)
```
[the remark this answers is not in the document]
ogenstad: This looks like a valid comment, we should only send the event after the new state has been saved to the database.

backend/infrahub/events/proposed_change_action.py
```diff
@@ -0,0 +1,112 @@
...
+    requested_by_account_name: str = Field(
+        ..., description="The name of the user who requested the proposed change to be reviewed"
+    )
+
+    def get_resource(self) -> dict[str, str]:
+        return {
+            **super().get_resource(),
+            "infrahub.proposed_change.review_requested_by_account_id": self.requested_by_account_id,
+            "infrahub.proposed_change.review_requested_by_account_name": self.requested_by_account_name,
+        }
```
ogenstad: Looking at this again I realize that we typically use get_related() for the event specific data points. Would perhaps be nice in terms of consistency. For some events like the group or node mutation ones it's required as we can have multiple entries with the same name. In this case I'm not sure it's really needed. What do you think?

gmazoyer: I haven't thought about this, but this is a good idea. If we leverage `get_related()`, would it also simplify the GraphQL exposed objects?

ogenstad: I would say that it's fairly similar, an example is the artifacts with the method `_return_artifact_event()`

ogenstad (approved):

PR #6883, opened 2025-07-23
fix(backend): create node with many rels faster

fatih-acar:
Avoid fetching each peer during node creation by fetching them all at once prior to calling get_create_data().

## Summary by CodeRabbit

* **Refactor**
  * Simplified relationship management by converting several methods from asynchronous to synchronous, improving code clarity and reducing unnecessary async usage.
  * Updated related logic in queries and mutations to align with synchronous method calls.
* **Tests**
  * Updated test fixtures and test cases to use synchronous relationship methods.
  * Refactored schema setup in migration test fixtures for improved consistency and maintainability.
  * Adjusted assertions in node relationship tests for more precise identity checks.

### DIFF
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -157,10 +157,20 @@ class NodeCreateAllQuery(NodeQuery):
         relationships: list[RelationshipCreateData] = []
         for rel_name in self.node._relationships:
             rel_manager: RelationshipManager = getattr(self.node, rel_name)
+            # Fetch all relationship peers through a single database call for performances.
             peers = await rel_manager.get_peers(db=db, branch_agnostic=self.branch_agnostic)
+
             for rel in rel_manager._relationships:
-                if rel.get_peer_id() in peers:
-                    await rel.set_peer(value=peers[rel.get_peer_id()])
+                try:
+                    rel.set_peer(value=peers[rel.get_peer_id()])
+                except KeyError:
+                    pass
+                except ValueError:
+                    # Relationship has not been initialized yet, it means the peer does not exist in db yet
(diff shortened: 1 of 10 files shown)


backend/infrahub/core/query/node.py
```diff
@@ -157,7 +157,10 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
         relationships: list[RelationshipCreateData] = []
         for rel_name in self.node._relationships:
             rel_manager: RelationshipManager = getattr(self.node, rel_name)
+            peers = await rel_manager.get_peers(db=db, branch_agnostic=self.branch_agnostic)
             for rel in rel_manager._relationships:
+                if rel.get_peer_id() in peers:
+                    await rel.set_peer(value=peers[rel.get_peer_id()])
```
LucasG0:
```suggestion
for rel in rel_manager.get_relationships(db=db, branch_agnostic=self.branch_agnostic):
```

This should fetch all relationships at once, and it avoids accessing private members `_relationships`

fatih-acar: I already tried that and the thing is that `get_relationships()` only sets the peer ids (and does not call a get_peers equivalent to fetch the full peer node).

I wonder if we really need the peer attributes for the NodeCreateAll query to work, maybe we can just avoid the `peer.resolve()` call within the `rel.get_create_data` function.

LucasG0: >I already tried that and the thing is that get_relationships() only sets the peer ids (and does not call a get_peers equivalent to fetch the full peer node).

Right 

It seems we need to fetch to get the branch peer

ajtmccarty (approved):

PR #6888, opened 2025-07-24
Lock names include attributes values

LucasG0:
IFC-1874

This PR makes the current locks more granular, ie instead of locking only on object kind, we also lock on values of the attributes contained within uniqueness constraint. In this way, we allow concurrent creation/updates of objects having different values on these key attributes. We cannot do it for relationships at it would require fetching relationships peers before locking.

EDIT 2025-09-30: this PR also contains fixes for the resource managers to prevent duplicate allocations. Locking is now done during the mutation but also during the allocation of the resource, thus making the locks "re-entrant" was required to avoid deadlocks within a same request.

## Summary by CodeRabbit

* **New Features**
  * Option to skip resource-pool allocation during object create/update.
  * Unified constraint-and-save operation that runs under managed multi-locks.

* **Refactor**
  * Preview-before-persist flow to derive validation/save fields and locks before saving.
  * Centralized lock-name computation and multi-lock orchestration across mutations.
  * Exposed process_pools flag to control pool handling.

* **Bug Fixes**
  * Branch-aware locking and improved reentrant lock behavior.

* **Tests**
  * Added tests for lock-name generation and reentrant lock behavior.

### DIFF
--- a/backend/infrahub/graphql/mutations/ipam.py
+++ b/backend/infrahub/graphql/mutations/ipam.py
@@ -223,7 +223,7 @@ class InfrahubIPAddressMutation(InfrahubMutationMixin, Mutation):
 
         namespace_lock_names = cls._get_lock_names(namespace_id)
         async with InfrahubMultiLock(lock_registry=lock.registry, locks=namespace_lock_names):
-            async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names):
+            async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names, metrics=False):
                 async with db.start_transaction() as dbt:
                     reconciled_address = await cls._mutate_update_object_and_reconcile(
                         info=info,
@@ -400,7 +400,7 @@ class InfrahubIPPrefixMutation(InfrahubMutationMixin, Mutation):
 
         namespace_lock_names = cls._get_lock_names(namespace_id)
         async with InfrahubMultiLock(lock_registry=lock.registry, locks=namespace_lock_names):
-            async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names):
+            async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names, metrics=False):
                 async with db.start_transaction() as dbt:
                     reconciled_prefix = await cls._mutate_update_object_and_reconcile(
                         info=info,
--- a/backend/infrahub/graphql/mutations/ipam.py
+++ b/backend/infrahub/graphql/mutations/ipam.py
@@ -1,5 +1,4 @@
 import ipaddress
-from ipaddress import IPv4Interface
 from typing import TYPE_CHECKING, Any
 
 from graphene import InputObjectType, Mutation
@@ -13,12 +12,14 @@ from infrahub.core.constants import InfrahubKind
 from infrahub.core.ipam.reconciler import IpamReconciler
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
+from infrahub.core.node.create import get_profile_ids
 from infrahub.core.schema import NodeSchema
 from infrahub.database import InfrahubDatabase, retry_db_transaction
 from infrahub.exceptions import NodeNotFoundError, ValidationError
 from infrahub.lock import InfrahubMultiLock, build_object_lock_name
 from infrahub.log import get_logger
 
+from ...lock_getter import get_lock_names_on_object_mutation
 from .main import DeleteResult, InfrahubMutationMixin, InfrahubMutationOptions
 from .node_getter.by_default_filter import MutationNodeGetterByDefaultFilter
 
@@ -118,7 +119,7 @@ class InfrahubIPAddressMutation(InfrahubMutationMixin, Mutation):
         data: InputObjectType,
         branch: Branch,
         db: InfrahubDatabase,
-        ip_address: IPv4Interface | ipaddress.IPv6Interface,
+        ip_address: ipaddress.IPv4Interface | ipaddress.IPv6Interface,
         namespace_id: str,
     ) -> Node:
         address = await cls.mutate_create_object(data=data, db=db, branch=branch)
@@ -166,8 +167,24 @@ class InfrahubIPAddressMutation(InfrahubMutationMixin, Mutation):
         db: InfrahubDatabase,
         address: Node,
         namespace_id: str,
+        fields_to_validate: list[str],
+        fields: list[str],
+        previous_profile_ids: set[str],
+        lock_names: list[str],
     ) -> Node:
-        address = await cls.mutate_update_object(db=db, info=info, data=data, branch=branch, obj=address)
+        address = await cls.mutate_update_object(
+            db=db,
+            info=info,
+            data=data,
+            branch=branch,
+            obj=address,
+            fields_to_validate=fields_to_validate,
+            fields=fields,
+            previous_profile_ids=previous_profile_ids,
+            lock_names=lock_names,
+            manage_lock=False,
+            apply_data=False,
+        )
         reconciler = IpamReconciler(db=db, branch=branch)
         ip_address = ipaddress.ip_interface(address.address.value)
         if lock_name := cls._get_lock_name(namespace_id, branch):
@@ -205,11 +222,33 @@ class InfrahubIPAddressMutation(InfrahubMutationMixin, Mutation):
         namespace = await address.ip_namespace.get_peer(db)
         namespace_id = await validate_namespace(db=db, branch=branch, data=data, existing_namespace_id=namespace.id)
 
-        async with db.start_transaction() as dbt:
-            reconciled_address = await cls._mutate_update_object_and_reconcile(
-                info=info, data=data, branch=branch, address=address, namespace_id=namespace_id, db=dbt
-            )
-            result = await cls.mutate_update_to_graphql(db=dbt, info=info, obj=reconciled_address)
+        before_mutate_profile_ids = await get_profile_ids(db=db, obj=address)
+        await address.from_graphql(db=db, data=data)
+        fields_to_validate = list(data)
+        fields = list(data.keys())
+
+        for field_to_remove in ("id", "hfid"):
+            if field_to_remove in fields:
+                fields.remove(field_to_remove)
+
+        schema_branch = db.schema.get_schema_branch(name=branch.name)
+        lock_names = get_lock_names_on_object_mutation(node=address, branch=branch, schema_branch=schema_branch)
+
+        async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names):
+            async with db.start_transaction() as dbt:
+                reconciled_address = await cls._mutate_update_object_and_reconcile(
+                    info=info,
+                    data=data,
+                    branch=branch,
+                    db=dbt,
+                    address=address,
+                    namespace_id=namespace_id,
+                    fields_to_validate=fields_to_validate,
+                    fields=fields,
+                    previous_profile_ids=before_mutate_profile_ids,
+                    lock_names=lock_names,
+                )
+                result = await cls.mutate_update_to_graphql(db=dbt, info=info, obj=reconciled_address)
 
         return address, result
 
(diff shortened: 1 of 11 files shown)


backend/infrahub/lock_utils.py
ajtmccarty, line 1: usually `utils` means that a file contains a bunch of things that don't really belong together and I don't think that is the case here. I think this file can be `locks.py` or even `lock_getter.py`

backend/infrahub/lock_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    if isinstance(node_schema, GenericSchema):
+        return False
+
+    for generic_kind in node_schema.inherit_from:
+        if generic_kind in KINDS_CONCURRENT_MUTATIONS_NOT_ALLOWED:
+            return True
+    return False
+
+
+def _get_lock_names_on_object_mutation(node: Node, branch: Branch, schema_branch: SchemaBranch) -> list[str]:
```
ajtmccarty: at least one of these should be a public method and I think it's this one

ajtmccarty (commented):
could you please add a PR description that explains why the changes in `_get_kind_lock_names_on_object_mutation` will help with the locking slowness?

fatih-acar:
Quick remark: we should be very careful with user inputs, in this case you are lucky that Redis keys are very permissive regarding character set. I'm not sure it would be the case with all implementations.

We can instead of using the attribute value directly, use some kind of hex hash of it.

Edit: not saying we should do it right now, I don't even know performance implications of such a change.

backend/infrahub/lock_utils.py
LucasG0: Right

LucasG0:
@fatih-acar good point, I'll update to use the hash, I don't expect having performances issues on that, hashing a 1 million length string takes less than 1ms.

fatih-acar:
I think there's an issue with objects depending on a resource manager such as NumberPool... I end up with duplicate numbers allocated.

I think we still require to lock exclusively on kind when the object uses resource manager...

Edit: this issue is NOT related to https://github.com/opsmill/infrahub/pull/6866

fatih-acar:
> I think there's an issue with objects depending on a resource manager such as NumberPool... I end up with duplicate numbers allocated.
> 
> I think we still require to lock exclusively on kind when the object uses resource manager...
> 
> Edit: this issue is NOT related to #6866

https://github.com/opsmill/infrahub/pull/7025 is not enough to fix the issue because the `get_resource()` call is done within a DB transaction when we're inside the Create/Upsert mutation path...
One solution to fix this would be to apply the same filter we do on uniqueness constraints, but also on attributes that depend on a resource manager.

fatih-acar:
Fixes https://github.com/opsmill/infrahub/issues/7254

fatih-acar:
We can probably re-enable locking within branches now that they are more granular?

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+    # different processes.
+    return hashlib.sha256(value.encode()).hexdigest()
+
+
+def get_lock_names_on_object_mutation(node: Node, schema_branch: SchemaBranch) -> list[str]:
+    """
+    Return lock names for object on which we want to avoid concurrent mutation (create/update). Except for some specific kinds,
+    concurrent mutations are only allowed on non-main branch as objects validations will be performed at least when merging in main branch.
+    Lock names include kind, some generic kinds, and values of attributes of corresponding uniqueness constraints.
+    """
```
ajtmccarty, line 57: pretty sure this is leftover from months ago, but I don't think this is accurate any longer
also, I don't think this docstring belongs here as it references logic way outside of its scope

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+
+    lock_names: list[str] = []
+
+    def _add_pool_lock(pool_id: str) -> None:
+        lock_name = f"resource_pool.{pool_id}"
+        if lock_name not in lock_names:
+            lock_names.append(lock_name)
+
+    # Check if node is using resource manager allocation via attributes
+    for attr_name in getattr(node, "_attributes", []):
```
ajtmccarty: we probably do this in many other places, but I think the more correct way to get these names is using the schema linked to the `node` object
```python
for attr_name in node.get_schema().attribute_names:
```

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+    concurrent mutations are only allowed on non-main branch as objects validations will be performed at least when merging in main branch.
+    Lock names include kind, some generic kinds, and values of attributes of corresponding uniqueness constraints.
+    """
+
+    lock_names: list[str] = []
+
+    def _add_pool_lock(pool_id: str) -> None:
+        lock_name = f"resource_pool.{pool_id}"
+        if lock_name not in lock_names:
+            lock_names.append(lock_name)
```
ajtmccarty: if `lock_names` was a `set` you could probably remove this `_add_pool_lock` function-in-a-function

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+    # Check if node is using resource manager allocation via attributes
+    for attr_name in getattr(node, "_attributes", []):
+        attribute = getattr(node, attr_name, None)
+        if attribute is not None and getattr(attribute, "from_pool", None) and "id" in attribute.from_pool:
+            _add_pool_lock(attribute.from_pool["id"])
+
+    # Check if relationships allocate resources
+    for rel_name in node._relationships:
+        rel_manager: RelationshipManager = getattr(node, rel_name)
+        for rel in rel_manager._relationships:
```
ajtmccarty, line 70: maybe we need a `Node.get_relationship_managers` method and a `RelationshipManager.get_fetched_relationships` method to avoid accessing these private variables outside of the classes

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+            if rel.from_pool and "id" in rel.from_pool:
+                _add_pool_lock(rel.from_pool["id"])
+
+    lock_kinds = _get_kinds_to_lock_on_object_mutation(node.get_kind(), schema_branch)
+    for kind in lock_kinds:
+        schema = schema_branch.get(name=kind)
```
ajtmccarty: `duplicate=False`

backend/infrahub/lock_getter.py
```diff
@@ -0,0 +1,124 @@
...
+                # Check whether we should remove original schema kind as it simply overrides uniqueness_constraint
+                # of a generic
+                kinds.pop(0)
+                node_schema_kind_removed = True
+    return kinds
+
+
+def _hash(value: str) -> str:
+    # Do not use builtin `hash` for lock names as due to randomization results would differ between
+    # different processes.
```
ajtmccarty, line 49: I learned something new today

backend/infrahub/lock.py
ajtmccarty, line 1: is this all to support the "re-entrant" locks you mentioned in the description? meaning a lock that can be acquired multiple times?
how does this prevent different workers from acquiring the same lock if the lock can be acquired more than once?

backend/infrahub/graphql/mutations/ipam.py
ajtmccarty, line 1: it would be nice to reduce some of the duplicated code in here, but that is seems out of scope for this one

backend/infrahub/graphql/mutations/main.py
```diff
@@ -220,42 +210,52 @@ async def _call_mutate_update(
         """
         Wrapper around mutate_update to potentially activate locking and call it within a database transaction.
         """
+        before_mutate_profile_ids = await get_profile_ids(db=db, obj=obj)
+        fields_to_validate = list(data)
+        fields = list(data.keys())
```
ajtmccarty, line 215: are these different?

backend/infrahub/graphql/mutations/main.py
```diff
@@ -220,42 +210,52 @@ async def _call_mutate_update(
...
+                fields.remove(field_to_remove)
+
+        # Prepare a clone to compute locks without triggering pool allocations
+        preview_obj = await NodeManager.get_one_by_id_or_default_filter(
+            db=db,
+            kind=obj.get_kind(),
+            id=obj.get_id(),
+            branch=branch,
+            include_owner=True,
+            include_source=True,
```
ajtmccarty: are owner and source necessary for this?

ajtmccarty (commented):
some questions and comments (mostly on things you did not write)
and a couple more questions
1. is it possible to add any sort of graphql-level tests that actually cover the locks as they will be used? seems like it would be hard, but maybe it's possible by creating/updating overlapping objects and checking timestamps of the changes or something
2. is this change big enough to be on develop instead of stable?

backend/infrahub/lock_getter.py
fatih-acar: Indeed, let me know what you think of the new version

backend/infrahub/lock_getter.py
fatih-acar: Thanks, that's the type of thing I'm not aware of :)

backend/infrahub/lock_getter.py
fatih-acar: Smarter than AI :p

backend/infrahub/lock_getter.py
fatih-acar: Yes I believe we do this in many places to access relationships without fetching them. I agree on this point but we can probably do that in another PR?

backend/infrahub/lock_getter.py
fatih-acar: Good catch!

backend/infrahub/lock_getter.py
fatih-acar: Thanks coderabbit :)

backend/infrahub/graphql/mutations/ipam.py
fatih-acar: Agree

backend/infrahub/graphql/mutations/main.py
fatih-acar: I believe `fields_to_validate` is used for the constraints checker, this is what I understand from the previous state of the code. So yes they are different and required?

backend/infrahub/graphql/mutations/main.py
fatih-acar: Indeed good catch! AI not smart enough

backend/infrahub/lock.py
fatih-acar: Yes exactly
We track this behavior using a counter within a Contextvar: this is to have a asyncio context scoped counter (this is similar to TLS (thread local storage) but for asyncio coroutines, see https://peps.python.org/pep-0567/)

Different worker is necessarily within another coroutine, so in this case since the refcount is 0, it will try to acquire the global lock, which will be held by another worker.
A different coroutine within the same worker will also try to acquire the lock (in case of a concurrent graphql mutation running on the same worker for example).

fatih-acar: The main goal is just to fix the following call stack:
Create mutation holds resource pool lock -> create new node internally using Node.new() -> triggers resource pool get_resource() -> get_resource locks again using the same resource pool id -> deadlock

We introduced `get_resource locks again using the same resource pool id` in https://github.com/opsmill/infrahub/pull/7025 to lock direct mutation calls to the resource pools...
A cleaner solution would be to move the lock from the `get_resource` call up to the different top-level mutation handlers.

fatih-acar:
Thanks for reviewing @ajtmccarty 

> some questions and comments (mostly on things you did not write) and a couple more questions
> 
> 1. is it possible to add any sort of graphql-level tests that actually cover the locks as they will be used? seems like it would be hard, but maybe it's possible by creating/updating overlapping objects and checking timestamps of the changes or something
> 2. is this change big enough to be on develop instead of stable?

1. I think yes, but it would be a bit hard as you wonder. It's pretty easy to stress test this using our internal scale testing suite since it massively creates objects concurrently (this is how I caught most of the issues that are resolved in this PR).
2. I think this should go in develop for the 1.6 release, but we might need to cherry-pick some changes to fix https://github.com/opsmill/infrahub/issues/7254 in stable...

fatih-acar:
Replaced by https://github.com/opsmill/infrahub/pull/7508

PR #6914, opened 2025-07-30
command to clean up incorrect inheritance/database labels

ajtmccarty:
IFC-1691

new `infrahub db check-inheritance [--fix/--no-fix]` to check and clean up any incorrect inheritance at the database level
the root cause of the problem remains unidentified, but it is somehow possible to have the Node vertices with labels that do not match their current inheritance structure. for example, if I try to migrate the `NodeOne` schema to inherit from `CoreArtifactTarget` and that migration silently fails in some way then the schema on the database will indicate that all active `NodeOne` vertices should also have the `CoreArtifactTarget` label, but that will not actually be true

this command can be used to identify schemas with this problem and correct them

this version only works for inheritance misalignment. it does not identify or correct schema name/namespace/kind misalignment (mostly because that was going to be a harder problem to solve and I do not know if it is necessary)

TODO

- [x] unit tests (manual testing on a database dump with the problem is successful)
- [x] check if we can wait to save inheritance updates until the migration has completed successfully. made a new issue for it #6948 

## Summary by CodeRabbit

* **New Features**
  * Introduced a new CLI command, infrahub db check-inheritance, to detect and optionally fix schema inheritance inconsistencies in the database.
* **Documentation**
  * Added detailed documentation for the new check-inheritance command, including usage instructions and options.
* **Tests**
  * Added comprehensive tests to verify the detection and correction of inheritance issues across schema branches.

### DIFF
--- a/backend/infrahub/cli/db.py
+++ b/backend/infrahub/cli/db.py
@@ -56,6 +56,7 @@ from infrahub.services.adapters.message_bus.local import BusSimulator
 from infrahub.services.adapters.workflow.local import WorkflowLocalExecution
 
 from .constants import ERROR_BADGE, FAILED_BADGE, SUCCESS_BADGE
+from .db_commands.check_inheritance import check_inheritance
 from .patch import patch_app
 
 
@@ -174,6 +175,28 @@ async def migrate_cmd(
     dbdriver = await context.init_db(retry=1)
 
     await migrate_database(db=dbdriver, initialize=True, check=check, migration_number=migration_number)
+
+    await dbdriver.close()
+
+
+@app.command(name="check-inheritance")
+async def check_inheritance_cmd(
+    ctx: typer.Context,
+    fix: bool = typer.Option(False, help="Fix the inheritance of any invalid nodes."),
+    config_file: str = typer.Argument("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+) -> None:
+    """Check the database for any vertices with incorrect inheritance"""
+    logging.getLogger("infrahub").setLevel(logging.WARNING)
+    logging.getLogger("neo4j").setLevel(logging.ERROR)
+    logging.getLogger("prefect").setLevel(logging.ERROR)
+
+    config.load_and_exit(config_file_name=config_file)
+
+    context: CliContext = ctx.obj
+    dbdriver = await context.init_db(retry=1)
+    await initialize_registry(db=dbdriver)
+
+    await check_inheritance(db=dbdriver, fix=fix)
 
     await dbdriver.close()
 
--- a/backend/infrahub/cli/db.py
+++ b/backend/infrahub/cli/db.py
@@ -196,7 +196,9 @@ async def check_inheritance_cmd(
     dbdriver = await context.init_db(retry=1)
     await initialize_registry(db=dbdriver)
 
-    await check_inheritance(db=dbdriver, fix=fix)
+    success = await check_inheritance(db=dbdriver, fix=fix)
+    if not success:
+        raise typer.Exit(code=1)
 
     await dbdriver.close()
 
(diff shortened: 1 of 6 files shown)


backend/infrahub/cli/db.py
```diff
@@ -178,6 +179,30 @@ async def migrate_cmd(
     await dbdriver.close()
 
 
+@app.command(name="check-inheritance")
```
LucasG0, line 182: Should these checks be part of the `check` command as well? To avoid `check` not complaining while the db is broken

backend/infrahub/core/migrations/query/node_duplicate.py
```diff
@@ -38,11 +38,23 @@ def __init__(
...
+            ORDER BY n_label ASC
+            WITH collect(n_label) AS sorted_labels
+
+            RETURN (
+                node.kind = $new_node.kind AND
+                sorted_labels = $new_sorted_labels
+            ) AS already_migrated
+        }
+        WITH node WHERE already_migrated = FALSE
+        """ % {"labels_str": labels_str}
```
LucasG0, line 67: Is it possible to add a docstring to `class NodeDuplicateQuery(Query)` to indicate what's it's doing? It took me some time to re-understand that we are copying all nodes from a given kind to another.

Also a comment around above block would be helpful to indicate why we need this block, ie because this migration might have silently failed for unkown reasons resulting in nodes not having being migrated

docs/docs/reference/infrahub-cli/infrahub-db.mdx
```diff
@@ -17,6 +17,7 @@ $ infrahub db [OPTIONS] COMMAND [ARGS]...
 **Commands**:
 
 * `check`: Run database sanity checks and output the...
+* `check-inheritance`: Check the database for any vertices with...
```
LucasG0, line 20: Just wondering why we add a dedicated command versus a `patch` here?

backend/infrahub/cli/db.py
ajtmccarty: I could try to add the `check-inheritance` logic into the `check` command, although it would require some changes to produce a CSV file of the problem nodes
I did not do this initially b/c the `check` command looks for things that are illegal at the database level and that can be identified by just examining the structure of the data in the graph: Relationship nodes with the same UUID, Relationship nodes with only one peer, etc. This `check-inheritance` command is different b/c the database is not really in an illegal state, but the schema and data are out of sync

backend/infrahub/cli/db_commands/check_inheritance.py
```diff
@@ -0,0 +1,284 @@
...
+
+    if registry.default_branch in kind_label_counts_by_branch:
+        kinds = [kind_label_count.kind for kind_label_count in kind_label_counts_by_branch[registry.default_branch]]
+        rprint(
+            "[bold cyan]Note that migrations were run on the default branch for the following schema kinds: "
+            f"{', '.join(kinds)}. You should rebase any branches that include/will include changes using "
+            "the migrated schemas[/bold cyan]"
+        )
+
+    return True
```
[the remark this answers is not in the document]
ajtmccarty: the migration is idempotent, no transaction is required
and a transaction could cause the migration to fail with an out-of-memory error if there were too many nodes that required changes

backend/infrahub/core/migrations/query/node_duplicate.py
ajtmccarty: I'll add a comment or two, but the changes are really just to make the migration idempotent so that we don't try to migrate nodes unnecessarily. for example, if the same schema changes are made on `main` and a branch before that branch is merged into `main`, then we could skip basically all of this migration logic during the merge process

docs/docs/reference/infrahub-cli/infrahub-db.mdx
ajtmccarty: I think I explained it above `check-inheritance` is looking at both schema and data and `check` is only looking at the database structure

LucasG0 (approved):

PR #6915, opened 2025-07-30
Prevent generic relationship peer to be overridden

### LINKED ISSUE #6699: bug: Overriding the peer of a generic relationship on the node level can crash the API server
### Component

API Server / GraphQL

### Infrahub version

1.3.0

### Current Behavior

There's a lack of validation when overriding the peer of a generic relationship which makes it possible to crash the GraphQL api by loading such a schema.

### Expected Behavior

This type of schema should be rejected as we wouldn't be able to produce a valid GraphQL schema as the relationship on a generic needs to point to the same kind of peer as all nodes inheriting from this generic.

### Steps to Reproduce

1. Load this schema into a test environment.

```yaml
---
version: "1.0"

nodes:
  - name: TagGroup
    namespace: Test
    inherit_from:
      - CoreGroup
    relationships:
      - name: members
        peer: BuiltinTag
        optional: false
        cardinality: many
        identifier: "group_member"
```

2. Note that after this the GraphQL API no longer works.

In the server logs this error can be observed:

```python
  File "/Users/patrick/Code/opsmill/infrahub/backend/infrahub/graphql/app.py", line 287, in _handle_http_request
    GRAPHQL_QUERY_ERRORS_METRICS.labels(**labels).observe(len(errors))
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/infrahub/python_sdk/infrahub_sdk/analyzer.py", line 44, in is_valid
    errors = validate(schema=self.schema, document_ast=self.document)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/.virtualenvs/infrahub/lib/python3.12/site-packages/graphql/validation/validate.py", line 47, in validate
    assert_valid_schema(schema)
  File "/Users/patrick/.virtualenvs/infrahub/lib/python3.12/site-packages/graphql/type/validate.py", line 90, in assert_valid_schema
    raise TypeError("\n\n".join(error.message for error in errors))
TypeError: Interface field CoreGroup.members expects type NestedPaginatedCoreNode! but TestTagGroup.members is type NestedPaginatedBuiltinTag!.
```

### Additional Information

This should be fixed in the SchemaBranch class so that we raise an error if such a schema is loaded.

ogenstad:
Fixes #6699

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema validation to ensure that inherited relationships cannot override the peer attribute of a generic relationship. An error is now raised during schema loading if this rule is violated.

* **Tests**
  * Added a new test case to verify that overriding the peer of an inherited generic relationship correctly raises an error.

* **Documentation**
  * Updated the changelog to reflect the stricter validation for inherited relationship peers.

### DIFF
--- a/backend/infrahub/core/schema/node_schema.py
+++ b/backend/infrahub/core/schema/node_schema.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from infrahub.core.constants import AllowOverrideType, InfrahubKind
+from infrahub.core.constants import AllowOverrideType, InfrahubKind, RelationshipKind
 
 from .generated.node_schema import GeneratedNodeSchema
 from .generic_schema import GenericSchema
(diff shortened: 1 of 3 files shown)


gmazoyer (approved):

backend/infrahub/core/schema/node_schema.py
```diff
@@ -70,14 +70,16 @@ def validate_inheritance(self, interface: GenericSchema) -> None:
...
-                )
+            if relationship.name in interface.relationship_names and not relationship.inherited:
+                interface_relationship = interface.get_relationship(relationship.name)
+                if interface_relationship.allow_override == AllowOverrideType.NONE:
+                    raise ValueError(
+                        f"{self.kind}'s relationship {relationship.name} inherited from {interface.kind} cannot be overriden"
+                    )
+                if relationship.kind != RelationshipKind.HIERARCHY and relationship.peer != interface_relationship.peer:
+                    raise ValueError(
+                        f"{self.kind}'s relationship {relationship.name} inherited from {interface.kind} must have the same peer"
```
ajtmccarty: might be helpful to include the value of `peer` in this message

ajtmccarty (approved):

backend/infrahub/core/schema/node_schema.py
ogenstad: Sounds like a good idea, fixed!

PR #6919, opened 2025-07-30
Add ProposedChange.total_comments readonly property

LucasG0:
## Summary by CodeRabbit

* **New Features**
  * Added a new optional, read-only field displaying the total number of comments (including both direct and thread comments) for each proposed change.
* **Tests**
  * Introduced functional tests to verify accurate calculation and display of the total comments count for proposed changes.


backend/infrahub/core/node/proposed_change.py
```diff
@@ -0,0 +1,43 @@
...
+                total_comments = 0
+                proposed_change = cast(CoreProposedChangeProtocol, self)
+                change_comments = await proposed_change.comments.get_relationships(db=db)
+                total_comments += len(change_comments)
+
+                threads = await proposed_change.threads.get_peers(db=db)
+                for thread in threads.values():
+                    thread_ = cast(CoreChangeThread, thread)
+                    thread_comments = await thread_.comments.get_relationships(db=db)
+                    total_comments += len(thread_comments)
```
ajtmccarty: not sure if this is necessary, but I think you could avoid the nested `get_relationships` call if you used `NodeManager.query` for the thread comments with a filter for the IDs of the threads retrieved on line 35

ajtmccarty (approved):

backend/infrahub/core/protocols.py
```diff
@@ -482,6 +482,7 @@ class CoreProposedChange(CoreTaskTarget):
     destination_branch: String
     state: Enum
     is_draft: Boolean
+    total_comments: IntegerOptional
```
ogenstad, line 485: I'm wondering if this will be problematic in some way later. I'd prefer to have this as an Integer instead of optional. The read_only attribute could have a default value of 0. However I'm not certain if the schema upgrade will handle that. It could be that we should revisit this part later. Not a high priority it's just nicer to know that an empty set would always return a `0` which I think it would but according to the schema it would be valid with `0 | None`. Given the time if we need an extra migration for this we can perhaps create a housekeeping issue to change it later.

ogenstad (approved):

ogenstad (approved):
Let's go with this for now and revisit if we come up with a "runtime attribute" in the future that isn't stored in any way in the database.

backend/infrahub/core/node/proposed_change.py
LucasG0: yes nice

PR #6922, opened 2025-07-30
Remove prefect events ASC flaky test

LucasG0:
## Summary by CodeRabbit

* **Tests**
  * Updated event query test by removing filtering and count assertion steps.


ogenstad (approved):

ogenstad:
This doesn't remove the test only the check for the number of results which I assume is failing in CI due to pagination issues when we revert the sort order. However we only really need to test the query here as we're just using a feature in Prefect, so removing these two lines should be fine.

PR #6926, opened 2025-07-30
IFC-1539 Add query to get actions on a proposed change

gmazoyer:
This change introduces a new `CoreProposedChangeAvailableActions` which can be used to understand what are the actions that can be taken on a proposed change. While the way the code is built can certainly be improved, this PR is mostly here to propose and expose an API that we should agree on.

The list of possible actions to take and how they are evaluated may change. If it comes to that it should be fairly simple to adapt.

## Summary by CodeRabbit

* **New Features**
  * Introduced a GraphQL query to display available actions for proposed changes, including reasons for action unavailability.
  * Added support for evaluating user permissions and proposed change states to determine which actions are accessible.
  * Expanded the schema with a new entry point for querying proposed change actions.
  * Defined a comprehensive framework to validate action availability based on user roles, proposed change state, and permissions.

* **Tests**
  * Added unit tests to verify the correct availability of actions based on different proposed change states and user roles.


backend/infrahub/graphql/queries/proposed_change.py
```diff
@@ -0,0 +1,156 @@
...
+    ),
+    ActionRule(
+        action=ProposedChangeAction.MERGE,
+        checks=[
+            StateIs(expected=[ProposedChangeState.OPEN]),
+            DraftIs(expected=False),
+            HasPermission(permission=MERGE_PROPOSED_CHANGE_PERMISSION),
+        ],
+    ),
+]
```
ogenstad: I wonder if these should be placed in the infrahub.proposed_change module instead. We can leave them here for now just raising the topic.

backend/infrahub/graphql/queries/proposed_change.py
```diff
@@ -0,0 +1,156 @@
...
+)
+REVIEW_PROPOSED_CHANGE_PERMISSION = GlobalPermission(
+    action=GlobalPermissions.REVIEW_PROPOSED_CHANGE.value,
+    decision=PermissionDecision.ALLOW_ALL.value,
+)
+
+ACTION_RULES = [
+    ActionRule(
+        action=ProposedChangeAction.OPEN,
+        checks=[StateIs(expected=[ProposedChangeState.CLOSED, ProposedChangeState.CANCELED])],
```
ogenstad: We shouldn't be allowed to open a canceled proposed change as that typically only happens when a branch gets deleted. So only closed -> open would be valid.

ogenstad (approved):
LGTM, added a small issue about the state transition.

backend/infrahub/proposed_change/action_checker.py
```diff
@@ -0,0 +1,186 @@
...
+            StateIs(expected=[ProposedChangeState.OPEN]),
+            HasPermission(permission=REVIEW_PROPOSED_CHANGE_PERMISSION),
+        ],
+    ),
+    ActionRule(
+        action=ProposedChangeAction.MERGE,
+        checks=[
+            StateIs(expected=[ProposedChangeState.OPEN]),
+            DraftIs(expected=False),
+            HasPermission(permission=MERGE_PROPOSED_CHANGE_PERMISSION),
```
LucasG0, line 198: For `MERGE` we might want to check for `verify_proposed_change_is_mergeable`?

LucasG0 (approved):

PR #6927, opened 2025-07-31
fix for unlikely diff bug

### LINKED ISSUE #6928: bug: `Cannot identify peer ID for relationship property` error during diff
### Component

API Server / GraphQL

### Infrahub version

1.2.12

### Current Behavior

A `Cannot identify peer ID for relationship property` error is raised when trying to generate a diff

### Expected Behavior

This error should not be raised. It means that some information if missing from one of the cypher queries that we run to calculate a diff. Specifically, it means that the logic for combining the results of various cypher queries into a single diff object in memory is missing a peer for a relationship that has metadata included in the diff. So the cypher queries include information for `IS_VISIBLE` or `IS_PROTECTED` properties, but no peer Node to group them together

### Steps to Reproduce

to reproduce this issue you need to have multiple branches with different inheritance updates for the same schema and then make a new branch that includes a delete of a node for which one of the migrated-kind schema is a peer

1. migrate `SchemaA` (this can be any schema) to update its inheritance on the default branch
2. make a new branch `br1` and migrate `SchemaA` to reset its inheritance to what it was before step 1
3. make another new branch `br2` and delete a node that has a relationship to an instance of `SchemaA`
4. try to generate a diff for `br2` and you might see the `Cannot identify peer` error

### Additional Information

The database state for the Relationship vertex that raises this issue is uncommon, but not illegal. The Relationship vertex in question is linked to 4 Node vertices (the usual is 2), 3 of which are the result of updating the inheritance of a NodeSchema.

ajtmccarty:
fixes #6928 

if a deleted node included in the diff for a given branch had a peer on one of its relationships that had gone through the NodeKindUpdate migration (changing its kind or inheritance) multiple times on different branches then the diff _could_ fail to generate depending on how results were ordered

the fix is
- to be more deterministic when ordering the results of certain queries to ensure that active edges are preferred over deleted edges with the same from time
- to allow deeper edges in a path to have an earlier from time than shallower edges, which can happen if the peer of a relationship had its kind/inheritance updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where diffs could not be generated if a peer of a deleted node had changes in its kind or inheritance across branches. Diffs now correctly reflect these changes.
* **Tests**
  * Added a new test to validate diff calculation when a node kind migration is reverted on a branch and a related node is deleted on a further branch.

### DIFF
--- a/backend/infrahub/core/query/diff.py
+++ b/backend/infrahub/core/query/diff.py
@@ -148,7 +148,14 @@ CALL (diff_path) {
         OR type(base_r_node) <> "IS_RELATED" OR type(base_r_prop) <> "IS_RELATED"
     )
     WITH latest_base_path, base_r_root, base_r_node, base_r_prop
-    ORDER BY base_r_prop.from DESC, base_r_node.from DESC, base_r_root.from DESC
+    // status="active" ordering is for tie-breaking edges added and deleted at the same time, we want the active one
+    ORDER BY
+        base_r_prop.from DESC,
+        base_r_prop.status = "active" DESC,
+        base_r_node.from DESC,
+        base_r_node.status = "active" DESC,
+        base_r_root.from DESC,
+        base_r_root.status = "active" DESC
     LIMIT 1
     RETURN latest_base_path
 }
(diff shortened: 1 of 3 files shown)


backend/infrahub/core/query/diff.py
```diff
@@ -172,8 +179,6 @@ def __init__(
     AND %(id_func)s(peer_r_node) = %(id_func)s(r_node)
     AND [%(id_func)s(n), type(peer_r_node)] <> [%(id_func)s(peer), type(r_peer)]
     AND r_peer.from < $to_time
-    // filter out paths where an earlier from time follows a later from time
-    AND peer_r_node.from <= r_peer.from
```
ajtmccarty, line 176: these paths are actually legal in the case of a peer node with a migrated kind, so we cannot ignore them

ogenstad (approved):

PR #6945, opened 2025-08-01
IFC-1686 Add support for `kinds` filter for generic IP query

gmazoyer:
This change adds a `kinds` filter to the generic IPAM queries in order to only get nodes which kinds are matching a filter provided one.

When performing the database query, if a limit is given and kinds must be filtered, we query for a larger dataset to be able to fill in the return page as much as possible with the kinds that need to be in.

## Summary by CodeRabbit

* **New Features**
  * Added support for filtering IP address and prefix results by specific kinds in GraphQL queries, allowing more precise queries.
* **Bug Fixes**
  * Ensured that available IP prefix and range nodes are always included in filtered results.
* **Tests**
  * Introduced comprehensive tests to verify correct filtering and inclusion of nodes by kind and limit in query results.

### DIFF
--- a/backend/infrahub/graphql/resolvers/ipam.py
+++ b/backend/infrahub/graphql/resolvers/ipam.py
@@ -231,6 +231,23 @@ async def _resolve_available_prefix_nodes(
     return available_nodes
 
 
+def _filter_kinds(nodes: list[Node], kinds: list[str], limit: int | None) -> list[Node]:
+    filtered: list[Node] = []
+    available_node_kinds = [InfrahubKind.IPPREFIXAVAILABLE, InfrahubKind.IPRANGEAVAILABLE]
+    kinds += available_node_kinds
+
+    limit_with_available = limit
+    for node in nodes:
+        if node.get_schema().kind not in kinds:
+            continue
+        # Adapt the limit of nodes to return by always including available ones
+        if node.get_schema().kind in available_node_kinds:
+            limit_with_available += 1
+        filtered.append(node)
+
+    return filtered[:limit_with_available] if limit else filtered
+
+
 async def _annotate_result(
     db: InfrahubDatabase,
     branch: Branch,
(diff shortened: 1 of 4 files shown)


backend/infrahub/graphql/resolvers/ipam.py
```diff
@@ -231,38 +232,58 @@ async def _resolve_available_prefix_nodes(
...
+def _filter_kinds(nodes: list[Node], kinds: list[str], limit: int | None) -> list[Node]:
+    filtered: list[Node] = []
+    available_node_kinds = [InfrahubKind.IPPREFIXAVAILABLE, InfrahubKind.IPRANGEAVAILABLE]
+    kinds_with_available = kinds + available_node_kinds
+
+    limit_with_available = limit
+    for node in nodes:
+        if node.get_schema().kind not in kinds_with_available:
+            continue
+        # Adapt the limit of nodes to return by always including available ones
```
LucasG0, line 245: Why is it ok that we would return more nodes than expected in that case?

gmazoyer: This is expected. We decided, when working on a previous PR, that the limit should be honoured for existing nodes, and that the ones we create on the fly (in-memory for a query) should not count into the pagination limit.

backend/infrahub/graphql/resolvers/ipam.py
```diff
@@ -275,14 +296,15 @@ async def ipam_paginated_list_resolver(  # noqa: PLR0915
     partial_match: bool = False,
     **kwargs: dict[str, Any],
 ) -> dict[str, Any]:
-    schema: NodeSchema = (
+    schema: NodeSchema | GenericSchema = (
```
LucasG0, line 300: I'm not clear on which schema this could be here? Are kinds_to_filter supposed to be only kinds inheriting from this schema?

LucasG0 (approved):

PR #4582, opened 2024-10-08
fix: exclude infrahub.toml in dockerfile

wartraxx51:

PR #6956, opened 2025-08-03
Cleanup references to infrahub.toml

petercrocker:
Cleanup based on this PR: https://github.com/opsmill/infrahub/pull/4582

## Summary by CodeRabbit

* **Documentation**
  * Updated installation guide by removing references to a deprecated environment variable in Helm chart examples.
  * Simplified SSO configuration guide by displaying only environment variable examples and removing tabbed TOML configuration sections.


fatih-acar (approved):

ogenstad:
Personally I think it would make sense to remove the support for the infrahub.toml files alltogether and only use environment variables. But I know that at least some customers wants to use these files. If we keep them I think it makes sense to have the documentation for them intact as well. But I might have missed a discussion around this if there were some changes to our thoughts around this?

PR #6984, opened 2025-08-06
IFC-1718 Add specific events for threads in proposed change

gmazoyer:
This change adds a couple of events that are triggered when a thread is created or updated in a proposed change.

These events are triggered in addition to the classic node mutation events.

## Summary by CodeRabbit

* **New Features**
  * Introduced support for events when a proposed change thread is created or updated, enabling improved tracking of thread lifecycle actions.
  * Added corresponding event types to the user-facing event stream and GraphQL API.

* **Documentation**
  * Updated event reference documentation to describe the new "Proposed Change Thread Created" and "Proposed Change Thread Updated" events, including their payload structure and usage.

* **Tests**
  * Added functional tests to verify correct event firing for creation and update of proposed change threads.

### DIFF
--- a/backend/infrahub/events/generator.py
+++ b/backend/infrahub/events/generator.py
@@ -1,14 +1,16 @@
 from infrahub.context import InfrahubContext
 from infrahub.core.branch import Branch
 from infrahub.core.changelog.models import RelationshipChangelogGetter
-from infrahub.core.constants import MutationAction
+from infrahub.core.constants import InfrahubKind, MutationAction
 from infrahub.core.node import Node
+from infrahub.core.protocols import CoreProposedChange
 from infrahub.database import InfrahubDatabase
 from infrahub.events.node_action import NodeDeletedEvent, NodeMutatedEvent, NodeUpdatedEvent, get_node_event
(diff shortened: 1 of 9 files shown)


backend/infrahub/events/generator.py
```diff
@@ -68,4 +70,29 @@ async def generate_node_mutation_events(
...
-    return events + group_events
+
+    specific_events: list[InfrahubEvent] = []
+    if (kind := node.get_kind()) in [
+        InfrahubKind.CHANGETHREAD,
+        InfrahubKind.OBJECTTHREAD,
+        InfrahubKind.ARTIFACTTHREAD,
+        InfrahubKind.FILETHREAD,
+    ]:
+        proposed_change: CoreProposedChange = await node.change.get_peer(db=db, peer_type=CoreProposedChange)  # type: ignore[attr-defined]
```
ogenstad, line 81: As a follow up I think for newly created threads we should be able to access the peer proposed change ID from the changelog and avoid this DB lookup. It would be problematic for update events though. I'd suggest we get this merged first and then evaluate this small change.

ogenstad (approved):

Issue #6587, opened 2025-06-02
task: leverage attribute value database indexing


fatih-acar:
### Component

API Server / GraphQL

### Task Description

In #6250 we reintroduced AttributeValue node indexing using the `value` property. This significantly improves performance at high scale.
The current implementation is gated behind a feature flag because it does not support large attribute types (as defined here https://github.com/opsmill/infrahub/blob/infrahub-v1.2.11/backend/infrahub/types.py#L369).

In order to remove the feature flag, we must support these attributes (without indexing, since the database cannot index large values anyway).

One solution would be to have a dedicated node label (like `AttributeLargeValue`) where we would not index the `value` property.
Another solution is to have dedicated node label per attribute type (`AttributeIntegerValue`, etc), that would potentially lead to better indexing (since we would index per type, and not mix all types in a single index): this needs to be confirmed with benchmarks.

PR #6991, opened 2025-08-08
adds migration for new AttributeValueIndexed vertices

ajtmccarty:
completes #6587 

adds a migration and test for updating the correct AttributeValue vertices to AttributeValueIndexed vertices

this is actually somewhat more complex than it might seem at first glance b/c we want to use `:AttributeValue` vertices for attributes of kind `List`, `JSON`, and `TextArea` and we want to use `:AttributeValue:AttributeValueIndexed` for all other attribute kinds.

this migration will update all currently active attributes on all branches based on the current schema on each branch to have the correct `AttributeValue` vertex labels, but it won't make any changes for attributes that have been deleted. even this non-complete solution is still more complex than you might expect at first glance b/c schema and data can change independently of one another.

Here's an example of one situation we might enounter
1. update an attribute schema from `Text` to `TextArea` on `main`
2. create some nodes that use this attribute on `main`
3. create a branch
4. create some nodes that use the above attribute on the branch
5. update the attribute from `TextArea` back to `Text` on the branch

In this case we have to update some of the attribute values to be indexed and some not to be and the attribute values need to have different from times

TODO:

- [x]  test for attribute updates on branch for nodes created on main
- [x]  test for reverting large attribute type on branch where value is too large to be indexed
- [x] check migration on existing backups

## Summary by CodeRabbit

* **New Features**
  * Improved handling of large attribute values to optimize indexing and performance across branches.

* **Bug Fixes**
  * Ensured attribute values remain consistent and correctly indexed after migration across all branches, preventing duplicate relationships.

* **Tests**
  * Added comprehensive automated tests to validate migration behavior, data integrity, and correct indexing across branch scenarios.


ogenstad (approved):
Nice work!

Regarding:
```python
class GetLargeAttributeTypesQuery(Query):
    """For every active attribute on every branch, return a SchemaAttributeTimeframe object"""
```
As you tested this against existing backups, do we have a sense of how much time the migration will take to complete?

ajtmccarty:
> As you tested this against existing backups, do we have a sense of how much time the migration will take to complete?

Yes, it is actually pretty reasonable for a migration. The whole thing takes ~5m on my machine for a large customer dataset (6 million vertices) that includes customer schemas with List/JSON/TextArea attributes

PR #7007, opened 2025-08-11
IFC-1499 Check schema permission for dropdown/enum operations

### LINKED ISSUE #6410: bug: A user without schema rights can still update a dropdown options
### Component

API Server / GraphQL

### Infrahub version

1.2.8

### Current Behavior

A user with no rights regarding the schema, should not be able to add/delete dropdown options.

### Expected Behavior

A user with no rights regarding the schema, should not be able to add/delete dropdown options.

### Steps to Reproduce

n/a

### Additional Information

_No response_

gmazoyer:
This change ensures that the `MANAGE_SCHEMA` global permission is checked when trying to add or remove a dropdown/enum value.

Unit tests involving permissions have been slightly changed to use a fixture to set the permission backend to use.

## Summary by CodeRabbit

- Bug Fixes
  - Enforced that only users with the "Manage schema" permission can add or remove dropdown and enum values via GraphQL.
  - Schema enum mutations now require an explicit enum name in the input.

- Documentation
  - Changelog updated to note the "Manage schema" permission requirement for dropdown/enum changes.

- Tests
  - Test suite switched to a fixture-driven permission backend and updated test signatures to rely on the default permission fixture.

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -44,6 +44,7 @@ from infrahub.database import InfrahubDatabase, get_db
 from infrahub.lock import initialize_lock
 from infrahub.message_bus import InfrahubMessage, InfrahubResponse
 from infrahub.message_bus.types import MessageTTL
+from infrahub.permissions import LocalPermissionBackend
 from infrahub.services import InfrahubServices
 from infrahub.services.adapters.message_bus import InfrahubMessageBus
 from tests.adapters.log import FakeLogger
@@ -165,6 +166,14 @@ async def default_ipnamespace(db: InfrahubDatabase, register_core_models_schema)
         registry.default_ipnamespace = ip_namespace.id
         return ip_namespace
     return None
+
+
+@pytest.fixture
+def default_permission_backend() -> Generator[None, Any, Any]:
+    if not registry.permission_backends:
+        registry.permission_backends = [LocalPermissionBackend()]
+    yield
+    registry.permission_backends = []
 
 
 @pytest.fixture
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -170,10 +170,10 @@ async def default_ipnamespace(db: InfrahubDatabase, register_core_models_schema)
 
 @pytest.fixture
 def default_permission_backend() -> Generator[None, Any, Any]:
-    if not registry.permission_backends:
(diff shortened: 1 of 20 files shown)


backend/tests/conftest.py
```diff
@@ -167,6 +168,14 @@ async def default_ipnamespace(db: InfrahubDatabase, register_core_models_schema)
     return None
 
 
+@pytest.fixture
+def default_permission_backend() -> Generator[None, Any, Any]:
+    if not registry.permission_backends:
+        registry.permission_backends = [LocalPermissionBackend()]
+    yield
+    registry.permission_backends = []
```
ogenstad: What is it we want to do with this fixture? To me this look wrong in the sense that if another permissions_backend is set it will not change it back to the default and then once done it will remove all permissions backends.

Don't we want a copy to begin with and then assign `[LocalPermissionBackend()]` followed by a yield and then we revert to the original value of the permissions_backend?

backend/tests/unit/graphql/auth/query_permission_checker/test_object_permission_checker.py
```diff
@@ -263,7 +261,9 @@ async def test_setup(
 
         permissions_helper._first = first_account
 
-    async def test_first_account_tags(self, db: InfrahubDatabase, permissions_helper: PermissionsHelper) -> None:
+    async def test_first_account_tags(
+        self, db: InfrahubDatabase, default_permission_backend: None, permissions_helper: PermissionsHelper
```
ajtmccarty: could the `default_permission_backend` fixture be moved to the `test_setup` method so it isn't required in all of the `test_...` methods?

backend/tests/unit/graphql/queries/test_list_permissions.py
```diff
@@ -168,7 +165,9 @@ async def test_setup(
         await group.members.add(db=db, data={"id": first_account.id})
         await group.members.save(db=db)
 
-    async def test_first_account_tags(self, db: InfrahubDatabase, permissions_helper: PermissionsHelper) -> None:
+    async def test_first_account_tags(
+        self, db: InfrahubDatabase, default_permission_backend: None, permissions_helper: PermissionsHelper
```
ajtmccarty: same here

ajtmccarty (approved):

PR #7014, opened 2025-08-12
Wrap resource pool assignments within a lock context

ogenstad:
Due to timing issues we might end up with the same resource from a pool if multiple requests come in at once. Due to this we're wrapping the resource allocations within a global lock to prevent this behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of resource pool allocations by serializing concurrent requests per pool, preventing duplicate or invalid assignments under heavy load.
  * Applies to IP address, prefix, and number allocations.
  * No changes to public APIs.

* **Documentation**
  * Added changelog entry describing the fix for concurrent allocation issues.


gmazoyer (approved):

backend/infrahub/core/node/resource_manager/ip_address_pool.py
```diff
@@ -62,25 +63,26 @@ async def get_resource(
...
-        await reconciler.reconcile(ip_value=next_address, namespace=ip_namespace.id, node_uuid=node.get_id())
-
-        if identifier:
-            query_set = await IPAddressPoolSetReserved.init(
-                db=db, pool_id=self.id, identifier=identifier, address_id=node.id, at=at
-            )
-            await query_set.execute(db=db)
-
-        return node
+        async with lock.registry.get(name=self.get_id(), namespace="resource_pool"):
```
fatih-acar: I think whole `get_resource()` has to be locked, there might be TOCTOU issues when concurrent mutations occur with an identifier defined? (regarding line 40 to 51)

ogenstad: I might be missing something here, can you walk me through the scenario you see? The way I was thinking was that if we'd already had a reservation in the database/pool we wouldn't need to lock the pool to get that reservation. However, if we don't have a previous reservation, we'd want to have a lock while making the reservation.

Are you thinking if we have two mutations that try to update the same node at the same time and as such have the same `identifier`?

fatih-acar: > However, if we don't have a previous reservation, we'd want to have a lock while making the reservation.

For that specific statement, I think you need to lock early (before checking for a reservation).

> Are you thinking if we have two mutations that try to update the same node at the same time and as such have the same identifier?

Exactly, to solve this specific issue.
EDIT: I don't know if that can happen in a real world scenario though... I don't know how `identifier` is generated

ogenstad: I'm not sure either if we'll ever run into this but I moved the lock regardless. I think the main issue might be multiple mutations on against the same node. But better safe than sorry, it might take a slight performance hit but I don't think this in itself will be in the hot path.

Also I messed up this PR as I brought in the wrong branch.. Opened #7025 as a replacement instead.

fatih-acar: Thanks, looks good.
I'll run additional tests with a branch containing this and https://github.com/opsmill/infrahub/pull/6888 together

PR #7018, opened 2025-08-12
Fix unset draft (open) action for proposed change

pa-lem:
Also ensure the proposed change details are up to date after action submit


pa-lem:
@coderabbitai review

PR #7031, opened 2025-08-13
IFC-1745 Prevent draft proposed change merged via mutation

gmazoyer:
This change prevents a proposed change that is marked as a draft to be merged by using the proposed change update mutation.

There is also a slight change to an permission error message, to make it a bit more friendly, which is so small I think it can go inside that PR too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents merging proposed changes that are drafts or would become drafts during merge.
  * Improves permission error message for users lacking review rights on proposed changes.
* **Tests**
  * Added unit tests verifying draft proposed changes cannot be merged.
  * Updated functional tests to expect the new review permission denial message.

### DIFF
--- a/backend/infrahub/graphql/mutations/proposed_change.py
+++ b/backend/infrahub/graphql/mutations/proposed_change.py
@@ -137,6 +137,9 @@ class InfrahubProposedChangeMutation(InfrahubMutationMixin, Mutation):
             updated_state = ProposedChangeState(state_update)
             state.validate_state_transition(updated_state)
 
+        # Check if the draft state will change (defaults to current draft state)
+        will_be_draft = data.get("is_draft", {}).get("value", obj.is_draft.value)
+
         # Check before starting a transaction, stopping in the middle of the transaction seems to break with memgraph
         if updated_state == ProposedChangeState.MERGED and graphql_context.account_session:
             try:
@@ -150,7 +153,7 @@ class InfrahubProposedChangeMutation(InfrahubMutationMixin, Mutation):
                 raise ValidationError(str(exc)) from exc
 
         if updated_state == ProposedChangeState.MERGED:
-            if obj.is_draft.value:
+            if obj.is_draft.value or will_be_draft:
                 raise ValidationError("A draft proposed change is not allowed to be merged")
             data["state"]["value"] = ProposedChangeState.MERGING.value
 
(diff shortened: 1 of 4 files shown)


backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -150,6 +150,8 @@ async def mutate_update(
                 raise ValidationError(str(exc)) from exc
 
         if updated_state == ProposedChangeState.MERGED:
+            if obj.is_draft.value:
+                raise ValidationError("A draft proposed change is not allowed to be merged")
```
gmazoyer: I wonder if the checks performed allowing a PR to be merge inside the mutation should be a some point moved to the `ProposedChangeChecker.verify_proposed_change_is_mergeable` method instead; and call that method in the mutation.

If so, it's probably for another PR.

ogenstad: Can you create an internal housekeeping issue around this?

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -150,6 +150,8 @@ async def mutate_update(
                 raise ValidationError(str(exc)) from exc
 
         if updated_state == ProposedChangeState.MERGED:
+            if obj.is_draft.value:
```
ogenstad: I wonder if this is enough. I think we might end up with an issue if some creative person sent in a mutation that changed a PC from not being in draft mode to being a draft and tried to merge it in the same mutation. I.e. we might need something similar to what we do with updated_state?

ogenstad (commented):
Added a comment, I think we need a small change to this to ensure that it works even with the corner case.

backend/infrahub/graphql/mutations/proposed_change.py
gmazoyer: Good catch

gmazoyer: Should be covered now

ajtmccarty (approved):

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -150,6 +153,8 @@ async def mutate_update(
                 raise ValidationError(str(exc)) from exc
 
         if updated_state == ProposedChangeState.MERGED:
+            if obj.is_draft.value or will_be_draft:
```
ogenstad: Shouldn't this just be:

```python
if will_be_draft:
```
It could be that the PC was a draft but the mutation changes it to not be a draft in the same mutation as it's changed to the "merged" state. It looks like `will_be_draft` will always have the correct and current value?

PR #7033, opened 2025-08-13
Add proposed change event when multiple approvals are revoked

gmazoyer:
This change adds a new event to be raised when multiple approvals were revoked during a single transaction.

## Summary by CodeRabbit

* **New Features**
  * Added an “Approvals Revoked” event for proposed changes that records revoked approvals as a map of reviewer IDs to names and is exposed via the public events API and GraphQL.

* **Improvements**
  * Task processing now includes the new event in proposed-change grouping and aggregates per-reviewer related data for multi-review revocations.

* **Chores**
  * Registered the new event type in enums and exported it publicly; updated documentation and event mappings.

### DIFF
--- a/backend/infrahub/events/proposed_change_action.py
+++ b/backend/infrahub/events/proposed_change_action.py
@@ -1,6 +1,6 @@
-from typing import ClassVar
+from typing import ClassVar, Self
 
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from infrahub.core.constants import InfrahubKind, MutationAction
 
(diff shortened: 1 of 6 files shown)


ajtmccarty (approved):

ogenstad:
Wouldn't we need to have that info populated in order to see who's approval that got revoked? I think we should have that info as we are revoking the previous approval?

gmazoyer:
> Wouldn't we need to have that info populated in order to see who's approval that got revoked? I think we should have that info as we are revoking the previous approval?

Do you mean that we should raise one event per revoked approval? E.g. if a proposed change has been approved 5 times, raise 5 approval revoked events.

backend/infrahub/events/proposed_change_action.py
```diff
@@ -150,6 +150,33 @@ class ProposedChangeRejectionRevokedEvent(ProposedChangeReviewRevokedEvent):
     event_name: ClassVar[str] = f"{EVENT_NAMESPACE}.proposed_change.rejection_revoked"
 
 
+class ProposedChangeApprovalsRevokedEvent(ProposedChangeEvent):
+    reviewer_account_ids: list[str] = Field(..., description="IDs of accounts whose approval was revoked")
+    reviewer_account_names: list[str] = Field(..., description="Names of accounts whose approval was revoked")
```
ogenstad: I wonder if a `dict[str,str]` would be better here. It doesn't really matter from a Prefect point of view as we're iterating over each list to match them. But these fields would also be included in the payload of the event, as such a receiver would have to combine them to match the ids to the names. What do you think?

ogenstad (commented):
Added a small comment, other than that I think it looks good.

ogenstad (approved):

PR #7042, opened 2025-08-14
schema update migration for attribute kind update to/from JSON/List/TextArea

ajtmccarty:
IFC-1734

add a schema update migration (the kind that runs when a new schema is loaded, not a database upgrade migration) to handle correctly updating `AttributeValue[Indexed]` vertices so that "large" attribute types (`JSON` / `List` / `TextArea`) are _not_ indexed and all other attribute types _are_ indexed.

also fixes a couple of other issues noted in my comments

## Summary by CodeRabbit

- New Features
  - Per-branch migration for attribute-kind changes with automatic reindexing of attribute values.

- Bug Fixes
  - Safer model updates: incompatible field assignments are skipped.
  - String content validation enforces length limits for non-large text attributes.
  - Constraint validation now accepts migration-required rules when a validator exists.
  - Pre-migration truncation to avoid oversize data failures.

- Documentation
  - Validator docs: attribute "kind" status changed to migration_required.

- Tests
  - New and updated integration/unit tests covering migrations, indexing state, truncation, and validator edge cases.

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -680,6 +680,12 @@ class String(BaseAttribute):
     type = str
     value: str
 
+    @classmethod
+    def validate_content(cls, value: Any, name: str, schema: AttributeSchema) -> None:
+        if value is not None and not is_large_attribute_type(schema.kind):
+            validate_string_length(value=str(value))
+        super().validate_content(value=value, name=name, schema=schema)
+
 
 class StringOptional(String):
     value: str | None
--- a/backend/infrahub/core/models.py
+++ b/backend/infrahub/core/models.py
@@ -569,7 +569,10 @@ class HashableModel(BaseModel):
 
         for field_name in other.model_fields.keys():
             if not hasattr(self, field_name):
-                setattr(self, field_name, getattr(other, field_name))
+                try:
+                    setattr(self, field_name, getattr(other, field_name))
+                except ValueError:
+                    pass
                 continue
 
             attr_other = getattr(other, field_name)
--- a/backend/infrahub/core/models.py
+++ b/backend/infrahub/core/models.py
@@ -572,6 +572,7 @@ class HashableModel(BaseModel):
                 try:
                     setattr(self, field_name, getattr(other, field_name))
                 except ValueError:
+                    # handles the case where self and other are different types and other has fields that self does not
                     pass
                 continue
 
--- a/backend/infrahub/core/validators/attribute/kind.py
+++ b/backend/infrahub/core/validators/attribute/kind.py
@@ -65,8 +65,12 @@ class AttributeKindUpdateValidatorQuery(AttributeSchemaValidatorQuery):
             if value in (None, NULL_VALUE):
                 continue
             try:
+                attr_value = result.get("attribute_value")
                 infrahub_attribute_class.validate_format(
-                    value=result.get("attribute_value"), name=self.attribute_schema.name, schema=self.attribute_schema
+                    value=attr_value, name=self.attribute_schema.name, schema=self.attribute_schema
+                )
+                infrahub_attribute_class.validate_content(
+                    value=attr_value, name=self.attribute_schema.name, schema=self.attribute_schema
                 )
             except ValidationError:
                 grouped_data_paths.add_data_path(
--- a/backend/infrahub/core/validators/determiner.py
+++ b/backend/infrahub/core/validators/determiner.py
@@ -98,7 +98,10 @@ class ConstraintValidatorDeterminer:
                 continue
 
             prop_field_update = prop_field_info.json_schema_extra.get("update")
-            if prop_field_update != UpdateSupport.VALIDATE_CONSTRAINT.value:
+            if prop_field_update not in (
+                UpdateSupport.VALIDATE_CONSTRAINT.value,
+                UpdateSupport.MIGRATION_REQUIRED.value,
+            ):
                 continue
 
             if getattr(schema, prop_name) is None:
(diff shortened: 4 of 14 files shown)


backend/infrahub/core/models.py
```diff
@@ -569,7 +569,10 @@ def update(self, other: HashableModel) -> Self:
 
         for field_name in other.model_fields.keys():
             if not hasattr(self, field_name):
-                setattr(self, field_name, getattr(other, field_name))
+                try:
+                    setattr(self, field_name, getattr(other, field_name))
+                except ValueError:
+                    pass
```
ajtmccarty, line 576: I came across an issue here during my testing when updating the kind of an Attribute such that the new kind used a different sub-class of `AttributeParameters` than the old kind. For example, updating a attribute's type from `Text` (uses the `TextAttributeParameters` class that includes, `regex`, `min_length`, and `max_length`) to `URL` (uses the `AttributeParameters` class that has no actual parameter options)
the logic here in  in `HashableModel.update` ends up trying to set the `regex` attribute on the `AttributeParameters` class and fails with a `ValueError` (`ValueError: "AttributeParameters" object has no field "regex"`)
I'm not sure if the code here before was never reached or if I am misunderstanding it, but it seems like it would never work
```
if not hasattr(self, "x"):
    setattr(self, "x", "y")
```
please let me know if it looks like I need to change this back and find a different way to handle the `AttributeParameters`

backend/infrahub/core/validators/determiner.py
ajtmccarty, line 1: the updates in here are to support the case where we want to use both a constraint validator _and_ a migration for a particular schema change. this used to only be the case for 'node.attribute.add' and 'node.inherit_from.update', but it now also includes 'attribute.kind.update'.
the issue is that `ConstraintValidatorDeterminer` was not correctly searching for properties that had `json_schema_extra.update = "migration_required"` _and_ had an entry in the `CONSTRAINT_VALIDATOR_MAP` dictionary
a few of our existing unit tests actually caught the issue and the small changes in this module seem to fix it

backend/infrahub/core/models.py
ogenstad: I'm not completely sure about this. Part of me wants to say that if you have a Text kind with parameters and you want to convert that to another kind that doesn't support the parameters that you have set that you'd first do a migration to remove the setting of that parameter, then after this the migration wouldn't be a problem.

Any thoughts around that?

ogenstad (approved):
LGTM, not sure about my comment but what I think would make most sense for now is that the schema update is rejected if someone tries to migrate a kind to another where the parameters doesn't match i.e. that this failure would happen before the migration takes place (the alternative could be to silently succeed which I take is the behaviour now but it feels a bit wrong to me).

One thing around this. What happens if someone tries to update the kind from TextArea to Text and we have nodes with data that are too large? It would be nice if we can reject the schema and indicate the problematic nodes. What are your thoughts around this?

backend/infrahub/core/attribute.py
```diff
@@ -680,6 +680,12 @@ class String(BaseAttribute):
     type = str
     value: str
 
+    @classmethod
+    def validate_content(cls, value: Any, name: str, schema: AttributeSchema) -> None:
+        if value is not None and not is_large_attribute_type(schema.kind):
+            validate_string_length(value=str(value))
+        super().validate_content(value=value, name=name, schema=schema)
```
ajtmccarty, line 687: added this check to the `String` attribute type class so that we can raise an error during an attribute type change in `AttributeKindChecker`

backend/infrahub/core/validators/attribute/kind.py
```diff
@@ -65,8 +65,12 @@ async def get_paths(self) -> GroupedDataPaths:
             if value in (None, NULL_VALUE):
                 continue
             try:
+                attr_value = result.get("attribute_value")
                 infrahub_attribute_class.validate_format(
-                    value=result.get("attribute_value"), name=self.attribute_schema.name, schema=self.attribute_schema
+                    value=attr_value, name=self.attribute_schema.name, schema=self.attribute_schema
+                )
+                infrahub_attribute_class.validate_content(
+                    value=attr_value, name=self.attribute_schema.name, schema=self.attribute_schema
```
ajtmccarty, line 73: added this b/c we _should_ be checking and rejecting an attribute kind update if the change would cause values to violate the parameters of the new kind, e.g. a regex or min/max_length update. this will also catch the case when a `TextArea` is updated to `Text`, but one of the values is too long to be indexed

ajtmccarty (commented):
@ogenstad good point about the too-large `TextArea` issue during an attribute kind update. the changes (and tests) in my latest commit should handle that correctly.

backend/infrahub/core/models.py
ajtmccarty: I think we would still encounter the same issue even if the user set all the parameter values to `null`/`None` b/c the logic here would still see that `regex`, for example, is a field on the `TextAttributeParameters` and try to set the `regex` attribute on the `AttributeParameters` class, where it does not exist

I know that we want to get away from doing "automagic" schema updates to make things "easier" (and harder to understand) for the user, but I think this it is a reasonable assumption that if you change the kind of an attribute to something that does not support the old parameters, then those old parameters will be removed.

ogenstad: Ok, sounds fine to me. However can you add a comment with a brief explanation of this behaviour after catching the ValueError and above the `pass` statement so that we know why it looks like this when we revisit it later?

ogenstad (approved):

PR #7054, opened 2025-08-18
Automatic revoke event in proposed change - IHP-22

pa-lem:
Add event for automatic approval revoke after changes

## Summary by CodeRabbit

- **New Features**
  - Global event view now recognizes and displays proposed-change events with a dedicated proposed-change title.
  - Added support for a new "approvals revoked" proposed-change event type.

- **Bug Fixes**
  - For the plural "approvals revoked" event, the UI shows a gray status message ("Changes occured in the source branch, approval revoked") instead of the discussion thread.
  - Existing singular "approval revoked" and other event renderings remain unchanged.


ogenstad:
This is what it looks like when I run this:

Screenshot 2025-08-21 at 16 43 19

One small caveat is the text "Changes occurred in the proposed change", this could be clarified with "Changes occurred in the source branch" as it's not the proposed change itself that has changed.

PR #7060, opened 2025-08-18
make keyword scanner play nice

petercrocker:
## Summary by CodeRabbit

* **Chores**
  * Made the CI keyword scan non-blocking so scan failures no longer break builds.
  * Added conditions so scans run only on pushes and pull requests from the same repository, skipping fork PRs.
  * Reduces CI noise and unnecessary runs for projects without configured keywords.
  * Improves CI reliability without affecting user-facing behavior.


.github/workflows/keyword-scan.yml
```diff
@@ -9,13 +9,18 @@ env:
 jobs:
   scan:
     runs-on: ubuntu-latest
+    continue-on-error: true  # Make the whole job non-blocking
```
FragmentedPacket: Could we move the `if` up here so the `if` does have to be on each step?

ogenstad: Unfortunately I'm not sure this will help us that much. In my mind GitHub's `continue-on-error` is pretty useless as it doesn't give any kind of indication that something didn't go as expected this is unlike allow_failure from GitLab (https://docs.gitlab.com/ci/yaml/#allow_failure) where you can clearly see that the pipeline failed.

Using `continue-on-error` will cause a green pipeline even if someone internally adds any of the keywords within a commit as the job will fail but still be marked as success.

I don't know if this works but we'd want something like this:

```yaml
if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
```

The idea would be to only run this job if the PR originated from our own repo. The syntax here might be incorrect.

Probably the action should even fail if the keyword environment was empty as a sanity check if someone on our end had cleared out that variable for whatever reason.

BeArchiTek (approved):

PR #7062, opened 2025-08-19
fixes for diffing and merging a branch into an emtpy default branch

### LINKED ISSUE #6484: bug: Cannot merge schema/data if main has schema/data (Fresh instance)
h3.  Component

[| #component]

API Server / GraphQL

h3.  Infrahub version

[| #infrahub-version]

1.2.9/1.2.10

h3.  Current Behavior

[| #current-behavior]

Whenever I spin up an empty instance of Infrahub and create a new branch and populate with a new schema and data I get a merge failure message.

This seems to fail with a larger initial data set, but a single model and single created object of the model doesn't seem to reproduce the issue.

h3.  Expected Behavior

[| #expected-behavior]

Be able to spin up Infrahub and build a schema in a branch and populate data then being able to merge branch into `main` from a fresh instance.

h3.  Steps to Reproduce

[| #steps-to-reproduce]

# Launch demo instance of Infrahub
# Create a new branch: `infrahubctl branch create test`
# Clone down [https://github.com/opsmill/infrahub-demo-dc-fabric.git| https://github.com/opsmill/infrahub-demo-dc-fabric.git]
# `infrahubctl schema load models --branch test`
# Populate basic data: `infrahubctl run bootstrap/create_basic.py --branch test`
# Attempt to merge in branch

h3.  Additional Information

[| #additional-information]

[!https://github.com/user-attachments/assets/9dd40e82-ae64-4ff1-aeca-ca47800ac1e8!| https://github.com/user-attachments/assets/9dd40e82-ae64-4ff1-aeca-ca47800ac1e8]

ajtmccarty:
fixes #6484 

there were a few small bugs that would prevent retrieving a diff of and merging a branch with new schema and data on a fresh instance of Infrahub
- update `SchemaBranch.update` to include `Profile` and `Template` schemas, otherwise adding a new schema and new instance of a `Profile` or `Template` schema would fail
- update logic for retrieving a diff to use the correct branch when getting information about parents of nodes in the diff
- add a big old test to make sure this does not regress

## Summary by CodeRabbit

- New Features
  - Expanded schema updates to include profiles and templates during diff/merge.
  - Improved diff tree to determine parent relationship names with branch-aware context.

- Bug Fixes
  - Fixed failures when generating diffs and merging a branch that introduces new schema and data into a fresh instance.

- Tests
  - Added end-to-end integration tests covering proposed changes, diff tree inspection, and merge verification.

- Refactor
  - Minor internal metrics and request-handling cleanup; removed an obsolete public helper for parent-relationship info.

### DIFF
--- a/backend/infrahub/core/diff/model/path.py
+++ b/backend/infrahub/core/diff/model/path.py
@@ -22,8 +22,6 @@ if TYPE_CHECKING:
     from neo4j.graph import Relationship as Neo4jRelationship
     from whenever import TimeDelta
 
-    from infrahub.graphql.initialization import GraphqlContext
-
 
 @dataclass
 class TimeRange:
@@ -315,12 +313,6 @@ class EnrichedDiffRelationship(BaseSummary):
 
 
 @dataclass
-class ParentNodeInfo:
-    node: EnrichedDiffNode
-    relationship_name: str = "undefined"
-
-
-@dataclass
 class EnrichedDiffNode(BaseSummary):
     identifier: NodeIdentifier
     label: str
@@ -363,37 +355,6 @@ class EnrichedDiffNode(BaseSummary):
         for rel in self.relationships:
             rel.clear_conflicts()
         self.conflict = None
-
-    def get_parent_info(self, graphql_context: GraphqlContext | None = None) -> ParentNodeInfo | None:
-        for r in self.relationships:
-            for n in r.nodes:
-                relationship_name: str = "undefined"
-
-                if not graphql_context:
-                    return ParentNodeInfo(node=n, relationship_name=relationship_name)
-
-                node_schema = graphql_context.db.schema.get(name=self.kind)
-                rel_schema = node_schema.get_relationship(name=r.name)
-
-                parent_schema = graphql_context.db.schema.get(name=n.kind)
-                rels_parent = parent_schema.get_relationships_by_identifier(id=rel_schema.get_identifier())
-
-                if rels_parent and len(rels_parent) == 1:
-                    relationship_name = rels_parent[0].name
-                elif rels_parent and len(rels_parent) > 1:
-                    for rel_parent in rels_parent:
-                        if (
-                            rel_schema.direction == RelationshipDirection.INBOUND
-                            and rel_parent.direction == RelationshipDirection.OUTBOUND
-                        ) or (
-                            rel_schema.direction == RelationshipDirection.OUTBOUND
-                            and rel_parent.direction == RelationshipDirection.INBOUND
-                        ):
-                            relationship_name = rel_parent.name
-                            break
-
-                return ParentNodeInfo(node=n, relationship_name=relationship_name)
-        return None
 
     def get_all_child_nodes(self) -> set[EnrichedDiffNode]:
         all_children = set()
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -226,8 +226,18 @@ class SchemaBranch:
     def update(self, schema: SchemaBranch) -> None:
         """Update another SchemaBranch into this one."""
 
-        local_kinds = list(self.nodes.keys()) + list(self.generics.keys())
-        other_kinds = list(schema.nodes.keys()) + list(schema.generics.keys())
+        local_kinds = (
+            list(self.nodes.keys())
+            + list(self.generics.keys())
+            + list(self.profiles.keys())
+            + list(self.templates.keys())
+        )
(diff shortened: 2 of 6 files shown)


backend/infrahub/core/diff/model/path.py
```diff
@@ -364,37 +356,6 @@ def clear_conflicts(self) -> None:
             rel.clear_conflicts()
         self.conflict = None
 
-    def get_parent_info(self, graphql_context: GraphqlContext | None = None) -> ParentNodeInfo | None:
```
ajtmccarty, line 367: moved to `backend/infrahub/graphql/queries/diff/tree.py` where it belongs

backend/infrahub/graphql/queries/diff/tree.py
```diff
@@ -166,6 +180,43 @@ async def to_diff_tree(
             num_conflicts=enriched_diff_root.num_conflicts,
         )
 
+    def _get_parent_info(
```
ajtmccarty, line 183: moved from `backend/infrahub/core/diff/model/path.py` and updated to use the correct branch for retrieving schema

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -226,8 +226,18 @@ def diff(self, other: SchemaBranch) -> SchemaDiff:
...
+            + list(self.generics.keys())
+            + list(self.profiles.keys())
+            + list(self.templates.keys())
+        )
+        other_kinds = (
+            list(schema.nodes.keys())
+            + list(schema.generics.keys())
+            + list(schema.profiles.keys())
+            + list(schema.templates.keys())
+        )
```
ajtmccarty: fixed to include all types of schemas

backend/infrahub/graphql/app.py
```diff
@@ -231,6 +231,7 @@ async def _handle_http_request(
                 operation_name=operation_name,
                 branch=branch,
             )
+        impacted_models = analyzed_query.query_report.impacted_models
```
ajtmccarty, line 234: made this change to fix an error that came up in the integration test in this PR
when sending the GraphQL mutation to merge a proposed change, the call to `analyzed_query.query_report.impacted_models` below would run after the mutation was completed and would raise an error when trying to retrieve the `CoreProposedChange` schema from the `SchemaBranch`
I did not observe this error when testing manually
I think something in our testing setup is causing the `_cache` of different `SchemaBranch`s to be shared and updated in a manner that can leave them out of sync. I could not find the exact cause, but I think making this change should make no functional difference to this logic

ogenstad: Sounds weird but I don't see any harm with making this change.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -226,8 +226,18 @@ def diff(self, other: SchemaBranch) -> SchemaDiff:
...
+            list(self.nodes.keys())
+            + list(self.generics.keys())
+            + list(self.profiles.keys())
+            + list(self.templates.keys())
+        )
+        other_kinds = (
+            list(schema.nodes.keys())
+            + list(schema.generics.keys())
+            + list(schema.profiles.keys())
+            + list(schema.templates.keys())
```
ogenstad: Potentially we could have the listing of these keys as a private function, though perhaps mostly relevant if we have even more variations in the future.

backend/infrahub/graphql/queries/diff/tree.py
```diff
@@ -166,6 +180,43 @@ async def to_diff_tree(
...
+        self, diff_node: EnrichedDiffNode, graphql_context: GraphqlContext | None = None
+    ) -> ParentNodeInfo | None:
+        for r in diff_node.relationships:
+            for n in r.nodes:
+                relationship_name: str = "undefined"
```
ogenstad, line 188: I know you've only moved code around here, just highlighting. Might we have any weird scenarios if we have a schema where the relationship name is "undefined"? I'm assuming this variable will almost always be replaced.

ogenstad (approved):

backend/infrahub/graphql/queries/diff/tree.py
ajtmccarty: I think this would be all right if the parent relationship was actually named "undefined" b/c the response would still be correct, but that's not something I considered, so thanks for the question

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: we actually already have this as a property (`all_names`), so I've updated this to use it

PR #7067, opened 2025-08-20
Fix artifacts generation validation

### LINKED ISSUE #6976: bug: Artifact generation fails when device is removed from the artifact definition group
h3.  Component

[| #component]

API Server / GraphQL

h3.  Infrahub version

[| #infrahub-version]

1.3.2/3/4

h3.  Current Behavior

[| #current-behavior]

Artifact generation failure within a Proposed Change if a device is removed from the artifact definition's target group.

[!https://github.com/user-attachments/assets/7218bdc0-dcb6-4e62-aeed-95e4d000673f|width=1518,height=256!| https://github.com/user-attachments/assets/7218bdc0-dcb6-4e62-aeed-95e4d000673f]

h3.  Expected Behavior

[| #expected-behavior]

Artifact generation to succeed and not run against the device that was removed from the target group

h3.  Steps to Reproduce

[| #steps-to-reproduce]

# Fresh instance of Infrahub on 1.3.4
# invoke load-schema load-data ([https://github.com/opsmill/infrahub-demo-dc-fabric| https://github.com/opsmill/infrahub-demo-dc-fabric])
# poetry run infrahubctl run bootstrap/generate_topology.py topology=fra05-pod1
# Add repository to Infrahub (artifacts get generated)
# Create new branch
# Remove `fra05-pod1-leaf1` from `arista_devices` group
# Create a PC with the artifact still present
# it should fail on artifact generation

h3.  Additional Information_No response_

[| #additional-information_no-response_]

LucasG0:
Fixes https://github.com/opsmill/infrahub/issues/6976

Issue is that `artifact.object` was not fetched as `include` does not work, so if the node was not in store, it wouldn't work. This PR relies on fixing sdk `include` here: https://github.com/opsmill/infrahub-sdk-python/pull/513

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed intermittent UI/data inconsistencies by changing data-loading order for generators, artifacts, and validations to preserve object relationships.

* **Performance**
  * Eliminated redundant data fetches during generator and artifact processing to reduce network calls and improve responsiveness.

* **Refactor**
  * Reordered internal fetch logic for generators, artifacts, and validation flows without altering public APIs or behavior.

* **Chores**
  * Updated Python SDK submodule.

### DIFF
--- a/backend/infrahub/generators/tasks.py
+++ b/backend/infrahub/generators/tasks.py
@@ -168,6 +168,15 @@ async def request_generator_definition_run(
 
     client = get_client()
 
+    # Needs to be fetched before fetching group members otherwise `object` relationship would override
+    # existing node in client store without the `name` attribute due to #521
+    existing_instances = await client.filters(
+        kind=InfrahubKind.GENERATORINSTANCE,
+        definition__ids=[model.generator_definition.definition_id],
+        include=["object"],
+        branch=model.branch,
+    )
+
     group = await client.get(
         kind=InfrahubKind.GENERICGROUP,
         prefetch_relationships=True,
@@ -177,12 +186,6 @@ async def request_generator_definition_run(
     )
     await group.members.fetch()
 
-    existing_instances = await client.filters(
-        kind=InfrahubKind.GENERATORINSTANCE,
-        definition__ids=[model.generator_definition.definition_id],
-        include=["object"],
-        branch=model.branch,
-    )
     instance_by_member = {}
     for instance in existing_instances:
         instance_by_member[instance.object.peer.id] = instance.id
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -646,16 +646,19 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
         context=context,
     )
 
-    await artifact_definition.targets.fetch()
-    group = artifact_definition.targets.peer
-    await group.members.fetch()
-
+    # Needs to be fetched before fetching group members otherwise `object` relationship would override
+    # existing node in client store without the `name` attribute due to #521
     existing_artifacts = await client.filters(
         kind=InfrahubKind.ARTIFACT,
         definition__ids=[model.artifact_definition.definition_id],
         include=["object"],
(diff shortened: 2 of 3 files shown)


backend/infrahub/proposed_change/tasks.py
```diff
@@ -617,6 +617,7 @@ async def validate_artifacts_generation(
         definition__ids=[model.artifact_definition.definition_id],
         include=["object"],
         branch=model.source_branch,
+        prefetch_relationships=True,
```
LucasG0: Only this line is necessary to fix the actual issue
We later try to access the `object` relationship while it has not been fetched. I am not sure how this has ever worked, it might be that previously the `include` parameters would enforce fetching specified relationships, or that `object` is already located in the store

backend/infrahub/generators/tasks.py
```diff
@@ -178,6 +178,7 @@ async def request_generator_definition_run(
         definition__ids=[model.generator_definition.definition_id],
         include=["object"],
         branch=model.branch,
+        prefetch_relationships=True,
```
LucasG0: Adding this line actually introduces these CI failures when executing generator's graphql query because `variables={"name": None}`. However, `graphql_query_post` payload does not accept non-null values for `variables` field, thus raising a 422 error. The root cause for `name` being `None` is that fetching relationships here overrides the `name` previously stored in the sdk store.

backend/infrahub/proposed_change/tasks.py
LucasG0: We actually should patch `include` within sdk instead of using `prefetch_relationships` here.

backend/infrahub/generators/tasks.py
```diff
@@ -168,6 +168,15 @@ async def request_generator_definition_run(
 
     client = get_client()
 
+    # Needs to be fetched before fetching group members otherwise `object` relationship would override
+    # existing node in client store without the `name` attribute due to #521
```
ajtmccarty, line 172: I don't like adding the same comment in 4 different places...
but I also cannot think of a better option right now

ajtmccarty (approved):

ogenstad (approved):

PR #4940, opened 2024-11-15
Forbid changing optional property of inherited attributes

gmazoyer:
Fixes #4936

Issue #4936, opened 2024-11-14
bug: Overriding `Optional` between Generic and Node crash server


BeArchiTek:
### Component

API Server / GraphQL

### Infrahub version

1.0.4

### Current Behavior

When loading a schema and having the same attribute in both the Generic and the node inheriting from it, it's possible to load the schema, *BUT* if the `optional` data of the attribute changes, the server will loop on the TypeError.

### Expected Behavior

Schema should be refuse if the `inherited: true` is not present

### Steps to Reproduce

1. Start infrahub `1.0.4` and load this schema :
```yaml
# yaml-language-server: $schema=https://schema.infrahub.app/infrahub/schema/latest.json
---
version: '1.0'

generics:
  - name: Sandwich
    namespace: Meal
    attributes:
      - name: description
        kind: Text
        optional: false

nodes:
  - name: Grilled
    namespace: Sandwich
    inherit_from:
      - MealSandwich
    attributes:
      - name: description
        kind: Text
        optional: true

```
2. Check the log of the server, you should see
```console
2024-11-14 14:56:04 TypeError: Interface field MealSandwich.description expects type TextAttribute! but 
2024-11-14 14:56:04 SandwichGrilled.description is type TextAttribute.
```

PR #7079, opened 2025-08-25
allow updating mandatory attr on a generic to be optional

### LINKED ISSUE #6800: bug: Inherited attribute cannot be changed from required to not required
h3.  Component

[| #component]

API Server / GraphQL

h3.  Infrahub version

[| #infrahub-version]

1.3.2

h3.  Current Behavior

[| #current-behavior]

Cannot update a field defined on a generic that was originally required, but no longer is.

    ❯ infrahubctl schema check models/base --branch test
    Unable to load the schema:
      InfraInterfaceL2.speed inherited from InfraInterface must have the same value for property "optional" ["True", "False"]

h3.  Expected Behavior

[| #expected-behavior]

Change an attribute on a generic from required to not required and be able to check/load the schema.

h3.  Steps to Reproduce

[| #steps-to-reproduce]

# `invoke demo.start`
# `infrahubctl schema load models/base`
# Update `InfraInterface.speed` in `models/base/dcim.yml` to `optional: true`
# `infrahubctl schema load models/base`

h3.  Additional Information

[| #additional-information]

To get around this, you either need to:

# Delete the attribute and then re-add it as being optional
# Set the attribute on all inheriting schemas

ajtmccarty:
fixes #6800 

remove logic that prevented updating a mandatory generic attribute to optional if any schema inherited from the generic

update integration test to include test cases for
- creating a generic schema with a mandatory attribute and inheriting node schema that override the attribute to be optional
- updating an inheriting node schema to set the mandatory attribute to optional
- updating the generic schema attribute from mandatory to optional
- deleting the attribute from the generic schema and verifying that the overridden node-level attributes remain

also did manual testing of this to double verify that it all works as expected

changing a generic schema's attribute from mandatory to optional used to cause this bug (#4936), which was resolved by preventing this change in #4940, but we must have fixed the underlying problem between then and now because just removing the logic to prevent the mandatory -> optional update seems to have fixed the issue with no other changes required

### DIFF
--- a/backend/infrahub/core/schema/node_schema.py
+++ b/backend/infrahub/core/schema/node_schema.py
@@ -53,11 +53,6 @@ class NodeSchema(GeneratedNodeSchema):
                         f"{self.kind}.{attribute.name} inherited from {interface.namespace}{interface.name} must be the same kind "
                         f'["{interface_attr.kind}", "{attribute.kind}"]'
                     )
-                if attribute.optional != interface_attr.optional:
-                    raise ValueError(
-                        f"{self.kind}.{attribute.name} inherited from {interface.namespace}{interface.name} must have the same value for property "
-                        f'"optional" ["{interface_attr.optional}", "{attribute.optional}"]'
-                    )
 
         for relationship in self.relationships:
             if (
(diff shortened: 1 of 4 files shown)


ajtmccarty:
can be merged for the release if approved

gmazoyer (approved):

PR #7080, opened 2025-08-26
Update screenshots in docs

wvandeun:


docs/docs/media/topics/activity-logs/activity_log_device.png
dgarros, line 1: Looks like there was a timing issue on this one

dgarros:
Thanks for regenerating the screenshots @wvandeun, overall I agree there isn't many changes so not sure if we need to merge this PR for this release
The take away from this for me is that we need to plan to update the current process to capture more Screenshots from the Frontend because there was a lot of changes in this release but it's not showing up here

PR #7088, opened 2025-08-26
handle overridden choice attributes on generics during schema integrity check

### LINKED ISSUE #7086: bug: overridden choices attribute schema is incorrectly checked against generic's attribute schema
h3\. Component

[| #component]

API Server / GraphQL

h3\. Infrahub version

[| #infrahub-version]

1.3.8

h3\. Current Behavior

[| #current-behavior]

If a node schema inherits from a generic schema and overrides the choices of a `Dropdown` attribute, the values of the node's attribute will be incorrectly checked against the choices defined on the generic schema during the proposed change pipeline's schema integrity check

h3\. Expected Behavior

[| #expected-behavior]

We should check the values of inherited `Dropdown` attributes against the values defined on the generic schema only if they are not overridden at the node level

h3\. Steps to Reproduce

[| #steps-to-reproduce]

h1.  Create and load a schema that includes a generic with choices and a node that inherits from the generic and overrides the choices. something like this

[| #create-and-load-a-schema-that-includes-a-generic-with-choices-and-a-node-that-inherits-from-the-generic-and-overrides-the-choices-something-like-this]

    generics:
      - name: GenericRack
        namespace: Location
        attributes:
          - name: status
            kind: Dropdown
            choices:
              - name: generic_active
              - name: generic_planned
    nodes:
      - name: Rack
        namespace: Location
        inherit_from:
          - "LocationGenericRack"
        attributes:
          - name: status
            kind: Dropdown
            choices:
              - name: active
              - name: planned

h1.  create a branch

[| #create-a-branch]

h1.  add an instance of `LocationRack` on the branch

[| #add-an-instance-of-locationrack-on-the-branch]

h1.  create a proposed change

[| #create-a-proposed-change]

h1.  the schema integrity check will fail with a message like

[| #the-schema-integrity-check-will-fail-with-a-message-like]

    Attribute-level 'choices' constraint violation on schema 'LocationGenericRack'. Node (LocationGenericRack: <UUID>) is not compliant. The error relates to field role='active'.

h3\. Additional Information

[| #additional-information]

_No response_

ajtmccarty:
fixes #7086 

there is a bug in the `AttributeChoicesChecker` that we use during schema validation to check if the values of a choice-type attribute (`Dropdown`) that incorrectly checks the allowed values of a generic schema for inheriting schema that have overridden the generic attribute with their own choices.

For example, say I have this schema
- a generic schema `GenericA` with a `role` attribute that allows these values: [`one`, `two`]
- a node schema `NodeB` that inherits from `GenericA` and has overridden the `role` attribute to allow these values [`three`, `four`]
when schema validation runs, it will incorrectly identify instances of `NodeB` as having invalid `role` values b/c the role is not `one` or `two`

the fix is to update `AttributeChoicesChecker` to ignore Nodes that have overridden the attribute being checked

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation for dropdown/choice attributes when node schemas override a generic attribute, preventing false “illegal value” errors.
  * Improved handling of schema inheritance so generic and node-specific choices can differ without triggering validation issues.
  * Ensures accurate validation results across branches during attribute updates and deletions.

* **Documentation**
  * Added changelog entry describing the fix for dropdown attribute validation on overridden generic attributes.

### DIFF
--- a/backend/tests/unit/core/constraint_validators/test_attribute_choices_update.py
+++ b/backend/tests/unit/core/constraint_validators/test_attribute_choices_update.py
@@ -13,7 +13,7 @@ from infrahub.core.validators.model import SchemaConstraintValidatorRequest
 from infrahub.database import InfrahubDatabase
 
 
-async def test_query_new_choice_value(db: InfrahubDatabase, default_branch: Branch, criticality_schema):
+async def test_new_choice_value(db: InfrahubDatabase, default_branch: Branch, criticality_schema):
     crit_low = await Node.init(db=db, schema=criticality_schema)
     await crit_low.new(db=db, name="low", level=4, status="active")
     await crit_low.save(db=db)
@@ -41,7 +41,7 @@ async def test_query_new_choice_value(db: InfrahubDatabase, default_branch: Bran
     assert len(all_data_paths) == 0
 
 
-async def test_query_remove_choice(db: InfrahubDatabase, default_branch: Branch, criticality_schema):
+async def test_remove_choice(db: InfrahubDatabase, default_branch: Branch, criticality_schema):
     crit_low = await Node.init(db=db, schema=criticality_schema)
     await crit_low.new(db=db, name="low", level=4, status="active")
     await crit_low.save(db=db)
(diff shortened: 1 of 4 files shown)


backend/tests/unit/core/constraint_validators/test_attribute_choices_update.py
ajtmccarty, line 1: I added a new test to this file and updated all of the other tests to use `AttributeChoicesChecker` instead of just the `AttributeChoicesUpdateValidatorQuery` b/c the `AttributeChoicesChecker` actually includes some logic that should be tested now

gmazoyer (approved):

Issue #6934, opened 2025-08-01
bug: it is possible for an IP prefix to be its own parent


ajtmccarty:
# 
## 
### Component

API Server / GraphQL

# 
## 
### Infrahub version

1.3.3

# 
## 
### Current Behavior

It is possible for an IPPrefix to be its own parent

# 
## 
### Expected Behavior

This should not be possible

# 
## 
### Steps to Reproduce

1. Create a prefix `100.64.37.0/24` on the default branch

2. Create a prefix `100.64.0.0/16` on the default branch
3. Create and address `100.64.37.1/24` on the default branch
4. Create and address `100.64.37.2/24` on the default branch
5. Create and address `100.64.37.3/24` on the default branch
6. Create a new branch, let's say `branch2`
7. Edit the prefix you created in step 1 (`100.64.37.0/24`) and change its prefix to `100.64.37.0/25`
8. You will now not be able to see the prefix in the UI
9. This cypher query will show the self-parenting prefix

{noformat}MATCH p = (n:Node)-[r1:IS_RELATED]->(:Relationship)-[r2:IS_RELATED {status: r1.status, branch:r1.branch}]->(n)
RETURN p{noformat}

# 
## 
### Additional Information

This bug can cause diff generation to fail b/c the diff logic expects a Relationship vertex to always be linked to at least two Node vertices.

This will require a migration to clean up illegal self-parenting IP prefixes

ajtmccarty:
failing test for this issue exists on this branch [ip-self-parent-IFC-1712](https://github.com/opsmill/infrahub/compare/ip-self-parent-IFC-1712)
the problem is somewhere in how the `IPPrefixReconcileQuery` determines the calculated/new parent (`IPPrefixReconcileQuery.get_calculated_parent_uuid`)
the fix should include a migration to clean up any instances of this bad data, probably by identifying any IP prefix self-parents and replacing the illegal edges with the correct ones

ajtmccarty:
fixed in #7105 and #7164

PR #7105, opened 2025-08-29
fix branch-support bugs in IP reconciliation

ajtmccarty:
part of #6934 
IFC-1712

there were a few bugs in the IP reconciliation cypher query that could cause prefixes or addresses to have incorrect parents/children if their address/prefix/namespace/parent/children were updated on a branch
good news is that the bug(s) were not an issue when running on the default branch AND we always run IP reconciliation for any updated IP nodes when merging a branch

the bugs were that we were only checking if edges were active in a few places instead of selecting only the latest active value for a given attribute or relationship. for example, if I have a prefix `10.10.0.0/16` on main and I create a branch and update this prefix to `10.10.0.0/15` on the branch, then the IP reconciliation logic _could_ still consider this node as having the value `10.10.0.0/16` in certain circumstances.

The fixes are to update the query to do an initial wider filter to get all the possible correct nodes for each stage and then filter them to make sure that their latest value and namespace are the correct ones.

TODO:

will need a migration in a follow-up PR to identify prefix/address nodes that have been updated on a branch that has not been merged back into the default branch and reconcile them on the branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect parent/child assignment during branched IP prefix/address updates; reconciliation now respects branch-specific state and active relationships.

* **Improvements**
  * More robust and modular IP lookup, namespace selection, and parent/child reconciliation with unified handling for prefixes and addresses and stricter status/branch filtering.

* **Tests**
  * Added branch-focused reconciliation tests, migration-aware reconciliation test, and a missing-prefix error test.

* **Documentation**
  * Changelog updated to reflect the IP reconciliation fix.

### DIFF
--- a/backend/infrahub/core/query/ipam.py
+++ b/backend/infrahub/core/query/ipam.py
@@ -673,8 +673,6 @@ class IPPrefixReconcileQuery(Query):
             "ip_prefix_kind": InfrahubKind.IPPREFIX,
             "ip_address_kind": InfrahubKind.IPADDRESS,
             "branch_filter": branch_filter,
-            "ip_prefix_kind": InfrahubKind.IPPREFIX,
-            "ip_address_kind": InfrahubKind.IPADDRESS,
             "ip_prefix_attribute_kind": PREFIX_ATTRIBUTE_LABEL,
             "ip_address_attribute_kind": ADDRESS_ATTRIBUTE_LABEL,
         }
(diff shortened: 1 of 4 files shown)


ajtmccarty:
setting to draft while I investigate test failures

gmazoyer (approved):

ogenstad (approved):
FYI the failure for Cloudflare in the CI is fine to ignore that has been fixed in `stable` already.

PR #7115, opened 2025-09-01
Fix branched_from property during rebase operations

### LINKED ISSUE #7113: bug: rebasing a branch does not update its timestamp in the registry for the API server(s)
h3\. Component

[| #component]

API Server / GraphQL

h3\. Infrahub version

[| #infrahub-version]

1.4.3

h3\. Current Behavior

[| #current-behavior]

when a branch is rebased we update its `branched_from` timestamp, which is used to determine which values on the default branch are visible on this branch, but we do now update the `branched_from` timestamp in the in-memory registry of the API server(s)

the branch rebase action runs in a prefect flow and does not affect the in-memory branches of the other prefect server(s) or API server(s)

h3\. Expected Behavior

[| #expected-behavior]

we need to update the `branched_from` time correctly for a rebased branch everywhere once the rebase is complete

h3\. Steps to Reproduce

[| #steps-to-reproduce]

h1.  add a tag on `main`, let's call it `Red`

[| #add-a-tag-on-main-lets-call-it-red]

h1.  create a branch `branch2`

[| #create-a-branch-branch2]

h1.  rebase `branch2`

[| #rebase-branch2]

h1.  view the tags on `branch2` and you will not see `Red`

[| #view-the-tags-on-branch2-and-you-will-not-see-red]

if you do one of the following, you will see `Red` appear on `branch2`

h1.  restart the API server

[| #restart-the-api-server]

h1.  make a change to the schema on `main` and rebase `branch2` again

[| #make-a-change-to-the-schema-on-main-and-rebase-branch2-again]

h3\. Additional Information

[| #additional-information]

_No response_

ogenstad:
Fixes #7113

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly updates a branch’s origin after rebasing, even when only data changes occur.
  * Ensures branch state stays accurate and up to date after rebase, reducing inconsistencies users may see in branch information.
  * Improves reliability of branch metadata shown in tools and queries post-rebase.

* **Documentation**
  * Added a changelog entry describing the fix for rebase handling to ensure accurate branch origin tracking.

### DIFF
--- a/backend/infrahub/tasks/registry.py
+++ b/backend/infrahub/tasks/registry.py
@@ -42,7 +42,7 @@ async def create_branch_registry(db: InfrahubDatabase, branch: Branch) -> None:
 
 
 async def update_branch_registry(db: InfrahubDatabase, branch: Branch) -> None:
-    """Update the registry for a branch if the schema hash has changed."""
+    """Update the registry for a branch if the schema hash has changed or the branch was rebased."""
 
     existing_branch: Branch = registry.branch[branch.name]
 
@@ -50,13 +50,23 @@ async def update_branch_registry(db: InfrahubDatabase, branch: Branch) -> None:
         log.warning("Branch schema hash is not set, cannot update branch registry")
         return
 
-    if existing_branch.schema_hash and existing_branch.schema_hash.main == branch.active_schema_hash.main:
+    if existing_branch.schema_hash.main == branch.active_schema_hash.main:
         log.debug(
-            "Branch schema hash is the same, no need to update branch registry",
+            "Branch schema hash is the same, no need to refresh the GraphQL schema within the registry",
             branch=branch.name,
             hash=existing_branch.schema_hash.main,
             worker=WORKER_IDENTITY,
         )
+        if existing_branch.branched_from != branch.branched_from:
+            # If the hash is the same but the branched_from timestamp differs it means
+            # that the branch has been rebased and these timestamps need to be refreshed
+            # in the registry even though the schema doesn't need to be reloaded.
+            log.info(
+                "Updating branched_from property in registry for rebased branch",
+                branch=branch.name,
+                worker=WORKER_IDENTITY,
+            )
+            registry.branch[branch.name] = branch
         return
 
     log.info(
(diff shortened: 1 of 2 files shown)


backend/infrahub/tasks/registry.py
```diff
@@ -57,6 +57,12 @@ async def update_branch_registry(db: InfrahubDatabase, branch: Branch) -> None:
             hash=existing_branch.schema_hash.main,
             worker=WORKER_IDENTITY,
         )
+        if existing_branch.branched_from != branch.branched_from:
```
dgarros, line 60: Would be good to display a single log message, here we'll have 2 somewhat contradicting log messages

dgarros (commented):
I think we should refactor the function to be not just about schema hash, for example the description of the function should be updated

backend/infrahub/tasks/registry.py
```diff
@@ -42,21 +42,31 @@ async def create_branch_registry(db: InfrahubDatabase, branch: Branch) -> None:
...
+    """Update the registry for a branch if the schema hash has changed or the branch was rebased."""
 
     existing_branch: Branch = registry.branch[branch.name]
 
     if not existing_branch.schema_hash:
         log.warning("Branch schema hash is not set, cannot update branch registry")
         return
 
-    if existing_branch.schema_hash and existing_branch.schema_hash.main == branch.active_schema_hash.main:
+    if existing_branch.schema_hash.main == branch.active_schema_hash.main:
```
ogenstad, line 53: I removed the first part of the if statement here for the conditional `existing_branch.schema_hash` as we already have a check for this just above and return early if the schema hash is not set.

backend/infrahub/tasks/registry.py
ogenstad: I reworded the first one but otherwise kept the two of them as the first one is logged at the debug level.

ogenstad:
> I think we should refactor the function to be not just about schema hash, for example the description of the function should be updated

Yes, good point. I've updated the doc string.

dgarros (approved):

PR #7125, opened 2025-09-02
Refactor 'infrahub shell'

ogenstad:
This PR refactors the `infrahub shell` command which currently isn't working at all. The command is meant to be used for local debugging or troubleshooting and provides access to internal components within Infrahub.

It uses IPython to you have access to asyncio directly from the shell without doing anything special. I.e. you can just run async code as if it was the sync version.

Note that this shell uses IPython which is a develop dependency of Infrahub and as such won't be included in a purely production build.

In order to test it you run `infrahub shell` it's using the same API as the other command line tools such as `infrahub server start`.

A few objects are made available from the start and more can be added.

Just after running the shell you can for example paste the code:

```python
>>> tags = await NodeManager.query(schema='BuiltinTag', db=db)
```

In order to query for all of the tags using the NodeManager class.

## Summary by CodeRabbit

- New Features
  - Interactive shell now fully auto-initializes the runtime, preloads services and data access, and exposes helpful helpers for immediate interaction.
  - Improved user experience with a welcome message and a consistent Neutral theme.

- Refactor
  - Simplified, in-process startup sequence for faster, more reliable launches.

- Changes
  - Shell command is now parameterless and auto-loads configuration (no config-file argument).

### DIFF
--- a/backend/infrahub/cli/__init__.py
+++ b/backend/infrahub/cli/__init__.py
@@ -1,5 +1,3 @@
-from asyncio import run as aiorun
-
 import typer
 from infrahub_sdk.async_typer import AsyncTyper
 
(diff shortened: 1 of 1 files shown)


backend/infrahub/cli/__init__.py
```diff
@@ -43,18 +41,90 @@ async def _init_shell(config_file: str) -> None:
...
+        get_database,
+        get_workflow,
+        set_component_type,
+    )
+
+    async def initialize_service() -> InfrahubServices:
+        config.load_and_exit()
+        client = InfrahubClient()
+
+        database = InfrahubDatabase(mode=InfrahubDatabaseMode.DRIVER, driver=await get_db())
```
ajtmccarty: I think code rabbit is right about removing this line, looks like it is overwritten on 77 below

ajtmccarty (approved):

backend/infrahub/cli/__init__.py
ogenstad: Ah yes, removed.

PR #7150, opened 2025-09-04
Fixed filtering options in object table schema selector

bilalabbad:
## Summary by CodeRabbit

* **New Features**
  * Added client-side filtering to the object schema selector dropdown, enabling type-to-filter and keyword-based matching for faster option discovery.

* **Bug Fixes**
  * Resolved an issue where the kind selector did not filter its options correctly, ensuring accurate results when narrowing down object kinds.

* **Tests**
  * Expanded end-to-end coverage to validate substring-based filtering behavior in the schema selector.


minitriga (approved):
👍

PR #7155, opened 2025-09-05
IFC-1796 Add method to get changes between 2 commits

gmazoyer:
This method takes a first commit parameter and an optional second commit parameter, if not set the repo's head will be used instead. It then computes a diff and returns a Pydantic model with list of filenames for fiels that have been changed, added or deleted.

## Summary by CodeRabbit

- New Features
  - Added the ability to list changed files between two commits, grouped as added, copied, deleted, renamed, modified, and type-changed.
  - Returns structured results and reports clear errors when a requested commit is not found.

- Tests
  - Added integration tests and fixtures to validate changed-file detection across multiple commits and error handling for unknown commits.

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -977,13 +977,11 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         """Return the changes between two commits in this repo."""
         files: dict[str, list[str | tuple[str, str]]] = {
             "added": [],
-            "modified": [],
+            "copied": [],
             "deleted": [],
             "renamed": [],
-            "copied": [],
+            "modified": [],
             "type_changed": [],
-            "unmerged": [],
-            "unknown": [],
         }
 
         repo = self.get_git_repo_main()
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, NoReturn
 from uuid import UUID  # noqa: TC003
 
 import git
-from git import Blob, Repo
+from git import BadName, Blob, Repo
 from git.exc import GitCommandError, InvalidGitRepositoryError
 from git.refs.remote import RemoteReference
 from infrahub_sdk import InfrahubClient  # noqa: TC002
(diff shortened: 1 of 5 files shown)


backend/infrahub/git/base.py
```diff
@@ -970,3 +970,37 @@ def _get_mapped_target_branch(self, branch_name: str) -> str:
         if branch_name == self.default_branch and branch_name != registry.default_branch:
             return registry.default_branch
         return branch_name
+
+    def get_changed_files(
+        self, first_commit: str, second_commit: str | None = None
+    ) -> dict[str, list[str | tuple[str, str]]]:
```
ogenstad: I'd suggest instead returning a dataclass with named properties for "added", "copied" etc instead since it's so much easier to interact with compared to a dictionary where we aren't completely sure about the name of the keys if we don't go through the `get_changed_files()` function.

gmazoyer: I used a pydantic class as this is what we use for `RepoFileInformation`.

ogenstad (approved):
LGTM, I'm just surprised around the file paths with `backend/tests/fixtures/repos/changed-files/initial__main/commit1/test.gql` 

I thought we had a system where follow up commits would be within their own directory of `backend/tests/fixtures/repos/changed-files/` such as `backend/tests/fixtures/repos/changed-files/pr_001`. 

Might just be that I'm unfamiliar with the `MultipleStagesFileRepo` class.

backend/tests/integration/git/test_repository.py
```diff
@@ -123,3 +123,72 @@ async def test_repository_operational_status(
...
+            repository_id=repository.id,
+            name=repository.name.value,
+            repository_kind=InfrahubKind.REPOSITORY,
+        )
+
+        # Have commits from oldest to youngest
+        commits = list(reversed(list(infrahub_repo.get_git_repo_main().iter_commits())))
+        assert len(commits) == 3
+
+        diff_1_to_3 = infrahub_repo.get_changed_files(first_commit=commits[0].hexsha, second_commit=commits[1].hexsha)
```
ajtmccarty: I think this is `diff_1_to_2`

ajtmccarty (approved):

PR #7163, opened 2025-09-05
updates and tests for Relationship queries to not return deleted rels

ajtmccarty:
found a couple of issues with a `Relationship`-related query while investigating IFC-1836 and looking at what other queries used `Branch.get_query_filter_relationships`

`RelationshipGetByIdentifierQuery` and `RelationshipGetQuery` both use that method. `RelationshipGetByIdentifierQuery` was correctly accounting for deleted edges, but `RelationshipGetQuery` could return deleted relationships. Fortunately, this query is only used when deleting things, so the worst case would be that something gets double-deleted, which is usually not a big deal

The updates to the query is to make sure we only get the latest Relationship path (ie `(Node)-[:IS_RELATED]-(:Relationship)-[:IS_RELATED]-(:Node)`) and then account for whether that path is deleted or not

## Summary by CodeRabbit

- Bug Fixes
  - Fix handling of deleted relationships to avoid duplicate deletion paths and ensure accurate active status; deletions are now idempotent.

- Documentation
  - Added changelog entry describing the corrected query behavior for deleted relationships.

- Tests
  - Added tests for single active relationship retrieval, status after deletion, idempotent deletes, and additional query coverage when updating ownership and after node deletions.

### DIFF
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -871,6 +871,9 @@ class RelationshipGetQuery(RelationshipQuery):
         query = """
         MATCH (s)%s(rl:Relationship { name: $name })%s(d)
         WHERE %s
+        ORDER BY r1.branch_level DESC, r1.from DESC, r1.status ASC, r2.branch_level DESC, r2.from DESC, r2.status ASC
+        WITH *, r1.status = "active" AND r2.status = "active" AS is_active
+        LIMIT 1
         """ % (
             r1,
             r2,
@@ -880,7 +883,7 @@ class RelationshipGetQuery(RelationshipQuery):
         self.params["at"] = self.at.to_string()
 
         self.add_to_query(query)
-        self.return_labels = ["s", "d", "rl", "r1", "r2"]
+        self.return_labels = ["s", "d", "rl", "r2", "is_active"]
 
 
 class RelationshipGetByIdentifierQuery(Query):
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -885,6 +885,19 @@ class RelationshipGetQuery(RelationshipQuery):
         self.add_to_query(query)
         self.return_labels = ["s", "d", "rl", "r2", "is_active"]
 
+    def is_already_deleted(self) -> bool:
+        result = self.get_result()
+        if not result:
+            return False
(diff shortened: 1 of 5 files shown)


ajtmccarty:
can be merged if approved

backend/infrahub/core/relationship/model.py
```diff
@@ -410,6 +410,10 @@ async def delete(self, db: InfrahubDatabase, at: Timestamp | None = None) -> Non
                 f"Unable to find the relationship to delete. id: {self.id}, source: {node.id}, destination: {peer.id}"
             )
 
+        # this Relationship has already been deleted
+        if result.get("is_active") is False:
```
LucasG0: Would this be cleaner that `RelationshipGetQuery` has a method `get_rel` which returns the relationship, which we could use instead of having a mix of `get_result` / `get` / `result.get_rels`?

LucasG0 (approved):

backend/infrahub/core/relationship/model.py
ajtmccarty: it is kind of a mess in here
I tried to clean it up a little with my last commit

backend/infrahub/core/query/relationship.py
```diff
@@ -880,7 +883,20 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
+        if not result:
+            return False
+        return result.get("is_active") is False
+
+    def get_relationships_ids_for_branch(self, branch_name: str) -> list[str] | None:
+        result = self.get_result()
+        if not result:
+            return None
+
+        return [rel.element_id for rel in result.get_rels() if rel.get("branch") == branch_name]
```
LucasG0, line 899: From the `LIMIT 1` introduced above, this function should not return more than 1 rel, is that right? Ifso, should this return an optional `str`?

ajtmccarty: this will return multiple relationships if `r1` and `r2` are non-null, which will usually (or maybe always) be the case b/c `get_rels` returns all edges in a result row
I think that it would be good to rewrite most of the queries in this module so that they are easier to read and use, but I don't want to get sucked into that right now b/c I think it will take a decent amount of time

LucasG0 (approved):

PR #7175, opened 2025-09-08
IFC-1801 Add mutation to trigger computed attribute re-computation

gmazoyer:
## Summary by CodeRabbit

- New Features
  - Added a GraphQL mutation to recompute computed attributes by kind or node IDs, supporting both Jinja- and Python-based computed attributes for targeted or bulk updates.

- Bug Fixes
  - Corrected a typo in the error message shown when a node lacks the specified attribute.

- Documentation
  - Changelog updated to describe the new recomputation mutation and its usage.

- Tests
  - Functional tests refactored to invoke the new mutation instead of triggering workflows directly.

### DIFF
--- a/backend/infrahub/graphql/mutations/computed_attribute.py
+++ b/backend/infrahub/graphql/mutations/computed_attribute.py
@@ -177,7 +177,8 @@ class RecomputeComputedAttribute(Mutation):
                 message=f"The indicated node does not use a computed attribute for the specified attribute '{data.attribute}'"
             )
 
-        workflow = COMPUTED_ATTRIBUTE_PROCESS_JINJA2 if data.node_ids else TRIGGER_UPDATE_JINJA_COMPUTED_ATTRIBUTES
+        recalculate_single_workflow = COMPUTED_ATTRIBUTE_PROCESS_JINJA2
+        recalculate_all_workflow = TRIGGER_UPDATE_JINJA_COMPUTED_ATTRIBUTES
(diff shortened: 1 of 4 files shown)


backend/infrahub/graphql/mutations/computed_attribute.py
```diff
@@ -117,3 +125,101 @@ async def mutate(
...
+        if not attribute:
+            raise ProcessingError(
+                message=f"The indicated node does not have the specified attribute '{data.attribute}'"
+            )
+        if not attribute.computed_attribute:
+            raise ProcessingError(
+                message=f"The indicated node does not use a computed attribute for the specified attribute '{data.attribute}'"
+            )
+
+        workflow = COMPUTED_ATTRIBUTE_PROCESS_JINJA2 if data.node_ids else TRIGGER_UPDATE_JINJA_COMPUTED_ATTRIBUTES
```
ogenstad: I think this `workflow` variable would be slightly easier to read if we instead had two of them:

```python
recalculate_single_workflow = COMPUTED_ATTRIBUTE_PROCESS_JINJA2
recalculate_all_workflow = TRIGGER_UPDATE_JINJA_COMPUTED_ATTRIBUTES
```
Then we reuse the `recalculate_single_workflow` and `recalculate_all_workflow` names for the transform based one. Feel free to ignore this comment if you don't agree.

ogenstad (approved):

PR #7178, opened 2025-09-08
Refactor generator actions to execute generator directly

dgarros:
This PR refactor the flow `run_generator_group_event` and `run_generator` to execute the generators directly instead of calling the `CoreGeneratorDefinitionRun` mutation.

In the current implementation, the mutation is not async and as a result the execution could fail if the execution time is bigger than the timeout of the GraphQL API

Also in order to improve the performance of this workflow, the implementation is not using the SDK and instead it's using a single GraphQL query to pull all the information necessary from the backend.

## Summary by CodeRabbit

* **New Features**
  * Generator runs now execute per-target: system fetches generator and target metadata and submits a dedicated run per target.

* **Improvements**
  * Generator definitions surface parameters consistently and validate target membership before execution.
  * Some generator-related workflows reclassified to internal handling.

* **Bug Fixes**
  * Clear errors when a generator definition is missing or requested targets don’t apply.

* **Refactor**
  * Switched to a workflow-driven, per-target orchestration for generator execution.

* **Tests**
  * Added functional and integration tests for generator and action-triggered runs.

* **Chores**
  * Added housekeeping changelog entry.

### DIFF
--- a/backend/infrahub/actions/tasks.py
+++ b/backend/infrahub/actions/tasks.py
@@ -1,16 +1,107 @@
 from __future__ import annotations
 
-from infrahub_sdk.graphql import Mutation
+from collections import defaultdict
+from typing import TYPE_CHECKING, Any
+
+from infrahub_sdk.graphql import Mutation, Query
+from infrahub_sdk.types import Order
 from prefect import flow
 
 from infrahub.context import InfrahubContext  # noqa: TC001  needed for prefect flow
+from infrahub.core.constants import InfrahubKind
+from infrahub.generators.models import (
+    GeneratorDefinitionModel,
+    RequestGeneratorRun,
+)
 from infrahub.services import InfrahubServices  # noqa: TC001  needed for prefect flow
 from infrahub.trigger.models import TriggerType
 from infrahub.trigger.setup import setup_triggers_specific
+from infrahub.workers.dependencies import get_client, get_workflow
+from infrahub.workflows.catalogue import REQUEST_GENERATOR_RUN
 from infrahub.workflows.utils import add_tags
 
 from .gather import gather_trigger_action_rules
 from .models import EventGroupMember  # noqa: TC001  needed for prefect flow
+
+if TYPE_CHECKING:
+    from infrahub_sdk.client import InfrahubClient
+    from infrahub_sdk.node import InfrahubNode
+
+
+def get_generator_run_query(definition_id: str, target_ids: list[str]) -> Query:
+    return Query(
+        name=InfrahubKind.GENERATORDEFINITION,
+        query={
+            InfrahubKind.GENERATORDEFINITION: {
+                "@filters": {
+                    "ids": [definition_id],
+                },
+                "edges": {
+                    "node": {
+                        "id": None,
+                        "name": {
+                            "value": None,
+                        },
+                        "class_name": {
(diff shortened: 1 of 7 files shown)


backend/infrahub/actions/tasks.py
```diff
@@ -104,16 +205,96 @@ async def configure_action_rules(
...
+async def _get_targets(
+    branch_name: str,
+    targets: list[dict[str, Any]],
+    client: InfrahubClient,
+) -> dict[str, dict[str, InfrahubNode]]:
+    """Get the targets per kind in order to extract the variables.
+
+    NOTE: Currently this will only work for attributes because we are not fetching the relationships out of the box.
+    To avoid performance issues, we need to analyse the generator and fetch the right relationships
+    """
```
ogenstad: I'm not quite sure I understand this part. This is to find the input parameter to the query? I.e for this generator query https://github.com/opsmill/infrahub-demo-edge/blob/023f438580ef6c7ec6773eba0e4264188e66eefa/generators/circuit_endpoints.gql#L2 we want to validate that we specifically query the node for the field that would contain `circuit_id__value` is that correct or are we talking about something else?

backend/infrahub/actions/tasks.py
```diff
@@ -104,16 +205,96 @@ async def configure_action_rules(
...
+
+    targets_per_kind: dict[str, dict[str, InfrahubNode]] = defaultdict(dict)
+
+    for target in targets:
+        targets_per_kind[target["node"]["__typename"]][target["node"]["id"]] = None
```
ogenstad, line 222: I think this highlights the need to move forward with those static queries that generate bindings we talked about last week. The problem for me here (which I think we also still have in a number of other places throughout the code is that we assume that `targets` is `list[dict[str, Any]]` but within this part of the code we make the assumption that the dict will always have a key for "node" which contains sub-keys for "__typename" and "id", which would be hard to spot if anything in the source changes.

I think this is fine for now but something that we should go back to and refactor later with those generated bindings.

backend/infrahub/actions/tasks.py
```diff
@@ -104,16 +205,96 @@ async def configure_action_rules(
...
+
+    Raises:
+        ValueError: If the generator definition is not found or none of the requested
+            targets are members of the target group.
+    """
+    response = await client.execute_graphql(
+        query=get_generator_run_query(definition_id=generator_definition_id, target_ids=node_ids).render(),
+        branch_name=branch_name,
     )
+    data = response[InfrahubKind.GENERATORDEFINITION]["edges"][0]["node"]
```
ogenstad, line 264: While it's unlikely to happen if the generator was deleted after triggering it this could cause a confusing error. Like below we could raise an error if we didn't have anything within `response[InfrahubKind.GENERATORDEFINITION]["edges"]`

ogenstad (approved):

backend/infrahub/actions/tasks.py
dgarros: This was a misunderstanding on my side, I removed the comment

backend/infrahub/actions/tasks.py
dgarros: Good point, I added a check to ensure that edges is not empty

PR #7179, opened 2025-09-08
Add option Any to Trigger Attribute Match

dgarros:
Add an option for `CoreNodeTriggerAttributeMatch` to match on Any attribute values.

## Summary by CodeRabbit

* **New Features**
  * Added an “Any” option to value-matching dropdowns, allowing triggers to match against any attribute value where value-match criteria apply.

* **Documentation**
  * Added a changelog entry announcing the new “Any” matching option.

* **Tests**
  * Expanded unit tests to validate attribute-based triggers across different value-match modes, branch-scoping (including not-main branches), multiple attribute matches, and trigger activation states.

### DIFF
--- a/backend/infrahub/actions/constants.py
+++ b/backend/infrahub/actions/constants.py
@@ -118,6 +118,12 @@ class ValueMatch(DropdownEnum):
         description="Match against both the current and previous values",
         color="#5ac8fa",
     )
+    ANY = DropdownChoice(
+        name="any",
+        label="Any",
+        description="Match against any value",
+        color="#276cc2",
+    )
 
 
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
LGTM. Since just by looking at this new option it can be hard to figure out what it's supposed to actually do I think it would be nice to add an extra condition to this in `test_gather_trigger_gather_trigger_action_rules_node_attribute()` to validate that the `match_related` are populated correctly if we change things in the future. It would also be fine with an issue that points to this comment to add this check to the test in a follow up PR.

dgarros:
thanks @ogenstad I added an extra condition in `test_gather_trigger_gather_trigger_action_rules_node_attribute()`

PR #7197, opened 2025-09-10
IFC-1812 Refactor HFID & display_labels

gmazoyer:
This PR brings several changes to how `display_label` and `human_friendly_id` for nodes are stored and retrieved. While it tries to change the least amount of code to keep code compatible as much as possible, it now stores the `display_label` and `human_friendly_id` of nodes into the database to be able to retrieve them when looking up nodes instead of computing them every times. It also introduces events to help in `display_label` computation.

This PR provides implementations for IFC-1820, IFC-1821, IFC-1822, IFC-1823, IFC-1824, IFC-1825 and IFC-1826.

@coderabbitai ignore


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -751,6 +753,26 @@ def validate_uniqueness_constraints(self) -> None:
                         element_name=element_name,
                     )
 
+    def validate_display_label(self) -> None:
+        for name in self.all_names:
+            node_schema = self.get(name=name, duplicate=False)
+
+            if node_schema.display_label and node_schema.display_labels:
+                raise ValidationError(f"{node_schema.kind}: cannot defined both `display_label` and `display_labels`")
```
ogenstad: Small typo here "defined" -> "define"

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -751,6 +753,26 @@ def validate_uniqueness_constraints(self) -> None:
                         element_name=element_name,
                     )
 
+    def validate_display_label(self) -> None:
+        for name in self.all_names:
+            node_schema = self.get(name=name, duplicate=False)
+
+            if node_schema.display_label and node_schema.display_labels:
```
ogenstad: I think for clarity it should probably be:

```diff
-             if node_schema.display_label and node_schema.display_labels:
+             if node_schema.display_label and node_schema.display_labels is not None:
```

I.e. to ensure that we don't have an empty array.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1136,6 +1158,47 @@ def validate_computed_attributes(self) -> None:
...
 
+    def _validate_display_label(self, node: MainSchemaTypes) -> None:
+        if not node.display_label:
+            return
+
+        allowed_path_types = (
+            SchemaElementPathType.ATTR_WITH_PROP
+            | SchemaElementPathType.REL_ONE_MANDATORY_ATTR_WITH_PROP
+            | SchemaElementPathType.REL_ONE_ATTR_WITH_PROP
+        )
```
ogenstad: It could be that we want additional allowed options here later.

backend/infrahub/core/utils.py
```diff
@@ -214,6 +214,14 @@ def build_regex_attr(value: str | int | bool) -> str:
     raise ValueError("value was neither a string, an int or a boolean")
 
 
+def is_jinja2_template(value: str) -> bool:
+    """Tell if a string is supposed to be a Jinja2 template.
+
+    This is done by looking at Jinja2 control structures such as `{{ variable }}` or `{% if ... %}`.
+    """
+    return bool(re.search(r"{{.+}}", value) or re.search(r"{%.+%}", value))
```
ogenstad: As this will be used for the display_label where anything outside of Jinja2 would be single fields such as `name__value` I think we can change this slightly to be any string that includes `{` or `}` that way we'd also find invalid templates such as `{ name__value|upper }}` which should be simpler to validate when we try to process this as a Jinja2 template.

backend/tests/unit/core/schema_manager/test_manager_schema.py
```diff
@@ -1332,6 +1353,47 @@ async def test_validate_display_labels_error(schema_all_in_one, display_labels,
         schema.validate_display_labels()
 
 
+@pytest.mark.parametrize(
+    "display_label,expected_error",
+    [
+        ("{{ mybool }}", ""),
```
ajtmccarty: is there no error message for this one?

ajtmccarty (approved):

PR #7200, opened 2025-09-10
feat: allow group mapping for google sso

fatih-acar:
Google does not include groups within the tokens.
We can only fetch groups through their APIs.
One available API is the Admin SDK/Directory API which requires admin permissions, which would in turn require extra configuration for Infrahub.
Another API is the Cloud Identity API which provides endpoints to fetch group memberships using the user token.

## Summary by CodeRabbit

* **New Features**
  * Improved SSO group resolution: when an identity provider omits groups, the system can fetch Google group memberships to better map user groups at login.
  * Falls back to configured group defaults if no groups are found.

* **Configuration**
  * New per-provider settings to enable Cloud Identity group fetching and to customize the Cloud Identity endpoint.
  * Disabled by default; requires appropriate Cloud Identity scope.

### DIFF
--- a/backend/infrahub/api/oidc.py
+++ b/backend/infrahub/api/oidc.py
@@ -13,8 +13,7 @@ from pydantic import BaseModel, HttpUrl
 
 from infrahub import config, models
 from infrahub.api.dependencies import get_db
-from infrahub.auth import signin_sso_account
-from infrahub.config import SecurityOIDCGoogle
+from infrahub.auth import get_groups_from_provider, signin_sso_account
 from infrahub.exceptions import GatewayError, ProcessingError
 from infrahub.log import get_logger
 from infrahub.message_bus.types import KVTTL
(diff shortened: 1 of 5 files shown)


backend/infrahub/api/oidc.py
```diff
@@ -152,6 +153,16 @@ async def token(
...
+    if not sso_groups and isinstance(provider, SecurityOIDCGoogle):
+        # Poor man's workaround to fetch user groups from Google
+        if provider.fetch_groups:
+            groups_response = await service.http.get(
+                f"{provider.cloudidentity_url}?query=member_key_id == '{user_info['email']}'",
+                headers=headers,
+            )
+            group_memberships = groups_response.json()
+            if "memberships" in group_memberships:
+                sso_groups = [membership["groupKey"]["id"] for membership in group_memberships["memberships"]]
```
ogenstad: I think we should have a function that takes a provider where we check if it's either a `SecurityOAuth2Google` or a `SecurityOIDCGoogle` and if that's the case run this code and return the sso_groups.

Could be something like:

```python
    sso_groups = user_info.get("groups") or await _get_id_token_groups(
        oidc_config=oidc_config, service=service, payload=payload, client_id=provider.client_id
    ) or await this_other_lookup_function(userinfo, provider)
```

That way we don't have to duplicate this code that is pretty much the same thing. (The same thing could be said for these two methods in general i.e. oauth and oidc. But I don't think we should add more variations to them.

ogenstad (approved):

PR #7213, opened 2025-09-12
Add form from selected kind for object convert type

pa-lem:
* start to fetch mappings according to source and target kinds
* display form from the target kind selection
* provide object details to component

image

## Summary by CodeRabbit

- New Features
  - Introduced an object type conversion flow with a dynamic form that adapts to the selected target type.
  - Added a schema picker to choose the destination type.
  - Provides loading and error states during field-mapping retrieval.

- Improvements
  - Clearer destination labeling using the selected schema’s name.
  - Placeholder view with icon and guidance when no target type is selected.
  - More responsive combobox behavior with controlled open/close for smoother selection.


frontend/app/src/entities/nodes/object/api/get-fields-mapping-type-conversion-from-api.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const query = gql`
+query($sourceKind: String, $targetKind: String) {
+  FieldsMappingTypeConversion(source_kind: $sourceKind, target_kind: $targetKind) {
+    mapping
+  }
+}
+`;
+
+export function getFieldsMappingFromApi({
```
bilalabbad: name is too generic. Make it more specific so we know for which mapping

frontend/app/src/entities/nodes/object/domain/get-fields-mapping-type-conversion.query.ts
```diff
@@ -0,0 +1,56 @@
...
+interface FieldsMappingOptionsParams extends ContextParams, FieldsMappingProps {}
+
+export function getFieldsMappingOptions({
+  sourceKind,
+  targetKind,
+  branchName,
+  atDate,
+}: FieldsMappingOptionsParams) {
+  return queryOptions({
+    queryKey: [branchName, atDate, "fields-mapping-type-conversion", sourceKind, targetKind],
```
bilalabbad: use queryKey factory instead of hardcoding it

frontend/app/src/entities/nodes/object/domain/get-fields-mapping-type-conversion.query.ts
```diff
@@ -0,0 +1,56 @@
...
+    queryFn: () => {
+      return getFieldsMappingFromApi({
+        sourceKind,
+        targetKind,
+        branchName,
+        atDate,
+      });
+    },
+    enabled: Boolean(sourceKind && targetKind),
```
bilalabbad: this condition is not needed if sourceKind and targetKind and both defined

frontend/app/src/entities/nodes/object/domain/get-fields-mapping-type-conversion.query.ts
```diff
@@ -0,0 +1,56 @@
...
+  const { data, ...props } = useQuery(
+    getFieldsMappingOptions({
+      sourceKind,
+      targetKind,
+      branchName: currentBranch.name,
+      atDate: timeMachineDate,
+    })
+  );
+
+  const result = data?.data?.FieldsMappingTypeConversion?.mapping ?? {};
```
bilalabbad: Here it's mixing concern, this query does not respect the standard way of interacting with api.

this should always be:
```
[XFromApi] -> [MyDomainSpecificUseCase] -> [MyDomainSpecificUseCase.query]
```
- fromApi: contains only the how to call API logic
- Domain layer: adapts API data into app-compatible structures + business logic
- query: is just the binding layer that wires a domain use case into React component

frontend/app/src/entities/nodes/object/domain/get-fields-mapping-type-conversion.query.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return {
+    data: result,
+    ...props,
+  };
```
bilalabbad: + this destructing will cause re-render at each props changes. all props are no longer stable

frontend/app/src/shared/components/form/convert-form.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import ErrorScreen from "../errors/error-screen";
+import { LoadingIndicator } from "../loading/loading-indicator";
+import DynamicForm from "./dynamic-form";
+import { getFormFieldsFromSchema } from "./utils/getFormFieldsFromSchema";
+
+export type ConvertFormProps = {
+  objectDetailsData: NodeObject;
+  sourceKind: string;
+  targetKind: string;
```
bilalabbad: passing schema directly will make the component simpler

bilalabbad (changes requested):

bilalabbad (approved):

PR #7234, opened 2025-09-18
IFC-499: Clean up dead locks at set intervals

solababs:
## Description
Clean up dead locks at set intervals determined by `clean_up_deadlocks_interval_mins` setting default to `15`
Deadlocks are checked in the cache via a `lock.` prefix then the worker is checked if it is still alive before releasing the lock

[x] - Run test
[x] - Add/Update test

## Summary by CodeRabbit

- New Features
  - Automatic cleanup of stale locks owned by inactive workers on a scheduled interval.
  - Config option to adjust cleanup interval (default 15 minutes).

- Documentation
  - Added reference docs for the new cleanup configuration option.

- Refactor
  - Centralized and standardized lock naming/handling for improved reliability.

- Tests
  - Added unit and end-to-end tests for lock naming and the cleanup behavior.


backend/infrahub/workflows/catalogue.py
```diff
@@ -530,6 +532,14 @@
     function="validate_schema_number_pools",
 )
 
+CLEAN_UP_DEADLOCKS = WorkflowDefinition(
+    name="clean_up_deadlocks",
+    type=WorkflowType.CORE,
```
dgarros: I think this should an INTERNAL workflow to avoid generating too many tasks visible to the users

backend/infrahub/config.py
```diff
@@ -199,6 +199,10 @@ class MainSettings(BaseSettings):
         description="Enable strict schema validation. When set to `False`, "
         "`human_friendly_id` schema fields should not necessarily target a unique combination of peer attributes.",
     )
+    clean_up_deadlocks_interval_mins: int = Field(
```
dgarros: I think we could move this settings to CacheSettings

backend/infrahub/lock.py
```diff
@@ -179,6 +179,49 @@ async def locked(self) -> bool:
         return self.local.locked()
 
 
+class InfahubLockNameGenerator:
```
fatih-acar:
```suggestion
class InfrahubLockNameGenerator:
```

Small typo

backend/infrahub/workflows/catalogue.py
```diff
@@ -530,6 +530,14 @@
     function="validate_schema_number_pools",
 )
 
+CLEAN_UP_DEADLOCKS = WorkflowDefinition(
+    name="clean_up_deadlocks",
+    type=WorkflowType.INTERNAL,
+    cron="* * * * *",
```
fatih-acar, line 539: For such intervals I think you can make use of concurrency limits, as seen here: https://github.com/opsmill/infrahub/pull/7244

backend/infrahub/lock.py
```diff
@@ -179,6 +179,49 @@ async def locked(self) -> bool:
         return self.local.locked()
 
 
+class InfrahubLockNameGenerator:
+    local = "local"
+    _global = "global"
+
+    @staticmethod
```
ajtmccarty: I think it would be better if these were class or instance methods so that you can use `self` or `cls` below instead of `InfrahubLockNameGenerator`

backend/infrahub/lock.py
```diff
@@ -179,6 +179,49 @@ async def locked(self) -> bool:
         return self.local.locked()
 
 
+class InfrahubLockNameGenerator:
```
ajtmccarty: moving this logic to its own class is a great change 👍
I think you can probably just call this `LockNameGenerator`. We've got some older class that start with `Infrahub`, but it is usually not necessary

You could even pass this into `InfrahubLockRegistry` as a dependency and update `initialize_lock()` in `lock.py`, which seems to be the only place we instantiate `InfrahubLockRegistry`

```python
class InfrahubLockRegistry:
    def __init__(
        ...
        name_generator: InfrahubLockNameGenerator,
    ):
        ...
        self.name_generator = name_generator
```

backend/infrahub/pools/tasks.py
```diff
@@ -124,3 +126,32 @@ async def _create_number_pool(
...
+
+    for key, value in zip(keys, values, strict=False):
+        if not key or not value:
+            continue
+
+        timestamp, worker_id = value.split("::", 1)
+        if not workers_active_map.get(worker_id) and Timestamp() > Timestamp(timestamp).add(
+            minutes=config.SETTINGS.cache.clean_up_deadlocks_interval_mins
+        ):
+            name, namespace, local = InfrahubLockRegistry.unpack_name(name=key.replace("lock.", ""))
```
ajtmccarty: I think you can call `InfrahubLockNameGenerator.unpack_name` here instead of going through `InfrahubLockRegistry`, which maybe doesn't need the `unpack_name` method

backend/infrahub/lock.py
```diff
@@ -179,9 +179,54 @@ async def locked(self) -> bool:
...
+        elif local is False:
+            new_name = f"{self._global}."
+
+        if namespace is not None:
+            new_name += f"{namespace}."
+        new_name += name
+
+        return new_name
+
+    def unpack_name(self, name: str) -> tuple[str, str | None, bool | None]:
```
ogenstad, line 203: It looks like this function will always return a bool for the `local` variable and never a `None`. I'm guessing the `None` comes from the fact that we're assigning this value to the variable to avoid potential unbound errors. Could be a bit cleaner to ensure that we always return a boolean to clarify the expected response from this method.

backend/infrahub/lock.py
```diff
@@ -201,43 +246,36 @@ def __init__(
...
         return self.locks[lock_name]
 
+    async def pop_lock(
+        self,
+        name: str,
+        namespace: str | None,
+        local: bool | None = None,
+    ) -> InfrahubLock | None:
+        lock_name = self.name_generator.generate_name(name=name, namespace=namespace, local=local)
+        return self.locks.pop(lock_name, None)
```
ogenstad: I'm wondering if we need to pop the lock from the registry like this. It would only be removed from the worker that runs the cleanup job, where a local reference to the lock would be held by the worker that originally acquired the lock. I'm also not completely sure when we assign anything to `self.locks` if popping it might be an issue if the current worker is waiting to acquire the lock once it has been released. Was this something you saw during local testing that we needed? It could be that we still want it, I'm just not completely sure.

backend/infrahub/pools/tasks.py
```diff
@@ -124,3 +126,33 @@ async def _create_number_pool(
...
                 await number_pool.save(db=dbs)
+
+
+@flow(
+    name="clean_up_deadlocks",
+    flow_run_name="Clean up deadlocks",
+)
+async def clean_up_deadlocks(service: InfrahubServices) -> None:
+    """Remove stale distributed locks left behind by inactive workers"""
+    lock_prefix = "lock."
```
ogenstad: I think we want to instead update the code within `infrahub/locks.py` and create a `lock_prefix` variable there and then use that constant in:

```python
            self.remote = GlobalLock(redis=self.connection, name=f"lock.{self.name}")
        else:
            self.remote = NATSLock(service=self.connection, name=f"lock.{self.name}")
```

Just to avoid issues if we rename something.

backend/infrahub/pools/tasks.py
```diff
@@ -124,3 +126,33 @@ async def _create_number_pool(
...
+            continue
+
+        timestamp, worker_id = value.split("::", 1)
+        if not workers_active_map.get(worker_id) and Timestamp() > Timestamp(timestamp).add(
+            minutes=config.SETTINGS.cache.clean_up_deadlocks_interval_mins
+        ):
+            name, namespace, local = LockNameGenerator().unpack_name(name=key.replace(lock_prefix, ""))
+            await lock.registry.pop_lock(name=name, namespace=namespace, local=local)
+            await service.cache.delete(key)
+            log.debug(f"Deleted deadlock key={key} worker={worker_id}")
```
ogenstad: As this is something that should not happen under normal conditions I think we can have a `log.info` here instead of `log.debug`.

backend/infrahub/pools/tasks.py
```diff
@@ -124,3 +126,33 @@ async def _create_number_pool(
...
+    """Remove stale distributed locks left behind by inactive workers"""
+    lock_prefix = "lock."
+    keys = await service.cache.list_keys(filter_pattern=f"{lock_prefix}*")
+    if not keys:
+        return
+
+    log = get_run_logger()
+    values = await service.cache.get_values(keys=keys)
+    workers = await service.component.list_workers(branch=registry.default_branch, schema_hash=False)
+    workers_active_map = {worker.id: worker.active for worker in workers}
```
ogenstad: I think we can rewrite this slightly to simplify the code below.

```diff
-    workers_active_map = {worker.id: worker.active for worker in workers}
+    workers_active = [worker.id for worker in workers if worker.active]
```

backend/infrahub/pools/tasks.py
```diff
@@ -124,3 +126,33 @@ async def _create_number_pool(
                     pool_type=NumberPoolType.SCHEMA.value,
                 )
                 await number_pool.save(db=dbs)
+
+
+@flow(
+    name="clean_up_deadlocks",
+    flow_run_name="Clean up deadlocks",
+)
+async def clean_up_deadlocks(service: InfrahubServices) -> None:
```
ogenstad: The module `infrahub.pools` is used for resource pools, currently we have number pools and ip address/prefix pools. As such we should move this code to some other file. For now I think the best option might be `infrahub/lock.py` (later I think we want to split up `lock.py` into separate files so this code would live under `infrahub/lock/tasks.py` but that can come in a later iteration.

If this would be problematic from an import point of view I think we can go ahead and create:

```
infrahub/locks/__init__.py
infrahub/locks/tasks.py (where the above code would live
```

Then we can later look at moving `infrahub/lock.py` into this new location.

backend/tests/functional/pools/test_clean_up_deadlocks.py
```diff
@@ -0,0 +1,25 @@
...
+async def test_clean_up_deadlocks(default_branch):
+    redis_cache = AsyncMock(RedisCache)
+    redis_cache.list_keys.return_value = ["lock.repository.sample-infrahub"]
+    redis_cache.get_values.return_value = ["2025-09-18T12:07:24.654862Z::800b4a90-87cd-458f-8c68-b0b5ebb81046"]
+    redis_cache.delete.return_value = True
+
+    component = AsyncMock(InfrahubComponent)
+    component.list_workers.return_value = [WorkerInfo(identity="800b4a90-87cd-458f-8c68-b0b5ebb81046")]
+
+    service = await InfrahubServices.new(cache=redis_cache, component=component)
```
ogenstad: FYI just also wanted to highlight that under tests/adapters/cache we have dummy adapters that can be used as a replacement for Redis (or NATS Cache) where you don't have to mock the values. The above code also works I just wanted to show that we had other options.

ogenstad:
Oh, we should also add a newsfragment to notify that the GitHub issue has been resolved.

backend/infrahub/lock.py
solababs: it will sometimes return `None` because the one i tested did not has the `local` or `global` prefix which is the case that it can be `None`

ogenstad: Ah sorry you're correct I glanced through it too quickly.

ajtmccarty:
I think something went wrong with this PR
looks like a lot of unrelated changes in here

solababs:
> I think something went wrong with this PR looks like a lot of unrelated changes in here

Yes, moved mypy fixes to https://github.com/opsmill/infrahub/pull/7255

backend/infrahub/workflows/catalogue.py
```diff
@@ -533,6 +533,16 @@
     function="validate_schema_number_pools",
 )
 
+CLEAN_UP_DEADLOCKS = WorkflowDefinition(
+    name="clean_up_deadlocks",
```
ogenstad: Unfortunately the naming isn't consistent across all of these workflow definitions as some of the git sync tasks are using underscores. But the bulk of the names are using a "-" so the desired name here would be "clean-up-deadlocks", you'd also need to change the name defined within the flow for this task so they match up.

backend/infrahub/locks/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+    values = await service.cache.get_values(keys=keys)
+    workers = await service.component.list_workers(branch=registry.default_branch, schema_hash=False)
+    workers_not_active = {worker.id for worker in workers if not worker.active}
+
+    for key, value in zip(keys, values, strict=False):
+        if not key or not value:
+            continue
+
+        timestamp, worker_id = value.split("::", 1)
+        if worker_id in workers_not_active and Timestamp() > Timestamp(timestamp).add(
```
ogenstad: I think we should switch this around so that we have a list of workers that are active. Old information is left in the cache for two hours and there could be situation where the worker information is no longer available in the cache.

So if we have:

```python
workers_active = {worker.id for worker in workers if worker.active}
...
if worker_id not in workers_active and Timestamp() > Timestamp(timestamp).add
```
We should avoid situations like that.

This would also require a slight update to the test.

ogenstad (approved):

PR #7237, opened 2025-09-20
saving profile attr values on nodes

ajtmccarty:
IFC-1808

new `NodeProfilesApplier` component to set the correct value, is_default, is_from_profile, and source metadata when a Node is created or updated

logic to handle updating all linked nodes when the attribute(s) of a profile are updated is up next


backend/infrahub/core/node/create.py
```diff
@@ -260,6 +236,8 @@ async def create_node(
             )
 
     if await get_profile_ids(db=db, obj=obj):
```
ajtmccarty, line 238: if we are creating a node, then we only need to apply profiles if the current state of the node includes `profiles` relationships. we don't need to worry about the pre-update state of the node like we do in an update

backend/infrahub/core/query/node.py
ajtmccarty, line 1: we include `is_from_profile` whenever we are retrieving an attribute from the database now b/c it affects whether we execute the `NodeProfilesUpdater` during an update or not

backend/infrahub/core/attribute.py
```diff
@@ -324,7 +324,7 @@ async def save(self, db: InfrahubDatabase, at: Timestamp | None = None) -> Attri
 
         save_at = Timestamp(at)
 
-        if not self.id or self.is_from_profile:
+        if not self.id:
```
ajtmccarty, line 327: we want to save attributes that are from profiles now

backend/infrahub/graphql/mutations/main.py
```diff
@@ -264,15 +260,14 @@ async def mutate_update_object(
             if field_to_remove in fields:
                 fields.remove(field_to_remove)
 
-        await obj.save(db=db, fields=fields)
+        after_mutate_profile_ids = await get_profile_ids(db=db, obj=obj)
+        if after_mutate_profile_ids or (not after_mutate_profile_ids and obj.uses_profiles()):
```
ajtmccarty, line 264: if the object either is linked to profiles _or_ has any attributes on it that use a profile (to handle the case when the `profiles` relationship for a Node is cleared out), then we need to run the `NodeProfileApplier`

backend/infrahub/profiles/queries/get_profile_data.py
ajtmccarty, line 1: new query to quickly get just the attribute values and priorities of a given list of profiles

backend/infrahub/profiles/node_applier.py
```diff
@@ -0,0 +1,63 @@
...
+from infrahub.core.node import Node
+from infrahub.database import InfrahubDatabase
+
+from .queries.get_profile_data import GetProfileDataQuery
+
+if TYPE_CHECKING:
+    from infrahub.core.attribute import BaseAttribute
+
+
+class NodeProfilesApplier:
```
ajtmccarty, line 11: pretty simple class to get all the profiles on a given node and then apply the profile attribute values to any attributes on the node where `is_default` or `is_from_profile` is True
optimization might be necessary if we end up using this same class to update all nodes linked to a given profile when a profile is updated

backend/tests/integration/profiles/test_profile_lifecycle.py
ajtmccarty, line 1: mostly changes to make this test case more thorough, but there are also some temporary changes that will be removed once the logic to handle a profile attribute update is in place. for now, we have to run the `NodeProfilesApplier` manually in a few cases

backend/tests/unit/core/test_manager_node.py
```diff
@@ -338,72 +338,6 @@ async def test_get_many_prefetch_hierarchical(
     assert parent_europe is None
 
 
-@pytest.mark.skip(reason="profile refactoring")
```
ajtmccarty, line 341: deleted tests are covered by the updated `TestProfileLifecycle` or new unit tests for `NodeProfilesApplier`

backend/infrahub/core/node/__init__.py
```diff
@@ -100,6 +100,13 @@ def get_id(self) -> str:
     def get_updated_at(self) -> Timestamp | None:
         return self._updated_at
 
+    def uses_profiles(self) -> bool:
+        for attr_name in self.get_schema().attribute_names:
+            node_attr = getattr(self, attr_name, None)
```
ogenstad:
```diff
- node_attr = getattr(self, attr_name, None)
+ node_attr = self.get_schema().get_attribute(name=attr_name)
```
This would be cleaner from a typing point of view. It would raise an error if we tried to get an attribute that didn't exist but given the context that should never happen.

backend/infrahub/core/query/node.py
```diff
@@ -634,20 +634,28 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
         MATCH (a)-[r:HAS_VALUE]-(av:AttributeValue)
         WHERE %(branch_filter)s
+        CALL (a) {
+            OPTIONAL MATCH (a)-[r:HAS_SOURCE]->(:CoreProfile)
+            WHERE %(branch_filter)s
+            RETURN r.status = "active" AS has_active_profile
+            ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+            LIMIT 1
+        }
+        WITH *, has_active_profile = TRUE AS is_from_profile
```
ogenstad, line 647: Would we want some kind of parameter to indicate if we want this additional subquery? I'm thinking when loading Schema type nodes that never should use a profile, or if the node has `generate_profile: false` within the schema. (this could be a later optimization I don't know how heavy this is)

backend/infrahub/graphql/mutations/main.py
```diff
@@ -264,15 +260,14 @@ async def mutate_update_object(
...
+            node_profiles_applier = NodeProfilesApplier(db=db, branch=branch)
+            await node_profiles_applier.apply_profiles(node=obj)
+            await obj.save(db=db)
```
ogenstad: Here we don't have the `fields=fields` part which means that all attributes (and I think collected relationships) of a node will be saved even though that might not have been desired. We've seen cases where multiple mutations step on each others toes without this. I.e. if something sends an update mutation for NodeA.name and another one sends an update mutation for NodeA.description, when these are applied during the same time we might end up reverting one of the changes. I'm not sure if `.apply_profiles()` can return a `list[str]` containing any "field" i.e. name of attribute or relationship updated and we can combine this list with the regular `fields` one?

backend/infrahub/profiles/node_applier.py
```diff
@@ -0,0 +1,63 @@
...
+            has_profile_data = False
+            node_attr = getattr(node, attr_name)
+            for profile_data in sorted_profile_data:
+                profile_value = profile_data.attribute_values.get(attr_name)
+                if profile_value is not None:
+                    has_profile_data = True
+                    node_attr.value = profile_value
+                    node_attr.is_default = False
+                    node_attr.is_from_profile = True
+                    node_attr.set_source(value=profile_data.uuid)
```
ogenstad: Based on the earlier comment regarding returning of field names. I think we only want to return the name of an attribute if it has in fact changed.

ogenstad (approved):
Added some comments but LGTM overall.

backend/infrahub/core/node/__init__.py
ajtmccarty: this wouldn't actually work in this case b/c we have the get the attribute from the `Node` and not the `NodeSchema`
that said, I hate calling `getattr` on nodes to get an attribute or relationship, so I added `get_attribute` and `get_relationship` to the `Node` class

backend/infrahub/core/query/node.py
ajtmccarty: I've made a small change so that we only try to get this `is_from_profile` value if the `Node` has ever been linked to a `node__profile` `Relationship`
so if profiles are not used at all for this node, then we can mostly skip this subquery
although I think the subquery should not add too much unless there are a lot of profile relationship changes on the node
I don't think there is a way to exclude the subquery based on a flag in the `NodeListGetAttribute` query b/c we need `is_from_profile` to be set correctly on a Node in the base update mutation and we won't know if we need to `is_from_profile` until we've already retrieved the Node and looked at the `profiles` relationship

backend/infrahub/graphql/mutations/main.py
ajtmccarty: good point. updated `apply_profiles` to return the names of the updated fields and add them to the existing `fields` list here

PR #7240, opened 2025-09-20
Fix HFID not being accepted when using the prefix/address mutations.

### LINKED ISSUE #7239: bug: Cannot create prefix/ip address when namespace RelatedNode is passed in using HFID
### Component

API Server / GraphQL

### Infrahub version

1.4.5/1.4.7

### Current Behavior

We get the following error when attempting to pass in the `ip_namespace` using HFID when attempting to create either an IP address or prefix.

```
{
  "data": {
    "IpamIPAddressCreate": null
  },
  "errors": [
    {
      "message": "A valid ip_namespace must be provided or ip_namespace should be left empty in order to use the default value.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "IpamIPAddressCreate"
      ]
    }
  ]
}
```

### Expected Behavior

We're able to create an IP Address or Prefix with a namespace defined using the namespace's HFID.

### Steps to Reproduce

1. Create new namespace on demo.infrahub.app
2. Attempt to create an IP address or Prefix using the namespace's HFID.
```
mutation CreatePrefix {
  IpamIPAddressCreate(
    data: {
      description: {value: "Test IP Address"}
      address: {value: "192.168.1.1/24"}
      ip_namespace: {hfid: ["Test1"]}
    }
  ) {
    ok
    object {
      id
    }
  }
}
```

### Additional Information

_No response_

FragmentedPacket:
Fixes #7239 

## Summary by CodeRabbit

- New Features
  - GraphQL mutations for creating and updating IP prefixes now accept IP namespaces referenced by HFID in addition to ID, while preserving existing ID-based behavior.

- Bug Fixes
  - Fixes creation/update of prefixes and IP addresses when an IP namespace is supplied via HFID.

- Tests
  - Added tests covering HFID-based create/update flows, ensuring successful operations, no errors, and correct namespace association on retrieval.

### DIFF
--- a/backend/tests/unit/graphql/mutations/test_ipam.py
+++ b/backend/tests/unit/graphql/mutations/test_ipam.py
@@ -376,7 +376,9 @@ async def test_ipprefix_create_with_ipnamespace(
     await ns.new(db=db, name="ns1")
     await ns.save(db=db)
 
-    query = """
+    ns_hfid = await ns.get_hfid_as_string(db=db)
+    # Test with namespace by ID
+    query_id = """
     mutation CreatePrefix($prefix: String!, $namespace: String!) {
         IpamIPPrefixCreate(
             data: {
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
Can you also add a news fragment to indicate that #7239 was fixed?

backend/tests/unit/graphql/mutations/test_ipam.py
```diff
@@ -416,6 +418,44 @@ async def test_ipprefix_create_with_ipnamespace(
...
+        }
+    }
+    """
+
+    result = await graphql(
+        schema=gql_params.schema,
+        source=query_hfid,
+        context_value=gql_params.context,
+        variable_values={"prefix": str(supernet), "namespace_hfid": ns_hfid},
+    )
```
ajtmccarty, line 447: is this creating the same IP prefix in the same namespace as the `CreatePrefix` mutation above? if so, should this fail with a uniqueness constraint violation?

ajtmccarty: looks like we don't set a uniqueness constraint on the builtin Ipam generics and the Ipam schemas for this test don't set one either, so I guess there is no issue

ajtmccarty (approved):

PR #7245, opened 2025-09-22
IFC-1823 Store the display_label and HFID of node attributes in the database

gmazoyer:
This PR brings several changes to how `display_label` and `human_friendly_id` for nodes are stored and retrieved. While it tries to change the least amount of code to keep code compatible as much as possible, it now stores the `display_label` and `human_friendly_id` of nodes into the database to be able to retrieve them when looking up nodes instead of computing them every times. It also introduces events to help in `display_label` computation.

This PR aims to be merged into https://github.com/opsmill/infrahub/pull/7197.

This PR provides implementations for IFC-1821, IFC-1822 and IFC-1823.


backend/infrahub/core/node/__init__.py
```diff
@@ -785,9 +789,18 @@ async def _update(
 
     async def save(self, db: InfrahubDatabase, at: Timestamp | None = None, fields: list[str] | None = None) -> Self:
         """Create or Update the Node in the database."""
-
         save_at = Timestamp(at)
 
+        if self._schema.namespace != "Schema":
+            self._display_label = String(
+                name="display_label",
+                schema=TextAttributeSchema(name="display_label", kind="Text", branch=BranchSupportType.AWARE),
```
ogenstad: I think we might want `self._schema.branch` here instead of hardcoding `branch=BranchSupportType.AWARE)`

backend/infrahub/core/node/__init__.py
```diff
@@ -785,9 +789,18 @@ async def _update(
 
     async def save(self, db: InfrahubDatabase, at: Timestamp | None = None, fields: list[str] | None = None) -> Self:
         """Create or Update the Node in the database."""
-
         save_at = Timestamp(at)
 
+        if self._schema.namespace != "Schema":
+            self._display_label = String(
```
ogenstad: One thing here is that we'd run this each time we save the node even if the change isn't required. I think it makes sense to have this on node creation and then within the update it should only happen if the fields in use by the display label have been updated.

backend/infrahub/core/diff/merger/serializer.py
```diff
@@ -155,6 +156,10 @@ async def serialize_diff(
                 continue
             serial_attr_diffs = []
             for attr_diff in node.attributes:
+                if attr_diff.name in NODE_ATTRIBUTES_TO_IGNORE and not node.kind.startswith("Schema"):
+                    # We ignore attributes on actual node and not schema nodes
+                    # These ignored attributes are ignored because we rely on recomputing them after a merge operation
+                    continue
```
gmazoyer, line 162: I wonder if we have a better way to handle this.

backend/infrahub/core/node/node_property_attribute.py
```diff
@@ -0,0 +1,250 @@
...
+        self._value = value
+
+        super().__init__(node_schema=node_schema)
+
+    @property
+    def is_jinja2_template(self) -> bool:
+        if self.template is None:
+            return False
+
+        return bool(re.search(r"{{.+}}", self.template) or re.search(r"{%.+%}", self.template))
```
ogenstad: Here we'd also want some broader definition of what a Jinja2 template is to account for broken templates.

```python
return bool("{" in self.template or "}" in self.template)
```

backend/infrahub/core/node/node_property_attribute.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def _analyze_plain_value(self) -> None:
+        if self.template is None or "__" not in self.template:
+            return
+
+        items = self.template.split("__", maxsplit=1)
+        if items[0] in self.node_schema.attribute_names:
+            self.node_attributes.append(items[0])
+        elif items[0] in self.node_schema.relationship_names:
+            self.node_relationships.append(items[0])
```
ogenstad: I don't think we support having single attribute display labels be relationships.

ogenstad (approved):

backend/infrahub/core/diff/merger/serializer.py
ajtmccarty: this is probably the easiest way to do it. another option would be to exclude these kind of changes from the cypher queries to calculate a diff, but that would be significantly harder. you could also maybe use an enricher like `DiffLabelsEnricher` to remove the display labels and HFID earlier in the diff process, but it would be the same result
it would be good to add an `is_schema` method/property to the `EnrichedDiffNode` to use in this conditional

backend/infrahub/core/node/__init__.py
```diff
@@ -710,12 +720,32 @@ async def load(
             )
             self._updated_at = Timestamp(updated_at)
 
+        if self._schema.namespace != "Schema":
```
ajtmccarty, line 723: also feels like it should be a property/method

backend/infrahub/core/node/__init__.py
```diff
@@ -978,6 +1021,20 @@ async def render_display_label(self, db: InfrahubDatabase | None = None) -> str:
             return repr(self)
         return display_label.strip()
 
+    async def update_human_friendly_id(self, value: list[str] | None) -> None:
```
ajtmccarty, line 1024: perhaps calling these methods `set_human_friendly_id/display_label` would make it more clear that they do not save the change to the database
not totally sure about that, just my first thought

backend/infrahub/core/node/node_property_attribute.py
```diff
@@ -0,0 +1,247 @@
...
+        if isinstance(self._value, AttributeFromDB):
+            return self._value.value
+        return self._value
+
+    def set_value(self, value: str | None) -> None:
+        """Force the value of the display label to the given one."""
+        if isinstance(self._value, AttributeFromDB):
+            self._value.value = value
+        else:
+            self._value = value
```
ajtmccarty, line 106: seems like these ones should be `abstractmethod`s too

backend/infrahub/core/node/__init__.py
```diff
@@ -196,6 +203,9 @@ def __init__(self, schema: NodeSchema | ProfileSchema | TemplateSchema, branch:
         self._is_protected: bool = None
         self._computed_jinja2_attributes: list[str] = []
 
+        self._display_label: DisplayLabel | None = None
+        self._human_friendly_id: HumanFriendlyIdentifier | None = None
```
ajtmccarty, line 207: would be good to have getters for these two instance variables so that you don't have to access private variables outside of this class

backend/infrahub/core/schema/schema_branch_display.py
```diff
@@ -0,0 +1,123 @@
...
+    ) -> None:
+        self._template_based_display_labels: dict[str, TemplateLabel] = template_based_display_labels or {}
+        self._template_relationship_triggers: dict[str, RelationshipTriggers] = template_relationship_triggers or {}
+
+    def duplicate(self) -> DisplayLabels:
+        """Clone the current object."""
+        return self.__class__(
+            template_based_display_labels=deepcopy(self._template_based_display_labels),
+            template_relationship_triggers=deepcopy(self._template_relationship_triggers),
+        )
```
ajtmccarty, line 61: could it be a problem if that the duplicated class will be linked to the same objects on the original class? so that a change on an object in the original will affect the object on the duplicate and vice versa?

backend/infrahub/graphql/mutations/display_label.py
```diff
@@ -0,0 +1,86 @@
...
+                node_type="target_node",
+                identifier=str(data.id),
+                message="The indicated node was not found in the database",
+            )
+
+        existing_label = target_node._display_label.value if target_node._display_label else None
+        print(existing_label)
+
+        # TODO: Add the proper code here and send the correct event to
+        # indicate that the display_label was updated
```
ajtmccarty, line 82: I guess this is coming in a later PR?

backend/infrahub/core/node/node_property_attribute.py
gmazoyer: That's what I wanted to do at first, but since `_value` is different between the classes, I could not figure what the typing of these should look like. Any idea?

ajtmccarty: I think using `TypeVar` and `Generic` would work
we use it in a few other places
something like this, have not tested this, but I think it's about right

```python
T = TypeVar("T")



class NodePropertyAttribute(Generic[T]):
    ...

    @abstractmethod
    def set_value(self, value: T) -> None: ...

class DisplayLabel(NodePropertyAttribute[str | None]):

    ...
    def set_value(self, value: str | None) -> None: ...

class HumanFriendlyIdentifier(NodePropertyAttribute[list[str] | None]):
    ...
    def set_value(self, value: list[str] | None) -> None: ...

```

PR #7246, opened 2025-09-22
IFC-1487: Allow min_count only for node creation

solababs:
## Summary by CodeRabbit

* **New Features**
  * Relationship cardinality accepts unspecified (nullable) maximums so a required minimum can be enforced without a max.

* **Bug Fixes**
  * Initialization and validation now handle nullable min/max bounds consistently, avoiding invalid comparisons when bounds are absent and normalizing unspecified bounds.

* **Tests**
  * Added unit tests for min-only counts and for optional vs non-optional relationship cardinality handling.


ogenstad (commented):
As I understood the problem is that currently if you set a `min_count` you are forced to also set a `max_count`. The error being that there's a check to validate that `min_count` has to be higher than than `max_count`.

With the current change the `max_count` would be set to that of `min_count` if not defined. I'm not sure that is what we want here. I.e. it could be that someone wants to have 2 relationships as a minimum but they don't care about the maximum number of relationships, if it's 0, 2, 10 or 20.

It might have been better if "min_count" and "max_count" had defaulted to `None` instead of `0` if not defined. 

```python
class RelationshipValidatorList:
    """Provides a list/set like interface to the RelationshipManager's _relationships but with validation against min/max count and no duplicates.

    Raises:
        ValidationError: If the number of relationships is not within the min and max count.
    """

    def __init__(self, *relationships: Relationship, name: str, min_count: int = 0, max_count: int = 0) -> None:
        """Initialize list for Relationship but with validation against min/max count.

        Args:
            min_count (int, optional): Min count of relationships required. Defaults to 0.
            max_count (int, optional): Max count of relationships allowed. Defaults to 0. 0 provides no limit.

        Raises:
            ValidationError: The number of relationships is not within the min and max count.
        """
        if max_count < min_count:
            raise ValidationError({"msg": "max_count must be greater than min_count"})
        self.min_count: int = min_count
        self.max_count: int = max_count
        self.name = name

        self._relationships: list[Relationship] = [rel for rel in relationships if isinstance(rel, Relationship)]
        self._relationships_count: int = len(self._relationships)

        # Validate the initial relationships count is within the min and max count if relationships were provided
        # Allow this class to be instantiated without relationships
        if self._relationships:
            if self.max_count and self._relationships_count > self.max_count:
                self._raise_too_many()
            if self.min_count and self._relationships_count < self.min_count:
                self._raise_too_few()
```

Looking at this validator I wonder if we can just change is so that `if max_count < min_count:` is only evaluated if both `max_count` and `min_count` has a non zero value.

backend/infrahub/core/relationship/model.py
```diff
@@ -570,7 +570,9 @@ class RelationshipValidatorList:
         ValidationError: If the number of relationships is not within the min and max count.
     """
 
-    def __init__(self, *relationships: Relationship, name: str, min_count: int = 0, max_count: int = 0) -> None:
+    def __init__(
+        self, *relationships: Relationship, name: str, min_count: int | None = 0, max_count: int | None = 0
+    ) -> None:
```
ajtmccarty, line 575: I agree that it makes more sense for `max_count=None` to indicate no maximum exists as opposed to `max_count=0`. But given that we use `max_count=0` on the schema and on this class following the initial validation, maybe it makes more sense to just stick with `max_count=0` all the way through instead of allowing `max_count=None` at instantiation and then converting it to `max_count=0` a few lines later
I think there's a good argument either way, so up to you

solababs: @ajtmccarty I thought so, which is how i did it initially here https://github.com/opsmill/infrahub/pull/7246/commits/7a3db42ee49f7807ce06e0a322a368604527d2ca before @ogenstad's comment here https://github.com/opsmill/infrahub/pull/7246#pullrequestreview-3256750320, doing it that way requires us to set a `max_count` because the `min_count` only way(schema sets only min_count) for when it is not optional will hit the validation checking the `max_count > min_count` and we do not want to set any value for the `max_count` if it is not explicitly set by the user

ogenstad: What I was thinking was something like.

```python
if max_count > 0 and max_count < min_count:
```

Then we'd only implement this check if the user had made modifications from the default of `0`.

solababs: Those conditions are already considered here https://github.com/opsmill/infrahub/pull/7246/files#diff-0ff18534de790bccd074cc2d53a09a39af514af3935d66038b72f26b44d98073R748 in `RelationshipManager` and the check is done here https://github.com/opsmill/infrahub/pull/7246/files#diff-0ff18534de790bccd074cc2d53a09a39af514af3935d66038b72f26b44d98073R585 in `RelationshipValidatorList`

backend/infrahub/core/relationship/model.py
```diff
@@ -726,15 +734,24 @@ def __init__(self, schema: RelationshipSchema, branch: Branch, at: Timestamp, no
...
         self.has_fetched_relationships: bool = False
         self.lock = asyncio.Lock()
 
+    def _get_init_relationships(self) -> RelationshipValidatorList:
+        min_count = self.schema.min_count
+        max_count: int | None = self.schema.max_count
+        if self.schema.optional:
+            min_count = 0
+        else:
+            max_count = self.schema.max_count if self.schema.max_count > 0 else None
```
ajtmccarty: I think this should be outside of the conditional so that `max_count` can still be set to `None` even if the relationship is optional

solababs: @ajtmccarty  We still need the default to be https://github.com/opsmill/infrahub/pull/7246/files#diff-0ff18534de790bccd074cc2d53a09a39af514af3935d66038b72f26b44d98073R743-R744 making it `None` would not pass some tests
I added more tests to confirm the behaviour, see updated.

ajtmccarty: it looks like all the tests in `test_relationship_query` and `test_relationship_manager` pass if I make these changes locally
```
     def _get_init_relationships(self) -> RelationshipValidatorList:
         min_count = self.schema.min_count
-        max_count: int | None = self.schema.max_count
+        max_count: int | None = self.schema.max_count if self.schema.max_count > 0 else None
         if self.schema.optional:
             min_count = 0
-        else:
-            max_count = self.schema.max_count if self.schema.max_count > 0 else None
```
do you know which tests fails with this change?

solababs: @ajtmccarty see updated

backend/tests/unit/core/test_relationship_manager.py
ajtmccarty, line 1: you can use the `pytest.mark.parametrize` decorator to run these 4 tests with fewer lines
something like this (I didn't review the below code, it's just an example)

```python
@pytest.mark.parametrize(
    "optional,schema_min_count,schema_max_count,expected_min_count,expected_max_count",
    [
        (False, 2, 0, 2, 0),  # not optional, min_count only
        (True, 2, 0, 0, 0),   # optional, min_count only
        (False, 2, 5, 2, 5),  # not optional, both min and max
        (True, 2, 5, 0, 5),   # optional, both min and max
    ],
)
async def test_can_create_relationship_manager_with_optional_and_count_constraints(
    db: InfrahubDatabase,
    tag_blue_main: Node,
    person_jack_primary_tag_main: Node,
    branch: Branch,
    optional: bool,
    schema_min_count: int,
    schema_max_count: int,
    expected_min_count: int,
    expected_max_count: int,
):
    person_schema = registry.schema.get(name="TestPerson")
    rel_schema = person_schema.get_relationship("primary_tag")
    rel_schema.optional = optional
    rel_schema.min_count = schema_min_count
    rel_schema.max_count = schema_max_count

    relm = await RelationshipManager.init(
        db=db,
        schema=rel_schema,
        branch=branch,
        at=Timestamp(),
        node=person_jack_primary_tag_main,
    )
    await relm.save(db=db)

    assert relm._relationships.min_count == expected_min_count
    assert relm._relationships.max_count == expected_max_count
```

ajtmccarty (approved):

backend/tests/unit/core/test_relationship_manager.py
ogenstad: An approach for this that I think we should use more is to use objects when defining the parameters we have in tests similar to how we do it here:
https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/tests/unit/ctl/test_render_app.py#L30-L59

```python
@dataclass
class RenderAppFailure:
    name: str
    template: str
    error: str


RENDER_APP_FAIL_TEST_CASES = [
    RenderAppFailure(
        name="main-template-not-found",
        template="tag_format_missing",
        error="Missing template: tag_format.file-is-missing",
    ),
    RenderAppFailure(
        name="has-undefined-variables",
        template="undefined_variables",
        error="'host' is undefined",
    ),
    RenderAppFailure(
        name="has-syntax-error",
        template="syntax_error",
        error="unexpected '}'",
    ),
    RenderAppFailure(
        name="invalid-filter",
        template="missing_filter",
        error="No filter named 'my_filter_is_missing'.",
    ),
]


@pytest.mark.parametrize(
    "test_case",
    [pytest.param(tc, id=tc.name) for tc in RENDER_APP_FAIL_TEST_CASES],
)
@requires_python_310
def test_validate_template_not_found(test_case: RenderAppFailure, httpx_mock: HTTPXMock) -> None:
```

Since we in this case get a single object as a test_case using the RenderAppFailure it's easier to access properties of this object within the test and understand where the variables are coming from. And by assigning an `id` to the pytest.param it's easier to identify each job in the output when running the tests.

If we look at parameters like this:

```python
    "optional,schema_min_count,schema_max_count,expected_min_count,expected_max_count",
    [
        (False, 2, 0, 2, 0),  # not optional, min_count only
        (True, 2, 0, 0, 0),   # optional, min_count only
        (False, 2, 5, 2, 5),  # not optional, both min and max
        (True, 2, 5, 0, 5),   # optional, both min and max
    ],
)
```

It requires more mental energy to figure out what each testcase is and what it is we are testing (at least for me)

ajtmccarty (approved):

PR #7255, opened 2025-09-23
Refactor and move locks file into locks package and fix all mypy issues

solababs:


ajtmccarty:
I appreciate that `core_registry` is a better name than `registry` (especially if we also have `lock_registry`), but this seems more difficult to troubleshoot than it is worth. I think the best solution would actually be to remove the `registry` entirely and have classes that follow the Single Responsibility Principle that we can inject where they are required, but that would be an even bigger change than this.

So maybe just go with the changes in `infrahub.lock` for now and any others that are absolutely required to resolve the mypy issues

ogenstad:
> I appreciate that `core_registry` is a better name than `registry` (especially if we also have `lock_registry`), but this seems more difficult to troubleshoot than it is worth. I think the best solution would actually be to remove the `registry` entirely and have classes that follow the Single Responsibility Principle that we can inject where they are required, but that would be an even bigger change than this.
> 
> So maybe just go with the changes in `infrahub.lock` for now and any others that are absolutely required to resolve the mypy issues

I agree with Aaron here, let's just go with the required changes to locks and take a step back for the other changes. I think the registry needs a refactoring in general but let's do it later in another way.

solababs:
> > I appreciate that `core_registry` is a better name than `registry` (especially if we also have `lock_registry`), but this seems more difficult to troubleshoot than it is worth. I think the best solution would actually be to remove the `registry` entirely and have classes that follow the Single Responsibility Principle that we can inject where they are required, but that would be an even bigger change than this.
> > So maybe just go with the changes in `infrahub.lock` for now and any others that are absolutely required to resolve the mypy issues
> 
> I agree with Aaron here, let's just go with the required changes to locks and take a step back for the other changes. I think the registry needs a refactoring in general but let's do it later in another way.

@ogenstad this is the required change because moving the locks into its package exposes the `registry` name in both locks and the core registry that is why we have to rename both

ogenstad:
> @ogenstad this is the required change because moving the locks into its package exposes the `registry` name in both locks and the core registry that is why we have to rename both

In which file is it we have this conflict?

What we have in some places where we need to import a name that already exists is to do something like moving from:

```python
from infrahub.core.registry import registry
```

to:

```python
from infrahub.core.registry import registry as core_registry
```

solababs:
> > @ogenstad this is the required change because moving the locks into its package exposes the `registry` name in both locks and the core registry that is why we have to rename both
> 
> In which file is it we have this conflict?
> 
> What we have in some places where we need to import a name that already exists is to do something like moving from:
> 
> ```python
> from infrahub.core.registry import registry
> ```
> 
> to:
> 
> ```python
> from infrahub.core.registry import registry as core_registry
> ```

All the files that has the registry, since we have removed the mypy overrides for locks and moved the locks file into its package and also the registry is exposed from core's `__init__` file as registry, which makes the registry available from core causing the mypy conflicts
that is why the change `from infrahub.core.registry import registry as core_registry` is done in the `__init__` file

ajtmccarty:
I like the change to use `get_lock_registry()`. that is a big improvement
If `InfrahubLockRegistry` is now saved in the global variable `lock_registry` instead of `registry`, then I'd think it would be safe to leave `core.registry` as just `registry` instead of `core_registry`, which will reduce the number of changed lines in this PR significantly

PR #5971, opened 2025-03-07
allow disabling profile logic in NodeGetListQuery

ajtmccarty:
adds a new env var `INFRAHUB_EXPERIMENTAL_NO_PROFILES` that removes the profile-related logic from `NodeGetListQuery` and `NodeManager.get_many`. this results in a substantially simpler `NodeGetListQuery` cypher query and an improvement in performance. my local testing showed perhaps a 10% improvement in performance when running a single query with and without the profiles logic, but I am not completely certain that there is actually a performance improvement. perhaps running one of the large generators the team has been working on is the best way to determine if this update actually improves performance

~this new env var also removes any profile-related mutations from GraphQL schema.~ this caused toast errors on the front end, so removed it

if a user has profiles linked to objects and then turns activates this setting, queries for objects linked to existing profiles will no longer be correct. I'm not sure if there are more guardrails that we should put in place for this

PR #7256, opened 2025-09-23
remove profiles on read

ajtmccarty:
targets the new feature branch for the profiles refactoring

delete the logic to apply profiles while reading from the database and skip the affected failing tests. tests will be updated and reactivated as the following PRs come in to support creating/updating nodes using profiles at write time


fatih-acar:
:tada: I think you can close https://github.com/opsmill/infrahub/pull/5971

ogenstad (approved):

PR #7258, opened 2025-09-24
Use ConversionFieldInput from SDK

LucasG0:
Use classes within SDK to avoid duplication.
ConvertObjectType mutation now accepts `use_default_value` in order to specify when we want of a field to use the default value instead of potentially reusing the value of an existing matching input field. Corresponding sdk PR: https://github.com/opsmill/infrahub-sdk-python/pull/558

## Summary by CodeRabbit

* **New Features**
  * Improved object-type conversion with direct field mappings, raw-value extraction, and explicit default-value handling; relationship conversions (single/multiple) more robust.
  * Converted objects now preserve additional optional attributes (e.g., favorite_color).

* **Refactor**
  * Public conversion input types replaced with new SDK-provided wrappers; callers and integrations should adopt the new input structures.

* **Tests**
  * Expanded tests to cover new input types, default-value behavior, and optional-attribute preservation.

* **Chores**
  * SDK subproject reference bumped.

### DIFF
--- a/backend/infrahub/core/convert_object_type/object_conversion.py
+++ b/backend/infrahub/core/convert_object_type/object_conversion.py
@@ -1,6 +1,6 @@
-from typing import Any, Self, assert_never
+from typing import Any, assert_never
 
-from pydantic import BaseModel, model_validator
+from infrahub_sdk.convert_object_type import ConversionFieldInput, ConversionFieldValue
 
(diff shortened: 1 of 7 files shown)


ajtmccarty (approved):
can you link the PR for the sdk change in the description of this one please?

PR #7280, opened 2025-09-26
allow attribute properties (source, owner) to be deleted

ajtmccarty:
we don't provide a way to delete a source/owner property on an attribute. we only allow updating to a different source/owner. this PR adds the ability to delete the properties

## Summary by CodeRabbit

- New Features
  - Explicitly clear property values.
  - Branch-aware updates and retrievals for node and relationship metadata (source, owner, visibility, protection).

- Bug Fixes
  - More accurate change tracking and updates only when needed, reducing unnecessary writes.
  - Correct selection of active relationships and associated metadata during queries.
  - Ensures related relationship timestamps are updated after property changes.
  - Improved consistency when saving nodes after related ownership updates.

### DIFF
--- a/backend/infrahub/core/property.py
+++ b/backend/infrahub/core/property.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from enum import Enum
 from typing import TYPE_CHECKING
 from uuid import UUID
 
@@ -24,6 +25,10 @@ class ValuePropertyData(BaseModel):
 class NodePropertyData(BaseModel):
     name: str
     peer_id: str
+
+
+class ClearValue(Enum):
+    CLEAR = "clear"
 
 
 class FlagPropertyMixin:
@@ -51,6 +56,7 @@ class NodePropertyMixin:
         for node in self._node_properties:
             setattr(self, f"_{node}", None)
             setattr(self, f"{node}_id", None)
+            setattr(self, f"_clear_{node}", False)
 
         if not kwargs:
             return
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/query/node.py
```diff
@@ -668,16 +668,32 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
 
         if self.include_source:
             query = """
-            OPTIONAL MATCH (a)-[rel_source:HAS_SOURCE]-(source)
-            WHERE all(r IN [rel_source] WHERE ( %(branch_filter)s ))
```
ajtmccarty, line 672: this cypher and the similar bit under `if self.include_owner` does not account for a source being deleted or updated on a branch, in which case there could be multiple active paths to different `source` `Node`s

backend/infrahub/core/query/attribute.py
```diff
@@ -144,25 +144,43 @@ def __init__(
...
+        else:
+            node_query = """
+        MATCH (np:Node { uuid: $prop_id })-[r:IS_PART_OF]->(:Root)
+        WHERE %(branch_filter)s
+        ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+        LIMIT 1
+        WITH np
+        WHERE r.status = "active"
+            """ % {"branch_filter": branch_filter}
+        self.add_to_query(node_query)
```
ajtmccarty, line 176: the changes to get the correct `np` `Node` are necessary b/c there can be multiple `Node`s with the same UUID in the case of a node schema kind/inheritance update, so we need to make sure that we only get the latest active one on this branch

backend/infrahub/core/property.py
```diff
@@ -26,6 +27,10 @@ class NodePropertyData(BaseModel):
     peer_id: str
 
 
+class ClearValue(Enum):
+    CLEAR = "clear"
```
ajtmccarty, line 31: this functions as a sentinel value for marking a property as needing to be NULLed

dgarros (approved):

PR #7285, opened 2025-09-29
Fix payload for transform webhook to align with other webhooks

### LINKED ISSUE #6815: bug: Event Metadata Not Available in Transform for Custom Webhooks
### Component

_No response_

### Infrahub version

1.3.2

### Current Behavior

When using the built-in branch-merged event webhook, the expected metadata is included in the payload. However, for custom webhooks, the metadata is not available in the transform, limiting the data that can be used to build the payload.

### Expected Behavior

Metadata should be accessible in the transform for custom webhooks, allowing it to be used when building the payload.

### Steps to Reproduce

Trigger a built-in event webhook (e.g., branch-merged) and observe that metadata is present in the payload.
Create and trigger a custom webhook.
Attempt to access metadata in the transform for the custom webhook.

### Additional Information

This will be a breaking change when adding `_metadata` in the transforms

ogenstad:
Fixes #6815

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized webhook payload structure for transform-based webhooks to match standard/custom webhooks. The original payload is now nested under data.data, improving consistency across webhook types. Existing transforms may need updates to handle the new structure.

* **Documentation**
  * Added a changelog entry noting the breaking change and advising users to update transforms to the new payload format.

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -231,7 +231,7 @@ class TransformWebhook(Webhook):
             commit=commit,
             location=f"{self.transform_file}::{self.transform_class}",
             convert_query_response=self.convert_query_response,
-            data={"data": data, **context.model_dump()},
+            data={"data": {"data": data, **context.model_dump()}},
             client=client,
         )  # type: ignore[misc]
 
--- a/changelog/6815.fixed.md
+++ b/changelog/6815.fixed.md
@@ -1,0 +1,1 @@
+Breaking change: The format of the data payload has been corrected for transform based webhooks so that they are consistend with standard webhooks as well as custom webhooks (without an attached transform). Due to this change any transforms attached to a webhook needs to be updated to account for the new format.


backend/infrahub/webhook/models.py
```diff
@@ -231,7 +231,7 @@ async def _prepare_payload(self, data: dict[str, Any], context: EventContext, cl
             commit=commit,
             location=f"{self.transform_file}::{self.transform_class}",
             convert_query_response=self.convert_query_response,
-            data={"data": data, **context.model_dump()},
+            data={"data": {"data": data, **context.model_dump()}},
```
ajtmccarty, line 234: data data data data

ajtmccarty (approved):

backend/infrahub/webhook/models.py
ogenstad: Yes a requirement that comes from here:
https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/infrahub_sdk/transforms.py#L75

It's the Matryoshka-data design model. I think it's one of those forgotten appendixes to SOLID.

Issue #6125, opened 2025-03-24
bug: parent, children, ancestors and descendants should be added to the restricted attributes and relationship names


wvandeun:
### Infrahub version

1.2.0

### Current Behavior

You can define a schemanode with one of the following names or attributes: parent, children, ancestors and descendants. These relationship names are used internally for hierarchical nodes and should be restricted.

### Expected Behavior

A user should not be allowed to create relationships or attributes with the name parent/children/ancestors/descendants.

### Steps to Reproduce

- spin up instance of Infrahub
- load this schema
```yaml
---
version: "1.0"
generics:
  - name: Location
    namespace: Generic
    hierarchical: true
    attributes:
      - name: name
        unique: true
        optional: false
        kind: Text
nodes:
  - name: Site
    namespace: Location
    inherit_from:
      - GenericLocation
    children: "TestingParent"
    parent: ""

  - name: Parent
    namespace: Testing
    inherit_from:
      - GenericLocation
    children: ""
    parent: LocationSite
    relationships:
      - name: children
        kind: Generic
        optional: true
        peer: TestingChild
        cardinality: many

  - name: Child
    namespace: Testing
    attributes:
      - name: name
        unique: true
        optional: false
        kind: Text
    relationships:
      - name: parent
        kind: Attribute
        optional: false
        peer: TestingParent
        cardinality: one
```
- Create a site node
- Create a parent node, set the parent to the site
- Create a child node, set the parent to the parent
- Navigate to the parent detail view, click on the children tab
- An error message will show
```
Fragment cannot be spread here as objects of type 'GenericLocation' can never be of type 'TestingChild'.
```

### Additional Information

Did an attempt to parent/children/ancestors and descendants to `RESERVED_ATTR_REL_NAMES` but after it the server would not start
https://github.com/opsmill/infrahub/blob/d47d0a47d4827bc7690acc3742e9f99c6f4f1986/backend/infrahub/core/constants/__init__.py#L28

```
[...]

PR #7287, opened 2025-09-29
IFC-1405: Update restricted words for generics attributes and relationships

solababs:
Fixes: https://github.com/opsmill/infrahub/issues/6125

## Summary by CodeRabbit

* **New Features**
  * Added schema validation that disallows reserved hierarchical names ("ancestors", "descendants") as attribute or relationship names on hierarchical nodes; offending schemas are rejected with descriptive errors.

* **Tests**
  * Added component and core tests plus new fixtures to ensure hierarchical-name restrictions are enforced and error messages surface correctly.


backend/infrahub/core/schema/relationship_schema.py
```diff
@@ -23,6 +23,15 @@
...
     _sort_by: list[str] = ["name"]
+    _loaded_from_api: bool = PrivateAttr(default=False)
+
+    @property
+    def loaded_from_api(self) -> bool:
+        return self._loaded_from_api
+
+    @loaded_from_api.setter
+    def loaded_from_api(self, value: bool) -> None:
+        self._loaded_from_api = value
```
ajtmccarty: this violates the Single Responsibility Principle, which is something we do all over the place, but we are trying not to add any more
the `RelationshipSchema` should not need data indicating whether it was loaded from the API or not
if we need to do validation of data from the API before it is instantiated into a `RelationshipSchema`, then that should happen in the API endpoint somewhere

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -989,6 +995,28 @@ def validate_names(self) -> None:
...
 
+    def validate_hierarchical_nodes_restricted_words(self) -> None:
+        for name in self.all_names:
+            node = self.get(name=name, duplicate=False)
+            is_hierarchical_node = (isinstance(node, GenericSchema) and node.hierarchical) or (
+                isinstance(node, NodeSchema) and node.hierarchy
+            )
+
+            if not is_hierarchical_node or node.kind in INTERNAL_SCHEMA_NODE_KINDS:
+                continue
```
ajtmccarty: I think that we want this restriction to be more general and prevent using any of the `RESERVED_ATTR_REL_HIERARCHICAL_NAMES` names on any attribute or relationship that is defined by the user, not just on hierarchical nodes

ajtmccarty (commented):
I appreciate that you are following the existing patterns that we use for schema validation and that you are restricting values as minimally as possible, but I think we want a slightly different approach for this logic.
I suspect that we cannot add this restriction logic anywhere in `SchemaBranch.process` b/c we could be loading an existing schema that has valid `parent`, `children`, etc. relationships on it
I think we probably need to add this restriction either somewhere in the `load_schema` endpoint or on the `SchemasLoadAPI` object. I think the most correct approach is to add a new `SchemaApiValidator` class (or some other name) that runs this type of API-level validation on an incoming schema and can be extended to add more validations in the future, but for this small check, maybe something on `SchemasLoadAPI` would be okay too

also, can you add `fixes #7287` to the PR description so that it is correctly linked in GH please?

backend/infrahub/core/schema/schema_branch.py
ogenstad: If we stick to restricting `RESERVED_ATTR_REL_HIERARCHICAL_NAMES` globally that would be the easiest solution as we can just iterate over the input from the user and restrict these attribute or relationship names from entering the system. My main concern would be if there are users out there that have already used these names anywhere.

I.e. today this is a perfectly valid schema:

```yaml
---
version: '1.0'

nodes:
  - name: Person
    namespace: Test
    label: "Person"
    default_filter: name__value
    display_labels:
      - name__value
    attributes:
      - name: name
        kind: Text
      - name: children
        kind: Number
      - name: description
        kind: Text
        optional: true
```

It's fine to use "children" as the name of an attribute as long as the node itself isn't hierarchical. It might be perfectly fine but it could also be that some installations face problems due to this.

ogenstad:
> I appreciate that you are following the existing patterns that we use for schema validation and that you are restricting values as minimally as possible, but I think we want a slightly different approach for this logic. I suspect that we cannot add this restriction logic anywhere in `SchemaBranch.process` b/c we could be loading an existing schema that has valid `parent`, `children`, etc. relationships on it I think we probably need to add this restriction either somewhere in the `load_schema` endpoint or on the `SchemasLoadAPI` object. I think the most correct approach is to add a new `SchemaApiValidator` class (or some other name) that runs this type of API-level validation on an incoming schema and can be extended to add more validations in the future, but for this small check, maybe something on `SchemasLoadAPI` would be okay too
> 
> also, can you add `fixes #7287` to the PR description so that it is correctly linked in GH please?

I think it would be problematic to only have a validation within the `SchemasLoadAPI` object if we only want to restrict this nodes that are hierarchical. I.e. if a user submits a schema that inherits from some existing generic we'd need to merge those schemas in order to determine if the generic is a hierarchical node. Potentially we could use the registry to do a lookup of this during the load phase though.

ajtmccarty:
I stand by my previous comment that this check belongs in the schema loading logic somewhere. I made a similar change recently to add `_reconcile_text_attribute_parameters` to `SchemaBranch.load_schema` and this wouldn't be exactly the same thing because it can raise an error.

I think a `UserSchemaValidator` class with a `validate(self, user_schema: SchemaRoot) -> SchemaRoot` or something similar would be a good place to have both `_reconcile_text_attribute_parameters` and `protect_hierarchical_relationship_names`

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1134,6 +1136,28 @@ def validate_names(self) -> None:
...
+                raise ValueError(f"{node.kind}: {attr.name} isn't allowed as an attribute name on hierarchical nodes.")
+
+        for rel in node.relationships:
+            if rel.name in RESERVED_ATTR_REL_HIERARCHICAL_NAMES:
+                raise ValueError(f"{node.kind}: {rel.name} isn't allowed as a relationship name on hierarchical nodes.")
+
+    def validate_hierarchical_nodes_restricted_words(self) -> None:
+        for name in self.all_names:
+            node = self.get(name=name, duplicate=False)
+            if not node.id:
```
ajtmccarty: this looks like a pretty good solution to allowing loading existing valid hierarchical schemas
there is the edge case where a user loads a new schema with an ID defined and an attribute named `"parent"`, which I think we would incorrectly allow in this case, but that feels like a very corner-y case
please add a test that covers the case where ID is defined on a schema that already has `parent/children` relationships to validate it is allowed

backend/infrahub/core/validators/schema_branch/__init__.py
ajtmccarty, line 1: I think these files should be left empty unless an existing component is getting moved and the import is required for backwards compatibility. forcing the import of everything in a module makes us more likely to encounter import errors

backend/infrahub/core/schema/definitions/core/__init__.py
```diff
@@ -197,3 +197,5 @@ class CoreModelsMixedType(TypedDict):
     "generics": [item.to_dict() if isinstance(item, GenericSchema) else item for item in core_models_mixed["generics"]],
     "nodes": [item.to_dict() if isinstance(item, NodeSchema) else item for item in core_models_mixed["nodes"]],
 }
+
+CORE_MODELS_NODE_KINDS = [node["namespace"] + node["name"] for node in core_models["nodes"] + core_models["generics"]]
```
ogenstad: I don't think we need to create this variable, but we can instead just check if the node is within the "Core" namespace.

backend/infrahub/core/constants/__init__.py
```diff
@@ -40,6 +40,13 @@
     "process_pools",
 ]
 
+RESERVED_ATTR_REL_HIERARCHICAL_NAMES = [
+    # "parent",
+    # "children",
```
ogenstad: If we don't need the parent and children here I think we should just remove them to avoid confusion. Alternatively add a comment.

backend/infrahub/core/validators/schema_branch/hierarchical_nodes_restricted_words_validator.py
```diff
@@ -0,0 +1,41 @@
...
+            node = schema_branch.get(name=name, duplicate=False)
+            if not node.id:
+                self._validate_hierarchical_node_restricted_words(node=node)
+
+    def _validate_hierarchical_node_restricted_words(self, node: MainSchemaTypes) -> None:
+        is_hierarchical_node = (isinstance(node, GenericSchema) and node.hierarchical) or (
+            isinstance(node, NodeSchema) and node.hierarchy
+        )
+
+        if not is_hierarchical_node or node.kind in INTERNAL_SCHEMA_NODE_KINDS or node.kind in CORE_MODELS_NODE_KINDS:
```
ogenstad: I think we can just have this instead and get rid of that variable:

```diff
- if not is_hierarchical_node or node.kind in INTERNAL_SCHEMA_NODE_KINDS or node.kind in CORE_MODELS_NODE_KINDS:
+ if not is_hierarchical_node or node.kind in INTERNAL_SCHEMA_NODE_KINDS or node.namespace == "Core:
```

backend/tests/component/api/test_40_schema.py
```diff
@@ -443,3 +443,49 @@ async def test_schema_read_endpoints_anonymous_account(
...
+    db: InfrahubDatabase,
+    client: TestClient,
+    admin_headers,
+    default_branch: Branch,
+    prefect_test_fixture,
+    workflow_local,
+    authentication_base,
+    helper,
+    reserved_name,
+):
```
ogenstad: When adding new tests it would be good if we had typehints for all of the parameters and also the "-> None" return value.

ogenstad (approved):

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -110,6 +111,8 @@ def __init__(
             self.profiles = data.get("profiles", {})
             self.templates = data.get("templates", {})
 
+        self.validators: list[SchemaBranchValidator] = [HierarchicalNodesRestrictedWords()]
```
ajtmccarty, line 119: nice first step towards how the SchemaBranch validation should be organized
some day we'll have a `AggregatedSchemaBranchValidatior` or something that accepts a list of `SchemaBranchValidators` in its constructor and has a `validate(self, schema_branch: SchemaBranch)` method

ajtmccarty (approved):

PR #7291, opened 2025-09-29
Add missing parts for display label mutation

ogenstad:
Changes in this PR:

* Updates some internal logic within `Node` so that we can update the display label in a manual way that doesn't regenerate the display_label from scratch. This is used when the update is done from a related node.
* Also updates the graphql mutation to update display labels by using the .set_display_label method on the Node objects and then sends an event with the update.
* Added unittests for the display label mutation
* Added integration tests as part of the old computed attribute test where the display label of a node is modified when triggered by a change in a related node.
* Slight change to the computed attribute test logic as we also rename the color of the tshirt (as the color got added to the display label of the tshirt nodes and we needed to get an updated value)

### DIFF
--- a/backend/infrahub/core/node/node_property_attribute.py
+++ b/backend/infrahub/core/node/node_property_attribute.py
@@ -29,11 +29,14 @@ class NodePropertyAttribute:
         self.node_schema = node_schema
         self.node_attributes: list[str] = []
         self.node_relationships: list[str] = []
+        self._manually_assigned = False
 
         self.analyze_variables()
 
     def needs_update(self, fields: list[str] | None) -> bool:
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/node/node_property_attribute.py
```diff
@@ -81,13 +84,16 @@ def value(self) -> str | None:
             return self._value.value
         return self._value
 
-    def set_value(self, value: str | None) -> None:
+    def set_value(self, value: str | None, manually_assigned: bool = False) -> None:
```
gmazoyer, line 87: 👍🏻 this is something I wanted to add, nice one

gmazoyer (approved):

PR #7292, opened 2025-09-30
handle profile attribute and priority updates

ajtmccarty:
IFC-1809

add logic for handling updates to profile attributes and priority so that all nodes using a given profile are correctly updated when that node is added/updated/deleted

there are a few pieces to this new logic
- a new `InfrahubProfilesRefresh` mutation that accepts a single ID and will refresh the profile values for that particular node
- a new `object_profiles_refresh_multiple` workflow that accepts a list of object IDs to be refreshed
- a new `object_profiles_refresh` workflow that accepts a single ID and calls the `InfrahubProfilesRefresh` mutation on that node
- a new `InfrahubProfileMutation` class that handles submitting the `object_profiles_refresh_multiple` workflow, if necessary. the objects associated with a given profile require a refresh in the following cases
  - a new profile is created and includes `related_nodes`. in this case all `related_nodes` are submitted for a profile refresh
  - a profile is updated and the update includes changes to an attribute (except `profile_name`). in this case, all objects linked to the profile are submitted for a profile refresh
  -  a profile is updated and the update includes changes to the `related_nodes` relationship on the profile. in this case, any nodes added to or removed from the `related_nodes` relationship are submitted for a refresh
  - a profile is deleted. in this case, any `related_nodes` are submitted for a refresh

### DIFF
--- a/backend/infrahub/graphql/mutations/main.py
+++ b/backend/infrahub/graphql/mutations/main.py
@@ -379,6 +379,15 @@ class InfrahubMutationMixin:
             return updated_obj, mutation, False
 
     @classmethod
+    async def _delete_obj(cls, graphql_context: GraphqlContext, branch: Branch, obj: Node) -> list[Node]:
+        db = graphql_context.db
+        async with db.start_transaction() as dbt:
+            deleted = await NodeManager.delete(db=dbt, branch=branch, nodes=[obj])
+        deleted_str = ", ".join([f"{d.get_kind()}({d.get_id()})" for d in deleted])
+        log.info(f"nodes deleted: {deleted_str}")
+        return deleted
+
+    @classmethod
     @retry_db_transaction(name="object_delete")
     async def mutate_delete(
         cls,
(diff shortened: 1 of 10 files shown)


backend/infrahub/graphql/mutations/main.py
ajtmccarty, line 1: small refactoring to make handling a delete in the new `InfrahubProfileMutation` class DRYer

backend/infrahub/graphql/mutations/profile.py
```diff
@@ -0,0 +1,195 @@
...
+
+class InfrahubProfilesRefresh(Mutation):
+    class Arguments:
+        data = ProfilesRefreshInput(required=True)
+        context = ContextInput(required=False)
+
+    ok = Boolean()
+
+    @classmethod
+    @trace.get_tracer(__name__).start_as_current_span("profiles_refresh")
```
ajtmccarty, line 173: not sure if this is necessary. maybe you can say @fatih-acar ?

fatih-acar: If you believe this mutation can be slow sometimes yes, it's good to have extra spans :+1:

backend/infrahub/graphql/mutations/profile.py
```diff
@@ -0,0 +1,195 @@
...
+
+log = get_logger()
+
+
+class InfrahubProfileMutation(InfrahubMutationMixin, Mutation):
+    @classmethod
+    def __init_subclass_with_meta__(
+        cls,
+        schema: ProfileSchema,
+        _meta: Any | None = None,
```
ogenstad: Does this work here?

```python
_meta: InfrahubMutationOptions | None = None,
```

ogenstad (approved):

backend/infrahub/graphql/mutations/profile.py
ajtmccarty: looks like it does. made the change

PR #7299, opened 2025-10-01
Increase default timeout to 60 seconds for transforms.

FragmentedPacket:
We believe it may be good to increase the timeout of transforms from 10 seconds to 60 seconds which should give customers enough room for most artifact transforms.

## Summary by CodeRabbit

- **New Features**
  - Increased the default timeout for core transforms and checks from 10s to 60s. This change applies where defaults are used and may reduce premature timeouts for longer-running operations; any explicit/custom timeouts remain unchanged. Users should experience fewer interruptions during execution.

### DIFF
--- a/backend/infrahub/core/schema/definitions/core/check.py
+++ b/backend/infrahub/core/schema/definitions/core/check.py
@@ -29,7 +29,7 @@ core_check_definition = NodeSchema(
         Attr(name="description", kind="Text", optional=True),
         Attr(name="file_path", kind="Text"),
         Attr(name="class_name", kind="Text"),
-        Attr(name="timeout", kind="Number", default_value=10),
+        Attr(name="timeout", kind="Number", default_value=60),
         Attr(name="parameters", kind="JSON", optional=True),
     ],
     relationships=[
--- a/backend/infrahub/core/schema/definitions/core/transform.py
+++ b/backend/infrahub/core/schema/definitions/core/transform.py
@@ -29,7 +29,7 @@ core_transform = GenericSchema(
         Attr(name="name", kind="Text", unique=True),
         Attr(name="label", kind="Text", optional=True),
         Attr(name="description", kind="Text", optional=True),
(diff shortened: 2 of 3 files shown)


ogenstad:
I'm fine with this change and I think we can merge it. Having said that we also have other issues to refactor this timeout alltogether. I don't remember the issue id now, but the idea would be to mark all of these timeouts as optional and default to being optional. That would mean that if a user hasn't set a timeout we'd instead just use the timeoout of the SDK used within the prefect runner. One comment though I don't think the change for the Python Transform is enough, even if we update the default is ok for now if I don't remember incorrectly the change needs to happen here too: https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/infrahub_sdk/transforms.py#L20

Also we should have a newsfragment.

FragmentedPacket:
> I'm fine with this change and I think we can merge it. Having said that we also have other issues to refactor this timeout alltogether. I don't remember the issue id now, but the idea would be to mark all of these timeouts as optional and default to being optional. That would mean that if a user hasn't set a timeout we'd instead just use the timeoout of the SDK used within the prefect runner. One comment though I don't think the change for the Python Transform is enough, even if we update the default is ok for now if I don't remember incorrectly the change needs to happen here too: https://github.com/opsmill/infrahub-sdk-python/blob/v1.14.0/infrahub_sdk/transforms.py#L20
> 
> Also we should have a newsfragment.

I will open a PR today to increase those timeouts as well.

ogenstad (approved):
LGTM, however I think we should wait to merge this until the code from https://github.com/opsmill/infrahub-sdk-python/pull/563 also lives within the stable branch of Infrahub to avoid confusion.

I think we can target this for Infrahub 1.5.0. The alternative would be if we released a 1.14.1 version of the SDK and brought that commit into `stable`. The problem would be if we wanted to release a new 1.4.x version of Infrahub that didn't include that SDK commit.

PR #7311, opened 2025-10-03
migration for saving profile attributes on nodes

ajtmccarty:
IFC-1810

add migration 40 that runs `NodeProfilesApplier` on every node that needs it on each branch

    For any profile that has updates on a given branch (including default branch)
    - run NodeProfilesApplier.apply_profiles on each node related to the profile on that branch
    For any node that has an updated relationship to a profile on a given branch
    - run NodeProfilesApplier.apply_profiles on the node on that branch

TODO:

- [x] test on big database


backend/infrahub/core/migrations/graph/m040_profile_attrs_in_db.py
```diff
@@ -0,0 +1,165 @@
...
+    name = "get_nodes_with_profile_updates_by_branch"
+    type = QueryType.READ
+    insert_return = False
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        query = """
+MATCH (node:Node)-[e1:IS_RELATED]->(:Relationship {name: "node__profile"})
+WHERE NOT node:CoreProfile
+WITH DISTINCT node.uuid AS node_uuid, e1.branch AS branch
+RETURN node_uuid, collect(branch) AS branches
```
ogenstad, line 71: It doesn't look like we take deleted nodes into account in this or the previous query. But given that we use the NodeManager to process the results below I assume that we'd still only target anything that's active when it comes to the actual data that is migrated?

ogenstad (approved):

backend/infrahub/core/migrations/graph/m040_profile_attrs_in_db.py
ajtmccarty: good point. it's okay not to check for deleted edges in these queries, b/c we actually want to include deleted edges. for example, if a node is linked to two profiles and then one of the profiles is removed on a branch, we want to account for that and run the node through the `NodeProfilesApplier` on that branch

but I don't think this correctly handles the case when a node has been deleted, so I'll add a test for that

PR #7318, opened 2025-10-03
IFC-1824 Add migrations for display_label and HFID

gmazoyer:


backend/infrahub/core/migrations/graph/__init__.py
```diff
@@ -41,6 +41,7 @@
 from .m037_index_attr_vals import Migration037
 from .m038_redo_0000_prefix_fix import Migration038
 from .m039_ipam_reconcile import Migration039
+from .m040_create_hfid_display_label_in_db import Migration040
```
ogenstad: With https://github.com/opsmill/infrahub/pull/7311 in mind I think this should be migration 41 instead. Possible we'd have one migration 41 that updated the schema as a schema migration and then this migration with the logic for the nodes would be migration 42.

backend/infrahub/core/query/node.py
```diff
@@ -143,11 +143,13 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
         attributes_indexed: list[AttributeCreateData] = []
 
         if self.node.has_display_label():
-            display_label_attr = await self.node._display_label.get_node_attribute(node=self.node, at=at)
-            attributes_indexed.append(display_label_attr.get_create_data())
+            attributes_indexed.append(
+                self.node._display_label.get_node_attribute(node=self.node, at=at).get_create_data()
```
ajtmccarty, line 147: would be nice to avoid accessing `._display_label` and `._human_friendly_id` outside of the `Node` class

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -245,6 +246,12 @@ def get_field(self, name: str, raise_on_error: bool = True) -> AttributeSchema |
         return None
 
     def get_attribute(self, name: str) -> AttributeSchema:
+        # FIXME: Clean this up somehow
```
ajtmccarty, line 249: looks like this remains to be cleaned up somehow

backend/tests/unit/core/diff/test_diff_calculator.py
ajtmccarty, line 1: thanks for updating all of these. this must have been... not fun

backend/infrahub/core/migrations/graph/m041_create_hfid_display_label_in_db.py
```diff
@@ -0,0 +1,99 @@
...
+        schema_branch = await registry.schema.load_schema_from_db(db=db, branch=default_branch)
+
+        migrations = list(self.migrations)
+
+        for node_schema_kind in schema_branch.node_names:
+            schema = schema_branch.get_node(name=node_schema_kind, duplicate=False)
+            if not isinstance(schema, NodeSchema):
+                continue
+
+            if schema.human_friendly_id:
```
ajtmccarty: I would think that we want to add `human_friendly_id` and `display_label` to every `SchemaNode`  in the database even if the schema does not have a `human_friendly_id` or `display_label` defined. They would just be set to `NULL` or `[]` in that case

backend/infrahub/core/migrations/graph/m041_create_hfid_display_label_in_db.py
```diff
@@ -0,0 +1,99 @@
...
+                    NodeAttributeAddMigration(
+                        new_node_schema=schema,
+                        previous_node_schema=schema,
+                        schema_path=SchemaPath(
+                            schema_kind=schema.kind, path_type=SchemaPathType.ATTRIBUTE, field_name="display_label"
+                        ),
+                    )
+                )
+
+        for migration in migrations:
```
ajtmccarty, line 81: progress bar might be nice here too

backend/infrahub/core/migrations/graph/m042_backfill_hfid_display_label_in_db.py
```diff
@@ -0,0 +1,72 @@
...
+        result = MigrationResult()
+        # load schemas from database into registry
+        initialize_lock()
+        await initialization(db=db)
+
+        for node_schema in registry.get_full_schema(duplicate=False).values():
+            if node_schema.is_generic_schema:
+                continue
+
+            nodes: list[Node] = await NodeManager.query(db=db, schema=node_schema)
```
ajtmccarty, line 67: this could be 20,000 nodes
might want to get a count of the nodes to start with and then use limit and offset to page through them
getting a count of all the node schema before starting would also let you give a more accurate total progress bar that could show the expected time for updating _all_ nodes instead of showing the progress for all the node of a given kind and then resetting back to 0

backend/infrahub/core/query/node.py
ogenstad: I think we should continue using `._display_label` and `._human_friendly_id` like this for now. As the Node class is a mix of data and methods where the data aspect is dynamic. Then It would be nice with a refactoring where we can remove all methods from the `Node` class and instead only have it be a data container so that we don't have methods and properties on `Node` that might cause conflicts with the names of attributes and relationships. Specifically for these two it might be fine to have them as `.display_label` or `.human_friendly_id` directly.

This might fall under the work with the general refactoring with that `NodeModifier` class I mentioned.

backend/infrahub/core/migrations/graph/m041_create_hfid_display_label_in_db.py
ogenstad: This is a good point.

ogenstad:
I think it would be nice if we could merge this into `gma-20250910-ifc1820` and then fix Aarons comments within that branch. Those fixes should not impact the work I'm doing for the inline display_label -> display_labels migrations and would help to avoid that being a moving target with the tests.

PR #6393, opened 2025-05-01
Distribute Jinja2 computed attribute processing across workers

dgarros:

PR #7320, opened 2025-10-03
fix(backend): performance issues when generating artifacts

fatih-acar:
Many small changes here:

- enable caching of the git repositories by leveraging Prefect's cache policy, I set a hardcoded TTL of 30s to match the Git Sync scheduled job of 1 minute
- AI generated commit to improve fetching of artifact targets attributes/rels: we were previously fetching ALL attributes even though we don't need them
- batch the artifact generation workflow submits
- ensure the schema cache is set for the infrahub sdk client, avoids unnecessary calls to the API server to retrieve the schema on each task

We are now able to generate 1000 artifacts under 2 minutes without errors, whereas it would take more than five minutes with failures.

## Release Notes

* **Performance Improvements**
  * Artifact generation and proposed change checks now execute faster through optimized caching and reduced GraphQL queries.

* **Changes**
  * Query field is deprecated in artifact-related definitions; use `query_id` instead for future compatibility.


backend/infrahub/git/tasks.py
```diff
@@ -385,10 +468,16 @@ async def generate_request_artifact_definition(
             context=context,
         )
 
-        await get_workflow().submit_workflow(
-            workflow=REQUEST_ARTIFACT_GENERATE, context=context, parameters={"model": request_artifact_generate_model}
+        batch.add(
```
dgarros, line 390: is this batch really required ? it will create them faster in Prefect but it won't really speed up the execution

fatih-acar: you are right, it's similar to the discussion here: https://github.com/opsmill/infrahub/pull/6393#issuecomment-3193535718

it just allows shifting the bottleneck on the Prefect API servers (and then scale them out if necessary to achieve better performance)

backend/infrahub/core/schema/manager.py
```diff
@@ -40,6 +42,7 @@ class SchemaManager(NodeManager):
     def __init__(self) -> None:
         self._cache: dict[int, Any] = {}
         self._branches: dict[str, SchemaBranch] = {}
+        self._sdk_branches: dict[str, SDKBranchSchema] = {}
```
dgarros: How is that gonna impact the memory consumption ? is there an issue with converting the schema on the fly ? if that's too cpu intensive can we use a LRU cache instead ?

backend/infrahub/core/schema/manager.py
```diff
@@ -140,12 +143,16 @@ def get_schema_branch(self, name: str) -> SchemaBranch:
...
+        self.set_schema_branch(name, schema=SchemaBranch(cache=self._cache, name=name))
         return self._branches[name]
 
+    def get_sdk_schema_branch(self, name: str) -> SDKBranchSchema:
+        return self._sdk_branches[name]
+
     def set_schema_branch(self, name: str, schema: SchemaBranch) -> None:
         schema.name = name
         self._branches[name] = schema
+        self._sdk_branches[name] = SDKBranchSchema.from_api_response(data=schema.to_dict_api_schema_object())
```
ogenstad: I'm not sure we want to have this globally i.e. on API workers we shouldn't need to care about the schema from an API point of view. It could be that it would impact some tests, but there's a function called `get_component_type()` that we could use here and only set the SDK schema for the Prefect workers (currently `ComponentType.GIT_AGENT` we should change that name).

Also `SchemaManager` has a method called `.purge_inactive_branches()` , in that method we should clear out this key when applicable.

Finally, I'm not sure about the size of these SDK schemas. I think they are a lot smaller than the GraphQL schemas but in an environment with a high number of branches this would consume a lot of extra memory. As everything is passed as dictionaries each entry would be their own objects instead of a shared reference. You can probably determine the memory footprint of this yourself and it might be fine for now. But I think we should consider storing the branches within the SDK based on the hash of the schema just like we started to do for the GraphQL schema. (This of course applies to the other branches within SchemaManager `self._branches: dict[str, SchemaBranch] = {}` but this is code we already have in place and won't worsen the situation). It could be that the memory footprint is fine and we can tackle that in an upcoming PR, but would be good with a sanity check.

backend/infrahub/core/schema/manager.py
fatih-acar: Converting the schema on the fly (like done in a previous version of this PR) leads to intensive CPU and memory usage due to pydantic conversion. It looks like the GC is having a bad time freeing memory due to cyclic references in the SDK client here: https://github.com/opsmill/infrahub-sdk-python/blob/stable/infrahub_sdk/client.py#L298

My test scenario was to run a artifact definition generation creating 1000 "Generate artifact" subtasks:

- when converting the schema on the fly, besides CPU usage, task worker memory goes from 500MB to 2GB, and stays at 2GB without ever releasing memory
- with this version, memory stays at 600MB steady during the execution

I think a LRU cache is only useful if you have a lot of branches and a fixed-size cache. We don't need that here.

backend/infrahub/core/schema/manager.py
fatih-acar: Good catch for `purge_inactive_branches`, thanks!

Regarding using `get_component_type()`, as you mentioned it requires some changes to the tests and I'm not sure if it's worth the change... we can keep that cache global to API workers and task workers even though API workers won't use it, the trade off looks fine to me...

fatih-acar:
There's a follow up PR related to memory usage when generating artifacts: https://github.com/opsmill/infrahub/pull/7327

backend/infrahub/core/schema/manager.py
fatih-acar: Just saw your edit: I agree with this, I think we can revisit this later as we would have to do the same change to `self._branches: dict[str, SchemaBranch] = {}` at some point. Memory footprint is way better like this than the alternative that regenerates the schema on the fly...

ogenstad: Works for me if you think it's fine. Just wanted to highlight the issue.

dgarros: I'm really not confident about this change, it will again increase the memory if we have a large number of branches.
We should be able to use a LRU cache here with the hash_id of the schema to identify when we need to regenerate the schema or not

fatih-acar: @dgarros are you looking for something like this? https://github.com/opsmill/infrahub/pull/7320/commits/9dcedfc5fa8e8bc1da6270604ea3c8c43cf453b2

I think this makes sense, yes.
Especially in this scenario where we know that regenerating artifacts in parallel will likely happen in a single branch.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -162,6 +162,14 @@ def to_dict_schema_object(self, duplicate: bool = False) -> dict[str, dict[str,
             "templates": {name: self.get(name, duplicate=duplicate) for name in self.templates},
         }
 
+    def to_dict_api_schema_object(self, duplicate: bool = False) -> dict[str, list[dict]]:
```
ajtmccarty: I don't think you need `duplicate` here if all the output is going through `.model_dump`

backend/infrahub/artifacts/models.py
```diff
@@ -25,7 +25,8 @@ class CheckArtifactCreate(BaseModel):
     target_kind: str = Field(..., description="The kind of the target object for this artifact")
     target_name: str = Field(..., description="Name of the artifact target")
     artifact_id: str | None = Field(default=None, description="The id of the artifact if it previously existed")
-    query: str = Field(..., description="The name of the query to use when collecting data")
+    query: str = Field(..., description="The name of the query to use when collecting data")  # Deprecated
+    query_id: str = Field(..., description="The name of the query to use when collecting data")
```
ajtmccarty: should it be "The id of the query"?

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+        batch.add(task=client.filters, **kwargs)
+
+    async for _ in batch.execute():
+        pass
+
+
+async def fetch_definition_targets(
+    client: InfrahubClient,
+    branch: str,
+    definition: CoreArtifactDefinition | CoreCheckDefinition | ProposedChangeGeneratorDefinition,
```
ajtmccarty: should the typing here include `InfrahubNode`?

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+        if keep_relationships:
+            kwargs["include"] = list(keep_relationships)
+
+        batch.add(task=client.filters, **kwargs)
+
+    async for _ in batch.execute():
+        pass
+
+
+async def fetch_definition_targets(
```
ajtmccarty: this feels like it should be several public functions for different types of `definition` that all call a `_fetch_definition_targets` function with the correct `group_id` and `parameters`

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+                _walk(nested)
+
+    _walk(params)
+    return segments
+
+
+async def _prefetch_group_member_nodes(
+    *,
+    client: InfrahubClient,
+    members: Any,
```
ajtmccarty: can this be typed as the SDK's `RelationshipManager` class?

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+    batch = await client.create_batch()
+
+    for kind, ids in ids_per_kind.items():
+        schema = await client.schema.get(kind=kind, branch=branch)
+
+        keep_attributes = {field for field in required_fields if field in schema.attribute_names}
+        keep_relationships = {field for field in required_fields if field in schema.relationship_names}
+
+        exclude_attributes = [attr for attr in schema.attribute_names if attr not in keep_attributes]
+        exclude_relationships = [rel for rel in schema.relationship_names if rel not in keep_relationships]
```
ajtmccarty: I'd do this b/c `required_fields` is already a `set`
```
valid_fields = set(schema.attribute_names) | set(schema.relationship_names)
cleaned_fields = required_fields & valid_fields
```
then you can get rid of these 4 lists and use `cleaned_fields` in the batched call to `client.filters`

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+
+        unique_ids = list(dict.fromkeys(ids))
```
ajtmccarty: I'd say make `ids_per_kind` a `dict[str, set[str]]` and then you can get rid of this weird deduplication

ajtmccarty (approved):
approving b/c it sounds like you need this, but I think whatever LLM wrote the functions in `infrahub.git.utils` did not do a good job

backend/infrahub/git/repository.py
```diff
@@ -248,12 +251,13 @@ async def sync_from_remote(self, commit: str | None = None) -> None:
         await self.update_commit_value(branch_name=self.infrahub_branch_name, commit=commit)
 
 
-@task(
-    name="Fetch repository commit",
-    description="Retrieve a git repository at a given commit, if it does not already exist locally",
-    cache_policy=NONE,
+@cached(
```
dgarros, line 254: Not a blocker for this PR but I'm wondering if we could leverage prefect for the caching ? and what is the limitation if you already tried

pyproject.toml
```diff
@@ -76,6 +76,8 @@ dulwich = "^0.22.7"
 whenever = "0.7.3"
 netutils = "1.12.0"
 copier = "^9.8.0"
+cachetools-async = "^0.0.5"
+types-cachetools = "^6.2.0.20250827"
```
dgarros: This should go in `[tool.poetry.group.dev.dependencies]`

pyproject.toml
```diff
@@ -76,6 +76,8 @@ dulwich = "^0.22.7"
 whenever = "0.7.3"
 netutils = "1.12.0"
 copier = "^9.8.0"
+cachetools-async = "^0.0.5"
```
dgarros: This should go up in the first section since it's not specific to the SDK

dgarros (approved):
Please can you add a fragment to include something in the release note

backend/infrahub/core/schema/schema_branch.py
fatih-acar: Indeed, thanks!

backend/infrahub/artifacts/models.py
fatih-acar: Good catch!

backend/infrahub/git/utils.py
fatih-acar: Yes, but I changed everything in favor of your suggestion below...

backend/infrahub/git/utils.py
fatih-acar: Yes much cleaner

backend/infrahub/git/utils.py
fatih-acar: Indeed, bad LLM...

backend/infrahub/git/utils.py
fatih-acar: Nice one! LLM was not clever enough to figure this out...
But I think we want a XOR ^ operation instead here for the exclude.
Also I added an inline comment about https://github.com/opsmill/infrahub-sdk-python/pull/205 that should avoid  having to do this in the future...

backend/infrahub/git/utils.py
fatih-acar: Good catch!

backend/infrahub/git/repository.py
fatih-acar: It was done using the Prefect caching mechanism in the first iteration, but as described in the following commit: https://github.com/opsmill/infrahub/pull/7320/commits/02b289d4f8f77dd976721c08e5883c8314cde527

```Unfortunately we cannot leverage Prefect's caching because:
- local caching requires a lot of setup (wrt Blocks, etc)
- it's memory intensive due to excessive pickle serialization in our case (we don't need it)

Revert to basic local TTL cache.```

Comment (fatih-acar) [reply backend/infrahub/git/repository.py:254]: Basically Prefect caching is useful for distributed caching but here we want a local, efficient cache.

File: pyproject.toml (hunk 12 above)
Comment (fatih-acar) [reply pyproject.toml]: Good catch thanks!

File: backend/infrahub/git/utils.py (hunk 9 above)
Comment (fatih-acar) [reply backend/infrahub/git/utils.py]:
```
cleaned_fields = valid_fields - required_fields
```
is even better according to coderabbit, so I will use that :smile:

File: backend/infrahub/git/repository.py [hunk 14]
```diff
@@ -248,12 +251,13 @@ async def sync_from_remote(self, commit: str | None = None) -> None:
...
+    TTLCache(maxsize=100, ttl=30),
+    key=lambda *_, **kwargs: hashkey(
+        kwargs.get("repository_id"), kwargs.get("name"), kwargs.get("repository_kind"), kwargs.get("commit")
+    ),
```
ogenstad, line 258: Given the TTL it might not have a practical meaning now, but given that commit can be `None` I think we might occasionally end up with a repo using an earlier commit that we would have expected. In another iteration we could instead cache after we have the commit. Repository_id and commit is probably enough to use for the key.

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,118 @@ async def get_repositories_commit_per_branch(
             )
 
     return repositories
+
+
+def _collect_parameter_first_segments(params: Any) -> set[str]:
```
ogenstad, line 58: Do we need an `Any` here looks like we expect `str | list | tuple | set | dict` and won't use anything else.

backend/infrahub/git/utils.py
```diff
@@ -46,3 +53,103 @@ async def get_repositories_commit_per_branch(
...
+                segments.add(segment)
+        elif isinstance(value, dict):
+            for nested in value.values():
+                _walk(nested)
+        elif isinstance(value, (list, tuple, set)):
+            for nested in value:
+                _walk(nested)
+
+    _walk(params)
+    return segments
```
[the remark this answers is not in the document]
ogenstad: Hey coderabbit, where is this requirement for docstring coming from?

ogenstad (approved):

backend/infrahub/git/utils.py
fatih-acar: I think LLM used `Any` here because we pass `parameters` straight from the `JSONAttribute` protocol where `value` is typed as `Any`. This looks legit?

backend/infrahub/git/repository.py
fatih-acar: The TTL is lower than the git sync scheduled job on purpose because I'd expect this kind of edge case... I think it's fine like that for the moment.
+1 for the key.

Issue #6397, opened 2025-05-02
bug: Can't unselect node kind for webhook


BaptisteGi:
### Component

Frontend UI

### Infrahub version

1.2.7

### Current Behavior

If I select a `node kind` for a webhook I can't unselect it later. I need to delete the webhook and create it again. This is particularly impacting due to https://github.com/opsmill/infrahub/issues/6325.

### Expected Behavior

As the attribute is optional I should be able to put it back to empty.

### Steps to Reproduce

- Navigate to webhook
- Create a webhook and specify a node kind
- Then edit the webhook
=> You can unselect but then it doesn't allow you to save
=> You can edit another field to unlock the save button but value in node kind is ignored

### Additional Information

- The current process to empty the field (search selected value to click on it) isn't intuitive. We should find something better (have the selected value at the top, a button to empty the field ...)

ogenstad:
I think this might actually be a backend issue.

Possible fix:

```diff
diff --git a/backend/infrahub/graphql/mutations/webhook.py b/backend/infrahub/graphql/mutations/webhook.py
index 66880446e..fb96ddf51 100644
--- a/backend/infrahub/graphql/mutations/webhook.py
+++ b/backend/infrahub/graphql/mutations/webhook.py
@@ -127,7 +127,7 @@ class InfrahubWebhookMutation(InfrahubMutationMixin, Mutation):


 def _validate_input(graphql_context: GraphqlContext, branch: Branch, input_data: WebhookCreate) -> None:
-    if input_data.node_kind:
+    if input_data.node_kind and input_data.node_kind.value:
         # Validate that the requested node_kind exists, will raise an error if not
         graphql_context.db.schema.get(name=input_data.node_kind.value, branch=branch, duplicate=False)
```

PR #7322, opened 2025-10-06
IFC-1496: Check optional node kind value

solababs:
Fixes https://github.com/opsmill/infrahub/issues/6397

## Summary by CodeRabbit

* **Bug Fixes**
  * Updating or creating a webhook no longer fails when the node type field is omitted or explicitly cleared; empty node kind is now treated as optional to avoid unnecessary validation errors.

* **Tests**
  * Added a test verifying webhooks can be created and updated with the node type removed without errors.

* **Documentation**
  * Added a changelog entry documenting the optional node kind behavior for webhook create/update.

### DIFF
--- a/backend/tests/unit/graphql/mutations/test_webhook.py
+++ b/backend/tests/unit/graphql/mutations/test_webhook.py
@@ -82,6 +82,9 @@ async def test_update_webhook_with_optional_node_kind(
     assert not result.errors
     assert result.data
     webhook_id = result.data["CoreStandardWebhookCreate"]["object"]["id"]
+    webhook = await NodeManager.get_one_by_id_or_default_filter(db=db, id=webhook_id, kind=CoreStandardWebhook)
+    assert webhook.name.value == "builtin-tag-update-1"
+    assert webhook.node_kind.value == "BuiltinTag"
 
     result = await graphql(
         schema=gql_params.schema,
@@ -96,7 +99,7 @@ async def test_update_webhook_with_optional_node_kind(
     assert not result.errors
     assert result.data
     updated_webhook = await NodeManager.get_one_by_id_or_default_filter(db=db, id=webhook_id, kind=CoreStandardWebhook)
-    assert not updated_webhook.node_kind.value
+    assert updated_webhook.node_kind.value is None
 
 
 async def test_create_webhook_with_node_kind_and_all_events(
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
LGTM, can you please also add a newsfragment that links to the GitHub issue and indicate what bug was fixed?

changelog/6397.fixed.md
```diff
@@ -0,0 +1 @@
+Allow optional node kind value
```
bilalabbad: It’s a bit too vague, let’s make it explicit that it's a bugfix that allow updating  the node kind to null in webhook

backend/tests/unit/graphql/mutations/test_webhook.py
```diff
@@ -68,6 +68,37 @@ async def test_create_webhook_with_node_kind_and_valid_node_event(
...
+        root_value=None,
+        variable_values={
+            "id": webhook_id,
+            "node_kind": None,
+        },
+    )
+    assert not result.errors
+    assert result.data
+    updated_webhook = await NodeManager.get_one_by_id_or_default_filter(db=db, id=webhook_id, kind=CoreStandardWebhook)
+    assert not updated_webhook.node_kind.value
```
bilalabbad: ` assert updated_webhook.node_kind.value is None`

 make it more explicit about the intent

backend/tests/unit/graphql/mutations/test_webhook.py
```diff
@@ -68,6 +68,37 @@ async def test_create_webhook_with_node_kind_and_valid_node_event(
...
+    result = await graphql(
+        schema=gql_params.schema,
+        source=CREATE_WEBHOOK,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"name": "builtin-tag-update-1", "node_kind": "BuiltinTag", "event_type": "all"},
+    )
+    assert not result.errors
+    assert result.data
+    webhook_id = result.data["CoreStandardWebhookCreate"]["object"]["id"]
```
bilalabbad, line 84: I would be nice to add
```
webhook = await NodeManager.get_one_by_id_or_default_filter(db=db, id=webhook_id, kind=CoreStandardWebhook)
assert webhook.name.value == "builtin-tag-update-1"
assert webhook.node_kind.value == "BuiltinTag"
```
to see explicitly the before/after update

bilalabbad (approved):

PR #7327, opened 2025-10-06
fix(backend): re-use tls contexts where possible

fatih-acar:
- On task workers when using the SDK client
- On task workers when querying Prefect API
- On api/task workers when executing Prefect tasks/flows locally
- On api workers when submitting workflows

Note: this depends on https://github.com/PrefectHQ/prefect/pull/19106 for Prefect client changes to be effective

We have seen that creating SSL/TLS contexts are resource intensive (CPU and memory).

EDIT: sdk client change is now handled directly by https://github.com/opsmill/infrahub-sdk-python/pull/581

## Summary by CodeRabbit

- Bug Fixes
  - TLS verification is now consistently applied during workflow deployments and async execution, reducing HTTPS-related failures.

- Refactor
  - Centralizes HTTP/TLS configuration and uses a shared async HTTP client across tasks and flows to improve stability and reduce overhead.

- Chores
  - Adds a configurable TLS verification interface to standardize how verification settings are provided and consumed.

### DIFF
--- a/backend/infrahub/workflows/utils.py
+++ b/backend/infrahub/workflows/utils.py
@@ -9,6 +9,7 @@ from prefect.runtime import flow_run
 from infrahub.core.constants import GLOBAL_BRANCH_NAME
 from infrahub.core.registry import registry
 from infrahub.tasks.registry import refresh_branches
+from infrahub.workers.dependencies import get_http
 
 from .constants import TAG_NAMESPACE, WorkflowTag
 
@@ -26,7 +27,7 @@ async def add_tags(
     namespace: bool = True,
     db_change: bool = False,
 ) -> None:
-    client = get_client(sync_client=False)
+    client = get_client(httpx_settings={"verify": get_http().verify_tls()}, sync_client=False)
     current_flow_run_id = flow_run.id
     current_tags: list[str] = flow_run.tags
     branch_tags = (
--- a/backend/infrahub/services/adapters/http/__init__.py
+++ b/backend/infrahub/services/adapters/http/__init__.py
@@ -3,10 +3,15 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
+    import ssl
+
     import httpx
 
 
 class InfrahubHTTP:
+    def verify_tls(self, verify: bool | None = None) -> bool | ssl.SSLContext:
+        raise NotImplementedError()
+
     async def get(
         self,
         url: str,
--- a/backend/infrahub/services/adapters/workflow/worker.py
+++ b/backend/infrahub/services/adapters/workflow/worker.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, overload
 
 from prefect.client.schemas.objects import StateType
+from prefect.context import AsyncClientContext
 from prefect.deployments import run_deployment
 
+from infrahub.services.adapters.http.httpx import HttpxAdapter
 from infrahub.workers.utils import inject_context_parameter
 from infrahub.workflows.initialization import setup_task_manager
 from infrahub.workflows.models import WorkflowInfo
@@ -19,6 +21,11 @@ if TYPE_CHECKING:
 
 
 class WorkflowWorkerExecution(InfrahubWorkflow):
+    # This is required to grab a cached SSLContext from the HttpAdapter.
+    # We cannot use the get_http() dependency since it introduces a circular dependency.
+    # We could remove this later on by introducing a cached SSLContext outside of this adapter.
+    _http_adapter = HttpxAdapter()
+
     @staticmethod
     async def initialize(component_is_primary_server: bool) -> None:
         if component_is_primary_server:
@@ -79,5 +86,6 @@ class WorkflowWorkerExecution(InfrahubWorkflow):
         parameters = dict(parameters) if parameters is not None else {}
         inject_context_parameter(func=flow_func, parameters=parameters, context=context)
 
-        flow_run = await run_deployment(name=workflow.full_name, timeout=0, parameters=parameters or {}, tags=tags)  # type: ignore[return-value, misc]
+        async with AsyncClientContext(httpx_settings={"verify": self._http_adapter.verify_tls()}):
+            flow_run = await run_deployment(name=workflow.full_name, timeout=0, parameters=parameters or {}, tags=tags)  # type: ignore[return-value, misc]
         return WorkflowInfo.from_flow(flow_run=flow_run)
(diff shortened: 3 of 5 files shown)


fatih-acar:
@ogenstad is that good use of the http adapter? :smile:

ogenstad:
> @ogenstad is that good use of the http adapter? 😄

Well, good find with the performance issue :) It was perhaps not how I envisioned it being used, but I don't know if I have a strong objection to it. Possibly `self._http_adapter.verify_tls()` could live somewhere else, or rather if verify_tls() was more of a generic function perhaps we wouldn't have to inject the HTTP adapter to Prefect as we're not really using it aside from the TLS caching. But it's perhaps fine.

So the real reason why I wanted this to be cached was actually during application startup.

```python
    def set_tls_context(self) -> Self:
        try:
            # Validate that the context can be created, we want to raise this error during application start
            # instead of running into issues later when we first try to use the tls context.
            self.get_tls_context()
        except ssl.SSLError as exc:
            raise ValueError(f"Unable load CA bundle from {self.tls_ca_bundle}: {exc}") from exc

        return self
```

I wanted the above to produce an error if you tried to load a custom CA bundle that didn't exist as it would be easier to spot during startup then have it happen some time later when you wanted to do something with HTTP.

But were you using a CA bundle or was it just any TLS call that benefitted from not having to recreate the context?

For the SDK. I also did not have that in mind when I added the ability to provide your custom transform but I like that it works :) I'm wondering though if it might cause any confusion with regards to TLS settings in the SDK if users have environment variables for the SDK that would set this to a different value. I.e. the TLS options in the SDK would be ignored which can be confusing?

fatih-acar:
> 
> Possibly `self._http_adapter.verify_tls()` could live somewhere else, or rather if verify_tls() was more of a generic function perhaps we wouldn't have to inject the HTTP adapter to Prefect as we're not really using it aside from the TLS caching. But it's perhaps fine.
> 

Indeed, I only need the cached SSLContext thus I tried to reuse the current http adapter as much as possible. It can probably live somewhere else as you mentioned but don't know if it's worth the effort.

> 
> But were you using a CA bundle or was it just any TLS call that benefitted from not having to recreate the context?
> 

The thing is that SSL contexts are ALWAYS created even if you do not use TLS at all... This is the case for both the SDK client and the Prefect client. The only way to disable creating a SSL context for the SDK client is to disable TLS verification. For the Prefect Client there's no way to avoid the context creation even when TLS is not used anywhere... (thus the upstream PR I opened).

> For the SDK. I also did not have that in mind when I added the ability to provide your custom transform but I like that it works :) I'm wondering though if it might cause any confusion with regards to TLS settings in the SDK if users have environment variables for the SDK that would set this to a different value. I.e. the TLS options in the SDK would be ignored which can be confusing?

Glad you mention the part I'm not convinced of... I believe this can be a breaking change. I tried to move the `timeout` parameter into the HTTP adapter so we can use timeouts either configured in the SDK or passed to SDK calls. But I won't be able to do this for all the SDK client config...
The other option would be to allow the SDK client to grab a pre-built SSL context in that case, thus requiring a change in the SDK (which would be safer?). But I liked using the HTTP adapter for the SDK using the `requester` wire, looked elegant :)

ogenstad:
> Indeed, I only need the cached SSLContext thus I tried to reuse the current http adapter as much as possible. It can probably live somewhere else as you mentioned but don't know if it's worth the effort.

I think an inline comment explaining why it's there and mention that if we're doing a refactor we could separate the SSLContext part out to some other shared location.

> The thing is that SSL contexts are ALWAYS created even if you do not use TLS at all... This is the case for both the SDK client and the Prefect client. The only way to disable creating a SSL context for the SDK client is to disable TLS verification. For the Prefect Client there's no way to avoid the context creation even when TLS is not used anywhere... (thus the upstream PR I opened).

Based on this I'd say that we should implement a fix for this directly in the SDK so that we always create an SSL context that is cached within the SDK config. That way anyone could benefit from it not just internally when used within Infrahub. I.e. when someone runs a generator outside of Infrahub or just uses the SDK we'd have this in place already. Then what we can do is to cache the first SSLContext somewhere reached by `build_client()` so that we don't have to recreate the SDK context each time. Or provide a way to set the SSLContext from Infrahub into the SDK. I.e. a method within config.py in the SDK.

How does that sound?

fatih-acar:
> 
> Based on this I'd say that we should implement a fix for this directly in the SDK so that we always create an SSL context that is cached within the SDK config. That way anyone could benefit from it not just internally when used within Infrahub. I.e. when someone runs a generator outside of Infrahub or just uses the SDK we'd have this in place already. Then what we can do is to cache the first SSLContext somewhere reached by `build_client()` so that we don't have to recreate the SDK context each time. Or provide a way to set the SSLContext from Infrahub into the SDK. I.e. a method within config.py in the SDK.
> 
> How does that sound?

Sounds good, I think we can do both: create a cached SSL context within the SDK (for the SDK end-users) and allow passing an external SSLContext to the SDK client (for the tasks workers).

fatih-acar:
Pushed https://github.com/opsmill/infrahub-sdk-python/pull/581 for the SDK change
I removed the previous commit for the SDK client change within this PR (it's not required anymore).

I also added an inline comment for the HttpAdapter usage.

backend/infrahub/workflows/utils.py
```diff
@@ -26,7 +27,7 @@ async def add_tags(
     namespace: bool = True,
     db_change: bool = False,
 ) -> None:
-    client = get_client(sync_client=False)
+    client = get_client(httpx_settings={"verify": get_http().verify_tls()}, sync_client=False)
```
ogenstad, line 30: It's a bit confusing in the codebase that we use `get_client()`  where we sometimes want an InfrahubClient and in other locations we want a Prefect client. Did you just update this call as a hotspot where you needed this the most? Wondering if we should have a helper function to get the prefect client where we can update these settings in one location. I think we're good to go for this PR but that we should consider the approach for this for all the other calls.

ogenstad (approved):
LGTM, did we also need a followup PR to allow the use of the Infrahub TLS context when we initialize a new SDK client?

I don't think this is supported in the Prefect version we have in `stable`, but I think you also said that it would just be silently ignored until we have the correct version.

fatih-acar:
> LGTM, did we also need a followup PR to allow the use of the Infrahub TLS context when we initialize a new SDK client?
> 
> I don't think this is supported in the Prefect version we have in `stable`, but I think you also said that it would just be silently ignored until we have the correct version.

Yes, follow up PR is here: https://github.com/opsmill/infrahub/pull/7396 (waiting for an SDK update)

Yes, the changes here will be ignored in stable until Prefect is upgraded (which is the case in develop/release-1.5).

backend/infrahub/workflows/utils.py
fatih-acar: > Did you just update this call as a hotspot where you needed this the most?

Exactly, `add_tags()` is called within almost all workflows.
I agree it's confusing to have `get_client` for both InfrahubClient and PrefectClient...

PR #7339, opened 2025-10-07
add profile support for schema changes that add or remove attributes

ajtmccarty:
IFC-1887
IFC-1897
`ProfileSchema`s are dynamically generated when a schema is updated and are never saved to the database, so we do not need to worry about migrating `ProfileSchema` when the associated `Node` or `GenericSchema`s are updated.
HOWEVER, we have not been correctly handling existing `Profile` **objects** when the profile's associated `Node` or `Generic` schema is updated.  when an optional (and `read_only=False`) attribute is added, an attribute is updated to become `optional=True` and `read_only=False`, then we need to add that attribute to any existing `Profile` objects for that schema so that it can be used. we also need to handle the reverse case, when an attribute is removed from a `Node`/`Generic` schema and when an attribute is updated to be `optional=False` or `read_only=True`.

handling these cases required updating the existing schema migrations for adding and deleting attributes and adding new schema migrations for changes to `optional` and `read_only` properties.

TODO:
 - [x] add profile based on generic to test case
 - [x] test large-type attribute

### DIFF
--- a/backend/infrahub/core/migrations/query/__init__.py
+++ b/backend/infrahub/core/migrations/query/__init__.py
@@ -8,7 +8,12 @@ if TYPE_CHECKING:
     from ..shared import AttributeSchemaMigration, SchemaMigration
 
 
-class MigrationQuery(Query):
+class MigrationBaseQuery(Query):
+    def get_nbr_migrations_executed(self) -> int:
+        return self.num_of_results
+
+
+class MigrationQuery(MigrationBaseQuery):
     type: QueryType = QueryType.WRITE
 
     def __init__(
@@ -19,11 +24,8 @@ class MigrationQuery(Query):
         self.migration = migration
         super().__init__(**kwargs)
 
-    def get_nbr_migrations_executed(self) -> int:
-        return self.num_of_results
 
-
-class AttributeMigrationQuery(Query):
+class AttributeMigrationQuery(MigrationBaseQuery):
     type: QueryType = QueryType.WRITE
 
     def __init__(
@@ -33,6 +35,3 @@ class AttributeMigrationQuery(Query):
     ):
         self.migration = migration
         super().__init__(**kwargs)
-
-    def get_nbr_migrations_executed(self) -> int:
-        return self.num_of_results
--- a/backend/infrahub/core/migrations/query/attribute_add.py
+++ b/backend/infrahub/core/migrations/query/attribute_add.py
@@ -17,14 +17,14 @@ class AttributeAddQuery(Query):
 
     def __init__(
         self,
-        node_kind: str,
+        node_kinds: list[str],
         attribute_name: str,
         attribute_kind: str,
         branch_support: str,
         default_value: Any | None = None,
         **kwargs: Any,
     ) -> None:
-        self.node_kind = node_kind
+        self.node_kinds = node_kinds
         self.attribute_name = attribute_name
         self.attribute_kind = attribute_kind
         self.branch_support = branch_support
@@ -36,7 +36,7 @@ class AttributeAddQuery(Query):
         branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
         self.params.update(branch_params)
 
-        self.params["node_kind"] = self.node_kind
+        self.params["node_kinds"] = self.node_kinds
         self.params["attr_name"] = self.attribute_name
         self.params["branch_support"] = self.branch_support
         self.params["current_time"] = self.at.to_string()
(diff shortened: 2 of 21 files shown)


backend/infrahub/core/migrations/query/__init__.py
ajtmccarty, line 1: to make typing more correct

backend/infrahub/core/migrations/query/attribute_add.py
```diff
@@ -17,14 +17,14 @@ class AttributeAddQuery(Query):
 
     def __init__(
         self,
-        node_kind: str,
+        node_kinds: list[str],
```
ajtmccarty, line 20: so we can run the same updates on instances of multiple schemas at the same time. such as when we're updating a generic's profile and any node profiles that inherit from the generic's schema

backend/infrahub/core/migrations/query/attribute_remove.py
ajtmccarty, line 1: relocating the query to be reused.
also updates the `node_kind` argument to `node_kinds` and adds the `profile_kinds_to_update` variable

backend/infrahub/core/migrations/schema/attribute_supports_profile.py
```diff
@@ -0,0 +1,90 @@
...
+            self.previous_attribute_schema.support_profiles == self.new_attribute_schema.support_profiles
+            # the attribute is new, so there cannot be existing profiles to update
+            or self.previous_attribute_schema.id is None
+        ):
+            return MigrationResult()
+        profiles_queries: list[type[AttributeMigrationQuery]] = []
+        if self.new_attribute_schema.support_profiles:
+            profiles_queries.append(ProfilesAttributeAddMigrationQuery)
+        if not self.new_attribute_schema.support_profiles:
+            profiles_queries.append(ProfilesAttributeRemoveMigrationQuery)
```
ajtmccarty, line 88: this is a new pattern that we have not used in SchemaMigrations before.
here we determine which query to run based on the changes to the schema. we add an attribute to profile schema(s) if it `supports_profile` changes from `False` to `True` and we remove an attribute in the opposite case

backend/infrahub/core/migrations/schema/attribute_supports_profile.py
```diff
@@ -0,0 +1,90 @@
...
+    from infrahub.core.timestamp import Timestamp
+    from infrahub.database import InfrahubDatabase
+
+
+def _get_node_kinds(schema: MainSchemaTypes) -> list[str]:
+    if not isinstance(schema, (NodeSchema, GenericSchema)):
+        return [schema.kind]
+    schema_kinds = [f"Profile{schema.kind}"]
+    if isinstance(schema, NodeSchema) or not schema.used_by:
+        return schema_kinds
```
ogenstad: Nitpicking but I think we can get rid of the `schema_kinds` variable and return `[f"Profile{schema.kind}"]` directly, it makes it a bit more clear that the returned value is only used within the if-statement.

backend/infrahub/core/migrations/schema/node_attribute_remove.py
```diff
@@ -1,123 +1,27 @@
...
 from ..shared import AttributeSchemaMigration
 
-if TYPE_CHECKING:
-    from pydantic.fields import FieldInfo
-
-    from infrahub.database import InfrahubDatabase
-
 
-class NodeAttributeRemoveMigrationQuery01(AttributeMigrationQuery):
+class NodeAttributeRemoveMigrationQuery01(AttributeMigrationQuery, AttributeRemoveQuery):
```
ogenstad, line 10: I think we in general should be careful with adding additional mixin-classes. Can be fine if we don't have any overlapping methods in them, but I think it generally leads to confusion later on.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2255,7 +2255,7 @@ def generate_profile_from_node(self, node: NodeSchema) -> ProfileSchema:
         )
 
         for node_attr in node.attributes:
-            if node_attr.read_only or node_attr.optional is False:
+            if not node_attr.support_profiles:
```
ogenstad, line 2290: This is a change within the logic as per what we had before:

```python
    @property
    def support_profiles(self) -> bool:
        return self.read_only is False and self.optional is True
```

I.e. we have `and` where we previously had `or`, was that intentional?

ogenstad (approved):

backend/infrahub/core/migrations/schema/attribute_supports_profile.py
ajtmccarty: there was actually a bug in here unrelated to your nitpick
if the schema being updated was a generic schema, then this would just return a list of Profile schemas for the inheriting schemas, but would not include the Profile of the generic schema
so if the input was `NetworkGenericDevice`, then the output would have been `[ProfileNetworkRealDevice]` instead of `[ProfileNetworkGenericDevice, ProfileNetworkRealDevice]`
I've updated the tests to catch the issue too
so this was a good nit to pick

backend/infrahub/core/migrations/schema/node_attribute_remove.py
ajtmccarty: good point and I agree. we do this in a few other places with `AttributeMigrationQuery`, so I was following the existing pattern, which isn't the greatest excuse. but I think it's working for now, so I'm inclined to leave it

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: yes
the conditional on this line used to be checking if an attribute did NOT support profiles, but the new property is to check if an attribute DOES support profiles, so the inversion is intentional
I did have to envision a truth table though

ajtmccarty (commented):
thanks for the review, your "nitpick" actually uncovered a bug

PR #7341, opened 2025-10-08
fix: small memory/cpu optimizations

fatih-acar:
Measured a rough 10% speedup in my test case with those, with lesser cpu/memory usage.

- missed a duplicate=False

- for the gzip change, here's the commit message:

> During testing, we noticed that all graphql queries were compressed even
> though the size of the response was less than the defined 100kB size.
> It seems compression is always enabled for streamed response, thus
> compressing almost ALL requests toward APIs.
> 
> Introduce a new middleware that enables compression only on specific,
> useful paths.

Also change default compression level to 1 (instead of the default 9) for better CPU usage/compression ratio

## Summary by CodeRabbit

- New Features
  - Selective HTTP compression for specified API paths (path-based inclusion).
  - Configurable compression level and minimum size for compressed responses.

- Bug Fixes
  - Fixed GraphQL inline fragment handling to prevent unintended duplication and ensure consistent query results.

### DIFF
--- a/backend/infrahub/graphql/analyzer.py
+++ b/backend/infrahub/graphql/analyzer.py
@@ -639,7 +639,7 @@ class InfrahubGraphQLQueryAnalyzer(GraphQLQueryAnalyzer):
         self, node: InlineFragmentNode, query_node: GraphQLQueryNode
     ) -> GraphQLQueryNode:
         context_type = query_node.context_type
-        infrahub_model = self.schema_branch.get(name=node.type_condition.name.value)
+        infrahub_model = self.schema_branch.get(name=node.type_condition.name.value, duplicate=False)
         context_type = ContextType.DIRECT
         current_node = GraphQLQueryNode(
             parent=query_node,
--- a/backend/infrahub/core/registry.py
+++ b/backend/infrahub/core/registry.py
@@ -113,8 +113,8 @@ class Registry:
             return True
         return False
 
-    def get_node_schema(self, name: str, branch: Branch | str | None = None) -> NodeSchema:
-        return self.schema.get_node_schema(name=name, branch=branch)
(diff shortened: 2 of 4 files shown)


backend/infrahub/graphql/analyzer.py
```diff
@@ -639,7 +639,7 @@ def _populate_inline_fragment_node(
         self, node: InlineFragmentNode, query_node: GraphQLQueryNode
     ) -> GraphQLQueryNode:
         context_type = query_node.context_type
-        infrahub_model = self.schema_branch.get(name=node.type_condition.name.value)
+        infrahub_model = self.schema_branch.get(name=node.type_condition.name.value, duplicate=False)
```
ogenstad, line 642: Good catch! I'm wondering about these methods where where we default to duplicate=True.. I'm wondering if we really want that to be the default.

ogenstad (approved):

backend/infrahub/graphql/analyzer.py
fatih-acar: Also pushed this change (not sure if you saw it): https://github.com/opsmill/infrahub/pull/7341/files#diff-8fe408583ad84711bfe2ed274c9b892ab0b685027e716d716c375e5b64a69b04R116

Back in the time when we introduced `duplicate = True` for optimization, we wanted to selectively disable duplication to be safe and not introduce breakage.
I think we are at a point where we have more `duplicate=False` statements than the other way around, so we can probably switch to `duplicate=False` by default and use `duplicate=True` where it makes sense?

ogenstad: Yeah we can have that as a tech debt item for Infrahub 1.6.

ogenstad: Yes saw your other change as well and I think it highlights the problem of defaulting to False..

PR #7343, opened 2025-10-08
IFC-1883: Validate inherited relationships fields match

solababs:
## Summary by CodeRabbit

- New Features
  - Schema validation now enforces consistent settings for inherited relationships (name, identifier, cardinality, direction, deletion behavior, read-only status, etc.) and reports precise mismatch details.
- Tests
  - Added parameterized test coverage to verify inherited relationship validations and exact error messages for improved reliability and diagnostics.


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,67 @@ def validate_count_against_cardinality(self) -> None:
...
+                        relationship_a = node_a.get_relationship(name=relationship_name)
+                        relationship_b = node_b.get_relationship(name=relationship_name)
+                    except ValueError:
+                        continue
+
+                    for field in compulsorily_matching_fields:
+                        if not hasattr(relationship_a, field) or not hasattr(relationship_b, field):
+                            continue
+
+                        equal_delete_actions = (None, RelationshipDeleteBehavior.NO_ACTION)
```
solababs: We are setting `on_delete` actions that are `None` to `RelationshipDeleteBehavior.NO_ACTION` here https://github.com/opsmill/infrahub/blob/d834be3d583a90a050d02ffc0e25fbb64a1cf00c/backend/infrahub/core/schema/schema_branch.py#L1362-L1372 in `post_processing` which sets the db value for `None` to `RelationshipDeleteBehavior.NO_ACTION` and when this is loaded to compare it would not come as `None` causing a mismatch

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,67 @@ def validate_count_against_cardinality(self) -> None:
...
+    def validate_inherited_relationships_fields(self) -> None:
+        for name in self.node_names:
+            node_schema = self.get(name=name, duplicate=False)
+            if not node_schema.inherit_from:
+                continue
+
+            self.validate_node_inherited_relationship_fields(node_schema)
+
+    def validate_node_inherited_relationship_fields(self, node_schema: NodeSchema) -> None:
+        nodes = [self.get(name=node_name, duplicate=False) for node_name in node_schema.inherit_from]
```
ajtmccarty: might be more clear to call these `generics` instead of `nodes`

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,67 @@ def validate_count_against_cardinality(self) -> None:
...
+        for index, node_a in enumerate(nodes):
+            for node_b in nodes[index + 1 :]:
+                for relationship_name in related_relationship_names:
+                    try:
+                        relationship_a = node_a.get_relationship(name=relationship_name)
+                        relationship_b = node_b.get_relationship(name=relationship_name)
+                    except ValueError:
+                        continue
+
+                    for field in compulsorily_matching_fields:
```
ajtmccarty: I'd call these `property` and `compulsorily_matching_properties`
our vocabulary on this is kind of mixed, but `field` is generally used to mean an attribute or a relationship on a schema, while `property` is an element of a `field`
for example, the `ip_prefix` relationship is a field that has an `peer` property of `BuiltinIPPrefix`

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,67 @@ def validate_count_against_cardinality(self) -> None:
...
+            "read_only",
+        )
+        for index, node_a in enumerate(nodes):
+            for node_b in nodes[index + 1 :]:
+                for relationship_name in related_relationship_names:
+                    try:
+                        relationship_a = node_a.get_relationship(name=relationship_name)
+                        relationship_b = node_b.get_relationship(name=relationship_name)
+                    except ValueError:
+                        continue
```
ajtmccarty: I found the set logic a little hard to read. I think it'd be easier to understand a `dict` of the form `{relationship_name: [NodeSchema, ...]}` or maybe `{relationship_name: [(NodeSchema, RelationshipSchema), ...]}`
then you can just run these checks if the list in the value is of length greater than 1

backend/tests/unit/core/schema_manager/test_manager_schema.py
```diff
@@ -3022,6 +3024,362 @@ def test_schema_branch_conflicting_required_relationships(schema_all_in_one):
...
+    name: str
+    schema: dict[str, Any]
+    error_message: str
+
+
+INHERITED_RELATIONSHIPS_TEST_CASES = [
+    InheritedRelationshipsTestData(
+        name="inherit-from-2-generics-peer-fail",
+        schema={
+            "generics": [
```
ajtmccarty: is it possible to make some of the generics or nodes in this test case into constants that can be reused to reduce the size of `INHERITED_RELATIONSHIPS_TEST_CASES`?

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,79 @@ def validate_count_against_cardinality(self) -> None:
...
+                    if not matched:
+                        raise ValueError(
+                            f"{node_schema.kind} inherits from '{generic_a.kind}' & '{generic_b.kind}'"
+                            f" with different '{_property}' on the '{relationship_name}' relationship"
+                        )
+
+    def _check_relationship_properties_match(
+        self, relationship_a: RelationshipSchema, relationship_b: RelationshipSchema
+    ) -> tuple[bool, str | None]:
+        compulsorily_matching_properties = (
```
ajtmccarty, line 1372: is there a reason that `hierarchical`, `read_only`, and `allow_override` are not included in this list?

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1231,6 +1232,79 @@ def validate_count_against_cardinality(self) -> None:
...
 
+    def validate_inherited_relationships_fields(self) -> None:
+        for name in self.node_names:
+            node_schema = self.get(name=name, duplicate=False)
+            if not node_schema.inherit_from:
+                continue
+
+            self.validate_node_inherited_relationship_fields(node_schema)
+
+    def validate_node_inherited_relationship_fields(self, node_schema: NodeSchema) -> None:
```
ajtmccarty, line 1339: something to think about is that we allow overriding inherited relationships and attributes
say, 2 generics define different versions of a `device` relationship and a node inherits from both of those generics AND overrides the `device` relationship. do we still want to raise a validation error in this case?
I'd say yes b/c it still seems incorrect for a node to inherit from generics with different definitions of the same field (and it is simpler), but just want to make sure it is considered

ajtmccarty (commented):
looks good other than one comment

backend/infrahub/core/schema/schema_branch.py
ogenstad: When we query each generic the relationship on each node inheriting from the generic **must** look the same. So depending on the override we have to do it, i.e. we can't change the type or the peer.

backend/infrahub/core/schema/schema_branch.py
solababs: @ajtmccarty I have updated the list to add those

ajtmccarty (approved):

PR #7348, opened 2025-10-09
add uniqueness constraints for SchemaAttribute/Relationship

### LINKED ISSUE #7346: bug: duplicated attrs/rels on a schema on the default branch crash the application
### Component

API Server / GraphQL

### Infrahub version

1.4.5

### Current Behavior

It is somehow possible to end up with duplicated attributes/relationships on a NodeSchema on the default branch

### Expected Behavior

This should be impossible

### Steps to Reproduce

not entirely sure right now, but this might be it
1. add an attribute/relationship to a schema on a branch
2. add the same attribute/relationship to a schema on the default branch
3. merge the branch

you will end up with 500 errors on the server and a traceback that looks like this
```
  File "/source/community/backend/infrahub/core/schema/manager.py", line 708, in load_schema_from_db
    schema.process(validate_schema=validate_schema)
  File "/source/community/backend/infrahub/core/schema/schema_branch.py", line 502, in process
    self.process_validate()
  File "/source/community/backend/infrahub/core/schema/schema_branch.py", line 521, in process_validate
    self.validate_names()
  File "/source/community/backend/infrahub/core/schema/schema_branch.py", line 974, in validate_names
    raise ValueError(
ValueError: InfraDevice: Names of attributes and relationships must be unique : ['duplicated_field_1', 'duplicated_field_2']
```

### Additional Information

will require either a migration or a script of some sort to fix the issue on existing instances

ajtmccarty:
fixes #7346 

it was possible to upload identical schema changes on a branch and on the default branch, then merge that branch, and break the application in such a manner that it could not start without running a cypher query to clean things up.

the root cause is fixed by adding uniqueness constraints to both `AttributeSchema` and `RelationshipSchema` (`[node, name__value]`) so that attempting to merge a branch with duplicated attribute or relationships schemas will fail with uniqueness constraint violation errors that look like this 

```
Node (SchemaAttribute: <UUID>) is not compliant. The error relates to field name.value='smell'
Node (SchemaAttribute: <UUID>) is not compliant. The error relates to field node.id='<UUID>'"
```

also adds a new `infrahub db check-duplicate-schema-fields` command that shows any duplicates in a nice table and allows fixing them with the `--fix` flag.
Fixing, in this case, means just hard-deleting the earlier version of the duplicate attribute or relationship schemas

TODO:
 - [x] CLI command to fix broken instances
 - [x] test upgrade on existing instance

## Summary by CodeRabbit

- New Features
  - Added CLI command to detect and optionally fix duplicate schema attributes/relationships on the default branch.
- Bug Fixes
  - Prevented merging duplicate attributes/relationships into the default branch to avoid crashes and improve merge errors.
  - Exposed multi-field uniqueness constraints in schema representations to strengthen integrity checks.
- Documentation
  - Documented the new CLI command and its usage in the CLI reference.
- Tests
  - Added functional and unit tests covering detection, fixing, and merge behavior for duplicate schema fields.

### DIFF
--- a/backend/infrahub/cli/db.py
+++ b/backend/infrahub/cli/db.py
@@ -54,6 +54,7 @@ from infrahub.log import get_logger
 
 from .constants import ERROR_BADGE, FAILED_BADGE, SUCCESS_BADGE
 from .db_commands.check_inheritance import check_inheritance
+from .db_commands.clean_duplicate_schema_fields import clean_duplicate_schema_fields
 from .patch import patch_app
 
 
@@ -194,6 +195,29 @@ async def check_inheritance_cmd(
     await initialize_registry(db=dbdriver)
 
     success = await check_inheritance(db=dbdriver, fix=fix)
+    if not success:
+        raise typer.Exit(code=1)
+
+    await dbdriver.close()
+
+
+@app.command(name="check-duplicate-schema-fields")
+async def check_duplicate_schema_fields_cmd(
+    ctx: typer.Context,
+    fix: bool = typer.Option(False, help="Fix the duplicate schema fields on the default branch."),
+    config_file: str = typer.Argument("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+) -> None:
+    """Check for any duplicate schema attributes or relationships on the default branch"""
+    logging.getLogger("infrahub").setLevel(logging.WARNING)
+    logging.getLogger("neo4j").setLevel(logging.ERROR)
+    logging.getLogger("prefect").setLevel(logging.ERROR)
+
+    config.load_and_exit(config_file_name=config_file)
+
+    context: CliContext = ctx.obj
+    dbdriver = await context.init_db(retry=1)
+
+    success = await clean_duplicate_schema_fields(db=dbdriver, fix=fix)
     if not success:
         raise typer.Exit(code=1)
 
--- a/backend/tests/functional/schemas/test_load_on_branch_and_main.py
+++ b/backend/tests/functional/schemas/test_load_on_branch_and_main.py
@@ -7,7 +7,9 @@ from infrahub_sdk.exceptions import GraphQLError
 
 from infrahub.core.constants import HashableModelState, RelationshipCardinality
 from infrahub.core.schema import AttributeSchema, NodeSchema, RelationshipSchema, SchemaRoot
-from tests.helpers.schema import CAR_SCHEMA
+from infrahub.core.schema.definitions.core.artifact import core_artifact_target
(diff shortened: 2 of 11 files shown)


backend/infrahub/cli/db.py
```diff
@@ -200,6 +201,29 @@ async def check_inheritance_cmd(
...
+    config.load_and_exit(config_file_name=config_file)
+
+    context: CliContext = ctx.obj
+    dbdriver = await context.init_db(retry=1)
+
+    success = await clean_duplicate_schema_fields(db=dbdriver, fix=fix)
+    if not success:
+        raise typer.Exit(code=1)
+
+    await dbdriver.close()
```
ogenstad, line 224: Not sure if it makes a practical difference but if there's a failure and we raise typer.Exit we never close the database.

Could be nice with a context manager in the future.

backend/tests/functional/schemas/test_load_on_branch_and_main.py
```diff
@@ -0,0 +1,184 @@
...
+        return await self._load_data(client, default_branch.name)
+
+    async def test_merge_fails_with_duplicates(
+        self,
+        client: InfrahubClient,
+        default_branch: Branch,
+        load_schema_on_branch,
+        load_schema_on_main,
+        # load_data_on_branch: dict[str, InfrahubNode],
+        # load_data_on_main: dict[str, InfrahubNode],
```
ogenstad: Leftovers?

It doesn't look like the `load_data_on_branch` fixture is used. And we only seem to call self._load_data targeting the main branch.

What I'm wondering is what happens to the data that was attached to any deleted attributes. I.e. if we create "car-2-branch1" and then merge it into main. Did we also switch the attribute values for this node to the new schema attributes?

ajtmccarty: yes those are leftovers, removing 

if `car-2-branch1` is created on a branch that contains a duplicated schema, then it would not be possible to merge `branch1` into `main` b/c the new uniqueness constraints on the `AttributeSchema` and `RelationshipSchema` models would block it.
at this point, the only way forward would be to delete the duplicated attributes and relationships from the schema in either `branch1` or `main`, which would also delete the data from the `car` objects using the schema on either `branch1` or `main`, and then merge or rebase

it's not perfect and we could handle it better with schema-specific logic in the diff, which we'll need to add at some point, but that would be a much bigger effort and this way at least blocks the thing that breaks the app

ogenstad: Ok, I think that's fine as long as we don't end up with any nodes in main that we assumed would have data on them.

ogenstad (approved):

Issue #7353, opened 2025-10-09
bug: loading a schema that has an atribute with state absent=property will still create the attribute if the node does not exist


ogenstad:
### Component

API Server / GraphQL

### Infrahub version

v1.5.0b0

### Current Behavior

When you load a schema, that defines a new node, which has an attribute defined with the state: absent property, then the attribute will be created on the node.

This is a situation that could happen when someone loads existing schema files into a new instance of Infrahub.

### Expected Behavior

Since the attribute is defined with the state: absent property, it should not be created.

### Steps to Reproduce

1. Start a new instance of Infrahub
2. Load this schema
```yaml
# yaml-language-server: $schema=https://schema.infrahub.app/infrahub/schema/latest.json
---
version: '1.0'

nodes:
  - name: Widget
    namespace: Test
    label: "Widget"
    display_labels: [name__value]
    attributes:
      - name: name
        kind: Text
        unique: true
      - name: description
        kind: Text
        state: absent
```
3. Check the schema for TestWidget and note that it has an attribute called `description` even though it shouldn't.

### Additional Information

This is a follow up for the issue #4302, where the relationship bug was fixed in #7345.

PR #7357, opened 2025-10-09
IFC-1893: Handle processing schema attribute state on creation and update

solababs:
Fixes https://github.com/opsmill/infrahub/issues/7353

## Summary by CodeRabbit

* **Bug Fixes**
  * Attributes marked as “absent” are now pruned during schema pre-validation and will no longer be created, stored, or shown in loaded schemas (applies to nodes and generic interfaces).

* **Tests**
  * Added unit tests verifying absent attributes are removed for both nodes and generics.

* **Documentation**
  * Changelog entry added noting prevention of creating attributes with state "absent".

* **Chore**
  * Updated test fixture to remove an absent attribute flag.


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1628,6 +1629,21 @@ def process_relationships_state(self) -> None:
             updated_node.relationships = filtered_relationships
             self.set(name=name, schema=updated_node)
 
+    def process_attributes_state(self) -> None:
+        for name in self.node_names + self.generic_names_without_templates:
+            node = self.get(name=name, duplicate=False)
+            if node.id or (not node.id and not node.attributes):
+                continue
```
ajtmccarty: I think this will prevent you from checking `ABSENT` attributes added to an existing schema
it would be good to add a test for that case

solababs: Updates already work, existing schema once updated to absent already removes the attribute, this case it is only for creation

ajtmccarty: if I load a schema like this
```yaml
---
version: "1.0"

nodes:
  - name: PersonTwo
    namespace: Random
    include_in_menu: true
    attributes:
      - name: name
        kind: Text
        optional: true
      - name: should_be_absent
        state: absent
        kind: Text
        optional: true
```

then the `should_be_absent` schema is still loaded in the schema explorer in the UI

Image

so I think you need to add support for a new `state=absent` attribute being added to an existing Node or Generic schema

solababs: I think you loaded it on `develop` branch which currently has the issue, if you load it on this branch it would not be present

Image

ogenstad: This looks good to me. @ajtmccarty, might you have been validating this on another branch as Sola suggested?

ogenstad (approved):
LGTM, we can wait to see what it was Aaron tested before merging but as far as I can see this works as expected.

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: no, this is still a problem
my instructions for reproducing it weren't clear enough
you need to load a schema and then update the existing schema with a new attribute that has `state=absent`
this conditional will always skip an existing schema, even if the schema has a new attribute defined on it

solababs: The logic in the `diff` function here https://github.com/opsmill/infrahub/blob/403f79fda0dca1106ee41370a699f67c6ad27474/backend/infrahub/core/schema/schema_branch.py#L220-L223 works correctly for updates but I have also updated the logic in this PR to work for both update and new attributes.

ogenstad: I can't reproduce this but perhaps the latest commit fixed the remaining issue?

ajtmccarty: looks good to me now. approved.

ajtmccarty (approved):

Issue #7354, opened 2025-10-09
bug: Loading a schema where the node or generic state=absent will create the node in the database if it doesn't exist


ogenstad:
### Component

API Server / GraphQL

### Infrahub version

v1.5.0b0

### Current Behavior

When you load a schema and define a node or generic to have state=absent the node will still be created provided that it didn't previously exist in the database.

This is a situation that could happen when someone loads existing schema files into a new instance of Infrahub.

### Expected Behavior

The node or generic should not be created when if defined as `state: absent`.

### Steps to Reproduce

1. Start a new instance of Infrahub
2. Load this schema:

```yaml
# yaml-language-server: $schema=https://schema.infrahub.app/infrahub/schema/latest.json
---
version: '1.0'

nodes:
  - name: Widget
    namespace: Test
    label: "Widget"
    state: absent
    display_labels: [name__value]
    attributes:
      - name: name
        kind: Text
        unique: true
      - name: description
        kind: Text
```
3. Verify that a TestWidget node schema has been added even though it shouldn't.
4. Loading the schema again will remove the node.

### Additional Information

Originated from https://github.com/opsmill/infrahub/issues/4302#issuecomment-2729571797 and also related to #7353.

PR #7358, opened 2025-10-09
IFC-1894: Handle processing schema node state on creation

solababs:
Fixes https://github.com/opsmill/infrahub/issues/7354

## Summary by CodeRabbit

- Bug Fixes
  - Prevents creation/registration of nodes marked as "absent".
  - Removes absent nodes during schema load via an added pre-validation cleanup, so they no longer interfere with processing or validation.
  - Ensures absent nodes cannot be retrieved from the registry.

- Tests
  - Added a unit test verifying absent nodes are not accessible after schema loading.

- Documentation
  - Added changelog entry noting the fix for "absent" nodes.

### DIFF
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -1632,9 +1632,8 @@ class SchemaBranch:
     def process_nodes_state(self) -> None:
         for name in self.node_names + self.generic_names_without_templates:
             node = self.get(name=name, duplicate=False)
-            if node.id or (not node.id and node.state != HashableModelState.ABSENT):
-                continue
-            self.delete(name=name)
+            if not node.id and node.state == HashableModelState.ABSENT:
+                self.delete(name=name)
 
     def _generate_weight_generics(self) -> None:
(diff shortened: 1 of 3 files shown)


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1628,6 +1629,13 @@ def process_relationships_state(self) -> None:
             updated_node.relationships = filtered_relationships
             self.set(name=name, schema=updated_node)
 
+    def process_nodes_state(self) -> None:
+        for name in self.node_names + self.generic_names_without_templates:
+            node = self.get(name=name, duplicate=False)
+            if node.id or (not node.id and node.state != HashableModelState.ABSENT):
+                continue
+            self.delete(name=name)
```
ajtmccarty: normally, I prefer an early return/continue too, but I think this is easier to read when it's only one line after the conditional. your choice, maybe it's just me
```python
if not node.id and node.state is HashableModelState.ABSENT:
    self.delete(name=name)
```

ajtmccarty (approved):

ajtmccarty (approved):

PR #7360, opened 2025-10-09
Phase out subscription=False when generating GraphQL schema

ogenstad:
This PR aims to phase out the use of subscription=False and mutations=False within the unittests. These parameters was at some point used to speed up the generation of the GraphQL schema, but since we cache the GraphQL schema by hash it's instead a source of flaky tests and can actually slow the CI down as we might end up regenerating the GraphQL schema even though we don't need to (the use of subscription=False or mutations=False causes the schema to be regenerated each time). In production we never use these parameters, it's only within the tests and I think the next step is to remove them completely.

Hopefully this will fix the occasionally flaky tests that we've observed in the unittests.

As part of this fix I introduced a new fixture that forces the regeneration of the schema for tests using GraphQL enums (an experimental feature) as tests might get unreliable without them, the fixture also resets the graphql registry afterwards to avoid impacting other tests.

As part of this PR I've also made some slight fixes to typing where applicable, things like `assert result.data` is because result.data is an optional dictionary and we tend to have a lot of typing violations which are not currently reported but listed as errors in the file by pylance, the problem is when we try to access a dictionary within data while not knowing if data is `None`.

We might be using the `branch.update_schema_hash()` call a bit excessively in this PR, but it will still be faster than a complete regeneration of the GraphQL schema in all of these tests.

Note there could still potentially be sone areas where test issues comes up as we run some of the tests in an async loop that's updating the global objects (registry), if this happens we might need to reconsider how we approach these tests.

## Summary by CodeRabbit

* **Tests**
  * Refactored GraphQL test infrastructure with improved schema state management and type safety in test signatures.
  * Enhanced test isolation through schema cache reset mechanisms between test runs.

### DIFF
--- a/backend/tests/benchmark/test_graphql_query.py
+++ b/backend/tests/benchmark/test_graphql_query.py
@@ -73,10 +73,8 @@ def test_query_one_model(exec_async, aio_benchmark, db: InfrahubDatabase, defaul
         }
     }
     """
-
-    gql_params = exec_async(
-        prepare_graphql_params, db=db, include_mutation=False, include_subscription=False, branch=default_branch
-    )
+    default_branch.update_schema_hash()
+    gql_params = exec_async(prepare_graphql_params, db=db, branch=default_branch)
 
     for _ in range(NBR_WARMUP):
         exec_async(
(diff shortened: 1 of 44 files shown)


ajtmccarty (approved):
is there some way to avoid adding so many `update_schema_hash()` calls? maybe a helper function we can call that calls `branch.update_schema_hash()` and `prepare_graphql_params`?

PR #7361, opened 2025-10-09
Draft: Template profile support

minitriga:
## Summary by CodeRabbit

* **New Features**
  * Templates now support attaching profiles; profile defaults are applied when creating nodes (multi-profile precedence honored; explicit template values win). UI surfaces current profiles and badges in template and create flows.

* **Documentation**
  * Updated docs to clarify template → profile inheritance.

* **Tests**
  * Added broad unit, integration and end-to-end tests for templates, profiles, precedence, metadata origins, and create-from-template flows.


frontend/app/src/shared/components/form/profiles-selector.tsx
```diff
@@ -70,7 +64,9 @@ export const ProfilesSelector = ({
   const profilesList = kindList
     .map((profile) => {
       // Get the profile schema for the current kind
-      const profileSchema = profileSchemas.find((profileSchema) => profileSchema.name === profile);
+      const profileSchema = profileSchemas.find(
+        (profileSchema) => profileSchema.name === profile?.replace("Template", "")
```
bilalabbad: I don't understand why we need `profile?.replace("Template", "")`?

backend/infrahub/profiles/node_applier.py
```diff
@@ -29,7 +30,12 @@ async def _get_attr_names_for_profiles(self, node: Node) -> list[str]:
         for attr_schema in node_schema.attributes:
             attr_name = attr_schema.name
             node_attr: BaseAttribute = getattr(node, attr_name)
-            if node_attr.is_from_profile or node_attr.is_default:
+            is_template = None
+            if node_attr.source_id:
+                await node_attr.get_source(db=self.db)
+                if isinstance(node_attr.source.get_schema(), TemplateSchema):
+                    is_template = True
+            if node_attr.is_from_profile or node_attr.is_default or is_template:
```
ajtmccarty, line 38: does this mean that profile values will override template values? if so, that's not what I would expect, but maybe you could explain what the expected behavior is in the PR description

I would think that template values would take precedence over profile values. and I'd think that adding the `profiles` relationship to a template would be enough and this change to `NodeProfilesApplier` wouldn't be necessary, but I could definitely be missing something

frontend/app/src/shared/components/form/profiles-selector.tsx
minitriga: So here we are linking the template to the profile of a node there is no profile for the template

minitriga:
@copilot can you have a look at this PR and create update tests where necessary?

frontend/app/src/entities/nodes/object-template/object-template-form.tsx
```diff
@@ -80,7 +80,15 @@ export default function ObjectTemplateForm({
   }
 
   if (selectedObjectTemplate !== undefined) {
-    return <ObjectForm {...props} objectTemplate={selectedObjectTemplate} />;
+    return (
+      <ObjectForm
+        {...props}
+        currentProfiles={
+          selectedObjectTemplate?.profiles?.edges?.map((edge: any) => edge?.node) ?? []
```
bilalabbad, line 87: profile input remains empty when I select a template with profiles

gmazoyer (approved):

fatih-acar:
Can we rebase this branch correctly?

minitriga:
New PR: https://github.com/opsmill/infrahub/pull/7698

PR #7367, opened 2025-10-10
Convert old display label format to new when processing the schema

ogenstad:


backend/infrahub/core/node/__init__.py
```diff
@@ -180,8 +180,10 @@ async def add_human_friendly_id(self, db: InfrahubDatabase) -> None:
 
     async def get_display_label(self, db: InfrahubDatabase) -> str:
         if self._display_label:
-            if display_label := self._display_label.get_value(node=self, at=self._at):
-                return display_label
+            if isinstance(self._display_label._value, str):
+                return self._display_label._value
```
gmazoyer, line 184: Did `get_value()` return an unexpected value in that case?

gmazoyer (approved):

backend/infrahub/core/node/__init__.py
ogenstad: It returned the correct values but codspeed identified it as slowing things down. Probably because we were creating two new objects that seem to take time to create.

Issue #7372, opened 2025-10-11
Boolean attributes always show true in details view


dgarros:
## Current Behavior

In the frontend Details view of an object, attributes of type boolean are always displayed with a checkmark (✅), regardless of their actual value. This visually implies that the attribute is true even when it is false.

## Expected Behavior

Boolean attributes should accurately reflect their stored value:
	•	true → checkmark (or “True”)
	•	false → empty/unchecked state (or “False”)

## Steps to Reproduce

1.	Go to the Infrahub sandbox at https://sandbox.infrahub.app.
2.	Navigate to any Python Transform object (these have a boolean attribute named convert_query_response).
3.	In the object Details view, observe that the convert_query_response attribute is displayed with a checkmark, indicating true.
4.	Click Edit for the same object.
5.	Observe that in the edit form, convert_query_response is actually set to false.
6.	Save without making changes — the Details view continues to show a checkmark, even though the stored value is false.

bilalabbad:
Fixed in 1.4.11

PR #7378, opened 2025-10-13
Fix(UI): Boolean fields now correctly display true or false in the Details view

bilalabbad:
Issue:
- #7372 

## Summary by CodeRabbit

- Bug Fixes
  - Boolean fields in the Details view now show a checkmark only for true; false values no longer display a checkmark.
  - Improved consistency of icon rendering for boolean values.

- Documentation
  - Added a changelog entry describing the boolean display fix.
  - Embedded explanatory videos in branching and resource-manager documentation.

### DIFF
--- a/changelog/7372.fixed.md
+++ b/changelog/7372.fixed.md
@@ -1,0 +1,1 @@
+Fixed an issue where boolean fields in the object Details view always displayed a checkmark, even when the value was false.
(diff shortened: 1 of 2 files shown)


dgarros:
@bilalabbad thank you for the quick fix
if possible it would be good to fix this one in `stable` instead instead of `develop`

pa-lem (approved):

PR #7389, opened 2025-10-15
Clean up GraphQL schema export 

dgarros:
This PR refactor how we generate and validate the GraphQL schema by replacing the current invoke-based approach with a pure python approach with less dependencies

### ⚡ Faster CI Pipeline

Before: CI had to build Docker images, start a Neo4j database, and install npm packages just to generate a schema
After: CI now generates schemas directly without any database or Docker setup
Result: Significantly faster CI execution and reduced resource usage

### 🧹 Cleaner Dependencies
Removed: npm packages (format-graphql) that were only used for schema formatting
Removed: Database connection requirements for schema generation
Result: Simpler project setup and fewer dependencies to maintain

### 🎯 Better Schema Sorting
Before: Used regex patterns to sort GraphQL interfaces (unreliable)
After: Uses proper AST-based sorting for all schema elements
Result: More accurate and comprehensive sorting of interfaces, types, fields, and enum values

### 🔄 Maintained Compatibility
Invoke tasks still work: 
`poetry run invoke schema.generate-graphqlschema` and `poetry run invoke schema.validate-graphqlschema`
Same output: Generated schemas are identical but now properly sorted
No breaking changes: Existing workflows continue to work

## What This Means for You
Faster development: CI runs faster, so you get feedback quicker
More reliable: Schema sorting is now more accurate and comprehensive
Easier maintenance: Fewer dependencies and simpler codebase
Same experience: All existing commands and workflows continue to work

## Summary by CodeRabbit

* **New Features**
  * GraphQL schema endpoint now supports an optional sorting parameter for alphabetically organized schema output.
  * New internal development CLI commands added for schema export, database initialization, and test data loading.

* **Chores**
  * Simplified CI workflow by removing redundant setup steps.
  * Reorganized CLI database commands and development utilities.


backend/infrahub/cli/schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+@app.command(name="export-graphql-schema")
+async def export_graphql_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),  # noqa: ARG001
+    out: Path = typer.Option("schema.graphql"),  # noqa: B008
+) -> None:
+    """Export the GraphQL schema to a file."""
+
+    config.load_and_exit()
```
ogenstad: As code rabbit indicates we should include the config file as an argument to config.load_and_exit()

backend/infrahub/cli/schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+app = AsyncTyper()
+
+
+@app.command(name="export-graphql-schema")
+async def export_graphql_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),  # noqa: ARG001
+    out: Path = typer.Option("schema.graphql"),  # noqa: B008
+) -> None:
```
ogenstad: I'm wondering if this command might lead to some confusion. If I run the command `infrahub schema export-graphql-schema` on my own server my expectation would be that we'd export the current GraphQL schema for the installation. I like the fact that we have this command that is a lot simpler to use than what I originally started with. But I wonder if this is the correct command. I'd argue that it should clearly indicate that we're exporting the base schema, if there's a need we could perhaps later add an option to export the GraphQL schema as it exists in the database (As we can do this through the API I'm not sure it will ever be required).

It might be a bit more verbose but if we had a command like: `infrahub schema export-graphql-core-schema` I think it would avoid confusion. The docstring could also be updated to reflect this, if you have a suggestion for a shorter name I think that's fine too, but it would be good to avoid this area of confusion.

backend/infrahub/cli/schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+    schema_branch.process()
+
+    gqlm = GraphQLSchemaManager(schema=schema_branch)
+    gql_schema = gqlm.generate(
+        include_query=True,
+        include_mutation=True,
+        include_subscription=True,
+        include_types=True,
+    )
```
ogenstad: I'd like to start to phase out these arguments to generate. As we can how cache the schema in by hash we end up with some occational flaky unittests and I don't think we ever use anything other than `=True` in the production we only ever use these arguments within the tests.

This could be mitigated if we had a unittest that would run this command similar to how we test the CTL scripts in the SDK and later validate that we can correctly read the output GraphQL file if we write it to a temp folder.

backend/infrahub/graphql/schema_sort.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Returns:
+        Sorted list of named type nodes, or None if input was None.
+    """
+    if not interfaces:
+        return None
+    return sorted(interfaces, key=lambda i: i.name.value)
+
+
+def sort_schema_ast(document: DocumentNode) -> DocumentNode:
```
ogenstad, line 104: I think we should call this function from this endpoint as well and always print out the sorted version:

https://github.com/opsmill/infrahub/blob/infrahub-v1.5.0b0/backend/infrahub/graphql/api/endpoints.py#L37

backend/infrahub/graphql/schema_sort.py
```diff
@@ -0,0 +1,157 @@
...
+            sorted_inputs = _sort_input_fields(definition.fields)
+            sorted_definitions.append(
+                InputObjectTypeDefinitionNode(
+                    name=definition.name,
+                    directives=definition.directives,
+                    fields=sorted_inputs,
+                    description=definition.description,
+                    loc=definition.loc,
+                )
+            )
```
ogenstad, line 166: We are missing a few types here just like CodeRabbit indicates. The `@expand` directive for example has been removed from the output file. I'd suggest to do a:

```python
else:
    breakpoint()
```

To catch missing definition types.

I'm not sure if all of the below are required as some of them might be child nodes to the ones that you have included.

```python
>>> from collections import deque
>>>
>>> from graphql.language.ast import DefinitionNode
>>>
>>> def all_subclasses(cls):
...     seen = set()
...     q = deque(cls.__subclasses__())
...     while q:
...         sub = q.popleft()
...         if sub in seen:
...             continue
...         seen.add(sub)
...         q.extend(sub.__subclasses__())
...     return seen
...
>>>
>>> for subclass in all_subclasses(DefinitionNode):
...     print(subclass)
...
<class 'graphql.language.ast.InputValueDefinitionNode'>
<class 'graphql.language.ast.OperationDefinitionNode'>
<class 'graphql.language.ast.InterfaceTypeDefinitionNode'>
<class 'graphql.language.ast.InterfaceTypeExtensionNode'>
<class 'graphql.language.ast.EnumTypeExtensionNode'>
<class 'graphql.language.ast.EnumValueDefinitionNode'>
<class 'graphql.language.ast.FieldDefinitionNode'>
<class 'graphql.language.ast.SchemaDefinitionNode'>
<class 'graphql.language.ast.DirectiveDefinitionNode'>
<class 'graphql.language.ast.UnionTypeDefinitionNode'>
<class 'graphql.language.ast.ObjectTypeDefinitionNode'>
<class 'graphql.language.ast.ObjectTypeExtensionNode'>
<class 'graphql.language.ast.TypeSystemDefinitionNode'>
<class 'graphql.language.ast.ScalarTypeDefinitionNode'>
<class 'graphql.language.ast.ExecutableDefinitionNode'>
<class 'graphql.language.ast.FragmentDefinitionNode'>
<class 'graphql.language.ast.TypeDefinitionNode'>
<class 'graphql.language.ast.TypeExtensionNode'>
<class 'graphql.language.ast.UnionTypeExtensionNode'>
<class 'graphql.language.ast.InputObjectTypeExtensionNode'>
<class 'graphql.language.ast.InputObjectTypeDefinitionNode'>
<class 'graphql.language.ast.EnumTypeDefinitionNode'>
<class 'graphql.language.ast.ScalarTypeExtensionNode'>
>>>
```

backend/infrahub/graphql/schema_sort.py
```diff
@@ -0,0 +1,157 @@
...
+
+    return DocumentNode(definitions=sorted_definitions)
```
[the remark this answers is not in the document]
ogenstad: The final

```python
else:
    sorted_definitions.append(definition)
```

Could make sense to have something that produces a complete schema even though some parts might not be sorted if we're missing sort options for existing or new definition types for later versions of GraphQL.

ogenstad (commented):
Looks like a good change to me though some work remaining. A though could be to take the existing `./schema/schema.graphql` file from `develop` and run it through this function and then see if the output is the same as what we get from running `infrahub schema export-graphql-schema`.

backend/infrahub/graphql/schema_sort.py
dgarros: Not sure about this one, sorting the schema takes some time/resource and in most cases it isn't required so it could add some unnecessary overhead. 

Maybe we can keep it unsorted by default and add an option to sort it, something like `/schema.graphql?sort=true`

backend/infrahub/cli/schema.py
dgarros: This is a very good point, indeed, as it, it would be confusing to the user.
Instead of having a `infrahub schema` command maybe we could create a `infrahub dev` or `infrahub internal` command where we would have all the sub commands we are using internally during the development cycle ? 
We could also migrate the `infrahub db init` to `infrahub dev|internal db-init`  or similar ?

ogenstad: Yes having these under `infrahub dev` I think would do the trick.

backend/infrahub/graphql/schema_sort.py
ogenstad: We can do that, on the other hand it's not a url that's used a lot. I think the only thing that actively uses it would be the tool you did for the SDK along with the VS Code plugin. But providing a sort query argument would work as well.

backend/infrahub/cli/dev.py
```diff
@@ -0,0 +1,118 @@
...
+app = AsyncTyper()
+
+
+@app.command(name="export-graphql-schema")
+async def export_graphql_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("schema.graphql"),  # noqa: B008
+) -> None:
+    """Export the GraphQL schema to a file."""
```
ogenstad: For clarity I think we should still say "Export the core GraphQL schema to a file."

ogenstad (approved):

PR #7390, opened 2025-10-15
feat(cli): add tasks flush commands

fatih-acar:
This will be used by the ops tools (infrahub-taskmanager).

This can be included in an internal scheduled workflow later on, if required.

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI subcommand group: flush.
    - flush flow-runs: bulk-remove flow runs older than a configurable retention (days-to-keep, default 30) with batching (batch-size, default 100).
    - flush stale-runs: target stale RUNNING flow runs (days-to-keep, default 2; batch-size, default 100) and mark them instead of deleting.
  * Both commands load configuration and allow adjustable logging during cleanup.

### DIFF
--- a/backend/infrahub/task_manager/task.py
+++ b/backend/infrahub/task_manager/task.py
@@ -1,7 +1,10 @@
+import asyncio
 import uuid
+from datetime import datetime, timedelta, timezone
 from typing import Any
 from uuid import UUID
 
+from prefect import State
 from prefect.client.orchestration import PrefectClient, get_client
 from prefect.client.schemas.filters import (
     ArtifactFilter,
@@ -12,6 +15,7 @@ from prefect.client.schemas.filters import (
     FlowRunFilter,
     FlowRunFilterId,
     FlowRunFilterName,
+    FlowRunFilterStartTime,
     FlowRunFilterState,
     FlowRunFilterStateType,
     FlowRunFilterTags,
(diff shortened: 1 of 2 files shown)


backend/infrahub/task_manager/task.py
```diff
@@ -311,3 +314,68 @@ async def query(
                     )
 
         return {"count": count or 0, "edges": nodes}
+
+    @classmethod
+    async def delete_flow_runs(
+        cls,
+        states: list[StateType] | None = None,
```
ajtmccarty: could you give this the default used below
```
states: list[StateType] =  [StateType.COMPLETED, StateType.FAILED, StateType.CANCELLED],
```
and then delete the `if states is None:` conditional?

backend/infrahub/task_manager/task.py
```diff
@@ -311,3 +314,68 @@ async def query(
...
+                        logger.warning(f"Failed to delete flow run {flow_run.id}: {e}")
+                        failed_deletes.append(flow_run.id)
+
+                    # Rate limiting
+                    if batch_deleted % 10 == 0:
+                        await asyncio.sleep(0.5)
+
+                logger.info(f"Delete {batch_deleted}/{len(flow_runs)} flow runs (total: {deleted_total})")
+                if failed_deletes:
+                    logger.warning(f"Failed to delete {len(failed_deletes)} flow runs")
```
ajtmccarty: this mostly duplicates the logger message on 364
not sure if you need both

backend/infrahub/task_manager/task.py
```diff
@@ -311,3 +314,68 @@ async def query(
...
+
+                # Get next batch
+                flow_runs = await client.read_flow_runs(flow_run_filter=flow_run_filter, limit=batch_size)
```
ajtmccarty, line 373: if there are flows that fail to delete or set to `CRASHED` will this query keep retrieving them and trying to delete/CRASH them?
maybe that is expected and fine, but if there are enough runs that cannot be deleted/CRASHed and the batch size is small enough, then I think this could end up in an infinite loop where it keeps retrieving the same runs and failing to delete them

ajtmccarty (approved):

backend/infrahub/task_manager/task.py
fatih-acar: ruff was complaining about this, so I followed its recommendation. But it looks cleaner to me too, so I added an ignore for now...

backend/infrahub/task_manager/task.py
fatih-acar: Indeed, removed the other one. This one can be useful since it shows the actual error within the exception.

backend/infrahub/task_manager/task.py
fatih-acar: Yes good catch! I've seen that behavior while testing CLI mode (where I could just break execution). But it can be an issue if we integrate this within a scheduled task, so I've added an extra check to see if we are processing the same flow runs and abort.

ajtmccarty (approved):

Issue #7157, opened 2025-09-05
bug: Templates might cause unpredictable schema hashes


ogenstad:
### Infrahub version

1.4.4

### Current Behavior

This feature was raised in the SDK as https://github.com/opsmill/infrahub-sdk-python/issues/526 where the `infrahubctl schema load schemas/ --wait 30 ` command would time out with:

```
 Schema is still not in sync after the specified waiting time
```

The problem is not however an issue within the SDK code but a problem within Infrahub.

It seems to relate to unpredictable hashing of auto generated Template nodes within the schema and these nodes can have different hash values, which leads to the impression that the schema never converges across the workers.

### Expected Behavior

The schemas should be hashed in exactly the same way in all circumstances.

### Steps to Reproduce

1. spin infrahub
2. clone this repo at this ref https://github.com/opsmill/infrahub-demo-service-catalog/tree/1035dbec5f8076d820fe4d1292d598c8639eec26
3. poetry install
4. poetry shell infrahubctl schema load schemas/ --wait 30
5. Note that the command will fail
6. Check the schema on the API endpoint /api/schema/summary
7. Run an `invoke dev.stop dev.start`
8. Check the schema again on the API endpoint /api/schema/summary (this is mainly to highlight that a difference can occur, but it should more or less replicate what we see when looking at different workers.

### Additional Information

-

PR #7406, opened 2025-10-15
Generate order_weight for generic templates

solababs:
Fixes https://github.com/opsmill/infrahub/issues/7157

## Description
Generate order_weight for generic templates.
The order of the fields for the template must respect the order of the node.

## Summary by CodeRabbit

* **New Features**
  * Templates now inherit ordering weights from their generic definitions so attributes and relationships maintain consistent display/order.

* **Tests**
  * Added a unit test verifying order-weight propagation from generics to templates, including scenario where template weights are missing and then populated from generics.

* **Documentation**
  * Changelog entry documenting automatic order-weight generation for generic templates.


backend/tests/unit/core/schema_manager/test_manager_schema.py
```diff
@@ -603,6 +603,69 @@ def extract_weights(schema: SchemaBranch):
...
+    }
+    schema_branch = SchemaBranch(cache={}, name="test")
+    schema_branch.load_schema(schema=SchemaRoot(**schema))
+    schema_branch.process()
+
+    template = schema_branch.get(name="TemplateDcimGenericDevice", duplicate=False)
+
+    assert template.get_attribute(name="template_name").order_weight == 1000
+    assert template.get_attribute(name="description").order_weight == 2000
+    assert template.get_attribute(name="os_version").order_weight == 3000
```
ogenstad: In this case the expectation would be that the order_weight for os_version matches what was set on the generic itself i.e. 5200.

backend/tests/unit/core/schema_manager/test_manager_schema.py
```diff
@@ -661,9 +661,9 @@ def test_schema_branch_processes_generic_template_schema_weight(register_core_mo
     template = schema_branch.get(name="TemplateDcimGenericDevice", duplicate=False)
 
     assert template.get_attribute(name="template_name").order_weight == 1000
-    assert template.get_attribute(name="description").order_weight == 2000
-    assert template.get_attribute(name="os_version").order_weight == 3000
-    assert template.get_relationship(name="tags").order_weight == 4000
+    assert template.get_attribute(name="description").order_weight == 8000
+    assert template.get_attribute(name="os_version").order_weight == 5200
+    assert template.get_relationship(name="tags").order_weight == 3000
```
ogenstad: I'd like some additional asserts here if we start by grabbing the generic:

```python
generic_node = schema_branch.get(name="DcimGenericDevice", duplicate=False)
```
Then aside from the "template_name" we validate that the order_weight on all the other attributes and relationships match both the expected value and that the order_weight is the same on both the generic and the template.

I'd also like one additional test here, if we remove the order weight for the name attribute from above, i.e. instead of:

```json
{"name": "name", "kind": "Text", "unique": True, "order_weight": 7000},
```
we have:

```json
{"name": "name", "kind": "Text", "unique": True},
```

Then we have an assert that validates that the "name" attribute has the same order_weight on both the generic and the template.

backend/tests/unit/core/schema_manager/test_manager_schema.py
```diff
@@ -605,6 +605,110 @@ def extract_weights(schema: SchemaBranch):
...
+        template.get_relationship(name="tags").order_weight
+        == dcim_generic_device.get_relationship(name="tags").order_weight
+        == 3000
+    )
+
+    schema_2 = copy.deepcopy(schema)
+    schema_2["generics"][0]["attributes"] = [
+        {"name": "name", "kind": "Text", "unique": False, "order_weight": 7000},
+        {"name": "description", "kind": "Text", "optional": True, "order_weight": 5650},
+        {"name": "os_version", "kind": "Text", "optional": True, "order_weight": 2500},
```
ogenstad: What I meant was that if we could have this line instead be:
```diff
- {"name": "os_version", "kind": "Text", "optional": True, "order_weight": 2500},
+ {"name": "os_version", "kind": "Text", "optional": True},
```

Then below we also remove the hardcoded value but still ensure that we get the same value even though it's dynamically assigned.:

```python
assert template.get_attribute(name="os_version").order_weight == dcim_generic_device.get_attribute(name="os_version").order_weight
```

ogenstad (approved):

PR #7408, opened 2025-10-16
duplicated updated generic attributes

### LINKED ISSUE #7407: bug: adding an attribute to a generic creates duplicates on schemas using the generic
### Component

API Server / GraphQL

### Infrahub version

1.4.10

### Current Behavior

updating a schema to add an attribute to a generic will result in duplicate attributes being added to any objects using schemas that inherit from that generic at the database level

this will cause inconsistent results when updating/querying the object with duplicated attributes b/c we only update one attribute during an update and there is no guarantee that we will return the correct version of the attribute when querying it

### Expected Behavior

only one instance of an attribute should be added to each object that inherits from a given generic when the generic schema is updated

### Steps to Reproduce

1. load a schema that includes a node that inherits from a generic
2. add at least one object that uses the inheriting node schema
3. load an updated version of the schema that adds a new attribute to the generic
4. 
   a. either check on the object you created in step 2 at the database level to see that it has multiple instances of the attribute you added in step 3 
   b. or try to update the object you created in step 2 and then retrieve its value and see that it is inconsistent

### Additional Information

_No response_

ajtmccarty:
fixes #7407 

fixes a bug that would cause duplicated attributes on nodes when adding an attribute to a generic schema that the node inherits from. the problem is that the `node.attribute.add` migration was run for both the node and generic schemas, which can result in 2 new attributes with the same name. this is not guaranteed to happen and the `NodeAttributeAddMigration` actually contains logic to avoid this issue, but it the problem can still occur if multiple migrations adding the same attribute to overlapping groups of nodes run at the same time because they run inside of transactions and if both cypher queries start before the other is committed, then they will create duplicates.

also fixes an unrelated bug in the `NodeCreateAllQuery` that was found during testing

TODO:
- [x] add migration

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where duplicate attributes could be created when updating generic schemas with new attributes
  * Improved attribute handling for inherited schemas to prevent duplicate attribute creation

* **Migrations**
  * Added graph schema migration (v40) to identify and clean up any existing duplicate attributes while maintaining data consistency across all branches


backend/infrahub/core/migrations/schema/node_attribute_add.py
ajtmccarty, line 1: we only want to run this schema migration on non-inherited attributes, b/c if we run it on an inherited attribute then it will be run twice (once for the generic and once for the node schema) and we'll end up with duplicates

backend/infrahub/core/query/node.py
```diff
@@ -233,11 +233,15 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
         UNWIND $attrs AS attr
         // Try to find a matching vertex
-        OPTIONAL MATCH (existing_av:AttributeValue {value: attr.content.value, is_default: attr.content.is_default})
-        WHERE NOT existing_av:AttributeValueIndexed
+        CALL (attr) {
+            OPTIONAL MATCH (existing_av:AttributeValue {value: attr.content.value, is_default: attr.content.is_default})
+            WHERE NOT existing_av:AttributeValueIndexed
+            RETURN existing_av
+            LIMIT 1
+        }
```
ajtmccarty, line 244: in certain circumstances we can end up with duplicated `AttributeValue` nodes, usually when two requests that involve a new attribute with the same value are handled at the same time. this query did not account for the this quirk and could end up creating duplicate paths to AttributeValues

ogenstad (approved):

PR #7418, opened 2025-10-16
IFC-1886: Paginated branch graphql query

solababs:
## Summary by CodeRabbit

* **New Features**
  * Added InfrahubBranch GraphQL types, edge/pagination wrapper, and a paginated InfrahubBranch query (offset/limit) exposing detailed branch metadata.

* **Bug Fixes / Data**
  * Branch listings now exclude deleting branches and return corrected total counts; pagination parameters are validated.

* **Tests**
  * Added unit tests for paginated InfrahubBranch query and field-mapping consistency.

* **Chores**
  * Updated pre-commit config to ignore the generated schema file for EOF fixes.


backend/infrahub/core/branch/models.py
```diff
@@ -165,6 +165,18 @@ async def get_list(
...
+    @classmethod
+    async def get_list_and_count(
+        cls,
+        db: InfrahubDatabase,
+        limit: int = 1000,
+        ids: list[str] | None = None,
+        name: str | None = None,
+        **kwargs: Any,
+    ) -> tuple[list[Self], int]:
+        kwargs["raw_filter"] = f"n.status <> '{BranchStatus.DELETING.value}'"
```
ajtmccarty: It's be better to use a named parameter for `raw_filter` instead of `kwargs`

backend/infrahub/core/node/standard.py
```diff
@@ -227,3 +227,19 @@ async def get_list(
...
+        ids: list[str] | None = None,
+        name: str | None = None,
+        **kwargs: dict[str, Any],
+    ) -> tuple[list[Self], int]:
+        query: Query = await StandardNodeGetListQuery.init(
+            db=db, node_class=cls, ids=ids, node_name=name, limit=limit, **kwargs
+        )
+        await query.execute(db=db)
+
+        return [cls.from_db(result.get("n")) for result in query.get_results()], await query.count(db=db)
```
ajtmccarty: this will actually cause the query to be executed two times, so I don't think that we don't gain much by putting the two queries into the same method
and this also makes it impossible to skip getting all the branches if the query is only for a count of all the branches

backend/infrahub/core/query/__init__.py
```diff
@@ -378,6 +379,7 @@ def __init__(
         self.has_errors: bool = False
 
         self.stats: QueryStats = QueryStats()
+        self.kwargs = kwargs
```
ajtmccarty: I'd remove these and add a `raw_filter` or `extra_filters` instance variable to `StandardNodeGetListQuery`

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,33 @@ async def branch_resolver(
...
     required=True,
 )
+
+
+async def infrahub_branch_resolver(
+    root: dict,  # noqa: ARG001
+    info: GraphQLResolveInfo,
+    **kwargs: Any,
+) -> dict[str, Any]:
+    limit = kwargs.pop("limit", 100)
```
ajtmccarty: instead of using `100` or `1000` as a limit, you can use `config.SETTINGS.database.query_size_limit`
it defaults to `5,000`, but I don't think that will be a problem for this query b/c retrieving the branches is pretty simple and other things are going to break way before anyone reaches 5000 branches

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,33 @@ async def branch_resolver(
...
+    offset = kwargs.pop("offset", 0)
+    fields = extract_graphql_fields(info)
+    branches, count = await InfrahubBranchType.get_list_and_count(
+        fields=fields.get("edges", {}).get("node", {}),
+        graphql_context=info.context,
+        limit=limit,
+        offset=offset,
+        **kwargs,
+    )
+    return {"count": count, "edges": {"node": branches}}
```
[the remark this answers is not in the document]
ajtmccarty: I think the robot is right about this. the structure should be 
```
edges: [
    {
        node: {
            name: "branch-2", ...
        }
    }, ...
]
```
you'll need to update the test too

backend/infrahub/graphql/types/branch.py
```diff
@@ -44,3 +44,27 @@ async def get_list(
...
+
+    @classmethod
+    async def get_list_and_count(
+        cls,
+        fields: dict,
+        graphql_context: GraphqlContext,
+        **kwargs: Any,
+    ) -> tuple[list[dict[str, Any]], int]:
+        async with graphql_context.db.start_session(read_only=True) as db:
+            objs, count = await Branch.get_list_and_count(db=db, **kwargs)
```
ajtmccarty: I think it would be best to separate getting the branch objects from getting the total count of all branch objects and only do each part if it is actually requested

there is probably some efficiency to be gained if the request is for all branches and a count of all branches, in which case you could skip calling a separate query to get the count and just use the length of the list of branch objects, but I don't think that's absolutely required

backend/infrahub/graphql/queries/branch.py
ogenstad: I don't think we should set a default at all, mostly as it would be inline with the rest of the queries we have that are dynamically created. Also while it might not be clear with regards to our other queries since this is a query we define all by ourselves we can get rid of the `**kwargs` completely here:

```python
InfrahubBranchQueryList = Field(
    InfrahubBranchType,
    ids=List(ID),
    name=String(),
    offset=Int(default_value=0),
    limit=Int(default_value=100),
    description="Retrieve paginated information about active branches.",
    resolver=infrahub_branch_resolver,
    required=True,
)
```

Given this definition we know that we'll send in `ids`, `name`, `offset`, `limit` to the resolver function `infrahub_branch_resolver()` as such we can have all of these named parameters, that currently all of them would be optional and default to `None`.

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,33 @@ async def branch_resolver(
...
+
+
+InfrahubBranchQueryList = Field(
+    InfrahubBranchType,
+    ids=List(ID),
```
ogenstad: This should probably be a List of ID but defined as NotNull for the content of the list/array.

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,33 @@ async def branch_resolver(
...
+    name=String(),
+    offset=Int(default_value=0),
+    limit=Int(default_value=100),
```
ogenstad: As per above I'd remove the default values for these. If we at some point change it we should do that for all of the kinds and implement some default limit.

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -148,3 +148,93 @@ async def test_branch_query(
...
+                        node {
+                            name
+                            description
+                        }
+                    }
+                }
+            }
+        """
+        gql_params = await prepare_graphql_params(
+            db=db, include_subscription=False, branch=default_branch, service=service
```
ogenstad: FYI, as per #7360 we should avoid creating new graphql params where we use `include_subscription=False`.

backend/infrahub/graphql/queries/branch.py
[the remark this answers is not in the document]
solababs: @ajtmccarty we are using this because we want to maintain a consistent response like some of the other queries i.e.
```
query MyQuery {
  BuiltinTag {
    count
    edges {
      node {
        name {
          value
        }
        description {
          value
        }
      }
    }
  }
}
```
cc @ogenstad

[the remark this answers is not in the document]
ogenstad: I think the confusing part here is that coderabbit spots an issue and then hallucinates a solution that is incorrect. What we want is that "edges" should be a list/array and each "node" should be an object as an entry in that array. A potential area of confusion here is that you don't see this within the GraphQL query which is correct as per the above however I think the response might not be? What I'd do is to create an environment with a few BuiltinTag objects and issue the above query and observe what the response looks like and then see if the structure matches what we now get from the `InfrahubBranch` query.

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,29 @@ async def branch_resolver(
...
+    info: GraphQLResolveInfo,
+    limit: int | None = None,
+    offset: int | None = None,
+) -> dict[str, Any]:
+    fields = extract_graphql_fields(info)
+    branches = await BranchType.get_list(
+        graphql_context=info.context, fields=fields.get("edges", {}).get("node", {}), limit=limit, offset=offset
+    )
+    count = await InfrahubBranchType.get_list_count(graphql_context=info.context)
+    return {"count": count, "edges": {"node": branches}}
```
ogenstad: I think what Aaron meant earlier regarding the separation of count and edges is that users can send in queries that request different information:

In this case we'd expect to return both edges and branches:
```graphql
query MyQuery {
  InfrahubBranch {
    count
    edges {
      node {
        name
      }
    }
  }
}
```

If we however have a query like the following one we don't need to send an additional query for the "count" since the request doesn't expect to have this information:
```graphql
query MyQuery {
  InfrahubBranch {
    edges {
      node {
        name
      }
    }
  }
}
```

Finally it could be that someone just wants to count the branches, and in this case we don't need to query the database for the details of each branch.

```graphql
query MyQuery {
  InfrahubBranch {
    count
  }
}
```

One thing that I noticed is that we have different filters for `BranchType.get_list` compared to `InfrahubBranchType.get_list_count` where we filter out the `-global-` branch with .get_list() but not with `get_list_count()`. So the count returns an extra number which is never seen.

A comment regarding the structure. We'd expect "edges" to be an array, instead of having "node" be an array. I.e. "edges" should be an array and "node" should be an object.
```python
return {"count": count, "edges": {"node": branches}}
```

On the topic of structure I think we should probably align the output that we get so that it matches the other nodes. I.e. more of a move away from what we had in the original `Branch` query. I.e. I think that a query would instead look like this:

```graphql
```graphql
query MyQuery {
  InfrahubBranch {
    edges {
      node {
        name {
           value   # <-- This is the new part we don't currently have.
        }
      }
    }
  }
}
```
```

File: backend/tests/unit/graphql/queries/test_branch.py [hunk 11]
```diff
@@ -148,3 +141,90 @@ async def test_branch_query(
...
+        all_branches = await graphql(
+            schema=gql_params.schema,
+            source=query,
+            context_value=gql_params.context,
+            root_value=None,
+            variable_values={},
+        )
+        assert all_branches.errors is None
+        assert all_branches.data
+        assert all_branches.data["InfrahubBranch"]["count"] == 13
```
ogenstad: This highlights the issue I mentioned above with the count. When we start out with the default branch and then create 10 branches. Probably the `-global-` branch is part of the count. And also one other branch. I'm not sure if this is due to the testing environment an we'd have some duplicate of the default branch. It would be good to determine where the extra branch that isn't the global branch is.

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -148,3 +141,90 @@ async def test_branch_query(
...
+            },
+            *[
+                {
+                    "description": f"sample description {i}",
+                    "name": f"sample-branch-{i}",
+                }
+                for i in range(10)
+            ],
+        ]
+        assert all_branches.data["InfrahubBranch"]["edges"]["node"].sort(
```
ogenstad: As CodeRabbit is mentioning `all_branches.data["InfrahubBranch"]["edges"]["node"]` should not be a valid field as `all_branches.data["InfrahubBranch"]["edges"]` should be an array.

ogenstad:
@solababs, I think we're close to the first iteration of this query now. I see that there's a slight issue with how the branch IDs are presented if you start an environment and compare the responses to these two queries:

```graphql
query MyQuery {
  InfrahubBranch {
    count
    edges {
      node {
        id
        name {
          value
        }
      }
    }
  }
  Branch {
    id
    name
  }
}
```

I think you're returning the internal Neo4j ID of these nodes which is typically different from the ID we present to the external world.

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,33 @@ async def branch_resolver(
...
+        result["edges"] = [{"node": branch} for branch in branches]
+    if "count" in fields:
+        result["count"] = await InfrahubBranchType.get_list_count(graphql_context=info.context)
+    return result
+
+
+InfrahubBranchQueryList = Field(
+    InfrahubBranchType,
+    ids=List(of_type=NonNull(ID)),
+    name=String(),
```
ogenstad: These parameters doesn't seem to work now.

```graphql
query MyQuery {
  InfrahubBranch(name: "branch2") {
    count
    edges {
      node {
        id
        name {
          value
        }
      }
    }
  }
}
```

I.e `infrahub_branch_resolver()` doesn't accept the parameters `ids` or `name` so the above query would fail. I think perhaps we can just remove both of them in this first iteration and then consider which filter options we want and need. We will want to add some of them for sure. But it could potentially be quite verbose if we start to add all of the variations that we dynamically generate for other node types, so I think that we can have a discussion around that after this PR is merge.

I.e. if we were to follow the current approach we'd have:

* name__value
* name__values
* description__value
* description__values
* etc for all attributes.

backend/infrahub/graphql/types/branch.py
```diff
@@ -46,4 +51,80 @@ async def get_list(
...
+
+class InfrahubBranch(BranchType):
+    id = String(required=True)
+    created_at = String(required=False)
+
+    name = Field(RequiredStringValueField, required=True)
+    description = Field(NonRequiredStringValueField, required=False)
+    origin_branch = Field(NonRequiredStringValueField, required=False)
+    branched_from = Field(NonRequiredStringValueField, required=False)
+    status = Field(StatusField, required=True)
```
ogenstad, line 83: Here we override `status` from the parent `BranchType` class but we don't override `graph_version`

So a query ends up being:

```graphql
query MyQuery {
  InfrahubBranch {
    count
    edges {
      node {
        graph_version
        status {
          value
        }
        name {
          value
        }
      }
    }
  }
}
```

backend/infrahub/graphql/queries/branch.py
```diff
@@ -28,3 +28,31 @@ async def branch_resolver(
     resolver=branch_resolver,
     required=True,
 )
+
+
+async def infrahub_branch_resolver(
+    root: dict,  # noqa: ARG001
+    info: GraphQLResolveInfo,
+    limit: int | None = None,
+    offset: int | None = None,
```
ogenstad, line 38: As CodeRabbit suggests it would be good to catch any negative values for `limit` or `offset` and if present raise a ValidationError back to the user.

ogenstad (approved):

PR #7419, opened 2025-10-16
fix(tasks): version bump for helm charts

fatih-acar:
When bumping X.Y.9 to X.Y.10, the Helm chart version would not get bumped since the comparison was done on plain strings (it would not detect a new version).
Make use of the Python packaging Version type to compare versions instead and solve this issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version comparison and parsing logic to properly handle prerelease versions and various edge cases in release version calculations

* **Refactor**
  * Enhanced internal version management system with more robust version comparisons and accurate version increment calculations
  * Optimized dependency handling in release tooling to reduce unnecessary imports

### DIFF
--- a/tasks/release.py
+++ b/tasks/release.py
@@ -100,8 +100,11 @@ def update_helm_chart(context: Context, chart_repo: str | None = "helm/") -> Non
     """Update helm/Chart.yaml with the current version from pyproject.toml."""
     print(" - [release] Update Helm chart")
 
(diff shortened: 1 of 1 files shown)


dgarros:
@fatih-acar what is the status on this PR, do we still need this change ?
please update it or close it, thanks

fatih-acar:
@dgarros yes this change is still required
the PR was just missing an approval... can I get one?

dgarros (approved):

PR #7424, opened 2025-10-17
Send deprecation warnings for schema API

ogenstad:
This PR adds the concept of returning deprecation warnings through the API endpoints when loading or checking a schema. The idea is to help users to determine when they are using deprecated features.

It will go together with this PR in the SDK repo: https://github.com/opsmill/infrahub-sdk-python/pull/582

<del>The change is quite simple, one thing I'm not completely sure of is the structure of the warning that currently looks like this:</del>

<del>

class SchemaWarningType(Enum):
    DEPRECATION = "deprecation"

class SchemaWarning(BaseModel):
    type: SchemaWarningType = Field(..., description="The type of warning")
    kind: str = Field(..., description="The kind impacted by the warning")
    message: str = Field(..., description="The message that describes the warning")

</del>

<del>I'm thinking that later we could add other warning types such as `STRICT` for when Infrahub isn't running in strict mode and we want to report back on those issues.</del>

<del>What I'm unsure of is the `kind: str` if we think that any future schema warning will always be tied to a kind, just now I can't think of other variations but it could just be that I'm missing it now. Any thoughts? We can always modify it later, but that might be a breaking change. I think it's fine as it is now, but not certain..</del>

### DIFF
--- a/backend/infrahub/core/schema/__init__.py
+++ b/backend/infrahub/core/schema/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import uuid
+from enum import Enum
 from typing import Any, TypeAlias
 
 from infrahub_sdk.utils import deep_merge_dict
@@ -44,6 +45,21 @@ class SchemaExtension(HashableModel):
     nodes: list[NodeExtensionSchema] = Field(default_factory=list)
 
 
+class SchemaWarningType(Enum):
+    DEPRECATION = "deprecation"
+
+
+class SchemaWarningKind(BaseModel):
+    kind: str = Field(..., description="The kind impacted by the warning")
+    field: str | None = Field(default=None, description="The attribute or relationship impacted by the warning")
+
+
+class SchemaWarning(BaseModel):
+    type: SchemaWarningType = Field(..., description="The type of warning")
+    kinds: list[SchemaWarningKind] = Field(default_factory=list, description="The kinds impacted by the warning")
+    message: str = Field(..., description="The message that describes the warning")
+
+
 class SchemaRoot(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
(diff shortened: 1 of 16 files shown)


dgarros:
Great improvement
Regarding the format, in addition to the kind, I think it would be good to also report the complete path to the warning (or as details as possible). For example, if the issue is on an attribute we should be able to report `kind` & `attribute`  etc ..
We have something similar for the error currently.

Also I don't think we should assume that a warning will always be associated with a single kind, I'm sure we'll find situations where a warning needs to be associated with multiple kinds or none at all. it would be good to ensure that the structure will support that.
we could return a list of paths, associated with a warning, this way we can have 0, 1 or many

ogenstad:
> Also I don't think we should assume that a warning will always be associated with a single kind, I'm sure we'll find situations where a warning needs to be associated with multiple kinds or none at all. it would be good to ensure that the structure will support that. we could return a list of paths, associated with a warning, this way we can have 0, 1 or many

Yes that was the part I wasn't sure about from the description. I don't see any situation just now what kind of error it would be that wouldn't include a kind. Might be that `kind` should instead be a list of objects that can be empty and that `attribute` is an optional value within that context?

backend/infrahub/core/schema/__init__.py
```diff
@@ -49,9 +49,14 @@ class SchemaWarningType(Enum):
     DEPRECATION = "deprecation"
 
 
+class SchemaWarningKind(BaseModel):
+    kind: str = Field(..., description="The kind impacted by the warning")
+    attribute: str | None = Field(default=None, description="The attribute impacted by the warning")
```
dgarros: I would replace `attribute` with `field` since it could also be a relationship.
We are using `field` or `element` in a few places to represent either an attribute or a relationship

dgarros (approved):

PR #7433, opened 2025-10-20
Refactor recurring setup jobs for Prefect automations

ogenstad:
This PR refactors how prefect automations are setup for display_labels and hfid, it also fixes an issue with order of operations as well as using the correct schema after modifying a schema within the validation for display_label.

Previously we'd setup the prefect automations when the server started and ensure that automations for display_label and hfid were up to date in Prefect during startup (this was done on the primary worker), a problem with this is that it could add some delay for the startup on the API server that got the primary role. The change here is that we instead check to see if the startup is part of the first initialization and if so we trigger a new function specifically to setup the display_labels and hfid, now this only ever happens once instead of at each startup.

Additionally the gathering and setup of HFID automations was missing from the upgrade command, this omission has now been corrected.

## Summary by CodeRabbit

* **New Features**
  * Initial system setup now automatically configures human-friendly identifiers and display labels for task managers on first deployment.

* **Improvements**
  * Enhanced initialization process to track first-run status and execute comprehensive setup procedures accordingly.
  * Optimized display label validation sequence for improved configuration handling.

### DIFF
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -540,8 +540,8 @@ class SchemaBranch:
         self.validate_identifiers()
         self.sync_uniqueness_constraints_and_unique_attributes()
         self.validate_uniqueness_constraints()
-        self.validate_display_label()
         self.validate_display_labels()
+        self.validate_display_label()
         self.validate_order_by()
         self.validate_default_filters()
         self.validate_parent_component()
@@ -789,6 +789,7 @@ class SchemaBranch:
                     )
                 self.set(name=name, schema=update_candidate)
 
+            node_schema = self.get(name=name, duplicate=False)
             if not node_schema.display_label:
                 continue
 
--- a/backend/infrahub/services/adapters/workflow/worker.py
+++ b/backend/infrahub/services/adapters/workflow/worker.py
@@ -6,7 +6,7 @@ from prefect.client.schemas.objects import StateType
 from prefect.deployments import run_deployment
 
 from infrahub.workers.utils import inject_context_parameter
-from infrahub.workflows.initialization import setup_task_manager
+from infrahub.workflows.initialization import setup_task_manager, setup_task_manager_identifiers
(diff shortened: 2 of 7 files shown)


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -540,8 +540,8 @@ def process_validate(self) -> None:
         self.validate_identifiers()
         self.sync_uniqueness_constraints_and_unique_attributes()
         self.validate_uniqueness_constraints()
-        self.validate_display_label()
         self.validate_display_labels()
+        self.validate_display_label()
```
ogenstad, line 544: I moved validate_display_label() to be after validate_display_labels() as there might be updates to the schema done in validate_display_labels that we want to have in place when converting.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -789,6 +789,7 @@ def validate_display_label(self) -> None:
                     )
                 self.set(name=name, schema=update_candidate)
 
+            node_schema = self.get(name=name, duplicate=False)
```
ogenstad, line 792: Since we update the schema using `update_candidate` we need to get the latest version from the schema before determining if `node_schema.display_label` is set.

backend/infrahub/services/adapters/workflow/worker.py
```diff
@@ -20,10 +20,13 @@
 
 class WorkflowWorkerExecution(InfrahubWorkflow):
     @staticmethod
-    async def initialize(component_is_primary_server: bool) -> None:
+    async def initialize(component_is_primary_server: bool, is_initial_setup: bool = False) -> None:
         if component_is_primary_server:
             await setup_task_manager()
 
+        if is_initial_setup:
+            await setup_task_manager_identifiers()
```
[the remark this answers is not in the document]
ogenstad: This is incorrect only one worker will ever set `is_initial_setup` to True as that part happens within a lock.

gmazoyer (approved):

PR #7434, opened 2025-10-20
Fix array based conversion of display labels for obsolete format

ogenstad:
For `display_labels` we supported both short and verbose attribute values i.e. both `["name"]` and `["name__value"]`, this change ensures that we also support the same when migrating `display_labels` to `display_label` within the SchemaBranch.

## Summary by CodeRabbit

* **Refactor**
  * Improved consistency of display label formatting logic through standardized component handling.

* **Tests**
  * Added comprehensive unit test coverage for display label computation, validating multiple formatting scenarios including single and multi-component labels with various configurations.

### DIFF
--- a/backend/tests/unit/core/schema/schema_branch/test_schema_convert_display_labels.py
+++ b/backend/tests/unit/core/schema/schema_branch/test_schema_convert_display_labels.py
@@ -1,0 +1,103 @@
+from dataclasses import dataclass
+
+import pytest
+
+from infrahub.core.schema import AttributeSchema, NodeSchema, SchemaRoot
+from infrahub.core.schema.schema_branch import SchemaBranch
+
+
+@dataclass
+class DisplayLabelTestCase:
+    name: str
+    schema_root: SchemaRoot
+    display_label: str
+
+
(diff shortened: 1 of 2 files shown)


backend/tests/unit/core/schema/schema_branch/test_schema_convert_display_labels.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@dataclass
+class DisplayLabelTestCase:
+    name: str
+    schema_root: SchemaRoot
+    display_label: str
+
+
+DISPLAY_LABEL_TEST_CASES: list[DisplayLabelTestCase] = [
```
ajtmccarty, line 16: good pattern. easy to read

ajtmccarty (approved):

backend/tests/unit/core/schema/schema_branch/test_schema_convert_display_labels.py
ogenstad: Yeah, I agree we have it in a few places and I think we should do more of it. Especially for pytest.params with multiple variables.

PR #7447, opened 2025-10-20
Allow rebase of branches during migration

gmazoyer:
These changes add a new flow to apply migrations (schema/data ones) inside branches. The upgrade process (via the provided command) is also updated with a new `--rebase-branches` flag to trigger branch rebases as part of the upgrade process.

A new `MigrationRequiringRebase` class is introduced with a couple of methods that need to be defined on classes inheriting it. One method defines what to do on the default branch while the other one defines what to do for all other branches.

The branch nodes can have a new `NEED_UPGRADE_REBASE` status and they also now have a `graph_version` property to know if a branch is up-to-date regarding the graph structure.

A new prefect event is also introduced and triggered when the `branch-migrate` flow completes for a branch.

## Summary by CodeRabbit

* **New Features**
  * Per-branch migration and rebase workflows; upgrade CLI gains --rebase-branches and can emit a Branch Migrated event.
  * Migration runner and workflow to detect/apply required branch migrations and mark branches as migrated.

* **Schema Updates**
  * Branch type adds graph_version field; BranchStatus adds NEED_UPGRADE_REBASE.

* **Documentation**
  * CLI and events docs updated for --rebase-branches and the Branch Migrated event.

* **Tests**
  * Unit tests added for the migration runner behavior.


schema/schema.graphql
```diff
@@ -323,6 +324,7 @@ enum BranchRelativePermissionDecision {
 enum BranchStatus {
   DELETING
   NEED_REBASE
+  NEED_UPGRADE_REBASE
   OPEN
 }
```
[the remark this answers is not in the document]
ogenstad: This is not possible as we might need to be able to resolve conflicts in the NEED_UPGRADE_REBASE branches.

backend/infrahub/core/branch/models.py
```diff
@@ -261,6 +261,10 @@ def get_branches_and_times_for_range(
 
         return start, end
 
+    async def create(self, db: InfrahubDatabase) -> bool:
+        self.graph_version = GRAPH_VERSION
```
ogenstad, line 265: One question here is how we manage this number after a rebase operation. Do we then bump the graph version? (I haven't yet checked if we already do this, just commenting some thoughts)

backend/infrahub/core/branch/tasks.py
```diff
@@ -49,8 +59,53 @@
...
 
 
+@flow(name="branch-migrate", flow_run_name="Apply migrations to branch {branch}")
+async def migrate_branch(branch: str, context: InfrahubContext, send_events: bool = True) -> None:
+    await add_tags(branches=[branch])
+
+    db = await get_database()
+    log = get_run_logger()
+
+    obj = await Branch.get_by_name(db=db, name=branch)
```
ogenstad: I think we always want this within a db.start_session() context manager?

backend/infrahub/core/branch/tasks.py
```diff
@@ -49,8 +59,53 @@
...
+    obj.status = BranchStatus.NEED_UPGRADE_REBASE
+    await obj.save(db=db)
+
+    try:
+        log.info(f"Running migrations for branch '{obj.name}'")
+        await migration_runner.run(db=db)
+    except MigrationFailureError as exc:
+        log.error(f"Failed to run migrations for branch '{obj.name}': {exc.errors}")
+        return
+
```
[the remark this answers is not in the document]
ogenstad: In what way would the CLI mark the branch as successfully rebased?

backend/infrahub/core/branch/tasks.py
```diff
@@ -49,8 +59,53 @@
...
+    migration_runner = MigrationRunner(branch=obj)
+    if not migration_runner.has_migrations():
+        log.info(f"No migrations detected for branch '{obj.name}'")
+        return
+
+    # Branch status will remain as so if the migration process fails
+    # This will help user to know that a branch is in an invalid state to be used properly and that actions need to be taken
+    obj.status = BranchStatus.NEED_UPGRADE_REBASE
+    await obj.save(db=db)
```
ogenstad: It could be that `obj.status` already is `BranchStatus.NEED_UPGRADE_REBASE` going into this right? If that's the case we don't need to save the branch back to the database. I.e. we only do that provided that the status is something else.

backend/infrahub/core/branch/tasks.py
```diff
@@ -49,8 +59,53 @@
...
+async def migrate_branch(branch: str, context: InfrahubContext, send_events: bool = True) -> None:
+    await add_tags(branches=[branch])
+
+    db = await get_database()
+    log = get_run_logger()
+
+    obj = await Branch.get_by_name(db=db, name=branch)
+
+    if obj.graph_version == GRAPH_VERSION:
+        log.info(f"Branch '{obj.name}' is up-to-date")
```
ogenstad: I think we can be a bit more verbose here and say that the graph version of the branch matches the expected version instead. Of it only says that the branch is up to date it's unclear what we mean if the data or schema or something else is up to date.

backend/infrahub/core/branch/tasks.py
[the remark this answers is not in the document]
ogenstad: @gmazoyer, this could be good to catch. Perhaps a query to check the branch graph version instead of just assuming it was successful?

backend/infrahub/core/migrations/graph/__init__.py
```diff
@@ -113,7 +111,7 @@ async def get_graph_migrations(
 
 def get_migration_by_number(
     migration_number: int | str,
-) -> GraphMigration | InternalSchemaMigration | ArbitraryMigration:
+) -> GraphMigration | InternalSchemaMigration | ArbitraryMigration | MigrationRequiringRebase:
```
ajtmccarty: might be good to add a `MigrationTypes` constant for these classes. they show up together a lot

backend/infrahub/events/branch_action.py
```diff
@@ -128,3 +128,29 @@ def get_messages(self) -> list[InfrahubMessage]:
...
+            "infrahub.branch.id": self.branch_id,
+            "infrahub.branch.name": self.branch_name,
+        }
+
+    def get_messages(self) -> list[InfrahubMessage]:
+        events: list[InfrahubMessage] = [
+            # EventBranchMigrated(
+            #     branch=self.branch,
+            #     meta=self.get_message_meta(),
+            # ),
```
ajtmccarty, line 153: leftover?

backend/tests/unit/core/migrations/test_runner.py
```diff
@@ -0,0 +1,59 @@
...
+    branch = await create_branch(db=db, branch_name="foo")
+
+    runner = MigrationRunner(branch=branch)
+    assert not runner.applicable_migrations
+
+    branch.graph_version = None
+    runner = MigrationRunner(branch=branch)
+    await branch.save(db=db)
+    assert runner.applicable_migrations
+    assert [m.name for m in runner.applicable_migrations] == ["043_backfill_hfid_display_label_in_db"]
```
ajtmccarty: we're going to have to update this every time we add a new MigrationRequiringRebase

ajtmccarty (approved):
this looks pretty good to me
just want to make sure that prefect does not need to be running during a migration with rebase. looks like it shouldn't be necessary

backend/infrahub/events/branch_action.py
gmazoyer: It's actually not, all the other `get_messages` methods have this in their bodies, so I put this for consistency. I don't know what the initial intent was.

backend/infrahub/core/branch/models.py
gmazoyer: This number only changes in the `branch-migrate` flow, if there are no migrations to apply or if migrations were applied succesfully. This means that if a branch is created at graph version `n` and the global graph version is changed to `m`, the branch's graph version will never be `m` unless a rebase operation happens.

PR #7461, opened 2025-10-22
Avoid importing schema from demo edge

ogenstad:
Due to a recent change in the bundled `infrahub-demo-edge` repository within the test fixtures we were importing a schema update with the repository. This is currently causing a lot of noise in the prefect environment for the end-to-end tests with Playwright as the schema now differs from the main branch and the other branches (this is a separate issue that needs to be resolved)

In this PR I'm reverting the previous change to this `infrahub-demo-edge` repository so it matches what we expect to see there from an end to end test perspective. As the repository was used in an integration test I created a copy of the repository so that the integration test can still use it.


ajtmccarty (approved):
so you saw that I had 1700 line PR and felt like you had to do a bigger one

PR #7463, opened 2025-10-22
fix(backend, tests): use DI for database instance

fatih-acar:


ajtmccarty:
@fatih-acar the diff query pagination issue should be fixed in https://github.com/opsmill/infrahub/pull/7566, but that PR does not include the dependency changes that you've made here. not sure if you want to keep those or not

fatih-acar:
Thanks @ajtmccarty for taking care of the real issue, I will let only the DI changes in this PR then

fatih-acar:
Done in https://github.com/opsmill/infrahub/pull/7746

PR #7464, opened 2025-10-23
apply non-template relationships in sub-templates

### LINKED ISSUE #7430: bug: Component Template relationship missing when creating parent object
### Component

_No response_

### Infrahub version

1.4.11

### Current Behavior

When a relationship is defined on a `component` relationship within a parent object template, that relationship doesn't show up after object creation on the `component` relationship.

Device -> Interface (Component) -> Tagged VLAN (Won't have the relationship on device creation)

### Expected Behavior

The defined relationship is defined on a component template when an object is instantiated selecting a parent object template.

### Steps to Reproduce

1. Clone down `infrahub` repo and spin up infrahub using `invoke demo.start`
2. Add `generate_template: true` to the `InfraDevice` model found in `models/base/dcim.yml`
3. Load the schemas
4. Load the following object file:

```
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationContinent
  data:
    - name: North America
      children:
        kind: LocationCountry
        data:
          - name: USA
            children:
              kind: LocationSite
              data:
                - name: DEN01
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: InfraVLAN
  data:
    - name: "SERVERS"
      vlan_id: 100
      site: DEN01
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateInfraDevice
  data:
    - template_name: Core Device
      status: provisioning
      role: core
      site: DEN01
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateInfraInterfaceL2
  data:
    - template_name: "Ethernet1/1"
      l2_mode: Access
      speed: 1000
      name: Ethernet1/1
      device: Core Device
      tagged_vlan:
        data:
          - name: SERVERS
            vlan_id: 100
```

5. Check that the template has an interface and the interface has the VLAN assigned to it
6. Create a new device using the `Core Device` template
7. Check that the VLAN is not associated with the interface of the newly created device.

### Additional Information

_No response_

ajtmccarty:
fixes #7430
IFC-1931

add handling for non-template relationships on sub-templates

not completely certain this is the correct way to handle it, but it might be

long description of the problem from the issue below

> When we generate a `TemplateSchema` for a given Node or GenericSchema, we also generate `TemplateSchema`s for any relationships of `kind=Component` on the initial schema. And we do this recursively down through `kind=Component` relationships on the sub-templates, and the sub-sub-templates, and so on.

> For example, if I have a `Continent` schema that includes a `Component` relationship to a `Country` schema that includes a `Component` relationship to a `State` schema, then the `TemplateContinent` schema would include a relationship to a `TemplateCountry` schema that would include a relationship to a `TemplateState` schema

> If any of these `Template` schemas have non-Component relationships, then those relationships should be able to connect to regular existing objects as we would expect on any relationship of any object. So if we update the `State` schema from before to include a `kind=Generic` relationship to a `City` schema, then we would be able to link an instance of `City` to a `TemplateState`. If I have an instance of a `TemplateState` schema (say, `WashingtonStateTemplate`) linked to an instance of a non-template `City` schema (say, `Olympia`), then I would expect any `State` instance created using the `WashingtonStateTemplate` to be linked to the same `Olympia` instance of the `City` schema.

> This works at the top-level of the template schema, meaning it works for relationships defined on the template being used. But it does not work for sub-templates. For example, I want to create an instance of the `Country` schema that is linked to the `WashingtonStateTemplate` schema, the end result should be a new instance of the `Country` schema linked to a new instance of the `State` schema linked to the existing `Olympia` instance of the `City` schema, but instead this final link from `State` to `Olympia` is not created

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed object-creation relationship handling so templates and real objects link correctly, improving accuracy for nested/sub-template relationships.

* **Tests**
  * Added GraphQL tests validating creation and linking between template-based and real objects (includes an additional/duplicate test case).

* **Documentation**
  * Added changelog entry documenting the sub-template object linking fix.

### DIFF
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -57,14 +57,25 @@ async def extract_peer_data(
 
     for rel in template_peer.get_schema().relationship_names:
         rel_manager: RelationshipManager = getattr(template_peer, rel)
-        if (
-            rel_manager.schema.kind not in [RelationshipKind.COMPONENT, RelationshipKind.PARENT]
-            or rel_manager.schema.name not in obj_peer_schema.relationship_names
-        ):
+
+        if rel_manager.schema.name not in obj_peer_schema.relationship_names:
             continue
 
-        if list(await rel_manager.get_peers(db=db)) == [current_template.id]:
+        peers_map = await rel_manager.get_peers(db=db)
+        if rel_manager.schema.kind in [RelationshipKind.COMPONENT, RelationshipKind.PARENT] and list(
+            peers_map.keys()
+        ) == [current_template.id]:
             obj_peer_data[rel] = {"id": parent_obj.id}
+            continue
+
+        rel_peer_ids = []
+        for peer_id, peer_object in peers_map.items():
+            # deeper templates are handled in the next level of recursion
+            if peer_object.get_schema().is_template_schema:
+                continue
+            rel_peer_ids.append({"id": peer_id})
+
+        obj_peer_data[rel] = rel_peer_ids
 
     return obj_peer_data
 
--- a/backend/tests/unit/graphql/test_mutation_create.py
+++ b/backend/tests/unit/graphql/test_mutation_create.py
@@ -3,7 +3,7 @@ import pytest
 from infrahub import config
 from infrahub.core import registry
 from infrahub.core.branch.models import Branch
-from infrahub.core.constants import InfrahubKind, SchemaPathType
+from infrahub.core.constants import InfrahubKind, RelationshipKind, SchemaPathType
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager
 from infrahub.core.migrations.schema.node_kind_update import NodeKindUpdateMigration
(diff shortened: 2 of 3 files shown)


backend/infrahub/core/node/create.py
```diff
@@ -57,14 +57,25 @@ async def extract_peer_data(
...
-        ):
+
+        if rel_manager.schema.name not in obj_peer_schema.relationship_names:
             continue
 
-        if list(await rel_manager.get_peers(db=db)) == [current_template.id]:
+        peers_map = await rel_manager.get_peers(db=db)
+        if rel_manager.schema.kind in [RelationshipKind.COMPONENT, RelationshipKind.PARENT] and list(
+            peers_map.keys()
+        ) == [current_template.id]:
```
ajtmccarty, line 67: pre-existing logic for linking a sub-template to its parent

backend/infrahub/core/node/create.py
```diff
@@ -57,14 +57,25 @@ async def extract_peer_data(
...
+            continue
+
+        rel_peer_ids = []
+        for peer_id, peer_object in peers_map.items():
+            # deeper templates are handled in the next level of recursion
+            if peer_object.get_schema().is_template_schema:
+                continue
+            rel_peer_ids.append({"id": peer_id})
+
+        obj_peer_data[rel] = rel_peer_ids
```
ajtmccarty, line 78: new logic to allow linking existing objects to the instance being created. ignores child template schema because they will be handled in the next level of recursion in the `handle_template_relationships` stack

gmazoyer (approved):

PR #7471, opened 2025-10-24
improve performance of migration to set HFID/display label (only on default/global branches)

ajtmccarty:
a few new queries to improve the performance of the migration to set display label and human_friendly_id values for all active nodes on the default and global branches

logic to handle non-default branches will come later. it first requires logic to identify which nodes need to be updated on a given branch

migration works basically like this
- for each node schema that is not a SchemaNode, SchemaGeneric, SchemaRelationship, or SchemaAttribute
  - if this node schema has display_labels or human_friendly_id defined
    - in batches of 1000, do the following
      - get the current attribute values or attribute values of relationships that are used in the display labels/ HFID
      - set the display labels
      - set the HFID

there is some finagling required in the query to set attribute values when working with large-type attributes (List, JSON, TextArea), of which human_friendly_id happens to be one b/c these attributes are not indexed. Looking for an AttributeValue vertex with a particular value in the whole database requires looking at every AttributeValue vertex and, in a large database, this will take a long time if we need to do it once for every node. My solution is probably a little hard to read, but it is effectively building our own index for each batch of 1000 nodes

one thing to please keep an eye out for. I am not sure if I am using `display_label` vs `display_labels` in the correct places.

### DIFF
--- a/backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
+++ b/backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
@@ -2,14 +2,15 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
-from rich.progress import Progress
+import ujson
+from rich.progress import Progress, TaskID
 
 from infrahub.core import registry
 from infrahub.core.constants import GLOBAL_BRANCH_NAME, BranchSupportType, RelationshipDirection
 from infrahub.core.initialization import get_root_node
 from infrahub.core.migrations.shared import MigrationResult
 from infrahub.core.query import Query, QueryType
-from infrahub.core.schema import AttributeSchema, SchemaRoot, internal_schema
+from infrahub.core.schema import AttributeSchema, NodeSchema, SchemaRoot, internal_schema
 from infrahub.core.schema.manager import SchemaManager
 from infrahub.types import is_large_attribute_type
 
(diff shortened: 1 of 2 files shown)


backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
```diff
@@ -1,86 +1,489 @@
...
+WHERE e.branch IN $branch_names
+AND e.to IS NULL
+AND e.status = "active"
+// ------------
+// Order and limit the Nodes
+// ------------
+WITH DISTINCT n
+ORDER BY elementId(n)
+SKIP toInteger($offset)
+LIMIT toInteger($limit)
```
ogenstad, line 123: This is not really a comment on this PR as such but something that I'd like to highlight for the future. To me it looks like the `GetPathDetailsDefaultBranch` requires us to set `offset` and `limit`. However within `GetPathDetailsDefaultBranch.init` this requirement isn't clear.

I don't actually know what happens here if we'd call this without providing these values if it would be `SKIP 0` or `SKIP null`. Even if it might be more verbose I think we should consider having either a unique entrypoint for all queries. Alternatively just using the `__init__()` method of each query to provide any required information and get rid of all of the `**kwargs` that just masks any requirements.

ogenstad:
Regarding the failing test.

```python
FAILED backend/tests/unit/core/migrations/graph/test_042_043.py::TestMigration042::test_migration_042_043 - infrahub.exceptions.SchemaNotFoundError: 
        Unable to find the schema 'TestingWidget' in the registry
```

I wonder if this is because load the node kind to the registry within the test setup and then after that load the schema from the database during the migration. I'd guess that `TestingWidget` gets removed. Might be solved by adding the schema to the database in the test setup phase or review how the registry is cleared out during the migrations.

backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
ajtmccarty: I agree. I don't like how the queries are built or instantiated. It's too much inheritance and typing basically does not work for any of them
if you try to run this particular query without an offset of limit, then you get this error when it tries to run
```
  {code: Neo.ClientError.Statement.TypeError} {message: Expected a numeric value for org.neo4j.cypher.internal.runtime.interpreted.commands.expressions.NumericHelper$@5bb30d10, but
got null}
```
which means nothing to the user running the migration, but this query should only be used in this migration and we'll always set the limit and offset for it

so my conclusion is that this is a problem for a different day

ogenstad: For sure, I didn't mean that we should change this query. More a reflection and something that could go into that "redesign query engine" idea.

PR #7482, opened 2025-10-27
IFC-1983: Add config parameter for selective git sync

solababs:
This PR adds `sync_branch_names` parameter to git settings so that users can add branches that they want to be synced with git instead of syncing all branches like we currently do. This parameter can also be a regex so that we can match branch names when syncing

## Summary by CodeRabbit

* **New Features**
  * Git branch synchronization now accepts configurable branch-name patterns (plain names or regex) with validation and clear errors for invalid patterns.
* **Documentation**
  * Added docs and examples for the new Git branch sync configuration option, including pattern examples and default behavior.
* **Tests**
  * Added unit tests covering valid and invalid branch-pattern configurations.


backend/infrahub/config.py
```diff
@@ -444,6 +445,28 @@ class GitSettings(BaseSettings):
...
+        default_factory=list,
+        description="Names or regex of branches to sync with git e.g. 'main', '/infrahub/.*/', '/release/.*/' ",
+    )
+    _compiled_branch_names: list[str] = PrivateAttr(default_factory=list)
+
+    @model_validator(mode="after")
+    def validate_sync_branch_names(self) -> Self:
+        compiled_branch_names = []
+        for branch_name in self.sync_branch_names:
+            if not branch_name.startswith("/") and not branch_name.endswith("/"):
```
ogenstad: I think we can assume that any of the entries should be compiled to regex that way we could treat them all the same. I.e. if we have a branch called "my-branch1" it would work just the same to do a `re.compile("my-branch1")` and use regex to validate that name. Or people could write regular expressions using a different format such as `"^branch-"`.

backend/infrahub/config.py
```diff
@@ -444,6 +445,19 @@ class GitSettings(BaseSettings):
         default_factory=default_append_git_suffix_domains,
         description="Automatically append '.git' to HTTP URLs if for these domains.",
     )
+    sync_branch_names: list[str] = Field(
+        default_factory=list,
+        description="Names or regex of branches to sync with git e.g. 'main', 'infrahub/.*', 'release/.*', '^branch-'",
```
ogenstad: The issue we're trying to solve is to allow users to configure the names of branches that will be created and imported in Infrahub provided that they match this setting. So if a user creates a branch within Infrahub and configures it to sync with Git that branch would still be expected to sync with Git, as such I think we should slightly update the description here to indicate the filter controls which branches from Git that are automatically created within Infrahub during import.

backend/tests/unit/test_config.py
```diff
@@ -19,3 +21,14 @@ def test_load_sso_config(helper: TestHelper) -> None:
...
+
+def test_valid_git_settings__sync_branch_names():
+    sync_branch_names = ["main", "infrahub/.*", "release/.*"]
+    git_settings = GitSettings(sync_branch_names=sync_branch_names)
+    assert git_settings.sync_branch_names == sync_branch_names
+
+
+def test_invalid_git_settings__sync_branch_names():
+    with pytest.raises(ValueError):
+        GitSettings(sync_branch_names=["main", "infrahub/.*", "release/.*", "a[b"])
```
ogenstad: For a test like this I'd also add a match statement so that it's clear that it's tracking the expected error, i.e. "Invalid regex pattern for sync_branch_names"

backend/infrahub/config.py
```diff
@@ -444,6 +445,19 @@ class GitSettings(BaseSettings):
...
+    )
+
+    @model_validator(mode="after")
+    def validate_sync_branch_names(self) -> Self:
+        for branch_name in self.sync_branch_names:
```
ogenstad: For clairty since this might not be a branch_name I'd suggest we have something like `branch_filter` instead of `branch_name`.

backend/infrahub/config.py
```diff
@@ -444,6 +445,19 @@ class GitSettings(BaseSettings):
         default_factory=default_append_git_suffix_domains,
         description="Automatically append '.git' to HTTP URLs if for these domains.",
     )
+    sync_branch_names: list[str] = Field(
```
ogenstad: Based on the comment below regarding the description, I'm not sure about this name if it's fine or if we could have something that clarifies the intent a bit more?

backend/infrahub/config.py
```diff
@@ -444,6 +445,25 @@ class GitSettings(BaseSettings):
         default_factory=default_append_git_suffix_domains,
         description="Automatically append '.git' to HTTP URLs if for these domains.",
     )
+    import_sync_branch_names: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Names or regex of branches to be created in infrahub during import "
+            "e.g. 'main', 'infrahub/.*', 'release/.*', '^branch-'. "
```
ogenstad: The "main" branch (or default branch really) would always be synced and should be outside of this setting, as such I think we should remove 'main' from the examples here and just keep the rest of them.

ogenstad (approved):

PR #7483, opened 2025-10-27
chore: update click version to 8.2.1 in poetry.lock and pyproject.toml

minitriga:


fatih-acar:
Looks like CI is failing...
We should probably revert `click` to the latest working version we had (until we upgrade typer AND click)
https://github.com/opsmill/infrahub/blob/stable/poetry.lock#L648
`8.1.7` seems to be the right one

ajtmccarty (approved):

PR #7486, opened 2025-10-27
IFC-1984: Update git sync to consider new configuration

solababs:
This PR updates the git sync to check the new configuration `import_sync_branch_names` and if it is present it matches the branch name with the names in the config and skips adding it as a remote branch if it does not match with exception to allow
- branch name that == default branch
- branch with `sync_with_git` == `True`
- branch that `is_default` == `True`

## Summary by CodeRabbit

* **New Features**
  * Added branch filtering for remote sync: repository sync now respects configurable import-sync patterns, so only matching branches are considered while others are skipped during synchronization.

* **Tests**
  * Added tests and fixtures covering various branch-filtering scenarios to ensure correct behavior with and without import-sync patterns.


backend/infrahub/core/branch/tasks.py
```diff
@@ -355,6 +356,16 @@ async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> N
...
+            if (
+                config.SETTINGS.git.import_sync_branch_names
+                and obj.name not in {GLOBAL_BRANCH_NAME, registry.default_branch}
+                and not obj.is_default
+                and not obj.sync_with_git
+                and not branch_name_in_sync_branches(branch_short_name=obj.name)
+            ):
+                raise ValidationError(
+                    f"Branch name '{obj.name}' does not match import_sync_branch_names: {config.SETTINGS.git.import_sync_branch_names}"
+                )
```
ogenstad: I don't think we should restrict the creation of these branches within Infrahub itself. I.e. the setting should only restrict which branches are automatically created due to imports from Git, not control the naming of branches that users can create within Infrahub.

backend/infrahub/core/utils.py
```diff
@@ -278,3 +279,10 @@ def __init_subclass__(cls, **meta_options: dict[str, Any]) -> None:
     @classmethod
     def __init_subclass_with_meta__(cls, **meta_options: dict[str, Any]) -> None:
         """This method just terminates the super() chain"""
+
+
+def branch_name_in_sync_branches(branch_short_name: str) -> bool:
+    for branch_filter in config.SETTINGS.git.import_sync_branch_names:
+        if re.fullmatch(branch_filter, branch_short_name) or branch_filter == branch_short_name:
+            return True
+    return False
```
ogenstad: I think we want to avoid adding more stuff within the infrahub.core namespace so I'd instead move this to infrahub.git.

One thought would be if we saved the compiled regex entries directly within the config object in a private attribute we could access them all and just do a .match for all entries within the list as we'd already have them compiled.

Also I'd suggest that we combine this function with the logic we have in git/base.py:

```python
if config.SETTINGS.git.import_sync_branch_names and not branch_name_in_sync_branches
```

Currently the branch_name_in_sync_branches() function might behave in an unexpected way if used in isolation and `config.SETTINGS.git.import_sync_branch_names` didn't contain any entries. If we instead had a function that would return True if we should skip the import, if users hadn't configured this option it would always return True, otherwise it would return True if the branch name didn't contain a match for the regex filters.

backend/infrahub/git/base.py
```diff
@@ -500,8 +503,20 @@ def get_branches_from_remote(self) -> dict[str, BranchInRemote]:
             if short_name == "HEAD":
                 continue
 
+            if config.SETTINGS.git.import_sync_branch_names and not branch_name_in_sync_branches(
+                branch_short_name=short_name
+            ):
+                skipped_branch_names.append(short_name)
+                continue
```
ogenstad: Here we must also account for the default branch, any branch in Infrahub marked as "sync with git", and potentially any "default branch" configured on the repo itself (this is typically always "main" but it can differ).

backend/tests/unit/git/test_git_repository.py
```diff
@@ -279,6 +280,17 @@ async def test_get_branches_from_remote(git_repo_01: InfrahubRepository):
...
 
+async def test_get_branches_from_remote_in_import_sync_branch_names(git_repo_01: InfrahubRepository):
+    config.SETTINGS.git.import_sync_branch_names = ["branch.*"]
+    repo = git_repo_01
+
+    remote_branches = repo.get_branches_from_remote()
+    assert isinstance(remote_branches, dict)
+    assert sorted(remote_branches.keys()) == ["branch01", "branch02"]
+
+    config.SETTINGS.git.import_sync_branch_names = []
```
ogenstad: Instead of changing the config like this the test and resetting it back to another value afterwards I'd suggest that we use a fixture similar to this one:

```python
@pytest.fixture
def query_limit_of_one() -> Generator[None, None, None]:
    original_query_size_limit = config.SETTINGS.database.query_size_limit
    config.SETTINGS.database.query_size_limit = 1
    yield
    config.SETTINGS.database.query_size_limit = original_query_size_limit
```

backend/tests/unit/git/test_git_repository.py
```diff
@@ -279,6 +280,17 @@ async def test_get_branches_from_remote(git_repo_01: InfrahubRepository):
     assert sorted(remote_branches.keys()) == ["branch01", "branch02", "clean-branch", "main"]
 
 
+async def test_get_branches_from_remote_in_import_sync_branch_names(git_repo_01: InfrahubRepository):
```
ogenstad: Minor detail here, I just merged a PR to `develop` that went through and added return values to all of the tests that didn't have them. There's still some work to be done with other functions before it's enforced but please add a `) -> None:` as the return value here.

backend/infrahub/git/base.py
```diff
@@ -733,6 +774,10 @@ async def fetch(self) -> bool:
 
         return True
 
+    async def get_filtered_remote_branches(self) -> dict[str, BranchInRemote]:
+        remote_branches = self.get_branches_from_remote()
+        return await self.filter_remote_branches(branches=remote_branches)
```
ogenstad: As we don't call filter_remote_branches in other locations and always call .get_filtered_remote_branches() I wonder if it's clearer if we remove the filter_remote_branches() method and just have that code inline within get_filtered_remote_branches()?

backend/infrahub/git/base.py
```diff
@@ -484,6 +487,44 @@ async def get_branches_from_graph(self) -> dict[str, BranchInGraph]:
...
+            return branches
+
+        filtered_branches = {}
+        skipped_branch_names = []
+
+        for short_name, branch_data in branches.items():
+            branch = None
+
+            try:
+                branch = await self.sdk.branch.get(branch_name=short_name)
```
ogenstad: I think this could potentially be problematic if the imported repository has a very high number of branches as we'd always try to get the branch name from the API using the SDK. The branches should always be loaded in the registry. so if we instead have something like this we'd just look in a local dictionary and avoid the delay.

```python
branch = registry.get_branch_from_registry(branch=short_name)
```

Alternatively we could add a method in the registry that would return all registry branches (i.e. just return the self.branch dictionary)

backend/infrahub/git/base.py
```diff
@@ -484,6 +487,44 @@ async def get_branches_from_graph(self) -> dict[str, BranchInGraph]:
...
+            except BranchNotFoundError:
+                ...
+
+            branch_exists_import_sync_condition = branch and (
+                branch.name not in {GLOBAL_BRANCH_NAME, registry.default_branch}
```
ogenstad: The `GLOBAL_BRANCH_NAME` is `-global-` which is an invalid branch name in Git and as such should never come from a git remote. However the registry itself could have a default branch (self.default_branch) in most cases it will be exactly the same but it doesn't have to be, for that reason we should also add the default branch of the repository to this entry.

backend/infrahub/git/base.py
```diff
@@ -484,6 +487,44 @@ async def get_branches_from_graph(self) -> dict[str, BranchInGraph]:
...
+                and not branch.is_default
```
ogenstad: As we already check for `registry.default_branch` perhaps this "not branch.is_default` is redundant?

ogenstad (approved):
Can you also please add a towncrier newsfragment of the "added" kind that adds this as a new feature?

ogenstad (approved):

PR #7508, opened 2025-10-29
fix(backend): granular locking and transaction fixes

### LINKED ISSUE #7254: bug: number resource pools can allocate the same resource when allocation requests run in parallel
### Component

API Server / GraphQL

### Infrahub version

1.4.7

### Current Behavior

When running 2 or more mutations in parallel, that will create an object for which a resource needs to be allocated from the same number pool, it might happen that the same number out of the pool is allocated to both objects.  

### Expected Behavior

Each object gets a unique number allocated from the number pool

### Steps to Reproduce

- spin up an instance of Infrahub
- install the latest version of `infrahub-sdk`
- load this schema
```yaml
---
version: "1.0"
nodes:
  - name: Object
    namespace: Infra
    attributes:
      - name: name
        kind: Text
      - name: number
        kind: NumberPool
        read_only: true
        optional: false
        parameters:
          start_range: 100
          end_range: 999
```
- create the following script
```python
import os
from infrahub_sdk import InfrahubClientSync, Config

config = Config(default_branch=os.getenv("BRANCH", "main"))
client = InfrahubClientSync(config=config)

obj = client.create("InfraOtherObject", name="test")
obj.save(allow_upsert=True)
```
- run 2 instance of the script in parallel
```bash
.venv/bin/python test.py & .venv/bin/python test.py &
```
- look at the resulting objects in Infrahub, they likely will have the same value allocated for the number attribute (if not, repeat running multiple instance of the script in parallel)

### Additional Information

We have also seen this happen with a manually created number pool (not using attribute kind numberpool).

We were also able to reproduce this running 2 instances of the same generator concurrently in Infrahub (allocating out of the same number pool)

fatih-acar:
Replaces #6888 (rebase of this branch on top of release-1.5/develop, with a bit of refactor).
Fixes #7254 

TODO:
- get rid of re-entrant locks? as per https://github.com/opsmill/infrahub/pull/6888#discussion_r2396118599

> The main goal is just to fix the following call stack:
Create mutation holds resource pool lock -> create new node internally using Node.new() -> triggers resource pool get_resource() -> get_resource locks again using the same resource pool id -> deadlock
We introduced get_resource locks again using the same resource pool id in https://github.com/opsmill/infrahub/pull/7025 to lock direct mutation calls to the resource pools...
A cleaner solution would be to move the lock from the get_resource call up to the different top-level mutation handlers.

- avoid locking on update mutations when we do not touch unique attributes and/or resource pools

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented resource pools from allocating duplicate values during concurrent mutations.

* **Performance**
  * Improved mutation performance via more granular multi-locking and reduced contention.

* **Behavioral**
  * Mutation previewing now avoids triggering pool allocations, reducing side effects when computing required locks.

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -314,7 +314,9 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
 
         return cls(**attrs)
 
-    async def handle_pool(self, db: InfrahubDatabase, attribute: BaseAttribute, errors: list) -> None:
+    async def handle_pool(
+        self, db: InfrahubDatabase, attribute: BaseAttribute, errors: list, allocate_resources: bool = True
+    ) -> None:
         """Evaluate if a resource has been requested from a pool and apply the resource
 
         This method only works on number pools, currently Integer is the only type that has the from_pool
@@ -325,7 +327,7 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
             attribute.from_pool = {"id": attribute.schema.parameters.number_pool_id}
             attribute.is_default = False
 
-        if not attribute.from_pool:
+        if not attribute.from_pool or not allocate_resources:
             return
 
         try:
@@ -485,7 +487,7 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
             elif relationship_peers := await relationship.get_peers(db=db):
                 fields[relationship_name] = [{"id": peer_id} for peer_id in relationship_peers]
 
-    async def _process_fields(self, fields: dict, db: InfrahubDatabase) -> None:
+    async def _process_fields(self, fields: dict, db: InfrahubDatabase, process_pools: bool = True) -> None:
         errors = []
 
         if "_source" in fields.keys():
@@ -539,7 +541,7 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
         # Generate Attribute and Relationship and assign them
         # -------------------------------------------
         errors.extend(await self._process_fields_relationships(fields=fields, db=db))
-        errors.extend(await self._process_fields_attributes(fields=fields, db=db))
+        errors.extend(await self._process_fields_attributes(fields=fields, db=db, process_pools=process_pools))
 
         if errors:
             raise ValidationError(errors)
@@ -576,7 +578,9 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
 
         return errors
 
-    async def _process_fields_attributes(self, fields: dict, db: InfrahubDatabase) -> list[ValidationError]:
+    async def _process_fields_attributes(
+        self, fields: dict, db: InfrahubDatabase, process_pools: bool
+    ) -> list[ValidationError]:
         errors: list[ValidationError] = []
 
         for attr_schema in self._schema.attributes:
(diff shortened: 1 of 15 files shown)


backend/infrahub/core/node/__init__.py
```diff
@@ -731,7 +736,7 @@ async def process_label(self, db: InfrahubDatabase | None = None) -> None:  # no
                 self.label.value = " ".join([word.title() for word in self.name.value.split("_")])
                 self.label.is_default = False
 
-    async def new(self, db: InfrahubDatabase, id: str | None = None, **kwargs: Any) -> Self:
+    async def new(self, db: InfrahubDatabase, id: str | None = None, process_pools: bool = True, **kwargs: Any) -> Self:
```
ogenstad, line 739: An issue here which has more to do with how the Node class works than the PR is that because of the `**kwargs` stuff we make `process_pools` a reserved keyword that can't be used for attribute or relationship names. There's a `RESERVED_ATTR_REL_NAMES` in the code I think we should add `process_pools` to that list. It will potentially cause issues if anyone is using an attribute with this name. I think it should be fairly safe to do in this instance..

We should refactor the Node class at some point and or get rid of that utterly awful `**kwargs: Any` and instead have a data fields where all the attribute and relationship data goes.

backend/infrahub/core/node/lock_utils.py
```diff
@@ -43,55 +43,78 @@ def _get_kinds_to_lock_on_object_mutation(kind: str, schema_branch: SchemaBranch
...
-    lock_kinds = _get_kinds_to_lock_on_object_mutation(kind, schema_branch)
-    lock_names = [build_object_lock_name(kind) for kind in lock_kinds]
-    return lock_names
+    lock_names: set[str] = set()
+
+    # Check if node is using resource manager allocation via attributes
+    for attr_name in node.get_schema().attribute_names:
+        attribute = getattr(node, attr_name, None)
+        if attribute is not None and getattr(attribute, "from_pool", None) and "id" in attribute.from_pool:
+            lock_names.add(f"resource_pool.{attribute.from_pool['id']}")
```
ogenstad: We need to take care that this name of the lock remains the same as we have within the resource manager. I'm not sure we need to do it now but as you say we should move them later. What we need to do at that time is to also implement the resource locks within the `InfrahubIPPrefixPoolGetResource` and `InfrahubIPAddressPoolGetResource` mutations if we move them out from the resources themselves.

ogenstad (approved):
LGTM, I think we should go with this for now. The main comment is around the added restricted field name.

Also can you add a news fragment to indicate the bug that was fixed? Also it would be great if you could create housekeeping issues for the optional tasks. Then we can take a step back to consider how we want these locks to behave and see if we can simplify it some more.

One thing we'd need to consider if we move the resource lock up in the stack is that we'll have to remember to use it whenever we access those resources from a new mutation. Still I think it's clear that the current implementation didn't work as desired.

backend/infrahub/core/node/__init__.py
fatih-acar: Thanks I wasn't aware of this...

backend/infrahub/core/node/lock_utils.py
fatih-acar: Added a constant to solve that.

fatih-acar:
@coderabbitai review

PR #7514, opened 2025-10-30
HFID/display label migration updates

ajtmccarty:
update the migrations to create (migration 042) and set (migration 043) `display_label` and `human_friendly_id` attributes so that they work with the new approach to migrate `main`, rebase branches, then migrate only the necessary attributes on each branch

it's a lot of code and a lot of tests. here's the outline of how each migration works

### Migration 042 - add display_label and human_friendly_id attributes to all Nodes

- run the existing migration on the default branch, adding the new attributes to every node that exists on those branches
- for each branch
  - rebase the branch (handled outside of the migration)
  - identify which nodes have been created on the branch
  - run the migration to add the new attributes on only those nodes created on the branch, the rebase will take care of the nodes created on `main` that exist on the branch

### Migration 043 - set the display_label and human_friendly_id attributes on all Nodes

- run the existing migration on the default branch, setting the display_label and HFIDs to the correct values
- for each branch
  - rebase the branch (handled outside of the migration)
  - for each node schema
    - if the display_label or HFID was updated at the schema-level, then update the appropriate attribute(s) for ALL nodes that exist on this branch
    - identify which nodes have updates to attributes or relationships or attributes on peers that would affect the HFID or display label value and run the updates on just those nodes for just the attribute(s) that need it

### DIFF
--- a/backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
+++ b/backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
+from itertools import chain
 from typing import TYPE_CHECKING, Any
 
 import ujson
@@ -119,17 +120,20 @@ class GetResultMapQuery(Query):
         return result_map
 
 
-class GetUpdatedPathDetailsBranchQuery(GetResultMapQuery):
-    name = "get_updated_path_details_branch"
-    type = QueryType.WRITE
+class GetPathDetailsBranchQuery(GetResultMapQuery):
+    name = "get_path_details_branch"
+    type = QueryType.READ
     insert_limit = False
 
-    def __init__(self, schema_kind: str, schema_paths: list[SchemaAttributePath], **kwargs: Any) -> None:
+    def __init__(
+        self, schema_kind: str, schema_paths: list[SchemaAttributePath], updates_only: bool = True, **kwargs: Any
+    ) -> None:
         super().__init__(**kwargs)
 
         if self.branch.name in [registry.default_branch, GLOBAL_BRANCH_NAME]:
             raise ValueError("This query can only be used on non-default branches")
         self.schema_kind = schema_kind
+        self.updates_only = updates_only
         self.attribute_names = []
         self.bidir_rel_attr_map: dict[str, list[str]] = defaultdict(list)
         self.outbound_rel_attr_map: dict[str, list[str]] = defaultdict(list)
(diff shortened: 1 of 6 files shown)


backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
```diff
@@ -61,7 +64,235 @@ def get_num_nodes(self) -> int:
...
+            node_uuid = result.get_as_type(label="n_uuid", return_type=str)
+
+            # for each node, build a map of the schema path key to value so that they
+            # can be ordered correctly for the input `schema_paths`
+            schema_path_value_map: dict[str | tuple[str, RelationshipDirection, str], Any] = {}
+            attr_values_tuples: list[tuple[str, Any]] = result.get_as_type(label="attr_vals_list", return_type=list)
+            for attr_value_tuple in attr_values_tuples:
+                attr_name = attr_value_tuple[0]
+                attr_value = attr_value_tuple[1]
+                schema_path_value_map[attr_name] = attr_value
```
ogenstad, line 97: Would it work to instead do:

```python
            schema_path_value_map: dict[str | tuple[str, RelationshipDirection, str], Any] = dict(
                result.get_as_type(label="attr_vals_list", return_type=list)
            )
```

backend/tests/unit/core/migrations/graph/test_042_043.py
```diff
@@ -1,78 +1,680 @@
...
+            human_friendly_id=[
+                str(v)
+                for v in (secondary_thing.name.value, secondary_thing.color.value, secondary_thing.agnostic_smell.value)
+            ],
+            status=NodeStatus.ACTIVE,
+        )
+
+    @pytest.fixture
+    async def secondary_thing_three_agnostic_value_update(self) -> str:
+        return "more okay"
```
ogenstad, line 164: Not really an objection but is there some value of having this as a fixture instead of it just being a constant?

ogenstad (approved):

backend/infrahub/core/migrations/graph/m043_backfill_hfid_display_label_in_db.py
ajtmccarty: I did forget that you can initialize a `dict` with a list of `tuple`s, but I think this is easier to read and understand with the variable names `attr_name` and `attr_value`

backend/tests/unit/core/migrations/graph/test_042_043.py
ajtmccarty: I made it a fixture because it's only used in other fixtures
a constant would be fine too, but I was in fixture mode

PR #7516, opened 2025-10-30
Set transform timeout to 60 seconds

ogenstad:
Used for the repository that gets imported by the end to end tests

## Summary by CodeRabbit

* **Chores**
  * Updated data transform configurations to include timeout settings for improved reliability and stability.


fatih-acar:
What about https://github.com/opsmill/infrahub/pull/7299?

ogenstad:
> What about #7299?

That one should be merged too, however we want to do it when we bring the `stable` branch of the sdk into `stable` of the Infrahub repo. So it would be around the time of the release.
https://github.com/opsmill/infrahub/pull/7299#pullrequestreview-3313181292

ajtmccarty (approved):

PR #7518, opened 2025-10-30
IFC-2023: Export REST API schema to `./schema/openapi.json` and repository config schema to output path

solababs:
This PR adds ability to export REST API schema to `./schema/openapi.json` and also do a git diff comparison to know when the schema has changes that should be updated. This can be run using `invoke schema.validate-jsonschema`

Also adds the repository config to the specified output path and that can also be run using `invoke schema.validate-repositoryconfig`

CI for github is also updated to generate the validations

## Summary by CodeRabbit

* **New Features**
  * CLI: added export-json-schema and export-node-schema to export OpenAPI/JSON schemas and repository config.
  * Build tasks: added generation and validation flows for JSON Schema, repository-config and node schema; updated schema output locations.

* **Documentation**
  * CLI docs updated with usage and options for the new export commands.

* **Chores**
  * CI: added a job to validate JSON Schema alongside existing schema checks.
  * Pre-commit: excluded the generated schema file from EOF fixer.


backend/infrahub/cli/dev.py
```diff
@@ -57,6 +60,20 @@ async def export_graphql_schema(
...
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("schema.json"),  # noqa: B008
+) -> None:
+    """Export the Core GraphQL schema to a file."""
+    config.load_and_exit(config_file_name=config_file)
+    async with AsyncClient(app=server_app, base_url=config.SETTINGS.main.internal_address) as client:
+        response = await client.get("/api/openapi.json")
+        content = json.dumps(response.json(), indent=4)
+        out.write_text(content)
```
ogenstad: I think we can simplify this a bit and just have:

```python
>>> from infrahub.server import app as server_app
>>> server_app.openapi()
```

backend/infrahub/cli/dev.py
```diff
@@ -57,6 +60,20 @@ async def export_graphql_schema(
     out.write_text(sorted_schema_str)
 
 
+@app.command(name="export-json-schema")
+async def export_json_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("schema.json"),  # noqa: B008
```
ogenstad: I'd have the default name as `openapi.json` instead of `schema.json`

schema/openapi.json
```diff
@@ -0,0 +1,6193 @@
+{
+    "openapi": "3.1.0",
+    "info": {
+        "title": "Infrahub",
+        "version": "1.5.0b1"
```
ogenstad: I hadn't considered the version here. It feels like something that will pop up as merge conflicts when we merge between `stable` and `develop`. Not really a huge issue but something to consider. I'm not sure we strictly need the version here in order to do bindings for the frontend. Might be that we should remove it before saving the file and just have a blank string, or "latest" as the version. Do you have any thoughts around this?

tasks/schema.py
```diff
@@ -1,66 +1,51 @@
...
-def generate_jsonschema(context: Context) -> None:  # noqa: ARG001
-    """Generate JSON schemas into ./generated"""
+def generate_jsonschema(context: Context) -> None:
+    """Generate JSON schemas into ./schema"""
+    with context.cd(ESCAPED_REPO_PATH):
+        context.run(f"poetry run infrahub dev export-json-schema --out {OPENAPI_PATH}")
+
 
-    generate_sdk_repository_config()
-    generate_infrahub_node_schema()
```
ogenstad, line 26: We still want to keep these tasks in here even if we don't want to validate that the schema files under `./generated` are up to date within the Git repository. It's fine if we also have them as a command under `infrahub dev` but I think they should remain within this job even if we call the infrahub dev command. Also it would be nice with some output like we had before with:

```bash
❯ inv schema.generate-jsonschema
Wrote to /Users/patrick/Code/opsmill/infrahub/generated/python-sdk/repository-config/develop.json
Wrote to /Users/patrick/Code/opsmill/infrahub/generated/infrahub/schema/develop.json
```

So it's clear for the person who'd generate these specific files where they are. And because we specifically want them to go into this folder it might be better to keep some logic within these invoke tasks as we want to hardcode the output path as per the above (within the context of the current root folder)

tasks/schema.py
```diff
@@ -1,66 +1,51 @@
...
+        context.run(f"poetry run infrahub dev export-graphql-schema --out {SCHEMA_PATH}")
 
 
 @task
-def generate_jsonschema(context: Context) -> None:  # noqa: ARG001
-    """Generate JSON schemas into ./generated"""
+def generate_jsonschema(context: Context) -> None:
+    """Generate JSON schemas into ./schema"""
+    with context.cd(ESCAPED_REPO_PATH):
+        context.run(f"poetry run infrahub dev export-json-schema --out {OPENAPI_PATH}")
```
ogenstad, line 33: I think it would be nice with a print output to indicate where the file was stored.

backend/infrahub/cli/dev.py
```diff
@@ -57,6 +60,31 @@ async def export_graphql_schema(
...
 
 
+@app.command(name="export-json-schema")
+async def export_json_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("openapi.json"),  # noqa: B008
+) -> None:
+    """Export the JSON schema to a file."""
+    config.load_and_exit(config_file_name=config_file)
```
ogenstad: Since we don't really use the server now I think we can skip the config all together and remove that as a parameter and skip loading it?

backend/infrahub/cli/dev.py
```diff
@@ -57,6 +60,31 @@ async def export_graphql_schema(
     out.write_text(sorted_schema_str)
 
 
+@app.command(name="export-json-schema")
+async def export_json_schema(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("openapi.json"),  # noqa: B008
+) -> None:
+    """Export the JSON schema to a file."""
```
ogenstad: As per CodeRabbits comment I think we can be a bit more explicit around which JSON schema we're exporting as there are a few of them. I.e. we should indicate that it's the OpenAPI one.

ogenstad (approved):
LGTM but I added a few minor comments.

schema/openapi.json
solababs: @ogenstad I think "latest" works

backend/infrahub/cli/dev.py
```diff
@@ -57,6 +60,31 @@ async def export_graphql_schema(
...
+async def export_repository_config(
+    ctx: typer.Context,  # noqa: ARG001
+    config_file: str = typer.Option("infrahub.toml", envvar="INFRAHUB_CONFIG"),
+    out: Path = typer.Option("develop.json"),  # noqa: B008
+) -> None:
+    """Export the repository configuration to a file."""
+    config.load_and_exit(config_file_name=config_file)
+    schema = json.dumps(InfrahubRepositoryConfig.model_json_schema(), indent=4)
+    out.parent.mkdir(parents=True, exist_ok=True)
+    out.write_text(schema)
```
ogenstad: Hey I think that we might want a small change here in the sense that within the `infrahub` CLI app I think we might want the Node schema export, i.e the `generate_infrahub_node_schema` job which I think is now deleted. However for `InfrahubRepositoryConfig` we don't want it as an option of the infrahub CLI command, I think fairly soon we'll actually want to move the generation of the SDK json schema into the SDK repository so it doesn't live here at all.

ogenstad (approved):

PR #7536, opened 2025-11-01
no duplicated edges during merge involving migrated-kind nodes

ajtmccarty:
a little background, schema updates that include a change to name, namespace, or inheritance are handled in a special way at the database level. we create duplicate vertices of all Nodes using the updated schema, update the label(s) on the new version, and soft-delete the old version. the end result is two nodes with the same UUID linked to all of the same attribute and relationship vertices. and some of our cypher queries have not been updated to handle this case.

### Diff merge bug

there was a bug in the diff merge logic that affected relationship updates that included peers that had their kind/inheritance updated on the default branch by creating duplicated edges to the older, deleted versions of migrated nodes. this PR fixes that bug and adds a test for it.

this PR also includes a migration and test for removing these duplicated/illegal edges on the default branch.

### Diff calculation bug

while working on this, I also found that the diff calculation logic included a related bug that would cause edges linked to the older versions of migrated kind nodes to be included in the diff when they should be excluded. this PR includes a fix and test for that bug as well.

### Diff delete bug for huge diffs

and one final thing, while testing this against a very large diff, I found that the cypher query can run out of memory and crash, so there is one final update to reduce the memory usage of the query to delete a diff

## Release Notes

* **Bug Fixes**
  * Fixed branch merge logic that could create duplicate edges when node kinds or inheritance were updated, preventing relationship constraint violations.
  * Fixed diff calculation to properly handle nodes with migrated kinds or inheritance when determining branch-specific changes.


backend/infrahub/core/diff/query/delete_query.py
```diff
@@ -20,10 +20,14 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
-        DETACH DELETE d_root
+MATCH (d_root:DiffRoot)
+%(diff_filter)s
+OPTIONAL MATCH (d_root)-[*]->(diff_thing)
+WITH DISTINCT d_root, diff_thing
+ORDER BY elementId(diff_thing)
+CALL (diff_thing) {
+    DETACH DELETE diff_thing
+} IN TRANSACTIONS
+DETACH DELETE d_root
```
ajtmccarty, line 31: reduce memory usage of this query so that it can delete a very large diff (over 1 million database vertices)

backend/infrahub/core/diff/query/merge.py
ajtmccarty, line 1: fixes to make sure we use the latest Node with a given UUID in case it had its kind/inheritance updated in the past, resulting in multiple Nodes with the same UUID

backend/infrahub/core/query/diff.py
```diff
@@ -641,8 +641,28 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
         )
         // skip paths where nodes/attrs/rels are updated after $from_time, those are handled in other queries
         AND (
-            r_root.from <= $from_time AND (r_root.to IS NULL OR r_root.branch <> diff_rel.branch OR r_root.to <= $from_time)
-            AND r_node.from <= $from_time AND (r_node.to IS NULL OR r_node.branch <> diff_rel.branch OR r_node.to <= $from_time)
```
ajtmccarty, line 645: there are three diff calculation queries that work together: one for node-level adds/deletes, one for attribute/relationship-level changes, and one for property-level changes. each query excludes changes covered by any higher-level queries. for example, the attribute/relationship query excludes changes covered by the node query because we don't need to track which individual attributes and relationships were deleted for a given node if the whole node was deleted

this filter in the property-level query worked _almost_ all of the time to exclude properties handled at the node or attribute/relationship levels, but it did not work correctly if a peer node had its kind/inheritance updated. in this case, the query would erroneously include some paths to the deleted version of the migrated node. making this chunk of the query (and the matching one below) more specific resolved the issue

backend/tests/unit/core/diff/test_diff_and_merge.py
```diff
@@ -1207,3 +1207,111 @@ async def test_diff_and_merge_with_migrated_node_kind_and_migrated_inheritance(
         # get undeleted node
         undeleted_car = await NodeManager.get_one(db=db, branch=default_branch, id=e_car_2.id)
         assert undeleted_car.get_kind() == "TestElectricCar"
+
+    async def test_diff_and_merge_with_migrated_node_kind_peer(
```
ajtmccarty, line 1211: test for merging a branch that includes relationships to a migrated-kind node

backend/tests/unit/core/diff/test_diff_calculator.py
```diff
@@ -4296,3 +4296,159 @@ async def test_migrated_kind_with_property_level_changes(
     assert is_protected_prop.action is DiffAction.UPDATED
     assert is_protected_prop.new_value is True
     assert is_protected_prop.previous_value is False
+
+
+async def test_migrated_kind_on_main_then_relationship_update_on_branch(
```
ajtmccarty, line 4301: test for calculating the diff correctly when it includes relationships to a migrated-kind node

backend/tests/unit/core/migrations/graph/test_041/data_export/edges.csv
ajtmccarty, line 1: dataset created by the `infrahub db selected-export` command using the old code that created the illegal edges after a merge including relationships to a migrated-kind peer

ogenstad (approved):

PR #7544, opened 2025-11-03
IFC-1977: Add git config for user name and email, set config on repo

solababs:
This PR creates new git configurations that are then set on the repository. The created configurations are
- [x] `user_name` - The user name of the git user
- [x] `user_email` - The email of the git user

## Summary by CodeRabbit

* **New Features**
  * New Git configuration options: user name, user email, global config file path, and dev credential helper
  * Option to enable explicit non-fast-forward merge commits
  * Git global configuration is now initialized automatically during worker startup

* **Documentation**
  * Configuration reference updated with the new Git and development settings

* **Tests**
  * Added integration and unit tests validating Git user config and explicit-merge behavior

* **Changelog**
  * Entry documenting explicit merge commits and new Git settings


backend/infrahub/git/base.py
```diff
@@ -415,6 +417,13 @@ async def create_locally(
 
         return True
 
+    @staticmethod
+    def create_repo_config(repo: Repo) -> None:
+        if config.SETTINGS.git.user_name and config.SETTINGS.git.user_email:
+            with repo.config_writer() as git_config:
+                git_config.set_value("user", "name", config.SETTINGS.git.user_name)
+                git_config.set_value("user", "email", config.SETTINGS.git.user_email)
```
ajtmccarty: is it possible to have a username without an email or vice versa?

ogenstad: I think it would be possible if a someone just configures an override for one of them.

I'm not sure where this should be implemented. I was considering to add it as a global commit option when someone starts up the prefect worker. I.e. that they would just run "git config --global" once at that point during startup. Then we don't need to deal with it each time we do something locally in the repo. It would also allow users to change it globally with a config setting.

The annoying thing with this would be that if you ran the infrahub prefect worker locally on your laptop and had this option enabled it would change the global git config...

What do you guys think of this?

solababs: That works, I can move the implementation if we are agreed on that

ajtmccarty: sounds okay to me

backend/infrahub/config.py
```diff
@@ -453,6 +454,8 @@ class GitSettings(BaseSettings):
             "Note: other branches created with sync with git will be imported also"
         ),
     )
+    user_name: str | None = Field(default=None, description="User name of the git user")
+    user_email: EmailStr | None = Field(default=None, description="Email of the git user")
```
ogenstad: I think for the description we should clarify where it will be used as this is something that shows up in the documentation. I.e. we should state that the username and email provided here will be used when Infrahub commits code to a repository.

backend/infrahub/workers/infrahub_async.py
```diff
@@ -40,6 +42,9 @@
 WORKER_QUERY_SECONDS = "2"
 WORKER_DEFAULT_RESULT_STORAGE_BLOCK = f"redisstoragecontainer/{TASK_RESULT_STORAGE_NAME}"
 DEFAULT_TASK_LOGGERS = ["infrahub.tasks"]
+DEFAULT_GIT_GLOBAL_CONFIG_DIRECTORY = "/opt/infrahub"
+DEFAULT_GIT_GLOBAL_CONFIG_FILE = f"{DEFAULT_GIT_GLOBAL_CONFIG_DIRECTORY}/git-config"
```
ogenstad: I'd suggest that we also move this to the config object instead of having it as a hardcoded constant.

backend/infrahub/workers/infrahub_async.py
```diff
@@ -204,3 +210,35 @@ async def _init_services(self, client: InfrahubClient) -> None:
         )
 
         self.service = service
+
+    async def set_git_global_config(self) -> None:
+        if INFRAHUB_PRODUCTION and not os.getenv("GIT_GLOBAL_CONFIG"):
+            os.makedirs(DEFAULT_GIT_GLOBAL_CONFIG_DIRECTORY, exist_ok=True)  # noqa: PTH103
```
ogenstad: Instead of ignoring the error here we should use Pathlib as suggested by the rule: https://docs.astral.sh/ruff/rules/os-makedirs/

backend/infrahub/workers/infrahub_async.py
```diff
@@ -204,3 +210,35 @@ async def _init_services(self, client: InfrahubClient) -> None:
         )
 
         self.service = service
+
+    async def set_git_global_config(self) -> None:
+        if INFRAHUB_PRODUCTION and not os.getenv("GIT_GLOBAL_CONFIG"):
```
ogenstad: As we discussed in Slack I think we should remove the `INFRAHUB_PRODUCTION` variable from here.

backend/tests/integration/workers/test_infrahubasync.py
```diff
@@ -107,3 +112,64 @@ async def test_broken_flow(
...
+        assert os.getenv("GIT_CONFIG_GLOBAL") and os.getenv("GIT_CONFIG_GLOBAL") == tmp_git_config
+
+        initial_user_name = config.SETTINGS.git.user_name
+        initial_user_email = config.SETTINGS.git.user_email
+
+        user_name = "Test User"
+        user_email = "test@email.com"
+
+        await set_git_config("user.name", user_name)
+        await set_git_config("user.email", user_email)
```
ogenstad: As CodeRabbit indicates I'm wondering if this is correct, i.e. the idea would be that the test `test_worker_has_set_git_user_config()` would validate that the worker configures these values but here we have already added them in the fixture. Is this a leftover from a previous iteration?

backend/infrahub/workers/infrahub_async.py
```diff
@@ -204,3 +210,35 @@ async def _init_services(self, client: InfrahubClient) -> None:
...
+            proc_email = await asyncio.create_subprocess_exec(
+                "git",
+                "config",
+                "--global",
+                "user.email",
+                config.SETTINGS.git.user_email,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.PIPE,
+            )
+            await proc_email.wait()
```
ogenstad: For these two commands we assume that everything will work fine. We should consider errors as well, what would it look like if any of these calls failed? Would the app crash or would the error be silently ignored?

backend/infrahub/config.py
```diff
@@ -453,6 +454,26 @@ class GitSettings(BaseSettings):
...
+        description=(
+            "User name of the git user. This will be used as the user name when Infrahub commits code to a repository"
+        ),
+    )
+    user_email: EmailStr | None = Field(
+        default=None,
+        description=(
+            "Email of the git user. This will be used as the user email when Infrahub commits code to a repository"
+        ),
+    )
```
ogenstad, line 472: I wonder if we should set defaults here to match what we have in:
https://github.com/opsmill/infrahub/blob/stable/development/Dockerfile#L81-L82

That way we could validate the settings on each startup of the agent. Could also remove the need to set it within the Dockerfile.

backend/infrahub/workers/infrahub_async.py
```diff
@@ -204,3 +207,48 @@ async def _init_services(self, client: InfrahubClient) -> None:
         )
 
         self.service = service
+
+    async def set_git_global_config(self) -> None:
```
ogenstad, line 211: When we are creating a new config that will be used we'd also need to include the other settings, i.e. this config:
https://github.com/opsmill/infrahub/blob/stable/development/Dockerfile#L83-L85

backend/infrahub/config.py
```diff
@@ -453,6 +454,26 @@ class GitSettings(BaseSettings):
...
+    global_config_file: str | None = Field(
+        default="/opt/infrahub/.gitconfig",
```
ogenstad, line 474: As we set a default I don't thing it the setting should be an optional string. Using an environment variable I'm not sure we'd be able to override this and set it ot a `null` value.

backend/infrahub/workers/infrahub_async.py
```diff
@@ -252,3 +252,52 @@ async def set_git_global_config(self) -> None:
...
+            self._logger.error("Failed to set git credential.usehttppath: %s", error_msg)
+        else:
+            self._logger.info("Git credential.usehttppath set")
+
+        proc = await asyncio.create_subprocess_exec(
+            "git",
+            "config",
+            "--global",
+            "credential.helper",
+            "/usr/local/bin/infrahub-git-credential",
```
ogenstad: I think this would actually be good to be able to control for local development, but I'm not sure if it's useful elsewhere. Could you add a git_credential_helper setting under the `DevelopmentSettings` config class and have the value default to `/usr/local/bin/infrahub-git-credential`?

ogenstad (approved):

development/Dockerfile
```diff
@@ -12,6 +12,8 @@ ENV PATH="${PATH}:/root/.local/bin" \
     INFRAHUB_FRONTEND_DIRECTORY="/opt/infrahub/frontend/app" \
     INFRAHUB_DOCS_DIRECTORY="/opt/infrahub/docs"
 
+RUN groupadd -r -g 1000 infrahub && useradd --no-log-init -u 1000 --home-dir /opt/infrahub -r -g infrahub infrahub
+
```
fatih-acar: I don't think the changes in `Dockerfile` are required.
Doing this is a breaking change for existing Infrahub instances regarding permissions (since the directories and files were created using `root`), I'd be careful with that... if you really need those changes.

ogenstad: Ok, let's skip the changes to the Dockerfile.

ogenstad (approved):

PR #7545, opened 2025-11-03
IFC-1979: Conditionally merge with no fast forward with `allow_explicit_merge_commit` git config

solababs:
This PR adds a merge with no fast forward based on the setting from git.
Git setting is boolean - `allow_explicit_merge_commit`

### DIFF
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -456,6 +456,9 @@ class GitSettings(BaseSettings):
     )
     user_name: str | None = Field(default=None, description="User name of the git user")
     user_email: EmailStr | None = Field(default=None, description="Email of the git user")
+    allow_explicit_merge_commit: bool = Field(
+        default=False, description="Whether to allow explicit merge commits when infrahub merges branches"
+    )
 
     @model_validator(mode="after")
     def validate_sync_branch_names(self) -> Self:
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -326,6 +326,10 @@ class DevelopmentSettings(BaseSettings):
         default=False,
         description="Allow enterprise configuration in development mode, this will not enable the features just allow the configuration.",
     )
+    git_credential_helper: str = Field(
+        default="/usr/local/bin/infrahub-git-credential",
+        description="Location of git credential helper",
+    )
 
 
 class BrokerSettings(BaseSettings):
@@ -470,7 +474,7 @@ class GitSettings(BaseSettings):
         default="/opt/infrahub/.gitconfig",
         description=(
             "The location of the git config file. "
-            "This will be set as the system `GIT_GLOBAL_CONFIG` environment variable "
+            "This will be set as the system `GIT_CONFIG_GLOBAL` environment variable "
             "if the environment variable is not initially set"
         ),
     )
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -474,7 +474,7 @@ class GitSettings(BaseSettings):
             "if the environment variable is not initially set"
         ),
     )
-    allow_explicit_merge_commit: bool = Field(
+    use_explicit_merge_commit: bool = Field(
         default=False, description="Whether to allow explicit merge commits when infrahub merges branches"
     )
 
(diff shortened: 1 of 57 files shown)


wvandeun:
Wouldn't it be better if we rename this configuration setting to `use_explicit_merge_commit`?

backend/infrahub/git/repository.py
```diff
@@ -170,7 +171,12 @@ async def merge(self, source_branch: str, dest_branch: str, push_remote: bool =
         commit = self.get_commit_value(branch_name=source_branch, remote=False)
 
         try:
-            repo.git.merge(commit)
+            if config.SETTINGS.git.allow_explicit_merge_commit:
+                if not config.SETTINGS.git.user_name:
+                    raise ValueError("Cannot allow explicit merge commit if git user_name setting is not set.")
```
wvandeun: What would be the user behavior when we raise `ValueError` here?
How would the user become aware that there was an issue and what action would we expect the user to take?

ogenstad: I'd suggest that we ignore this setting and if not configured we'll just use the default username that's set in the Docker image.

backend/infrahub/config.py
```diff
@@ -474,6 +474,9 @@ class GitSettings(BaseSettings):
             "if the environment variable is not initially set"
         ),
     )
+    allow_explicit_merge_commit: bool = Field(
+        default=False, description="Whether to allow explicit merge commits when infrahub merges branches"
+    )
```
ogenstad, line 483: I'd suggest another name than `allow_` here. Could be `add_explicit_merge_commit`. I.e. we're not really allowing something but configuring Infrahub to add this extra commit.

backend/infrahub/git/repository.py
```diff
@@ -170,7 +172,11 @@ async def merge(self, source_branch: str, dest_branch: str, push_remote: bool =
         commit = self.get_commit_value(branch_name=source_branch, remote=False)
 
         try:
-            repo.git.merge(commit)
+            if config.SETTINGS.git.allow_explicit_merge_commit:
+                user_name = config.SETTINGS.git.user_name or get_git_user_config(repo)[0]
+                repo.git.merge(commit, "--no-ff", m=f"Merged by Infrahub by {user_name}")
```
ogenstad: For now I think we should just go with `repo.git.merge(commit, "--no-ff", m=f"Merged by Infrahub by Infrahub")`

If we'll have a username here it would be the user within Infrahub that merged the branch, or proposed change and not the user set within the Infrahub configuration. The goal of this epic is to add this none ff commit and we haven't gotten a clear answer with regards to what the commit message should be at this point. So let's just keep it simple for now.

backend/infrahub/config.py
solababs: I actually changed it to `use_` as stated here https://github.com/opsmill/infrahub/pull/7545#issuecomment-3479107823 I think i lost it while trying to fix conflicts, ill add that back

ogenstad: Ok, `use_` works for me!

backend/infrahub/git/repository.py
```diff
@@ -170,7 +171,10 @@ async def merge(self, source_branch: str, dest_branch: str, push_remote: bool =
         commit = self.get_commit_value(branch_name=source_branch, remote=False)
 
         try:
-            repo.git.merge(commit)
+            if config.SETTINGS.git.use_explicit_merge_commit:
+                repo.git.merge(commit, "--no-ff", m="Merged by Infrahub by Infrahub")
```
ogenstad: Sorry I don't know if I duplicated this in my previous comment but the extra "by Infrahub" is a bit redundant here. We should start with just "Merged by Infrahub"

backend/infrahub/git/utils.py
```diff
@@ -177,3 +178,15 @@ def branch_name_in_import_sync_branches(branch_short_name: str) -> bool:
...
+def get_git_user_config(repo: Repo) -> tuple[str, str]:
+    """
+    Checks for the git configuration of the user on the global level
+    Returns the git (username, email,)
+    """
+    with repo.config_reader(config_level="global") as git_config:
+        return (
+            str(git_config.get_value("user", "name")),
+            str(git_config.get_value("user", "email")),
+        )
```
ogenstad: I don't think we'll use this function based on the refactor?

ogenstad (approved):

PR #7551, opened 2025-11-03
Add return type annotations to tests

ogenstad:
Auto fixes for ruff violations returning None or Never

```toml
"backend/tests/**.py" = [
    "ANN201",   # Missing return type annotation for public function
]
```


gmazoyer (approved):
Of course I read every single lines touched by this PR

PR #7566, opened 2025-11-05
test and fixes for diff query in parallel mode

ajtmccarty:
IFC-2017

add tests and fix for diff calculation queries using parallel runtime
- needed to add a NULL row to the query before sorting and filtering for pagination, b/c if all results were filtered out after pagination, then nothing would be returned and the `DiffCalculator` would consider the query complete. but there could actually be more rows to read on later pages

also found an issue in the `NodeListGetAttributeQuery` that affected nodes that had their kind/inheritance migrated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of pagination limits to ensure consistent minimum thresholds, preventing edge cases with small result sets.

* **Performance**
  * Optimized database query patterns for more efficient data retrieval and pagination.
  * Enhanced data consistency calculations to ensure accurate result availability indicators across filtered datasets.

### DIFF
--- a/backend/infrahub/core/diff/calculator.py
+++ b/backend/infrahub/core/diff/calculator.py
@@ -141,8 +141,8 @@ class DiffCalculator:
             to_time=to_time,
             previous_node_field_specifiers=previous_node_specifiers,
         )
-        node_limit = int(config.SETTINGS.database.query_size_limit / 10)
-        fields_limit = int(config.SETTINGS.database.query_size_limit / 3)
+        node_limit = max(int(config.SETTINGS.database.query_size_limit / 10), 1)
+        fields_limit = max(int(config.SETTINGS.database.query_size_limit / 3), 1)
         properties_limit = config.SETTINGS.database.query_size_limit
 
         calculation_request = DiffCalculationRequest(
--- a/backend/infrahub/core/query/diff.py
+++ b/backend/infrahub/core/query/diff.py
@@ -301,13 +301,14 @@ WITH p, q, diff_rel, CASE
     ELSE $from_time
 END AS row_from_time
 ORDER BY %(id_func)s(p) DESC
-SKIP $offset
-LIMIT $limit
+SKIP toInteger($offset)
+LIMIT toInteger($limit)
 // -------------------------------------
 // Add flag to indicate if there is more data after this
 // -------------------------------------
 WITH collect([p, q, diff_rel, row_from_time]) AS limited_results
-WITH limited_results, size(limited_results) = $limit AS has_more_data
+WITH limited_results + [[NULL, NULL, NULL, NULL]] AS limited_results
+WITH limited_results, size(limited_results) = ($limit + 1) AS has_more_data
 UNWIND limited_results AS one_result
 WITH one_result[0] AS p, one_result[1] AS q, one_result[2] AS diff_rel, one_result[3] AS row_from_time, has_more_data
 // -------------------------------------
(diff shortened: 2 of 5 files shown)


backend/infrahub/core/diff/calculator.py
```diff
@@ -141,8 +141,8 @@ async def calculate_diff(
             to_time=to_time,
             previous_node_field_specifiers=previous_node_specifiers,
         )
-        node_limit = int(config.SETTINGS.database.query_size_limit / 10)
-        fields_limit = int(config.SETTINGS.database.query_size_limit / 3)
+        node_limit = max(int(config.SETTINGS.database.query_size_limit / 10), 1)
+        fields_limit = max(int(config.SETTINGS.database.query_size_limit / 3), 1)
```
ajtmccarty, line 145: if the query size limit is set low enough (< 10) then these values could be 0, which resulted in an infinite loop that incremented by 0

backend/infrahub/core/query/diff.py
```diff
@@ -301,13 +301,14 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
     ELSE $from_time
 END AS row_from_time
 ORDER BY %(id_func)s(p) DESC
-SKIP $offset
-LIMIT $limit
+SKIP toInteger($offset)
+LIMIT toInteger($limit)
```
ajtmccarty, line 305: not actually necessary, but does not hurt and makes testing easier b/c the neo4j browser turns all numbers set as parameters into `float`s and these need to be integers

backend/infrahub/core/query/node.py
```diff
@@ -628,44 +628,45 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
 
         query = """
```
ajtmccarty, line 629: this query is old and was written in such a manner that it could return multiple results for the same attributes when operating on a node with a migrated kind. This could cause the wrong value to be returned for a given attribute. I'm not exactly sure why no unit tests surfaced this issue before. It could be that my recent changes and testing around diff queries and the neo4j parallel runtime revealed it.
I've made some changes below to make the query easier to read and made some functional changes so that it only returns one result for each attribute

backend/infrahub/core/query/node.py
```diff
@@ -628,44 +628,45 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
-            ORDER BY r.branch_level DESC, r.from DESC
+            RETURN r.status = "active" AS is_active
+            ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
             LIMIT 1
         }
-        WITH n1 as n, r1, a1 as a
-        WHERE r1.status = "active"
-        WITH n, r1, a
-        MATCH (a)-[r:HAS_VALUE]-(av:AttributeValue)
-        WHERE %(branch_filter)s
```
ajtmccarty, line 641: this could cause there to be multiple results for each Attribute, but a given Attribute can only have a single value on a branch at a point in time

backend/infrahub/core/query/node.py
```diff
@@ -628,44 +628,45 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
-            at=self.at, branch_agnostic=self.branch_agnostic, variable_name="rel_isv"
-        )
-        rel_isp_branch_filter, _ = self.branch.get_query_filter_path(
-            at=self.at, branch_agnostic=self.branch_agnostic, variable_name="rel_isp"
-        )
         query = """
-        MATCH (a)-[rel_isv:IS_VISIBLE]-(isv:Boolean)
-        MATCH (a)-[rel_isp:IS_PROTECTED]-(isp:Boolean)
-        WHERE (%(rel_isv_branch_filter)s) AND (%(rel_isp_branch_filter)s)
-        """ % {"rel_isv_branch_filter": rel_isv_branch_filter, "rel_isp_branch_filter": rel_isp_branch_filter}
```
ajtmccarty, line 668: this approach would return multiple results if the IS_VISIBLE or IS_PROTECTED properties were updated

backend/tests/unit/core/test_node_query.py
ajtmccarty, line 1: my changes to the `NodeListGetAttributeQuery` mean that only the correct and expected results are returned instead of extra results that we then filter out in memory

gmazoyer (approved):

Issue #6364, opened 2025-04-25
bug: you can load a schema with an optional and unique attribute


wvandeun:
### Component

API Server / GraphQL

### Infrahub version

1.2.6

### Current Behavior

You can load a schema with an attribute defined on a node, that is both optional and unique.
When this is the only unique attribute on the node, it will also be taken as the HFID implicitly.

Once the schema is loaded, you can then create multiple objects with a `null` value for that optional/unique attribute.

### Expected Behavior

You cannot load a schema that has an optional and unique attribute. For backwards compatibility we could consider disabling this with the `INFRAHUB_STRICT_MODE` environment variable if required.

### Steps to Reproduce

- start instance of Infrahub
- load this schema
```yaml
---
version: "1.0"
nodes:
  - name: Object
    namespace: Testing
    attributes:
      - name: name
        kind: Text
        optional: true
        unique: true
      - name: other
        kind: Text
        optional: true
```
- check the schema, notice that the HFID is set to the value of the name attribute
- create 2 nodes with a null value for the name attribute
```python
o1 = client.create("TestingObject", name="", other="foo")
o1.save()
o2 = client.create("TestingObject", name="", other="bar")
o2.save()
```

saltas888:
Recording the decision here, since it was discussed outside the issue.

The plan is to reject the schema shape rather than change the NULL semantics:

- an attribute that is both optional and unique raises a validation error at schema load when `INFRAHUB_SCHEMA_STRICT_MODE` is enabled, which is the default
- `INFRAHUB_SCHEMA_STRICT_MODE=false` keeps today's behavior for anyone who needs a temporary bridge
- NULL stays a regular value for uniqueness checks, unchanged since #5383, so the alternative that was discussed, letting optional unique attributes hold multiple NULLs, isn't part of the fix
- an upgrade check lists any existing schemas that hit this, so it surfaces at `infrahub db migrate` rather than at runtime

#6191 is the same root cause surfacing at branch merge time, closed in favour of this one.

PR #7569, opened 2025-11-05
IFC-1480: Validate optional and unique attribute parameters

solababs:
Fixes https://github.com/opsmill/infrahub/issues/6364

Validation to check the `optional` and `unique` attribute parameters that they are not both `True` for generics and nodes attributes
This check only occurs if `schema_strict_mode` configuration setting is `True`

## Summary by CodeRabbit

* **New Features**
  * Added validation to prevent an attribute being configured as both optional and unique; enforcement respects the system's strict-schema validation setting.
* **Tests**
  * Added unit tests covering the new validation and behavior when strict-schema validation is disabled.
* **Documentation**
  * Changelog entry documenting the new attribute-parameter validation.


saltas888:
I'm closing this as its too old and the conflicts resolution might be more work than rework on it

PR #7580, opened 2025-11-06
docs(backend): clarify instructions in backend development

iddocohen:
This PR updates the documentation with corrected development commands, clarifies the local build process, and standardizes test execution.

### Changes

- Added clear instruction to build services locally using poetry run invoke dev.build
- Fixed bug in service management commands to use poetry run invoke ... 
- Replaced direct pytest execution with poetry run invoke backend for running tests

## Summary by CodeRabbit

* **Documentation**
  * Updated backend development docs with steps to build required services locally and an install step before running tests.
  * Standardized developer workflow so service and status commands run consistently under the project environment.
  * Revised test instructions to run tests via the project environment.
  * Minor formatting and clarity improvements.

### DIFF
--- a/docs/docs/development/backend.mdx
+++ b/docs/docs/development/backend.mdx
@@ -65,10 +65,16 @@ It is also possible to disable the automatic startup of services and to rely on 
 export INFRAHUB_USE_TEST_CONTAINERS=false
 ```
 
+To build the required services locally, run the following command:
+
+```shell
+poetry run invoke dev.build
+```
+
 The required services now need to be started using dedicated commands:
 
 ```shell
-invoke dev.destroy dev.deps
+poetry run invoke dev.destroy dev.deps
 ```
 
 This will actually pass two commands, one to destroy any remains of a previous run and one to start services. So this will effectively bring up clean services without leftovers. We can see which services are running by using:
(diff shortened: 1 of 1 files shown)


petercrocker (approved):
Thanks for the corrections, they're how we run this, and really appreciate the contribution!!

petercrocker:
One suggestion @iddocohen (that you wouldn't have known) is if you rebase this on `stable` then the docs get updated immediately. For docs changes/corrections, we like having them updated immediately after merge.

iddocohen:
Thank you @petercrocker for the merge and the feedback.I am happy to add the best-practice workflow for documentation updates or fixes, to make it more clear in the future - this page seems like the right place to include it: https://docs.infrahub.app/development/git-best-practices

Before I do, I just want to double-check what the team is preferring:

1. For **all** documentation fixes only, should we branch off from **stable**?
2. Or should **all** documentation changes still branch from **develop**, and then be **rebased** onto **stable** in the **PR**?
3. Alternatively, should we branch from stable for fixes that apply to current docs, but branch from develop for documentation tied to new features?

ogenstad:
Looks like Pete already merged this to `develop`, the main difference is time for when it shows up live. We have a `release-1.5` branch which will be what will be merged into `stable` with the Infrahub 1.5 release. So anything that goes into `develop` will be released as part of Infrahub 1.6. For minor changes such as typos or corrections it's fine to target the documentation in `stable` directly which means it will be published directly to the documentation website and included in the next release we publish of Infrahub.

In this case we'll probably migrate from poetry to uv in the 1.6 release cycle but your change will still make it easier to find where we need to update the documentation.

So anything minor could be branched from `stable` directly and targeted to `stable`. For any other documentation update that mentions new features they should go either into `develop` or the current release branch.

Does this clarify things for you? (this process is also something we should document properly).

iddocohen:
Thanks @ogenstad for clarification. I will try to capture the workflow and logic around (`develop` -> `release-X` -> `stable`) in documentation and when/if doc fixes should go into stable vs develop. Stay tuned.

PR #7590, opened 2025-11-06
Release 1.5 to develop

ajtmccarty:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate edge creation during branch merges for nodes with migrated kinds.
  * Improved diff calculation to correctly exclude older versions of migrated nodes.

* **Documentation**
  * Added migration guide for updating schemas from `display_labels` to single `display_label` templates.
  * Updated schema documentation and examples to reflect new template-based display labels.


ajtmccarty:
this branch has a name that makes it protected, so cannot be updated
replaced by #7597

PR #7595, opened 2025-11-06
IFC-1891: Update branch query filter path time condition

### LINKED ISSUE #7338: bug: updated cardinality=one relationship on branch after merge has too many related nodes
### Component

API Server / GraphQL

### Infrahub version

1.4.10

### Current Behavior

if a branch with an update to a cardinality=one relationship is merged and then that relationship is updated on the branch after the merge, it will fail with an error like `ValidationError: Too many relationships, max 1 at field_name`

### Expected Behavior

it should be possible to correctly update a cardinality-one relationships on a branch after that branch is merged

### Steps to Reproduce

1. `branch1` that includes updated Relationship A->B is merged (this Relationship must be updated on the branch in order to trigger the failure condition)
2. on `branch1`, try to updated Relationship A->B to use a new peer and you will get an error

### Additional Information

this is the problem behind the scenes
1. `branch1` that includes Relationship A->B is merged
2. now `branch1.branched_from` has the same timestamp as the A->B Relationship on `main`
3. A->B is updated to A->C on `branch1`
4. now the standard cypher query branch-filtering logic we use all over the place will include both the A->B Relationship and the A->C Relationship
  a. the A->B Relationship has a timestamp equal to `branch1.branched_from` on `main` and the cypher filtering is `edge.from <= branch.branched_from` (changing `<=` to `<` might fix the issue)
  b. the A-> C Relationship exists on the branch with a timestamp between `branch1.branched_from` and Right Now

solababs:
Fixes https://github.com/opsmill/infrahub/issues/7338

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened time-bound path relationship filtering to fix incorrect boundary handling.
  * Deletions and queries now consistently use an explicit timestamp when none is provided, improving time-aware behavior.

* **Tests**
  * Added a unit test validating relationship merge and synchronization across branches.

* **Documentation**
  * Added a changelog entry documenting the path relationship filter fix.

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -335,7 +335,7 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin):
 
         delete_at = Timestamp(at)
 
-        query = await AttributeGetQuery.init(db=db, attr=self)
+        query = await AttributeGetQuery.init(db=db, attr=self, at=delete_at)
         await query.execute(db=db)
         results = query.get_results()
 
--- a/backend/infrahub/core/query/attribute.py
+++ b/backend/infrahub/core/query/attribute.py
@@ -177,7 +177,7 @@ class AttributeGetQuery(AttributeQuery):
         at = self.at or self.attr.at
         self.params["at"] = at.to_string()
 
-        rels_filter, rels_params = self.branch.get_query_filter_path(at=at.to_string(), include_exact_date_from=True)
+        rels_filter, rels_params = self.branch.get_query_filter_path(at=at.to_string(), include_exact_from_date=True)
         self.params.update(rels_params)
 
         query = (
--- a/backend/infrahub/core/query/attribute.py
+++ b/backend/infrahub/core/query/attribute.py
@@ -38,7 +38,7 @@ class AttributeQuery(Query):
         if at:
             self.at = Timestamp(at)
         else:
-            self.at = self.attr.at
+            self.at = Timestamp()
 
         self.branch = branch or self.attr.get_branch_based_on_support_type()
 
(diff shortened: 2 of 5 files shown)


backend/infrahub/core/query/attribute.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         if at:
             self.at = Timestamp(at)
         else:
-            self.at = self.attr.at
+            self.at = Timestamp()
```
ajtmccarty, line 41: I don't think it makes sense to default to the timestamp of the attribute in the query. defaulting to the current timestamp makes more sense

backend/infrahub/core/attribute.py
```diff
@@ -335,7 +335,7 @@ async def delete(self, db: InfrahubDatabase, at: Timestamp | None = None) -> Att
 
         delete_at = Timestamp(at)
 
-        query = await AttributeGetQuery.init(db=db, attr=self)
+        query = await AttributeGetQuery.init(db=db, attr=self, at=delete_at)
```
ajtmccarty, line 338: we want to get the attributes at the moment that we plan to delete them

changelog/7338.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed filter based on path relationship condition.
```
ajtmccarty: I don't think this means anything to someone unfamiliar with the code base.
I'd say something more like 

> Fixed bug that prevented retrieving cardinality-one relationships on a branch that was already merged and included changes to the relationship. This bug would be visible to the user as errors that look like `ValidationError: Too many relationships, max 1 at field_name`

ajtmccarty (approved):
added a couple comments explaining my commit and a suggestion for the changelog

I made some changes to how we instantiate `AttributeQuery` and `AttributeGetQuery` so that they default to using the current timestamp instead of the timestamp of the attribute being queried. This timestamp issue was causing the `AttributeGetQuery` to fail when retrieving attributes and properties to delete

ogenstad (approved):

PR #7610, opened 2025-11-07
fix: add NonNegativeInt scalar and validate GraphQL pagination args

iddocohen:
Fixing #7474 with new scalar class called NonNegativeInt()

## 
## Summary by CodeRabbit

* **Improvements**
  * Pagination arguments (limit, offset and related paging params) now enforce non-negative integers across GraphQL endpoints, improving validation and error clarity.

* **New Features**
  * Introduced a NonNegativeInt scalar to standardize non-negative integer validation for GraphQL inputs.

* **Tests**
  * Added unit tests verifying negative or invalid pagination inputs are rejected with "non-negative integer" validation errors.
 by CodeRabbit

* **New Features**
  * Added a non-negative integer input type for GraphQL, enabling explicit non-negative validation for numeric arguments.
* **Bug Fixes**
  * Pagination parameters (limit, offset and related paging args) now enforce non-negative values and return clearer errors for invalid inputs.
* **Tests**
  * Added unit tests to verify negative limit/offset values are rejected with appropriate error messages.

## Tested
  - pytest backend/tests/unit/graphql/test_query.py and passing:

image


backend/infrahub/graphql/types/scalars.py
```diff
@@ -0,0 +1,86 @@
+from typing import Any
+
+from graphene import Scalar
+from graphql import GraphQLError, language
```
ogenstad: Raising a GraphQLError in this case will hide some context of the query, while the change isn't significant if you instead had:

```python
from infrahub.exceptions import ValidationError
```

And raised a ValidationError some additional context with regards to where the error occurred would be provided as feedback to the user.

ogenstad (commented):
Thank you for picking this up! I added a small comment about the exceptions being raised. Also we use [towncrier](https://towncrier.readthedocs.io/en/stable/tutorial.html) when applicable to indicate what we've changed.

Can you create a towncrier newsfragment and include with this PR?

It's done by running `towncrier create` and when prompted for the issue you indicate the issue ID of the bug this is targeting and then select "fixed" as the type of change. Then just write what it was that was updated.

backend/infrahub/graphql/types/scalars.py
iddocohen: Thanks @ogenstad for the feedback!
I chose `GraphQLError` to stay consistent with Graphene’s internal error handling (e.g. `graphene.Int()`) allowing the executor to automatically attach locations and AST nodes and ensuring a GraphQL-compliant error structure.

My hesitation was with `ValidationError` is that it would surface a domain-specific exception into the Graphene layer, which I wanted to avoid.

That said, I am happy to switch if you prefer — I will follow your lead here.

ogenstad: The reasoning in this case is around what the GraphQL response would look like, if we look at a query that looks like this:

```graphql
query MyQuery {
  BuiltinTag(limit: -1, name__value: "red") {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
}
```

If we raise a `GraphQLError` the response will be:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Value must be a non-negative integer"
    }
  ]
}
```

If we instead raise a Validation error the output would be:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Expected value of type 'NonNegativeInt', found -1; Value must be a non-negative integer",
      "locations": [
        {
          "line": 2,
          "column": 21
        }
      ]
    }
  ]
}
```

Which while perhaps not pretty is more helpful as there's at least a chance for the user to track down where the error is happening within the provided query. It's also more in line with what Graphene would return if we have a type violation of the builtin types such as for a query like this:

```graphql
query MyQuery {
  BuiltinTag(name__value: -1) {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
}
```

Where the response would be:

```json
{
  "data": null,
  "errors": [
    {
      "message": "String cannot represent a non string value: -1",
      "locations": [
        {
          "line": 2,
          "column": 27
        }
      ]
    }
  ]
}
```

iddocohen: Thanks @ogenstad, it makes sense. Changed it to `ValidationError` and added changelog with `towncrier` accordingly.

petercrocker:
@iddocohen this still a valid PR? If so, can you resolve the conflicts, and I'll make sure it gets reviewed.

iddocohen:
Superseded by #9965. That PR rebases this work onto current `develop` and reworks it per your review here (@ogenstad): the scalar now raises `ValidationError` (so the executor attaches the argument type name and source location), and validation is extended to every paginated query. Note #7474 itself was already fixed and released in v1.9.5 by #9203 at the resolver level — #9965 complements that by validating declaratively at the schema level across all endpoints. Closing this in favour of #9965.

PR #7614, opened 2025-11-07
NodeManager overload cleanup for get methods

ogenstad:
Additional cleanups to the overload decorators on NodeManager.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for IP namespace validation to ensure explicit error reporting when namespace lookup fails.

* **Refactor**
  * Improved type system flexibility for internal object retrieval methods to support both string and type-based identifiers with nullable return types.

### DIFF
--- a/backend/infrahub/graphql/mutations/ipam.py
+++ b/backend/infrahub/graphql/mutations/ipam.py
@@ -48,7 +48,7 @@ async def validate_namespace(
         namespace_id = namespace.id
     elif "hfid" in data["ip_namespace"]:
         namespace = await registry.manager.get_one_by_hfid(
-            db=db, branch=branch, kind=InfrahubKind.IPNAMESPACE, hfid=data["ip_namespace"]["hfid"]
+            db=db, branch=branch, kind=InfrahubKind.IPNAMESPACE, hfid=data["ip_namespace"]["hfid"], raise_on_error=True
         )
         namespace_id = namespace.id
     else:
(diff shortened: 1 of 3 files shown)


backend/infrahub/graphql/mutations/ipam.py
```diff
@@ -48,7 +48,7 @@ async def validate_namespace(
         namespace_id = namespace.id
     elif "hfid" in data["ip_namespace"]:
         namespace = await registry.manager.get_one_by_hfid(
-            db=db, branch=branch, kind=InfrahubKind.IPNAMESPACE, hfid=data["ip_namespace"]["hfid"]
+            db=db, branch=branch, kind=InfrahubKind.IPNAMESPACE, hfid=data["ip_namespace"]["hfid"], raise_on_error=True
```
ogenstad, line 51: This change is required to make the type checkers happy, it looks like we always assumed that we'd get a return value here as we on line 53 tro to assign namespace_id from `namespace.id` if no object was found and we'd returned a `None` value we'd instead get an `AttributeError`. Before when we returned an `Any` the typechecker was not able to spot this. As such it should be fine to always raise an error if we don't find the node here.

backend/tests/integration/proposed_change/test_proposed_change_conflict.py
```diff
@@ -199,7 +199,7 @@ async def test_conflict_pipeline(
         john = await NodeManager.get_one_by_default_filter(db=db, id="John", kind=TestKind.PERSON)
         # The value of the description should match that of the source branch that was selected
         # as the branch to keep in the data conflict
-        assert john.description.value == "Oh boy"  # type: ignore[attr-defined]
+        assert john.description.value == "Oh boy"  # type: ignore[attr-defined, union-attr]
```
ogenstad, line 202: The idea is that we'll generate additional protocols for the test classes (we have a small number now) then when we start to use those within the tests we can phase out these type ignore statements.

ajtmccarty (approved):

PR #7616, opened 2025-11-08
add MetadataOptions flag, replace include_source/owner

ajtmccarty:
replaces `include_source` and `include_owner` params with a new `MetadataOptions` `Flag` that can capture all requested metadata in a single parameter

## Summary by CodeRabbit

* **Refactor**
  * Consolidated separate source/owner flags into a single metadata-options parameter. Callers across loaders, resolvers, mutations and tests were updated to use the new option. Result: related metadata returned by queries/updates is now selected via metadata options (e.g., source, owner, linked nodes); overall behavior preserved aside from this configurable metadata inclusion.


dgarros (approved):
Looks good

PR #7628, opened 2025-11-10
IFC-2043: Remove is_visible property from attribute metadata

solababs:
This PR removes `is_visible` property from attribute metadata.

## Summary by CodeRabbit

* **Chores**
  * Removed the per-attribute and per-relationship "visibility" flag — it no longer appears in APIs, GraphQL types, changelogs, or UI-exposed property lists.
* **Migrations**
  * Added a migration to delete existing visibility relationships from the graph.
* **Tests**
  * Updated tests and snapshots to reflect the absence of the visibility flag and the reduced set of property types.


backend/tests/integration/diff/test_diff_incremental_addition.py
ajtmccarty: should be able to just remove this `for` loop completely now

backend/tests/unit/core/diff/test_conflict_transferer.py
```diff
@@ -62,18 +62,10 @@ async def test_attr_conflicts(self, conflict_transferer: DiffConflictTransferer)
...
             property_type=DatabaseEdgeType.IS_PROTECTED, conflict=conflict_later_only
         )
-        conflict_transfer_1 = EnrichedConflictFactory.build(selected_branch=ConflictSelection.DIFF_BRANCH)
-        conflict_transfer_2 = replace(conflict_transfer_1, selected_branch=None)
-        prop_transfer_1 = EnrichedPropertyFactory.build(
-            property_type=DatabaseEdgeType.IS_VISIBLE, conflict=conflict_transfer_1
-        )
-        prop_transfer_2 = EnrichedPropertyFactory.build(
-            property_type=DatabaseEdgeType.IS_VISIBLE, conflict=conflict_transfer_2
-        )
```
ajtmccarty, line 72: please change these properties to be `DatabaseEdgeType.IS_PROTECTED` instead of deleting this part of the test

backend/tests/unit/core/diff/test_conflict_transferer.py
```diff
@@ -132,20 +123,10 @@ async def test_rel_prop_conflicts(self, conflict_transferer: DiffConflictTransfe
...
```
ajtmccarty, line 142: same here

backend/tests/unit/core/diff/test_conflicts_enricher.py
ajtmccarty, line 1: and same in this module

backend/tests/unit/core/diff/test_diff_and_merge.py
```diff
@@ -628,7 +628,7 @@ async def test_update_individual_relationship_properties_one_at_a_time(
         await diff_coordinator.update_branch_diff(base_branch=default_branch, diff_branch=branch2)
 
         car_branch = await NodeManager.get_one(db=db, branch=branch2, id=car_accord_main.id)
-        await car_branch.owner.update(db=db, data={"id": person_john_main.id, "_relation__is_visible": False})
```
ajtmccarty, line 635: please change this to be `_relation__is_protected: True` and update the checks below

backend/tests/unit/core/diff/test_diff_calculator.py
```diff
@@ -853,11 +827,6 @@ async def test_relationship_one_property_branch_update(
     assert single_relationship.action is DiffAction.UPDATED
     assert len(single_relationship.properties) == 2
     property_diff_by_type = {p.property_type: p for p in single_relationship.properties}
-    property_diff = property_diff_by_type[DatabaseEdgeType.IS_VISIBLE]
-    assert property_diff.property_type == DatabaseEdgeType.IS_VISIBLE
-    assert property_diff.previous_value is True
-    assert property_diff.new_value is False
-    assert before_branch_change < property_diff.changed_at < after_branch_change
```
ajtmccarty, line 860: please update this to check that `IS_PROTECTED` is updated from `False` to `True`

backend/tests/unit/core/diff/test_diff_calculator.py
```diff
@@ -881,11 +850,6 @@ async def test_relationship_one_property_branch_update(
...
```
ajtmccarty, line 888: same here

backend/tests/unit/core/diff/test_diff_calculator.py
```diff
@@ -2307,13 +2227,8 @@ async def test_branch_relationship_delete_with_property_update(
...
     assert set(prop_diff_by_type.keys()) == {
         DatabaseEdgeType.IS_RELATED,
-        DatabaseEdgeType.IS_VISIBLE,
+        DatabaseEdgeType.IS_PROTECTED,
     }
-    visible_prop = prop_diff_by_type[DatabaseEdgeType.IS_VISIBLE]
-    assert visible_prop.action is DiffAction.UPDATED
-    assert visible_prop.new_value is False
-    assert visible_prop.previous_value is True
-    assert before_main_change < visible_prop.changed_at < after_main_change
```
ajtmccarty, line 2316: same here

backend/tests/unit/core/diff/test_diff_calculator.py
ajtmccarty: can remove this `for` loop now

backend/tests/unit/core/test_branch_diff.py
```diff
@@ -601,16 +589,7 @@ async def test_diff_get_relationships(db: InfrahubDatabase, base_dataset_02) ->
...
-        "properties": {
-            "IS_VISIBLE": {
-                "branch": "branch1",
-                "type": "IS_VISIBLE",
-                "path": None,
-                "action": DiffAction.UPDATED,
-                "value": {"previous": True, "new": False},
-                "changed_at": Timestamp(base_dataset_02["time_m20"]),
-            }
-        },
```
ajtmccarty, line 613: it would be nice to update an `IS_PROTECTED` edge in the `base_dataset` so that this test still includes a property update

ajtmccarty (commented):
looks pretty good
just a few comments on updating some tests. sorry you have to go through those `DiffCalculator` tests. those are not fun

I think it would be good to also add a migration that deletes all `IS_VISIBLE` edges from the database. for a large database, this could be many edges, so it would be good to call the delete in transactions in a cypher query pretty similar to this one from the neo4j docs https://neo4j.com/docs/cypher-manual/current/subqueries/subqueries-in-transactions/#delete-with-call-in-transactions

backend/infrahub/core/migrations/graph/m048_remove_is_visible_relationship.py
```diff
@@ -0,0 +1,33 @@
...
+class RemoveIsVisibleRelationshipQuery(Query):
+    name = "remove_is_visible_relationship"
+    type: QueryType = QueryType.WRITE
+    insert_return = False
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        query = """
+        MATCH ()-[rel:IS_VISIBLE]->()
+        DELETE rel
+        """
```
ajtmccarty, line 25: tested this out on an old very large backup with over 5 million IS_VISIBLE edges and it crashes the database with an OOM memory. at least is does locally for me. not sure if it would actually be an issue for a production database, but better safe than sorry
this query will prevent the OOM error by committing batches as it runs
```
MATCH ()-[rel:IS_VISIBLE]->()
CALL (rel) {
  DELETE rel
} IN TRANSACTIONS
```

backend/tests/unit/core/migrations/graph/test_045.py
```diff
@@ -0,0 +1,57 @@
...
+    RETURN count(DISTINCT n) AS visible_nodes_count;
+    """
+    visible_nodes_count = await db.execute_query(query=count_is_visible_relationship_query)
+    assert visible_nodes_count[0].get("visible_nodes_count") == 0
+
+    car_name_attr = car_accord_main.get_attribute("name")
+    person_name_attr = person_john_main.get_attribute("name")
+
+    add_is_visible_relationship_query = """
+    CREATE (bool_true:Boolean { value: true })
```
ajtmccarty: doesn't really matter for this test, but you can do
```
MERGE (bool_true:Boolean { value: true })
LIMIT 1
```
to avoid creating a new Boolean vertex when one already exists

backend/tests/unit/core/migrations/graph/test_045.py
```diff
@@ -0,0 +1,57 @@
+from infrahub.core.migrations.graph import Migration045
+from infrahub.core.timestamp import current_timestamp
+from infrahub.database import InfrahubDatabase
+
+
+async def test_migration_045(db: InfrahubDatabase, default_branch, person_john_main, car_accord_main) -> None:
+    count_is_visible_relationship_query = """
+    MATCH (n)-[rel:IS_VISIBLE]-()
+    RETURN count(DISTINCT n) AS visible_nodes_count;
```
ajtmccarty: does not really matter for this test, but it might be more straightforward to just count number of IS_VISIBLE edges
```
MATCH ()-[:IS_VISIBLE]->() RETURN COUNT(*) as is_visible_count
```

ajtmccarty (approved):
I pushed a commit to fix a flaky test in `test_conflict_transferer`
local testing shows that getting your related PR in the SDK merged and then using that commit on this branch will solve most (all?) of the remaining functional and integration test failures

ajtmccarty:
looks like this one needs to have its migration bumped up by one b/c a migration 45 was added in stable
but after that is resolved, can this be merged?

backend/infrahub/core/diff/merger/serializer.py
```diff
@@ -93,9 +93,7 @@ def _convert_property_value(
         if property_type in (DatabaseEdgeType.HAS_OWNER, DatabaseEdgeType.HAS_SOURCE, DatabaseEdgeType.IS_RELATED):
             return raw_value
         # these are boolean
-        if (property_type in (DatabaseEdgeType.IS_VISIBLE, DatabaseEdgeType.IS_PROTECTED)) and isinstance(
-            raw_value, str
-        ):
+        if (property_type in (DatabaseEdgeType.IS_PROTECTED,)) and isinstance(raw_value, str):
```
ogenstad: This will eventually lead to linting errors, perhaps even if we rebase against `develop`. The reason is that we're checking to see if a property is within a single instance container.

The above could be rewritten as:

```python
if property_type == DatabaseEdgeType.IS_PROTECTED and isinstance(raw_value, str):
```

backend/infrahub/core/diff/query/merge.py
```diff
@@ -427,12 +427,12 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
             AND (peer_db_id IS NULL OR %(id_func)s(peer) = peer_db_id)
             // ------------------------------
-            // the serialized diff might not include the values for IS_VISIBLE and IS_PROTECTED in
+            // the serialized diff might not include the values for IS_PROTECTED in
             // some cases, so we need to figure them out here
             // ------------------------------
             CALL (attr_rel, property_diff) {
                 OPTIONAL MATCH (attr_rel)-[r_vis_pro]->(bool:Boolean)
-                WHERE property_diff.property_type IN ["IS_VISIBLE", "IS_PROTECTED"]
+                WHERE property_diff.property_type IN ["IS_PROTECTED"]
```
ogenstad: Kind of similar comment as above. Not sure if the IN is more expensive in Neo4j compared to a direct comparison?

backend/infrahub/core/query/relationship.py
```diff
@@ -996,10 +977,7 @@ def get_peers(self) -> Generator[RelationshipPeerData, None, None]:
                 properties={},
             )
 
-            for prop, metadata_option in [
-                ("is_protected", MetadataOptions.IS_PROTECTED),
-                ("is_visible", MetadataOptions.IS_VISIBLE),
-            ]:
+            for prop, metadata_option in [("is_protected", MetadataOptions.IS_PROTECTED)]:
```
ogenstad: The for-loop could be removed and refactored as we only have one entry ni the loop now.

backend/tests/integration/diff/test_diff_rebase.py
```diff
@@ -308,15 +308,14 @@ async def test_no_conflicts_before_merge(
...
-                DatabaseEdgeType.IS_VISIBLE,
                 DatabaseEdgeType.IS_PROTECTED,
             }
             related_prop = properties_by_type[DatabaseEdgeType.IS_RELATED]
             assert related_prop.action is DiffAction.UPDATED
             assert related_prop.previous_value == koenigsegg_id
             assert related_prop.new_value == new_peer_id
             assert related_prop.conflict is None
-            for prop_type, value in ((DatabaseEdgeType.IS_VISIBLE, "True"), (DatabaseEdgeType.IS_PROTECTED, "False")):
+            for prop_type, value in ((DatabaseEdgeType.IS_PROTECTED, "False"),):
```
ogenstad: Similar comment as above with the for loop

ogenstad (approved):

PR #7629, opened 2025-11-11
node-level create and read user timestamps

ajtmccarty:
part of IFC-1942

adds created/updated_at/by metadata to nodes when created, read, and deleted
mostly updates logic in the cypher queries to create, delete, and read a node from the database to support the new metadata on edges (`from_user_id` and `to_user_id`) and `Node`s (`updated_at`, `updated_by`, `created_at`, `created_by`)

logic for writing the data correctly during a non-create update will come later after Attributes and Relationships have the new metadata fully incorporated


backend/infrahub/core/constants/__init__.py
```diff
@@ -46,6 +46,8 @@
 
 EVENT_NAMESPACE = "infrahub"
 
+SYSTEM_USER_ID = "__system__"
```
ajtmccarty, line 49: default "user ID" to use for updated/created_by/at metadata updates

backend/infrahub/core/metadata/base.py
ajtmccarty, line 1: new, very simple base class to use on `Node`, `Attribute`, and `Relationship` classes to track the new metadata

backend/infrahub/core/node/__init__.py
```diff
@@ -120,9 +121,6 @@ def get_id(self) -> str:
 
         raise InitializationError("The node has not been saved yet and doesn't have an id")
 
-    def get_updated_at(self) -> Timestamp | None:
```
ajtmccarty, line 123: replaced by version in `MetadataBase`

backend/infrahub/core/node/__init__.py
```diff
@@ -856,18 +858,20 @@ async def _create(self, db: InfrahubDatabase, at: Timestamp | None = None) -> No
...
         node_changelog = NodeChangelog(node_id=self.get_id(), node_kind=self.get_kind(), display_label="")
+        self.set_updated_at(update_at)
+        self.set_updated_by(user_id)
 
         # Go over the list of Attribute and update them one by one
         for name in self._attributes:
             if (fields and name in fields) or not fields:
-                attr: BaseAttribute = getattr(self, name)
-                updated_attribute = await attr.save(at=update_at, db=db)
+                attr = self.get_attribute(name=name)
```
ajtmccarty, line 888: just a more type-hint friendly way of getting an `Attribute` or `RelationshipManager` for a node

backend/infrahub/core/node/__init__.py
```diff
@@ -962,22 +972,7 @@ async def delete(self, db: InfrahubDatabase, at: Timestamp | None = None) -> Non
...
-            filters={"id": self.id},
-            branch=self._branch,
-            at=delete_at,
-            order=OrderModel(disable=True),
-        )
-        await query.execute(db=db)
-        result = query.get_result()
-
-        if result and result.get("rb.branch") == branch.name:
-            await update_relationships_to([result.get("rb_id")], to=delete_at, db=db)
```
ajtmccarty, line 978: I'm trying to remove this wherever I find them
the `NodeDeleteQuery` should handle this at the database level inside of its cypher query instead of requiring this extra processing in memory and a separate trip to the database

backend/infrahub/core/query/node.py
```diff
@@ -219,14 +222,38 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
         }
+        if self.branch.is_default or self.branch.is_global:
+            self.params["node_prop"].update(
+                {
+                    "created_at": at.to_string(),
+                    "created_by": self.user_id,
+                    "updated_at": at.to_string(),
+                    "updated_by": self.user_id,
+                }
+            )
```
ajtmccarty, line 233: we only add these properties to the `Node` vertex if we are on the default or global branch to make querying for metadata on those branches faster than on user-created branches

backend/infrahub/core/query/node.py
```diff
@@ -219,14 +222,38 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
+
+        rel_edge_prop_str = "{ branch: rel.branch, branch_level: rel.branch_level, status: rel.status, from: $at, from_user_id: $user_id }"
+        rel_edge_prop_str_hierarchy = (
+            "{ branch: rel.branch, branch_level: rel.branch_level, "
+            "status: rel.status, hierarchy: rel.hierarchical, from: $at, from_user_id: $user_id }"
+        )
+        rel_vertex_prop_str = "{ uuid: rel.uuid, name: rel.name, branch_support: rel.branch_support"
+        if self.branch.is_default or self.branch.is_global:
+            rel_vertex_prop_str += ", created_at: $at, created_by: $user_id, updated_at: $at, updated_by: $user_id"
+        rel_vertex_prop_str += " }"
```
ajtmccarty, line 259: mostly refactoring to move a bunch of repeated property create strings into one place
and add the updated/created_by/at metadata to them

backend/infrahub/core/query/node.py
```diff
@@ -525,48 +575,55 @@ def get_ids(self) -> dict[str, tuple[str, str]]:
...
         query = """
-        MATCH (root:Root)
-        CREATE (n)-[r:IS_PART_OF { branch: $branch, branch_level: $branch_level, status: "deleted", from: $at }]->(root)
+MATCH (root:Root)
+LIMIT 1
+CREATE (n)-[delete_edge:IS_PART_OF { branch: $branch, branch_level: $branch_level, status: "deleted", from: $at, from_user_id: $user_id }]->(root)
+WITH r
+WHERE r.branch = $branch
+SET r.to = $at
+SET r.to_user_id = $user_id
```
ajtmccarty, line 628: now handles setting the `to` time on edges that used to be handled in the `update_relationships_to` function

backend/infrahub/core/query/node.py
```diff
@@ -980,11 +1037,63 @@ class NodeListGetInfoQuery(Query):
...
         self.ids = ids
+        self.include_metadata = include_metadata
         super().__init__(**kwargs)
 
+    def _needs_user_timestamp_metadata(self) -> bool:
+        return bool(self.include_metadata & MetadataOptions.USER_TIMESTAMPS)
+
+    def _add_updated_metadata_to_query(self, branch_filter_str: str) -> None:
+        if self.branch.is_default or self.branch.is_global:
+            last_update_query = """
```
ajtmccarty, line 1054: we can use this very simple addition to the query on the default or global branch

backend/infrahub/core/query/node.py
```diff
@@ -980,11 +1037,63 @@ class NodeListGetInfoQuery(Query):
...
+WITH *, n.updated_at AS updated_at, n.updated_by AS updated_by
+            """
+        else:
+            last_update_query = """
```
ajtmccarty, line 1058: we have to use this more complex query on a user-created branch b/c we have to look at every edge under the node to find its latest update_at/by

backend/infrahub/core/manager.py
```diff
@@ -1206,6 +1205,10 @@ async def get_many(
             node_branch = await registry.get_branch(db=db, branch=node.branch)
             item = await node_class.init(schema=node.schema, branch=node_branch, at=at, db=db)
             await item.load(**new_node_data, db=db)
+            item.set_created_at(node.created_at)
+            item.set_created_by(node.created_by)
+            item.set_updated_at(node.updated_at)
+            item.set_updated_by(node.updated_by)
```
ogenstad: Even though it would be ugly to use private methods here, if we do it like this for correctness we should really add these four names as restricted names for attribute and relationship names.

backend/infrahub/core/metadata/base.py
```diff
@@ -0,0 +1,42 @@
...
+
+@dataclass
+class MetadataInfo:
+    created_at: Timestamp | None = None
+    created_by: str | None = None
+    updated_at: Timestamp | None = None
+    updated_by: str | None = None
+
+
+class MetadataBase:
```
ogenstad: Do we really need this to be a mixin? Wouldn't it be cleaner to initialize a ? .meta (or ._meta) within the Node class?

backend/infrahub/core/node/__init__.py
```diff
@@ -786,19 +783,12 @@ async def load(
...
         self.id = id
         self.db_id = db_id
         self._existing = True
 
-        if updated_at:
-            kwargs["updated_at"] = (
-                updated_at  # FIXME: Allow users to use "updated_at" named attributes until we have proper metadata handling
-            )
-            self._updated_at = Timestamp(updated_at)
-
```
ogenstad, line 801: 🎉

backend/infrahub/core/query/node.py
```diff
@@ -219,14 +222,40 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
             "namespace": self.node._schema.namespace,
             "branch_support": self.node._schema.branch,
         }
+        if self.branch.is_default or self.branch.is_global:
+            self.params["node_prop"].update(
+                {
+                    "created_at": at.to_string(),
+                    "created_by": self.user_id,
+                    "updated_at": at.to_string(),
+                    "updated_by": self.user_id,
```
ogenstad, line 231: I might be missing some context here but it looks strange that created_by would always be the same as updated_by, is this only if the user id is included as part of the query when looking for nodes created (and updated) last by that user?

ogenstad (commented):
LGTM overall, but I think it would be nice if we got solve it in a different way than using a Mixin and also if we could avoid adding names that would restrict valid names for attributes and relationships.

backend/infrahub/core/query/node.py
ajtmccarty: this is in the `NodeCreateAllQuery`, so it will only run when creating a node
updates and deletes use different queries
so we only set the updated/created metadata to the same value when a node is being created on main/global

backend/infrahub/core/metadata/base.py
ajtmccarty: you're right
I was thinking it would be easier to use this in `BaseAttribute` and `Relationship` if it were a base class, but it is more solid to make this an interface and then have `Node`, `BaseAttribute`, and `Relationship` responsible for setting and getting the values

backend/infrahub/core/manager.py
ajtmccarty: moving to private methods

ajtmccarty (commented):
thanks for the review Patrick

ajtmccarty:
@ogenstad please re-review if you have time

ogenstad (approved):
LGTM. I'm still not convinced we need the MetadataInterface mixin in `class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta)`. But I think we can reconsider how to manage things like that once we refactor this at an overall level with that NodeModifier (name pending) class.

PR #7635, opened 2025-11-12
attribute-level immutable metadata

ajtmccarty:
adds updated/created_at/by metadata to Attributes when they are created, updated, read, and deleted

adds new more granular support for requesting specific metadata when retrieving objects using `MetadataQueryOptions` so that you can request one level of metadata for the retrieved nodes and a different level for retrieved attributes

adds a new class that determines which metadata is required based on the requested node fields

updates all of the Attribute CRUD cypher queries to do everything in one query instead of multiple queries


backend/infrahub/core/query/attribute.py
ajtmccarty, line 1: the changes to these queries are mostly two things
- updates to set the `to` time on edges when making an update so that calling `update_relationship_to` from Python is no longer required
- updates to set the `updated_at` and `updated_by` properties on the `:Attribute` if the update is on the main or global branch

backend/infrahub/core/query/attribute.py
```diff
@@ -228,43 +263,119 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
+    WITH a
+    WHERE $branch_level = 1
+    LIMIT 1
+    SET a.updated_at = $at, a.updated_by = $user_id
 }
         """ % {"branch_filter": branch_filter, "rel_label": rel_label}
         self.add_to_query(query)
 
 
-class AttributeGetQuery(AttributeQuery):
```
ajtmccarty, line 242: this was only used as part of the logic to delete an Attribute
the previous flow was to get the attribute, then call a different cypher query for each edge that we needed to delete

the new approach is to just do it all in cypher and return whatever the previous value was so that it can go into the attribute's changelog

backend/infrahub/core/query/node.py
ajtmccarty, line 673: these changes to the `NodeListGetAttributeQuery` are to add support for querying for updated/created_at/by metadata on attributes and to only query for the specific metadata requested b/c querying for updated/created_at/by metadata on a branch will be slow, so we only want to do it if absolutely necessary

backend/infrahub/core/attribute.py
```diff
@@ -326,78 +357,44 @@ def deserialize_value(self, data: AttributeFromDB) -> Any:
...
-            dst_node_id=self.db_id,
-            rel_type="HAS_ATTRIBUTE",
-            branch_name=branch.name,
-            branch_level=branch.hierarchy_level,
-            at=delete_at,
-            status=RelationshipStatus.DELETED,
-            db=db,
-        )
-
-        return changelog
```
ajtmccarty, line 398: this is what we were doing to delete an attribute

backend/infrahub/core/manager.py
```diff
@@ -41,6 +42,13 @@
 SchemaProtocol = TypeVar("SchemaProtocol")
 
 
+@dataclass
+class MetadataQueryOptions:
+    node_level: MetadataOptions = MetadataOptions.NONE
+    attribute_level: MetadataOptions = MetadataOptions.NONE
+    relationship_level: MetadataOptions = MetadataOptions.NONE
```
ajtmccarty, line 49: all the changes in this module are just to add support for using this in the `include_metadata` parameter so that we can specify different levels of metadata for the node, attributes, and relationships

backend/infrahub/graphql/loaders/node.py
```diff
@@ -61,7 +57,6 @@ async def batch_load_fn(self, keys: list[Any]) -> list[Node | None]:
                 branch=self.query_params.branch,
                 include_metadata=self.query_params.include_metadata,
                 prefetch_relationships=self.query_params.prefetch_relationships,
-                account=self.query_params.account,
```
ajtmccarty, line 64: turns out `account` is not actually used for anything

backend/infrahub/graphql/resolvers/resolver.py
```diff
@@ -189,6 +186,14 @@ async def default_paginated_list_resolver(
 
         objs = []
         if edges or "hfid" in filters:
+            metadata_determiner = MetadataDeterminer(
+                schema_branch=db.schema.get_schema_branch(name=graphql_context.branch.name)
+            )
+            include_metadata = await metadata_determiner.determine_metadata(
+                schema=schema,
+                node_fields=node_fields,
+            )
```
ajtmccarty, line 195: use the actual fields requested to determine what metadata we need for the node and attribute levels

gmazoyer (approved):

PR #7652, opened 2025-11-14
Fix query key for object details to include attributes and relationships

pa-lem:
This will avoid cache collision for queries on object details

## Summary by CodeRabbit

* **Refactor**
  * Enhanced object detail query key generation to better support attribute and relationship visibility customization, improving cache handling for object detail views.

### DIFF
--- a/frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
+++ b/frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
@@ -1,6 +1,10 @@
 import type { ContextParams } from "@/shared/api/types";
 import type { Filter } from "@/shared/hooks/useFilters";
 
+import { getAttributesVisibleInDetailedView } from "@/entities/nodes/object/utils/get-attributes-visible-in-detailed-view";
+import { getRelationshipsVisibleInDetailedView } from "@/entities/nodes/object/utils/get-relationships-visible-in-detailed-view";
+import type { AttributeSchema, ModelSchema, RelationshipSchema } from "@/entities/schema/types";
+
 export interface ObjectKeysBaseParams extends ContextParams {
   objectKind: string;
 }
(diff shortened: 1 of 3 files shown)


frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
```diff
@@ -11,6 +15,9 @@ export interface ObjectListKeysParams extends ObjectKeysBaseParams {
 
 export interface ObjectDetailKeysParams extends ObjectKeysBaseParams {
   objectId: string;
+  objectSchema?: ModelSchema;
+  getAttributesVisible?: (attributes: AttributeSchema[]) => AttributeSchema[];
+  getRelationshipsVisible?: (relationships: RelationshipSchema[]) => RelationshipSchema[];
```
bilalabbad, line 20: args should never be optional on query keys.
Here for example you could split in 2:
- details => all the details queries
- detail => details page

frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
```diff
@@ -38,9 +45,38 @@ export const objectQueryKeys = {
...
+  if (!params?.objectSchema?.relationships) {
+    return [];
+  }
+
+  const getRelationships = params?.getRelationshipsVisible ?? getRelationshipsVisibleInDetailedView;
+
+  return getRelationships(params.objectSchema.relationships).map((relationship) => {
+    return relationship.name;
+  });
+};
```
bilalabbad, line 82: Tanstack do a deep comparaison already. No need to spread nor map a arrray of string

frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
pa-lem: it's already optional for other params, and getAttributesVisible and getRelationshipsVisible are optional in the main query

frontend/app/src/entities/nodes/object/domain/object.query-keys.ts
pa-lem: it's to make it consistent to all the other keys and to reuse the same cache if the same attributes and relationships are used, even if some config changes in one of those attributes

PR #7657, opened 2025-11-14
IFC-1951 Support relationships in profiles

gmazoyer:
### IFC-1952 Add schema part to handle relationships in profile
This change is relatively simple and just adds relationships of kind generic and kind attribute to profile schemas corresponding to node schemas having these relationships.

When we load a schema, we make sure to clear all the relationships, except the ones of kind PROFILE such as related_nodes, to create a brand new list of relationships which will account for deleted and new ones. This is very similar to what we do for templates already.

### IFC-1961 Add ability to clear out profile relationships on updates
This change makes sure that relationships coming from profile are clearout when an update is made on the relationship itself. This works combined with IFC-1953 which actually apply the profiles. This change, in the end, was mostly about covering the `RelationshipAdd`/`RelationshipRemove` mutations endpoints when the `profiles` or `related_nodes` relationships are mutated.

### IFC-1953 Add the required logic to apply relationships of profiles
This change allows to actually set or clear peers to the relationships of a node if that node has one or more profiles. If there are more than one profile we follow the profile priority value as we used to do before. If a node has different peers for a relationship, the profile peer set won't be applied.

### IFC-1954 Add events to trigger applying profiles
This change adds the required prefect automation mechanisms to trigger re-applying a profile on its related nodes when an attribute or a relationship changes on the profile.

Note: we do not support mandatory relationship in this PR, and this is by design.

## Summary by CodeRabbit

* **New Features**
  * Profiles can now drive and manage relationships and attributes across objects; manual and automated profile-refresh workflows and triggers added; frontend hook/query and selector integration to fetch profiles.

* **Improvements**
  * Profile provenance and priority surfaced and respected for defaults and relationship fallbacks; forms resolve relationship defaults from profiles; IPAM entities now reference namespaces.

* **Tests**
  * Expanded unit, integration and e2e coverage and new fixtures; test assertions adjusted for profile-driven UI changes.


ogenstad (approved):

frontend/app/src/entities/nodes/api/getProfiles.ts
```diff
@@ -22,7 +22,31 @@ export const getProfiles = Handlebars.compile(`
...
+              node {
+                id
+                display_label
+                hfid
+                __typename
+              }
+              {{/if}}
+            }
+
+          {{/each}}
```
bilalabbad: It could have been nicer to migrate it to tanstack query instead of reusing legacy code...

frontend/app/src/shared/api/graphql/utils.ts
```diff
@@ -70,31 +70,33 @@ export const addRelationshipsToRequest = (
...
+      },
+      owner: {
+        id: true,
+        hfid: true,
+        display_label: true,
+        __typename: true,
+      },
+    },
+  };
+
```
bilalabbad: Nothing changes right? If so, we should just remove it rather than adding unnecessary noise to the git history.

bilalabbad (commented):
- When opening metadata popover for relationship of cardinality one, the value is not styled correctly

frontend/app/src/shared/components/display/meta-details-tooltips.tsx
```diff
@@ -1,4 +1,5 @@
 import { Icon } from "@iconify-icon/react";
+import { FileBoxIcon } from "lucide-react";
```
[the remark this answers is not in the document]
bilalabbad: in lucide-react, both both with and without "icon" exists

bilalabbad (approved):

PR #7660, opened 2025-11-15
IFC-1953 Apply relationships from profiles

gmazoyer:
This change brings some logic so that the `NodeProfilesApplier`, which is in charge of applying profiles attributes to nodes, works on applying profiles relationships to nodes too.

This works by looking at which profiles to apply, resolving relationships on them, if any and creating relationships to the same peers for the node on which profiles are applied on.

### DIFF
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -12,7 +12,7 @@ from infrahub.core.changelog.models import (
     RelationshipCardinalityManyChangelog,
     RelationshipCardinalityOneChangelog,
 )
-from infrahub.core.constants import RelationshipDirection, RelationshipStatus
+from infrahub.core.constants import InfrahubKind, RelationshipDirection, RelationshipStatus
 from infrahub.core.constants.database import DatabaseEdgeType
 from infrahub.core.query import Query, QueryType
 from infrahub.core.query.subquery import build_subquery_filter, build_subquery_order
@@ -101,6 +101,9 @@ class RelationshipPeerData:
     """Both relationships pointing at this Relationship Node."""
 
     updated_at: str | None = None
+
+    is_from_profile: bool = False
+    profile_id: UUID | None = None
 
     def rel_ids_per_branch(self) -> dict[str, list[str | int]]:
         response = defaultdict(list)
@@ -850,6 +853,9 @@ class RelationshipGetPeerQuery(Query):
 
             if hasattr(self.rel, "_node_properties"):
                 for prop in self.rel._node_properties:
+                    if prop == "source" and (source := result.get("source")):
+                        data.is_from_profile = InfrahubKind.PROFILE in source.labels
+                        data.profile_id = source._properties["uuid"]
                     if prop_node := result.get(prop):
                         data.properties[prop] = NodePropertyData(
                             name=prop,
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -853,8 +853,8 @@ class RelationshipGetPeerQuery(Query):
 
             if hasattr(self.rel, "_node_properties"):
                 for prop in self.rel._node_properties:
-                    if prop == "source" and (source := result.get("source")):
-                        data.is_from_profile = InfrahubKind.PROFILE in source.labels
+                    if prop == "source" and (source := result.get("source")) and InfrahubKind.PROFILE in source.labels:
+                        data.is_from_profile = True
                         data.profile_id = source._properties["uuid"]
                     if prop_node := result.get(prop):
                         data.properties[prop] = NodePropertyData(
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -748,12 +748,15 @@ class RelationshipValidatorList:
 
 
 class RelationshipManager:
-    def __init__(self, schema: RelationshipSchema, branch: Branch, at: Timestamp, node: Node) -> None:
+    def __init__(
+        self, schema: RelationshipSchema, branch: Branch, at: Timestamp, node: Node, is_from_profile: bool = False
+    ) -> None:
         self.schema: RelationshipSchema = schema
         self.name: str = schema.name
         self.node: Node = node
         self.branch: Branch = branch
         self.at = at
+        self.is_from_profile = is_from_profile
 
         # TODO Ideally this information should come from the Schema
         self.rel_class = Relationship
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -1027,6 +1027,14 @@ class RelationshipManager:
         for peer_id in details.peer_ids_present_local_only:
             await self.remove_locally(peer_id=peer_id, db=db)
 
+        has_profile_source = False
+        for rel in self._relationships:
+            source_node = await rel.get_source(db=db)
+            if source_node and source_node.get_schema().is_profile_schema:
+                has_profile_source = True
+                break
+        self.is_from_profile = has_profile_source
+
     async def get(self, db: InfrahubDatabase) -> Relationship | list[Relationship] | None:
         rels = await self.get_relationships(db=db)
 
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -1072,6 +1072,12 @@ class RelationshipManager:
 
         return self._relationships.as_list()
 
+    async def get_relationship(self, db: InfrahubDatabase, peer_id: str) -> Relationship | None:
+        for rel in await self.get_relationships(db=db):
+            if rel.peer_id == peer_id:
+                return rel
+        return None
+
     async def update(self, data: list[str | Node] | dict[str, Any] | str | Node | None, db: InfrahubDatabase) -> bool:
         """Replace and Update the list of relationships with this one."""
         if not isinstance(data, list):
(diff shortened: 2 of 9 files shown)


backend/infrahub/profiles/node_applier.py
```diff
@@ -173,11 +173,9 @@ async def _remove_profile_from_relationship(self, relationship_manager: Relation
     async def apply_profiles(self, node: Node) -> list[str]:
         profile_ids = await self._get_profile_ids(node=node)
         attr_names_for_profiles = await self._get_attr_names_for_profiles(node=node)
-        if not attr_names_for_profiles:
-            return []
         rel_names_for_profiles = await self._get_rel_names_for_profiles(node=node)
         rel_filters_for_profiles = await self._get_rel_filters_for_profiles(node=node, rel_names=rel_names_for_profiles)
-        if not rel_filters_for_profiles:
+        if not attr_names_for_profiles and not rel_filters_for_profiles:
```
gmazoyer, line 174: Nice catch 👍

backend/infrahub/core/relationship/model.py
```diff
@@ -1024,6 +1027,14 @@ async def _fetch_relationships(
         for peer_id in details.peer_ids_present_local_only:
             await self.remove_locally(peer_id=peer_id, db=db)
 
+        has_profile_source = False
+        for rel in self._relationships:
+            source_node = await rel.get_source(db=db)
```
ajtmccarty: I believe that this could end up sending a database query for every peer in a relationship if the source is `None`, which it usually will be
we might need to make some changes to how `fetch_relationship_ids` works to retrieve whether a relationship is from a profile or not in order to avoid this

backend/infrahub/profiles/node_applier.py
```diff
@@ -33,13 +36,55 @@ async def _get_attr_names_for_profiles(self, node: Node) -> list[str]:
...
+    async def _get_rel_filters_for_profiles(self, node: Node, rel_names: list[str]) -> list[RelationshipFilter]:
+        node_schema = node.get_schema()
+
+        identifiers: list[RelationshipFilter] = []
+        for rel_name in rel_names:
+            rel_schema = node_schema.get_relationship(name=rel_name)
+
+            # We are past schema validation so we should have an identifier
+            if not rel_schema.identifier:
+                raise ValueError(f"Relationship {rel_name} has no identifier")
```
ajtmccarty: you can use `rel_schema.get_identifier()` to do this same thing

backend/infrahub/profiles/node_applier.py
```diff
@@ -67,35 +112,117 @@ def _remove_profile_from_attribute(self, node_attr: BaseAttribute) -> None:
...
+        if current_peer_ids and peer_ids:
+            for peer_id in peer_ids:
+                if peer_id in current_peer_ids:
+                    await node_rel.remove_locally(db=self.db, peer_id=peer_id)
+                    is_changed = True
+
+            if is_changed:
+                node_rel.is_from_profile = False
+                await node_rel.save(db=self.db)
+            return is_changed
```
ajtmccarty: if I have an object that includes a relationship from a profile to objects `B` and `C` and I want to override this relationship to be to objects `C` and `D`, is this possible? or will the relationship to object `C` always be removed?

backend/infrahub/profiles/node_applier.py
```diff
@@ -67,35 +112,117 @@ def _remove_profile_from_attribute(self, node_attr: BaseAttribute) -> None:
...
+
+        if node_rel.schema.cardinality == RelationshipCardinality.ONE:
+            target_peer_ids = {peer_ids[0]} if peer_ids else set()
+        else:
+            target_peer_ids = set(peer_ids)
```
ajtmccarty: can these both just be `target_peer_ids = set(peer_ids)`?

backend/infrahub/profiles/node_applier.py
```diff
@@ -67,35 +112,117 @@ def _remove_profile_from_attribute(self, node_attr: BaseAttribute) -> None:
...
+
+        # Remove relationships that are from this profile but not in target
+        for rel in current_rels:
+            source = await rel.get_source(db=self.db)
+            if node_rel.is_from_profile and source and source.id == profile_id:
+                if rel.peer_id and rel.peer_id not in target_peer_ids:
+                    await node_rel.remove_locally(peer_id=rel.peer_id, db=self.db)
+                    node_rel.is_from_profile = True
+                    is_changed = True
```
ajtmccarty: can this be refactored so that `get_source` is only called when necessary, as it can trigger a database query?

backend/tests/unit/core/profiles/test_node_applier.py
```diff
@@ -214,3 +218,389 @@ async def test_get_many_with_multiple_profiles_same_priority(
...
+    await missing_child_profile.new(
+        db=db, profile_name="missing", things=[child_and_thing_nodes.thing_nodes[2]], profile_priority=200
+    )
+    await missing_child_profile.save(db=db)
+
+    await child_one.profiles.update(db=db, data=[augmented_child_profile])
+    await child_one.save(db=db)
+
+    await child_two.profiles.update(db=db, data=[missing_child_profile])
+    await child_two.save(db=db)
```
ajtmccarty, line 443: should `child_two` also be validated with `_validate_node_profile_relationships` below?

ajtmccarty (approved):
looks good, just a few small comments
I am a little concerned about performance b/c it looks like we are basically handling one relationship peer at a time, so if a many relationships are added to a profile used by many nodes, then I'm worried this will execute at least one db query per relationship per node, but I might be missing something

backend/infrahub/core/relationship/model.py
gmazoyer: I tested a tweak that seems to work at avoiding the database queries:
```diff
diff --git a/backend/infrahub/core/query/relationship.py b/backend/infrahub/core/query/relationship.py
index 3c69a7d4c..3684cd7a2 100644
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -12,7 +12,7 @@ from infrahub.core.changelog.models import (
     RelationshipCardinalityManyChangelog,
     RelationshipCardinalityOneChangelog,
 )
-from infrahub.core.constants import RelationshipDirection, RelationshipStatus
+from infrahub.core.constants import InfrahubKind, RelationshipDirection, RelationshipStatus
 from infrahub.core.constants.database import DatabaseEdgeType
 from infrahub.core.query import Query, QueryType
 from infrahub.core.query.subquery import build_subquery_filter, build_subquery_order
@@ -102,6 +102,8 @@ class RelationshipPeerData:

     updated_at: str | None = None

+    is_from_profile: bool = False
+
     def rel_ids_per_branch(self) -> dict[str, list[str | int]]:
         response = defaultdict(list)
         for rel in self.rels:
@@ -850,6 +852,8 @@ class RelationshipGetPeerQuery(Query):

             if hasattr(self.rel, "_node_properties"):
                 for prop in self.rel._node_properties:
+                    if prop == "source" and (source := result.get("source")):
+                        data.is_from_profile = InfrahubKind.PROFILE in source.labels
                     if prop_node := result.get(prop):
                         data.properties[prop] = NodePropertyData(
                             name=prop,
diff --git a/backend/infrahub/core/relationship/model.py b/backend/infrahub/core/relationship/model.py
index 9c4a5e206..1d2640a28 100644
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -983,7 +983,9 @@ class RelationshipManager:

         peers = await self.get_db_peers(db=db, at=at, branch_agnostic=branch_agnostic)

-        peers_database: dict = {str(peer.peer_id): peer for peer in peers}
+        self.is_from_profile = any(peer.is_from_profile for peer in peers)
+
+        peers_database = {str(peer.peer_id): peer for peer in peers}
         peer_ids = list(peers_database.keys())

         # Calculate which peer should be added or removed
@@ -1027,14 +1029,6 @@ class RelationshipManager:
         for peer_id in details.peer_ids_present_local_only:
             await self.remove_locally(peer_id=peer_id, db=db)

-        has_profile_source = False
-        for rel in self._relationships:
-            source_node = await rel.get_source(db=db)
-            if source_node and source_node.get_schema().is_profile_schema:
-                has_profile_source = True
-                break
-        self.is_from_profile = has_profile_source
-
     async def get(self, db: InfrahubDatabase) -> Relationship | list[Relationship] | None:
         rels = await self.get_relationships(db=db)
```
How does that look to you?

gmazoyer: Added these changes as a commit to this PR

backend/infrahub/profiles/node_applier.py
gmazoyer: @ajtmccarty you got that right. I reproduced this with a test. I think we can have a fix that would work if the `C` is removed to be then re-added. If we kept `C` as is, it will have a profile as a source which would be wrong. Unless we find a way to `clear_source()` of `C`.

ajtmccarty: I'm making some changes to how the relationship update queries work in my immutable metadata branch that will hopefully be merged back to develop early next week and those should allow clearing a source on a relationship
might be worth adding a small issue to the epic to track this

backend/infrahub/core/query/relationship.py
```diff
@@ -850,6 +853,9 @@ def get_peers(self) -> Generator[RelationshipPeerData, None, None]:
 
             if hasattr(self.rel, "_node_properties"):
                 for prop in self.rel._node_properties:
+                    if prop == "source" and (source := result.get("source")):
+                        data.is_from_profile = InfrahubKind.PROFILE in source.labels
+                        data.profile_id = source._properties["uuid"]
```
ajtmccarty, line 858: I think this could end up setting `data.profile_id` even if the source is not a `Profile` Node

ajtmccarty (approved):
looks good

PR #7680, opened 2025-11-18
IFC-1961 Add ability to clear out profile relationships on updates

gmazoyer:
This change makes sure that we run the `NodeProfilesApplier` when we mutate a relationship of some nodes by using the `RelationshipAdd` or `RelationshipRemove` mutations.

Previously, when adding or removing a profile via these mutations, the values of attributes were not updated according to what was in the profiles. This change fix this behaviour. Combined with changes from IFC-1953, it will also take care of applying or un-applying peers to relationships coming from profiles.

### DIFF
--- a/backend/infrahub/graphql/mutations/relationship.py
+++ b/backend/infrahub/graphql/mutations/relationship.py
@@ -32,6 +32,7 @@ from infrahub.graphql.context import apply_external_context
 from infrahub.graphql.types.context import ContextInput
 from infrahub.groups.ancestors import collect_ancestors
 from infrahub.permissions import get_global_permission_for_kind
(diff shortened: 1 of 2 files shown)


ajtmccarty (commented):
good catch
I think we should also support updating profiles when a related node is added to or removed from a profile from the other side. ie I use the RelationshipAdd mutation where the source is a `ProfileInterface` instance, the relationship name is "related_nodes", and the peers are `Interface` instances

ogenstad (approved):
Would be good with some tests that covers these paths.

gmazoyer:
Added tests to cover these new paths.

PR #7683, opened 2025-11-19
IFC-1940: Add metadata to graphql response

solababs:


backend/infrahub/core/node/__init__.py
```diff
@@ -1076,6 +1081,23 @@ async def to_graphql(
 
         return response
 
+    async def _build_meta_response(self, db: InfrahubDatabase, field_name: str, fields: dict) -> dict:
+        data = {}
+        for meta_field in fields.get(field_name).keys():
```
ogenstad: This looks like it could potentially be problematic if `fields.get(field_name)` were to return `None` and we then call `.keys()` on None. It might be that within this context it should never happen. But in the function signature we don't have any type of guarantee that this would be the case. Is there something we can change around this?

backend/infrahub/core/node/__init__.py
```diff
@@ -1076,6 +1081,23 @@ async def to_graphql(
...
+            meta_field_data: RelationshipManager | str | None = getattr(self, meta_field, None)
+
+            if meta_field_data is not None and isinstance(meta_field_data, RelationshipManager):
+                peer = await meta_field_data.get_peer(db=db)
+                data[meta_field] = peer.id if peer else None
+                continue
+
+            if meta_field_data is not None and isinstance(meta_field_data, str):
```
ogenstad: We can skip both of the `meta_field_data is not None` checks here as we're already checking the variable with `isinstance`. I.e. if it's a string or a RelationshipManager it will never be `None`.

backend/infrahub/graphql/types/attribute.py
```diff
@@ -73,7 +73,19 @@ class AttributeInterface(InfrahubInterface):
     # owner = Field("DataOwner")
 
 
-class BaseAttribute(ObjectType):
+class InfrahubAttributeMetaObject(ObjectType):
+    updated_by = String(required=False, description="UUID of the user that last modified the attribute or relationship")
```
ogenstad: Small comment here regarding the UUID. I think that Aaron had a hardcoded user with the name `-system-` or similar. Not sure if we want to mention something like that in these descriptions? As it's just a description and not a strict type other systems wouldn't expect there to always be a UUID behind this response.

This is more of an observation, and I'm not sure about the correct approach if we should just leave it as it is now.

backend/infrahub/graphql/types/attribute.py
```diff
@@ -73,7 +73,15 @@ class AttributeInterface(InfrahubInterface):
     # owner = Field("DataOwner")
 
 
-class BaseAttribute(ObjectType):
+class InfrahubAttributeMetaObject(ObjectType):
+    updated_by = String(required=False, description="User that last modified the attribute")
+    updated_at = DateTime(
+        required=False,
```
ogenstad, line 79: These are both set to be `required=False`, I think we have some errors around this in some places in Infrahub. I.e. #7778, if this gets used within the GraphQL response it would mean that both of these are optional and that `null` is a valid response. It could of course be what we would need given that we won't have the information for older data sets. With this in mind having required=False here could make sense. Just highlighting it here.

backend/infrahub/graphql/types/branch.py
```diff
@@ -67,23 +66,31 @@ async def get_by_name(
         raise BranchNotFoundError(f"Branch with name '{name}' not found")
 
 
-class RequiredStringValueField(InfrahubObjectType):
+class InfrahubBranchMetaObject(ObjectType):
+    updated_by = String(required=False, description="User that last modified the attribute or relationship")
+    updated_at = DateTime(
+        required=False,
+        description="Date/Time when the attribute or relationship was last modified by a user or a system task",
```
ogenstad, line 73: Small comment regarding the descriptions here, as this relates to the branch objects themselves they won't have relationships.

backend/infrahub/graphql/types/branch.py
```diff
@@ -67,23 +66,31 @@ async def get_by_name(
...
+    )
+
+
+class RequiredStringValueField(InfrahubBranchMetaObject):
     value = String(required=True)
 
 
-class NonRequiredStringValueField(InfrahubObjectType):
+class NonRequiredStringValueField(InfrahubBranchMetaObject):
```
ogenstad, line 81: This is also used for things such as origin_branch, branched_from and name which would never change. I don't know if we need those fields there or if it matters to have them. It might not hurt to have them just for API consistency though.

backend/tests/unit/graphql/queries/test_proposed_change_actions.py
```diff
@@ -26,6 +26,57 @@
...
+    query {
+        CoreProposedChange {
+            edges {
+                node {
+                    node_metadata {
+                        created_by
+                        created_at
+                        updated_by
+                        updated_at
+                    }
```
ogenstad, line 39: I think we said that we'd want this at the same level as `node` in order to not have "node_metadata", be a restricted name for attributes and or relationships.

backend/tests/unit/graphql/queries/test_proposed_change_actions.py
```diff
@@ -26,6 +26,57 @@
...
+                    description {
+                        value
+                        updated_by
+                        updated_at
+                    }
+                    reviewers {
+                        node_metadata {
```
ogenstad, line 52: It's not clear to me what metadata would be displayed here. It looks like it would be the metadata for the relationship container itself?

ogenstad:
Closing as this PR was continued in #7846 and them merged.

PR #7685, opened 2025-11-20
immutable metadata on relationships

ajtmccarty:
support created/updated_at/by metadata on relationships

the big pieces are
- updates to various relationship CRUD cypher queries  to add setting and getting metadata (as well as consolidating logic so that everything happens in one query instead of several)
- updates to how we load relationships from the database in `NodeManager` and `Relationship` so that metadata can be retrieved and passed on to the correct objects

there's definitely more in here, but they are smaller changes in service of these two larger changes. I added some comments throughout to try to give some context


backend/infrahub/core/constants/__init__.py
```diff
@@ -367,6 +367,7 @@ class MetadataOptions(Flag):
     OWNER = auto()
     LINKED_NODES = SOURCE | OWNER
     IS_PROTECTED = auto()
+    IS_VISIBLE = auto()
```
ajtmccarty, line 370: temporary until the PR to remove IS_VISIBLE everywhere is merged

backend/infrahub/core/query/diff.py
```diff
@@ -77,7 +77,7 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
         AND diff_rel.branch in $branch_names
         AND (
             (diff_rel.from >= $from_time AND diff_rel.from < $to_time)
-            OR (diff_rel.to >= $to_time AND diff_rel.to < $to_time)
+            OR (diff_rel.to >= $from_time AND diff_rel.to < $to_time)
```
ajtmccarty, line 80: this one has been incorrect for along time, but it would only be apparent if a user updated an existing diff by removing only source or owner metadata

backend/infrahub/core/query/node.py
ajtmccarty, line 934: this data structure is returned by the `NodeListGetRelationshipsQuery` below. these changes update it to include timestamp metadata

backend/infrahub/core/query/node.py
```diff
@@ -955,26 +979,82 @@ def get_all_peers(self) -> set[str]:
     def has_node(self, node_id: str) -> bool:
         return node_id in self._rel_names_by_node_id
 
+    def get_metadata_map(
+        self, node_id: str, rel_name: str, direction: RelationshipDirection
+    ) -> dict[MetadataOptions, Timestamp | str | None]:
+        return self._metadata_map.get((node_id, rel_name, direction), {})
+
 
 class NodeListGetRelationshipsQuery(Query):
```
ajtmccarty, line 992: updated to include timestamp metadata
still does not include source/owner metadata, but we actually retrieve that using a different query: `RelationshipGetPeerQuery` in `NodeManager.query_peers`
maybe we will need owner/source metadata in this query at some point, but not right now

backend/infrahub/core/query/relationship.py
```diff
@@ -100,7 +100,10 @@ class RelationshipPeerData:
     rels: list[RelData] | None = None
     """Both relationships pointing at this Relationship Node."""
 
-    updated_at: str | None = None
+    created_at: Timestamp | None = None
+    created_by: str | None = None
+    updated_at: Timestamp | None = None
+    updated_by: str | None = None
```
ajtmccarty, line 107: add timestamp metadata to `RelationshipPeerData`

backend/infrahub/core/query/relationship.py
```diff
@@ -144,7 +147,7 @@ class RelationshipQuery(Query):
     def __init__(
         self,
         rel: type[Relationship] | Relationship | None = None,
-        rel_type: str | None = None,
```
ajtmccarty, line 147: `rel_type` did nothing, it was always `"IS_RELATED"`

backend/infrahub/core/query/relationship.py
```diff
@@ -336,47 +385,71 @@ def query_add_all_node_property_create(self) -> None:
...
 
 
-class RelationshipUpdatePropertyQuery(RelationshipQuery):
+class RelationshipUpdatePropertyQuery(RelationshipWriteQuery):
     name = "relationship_property_update"
     type = QueryType.WRITE
 
     def __init__(
         self,
-        rel_node_id: str,
```
ajtmccarty, line 353: replaced with `rel_id` in `RelationshipQuery`

backend/infrahub/core/query/relationship.py
```diff
@@ -421,172 +492,156 @@ def query_add_all_flag_property_create(self) -> None:
...
+    WHERE $prop_%(var_name)s IS NULL
+    AND $branch_level > 1
+    MATCH (rl)-[r:%(edge_type)s]->(current_prop)
+    WHERE %(branch_filter)s
+    AND r.branch_level = 1
+    AND r.status = "active"
+    CREATE (rl)
+        -[:%(edge_type)s { branch: $branch, branch_level: $branch_level, status: "deleted", from: $at, from_user_id: $user_id }]
+        ->(current_prop)
+}
```
ajtmccarty, line 527: only add the status=deleted edge if we have to, which is if we are making a change on a branch and there is a current property of this type on the default branch

backend/infrahub/core/query/relationship.py
```diff
@@ -421,172 +492,156 @@ def query_add_all_flag_property_create(self) -> None:
...
+        """ % {"edge_type": "HAS_" + name.upper(), "var_name": name, "branch_filter": branch_filter}
         self.add_to_query(query)
 
 
-class RelationshipDataDeleteQuery(RelationshipQuery):
```
ajtmccarty, line 445: this did the same thing as `RelationshipDeleteQuery`, but used different arguments. I've updated `RelationshipDeleteQuery` to work for this queries arguments too

backend/infrahub/core/query/relationship.py
```diff
@@ -738,47 +903,20 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
-            ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
-            LIMIT 1
-        }
-            """ % {
-                "node_prop": node_prop,
-                "node_prop_type": node_prop.upper(),
-                "branch_filter": branch_filter,
-            }
-            self.add_to_query(query)
-            self.update_return_labels([f"rel_{node_prop}", node_prop])
```
ajtmccarty, line 781: moved up above and made adding the metadata parts of the query conditional on the value of `include_metadata`

backend/infrahub/core/query/relationship.py
```diff
@@ -832,88 +981,45 @@ def get_peers(self) -> Generator[RelationshipPeerData, None, None]:
...
-                            name=prop,
-                            prop_db_id=prop_node.element_id,
-                            rel=RelData.from_db(result.get(f"rel_{prop}")),
-                            value=prop_node.get("uuid"),
-                        )
-
-            yield data
-
-
-class RelationshipGetQuery(RelationshipQuery):
```
ajtmccarty, line 864: was only used as part of deleting a relationship. so we would retrieve the relationship and then do some processing in memory and then call a query to delete the relationship. now we just do the whole delete in one query

backend/infrahub/core/relationship/model.py
```diff
@@ -361,23 +400,14 @@ async def update(
...
         update_at = Timestamp(at)
         branch = self.get_branch_based_on_support_type()
 
-        rel_ids_to_update = []
-        for prop_name, prop in data.properties.items():
-            if prop_name in properties_to_update and prop.rel.branch == self.branch.name:
-                rel_ids_to_update.append(prop.rel.db_id)
-
-        if rel_ids_to_update:
-            await update_relationships_to(rel_ids_to_update, to=update_at, db=db)
```
ajtmccarty, line 377: the RelationshipPropertyUpdateQuery handles setting the `to` times now

backend/infrahub/core/relationship/model.py
```diff
@@ -952,8 +995,11 @@ async def get_db_peers(
             db=db,
             source=self.node,
             at=at or self.at,
-            rel=self.rel_class(schema=self.schema, branch=self.branch, node=self.node),
+            rel=self.rel_class(
+                schema=self.schema, branch=self.branch, source_kind=self.node.get_kind(), node=self.node
+            ),
             branch_agnostic=branch_agnostic,
+            include_metadata=MetadataOptions.IS_PROTECTED | MetadataOptions.IS_VISIBLE | MetadataOptions.LINKED_NODES,
         )
```
ajtmccarty, line 1045: default to getting this metadata b/c that is what the query used to return, but it might be possible to default to no metadata and allow overriding it in a parameter

backend/infrahub/core/relationship/model.py
```diff
@@ -85,13 +84,14 @@ class RelationshipUpdateDetails:
...
 
-class Relationship(FlagPropertyMixin, NodePropertyMixin):
+class Relationship(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
     rel_type: str = "IS_RELATED"
 
     def __init__(
         self,
         schema: RelationshipSchema,
         branch: Branch,
+        source_kind: str,
```
ajtmccarty, line 113: we need this to determine whether the source node of relationship is branch-agnostic so that we can look at the correct branch to find it when updating a relationship

backend/infrahub/core/relationship/model.py
```diff
@@ -166,44 +185,104 @@ def get_branch_based_on_support_type(self) -> Branch:
...
+            self.data = data.peer
+        else:
+            self.data = data
+
+        if isinstance(data, RelationshipPeerData):
+            self._process_relationship_peer_data(data=data)
+        elif isinstance(data, dict):
+            self._process_dict_data(data=data)
+        elif isinstance(data, PeerWithRelationshipMetadata):
+            self._process_peer_with_relationship_metadata(data=data)
```
ajtmccarty, line 284: refactored to make more readable after adding support for the new `PeerWithRelationshipMetadata` dataclass. added the dataclass because we needed a way to pass both an instance of a `Node` and metadata about the relationship into `_process_data`

backend/infrahub/core/utils.py
```diff
@@ -62,28 +62,6 @@ async def delete_all_relationships_for_branch(branch_name: str, db: InfrahubData
...
-    MATCH ()-[r]->()
-    WHERE %(id_func)s(r) IN $ids
-    AND r.to IS NULL
-    SET r.to = $to
-    RETURN %(id_func)s(r)
-    """ % {"id_func": db.get_id_function_name()}
-
-    params = {"to": to.to_string(), "ids": [db.to_database_id(_id) for _id in ids]}
-
-    return await db.execute_query(query=query, params=params, name="update_relationships_to")
```
ajtmccarty, line 84: finally vanquished

backend/infrahub/graphql/resolvers/resolver.py
```diff
@@ -186,14 +185,6 @@ async def default_paginated_list_resolver(
...
         objs = []
         if edges or "hfid" in filters:
-            metadata_determiner = MetadataDeterminer(
-                schema_branch=db.schema.get_schema_branch(name=graphql_context.branch.name)
-            )
-            include_metadata = await metadata_determiner.determine_metadata(
-                schema=schema,
-                node_fields=node_fields,
-            )
-
```
ajtmccarty, line 196: this has moved inside of `get_many`

backend/infrahub/graphql/metadata_determiner.py
ajtmccarty, line 1: moved from `graphql/` to `core/metadata` and changed the logic to look at the `fields` dict without requiring the schema branch

backend/tests/unit/graphql/test_mutation_relationship.py
```diff
@@ -735,14 +733,10 @@ async def test_relationship_groups_add_remove(
 
     assert result.errors is None
 
-    nbr_rels_after = await count_relationships(db=db)
-    assert nbr_rels_after - nbr_rels_before == 8
```
ajtmccarty, line 739: this kind of test is bad
a count of relationships at the database level tells us almost nothing about how saving and retrieving the actual data is functioning

backend/infrahub/core/query/node.py
```diff
@@ -1054,29 +1134,63 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
-        RETURN DISTINCT n_uuid, rel_name, peer_uuid, direction
         """ % {"filters": rels_filter}
         self.add_to_query(query)
+
+        self.add_to_query("WITH DISTINCT n_uuid, rel, peer_uuid, direction")
         self.order_by = ["n_uuid", "rel_name", "peer_uuid", "direction"]
         self.return_labels = ["n_uuid", "rel_name", "peer_uuid", "direction"]
 
+        if self.include_metadata & (MetadataOptions.CREATED_AT | MetadataOptions.CREATED_BY):
+            self._add_created_metadata_to_query()
```
ogenstad, line 1153: Potentially the query_init could be easier to read if this if logic happened within `_add_created_metadata_to_query()`

backend/infrahub/core/query/node.py
```diff
@@ -1054,29 +1134,63 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
...
         for result in self.get_results():
             node_id = result.get("n_uuid")
             rel_name = result.get("rel_name")
             peer_id = result.get("peer_uuid")
             direction = str(result.get("direction"))
+            if self.include_metadata & MetadataOptions.CREATED_AT:
+                created_at_str = result.get("created_at")
+                created_at = Timestamp(created_at_str) if created_at_str else None
+            else:
+                created_at = None
```
ogenstad: I think something like this can be a bit easier to read, but there might be situations where it's nice to have the different code branches so that you can see which paths we validate in a code coverage report. This is more of an observation than anything else.


```python
            created_at = None
            if self.include_metadata & MetadataOptions.CREATED_AT:
                created_at_str = result.get("created_at")
                created_at = Timestamp(created_at_str) if created_at_str else None
```

backend/infrahub/core/query/relationship.py
```diff
@@ -832,88 +981,45 @@ def get_peers(self) -> Generator[RelationshipPeerData, None, None]:
...
+                    continue
+                prop_node = result.get(prop)
+                if not prop_node:
+                    continue
+                data.properties[prop] = FlagPropertyData(
+                    name=prop,
+                    prop_db_id=prop_node.element_id,
+                    rel=RelData.from_db(result.get(f"rel_{prop}")),
+                    value=prop_node.get("value"),
+                )
```
ogenstad, line 1013: This looks a lot cleaner!

backend/infrahub/graphql/resolvers/many_relationship.py
```diff
@@ -68,6 +68,24 @@ async def get_peer_count(
...
+        if property_fields and "updated_at" in property_fields:
+            include_metadata |= MetadataOptions.UPDATED_AT
+        if property_fields and "updated_by" in property_fields:
+            include_metadata |= MetadataOptions.UPDATED_BY
+        if property_fields and "source" in property_fields:
+            include_metadata |= MetadataOptions.SOURCE
+        if property_fields and "owner" in property_fields:
+            include_metadata |= MetadataOptions.OWNER
+        if property_fields and "is_protected" in property_fields:
+            include_metadata |= MetadataOptions.IS_PROTECTED
```
ogenstad, line 86: As property_fields is a required dictionary I think we can remove all of the "property_fields and" in the if-statements.

backend/infrahub/graphql/resolvers/single_relationship.py
```diff
@@ -25,6 +25,24 @@ class SingleRelationshipResolver:
...
+            include_metadata |= MetadataOptions.UPDATED_AT
+        if property_fields and "updated_by" in property_fields:
+            include_metadata |= MetadataOptions.UPDATED_BY
+        if property_fields and "source" in property_fields:
+            include_metadata |= MetadataOptions.SOURCE
+        if property_fields and "owner" in property_fields:
+            include_metadata |= MetadataOptions.OWNER
+        if property_fields and "is_protected" in property_fields:
+            include_metadata |= MetadataOptions.IS_PROTECTED
+        return include_metadata
```
ogenstad, line 44: Same comment as above.

backend/tests/unit/graphql/test_mutation_relationship.py
ogenstad: I think we do things like this in a number of places where we don't explain the purpose of the assert. Especially during later refactors when we switch those numbers around..

ogenstad (approved):

backend/infrahub/core/query/node.py
ajtmccarty: good point. made the change in the two queries that use this structure

backend/infrahub/core/query/node.py
ajtmccarty: made the coverage-friendly change
seems about as readable to me

backend/infrahub/graphql/resolvers/many_relationship.py
ajtmccarty: removed unnecessary checks here

backend/infrahub/graphql/resolvers/single_relationship.py
ajtmccarty: and here

PR #7696, opened 2025-11-21
Refactor branch defined triggers

ogenstad:
This PR fixes an issue related to the tasks that are being triggered as part of schema changes related to jinja2 based computed attributes, display labels and HFIDs. The idea was that Infrahub would trigger new tasks to be executed when the schema within a branch is updated, the triggering action was when the prefect task tied to an action within a given branch was updated. A problem around this setup was that branches with the same schema as the default branch didn't contain these tasks until the schema diverged, at that point a Prefect task would be created within a branch and the act of this creation would be a trigger to re-evaluate all actions tied to that specific task. This problem combined with the fact that the Prefect SDK by default limited the automations collected to 200 would result in an invalid view of the tasks that existed in Prefect.

As part of this PR I've also implemented some other minor readability changes.

The main change is that if a task is modified in a branch (created or updated) Infrahub will only trigger a re-computation of that task if the signature of the task differs from the default branch, if it doesn't differ from the default branch data would remain as it was within the main branch. 

In a follow up PR once this issue is resolved I'd like to take a look at how tests around these setups are actually managed and see what we can improve there.

## Summary by CodeRabbit

* **Chores**
  * Optimized trigger and task scheduling to avoid redundant updates when branch configurations match the default branch.
  * Added finer-grained comparison outcomes to drive update decisions and reduce unnecessary workflow submissions.
  * Consolidated automation retrieval to ensure complete aggregation before processing.

* **Tests**
  * Updated tests to use the new automation aggregation approach and reflect refined scheduling behavior.

* **Changelog**
  * Documented reduced redundant task scheduling across branches.

### DIFF
--- a/backend/infrahub/computed_attribute/tasks.py
+++ b/backend/infrahub/computed_attribute/tasks.py
@@ -29,6 +29,7 @@ from .gather import gather_trigger_computed_attribute_jinja2, gather_trigger_com
 from .models import (
     ComputedAttrJinja2GraphQL,
     ComputedAttrJinja2GraphQLResponse,
+    ComputedAttrJinja2TriggerDefinition,
     PythonTransformTarget,
 )
 
@@ -312,21 +313,46 @@ async def computed_attribute_setup_jinja2(
         )  # type: ignore[misc]
         # Configure all ComputedAttrJinja2Trigger in Prefect
 
+        all_triggers = report.triggers_with_type(trigger_type=ComputedAttrJinja2TriggerDefinition)
+
         # Since we can have multiple trigger per NodeKind
-        # we need to extract the list of unique node that should be processed
-        unique_nodes: set[tuple[str, str, str]] = {
-            (trigger.branch, trigger.computed_attribute.kind, trigger.computed_attribute.attribute.name)  # type: ignore[attr-defined]
-            for trigger in report.updated + report.created
-        }
-        for branch, kind, attribute_name in unique_nodes:
-            if event_name != BranchDeletedEvent.event_name and branch == branch_name:
+        # we need to extract the list of unique node that should be processed, this is done by filtering the triggers that targets_self
+        modified_triggers = [
+            trigger
+            for trigger in report.modified_triggers_with_type(trigger_type=ComputedAttrJinja2TriggerDefinition)
+            if trigger.targets_self
+        ]
+
+        for modified_trigger in modified_triggers:
+            if event_name != BranchDeletedEvent.event_name and modified_trigger.branch == branch_name:
+                if branch_name != registry.default_branch:
+                    default_branch_triggers = [
+                        trigger
+                        for trigger in all_triggers
+                        if trigger.branch == registry.default_branch
+                        and trigger.targets_self
+                        and trigger.computed_attribute.kind == modified_trigger.computed_attribute.kind
+                        and trigger.computed_attribute.attribute.name
+                        == modified_trigger.computed_attribute.attribute.name
+                    ]
+                    if (
+                        default_branch_triggers
+                        and len(default_branch_triggers) == 1
+                        and default_branch_triggers[0].template_hash == modified_trigger.template_hash
+                    ):
+                        log.debug(
+                            f"Skipping computed attribute updates for {modified_trigger.computed_attribute.kind}."
+                            f"{modified_trigger.computed_attribute.attribute.name} [{branch_name}], schema is identical to default branch"
+                        )
+                        continue
+
                 await get_workflow().submit_workflow(
                     workflow=TRIGGER_UPDATE_JINJA_COMPUTED_ATTRIBUTES,
                     context=context,
                     parameters={
-                        "branch_name": branch,
-                        "computed_attribute_name": attribute_name,
-                        "computed_attribute_kind": kind,
+                        "branch_name": modified_trigger.branch,
+                        "computed_attribute_name": modified_trigger.computed_attribute.attribute.name,
+                        "computed_attribute_kind": modified_trigger.computed_attribute.kind,
                     },
                 )
 
--- a/backend/infrahub/telemetry/task_manager.py
+++ b/backend/infrahub/telemetry/task_manager.py
@@ -8,6 +8,7 @@ from prefect.client.schemas.objects import WorkerStatus
 from infrahub.events.utils import get_all_events
 from infrahub.trigger.constants import NAME_SEPARATOR
 from infrahub.trigger.models import TriggerType
+from infrahub.trigger.setup import gather_all_automations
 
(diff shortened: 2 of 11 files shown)


backend/infrahub/computed_attribute/tasks.py
```diff
@@ -312,21 +313,46 @@ async def computed_attribute_setup_jinja2(
...
-            for trigger in report.updated + report.created
-        }
-        for branch, kind, attribute_name in unique_nodes:
-            if event_name != BranchDeletedEvent.event_name and branch == branch_name:
+        # we need to extract the list of unique node that should be processed, this is done by filtering the triggers that targets_self
+        modified_triggers = [
+            trigger
+            for trigger in report.modified_triggers_with_type(trigger_type=ComputedAttrJinja2TriggerDefinition)
+            if trigger.targets_self
+        ]
```
ogenstad, line 324: Previously we created a set to exclude duplicates of computed attributes that spanned multiple node kinds, with this update where we instead check for triggers that target the actual kind we can remove duplicates in a way that instead keep the correct object type of the reports.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -312,21 +313,46 @@ async def computed_attribute_setup_jinja2(
...
+                        and trigger.targets_self
+                        and trigger.computed_attribute.kind == modified_trigger.computed_attribute.kind
+                        and trigger.computed_attribute.attribute.name
+                        == modified_trigger.computed_attribute.attribute.name
+                    ]
+                    if (
+                        default_branch_triggers
+                        and len(default_branch_triggers) == 1
+                        and default_branch_triggers[0].template_hash == modified_trigger.template_hash
+                    ):
```
ogenstad, line 342: This pattern is the same for display_labels and HFIDs, though the fields are a bit different. Instead of always triggering tasks if the trigger has been modified (created or updated, not refreshed) we first check if we're on a non default branch and if so try to find the same trigger on the default branch and if we find it and the hash of the trigger is the same we don't schedule a recalculation to happen.

backend/infrahub/telemetry/task_manager.py
```diff
@@ -53,7 +54,7 @@ async def count_events(event_name: str) -> int:
 
 @task(name="telemetry-gather-automations", task_run_name="Gather Automations", cache_policy=NONE)
 async def gather_prefect_automations(client: PrefectClient) -> dict[str, Any]:
-    automations = await client.read_automations()
+    automations = await gather_all_automations(client=client)
```
ogenstad, line 57: These updates are to ensure that we can find more than the default return value for automations in the Prefect API (200)

backend/infrahub/trigger/models.py
```diff
@@ -18,16 +18,78 @@
...
+        Returns:
+            List of triggers of the specified type from all categories
+        """
+        created = self._created_triggers_with_type(trigger_type=trigger_type)
+        updated = self._updated_triggers_with_type(trigger_type=trigger_type)
+        refreshed = self._refreshed_triggers_with_type(trigger_type=trigger_type)
+        unchanged = self._unchanged_triggers_with_type(trigger_type=trigger_type)
+        return created + updated + refreshed + unchanged
+
+    def modified_triggers_with_type(self, trigger_type: type[T]) -> list[T]:
```
ogenstad, line 81: Previously we had some issues where the `TriggerSetupReport` contains entries of type `TriggerDefinition` but what we're typically interested in within the setup are specific child classes of TriggerDefinition. What we did was to either use `cast` or ignore the type, with this change we can get the typing correct.

backend/infrahub/trigger/setup.py
```diff
@@ -145,15 +158,34 @@ async def setup_triggers(
...
-        )
-    else:
-        log.info(
-            f"Processed all triggers: "
-            f"{len(report.created)} created, {len(report.updated)} updated, {len(report.unchanged)} unchanged, {len(report.deleted)} deleted"
-        )
+    log.info(
+        f"Processed {trigger_log_message}: {len(report.created)} created, {len(report.updated)} updated, "
+        f"{len(report.refreshed)} refreshed, {len(report.unchanged)} unchanged, {len(report.deleted)} deleted"
+    )
```
ogenstad, line 162: Aside from the changes to the input to compare_automations and its updated return values the changes in this function is mostly for readability, in order to avoid if-statements just for logging and to gather existing automations once and only filter them if required.

backend/infrahub/trigger/setup.py
```diff
@@ -145,15 +158,34 @@ async def setup_triggers(
...
+
+
+async def gather_all_automations(client: PrefectClient) -> list[Automation]:
+    """Gather all automations from the Prefect server
+
+    By default the Prefect client only retrieves a limited number of automations, this function
+    retrieves them all by paginating through the results. The default within Prefect is 200 items,
+    and client.read_automations() doesn't support pagination parameters.
+    """
+    automation_count_response = await client.request("POST", "/automations/count")
```
dgarros, line 174: I believe we can skip this call
We should be able to identify if we have reach the end of the list by counting the number of results returned by the query
If the limit is set at 200 and we get 200 then most likely we are missing some results

backend/infrahub/trigger/setup.py
```diff
@@ -145,15 +158,34 @@ async def setup_triggers(
...
+    automation_count_response.raise_for_status()
+    automation_count: int = automation_count_response.json()
+    offset = 0
+    limit = 200
```
fatih-acar, line 178: Note for later: I don't know if we have other places that would be impacted by Prefect's internal pagination limit, but should we have a wrapper on top of Prefect's client to automatically paginate such queries?

backend/infrahub/trigger/setup.py
```diff
@@ -12,22 +12,38 @@
...
 
+    if force_update:
+        return TriggerComparison.UPDATE
+
     target_dump = target.model_dump(exclude_defaults=True, exclude_none=True)
     existing_dump = existing.model_dump(exclude_defaults=True, exclude_none=True, exclude={"id"})
 
-    return target_dump == existing_dump
+    if not trigger_type or not trigger_type.is_branch_specific:
+        if target_dump == existing_dump:
```
fatih-acar: I think this is redundant with line 39-40, if you move the statement above line 34.

ogenstad: Ah yes good point. I've moved the code.

dgarros (approved):
My comment is not a blocker since this is an important fix but it would be good to revisit soon

backend/infrahub/trigger/setup.py
ogenstad: Fair enough. I'll add a note around this and update later.

PR #7698, opened 2025-11-22
Template profile support

minitriga:
READY FOR REVIEW. 

## Summary by CodeRabbit

* **New Features**
  * Assign profiles to templates; objects created from templates inherit profiles (badges, pre-filled values), including component support.

* **Behavior Changes**
  * Template values take precedence over profile values; multiple profiles merge by priority; source metadata for profile/template origins is tracked and preserved.

* **Frontend UX**
  * Forms, autocomplete, and edit flows show profile/template-aware data and profile visibility for templates.

* **Tests**
  * Expanded unit, integration, and e2e coverage for template↔profile scenarios.

* **Documentation**
  * New and updated guides and topics describing profiles-with-templates usage.


backend/infrahub/graphql/resolvers/many_relationship.py
```diff
@@ -209,6 +209,8 @@ async def _get_entities_simple(
                 branch=branch,
                 branch_agnostic=rel_schema.branch is BranchSupportType.AGNOSTIC,
                 fetch_peers=True,
+                include_source=True,
```
fatih-acar, line 212: @opsmill/backend isn't there a performance impact? same for line 251-252

bilalabbad (commented):
I tested the features through the UI and found that while the backend behavior appears correct, the form does not accurately reflect the resolved field values when both a template and a profile are involved:

1. Create a **profile** for a patch panel  
   - Description: `"desc for profile"`  
   - Module capacity: **100**

2. Create a **template** for a patch panel using that profile, and **override** the profile value  
   - Module capacity: **200**

3. Create a new **patch panel** using that template  
   - **Issue:** The form displays Module capacity as **100** (from the profile) instead of the overridden **200** (from the template).

E2E test should be updated with this usecase.

backend/infrahub/core/node/create.py
```diff
@@ -136,7 +164,7 @@ async def handle_template_relationships(
             )
 
 
-async def get_profile_ids(db: InfrahubDatabase, obj: Node) -> set[str]:
+async def get_profile_ids(db: InfrahubDatabase, obj: Node | CoreObjectTemplate) -> set[str]:
```
gmazoyer, line 167: A `CoreObjectTemplate` is somewhat a `Node` so I'm not sure the typing need to be changed here. But that's minor comment.

gmazoyer (approved):

backend/infrahub/core/node/create.py
minitriga: I have reverted this.

minitriga: Its actually required. Because a profile can be attached to a template we need to be able to find the profile associated with a template: https://github.com/opsmill/infrahub/actions/runs/19658208448/job/56299328493?pr=7698

models/base/dcim.yml
```diff
@@ -176,6 +176,7 @@ nodes:
     order_by:
       - name__value
     display_label: "{{ name__value }}"
+    generate_template: true
```
bilalabbad, line 179: any reason to add that without creating a template with it?

bilalabbad (approved):

models/base/dcim.yml
minitriga: We create the template in the tests to check everything is working.

minitriga: We create the template in the tests to check everything is working.

Issue #7568, opened 2025-11-05
bug: cannot open task list view with too much history


fatih-acar:
### Component

API Server / GraphQL

### Infrahub version

1.4.12

### Current Behavior

When trying to open the [task list views](https://sandbox.infrahub.app/tasks), users may face errors (page won't show) or slowness when the task history is long (1M+ tasks).

### Expected Behavior

We should be able to show the task log regardless of the actual retention.

### Steps to Reproduce

Load dataset with 1M+ task history.
Open task list view

### Additional Information

First analysis shows that the default database timeout value (10 seconds) may not be enough to handle such workloads (see https://docs.prefect.io/v3/advanced/self-hosted#handling-large-databases).

Further analysis shows that the queries to the Prefect API we do are including default filters (as seen here: https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/task_manager/task.py#L183) which always trigger a SQL full table scan (since the index cardinality over the filtered tag is high) when doing count operations on task history or event logs. This does not apply when fetching rows with pagination (works fine but performance decreases when reaching oldest entries - few ms for the first entries, >1 second for the last entries in 1M rows).

Thus even if we do pass (offset: 0, limit: 10) as default parameters to the query, the Prefect backend and SQL database will always do a full scan of the table (and thus process all the entries in the task logs).

Few workarounds are available:
- increase the default database timeout (10s -> 60s or 120s?)
- apply task history retention policy (clean task logs older than 90 days?)
- investigate and fix Prefect backend issues regarding our queries (either optimize on the Infrahub side and/or Prefect side)

fatih-acar:
Even though we merged #7699, we still need to run more tests to confirm this is fixed.

PR #7699, opened 2025-11-23
fix(backend): IFC-2040 cache task count results

fatih-acar:
When the task history is large, the count query can take a few seconds to complete due to how the Prefect database is handling the query (it will scan each row in the table).
We can cache the count result when the history is larger than a statically defined 100k entries.

Also update the default db timeout value for Prefect since the count query can sometimes take more time than the default 10 seconds value.

Related to https://github.com/opsmill/infrahub/issues/7568

## Summary by CodeRabbit

* **New Features**
  * Implemented flow run count caching with configurable threshold (default: 100,000)
  * Cache entries expire after one minute

* **Configuration**
  * New setting: INFRAHUB_WORKFLOW_FLOW_RUN_COUNT_CACHE_THRESHOLD to control caching
  * New environment variable: PREFECT_API_DATABASE_TIMEOUT=60 for backend DB operations

* **Documentation**
  * Updated configuration reference with the new cache threshold

* **Changelog**
  * Added entry noting task count caching to mitigate long-history performance issues

### DIFF
--- a/backend/infrahub/task_manager/task.py
+++ b/backend/infrahub/task_manager/task.py
@@ -1,5 +1,6 @@
 import asyncio
-import uuid
+import hashlib
+import json
 from datetime import datetime, timedelta, timezone
 from typing import Any
 from uuid import UUID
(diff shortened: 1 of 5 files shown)


backend/infrahub/task_manager/task.py
```diff
@@ -59,10 +69,24 @@ async def count_flow_runs(
...
         }
+        cache_key = cls._build_flow_run_count_cache_key(body)
+
+        cache = await get_cache()
+        cached_value_raw = await cache.get(key=cache_key)
+        if cached_value_raw is not None:
+            try:
+                return int(cached_value_raw)
+            except (TypeError, ValueError):
+                await cache.delete(key=cache_key)
```
ogenstad, line 80: Not sure if it's currently useful elsewhere but this could potentially at some point move into a `cache.get_as_int()`.

ogenstad (approved):
Could be nice with a news-fragment as well.

backend/infrahub/task_manager/task.py
fatih-acar: It seems we do not convert to int anywhere else so I'll let it as-is for the moment...

PR #7701, opened 2025-11-24
IFC-1948: Update StandardNode branch to save metadata fields

solababs:


backend/infrahub/core/branch/models.py
```diff
@@ -37,6 +37,9 @@ class Branch(StandardNode):
     branched_from: Optional[str] = Field(default=None, validate_default=True)
     hierarchy_level: int = 2
     created_at: Optional[str] = Field(default=None, validate_default=True)
+    created_by: Optional[str] = Field(default=None, validate_default=True)
+    updated_at: Optional[str] = Field(default=None, validate_default=True)
+    updated_by: Optional[str] = Field(default=None, validate_default=True)
```
ajtmccarty: I think these fields should be at the `StandardNode` level, b/c we are going to want to be able to support this data on every `StandardNode`

the CRUD methods on the `StandardNode` should also be updated to accept a `user_id: str = SYSTEM_USER_ID` parameter (defined in [my immutable metadata PR here](https://github.com/opsmill/infrahub/blob/ajtm-immutable-metadata/backend/infrahub/core/constants/__init__.py#L49)) that is passed onto any cypher queries that those methods call

backend/infrahub/graphql/mutations/models.py
```diff
@@ -9,3 +9,8 @@ class BranchCreateModel(BaseModel):
     branched_from: str | None = None
     sync_with_git: bool = True
     is_isolated: bool = True
+
+
+class BranchCreateModelWithMetaFields(BranchCreateModel):
+    created_at: str | None = None
+    created_by: str | None = None
```
ajtmccarty: I don't think you need `created_at` here. the timestamp of an action should be handled in the backend where the object is actually created instead of at the graphql level

I also don't know if you need this new class at all. it might clearer to add a `user_id` parameter to the `create_branch` flow. but if you keep `created_by` in this model, then it should be mandatory b/c we should always know what user is creating a branch

backend/infrahub/core/branch/models.py
```diff
@@ -553,4 +560,344 @@ async def rebase_graph(self, db: InfrahubDatabase, at: Optional[Timestamp] = Non
...
+class DescriptionValueField(FieldMetadata):
+    value: str = ""
+
+
+class StatusValueField(FieldMetadata):
+    value: BranchStatus = BranchStatus.OPEN
+
+
+class OriginBranchValueField(FieldMetadata):
+    value: str = "main"
```
ogenstad, line 583: In 99% of the time this would default to "main", but it is still possible to setup Infrahub to use another name for the default branch during first initialization, as such this should probably be a Field with a default_factory which would return `registry.default_branch` when called (not sure if that might lead to some circular imports).

ogenstad:
Closing as this work was completed in #7859.

PR #7712, opened 2025-11-25
Fix `display_label` not being stored in the db

### LINKED ISSUE #7704: bug: Display label trigger not setup for existing nodes
### Component

API Server / GraphQL

### Infrahub version

1.5.2

### Current Behavior

Display label for some of the node is never updated:

Image

Even tho my schema is correct:

Image

### Expected Behavior

Display label should be adjusted to match with what's in my schema

### Steps to Reproduce

- Load this schema

```yml
---
version: "1.0"

nodes:
  - name: Stuff
    namespace: Random
    description: Random stuff
    attributes:
      - name: name
        kind: Text
        computed_attribute:
          kind: Jinja2
          jinja2_template: "RANDOM_{{ index__value }}"
        read_only: true
        optional: false
        unique: true
      - name: index
        kind: Number

```

- Create a couple of records directly in infrahub
- Then load this schema (that introduces the display label)

```yml
---
version: "1.0"

nodes:
  - name: Stuff
    namespace: Random
    description: Random stuff
    display_label: "{{ name__value}}"
    attributes:
      - name: name
        kind: Text
        computed_attribute:
          kind: Jinja2
          jinja2_template: "RANDOM_{{ index__value }}"
        read_only: true
        optional: false
        unique: true
      - name: index
        kind: Number

```

- Now getting back to the UI 
=> Existing records won't ever have a proper display label, even when editing it and so on
=> When creating a new record display label is correct

### Additional Information

I understand it as "display label isn't setup for existing records". Please make sure that this is not impacting HFID.

gmazoyer:
Fixes #7704 

When the `display_label` was not set for a schema node, a null value was not stored as display label in the database. This issue prevented the display label from being updated after a schema change.

This fix ensures that a `display_label` vertex is always created in the database even for nodes that do not have a display label at the schema level.

## Summary by CodeRabbit

* **New Features**
  * Adds Migration 047 to backfill and initialize missing display_label values with branch-scoped, batched execution and progress reporting.

* **Improvements**
  * Graph version incremented.
  * display_label and human_friendly_id are now created/computed when absent; behavior is branch-aware and idempotent.

* **Bug Fixes**
  * Consistent NULL handling for display_label in storage and diffs; diffs accept explicit NULL markers.

* **Tests**
  * New and updated tests covering migration, idempotency, branching, diffs, and default display-label behavior.

* **Changelog**
  * Entry added documenting the NULL-handling fix.


backend/tests/integration/diff/test_diff_update.py
```diff
@@ -319,7 +319,7 @@ async def test_diff_deleted_ed_209(
             }
             for prop_diff in properties_by_type.values():
                 assert prop_diff.action is DiffAction.REMOVED
-                assert prop_diff.new_value is None
+                assert prop_diff.new_value in [None, NULL_VALUE]
```
gmazoyer, line 322: Seems that if the value is already `"NULL"` we do not use `None`, from https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/core/diff/query_parser.py#L140-L146

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+
+    seen: set[str] = set()
+    result: list[str] = []
+    for match in matches:
+        var_name = match.group(1)
+        if var_name not in seen:
+            seen.add(var_name)
+            result.append(var_name)
+
+    return result
```
ogenstad: I think this would be problematic if Jinja2 filters were used such as:

```python
extract_jinja2_variables_in_order("{{ title|upper }} - {{ description }}")
```

Do we have to have them in order? I.e. we can't use the Jinja2Template class from the SDK? We haven't cared about that before, and perhaps it's just not relevant in other contexts..

gmazoyer: This will probably fail indeed. That said, I think that I have to do some more testing with `Jinja2Template`. Last time I checked the order of variables were not guaranteed. Though, I should write a proper test for this to see, if the order is "random" or just reversed (as I only tested with 2 vars).

Also this echos to another issue in the migration. We do not backfill the display label by rendering the Jinja2 template, we just join the values using spaces. This should be addressed too.

ogenstad: No the variables are returned as a sorted list, and they come back from Jinja2 as a `set`. I think the previous migrations were probably fine when it comes to that as it wasn't possible to have Jinja2 filters prior to 1.4. So any of the older schemas would have just been "attribute1" + " " "attribute2".

backend/infrahub/core/node/__init__.py
```diff
@@ -179,19 +179,16 @@ async def add_human_friendly_id(self, db: InfrahubDatabase) -> None:
         await self._human_friendly_id.compute(db=db, node=self)
 
     async def get_display_label(self, db: InfrahubDatabase) -> str:
-        if self._display_label:
-            if isinstance(self._display_label._value, str):
-                return self._display_label._value
-            if self._display_label._value:
-                return self._display_label._value.value
+        if self._display_label and (value := self._display_label.get_value(node=self, at=self._at)):
```
ajtmccarty, line 182: I know it's not part of this PR, but passing `self` into a method of `self` is bad

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+AND e.to IS NULL
+AND NOT exists((n)-[:IS_PART_OF {branch: e.branch, status: "deleted"}]->(:Root))
+WITH DISTINCT n, e AS is_part_of_e
+OPTIONAL MATCH (n)-[r:HAS_ATTRIBUTE]->(attr:Attribute {name: $attribute_name})
+WHERE r.branch IN $branch_names
+AND r.status = "active"
+AND r.to IS NULL
+WITH n, is_part_of_e, r AS has_attr_e
+WHERE is_part_of_e.status = "active" AND (has_attr_e IS NULL OR has_attr_e.status = "deleted")
+WITH n.uuid AS node_uuid
```
ajtmccarty: - the `is_part_of_e` has already been filtered to only `status="active"` edges, so that can be removed
- `has_attr_e` will never be `status="deleted"` b/c it is filtered to `status="active"` in the first `WHERE` clause in the snippet below. I think you want to get the latest edge linking `n` to `:Attribute {name: $attribute_name}` and filter to only those rows where the latest `has_attr_e` is NULL or is deleted

I have not tested this, but I think this will do it

```suggestion
WITH DISTINCT n
CALL (n) {
    OPTIONAL MATCH (n)-[r:HAS_ATTRIBUTE]->(attr:Attribute {name: $attribute_name})
    WHERE r.branch IN $branch_names
    RETURN r AS has_attr_e
    ORDER BY r.from DESC, r.status ASC
    LIMIT 1
}
WITH n, has_attr_e
WHERE (has_attr_e IS NULL OR has_attr_e.status = "deleted")
WITH n.uuid AS node_uuid
```

also, not sure how many UUIDs this might be, but it could be necessary to paginate this if it could be more than roughly 10k

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+        # Count nodes that will get computed values
+        kinds_to_backfill: list[str] = []
+        for node_schema_name in (
+            main_schema_branch.node_names + main_schema_branch.profile_names + main_schema_branch.template_names
+        ):
+            if node_schema_name in self.kinds_to_skip:
+                continue
+
+            node_schema = main_schema_branch.get(name=node_schema_name, duplicate=False)
+            if node_schema.branch != BranchSupportType.AWARE or not node_schema.display_label:
```
ajtmccarty: I suspect that I am missing something, but I thought that we wanted to backfill the data for schemas where `node_schema.display_label` is falsy and this looks like it skips the display_label=None/[] schemas

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+            "is_visible_default": True,
+        }
+        branch_filter, branch_filter_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_filter_params)
+
+        # Create the NULL AttributeValue and Boolean values first
+        create_av_query = """
+MERGE (av:AttributeValue&AttributeValueIndexed {is_default: false, value: $null_value})
+MERGE (is_protected_value:Boolean { value: $is_protected_default })
+MERGE (is_visible_value:Boolean { value: $is_visible_default })
```
ajtmccarty, line 319: should add `LIMIT 1` here
there _could_ be multiple of these `(av:AttributeValue&AttributeValueIndexed {is_default: false, value: $null_value})`
unlikely, but possible

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+            "branch_support": "aware",
+            "is_protected_default": False,
+            "is_visible_default": True,
+        }
+        branch_filter, branch_filter_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_filter_params)
+
+        # Create the NULL AttributeValue and Boolean values first
+        create_av_query = """
+MERGE (av:AttributeValue&AttributeValueIndexed {is_default: false, value: $null_value})
```
ajtmccarty: should this be `is_default: true`?

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+// ------------
+// Create the display_label attribute with NULL value for nodes
+// ------------
+WITH av, is_protected_value, is_visible_value
+MATCH (n:Node)-[e:IS_PART_OF]->(:Root)
+WHERE n.uuid IN $node_uuids
+AND e.branch IN [$default_branch, $global_branch]
+AND e.to IS NULL
+AND e.status = "active"
+CREATE (a:Attribute { name: $attribute_name, branch_support: $branch_support })
```
ajtmccarty: Attribute vertices need a uuid property too.
I think `uuid: randomUUID()` will do it

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+CALL (n) {
+    MATCH (n)-[r:IS_PART_OF]->(:Root)
+    WHERE %(branch_filter)s
+    RETURN r.status = "active" AS is_active
+    ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+    LIMIT 1
+}
+WITH n, is_active, av, is_protected_value, is_visible_value
+WHERE is_active = TRUE
+CREATE (a:Attribute { name: $attribute_name, branch_support: $branch_support })
```
ajtmccarty: need uuid here too

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+            "default_branch": registry.default_branch,
+            "global_branch": GLOBAL_BRANCH_NAME,
+            "branch": GLOBAL_BRANCH_NAME if self.is_branch_agnostic else self.branch.name,
+            "branch_level": 1 if self.is_branch_agnostic else self.branch.hierarchy_level,
+            "at": self.at.to_string(),
+        }
+        branch_filter, branch_filter_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_filter_params)
+
+        if self.is_large_type_attribute:
```
ajtmccarty: I think you can delete these `if self.is_large_type_attribute` conditionals b/c `display_label` is a string, which is a regular, indexed attribute

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+                            attribute_schema=display_label_attribute_schema,
+                            progress=progress,
+                            update_task=backfill_task,
+                        )
+
+        except Exception as exc:
+            return MigrationResult(errors=[str(exc)])
+        return MigrationResult()
+
+    async def _do_one_schema_branch(
```
ajtmccarty: this method appears to be unused

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,645 @@
...
+            offset += self.update_batch_size
+
+    async def execute_against_branch(self, db: InfrahubDatabase, branch: Branch) -> MigrationResult:
+        schema_branch = await get_or_load_schema_branch(db=db, branch=branch)
+
+        base_node_schema = schema_branch.get("SchemaNode", duplicate=False)
+        display_label_attribute_schema = base_node_schema.get_attribute("display_label")
+
+        try:
+            get_nodes_without_dl_query = await GetNodesWithoutDisplayLabelQuery.init(
```
ajtmccarty: this query only has logic to run against the default/global branches. I don't think it will do anything when run on a user-created branch in its current state

backend/tests/unit/core/migrations/graph/test_047.py
```diff
@@ -0,0 +1,175 @@
...
+        person_jane_branch.height.value = 165
+        await person_jane_branch.save(db=db)
+
+        initial_values = await self.get_attribute_values_from_db(
+            db=db, branch=test_branch, node_ids=[person_john_branch.id, person_jane_branch.id]
+        )
+        assert initial_values[person_john_branch.id] is None
+        assert initial_values[person_jane_branch.id] is None
+
+        async with db.start_session() as dbs:
```
ajtmccarty, line 160: with migration's current code, I believe that this test would fail if you ran the migration on `main`, rebased branch `"test-branch-m047"`, then tried to run the migration against the branch

gmazoyer: I added some tests to cover this path:
* `test_migration_047_execute_against_branch_after_main`
* `test_migration_047_execute_against_branch_with_name_change`

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,624 @@
...
+        update_task: TaskID | None = None,
+    ) -> None:
+        if not schema.display_label:
+            return
+
+        schema_paths = self._extract_schema_paths_from_display_label(schema=schema, schema_branch=schema_branch)
+        if not schema_paths:
+            return
+
+        variable_names = _extract_jinja2_variables_in_order(schema.display_label)
```
ajtmccarty: this looks like it is doing almost the same thing as line 418
can you just get these from `schema_paths`?

backend/infrahub/core/migrations/graph/m047_backfill_or_null_display_label.py
```diff
@@ -0,0 +1,624 @@
...
+            "default_branch": registry.default_branch,
+            "global_branch": GLOBAL_BRANCH_NAME,
+            "branch": GLOBAL_BRANCH_NAME if self.is_branch_agnostic else self.branch.name,
+            "branch_level": 1 if self.is_branch_agnostic else self.branch.hierarchy_level,
+            "at": self.at.to_string(),
+        }
+        branch_filter, branch_filter_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_filter_params)
+
+        self.add_to_query("WITH [] AS value_id_pairs")
```
ajtmccarty: I think you can remove all references to `value_id_pairs` in this one

ajtmccarty (approved):
looks good to me
nice work on the cypher queries

PR #7718, opened 2025-11-25
fix: allow List attributes to use regex parameter

### LINKED ISSUE #7717: bug: unable to load schema with an attribute of kind list with regex parameter
### Component

API Server / GraphQL

### Infrahub version

1.5.3

### Current Behavior

You are not able to load a schema that defines an attribute of kind List, with a regular expression constraint.

For example:
```yaml
---
version: "1.0"
nodes:
  - name: Node
    namespace: Testing
    attributes:
      - name: protocols
        kind: List
        optional: true
        parameters:
          regex: "ssh|ping|telnet"
```

You get the following error message:
```
Unable to load the schema:
  Node: TestingNode | attributes ({'name': 'protocols', 'kind': 'List', 'optional': True, 'parameters': {'regex': 'ssh|ping|telnet'}}) | Value error, TextAttributeParameters 
can't be used as parameters for List (value_error)
```

When you use the old syntax, the schema loads, and you can create new objects. But then you run into issue #7654

### Expected Behavior

You can load a schema that defines a regex constraint on an attribute of kind List, using the parameters.regex syntax.

### Steps to Reproduce

Spin up instance of Infrahub
Try loading this schema
```yaml
---
version: "1.0"
nodes:
  - name: Node
    namespace: Testing
    attributes:
      - name: name
        kind: Text
        optional: false
      - name: protocols
        kind: List
        optional: true
        parameters:
          regex: "ssh|ping|telnet"
```

### Additional Information

_No response_

wvandeun:
Add support for List attributes to accept regex validation via parameters.regex syntax. List attributes can now use TextAttributeParameters which includes regex support.

Changes:
- Added List to parameter class mapping
- Created ListAttributeSchema with regex reconciliation
- Updated validation to allow TextAttributeParameters for List
- Added tests for List with regex parameter

Fixes #7717

## Summary by CodeRabbit

* **New Features**
  * List attributes now support regex validation for individual list items across the platform.

* **Tests**
  * Added coverage to validate list regex behavior and ensure invalid list items are rejected.

* **Schema / API**
  * Public schemas and API types updated to include the new list-parameter shape so UIs and integrations can use the new validation option.

### DIFF
--- a/backend/tests/component/core/schema/test_attribute_parameters.py
+++ b/backend/tests/component/core/schema/test_attribute_parameters.py
@@ -12,7 +12,11 @@ from infrahub.core.node.resource_manager.number_pool import CoreNumberPool
 from infrahub.core.registry import registry
 from infrahub.core.schema import GenericSchema, NodeSchema, SchemaRoot
 from infrahub.core.schema.attribute_parameters import (
+<<<<<<< HEAD:backend/tests/component/core/schema/test_attribute_parameters.py
     AttributeParameters,
+=======
+    ListAttributeParameters,
+>>>>>>> 13a243097 (Use ListAttributeParameters for List):backend/tests/unit/core/schema/test_attribute_parameters.py
     NumberAttributeParameters,
     NumberPoolParameters,
     TextAttributeParameters,
@@ -396,7 +400,7 @@ def test_list_attribute_with_regex_parameter() -> None:
 
     node = NodeSchema(**node_schema)
     protocols_attribute = node.get_attribute("protocols")
-    assert isinstance(protocols_attribute.parameters, TextAttributeParameters)
+    assert isinstance(protocols_attribute.parameters, ListAttributeParameters)
     assert protocols_attribute.parameters.regex == "ssh|ping|telnet"
     assert protocols_attribute.get_regex() == "ssh|ping|telnet"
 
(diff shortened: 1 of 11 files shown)


backend/tests/component/core/schema/test_attribute_parameters.py
```diff
@@ -260,3 +260,48 @@ def test_validate_min_max_text_attribute() -> None:
         TextAttributeParameters(min_length=10, max_length=5)
 
     assert config.SETTINGS.main.schema_strict_mode
+
+
+def test_list_attribute_with_regex_parameter() -> None:
```
dgarros, line 384: I would be good to add a test to ensure that the regex is properly applied to the values in the list

dgarros (approved):
@wvandeun sorry I highjacked your PR to do some tests around Claude Code. in the process I think got it in a state where it can be merged

ajtmccarty (commented):
I think this needs a graphql test

dgarros:
> I think this needs a graphql test

@ajtmccarty Please can you elaborate on this one, this PR only touches the schema and it shouldn't have any impact on GraphQL

ajtmccarty:
> > I think this needs a graphql test
> 
> @ajtmccarty Please can you elaborate on this one, this PR only touches the schema and it shouldn't have any impact on GraphQL

I suppose this is okay as is since it's not really changing how the regex parameters are evaluated

ajtmccarty (approved):

ajtmccarty:
we started reconciling the old-style top-level parameters (AttributeSchema.regex/min_length/max_length) with the new parameter version (AttributeSchema.parameters.regex/min_length/max_length) in a new way in #7998 , so I merged `release-1.7` into this branch to pull in those changes and updated the PR to use the new logic. 

@dgarros you should review it again before merging.

wvandeun:
@dgarros are we good to merge this one?

dgarros:
yes LGTM

PR #7731, opened 2025-11-26
IFC-1896 Don't query fields to compute display label/HFID 

gmazoyer:
After the HFID and display label refactoring work, we started storing the values for these properties as attributes in the database. We therefore do not require any logic to try to compute them on the fly and so we do not need to find out which fields to query for alongside those that the user requested.

This change fixes cases where HFID and display label may not be retrieved from the database when trying to access them for relationships via GraphQL queries.

There are workarounds in this PR to make IPAM and permissions responses correct. For IPAM, the issue is that we do not store some of the "context" nodes in the database, so we therefore have to query for fields to reconstruct their display labels (they do not have HFIDs). For permissions, the `identifier` field is computed on the fly too so we need to query for the fields to reconstruct it (which are the same as the one used in the display label). Ideally, we'd need to move the permission `identifier` field to a read-only computed attribute and maybe move the display label to be Jinja2 computed too (could be we'll end up with identifier == display label).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GraphQL retrieval of human-friendly ID and display label for relationship nodes so stored values are returned correctly.
  * Ensured display labels for available IP prefix/range nodes are computed and returned properly in GraphQL listings.

### DIFF
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -3,9 +3,9 @@ from __future__ import annotations
 from copy import copy
 from typing import TYPE_CHECKING, Any, Iterable, Literal, TypeVar, overload
 
-from infrahub_sdk.utils import is_valid_uuid
+from infrahub_sdk.utils import deep_merge_dict, is_valid_uuid
 
-from infrahub.core.constants import RelationshipCardinality, RelationshipDirection
+from infrahub.core.constants import InfrahubKind, RelationshipCardinality, RelationshipDirection
 from infrahub.core.node import Node
 from infrahub.core.node.delete_validator import NodeDeleteValidator
 from infrahub.core.query.node import (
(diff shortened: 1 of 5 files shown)


backend/infrahub/core/manager.py
```diff
@@ -188,18 +188,23 @@ async def query(
...
+            and "identifier" in fields
+            and node_schema.kind
+            in [
+                InfrahubKind.BASEPERMISSION,
+                InfrahubKind.GLOBALPERMISSION,
+                InfrahubKind.OBJECTPERMISSION,
+            ]
+        ):
+            # This is a workaround to ensure we are querying the right fields for permissions
+            # The identifier for permissions needs the same fields as the display label
```
ogenstad, line 202: Did we not need to have this fallback any more? I guess all old data should be populated with the migrations? I.e. that we're removing the "if fields and 'display_label' in fields"?

ogenstad (approved):

PR #7746, opened 2025-11-28
Python 3.13

ogenstad:
## Summary by CodeRabbit

* **Infrastructure**
  * Bumped Python baseline to 3.13.x and parameterized CI/workflows to use a central Python version.

* **API Changes**
  * Broadened OpenAPI response schemas to allow additional properties and more flexible JSON payloads.

* **Dependencies**
  * Updated and pinned multiple libraries and packaging metadata to support newer Python versions.

* **Tests**
  * Modernized async test fixtures and moved tests to use a scoped database provider for isolation.

* **Frontend**
  * Widened generated API types to accept generic key-value JSON payloads.


backend/infrahub/core/diff/model/diff.py
```diff
@@ -26,7 +26,7 @@ def to_graphql(self) -> dict[str, Any]:
         """
         resp: dict[str, Any] = {}
         for key, value in self:
-            field_info = self.model_fields[key]
+            field_info = type(self).model_fields[key]
```
ogenstad: I wonder if `field_info = self.__class__.model_fields[key]` would be cleaner here. Have some vague memory that we're using that somewhere else. Assuming that the self call to model_fields got deprecated?

fatih-acar: I think both are equivalent. The documentation just states that model_fields should be accessed on the class directly and not the instance. so __class__ would also work?

We can switch to __class__ if you want?

.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.12-bookworm
+FROM python:3.14-bookworm
```
gmazoyer: Should this be `python:3.14-trixie` now?

fatih-acar: Good catch thanks! I'll just switch to `3.14` to be consistent with `development/Dockerfile`

backend/infrahub/core/diff/model/diff.py
fatih-acar: Switched to `__class__`

.devcontainer/Dockerfile
fatih-acar: Now using stricter pinning `3.14.2`

wvandeun (approved):

PR #7794, opened 2025-12-02
Add Claude command to fix mypy violations

ogenstad:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for resolving type checking violations, including step-by-step procedures, common error patterns, fix strategies, and validation workflows.

### DIFF
--- a/.claude/commands/fix-mypy-module.md
+++ b/.claude/commands/fix-mypy-module.md
@@ -1,0 +1,234 @@
+---
+description: Find and fix mypy typing violations for a specific module override in pyproject.toml
+allowed-tools: Bash(uv run mypy:*), Read, Edit, Glob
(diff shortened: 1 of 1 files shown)


dgarros (approved):
The command looks good but I'm wondering if some of the best practices in there should be stored somewhere else in the repo to be used / referenced outside of this command

ogenstad:
> The command looks good but I'm wondering if some of the best practices in there should be stored somewhere else in the repo to be used / referenced outside of this command

@dgarros, I merged it for now just to have it in the repo. Then we can reiterate. What section was it you were thinking of moving? While it might take a while to me all of these should eventually be covered by linters, so hopefully we can delete this command in the not too distant future. :)

PR #7807, opened 2025-12-03
IFC-1954 Add events to apply profiles on attr/rel change

gmazoyer:
This change adds the required prefect automation mechanisms to trigger re-applying a profile on its related nodes when an attribute or a relationship changes on the profile.

### DIFF
--- a/backend/infrahub/profiles/gather.py
+++ b/backend/infrahub/profiles/gather.py
@@ -1,0 +1,58 @@
+from __future__ import annotations
+
+from prefect import task
+from prefect.cache_policies import NONE
+
+from infrahub.core.constants import RelationshipKind
+from infrahub.core.registry import registry
+from infrahub.database import InfrahubDatabase  # noqa: TC001  needed for prefect flow
+from infrahub.workflows.catalogue import PROFILE_REFRESH_PROCESS
+
+from .models import ProfileRefreshTriggerDefinition
+
+
+@task(name="gather-trigger-profile-refresh", cache_policy=NONE)
+async def gather_trigger_profile_refresh(
+    db: InfrahubDatabase | None = None,  # noqa: ARG001 Needed to have a common function signature for gathering functions
+) -> list[ProfileRefreshTriggerDefinition]:
+    """Gather profile refresh triggers for all profile schemas.
+
+    This function creates trigger definitions for each profile schema that will
+    listen for `NodeUpdatedEvent` on profiles. When a profile's attributes or
+    relationships change, the trigger will fire and execute the profile refresh
+    workflow to re-apply profiles to all related nodes.
+    """
+    branches_with_diff_from_main = registry.get_altered_schema_branches()
+    branches_to_process: list[tuple[str, list[str]]] = [(branch, []) for branch in branches_with_diff_from_main]
+    branches_to_process.append((registry.default_branch, branches_with_diff_from_main))
+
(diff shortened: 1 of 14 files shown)


backend/tests/unit/profiles/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    SCHEMA = SchemaRoot(nodes=[location_schema, device_schema])
+
+    registry.schema.register_schema(schema=SCHEMA, branch=default_branch.name)
+    registry.schema.process_schema_branch(name=default_branch.name)
+    default_branch.update_schema_hash()
+    await default_branch.save(db=db)
+
+    return SCHEMA
```
ogenstad: I'm wondering about the return of a SchemaRoot object here. I think we might have this in other places too. Not sure if it's problematic and I think for this test it seems fine. But I'm wondering if we're processing the schema if we'd end up with a different SchemaRoot within the registry just based on the processing. I.e. the object that we return might have slight differences compared to the one in the registry.

As we don't use the object directly in any of the tests it doesn't matter in this case. Mostly thinking out loud and wondering if things like this might be an area of confusion in the future.

backend/infrahub/profiles/gather.py
```diff
@@ -0,0 +1,58 @@
...
+
+            trigger_attr = [
+                attr.name for attr in profile_schema.attributes if attr.name not in ("profile_name", "profile_priority")
+            ]
+            trigger_rels = [
+                rel.name
+                for rel in profile_schema.relationships
+                if rel.kind in (RelationshipKind.GENERIC, RelationshipKind.ATTRIBUTE) and rel.name != "related_nodes"
+            ]
+            trigger_fields = trigger_attr + trigger_rels
```
ogenstad: I wonder if this should be a sorted list. I'm not sure that the attributes always come back in the same order after a schema has been reloaded and it might trigger an update to the automations just because of that even if we didn't need one.

backend/infrahub/profiles/gather.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for branch_scope, branches_out_of_scope in branches_to_process:
+        schema_branch = registry.schema.get_schema_branch(name=branch_scope)
+
+        for profile_name in schema_branch.profile_names:
+            profile_schema = schema_branch.get_profile(name=profile_name, duplicate=False)
+
+            trigger_attr = [
+                attr.name for attr in profile_schema.attributes if attr.name not in ("profile_name", "profile_priority")
+            ]
```
ogenstad, line 39: One consideration here if we trigger on attributes is that I thought that this was already handled within the existing code?

It might be better to move it for consistency. I'm just not sure if that is the case now.

backend/infrahub/profiles/tasks.py
```diff
@@ -20,44 +30,87 @@
...
+    """
     log = get_run_logger()
+    client = get_client()
 
     await add_tags(branches=[branch_name])
 
-    for node_id in node_ids:
+    profile = await client.get(kind=profile_kind, ids=[profile_id], branch=branch_name)
+    related_nodes: RelationshipManager = profile.related_nodes  # type: ignore
+    await related_nodes.fetch()
```
ogenstad, line 104: Can we have this happen in one query instead of requiring that second fetch?

ogenstad (approved):

PR #7810, opened 2025-12-03
Fix schema cache issue after adding/removing enums or dropdowns

### LINKED ISSUE #7780: bug: Adding dropdown option via UI without schema update causes validation error and schema hash mismatch
## Component

- Frontend UI
- API Server / GraphQL

## Infrahub version

1.5.3

## Current Behavior

When adding a dropdown option through the UI without updating the schema:

1. The UI indicates it is authorizing the changes (appears to succeed)
2. The node does not appear in the list view
3. A validation error is displayed:
   ```
   xxx must be one of 'core, cpe, edge, firewall, leaf, spine' at role
   ```
   This indicates the new option `xxx` is not valid against the schema constraints.

4. Following this, after several page refreshes, an HTTP 500 error appears on the website with the following server log:
   ```
   ValueError: Schema 'InfraDevice' on branch main has incorrect hash: '76c7bb61c3fd0e676812c357786c82eb'
   ```
   Note: This schema hash error does not occur consistently (intermittent).

## Expected Behavior

1. The UI should either:
   - Prevent adding dropdown values that are not defined in the schema, OR
   - Clearly indicate that the operation failed due to schema constraints

2. The schema hash mismatch error should not occur - the system should maintain schema consistency

3. No HTTP 500 errors should be returned to the user for validation-related issues

## Steps to Reproduce

Using the demo schema and demo data
1. Navigate to a node type with a dropdown field (e.g., `InfraDevice` with a `role` field)
2. Attempt to add a new option value to the dropdown via the UI (e.g., adding a custom value not in the schema-defined options)
3. Observe that the UI appears to accept the change
4. Check that the node does not appear in the list view
5. Observe the validation error message
6. Refresh the page several times
7. Observe the intermittent 500 error and check server logs for the schema hash mismatch error

## Additional Information

The schema hash mismatch error suggests a potential race condition or cache inconsistency issue where the in-memory schema state diverges from the persisted state. This could be related to how dropdown options are being added without proper schema validation or synchronization.

ogenstad:
Some previous optimizations to avoid duplicating the schema was causing issues for the mutations where we actually want to duplicate the schema node, i.e. when we are modifying them.

(the codspeed issue is not related to anything within this PR)

Fixes #7780.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed intermittent "incorrect hash" errors after adding or removing dropdown or enum options in the UI; schema cache now preserves correct state across page refreshes.

* **Tests**
  * Added a unit test that verifies schema cache integrity when dropdown options are added, preventing in-place corruption of cached schema objects.

### DIFF
--- a/backend/infrahub/graphql/mutations/schema.py
+++ b/backend/infrahub/graphql/mutations/schema.py
@@ -81,7 +81,7 @@ class SchemaDropdownAdd(Mutation):
         _validate_schema_permission(graphql_context=graphql_context)
         await apply_external_context(graphql_context=graphql_context, context_input=context)
 
-        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=False)
+        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=True)
         attribute = str(data.attribute)
         validate_kind_dropdown(kind=kind, attribute=attribute)
         dropdown = str(data.dropdown)
@@ -141,7 +141,7 @@ class SchemaDropdownRemove(Mutation):
         graphql_context: GraphqlContext = info.context
 
         _validate_schema_permission(graphql_context=graphql_context)
-        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=False)
+        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=True)
         await apply_external_context(graphql_context=graphql_context, context_input=context)
 
         attribute = str(data.attribute)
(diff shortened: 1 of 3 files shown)


backend/infrahub/graphql/mutations/schema.py
```diff
@@ -104,7 +104,7 @@ async def mutate(
             context=graphql_context.get_context(),
         )
 
-        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=False)
+        kind = graphql_context.db.schema.get(name=str(data.kind), branch=graphql_context.branch.name, duplicate=True)
```
ajtmccarty: could this one stay `false`?
it looks like a read-only access from here

ogenstad: @ajtmccarty, yes that looks correct. I've reverted that last one. Could you take another look?

ajtmccarty (approved):

PR #6712, opened 2025-06-16
fix(backend): add database version check on startup

fatih-acar:

PR #7827, opened 2025-12-05
chore(deps): Upgrade backend infrastructure dependencies

fatih-acar:
## 
## Summary by CodeRabbit

* **New Features**
  * Added an automated workflow guide for upgrading backend infrastructure dependencies.

* **Chores**
  * Upgraded backend service images: Neo4j → 2025.10.1, Redis → 8.4.0, RabbitMQ → 4.2.1, PostgreSQL → 18.
  * Updated local demo and test environment configurations to use the newer images and updated volume path for Postgres.

* **Documentation**
  * Added changelog entry and updated demo docs to reflect upgraded dependencies.

This PR upgrades backend infrastructure dependencies to their latest versions.

## Version Changes

| Component | Previous | New |
|-----------|----------|-----|
| Neo4j | 2025.03.0 | 2025.10.1 |
| Redis | 7.2.11 | 8.4.0 |
| RabbitMQ | 3.13.7 | 4.2.1 |
| PostgreSQL | 16-alpine | 18-alpine (pgautoupgrade) |

## Files Updated

- [x] `backend/tests/helpers/constants.py` - Neo4j image versions
- [x] `backend/tests/conftest.py` - Redis and RabbitMQ test container images
- [x] `docker-compose.yml` - Main compose file defaults
- [x] `development/docker-compose-deps.yml` - Development compose defaults
- [x] `development/docker-compose-deps-nats.yml` - NATS compose PostgreSQL
- [x] `tasks/shared.py` - Invoke task defaults
- [x] `python_testcontainers/infrahub_testcontainers/container.py` - Test container env vars
- [x] `python_testcontainers/infrahub_testcontainers/docker-compose.test.yml` - Test compose
- [x] `python_testcontainers/infrahub_testcontainers/docker-compose-cluster.test.yml` - Cluster test compose
- [x] `utilities/db_backup/__main__.py` - Backup utility Neo4j image
- [x] `.github/workflows/version-upgrade.yml` - CI workflow Neo4j image
- [x] `docs/docs/topics/local-demo-environment.mdx` - Documentation (fixed outdated versions)

## Testing

- [x] CI passes all tests
- [x] Local docker-compose starts successfully
- [x] Integration tests pass
- [x] No performance regression when running infrahub-private-tests (scale tests)
- [x] Migration and auto-upgrade of persistent containers pass (version-upgrade job succeeds)
- [x] Helm chart has been tested with upgraded dependencies


.claude/commands/upgrade-backend-dependencies.md
dgarros, line 1: I think it would be best if we could have one PR per component by default,
it would be easier to identify regression / issues and troubleshoot.

Also if possible, I think it would be good to check the release note and identify if we need to change something in the code, and potentially open an issue to track that work if it's not mandatory to fix it right away

I understand the second point might be out of scope of this PR but I'd love to see us get there at some point

fatih-acar: > I think it would be best if we could have one PR per component by default,
it would be easier to identify regression / issues and troubleshoot.

I will make the command ask for a specific component by default (instead of defaulting to ALL).
For this specific one, upgrading all components at once was useful to not waste time on multiple PRs (luckily it looks like there's no compat issue at all, only a small RabbitMQ change required)

> Also if possible, I think it would be good to check the release note and identify if we need to change something in the code, and potentially open an issue to track that work if it's not mandatory to fix it right away

I think the LLM can do most of this for us, I'll add a step in the command.

.github/workflows/version-upgrade.yml
```diff
@@ -40,9 +40,16 @@ jobs:
           # - name: From 1.1.0
           #   source_version: 1.1.0
           #   neo4j_image: "neo4j:5.20.0-enterprise"
-          - name: From 1.3.0
-            source_version: 1.3.0
+          # - name: From 1.3.0
+          #   source_version: 1.3.0
+          #   neo4j_image: "neo4j:2025.03.0-enterprise"
```
ajtmccarty: leftover?

changelog/+upgrade-deps-20251206.changed.md
```diff
@@ -0,0 +1 @@
+Upgraded backend dependencies: Neo4j to 2025.10.1, Redis to 8.4.0, RabbitMQ to 4.2.1, PostgreSQL to 18
```
ogenstad, line 1: This gets a bit weird as the versions aren't strictly tied to the code and anyone could run Infrahub with some undefined version of these backend services. It's correct in terms of what we default to throughout the repo and it's probably better to include a statement of the changes to the default services. But we could consider how or where this should be presented in the future.

.claude/commands/upgrade-backend-dependencies.md
```diff
@@ -0,0 +1,586 @@
...
+## Step 10: Commit Changes
+
+```bash
+git add -A
+git commit -m "chore(deps): upgrade {component(s)} to {version(s)}
+
+- Updated {component} from {old} to {new}
+- Updated all relevant configuration files
+- Updated documentation
+- Applied required code changes per migration guide"
```
ogenstad, line 452: In general I'm not a fan of having the AI automatically create commits. In this case since it's so isolated to the updating of components it might be fine given the overall workflow of this. But outside of this when to commit changes to a branch and what to write is something that I want to be in control of.

ogenstad (approved):

.github/workflows/version-upgrade.yml
fatih-acar: Indeed thanks!

changelog/+upgrade-deps-20251206.changed.md
fatih-acar: I agree, this is because we are lucky enough to keep compatibility with older versions.
I believe we will need to add hard constraints to Infrahub at some point because of compat issues. Similar to what was done here: https://github.com/opsmill/infrahub/pull/6712

.claude/commands/upgrade-backend-dependencies.md
fatih-acar: Shouldn't this be a global instruction to the repo? i.e. "never create commits" or "ask for confirmation before commit"

ogenstad: That makes sense. Not sure if specific instructions in a command would override that though.

PR #7828, opened 2025-12-08
include created/updated_at/by metadata in merge logic

ajtmccarty:
IFC-1945

Adds support for updating created/updated_at/by metadata during a merge.

- the existing merge-related queries needed to be updated to include the correct from_user_id and to_user_id as taken from the corresponding edge on the branch being merged.
- needed a new query (`DiffMergeMetadataQuery`) to handle updating the updated_at/by properties on Node, Relationship, and Attribute vertices affected by the merge as well as setting the `previous_updated_by/at` property in case the merge needs to be rolled back before it completes. on the default and global branches, we store the metadata as properties directly on these vertices so that querying is faster as it does not require checking every edge connected to the vertex
- the DiffMergeRollbackQuery needed to be updated to handle rolling back updated_at/by properties
- had to update a whole of tests to cover these changes effectively

there is an existing edge case that we do not handle and I'm not sure if it is important enough to figure out.
consider the following:
- create `NodeA` and `NodeB`
- create branch `br2`
- on `main`, add a relationship from `NodeA` to `NodeB`
- on `br2`, delete `NodeA`
- merge `br2`. the merge logic will handle deleting all relationships touching `NodeA`, including the relationship to `NodeB` that does not exist on `br2`
- the metadata for this relationship and `NodeB` will not be updated as part of the merge, which is technically incorrect I think

TODO:
 - test performance on very big branch

## Summary by CodeRabbit

* **New Features**
  * Enhanced metadata tracking: created/updated timestamps and user attribution for nodes, attributes, and relationships.
  * Batch-applied metadata updates during merges and branch-agnostic metadata retrieval across queries.
  * Merge now propagates user/timestamp info through relationship and node operations and tracks affected nodes for more accurate rollback.

* **Bug Fixes**
  * Corrected relationship metadata detection and improved migrated-kind handling.
  * Adjusted time-tracking semantics for updated timestamps.

* **Tests**
  * Expanded tests covering metadata propagation, branch-agnostic queries, merge/rollback, and conflict scenarios.

### DIFF
--- a/backend/infrahub/core/diff/merger/merger.py
+++ b/backend/infrahub/core/diff/merger/merger.py
@@ -10,6 +10,7 @@ from infrahub.core.diff.query.merge import (
     DiffMergePropertiesQuery,
     DiffMergeQuery,
     DiffMergeRollbackQuery,
+    MergeMetadataQuery,
 )
 from infrahub.log import get_logger
 
@@ -39,6 +40,7 @@ class DiffMerger:
         self.db = db
         self.diff_repository = diff_repository
         self.serializer = serializer
+        self._affected_node_uuids: list[str] = []
 
     async def merge_graph(self, at: Timestamp) -> EnrichedDiffRoot:
         tracking_id = BranchTrackingId(name=self.source_branch.name)
@@ -69,6 +71,7 @@ class DiffMerger:
                 # make sure that we use the ADDED db_id if it exists
                 # it will not if a node was migrated and then deleted
                 migrated_kinds_id_map[n.uuid] = n.identifier.db_id
+
         async for node_diff_dicts, property_diff_dicts in self.serializer.serialize_diff(diff=enriched_diff):
             if node_diff_dicts:
                 log.info(f"Merging batch of nodes #{batch_num}")
@@ -105,13 +108,31 @@ class DiffMerger:
             )
             await migrated_merge_query.execute(db=self.db)
 
+        affected_node_uuids = [n.uuid for n in enriched_diff.nodes]
+        self._affected_node_uuids = affected_node_uuids
+        if affected_node_uuids:
+            finalize_query = await MergeMetadataQuery.init(
+                db=self.db,
+                branch=self.source_branch,
+                at=at,
+                target_branch=self.destination_branch,
+                node_uuids=affected_node_uuids,
+            )
+            await finalize_query.execute(db=self.db)
+
         self.source_branch.branched_from = at.to_string()
         await self.source_branch.save(db=self.db)
         registry.branch[self.source_branch.name] = self.source_branch
         return enriched_diff
 
     async def rollback(self, at: Timestamp) -> None:
+        if not self._affected_node_uuids:
+            return
         rollback_query = await DiffMergeRollbackQuery.init(
-            db=self.db, branch=self.source_branch, target_branch=self.destination_branch, at=at
+            db=self.db,
+            branch=self.source_branch,
+            target_branch=self.destination_branch,
+            at=at,
+            node_uuids=self._affected_node_uuids,
         )
         await rollback_query.execute(db=self.db)
--- a/backend/infrahub/core/diff/merger/merger.py
+++ b/backend/infrahub/core/diff/merger/merger.py
@@ -116,14 +116,14 @@ class DiffMerger:
             for i in range(0, len(affected_node_uuids), self.metadata_batch_size):
                 batch_uuids = affected_node_uuids[i : i + self.metadata_batch_size]
                 log.info(f"Updating metadata for batch {i // self.metadata_batch_size + 1} ({len(batch_uuids)} nodes)")
-                finalize_query = await DiffMergeMetadataQuery.init(
+                metadata_query = await DiffMergeMetadataQuery.init(
                     db=self.db,
                     branch=self.source_branch,
                     at=at,
                     target_branch=self.destination_branch,
                     node_uuids=batch_uuids,
                 )
-                await finalize_query.execute(db=self.db)
+                await metadata_query.execute(db=self.db)
 
         self.source_branch.branched_from = at.to_string()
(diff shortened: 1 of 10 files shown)


backend/infrahub/core/diff/merger/merger.py
```diff
@@ -105,13 +110,34 @@ async def merge_graph(self, at: Timestamp) -> EnrichedDiffRoot:
             )
             await migrated_merge_query.execute(db=self.db)
 
+        affected_node_uuids = [n.uuid for n in enriched_diff.nodes]
```
ajtmccarty, line 113: we use these in a rollback to increase performance of the rollback query

backend/infrahub/core/diff/query/merge.py
ajtmccarty, line 1: this is where most of the functional changes are
in the DiffMerge queries, we look find the responsible user ID and make sure to set it when updating the various edges on the default branch
we also set the `created_at/by` properties on added `Nodes`, `Relationships`, and `Attributes`
correctly updating the `updated_by/at` properties is a little harder and has its own new query

backend/infrahub/core/diff/query/merge.py
```diff
@@ -404,8 +446,8 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
             -[r2:IS_RELATED]
             -(rel_peer:Node {uuid: attr_rel_prop_diff.peer_uuid})
         WHERE attr_rel_prop_diff.relationship_id IS NOT NULL
-        AND (node_db_id IS NULL OR %(id_func)s(n) = node_db_id)
-        AND (peer_db_id IS NULL OR %(id_func)s(rel_peer) = peer_db_id)
+        AND (node_db_id IS NULL OR elementId(n) = node_db_id)
+        AND (peer_db_id IS NULL OR elementId(rel_peer) = peer_db_id)
```
ajtmccarty, line 450: since we've committed to neo4j, we don't need to to this `id_func` thing any longer

backend/infrahub/core/metadata/query/node_metadata.py
ajtmccarty, line 1: this new query is not used in production right now. it is only used in tests for verifying that metadata was correctly updated on deleted nodes, attributes, and relationships

backend/infrahub/core/metadata/determiner.py
```diff
@@ -47,11 +47,11 @@ async def determine_metadata_for_fields(self, node_fields: dict[str, Any]) -> Me
             if not (field_metadata_options & MetadataOptions.CREATED_BY) and "created_by" in metadata_properties_dict:
                 field_metadata_options |= MetadataOptions.CREATED_BY
             if not (field_metadata_options & MetadataOptions.SOURCE) and (
-                "source" in metadata_properties_dict or "_relation__owner" in metadata_properties_dict
+                "source" in metadata_properties_dict or "_relation__source" in metadata_properties_dict
```
ajtmccarty, line 50: these were reversed

backend/infrahub/core/node/__init__.py
ajtmccarty, line 1: forgot to pass user_id down in a few places

backend/tests/unit/core/diff/test_diff_and_merge.py
ajtmccarty, line 1: added a somewhat ridiculous amount of validation to these tests, but it helped catch a few issues, so probably worth it

backend/tests/unit/core/diff/test_diff_and_merge.py
```diff
@@ -117,6 +147,113 @@ async def test_diff_and_merge_schema_with_default_values(
...
+        "base_action,diff_action,selection,expect_deleted",
+        [
+            # DELETE on base + UPDATE on diff: selecting BASE_BRANCH keeps the node deleted
+            ("delete", "update", ConflictSelection.BASE_BRANCH, True),
+            # UPDATE on base + DELETE on diff: selecting BASE_BRANCH keeps the node with base value
+            ("update", "delete", ConflictSelection.BASE_BRANCH, False),
+            # Note: DIFF_BRANCH selection for node-level conflicts involving deletion is not tested here
+            # because un-deleting an object on the default branch is not supported
+        ],
+    )
```
ogenstad, line 160: Just a reminder of that pattern we use with dataclasses for a testcase, with something like that the comments could also go into a description of each test case. This test is fairly contained as it is to could be fine to use this format here, but in general I think the alternative is easier to read.

backend/tests/unit/core/diff/test_diff_and_merge.py
```diff
@@ -724,14 +1291,106 @@ async def test_branch_delete_with_added_base_relationship(
...
+            [r for r in car_meta.relationships if r.peer_uuid == person_john_main.id], key=lambda r: r.updated_at
+        )
+        assert len(car_rels_to_john) == 2
+        resolved_conflict_rel_to_john = car_rels_to_john[0]
+        assert resolved_conflict_rel_to_john.is_deleted is True
+        assert resolved_conflict_rel_to_john.created_by == "main-user-2"
+        assert before_owner_rel_resolved < resolved_conflict_rel_to_john.created_at < after_owner_rel_resolved
+        assert resolved_conflict_rel_to_john.updated_by == "main-user-2"
+        assert before_owner_rel_resolved < resolved_conflict_rel_to_john.updated_at < after_owner_rel_resolved
+        # original is actually updated later b/c it happens durign the merge
```
ogenstad: typo for during, looks like a copy-n-paste as it is repeated elsewhere.

ogenstad (approved):
Great work!

backend/tests/unit/core/diff/test_diff_and_merge.py
ajtmccarty: fixed

PR #7839, opened 2025-12-08
Prevent macros running if pool has not been processed yet

### LINKED ISSUE #7836: bug: Jinja2 Computed Attribute Error
### Component

_No response_

### Infrahub version

1.6.0

### Current Behavior

When using infrahub version 1.4.13 with the following schema attributes the result of the jinja2 template is `TEST000000100`

```yaml
      - name: number_sequence
        kind: NumberPool
        read_only: true
        optional: false
        parameters:
          start_range: 100
          end_range: 999999999
      - name: number
        kind: Text
        read_only: true
        optional: false
        computed_attribute:
          kind: Jinja2
          jinja2_template: "TEST{{ '%09d' | format(number_sequence__value) }}"
```

However after upgrading to 1.6.0 this jinja template breaks with infrahub giving the following error: 
`%d format: a real number is required, not NoneType`

### Expected Behavior

I would expect the jinja2 template to not be impacted between versions. 

### Steps to Reproduce

Load the following schema on 1.4.13

```
---
version: "1.0"
nodes:
  - name: Test
    namespace: Test
    description: "TestTest"
    label: "TEST"
    icon: "mdi:hub"
    display_labels: 
      - name__value
    human_friendly_id: []
    uniqueness_constraints: []
    attributes:
      # custom fields
      - name: name
        kind: Text
        unique: false
      - name: description
        kind: TextArea
        optional: true

      # auto-populated fields
      - name: number_sequence
        kind: NumberPool
        read_only: true
        optional: false
        parameters:
          start_range: 100
          end_range: 999999999
      - name: number
        kind: Text
        read_only: true
        optional: false
        computed_attribute:
          kind: Jinja2
          jinja2_template: "TEST{{ '%09d' | format(number_sequence__value) }}"
```

adding a new test should work. 

load the same schema on 1.6.0 and creating a node should fail. 

### Additional Information

_No response_

minitriga:
Fixes: #7836 

Logic Implemented: 

  1. For each variable in the Jinja2 template:
    - If it's from a relationship, handle it as before
    - If it's a local attribute:
        - Get the attribute object
      - Check if it's a pool attribute (from_pool is set) with an unallocated value (value is None)
      - If so, break out of the inner variable loop
      - Otherwise, get the property value and add to variables
  2. The for...else construct means:
    - If the inner loop completes normally (no break), execute the else block which renders the template and creates the attribute
    - If the inner loop breaks (pool value not allocated), skip the else block entirely, moving to the next macro in the outer loop
    
 I added a new regression test in backend/tests/unit/graphql/test_mutation_create_jinja2_attributes.py:

  - Test name: test_create_with_jinja2_format_filter_on_number_pool
  - Purpose: Tests that Jinja2 computed attributes using the format filter on NumberPool values work correctly
  - Specifically tests the GitHub issue #7836 scenario: A template like "TKT{{ '%09d' | format(sequence_number__value) }}" that formats a NumberPool value with leading zeros
  - The test passes (both for main branch and branch2 parametrized cases)

  The test creates:
  1. A TestingTicket schema with:
    - title - text field
    - sequence_number - NumberPool field
    - serial_number - Text field computed via Jinja2 using format filter on the NumberPool value
  2. Creates a ticket via GraphQL mutation
  3. Verifies the sequence_number gets allocated (value=1)
  4. Verifies the serial_number is correctly formatted as "TKT000000001"

## Summary by CodeRabbit

* **Bug Fixes**
  * Skip Jinja2 macro rendering and validation when a referenced attribute comes from a pool but hasn’t been allocated yet, preventing erroneous errors during attribute generation.

* **Tests**
  * Added a unit test validating Jinja2 computed attributes that format values allocated from a number pool (verifies zero-padded serial formatting).

* **Chores**
  * Added a changelog entry documenting the macro-processing skip behavior.

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -708,30 +708,32 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
                         variables[variable] = None
 
                 elif attribute_path.is_type_attribute:
-                    attribute = getattr(
-                        getattr(self, attribute_path.active_attribute_schema.name),
-                        attribute_path.active_attribute_property_name,
-                    )
+                    attr_obj: BaseAttribute = getattr(self, attribute_path.active_attribute_schema.name)
+                    # Skip macro if the referenced attribute is from a pool but hasn't been allocated yet
+                    if attr_obj.from_pool and attr_obj.value is None:
+                        break
+                    attribute = getattr(attr_obj, attribute_path.active_attribute_property_name)
                     variables[variable] = attribute
+            else:
+                # Only render if the inner loop completed without breaking (all variables resolved)
+                content = await jinja_template.render(variables=variables)
 
-            content = await jinja_template.render(variables=variables)
+                generator_method_name = "_generate_attribute_default"
+                if hasattr(self, f"generate_{attr_schema.name}"):
+                    generator_method_name = f"generate_{attr_schema.name}"
 
-            generator_method_name = "_generate_attribute_default"
-            if hasattr(self, f"generate_{attr_schema.name}"):
-                generator_method_name = f"generate_{attr_schema.name}"
+                generator_method = getattr(self, generator_method_name)
+                try:
+                    setattr(
+                        self,
+                        attr_schema.name,
+                        await generator_method(db=db, name=attr_schema.name, schema=attr_schema, data=content),
+                    )
+                    attribute = getattr(self, attr_schema.name)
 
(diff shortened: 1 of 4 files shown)


petercrocker:
@minitriga this still a valid PR? If so, can you resolve the conflicts, and I'll make sure it gets reviewed.

minitriga:
Closing this PR — the fix will be tracked via the existing bug report #7836, which has been updated with the root cause, proposed approach, and regression test plan from this PR. A new PR will be opened against that issue once the approach is agreed.

PR #7846, opened 2025-12-09
Restructure metadata query and response format

ogenstad:
This PR builds on #7683 but also replaces some parts as per the query defined in IFC-1813.

I.e the format looks like this:

```graphql
query CoreProposedChange {
  CoreProposedChange {
    edges {
      node_metadata {
          created_by {
            id
            display_label
          }
          updated_by {
            id
            display_label            
          }
          updated_at        
      }
      node {       
        name {
          value
          created_at
          created_by {
            id
            display_label
          }
          updated_at
          updated_by {
            id
            display_label
          }
        }
        reviewers {
          edges {
            properties { } Keep but deprecate
            relationship_metadata {
              updated_by
              updated_at
              * Also add existing properties fields
            }
            node_metadata {
              created_by
              created_at
              updated_by
              updated_at              
            }
            node { 
              name {
                value
                updated_at
                updated_by
              }
            }
          }
        }
      }
    }
  }
}
```

There are some parts of this PR which will be modified and combined with what we have in #7859, for the StandardNode's i.e. the Branch objects. I will work on those modifications once both of these PRs are merged.

One comment around this one is that the goal here is to get the query and response format correct, there's still an issue of looking up the display labels for any IDs under the created_by or updated_by fields. But we should be able to add that once the structure is in place.

## Summary by CodeRabbit

* **New Features**
  * GraphQL exposes per-item audit metadata (node_metadata and relationship_metadata) with created_at/created_by and updated_at/updated_by across nodes, relationships, attributes and branches.
* **Refactor**
  * Query/resolver/data-loader flows now accept per-call metadata selection so metadata is loaded only when requested.
  * Several list endpoints and mutations return flattened node representations.
* **Tests**
  * Added/updated tests validating metadata fields in queries and paginated responses.

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -552,6 +552,12 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
                     )
                 continue
 
+            if field_name == "meta" and isinstance(fields.get("meta"), dict):
+                response[field_name] = {
+                    meta_field: getattr(self, meta_field, None) for meta_field in fields.get(field_name).keys()
+                }
+                continue
+
             if field_name.startswith("_"):
                 field = getattr(self, field_name[1:])
             else:
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -527,6 +527,10 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
                 response[field_name] = updated_at.to_graphql() if updated_at else None
                 continue
 
+            if field_name == "updated_by":
+                response[field_name] = self._get_updated_by()
+                continue
+
             if field_name == "__typename":
                 response[field_name] = self.get_kind()
                 continue
@@ -550,12 +554,6 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
                         fields={"id": None, "display_label": None, "__typename": None},
                         related_node_ids=related_node_ids,
                     )
-                continue
-
-            if field_name == "meta" and isinstance(fields.get("meta"), dict):
-                response[field_name] = {
-                    meta_field: getattr(self, meta_field, None) for meta_field in fields.get(field_name).keys()
-                }
                 continue
 
             if field_name.startswith("_"):
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -20,6 +20,7 @@ from infrahub.core.constants import (
     SYSTEM_USER_ID,
     AttributeDBNodeType,
     BranchSupportType,
+    InfrahubKind,
     MetadataOptions,
 )
 from infrahub.core.metadata.interface import MetadataInterface
@@ -524,11 +525,18 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
         for field_name in field_names:
             if field_name == "updated_at":
                 updated_at = self._get_updated_at()
-                response[field_name] = updated_at.to_graphql() if updated_at else None
+                response[field_name] = await updated_at.to_graphql() if updated_at else None
                 continue
 
             if field_name == "updated_by":
-                response[field_name] = self._get_updated_by()
+                response[field_name] = (
+                    {
+                        "id": self._get_updated_by(),
+                        "__kind__": InfrahubKind.ACCOUNT,
+                    }
+                    if self._get_updated_by()
+                    else None
+                )
                 continue
 
             if field_name == "__typename":
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -37,7 +37,6 @@ class Branch(StandardNode):
     origin_branch: str = "main"
     branched_from: Optional[str] = Field(default=None, validate_default=True)
     hierarchy_level: int = 2
-    created_at: Optional[str] = Field(default=None, validate_default=True)
     is_default: bool = False
     is_global: bool = False
     is_protected: bool = False
@@ -92,11 +91,6 @@ class Branch(StandardNode):
         if not self.branched_from:
             raise RuntimeError(f"branched_from not set for branch {self.name}")
         return self.branched_from
-
-    @field_validator("created_at", mode="before")
-    @classmethod
-    def set_created_at(cls, value: str) -> str:
-        return Timestamp(value).to_string()
 
     def get_created_at(self) -> str:
         if not self.created_at:
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -75,13 +75,8 @@ def get_schema[SchemaProtocol](
 ) -> MainSchemaTypes:
     if isinstance(node_schema, str):
         return db.schema.get(name=node_schema, branch=branch.name, duplicate=duplicate)
-    if (
-        hasattr(node_schema, "_is_runtime_protocol")
-        and node_schema._is_runtime_protocol
-    ):
-        return db.schema.get(
-            name=node_schema.__name__, branch=branch.name, duplicate=duplicate
-        )
+    if hasattr(node_schema, "_is_runtime_protocol") and node_schema._is_runtime_protocol:
+        return db.schema.get(name=node_schema.__name__, branch=branch.name, duplicate=duplicate)
     if not isinstance(node_schema, (MainSchemaTypes)):
         raise ValueError(f"Invalid schema provided {node_schema}")
 
@@ -213,9 +208,7 @@ class NodeManager:
             # This is a workaround to ensure we are querying the right fields for permissions
             # The identifier for permissions needs the same fields as the display label
             schema_branch = db.schema.get_schema_branch(name=branch.name)
-            display_label_fields = schema_branch.generate_fields_for_display_label(
-                name=node_schema.kind
-            )
+            display_label_fields = schema_branch.generate_fields_for_display_label(name=node_schema.kind)
             if display_label_fields:
                 deep_merge_dict(dicta=fields, dictb=display_label_fields)
 
@@ -237,12 +230,7 @@ class NodeManager:
     async def count(
         cls,
         db: InfrahubDatabase,
-        schema: type[SchemaProtocol]
-        | NodeSchema
-        | GenericSchema
-        | ProfileSchema
-        | TemplateSchema
-        | str,
+        schema: type[SchemaProtocol] | NodeSchema | GenericSchema | ProfileSchema | TemplateSchema | str,
         filters: dict | None = None,
         at: Timestamp | str | None = None,
         branch: Branch | str | None = None,
@@ -358,9 +346,7 @@ class NodeManager:
                 InfrahubKind.OBJECTPERMISSION,
             ]:
                 schema_branch = db.schema.get_schema_branch(name=branch.name)
-                display_label_fields = schema_branch.generate_fields_for_display_label(
-                    name=peer_schema.kind
-                )
+                display_label_fields = schema_branch.generate_fields_for_display_label(name=peer_schema.kind)
                 if display_label_fields:
                     deep_merge_dict(dicta=fields, dictb=display_label_fields)
 
@@ -660,9 +646,7 @@ class NodeManager:
         kind_str = node_schema.kind
 
         if not node_schema.default_filter:
-            raise NodeNotFoundError(
-                branch_name=branch.name, node_type=kind_str, identifier=id
-            )
+            raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=id)
 
         items = await NodeManager.query(
             db=db,
@@ -839,9 +823,7 @@ class NodeManager:
 
         filters = {}
         for key, item in zip(node_schema.human_friendly_id, hfid, strict=False):
-            path = node_schema.parse_schema_path(
-                path=key, schema=registry.schema.get_schema_branch(name=branch.name)
-            )
+            path = node_schema.parse_schema_path(path=key, schema=registry.schema.get_schema_branch(name=branch.name))
 
             if path.is_type_relationship and path.related_schema:
                 rel_schema = path.related_schema
@@ -851,9 +833,7 @@ class NodeManager:
                     schema=registry.schema.get_schema_branch(name=branch.name),
                 )
 
-            filters[key] = (
-                path.active_attribute_schema.get_class().deserialize_from_string(item)
-            )
+            filters[key] = path.active_attribute_schema.get_class().deserialize_from_string(item)
 
         items = await NodeManager.query(
             db=db,
@@ -872,9 +852,7 @@ class NodeManager:
 
         if len(items) < 1:
             if raise_on_error:
-                raise NodeNotFoundError(
-                    branch_name=branch.name, node_type=kind_str, identifier=hfid_str
-                )
+                raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=hfid_str)
             return None
 
         if len(items) > 1:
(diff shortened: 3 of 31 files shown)


backend/infrahub/core/node/standard.py
```diff
@@ -77,10 +78,28 @@ async def to_graphql(self, fields: dict) -> dict:
             if field_name == "__typename":
                 response[field_name] = self.get_type()
                 continue
-            field = getattr(self, field_name)
+            if field_name == "node_metadata" and isinstance(fields.get("node_metadata"), dict):
+                _result = {}
+                for meta_field in fields.get(field_name, {}).keys():
+                    if meta_field == "created_at":
```
ajtmccarty: should this be applied to `updated_at` too?

backend/infrahub/core/metadata/determiner.py
```diff
@@ -6,19 +6,26 @@
...
+        self, node_fields: dict[str, Any], metadata_fields: dict[str, Any] | None = None
+    ) -> MetadataQueryOptions:
+        """Determine metadata required based on fields requested.
 
-        All attribute-level metadata is combined and all relationship-level metadata is combined
+        Args:
+            node_fields: The node fields requested in the query.
+            metadata_fields: The metadata fields requested for the node (e.g., created_at, updated_at).
+
+        All attribute-level metadata is combined and all relationship-level metadata is combined.
```
ajtmccarty: I left this one behind and it is not accurate any longer. can you remove it please?

backend/infrahub/core/node/__init__.py
```diff
@@ -1072,6 +1073,10 @@ async def to_graphql(
                     response[field_name] = None
                 continue
 
+            if field_name == "node_metadata" and isinstance(fields.get("node_metadata"), dict):
+                response[field_name] = await self._build_meta_response(field_name, fields)
+                continue
```
ajtmccarty: I find how we build these graphql responses to be confusing, so I might be wrong here, but this looks like the `node_metadata` field is at the same level as `hfid` and `display_label`, when I would expect it to be one layer up
```
query CoreProposedChange {
  CoreProposedChange {
    edges {      
      node_metadata { ... }   // <-- where it is in the example in the description
      node {
        node_metadata { ... }   // <-- where I think this code expects it to be
        name { ... }
        display_label { ... }
      }
  }  
}
```

backend/infrahub/graphql/resolvers/many_relationship.py
```diff
@@ -110,6 +132,10 @@ async def resolve(
         edges = fields.get("edges", {})
         node_fields = edges.get("node", {})
         property_fields = edges.get("properties", {})
+        metadata_fields = {
+            "node_metadata": edges.get("node_metadata", {}),
+            "relationship_metadata": edges.get("relationship_metadata", {}),
+        }
```
ajtmccarty, line 138: feels like `metadata_fields` should be a dataclass or some other better-defined data structure if it is going to hold these separate unrelated dictionaries

backend/infrahub/graphql/types/branch.py
```diff
@@ -67,23 +66,31 @@ async def get_by_name(
         raise BranchNotFoundError(f"Branch with name '{name}' not found")
 
 
-class RequiredStringValueField(InfrahubObjectType):
+class InfrahubBranchMetaObject(ObjectType):
+    updated_by = String(required=False, description="User that last modified the attribute or relationship")
```
ajtmccarty: "User that last modified the ~attribute or relationship~ branch(?)"

backend/infrahub/graphql/types/branch.py
```diff
@@ -67,23 +66,31 @@ async def get_by_name(
...
+    updated_at = DateTime(
+        required=False,
+        description="Date/Time when the attribute or relationship was last modified by a user or a system task",
+    )
```
ajtmccarty: should this include `created_at/by`? we should support that for `Branch` objects

backend/infrahub/graphql/types/attribute.py
```diff
@@ -73,7 +73,15 @@ class AttributeInterface(InfrahubInterface):
     # owner = Field("DataOwner")
 
 
-class BaseAttribute(ObjectType):
+class InfrahubAttributeMetaObject(ObjectType):
+    updated_by = String(required=False, description="User that last modified the attribute")
+    updated_at = DateTime(
+        required=False,
+        description="Date/Time when the attribute was last modified by a user or a system task",
+    )
```
ajtmccarty, line 78: should this include `created_at/by`? we save that information for Attribute objects

backend/infrahub/graphql/types/metadata.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    updated_at = DateTime(
+        required=False, description="Date/Time when the object was last modified by a user or a system task"
+    )
+
+
+class InfrahubRelationshipMetaData(ObjectType):
+    created_at = DateTime(required=False, description="Date/Time the object has been created")
+    created_by = String(
+        required=False, description="UUID of the user that created the object, even if the user is later deleted"
```
ajtmccarty: object -> relationship ?

backend/infrahub/graphql/manager.py
```diff
@@ -311,26 +321,44 @@ def generate_object_types(self) -> None:
...
         data_source = self.get_type(name=InfrahubKind.LINEAGESOURCE)
         data_owner = self.get_type(name=InfrahubKind.LINEAGEOWNER)
         self.define_relationship_property(data_source=data_source, data_owner=data_owner)
+
+        # Now that GENERICACCOUNT exists, create node_metadata and relationship_metadata
+        account_type = self.get_type(name=InfrahubKind.GENERICACCOUNT)
+        self.define_node_metadata(account_type=account_type)
+        self.define_relationship_metadata(account_type=account_type)
+        node_metadata = self.get_type(name="InfrahubNodeMetadata")
+        relationship_metadata = self.get_type(name="InfrahubRelationshipMetadata")
```
ajtmccarty: maybe these names should be constants somewhere? not really sure where, but that is my first thought when I see long string names in multiple places

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -342,3 +342,60 @@ async def test_paginated_branch_query__returns_error_on_invalid_offset_or_limit(
...
+        assert all_branches.data
+
+        for branch in all_branches.data["InfrahubBranch"]["edges"]:
+            assert branch["node"]["name"]["value"]
+            assert branch["node"]["name"]["updated_at"] is None
+            assert branch["node"]["name"]["updated_by"] is None
+
+            assert branch["node"]["description"]["value"]
+            assert branch["node"]["description"]["updated_at"] is None
+            assert branch["node"]["description"]["updated_by"] is None
```
ajtmccarty: is this just temporary? seems like the metadata fields should be populated

backend/tests/unit/graphql/queries/test_proposed_change_actions.py
```diff
@@ -239,3 +287,48 @@ async def test_proposed_change_draft(
...
+        service=service,
+        account_session=session_admin,
+    )
+
+    assert not response.errors
+    assert response.data
+
+    for prc in response.data["CoreProposedChange"]["edges"]:
+        assert prc["node"]["name"]["value"]
+        assert prc["node"]["name"]["updated_by"]
```
ajtmccarty: might be nice to either retrieve the PC and verify that the created/updated_by/at values are correct in here or set `Timestamp`s on either side of `proposed_change.save()` and verify that the timestamps that way

backend/tests/unit/graphql/test_graphql_query.py
ajtmccarty, line 1: I think you need a new test that verifies that requesting different metadata options at different nesting levels works correctly.
specifically, I am worried about getting different options at the relationship level and peer level because I find the different loaders and `to_graphql` methods difficult to fully understand

backend/infrahub/core/metadata/determiner.py
ogenstad: I'm not completely sure what you mean here. What is it that should be removed?

backend/infrahub/core/node/standard.py
```diff
@@ -81,14 +99,64 @@ async def to_graphql(self, fields: dict) -> dict:
...
+                case "created_at":
+                    meta_response["created_at"] = Timestamp(self.created_at).to_datetime() if self.created_at else None
+                case "updated_at":
+                    meta_response["updated_at"] = Timestamp(self.updated_at).to_datetime() if self.updated_at else None
+                case "created_by":
+                    if self.created_by and self.created_by != SYSTEM_USER_ID:
+                        meta_response["created_by"] = {"id": self.created_by}
+                case "updated_by":
+                    if self.updated_by and self.updated_by != SYSTEM_USER_ID:
+                        meta_response["updated_by"] = {"id": self.created_by}
```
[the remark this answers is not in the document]
ogenstad: fixed locally

backend/infrahub/core/metadata/determiner.py
ajtmccarty: line 18 in the docstring. this method returns separate metadata requirements for attributes and relationships, but the docstring says they are combined

backend/infrahub/core/node/__init__.py
ogenstad: No sorry my bad this was left since the earlier iteration and should have been removed, it wasn't being used in this context. Will have removed it when I push some more code.

backend/infrahub/core/metadata/determiner.py
ogenstad: Ok thanks.

backend/infrahub/graphql/types/branch.py
ogenstad: I've also removed this part since these extra metadata entries doesn't make much sense to have in the branch or standard node objects.

backend/tests/unit/graphql/queries/test_branch.py
ogenstad: This would have been a placeholder until it was possible to load those metadata values from the StandardNode objects, as we won't have them initially I've just removed them from the GraphQL API. Some parts might be relevant to have but not like this. One thing that comes to mind is that it could be useful to see when a schema was modified on a branch and perhaps other things. But that's out of scope for now.

backend/infrahub/graphql/manager.py
ogenstad: Yes, fixed in coming changes.

backend/infrahub/core/node/standard.py
ogenstad: I've refactored this so that it should be easier to read in general and also split the to_graphql method apart into separate things for the old and new "Branch" variations. At some point we'll also phase out the old mutations just called BranchCreate, BranchMerge and have InfrahubBranchCreate etc.

backend/infrahub/core/branch/models.py
```diff
@@ -93,11 +92,6 @@ def get_branched_from(self) -> str:
             raise RuntimeError(f"branched_from not set for branch {self.name}")
         return self.branched_from
 
-    @field_validator("created_at", mode="before")
-    @classmethod
-    def set_created_at(cls, value: str) -> str:
-        return Timestamp(value).to_string()
-
```
ogenstad, line 100: Moved to StandardNode to avoid some typing issues there.

backend/infrahub/core/attribute.py
```diff
@@ -524,11 +525,18 @@ async def to_graphql(
...
             if field_name == "updated_by":
-                response[field_name] = self._get_updated_by()
+                response[field_name] = (
+                    {
+                        "id": self._get_updated_by(),
+                        "__kind__": InfrahubKind.ACCOUNT,
+                    }
+                    if self._get_updated_by()
+                    else None
+                )
```
ogenstad, line 539: We want this to be a CoreAccount so we can't just return a string, will need to look at some resolver to grab the display label and such as a follow up.

backend/infrahub/core/task/task.py
```diff
@@ -70,7 +72,7 @@ async def query(
                 logs = [
                     {
                         "node": await TaskLog.from_db(result, extras={"task_id": task_result.get("uuid")}).to_graphql(
-                            fields=log_fields
+                            fields=StandardNodeQueryFields(node=log_fields)
```
ogenstad, line 75: This is just to make mypy happy for now. But this is just some old code that isn't being used and should be deleted.

backend/infrahub/graphql/types/branch.py
```diff
@@ -33,64 +35,35 @@ class Meta:
...
-        cls,
-        fields: dict,
-        graphql_context: GraphqlContext,
-        name: str,
-    ) -> dict[str, Any]:
-        branch_responses = await cls.get_list(fields=fields, graphql_context=graphql_context, name=name)
-
-        if branch_responses:
-            return branch_responses[0]
-        raise BranchNotFoundError(f"Branch with name '{name}' not found")
```
ogenstad, line 67: Moved to the correct class where it's actually being used.

backend/infrahub/graphql/types/branch.py
```diff
@@ -33,64 +35,35 @@ class Meta:
...
-
-
-class InfrahubBranchMetaObject(ObjectType):
-    updated_by = String(required=False, description="User that last modified the attribute or relationship")
-    updated_at = DateTime(
-        required=False,
-        description="Date/Time when the attribute or relationship was last modified by a user or a system task",
-    )
```
ogenstad: Remove attribute level metadata on standard node objects.

backend/infrahub/core/node/standard.py
```diff
@@ -30,9 +31,16 @@
     from infrahub.database import InfrahubDatabase
 
 
+@dataclass(slots=True)
+class StandardNodeQueryFields:
+    node: dict[str, Any] = field(default_factory=dict)
+    node_metadata: dict[str, Any] = field(default_factory=dict)
```
ajtmccarty, line 37: this is a good data structure

backend/infrahub/core/node/standard.py
```diff
@@ -81,14 +101,64 @@ async def to_graphql(self, fields: dict) -> dict:
...
+                case "created_at":
+                    meta_response["created_at"] = Timestamp(self.created_at).to_datetime() if self.created_at else None
+                case "updated_at":
+                    meta_response["updated_at"] = Timestamp(self.updated_at).to_datetime() if self.updated_at else None
+                case "created_by":
+                    if self.created_by and self.created_by != SYSTEM_USER_ID:
+                        meta_response["created_by"] = {"id": self.created_by}
+                case "updated_by":
+                    if self.updated_by and self.updated_by != SYSTEM_USER_ID:
+                        meta_response["updated_by"] = {"id": self.updated_by}
```
ajtmccarty, line 158: maybe an issue for later, but I'd think that we would want to be able to show when we don't know who made a change (`updated_by=null`) vs when Infrahub made a change (`updated_by=__system__`)

backend/infrahub/core/manager.py
```diff
@@ -87,6 +92,7 @@ async def query(
         schema: NodeSchema | GenericSchema | ProfileSchema | TemplateSchema | str,
         filters: dict | None = ...,
         fields: dict | None = ...,
+        metadata_fields: dict | None = ...,
```
ajtmccarty, line 95: I feel like we should be able to avoid passing `metadata_fields` all the way down into NodeManager from the graphql query. could we use either the existing MetadataDeterminer or a new MetadataDetermineGraphQL to set `include_metadata` at a higher level and then pass that down here?

backend/infrahub/graphql/queries/branch.py
```diff
@@ -48,27 +51,32 @@ async def infrahub_branch_resolver(
     fields = extract_graphql_fields(info)
     result: dict[str, Any] = {}
     if "edges" in fields:
+        query_fields = StandardNodeQueryFields(
+            node=fields.get("edges", {}).get("node", {}),
+            node_metadata=fields.get("edges", {}).get("node_metadata", {}),
+        )
```
ajtmccarty, line 57: I'll say it again. this is a good data structure. it's easy to understand and these two dictionaries belong together

ajtmccarty (approved):

PR #7854, opened 2025-12-09
fix(backend): do not delete rels when previewing mutation object

### LINKED ISSUE #7853: bug: nulling/removing a mandatory parent is allowed
### Component

API Server / GraphQL

### Infrahub version

1.6.0

### Current Behavior

In some cases, it might be possible to end up with a hierarchical node having no parent at all when it should be mandatory to have one.

### Expected Behavior

Nulling a parent should be correctly denied

### Steps to Reproduce

```
# load base schemas
infrahubctl schema load schema-library/base schema-library/extensions/location_minimal schema-library/extensions/compute/

infrahubctl object load objects/locations.yml
```

Then execute following graphql query:

```
mutation {
  LocationRackUpdate(data:  {
     hfid: ["test-rack"],
     parent: null
  }) {
    ok
  }
}
```

This should throw
```
{
  "data": {
    "LocationRackUpdate": null
  },
  "errors": [
    {
      "message": "Too few relationships, min 1 at parent",
      "locations": [
        {
          "line": 16,
          "column": 3
        }
      ],
      "path": [
        "LocationRackUpdate"
      ]
    }
  ]
}
```

But the node had its parent nulled:
```
{
  "data": {
    "LocationRack": {
      "edges": [
        {
          "node": {
            "hfid": [
              "test-rack"
            ],
            "parent": {
              "node": null
            }
          }
        }
      ]
    }
  }
}
```

```
# locations.yml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationCountry
  data:
    - name: test-country
      shortname: test-country
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationMetro
  data:
    - name: test-metro
      shortname: test-metro
      parent: test-country
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationSite
  data:
    - name: test-site
      shortname: test-site
      parent: test-metro
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: LocationRack
  data:
    - name: test-rack
      shortname: test-rack
      parent: test-site
```

### Additional Information

_No response_

fatih-acar:
PR https://github.com/opsmill/infrahub/pull/7508 introduced a regression where relationship constraints would not rollback the transaction in case of failures due to the introduction of a "preview object" executed outside of the transaction. This would leave objects with nulled relationships where it would not have been allowed.

Make sure those deletions are not processed when building this preview object.
Also add more coverage to the unit test.

Fixes #7853

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended removal of mandatory relationships; relationship updates now preserve consistent state when deletion is conditionally skipped.

* **Tests**
  * Added test coverage to confirm required relationships remain intact after failed update attempts.

* **Changelog**
  * Added a release note documenting the fix for mandatory relationship removal.

### DIFF
--- a/backend/tests/unit/graphql/test_mutation_update.py
+++ b/backend/tests/unit/graphql/test_mutation_update.py
@@ -1627,6 +1627,32 @@ async def test_removing_mandatory_relationship_not_allowed(
     assert len(result.errors) == 1
     assert result.errors[0].message == "Too few relationships, min 1 at owner"
 
+    query = """
+    query {
+        TestDog(ids: ["%(animal_id)s"]) {
+            edges {
+                node {
+                    owner {
+                        node {
+                            id
+                        }
+                    }
+                }
+            }
+        }
+    }""" % {"animal_id": dog1.id}
+    gql_params = await prepare_graphql_params(db=db, branch=default_branch)
+    result = await graphql(
+        schema=gql_params.schema,
+        source=query,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={},
+    )
+    assert result.data["TestDog"]["edges"][0]["node"]["owner"]["node"] is not None
+    retrieved_owner_id = result.data["TestDog"]["edges"][0]["node"]["owner"]["node"]["id"]
+    assert retrieved_owner_id == person1.id
+
 
 async def test_updating_relationship_when_peer_side_is_required(
     db: InfrahubDatabase, default_branch, animal_person_schema
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -1091,7 +1091,7 @@ class Node(BaseNode, metaclass=BaseNodeMeta):
 
             if key in self._relationships:
                 rel: RelationshipManager = getattr(self, key)
-                changed |= await rel.update(db=db, data=value)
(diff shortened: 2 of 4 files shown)


ogenstad (approved):
Can you also add a newsfragment?

backend/tests/unit/graphql/test_mutation_update.py
```diff
@@ -1627,6 +1627,32 @@ async def test_removing_mandatory_relationship_not_allowed(
...
+    result = await graphql(
+        schema=gql_params.schema,
+        source=query,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={},
+    )
+    assert result.data["TestDog"]["edges"][0]["node"]["owner"]["node"] is not None
+    retrieved_owner_id = result.data["TestDog"]["edges"][0]["node"]["owner"]["node"]["id"]
+    assert retrieved_owner_id == person1.id
```
ajtmccarty, line 1654: I think retrieving the TestDog with NodeManager.get_one would allow you verify the same thing and let's you skip the graphql overhead

fatih-acar: Thanks, I was not sure about that so I went with the graphql query instead.
If both works I'll keep the graphql query as its already done.

PR #7857, opened 2025-12-10
set updated_at/by for peer Nodes when updating rel on default/global

ajtmccarty:
when we update a Relationship on the default or global branch, we need to set the `updated_at/by` properties on both peer `Node`s, otherwise the destination peer Node will not get its updated_at/by metadata set correctly

also includes an update to `DATABASE.md` to help with cypher query generation

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive new documentation section with patterns and practical examples for identifying active nodes on default and global branches and managing their relationships.

* **Improvements**
  * Enhanced metadata synchronization to ensure modification timestamps are consistently propagated and properly maintained across all related entities whenever relationship operations (creation, updates, or deletions) occur.

### DIFF
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -340,6 +340,11 @@ class RelationshipCreateQuery(RelationshipWriteQuery):
             r1,
             r2,
         )
+        if self.branch.is_default or self.branch.is_global:
+            query_create += """
+        SET s.updated_at = $at, s.updated_by = $user_id
+        SET d.updated_at = $at, d.updated_by = $user_id
+            """
 
         self.add_to_query(query_create)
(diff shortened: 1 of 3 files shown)


backend/infrahub/core/query/relationship.py
```diff
@@ -340,6 +340,11 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
             r1,
             r2,
         )
+        if self.branch.is_default or self.branch.is_global:
```
ogenstad, line 343: Just noting that we seem to use `if self.branch.is_default or self.branch.is_global:` in a number of places. Might be worth considering having a property for the same thing on the branch objects.

ogenstad (approved):

PR #7859, opened 2025-12-10
Add branch metadata

ogenstad:
Adds metadata to StandardNode or rather Branch. This PR takes a more pragmatic approach than what we have in #7701. The reason is that there's not much that a user can change on a branch. Mainly the description of a branch. Instead the focus is to allow us to create the user that created the branch as well as the user who last updated it, this could be due to changing the description, doing a rebase or loading a schema. It also includes a new .updated_at timestamp.

This doesn't yet expose this information to the GraphQL layer, that will be done in a separate PR.

## Summary by CodeRabbit

* **New Features**
  * Added audit fields (created_by, updated_by, updated_at) and capture the acting user for branch/node create, update, rebase, save and schema load operations.

* **Tests**
  * Expanded unit and functional tests and fixtures to validate audit metadata and that acting account (admin/bot) is recorded correctly across branch and schema workflows.

### DIFF
--- a/backend/infrahub/api/schema.py
+++ b/backend/infrahub/api/schema.py
@@ -375,7 +375,7 @@ async def load_schema(
                 branch.is_isolated = True
                 log.info("Branch converted to isolated mode because the schema has changed", branch=branch.name)
 
-            await branch.save(db=dbt)
+            await branch.save(db=dbt, user_id=account_session.account_id)
             updated_branch = registry.schema.get_schema_branch(name=branch.name)
             updated_hash = updated_branch.get_hash()
 
--- a/backend/infrahub/core/node/standard.py
+++ b/backend/infrahub/core/node/standard.py
@@ -7,9 +7,9 @@ from uuid import UUID
 
 import ujson
 from infrahub_sdk.uuidt import UUIDT
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
-from infrahub.core.constants import NULL_VALUE
+from infrahub.core.constants import NULL_VALUE, SYSTEM_USER_ID
 from infrahub.core.query.standard_node import (
     StandardNodeCreateQuery,
     StandardNodeDeleteQuery,
@@ -18,6 +18,7 @@ from infrahub.core.query.standard_node import (
     StandardNodeQuery,
     StandardNodeUpdateQuery,
 )
+from infrahub.core.timestamp import current_timestamp
 from infrahub.exceptions import Error, InitializationError
 
 if TYPE_CHECKING:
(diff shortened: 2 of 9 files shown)


gmazoyer (approved):

backend/infrahub/core/node/standard.py
```diff
@@ -32,6 +33,9 @@
 class StandardNode(BaseModel):
     id: Optional[str] = None
     uuid: Optional[UUID] = None
+    created_by: str = Field(default=SYSTEM_USER_ID)
+    updated_by: str = Field(default=SYSTEM_USER_ID)
```
ajtmccarty: I'm not sure if this is what we want over `None`
when we actually release this, most Nodes and StandardNodes in the database will not have an `updated_by` user and I think defaulting to the SYSTEM_USER_ID in this case is kind of misleading b/c we don't actually know what user did the latest update

backend/tests/functional/branch/test_graphql_mutation.py
```diff
@@ -160,6 +167,12 @@ async def test_branch_rebase(self, initial_dataset: str, client: InfrahubClient)
         branch_after = await client.branch.get(branch_name=branch.name)
         assert branch.branched_from != branch_after.branched_from
 
+        # Validate updated_by field is correctly set until this can be queried via GraphQL and the metadata fields
+        async with db.start_session() as dbs:
+            rebase_branch = await Branch.get_by_name(db=dbs, name="branch_to_rebase")
+        assert rebase_branch.updated_by == admin_account.id
```
ajtmccarty: might be good to also test that `updated_at` is set correctly and that `created_at` is unchanged

backend/infrahub/api/schema.py
```diff
@@ -375,7 +375,7 @@ async def load_schema(
                 branch.is_isolated = True
                 log.info("Branch converted to isolated mode because the schema has changed", branch=branch.name)
 
-            await branch.save(db=dbt)
+            await branch.save(db=dbt, user_id=account_session.account_id)
```
ajtmccarty, line 378: would be good to add a test for this to a schema update test somewhere

ajtmccarty (approved):

backend/infrahub/core/node/standard.py
ogenstad: Makes sense, I've changed this. Not completely sure about the static nature of the SYSTEM_USER_ID either as we'll return a CoreAccount type within the GraphQL responses but it could be that anything with SYSTEM_USER_ID will just be a `null` value indicating the system. Alternatively we'd could populate things like display_label with system generated data so that it would always just be "system" or whatever.

backend/tests/functional/branch/test_graphql_mutation.py
ogenstad: Yes, I've added this along with some updated fixtures so we can decide which client we use within these tests for similar checks.

backend/infrahub/api/schema.py
ogenstad: Fixed

PR #7868, opened 2025-12-11
Added type NodeObjectWithMetadata for get-object

bilalabbad:
## Summary by CodeRabbit

* **New Features**
  * Per-attribute metadata added (source, owner, protection, timestamps) and shown in attribute tooltips.
  * Attribute display now prefers label over name when available.

* **Refactor**
  * Components updated to accept metadata-wrapped object data; attribute values, protection indicators, and relationship lists now read from per-attribute metadata.
  * Relationship rendering updated to handle single vs. multiple relations with metadata-aware handling.

### DIFF
--- a/frontend/app/src/entities/nodes/object/ui/object-details/object-data-display.tsx
+++ b/frontend/app/src/entities/nodes/object/ui/object-details/object-data-display.tsx
@@ -50,38 +50,36 @@ export function ObjectDataDisplay({
               <>
                 <ObjectAttributeValue attributeSchema={attribute} attributeData={attributeData} />
 
-                {attributeData && (
-                  <MetaDetailsTooltip
-                    updatedAt={attributeData.updated_at}
-                    source={attributeData.source}
-                    owner={attributeData.owner}
-                    isFromProfile={attributeData.is_from_profile}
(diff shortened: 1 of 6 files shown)


frontend/app/src/entities/nodes/object/ui/object-details/object-data-display.tsx
```diff
@@ -37,28 +39,24 @@ export function ObjectDataDisplay({
...
             value={
               <>
-                <ObjectAttributeValue
-                  attributeSchema={attribute}
-                  attributeData={objectData[attribute.name] as AttributeType}
-                />
+                <ObjectAttributeValue attributeSchema={attribute} attributeData={attributeData} />
 
-                {objectData[attribute.name] && (
+                {attributeData && (
```
pa-lem: no need to check attributeMetadata since there is an early return

pa-lem (approved):
1 tiny comment but lgtm

frontend/app/src/entities/nodes/object/ui/object-details/object-data-display.tsx
bilalabbad: oh yes good catch there is leftover

PR #7877, opened 2025-12-12
IFC-2111 Support mandatory attributes and relationships in profiles

gmazoyer:
This change add the support for mandatory fields (attributes and relationships) in profiles. This means that if a node is created with a profile that provides values for one or more mandatory fields, the user won't need to provide values for these fields directly.

To allow this behaviour to work, when a node is created or updated, we check if a profile provides any of the mandatory values, and if one does, we do not run the mandatory check constraint against the covered fields.

This change also introduces a new constraint checker to prevent profiles from being unapplied on nodes if the mandatory fields are still filled by those profiles. We intend to make sure that all fields are set by a user or another profile before allowing a profile to be deleted or unapplied.

## Summary by CodeRabbit

* **New Features**
  * Profiles can satisfy mandatory node attributes and relationships at creation.
  * New read path to locate nodes by human-friendly IDs (HFIDs).
  * IP address/prefix fields added to IP-related profile and core types.

* **Bug Fixes / Safety**
  * Prevents profile removals that would break nodes relying on inherited attributes/relationships.
  * Improved batch-safe relationship replace/update when profiles change.

* **Tests**
  * Expanded unit, integration, and functional tests covering profiles, mandatory fields, HFID lookups, and removal constraints.


backend/infrahub/core/query/node.py
```diff
@@ -1407,6 +1409,57 @@ async def get_nodes(self, db: InfrahubDatabase, duplicate: bool = False) -> Asyn
...
+        """ % {"branch_filter": branch_filter, "node_kind": self.node_kind}
+
+        self.add_to_query(query)
+        self.return_labels = ["n.uuid AS node_uuid", "av.value AS hfid"]
+
+    def get_node_uuids(self) -> list[str]:
+        """Get the list of node UUIDs from the query results."""
+        return [result.get_as_type(label="node_uuid", return_type=str) for result in self.get_results()]
+
+
```
gmazoyer, line 1475: Not sure I need this or if I can re-use an existing query

backend/infrahub/core/node/__init__.py
ajtmccarty, line 1: this might be the best approach given the current state of `Node` CRUD logic, but I think it is another item on the list of things to remove from the `Node` objects if we ever actually get around to refactoring them and `NodeManager`

backend/infrahub/core/query/node.py
```diff
@@ -1407,6 +1409,57 @@ async def get_nodes(self, db: InfrahubDatabase, duplicate: bool = False) -> Asyn
...
+        self.hfids = hfids
+        super().__init__(**kwargs)
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_params)
+        # The list is stored as a string in the database
+        self.params["hfid_values"] = [ujson.dumps(hfid) for hfid in self.hfids]
+
+        query = """
```
ajtmccarty, line 1445: Ideally, we would do this in reverse: start with a match on `AttributeValue.value IN `$hfid_values` and then go up to the `Node` from there b/c there are almost always going to be fewer matches for a specific HFID than all Nodes of a given kind.
But we cannot do that b/c hfids are stored as a `List` attribute and we do not index those. So if this is too slow, then we will need to come up with a fix to index HFIDs, which would probably not be too hard, but it would not be trivial

backend/infrahub/core/relationship/constraints/profiles_removal.py
```diff
@@ -0,0 +1,168 @@
...
+        self, relm: RelationshipManager, node_schema: NodeSchema, node: Node
+    ) -> None:
+        required_attr_names = self._get_required_attributes_names(schema=node_schema)
+        required_rel_names = self._get_required_relationship_names(schema=node_schema)
+        if not required_attr_names and not required_rel_names:
+            return
+
+        relm_update_details = await relm.fetch_relationship_ids(db=self.db, force_refresh=False)
+        if not relm_update_details.peer_ids_present_database_only:
+            return
```
ajtmccarty, line 90: nice early returns 👍

backend/tests/integration_docker/test_profiles.py
```diff
@@ -86,28 +87,32 @@ async def wait_for_relationship_peers(
...
+        manufacturer = await self._create_manufacturer(client=client, name="Manufacturer-attr-update")
+        device = await client.create(
+            kind="TestingDevice", name="device-1", manufacturer=manufacturer, part_number="MF-PN-001"
+        )
         await device.save()
 
         device_initial = await client.get(kind="TestingDevice", id=device.id)
         assert device_initial.height.value is None
         assert not device_initial.height.is_from_profile
-        assert device_initial.part_number.value is None
```
ajtmccarty, line 102: it seems like this test should have continued to work without removing these `part_number` checks

backend/tests/integration_docker/test_profiles.py
```diff
@@ -245,16 +254,17 @@ async def test_remove_node_from_profile_triggers_refresh(self, client: InfrahubC
         assert not device_after_removal.height.is_from_profile
 
     async def test_profile_relationship_cardinality_one_update(self, client: InfrahubClient) -> None:
-        manufacturer1 = await client.create(kind="TestingManufacturer", name="Manufacturer-1", country="France")
```
ajtmccarty, line 248: this does not appear to test that the manufacturer update on the profile is passed on to the device any longer

backend/tests/integration_docker/test_profiles.py
```diff
@@ -333,12 +341,16 @@ async def test_profile_relationship_cardinality_many_update(self, client: Infrah
         )
 
     async def test_profile_relationship_priority(self, client: InfrahubClient) -> None:
```
ajtmccarty: this doesn't test that the profile priority updates affect the device using the profiles

backend/tests/unit/core/constraint_validators/test_relationship_profiles_removal.py
```diff
@@ -0,0 +1,520 @@
...
+    await load_schema(db=db, schema=SchemaRoot(nodes=[CHILD, THING]), branch_name=branch.name)
+    profile_schema = registry.schema.get_profile_schema(name=f"Profile{TestKind.THING}", branch=branch, duplicate=False)
+
+    profile = await NodeManager.get_one(db=db, branch=branch, id=profile.id)
+    await profile.related_nodes.resolve(db=db)
+    await profile.related_nodes.update(db=db, data=[])
+
+    constraint = RelationshipProfileRemovalConstraint(db=db, branch=branch)
+    with pytest.raises(ValidationError) as exc:
+        await constraint.check(relm=profile.related_nodes, node_schema=profile_schema, node=profile)
```
ajtmccarty, line 244: I think the constraint uses a different method for validating changes to profiles: `validate_profile_deletion`

backend/tests/unit/core/constraint_validators/test_relationship_profiles_removal.py
```diff
@@ -0,0 +1,520 @@
...
+    updated_fields = await node_applier.apply_profiles(node=thing)
+    assert "owner" not in updated_fields
+    await thing.save(db=db)
+
+    profile = await NodeManager.get_one(db=db, branch=branch, id=profile.id)
+    await profile.related_nodes.resolve(db=db)
+    await profile.related_nodes.update(db=db, data=[])
+
+    constraint = RelationshipProfileRemovalConstraint(db=db, branch=branch)
+    await constraint.check(relm=profile.related_nodes, node_schema=profile_schema, node=profile)
```
ajtmccarty: same here

backend/tests/unit/core/constraint_validators/test_relationship_profiles_removal.py
```diff
@@ -0,0 +1,520 @@
...
+    assert "color" in updated_fields
+    await thing.save(db=db)
+
+    thing_required_color = copy.deepcopy(THING)
+    thing_required_color.relationships[0].optional = True
+    await load_schema(db=db, schema=SchemaRoot(nodes=[CHILD, thing_required_color]), branch_name=branch.name)
+    thing_schema = registry.schema.get_node_schema(name=TestKind.THING, branch=branch, duplicate=False)
+
+    thing = await NodeManager.get_one(db=db, branch=branch, id=thing.id)
+    await thing.color.from_graphql(data={"value": "user-set-green"}, db=db)
```
ajtmccarty: can this just be `thing.color.value = "user-set-green"`?

backend/tests/unit/core/constraint_validators/test_relationship_profiles_removal.py
```diff
@@ -0,0 +1,520 @@
...
+    await load_schema(db=db, schema=SchemaRoot(nodes=[CHILD, thing_required_color]), branch_name=branch.name)
+    profile_schema = registry.schema.get_profile_schema(name=f"Profile{TestKind.THING}", branch=branch, duplicate=False)
+
+    profile = await NodeManager.get_one(db=db, branch=branch, id=profile.id)
+    await profile.related_nodes.resolve(db=db)
+    await profile.related_nodes.update(db=db, data=[])
+
+    constraint = RelationshipProfileRemovalConstraint(db=db, branch=branch)
+    with pytest.raises(ValidationError) as exc:
+        await constraint.check(relm=profile.related_nodes, node_schema=profile_schema, node=profile)
```
ajtmccarty, line 451: `validate_profile_deletion` too and for the next two tests

backend/tests/unit/core/profiles/test_mandatory_fields_checker.py
```diff
@@ -0,0 +1,221 @@
...
+        thing_schema = registry.schema.get_node_schema(name=TestKind.THING, branch=default_branch)
+        profile_schema = registry.schema.get_profile_schema(f"Profile{TestKind.THING}", branch=default_branch)
+
+        profile1 = await Node.init(db=db, branch=default_branch, schema=profile_schema)
+        await profile1.new(db=db, profile_name="profile1", color="red", profile_priority=1000)
+        await profile1.save(db=db)
+
+        profile2 = await Node.init(db=db, branch=default_branch, schema=profile_schema)
+        await profile2.new(db=db, profile_name="profile2", profile_priority=2000)
+        await profile2.save(db=db)
```
ajtmccarty, line 191: this should include a mandatory attribute or relationship to really test the aggregation

backend/infrahub/profiles/mandatory_fields_checker.py
```diff
@@ -0,0 +1,108 @@
...
+            ids.append(item)
+        elif isinstance(item, dict):
+            # Dicts with `id` or `hfid` keys
+            if profile_id := item.get("id"):
+                ids.append(profile_id)
+            elif profile_hfid := item.get("hfid"):
+                hfids.append(profile_hfid)
+        elif hasattr(item, "id"):
+            # Avoid circular import by not importing Node directly
+            ids.append(item.id)
```
ajtmccarty, line 43: code smell

backend/infrahub/profiles/mandatory_fields_checker.py
```diff
@@ -0,0 +1,108 @@
...
+
+async def _resolve_hfids_to_ids(
+    db: InfrahubDatabase, branch: Branch, profile_kind: str, hfids: list[list[str]]
+) -> list[str]:
+    query = await NodeGetByHFIDQuery.init(db=db, branch=branch, node_kind=profile_kind, hfids=hfids)
+    await query.execute(db=db)
+    return query.get_node_uuids()
+
+
+async def get_mandatory_fields_from_profiles(
```
ajtmccarty: ideally, this would be its own component, `ProfilesMandatoryFieldGetter` or something, and the `db` and `branch` would be passed in to its constructor

backend/tests/unit/core/profiles/test_node_create_with_profiles.py
```diff
@@ -0,0 +1,174 @@
...
+        profile_schema = registry.schema.get_profile_schema(f"Profile{TestKind.CHILD}", branch=default_branch)
+
+        profile = await Node.init(db=db, branch=default_branch, schema=profile_schema)
+        await profile.new(db=db, profile_name="child_profile", name="from_profile", profile_priority=1000)
+        await profile.save(db=db)
+
+        node = await Node.init(db=db, branch=default_branch, schema=child_schema)
+        await node.new(db=db, profiles=[profile], name="user_provided")
+        await node.save(db=db)
+
```
ajtmccarty, line 77: should `NodeProfilesApplier.apply_profiles` run here?

backend/tests/unit/core/profiles/test_node_create_with_profiles.py
```diff
@@ -0,0 +1,174 @@
...
+        profile = await Node.init(db=db, branch=default_branch, schema=profile_schema)
+        await profile.new(db=db, profile_name="thing_profile", profile_priority=1000, owner=child)
+        await profile.save(db=db)
+
+        node = await Node.init(db=db, branch=default_branch, schema=thing_schema)
+        await node.new(db=db, name="test_thing", color="red", profiles=[profile])
+        await node.save(db=db)
+
+        assert node.id
+        assert node.name.value == "test_thing"
```
ajtmccarty, line 121: should test the `owner` relationship

backend/tests/unit/core/profiles/test_node_create_with_profiles.py
```diff
@@ -0,0 +1,174 @@
...
+        node = await Node.init(db=db, branch=default_branch, schema=thing_schema)
+        await node.new(db=db, name="test_thing", profiles=[profile1, profile2])
+        await node.save(db=db)
+
+        assert node.id
+
+        applier = NodeProfilesApplier(db=db, branch=default_branch)
+        await applier.apply_profiles(node)
+
+        assert node.color.value == "green"
```
ajtmccarty, line 163: should also test the `owner` relationship

ajtmccarty (commented):
It would be good to add a test for the case where a user wants to replace a profile that supplies mandatory attrs/rels with a new profile that supplies those same attrs/rels

backend/infrahub/profiles/node_applier.py
```diff
@@ -173,8 +173,9 @@ async def _apply_profile_to_relationship(
                 )
                 await new_rel.new(db=self.db, data=peer_id)
                 new_rel.set_source(value=profile_id)
-                node_rel._relationships.append(new_rel)
-                is_changed = True
+                relationships_to_replace["insert"].append(new_rel)
+
+        is_changed = await self._update_node_rel_relationships_to_replace(node_rel, relationships_to_replace)
```
ajtmccarty: I think this needs to be `|=`

backend/infrahub/profiles/node_applier.py
```diff
@@ -184,6 +185,43 @@ async def _apply_profile_to_relationship(
...
+                if rel_to_remove.peer_id:
+                    await node_rel.remove_locally(peer_id=rel_to_remove.peer_id, db=self.db)
+            is_changed = True
+        elif len_rels_to_insert > len_rels_to_remove:
+            for rel_to_insert, rel_to_remove in zip(rels_to_insert[:len_rels_to_remove], rels_to_remove, strict=True):
+                node_rel._relationships.replace(rel_to_insert=rel_to_insert, rel_to_remove=rel_to_remove)
+            for rel_to_insert in rels_to_insert[len_rels_to_remove:]:
+                node_rel._relationships.append(rel_to_insert)
+            is_changed = True
+        return is_changed
```
ajtmccarty: I _think_ this is all a workaround for validation on `RelationshipValidatorList`, which prevents us from making changes to its list of Relationships if any single change would violate a min/max count, even if the end result would not violate any constraints.

I think the best solution would be to remove that class and have the constraint validation logic only execute after all changes are made, but that is probably too much for this PR.
Maybe it is possible to add a method like `bulk_update` or something to `RelationshipManager` that allows you to pass in a new list of `Relationship`s and then sets them in a new instance of `RelationshipValidatorList` and then runs validation on it.

I could be missing something, but this method feels like it's doing too much work to avoid updating a class that barely works

backend/tests/integration/schema_lifecycle/test_migration_attribute_kind.py
```diff
@@ -112,8 +112,8 @@ def schema_step_01(
     @pytest.fixture(scope="class")
     def schema_thing_illegal_updates(self, schema_thing: dict[str, Any]) -> dict[str, Any]:
         updated_schema = deepcopy(schema_thing)
-        updated_schema["attributes"][0]["kind"] = "JSON"
-        updated_schema["attributes"][1]["kind"] = "List"
+        updated_schema["attributes"][0]["kind"] = "TextArea"
+        updated_schema["attributes"][1]["kind"] = "Text"
```
ajtmccarty, line 116: why did these need to change? is there a bug in our schema update logic that throws an unhandled error when trying to load a schema with this change?

backend/tests/integration_docker/test_profiles.py
```diff
@@ -273,6 +287,11 @@ async def test_profile_relationship_cardinality_one_update(self, client: Infrahu
             expected_peer_id=manufacturer1.id,
         )
 
+        device_after_profile = await client.get(
+            kind="TestingDevice", id=device.id, property=True, include=["manufacturer"]
+        )
+        assert device_after_profile.manufacturer.id == manufacturer1.id
```
ajtmccarty: this looks like the manufacturer on the profile overrides the manufacturer on the device even though the device already exists with a manufacturer, which is not what I would expect, but I could be missing something

backend/tests/unit/core/constraint_validators/test_relationship_profiles_removal.py
```diff
@@ -0,0 +1,519 @@
...
+    await constraint.check(relm=thing.profiles, node_schema=thing_schema, node=thing)
+
+
+async def test_constraint_blocks_removing_profile_with_inherited_required_relationship(
+    db: InfrahubDatabase, branch: Branch
+) -> None:
+    thing_optional = copy.deepcopy(THING)
+    thing_optional.relationships[0].optional = True
+
+    await load_schema(db=db, schema=SchemaRoot(nodes=[CHILD, thing_optional]), branch_name=branch.name)
```
ajtmccarty: I'd think that you wouldn't need to load this version of the THING schema with an optional relationship and could 
1. create the profile instance
2. create the Thing instance linked to the profile from 1
3. try to remove the profile from the Thing in 2

backend/tests/unit/core/profiles/test_node_applier.py
```diff
@@ -922,3 +923,132 @@ async def test_node_from_template_with_profile_precedence(
...
+                        "cardinality": "many",
+                        "kind": "Generic",
+                    }
+                ],
+            },
+        ],
+    }
+    schema_root = SchemaRoot(**SCHEMA)
+    registry.schema.register_schema(schema=schema_root, branch=default_branch.name)
+    default_branch.update_schema_hash()
```
ajtmccarty: seems like this could use the `car_person_schema` and avoid creating this whole new schema

backend/tests/integration/schema_lifecycle/test_migration_attribute_kind.py
ogenstad: It looks like this change comes from the changes to the attribute class.

```diff
    @property
    def support_profiles(self) -> bool:
-        return self.read_only is False and self.optional is True
+       return self.read_only is False and self.unique is False
```

Because of this change something with the above change fails. If we just revert this part we don't need to update the test as per above and the later line `assert len(error_messages) == 4`.

The actual error we get:

```json
{
    'data': None,
    'errors': [
        {
            'message': "1 validation error for AttributeSchema\n  Value error, TextAttributeParameters can't be used as parameters for JSON [type=value_error, input_value={'optional': True, 'state...e}, 'deprecation': None},
input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.12/v/value_error",
            'extensions': {'code': 422}
        }
    ]
}
```

So it looks like the first attribute is a `Text` kind and then we try to convert it to a `JSON` kind. This fails with this error:

```python
    def validate_parameters(self) -> Self:
        if isinstance(self.parameters, NumberPoolParameters) and not self.kind == "NumberPool":
            raise ValueError(f"NumberPoolParameters can't be used as parameters for {self.kind}")

        if isinstance(self.parameters, TextAttributeParameters) and self.kind not in ["Text", "TextArea"]:
            raise ValueError(f"TextAttributeParameters can't be used as parameters for {self.kind}")

        return self
```

I think there are multiple errors within this schema and that a change triggers some other error to happen. It would be ideal to not change this test, but I wonder if the test itself would need some updates with regards to how this is managed. But the change here would be that previously the call to `.support_profiles` was False and now it's True which leads to another error.

backend/tests/integration_docker/test_profiles.py
ajtmccarty: in its current state, this test is checking that updates to a profile do not override a relationship set on an object using that profile. I think either 

1. the name of this test needs to change to something like `test_profile_relationship_cardinality_one_update_does_not_override`
2. the test should be updated to create the device using a profile to set the manufacturer relationship

there might already be tests for option 2 further down the test file

backend/tests/integration_docker/test_profiles.py
ajtmccarty: same issue here as for `test_profile_relationship_cardinality_one_update`
either the test name should be updated to show this is a negative test. Or, if this is not tested elsewhere in the file, then device needs to be created using the profiles so that it inherits the manufacturer relationship

backend/tests/integration/schema_lifecycle/test_migration_attribute_kind.py
ajtmccarty: after looking some more, the error seems to come from how we create a profile for a Node in `SchemaBranch.generate_profile_from_node`

we use the attribute on the `Node`, which will include parameters specific to that type of attribute that we then pass into the `attr_schema_class` constructor and Pydantic tries to instantiate the correct parameters model based on the shape of the data in `node_attr.model_dump()["parameters"]`
```
            if not node_attr.support_profiles:
                continue
            attr_schema_class = get_attribute_schema_class_for_kind(kind=node_attr.kind)
            attr = attr_schema_class(
                optional=True,
                **node_attr.model_dump(exclude={"id", "unique", "optional", "read_only", "default_value", "inherited"}),
            )
            profile.attributes.append(attr)

```

I made a new task for cleaning this up here #7943
I think we can leave this as is for now, but I'd like to get this cleaned up soon

backend/infrahub/profiles/node_applier.py
ajtmccarty: just want to make sure this one is not forgotten

backend/infrahub/profiles/node_applier.py
```diff
@@ -184,6 +185,43 @@ async def _apply_profile_to_relationship(
 
         return is_changed
 
+    async def _update_node_rel_relationships_to_replace(
+        self, node_rel: RelationshipManager, relationships_to_replace: dict[str, list[Relationship]]
+    ) -> bool:
```
ajtmccarty: - this seems like it should be a method on the RelationshipManager class.
- this should accept two lists of Relationships, `relationships_to_remove` and `relationships_to_insert`, instead of a dictionary with `remove` and `insert` keys

ajtmccarty (approved):

PR #7878, opened 2025-12-12
Add partial_match filter for InfrahubBranch GraphQL query

wvandeun:
## Summary by CodeRabbit

* **New Features**
  * Added optional partial-match filtering for branch name searches (case-insensitive substring support).

* **Documentation**
  * GraphQL schema updated to expose the new partial_match argument on branch queries.

* **Tests**
  * Added end-to-end tests validating partial and exact match behaviors for branch queries.

### DIFF
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -156,9 +156,9 @@ class Branch(StandardNode):
         limit: int = 1000,
         ids: list[str] | None = None,
         name: str | None = None,
+        partial_match: bool = False,
         **kwargs: Any,
     ) -> list[Self]:
-        partial_match: bool = kwargs.pop("partial_match", False)
         query: Query = await BranchNodeGetListQuery.init(
             db=db,
             node_class=cls,
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -158,8 +158,15 @@ class Branch(StandardNode):
         name: str | None = None,
         **kwargs: Any,
     ) -> list[Self]:
+        partial_match: bool = kwargs.pop("partial_match", False)
         query: Query = await BranchNodeGetListQuery.init(
-            db=db, node_class=cls, ids=ids, node_name=name, limit=limit, **kwargs
+            db=db,
+            node_class=cls,
+            ids=ids,
+            node_name=name,
+            limit=limit,
+            partial_match=partial_match,
+            **kwargs,
         )
         await query.execute(db=db)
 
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/branch/models.py
```diff
@@ -158,8 +158,15 @@ async def get_list(
         name: str | None = None,
         **kwargs: Any,
     ) -> list[Self]:
+        partial_match: bool = kwargs.pop("partial_match", False)
```
ajtmccarty: `partial_match: bool = False` in the parameters would do the same thing and be more pythonic

backend/infrahub/core/query/branch.py
```diff
@@ -151,9 +151,19 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
 
 
 class BranchNodeGetListQuery(StandardNodeGetListQuery):
-    def __init__(self, exclude_global: bool = False, **kwargs: Any) -> None:
+    def __init__(self, exclude_global: bool = False, partial_match: bool = False, **kwargs: Any) -> None:
+        name: str | None = None
         self.raw_filter = f"n.status <> '{BranchStatus.DELETING.value}'"
+
+        if partial_match:
+            name = kwargs.pop("node_name", None)
```
ajtmccarty: `node_name` is a parameter of `StandardNodeGetListQuery.__init__`, so we should just let that class handle `node_name`
I think you can remove this `kwargs.pop` by moving the `super().__init__()` up above this `partial_match` boolean

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -387,3 +391,72 @@ async def test_paginated_branch_query_meta_data(
...
+        "search_term,partial_match,expected_count",
+        [
+            ("match", True, 10),
+            ("branch-zzz", True, 0),
+            ("auth", False, 0),
+            ("main", False, 1),
+            ("MAIN", False, 0),
+            ("MaTcH", True, 10),
+        ],
+    )
```
ajtmccarty: as this test it written, it will run 6 times and create 10 branches each time
I think you can accomplish the same thing by just creating 10 (or really 4 or 5 branches) and then run the 6 partial_match graphql queries with their various parameters

backend/tests/unit/graphql/conftest.py
```diff
@@ -394,3 +394,21 @@ async def account_bill(db: InfrahubDatabase, default_branch: Branch) -> Node:
...
+        edges {
+          node {
+            name {
+              value
+            }
+          }
+        }
+      }
+    }
+    """
```
ogenstad: Can you move this fixture so that the query is defined as a variable in the test_branch.py file instead? Mainly to make it easier to read the conftest.py files (we have a lot of cleanup that we need to do in those at some point.

ogenstad (approved):
LGTM!

PR #7879, opened 2025-12-12
Refactor to avoid having separate metadata_field dict

ogenstad:
Refactoring based on this comment: https://github.com/opsmill/infrahub/pull/7846#discussion_r2612124683 I did some additional refactoring to compose the metadata in fewer locations and as such the change was a bit larger than the initial scope but I hope it leads to an improvement overall.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated metadata handling into a single include_metadata flow and removed per-call metadata field parameters.
  * Updated loaders and resolvers to consistently use the new metadata query options.
  * Made metadata option objects immutable to improve predictability for caching and hashing.

* **New Features**
  * Centralized metadata option construction for GraphQL requests, enabling unified node and relationship metadata composition.


ogenstad:
@coderabbitai review

ogenstad:
@coderabbitai review

backend/infrahub/graphql/metadata.py
```diff
@@ -0,0 +1,105 @@
...
+        if not field_properties_dict or not isinstance(field_properties_dict, dict):
+            continue
+
+        # For attributes, metadata fields are at the same level as 'value'
+        # Check if this looks like an attribute (has metadata fields directly)
+        metadata_properties_dict = field_properties_dict
+
+        # Try relationship structures - if found, skip (handled elsewhere)
+        is_relationship = False
+        with contextlib.suppress(KeyError, TypeError):
```
ajtmccarty: I don't think we need this any longer

ajtmccarty (approved):
👏 👏 👏 
very nice. this is much better

backend/infrahub/graphql/metadata.py
ogenstad: Ah, yes good catch too much copy and paste. I removed that part.

PR #7885, opened 2025-12-13
handle updated/created_at/by metadata in schema migrations

ajtmccarty:
IFC-2091

- update all of our schema migrations to include handling for updating updated/created_at/by metadata correctly
- pass `user_id` all over the place in the schema migrations
- a few other small cypher changes fix bugs identified during testing
- a bunch of new tests and test updates

I think this PR is big enough already, so the code to pass the user_id down into the migrations will come in a separate PR

## Release Notes

* **Bug Fixes**
  * Improved metadata tracking for creation and modification timestamps during system migrations and operations.
  * Enhanced user attribution to properly record who performed system operations.
  * Fixed metadata propagation for node relationships and attributes during migrations.

### DIFF
--- a/backend/infrahub/core/migrations/graph/m018_uniqueness_nulls.py
+++ b/backend/infrahub/core/migrations/graph/m018_uniqueness_nulls.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING, Sequence
 
 from infrahub.core import registry
+from infrahub.core.constants import SYSTEM_USER_ID
 from infrahub.core.diff.payload_builder import get_display_labels_per_kind
 from infrahub.core.migrations.shared import MigrationResult
 from infrahub.core.schema import GenericSchema, NodeSchema, SchemaRoot, internal_schema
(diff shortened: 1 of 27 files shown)


backend/infrahub/core/query/__init__.py
```diff
@@ -367,6 +368,7 @@ def __init__(
         self.limit = limit
         self.offset = offset
         self.order_by = order_by
+        self.user_id = user_id
```
ajtmccarty, line 371: add `user_id` as an optional parameter at the lowest level to avoid screwups with the MRO when instantiating classes with multiple levels of inheritance between them and `Query`

backend/infrahub/core/migrations/graph/m018_uniqueness_nulls.py
```diff
@@ -95,5 +96,5 @@ class Migration018(InternalSchemaMigration):
     async def validate_migration(self, db: InfrahubDatabase) -> MigrationResult:  # noqa: ARG002
         return MigrationResult()
 
-    async def execute(self, db: InfrahubDatabase) -> MigrationResult:
+    async def execute(self, db: InfrahubDatabase, user_id: str = SYSTEM_USER_ID) -> MigrationResult:  # noqa: ARG002
```
ogenstad, line 99: Not for now but with this in mind and for future changes we could consider having some upgrade package that we include as an argument to execute where the database would be one attribute and the user_id would be an optional one would also take care of the ignore to the type.

ogenstad (approved):

PR #7895, opened 2025-12-15
Add account loader for metadata

ogenstad:
This PR adds an account loader specifically for metadata fields for created_by or updated_by. It will allow for lookups mainly for display labels for accounts within the metadata parts of the GraphQL queries.

It also adds a number of missing tests for previous PRs (when we weren't able to query and get back the correct information for metadata responses within the queries).

## Summary by CodeRabbit

* **New Features**
  * Account metadata fields (created_by, updated_by) now resolve to full account objects including display label and type info.

* **Improvements**
  * Batched, per-request loading for account metadata for better query performance and caching.
  * Consistent metadata shape across nodes, attributes, and relationships, with explicit handling for system and unknown accounts.

* **Tests**
  * Added comprehensive unit and functional tests covering resolution, batching, caching, and edge cases.

### DIFF
--- a/backend/infrahub/graphql/loaders/account.py
+++ b/backend/infrahub/graphql/loaders/account.py
@@ -1,0 +1,134 @@
+"""DataLoader for resolving account IDs to full account data."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any
+
+from aiodataloader import DataLoader
+
+from infrahub.core.constants import SYSTEM_USER_ID, InfrahubKind
+from infrahub.core.manager import NodeManager
+from infrahub.graphql.constants import KIND_GRAPHQL_FIELD_NAME
+
+if TYPE_CHECKING:
+    from infrahub.core.branch.models import Branch
+    from infrahub.core.timestamp import Timestamp
+    from infrahub.database import InfrahubDatabase
+
+SYSTEM_ACCOUNT_DISPLAY_LABEL = "[Infrahub System]"
+UNKNOWN_ACCOUNT_DISPLAY_LABEL = "[Unknown Account]"
+
+
+@dataclass
+class AccountLoaderParams:
+    branch: Branch
+    at: Timestamp | None
+    fields: dict[str, Any] = field(default_factory=dict)
+
+    def __hash__(self) -> int:
+        # Fields are not included in hash as they vary per query but we want to reuse loader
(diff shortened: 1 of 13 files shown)


backend/infrahub/graphql/loaders/account.py
```diff
@@ -0,0 +1,134 @@
...
+        real_account_ids = [k for k in keys if k != SYSTEM_USER_ID]
+
+        # Batch load real accounts
+        accounts_by_id: dict[str, dict[str, Any]] = {}
+        if real_account_ids:
+            async with self.db.start_session(read_only=True) as db:
+                nodes = await NodeManager.get_many(
+                    db=db,
+                    ids=real_account_ids,
+                    fields=self.params.fields,
```
ajtmccarty, line 111: there might be an issue with including the `fields` in here
if two different parts of the same GraphQL query use this loader and request different fields for the same user ID, then I think they will both end up with the same response instead of different responses for the different sets of fields. it shouldn't be too hard to add/update a test to confirm it.

I'm not sure that best way to handle the response though. It would be nice to have this return a list of `CoreAccount` objects, but that won't work with the system/unknown users. maybe this could return the full `id`, `display_label`, `__typename`, `name`, `label`, `description` dict here every time and then it would be pared down to just the requested fields somewhere within graphene

backend/tests/unit/graphql/loaders/test_account.py
```diff
@@ -0,0 +1,246 @@
...
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema,
+) -> None:
+    """Verify AccountLoaderParams hash is based on branch and timestamp, not fields.
+
+    The hash should be consistent across different field selections so that
+    loaders with the same branch/timestamp can be reused.
+    """
+    from infrahub.core.timestamp import Timestamp
```
ajtmccarty: can this import go to the top where it belongs?

ajtmccarty (approved):

backend/tests/unit/graphql/loaders/test_account.py
ogenstad: Yes of course.

backend/infrahub/graphql/loaders/account.py
ogenstad: Good point, I've added some tests to highlight that alternate lookups work.

PR #7904, opened 2025-12-16
Add GraphQL metadata sort options

ogenstad:
Add sort options for metadata to GraphQL layer, fully implemented for the standard nodes (i.e. only InfrahubBranch) and just included the API bindings for the dynamically defined schema nodes.

Example query for InfrahubBranch:

```graphql
        query {
            InfrahubBranch(order: {node_metadata: {updated_at: DESC}}) {
                edges {
                    node_metadata { updated_at }
                    node { name { value } description { value } }
                }
            }
        }
```

In this current implementation I've forbidden anyone from combining updated_at with created_at, specifically for these two fields I don't think it matters much as there won't be much overlap with anything that has the same timestamp. However it might come into play at some later time for other fields. It seems like the order of what the user requests (i.e. if created_at is defined before updated_at) is kept within the info object but not in the created models. At some later point we could analyze the actual query and retain the internal order, for now we raise an error if someone specifies both updated_at and created_at)

I've added the same "order" part to the other dynamically created options, but those doesn't currently do anything, they are just a dummy entry so that the frontend team can experiment with what it will look like for the other node types.

## Summary by CodeRabbit

* **New Features**
  * Clients can request ordered node/branch lists (by id, created_at, updated_at) with ASC/DESC via new GraphQL Order input and types. Ordering is validated (mutually exclusive timestamp fields).
* **Tests**
  * Added GraphQL tests verifying ordering behavior (asc/desc) and invalid ordering combinations.

### DIFF
--- a/backend/infrahub/core/node/standard.py
+++ b/backend/infrahub/core/node/standard.py
@@ -10,6 +10,7 @@ import ujson
 from infrahub_sdk.uuidt import UUIDT
 from pydantic import BaseModel, Field, field_validator
 
+from infrahub.constants.enums import OrderByField, OrderDirection
 from infrahub.core.constants import NULL_VALUE, SYSTEM_USER_ID, InfrahubKind
 from infrahub.core.query.standard_node import (
     StandardNodeCreateQuery,
@@ -29,6 +30,12 @@ if TYPE_CHECKING:
 
     from infrahub.core.query import Query
     from infrahub.database import InfrahubDatabase
+
+
+@dataclass
+class StandardNodeOrdering:
+    order_by: OrderByField = field(default=OrderByField.ID)
+    direction: OrderDirection = field(default=OrderDirection.ASC)
 
 
 @dataclass(slots=True)
@@ -296,10 +303,12 @@ class StandardNode(BaseModel):
         limit: int = 1000,
         ids: list[str] | None = None,
         name: str | None = None,
+        node_ordering: StandardNodeOrdering | None = None,
         **kwargs: dict[str, Any],
     ) -> list[Self]:
+        node_ordering = node_ordering or StandardNodeOrdering()
         query: Query = await StandardNodeGetListQuery.init(
-            db=db, node_class=cls, ids=ids, node_name=name, limit=limit, **kwargs
+            db=db, node_class=cls, ids=ids, node_name=name, limit=limit, node_ordering=node_ordering, **kwargs
         )
         await query.execute(db=db)
 
--- a/backend/infrahub/core/query/standard_node.py
+++ b/backend/infrahub/core/query/standard_node.py
@@ -1,14 +1,15 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, assert_never
 
+from infrahub.constants.enums import OrderByField
 from infrahub.core.query import Query, QueryType
 from infrahub.exceptions import InitializationError
 
 if TYPE_CHECKING:
     from uuid import UUID
 
-    from infrahub.core.node.standard import StandardNode
+    from infrahub.core.node.standard import StandardNode, StandardNodeOrdering
     from infrahub.database import InfrahubDatabase
 
 
@@ -137,6 +138,7 @@ class StandardNodeGetListQuery(Query):
     def __init__(
         self,
         node_class: StandardNode,
+        node_ordering: StandardNodeOrdering,
         ids: list[str] | None = None,
         node_name: str | None = None,
         partial_match: bool = False,
@@ -146,6 +148,7 @@ class StandardNodeGetListQuery(Query):
         self.node_name = node_name
         self.node_class = node_class
         self.partial_match = partial_match
+        self.node_ordering = node_ordering
 
         super().__init__(**kwargs)
 
@@ -178,4 +181,12 @@ class StandardNodeGetListQuery(Query):
         self.add_to_query(query)
 
         self.return_labels = ["n"]
-        self.order_by = [f"{db.get_id_function_name()}(n)"]
+        match self.node_ordering.order_by:
+            case OrderByField.ID:
+                self.order_by = [f"{db.get_id_function_name()}(n)"]
+            case OrderByField.CREATED_AT:
+                self.order_by = [f"n.created_at {self.node_ordering.direction.value}"]
+            case OrderByField.UPDATED_AT:
+                self.order_by = [f"n.updated_at {self.node_ordering.direction.value}"]
+            case _:
+                assert_never(self.node_ordering.order_by)
--- a/backend/tests/unit/graphql/queries/test_branch.py
+++ b/backend/tests/unit/graphql/queries/test_branch.py
@@ -480,3 +480,287 @@ class TestBranchQuery(TestInfrahubApp):
             assert result.errors is None
             assert result.data
             assert result.data["InfrahubBranch"]["count"] == test_case.expected_count
+
+    async def test_order_by_created_at_ascending(
+        self,
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        service: InfrahubServices,
+        session_admin: AccountSession,
+    ) -> None:
+        """Test that branches can be ordered by created_at in ascending order."""
+        branch_names = ["alpha-branch", "beta-branch", "gamma-branch"]
+        for branch_name in branch_names:
+            create_branch_query = """
+            mutation($branch_name: String!) {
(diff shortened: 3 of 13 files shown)


backend/infrahub/core/node/standard.py
```diff
@@ -29,6 +30,48 @@
 
     from infrahub.core.query import Query
     from infrahub.database import InfrahubDatabase
+    from infrahub.graphql.types.metadata import OrderInput
```
ajtmccarty: I think we should almost never import from something from the graphql layer into core. the core layer should not require any knowledge of things at the graphql level.

backend/infrahub/core/node/standard.py
```diff
@@ -29,6 +30,48 @@
...
+        if order is None:
+            return cls(order_by=OrderByField.ID)
+
+        node_metadata = getattr(order, "node_metadata", None)
+
+        if node_metadata is None:
+            return cls(order_by=OrderByField.ID)
+
+        created_at = getattr(node_metadata, "created_at", None)
+        updated_at = getattr(node_metadata, "updated_at", None)
```
ajtmccarty: do we really need to user `getattr` here?

backend/infrahub/core/node/standard.py
```diff
@@ -29,6 +30,48 @@
...
+            order: Optional ordering specification from GraphQL input.
+
+        Returns:
+            StandardNodeOrdering with the specified field and direction, or defaults to ID with no direction.
+
+        Raises:
+            ValidationError: If both created_at and updated_at are specified.
+        """
+        if order is None:
+            return cls(order_by=OrderByField.ID)
```
ajtmccarty: looks like the default is already set in the dataclass definition. defining it in multiple places will make it harder to change later.

same with the return at the end

backend/infrahub/core/query/standard_node.py
```diff
@@ -178,4 +181,12 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
-        self.order_by = [f"{db.get_id_function_name()}(n)"]
+        match self.node_ordering.order_by:
+            case OrderByField.ID:
+                self.order_by = [f"{db.get_id_function_name()}(n)"]
+            case OrderByField.CREATED_AT:
+                self.order_by = [f"n.created_at {self.node_ordering.direction.value}"]
+            case OrderByField.UPDATED_AT:
+                self.order_by = [f"n.updated_at {self.node_ordering.direction.value}"]
+            case _:
+                assert_never(self.node_ordering.order_by)
```
ajtmccarty, line 192: interesting. learned something new today

backend/infrahub/core/query/standard_node.py
```diff
@@ -178,4 +181,12 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
         self.add_to_query(query)
 
         self.return_labels = ["n"]
-        self.order_by = [f"{db.get_id_function_name()}(n)"]
+        match self.node_ordering.order_by:
+            case OrderByField.ID:
+                self.order_by = [f"{db.get_id_function_name()}(n)"]
```
ajtmccarty, line 186: ID usually means `Node.uuid` for us, not the database's `elementId`

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -480,3 +480,299 @@ async def test_partial_match_filter(
...
+
+    async def test_order_by_created_at_descending(
+        self,
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        service: InfrahubServices,
+        session_admin: AccountSession,
+    ) -> None:
+        """Test that branches can be ordered by created_at in descending order."""
+        branch_names = ["delta-branch", "epsilon-branch", "zeta-branch"]
```
ajtmccarty, line 549: fancy

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -480,3 +480,299 @@ async def test_partial_match_filter(
...
+        assert result.errors is None
+        assert result.data
+        edges = result.data["InfrahubBranch"]["edges"]
+
+        # Verify ordering by checking timestamps are descending
+        timestamps = [edge["node_metadata"]["created_at"] for edge in edges]
+        assert timestamps == sorted(timestamps, reverse=True), "Branches should be ordered by created_at descending"
+
+        # The last branch created should be first
+        assert edges[0]["node"]["name"]["value"] == "zeta-branch"
```
ajtmccarty, line 597: might as well check that all three are here in the expected order

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -480,3 +480,299 @@ async def test_partial_match_filter(
...
+                schema=gql_params.schema,
+                source=create_branch_query,
+                context_value=gql_params.context,
+                root_value=None,
+                variable_values={"branch_name": branch_name},
+            )
+            assert result.errors is None
+
+        # Update branches in a specific order: iota first, then theta, then eta
+        # This means eta will have the latest updated_at
```
ajtmccarty: sometimes these comments feel like they are directed at an idiot

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -480,3 +480,299 @@ async def test_partial_match_filter(
...
+        edges = result.data["InfrahubBranch"]["edges"]
+
+        # Verify ordering by checking timestamps are ascending
+        timestamps = [edge["node_metadata"]["created_at"] for edge in edges]
+        assert timestamps == sorted(timestamps), "Branches should be ordered by created_at ascending"
+
+        # The first should be main (default branch, created first)
+        assert edges[0]["node"]["name"]["value"] == "main"
+
+    async def test_order_by_created_at_descending(
```
ajtmccarty, line 541: feels like this could be combined with the ascending version of the test

backend/tests/unit/graphql/queries/test_branch.py
```diff
@@ -480,3 +480,299 @@ async def test_partial_match_filter(
...
+
+        # Verify ordering by checking timestamps are ascending
+        timestamps = [edge["node_metadata"]["updated_at"] for edge in test_edges]
+        assert timestamps == sorted(timestamps), "Branches should be ordered by updated_at ascending"
+
+        # Verify the expected order of our updated branches
+        result_branch_names = [e["node"]["name"]["value"] for e in test_edges]
+        assert result_branch_names == update_order, "Updated branches should appear in update order (ascending)"
+
+    async def test_order_by_updated_at_descending(
```
ajtmccarty, line 669: also feels like it can be combined

ajtmccarty (approved):

backend/infrahub/core/node/standard.py
ogenstad: Yes good point, this refactoring when combining the order for standard and dynamic nodes got a bit rushed. Will clean this up.

backend/infrahub/core/node/standard.py
ogenstad: Unfortunately yes we do at this point. It's to make the type checker happy.

```python
class OrderInput(InputObjectType):
    disable = Boolean(required=False)
    node_metadata = Field(InfrahubNodeMetadataOrder, required=False, description="Order settings for branch metadata")
```

The `node_metadata` attribute is defined as a `Field` and not what it actually is `InfrahubNodeMetadataOrder`.  In some places we just ignore this. We might have to update these definitions in the future to have a cleaner internal API. I agree it's not pretty. :(

backend/infrahub/core/node/standard.py
ogenstad: Yes good point.

backend/infrahub/core/query/standard_node.py
ogenstad: It's something that I noticed that they have in Rust and I thought it was a nice thing to check against so when looking for an alternative in Python I came across this. I asked Lucas to add it in one other place as well but I think it's something that we can probably use in more places moving forward.

backend/infrahub/core/query/standard_node.py
ogenstad: I will leave this as is for now since I didn't want to change the existing default behaviour. But we can backtrack later and switch this to the uuid instead. What do you think?

ajtmccarty: I missed it was already the default behavior
this is fine then

ajtmccarty (approved):

PR #7908, opened 2025-12-17
fix Migration041

ajtmccarty:
fix for Migration041.

added comments describing what each line change in Migration041 does, but basically 
1. we were not using enough information to determine if an edge was a duplicate of another edge. so some non-duplicate edges were erroneously deleted
2. we constrained our search space to only edges leaving a Relationship vertex, so duplicated edges entering a Relationship vertex were not deleted

the Migration has been updated to remove these problems.

this adds a new migration that runs the updated query from Migration041 on ALL Relationship vertices instead of just those linked to Nodes with a migrated kind/inheritance. the new migration also looks at Relationships missing IS_VISIBLE and IS_PROTECTED edges (some of these were erroneously deleted) and recreates them

the unit test involves a lot of custom cypher to duplicate the messed up database state that the migration fixes

I've tested on user backup for correctness and performance and everything looks good

## Summary by CodeRabbit

* **Bug Fixes**
  - Resolved missing relationship property metadata in graph database
  - Improved duplicate edge detection and removal during migrations

* **Chores**
  - Updated graph database version to 48
  - Enhanced database migration framework


backend/infrahub/core/migrations/graph/m041_deleted_dup_edges.py
```diff
@@ -87,14 +87,15 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
     // ------------
     MATCH (n:Node {uuid: node_uuid})-[:IS_RELATED]-(rel:Relationship)
     WITH DISTINCT rel
-    MATCH (rel)-[e]->(peer)
+    MATCH (rel)-[e]-(peer)
```
ajtmccarty, line 105: the duplication check logic in the `WITH` statement below includes a column for `direction`, so we can check both directions in this `MATCH`
checking only one direction will leave the other (i.e. `(rel)<-[e]-(peer)`) unchecked, so it would not be deduplicated

backend/infrahub/core/migrations/graph/m041_deleted_dup_edges.py
```diff
@@ -87,14 +87,15 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
     WITH
+        elementId(rel) AS rel_element_id,
```
ajtmccarty, line 107: we must consider include the unique `Relatioship` vertex when determining if an edge is duplicated

backend/infrahub/core/migrations/graph/m041_deleted_dup_edges.py
```diff
@@ -87,14 +87,15 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
         type(e) AS e_type,
         e.branch AS e_branch,
         e.from AS e_from,
         e.to AS e_to,
         e.status AS e_status,
-        e.peer AS e_peer,
+        elementId(peer) AS peer_element_id,
```
ajtmccarty, line 113: `e.peer` would have always been NULL. we also need to include a unique identifier for the `peer` vertex when determining if an edge is duplicated

ogenstad (approved):

PR #7913, opened 2025-12-17
Add GraphQL filters for metadata

ogenstad:
Exposes the filtering options for metadata in the GraphQL schema. Actual implementation will be done later in the NodeManager and query engine.

## New Features

* Added metadata filtering for query results, enabling searches by creation and modification timestamps (exact date, before, or after), as well as filtering by creator and modifier users using single or multiple user IDs.


backend/infrahub/graphql/manager.py
```diff
@@ -1047,6 +1050,47 @@ def generate_filters(
...
+            ),
+            # DateTime-based filters (updated_at)
+            "node_metadata__updated_at": graphene.DateTime(description="Filter by exact update timestamp"),
+            "node_metadata__updated_at__before": graphene.DateTime(
+                description="Filter for objects updated before this timestamp"
+            ),
+            "node_metadata__updated_at__after": graphene.DateTime(
+                description="Filter for objects updated after this timestamp"
+            ),
+        }
```
ajtmccarty, line 1092: judging from the description in the epic, these should also support a Date as well as a DateTime, but maybe those are the same thing in `graphene`

ogenstad: They are not exactly the same but the main difference should be that Date is more restrictive and only supports a date where as DateTime would accept an ISO timestamp that could be a timestamp or just a date without time. For input I think DateTime is what we want.

ajtmccarty (approved):

PR #7918, opened 2025-12-17
fix(backend): use new algorithm and method for resource manager allocations

### LINKED ISSUE #7897: bug: Resource pool allocation times out with large IPv6 prefix pools causing gunicorn worker crash
## Component

- API Server / GraphQL

## Infrahub version

1.6.1

## Current Behavior

When allocating resources from a "large" IPv6 prefix pool, the server times out and a gunicorn worker dies.

**Failing scenarios:**
- A `/48` prefix pool used to create `/127` prefixes → timeout + worker crash
- 2x `/56` prefix pools used to create `/127` prefixes → timeout + worker crash

**Working scenarios:**
- A `/64` prefix pool used to create `/127` prefixes → works fine
- 1x `/48` prefix pool used to create `/64` prefixes → works fine

The issue appears to be related to the size of the resource pool and/or the number of potential allocations that need to be computed.

## Expected Behavior

Resource allocation from large prefix pools should complete successfully without timing out or crashing the gunicorn worker, even if it takes longer to process.

## Steps to Reproduce

1. Create an IPv6 prefix pool with a `/48` prefix as the resource
2. Attempt to allocate `/127` prefixes from this pool
3. Observe timeout and gunicorn worker crash

## Additional Information

- A database backup has been provided to the engineering team for reproduction
- The issue seems to scale with the potential allocation space (e.g., `/48` → `/127` = 2^79 possible allocations vs `/64` → `/127` = 2^63 possible allocations)
- May require an investigation into how the resource pool allocation algorithm handles very large address spaces

fatih-acar:
Fixes #7897 

Makes IPv6 allocation in some real world use cases 500x faster (100ms vs 50s).
Performance benchmark on the "Add a new site" test case (which allocates v4 & v6 addresses a long with cable numbers) with large instances show a 2x speed up.
No improvement has been measured on small sized instances, but the Cypher queries work with small instances' memory constraints.

TODO: add more unit tests, especially to cover fragmentation of IPv6 prefix pools, but also fragmentation of all other resource pools.

## Design Goals

The primary goal of these allocation algorithms is to provide a **CPU and memory constrained solution** that scales efficiently. Key design decisions:

- **Database-side computation**: All gap detection, range calculations, and candidate selection are performed within Neo4j using Cypher queries
- **Minimal Python memory footprint**: Rather than fetching all allocated resources into Python and computing gaps in memory, the queries return only the next available value
- **Scalability**: This approach allows pools to contain millions of allocations without bloating Python's memory or causing CPU spikes during allocation requests
- **Single-value retrieval**: Queries are optimized to find and return a single free resource, avoiding the need to materialize large result sets

By pushing computation to the database layer, we ensure that allocation performance remains consistent regardless of pool utilization.

---

## Table of Contents

- [Number Pool Allocation](#number-pool-allocation)
- [IP Address Pool Allocation](#ip-address-pool-allocation)
- [IP Prefix Pool Allocation](#ip-prefix-pool-allocation)
- [IPv4 vs IPv6: The 128-bit Challenge](#ipv4-vs-ipv6-the-128-bit-challenge)

---

### Overview

Number pools manage integer allocations within a defined range (`start_range` to `end_range`). The algorithm finds the first available number by detecting gaps in the sequence of reserved values.

### Algorithm

1. **Query Reserved Values**: Fetch all reserved values from the database within the pool's range
2. **Sort and Prepend Sentinel**: Sort reserved values and prepend `start_range - 1` as a sentinel value
3. **Gap Detection**: Iterate through the sorted list comparing each value with its expected position:
   - If `expected ≠ current`, a gap exists → return `expected` as free
   - If at the last element and `value < end_range`, return `value + 1`
4. **Constraint Filtering**: Apply attribute parameter constraints (excluded values, excluded ranges, min/max)

### Cypher Query Logic (Simplified)

```cypher
WITH [$start_range - 1] + collect(used_value) AS nums
UNWIND range(0, size(nums) - 1) AS idx
WITH nums[idx] AS curr, idx - 1 + $start_range AS expected
RETURN expected AS number, expected <> curr AS is_free
```

### Exclusion Handling

Number pools support exclusion constraints defined in `NumberAttributeParameters`:

- **Single Excluded Values**: Individual numbers that cannot be allocated
- **Excluded Ranges**: Contiguous ranges of numbers to skip

The `get_next()` method iteratively:
1. Computes the effective range (pool range constrained by min/max parameters)
2. Skips past excluded values/ranges from the starting point
3. Queries for the next free number
4. Validates the candidate against exclusions; if excluded, advances and re-queries

---

### Overview

IP Address pools allocate individual IP addresses from one or more prefix resources. Resources are processed in order of their `allocation_weight` (highest first).

### Algorithm

1. **Resource Iteration**: Iterate through weighted prefix resources
2. **Address Discovery**: For each prefix, query for the next free IP address
3. **Gap Detection**: Use the same gap-detection pattern as number pools, but operating on binary address representations
4. **Pool vs Non-Pool Semantics**:
   - **Non-pool prefixes**: Exclude network address (first) and broadcast address (last)
   - **Pool prefixes**: All addresses including network/broadcast are allocatable

### Binary Address Storage

IP addresses are stored as binary strings for efficient prefix matching and ordering:

```python
# IPv4: 32-character string
"10.2.5.2" → "00001010000000100000010100000010"

# IPv6: 128-character string  
"2001:db8::1" → "00100000000000010000110110111000...0001"
```

### Query Structure

```cypher
-- Convert binary addresses to integers for gap calculation
WITH [x IN split(binary_address, "") | toInteger(x)] AS bits
WITH reduce(dec = 0, b IN bits | dec * 2 + b) AS int_value
-- ... gap detection similar to number pools
```

---

### Overview

IP Prefix pools carve out sub-prefixes of a specified length from larger parent prefixes. The algorithm finds contiguous address blocks that aren't already allocated.

### Algorithm

1. **Calculate Block Size**: `block_size = 2^(max_prefixlen - target_prefixlen)`
2. **Collect Existing Prefixes**: Query all existing prefixes within the parent
3. **Convert to Ranges**: Transform each prefix to start/end integer addresses
4. **Cursor-Based Gap Finding**:
   - Start cursor at parent prefix start
   - For each existing range:
     - If cursor + block_size fits before the range starts → **gap found!**
     - Otherwise, advance cursor past the range (aligned to block_size)
5. **Final Check**: If no gap found in loop, check remaining space after last range

### Cypher Algorithm (IPv4)

```cypher
reduce(acc = {cursor: parent_start, found: null}, r IN ranges_sorted |
    CASE
        WHEN acc.found IS NOT NULL THEN acc  -- Already found
        WHEN r.end < acc.cursor THEN acc     -- Range ends before cursor
        WHEN acc.cursor + block_size - 1 < r.start THEN 
            {cursor: acc.cursor, found: acc.cursor}  -- Gap found!
        ELSE -- Advance cursor past current range, aligned to block_size
            {cursor: ceil((r.end + 1) / block_size) * block_size, found: null}
    END
)
```

---

### The Problem

Neo4j (and most databases) use 64-bit integers, which can represent values up to ~9.2×10¹⁸. This is sufficient for:

- **IPv4**: 32-bit addresses (max ~4.3×10⁹) ✅
- **IPv6**: 128-bit addresses (max ~3.4×10³⁸) ❌ **Overflow!**

For IPv6 prefix allocation, integer arithmetic is impossible because:
- A /64 prefix contains 2⁶⁴ addresses (exceeds 64-bit integer)
- Block size calculations overflow for any reasonable prefix length
- Address comparisons and cursor arithmetic fail

### The Solution: Binary String Operations

For IPv6, Infrahub implements all arithmetic operations using **binary string manipulation** directly in Cypher:

#### 1. Binary String Representation

Instead of integers, addresses are stored and manipulated as 128-character binary strings:

```
"2001:db8::/32" → "00100000000000010000110110111000" + "0" × 96
```

#### 2. Binary String Increment

Custom increment logic implemented in Cypher (simulating a binary adder):

```cypher
// Increment binary string by 1 using carry propagation
reduce(
    inc = {bits: split(binary_str, ""), carry: 1, result: []},
    idx IN reverse(range(0, length - 1)) |
    {
        bits: inc.bits,
        carry: CASE
            WHEN inc.carry = 0 THEN 0
            WHEN inc.bits[idx] = "1" THEN 1  -- 1 + 1 = 10, carry continues
            ELSE 0                            -- 0 + 1 = 1, no carry
        END,
        result: CASE
            WHEN inc.carry = 0 THEN [inc.bits[idx]] + inc.result
            WHEN inc.bits[idx] = "1" THEN ["0"] + inc.result
            ELSE ["1"] + inc.result
        END
    }
).result
```

#### 3. String Comparison for Ordering

Binary strings can be compared lexicographically (since they're zero-padded to equal length):

```cypher
ORDER BY r.start_block ASC  -- String comparison works correctly
WHERE cursor_str <= parent_end_bin
```

#### 4. Overflow Detection

Since incrementing a binary string can overflow (all 1s + 1 = overflow), we detect this by checking string length:

```cypher
WHEN size(cursor_str) = target_prefixlen  -- Valid: length unchanged
```

### Comparison Table

| Aspect | IPv4 | IPv6 |
|--------|------|------|
| **Query Class** | `IPPrefixPoolGetNextIPv4` | `IPPrefixPoolGetNextIPv6` |
| **Address Representation** | 32-bit integer | 128-character binary string |
| **Gap Calculation** | Integer arithmetic | Binary string operations |
| **Block Size Calculation** | `2^(32 - prefixlen)` | Prefix truncation at target length |
| **Cursor Increment** | `cursor + block_size` | Binary string increment algorithm |
| **Overflow Handling** | Not needed (fits in 64-bit) | String length validation |
| **Comparison Operators** | `<`, `>`, `=` (numeric) | Lexicographic string comparison |

### Performance Considerations

The IPv6 binary string approach is:
- **More complex**: Requires custom Cypher logic for arithmetic
- **Slower**: String operations are less efficient than native integer math
- **Necessary**: No alternative exists within Neo4j's type system

## Summary by CodeRabbit

* **Refactor**
  * Centralized IPAM allocation into a single allocator for more consistent prefix/address selection and improved IPv6 handling.
  * Reworked number-pool allocation to support constrained free-number retrieval and more robust exclusion handling.

* **New Features**
  * Added "free resource" query paths to surface next-free prefixes and addresses.

* **Tests**
  * Added large-scale prefix allocation and fragmentation tests; removed tests tied to the prior availability utility.


backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -93,23 +92,21 @@ async def execute_post_queries(
...
+        async def allocate_numbers(db: InfrahubDatabase) -> None:
+            for node in nodes:
+                number = await number_pool.get_resource(
+                    db=db, branch=branch, node=node, attribute=self.new_attribute_schema
+                )
+                attr = getattr(node, self.new_attribute_schema.name)
+                attr.value = number
+                attr.source = number_pool.id
+
+                await node.save(db=db, fields=[self.new_attribute_schema.name])
```
ajtmccarty: please add `at=at` here

backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -93,23 +92,21 @@ async def execute_post_queries(
...
-            attr.value = number
-            attr.source = number_pool.id
-
-            await node.save(db=db, fields=[self.new_attribute_schema.name])
+        async def allocate_numbers(db: InfrahubDatabase) -> None:
+            for node in nodes:
+                number = await number_pool.get_resource(
+                    db=db, branch=branch, node=node, attribute=self.new_attribute_schema
+                )
+                attr = getattr(node, self.new_attribute_schema.name)
```
ajtmccarty: can use `node.get_attribute(self.name_attribute_schema.name)` here I believe

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +174,275 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+            $block_size AS block_size,
+            $parent_start AS parent_start,
+            $parent_end AS parent_end
+        WITH block_size, parent_start, parent_end,
+            [r IN ranges_raw |
+                {
+                    start: reduce(dec = 0, b IN split(r.binary, "") | dec * 2 + toInteger(b)),
+                    end: reduce(dec = 0, b IN split(r.binary, "") | dec * 2 + toInteger(b)) + toInteger(2 ^ (32 - r.prefixlen)) - 1
+                }
+            ] AS ranges
```
ajtmccarty: a comment for this might be nice
I have no idea what is going on here

same with the other big reduce/accumulator blocks

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +174,275 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+        WHERE ns_rel.name = "ip_namespace__ip_prefix"
+            AND av.binary_address STARTS WITH $prefix_binary
+            AND av.prefixlen > $maxprefixlen
+            AND av.version = $ip_version
+            AND all(r IN relationships(path2) WHERE (%(branch_filter)s) AND r.status = "active")
+        WITH
+            collect({binary: av.binary_address, prefixlen: av.prefixlen}) AS ranges_raw,
+            $block_size AS block_size,
+            $parent_start AS parent_start,
+            $parent_end AS parent_end
```
ajtmccarty: I would think you could just use `$block_size/parent_start/end` wherever needed and not have to assign the parameters to variables

backend/infrahub/core/query/ipam.py
```diff
@@ -242,6 +649,145 @@ async def get_ip_addresses(
...
+        namespace=namespace,
+        at=at,
+        branch_agnostic=branch_agnostic,
+        is_pool=is_pool,
+    )
+    await query.execute(db=db)
+    return query.get_address()
+
+
+async def get_next_free_prefix(
```
ajtmccarty: looks like it should be a class that accepts db, branch, namespace, and maybe branch_agnostic in the constructor instead of a chain of functions

backend/infrahub/core/query/ipam.py
```diff
@@ -38,6 +38,43 @@ class IPAddressData:
...
+
+
+@dataclass(frozen=True)
+class IPAddressFreeData:
+    free_addr: int
+    is_free: bool
+    is_last: bool
+
+    @classmethod
+    def from_db(cls, result: QueryResult) -> IPAddressFreeData:
```
ajtmccarty, line 66: these `from_db` methods reverse the dependency chain and violate the D in SOLID
it requires these dataclasses to have knowledge of the shape of the data returned by a query, which is going up the dependency chain instead of down. the query should know how to construct the data that it returns, not the other way around.
so instead of `from_db`, just build the class instances in the `get_addresses` or whatever methods of the queries

backend/infrahub/core/query/resource_manager.py
```diff
@@ -338,6 +354,100 @@ def iter_results(self) -> Generator[NumberPoolIdentifierData]:
...
+        query = """
+        MATCH (pool:%(number_pool)s { uuid: $pool_id })-[res:IS_RESERVED]->(av:AttributeValueIndexed)
+        WHERE toInteger(av.value) >= $start_range and toInteger(av.value) <= $end_range
+        CALL (pool, res, av) {
+            MATCH (pool)-[res]->(av)<-[hv:HAS_VALUE]-(attr:Attribute)<-[ha:HAS_ATTRIBUTE]-(n:%(node)s)
+            WHERE
+                n.uuid = res.identifier AND
+                attr.name = $attribute_name AND
+                all(r in [res, hv, ha] WHERE (%(branch_filter)s))
+            ORDER BY res.branch_level DESC, hv.branch_level DESC, ha.branch_level DESC, res.from DESC, hv.from DESC, ha.from DESC
```
ajtmccarty: I think this needs to be 
```
ORDER BY res.branch_level DESC, hv.branch_level DESC, ha.branch_level DESC,
res.from DESC, hv.from DESC, ha.from DESC
res.status ASC, hv.status ASC, ha.status ASC
```

to make sure that it correctly accounts for the case when an attribute is NULLed on a branch, which creates a `status="deleted"` and a `status="active"` edge with the same `from` timestamp

backend/infrahub/core/query/ipam.py
gmazoyer: I added these `from_db` methods because it seems to be the pattern that was approved in https://github.com/opsmill/infrahub/pull/8020.

backend/infrahub/core/migrations/schema/node_attribute_add.py
dgarros: Would be good to start creating an entry in the knowledge base for migration and capture that information at a minimum. Claude can usually generate a good starting point very quickly and we can later iterate on it.

backend/infrahub/core/query/ipam.py
dgarros: @ajtmccarty I get your point and we should probably move that logic within get_data instead but since the Dataclass and the Query are also colocated in the same file I thing we can live it that for now :D

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +172,263 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+        self.return_labels = ["free_start_bin"]
+        self.limit = 1
+
+    def get_prefix_data(self) -> IPv6PrefixFreeData | None:
+        result = self.get_result()
+        if not result:
+            return None
+
+        return IPv6PrefixFreeData.from_db(result=result)
+
```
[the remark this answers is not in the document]
gmazoyer: @fatih-acar do you think it's a legit feedback?

backend/infrahub/core/query/ipam.py
```diff
@@ -210,36 +502,231 @@ def get_addresses(self) -> list[IPAddressData]:
...
+        if result_data.is_free:
+            return ipaddress.ip_interface(result_data.free_addr)
+        if result_data.is_last and result_data.free_addr < self.params["end_range"]:
+            return ipaddress.ip_interface(result_data.free_addr + 1)
+
+        return None
+
+
+class IPAMResourceAllocator:
```
ajtmccarty: nice new component, very SOLID
I think this should move to `infrahub.core.ipam.resource_allocator.py`

backend/infrahub/core/node/resource_manager/number_pool.py
```diff
@@ -85,51 +128,98 @@ async def get_resource(
...
+            if parameters.min_value is not None:
+                effective_start = max(effective_start, parameters.min_value)
+            if parameters.max_value is not None:
+                effective_end = min(effective_end, parameters.max_value)
+
+        # Check if the effective range is valid
+        if effective_start > effective_end:
+            raise PoolExhaustedError("There are no more values available in this pool.")
+
+        min_value: int = effective_start
```
ajtmccarty: is this line necessary? it looks like `min_value` is always set again on 204

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +172,263 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+
+        # ruff: noqa: E501
+        query = """
+        // First match on IPNAMESPACE
+        MATCH (ns:%(ns_label)s)
+        WHERE ns.uuid = $ns_id
+        CALL (ns) {
+            MATCH (ns)-[r:IS_PART_OF]-(root:Root)
+            WHERE %(branch_filter)s
+            RETURN ns AS ns1, r AS r1
```
ajtmccarty: this works, but you don't need to do `RETURN r AS r1` and then `WITH r1 AS r` on line 220. this is an older pattern that we don't need going forward
you can just do `RETURN r` instead

backend/infrahub/core/query/ipam.py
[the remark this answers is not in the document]
dgarros: Would be good to create a test for it and validate it :)

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +172,263 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+                    ELSE
+                        {
+                            cursor: CASE
+                                WHEN toInteger((r.end + 1 + $block_size - 1) / $block_size) * $block_size > $parent_end + 1
+                                THEN $parent_end + 1
+                                ELSE toInteger((r.end + 1 + $block_size - 1) / $block_size) * $block_size
+                            END,
+                            found: null
+                        }
+                END
```
ajtmccarty, line 283: some comments in here would be helpful
some of this looks like it is doing nothing, but I think I am missing something.

`r.end + 1 + $block_size -1` seems like the  `+/- 1` can be removed unless it is supposed to make this easier to understand in some way

`(x / $block_size) * $block_size` must be doing something with rounding/flood/ceiling, but I'm not sure

backend/infrahub/core/query/ipam.py
```diff
@@ -137,6 +172,263 @@ def get_subnets(self) -> list[IPPrefixData]:
...
+            ) AS res
+        WITH CASE
+            WHEN res.found IS NOT NULL THEN res.found
+            WHEN res.cursor + $block_size - 1 <= $parent_end THEN res.cursor
+            ELSE NULL
+        END AS free_start
+        WHERE free_start IS NOT NULL
```
ajtmccarty, line 290: I think this means that the query could return nothing (not completely certain on that), which means that it should have `raise_error_if_empty = False` as a class variable

backend/infrahub/core/query/ipam.py
```diff
@@ -210,36 +502,231 @@ def get_addresses(self) -> list[IPAddressData]:
...
+        WITH ns
+        // MATCH all IPAddress that are IN SCOPE
+        MATCH path2 = (ns)-[:IS_RELATED]-(ns_rel:Relationship)-[:IS_RELATED]-(addr:%(node_label)s)-[:HAS_ATTRIBUTE]-(an:Attribute {name: "address"})-[:HAS_VALUE]-(av:AttributeIPHost)
+        WHERE ns_rel.name = "ip_namespace__ip_address"
+            AND av.binary_address STARTS WITH $prefix_binary
+            AND av.prefixlen >= $maxprefixlen
+            AND av.version = $ip_version
+            AND all(r IN relationships(path2) WHERE (%(branch_filter)s) and r.status = "active")
+        ORDER BY av.binary_address
+        WITH DISTINCT av.binary_address AS used_addresses
```
ajtmccarty: I don't think this variable name should be plural. it's only for 1 address, right? seems like it should be `used_address` or maybe even `used_address_str`

backend/infrahub/core/query/ipam.py
```diff
@@ -210,36 +502,231 @@ def get_addresses(self) -> list[IPAddressData]:
...
+        WITH [x IN split(used_addresses, "") | toInteger(x)] AS bits
+        WITH [$start_range - 1] + collect(reduce(dec = 0, b IN bits | dec * 2 + b)) AS nums
+        UNWIND range(0, size(nums) - 1) AS idx
+        CALL (nums, idx) {
+            WITH nums[idx] AS curr, idx - 1 + $start_range AS expected
+            RETURN expected AS addr, expected <> curr AS is_free, idx = size(nums) - 1 AS is_last
+        }
```
ajtmccarty, line 630: a comment to explain what is happening here would be helpful

backend/infrahub/core/query/ipam.py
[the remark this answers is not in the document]
fatih-acar: I think it's talking about the case when you reach `ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff`. The issue reported is legit but is not likely to happen in a real world use case?

Here's a test generated by Claude to verify this, feel free to add it:

The test will fail due to the issue reported by coderabbit.
If you change `ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff0/124` by `ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffe0/124` for example, the test will pass.

```
async def test_ipv6_small_prefix_pool_full_allocation(
    db: InfrahubDatabase, default_branch: Branch, default_ipnamespace: Node, register_ipam_schema: SchemaBranch
) -> None:
    """Test allocating /127 prefixes from a small /124 IPv6 prefix pool until exhausted.

    The prefix ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff0/124 contains 16 addresses.
    Allocating 8 x /127 prefixes (2 addresses each) should fill the pool completely.
    The 9th allocation should fail with IndexError.
    """
    prefix_schema = registry.schema.get_node_schema(name="IpamIPPrefix", branch=default_branch)

    ns_v6 = await Node.init(db=db, schema=InfrahubKind.NAMESPACE)
    await ns_v6.new(db=db, name="ns_v6_small")
    await ns_v6.save(db=db)

    parent_v6 = await Node.init(db=db, schema=prefix_schema)
    await parent_v6.new(db=db, prefix="ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff0/124", ip_namespace=ns_v6)
    await parent_v6.save(db=db)

    prefix_pool_schema = registry.schema.get_node_schema(name=InfrahubKind.IPPREFIXPOOL, branch=default_branch)

    pool = await CoreIPPrefixPool.init(schema=prefix_pool_schema, db=db)
    await pool.new(db=db, name="pool_v6_small", resources=[parent_v6], ip_namespace=ns_v6)
    await pool.save(db=db)

    allocated_prefixes: list[Node] = []
    expected_prefixes = [
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff0/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff2/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff4/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff6/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fff8/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fffa/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fffc/127",
        "ffff:ffff:ffff:ffff:ffff:ffff:ffff:fffe/127",
    ]

    # Allocate 8 /127 prefixes to fill the /124 pool
    for i in range(8):
        new_prefix = await pool.get_resource(
            db=db,
            prefixlen=127,
            prefix_type="IpamIPPrefix",
            member_type="prefix",
            identifier=f"v6_small_{i}",
            branch=default_branch,
        )
        allocated_prefixes.append(new_prefix)

    # Verify all 8 prefixes were allocated correctly
    for i, prefix in enumerate(allocated_prefixes):
        assert prefix.prefix.value == expected_prefixes[i]

    # The 9th allocation should fail as the pool is exhausted
    with pytest.raises(IndexError):
        await pool.get_resource(
            db=db,
            prefixlen=127,
            prefix_type="IpamIPPrefix",
            member_type="prefix",
            identifier="v6_small_overflow",
            branch=default_branch,
        )
```

fatih-acar (commented):
LGTM thanks @gmazoyer 
We must solve @ajtmccarty's relevant comments

backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -93,23 +93,21 @@ async def execute_post_queries(
...
+                attr.value = number
+                attr.set_source(number_pool)
+
+                await node.save(db=db, fields=[self.new_attribute_schema.name], at=at)
+
+        if db.is_transaction:
+            await allocate_numbers(db=db)
+        else:
+            async with db.start_transaction() as dbt:
+                await allocate_numbers(db=dbt)
```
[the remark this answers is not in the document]
gmazoyer: This is wrong. is_transaction is an property of InfrahubDatabase

ajtmccarty (approved):

PR #7949, opened 2025-12-23
sorting support for metadata when querying objects

ajtmccarty:
IFC-1947

add support for sorting Nodes by created_at or updated_at

## Summary by CodeRabbit

* **New Features**
  * Support ordering query results by metadata timestamps (created_at, updated_at) in ASC/DESC with stable pagination and branch-aware behavior.
* **Chores**
  * Centralized ordering model and validation into core and added a lightweight input deserializer for order inputs.
* **Bug Fixes / Performance**
  * Improved query construction for more reliable WHERE placement and stable ordering across pagination.
* **Tests**
  * Added extensive unit and GraphQL tests covering metadata ordering, pagination, and branch scenarios.


backend/infrahub/core/order.py
ajtmccarty, line 1: moved into core b/c it is used in core, so should not be in the `graphql` module
converted to a pydantic model so that it will be instantiated correctly when passed a nested dict

dgarros (approved):

backend/infrahub/core/query/node.py
```diff
@@ -1507,6 +1517,26 @@ def has_filter_by_id(self) -> bool:
...
+            return False
+        return bool(self.order.node_metadata.created_at or self.order.node_metadata.updated_at)
+
+    @property
+    def _metadata_order_fields(self) -> list[tuple[str, OrderDirection]]:
+        """Return the metadata field and direction to order by, or None."""
+        if not self._has_metadata_order:
+            return []
+        fields: list[tuple[str, OrderDirection]] = []
+        nm = self.order.node_metadata  # type: ignore[union-attr]
```
ogenstad: Could be cleaner to have a method that would return an optional `OrderModel` instead, then if that object is `None` we return an empty list.

backend/infrahub/core/query/node.py
```diff
@@ -1446,8 +1450,14 @@ def is_filter(self) -> bool:
...
         return FieldAttributeRequirementType.ORDER in self.types
 
+    @property
+    def is_metadata_order(self) -> bool:
+        return FieldAttributeRequirementType.METADATA_ORDER in self.types
+
     @property
     def node_value_query_variable(self) -> str:
+        if self.is_metadata_order:
+            return f"order_{self.field_name}"
```
ogenstad, line 1460: Looking ahead could the property name here cause confusion or conflict if we later want to support ordering by attribute value?

backend/infrahub/core/query/node.py
```diff
@@ -1531,6 +1561,78 @@ def _untrack_variable(self, variable: str) -> None:
...
+    WITH CASE
+        WHEN r.branch IN $branch0 AND r.from < $time0 THEN [r.from]
+        WHEN r.branch IN $branch1 AND r.from < $time1 THEN [r.from]
+        ELSE [NULL]
+    END AS from_details,
+    CASE
+        WHEN r.branch IN $branch0 AND r.to < $time0 THEN [r.to]
+        WHEN r.branch IN $branch1 AND r.to < $time1 THEN [r.to]
+        ELSE [NULL]
+    END AS to_details
```
ogenstad, line 1599: Given the nested if-statements above I'm wondering if it would be easier to read if we also indented that part of the cypher queries. In the above example all of the statements are quite short and it could be different for different queries. Not really an objection here more of an observation.

backend/infrahub/core/query/node.py
```diff
@@ -1749,24 +1858,51 @@ def _get_field_requirements(self, disable_order: bool) -> list[FieldAttributeReq
...
+                        field_attr_name=metadata_field,
+                        field_attr_value=None,
+                        index=index,
+                        types=[FieldAttributeRequirementType.METADATA_ORDER],
+                        order_direction=direction,
+                    )
+                index += 1
+
+        # Add schema order_by requirements
+        elif self.schema.order_by:
```
ogenstad: Given that this is an `elif` could this cause problems for instance if we create a device using a template and then have that template create 200 interfaces. If all of those interfaces have the same creation timestamp and we try to paginate the results. I think we're still always adding an `n.uuid` at the end as a tiebreaker?

ogenstad (approved):

backend/infrahub/core/query/node.py
ajtmccarty: yes good point. I actually just removed the `_has_metadata_order` and changed the `_metadata_order_fields` property to a `_get_metadata_order_fields` method

backend/infrahub/core/query/node.py
ajtmccarty: I don't think this will matter much. this variable is used consistently when building the query, so the name does not actually matter as long as it is unique

backend/infrahub/core/query/node.py
ajtmccarty: I do the spacing like this b/c it is the correct indenting when plugged into its parent `updated_at_query` string

backend/infrahub/core/query/node.py
ajtmccarty: I removed this `elif` and its preceding `if`, b/c they don't do much beyond add another layer of indenting, but I don't think this is an issue, because we always want to add `n.uuid` as an ordering tie-breaker

PR #7952, opened 2025-12-24
New details page (header + groups/profiles card + inline edit)

bilalabbad:


bilalabbad:
observation done. closing it.

PR #7957, opened 2025-12-24
allow running a single migration again

ajtmccarty:
an update to `infrahub db migrate` made it impossible to run one migration

this update re-implements that functionality and adds some tests to prevent it from regressing again in the future

## Summary by CodeRabbit

* **New Features**
  * Centralized migration entry point and an option to control whether the repository graph version is updated after migrations; improved handling and messaging when targeting a specific migration number, including support for re-applying a specific migration.

* **Tests**
  * Added end-to-end unit tests covering migration selection, re-run behavior, check mode, and graph version updates.

* **Documentation**
  * Clarified CLI help text for the --migration-number option.


dgarros (approved):
LGTM, would be good to add an entry in the changelog

Issue #6761, opened 2025-07-01
feature: Enable/Disable Webhooks


SudarshanVK:
### Component

Webhooks

### Describe the Feature Request

Right now there is no feature to enable or disable a configured webhook. The only way to achieve it is to delete the webhook which is not always ideal

### Describe the Use Case

Sometimes I may want to disable a webhooks and there should be a way to do it without having to delete the webhook.

ogenstad:
We have this for the event triggers and a similar option could be added to the webhooks.

dgarros:
Fix in #7961, will be included in the next release (1.7)

PR #7961, opened 2025-12-25
add ability to disable webhooks

ashleytayles:
## Release Notes

* **New Features**
  * Webhooks can now be enabled and disabled using the new "active" field.
  * Inactive webhooks will not trigger automations.
  * Deactivating a webhook automatically removes its associated automation; reactivating it recreates it.

* **Schema Updates**
  * GraphQL schema updated to expose and manage webhook active states.

Related Issue: [#6761](https://github.com/opsmill/infrahub/issues/6761)


changelog/6761.added.md
dgarros, line 1: Please can you rename the file 6761.added.md, we usually use the issue ID not the PR Id for the changelog

dgarros (approved):
Thank you @ashleytayles, looks great
Please can you rebase your branch to target `release-1.7` to include it in the next release and can you rename the changelog to reference the issue ID instead of the PR id

changelog/6761.added.md
ashleytayles: I have renamed the changelog and rebased to the 1.7 branch.

backend/infrahub/webhook/tasks.py
```diff
@@ -132,6 +132,21 @@ async def configure_webhook_one(
     trigger = WebhookTriggerDefinition.from_object(webhook)
 
     async with get_prefect_client(sync_client=False) as prefect_client:
+        existing_automations = await prefect_client.read_automations_by_name(trigger.generate_name())
```
ogenstad, line 135: Looking at this I realize we have a bug in the webhook code for setting up single webhooks. The call to `prefect_client.read_automations_by_name` will not work on systems with a lot of automations as the internal Prefect SDK doesn't allow for pagination. We should probably get rid of the task to just setup one automation like this. But before that's done we should instead use the `gather_all_automations` and then filter the results we get back to match against the expected automation name. Same with the unittests below, however in that context it's safer than within the actual code.

ashleytayles: > Looking at this I realize we have a bug in the webhook code for setting up single webhooks. The call to `prefect_client.read_automations_by_name` will not work on systems with a lot of automations as the internal Prefect SDK doesn't allow for pagination. We should probably get rid of the task to just setup one automation like this. But before that's done we should instead use the `gather_all_automations` and then filter the results we get back to match against the expected automation name. Same with the unittests below, however in that context it's safer than within the actual code.

Want me to take a stab at fixing for this PR, or shall we open a new issue?

ogenstad: As you didn't introduce the error we can leave it as is. I'll track it elsewhere. I do note that there's a new merge conflict for this PR that needs to be resolved though.

dgarros:
Thank you @ashleytayles

PR #7975, opened 2025-12-30
IFC-1949: Remove core proposed change relationship

solababs:
## Summary by CodeRabbit

* **Chores**
  * Added a graph migration to populate proposed-change creator metadata on existing nodes.

* **Breaking Changes**
  * ProposedChange no longer exposes a public created_by relationship in the schema or direct API node payloads.

* **Behavior Changes**
  * Creator ID is no longer auto-injected during proposed-change creation.
  * Review flow now prevents approvers from approving their own proposals.
  * GraphQL schema: expanded filtering and metadata options for richer queries.

* **Tests**
  * Added migration and review tests; updated tests to verify creator via edge metadata.

* **Frontend**
  * UI and queries now read creator from edge metadata; proposed-change list/detail props reshaped.


backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -274,10 +273,6 @@ async def mutate(
             state = ProposedChangeState(proposed_change.state.value.value)
             state.validate_reviewable()
 
-            created_by = await proposed_change.created_by.get_peer(db=graphql_context.db)
-            if created_by and created_by.id == graphql_context.active_account_session.account_id:
-                raise ValidationError(input_value="You cannot review your own proposed changes")
-
```
gmazoyer, line 280: Is this OK to be removed or should we change the conditional to use the metadata `created_by` info?

solababs: I added this https://github.com/opsmill/infrahub/pull/7975/changes#diff-8372cfd2a1ef2bdf93e3aed5070c4a2246e98207c456ea2e3e9669f6517e39b1R237 test to check that condition, I can add it using the metadata if this does not suffice

gmazoyer: I think the test passes because of the missing permission (given the error string). If you setup the test in a way that the user who created the proposed change has the `REVIEW_PROPOSED_CHANGE` global permission, it seems that the user would be able to review his own proposed change.

gmazoyer (approved):

backend/tests/unit/core/migrations/graph/test_051.py
```diff
@@ -0,0 +1,51 @@
...
+    CREATE (rel2)-[:IS_RELATED {branch: $branch, branch_level: 1, status: "active", from: $at}]->(account2)
+    """
+    await db.execute_query(
+        query=create_test_data_query,
+        params={
+            "account1_uuid": account1_uuid,
+            "account2_uuid": account2_uuid,
+            "pc1_uuid": pc1_uuid,
+            "pc2_uuid": pc2_uuid,
+            "branch": default_branch.name,
```
ajtmccarty, line 31: technically, this should be the global branch `GLOBAL_BRANCH_NAME`, but it does not really matter for this simple test

backend/tests/unit/graphql/mutations/test_proposed_change.py
```diff
@@ -218,6 +237,68 @@ async def test_create_invalid_state_combinations(
...
+                    decision=PermissionDecision.ALLOW_ALL.value,
+                )
+            ],
+            "object_permissions": [],
+        }
+
+
+async def test_cannot_approve_own_created_proposed_change(
+    db: InfrahubDatabase, default_branch: Branch, register_core_models_schema: None
+) -> None:
```
ajtmccarty, line 257: nice new test 👍

backend/tests/unit/graphql/mutations/test_proposed_change.py
```diff
@@ -218,6 +234,51 @@ async def test_create_invalid_state_combinations(
...
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={
+            "source": source_branch.name,
+            "destination": "main",
+            "name": "sample proposed change",
+            "state": "open",
+        },
+    )
+    assert not open_proposed_change.errors
```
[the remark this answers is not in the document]
ajtmccarty: this might be necessary to prevent some sort of flakiness in the test, but I can't recall the specifics exactly. @ogenstad do you know if the `update_schema_hash` call is required here?

ajtmccarty (approved):

backend/tests/unit/graphql/mutations/test_proposed_change.py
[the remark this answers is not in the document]
ogenstad: > this might be necessary to prevent some sort of flakiness in the test, but I can't recall the specifics exactly. @ogenstad do you know if the `update_schema_hash` call is required here?

I probably added this call a bit excessive when I added it pretty much everywhere. The problem is basically that we're sharing the registry across tests and within the registry we only use the name of the branch as a key, but in the GraphQL registry we use the hash of the branch as the key. Without this call to `update_schema_hash` we might end up using the incorrect hash for the current branch and as such end up with the incorrect GraphQL schema. I think it might be ok in this test as we're pretty much using a generic schema. I have some ideas for future updates to the testing infrastructure to make it easier to setup and load schemas to make it more consistent and less verbose. We might also want to consider what we should to within the registry around this. For now it can't hurt to add this call but I don't think it's required.

ogenstad (approved):

backend/infrahub/graphql/queries/proposed_change.py
```diff
@@ -37,17 +38,27 @@ async def resolve(
     ) -> dict:
         graphql_context: GraphqlContext = info.context
         proposed_change = await NodeManager.get_one(
-            kind=CoreProposedChange,
```
dgarros, line 40: just curious, why do we need to switch off using the protocol here ?

solababs: ruff and mypy complains about the protocol as it does not contain a `_get_created_by()` which is only available in a node  https://github.com/opsmill/infrahub/pull/7975/changes/BASE..e1ebfb140d9c650d2d0276f64826418b1f25a18f#diff-1c8ddacb2a23d8a9d0a1a4572cb46de35726b54e8aa52f80a802cef0d9ebf763R49 a node is actually returned by the `get_one` function instead of the protocol

ogenstad: The goal of the protocols are really to allow us to access attributes directly on the returned objects and ignore any typing violations when those attributes doesn't exist on the Node objects. In this case as we are shifting away from attributes we don't really need the protocol class here. Having said that regardless of this change we should add the `_get_created_by()` and similar metadata methods to the base protocol class.

dgarros: I disagree, the goal should be to use Protocols everywhere and move away from using NodeManager functions with kind defined as string.
It would have been faster to just update the Protocol here instead of implementing a workaround.

ajtmccarty: added in #8012

PR #7977, opened 2025-12-30
fix(backend): add task worker heartbeat detection

### LINKED ISSUE #8227: bug: git repositories sync blocked due to already running task
### Component

API Server / GraphQL

### Infrahub version

1.7.1

### Current Behavior

A known issue with task workers is that if they unexpectedly restart while having running tasks; those tasks will stay in "Running" state forever, requiring manual clearing.

When the git repositories sync task is concerned, this is even more problematic because we only allow one concurrent task to be ran; thus further sync tasks are blocked forever.

### Expected Behavior

Running tasks from restarted workers should be canceled/reset/cleaned automatically to prevent blocking git syncs.

### Steps to Reproduce

Restart a task worker during the execution of a git repositories sync task.

### Additional Information

_No response_

fatih-acar:
This will prevent zombie running tasks.

Basically this is an implementation of the official example: https://docs.prefect.io/v3/advanced/detect-zombie-flows#create-the-automation using our own workflow machinery.

No performance nor functional regressions were found during testing (ran generate "large" dataset test suite; no task was marked as crashed during the process).
Testing was done manually; I don't think it's required to implement our own functional tests for this since this is tested upstream.

Fixes #8227 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where task worker restarts could leave tasks stuck in Running, blocking Git repo sync.

* **New Features**
  * Added a system automation that detects flows missing heartbeats and marks them as crashed after ~90 seconds to prevent stuck runs.

* **Chores**
  * Added heartbeat frequency configuration for backend flows.

* **Documentation**
  * Added changelog entry describing the fix above.


backend/infrahub/trigger/models.py
```diff
@@ -218,3 +276,34 @@ def generate_name(self) -> str:
 
 class BuiltinTriggerDefinition(TriggerDefinition):
     type: TriggerType = TriggerType.BUILTIN
+
+
+class SystemTriggerDefinition(BaseModel):
```
ogenstad: Is there a reason why SystemTriggerDefinition can't inherit from TriggerDefinition? That way we wouldn't need `AnyTriggerDefinition` along with the update to other definitions such as:

```python
class TriggerSetupReport(BaseModel):
    created: list[TriggerDefinition | SystemTriggerDefinition] = Field(default_factory=list)
```

backend/infrahub/trigger/models.py
```diff
@@ -218,3 +273,24 @@ def generate_name(self) -> str:
...
+
+    def get_deployment_names(self) -> list[str]:
+        """Return empty list as system triggers don't use deployments."""
+        return []
+
+    def get_description(self) -> str:
+        return f"System Automation for {self.name} (v{__version__})"
+
+    def validate_actions(self) -> None:
+        """System triggers use ChangeFlowRunStateAction which doesn't require parameter validation."""
```
ogenstad: Do we need to define get_deployment_names and validate_actions here or would the ones from TriggerDefinition always return what we want given the conditionals there?

backend/infrahub/trigger/models.py
fatih-acar: I think it's slightly better now, thanks... should have done it before

ogenstad (approved):

backend/infrahub/trigger/models.py
fatih-acar: Good catch, Claude was not smart enough to figure this out... (probably because I did the changes in `TriggerDefinition` manually...)

PR #7978, opened 2025-12-30
Remove old Task and TaskLog based on standard nodes

ogenstad:
These tasks based on the StandardNode type have since way back been replaced with the logs within Prefect.

## Release Notes

* **Chores**
  * Removed task management functionality, including task creation and query operations
  * Removed task logging capabilities
  * Removed user task context manager
  * Removed task-related GraphQL input types
  * Cleaned up related package exports and test references


backend/tests/integration/proposed_change/test_proposed_change_profile.py
```diff
@@ -80,8 +79,5 @@ async def test_pipeline(self, db: InfrahubDatabase, update_profile: None, client
             validator.conclusion.value.value == ValidatorConclusion.SUCCESS.value for validator in peers.values()
         )
 
-        tasks = await Task.query(db=db, ids=[], fields={}, related_nodes=[proposed_change.id], limit=10, offset=0)
-        assert all(task["node"]["conclusion"] == TaskConclusion.SUCCESS for task in tasks["edges"])
```
ogenstad, line 84: It looks like I'm modifying the test here. But this part of the test didn't actually test anything. It looks like we're validating that all responses have the success conclusion. The problem is that `all([])` comes back as `True` and for a test like this in the future we should begin with an assert statement like this: `assert tasks["edges"]` to ensure that we actually check to see if this list has any entries to compare against.

gmazoyer (approved):

ajtmccarty (approved):

PR #7979, opened 2025-12-30
Remove abstract base class from query

ogenstad:
Remove ABC from the base Query class in order to start using the `__init__()` method directly for future queries. Also converted one query to the cleaner format as an example.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal query initialization to use explicit parameters, reducing ambiguity.
  * Converted the core query base to a concrete, configurable implementation for more predictable behavior.
  * Simplified permission-fetching instantiation to use direct parameter passing, improving clarity and reliability.

### DIFF
--- a/backend/infrahub/core/account.py
+++ b/backend/infrahub/core/account.py
@@ -54,11 +54,15 @@ class AccountGlobalPermissionQuery(Query):
     name: str = "account_global_permissions"
     type: QueryType = QueryType.READ
 
-    def __init__(self, account_id: str, **kwargs: Any):
+    def __init__(
+        self,
+        account_id: str,
+        branch: Branch | None = None,
+        branch_agnostic: bool = False,
+    ) -> None:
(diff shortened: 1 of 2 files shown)


backend/infrahub/core/account.py
```diff
@@ -293,7 +296,7 @@ def get_permissions(self) -> list[ObjectPermission]:
 
 
 async def fetch_permissions(account_id: str, db: InfrahubDatabase, branch: Branch) -> AssignedPermissions:
-    query1 = await AccountGlobalPermissionQuery.init(db=db, branch=branch, account_id=account_id, branch_agnostic=True)
+    query1 = AccountGlobalPermissionQuery(branch=branch, account_id=account_id, branch_agnostic=True)
```
gmazoyer, line 299: I like having it like this 👍

gmazoyer (approved):

solababs (approved):

ajtmccarty (approved):
this is much better

PR #7983, opened 2025-12-30
fix(backend, tests): too many db connections in integration tests

fatih-acar:
Since the neo4j driver upgrade, we must ensure the neo4j driver is properly closed and that a closed driver is not re-used. We disabled singletons during DI to get that behavior.

A side effect is that during some integration tests that leverage local workers, some tasks/flows depend on the injected database instance, which would trigger new driver instances with new db connections. Those instances would not get cleaned since it's not using a singleton nor used within a context manager that would close the underlying db driver.

Fix this by introducing a new class level fixture for the db instance and use it within the injected dependency.

## Summary by CodeRabbit

* **Tests**
  * Added a reusable database fixture to improve test stability and speed.
  * Simplified Prefect testing by replacing dynamic server orchestration with a direct client-based approach.
  * Updated test client and Prefect client interfaces to use async and dependency-driven setups for more consistent integration testing.

### DIFF
--- a/backend/tests/helpers/test_app.py
+++ b/backend/tests/helpers/test_app.py
@@ -121,10 +121,15 @@ class TestInfrahubAppBase(TestInfrahub):
         return schema_branch
 
     @pytest.fixture(scope="class")
+    async def class_db(self) -> InfrahubDatabase:
+        return await build_database(singleton=False)
+
+    @pytest.fixture(scope="class")
     async def test_client(
         self,
         dependency_provider: Provider,
         db: InfrahubDatabase,
+        class_db: InfrahubDatabase,
         initialize_registry: None,
         redis: dict[int, int] | None,
         nats: dict[int, int] | None,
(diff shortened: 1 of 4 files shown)


backend/tests/helpers/test_app.py
```diff
@@ -120,11 +120,16 @@ async def register_core_schema(
         await default_branch.save(db=db)
         return schema_branch
 
+    @pytest.fixture(scope="class")
```
ajtmccarty: could this fixture be defined next to the `db` fixture in `tests/conftest.py`
and maybe it should be called `db_class` or `db_class_scope`
I think we might want to have `scope="class"` versions of some other fixtures and they would be easier to read if they don't all start with `class_`

ajtmccarty (approved):

backend/tests/helpers/test_app.py
fatih-acar: Good catch thanks. Looks like I have not been consistent on the naming... I also moved the Prefect class fixtures to `tests/conftest.py`, using `_class` as a suffix

ajtmccarty (approved):

PR #7992, opened 2025-12-31
Add display label GraphQL filters

gmazoyer:
Now that we store the display labels of nodes in the database we can now perform search based on their values. Since it's stored as an indexed attribute the search efficiency should not be a problem.

## Summary by CodeRabbit

* **New Features**
  * GraphQL filtering by display label (exact, multi-value, partial-match), including across nested relationships.

* **Improvements**
  * Faster resolution of metadata fields used for node labels and human-friendly IDs.

* **UI**
  * Display-label configuration simplified to a single templated string, changing node label rendering.

* **Tests**
  * Added unit tests covering display-label filtering, nested relationships, and partial-match scenarios.


backend/tests/unit/graphql/queries/test_display_label.py
```diff
@@ -422,3 +423,231 @@ async def test_display_label_computed_attr(db: InfrahubDatabase, default_branch:
     assert result.data
     assert len(result.data["TestObjectB"]["edges"]) == 1
     assert result.data["TestObjectB"]["edges"][0]["node"]["display_label"] == "first SECOND"
+
+
+async def test_filter_by_display_label_value(
```
ajtmccarty: I think you could combine all these tests into one that loads all of the data and then runs the various queries against a single dataset. I know that one failure would prevent the following test cases from running, but our "unit" tests already take so long to run that I try to consolidate related tests that can share a dataset

ajtmccarty (approved):

PR #7993, opened 2025-12-31
Add file attachment display in object details view

pa-lem:


pa-lem:
all the work is done in https://github.com/opsmill/infrahub/pull/8009

PR #4883, opened 2024-11-07
Adds optional username/password authentication for Redis

wvandeun:

PR #7994, opened 2025-12-31
Add Redis Authentication

FragmentedPacket:
- Add password to redis class initialization. Update docker compose to support setting Redis password.

> Pending issue creation

## Summary by CodeRabbit

* **New Features**
  * Optional username/password authentication for Redis connections, enabling authenticated access when credentials are provided.

* **Changelog**
  * Added an entry documenting the new Redis credential option.

### DIFF
--- a/backend/infrahub/lock.py
+++ b/backend/infrahub/lock.py
@@ -286,6 +286,7 @@ class InfrahubLockRegistry:
                     port=config.SETTINGS.cache.service_port,
                     db=config.SETTINGS.cache.database,
                     credential_provider=credential_provider,
+                    ssl=config.SETTINGS.cache.tls_enabled,
                     ssl_cert_reqs="optional" if not config.SETTINGS.cache.tls_insecure else "none",
                     ssl_check_hostname=not config.SETTINGS.cache.tls_insecure,
                     ssl_ca_certs=config.SETTINGS.cache.tls_ca_file,
--- a/backend/infrahub/lock.py
+++ b/backend/infrahub/lock.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING
 
 import redis.asyncio as redis
 from prometheus_client import Histogram
+from redis import UsernamePasswordCredentialProvider
 from redis.asyncio.lock import Lock as GlobalLock
 
 from infrahub import config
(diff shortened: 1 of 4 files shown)


ogenstad:
We also have #4883 with an alternate approach.

Looks like we're missing the usernames in this PR, i.e. if we allow for users to use the password they would expect that username also gets passed. The other part is that Redis seems to accept a string or a None. As such I'm not if this change could indicate that it should try to login with an empty password instead of none at all. I.e. if there's ever a difference of sending an empty string `""` vs sending in `None`.  For the block storage for Redis used for Prefect we are doing this:

```python
@task(name="task-manager-setup-blocks", task_run_name="Setup Blocks", cache_policy=NONE)  # type: ignore[arg-type]
async def setup_blocks() -> None:
    log = get_run_logger()

    try:
        await RedisStorageContainer.register_type_and_schema()
    except ObjectAlreadyExists:
        log.warning(f"Redis Storage {TASK_RESULT_STORAGE_NAME} already registered ")

    redis_block = RedisStorageContainer.from_host(
        host=config.SETTINGS.cache.address,
        port=config.SETTINGS.cache.service_port,
        db=config.SETTINGS.cache.database,
        username=config.SETTINGS.cache.username or None,
        password=config.SETTINGS.cache.password or None,
    )
```

Other than these comments I think it should be pretty much ready to go.

backend/infrahub/lock.py
```diff
@@ -275,11 +276,16 @@ def __init__(
...
+                credential_provider: UsernamePasswordCredentialProvider | None = None
+                if config.SETTINGS.cache.username and config.SETTINGS.cache.password:
+                    credential_provider = UsernamePasswordCredentialProvider(
+                        username=config.SETTINGS.cache.username, password=config.SETTINGS.cache.password
+                    )
                 self.connection = redis.Redis(
                     host=config.SETTINGS.cache.address,
                     port=config.SETTINGS.cache.service_port,
                     db=config.SETTINGS.cache.database,
-                    ssl=config.SETTINGS.cache.tls_enabled,
```
ogenstad, line 282: This looks like some copy and paste error to remove the ssl param?

FragmentedPacket: Fixed.

ogenstad (approved):

dgarros:
would be good to add an entry in the changelog, also are we confident to add it directly to stable ? @ogenstad @fatih-acar

fatih-acar (approved):
LGTM for pushing into stable

ogenstad:
Yeah I think it should be fine to push to `stable`.

PR #7996, opened 2025-12-31
Fix false deprecation warnings for parameters in TextAttributeSchema

### LINKED ISSUE #7995: bug: False deprecation warnings for min_length/max_length when correctly using parameters
### Component

API Server / GraphQL

### Infrahub version

1.6.2

### Current Behavior

When loading a schema that correctly uses the parameters block for min_length, max_length, and regex on text attributes, the system incorrectly generates deprecation warnings claiming these values are set at the top level. Example deprecation warnings received:

```
deprecation: Use of 'max_length' on attributes is deprecated, use parameters instead [NetworkDevice.hostname]
deprecation: Use of 'min_length' on attributes is deprecated, use parameters instead [NetworkDevice.hostname]
deprecation: Use of 'max_length' on attributes is deprecated, use parameters instead [NetworkDevice.model]
deprecation: Use of 'max_length' on attributes is deprecated, use parameters instead [NetworkDevice.platform]
deprecation: Use of 'max_length' on attributes is deprecated, use parameters instead [NetworkInterface.name]
deprecation: Use of 'min_length' on attributes is deprecated, use parameters instead [NetworkInterface.name]
deprecation: Use of 'max_length' on attributes is deprecated, use parameters instead [NetworkInterface.description]
```

### Expected Behavior

No deprecation warnings should be generated when min_length, max_length, and regex are correctly defined within the parameters block. Deprecation warnings should only appear when these values are set at the attribute's top level (the deprecated syntax).

### Steps to Reproduce

1. Create a schema file with attributes using the correct parameters syntax:
```
version: "1.0"
nodes:
  - name: Device
    namespace: Network
    attributes:
      - name: hostname
        kind: Text
        unique: true
        parameters:
          regex: "^[a-zA-Z][a-zA-Z0-9._-]*$"
          min_length: 1
          max_length: 63
      - name: model
        kind: Text
        optional: true
        parameters:
          max_length: 128
  - name: Interface
    namespace: Network
    attributes:
      - name: name
        kind: Text
        parameters:
          regex: "^[a-zA-Z][a-zA-Z0-9/._-]*$"
          min_length: 1
          max_length: 64
      - name: description
        kind: TextArea
        optional: true
        parameters:
          max_length: 256
```
2. Load the schema via the API or infrahubctl schema load
3. Observe the deprecation warnings in the response/logs

### Additional Information

**Root Cause Analysis:** The issue occurs in the gather_warnings() method in backend/infrahub/core/schema/__init__.py. The deprecation check examines the top-level min_length/max_length fields on the attribute, but these fields get populated during schema processing due to the reconcile_parameters method in TextAttributeSchema. The reconcile_parameters method (in backend/infrahub/core/schema/attribute_schema.py) synchronizes values between the top-level fields and parameters:
- If a value is set in parameters, it gets copied to the top level
- If a value is set at top level only, it gets copied to parameters
By the time gather_warnings() runs, both locations have the same values, making it impossible to distinguish between correct usage (parameters) and deprecated usage (top-level).
**Affected Code Locations:**
- backend/infrahub/core/schema/__init__.py - SchemaRoot.gather_warnings() method (lines ~119-135)
- backend/infrahub/core/schema/attribute_schema.py - TextAttributeSchema.reconcile_parameters() method (lines ~241-252)

FragmentedPacket:
## 
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema attribute deprecation warnings to specifically identify which fields are deprecated rather than using generic messages.
  * Enhanced detection system to correctly distinguish between deprecated field usage and proper parameter-based configuration.

- Fixes false deprecation warnings when `min_length`, `max_length`, or `regex` are correctly defined in the `parameters` block
- Tracks deprecated field usage during `reconcile_parameters` to distinguish between top-level (deprecated) and parameters (correct) usage
- Adds test coverage for correct parameters usage

## Test plan

- [ ] Verify no deprecation warnings when using `parameters` block correctly
- [ ] Verify deprecation warnings still appear when using top-level fields (deprecated syntax)
- [ ] Run unit tests: `uv run pytest backend/tests/unit/core/test_schema.py::test_schema_warnings -v`

Fixes #7995

### DIFF
--- a/backend/infrahub/core/schema/__init__.py
+++ b/backend/infrahub/core/schema/__init__.py
@@ -117,20 +117,15 @@ class SchemaRoot(BaseModel):
                     )
                 )
             for attribute in model.attributes:
-                if attribute.max_length is not None:
+                # Check if the attribute has tracked any deprecated field usage
+                # This is set by reconcile_parameters when values are synced from top-level to parameters
+                deprecated_fields = getattr(attribute, "_deprecated_fields_used", set())
+                for field_name in deprecated_fields:
                     warnings.append(
                         SchemaWarning(
                             type=SchemaWarningType.DEPRECATION,
                             kinds=[SchemaWarningKind(kind=model.kind, field=attribute.name)],
-                            message="Use of 'max_length' on attributes is deprecated, use parameters instead",
-                        )
-                    )
-                if attribute.min_length is not None:
-                    warnings.append(
-                        SchemaWarning(
-                            type=SchemaWarningType.DEPRECATION,
-                            kinds=[SchemaWarningKind(kind=model.kind, field=attribute.name)],
-                            message="Use of 'min_length' on attributes is deprecated, use parameters instead",
+                            message=f"Use of '{field_name}' on attributes is deprecated, use parameters instead",
                         )
                     )
 
--- a/backend/infrahub/core/schema/__init__.py
+++ b/backend/infrahub/core/schema/__init__.py
@@ -119,8 +119,8 @@ class SchemaRoot(BaseModel):
             for attribute in model.attributes:
                 # Check if the attribute has tracked any deprecated field usage
                 # This is set by reconcile_parameters when values are synced from top-level to parameters
-                deprecated_fields = getattr(attribute, "_deprecated_fields_used", set())
-                for field_name in deprecated_fields:
+                deprecated_fields: set[str] = getattr(attribute, "_deprecated_fields_used", set())
+                for field_name in sorted(deprecated_fields):
                     warnings.append(
                         SchemaWarning(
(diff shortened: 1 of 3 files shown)


ajtmccarty:
superseded by #7998 

the approach in this PR is simpler, but it violates the Single Responsibility Principle in a way that I want to avoid adding to the code going forward. The `AttributeSchema` class (or any subclasses of it) should not be responsible for tracking what its initial state was. All of our schema-related classes should only be responsible for tracking the state of a schema (Node/Attribute/Relationship/etc.) right now. They should just be thin containers for holding data

ajtmccarty:
fixed in #7998

PR #7997, opened 2026-01-01
special filtering for created/updated_at at midnight

ajtmccarty:
IFC-1947

- add new special handling that allows filtering for a whole day's worth of metadata if the input time is midnight in any timezone
  - basically, if you filter to Interfaces with `node_metadata__created_at=2025-12-31T00:00:00`, then this will be translated to `node_metadata__created_at__after=2025-12-30T23:59:59.999999` and `node_metadata__created_at__before=2026-01-01T00:00:00`
- updates to use class-scoped fixtures for some existing metadata filtering tests to make them run faster. probably a good pattern to use in other places

## Summary by CodeRabbit

* **Bug Fixes**
  * Metadata datetime filtering: exact-midnight created_at/updated_at values are now treated as full-day ranges when querying.

* **Behavior**
  * Day-range interpretation is applied automatically during query handling so midnight datetimes return that day's results.

* **Tests**
  * Added comprehensive unit tests for midnight/day-range behavior and many edge cases.
  * Improved test infrastructure with class-scoped fixtures and reusable schema helpers for better isolation.

### DIFF
--- a/backend/infrahub/graphql/resolvers/resolver.py
+++ b/backend/infrahub/graphql/resolvers/resolver.py
@@ -148,6 +148,9 @@ def _transform_metadata_day_filters(filters: dict[str, Any]) -> dict[str, Any]:
     When a filter like `node_metadata__created_at="2025-02-03T00:00:00"` has a time
     of exactly midnight, transform it into __after and __before filters to match
     the entire day (inclusive of midnight).
+
+    If __after or __before filters are already explicitly defined, they will not be
+    overwritten by the generated day range filters.
     """
     result = dict(filters)
     metadata_datetime_fields = ("node_metadata__created_at", "node_metadata__updated_at")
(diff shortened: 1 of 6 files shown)


backend/infrahub/graphql/resolvers/resolver.py
```diff
@@ -167,6 +199,7 @@ async def default_paginated_list_resolver(
         filters = {
             key: value for key, value in kwargs.items() if ("__" in key and value is not None) or key in ("ids", "hfid")
         }
+        filters = _transform_metadata_day_filters(filters)
```
ajtmccarty, line 210: I don't think we do this kind of filter transform anywhere else and we don't have a great system for it b/c filters are just dictionaries with string keys.
it would be better if this `filters` input was more structured and then this logic could be linked to that structure

backend/tests/conftest.py
```diff
@@ -170,6 +180,13 @@ async def default_branch(reset_registry, local_storage_dir, empty_database, db:
     return await do_default_branch(db=db)
 
 
+@pytest.fixture(scope="class")
+async def default_branch_scope_class(
+    reset_registry_scope_class, local_storage_dir_scope_class, empty_database_scope_class, db: InfrahubDatabase
```
ogenstad: Can you add type annotations to these fixtures? It will soon be enforced based on other PRs and we could end up with failing pilelines if some branches are out of date before we merge them.

backend/tests/unit/conftest.py
```diff
@@ -1723,11 +1732,24 @@ async def criticality_schema(
...
-    )
+    return do_criticality_schema(branch=default_branch, schema_root=criticality_schema_root)
+
+
+@pytest.fixture(scope="class")
+async def criticality_schema_scope_class(
+    db: InfrahubDatabase,
+    default_branch_scope_class: Branch,
+    group_schema_scope_class,
+    data_schema_scope_class,
```
ogenstad: Same comment as above.

ogenstad (approved):

backend/tests/conftest.py
ogenstad: FYI: #7985

PR #8011, opened 2026-01-02
support migrating attribute kind on profiles

### LINKED ISSUE #7943: task: better support for switching attribute types with different parameters
### Component

API Server / GraphQL

### Task Description

Infrahub does not have great support for switching Attribute types if those Attributes use different parameters. For example, switching from a `kind="Text"` attribute to a `kind="JSON"` attribute will raise a `TextAttributeParameters can't be used as parameters for JSON` error, which the user cannot resolve

Using pydantic discriminators might help resolve these issues

ajtmccarty:
fixes #7943

- adds support to correctly migrate attributes on profiles when their associated node schema is updated
- adds support to correctly migrate attributes on templates when their associated node schema is updated
- adds support for turning one kind of AttributeParameters into another. for example, `TextAttributeParamters` to plain `AttributeParameters` or vice versa which would be necessary when converting Text to/from URL or something

## Summary by CodeRabbit

* **New Features**
  * Added cross-kind parameter conversion helpers to safely migrate attribute parameters between kinds.

* **Bug Fixes / Improvements**
  * Query selection expanded to include profile and template schema variants so migrations and validations apply consistently across variants.
  * Safer parameter handling when changing attribute kinds; non-matching fields are dropped and matching fields preserved.

* **Tests**
  * Added unit and integration tests covering parameter conversions and multi-kind (profile/template) migration scenarios.

### DIFF
--- a/backend/infrahub/core/migrations/schema/attribute_kind_update.py
+++ b/backend/infrahub/core/migrations/schema/attribute_kind_update.py
@@ -135,7 +135,9 @@ CALL (has_value_e) {
     SET has_value_e.to = $at
 }
         """ % {
-            "schema_kinds": f"{self.migration.previous_schema.kind}|Profile{self.migration.previous_schema.kind}",
+            "schema_kinds": (
+                f"{self.migration.previous_schema.kind}|Profile{self.migration.previous_schema.kind}|Template{self.migration.previous_schema.kind}"
+            ),
             "branch_filter": branch_filter,
             "new_attr_value_labels": new_attr_value_labels,
         }
--- a/backend/infrahub/core/migrations/schema/attribute_kind_update.py
+++ b/backend/infrahub/core/migrations/schema/attribute_kind_update.py
@@ -34,7 +34,7 @@ class AttributeKindUpdateMigrationQuery(AttributeMigrationQuery):
 // ------------
 // start with all the Attribute vertices we might care about
 // ------------
-MATCH (n:%(schema_kind)s)-[:HAS_ATTRIBUTE]->(attr:Attribute)
+MATCH (n:%(schema_kinds)s)-[:HAS_ATTRIBUTE]->(attr:Attribute)
 WHERE attr.name = $attr_name
 WITH DISTINCT n, attr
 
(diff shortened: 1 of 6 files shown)


backend/infrahub/core/schema/attribute_schema.py
```diff
@@ -114,13 +114,24 @@ def validate_dropdown_choices(cls, values: Any) -> Any:
...
+        Fields with the same name in both classes are preserved.
+        """
         kind = info.data["kind"]
         expected_parameters_class = get_attribute_parameters_class_for_kind(kind=kind)
         if value is None:
             return expected_parameters_class()
         if not isinstance(value, expected_parameters_class) and isinstance(value, AttributeParameters):
-            return expected_parameters_class(**value.model_dump())
+            return expected_parameters_class.convert_from(value)
+        if isinstance(value, dict):
```
dgarros, line 129: this could be converted into a `convert_from_dict()` method and it could be reused within `convert_from`

backend/infrahub/core/migrations/schema/attribute_kind_update.py
```diff
@@ -136,7 +135,7 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
     SET has_value_e.to = $at
 }
         """ % {
-            "schema_kind": self.migration.previous_schema.kind,
+            "schema_kinds": f"{self.migration.previous_schema.kind}|Profile{self.migration.previous_schema.kind}",
```
dgarros: While it's not directly related to this PR, I'm wondering if we should add a similar support for the Template ?

dgarros (approved):

backend/infrahub/core/schema/attribute_schema.py
ajtmccarty: good suggestion. added it

backend/infrahub/core/migrations/schema/attribute_kind_update.py
ajtmccarty: yes you are correct. added support for Templates with tests

PR #8015, opened 2026-01-03
schema update/migrations w/ rollback

### LINKED ISSUE #6948: bug: schema update with a failed migration can leave database in illegal state
### Component

API Server / GraphQL

### Infrahub version

1.3.3

### Current Behavior

if a schema update fails during the migration stage, then the schema will remain updated to the new state while the data will still exist in the old state

### Expected Behavior

we should make sure that the schema and data are in sync after a schema update

### Steps to Reproduce

this is very hard to reliably reproduce without editing code, but this is the order of operations
1. load an update to a schema that includes a migration, such as updating the inheritance or kind of a node or updating the name of an attribute
2. the schema load endpoint will update the schema (SchemaNode, SchemaAttribute, SchemaRelationship, etc. vertices at the database level) before running the migrations to update the data. If any of those migrations fail, the schema will be left in the updated state, while the data will still conform to the old state

### Additional Information

it would be nice if we could put the whole thing (schema update and migrations) into a database transaction, but we cannot because there could be so many changes that the transaction would crash

I think the ideal solution is to allow doing a rollback of the schema update and migration-related changes, so that if the schema update or any of the migrations fail, then we run a hard-delete/update on all of the added/updated edges

we do something similar for merging a branch, where all changes have the same microsecond timestamp and if there is an error, then we use that timestamp to rollback the merge

ajtmccarty:
fixes #6948 
IFC-1267

TODO:
- [x] check if this resolves https://opsmill.atlassian.net/browse/IFC-1267
  - it does
- [x] tests for schema rollback on merge and rebase failures
- [x] double check `origin_branch` argument

new `SchemaUpdateCoordinator` component that handles updating the schema in the database, updating the schema in the registry, and running any necessary migrations. Ensures that all of these changes have the same timestamps at the database level and will roll all of the changes back (schema and data) if any part fails.

One known outstanding issue. if a the schema update or migrations fail on the default branch, any `created/updated_at/by` metadata on the default branch updated during the migrations will NOT be reverted. this is only for the default branch and only for schema migrations.

## Summary by CodeRabbit

* **New Features**
  * Schema updates and associated data migrations are now coordinated and executed as a single atomic operation.
* **Bug Fixes**
  * Failures during migrations automatically roll back both database and schema registry state to preserve consistency.
* **Tests**
  * New unit and integration tests verify update/migration atomicity and rollback behavior.
* **Documentation**
  * Changelog entry added describing the atomic update-and-migrate behavior.


backend/infrahub/api/schema.py
```diff
@@ -422,6 +411,8 @@ async def load_schema(
     )
     await service.event.send(event=event)
 
+    updated_hash = update_result.updated_hash or ""
```
ogenstad: I'm guessing that the returned object can be generated when there's a failure where updated_hash could potentially be `None` and that we shouldn't get there in this case. But this looks a bit off as if we expect that the execute method would succeed but could potentially return an updated_hash that hasn't been set which looks to be the case if the migrations param is empty.

This might just be a typing issue but the hash will always have a value unless an exception is raised?

backend/infrahub/cli/db.py
```diff
@@ -503,42 +502,33 @@ async def update_core_schema(db: InfrahubDatabase, initialize: bool = True, debu
...
+            update_registry=True,
         )
-        default_branch.update_schema_hash()
-        get_migration_console().log(
-            "The Core Schema has been updated, make sure to rebase any open branches after the upgrade"
-        )
-        if debug:
-            get_migration_console().log(f"New schema hash: {default_branch.active_schema_hash.main}")
-        await default_branch.save(db=dbt)
+    except Exception as exc:
```
ogenstad, line 523: The coordinator.execute() method returns a SchemaUpdateResult object that contains a .exception attribute and a .success. It looks like success will be `False` if any exceptions exist within the object. Since we're catching any exceptions in this case I'm wondering about the relationship between this try/except clause and the .exceptions attribute on the object. The concern would be if we at some point populate the exceptions attribute instead of raising an error and then ignore the error when we get to this part of the code.

ogenstad (commented):
Looks good to me overall but raised some questions regarding the exception handling.

backend/infrahub/api/schema.py
ajtmccarty: updated the logic and type-hinting so that the `updated_hash: str` is always returned if the `update_db` or `update_registry` params are True.
it is possible for `SchemaUpdateCoordinator.execute` to be run with `update_db=False` and `update_registry=False`, in which case, only the migrations would be run and the method would not have an updated hash to return, so it would be `None`. We don't have any code that currently does this.

backend/infrahub/cli/db.py
ajtmccarty: made the return simpler for `SchemaUpdateCoordinator.execute` so that it just returns the updated hash and will raise any errors it encounters during the migrations

backend/infrahub/core/branch/tasks.py
```diff
@@ -166,45 +169,43 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
...
 
-        schema_in_main_before = merger.destination_schema.duplicate()
+        pre_rebase_schema = merger.destination_schema.duplicate()
         migrations = []
         async with lock.registry.global_graph_lock():
             async with db.start_transaction() as dbt:
-                rebase_at = Timestamp()
                 await obj.rebase(db=dbt, user_id=context.account.account_id, at=rebase_at)
-                log.info("Branch successfully rebased")
+                log.info("Branch graph successfully rebased")
```
ogenstad: Not sure if "branch graph" is easier to understand for users than just branch.

backend/infrahub/core/schema/update_coordinator.py
```diff
@@ -0,0 +1,329 @@
...
+                errors=error_msgs,
+            )
+
+        # Step 4: Rollback on failure
+        await self._rollback(at=at)
+        await self._restore_registry_state()
+
+        self.log.info("Schema rollback completed", branch=self.branch.name)  # type: ignore[call-arg]
+
+        # Re-raise the appropriate error
```
ogenstad: Gotta love these AI comments :)

backend/infrahub/core/schema/update_coordinator.py
```diff
@@ -0,0 +1,329 @@
...
+                "Schema migration returned errors, beginning rollback",
+                branch=self.branch.name,
+                errors=error_msgs,
+            )
+
+        # Step 4: Rollback on failure
+        await self._rollback(at=at)
+        await self._restore_registry_state()
+
+        self.log.info("Schema rollback completed", branch=self.branch.name)  # type: ignore[call-arg]
```
ogenstad: The type ignore here, is that something that relates to the type of logger used for the class? I.e. would the branch argument be problematic if the wrong type of logger gets used or why are we seeing this?

ogenstad (approved):

backend/infrahub/core/branch/tasks.py
ajtmccarty: true. not sure exactly what to put here. updated to just `Branch graph rebased` so that it doesn't sound like the whole rebase operation is complete

backend/infrahub/core/schema/update_coordinator.py
ajtmccarty: DELETED

backend/infrahub/core/schema/update_coordinator.py
ajtmccarty: the issue is that `get_run_logger()`, which we use for the logger in prefect tasks, can return a plain `Logger`, that does not except arbitrary kwargs
so I've updated these to use the older `extra={"branch", self.branch.name}` format instead
thanks for pointing it out

PR #7769, opened 2025-11-30
Add Claude Code command for refactoring Neo4j Query classes

dgarros:
## 
## Summary by CodeRabbit

* **Documentation**
  * Added detailed guidance for a standardized backend query pattern, migration steps, testing guidance, and performance recommendations.

* **Refactor**
  * Replaced ad-hoc result containers with typed data models and updated queries to return structured, typed results with richer docstrings and clearer public signatures.

* **Chores**
  * Expanded markdown linting/formatting targets to include dot-directories and apply fixes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

- Introduces a new Claude Code slash command (`/backend-query-refactor`) that automates the process of refactoring Neo4j Query classes to use typed Pydantic data classes
- Applies the refactoring pattern to three existing queries as examples: `AttributeUpdateValueQuery`, `NodeListGetAttributeQuery`, and `RelationshipGetByIdentifierQuery`

## Description

This PR establishes a standardized pattern for making Query class results more type-safe and performant by:

1. **Adding Pydantic data classes** to represent query return data with `Field(description=...)` for documentation
2. **Optimizing RETURN clauses** to return only specific values needed (e.g., `n.uuid AS node_uuid`) instead of entire Neo4j nodes/relationships
3. **Using `model_construct()`** for performance (skips validation since we control the data format)
4. **Converting existing `@dataclass` to Pydantic** for consistency across the codebase
5. **Adding docstrings** to Query classes explaining their purpose

### Changes Applied

| File | Change |
|------|--------|
| `backend/infrahub/core/query/attribute.py` | Added `AttributeUpdateValueQueryData` Pydantic class, optimized RETURN clause, added `get_data()` method |
| `backend/infrahub/core/query/node.py` | Added docstring to `NodeListGetAttributeQuery` (RETURN clause optimization not possible due to `_properties` requirement) |
| `backend/infrahub/core/query/relationship.py` | Converted `RelationshipPeersData` from `@dataclass` to Pydantic, optimized RETURN clause, added docstring |

### The Command

The `/backend-query-refactor` command provides a 12-step workflow:
1. Locate the Query class
2. Analyze usage patterns
3. Understand query purpose
4. Design appropriate Pydantic data class
5. Implement the data class
6. Add Query class documentation
7. Optimize RETURN clause
8. Add typed `get_data()` method
9. Analyze performance (Cypher + memory)
10. Check for existing tests
11. Propose new tests if needed
12. Optionally update callers

## Test plan
[...]

PR #8020, opened 2026-01-03
Refactor Database Queries to return structured data

dgarros:
Replaces https://github.com/opsmill/infrahub/pull/7769

This PR implement the change started in #7769 to create a cleaner boundary between the Query Result and the rest of the code by leveraging a dataclass to return all the results.

Since we already discussed it and it felt like there was a general agreement this is something we want to do, I went ahead and migrated all (most) of the queries. As such, it felt like the command wasn't needed anymore.
Instead, I created a new entry in `dev/knowledge/backend` to capture the best practices around Query. I purposefully didn't include the latest change that were introduced in https://github.com/opsmill/infrahub/pull/7979 since the design is still being discussed. Once we have finalized the design it would be good to update the KB. 

## Summary by CodeRabbit

* **Refactor**
  * Improved internal data access patterns across query and resource management modules for enhanced type safety and consistency.
  * Consolidated query result handling to use unified typed data access methods throughout the backend system.

* **Documentation**
  * Updated Python coding guidelines with frozen dataclass recommendations and version compatibility details.
  * Added comprehensive Query pattern documentation for backend development practices.

### DIFF
--- a/backend/infrahub/core/query/ipam.py
+++ b/backend/infrahub/core/query/ipam.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, Iterable
 from infrahub.core.constants import InfrahubKind
 from infrahub.core.graph.schema import GraphAttributeIPHostNode, GraphAttributeIPNetworkNode
 from infrahub.core.ipam.constants import AllIPTypes, IPAddressType, IPNetworkType
-from infrahub.core.query import QueryType
+from infrahub.core.query import QueryResult, QueryType
 from infrahub.core.registry import registry
 from infrahub.core.utils import convert_ip_to_binary_str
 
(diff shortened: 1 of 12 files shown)


backend/infrahub/graphql/queries/resource_manager.py
```diff
@@ -152,19 +151,21 @@ async def resolve(
                 identifier_query_class = identifier_query_map.get(pool.get_kind())
                 if not identifier_query_class:
                     raise ValidationError(input_value=f"This query doesn't get support {pool.get_kind()}")
-                identifier_query = await identifier_query_class.init(
-                    db=graphql_context.db, at=graphql_context.at, pool_id=pool_id, allocated=allocated_ids
+                identifier_query = cast(
```
ogenstad, line 154: Not a blocker, but I think we can get rid of this `cast()` by adding type-hints to describe the content of the `identifier_query_map` dictionary. But we can get back to that later. I think we should have a housekeeping issue at some point to remove `cast` wherever possible.

ogenstad (approved):

backend/infrahub/core/query/ipam.py
```diff
@@ -242,6 +242,48 @@ async def get_ip_addresses(
...
+        av = result.get_node("av")
+        return cls(
+            prefix_uuid=str(pfx.get("uuid")),
+            child_uuid=str(child.get("uuid")),
+            child_kind=child.get("kind"),
+            child_labels=tuple(child.labels),
+            ip_value=av.get("value"),
+            prefixlen=av.get("prefixlen"),
+            branch=str(result.get("branch")),
+        )
```
ajtmccarty, line 284: the use of dataclasses is a big improvement. 👍 
I'd say that the Query class should be responsible for constructing instances of this class. these data classes should only be responsible for holding data. they should not need to have logic related to interacting with `QueryResult`. the logic for interpreting cypher results and building a data transfer object is strongly coupled and should live in the Query class

PR #8032, opened 2026-01-05
IFC-2041: Add Builtin to restricted namespaces

solababs:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile and schema management so items in the Builtin namespace are treated differently when generating, skipping, or deleting profiles.

* **Tests**
  * Standardized test data to use a "Testing" namespace and updated tag-related tests, generators, and assertions.
  * Added a Testing tag kind, test constant, fixtures, and a helper tag schema to support the new test data.

* **Chores**
  * Added "Builtin" to the restricted namespaces configuration.


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2205,16 +2205,18 @@ def manage_profile_schemas(self) -> None:
             self.set(name=core_profile_schema_definition.kind, schema=core_profile_schema_definition)
 
         profile_schema_kinds = set()
+        restricted_namespace_profiles_to_create = (InfrahubKind.TAG, InfrahubKind.IPPREFIX, InfrahubKind.IPADDRESS)
         for node_name in self.node_names + self.generic_names_without_templates:
             node = self.get(name=node_name, duplicate=False)
             if (
                 node.namespace in RESTRICTED_NAMESPACES
```
ajtmccarty: this is missing support for profiles on `BuiltinIPNamespace`, which most likely no one is using, but it is possible
it would be easier and more future proof to change this line in the conditional to `(node.namespace in RESTRICTED_NAMESPACES and node.namespace != "Builtin")` and it accomplishes the same thing you're doing with `restricted_namespace_profiles_to_create`
you could even add a comment saying this is for legacy support of some sort
when we add new schemas to the `Builtin` namespace, we can turn off profile support for them manually using the `generate_profiles` property of the schema

schema/schema.graphql
ajtmccarty, line 1: looks like this needs to be updated to re-add the profiles for the `Builtin` schemas

ajtmccarty (approved):

ajtmccarty (approved):

ogenstad (approved):
LGTM, it would be good with a news fragment to highlight that we added Builtin to the restricted list of namespaces.

PR #8039, opened 2026-01-06
IFC-1305 Deprecate permission identifier and use computed display label

gmazoyer:
This changes convert the permissions' display labels so that they now use Jinja2 templates in order to have more meaningful strings. By doing this, we end up having `display_label == identifier` when it comes to permissions. So the change also deprecates the `identifier` attribute for permissions, code that uses it should be changed to use the display label now. Once this done, we will be able to cleanup the custom code that we wrote for permissions identifiers.

## Summary by CodeRabbit

* **New Features**
  * Infrahub-specific Jinja2 filters and a preconfigured InfrahubJinja2Template for computing permission display labels.
* **Migrations**
  * Migration to recompute and persist permission display_label values with branch-aware, batched updates.
* **Tests**
  * Component tests for display_label generation, migration behavior, idempotence, and branch handling.
* **Deprecations**
  * Permission identifier marked deprecated; use display_label instead.
* **Other**
  * Internal graph version incremented.


ogenstad:
I think we'd need to have a migration to populate the data in the updated `display_label` fields as well, should probably be done after https://github.com/opsmill/infrahub/pull/8092 is merged to have the correct version of the migration.

backend/infrahub/core/node/permissions.py
```diff
@@ -56,11 +52,7 @@ async def to_graphql(
...
         )
 
-        if fields:
-            if "identifier" in fields:
-                decision = PermissionDecisionFlag(value=self.decision.value.value)  # type: ignore[attr-defined]
-                response["identifier"] = {
-                    "value": f"object:{self.namespace.value}:{self.name.value}:{self.action.value.value}:{decision.name.lower()}"  # type: ignore[attr-defined,union-attr]
-                }
+        if fields and "identifier" in fields:
+            response["identifier"] = {"value": await self.get_display_label(db=db)}
```
ogenstad, line 56: I'm wondering if this is something that we just need as a temporary thing during a migration or if we can remove it later? One of the goals was to be able to have a user friendly label when selecting the permission nodes in the UI which I think we'd get with the display label, after that I wonder if the "identifier" provides much value but perhaps stuff would break if we remove it? I.e. frontend forms or something like that..

backend/infrahub/core/node/__init__.py
```diff
@@ -3,11 +3,11 @@
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Sequence, TypeVar, overload
 
-from infrahub_sdk.template import Jinja2Template
 from infrahub_sdk.utils import is_valid_uuid
 from infrahub_sdk.uuidt import UUIDT
 
 from infrahub import lock
+from infrahub.computed_attribute.jinja2 import InfrahubJinja2Template
```
ogenstad, line 10: We've talked about avoiding to import stuff outside of infrahub.core within the core itself, having said that we're violating this in a number of locations as it is now and it could be that this is something that would be refactored when we get to changing how Node and NodeManager interacts with each other. I'm not sure either that we'd want to add more stuff to infrahub.core now and include the Jinja logic there, so mainly an observartion.

backend/infrahub/core/node/permissions.py
gmazoyer: Yes this is just temporary until we remove the `identifier`. I was proposing a smooth transition by having identifier == display label so frontend can migrate as they wish. And once the migration done, we can have another PR removing the `identifier`.

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+    """Get permission nodes with their attribute values for computing display_label."""
+
+    name = "get_permission_attributes"
+    type = QueryType.READ
+    insert_return = False
+
+    def __init__(self, permission_kind: str, is_branch_agnostic: bool = False, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.permission_kind = permission_kind
+        self.is_branch_agnostic = is_branch_agnostic
```
ajtmccarty: `Query` base class actually already has a `branch_agnostic` parameter that can be used here I think

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+            "permission_kind": self.permission_kind,
+            "branch_names": [GLOBAL_BRANCH_NAME] if self.is_branch_agnostic else [registry.default_branch],
+        }
+
+        if self.is_branch_agnostic:
+            query = """
+MATCH (n:Node {kind: $permission_kind})-[e:IS_PART_OF]->(:Root)
+WHERE e.branch IN $branch_names
+AND e.status = "active"
+AND e.to IS NULL
```
ajtmccarty, line 68: I'd add a `AND NOT exists((n)-[:IS_PART_OF {branch: e.branch, status: "deleted"}]->())` here to make completely sure no deleted Nodes are included

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+    RETURN decision_val.value AS decision_value
+}
+RETURN n.uuid AS node_uuid, namespace_value, name_value, action_value, decision_value
+            """
+
+        self.add_to_query(query)
+        if self.is_branch_agnostic:
+            self.return_labels = ["node_uuid", "action_value", "decision_value"]
+        else:
+            self.return_labels = ["node_uuid", "namespace_value", "name_value", "action_value", "decision_value"]
```
ajtmccarty: not that big of a deal for a migration, but these seem like they should be two separate queries if they use completely separate cypher and return different data

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+    RETURN decision_val.value AS decision_value
+}
+RETURN n.uuid AS node_uuid, namespace_value, name_value, action_value, decision_value
+        """ % {"branch_filter_r": branch_filter_r, "branch_filter_r2": branch_filter_r2}
+
+        self.add_to_query(query)
+        self.return_labels = ["node_uuid", "namespace_value", "name_value", "action_value", "decision_value"]
+
+    def get_results_as_dicts(self) -> list[dict[str, Any]]:
+        return [{label: result.get(label) for label in self.return_labels} for result in self.get_results()]
```
ajtmccarty: best practice is to have these methods return more structured data, like a dataclass, so that you can do `row.name` instead of `row.get("name")` down below, but maybe it's easier to use dicts in this case if you are only returning the name/namespace for some of the queries

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+                "name__value": row.get("name_value"),
+                "action__value": row.get("action_value"),
+                "decision__value": row.get("decision_value"),
+            }
+
+            if any(v is None for v in variables.values()):
+                continue
+
+            jinja_template = InfrahubJinja2Template(template=core_object_permission.display_label)
+            rendered = await jinja_template.render(variables=variables)
```
ajtmccarty: might be worth catching possible errors here
recently saw  customer issue where a `JinjaTemplateUndefinedError` crashed a migration

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,425 @@
...
+    MigrationInput,
+    MigrationRequiringRebase,
+    MigrationResult,
+    get_migration_console,
+)
+from infrahub.core.query import Query, QueryType
+from infrahub.core.schema.definitions.core.permission import core_global_permission, core_object_permission
+
+from .load_schema_branch import get_or_load_schema_branch
+from .m047_backfill_or_null_display_label import UpdateAttributeValuesQuery
```
ajtmccarty: maybe this class should move to its own file. we keep reusing it

backend/tests/component/core/migrations/graph/test_059_recompute_permission_display_labels.py
ajtmccarty, line 1: Patrick had the good suggestion to give these test files names that correspond to the migration they are testing (so `test_059_recompute_permission_display_labels` in this case) so that it is easier to move them around when there are conflicts merging `stable` into `develop`

backend/tests/component/core/migrations/graph/test_059_recompute_permission_display_labels.py
```diff
@@ -0,0 +1,181 @@
...
+                continue
+            result_map[node_id] = node_attributes_map[node_id].attrs["display_label"].value
+        return result_map
+
+    async def set_display_label_value(self, db: InfrahubDatabase, node_uuid: str, value: str) -> None:
+        query = """
+        MATCH (n:Node {uuid: $node_uuid})-[:HAS_ATTRIBUTE]->(a:Attribute {name: "display_label"})
+        MATCH (a)-[hv:HAS_VALUE]->(av:AttributeValue)
+        WHERE hv.status = "active" AND hv.to IS NULL
+        SET av.value = $value
```
ajtmccarty, line 43: just FYI
this will change the value for every node currently linked to the `AttributeValue`
so if there are two nodes with display label `a` and you want to update just one of them to be `b`, this query is not capable of doing that

backend/tests/component/core/migrations/graph/test_059_recompute_permission_display_labels.py
gmazoyer: Good point

backend/infrahub/core/migrations/graph/m059_recompute_permission_display_labels.py
gmazoyer: Moved it to a dedicated file under the query sub-directory. While doing that I saw that we seem to use a variation of this query in migration 44 and 46. I didn't check if we could replace it with the one in the file yet, but this is something we might want to have a look at.

backend/infrahub/core/node/__init__.py
gmazoyer: I know, I somehow accepted this trade-off for now because of two reasons:
1. I did not really know where to put `InfrahubJinja2Template` heh :-)
2. I thought that we will probably solve this when we'll dig into refactoring the `Node` class, which we agreed needs refactoring

backend/tests/component/core/migrations/graph/test_059_recompute_permission_display_labels.py
gmazoyer: I think that is alright for this specific test. I don't really care about having distinct display labels, I only need to validate that after the migration has run, the display labels are properly set.

ajtmccarty (approved):

PR #8041, opened 2026-01-06
File object feature spec

wvandeun:
## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive File Object design spec covering a user-extensible core file type with standard metadata (name, checksum, size, type, storage id), attachment relationships and cardinalities, versioning and branch semantics, permission integration, configurable max file size, REST and GraphQL upload/download/query surfaces, Python SDK workflows, frontend rendering/permission considerations, and open UI/integration questions and future considerations.


dev/specs/2026-01-file-object.md
```diff
@@ -0,0 +1,634 @@
...
+      signed_by: {id: ""},
+    }
+  ) {
+    ok
+    __typename
+    object {
+      id
+    }
+  }
+}
```
ogenstad, line 368: I'd suggest that we instead have something like:

```graphql
mutation {
  NetworkCircuitContractCreate(
    data: {
      name: {value: ""}, 
      circuit: {id: ""},
      contract_start: {value: ""},
      contract_end: {value: ""},
      signed_by: {id: ""},
    }
    file: { <- new option specifically available on attachment types
      <- content of file
    }
  ) {
    ok
    __typename
    object {
      id
    }
  }
}
```

dev/specs/2026-01-file-object.md
```diff
@@ -0,0 +1,634 @@
...
+
+```
+POST /api/CoreFileObject/upload
+Body: {"file": "String"} # binary
+Return: {"identifier": {identifier}, "checksum": "String"}
+```
+
+#### Open questions
+
+- What is the right order of operation? Do we first create the `CoreFileObject` object and then the object in the storage system, or the other way around?
```
ogenstad, line 468: Also:

- How do we handle situations where someone tries to use the old storage API to download a file. I.e. just supplies the UUID of a file attachment node as the storage ID where there are no permissions. Immediate options would include a) introducing permissions for artifacts within the storage, this will slow things down for artifacts though b) have separate storage "folders" so that we can have UUIDs without any permissions attached to them and also UUIDs where we require permissions but stored in a different location.

dev/specs/2026-01-file-object.md
```diff
@@ -0,0 +1,634 @@
...
+from infrahub_sdk import InfrahubClientSync
+
+client = InfrahubClientSync()
+
+# Creating a CoreFileAttachment
+
+account = client.get(kind=CoreUserAccount, id=<uuid>)
+circuit = client.get(kind=InfraCircuit, id=<uuid>)
+
+identifier = client.file_object.upload(content=Path("/tmp/contract.pdf").read())
```
ogenstad, line 492: Small note on this as we're reading files using the pathlib module. The way this is written we just have an example for how to read a file using pathlib. If this would be the end state that is probably fine but if we want the SDK to do this internally we'd want to provide an async approach. We even have some cleanup tasks for this both in the SDK and Infrahub: https://github.com/opsmill/infrahub-sdk-python/blob/v1.17.0/pyproject.toml#L253

gmazoyer:
Branch rebased and target changed to `develop`

ogenstad (approved):

dev/specs/2026-01-file-object.md
wvandeun: added top open issues / questions

PR #8046, opened 2026-01-06
solve post-merge cardinality-one relationships bug on branch differently

### LINKED ISSUE #7338: bug: updated cardinality=one relationship on branch after merge has too many related nodes

ajtmccarty:
fixes #7338 

TODO:
- [x] needs a migration to move all `branched_from` times on merged branches back one microsecond

this issue was previously fixed in #7595, but after running into it again while making some unrelated changes, I believe we need to solve it in a different manner

see the bug for an explanation of the issue regarding updated cardinality-one relationships on a branch that has already been merged, but this is the basic issue.
-  after merging a branch, everything added on that branch, including cardinality-one relationships, now has the same `from` time
- and this `from` time is the same as the branch's `branched_from` time, which we use in filtering what data from `main` to include in this branch

the previous solution, which I suggested and approved, was to change the logic in one of our widely used filters from `<=` to just `<`. this works when querying the branch after it is merged because it excludes all of the changes just merged into `main` from the branch.
there are a couple of problems with this solution:
1. the other methods we use for generating branch filter queries still use the `<=` logic instead of the `<` logic. this has not been an issue b/c these methods are used in only a few places
2. this prevents us from making a batch of changes that depend on one another that we want to share the same `at` timestamp, which is how I re-encountered this issue. for example, say I want to make schema updates all with the same timestamp and some of the schema changes depend on other schema changes. I can only accomplish this using our `Branch.get_query_filter_path` method if the filter created by that method uses `<=` logic b/c `<` logic would exclude the changes from other queries with the same `at` timestamp

there is already a test that covers bug #7338 and it still passes with this change. there are many other tests for diffing and merging that also still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential duplicated relationships after branch merges by adjusting timestamp handling.

* **New Features**
  * Enhanced rebase functionality with optional timestamp parameter for more precise control.

* **Chores**
  * Updated graph database version to 51 with accompanying data migration.


ogenstad (approved):

ajtmccarty:
@opsmill/backend  added another commit that removes some dead filtering code and updates our `to` filter to be `>` instead of `>=` so requested another review

gmazoyer (approved):

PR #8047, opened 2026-01-06
'at' param in schema load and all migrations

ajtmccarty:
part of #6948 

- pass `at` parameter all the way through the schema load/update logic so that everything that is part of a single schema update has the same timestamp (so that we can roll it back if necessary)
- pass required `at` parameter to all migrations, schema and database
  - reviewed all schema migrations to make sure that they are using the `at` parameter when making changes
  - did not update database migrations b/c it did not seem worth the effort, but we should make sure that all changes in a given migration have the same timestamp if possible going forward
- tests for all changes getting the same timestamp during schema save and schema migrations

## Summary by CodeRabbit

* **Bug Fixes**
  * Migration and schema operations now consistently record explicit timestamps, improving audit accuracy for create/update/delete actions.

* **Chores**
  * Migration execution unified to accept a structured context object, streamlining propagation of timing and user metadata across flows.

* **Tests**
  * Added and updated tests and helpers to validate timestamp propagation and edge-timestamp invariants during schema/migration operations.

### DIFF
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -88,7 +88,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
 
         try:
             log.info(f"Running migrations for branch '{obj.name}'")
-            await migration_runner.run(db=db)
+            await migration_runner.run(db=db, at=Timestamp())
         except MigrationFailureError as exc:
             log.error(f"Failed to run migrations for branch '{obj.name}': {exc.errors}")
             raise
@@ -170,7 +170,8 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
         migrations = []
         async with lock.registry.global_graph_lock():
             async with db.start_transaction() as dbt:
-                await obj.rebase(db=dbt, user_id=context.account.account_id)
+                rebase_at = Timestamp()
+                await obj.rebase(db=dbt, user_id=context.account.account_id, at=rebase_at)
                 log.info("Branch successfully rebased")
 
             if obj.has_schema_changes:
@@ -199,6 +200,7 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
                         previous_schema=schema_in_main_before,
                         migrations=migrations,
                         user_id=context.account.account_id,
+                        at=rebase_at,
                     )
                 )
                 for error in errors:
@@ -291,7 +293,8 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
                 diff_locker=DiffLocker(),
                 workflow=get_workflow(),
             )
-            branch_diff = await merger.merge()
+            merge_at = Timestamp()
+            branch_diff = await merger.merge(at=merge_at)
             await merger.update_schema()
 
         changelog_collector = DiffChangelogCollector(diff=branch_diff, branch=obj, db=db)
@@ -304,6 +307,7 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
                     previous_schema=merger.initial_source_schema,
                     migrations=merger.migrations,
                     user_id=context.account.account_id,
+                    at=merge_at,
                 )
             )
             for error in errors:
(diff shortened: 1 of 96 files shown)


backend/infrahub/core/branch/tasks.py
```diff
@@ -170,7 +170,8 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
         migrations = []
         async with lock.registry.global_graph_lock():
             async with db.start_transaction() as dbt:
-                await obj.rebase(db=dbt)
+                rebase_at = Timestamp()
+                await obj.rebase(db=dbt, at=rebase_at)
```
ogenstad: Can you check to see if this is enough to fix IFC-1267?

backend/infrahub/core/migrations/graph/m025_uniqueness_nulls.py
```diff
@@ -22,5 +23,5 @@ class Migration025(InternalSchemaMigration):
     async def validate_migration(self, db: InfrahubDatabase) -> MigrationResult:  # noqa: ARG002
         return MigrationResult()
 
-    async def execute(self, db: InfrahubDatabase) -> MigrationResult:
+    async def execute(self, db: InfrahubDatabase, at: Timestamp) -> MigrationResult:  # noqa: ARG002
```
ogenstad: I'm not sure we have to do something about this now but we have a Liskov substitution violation on these methods. We actually already had one earlier as the base class has a `user_id` parameter and previously we overrode that to remove the parameter and just have the `db` param.

backend/tests/unit/core/schema_manager/test_schema_load_edge_timestamps.py
```diff
@@ -0,0 +1,169 @@
...
+    - A new relationship ("secondary_tag" on Criticality)
+    - An updated node property (Criticality label changed)
+    - An attribute with an updated property ("description" label changed)
+    - A relationship with an updated property ("tags" label changed)
+    """
+    current_schema = registry.schema.get_schema_branch(name=branch.name)
+    new_schema = current_schema.duplicate()
+
+    # Get a duplicate of the Criticality node to modify
+    node = new_schema.get(name="BuiltinCriticality", duplicate=False)
```
ogenstad: Can you check if the use of BuiltinCriticality here will cause later problems due to #8032?

ogenstad (approved):
I wonder if we should introduce some migration input datatype.
I.e. instead of

```python
execution_result = await migration.execute(db=db, branch=branch, at=Timestamp())
```

Something like:
```python
execution_result = await migration.execute(migration_input=MigrationInput(db=db, branch=branch))
```

Then any undefined timestamp would have a default_factory of `Timestamp`. I don't know if it might help future changes where and also the liskov issue.

backend/infrahub/core/migrations/graph/m025_uniqueness_nulls.py
ajtmccarty: I don't think this is actually a Liskov Substitution violation because I think the migrations all should include an `at` parameter that tells them when to set their from/to timestamps (even if they're not going to be used in the legacy migrations). But I _do_ like getting called out for not being SOLID enough.

backend/tests/unit/core/schema_manager/test_schema_load_edge_timestamps.py
ajtmccarty: good point. updated to `TestingCriticality`

backend/infrahub/core/migrations/graph/m025_uniqueness_nulls.py
ajtmccarty: I also added your suggested `MigrationInput` class, which will make adding `user_id` to all of these migrations much easier during conflict resolution, so thanks for the advice

backend/infrahub/core/branch/tasks.py
ajtmccarty: I'm going to add this as a TODO on the next PR that makes bigger changes to how rebase is called
#8015

backend/infrahub/core/migrations/graph/m025_uniqueness_nulls.py
ogenstad: Sorry that might not have been clear. What I meant was that going into it even before this PR it looked like this:

```python

class InternalSchemaMigration(BaseModel):

    async def execute(self, db: InfrahubDatabase, user_id: str = SYSTEM_USER_ID) -> MigrationResult:


class Migration025(InternalSchemaMigration):

    async def execute(self, db: InfrahubDatabase) -> MigrationResult: <-- Removes the user_id param from the base class
```

So based on the signature on `InternalSchemaMigration.execute()` we have a violation on `Migration025.execute()`

PR #8060, opened 2026-01-08
fix(frontend/backend): make search anywhere bar scale

### LINKED ISSUE #8059: task: make search anywhere scale
### Component

Frontend UI

### Task Description

When having a lot of objects in Infrahub, the search anywhere bar can be slow to respond.

Especially when the bar finds objects with many cardinality relationships with thousands of peers, the underlying graphql query will try to fetch all peers although it's not required at all.

Room for improvement is also present within the backend to make the search anywhere query scale better.

fatih-acar:
When having cardinality many relationships with thousands of peer, the underlying query can take seconds to complete. Avoid this and allow better responsiveness to the search anywhere bar by not fetching cardinality many relationships at all.

This PR also contains a backend change to introduce a new query for the search anywhere bar, leveraging a new index on the AttributeValue nodes.
Edit: testing shows there's no noticeable performance impact in adding this additional index...

~Important note: there's a small functional regression due to the backend changes, matching is now case sensitive with quirks to do a capitalized match, lowered match and uppered match. e.g. trying to find "SeRver" with "server" as a query won't work anymore. but finding "Server" with "server" will work
Another functional regression is that only indexed values will be searched: non-indexed types such as JSON, TextArea, List won't be found using search anywhere...~

The PR now adds a "case-sensitive" toggle for the search anywhere query, while keeping the old behavior the default one.

Fixes #8059

## Summary by CodeRabbit

* **New Features**
  * Optional case-sensitive toggle in "Search anywhere" for precise matching.
  * Added TEXT-backed attribute-value search enabling partial (contains/starts/ends) and exact matches.

* **Bug Fixes & Improvements**
  * Search scales better with large object counts.
  * Many-to-many relationships hidden in search results for clearer output.
  * UUID lookups and search edges return consistent id/kind payloads.

* **Tests**
  * Comprehensive tests for case sensitivity, branch-aware behavior, filters, and deletion semantics.


backend/infrahub/graphql/queries/search.py
```diff
@@ -105,9 +105,25 @@ async def search_resolver(
     limit: int = 10,
     partial_match: bool = True,
 ) -> dict[str, Any]:
+    """Search for nodes by attribute value.
+
+    This resolver uses the NodeGetListByAttributeValueQuery which is optimized for search
+    operations by starting from AttributeValueIndexed nodes and using a TEXT index for
+    efficient CONTAINS searches.
```
ajtmccarty: I don't think the docstring on this class should include implementation details of the query that it uses. Claude and company have a tendency to get too into comments and I think this takes it too far. the docstring on the query can, and does, describe this optimization, but it doesn't belong here

backend/infrahub/graphql/queries/search.py
```diff
@@ -116,25 +132,31 @@ async def search_resolver(
...
-                db=graphql_context.db,
-                branch=graphql_context.branch,
-                schema=kind,
-                filters={"any__value": q},
-                limit=limit,
-                partial_match=partial_match,
-            )
-            results.extend(objs)
+        # Use the optimized query that starts from AttributeValueIndexed nodes
+        # and leverages the TEXT index for efficient CONTAINS searches
```
ajtmccarty: samesies

ajtmccarty (commented):
a few component tests for the new query would be good

fatih-acar:
@ajtmccarty thanks for the review although I clicked on the wrong button yesterday and the PR went out of draft... :sweat_smile: 

I added tests that uncovered a missing case (I forgot to match for the exact search string...), I think it's ok now!

@wvandeun are you ok with the functional regressions we introduce on purpose? (see PR desc)

fatih-acar:
@opsmill/frontend I need a review on the (rather small) changes please

pa-lem (approved):

bilalabbad (approved):

PR #8061, opened 2026-01-08
Add mandatory 'db' attribute to Query objects

ogenstad:
Add mandatory 'db' param to the Query class prior to migrating other queries away from .init and .query_init methods.

## Summary by CodeRabbit

* **Refactor**
  * Restructured database parameter handling across core query classes to ensure consistent database context availability throughout the query lifecycle. Updated query constructors to require explicit database passing during initialization and propagated these changes across query subclasses. Modified the execute method to support optional database parameters with intelligent fallback behavior.

### DIFF
--- a/backend/infrahub/core/account.py
+++ b/backend/infrahub/core/account.py
@@ -56,11 +56,12 @@ class AccountGlobalPermissionQuery(Query):
 
     def __init__(
         self,
+        db: InfrahubDatabase,
         account_id: str,
         branch: Branch | None = None,
         branch_agnostic: bool = False,
     ) -> None:
-        super().__init__(branch=branch, branch_agnostic=branch_agnostic)
+        super().__init__(db=db, branch=branch, branch_agnostic=branch_agnostic)
         self.account_id = account_id
 
         self.params["account_id"] = self.account_id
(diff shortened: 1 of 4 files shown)


ajtmccarty (approved):
I think we should actually remove the `db` object from the `Query` object. Queries should not be executing other queries against the database. A `Query` class should just accept the data that it needs to construct a database query and then `execute` can accept the `db` argument (or really we should just use `InfrahubDatabase.execute(query: Query, ...)` but that's a different issue
I'm approving because addressing the typing issues seems wise, but I want to make it clear that this pattern is not SOLID

ogenstad:
> I think we should actually remove the `db` object from the `Query` object. Queries should not be executing other queries against the database. A `Query` class should just accept the data that it needs to construct a database query and then `execute` can accept the `db` argument (or really we should just use `InfrahubDatabase.execute(query: Query, ...)` but that's a different issue I'm approving because addressing the typing issues seems wise, but I want to make it clear that this pattern is not SOLID

I'm not sure that this PR solves any typing issues. It came from the discussion in Slack where Damien mainly wanted to have it for the schema lookup aspect. For my part I'd be happy to just have the current queries that use the db object to lookup the schema in an async way to have a database object as a required parameter as I agree that it looks a bit weird with this transition. What I mainly care about is getting rid of the .init and .query_init methods of the queries. I'd also at some point want to separate the schema objects from the database. Any additional thoughts from you @dgarros?

PR #8069, opened 2026-01-09
IFC-17662: Add permission for thread comment role on initialization for infrahub users

solababs:
## Summary by CodeRabbit

* **New Features**
  * Enhanced default user permissions to support thread commenting and change-thread management.
  * The new thread-comment permission is included by default for Infrahub Users.

* **Bug Fixes / Chores**
  * Minor cleanup and logging related to role creation during account setup.


backend/infrahub/core/initialization.py
```diff
@@ -455,6 +455,53 @@ async def create_proposed_change_reviewer_role(db: InfrahubDatabase) -> CoreAcco
...
+    thread_comment_create_permission = await Node.init(db=db, schema=InfrahubKind.OBJECTPERMISSION)
+    await thread_comment_create_permission.new(
+        db=db,
+        name="ThreadComment",
+        namespace="Core",
+        action=PermissionAction.CREATE.value,
+        decision=PermissionDecision.ALLOW_ALL.value,
+        description="Allow a user to create thread comments reply",
+    )
+    await thread_comment_create_permission.save(db=db)
```
ajtmccarty, line 469: do users need to be able to update a ThreadComment as well? I'm not really sure how the comments work

ajtmccarty (approved):

backend/infrahub/core/initialization.py
ogenstad: I'm not sure we should have individual permissions in this sense. I think the problem here is that these node kinds are branch agnostic. With the current permission system it means that if a user is restricted from making changes to the default branch they can only make changes to a node within another branch. This approach makes sense for nodes with branch=aware. However for nodes with branch=agnostic we currently have the weird situation where users can create a branch agnostic comment node if their current context is for a non default branch. However if the same user which is restricted from creating or updating nodes in the default branch tries to create a proposed change comment and the context is the default branch this will currently fail.

What I'd propose is that the "default branch permission" doesn't apply to branch agnostic nodes and that if a user has permission to modify these nodes that would apply to any branch.

ajtmccarty: makes sense to me

ajtmccarty:
this will be fixed in a different way. we need to update how we handle permissions for branch-agnostic schemas being updated on the default branch

PR #8070, opened 2026-01-09
fix(tasks): enterprise helm chart bump was broken

fatih-acar:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed release process to properly handle and update multiple Helm charts in a single run, instead of stopping after the first chart update.

### DIFF
--- a/tasks/release.py
+++ b/tasks/release.py
@@ -120,7 +120,7 @@ def update_helm_chart(context: Context, chart_repo: str | None = "helm/") -> Non
             print(
                 f"{str(chart_path)} updates not required, `appVersion` of {old_app_version} matches current from `pyproject.toml`"
             )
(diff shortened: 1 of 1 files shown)


tasks/release.py
```diff
@@ -145,11 +145,11 @@ def update_helm_chart(context: Context, chart_repo: str | None = "helm/") -> Non
             print(f"Warning: Unable to strictly compare versions, using default Helm chart version: {new_helm_version}")
 
         # Convert Version to str before passing to yaml
-        app_version = str(app_version)
```
fatih-acar, line 148: Issue was here, previous method worked with strings but we now need to keep the Version object for comparison

ogenstad (approved):

PR #8072, opened 2026-01-09
Remove heroicons library and use lucide

pa-lem:
## Summary by CodeRabbit

* **Chores**
  * Replaced the app's icon library across UI components and removed an unused icon package from the project manifest.  
  * All icons and visual behavior remain unchanged; no UI features or interactions were added or removed.


dgarros:
how will this change impact the name of the icon referenced in the user schema ? should we expect the name to change or not ?

pa-lem:
> how will this change impact the name of the icon referenced in the user schema ? should we expect the name to change or not ?

No, it's only the internal icons that are updated, we are still using Iconify in other places within the app + for the icons from the schema, it's mostly a cleanup without any breaking change

PR #8086, opened 2026-01-12
Rename unit tests to component

ogenstad:
Renames the current backend unit tests to "component" and start a fresh section for "unit" which should contain tests that doesn't require any external services across the network. I.e. a database or web service. Currently I've moved one of the test files from the component section into the new unit section. This is test_config.py.


dgarros (approved):
This is a great change
Don't forget to also update the doc in `dev/knowledge` to reflect the new structure of the doc

PR #8091, opened 2026-01-12
IFC-2154: Import read only repository last commit

solababs:
## Summary by CodeRabbit

* **New Features**
  * Added a GraphQL mutation and background workflow to import/synchronize the latest commit for read-only repositories so users can refresh repository references with recent remote changes.
* **Bug Fixes**
  * Removed unintended duplicate definitions that could trigger redundant workflow or mutation submissions.
* **Tests**
  * Added component and GraphQL tests covering the new import-last-commit flow and mutation.


backend/infrahub/git/tasks.py
```diff
@@ -587,6 +587,26 @@ async def import_objects_from_git_repository(model: GitRepositoryImportObjects)
...
+
+    client = get_client()
+
+    repo = await get_initialized_repo(
+        client=client,
+        repository_id=model.repository_id,
+        name=model.repository_name,
+        repository_kind=model.repository_kind,
+    )
+    await cast("InfrahubReadOnlyRepository", repo).sync_from_remote()
```
ogenstad: Since we know that we'll only support this on a read-only repo I'd suggest using `InfrahubReadOnlyRepository.init` instead and avoid the cast.

backend/infrahub/graphql/mutations/repository.py
```diff
@@ -202,6 +203,46 @@ async def mutate(
...
+        root: dict,  # noqa: ARG003
+        info: GraphQLResolveInfo,
+        data: IdentifierInput,
+    ) -> dict[str, bool]:
+        graphql_context: GraphqlContext = info.context
+        branch = graphql_context.branch
+        repository_id = str(data.id)
+        repo: CoreReadOnlyRepository = await NodeManager.get_one_by_id_or_default_filter(
+            db=graphql_context.db,
+            kind=InfrahubKind.READONLYREPOSITORY,
```
ogenstad: If we instead move CoreReadOnlyRepository out from the IF_TYPECHECKING part above and send it directly to `kind` instead of the string InfrahubKind.READONLYREPOSITORY it would look like we returned a `CoreReadOnlyRepository` automatically.

ogenstad (commented):
Looks good to me overall. I added some comments about typing.

I think CodeRabbit raises a valid concern and I see that we aren't actually implementing a lock in the other places where we call `sync_from_remote()` which we should. I.e. like we do here:

```python
@flow(
    name="git-repository-add-read-write",
    flow_run_name="Adding repository {model.repository_name} in branch {model.infrahub_branch_name}",
)
async def add_git_repository(model: GitRepositoryAdd) -> None:
    await add_tags(branches=[model.infrahub_branch_name], nodes=[model.repository_id])

    async with lock.registry.get(name=model.repository_name, namespace="repository"):
        repo = await InfrahubRepository.new(
            id=model.repository_id,
            name=model.repository_name,
            location=model.location,
            client=get_client(),
            infrahub_branch_name=model.infrahub_branch_name,
            internal_status=model.internal_status,
            default_branch_name=model.default_branch_name,
        )
        await repo.import_objects_from_files(  # type: ignore[call-overload]
            infrahub_branch_name=model.infrahub_branch_name, git_branch_name=model.default_branch_name
        )
```

It would also be nice with some tests for the actual mutation.

ogenstad (approved):
LGTM.

ogenstad (approved):

PR #8102, opened 2026-01-13
fix(backend): add graphql parse/validate caching

fatih-acar:
## Summary by CodeRabbit

* **Refactor**
  * Optimized GraphQL query execution infrastructure with caching mechanisms for parsed documents and schema validation, reducing redundant processing across execution paths. Enhanced async execution handling and standardized query parsing workflows across the application backend.

### DIFF
--- a/backend/infrahub/graphql/execution.py
+++ b/backend/infrahub/graphql/execution.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+from functools import lru_cache
+from inspect import isawaitable
+from typing import TYPE_CHECKING, Any, Callable
+
+from graphql import GraphQLSchema, execute, parse, validate
+from graphql.error import GraphQLError
+from graphql.execution import ExecutionResult
+from graphql.type import validate_schema
+
+if TYPE_CHECKING:
+    from graphql import ExecutionContext, GraphQLFieldResolver, GraphQLTypeResolver
+    from graphql.execution import Middleware
+    from graphql.language import Source
+    from graphql.language.ast import DocumentNode
+
+
+@lru_cache(maxsize=1024)
+def _cached_parse(query: str) -> DocumentNode:
+    """Internal cached parse function for queries without @expand directive."""
+    return parse(query)
+
+
+def cached_parse(query: str | Source) -> DocumentNode:
+    """Parse a GraphQL query string into a DocumentNode.
+
+    Queries containing the @expand directive are not cached because the parser
+    mutates the AST to add expanded fields, which would corrupt the cache.
+    """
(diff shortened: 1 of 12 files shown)


backend/infrahub/graphql/parser.py
```diff
@@ -117,7 +114,9 @@ def _process_expand_directive(self, path: str, directive: DirectiveNode) -> None
                     FieldNode(
                         kind="field",
                         name=NameNode(kind="name", value=attribute.name),
-                        selection_set=SelectionSetNode(selections=tuple(enrichers)),
+                        selection_set=SelectionSetNode(selections=tuple(e.node for e in enrichers)),
```
ogenstad, line 117: Do you remember why this was required?

backend/infrahub/graphql/utils.py
```diff
@@ -34,6 +73,57 @@ def extract_data(query_name: str, result: ExecutionResult) -> dict:
     return result.data or {}
 
 
+async def graphql_impl(
```
ogenstad: I'm not sure I like the name of this function or its placement. I'd rather have an execution.py within this folder and have that file contain both this function and the new cache_parse() function. Not sure what name I'd rather have instead of `graphql_impl` perhaps "execute_graphql` or `execute_graphql_query`.

ogenstad (approved):
Good find Fatih! Would ge nice with a news fragment as well.

backend/infrahub/graphql/parser.py
fatih-acar: Claude spotted this, `SelectionSetNode.selections` is supposed to be a tuple of `FieldNode`, I think it's right? (It was not caught by tests, I guess we're missing coverage there)

backend/infrahub/graphql/utils.py
fatih-acar: I was not sure either... was expecting some feedback on this, thank you will do the changes!

backend/infrahub/graphql/execution.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@lru_cache(maxsize=1024)
+def cached_validate(schema: GraphQLSchema, document_ast: DocumentNode) -> list[GraphQLError]:
+    return validate(schema, document_ast)
+
+
+@lru_cache(maxsize=1024)
+def cached_validate_schema(schema: GraphQLSchema) -> list[GraphQLError]:
+    return validate_schema(schema)
```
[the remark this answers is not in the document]
fatih-acar: Ignoring this as this is going to use the cached the instance of `GraphQLSchema` (using our own schema cache).

PR #8113, opened 2026-01-14
fix(docs): add size flavored enterprise charts install

fatih-acar:
## Summary by CodeRabbit

* **Documentation**
  * Installation rewired to favor flavored-chart installs from the registry with OCI URLs and explicit version pinning (production recommended).
  * Added sizing presets table (flavors with example chart versions and required memory) and links to review each preset.
  * Added migration steps for pre-4.0.0 charts, including cleanup of old preset values and Redis host configuration.
  * Preserved and reorganized base-chart and local-chart install guidance as an alternative.

### DIFF
--- a/docs/docs/guides/installation.mdx
+++ b/docs/docs/guides/installation.mdx
@@ -522,27 +522,59 @@ infrahub:
 Be sure to replace the placeholder values with your actual values.
 :::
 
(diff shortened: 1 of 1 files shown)


dgarros (approved):
I think it would be good to add the instructions regarding the migration from the current chart

fatih-acar:
> I think it would be good to add the instructions regarding the migration from the current chart

Done

PR #8121, opened 2026-01-14
link diff and pc during pc pipeline

ajtmccarty:
IFC-2194

link diff and proposed change during proposed change pipeline
once a proposed change is linked to a diff, they cannot be unlinked

## Summary by CodeRabbit

* **New Features**
  * Diffs now track and preserve their association with proposed changes during updates and recalculations, improving change management visibility.

* **Tests**
  * Added comprehensive tests verifying proposed change associations are maintained during diff operations and conflict resolution workflows.

### DIFF
--- a/backend/infrahub/core/diff/coordinator.py
+++ b/backend/infrahub/core/diff/coordinator.py
@@ -116,7 +116,9 @@ class DiffCoordinator:
             name=name,
         )
 
-    async def update_branch_diff(self, base_branch: Branch, diff_branch: Branch) -> EnrichedDiffRootMetadata:
+    async def update_branch_diff(
+        self, base_branch: Branch, diff_branch: Branch, proposed_change_id: str | None = None
+    ) -> EnrichedDiffRootMetadata:
         tracking_id = BranchTrackingId(name=diff_branch.name)
         log.info(f"Received request to update branch diff for {base_branch.name} - {diff_branch.name}")
         existing_incremental_lock = self.diff_locker.get_existing_lock(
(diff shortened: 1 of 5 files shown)


backend/infrahub/core/diff/coordinator.py
```diff
@@ -154,6 +156,12 @@ async def update_branch_diff(self, base_branch: Branch, diff_branch: Branch) ->
                 tracking_id=tracking_id,
                 force_branch_refresh=False,
             )
+
+            # set proposed change IDs
+            if proposed_change_id:
+                enriched_diffs.diff_branch_diff.proposed_change_id = proposed_change_id
+                enriched_diffs.base_branch_diff.proposed_change_id = proposed_change_id
```
ogenstad, line 163: I think this looks a bit weird that we need to set this both for the source and destination branch. It gives the feeling that these could be different. Not sure if it has a practical difference though. Potentially it would be a bit cleaner with some setter method on `EnrichedDiffs`. Not a big issue just an observation.

ogenstad (approved):

backend/infrahub/core/diff/coordinator.py
ajtmccarty: I agree that this is weird and I also want to change it, but I think it's probably a bigger PR outside the scope of this one.

Right now a diff for a given branch is really two diffs: one captures changes on the branch between `branched_from` and now, the other captures changes on the default branch between `branched_from` and now. At the database level, these two entities both have identical metadata properties (from_time, to_time, branch_name, etc.) and an edge linking one another. I think the more correct implementation would be to have a higher-level vertex, `DiffPair` or something, that links the two diffs and stores metadata that they share. It wouldn't be a huge change, but it's not trivial and too much for this PR

PR #8124, opened 2026-01-15
only allow 1 open PC per branch at a time

### LINKED ISSUE #8123: bug: should not allow multiple open proposed changes for the same branch
### Component

API Server / GraphQL

### Infrahub version

1.7.1

### Current Behavior

you can have multiple open proposed changes for the same branch
this does not really make sense b/c once the first proposed change is merged the second proposed change becomes worthless

### Expected Behavior

you should only be able to have 1 proposed change open per branch at a given time

### Steps to Reproduce

open multiple proposed changes on the same branch at the same time

### Additional Information

_No response_

ajtmccarty:
fixes #8123 
 only allow 1 open proposed change on a given branch at a time

## Release Notes

* **Bug Fixes**
  * Prevented multiple open proposed changes on the same source branch—attempting to create a second will now raise a validation error.

### DIFF
--- a/backend/infrahub/graphql/mutations/proposed_change.py
+++ b/backend/infrahub/graphql/mutations/proposed_change.py
@@ -84,7 +84,22 @@ class InfrahubProposedChangeMutation(InfrahubMutationMixin, Mutation):
         if state and state != ProposedChangeState.OPEN.value:
             raise ValidationError(input_value="A proposed change has to be in the open state during creation")
 
+        source_branch_name = data.get("source_branch", {}).get("value")
+
         async with graphql_context.db.start_transaction() as dbt:
+            existing_open_pcs = await NodeManager.query(
+                db=dbt,
+                schema=InfrahubKind.PROPOSEDCHANGE,
+                filters={
+                    "source_branch__value": source_branch_name,
+                    "state__value": ProposedChangeState.OPEN.value,
+                },
+            )
+            if existing_open_pcs:
+                raise ValidationError(
+                    input_value=f"An open proposed change already exists for branch '{source_branch_name}'"
+                )
+
             proposed_change, result = await super().mutate_create(
                 info=info, data=data, branch=branch, database=dbt, override_data=override_data
             )
--- a/backend/tests/integration/diff/test_diff_update.py
+++ b/backend/tests/integration/diff/test_diff_update.py
@@ -1125,44 +1125,6 @@ class TestDiffUpdateConflict(TestInfrahubApp):
             data_check = data_checks_by_conflict_id[tracked_conflict.conflict_id]
             assert data_check.keep_branch.value.value == tracked_conflict.keep_branch.value
 
-    async def test_create_another_proposed_change_data_checks_created(
-        self, db: InfrahubDatabase, initial_dataset, default_branch, client: InfrahubClient
-    ) -> None:
-        # verify duplicate data checks can be created
-        result = await client.execute_graphql(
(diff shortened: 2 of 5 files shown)


backend/tests/integration/diff/test_diff_update.py
```diff
@@ -1125,44 +1125,6 @@ async def test_expected_core_data_checks(
             data_check = data_checks_by_conflict_id[tracked_conflict.conflict_id]
             assert data_check.keep_branch.value.value == tracked_conflict.keep_branch.value
 
-    async def test_create_another_proposed_change_data_checks_created(
```
ajtmccarty, line 1128: test is no longer valid if we can't create multiple PCs for a single branch

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -84,7 +84,22 @@ async def mutate_create(
...
 
+        source_branch_name = data.get("source_branch", {}).get("value")
+
         async with graphql_context.db.start_transaction() as dbt:
+            existing_open_pcs = await NodeManager.query(
+                db=dbt,
+                schema=InfrahubKind.PROPOSEDCHANGE,
+                filters={
+                    "source_branch__value": source_branch_name,
+                    "state__value": ProposedChangeState.OPEN.value,
```
ogenstad, line 95: We also have the issue of reopening a proposed change that's in the CLOSED state. We could have a check on the update mutation for that. But I think a simpler solution would be to expact this filter to instead be "state__values" and include both the open and closed state for a proposed change.

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -84,7 +84,22 @@ async def mutate_create(
...
+                },
+            )
+            if existing_open_pcs:
+                raise ValidationError(
+                    input_value=f"An open proposed change already exists for branch '{source_branch_name}'"
```
ogenstad, line 100: If we change the above to be in either open or closed state this could be changed to "An active proposed change.."

ogenstad (approved):
LGTM, just added the comments that we need to also consider the CLOSED state.

backend/infrahub/graphql/mutations/proposed_change.py
ajtmccarty: @ogenstad looks like we actually do not allow re-opening a closed PC
verified this in the sandbox using graphql to try to open a closed PC
`validate_update` is the method used in `InfrahubProposedChangeMutation.mutate_update`
```
    @property
    def is_completed(self) -> bool:
        """Check if the proposed change is in a completed state."""
        return self != ProposedChangeState.OPEN

    def validate_updatable(self) -> None:
        if self.is_completed:
            raise ValidationError(
                input_value=f"A proposed change in the {self.value} state is not allowed to be updated"
            )
```
so I think we only need to worry about OPEN ones

ogenstad: Ok, not sure if that was done as a mistake, the `ProposedChangeState` class also has the `validate_state_transition()` method which is still in use but looks like that is just dead code now as the check happens later in the code and we'd already have raised this error. I'll check around with regards to expectations around this but let's merge this as is and then we can come back to it. Either to revert that previous change so that we can open a closed PC which would also require an update to this filter. Alternatively we just delete the dead code then.

(also need to open a bug regarding the upserts as we can currently circumvent this check by using an upsert instead of an update :/ )

PR #8142, opened 2026-01-16
Planning for implementation of file object feature

gmazoyer:
## Summary by CodeRabbit

* **New Features**
  * REST endpoints for binary file upload/download with metadata, size checks, and proper download headers; immutable per-upload storage with branch- and time-aware resolution

* **Configuration**
  * New max_file_size setting for uploads (default 50 MB)

* **GraphQL**
  * Upload strategies planned (custom Upload scalar and mutations); file metadata exposed read-only in queries

* **Documentation**
  * Architecture, design specs, README and implementation plan added

* **Tests**
  * Scaffolding for upload/download, schema, and permission tests


dev/wip/ifc-2140/00-architecture.md
```diff
@@ -0,0 +1,243 @@
...
+### Storage Layer
+
+- **No changes needed** to `InfrahubObjectStorage`
+- Storage remains a simple key-value store
+- Each upload creates a new entry (no deduplication)
+
+### REST API Must Handle
+
+- [ ] Generate UUID for new uploads
+- [ ] Calculate checksum (MD5)
```
ogenstad: I'd check to see what checksums are available from the storage backend. I.e does S3 still provide MD5 out of the box or is it some SHA variant that would be considered the default?

dev/wip/ifc-2140/00-architecture.md
```diff
@@ -0,0 +1,243 @@
...
+The existing storage system (`backend/infrahub/storage.py`) is a simple key-value store:
+
+```python
+class InfrahubObjectStorage:
+    def store(identifier: str, content: bytes) -> None  # Stores bytes
+    def retrieve(identifier: str) -> str                # Returns decoded string (text only!)
+```
+
+**Characteristics:**
+- Synchronous methods (called from async FastAPI endpoints)
```
ogenstad: The current API routes from FastAPI are acually sync methods as well. But we'll want to revisit this. At least the new ones should be async.

dev/wip/ifc-2140/00-architecture.md
```diff
@@ -0,0 +1,243 @@
...
+2. **FileObject nodes track references** - the `storage_id` attribute points to storage
+3. **Database tracks changes over time** - when `storage_id` changes, database records it
+4. **Time navigation via database** - query at time T returns `storage_id` as of time T
+
+### File Upload
+
+```
+1. User uploads file via REST API
+2. System generates new UUID as storage_id
+3. System calculates checksum (MD5)
```
ogenstad: Same comment as elsewhere with MD5 vs other algos.

dev/wip/ifc-2140/02-config.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Tasks
+
+### Configuration Setting
+
+- [ ] Modify `backend/infrahub/config.py`
+  - [ ] Add `max_file_size` field to `StorageSettings` class:
+    ```python
+    max_file_size: int = Field(
+        default=200,
```
ogenstad: I wonder if this is to high as the default, but not sure. Hopefully we'd not have this large files that needs to be loaded through the API.

dev/wip/ifc-2140/02-config.md
```diff
@@ -0,0 +1,73 @@
...
+
+- [ ] Update configuration documentation if needed (check `docs/` folder)
+
+### Verification
+
+- [ ] Run `uv run invoke backend.test-unit` to run all unit tests including new ones
+- [ ] Run `uv run invoke backend.test-component` to run all component tests
+- [ ] Manually test with environment variable:
+  ```bash
+  INFRAHUB_STORAGE_MAX_FILE_SIZE=50 uv run python -c "from infrahub import config; print(config.SETTINGS.storage.max_file_size)"
```
ogenstad: I don't think this line will work without actually loading the config first.

ogenstad (commented):
LGTM. One small comment though. I'd not rename the file under the specs comment as those files would remain after the implementation. But I do think it makes sense to the the files under the "wip" folder to the epics in Jira.

dev/specs/2026-01-file-object.md
```diff
@@ -463,6 +478,8 @@ Body: {"file": "String"} # binary
 Return: {"identifier": {identifier}, "checksum": "String"}
 ```
 
+We also have to explore if [uploading a file via our GraphQL API](https://graphql.org/learn/file-uploads/) is something sustainable.
```
dgarros: Would be good to determine if we should use GraphQL or REST before executing the plan,

dev/specs/2026-01-file-object.md
```diff
@@ -50,7 +50,7 @@ generics:
 The following attributes will have to be defined:
 
 - `file_name` (read-only, required) : the name of the file, as uploaded by the user
-- `checksum` (read-only, required): checksum (md5 or sha1) calculated on the uploaded file
+- `checksum` (read-only, required): SHA-1 checksum calculated on the uploaded file
```
dgarros, line 52: The checksum is already calculated by the existing object storage 
I think it would be good to add an entry in the `dev/knowledge` first that describe how the current implementation is working.
The idea is that the plan should show the changes that needs to be implemented.

dev/specs/2026-01-file-object.md
```diff
@@ -400,11 +408,12 @@ mutation {
     data: {
       id: "",
       hfid: [""],
-      name: {value: ""},
-      checksum: {value: ""},
-      file_size: {value: 123},
-      file_type: {value: ""},
-      storage_id: {value: ""},
+      # These are here for reference but as they are read-only field, they should not appear in the mutation
```
dgarros: We should remove them if they are not part of the mutation

dev/specs/2026-01-file-object.md
```diff
@@ -434,12 +443,18 @@ mutation {
...
+1. They are properly generated and exposed at the GraphQL layer
+2. They work as expected when we try using them
+
 ### REST API
 
 Infrahub already has the `/api/storage` API endpoints. However these endpoints do not take Infrahub's permission system into account.
 
 We need to add new REST API endpoints that allow you to upload and download `FileObject` objects, that also take the permission system into account.
 
+Endpoints for nodes implementing the `FileObject` generic should be exposed via the REST API like `/api/{node_kind}/object/`. We currently do not have any mechanism capable of doing that. This can create a significant amount of work to implement this. For instance, if we have `NetworkCircuitContract` inheriting from `CoreFileObject` in the schema, we should have endpoints like `/api/NetworkCircuitContract/object/{identifier}` and `/api/NetworkCircuitContract/upload` generated automatically.
```
dgarros: Not sure if we should use REST but if we do, the URL doesn't look correct to me

dev/wip/ifc-2140/00-architecture.md
```diff
@@ -0,0 +1,244 @@
...
+- **No `Upload` scalar defined** - Graphene doesn't have a built-in Upload type
+- **graphene-file-upload library is unmaintained** - No updates in 12+ months
+- Would need to define a custom `Upload` scalar to accept files in mutations
+
+### Options
+
+| Option | Pros | Cons |
+|--------|------|------|
+| **REST API only** | Simple, well-understood, already planned | Two APIs (REST for upload, GraphQL for CRUD) |
+| **GraphQL with custom Upload scalar** | Single API, consistent with mutations | Extra work, custom scalar needed |
```
dgarros: This would be my preferred option unless there is a good reason not to do that

PR #8144, opened 2026-01-16
Remove certificate_validation default from webhooks

### LINKED ISSUE #6308: bug: `Validate Certificates` attribute doesn't work for webhook
### Component

API Server / GraphQL

### Infrahub version

1.2.5

### Current Behavior

Setup a webhook with `Validate Certificates` set to false however I still face a SSL error

```console
task-worker-2      |   File "/source/backend/infrahub/services/adapters/http/httpx.py", line 73, in _request
task-worker-2      |     raise HTTPServerSSLError(message=f"Unable to validate TLS certificate for connection to {url}") from exc
task-worker-2      | infrahub.exceptions.HTTPServerSSLError
```

### Expected Behavior

If I set `Validate Certificates` to false SSL verification should be ignored

### Steps to Reproduce

- Create a webhook that points to an endpoint with unverified certificates

### Additional Information

```diff
diff --git a/backend/infrahub/webhook/models.py b/backend/infrahub/webhook/models.py
index e0704c084..38ec61ccb 100644
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -138,7 +138,9 @@ class Webhook(BaseModel):

     async def send(self, data: dict[str, Any], context: EventContext, service: InfrahubServices) -> Response:
         await self.prepare(data=data, context=context, service=service)
-        return await service.http.post(url=self.url, json=self.get_payload(), headers=self._headers)
+        return await service.http.post(
+            url=self.url, json=self.get_payload(), headers=self._headers, verify=self.validate_certificates
+        )
```

Setting the global `INFRAHUB_HTTP_TLS_INSECURE` flag to True fixed the issue

ogenstad:
This PR fixes a bug related to certificate validation for webhooks, however it also addresses an issue with the attributes for the webhook types.

A problem was that we had a global setting for HTTPS which determines if TLS certificates should be validated or not, at the same time we have an attribute on webhooks called `validate_certificates`. One issue here is that the attribut had a default value meaning that the global setting would always be overridden and never considered. The better approach here would be to make this attribute optional and instead default to not setting a value on the webhook. Moving forward the only reason to set a value would be if users wanted to override the global settings on an individual webhook. Because of this change we also have a migration to update the schema (in the future we'll want to have this managed internally with a schema migration that allows for the removal of default values on attributes)

For the actual bug the issue was that we didn't pass along setting for the webhooks to the HTTP adapter so the .validate_certificates attribute on the actual webhook was ignored. The change in this PR is to resolve this bug and ensure that we pass along the information to the HTTP adapter. Since we are caching TLS context in the HTTP adapter we also need to update how this caching is managed. Basically by just passing along the arguments for TLS validation we could have ended up in a situation where TLS verification was disabled at the global HTTP level but enabled for a specific webhook without these changes to how caching was managed we would end up not validating certificates under such a circumstance.

I've also added some unittests specifically around the config object.

Fixes #6308

## Summary by CodeRabbit

* **New Features**
  * Per-request TLS verification override to force certificate validation regardless of global insecure setting.

* **Bug Fixes**
  * Removed the default for webhook certificate validation so webhooks inherit global HTTP TLS settings unless explicitly set.
  * Migration added to update existing webhook schema defaults.

* **Chores**
  * Graph version bumped and migration registered.

* **Tests**
  * Added unit and component tests for TLS behavior, CA bundle handling, and the migration.


backend/infrahub/services/adapters/http/httpx.py
```diff
@@ -36,11 +36,38 @@ def settings(self) -> config.HTTPSettings:
...
+        return self.settings.get_tls_context(force_verify=True)
+
     def verify_tls(self, verify: bool | None = None) -> bool | ssl.SSLContext:
+        """Determine the TLS verification behavior for a request.
+
+        Args:
+            verify: Override for TLS verification behavior.
+                - None: Use global settings (may skip verification if tls_insecure=True)
+                - False: Explicitly disable certificate validation
+                - True: Force certificate validation, ignoring global tls_insecure setting
```
fatih-acar, line 62: LGTM!

fatih-acar (approved):

backend/infrahub/core/migrations/graph/m052_remove_webhook_validate_certificates_default.py
```diff
@@ -0,0 +1,58 @@
...
+from infrahub.core.query import Query, QueryType
+from infrahub.core.timestamp import Timestamp
+
+from ..shared import GraphMigration
+
+if TYPE_CHECKING:
+    from infrahub.database import InfrahubDatabase
+
+
+class Migration052Query01(Query):
```
ajtmccarty: docstring saying what this does would be nice

backend/infrahub/core/migrations/graph/m052_remove_webhook_validate_certificates_default.py
```diff
@@ -0,0 +1,58 @@
...
+    name = "migration_052_01"
+    type: QueryType = QueryType.WRITE
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
```
ajtmccarty: should accept an optional `at` param that can be used to set `$at` in the query

backend/infrahub/core/migrations/graph/m052_remove_webhook_validate_certificates_default.py
```diff
@@ -0,0 +1,58 @@
...
+            LIMIT 1
+        }
+        WITH av_sub as av2, r3, a3, path3
+        WHERE all(r IN relationships(path3) WHERE r.status = "active")
+        MERGE (new_value: AttributeValue { value: $null_value, is_default: false })
+        CREATE (a3)-[:HAS_VALUE { branch: r3.branch, branch_level: r3.branch_level, status: "active", from: $at }]->(new_value)
+        SET r3.to = $at
+        """ % {"filters": filters}
+        self.add_to_query(query)
+        self.return_labels = ["av2"]
```
ajtmccarty: I rewrote this query with some changes in a more readable format (at least in my opinion)
- it skips the filtering on `at` and just assumes we only care about the state of the database at this exact instant
- adds `AttributeValueIndexed` label where appropriate
- adds a filter to make sure the `SchemaGeneric` is `CoreWebhook` just in case a user created a `SpecialWebhook` or something generic
- makes it idempotent

```
// get the generic schema
MATCH p1 = (sg:SchemaGeneric)-[:HAS_ATTRIBUTE]->(:Attribute {name: "name"})-[:HAS_VALUE]->(:AttributeValueIndexed {value: "Webhook"})
WHERE all(r IN relationships(p1) WHERE r.status = "active" AND r.to IS NULL)

// for safety, also check that the sg is in the namespace "Core"
MATCH p2 = (sg)-[:HAS_ATTRIBUTE]->(:Attribute {name: "namespace"})-[:HAS_VALUE]->(:AttributeValueIndexed {value: "Core"})
WHERE all(r IN relationships(p2) WHERE r.status = "active" AND r.to IS NULL)

// there should only be 1 CoreWebhook schema generic
WITH sg
LIMIT 1

// find the validate_certificates attribute
MATCH p3 = (sg)-[:IS_RELATED]-(:Relationship {name: "schema__node__attributes"})
    -[:IS_RELATED]-(sa:SchemaAttribute)
    -[:HAS_ATTRIBUTE]->(:Attribute {name: "name"})
    -[:HAS_VALUE]->(:AttributeValueIndexed {value: "validate_certificates"})
WHERE all(r IN relationships(p3) WHERE r.status = "active" AND r.to IS NULL)
// there should only be 1 validate_certificates attribute
WITH sa
LIMIT 1

// get the default_value Attribute
MATCH (sa)-[ha:HAS_ATTRIBUTE]->(default_value_attr:Attribute {name: "default_value"})-[hv:HAS_VALUE]->(default_value)
WHERE all(r IN [ha, hv] WHERE r.status = "active" AND r.to IS NULL)
LIMIT 1

// skip if it is already NULL
WITH sa, default_value_attr, hv, default_value
WHERE default_value.value <> $null_value

// close the HAS_VALUE edge for the current default value
SET hv.to = $at

// get the new value
MERGE (new_value:AttributeValue:AttributeValueIndexed {value: $null_value, is_default: true})

// link the new value
CREATE (default_value_attr)-[new_hv:HAS_VALUE]->(new_value)
SET new_hv = properties(hv)
SET new_hv.from = $at, new_hv.to = NULL
```

backend/infrahub/core/migrations/graph/m052_remove_webhook_validate_certificates_default.py
```diff
@@ -0,0 +1,58 @@
...
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        at = Timestamp()
+        filters, params = at.get_query_filter_path()
+
+        self.params.update(params)
+        self.params["null_value"] = NULL_VALUE
+
+        # ruff: noqa: E501
+        query = """
```
ajtmccarty: this query doesn't really take branch into account at all, which is probably fine because the `CoreWebhook` generic should only exist on the default branch and not be editable by a user
not sure if it is worth making this a MigrationRequiringRebase to handle existing user branches. I'd say that users can choose to rebase their branches if they need/want this update

backend/tests/component/core/migrations/graph/test_052.py
ajtmccarty, line 1: could these tests use the existing `CoreWebhook` generic, set its default_value to `True`, then load it with `SchemaManager.load_schem_to_db()`? the validate by retrieving the `CoreWebhook` from the database before and after the migration?

backend/infrahub/core/migrations/graph/m052_remove_webhook_validate_certificates_default.py
ogenstad: Do we have any `.query_init()` methods that doesn't follow this exact method signature?

Looks like we could instead just access `self.at` in this context as per:

```python
class Query:
    name: str = "base-query"
    type: QueryType

    raise_error_if_empty: bool = False
    insert_return: bool = True
    insert_limit: bool = True

    def __init__(
        self,
        branch: Branch | None = None,
        at: Timestamp | str | None = None,
        limit: int | None = None,
        offset: int | None = None,
        order_by: list[str] | None = None,
        branch_agnostic: bool = False,
        user_id: str = SYSTEM_USER_ID,
    ) -> None:
        if branch:
            self.branch = branch

        self.branch_agnostic = branch_agnostic

        if not hasattr(self, "at"):
            self.at = Timestamp(at)

        if not self.at:
            self.at = Timestamp(at)

```

Is this what you meant?

ajtmccarty: yes this is the better way to do it

backend/tests/component/core/migrations/graph/test_052.py
ogenstad: Yes good point. I'll look at that.

backend/infrahub/core/migrations/graph/m055_remove_webhook_validate_certificates_default.py
```diff
@@ -0,0 +1,85 @@
...
+
+// skip if it is already NULL
+WITH sa, default_value_attr, hv, default_value
+WHERE default_value.value <> $null_value
+
+// close the HAS_VALUE edge for the current default value
+SET hv.to = $at
+
+// get the new value
+MERGE (new_value:AttributeValue:AttributeValueIndexed {value: $null_value, is_default: true})
```
ajtmccarty, line 66: another `LIMIT 1` here would be good in case there are multiple `(:AttributeValue:AttributeValueIndexed {value: "NULL", is_default: true})` vertices

backend/tests/component/core/migrations/graph/test_055.py
```diff
@@ -0,0 +1,115 @@
...
+    # Load only the CoreWebhook schema to the database
+    await registry.schema.load_schema_to_db(
+        schema=schema_branch,
+        db=db,
+        branch=default_branch,
+        limit=["CoreWebhook"],
+    )
+
+
+async def test_migration_055_query(
```
ajtmccarty, line 60: I'd say you only need one of these tests, but if you really want both then go ahead

ajtmccarty (approved):

PR #8155, opened 2026-01-19
IFC-2151 Add `CoreFileObject` to internal schema

gmazoyer:
This is the first step of the IFC-2140 feature. This adds a new generic to the internal schema so that user schemas can inherit from it.

This PR also iterates on the implementation plan to prioritise the file upload via GraphQL by passing files directly into object create mutations.

## Summary by CodeRabbit

* **New Features**
  * Added FileObject as a first-class attachment entity with system-managed, read-only metadata (file_name, checksum, file_size, file_type, storage_id) and GraphQL support (single-step create/update with file).

* **Workflow**
  * Single-step upload flow: GraphQL is primary (file parameter); REST remains an optional single-step fallback. Downloads remain REST-only. Responses include id and storage_id.

* **Documentation**
  * API, SDK, and design docs updated to reflect single-step uploads and read-only attributes.

* **Tests**
  * Updated tests and expectations to match the new FileObject schema and constraint counts.


backend/infrahub/core/schema/definitions/core/file_object.py
```diff
@@ -0,0 +1,54 @@
...
+            kind="Text",
+            description="The MIME type of the file",
+            read_only=True,
+            optional=False,
+            allow_override=AllowOverrideType.NONE,
+        ),
+        Attr(
+            name="storage_id",
+            kind="Text",
+            description="The ID of the uploaded file in Infrahub's storage",
```
ogenstad, line 48: Typo "Infrahub's" -> "Infrahubs"

dev/specs/2026-01-file-object.md
```diff
@@ -43,18 +43,21 @@ generics:
...
-- `storage_id` (required): the id of the uploaded file in Infrahub's storage - this is the only FileObject attribute that can be set via mutations to link the node to an uploaded file
+- `file_type` (read-only, required): the type of the file, detected from file content using magic bytes (via `puremagic`)
+- `storage_id` (read-only, required): the id of the uploaded file in Infrahub's storage - set automatically by the system when a file is uploaded via GraphQL mutation
+
+All attributes are read-only because they are system-managed. When creating a FileObject via GraphQL mutation with the `file` parameter, the system automatically stores the file and sets all these attributes.
 
-**Note:** The `storage_id` attribute only needs to remain writable if we keep REST upload as an option. If we use only the GraphQL combined mutation approach (with `file` parameter), `storage_id` could be made read-only since the system would set it automatically.
+**Fallback note:** If we need to fallback to REST API for file uploads, `storage_id` would need to be changed to writable (`read_only: false`) so users can link nodes to previously uploaded files via the two-step workflow.
```
ogenstad: If we provide a REST-API for uploads I think the expectation would be that the user would provide the data payload to also create the file in the same request and then we'd validate this json payload with a schema for the expected type. In that scenario we shouldn't need a two-step workflow.

dev/wip/ifc-2140/00-architecture.md
```diff
@@ -214,40 +214,28 @@ This provides:
...
 
-Both REST and GraphQL upload will be implemented initially:
+**GraphQL is the primary upload method.** Files are attached directly when creating/updating FileObject nodes via the `file` parameter.
 
 | Method | Endpoint | Use Case |
 |--------|----------|----------|
-| REST | `POST /api/file-object/{kind}/upload` | Simple HTTP clients, streaming, progress |
-| GraphQL | `mutation FileObjectUpload(file: Upload!)` | GraphQL-native clients, consistent API |
+| GraphQL | Create/Update mutations with `file` parameter | Primary - single-step file + node creation |
+| REST (optional) | `POST /api/file-object/{kind}/upload` | Fallback if needed; may be removed |
```
ogenstad: I think someone mentioned somewhere that the "/upload` suffix was a bit weird in terms of rest and that it could perhaps be just `POST /api/file-object/{kind}`. Don't have a super strong opinion around that but it might be a bit cleaner without the /upload as it's already a POST request and I don't think we'll want a POST without the file upload.

ogenstad (approved):

PR #8157, opened 2026-01-19
Add permission check to ReadOnlyRepositoryImportLastCommit mutation

ogenstad:
Adds a permission check for using the ReadOnlyRepositoryImportLastCommit. The requirements are that the user needs to have modify permissions on the current branch for CoreReadOnlyRepository.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository update operations now require stricter permission validation before execution.

### DIFF
--- a/backend/infrahub/graphql/mutations/repository.py
+++ b/backend/infrahub/graphql/mutations/repository.py
@@ -7,9 +7,11 @@ import httpx
 from graphene import Boolean, Field, InputObjectType, Mutation, String
 
 from infrahub import config
-from infrahub.core.constants import InfrahubKind, MetadataOptions
+from infrahub.core.account import ObjectPermission
+from infrahub.core.constants import InfrahubKind, MetadataOptions, PermissionAction, PermissionDecision
 from infrahub.core.manager import NodeManager
 from infrahub.core.protocols import CoreReadOnlyRepository
+from infrahub.core.registry import registry
 from infrahub.core.schema import NodeSchema
 from infrahub.git.models import (
     GitReadOnlyRepositoryImportCommit,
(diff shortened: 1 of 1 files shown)


backend/infrahub/graphql/mutations/repository.py
```diff
@@ -234,6 +236,20 @@ async def mutate(
         graphql_context: GraphqlContext = info.context
         branch = graphql_context.branch
         repository_id = str(data.id)
+
+        graphql_context.active_permissions.raise_for_permissions(
```
gmazoyer, line 240: Since you have only one permission to check here, you can use `raise_for_permission` (singular) which expects only one `permission` parameter.

backend/infrahub/graphql/mutations/repository.py
```diff
@@ -234,6 +236,20 @@ async def mutate(
...
+            permissions=[
+                ObjectPermission(
+                    namespace="Core",
+                    name="ReadOnlyRepository",
+                    action=PermissionAction.UPDATE.value,
+                    decision=PermissionDecision.ALLOW_DEFAULT.value
+                    if graphql_context.branch.name == registry.default_branch
+                    else PermissionDecision.ALLOW_OTHER.value,
+                ),
```
gmazoyer, line 249: Not for this PR, I think we could have a function (similar to the one for global permission like below) for getting a permission object in various places:
```python
def define_object_permission_from_branch(schema: MainSchemaTypes, action: PermissionAction, branch_name: str) -> ObjectPermission:
    if branch_name in (GLOBAL_BRANCH_NAME, registry.default_branch):
        decision = PermissionDecision.ALLOW_DEFAULT
    else:
        decision = PermissionDecision.ALLOW_OTHER

    return ObjectPermission(namespace=schema.namespace, name=schema.name, action=action.value, decision=decision.value)
```

gmazoyer (approved):

PR #8161, opened 2026-01-19
Enhance NodeKindUpdateMigration to update pool node references

### LINKED ISSUE #8160: bug: NumberPool node attribute not updated when node kind is renamed
## Component

- [x] API Server / GraphQL

## Infrahub Version

Latest (stable branch)

## Current Behavior

When a node kind is renamed in the schema (using `id` to identify the existing node), NumberPool nodes that are associated with that node via the `node` attribute are **not updated** to reference the new kind name.

The NumberPool stores the node kind as a plain text string in the `node` attribute (`backend/infrahub/core/schema/definitions/core/resource_pool.py:175`). When the node kind is renamed via `NodeKindUpdateMigration`, the migration:

1. Duplicates the node with the new kind name
2. Recreates all relationships
3. **Does NOT update NumberPool nodes** that reference the renamed kind

This causes:
- `registry.schema.get(name=self.node.value)` to fail with `SchemaNotFoundError` when the NumberPool tries to resolve the node
- The `validate_schema_number_pools` task may incorrectly delete the NumberPool or fail
- NumberPool allocation to stop working for the renamed node

## Expected Behavior

When a node is renamed, all NumberPool nodes that have `node` attribute pointing to the old kind name should be updated to reference the new kind name.

## Steps to Reproduce

1. Create a schema with a node that has a NumberPool attribute:
   ```yaml
   version: "1.0"
   nodes:
     - name: Device
       namespace: Infra
       attributes:
         - name: name
           kind: Text
         - name: vlan_id
           kind: NumberPool
           read_only: true
           optional: false
           parameters:
             start_range: 1
             end_range: 4094
   ```
2. Load the schema and note the node's `id` (e.g., from the API response or UI)
3. Verify the NumberPool is created with `node: "InfraDevice"`
4. Rename the node by providing the `id`:
   ```yaml
   version: "1.0"
   nodes:
     - name: NetworkDevice
       namespace: Infra
       id: <the-node-id-from-step-2>
       attributes:
         - name: name
           kind: Text
         - name: vlan_id
           kind: NumberPool
           read_only: true
           optional: false
           parameters:
             start_range: 1
             end_range: 4094
   ```
5. Load the updated schema
6. Query the NumberPool - observe that `node` still references `"InfraDevice"` (the old name)
7. Attempt to create a new NetworkDevice - NumberPool allocation fails

## Technical Details

**How rename detection works:**

The `SchemaBranch.diff()` method (`backend/infrahub/core/schema/schema_branch.py:214-251`) builds kind-to-id maps and detects when the same `id` has different names between old and new schemas.

**Current migration flow:**

1. `NodeKindUpdateMigration` is triggered (`backend/infrahub/core/migrations/__init__.py`)
2. `NodeDuplicateQuery` updates Neo4j nodes with new kind/namespace (`backend/infrahub/core/migrations/query/node_duplicate.py:175`)
3. All data instances and relationships are properly migrated

**What's missing:**

The migration does not query for and update NumberPool nodes where `node.value == old_kind_name`.

**Affected files:**
- `backend/infrahub/core/migrations/schema/node_kind_update.py` - Missing logic to update NumberPool nodes
- `backend/infrahub/core/node/resource_manager/number_pool.py:30` - Uses `registry.schema.get(name=self.node.value)` which fails with old kind name

**Suggested fix location:**

Add logic to `NodeKindUpdateMigration` (or as a post-migration step) to:
1. Query all `CoreNumberPool` nodes where `node.value == previous_kind`
2. Update `node` attribute to the new kind name

minitriga:
Fixes: #8160 

## Release Notes

* **Bug Fixes**
  * Node kind renames now automatically update all related pool nodes, including their references and auto-generated names, ensuring data consistency across the system.

* **Tests**
  * Added test coverage validating that pool nodes are correctly updated when their referenced node kind is renamed.

### DIFF
--- a/backend/infrahub/core/migrations/schema/node_kind_update.py
+++ b/backend/infrahub/core/migrations/schema/node_kind_update.py
@@ -1,10 +1,24 @@
 from __future__ import annotations
 
-from typing import Any, Sequence
+from typing import TYPE_CHECKING, Any, Sequence
+
+from infrahub.core import registry
+from infrahub.core.constants import infrahubkind
 
 from ..query import MigrationQuery
 from ..query.node_duplicate import NodeDuplicateQuery, SchemaNodeInfo
-from ..shared import SchemaMigration
+from ..shared import MigrationInput, MigrationResult, SchemaMigration
+
+if TYPE_CHECKING:
+    from infrahub.core.branch import Branch
+
+
+# Pool kinds and their text attributes that store node kind references
+POOL_NODE_KIND_ATTRIBUTES: list[tuple[str, str]] = [
+    (infrahubkind.NUMBERPOOL, "node"),
+    (infrahubkind.IPADDRESSPOOL, "default_address_type"),
+    (infrahubkind.IPPREFIXPOOL, "default_prefix_type"),
+]
 
 
 class NodeKindUpdateMigrationQuery01(MigrationQuery, NodeDuplicateQuery):
@@ -38,3 +52,40 @@ class NodeKindUpdateMigrationQuery01(MigrationQuery, NodeDuplicateQuery):
 class NodeKindUpdateMigration(SchemaMigration):
     name: str = "node.kind.update"
     queries: Sequence[type[MigrationQuery]] = [NodeKindUpdateMigrationQuery01]  # type: ignore[assignment]
+
+    async def execute_post_queries(
+        self,
+        migration_input: MigrationInput,
+        result: MigrationResult,
+        branch: Branch,
+    ) -> MigrationResult:
+        """Update pool nodes that reference the renamed node kind."""
+        old_kind = self.previous_schema.kind
+        new_kind = self.new_schema.kind
+
+        # Skip if kind hasn't actually changed (e.g., only inherit_from changed)
+        if old_kind == new_kind:
+            return result
+
+        db = migration_input.db
+
+        for pool_kind, attr_name in POOL_NODE_KIND_ATTRIBUTES:
+            # Skip if the pool schema isn't registered (e.g., in test environments)
+            if not registry.schema.has(name=pool_kind, branch=branch.name):
+                continue
+
+            # Query for pool nodes where the attribute value matches the old kind
+            pools = await registry.manager.query(
+                db=db,
+                branch=branch,
+                schema=pool_kind,
+                filters={attr_name: {"value": old_kind}},
+            )
+
+            for pool in pools:
+                attr = pool.get_attribute(name=attr_name)
+                attr.value = new_kind
+                await pool.save(db=db, fields=[attr_name], at=migration_input.at)
+                result.nbr_migrations_executed += 1
+
+        return result
--- a/backend/infrahub/core/migrations/schema/node_kind_update.py
+++ b/backend/infrahub/core/migrations/schema/node_kind_update.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Sequence
 
-from infrahub.core import registry
 from infrahub.core.constants import infrahubkind
+from infrahub.core.manager import NodeManager
 
 from ..query import MigrationQuery
 from ..query.node_duplicate import NodeDuplicateQuery, SchemaNodeInfo
@@ -71,11 +71,11 @@ class NodeKindUpdateMigration(SchemaMigration):
 
         for pool_kind, attr_name in POOL_NODE_KIND_ATTRIBUTES:
             # Skip if the pool schema isn't registered (e.g., in test environments)
-            if not registry.schema.has(name=pool_kind, branch=branch.name):
+            if not db.schema.has(name=pool_kind, branch=branch.name):
                 continue
 
             # Query for pool nodes where the attribute value matches the old kind
-            pools = await registry.manager.query(
+            pools = await NodeManager.query(
                 db=db,
                 branch=branch,
                 schema=pool_kind,
@@ -90,11 +90,16 @@ class NodeKindUpdateMigration(SchemaMigration):
                 attr.value = new_kind
 
                 # Update the pool name if it follows the auto-generated pattern
-                name_attr = pool.get_attribute(name="name")
-                old_prefix = f"{old_kind}."
-                if name_attr.value.startswith(old_prefix):
-                    name_attr.value = f"{new_kind}.{name_attr.value[len(old_prefix) :]}"
-                    fields_to_save.append("name")
+                try:
+                    name_attr = pool.get_attribute(name="name")
+                except ValueError:
+                    pass
+                else:
+                    old_prefix = f"{old_kind}."
+                    new_prefix = f"{new_kind}."
+                    if name_attr.value.startswith(old_prefix):
+                        name_attr.value = name_attr.value.replace(old_prefix, new_prefix, 1)
+                        fields_to_save.append("name")
 
                 await pool.save(db=db, fields=fields_to_save, at=migration_input.at)
                 result.nbr_migrations_executed += 1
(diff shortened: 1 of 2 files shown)


backend/infrahub/core/migrations/schema/node_kind_update.py
```diff
@@ -38,3 +52,51 @@ def get_nbr_migrations_executed(self) -> int:
...
+                # Update the kind reference attribute
+                attr = pool.get_attribute(name=attr_name)
+                attr.value = new_kind
+
+                # Update the pool name if it follows the auto-generated pattern
+                name_attr = pool.get_attribute(name="name")
+                old_prefix = f"{old_kind}."
+                if name_attr.value.startswith(old_prefix):
+                    name_attr.value = f"{new_kind}.{name_attr.value[len(old_prefix) :]}"
+                    fields_to_save.append("name")
```
ajtmccarty, line 97: can do `name_attr.value.replace(old_prefix, new_prefix)`

backend/infrahub/core/migrations/schema/node_kind_update.py
```diff
@@ -38,3 +52,51 @@ def get_nbr_migrations_executed(self) -> int:
...
+
+        # Skip if kind hasn't actually changed (e.g., only inherit_from changed)
+        if old_kind == new_kind:
+            return result
+
+        db = migration_input.db
+
+        for pool_kind, attr_name in POOL_NODE_KIND_ATTRIBUTES:
+            # Skip if the pool schema isn't registered (e.g., in test environments)
+            if not registry.schema.has(name=pool_kind, branch=branch.name):
```
ajtmccarty, line 74: `db.schema.has` seems safer here

backend/infrahub/core/migrations/schema/node_kind_update.py
```diff
@@ -38,3 +52,51 @@ def get_nbr_migrations_executed(self) -> int:
...
+                continue
+
+            # Query for pool nodes where the attribute value matches the old kind
+            pools = await registry.manager.query(
```
ajtmccarty, line 78: can you use `NodeManager.query` here instead of the registry? the `registry` is a crutch that masks some bad design, so we should avoid it wherever possible

backend/infrahub/core/migrations/schema/node_kind_update.py
```diff
@@ -38,3 +52,51 @@ def get_nbr_migrations_executed(self) -> int:
...
+                attr.value = new_kind
+
+                # Update the pool name if it follows the auto-generated pattern
+                name_attr = pool.get_attribute(name="name")
+                old_prefix = f"{old_kind}."
+                if name_attr.value.startswith(old_prefix):
+                    name_attr.value = f"{new_kind}.{name_attr.value[len(old_prefix) :]}"
+                    fields_to_save.append("name")
+
+                await pool.save(db=db, fields=fields_to_save, at=migration_input.at)
```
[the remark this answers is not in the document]
minitriga: Done — the `get_attribute(name="name")` call is now wrapped in try/except ValueError, so pool types that don't have a `name` attribute skip the name-update step cleanly while the kind attribute still gets updated.

backend/tests/component/core/migrations/schema/test_node_kind_update.py
```diff
@@ -327,3 +336,85 @@ async def test_migration_metadata(db: InfrahubDatabase, car_accord_main: Node, b
...
+                parameters=NumberPoolParameters(start_range=1000, end_range=9999),
+            ),
+        ],
+    )
+    schema = SchemaRoot(nodes=[device_schema])
+    await load_schema(db=db, schema=schema)
+
+    # Register the CoreNumberPool type for pool operations
+    registry.node[InfrahubKind.NUMBERPOOL] = CoreNumberPool
+
```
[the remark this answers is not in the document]
minitriga: Done — switched to `monkeypatch.setitem(registry.node, InfrahubKind.NUMBERPOOL, CoreNumberPool)` so the registry mutation is automatically reverted by pytest after the test, preventing cross-test pollution.

backend/infrahub/core/migrations/schema/node_kind_update.py
minitriga: Done — replaced the manual string slicing with `name_attr.value.replace(old_prefix, new_prefix, 1)`, which is simpler and limits the substitution to the first occurrence.

backend/infrahub/core/migrations/schema/node_kind_update.py
minitriga: Done — swapped `registry.manager.query()` to `NodeManager.query()` throughout `execute_post_queries` to avoid going through the global registry singleton.

backend/tests/component/core/migrations/schema/test_node_kind_update.py
```diff
@@ -327,3 +336,85 @@ async def test_migration_metadata(db: InfrahubDatabase, car_accord_main: Node, b
...
+    # 5. Verify pools with new kind name exist
+    pools_with_new_name = await registry.manager.query(
+        db=db,
+        branch=default_branch,
+        schema=InfrahubKind.NUMBERPOOL,
+        filters={"node": {"value": "Test2NetworkDevice"}},
+    )
+    assert len(pools_with_new_name) == 1
+    assert pools_with_new_name[0].id == pool_id
```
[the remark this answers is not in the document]
minitriga: Added `test_migration_updates_ip_address_pool_node_reference` and `test_migration_updates_ip_prefix_pool_node_reference`. Both follow the same pattern as the NumberPool test: register the ipam schema, monkeypatch the pool class into registry.node, create a pool pointing at a specific kind, run the migration that renames that kind, and assert that default_address_type / default_prefix_type (and the auto-generated pool name) are updated.

backend/infrahub/core/migrations/schema/node_kind_update.py
minitriga: Swapped to `db.schema.has()`. `db` is in scope as `migration_input.db`, and `DatabaseSchemaManager.has()` is a proper superset of `registry.schema.has()` — it checks any in-transaction schema overrides first and falls back to the registry. The unused registry import was removed as a side-effect.

backend/tests/component/core/migrations/schema/test_node_kind_update.py
```diff
@@ -329,3 +341,210 @@ async def test_migration_metadata(db: InfrahubDatabase, car_accord_main: Node, b
...
+
+async def test_migration_updates_number_pool_node_reference(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: SchemaBranch,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test that NumberPool.node is updated when the referenced kind is renamed."""
+    # Register the CoreNumberPool type for pool operations (monkeypatched to avoid shared-state pollution)
+    monkeypatch.setitem(registry.node, InfrahubKind.NUMBERPOOL, CoreNumberPool)
```
ajtmccarty, line 354: you should be able to safely set `registry.node[InfrahubKind.NUMBERPOOL] = CoreNumberPool` b/c we do that as part of initialization. you could set the values for `IPADDRESS/PREFIXPOOL` too

backend/tests/component/core/migrations/schema/test_node_kind_update.py
```diff
@@ -329,3 +341,210 @@ async def test_migration_metadata(db: InfrahubDatabase, car_accord_main: Node, b
...
+
+    # 4. Verify the pool's default_address_type was updated
+    updated_pool = await NodeManager.get_one(db=db, branch=default_branch, id=pool_id)
+    assert updated_pool.default_address_type.value == "IpamNewIPAddress"
+
+    # 5. Verify the pool name was updated (it starts with the old kind prefix)
+    assert updated_pool.name.value.startswith("IpamNewIPAddress.")
+
+
+async def test_migration_updates_ip_prefix_pool_node_reference(
```
ajtmccarty, line 495: would be good to have a test that verifies a non-default pool name does not get updated. or update the existing test(s) to verify it

PR #8170, opened 2026-01-20
Add filters to InfrahubBranch query

ogenstad:
Changes in this PR includes adding filters for the InfrahubBranch query. In order to avoid passing a growing number of parameters to a lot of different functions I introduced a dataclass to contain these parameters. In order to keep the existing API for standard node objects I've just added some overrides now. I think the StandardNode class should be rebased at some point so that we can have a cleaner API and something that could work on potential upcoming other types of standard nodes. However I didn't want to touch that part yet until we know more about the needs for other node kinds.

## Summary by CodeRabbit

* **New Features**
  * Added advanced branch filtering capabilities enabling queries by status, creation date range, creator ID, rebase timestamps, and last updated date.

* **Tests**
  * Added comprehensive test coverage for branch filtering by status, timestamp ranges, and combined filter criteria.


backend/infrahub/core/branch/filters.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from datetime import datetime
+
+    from infrahub.core.branch.enums import BranchStatus
+
+
+@dataclass(slots=True)
+class BranchListFilters:
```
gmazoyer, line 13: This is a good idea 👍

gmazoyer (approved):

PR #8181, opened 2026-01-20
fix branch-levels on NodeCreate

### LINKED ISSUE #8158: bug: incorrect branch_level property when creating branch-agnostic attributes or relationships
### Component

API Server / GraphQL

### Infrahub version

1.7.1

### Current Behavior

When an object with branch-agnostic attributes or relationships is created on a user-created branch, the edges of those attributes and relationships will have the wrong `branch_level` property. for example `[:IS_RELATED {branch: "-global-", branch_level: 2}]`. this is only when creating an object.

### Expected Behavior

we expect all edges on the `-global-` branch to have a branch-level of `1`. in some places, we filter on edges by `branch_level=1` ot identify if they are on the default or global branch

### Steps to Reproduce

1. create an instance of an object that includes branch-agnostic attributes or relationships. this can be just a branch-agnostic schema, including ones defined by infrahub, like a proposed change
2. run this query on the database `MATCH p = ()-[e {branch: "-global-", branch_level: 2}]->() RETURN p`
3. this 

### Additional Information

this will require a migration to fix existing edges with the problem

### LINKED ISSUE #8159: bug: incorrect branch_level property when creating an object with branch-local attributes or relationships
### Component

API Server / GraphQL

### Infrahub version

1.7.1

### Current Behavior

when creating an object that includes a branch-local attribute or relationship, the branch_level of the created edges for the attribute or relationship is `0`. this only happens when creating an object, not when updating.

### Expected Behavior

the branch_level of a branch-local attribute or relationship should be the same as the branch on which the object is created, so `1` if created on the default branch, `2` otherwise

### Steps to Reproduce

1. create an object that includes a branch-local attribute or relationship
2. run this database query `MATCH p = ()-[e {branch_level: 0}]->() RETURN p`

### Additional Information

requires a migration to fix edges with the problem

ajtmccarty:
fixes #8158 
fixes #8159 

a couple of bugs in the node create logic could cause illegal edges to be created in the database
- some edges on the global branch could have a `branch_level` of 2, which indicates a user-created branch, instead of a 1
- some edges for branch-local attributes could have a `branch_level` of 0, which does not happen anywhere else in the code and is not expected

this PR fixes the bugs and includes migrations to correct existing edges with the problems

## Release Notes

* **Bug Fixes**
  * Fixed node creation that was incorrectly assigning hierarchy levels to edges on the global branch
  * Fixed node creation for branch-local edges with incorrect hierarchy levels
  * Included database migrations to correct existing affected edges

* **Tests**
  * Added comprehensive validation tests for branch-agnostic edge properties to prevent regression


backend/infrahub/core/migrations/graph/m052_fix_global_branch_level.py
```diff
@@ -0,0 +1,50 @@
...
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        self.params["global_branch_name"] = GLOBAL_BRANCH_NAME
+
+        query = """
+        MATCH ()-[e {branch: $global_branch_name}]->()
+        WHERE e.branch_level <> 1
+        CALL (e) {
+            SET e.branch_level = 1
+        } IN TRANSACTIONS
```
ogenstad, line 28: Curious with this simple query what's the benefit of having this call statement instead of just a set?

ogenstad (approved):

backend/infrahub/core/migrations/graph/m052_fix_global_branch_level.py
ajtmccarty: it's just being extra safe
if this query ends up changing 1 million edges, which is very unlikely, then it could cause an out of memory error on the database. executing the write operations within transactions prevents that and maintains idempotency

ogenstad: Ok, thanks!

PR #8188, opened 2026-01-21
migration for linking open and merged PCs to the correct DiffRoot

ajtmccarty:
IFC-2209

should follow #8190

adds a migration to link open and merged proposed changes to the correct diff, if it can be found

for open PCs, we look for the current unmerged diff tracking this branch and use that one
for a merged PC, we look for a merged diff for this branch that has a `to` time between when the PC entered the `merging` state and when it entered the `merged` state

tested on several user backups and verified that the correct diffs are linked

## Summary by CodeRabbit

* **Chores**
  * Updated database schema version for graph migrations to version 57.
  * Implemented automatic linking of proposed changes to their corresponding diff roots, with specialized handling for both open and merged proposed changes based on branch tracking and merge status.

* **Tests**
  * Added comprehensive test coverage verifying the new database migration correctly links proposed changes to diff roots under various scenarios.


backend/infrahub/core/graph/__init__.py
```diff
@@ -1 +1 @@
-GRAPH_VERSION = 52
+GRAPH_VERSION = 53
```
ogenstad: This is a bit outdated now.

ogenstad (approved):

PR #8196, opened 2026-01-22
IFC-2174 Add support for uploading files via GraphQL

gmazoyer:
This change introduces a new scalar called `Upload` which can be used to upload a file via a GraphQL mutation. Mutations for nodes inheriting from `CoreObjectFile` will have a new parameter called `file` of type `Upload` exposed. This parameter accepts file in `multipart/form-data` which follows the GraphQL Multipart Request specification.

The `Upload` scalar is backed by Starlette's `UploadFile` which takes care of the heavy work of handling the file itself. This allows us to have as less custom code as possible keeping things easy to maintain.

After a file has been uploaded, it will be handed over to the `FileUploadProcessor` which will be in charge of gathering details about the files, which are useful to create the object in the database. It will also call the storage component so that the file can be stored according to the Infrahub configured storage backend.

## Summary by CodeRabbit

* **New Features**
  * GraphQL file uploads: Upload scalar and file argument on create/update/upsert for file-backed types; file metadata returned.

* **New Features / APIs**
  * New file upload processor to validate, checksum, detect MIME type, store files, and support cleanup.

* **Improvements**
  * Streamed, memory-efficient upload/storage handling with idempotent upserts and storage open/delete support.

* **Tests**
  * Extensive unit and component tests covering upload paths, validation, and storage.

* **Chores**
  * Added MIME-detection dependency and docs on reverse-proxy upload size limits.


backend/infrahub/core/file_processor.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+        magic_bytes = await self.file.read(2048)
+        file_type = self._detect_mime_type(content=magic_bytes)
+        checksum = await self._compute_checksum()
+        storage_id = str(UUIDT())
+        file_name = self.file.filename or storage_id
+
+        await self.file.seek(0)
+        registry.storage.store(identifier=storage_id, content=self.file.file)
```
ogenstad: Note here that we're calling a sync store method within an async function. For larger binary files it could perhaps be an issue for local storage but this could also be an S3 bucket which could kill the event loop. So I'm wondering if we need to have an alternate storage that's async. And perhaps phase out the other one.

dgarros: I agree Async would be ideal but I would prefer to do the improvement of the object store separately and not redesign it as part of this epic

backend/infrahub/core/file_processor.py
```diff
@@ -0,0 +1,134 @@
+from __future__ import annotations
```
ogenstad, line 1: It doesn't have to be now but I think we should avoid adding more stuff to the core module. Later can you consider the impact of moving infrahub/storage.py into a infrahub/storage module where the existing code perhaps lives as adapters for the storage.

backend/infrahub/core/file_processor.py
```diff
@@ -0,0 +1,134 @@
...
+
+        Returns:
+            The detected MIME type string.
+        """
+        try:
+            results = puremagic.magic_string(content)
+            if results:
+                return results[0].mime_type
+        except puremagic.PureError:
+            pass
```
ogenstad: Consider the use of contextlib.suppress here: https://github.com/opsmill/infrahub-sdk-python/pull/768

backend/infrahub/graphql/manager.py
```diff
@@ -950,7 +954,10 @@ def generate_graphql_mutation_update(
             meta_attrs: dict[str, Any] = {"schema": schema, "name": name, "description": schema.description}
             main_attrs["Meta"] = type("Meta", (object,), meta_attrs)
 
-            args_attrs = {"data": input_type(required=True), "context": ContextInput(required=False)}
+            args_attrs: dict[str, Any] = {"data": input_type(required=True), "context": ContextInput(required=False)}
+            if schema.is_file_object:
+                args_attrs["file"] = graphene.Argument(Upload, required=False)
```
ogenstad, line 959: One consideration when looking at this which I don't know if it might be covered elsewhere but for upsert mutations we'd want the operation to be idempotent. I don't think we can get around re uploading the file to the server itself. But we probably don't want to update the storage_id when we do so if the new file is an exact replica of the previous one.

backend/infrahub/graphql/mutations/main.py
```diff
@@ -446,3 +457,26 @@ async def build_graphql_response(info: GraphQLResolveInfo, db: InfrahubDatabase,
     if "object" in fields:
         result["object"] = await obj.to_graphql(db=db, fields=fields.get("object", {}))
     return result
+
+
+async def process_file_upload(file: UploadFile | None) -> tuple[dict[str, Any], FileUploadProcessor | None]:
```
ogenstad: We should have a return value for this function. However, I prefer functions that return one thing. I wonder if it would be easier if the mutations just called `FileUploadProcessor` directly if a file was provided and for the data parts we can just call `await processor.process()` directly?

backend/infrahub/graphql/mutations/main.py
```diff
@@ -119,6 +126,10 @@ async def mutate(
                 f"Unexpected class Name: {cls.__name__}, should end with Create, Update, Upsert, or Delete"
             )
 
+        if file_processor:
+            # The mutation succeeded so we can safely store the file permanently
+            await file_processor.store_file()
```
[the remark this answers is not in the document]
ogenstad: @gmazoyer, the above comment is a bit verbose and convoluted but it's a valid point. I.e. if the upload fails we probably want to revert the change in the database.

backend/tests/component/graphql/mutations/test_file_object.py
```diff
@@ -0,0 +1,429 @@
...
+
+        assert not result.errors, f"GraphQL errors: {result.errors}"
+        assert result.data is not None
+        assert result.data["TestingFileContractCreate"]["ok"] is True
+
+        obj_data = result.data["TestingFileContractCreate"]["object"]
+        assert obj_data["file_name"]["value"] == "contract.pdf"
+        assert obj_data["file_size"]["value"] == len(file_content)
+        assert obj_data["description"]["value"] == "Test contract"
+        assert obj_data["checksum"]["value"]
```
ogenstad: We should be able to calculate the checksum and compare against that too here?

backend/tests/component/graphql/mutations/test_file_object.py
```diff
@@ -0,0 +1,429 @@
...
+
+        result = await graphql(
+            schema=gql_params.schema,
+            source=query,
+            context_value=gql_params.context,
+            variable_values={},
+        )
+
+        assert result.errors is not None
+        assert "file" in str(result.errors[0])
```
ogenstad: Could we make this assert a bit more verbose? I'm guessing there's a longer version of the error message to indicate that no file was uploaded?

backend/tests/unit/core/test_file_processor.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+async def test_processor_unnamed_file(dummy_storage: DummyObjectStorage) -> None:
+    """Test that files without filename use storage_id as file_name."""
+    upload_file = create_upload_file(content=b"content", filename=None)
+
+    with (
+        patch("infrahub.core.file_processor.registry") as mock_registry,
+        patch("infrahub.core.file_processor.config") as mock_config,
+    ):
```
ogenstad: Do we need to mock all of the above tests?

backend/tests/adapters/storage.py
```diff
@@ -0,0 +1,16 @@
+from typing import BinaryIO
+
+from infrahub.storage import InfrahubObjectStorage
+
+
+class DummyObjectStorage(InfrahubObjectStorage):
```
ogenstad, line 7: 👍

ogenstad (approved):
Added a few comments but overall it looks good to me.

backend/infrahub/graphql/mutations/main.py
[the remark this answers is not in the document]
gmazoyer: Yes that's why I did not mark this conversation as resolved. I don't have an elegant solution so far. I wonder if we have to implement something at the storage level to have a proper error handling.

backend/infrahub/core/file_processor.py
gmazoyer: I put it here because, no idea where to put it actually. I considered creating a `infrahub.storage` module. But probably for another PR.

backend/infrahub/graphql/manager.py
gmazoyer: That is a good point, I think we will need some slightly more complex handling in the mutation code to make sure we do not duplicate files on upsert.

backend/infrahub/graphql/mutations/main.py
```diff
@@ -294,12 +347,17 @@ async def mutate_upsert(
         branch: Branch,
         node_getter_default_filter: MutationNodeGetterByDefaultFilter,
         database: InfrahubDatabase | None = None,
-    ) -> tuple[Node, Self, bool]:
+        file_processor: FileUploadProcessor | None = None,
+    ) -> tuple[Node, Self, bool, bool]:
```
gmazoyer: I'm not a fan of doing that, but this is the way I found best to avoid being too intrusive in the existing logic.

backend/infrahub/graphql/mutations/main.py
```diff
@@ -370,6 +432,21 @@ async def mutate_upsert(
...
+            if (
+                file_processor
+                and file_stored
+                and InfrahubKind.FILEOBJECT in node.get_schema().inherit_from
+                and node.checksum.value == file_processor.metadata.checksum  # type: ignore
+            ):
+                file_processor.delete_file()
+                file_stored = False
+                for key in ("file_name", "checksum", "file_size", "file_type", "storage_id"):
+                    data.pop(key, None)
```
ogenstad: Check the "hack" above regarding `data._pop`. A problem here is that "pop" can also be the name of an attribute as sometime used for a network location. As such we can run into problems if we call `data.pop`, specifically on such schemas. We need an overall refactoring of this to clean it up. But I think we can run into the same type of error here if we have an attribute called "pop" on a Fileobject node.

backend/infrahub/graphql/mutations/main.py
ogenstad: As we're adding more entries to this and now have two booleans I think it would be cleaner to return a new dataclass object for upserts where we have these four objects as named attributes instead. It's so easy to make errors when we have two booleans that we need to track the correct order of.

PR #8208, opened 2026-01-22
feat: add stale PRs workflow

fatih-acar:
## Summary by CodeRabbit

* **Chores**
  * Added automated stale pull request management: PRs inactive for 14 days are labeled "stale" (with guidance on reactivation); activity removes the label so PRs remain reactivated. The automation runs daily and can be triggered manually. Also added a visible "stale" label for consistent categorization.

### DIFF
--- a/.github/workflows/manage-stale-prs.yml
+++ b/.github/workflows/manage-stale-prs.yml
@@ -1,0 +1,40 @@
+---
+# yamllint disable rule:truthy
+name: "Manage Stale Pull Requests"
+
+on:
+  schedule:
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  stale:
+    if: github.repository == 'opsmill/infrahub'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/manage-stale-prs.yml
```diff
@@ -0,0 +1,40 @@
...
+
+          stale-pr-label: "stale"
+          stale-pr-message: |
+            This PR has been inactive for 2 weeks and has been marked as stale.
+
+            If you're still working on this, please:
+            - Push new commits, or
+            - Leave a comment to indicate this PR is still active
+
+            The stale label will be removed automatically when there's new activity.
```
dgarros, line 37: Is it gonna conflict with the workflow that manages labels ? I believe by default it will remove labels that are not defined in the yaml file but I could be wrong

dgarros (approved):

.github/workflows/manage-stale-prs.yml
fatih-acar: I was wondering the same and I think you're right... I've added this new label to the list, thanks!

PR #8239, opened 2026-01-26
add figma to file object spec

wvandeun:


ogenstad:
Do we know if the Figma link will remain active over time, if not perhaps we should close this PR?

PR #8240, opened 2026-01-26
set PC ID on diff even if locks collide

ajtmccarty:
if 2 different requests to update a diff come in at the same-ish time, the later request waits for the earlier request to complete and then just returns the diff generated by the earlier request.

this won't work correctly in the case where the earlier request does not include a proposed change ID (b/c it came from the branch view of the UI) and the later request _does_ include a proposed change ID (b/c it came from the PC view). in this case, we need the second diff request to link the diff to the correct proposed change, which is what this PR does

also
- cleans up a test that was not resetting data correctly: `test_convert_repositories.py`
- simplifies `DiffDataCheckSynchronizer.synchronize` now that we have a direct link from diff to proposed change

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure proposed-change IDs stay linked to diffs across concurrent updates, lock waits, and subsequent branch mutations.
  * Narrow deletion targets to avoid removing unrelated nodes and reduce unnecessary conflict/check updates; limit redundant writes.

* **Tests**
  * Added tests covering proposed-change linkage during lock contention and across branch mutations.
  * Updated tests to use a time-reset helper for more reliable state setup.

* **Chores**
  * Added a test database reset utility to restore state to a given point in time.

### DIFF
--- a/backend/infrahub/core/diff/combiner.py
+++ b/backend/infrahub/core/diff/combiner.py
@@ -440,7 +440,7 @@ class DiffCombiner:
                     to_time=later.to_time,
                     tracking_id=later.tracking_id,
                     nodes=combined_nodes,
-                    proposed_change_id=later.proposed_change_id,
+                    proposed_change_id=earlier.proposed_change_id,
                 )
             )
         base_branch_diff, diff_branch_diff = combined_diffs
--- a/backend/infrahub/core/diff/combiner.py
+++ b/backend/infrahub/core/diff/combiner.py
@@ -440,7 +440,7 @@ class DiffCombiner:
                     to_time=later.to_time,
                     tracking_id=later.tracking_id,
                     nodes=combined_nodes,
-                    proposed_change_id=earlier.proposed_change_id,
+                    proposed_change_id=later.proposed_change_id or earlier.proposed_change_id,
                 )
             )
         base_branch_diff, diff_branch_diff = combined_diffs
--- a/backend/infrahub/core/diff/coordinator.py
+++ b/backend/infrahub/core/diff/coordinator.py
@@ -212,6 +212,10 @@ class DiffCoordinator:
                 force_branch_refresh=False,
             )
 
+            # should never have a proposed change id for an arbitrary diff
+            enriched_diffs.diff_branch_diff.proposed_change_id = None
+            enriched_diffs.base_branch_diff.proposed_change_id = None
+
             await self.diff_repo.save(
                 enriched_diffs=enriched_diffs, node_identifiers_to_drop=list(node_identifiers_to_drop)
             )
--- a/backend/infrahub/core/diff/coordinator.py
+++ b/backend/infrahub/core/diff/coordinator.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass, field
+from dataclasses import asdict, dataclass, field
 from typing import TYPE_CHECKING, Iterable, Literal, Sequence, overload
 from uuid import uuid4
 
@@ -234,10 +234,12 @@ class DiffCoordinator:
         ):
             log.info(f"Acquired lock to recalculate diff for {base_branch.name} - {diff_branch.name}")
             # Log current diff roots to help debug race conditions
-            current_roots = await self.diff_repo.get_roots_metadata(diff_branch_names=[diff_branch.name])
+            current_roots = await self.diff_repo.get_roots_metadata(
+                diff_branch_names=[diff_branch.name], exclude_merged=False
+            )
             log.error(
                 f"DiffCoordinator.recalculate for diff_id={diff_id}: current diff roots for branch {diff_branch.name}: "
-                f"{[(dr.uuid, dr.tracking_id.serialize() if dr.tracking_id else None) for dr in current_roots]}"
+                f"{[asdict(dr) for dr in current_roots]}"
             )
             current_branch_diff = await self.diff_repo.get_one(diff_branch_name=diff_branch.name, diff_id=diff_id)
             current_base_diff = await self.diff_repo.get_one(
--- a/backend/infrahub/core/diff/coordinator.py
+++ b/backend/infrahub/core/diff/coordinator.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import asdict, dataclass, field
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Iterable, Literal, Sequence, overload
 from uuid import uuid4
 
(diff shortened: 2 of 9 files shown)


backend/infrahub/core/diff/combiner.py
```diff
@@ -440,7 +440,7 @@ async def combine(self, earlier_diffs: EnrichedDiffs, later_diffs: EnrichedDiffs
                     to_time=later.to_time,
                     tracking_id=later.tracking_id,
                     nodes=combined_nodes,
-                    proposed_change_id=later.proposed_change_id,
```
ajtmccarty, line 443: make sure to pass the PC ID forward if the later incremental diff does not have a PC ID

backend/infrahub/core/diff/query/delete_query.py
```diff
@@ -22,7 +22,7 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
         query = """
 MATCH (d_root:DiffRoot)
 %(diff_filter)s
-OPTIONAL MATCH (d_root)-[*]->(diff_thing)
+OPTIONAL MATCH (d_root)-[*]->(diff_thing:DiffRoot|DiffNode|DiffAttribute|DiffRelationship|DiffRelationshipElement|DiffProperty|DiffConflict)
```
ajtmccarty, line 25: `:DiffRoot` objects link to `:CoreProposedChange` objects at least some of the time now, so we need to give an allowed-list of vertices to delete during a diff delete or proposed changes and things linked to them will get mangled

backend/infrahub/core/diff/data_check_synchronizer.py
ajtmccarty, line 1: we used to need to get the proposed change for a diff by querying for `CoreProposedChange`s that matched the requirements, but now we can just use `EnrichedDiffRootMetadata.proposed_change_id` and remove the `for` loop

backend/tests/functional/convert_object_type/test_convert_repositories.py
```diff
@@ -126,6 +128,13 @@ async def local_repo(
         await people.new(db=db, name="people", members=[john])
         await people.save(db=db)
 
+    async def reset_to_time(self, db: InfrahubDatabase, default_branch: Branch, reset_time: Timestamp):
```
ajtmccarty: a more thorough clean up method for resetting the database to a given time

backend/tests/functional/convert_object_type/test_convert_repositories.py
```diff
@@ -187,12 +196,12 @@ async def test_convert_repo_to_read_only(
 
         # We want to test unidirectional relationship coming from validators towards a repository,
         # so we create a proposed change that would create validators attached to this repo.
-        branch_2_name = "branch_2"
-        _ = await create_branch(branch_name=branch_2_name, db=db)
+        branch_name = "branch_2"
```
ajtmccarty, line 188: also, updated the branches used in the various tests to different names b/c they were getting duplicated and causing unexpected behavior

backend/infrahub/core/diff/coordinator.py
```diff
@@ -116,6 +116,21 @@ async def run_update(
...
+        Updates the database link and the in-memory object's proposed_change_id.
+        """
+        if proposed_change_id and diff_root.proposed_change_id != proposed_change_id:
+            diff_uuids = [diff_root.uuid]
+            if diff_root.partner_uuid:
+                diff_uuids.append(diff_root.partner_uuid)
+            await self.diff_repo.link_to_proposed_change(diff_uuids=diff_uuids, proposed_change_id=proposed_change_id)
+        diff_root.proposed_change_id = proposed_change_id
+        return diff_root
+
```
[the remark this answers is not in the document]
ogenstad: @ajtmccarty, was this intentional that we'd reset the PC ID of an existing diff if we didn't provide one when calling this method?

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -50,60 +48,54 @@ async def _get_data_conflicts(self, enriched_diff: EnrichedDiffRoot) -> list[Dat
...
-                schema=CoreProposedChange,
-                filters={"source_branch": enriched_diff.diff_branch_name, "state": ProposedChangeState.OPEN},
-            )
-        except SchemaNotFoundError:
-            # if the CoreProposedChange schema does not exist, then there's nothing to do
-            proposed_changes = []
-        if not proposed_changes:
+        if not enriched_diff.proposed_change_id:
+            return []
+        pc = await NodeManager.get_one(db=self.db, kind=CoreProposedChange, id=enriched_diff.proposed_change_id)
```
ogenstad, line 54: Not that you changed the filter here which previously only checked for PCs in the OPEN state. I don't know if it has an impact on what we are doing here and if the operation would perhaps have been stopped earlier if that was the case. I.e. if the PC had already been merged.

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -50,60 +48,54 @@ async def _get_data_conflicts(self, enriched_diff: EnrichedDiffRoot) -> list[Dat
...
-            if not isinstance(enriched_diff, EnrichedDiffRoot):
-                has_validator = bool(await self.conflict_recorder.get_validator(proposed_change=pc))
-                # if this pc has a validator, then the conflicts for this diff have already been synchronized and we can be done
-                if has_validator:
-                    continue
+        # if the enriched_diff is EnrichedDiffRootMetadata, then it has no new data in it
+        if not isinstance(enriched_diff, EnrichedDiffRoot):
+            has_validator = bool(await self.conflict_recorder.get_validator(proposed_change=pc))
+            # if this pc has a validator, then the conflicts for this diff have already been synchronized and we can be done
+            if has_validator:
```
ogenstad, line 65: I know this is only moved code but the "has_validator" variable looks a bit redundant as get_validator returns a Node or `None`.

backend/infrahub/core/diff/query/delete_query.py
ogenstad: Can we run into a situation where we'd want to add another node type here but miss to update this specific query?

backend/tests/functional/convert_object_type/test_convert_repositories.py
ogenstad: Would this be useful elsewhere?

ogenstad (approved):
LGTM, I added some minor comments that you can disregard if you wish. The main one is related to the `null` value for a PC which also was highlighted by Coderabbit.

backend/infrahub/core/diff/data_check_synchronizer.py
ajtmccarty: this is here to make the logic more readable. `has_validator` is supposed to be a helpful variable name to make this early return easier to understand

backend/infrahub/core/diff/data_check_synchronizer.py
ajtmccarty: we _should_ only end up here for a diff linked to an OPEN proposed change, but I'll add a conditional for that to be safe

backend/infrahub/core/diff/coordinator.py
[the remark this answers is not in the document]
ajtmccarty: good point. updating this method to return early if `proposed_change_id` is None and not return the diff_root

backend/tests/functional/convert_object_type/test_convert_repositories.py
ajtmccarty: it could be. I think this is a good pattern for speeding up certain tests. I've moved this logic to a new `DatabaseResetter` class in tests.helper.db_reset.py

PR #8253, opened 2026-01-28
prevent duplicate CoreNumberPools

### LINKED ISSUE #8222: bug: duplicate CoreNumberPools can be created
### Component

API Server / GraphQL

### Infrahub version

1.7.1

### Current Behavior

it is possible to end up with multiple CoreNumberPools for the same schema and attribute name, this breaks various things as we expect only a single NumberPool to exist for a given NumberPool Attribute

### Expected Behavior

there should only ever be 1 active CoreNumberPool for a given schema kind/attribute combination

### Steps to Reproduce

1. create a branch
2. load a new schema that includes a NumberPool on the default branch
3. load the same schema that includes a NumberPool on the user branch
4. add an instance of the schema on the default branch
5. add an instance of the schema on the user branch
there will now be 2 instances of a CoreNumberPool for the same schema/attribute

### Additional Information

should update other resource pool types if required

ajtmccarty:
fixes #8222

TODO:
 - [x] check on error when loading data for E2E test
 - [x] review tests b/c there are probably some duplicates in there
 - [x] update description to include removal of setting a random `number_pool_id` during schema loading
 - [x] need to support loading multiple CoreNumberPools with the same `node` and `node_attribute` or provide a migration to remove the duplicates (added separate migration, #8384)

prevent duplicate `CoreNumberPool`s from being created when the same schema that includes at least one `NumberPool` attribute is created on multiple branches

- moves all schema-related NumberPool management logic into `SchemaNumberPoolSynchronizer`, which is mainly code moved from the `validate_schema_number_pools` task
- updates NumberPool attribute logic in `SchemaBranch.process` to stop setting a random UUID in the `number_pool_id` of the attribute's parameters when a NumberPool attribute is added or created. the UUID is now only set when the associated `CoreNumberPool` is actually created. new logic still correctly handles inheritance so that an inherited NumberPool attribute will use the same CoreNumberPool as the associated attribute on the associated GenericSchema
- removes `Node.fetch_or_create_number_pool`. this method used to create a `CoreNumberPool` during `Node` creation if the Node's schema included a `NumberPool` attribute and there was no matching `CoreNumberPool` with the ID defined in attribute.parameters.pool_id. this was an unexpected side effect that did not belong inside of the `Node` class
- now if we try to create a `Node` that includes a `NumberPool` attribute for which there is no matching `NumberPool`, we just raise a `pool request was not found` error
- the `NodeAttributeAdd` schema migration is also updated to call a method on `SchemaNumberPoolSynchronizer` instead of `Node.fetch_or_create_number_pool`
- update some tests that depended on `Node.fetch_or_create_number_pool` being called during Node creation to use `SchemaNumberPoolSynchronizer` instead

## Summary by CodeRabbit

* **New Features**
  * Schema-driven number pool synchronization added — pools defined in schemas are now ensured and reconciled automatically.

* **Bug Fixes**
  * Prevented duplicate number pools when the same schema is loaded on multiple branches; a single pool instance is maintained per schema attribute.

* **Tests**
  * Expanded test coverage and fixtures to validate synchronization, allocation, branching behavior, and migrations.


backend/tests/component/core/resource_manager/test_number_pool_query.py
```diff
@@ -46,6 +47,15 @@ async def register_test_schema(default_branch: Branch, register_core_models_sche
     return schema_branch
 
 
+@pytest.fixture
+async def run_number_pool_validation(db: InfrahubDatabase) -> None:
+    snps = SchemaNumberPoolSynchronizer(
+        db=db,
+        schema_manager=registry.schema,
+    )
+    await snps.run()
```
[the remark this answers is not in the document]
ajtmccarty: no. you can run the SchemaNumberPoolSynchronizer without any schemas registered and it should do nothing

backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -79,14 +83,21 @@ async def execute_post_queries(
...
+            schema_manager=registry.schema,
+        )
+        pool_id = await synchronizer.ensure_pool_for_attribute(
+            schema_node=self.new_schema,
+            attribute=self.new_attribute_schema,
             at=at,
         )
 
+        # Fetch the pool object to allocate numbers
+        number_pool = await registry.manager.get_one_by_id_or_default_filter(db=db, id=pool_id, kind=CoreNumberPool)
```
ogenstad: As we have "default_filter" semi deprecated and we know the ID in this case it could make sense to use `await registry.manager.get_one` instead.

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,269 @@
...
+                await self._update_pool_from_schema(schema_number_pool)
+            elif not defined_on_branches:
+                self.log.info(
+                    f"Deleting number pool (id={schema_number_pool.id}) as it is no longer defined in the schema"
+                )
+                await schema_number_pool.delete(db=self.db)
+
+    async def _update_pool_from_schema(self, schema_number_pool: CoreNumberPool) -> None:
+        """Update a pool's range parameters if they differ from the schema."""
+        schema = self.schema_manager.get(name=schema_number_pool.node.value, branch=registry.default_branch)
```
ogenstad: Do we need to raise an error here if the node kind doesn't exist in the default branch?

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,269 @@
...
+            original_pool_id = attribute.parameters.number_pool_id
+            actual_pool_id = await self.ensure_pool_for_attribute(
+                schema_node=schema_node,
+                attribute=attribute,
+            )
+
+            if actual_pool_id != original_pool_id:
+                self.log.info(
+                    f"Updating {schema_node.kind}.{attribute_name} number_pool_id "
+                    f"from {original_pool_id} to {actual_pool_id}"
```
ogenstad: What would happen to reservations in the pool if we switch the pool like this?

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,269 @@
...
+            name=lock_definition.lock_name, namespace=lock_definition.namespace_name, local=False
+        ):
+            if number_pool_id:
+                try:
+                    await registry.manager.get_one_by_id_or_default_filter(
+                        db=self.db, id=str(number_pool_id), kind=CoreNumberPool
+                    )
+                    return number_pool_id
+                except NodeNotFoundError:
+                    pass
```
ogenstad: Instead of having a try/except/pass can we use `contextlib.suppress`?

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,269 @@
...
+                db=self.db,
+                id=number_pool_id,
+                name=f"{pool_node}.{pool_attribute} [{number_pool_id}]",
+                node=pool_node,
+                node_attribute=pool_attribute,
+                start_range=start_range,
+                end_range=end_range,
+                pool_type=NumberPoolType.SCHEMA.value,
+            )
+            await number_pool.save(db=self.db, at=at)
```
ogenstad: Here we are saving the pool and above we have a delete operation as well. A consideration would be if we have a `context` object that we can access anywhere that would allow us to at least optionally pass the user account.

backend/infrahub/pools/tasks.py
```diff
@@ -3,124 +3,25 @@
...
-                    end_range=end_range,
-                    pool_type=NumberPoolType.SCHEMA.value,
-                )
-                await number_pool.save(db=dbs)
+    synchronizer = SchemaNumberPoolSynchronizer(
+        db=service.database,
+        log=log,
+        schema_manager=registry.schema,
+    )
+    await synchronizer.run()
```
ogenstad: Nice to see how this got simplified :) Answering my own question from above here we have the context object so we'd be able to pass it into the creation of the SchemaNumberPoolSynchronizer object.

changelog/8222.fixed.md
```diff
@@ -0,0 +1 @@
+Prevent creating multiple number pools for the same schema if it is separately loaded on multiple branches
```
ogenstad: Perhaps also mention that this could cause a slight delay after importing a schema that it might not be possible to immediately create a node even though the schemas are in sync between all of the workers because we have to have a short wait as the numberpool wouldn't automatically be created on node creation.

ogenstad (approved):
LGTM and good improvement. I added some comments.

backend/infrahub/core/migrations/schema/node_attribute_add.py
ajtmccarty: definitely correct
updated

backend/infrahub/pools/schema_number_pool_synchronizer.py
ajtmccarty: this is only called on line 125 where we have already checked that this schema exists with this attribute on the default branch and this should raise a SchemaNotFound error if the schema is missing for some reason, so I think it is okay. unless I'm missing something here

backend/infrahub/pools/schema_number_pool_synchronizer.py
ajtmccarty: updated

backend/infrahub/pools/schema_number_pool_synchronizer.py
ajtmccarty: updated to pass user_id through to the `save()` and `delete()` calls

changelog/8222.fixed.md
ajtmccarty: updated

backend/infrahub/pools/schema_number_pool_synchronizer.py
ajtmccarty: the current workflow is that a random UUID is loaded into the `number_pool_id` property of the NumberPool attribute's parameters during `SchemaBranch.process`
I've gone back to remove this logic and done quite a bit of refactoring to support it correctly along the way

ajtmccarty:
setting back to draft for now b/c of large refactoring

backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -83,12 +86,16 @@ async def execute_post_queries(
         db = migration_input.db
         at = migration_input.at
 
-        number_pool: CoreNumberPool = await Node.fetch_or_create_number_pool(
+        upserter = SchemaNumberPoolUpserter(
```
ajtmccarty, line 89: new component used here and in the new `SchemaNumberPoolSynchronizer` class

backend/infrahub/core/node/__init__.py
```diff
@@ -410,69 +420,6 @@ async def handle_pool(
                 )
             )
 
-    @staticmethod
-    async def fetch_or_create_number_pool(
```
ajtmccarty, line 414: no longer a part of the `Node` class. handled asynchronously in `SchemaNumberPoolValidator`

backend/infrahub/core/schema/schema_branch.py
ajtmccarty, line 1: changes to stop setting a random UUID for the `number_pool_id` during schema processing

backend/infrahub/pools/models.py
```diff
@@ -3,11 +3,12 @@
 
 @dataclass
 class NumberPoolLockDefinition:
-    pool_id: str
+    schema_kind: str
+    attribute_name: str
```
ajtmccarty, line 7: change the lock name to use the name of the schema and attribute instead of number pool ID b/c we won't know the number pool ID when creating a new number pool going forward

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,191 @@
...
+    from infrahub.core.schema import GenericSchema, NodeSchema
+    from infrahub.core.schema.manager import SchemaManager
+    from infrahub.core.schema.schema_branch import SchemaBranch
+    from infrahub.database import InfrahubDatabase
+    from infrahub.pools.schema_number_pool_upserter import SchemaNumberPoolUpserter
+
+default_log = get_logger()
+
+
+class SchemaNumberPoolSynchronizer:
```
ajtmccarty, line 27: new class to handle the logic that used to be in the `validate_schema_number_pools` task

backend/tests/component/core/constraint_validators/test_attribute_numberpool_constraints.py
```diff
@@ -25,6 +27,10 @@ async def snow_incident_01(db: InfrahubDatabase, default_branch: Branch, registe
     registry.schema.register_schema(schema=schema, branch=default_branch.name)
     registry.node[InfrahubKind.NUMBERPOOL] = CoreNumberPool
 
+    upserter = SchemaNumberPoolUpserter(db=db, schema_manager=registry.schema)
+    snps = SchemaNumberPoolSynchronizer(db=db, schema_manager=registry.schema, upserter=upserter)
+    await snps.run()
```
ajtmccarty, line 32: have to add this block or similar logic to a number of NumberPool attribute tests b/c the CoreNumberPool is no longer created by the Node class

backend/tests/component/core/schema/test_attribute_parameters.py
```diff
@@ -172,24 +182,6 @@ def test_number_pool_override_generic() -> None:
         schema_branch.process()
 
 
-def test_number_pool_assign_from_generics() -> None:
```
ajtmccarty, line 175: we don't do this any longer

backend/tests/component/graphql/mutations/test_resource_manager.py
```diff
@@ -919,57 +930,7 @@ async def test_delete_number_pool_in_use_by_numberpool_attribute(
 
     assert not query_before_creation.errors
     assert query_before_creation.data
-    assert query_before_creation.data["CoreNumberPool"]["count"] == 0
```
ajtmccarty, line 922: this wasn't necessary to test preventing deleting CoreNumberPools

backend/infrahub/core/migrations/schema/node_attribute_add.py
```diff
@@ -99,12 +106,12 @@ async def execute_post_queries(
 
         async def allocate_numbers(db: InfrahubDatabase) -> None:
             for node in nodes:
-                number = await number_pool.get_resource(
+                number = await number_pool.get_resource(  # type: ignore[attr-defined]
```
ogenstad, line 109: Not really an issue with this PR. But perhaps we should rethink this `.get_resource()` method that we add to some of the nodes.. One option could of course be to generate this method on any resource pool kind. Another approach would just be to instead use a separate function called "get_resource()` which took a node as its input. This is probably part of a larger refactoring, mainly just thinking out loud now.

backend/infrahub/core/node/__init__.py
```diff
@@ -351,39 +347,53 @@ async def handle_pool(
...
+                    )
+                )
+                return
+            attribute.from_pool = {"id": number_pool_id}
+            attribute.is_default = False
+        # Case 2: User-specified from_pool on a regular Number attribute
+        elif attribute.from_pool:
+            number_pool_id = attribute.from_pool.get("id")
+            if not number_pool_id:
+                errors.append(ValidationError({f"{attribute.name}.from_pool": "No pool ID specified in from_pool."}))
```
ogenstad, line 379: It could very well be that I wrote this originally. But I think we should start to transition away from these types of methods. My problem is that `handle_pool` takes a list of errors as input and we mutate that list within the code of that method. As a caller it's not very clear that this is something that happens at all. You don't have to change anything in this PR. However I think a cleaner approach for us in the future would be that `handle_pool()` returns a list of errors and that the caller would do something like:

```python
errors.extend(await self.handle_pool(db=db, attribute=attribute, allocate_resources=process_pools))
```

backend/infrahub/pools/schema_number_pool_synchronizer.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        updated_branches: set[str] = set()
+
+        for branch_name in self.schema_manager.get_branches():
+            schemas_to_update: list[str] = []
+            schema_branch = self.schema_manager.get_schema_branch(name=branch_name)
+
+            # Process generics first so their pool IDs are available for inheriting nodes
+            for generic_name in schema_branch.generic_names:
+                generic_schema = schema_branch.get_generic(name=generic_name, duplicate=False)
```
ogenstad, line 117: This might be fine but I just want to highlight a fix we merged in december: https://github.com/opsmill/infrahub/pull/7810, where the problem was that we were modifying the non duplicated node before storing it again. It looks like the changes in `self._process_schemanode()` might be safe in this regard as we modify the attributes and have a duplicate function there. Adding the comment just so you can have an extra sanity check.

backend/infrahub/pools/schema_number_pool_upserter.py
```diff
@@ -0,0 +1,260 @@
...
+                name=f"{pool_kind}.{attribute.name} [{number_pool_id}]",
+                node=pool_kind,
+                node_attribute=attribute.name,
+                start_range=attribute.parameters.start_range,
+                end_range=attribute.parameters.end_range,
+                pool_type=NumberPoolType.SCHEMA.value,
+            )
+            await number_pool.save(db=self.db, at=at, user_id=user_id)
+
+            # Re-fetch using _get_by_id to get the proper CoreNumberPool instance with its methods
```
ogenstad, line 176: Also a comment regarding the `.get_resource()` again. It seems a bit off that we'd need to query for the same node we just created just because we want to add some methods in a magical way. Not for this PR but to consider when we do the larger refactoring.

ogenstad (approved):

backend/infrahub/core/migrations/schema/node_attribute_add.py
ajtmccarty: yes I agree
we should get rid of the `get_resource` method completely
we should have a `ResourceGetterInterface` and different implementations for number pools, ip address pools, and ip prefix pools

backend/infrahub/core/node/__init__.py
ajtmccarty: yes I agree. we should almost never be mutating arguments within a function/method

backend/infrahub/pools/schema_number_pool_upserter.py
ajtmccarty: definitely agree. should not be on the `Node` class

PR #8254, opened 2026-01-28
IFC-2219 Add API endpoint to download files attached to `CoreFileObject`

gmazoyer:
This change adds an API endpoint to allow files attached to `CoreFileObject` nodes to be downloaded. The endpoint is now exposed as `/api/storage/files/<storage_id>` and also takes branch query parameters when required. This is needed, not to actually download a file, but rather to check for the permission to download it. Endpoint name can be discussed.

This covers IFC-2173 by adding the required method to handle non-text based files. Our existing implementation of file retrieval in storage only supports text files as it tries to decode the data stream. We now have another method that just read the data stream but does not try to decode it, thus handling any kind of files.

This also covers IFC-2175 as we are making sure that a user has the permission to view a `CoreFileObject` in order to download the file attached to it. We perform a simple lookup in the database given the file's storage ID to find the node it is attached to. We then do a simple object permission check.

## Summary by CodeRabbit

* **New Features**
  * Download files by storage ID with full binary support, correct Content-Type, and safe Content-Disposition (handles non‑ASCII and injection risks)
  * Permission‑enforced access requiring VIEW rights for downloads

* **API**
  * Added REST endpoint and OpenAPI entry for storage-based file downloads; frontend types updated for the new path

* **Tests**
  * Comprehensive unit and integration tests covering binary/text retrieval, headers, and permission scenarios


backend/infrahub/api/storage/file_object.py
```diff
@@ -0,0 +1,117 @@
...
+from infrahub.core.protocols import CoreFileObject
+from infrahub.database import InfrahubDatabase  # noqa: TC001
+from infrahub.exceptions import NodeNotFoundError
+from infrahub.permissions import define_object_permission_from_branch
+
+if TYPE_CHECKING:
+    from infrahub.auth import AccountSession
+    from infrahub.permissions import PermissionManager
+
+router = APIRouter(prefix=f"/{InfrahubKind.FILEOBJECT}")
```
ogenstad: Here the URL is hardcoded as CoreFileObject. Previously we talked about the option of having this name dynamic in the endpoint so that it would be possible to download artifacts here as well. I.e. we'd do a schema lookup for the "kind part" of the endpoint. Potentially we could even use that part for the permission check before doing a DB query to collect the kind. It would be nice to not require this query to the database but I'm not sure we can avoid that as we don't want someone to fool the system by just adding another "kind" in the URL and bypass the security that way.

It could also be that the above approach is better if a specific kind would be both a FILEOBJECT kind as well as being an artifact target.

I'm not asking for a change here, just that you can consider the impact.

backend/infrahub/api/__init__.py
```diff
@@ -35,6 +36,7 @@
 router.include_router(auth.router)
 router.include_router(diff.router)
 router.include_router(file.router)
+router.include_router(file_object.router)
```
ogenstad: From an organizational point of view I think it would make sense to have these routes under the /storage namespace. Even if it might not be a perfect fit I think it would be nice to not spread it out into different locations. What are your thoughts?

ogenstad (commented):
LGTM. Mainly have thoughts around the structure of the URL endpoints. What do you think about those?

backend/infrahub/api/__init__.py
gmazoyer: That is a good idea. I did not think of it while crafting this PR, but nesting this under `/storage` makes a lot of sense.

backend/infrahub/api/storage/file_object.py
gmazoyer: We'll need the database query in any case, because we need the file MIME type to be able to send the correct content type in the query response.

Still I will have a look if we can include artifacts in there, I think it's just a matter of figuring the dynamic routing thing out.

gmazoyer: Thinking about this a bit more. The advantage of hardcoding the URL with `CoreFileObject` or anything else, is that we avoid URL manipulation, it also makes the OpenAPI spec really clear. I'm not sure what the spec would look like if we'd have a dynamic kind.

The database query is unavoidable for several reasons:
1. We need the node's actual schema to check permissions correctly
2. We need the node's file_name and file_type attributes for response headers

That said, if we'd go with the dynamic kind, I believe the endpoint code would look roughly like this one (just for reference not working code):
```python
@router.get("/{kind}/{storage_id}")
async def download_file(
    kind: str,
    storage_id: str,
    db: InfrahubDatabase = Depends(get_db),
    branch_params: BranchParams = Depends(get_branch_params),
    permission_manager: PermissionManager = Depends(get_permission_manager),
    _: AccountSession = Depends(get_current_user),
) -> Response:
    # Validate kind exists in schema registry
    schema = registry.schema.get(name=kind, branch=branch_params.branch, raise_on_error=True)

    if InfrahubKind.FILEOBJECT in schema.inherit_from:
        base_schema = CoreFileObject
    elif InfrahubKind.ARTIFACT in schema.kind:
        base_schema = CoreArtifact
    else:
        raise HTTPException(400, f"{kind} is not a downloadable type")

    # Query node by storage_id
    nodes = await registry.manager.query(
        db=db, schema=base_schema,
        filters={"storage_id__value": storage_id},
        branch=branch_params.branch, at=branch_params.at, limit=1,
    )
    if not nodes:
        raise NodeNotFoundError(...)

    node = nodes[0]
    actual_schema = node.get_schema()

    # Check permission
    permission = define_object_permission_from_branch(
        schema=actual_schema,
        action=PermissionAction.VIEW,
        branch_name=branch_params.branch.name
    )
    permission_manager.raise_for_permission(permission=permission)

    # Return file with appropriate headers
    return Response(
        content=registry.storage.retrieve_binary(identifier=storage_id),
        media_type=node.file_type.value if base_schema == CoreFileObject else "text/plain",
        headers={"Content-Disposition": build_content_disposition(get_filename(node))},
    )
```

ajtmccarty (approved):

backend/infrahub/api/storage/__init__.py
```diff
@@ -0,0 +1,6 @@
+from infrahub.api.storage import file_object
```
dgarros, line 1: This looks like a pattern we want to avoid, we probably need to update the guidelines to ensure that Claude doesn't put content in the __init__ file

backend/infrahub/api/storage/file_object.py
```diff
@@ -0,0 +1,117 @@
...
+
+    https://developer.mozilla.org/docs/Web/HTTP/Reference/Headers/Content-Disposition
+    """
+    ascii_filename, encoded_filename = sanitize_filename(filename)
+    # Use both filename (for older clients) and filename* (for RFC5987 compliant clients)
+    return f"attachment; filename=\"{ascii_filename}\"; filename*=UTF-8''{encoded_filename}"
+
+
+@router.get(
+    "/{storage_id:str}",
```
dgarros, line 72: In addition to the storage id, we need to expose an API that is based on the UUID and/or the HFDI of the Object itself.
The storage id is gonna keep changing so from a consumer point of view it means that we'll need to do a lookup first to get the storage it.
It looks simple but it can lead to a lot of additional complexity on the client side.

PR #8260, opened 2026-01-28
Fixed misleading schema loading error pointing to the wrong attribute

ogenstad:
This PR fixes an issue related to the backend highlighting an incorrect index for attributes when there is a problem with a schema that users are loading. The issue is reported in IFC-2089. The root cause was that we had a schema with an attribute that looked like this:

```yaml
      - name: name
        kind: Text
      - name: description
        kind: Text
      - name: signaling
        kind: Dropdown
        label: "Signaling"
        optional: false
        choices:
          - name: tldp
            label: "TLDP"
          - name: off (due to yaml issues this needs to be "off")
            label: "Off"
          - name: bgp
            label: "BGP"
        default_value: tldp
```

The problem here is that the yaml parser reads the name "off" as a boolean and a boolean is an invalid type for a name. This coupled with the `.set_attribute_type()` validator modifies the attributes during validation and mixes up the ordering so we loose information about the index of the attributes within the error message. When combined with `infrahubctl` the above error would show up on attribute index 1 i.e. the second attribute of the node (i.e. "description"). This is because "off" is the second dropdown. However since the attribute index was lost the error message got very confusing. As such the error indicated that something was incorrect with the "description" attribute. The change in this PR is to ensure that the correct order of the attributes are preserved when there is an error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema validation error messages to preserve exact location (list index and attribute name) for nested/mis-typed attributes, fixing misleading highlights for Dropdown names parsed as booleans (e.g., "off").
* **Tests**
  * Added a test to ensure Dropdown choice-name validation errors include the attribute name and nested path.
* **Documentation**
  * Added changelog entry describing the fix.

### DIFF
--- a/backend/infrahub/core/schema/basenode_schema.py
+++ b/backend/infrahub/core/schema/basenode_schema.py
@@ -6,10 +6,10 @@ import os
 from collections import defaultdict
 from dataclasses import asdict, dataclass
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Iterable, Literal, overload
+from typing import TYPE_CHECKING, Any, Callable, Iterable, Literal, NoReturn, overload
 
 from infrahub_sdk.utils import compare_lists, intersection
-from pydantic import ConfigDict, field_validator
+from pydantic import ConfigDict, ValidationError, field_validator
 
 from infrahub.core.constants import HashableModelState, RelationshipCardinality, RelationshipKind
 from infrahub.core.models import HashableModel, HashableModelDiff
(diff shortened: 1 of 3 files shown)


backend/tests/component/core/test_schema.py
```diff
@@ -440,6 +440,50 @@ def test_dropdown_choice_sort() -> None:
     assert active < passive
 
 
+def test_dropdown_validation_error_with_invalid_choice_name() -> None:
+    """Test that validation errors for dropdown choices include attribute name and nested path.
+
+    This simulates the issue from IFC-2089 where YAML parsing converts 'off' to boolean False,
+    and the error message should clearly indicate which attribute and which choice field has the problem.
```
gmazoyer, line 447: This happened to me so many times… Good thing we can now have a clear error message for the user to understand the issue.

gmazoyer (approved):

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -124,19 +124,44 @@ def set_attribute_type(cls, raw_attributes: Any) -> Any:
...
+                    # Enhance error with attribute name and correct index in location path
+                    errors = []
+                    for error in e.errors():
+                        error_copy = error.copy()
+                        # Insert the correct list index at the start of the location path
+                        error_copy["loc"] = (idx,) + error["loc"]
+                        nested_path = ".".join(str(item) for item in error["loc"])
+                        error_copy["input"] = f"[{attr_name}.{nested_path}]: {error.get('input', 'N/A')}"
+                        errors.append(error_copy)
+                    raise ValidationError.from_exception_data(title=e.title, line_errors=errors) from e
```
ajtmccarty: could this be de-duplicated with a little private helper method in here?

ajtmccarty (approved):

backend/infrahub/core/schema/basenode_schema.py
ogenstad: Yes good point. I added a helper method.

PR #8266, opened 2026-01-29
IFC-1640: Rename generic unique attributes

solababs:
## Release Notes

* **New Features**
  * Enhanced schema inheritance system with improved tracking of inherited attributes and their source definitions.

* **Improvements**
  * Better attribute identification and mapping during schema migrations.
  * Optimized handling of inherited fields in schema processing workflows.
  * Improved support for attribute renames within inherited schema definitions.


backend/infrahub/core/schema/node_schema.py
ajtmccarty: ideally, the logic in this method would move to separate component outside of the the `NodeSchema` class. I see that you are following the pattern of the current state of the code, but the `NodeSchema` should not be handling this logic and this is only used in `SchemaBranch.process/process_inheritance`. I'd say that this logic and the changes in `AttributeSchema` and `BaseNodeSchema` should be moved into a new `NodeInheritanceHandler` (or something like that) class so all of this logic exist in the place where it is actually used.

backend/infrahub/core/schema/node_schema.py
```diff
@@ -135,6 +164,62 @@ def inherit_from_interface(self, interface: GenericSchema) -> None:
                 item_idx = existing_inherited_relationships[relationship.name]
                 self.relationships[item_idx].update_from_generic(other=new_relationship)
 
+        if renamed_attrs:
+            self._update_uniqueness_constraints_for_renamed_attributes(renamed_attrs)
+            self._update_hfid_for_renamed_attributes(renamed_attrs)
+            self._update_order_by_for_renamed_attributes(renamed_attrs)
+            self._update_default_filters_for_renamed_attributes(renamed_attrs)
+
+    def _update_default_filters_for_renamed_attributes(self, renamed_attrs: dict[str, str]) -> None:
```
ajtmccarty: these methods look like they could be DRYed out

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -478,6 +512,51 @@ async def test_step01_baseline_backend(
         )
         assert not errors
 
+    async def test_step01_5_rename_unique_generic_attr(
```
ajtmccarty: this should either become `step02` and all the following test steps should be incremented or it should move to the end of the test methods.
this test should validate the full output of  `client.schema.check` as the other tests in this file do
there should be another test after this one that validates loading the schema and then retrieves it from the database and verifies that it has been updated as expected, like the other tests in this file do

backend/infrahub/core/schema/__init__.py
ajtmccarty, line 1: we should avoid adding to these `__init__` files unless it is necessary for backwards compatibility. they make the import path a little shorter, but they make circular import errors more likely when too many things are all linked together in them

backend/infrahub/core/schema/attribute_schema.py
```diff
@@ -38,6 +38,17 @@ def get_attribute_schema_class_for_kind(kind: str) -> type[AttributeSchema]:
...
+    # Used for rename detection in diffs while keeping id=None for inherited attrs
+    _source_attribute_id: str | None = PrivateAttr(default=None)
+
+    @property
+    def source_attribute_id(self) -> str | None:
+        return self._source_attribute_id
+
+    @source_attribute_id.setter
+    def source_attribute_id(self, value: str | None) -> None:
+        self._source_attribute_id = value
```
ajtmccarty, line 51: I think this is okay because these seems like data an Attribute should track about itself: if it is inherited and the ID of the attribute it inherits from

backend/infrahub/core/schema/node_inheritance_handler.py
```diff
@@ -0,0 +1,135 @@
+from .generic_schema import GenericSchema
+from .node_schema import NodeSchema
+
+
+class NodeInheritanceHandler:
```
ajtmccarty, line 5: outstanding. moving logic like this into its own component is going to make our lives much easier

backend/infrahub/core/schema/node_inheritance_handler.py
```diff
@@ -0,0 +1,135 @@
...
+                    renamed_attrs[old_name] = new_attribute.name
+                    # Update tracking structures to prevent stale lookups
+                    existing_inherited_attributes.pop(old_name, None)
+                    existing_inherited_attributes[new_attribute.name] = item_idx
+                    existing_inherited_fields.remove(old_name)
+                    existing_inherited_fields.append(new_attribute.name)
+                node.attributes[item_idx].update_from_generic(other=new_attribute)
+                node.attributes[item_idx].name = new_attribute.name  # Update name explicitly
+                node.attributes[item_idx].source_attribute_id = new_attribute.source_attribute_id
+            elif attribute.name not in existing_inherited_attributes:
```
ajtmccarty, line 62: could you add comments for the `elif` and `else` here saying what each case is?
I think the `elif` is for a new attribute added on the generic and the `else` is for an existing inherited attribute that has not been renamed, but it'd be easier to understand if it was written here

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -1430,6 +1509,204 @@ async def test_step06_load_schema_with_override_deletes(
...
+        initial_dataset: dict[str, Node],
+        schema_step_07_rename_unique_generic_attr: dict[str, Any],
+    ) -> None:
+        await self._finalize_deleted_and_renamed_fields(
+            db=db,
+            branch=branch,
+            full_schema_dict=schema_step_07_rename_unique_generic_attr,
+            rename_map={
+                GENERIC_KIND: {
+                    "generic_unique_attr_new": "generic_unique_attr",
```
ajtmccarty, line 1526: I would expect the dictionary to go the other way, from current name to new name, but maybe that's just me

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -1430,6 +1509,204 @@ async def test_step06_load_schema_with_override_deletes(
...
+        assert generic_diff["changed"]["attributes"]["changed"]["generic_unique_attr_new"]["changed"] == {"name": None}
+
+        # All specific node schemas should show the inherited attribute rename
+        for specific_kind in [SPECIFIC_ONE_KIND, SPECIFIC_TWO_KIND, SPECIFIC_THREE_KIND]:
+            assert specific_kind in diff["changed"], f"{specific_kind} should be in changed"
+            specific_diff = diff["changed"][specific_kind]
+            assert "generic_unique_attr_new" in specific_diff["changed"]["attributes"]["changed"]
+            assert specific_diff["changed"]["attributes"]["changed"]["generic_unique_attr_new"]["changed"] == {
+                "name": None
+            }
```
ajtmccarty: I think this is all covered by the dictionary assertion above

ajtmccarty (commented):
looks very good to me
not approving yet, b/c I think the test failure indicates something is off with how we are saving the inherited schemas to the database and I want to review any changes there if they are necessary

backend/infrahub/core/models.py
```diff
@@ -260,6 +260,9 @@ def _process_attrs_rels(
         for field_name, sub_field_diff in node_field_diff.changed.items():
             field = schema.get_field(name=field_name)
 
+            if schema.is_node_schema and field.inherited:
+                continue
+
```
ajtmccarty, line 265: I'm worried this will change how we run migrations in an unexpected way.
It definitely _seems_ reasonable to skip adding migrations for an inherited field b/c the corresponding field on the GenericSchema should kick off a migration that handles any of its inheriting schemas too, but I am not totally sure we do that correctly everywhere
can you say why this change is necessary?

solababs: @ajtmccarty When an attribute is renamed on a Generic, data nodes have both labels (their specific kind AND the Generic's kind). 
The Generic's migration already handles all nodes with the Generic's label, so running migrations for both would create duplicate Attribute nodes.

ajtmccarty (approved):

PR #8272, opened 2026-01-29
freeze diff for a closed proposed change

ajtmccarty:
IFC-2224

adds support for "freezing" a diff when the proposed change it is associated with is closed to allow users to view the diff of the changes up to the point that the PC was closed and to prevent it from being updated

"freezing" a diff entails setting `is_frozen=True` on the `:DiffRoot` vertex in the database and updating the tracking_id to be a `FrozenTrackingId` instead of a `BranchTrackingId` b/c a `BranchTrackingId` implies that it is tracking all of the changes for the lifecycle of the branch and that is not the case once it is frozen

TODO:
- [x] test closing a PC and then viewing the diff in the branch to confirm that still works as expected
  - this actually fails, but it's a frontend issue and I've let them know 
- [x] test proposed change link is correctly removed when updating a branch diff with an earlier frozen diff

## Summary by CodeRabbit

* **New Features**
  * Proposed-change closure now freezes associated diffs; frozen diffs preserve historical tracking and remain queryable.
  * Repository API supports an option to include frozen diffs when deleting; bulk deletes can opt into removing frozen items.
  * Diff save/retrieve flows propagate and expose an is_frozen flag.

* **Tests**
  * New unit and integration tests validate freezing, protected deletion, coexistence of frozen and active diffs, and related tracking IDs.

### DIFF
--- a/backend/infrahub/core/diff/model/path.py
+++ b/backend/infrahub/core/diff/model/path.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from copy import deepcopy
 from dataclasses import asdict, dataclass, field
 from enum import Enum
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Self
 
 from infrahub.core.constants import (
     BranchSupportType,
@@ -41,7 +41,7 @@ class TrackingId:
         return f"{self.prefix}{self.delimiter}{self.name}"
 
     @classmethod
-    def deserialize(cls, id_string: str) -> TrackingId:
+    def deserialize(cls, id_string: str) -> Self:
         if not id_string.startswith(cls.prefix):
             raise ValueError(
                 f"Cannot deserialize TrackingId with incorrect prefix '{id_string}', expected prefix '{cls.prefix}{cls.delimiter}'"
--- a/backend/infrahub/core/diff/model/path.py
+++ b/backend/infrahub/core/diff/model/path.py
@@ -71,8 +71,12 @@ class NameTrackingId(TrackingId):
     prefix = "name"
 
 
+class FrozenTrackingId(TrackingId):
+    prefix = "frozen"
+
+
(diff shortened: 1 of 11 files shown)


backend/infrahub/core/diff/query/delete_query.py
```diff
@@ -9,15 +9,28 @@ class EnrichedDiffDeleteQuery(Query):
...
-            self.params = {"diff_root_uuids": self.enriched_diff_root_uuids}
-            diff_filter = "WHERE d_root.uuid IN $diff_root_uuids"
+            self.params["diff_root_uuids"] = self.enriched_diff_root_uuids
+            diff_filters.append("d_root.uuid IN $diff_root_uuids")
+        if not self.include_frozen:
+            diff_filters.append("(d_root.is_frozen IS NULL OR d_root.is_frozen <> TRUE)")
+
+        diff_filter = ""
+        if diff_filters:
+            diff_filter = "WHERE " + " AND ".join(diff_filters)
```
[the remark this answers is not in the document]
ajtmccarty: this is the intended behavior

backend/infrahub/core/diff/model/path.py
```diff
@@ -71,8 +71,12 @@ class NameTrackingId(TrackingId):
     prefix = "name"
 
 
+class FrozenTrackingId(TrackingId):
+    prefix = "frozen"
```
ogenstad, line 75: Not an issue with this code just highlighting as we're adding to a parent class. `TrackingId.deserialize()` returns a `TrackingId` but according to the classmethod `FrozenTrackingId.deserialize()` would also return a `TrackingId`. I don't know if we use it in this way anywhere but this might not be what we want..

ogenstad (approved):

backend/infrahub/core/diff/model/path.py
ajtmccarty: updated the signature on `TrackingId.deserialize` to return `Self`

PR #8223, opened 2026-01-24
add SchemaNode/Generic uniqueness constraints and test

ajtmccarty:
fixes #8221 

adds `[name, namespace]` uniqueness constraints to `SchemaNode` and `SchemaGeneric` so that it is not possible to create the same schemas on different branches and merge them. we assume that each `SchemaNode` and `SchemaGeneric` has a unique namespace-name combination, but we have not actually enforced it when merging a branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where duplicate schemas could be created on feature branches and unintentionally duplicated when merging back into the default branch.

* **New Features**
  * Enhanced schema node identification with human_friendly_id for improved lookup operations.
  * Implemented composite uniqueness constraints using namespace and name combinations to ensure schema uniqueness across branches.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #8285, opened 2026-02-03
migration for SchemaNode/Generic uniquness constraint updates

ajtmccarty:
follow-up to #8223 

had a chicken-egg situation for branches created on a system before #8223 was merged. #8223 updated uniqueness constraints to be correct for SchemaNode and SchemaGeneric (updated from `[name__value]` to `[namespace__value, name__value]`) and included the two schemas in uniqueness constraint validation. the problem is that the uniqueness constraint update will only happen on the default branch. if a branch was created before #8223 was applied, then rebasing the branch would fail because SchemaNode and SchemaGeneric would be included in uniqueness constraint validation, but using the old incorrect uniqueness constraint value of `[name__value]`, which would raise an error if any Schemas shared a name (not kind, just name, so LocationGeneric and OrganizationGeneric would be illegal).

the fix is to update the uniqueness constraints on any branch where they are not correct, which is what the new migration does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed uniqueness constraint on name fields in schema definitions, enabling proper branch rebasing for users who created branches before a previous fix.

* **Chores**
  * Updated schema migration framework to version 56 with improved constraint handling and database safety measures.


backend/infrahub/core/schema/manager.py
```diff
@@ -378,7 +378,8 @@ async def update_node_in_db(
 
         schema_dict = node.model_dump(exclude=IGNORE_FOR_NODE)
         for key, value in schema_dict.items():
-            getattr(obj, key).value = value
+            if obj_attr := getattr(obj, key, None):
+                obj_attr.value = value
```
ajtmccarty, line 382: added this change after trying to upgrade a backup of an internal instance and encountering this error 
```
  File "/infrahub/backend/infrahub/cli/db.py", line 361, in migrate_database
    execution_result = await migration.execute(migration_input=MigrationInput(db=db))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/infrahub/backend/infrahub/core/migrations/graph/m056_update_schema_node_generic_constraints.py", line 103, in execute
    await update_schema_on_branch(db=db, branch=branch, at=at, user_id=user_id)
  File "/infrahub/backend/infrahub/core/migrations/graph/m056_update_schema_node_generic_constraints.py", line 70, in update_schema_on_branch
    await manager.update_node_in_db(
  File "/infrahub/backend/infrahub/core/schema/manager.py", line 381, in update_node_in_db
    getattr(obj, key).value = value
    ^^^^^^^^^^^^^^^^^
AttributeError: 'Node' object has no attribute 'display_label'. Did you mean: '_display_label'?
```

seems like a safe change to make but if others disagree I can come back and do the branch updates to the `uniqueness_constraints`, `human_friendly_id`, and `name.unique` using plain cypher and revert this change

backend/tests/component/core/migrations/graph/test_056_update_schema_node_generic_constraints.py
ogenstad, line 1: I think we should start naming the files for the tests in the same way as the actual implementation. i.e. `test_m056_update_schema_node_generic_constraints.py` in this case. Mainly because it helps with rebase operations if we have two files called `test_m056.py` but they aren't related at all.

backend/infrahub/core/schema/manager.py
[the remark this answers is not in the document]
ogenstad: Since the code within the if-statement calls `obj_attr.value = value` this statement looks incorrect. I.e. for a falsy value if `if obj_attr := getattr(obj, key, None):` returned an empty list we'd run into a problem when setting the `.value` attribute of a list. This comment doesn't seem to take this into consideration.

ogenstad (approved):

PR #8297, opened 2026-02-03
Do not send proposed change id param for diff view on branch page

pa-lem:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed diff view to correctly resolve and display branch information when switching between viewing contexts, ensuring diffs are properly fetched and displayed with the appropriate context.


bilalabbad (approved):
In the future a better solution would be to remove the proposed change atom

PR #8299, opened 2026-02-03
IFC-21844: Freeze branch after merge

solababs:
Implements Freeze branch after merge from dev/specs/2026-01-branch-freeze.md

## Summary by CodeRabbit

* **New Features**
  * Introduces a MERGED read-only branch state: mutations, rebases, and proposed-change creation from merged sources are blocked; open proposed changes are auto-canceled; branch deletion remains allowed.

* **API & Middleware**
  * Status checks unified across REST and GraphQL; operations on blocked branches return clear validation errors.

* **Permissions**
  * Non-view actions are denied for MERGED and NEED_REBASE branches (view and delete exceptions preserved).

* **Tests**
  * Added extensive unit, functional, and component tests covering merged and need-rebase behaviors.


backend/infrahub/api/schema.py
```diff
@@ -324,6 +325,7 @@ async def load_schema(
     context: InfrahubContext = Depends(get_context),
 ) -> SchemaUpdate:
     check_need_rebase_status(branch)
+    check_merged_status(branch)
```
ogenstad: We have this pattern on multiple endpoints now:
```python
    check_need_rebase_status(branch)
    check_merged_status(branch)
```

I think it would be better to introduce a wrapper so that we only have `check_branch_status(branch)` then have this new function call the other checkers.

backend/infrahub/core/branch/merged_status.py
```diff
@@ -0,0 +1,7 @@
+from infrahub.core.branch import Branch
+from infrahub.core.branch.enums import BranchStatus
+
+
+def check_merged_status(branch: Branch) -> None:
+    if branch.status == BranchStatus.MERGED:
+        raise ValueError(f"Branch '{branch.name}' has been merged and is read-only. No modifications are allowed.")
```
ogenstad: Coderabbit complained about `ValueError` in some other location and I think it might be correct that we should use an error from infrahub.exceptions instead. Can we have a test that tries to use one of the REST API endpoints with a branch in a merged status and validate the return code. It should be a 4xx error and not a 5xx error that gets returned. It could be that the `ValueError` would instead yield a 500 internal server error.

backend/tests/unit/core/branch/test_merged_status.py
```diff
@@ -0,0 +1,32 @@
...
+    "status",
+    [
+        BranchStatus.OPEN,
+        BranchStatus.NEED_REBASE,
+        BranchStatus.DELETING,
+        BranchStatus.NEED_UPGRADE_REBASE,
+    ],
+)
+def test_check_merged_status_passes_for_non_merged_branch(status: BranchStatus) -> None:
+    mock_branch = MagicMock()
```
ogenstad: I don't think it adds much value to use MagicMock() here, it should be fine to just use a `Branch` object directly we don't need to store it in the database or anything just having the Pydantic object for a branch should be enough to test this.

backend/infrahub/permissions/report.py
```diff
@@ -28,6 +29,11 @@ def get_permission_report(  # noqa: PLR0911
     if global_permission_report[GlobalPermissions.SUPER_ADMIN]:
         return BranchRelativePermissionDecision.ALLOW
 
+    # Block mutations on merged branches
+    # Note: Branch delete is allowed via middleware, this covers node permissions
+    if branch.status == BranchStatus.MERGED and action != "view":
```
ogenstad: Can you also add the need rebase status here as we'd want to deny anything aside from "view" for that one as well?

ogenstad (commented):
Looks good overall to me. I think for the unittests we can have another look to see where we can remove the magic mocks and instead use real objects where it's possible to replace those that doesn't require a database or any other processing.

backend/infrahub/api/validators.py
ajtmccarty, line 1: I'd say this should live in `backend/infrahub/branch/status_checker.py`. I would think it will be used in places other than the API

and should be 
```
class BranchStatusChecker:
    def check(self, branch: Branch) -> None:
        if branch.status == BranchStatus.NEED_REBASE:
            raise BranchNeedsRebaseError(...)
        if branch.status == BranchStatus.MERGED:
            raise BranchAlreadyMergedError(...)
```
with no `__init__` required for now
where `BranchNeedsRebaseError` and `BranchAlreadyMergedError` inherit from a new `BranchStatusError` in infrahub.exceptions

backend/infrahub/graphql/middleware.py
```diff
@@ -1,12 +1,16 @@
+from infrahub.core.branch.merged_status import check_merged_status
 from infrahub.core.branch.needs_rebase_status import check_need_rebase_status
 
 ALLOWED_MUTATIONS_ON_NEED_REBASE_BRANCH = ["BranchRebase", "BranchDelete", "BranchCreate", "ProposedChangeCreate"]
+ALLOWED_MUTATIONS_ON_MERGED_BRANCH = ["BranchDelete"]
 
 
 def raise_on_mutation_on_branch_needing_rebase(next, root, info, **kwargs):  # type: ignore  # noqa
```
ajtmccarty: maybe `raise_on_mutation_for_branch_status`?

backend/tests/unit/graphql/test_middleware.py
ajtmccarty, line 1: this is a lot of mocking. I would think that the functional or integration tests can cover this

ajtmccarty (commented):
looks pretty good. I don't think I've ever made this comment before, but this might have too many tests, although too many is definitely better than not enough.
could you please add a PR description? even if it's just implements "/dev/spec/...."

backend/infrahub/api/validators.py
gmazoyer: I agree with this comment.

backend/infrahub/graphql/mutations/proposed_change.py
```diff
@@ -100,6 +100,17 @@ async def mutate_create(
...
+            try:
+                source_branch_obj = await Branch.get_by_name(db=dbt, name=source_branch_name)
+            except BranchNotFoundError:
+                raise ValidationError(
+                    input_value="The specified source branch for this proposed change was not found"
+                ) from None
+            if source_branch_obj.status == BranchStatus.MERGED:
+                raise ValidationError(
+                    input_value=f"Cannot create proposed change: branch '{source_branch_name}' has been merged"
+                )
```
ajtmccarty: this and the logic block above it (which I added, :shame) probably don't need to be in the transaction b/c they are just reading data

backend/infrahub/api/schema.py
```diff
@@ -323,7 +324,10 @@ async def load_schema(
     permission_manager: PermissionManager = Depends(get_permission_manager),
     context: InfrahubContext = Depends(get_context),
 ) -> SchemaUpdate:
-    check_need_rebase_status(branch)
+    try:
+        BranchStatusChecker.check(branch=branch)
+    except BranchStatusError as err:
+        raise SchemaNotValidError(message=str(err)) from err
```
ajtmccarty: this seems like the wrong error, judging from the name
the schema to load might be perfectly valid, but it is not allowed because of the state of the branch

backend/infrahub/branch/status_checker.py
```diff
@@ -0,0 +1,17 @@
+from infrahub.core.branch import Branch
+from infrahub.core.branch.enums import BranchStatus
+from infrahub.exceptions import BranchAlreadyMergedError, BranchNeedsRebaseError
+
+
+class BranchStatusChecker:
+    @staticmethod
```
ajtmccarty: it doesn't really matter in this case b/c the component is so simple, but to really support dependency injection/inversion, this should be a regular instance method so that it can make use of any components we might inject into a future `BranchStatusChecker.__init__` method

backend/infrahub/branch/status_checker.py
ajtmccarty, line 1: is it possible to consolidate the `check_merge_status` and `check_need_rebase_status` functions within this class?
I'm thinking some thing like
```
def check_merge_status(self, branch: Branch) -> None:
    if branch.status == BranchStatus.MERGED:
         raise BranchAlreadyMergedError(...)

def check_need_rebase_status(self, branch: Branch) -> None:
    if branch.status == BranchStatus.NEEDS_REBASE:
         raise BranchNeedsRebaseError(...)

def check(self, branch: Branch) -> None:
    self.check_needs_rebase_status(branch)
    self.check_merge_status(branch)
```
then it is also possible to call just one of the `check_xyz` methods where only one is required

ajtmccarty (approved):
looks great
thanks for handling my comments

PR #8315, opened 2026-02-05
add _from_resource_pool Relationships to template generation

ajtmccarty:
IFC-2248

updates TemplateSchema generation so that if a given schema has a relationship to a BuiltinIpPrefix/BuiltinIpAddress (or a schema inheriting from one of them), then the generated TemplateSchema gets an extra `_from_resource_pool` Relationship with a CoreIpAddressPool or CoreIpPrefixPool peer that will be used to handle provisioning relationships from resource pools when creating an object using an instance of this Template

also had to bump some character limits from 32 to 64 to support the `xyz_from_resource_pool` relationship name

## Summary by CodeRabbit

* **New Features**
  * Templates now automatically gain corresponding resource-pool relationships for IP address and IP prefix peers.

* **Chores**
  * Increased max length constraints for identifiers, names, labels, and namespaces (32 → 64).

* **Tests**
  * Added unit and functional tests verifying template resource-pool relationship creation and properties.


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2496,6 +2496,55 @@ def manage_object_template_relationships(self) -> None:
...
+            return None
+
+        pool_rel_name = f"{relationship.name}_from_resource_pool"
+        pool_identifier = f"{relationship.get_identifier()}_from_resource_pool"
+        pool_label = relationship.name.title() + " from Resource Pool"
+
+        return RelationshipSchema(
+            name=pool_rel_name,
+            peer=pool_peer,
+            kind=relationship.kind,
```
ogenstad: I think CodeRabbit has a valid point regarding the potential for causing problems with parent/component type relationships. I'm not sure we'd ever see this in the wild for any sensible schemas but since this is pretty much just a virtual type I think we can hardcode the kind. Not sure which one is the best to just use `RelationshipKind.GENERIC`. If we didn't need to export all kinds at the moment it could have made more sense to have some new virtual kind but I think it might just cause confusion to introduce a new relationship kind just for this. At some point it would be nice if we could revert this and instead store this kind of data in another way that didn't require this dummy relationship. So I don't think we should count on this being something long term (famous last words..)

schema/openapi.json
```diff
@@ -2245,7 +2245,7 @@
                     },
                     "namespace": {
                         "type": "string",
-                        "maxLength": 32,
+                        "maxLength": 64,
                         "minLength": 3,
                         "pattern": "^[A-Z][a-z0-9]+$",
```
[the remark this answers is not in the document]
ogenstad: The file got regenerated as part of the PR. How should we proceed to make code rabbit understand that this is normal. We have CI checks that would complain if the generated file wasn't valid so these review comments are completely incorrect.

ogenstad (approved):
LGTM. The only thing is the relationship kind highlighted by CodeRabbit. I think it would make sense to address those.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2496,6 +2496,55 @@ def manage_object_template_relationships(self) -> None:
...
+                pool_peer = InfrahubKind.IPPREFIXPOOL
+
+        if not pool_peer:
+            return None
+
+        pool_rel_name = f"{relationship.name}_from_resource_pool"
+        pool_identifier = f"{relationship.get_identifier()}_from_resource_pool"
+        pool_label = relationship.name.title() + " from Resource Pool"
+
+        return RelationshipSchema(
```
ogenstad, line 2535: Can we add a `description` attribut to this one that clearly highlights that it's one of these generated relationships?

ogenstad (approved):

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: good point, set to `GENERIC`

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: added

PR #8319, opened 2026-02-05
feat: IFC-2162 Add namespace restriction parameter to generic schemas

polmichel:
Add namespace restriction support for Generic schemas

### Why

**Problem statement**

The current data modeling of Git repositories in Infrahub allows users to create schemas that inherit from Git repositories. However, this is unexpected and could lead to further issues, as assumptions are made about the state of Git repositories when handling them in code. 

**Goal**
To prevent this from happening, a new attribute has been created for generic schemas (and, by extension, Git repositories) called "_restricted_namespaces_". This attribute can limit the values that namespaces can take on inherited nodes.

In the case of Git repositories schema, we will define the value of this new parameter so that the namespaces can only be "_Core_". The schema of these repositories is part of the core package. This will prevent any extension of the schema.

This parameter is not only meant for this situation, but it can also serve as additional data validation or control for users in their schema hierarchy.

**Non-goals**
The `infrahub db update-core-schema` command was not functioning properly when a modification was made to the core schema due to the non-initialization of the `infrahub.dependencies.component.registry.ComponentDependencyRegistry`.
This was not the case when calling `infrahub db upgrade` command, which contained the initialization.

### What changed

  - Generic schemas: a new attribute available _"restricted_namespaces"_, which is optional and a list of strings.
  - Core generic schema of git repositories now has a "restricted_namespaces" attribute defined to `["Core"]` which prevent any extension.

### How-to review

This PR is quite small: commit by commit, in the chronological order.

### How to test

  **Test plan**

Automated back-end tests: constraint validation logic, within existing or new nodes.
`uv run pytest backend/tests/integration/schema_lifecycle/test_restricted_namespaces_validation.py backend/tests/component/core/schema_manager/test_manager_schema.py -v -k "restricted_namespace or change_restriction"`

Front-end UI change: the new parameter should be visible
image

### Impact & rollout
- **Deployment notes:** this PR must be also merged https://github.com/opsmill/infrahub-sdk-python/pull/807

  Closes IFC-2162, IFC-2163, IFC-2164, IFC-2165, IFC-2166, IHS-190

## Summary by CodeRabbit

* **New Features**
  * Added namespace restrictions for generic schemas and enforcement during schema processing (violations now raise validation errors).

* **Documentation**
  * Updated schema docs and reference guides to describe the new restricted_namespaces parameter and its behavior.

* **Frontend**
  * Display restricted namespaces in the schema viewer UI for generic schemas.

* **Tests**
  * Added unit and integration tests covering restricted_namespaces validation and lifecycle behavior.

* **Changelog**
  * Added entry noting the new namespace restriction parameter.


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1158,6 +1159,24 @@ def validate_names(self) -> None:
                 ):
                     raise ValueError(f"{node.kind}: {rel.name} isn't allowed as a relationship name.")
 
+    def validate_restricted_namespaces_from_generic(self) -> None:
+        """Ensure that every node which inherit from a generic node containing restricted namespaces are following on
+        the rules"""
+        for name in self.nodes:
+            node: MainSchemaTypes = self.get(name=name, duplicate=False)
```
ogenstad: The call to `self.get()` already returns `MainSchemaTypes` so we wouldn't need to define the type with a hint like this. However it could be cleaner to instead call `self.get_node()` which specifically returns a `NodeSchema`

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1158,6 +1159,24 @@ def validate_names(self) -> None:
...
+
+            for generic_name in node.inherit_from:
+                generic_name_node: MainSchemaTypes = self.get(name=generic_name, duplicate=False)
```
ogenstad: Same comment as above but here we can use `self.get_generic()`.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1158,6 +1159,24 @@ def validate_names(self) -> None:
...
+                if (
+                    generic_name_node.restricted_namespaces is not None
+                    and node.namespace not in generic_name_node.restricted_namespaces
+                ):
+                    raise ValueError(
+                        f"Generic node '{generic_name}' on branch '{self.name}' has restricted namespaces: "
```
ogenstad: I'd avoid the bart with the branch name here. `self.name` is only optionally a string and I'm not sure if we always set a value for it which could cause weird error messages such as "on branch 'None'".

backend/infrahub/cli/db.py
```diff
@@ -443,6 +444,7 @@ async def update_core_schema(db: InfrahubDatabase, initialize: bool = True, debu
     if initialize:
         await initialize_registry(db=db)
         await initialize_internal_schema()
+        build_component_registry()
```
ogenstad, line 447: Do we need to add this?

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4567,3 +4568,92 @@ async def test_identify_object_templates_with_generics() -> None:
         TestKind.SFP,
         TestKind.VIRTUAL_INTERFACE,
     }
+
+
+async def test_generic_schema_with_restricted_namespace_fails_if_required(
+    incorrect_schema_generic_with_namespace_restriction,
+):
```
ogenstad: It might not be enforced everywhere yet but we should have the correct typehints when adding new tests i.e. a type for incorrect_schema_generic_with_namespace_restriction (even if it's `None`), this is mainly important when we actually use something from the fixtures within the test but good to have in place for when it is enforced. Also we should indicate that the test doesn't return anything. This goes for the other tests as well.

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4567,3 +4568,92 @@ async def test_identify_object_templates_with_generics() -> None:
...
+
+    schema: SchemaBranch = SchemaBranch(cache={}, name="test")
+    schema.load_schema(schema=test_schema)
+
+    # Act
+    with pytest.raises(ValueError) as error:
+        schema.validate_restricted_namespaces_from_generic()
+
+    # Assert
+    assert error.type is ValueError
```
ogenstad: I don't think this line does anything as the `with pytest.raises(ValueError) as error` should raise an error if a `ValueError` wasn't raised.

backend/tests/integration/schema_lifecycle/core-schema/update/test_update_restricted_namespaces.py
```diff
@@ -0,0 +1,126 @@
...
+            schema=schema_branch,
+            db=db,
+            branch=default_branch,
+        )
+
+    def _create_old_generic_schema_node(self) -> NodeSchema:
+        """Create SchemaGeneric node WITHOUT restricted_namespaces attribute."""
+        old_generic: NodeSchema = internal_generic_schema.model_copy(deep=True)
+        old_generic.attributes = [attr for attr in old_generic.attributes if attr.name != "restricted_namespaces"]
+        return old_generic
```
ogenstad: I think this test is a good lesson just to learn of the various components of Infrahub, but I'm not sure we actually need the test as the internal schema upgrade should already be tested elsewhere.

ogenstad (commented):
Added some comments but looks good to be overall.

backend/infrahub/cli/db.py
polmichel: When I was testing the upgrade_core_schema command `infrahub db update-core-schema`, I did not tweek it with a manual registry initialization as I did in the following test: _backend/tests/integration/schema_lifecycle/core-schema/update/test_update_restricted_namespaces.py:39_

I encountered an **infrahub.dependencies.component.exceptions.UntrackedDependencyError**, which could not be avoided due to the missing initialization of the registry. I assumed that if we wanted to call this method with _initialize_ set to true, that we would also want to correctly initialize the registry. As I understood it, this error occurred because I was adding a new node to the core schema.

I checked the different calls to _infrahub.cli.db.update_core_schema_, I think this would only impact this command.

This line isn't necessary for this feature. I can remove it if you think it would be better.

backend/tests/integration/schema_lifecycle/core-schema/update/test_update_restricted_namespaces.py
polmichel: You are right. I was thinking of this test as the basis for a global parameterized test, I will remove it and store it locally instead.

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4567,3 +4568,87 @@ async def test_identify_object_templates_with_generics() -> None:
         TestKind.SFP,
         TestKind.VIRTUAL_INTERFACE,
     }
+
+
+async def test_generic_schema_with_restricted_namespace_fails_if_required(
+    incorrect_schema_generic_with_namespace_restriction,
+) -> None:
+    # Arrange
+    test_schema: SchemaRoot = copy.deepcopy(incorrect_schema_generic_with_namespace_restriction)
```
ajtmccarty: I wouldn't expect the `deepcopy` to be necessary here
I think you could also get away with reusing the same SchemaRoot across these tests and just updating it in place before passing it to `load_schema()`

for example
```
async def test_generic_schema_with_restricted_namespace_fails_if_required(
    incorrect_schema_generic_with_namespace_restriction: SchemaRoot,
) -> None:
```
    incorrect_schema_generic_with_namespace_restriction.generics[0].restricted_namespaces = ["Something"]
    incorrect_schema_generic_with_namespace_restriction.nodes[0].namespace = ["Something"]
```

Comment (ajtmccarty) [inline backend/tests/component/core/schema_manager/test_manager_schema.py]: I wouldn't expect the `deepcopy` to be necessary here
I think you could also get away with reusing the same SchemaRoot across these tests and just updating it in place before passing it to `load_schema()`

for example
```
async def test_generic_schema_with_restricted_namespace_fails_if_required(
    incorrect_schema_generic_with_namespace_restriction: SchemaRoot,
) -> None:
```
    incorrect_schema_generic_with_namespace_restriction.generics[0].restricted_namespaces = ["Something"]
    incorrect_schema_generic_with_namespace_restriction.nodes[0].namespace = ["Something"]
```

ajtmccarty: I wouldn't expect the `deepcopy` to be necessary here
I think you could also get away with reusing the same SchemaRoot across these tests and just updating it in place before passing it to `load_schema()`

for example
```
async def test_generic_schema_with_restricted_namespace_fails_if_required(
    incorrect_schema_generic_with_namespace_restriction: SchemaRoot,
) -> None:
```
    incorrect_schema_generic_with_namespace_restriction.generics[0].restricted_namespaces = ["Something"]
    incorrect_schema_generic_with_namespace_restriction.nodes[0].namespace = ["Something"]
```

**Comment by ogenstad** [comment]
I think we'll need some additional work with the schema integrity checks for this one. The overall feature looks good however we have a situation with merge operations when if you:

1. Load a schema such as this one:

```yaml
version: '1.0'

generics:
  - name: Problem
    namespace: Base
    description: Generic Location, could be a country, city ...
    label: Base Problem
    icon: mingcute:location-line
    include_in_menu: false
    #restricted_namespaces:
    #  - Support
    order_by:
      - name__value
    display_labels:
      - name__value
    attributes:
      - name: name
        kind: Text
        order_weight: 1000
      - name: description
        kind: Text
        optional: true
        order_weight: 1200
      - name: problem_number
        kind: NumberPool
        description: "Give me the number!"
        unique: true
        optional: false
        read_only: true

nodes:
  - name: Problem
    namespace: Support
    label: "Problem"
    display_labels:
      - name__value
    generate_profile: true
    include_in_menu: true
    attributes:
      - name: name
        kind: Text
        order_weight: 1000
    inherit_from:
      - BaseProblem
  - name: Incident
    namespace: Support
    label: "Incident"
    display_labels:
      - name__value
    generate_profile: true
    include_in_menu: true
    inherit_from:
      - BaseProblem
    attributes:
      - name: name
        kind: Text
        order_weight: 1000
```

2. Create a branch: branch1
3. On the main branch add "Support" in `restricted_namespaces` on the generic
4. On branch1: Add a copy of one of the nodes but change the namespace to something else like "Baseline".
5. Create a proposed change, this should be a schema violation error but it's not actually picked up by the Schema Integrity job in the pipeline
6. Merge the branch, at this point we have an invalid schema within the main branch and reloading the application will fail.

bilalabbad (approved):
frontend LGTM

ajtmccarty:
I'd say the cross-branch schema integrity issue that Patrick identified is a separate issue because there are a lot of other cases where we can end up with a broken schema after a merge

a couple untested examples
- you update a Relationship's peer on a branch and delete the peer schema on the default branch
- update the HFID to include the color attribute on a branch and remove the color attribute on the default branch
etc

backend/infrahub/cli/db.py
polmichel: A internal issue has been created related to the `infrahub db update-core-schema` command removal https://opsmill.atlassian.net/browse/IFC-2254

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1160,6 +1161,24 @@ def validate_names(self) -> None:
...
+            for generic_name in node.inherit_from:
+                generic_name_node = self.get_generic(name=generic_name, duplicate=False)
+                if (
+                    generic_name_node.restricted_namespaces is not None
+                    and node.namespace not in generic_name_node.restricted_namespaces
+                ):
+                    raise ValueError(
+                        f"Generic node '{generic_name}' has restricted namespaces: "
+                        f"{generic_name_node.restricted_namespaces}. The node '{name}' does not comply "
+                        f"with this restriction as its namespace is '{node.namespace}'."
```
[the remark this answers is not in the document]
polmichel: **1. A generic does not have the _inherit_from_ in its schema so a Generic cannot inherit_from another Generic:**

 - GenericSchema → GeneratedGenericSchema → BaseNodeSchema → GeneratedBaseNodeSchema : no inherit_from
  - NodeSchema → GeneratedNodeSchema → BaseNodeSchema → GeneratedBaseNodeSchema:  has inherit_from
  
**2. A node cannot inherit from a non-generic node**

 Line 1171: self.get_generic(name=generic_name, ...). The validation assumes every entry in inherit_from is a generic and calls get_generic on it. So no, inherit_from 
  is specifically for inheriting from generics
  
So, I do not think this can happen

backend/infrahub/core/schema/generated/genericnode_schema.py
```diff
@@ -23,3 +23,8 @@ class GeneratedGenericSchema(BaseNodeSchema):
         description="List of Nodes that are referencing this Generic",
         json_schema_extra={"update": "not_applicable"},
     )
+    restricted_namespaces: list[str] | None = Field(
+        default=None,
+        description="Namespaces that Nodes inheriting from this Generic are restricted to",
```
ajtmccarty: maybe "Nodes inheriting from this Generic schema must belong to one of the listed namespaces"?

backend/infrahub/core/schema/definitions/internal.py
```diff
@@ -959,6 +959,14 @@ def to_dict(self) -> dict[str, Any]:
             optional=True,
             extra={"update": UpdateSupport.NOT_APPLICABLE},
         ),
+        SchemaAttribute(
+            name="restricted_namespaces",
+            kind="List",
+            internal_kind=str,
+            description="Namespaces that Nodes inheriting from this Generic are restricted to",
+            optional=True,
+            extra={"update": UpdateSupport.ALLOWED},
```
ajtmccarty, line 968: should this be `UpdateSupport.VALIDATE_CONSTRAINT`?
for example, if I have Generic schema `PrimaryA` and Node schema `SecondaryB` inherits from it and then I add `["Primary"]` as the `PrimaryA.restricted_namespaces`, would that break the schema?

UPDATE: this is fine as is. the schema update in the example above would be prevented by validation during the update

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4662,3 +4663,123 @@ async def test_manage_object_templates_component_relationship_to_excluded_kind()
...
+) -> None:
+    # Retrieve the schema and make it incorrect
+    test_schema: SchemaRoot = correct_schema_generic_with_namespace_restriction
+    test_schema.nodes[0].namespace = "NotACorrectNamespace"
+
+    schema: SchemaBranch = SchemaBranch(cache={}, name="test")
+    schema.load_schema(schema=test_schema)
+
+    # Act
+    with pytest.raises(ValueError) as error:
```
ajtmccarty: not required, but you should be able to do `pytest.raises(ValueError, match="does not comply with this restriction as its namespace")` and then you could skip the final assertion

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4662,3 +4663,123 @@ async def test_manage_object_templates_component_relationship_to_excluded_kind()
...
+
+    # Act
+    try:
+        schema.validate_restricted_namespaces_from_generic()
+
+    # Assert - no ValueError
+    except ValueError:
+        pytest.fail(
+            "A ValueError has been raised during process_restricted_namespaces check, however, the schema is correct"
+        )
```
ajtmccarty: do you still get the traceback in the pytest output with this structure?

backend/tests/component/core/schema_manager/test_manager_schema.py
ajtmccarty, line 1: this test file is way too big, which is not your fault
could you move the new tests to a new file on their own so this one doesn't get any bigger?

backend/tests/integration/schema_lifecycle/namespace_restriction/conftest.py
```diff
@@ -0,0 +1,68 @@
...
+        "inherit_from": ["TestingAnimal"],
+        "attributes": [
+            {"name": "breed", "kind": "Text", "optional": True},
+        ],
+    }
+
+
+@pytest.fixture(scope="class")
+def schema_cat_node() -> dict[str, Any]:
+    """Cat node in Cat namespace, inheriting from Animal generic (violates Dog restriction)."""
```
ajtmccarty: this doesn't depend on the schema, so I think the description is not quite accurate

backend/tests/integration/schema_lifecycle/namespace_restriction/test_restricted_namespaces_cross_branch.py
ajtmccarty, line 1: is this really an integration test? it looks like it just uses `SchemaBranch` and doesn't require the API or workers to be running. if it were using the client to load the schema, then it would be an integration or functional test, but I don't think that's really necessary. we have tests verifying that the schema is correctly validated when loaded on a branch. I think this is either a component test or we don't need it

backend/tests/component/core/schema_manager/namespace_restriction/test_restricted_namespaces_validation.py
```diff
@@ -0,0 +1,92 @@
...
+    from infrahub.core.schema.schema_branch import SchemaBranch
+from infrahub.database import InfrahubDatabase
+from tests.integration.schema_lifecycle.shared import TestSchemaLifecycleBase
+from tests.integration.shared import load_schema
+
+
+class TestRestrictedNamespacesValidation(TestSchemaLifecycleBase):
+    """Test that modifying restricted_namespaces is validated against existing nodes."""
+
+    async def test_change_restriction_should_fail(
```
ajtmccarty, line 25: this looks like it does basically the same thing as the other new test file. this should be a component test if it is required

backend/infrahub/core/schema/generated/genericnode_schema.py
polmichel: Yes this is better, thanks

backend/tests/component/core/schema_manager/test_manager_schema.py
polmichel: This was intended to be consistent with the other unit tests, but I guess this can be disturbing since it asserts only one element. Will do!

backend/tests/component/core/schema_manager/test_manager_schema.py
polmichel: I am not sure to understand what is expected there but I got this when I force the test to fail
```
correct_schema_generic_with_namespace_restriction = SchemaRoot(version=None, generics=[GenericSchema(id=None, state=<HashableModelState.PRESENT: 'present'>, name='Generic...nt=None, children=None)], extensions=SchemaExtension(id=None, state=<HashableModelState.PRESENT: 'present'>, nodes=[]))

    async def test_generic_schema_with_restricted_namespace_pass_if_same_namespace(
        self,
        correct_schema_generic_with_namespace_restriction,
    ) -> None:
        schema: SchemaBranch = SchemaBranch(cache={}, name="test")
        schema.load_schema(schema=correct_schema_generic_with_namespace_restriction)
    
        # Act
        try:
            schema.validate_restricted_namespaces_from_generic()
    
        # Assert - no ValueError
        except ValueError:
>           pytest.fail(
                "A ValueError has been raised during process_restricted_namespaces check, however, the schema is correct"
            )
E           Failed: A ValueError has been raised during process_restricted_namespaces check, however, the schema is correct
```

backend/tests/integration/schema_lifecycle/namespace_restriction/test_restricted_namespaces_cross_branch.py
polmichel: This was to illustrate the precedent issue regarding schema validation which was not taking in account the schema once merged and was only triggered on schema on both branches. I think we can remove it since a test already exists there https://github.com/opsmill/infrahub/pull/8428/changes.

And this should be a component test yes, my bad

backend/tests/component/core/schema_manager/namespace_restriction/test_restricted_namespaces_validation.py
polmichel: The other one was to verify that, given one correct schema on branch `A` and another correct schema on `main`, which would be invalid once merged together, the schema validation engine would detect this.
This one is just testing to load an invalid schema on a branch, without any merge process. I will move it into a dedicated component tests folder

backend/tests/component/core/schema_manager/test_manager_schema.py
polmichel: I removed this assertion and have let the test fail by itself if an error occurs

ajtmccarty (approved):

PR #8330, opened 2026-02-06
IFC-1305 Clean up permission `identifier` custom GraphQL code

gmazoyer:
This change cleans up the custom code we used to have at the graphql to build identifiers. We do not need it anymore as we use the same values as display labels (brought by #8039).

## Summary by CodeRabbit

* **New Features**
  * Permissions now include a required "Label" (display_label) shown throughout role management (global/object permissions, roles, comboboxes, tables).

* **Refactor**
  * Simplified internal permission handling to rely on the new display label field.

* **UI**
  * Replaced "Identifier" with "Label" in permission lists, columns, badges, and selection components.

* **Tests**
  * Updated permission tests to assert and sync display_label with permission objects.


backend/infrahub/core/schema/definitions/core/permission.py
```diff
@@ -27,15 +26,6 @@
...
         Attr(name="description", kind="Text", optional=True),
-        Attr(
-            name="identifier",
-            kind="Text",
-            description="Identifier for the permission",
-            read_only=True,
-            optional=True,
-            allow_override=AllowOverrideType.NONE,
-            deprecation="Use permission display_label instead",
-        ),
```
ogenstad, line 38: Do we need a migration or state=absent here for this to be removed from existing installations?

gmazoyer: We sure do we have the `backend/infrahub/core/schema/definitions/deprecated.py` file for that it seems.

ajtmccarty (approved):

PR #8343, opened 2026-02-09
add template support to schema migrations

ajtmccarty:
# Why

Template instances were not being correctly updated when the schema of the associated node/generic was updated to add/remove attributes or update attributes such that they needed to be added to or removed from existing Template instances. For example, updating `TestCar.color` from `read_only=True` to `False` would mean that any existing `TemplateTestCar` instances should now include the `color` attribute.

- Added Template support to attribute add, remove, and rename schema migrations
- Added support for adding/removing attribute to/from Templates when their properties change in a manner that affects whether they should be present on existing Templates or not. This logic already existed for Profiles, so the new Template-centered logic follows the pattern already set for Profiles.

## Checklist

- [x] Tests added/updated
- [ ] ~External docs updated (if user-facing or ops-facing change)~
- [ ] ~Internal .md docs updated (internal knowledge and AI code tools knowledge)~

## Summary by CodeRabbit

* **New Features**
  * Full support for updating Template instances via schema migrations when attributes are added/removed.
  * Attributes can now opt into Templates via a new support_templates setting.

* **Bug Fixes**
  * Template attribute generation now filters attributes by template support and treats templates in migration updates alongside profiles.

* **Tests**
  * Added comprehensive component and integration tests for template attribute lifecycle and migration flows; removed legacy profile-only tests.

* **Changelog**
  * Documented template migration support.


backend/infrahub/core/migrations/query/attribute_remove.py
```diff
@@ -33,20 +34,27 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
-        if isinstance(new_schema, GenericSchema):
-            for inheriting_schema_kind in new_schema.used_by:
-                node_schema = db.schema.get_node_schema(
-                    name=inheriting_schema_kind, branch=self.branch, duplicate=False
-                )
-                attr_schema = node_schema.get_attribute_or_none(name=self.attribute_name)
-                if attr_schema and not attr_schema.inherited:
-                    kinds_to_ignore.append(inheriting_schema_kind)
-                else:
-                    profile_kinds_to_update.append(f"Profile{inheriting_schema_kind}")
```
ajtmccarty, line 49: this whole block was not indented at the correct level, which would have prevented removing attributes from multiple profiles correctly

backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
ajtmccarty, line 1: this file replaces `attribute_supports_profile.py`. it adds support for templates to this migration in the same way that profiles were already supported. the name change is just because this migration now covers both Templates and Profiles

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -0,0 +1,320 @@
...
+    RETURN r2
+    ORDER BY r2.branch_level DESC, r2.from DESC
+    LIMIT 1
+}
+RETURN node.uuid AS node_id, r1.status AS has_attr_status, r2.status AS has_val_status
+"""
+
+
+async def assert_attribute_path_status(
+    db: InfrahubDatabase,
```
ajtmccarty, line 43: we do this verification with manual cypher b/c we do not have a good way to check if individual paths are added or deleted separate from changes to the schema when loading data using `NodeManager`
for example, if I delete the `color` attribute from a schema, we will never load that attribute even if the attribute path is not actually deleted at the database level. this cypher check confirms that these paths are added or deleted as expected

backend/tests/component/core/migrations/schema/test_attribute_supports_profile.py
ajtmccarty, line 1: moved into test_attribute_supports_generated_schema.py

backend/tests/integration/templates/test_template_attribute_lifecycle.py
ajtmccarty, line 1: new test to cover adding, removing, and updating properties of attributes such that they are added to or removed from existing Template instances

ogenstad (approved):
LGTM

PR #8344, opened 2026-02-09
fix: use compact json encoding for webhook payload signature calculation

wvandeun:
# Why

httpx dependency was updated from version 0.27.2 to 0.28.1 in Infrahub v1.7.0.
httpx seems to use compact json encoding since 0.28.0, which caused Infrahub to send an invalid signature for webhooks, breaking webhook authentication.

Goal: use compact json encoding for the webhook payload so that we calculate and set a correct signature for webhooks.

## Checklist

- [-] Tests added/updated
- [X] [Changelog entry](../dev/guidelines/changelog.md) added (`uv run towncrier create ...`)
- [-] External docs updated (if user-facing or ops-facing change)
- [-] Internal .md docs updated (internal knowledge and AI code tools knowledge)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed webhook authentication signature calculation to correctly handle compact JSON payload encoding in signature generation.

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -135,7 +135,7 @@ class Webhook(BaseModel):
         if self.shared_key:
             message_id = f"msg_{uuid.hex}" if uuid else f"msg_{uuid4().hex}"
             timestamp = str(at.to_timestamp()) if at else str(Timestamp().to_timestamp())
-            payload = json.dumps(self._payload or {})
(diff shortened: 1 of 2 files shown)


ajtmccarty (approved):

ogenstad:
I think this is fine as a quickfix but as we've already fixed this once before I think it's clear that we need a test that would catch a regression. If this has been verified I think it's fine to merge as is but we should create a follow up ticket and perhaps tie it to the housekeeping milestone for 1.9.

PR #8345, opened 2026-02-10
IFC-2251 Add resource pool allocation support for templates

gmazoyer:
# Why

This change allow to allocate relationships from pools (such as IP/Prefix pools) when creating a node from a template that references those pools.

This is also an attempt at extracting some logic out of our big `Node` class.

## What changed

**Behavioral changes:**
- Templates can now have `_from_resource_pool` relationships (e.g., `primary_ip_from_resource_pool`)
- Creating a template with a pool reference stores the reference but does NOT allocate
- Creating an object from that template allocates from the pool and sets the relationship
- User-provided values override pool allocations

**Implementation:**
- Added `PoolAllocator` ABC and `DefaultPoolAllocator` that allocates from NumberPools and IP/Prefix pools
- Added `allocate_from_resource_pools()` in `create.py` to handle template pool references
- Added `RESOURCE_POOL_REL_SUFFIX` constant to `constants/schema.py`
- Added `TemplateResourcePoolExclusiveConstraint` to avoid templates having both a resource relationship and a resource pool relationship set for the same relationship

**What stayed the same:**
- NumberPool allocation for regular (non-template) nodes unchanged
- Existing template behavior for attributes and non-pool relationships unchanged
- No schema changes required

**Limitation:**
- NumberPool allocation from templates is NOT supported yet

## How to review

Focus on:
1. `backend/infrahub/core/node/create.py` - `allocate_from_resource_pools()` function
2. `backend/infrahub/pools/` - PoolAllocator ABC and implementations
3. `backend/infrahub/templates/` - NodeTemplateApplier
4. `backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py` - TemplateResourcePoolExclusiveConstraint

Added `primary_ip` relationship to `tests/helpers/schema/device.py` which required some changes elsewhere.

## How to test

```bash
# Run template pool allocation tests

# Verify existing pool tests still pass
uv run pytest backend/tests/unit/core/ipam/ -v -k pool

# Verify template pool functional tests
```

## Impact & rollout

- **Backward compatibility**: Fully backward compatible. Existing templates without pool references should work unchanged.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Templates can allocate resources from configured resource pools when creating objects.
  * Template application now merges template-provided values with user-provided fields, preserving explicit overrides.

* **Validation**
  * Templates now enforce mutual exclusivity between direct resource assignment and pool-based allocation.

* **Tests**
  * Added extensive tests covering template pool allocation, precedence rules, exclusivity checks, and related behaviors.


backend/infrahub/core/node/__init__.py
```diff
@@ -350,12 +351,20 @@ async def handle_pool(
...
+        For TemplateSchema objects, we only store the pool reference without allocating resources.
+        The actual allocation happens when creating objects from the template.
         """
 
         if attribute.schema.kind == "NumberPool" and isinstance(attribute.schema.parameters, NumberPoolParameters):
             attribute.from_pool = {"id": attribute.schema.parameters.number_pool_id}
             attribute.is_default = False
 
+        # Templates should not allocate from pools - just store the reference
+        # Actual allocation happens when creating objects from the template
```
ajtmccarty: probably only need one of these comments

backend/infrahub/core/node/create.py
```diff
@@ -107,6 +109,39 @@ async def extract_peer_data(
...
+    for rel_schema in template_schema.relationships:
+        if not rel_schema.name.endswith(RESOURCE_POOL_REL_SUFFIX):
+            continue
+
+        original_rel_name = rel_schema.name.removesuffix(RESOURCE_POOL_REL_SUFFIX)
+
+        if original_rel_name not in obj_schema.relationship_names:
+            continue
+
+        pool_rel_manager: RelationshipManager = getattr(template, rel_schema.name)
```
ajtmccarty: can you do `template.get_relationship(rel_schema.name)`  here or does that not work with the `CoreObjectTemplate` type-hint in the signature?

backend/infrahub/core/node/create.py
```diff
@@ -147,6 +182,12 @@ async def handle_template_relationships(
             await constraint_runner.check(node=obj_peer, field_filters=list(obj_peer_data))
             await obj_peer.save(db=db)
```
ajtmccarty: could this `save()` move below `allocate_from_resource_pools` or does it need to run before the new function?

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2530,8 +2530,8 @@ def _create_resource_pool_relationship(self, relationship: RelationshipSchema) -
         if not pool_peer:
             return None
 
-        pool_rel_name = f"{relationship.name}_from_resource_pool"
-        pool_identifier = f"{relationship.get_identifier()}_from_resource_pool"
+        pool_rel_name = f"{relationship.name}{RESOURCE_POOL_REL_SUFFIX}"
+        pool_identifier = f"{relationship.get_identifier()}{RESOURCE_POOL_REL_SUFFIX}"
```
ajtmccarty, line 2534: 👍

backend/infrahub/pools/default_allocator.py
```diff
@@ -0,0 +1,64 @@
...
+        """Allocate from pool if attribute references one."""
+        if not attribute.from_pool:
+            return None
+
+        pool_id = attribute.from_pool.get("id")
+        if not pool_id:
+            return None
+
+        try:
+            number_pool = await registry.manager.get_one_by_id_or_default_filter(
```
ajtmccarty: can use `get_one` instead

backend/infrahub/pools/allocator.py
```diff
@@ -0,0 +1,21 @@
...
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from infrahub.core.attribute import BaseAttribute
+    from infrahub.core.node import Node
+    from infrahub.core.relationship.model import RelationshipManager
+
+
+class PoolAllocator(ABC):
```
ajtmccarty, line 12: are there other places that we use resource pools where this interface could be used in the future? My first thought was that there would be a `DefaultPoolAllocator` and a `TemplatePoolAllocator`, but it looks like the same logic is used in both cases. so this interface might not be necessary right now, but it's pretty slim, so do what brings you joy

backend/infrahub/core/node/create.py
```diff
@@ -107,6 +109,39 @@ async def extract_peer_data(
     return obj_peer_data
 
 
+async def allocate_from_resource_pools(
```
ajtmccarty, line 112: it feels like this function overlaps with some of the logic in DefaultPoolAllocator/NodeTemplateApplier. I'm not sure if there is a good/reasonable way to get them together

ajtmccarty (commented):
better and better
love the new components

backend/infrahub/pools/allocator.py
gmazoyer: You are actually right. I introduced a `NoOpPoolAllocator` which would I guess corresponds to your expected `TemplatePoolAllocator` in order to follow a null object pattern. But it was only used by a `NodeFactory` I wanted to introduce too. Although I thought it could be a bit too much for this PR.

backend/infrahub/core/node/create.py
gmazoyer: It is a bit tricky to unify the way it works for now.

`allocate_from_resource_pools` is used for nested template components (children created after parent), iterates over pool relationships, allocates, and updates the object directly. `NodeTemplateApplier` combined with `DefaultPoolAllocator` is used for top-level node field preparation (before creation).

We might find a way to combine all of this but not sure how hard it's going to be yet.

ajtmccarty: I think that's fine for now. this is definitely an improvement. I couldn't quite understand where each one fit into the Node creation logic, so it makes sense that it's not easy to completely de-duplicate

backend/infrahub/core/node/create.py
gmazoyer: CoreObjectTemplate doesn't have get_relationship in its protocol, we don't have it at the CoreNode level in fact. Could be we should update our protocols at some point because it's a bit annoying to work with them sometimes.

ogenstad: If we're just talking about adding `def get_relationship(self, name: str) -> RelationshipManager:` to CoreNode in `protocols_base.py` you can go ahead and just add it.

backend/infrahub/core/node/__init__.py
```diff
@@ -492,36 +498,17 @@ async def handle_object_template(self, fields: dict, db: InfrahubDatabase, error
...
-            relationship_schema = template._schema.get_relationship(name=relationship_name)
-            if (
-                relationship_name in list(fields)
-                or relationship_schema.kind
-                not in [RelationshipKind.ATTRIBUTE, RelationshipKind.GENERIC, RelationshipKind.PROFILE]
-                or relationship_name == OBJECT_TEMPLATE_RELATIONSHIP_NAME
-            ):
-                continue
+        pool_allocator = DefaultPoolAllocator(db=db, branch=self._branch)
+        applier = NodeTemplateApplier(db=db, branch=self._branch, pool_allocator=pool_allocator)
```
ajtmccarty, line 502: you could register these in the `build_component_registry()` and then use `get_component_registry()` and `ComponentRegistry.get_component()` to retrieve them so that the logic for building the component is encapsulated elsewhere

backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
```diff
@@ -0,0 +1,70 @@
...
+                await self._check_counterpart_not_set(node=node, counterpart_name=pool_rel_name, current_name=rel_name)
+
+    async def _check_counterpart_not_set(self, node: Node, counterpart_name: str, current_name: str) -> None:
+        """Check that the counterpart relationship is not set."""
+        try:
+            counterpart_relm = node.get_relationship(name=counterpart_name)
+        except ValueError:
+            return
+
+        update_details = await counterpart_relm.fetch_relationship_ids(db=self.db, force_refresh=True)
```
ajtmccarty, line 54: I guess this is necessary b/c `NodeConstraintRunner` only calls `fetch_relationship_ids` on the relationship it is checking and not the `conterpart_relm`
I guess `Node.resolve_relationships()`, which `NodeConstraintRunner` calls before running any relationship constraints, doesn't actually handle getting the peers of a relationship from the db, which is kind of confusing

backend/infrahub/templates/node_applier.py
```diff
@@ -0,0 +1,104 @@
...
+        await self._apply_attributes(template=template, fields=fields)
+        await self._apply_relationships(template=template, target_id=target_id, fields=fields)
+        return fields
+
+    async def _apply_attributes(self, template: CoreObjectTemplate, fields: dict[str, Any]) -> None:
+        for attr_name in template.get_schema().attribute_names:
+            if attr_name in fields or attr_name == OBJECT_TEMPLATE_NAME_ATTR:
+                continue
+
+            attr = getattr(template, attr_name)
```
ajtmccarty: template.get_attribute()

backend/infrahub/templates/node_applier.py
```diff
@@ -0,0 +1,104 @@
...
+            if rel_name in fields:
+                continue
+
+            if rel_schema.kind not in [RelationshipKind.ATTRIBUTE, RelationshipKind.GENERIC, RelationshipKind.PROFILE]:
+                continue
+
+            if rel_name == OBJECT_TEMPLATE_RELATIONSHIP_NAME:
+                continue
+
+            relationship: RelationshipManager = getattr(template, rel_name)
```
ajtmccarty: template.get_relationship()

backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
ajtmccarty, line 1: not sure how fast or slow this test file runs, but I think you could make this a class that loads the schema and fixtures, then run each test case very quickly w/o reloading all the setup data between each test

backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
```diff
@@ -0,0 +1,205 @@
...
+    await template.new(db=db, template_name="device-template-replace", primary_ip_from_resource_pool=ip_address_pool)
+    await template.save(db=db)
+
+    loaded_template = await registry.manager.get_one(db=db, id=template.id, branch=default_branch)
+    await loaded_template.primary_ip_from_resource_pool.update(db=db, data=None)
+    await loaded_template.primary_ip.update(db=db, data={"id": ip_address.id})
+
+    await constraint.check(
+        relm=loaded_template.primary_ip, node_schema=loaded_template.get_schema(), node=loaded_template
+    )
```
ajtmccarty: maybe one more test for when the template instance is already saved to the database with a `ip_address_from_resource_pool` relationship and the incoming data includes just the `ip_address`

backend/tests/component/core/node/test_template_pool_allocation.py
```diff
@@ -0,0 +1,179 @@
...
+        branch=default_branch,
+        schema=node_schema,
+    )
+
+    assert device.id is not None
+
+    primary_ip = await device.primary_ip.get_peer(db=db)
+    assert primary_ip is not None
+    assert primary_ip.id == explicit_address.id
+    assert primary_ip.address.value == "10.10.3.25/27"
```
ajtmccarty, line 180: maybe one more where the template instance has an `ip_address` relationship to check it is correctly set on the created device

backend/tests/component/templates/test_template_applier.py
```diff
@@ -0,0 +1,460 @@
...
+    value: Any
+    source_id: str
+
+
+@dataclass
+class ExpectedTemplateRelationship:
+    """Expected relationship peers from template application."""
+
+    name: str
+    peer_ids: list[str]
```
ajtmccarty, line 51: I think it might be worth also testing the `source` of the Relationship

backend/tests/component/templates/test_template_applier.py
```diff
@@ -0,0 +1,460 @@
...
+        user_fields = {"name": "my-device", "weight": 100, "airflow": "Front to rear"}
+
+        fields = await applier.apply(
+            template=template, target_schema=target_schema, target_id="new-device-id", user_fields=user_fields
+        )
+
+        _validate_template_fields(fields=fields, user_fields=user_fields)
+        assert "primary_ip" in fields
+        assert fields["primary_ip"]["_relation__source"] == ip_pool.id
+        assert fields["primary_ip"]["peer"] is not None
```
ajtmccarty: could this be `assert fields["primary_ip"]["peer"] == ip_pool.id`?

ajtmccarty (approved):
looks good
just a few small comments, mostly test-related

backend/infrahub/core/node/__init__.py
gmazoyer: Indeed, I tried it actually, but I ended up in a circular import thing :( like that `core/node -> dependencies/registry -> builders -> core/constraint -> core/branch -> core/node`. Hence the "manual" build for now

backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
gmazoyer: 💯

PR #8351, opened 2026-02-11
test for adding/removing _from_resource_pool rels on Templates

ajtmccarty:
IFC-2249

TODO:
- [x] test is failing until #8350 is merged and moved up into `develop`

turns out that we don't need to make any changes to existing schema migrations to correctly handle adding/removing the `xyz_from_resource_pool` Relationships on Templates correctly because TemplateSchemas are generated and not saved to the database

this PR adds an integration test that confirms Template schemas and instances are updated as expected
which is to say that if a Node or GenericSchema with `generate_template=True` has a relationship (let's say `xyz`) with a peer of BuiltinIPAddress, BuiltinIPPrefix, or a NodeSchema that inherits from either of those, then the generated TemplateSchema must include a `xyz_from_resource_pool` Relationship that links to either CoreIPAddressPool or CoreIPPrefixPool, depending on the peer of the `xyz` Relationship

one implementation detail to note: we don't actually delete the Relationships when we remove a RelationshipSchema from a Node/Generic/TemplateSchema, but this is nothing new.

## Release Notes

* **Tests**
  * Improved test coverage for template endpoint type validation to verify correct schema type responses.
  * Added comprehensive integration tests validating template resource pool lifecycle and inheritance behavior across different schema types.

### DIFF
--- a/backend/tests/integration/templates/test_schema_template_generic_inheritance.py
+++ b/backend/tests/integration/templates/test_schema_template_generic_inheritance.py
@@ -2,6 +2,7 @@ from typing import Any
 
 import pytest
 from infrahub_sdk import InfrahubClient
+from infrahub_sdk.schema.main import GenericSchemaAPI, TemplateSchemaAPI
 
 from infrahub.core.constants import OBJECT_TEMPLATE_RELATIONSHIP_NAME
 from infrahub.core.schema import AttributeSchema, GenericSchema, NodeSchema, SchemaRoot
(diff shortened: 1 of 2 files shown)


gmazoyer (approved):
Just a comment a top of my head, not for this PR. We have more and more specific schemas for our testing needs, probablu even more since our tests are now often LLM-generated (at least to save time writing the initial versions of them). I wonder if we should also try to consolidate them somehow, maybe assisted by a LLM too.

ogenstad:
> Just a comment a top of my head, not for this PR. We have more and more specific schemas for our testing needs, probablu even more since our tests are now often LLM-generated (at least to save time writing the initial versions of them). I wonder if we should also try to consolidate them somehow, maybe assisted by a LLM too.

I agree with this and we have some already in https://github.com/opsmill/infrahub/tree/stable/backend/tests/helpers/schema a benefit of those is that we also get the protocols for test schemas: https://github.com/opsmill/infrahub/blob/stable/backend/tests/protocols.py Which is an area that today causes a number of typing violations throughout the tests.

PR #8362, opened 2026-02-12
IFC-2156: Remove proposed change comments created_at and created_by attribute and relationship

solababs:
# Why

CoreComment and CoreThread explicitly define created_by (relationship) and created_at (attribute) in their schemas, but these are redundant — every node already has created_at and created_by as node metadata. This adds unnecessary schema complexity and graph relationship overhead.

This PR removes these redundant fields so that comments and threads rely on the built-in node metadata

Closes IFC-2156

## What changed

- Schema definitions: Removed created_at attribute and created_by relationship from CoreComment and CoreThread generic schemas, along with the order_by=["created_at__value"] that depended on the removed attribute

- Protocol definitions: Removed created_at and created_by from CoreComment and CoreThread in both protocols.py and tests/protocols.py

- Graph migration (m060): Backfills created_by UUID from the old comment__account and thread__account relationships directly onto the node property, preserving existing data

- Test fix: Removed created_at input from the CoreChangeThreadCreate mutation test since it's no longer an input field

## How to review

Start with the schema removal in propose_change_comment.py, then the protocol updates. 
The migration in m060_set_comment_thread_created_by_on_node.py. 
The test for the migration validates both comment and thread backfill.

## How to test

```bash
# Unit tests

# Migration test
uv run pytest backend/tests/component/core/migrations/graph/test_060_set_comment_thread_created_by_on_node.py -xvs

# Proposed change mutation test (verifies thread creation without created_at)
uv run pytest backend/tests/component/graphql/mutations/test_proposed_change.py -xvs
```

## Impact & rollout

- **Backward compatibility:** 
Migration m060 preserves existing created_by data by copying it from relationships to node properties. The created_at and created_by fields are no longer exposed as schema attributes/relationships in the GraphQL API — consumers should use node_metadata instead.

- **Performance:** 
Slight improvement — fewer relationship traversals needed for comment/thread author lookups

- **Deployment notes:** 
Migration m060 runs automatically on startup. Frontend will need a corresponding update to read created_by from node_metadata instead of from the comment/thread relationship.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Posting comments no longer requires approver/auth gating; submissions succeed when comment content is provided.

* **Chores**
  * Removed inline comment/thread timestamp and author fields from the API; client now reads author/date from edge-level node_metadata.
  * Backend migration added to backfill node-level author metadata.
  * UI updated to treat author/date as optional and render them only when present.


backend/infrahub/core/migrations/graph/m068_set_comment_thread_created_by_on_node.py
```diff
@@ -0,0 +1,58 @@
...
+    type: QueryType = QueryType.WRITE
+    insert_return = False
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        query = """
+        MATCH (comment:CoreComment)
+        CALL (comment) {
+            MATCH (comment)-[r1:IS_RELATED]-(:Relationship {name: "comment__account"})-[r2:IS_RELATED]-(account:CoreGenericAccount)
+            RETURN account
+            ORDER BY r1.from DESC, r2.from DESC
```
ajtmccarty, line 25: I don't think either of these suggestions is required b/c this is a required cardinality-one relationship that should never be updated, but I haven't reviewed the schema, so not totally sure
- if the relationship is capable of being updated, you'll need to use `r1.from DESC, r1.status ASC, r.from DESC, r2.status ASC` for ordering to account for relationships added and deleted at the same timestamp
- if the relationship is optional, you'll need to account for checking if the edges are status="active" b/c there the latest relationship could be deleted

ajtmccarty (approved):

ogenstad (changes requested):
As this work also requires frontend changes, let's wait with this one until after the feature freeze or when we cut a `release-1.8` branch this can be merged into `develop` for inclusion in 1.9 instead. So I don't have any objections to the code itself only that we should wait a few days.

ogenstad (approved):
Updating my review to approved as we now have the `release-1.8` branch and this should go into `develop` for 1.9. However it looks like we already have a new merge conflict due to additional migrations..

ogenstad:
@solababs, looks like we have some failing tests for the frontend with regards to the GraphQL schema update. Can you look at that one? After that I think this should be good to go.

frontend/app/src/entities/diff/ui/file-diff/file-content-diff.tsx
```diff
@@ -113,7 +111,6 @@ export function FileContentDiff({
   commitTo,
 }: FileContentDiffProps) {
   const { proposedChangeId } = useParams();
-  const auth = useAuth();
```
bilalabbad, line 116: auth variables seems to be missing, causing artifact-content-diff and file-content-diff to crash the page.

bilalabbad (changes requested):
- CI is failing, if you check logs you should see how to fix it
- Graphql generated types for frontend are not up to date
- New frontend code broke file and artifact diff

frontend/app/src/entities/proposed-changes/ui/conversations/thread.tsx
```diff
@@ -152,9 +140,11 @@ export const Thread = (props: tThread) => {
     );
   };
 
-  const comments = thread?.comments?.edges?.map((comment: any) => comment.node) ?? [];
-
-  const sortedComments = R.sortBy(comments, (x) => new Date(x.created_at.value).getTime());
```
pa-lem, line 157: are we sure that threads and comments are correctly sorted in the backend if we don;t do it in the frontend anymore?

after some tests it seems so, but I just wanted to raise this point

pa-lem (approved):

bilalabbad (approved):

PR #8363, opened 2026-02-12
Add Spec Kit

ogenstad:
# Why

Adds [Spec Kit](https://github.github.com/spec-kit/) files to the repository

## How to review

* Standard implementation of spec kit
* Exceptions contains two symlinks:
    *  ./specs -> ./dev/specs
    * ./specify/memory/constitution.md -> ./dev/constitution.md

Of note is that the prefix of the commands differs from what we have in other projects `speckit` vs `sk` I opted for speckit as it is what was suggested by the default installer and could perhaps make upgrades somewhat easier if we keep the exact naming. But I can revert to `sk` if desired.

## Summary by CodeRabbit

* **New Features**
  * Added spec-driven development workflow commands for creating, analyzing, planning, and clarifying feature specifications
  * Introduced automation for feature branch creation, implementation planning, and task management
  * Added prerequisites validation for development workflows

* **Documentation**
  * Added comprehensive templates for specifications, implementation plans, tasks, checklists, and governance
  * Added detailed command documentation for the new workflow system


ogenstad:
Tagging @dgarros and @fatih-acar, mainly to determine if we should go with this approach and change the prefix of the commands in the other repo or if I should update them in this PR instead.

fatih-acar:
I think it's ok to keep `/speckit.`
Did you manage to change the default branch naming? By default it's doing 'XXX-feature-name' where XXX is a sequential number
We probably want to use our current naming scheme with name initials + date and/or jira stuff?

ogenstad:
> I think it's ok to keep `/speckit.` Did you manage to change the default branch naming? By default it's doing 'XXX-feature-name' where XXX is a sequential number We probably want to use our current naming scheme with name initials + date and/or jira stuff?

No in this PR I just wanted to add the base templates as I don't know if we need much reviews on those (Coderabbit doesn't seem to agree though..) and then in a follow up PR once we have this in `develop` we can start to look at a constitution there as well.

So based on that we'll go ahead and change the prefix in that other repo so that we keep it consistent across the org?

fatih-acar (approved):
Yes, go ahead.
Ok for the changes in the other repo

PR #8372, opened 2026-02-12
IFC-2255 Add support for `NumberPool` in templates

gmazoyer:
# Why

Add support to reference `NumberPool`s in templates without allocating numbers from them while still allocating numbers when creating objects from those templates.

**Goal:** Enable templates to properly use `NumberPool` attributes that allocate unique values when objects are created, with proper source tracking.

### Behavioral changes
- Templates with `NumberPool` attributes now correctly pass pool references to objects created from them

### Implementation notes

**`backend/infrahub/core/node/__init__.py`**
- `handle_pool()`: Added source assignment for templates - when a template has a `from_pool` reference, the source is now set to the pool ID

**`backend/infrahub/core/node/create.py`**
- `get_template_relationship_peers()`: Added `include_metadata=MetadataOptions.SOURCE` when fetching MANY cardinality relationship peers. This ensures `source_id` is loaded on template attributes, which is required for NumberPool handling in `extract_peer_data()`
- `extract_peer_data()`: Added filter to skip attributes that don't exist on the target schema (prevents `template_name` from being passed to non-template objects)

### What stayed the same
- IPAddressPool and IPPrefixPool relationship allocation (already worked)
- Template creation behavior (pools still don't allocate on template save)
- API contract unchanged

## How to review

1. Start with `backend/infrahub/core/node/__init__.py` - the `handle_pool()` change is small but critical
2. Then `backend/infrahub/core/node/create.py` - the `MetadataOptions.SOURCE` fix for nested templates
3. Tests in `backend/tests/functional/templates/test_template_resource_pool.py` demonstrate all scenarios

**Key insight:** The `source_id` on template attributes must be loaded to detect NumberPool references. Without `include_metadata=MetadataOptions.SOURCE`, the code path that reconstructs `from_pool` from `source_id` never triggers. There is also a repeated logic for two code paths, which is expected as there is one code path for the top-level template and another for sub-level templates.

## How to test

```bash
# Run all template resource pool tests
uv run backend/tests/component/templates/test_template_applier.py -v
```

## Summary by CodeRabbit

* **New Features**
  * Templates can reference number-based resource pools; when applied, objects record pool sources and allocate values from pools while still honoring explicit user-provided values.

* **Tests**
  * Expanded unit and functional tests covering pool-backed attributes, unique allocations across instances, explicit-override precedence, exhaustion handling, and source tracking.

### DIFF
--- a/backend/tests/functional/templates/test_template_resource_pool.py
+++ b/backend/tests/functional/templates/test_template_resource_pool.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from ipaddress import ip_network
+from ipaddress import IPv4Interface, IPv4Network, IPv6Interface, IPv6Network, ip_network
 from typing import TYPE_CHECKING
 
 import pytest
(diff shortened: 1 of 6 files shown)


backend/tests/functional/templates/test_template_resource_pool.py
```diff
@@ -333,3 +294,398 @@ async def test_template_cannot_add_direct_when_pool_exists(
...
+        """Object created from template should allocate from pool."""
+        rack = await client.create(
+            kind="InfraRack", name="rack-from-pool-template", object_template=template_with_pool_slot.id
+        )
+        await rack.save()
+
+        retrieved = await client.get(kind="InfraRack", id=rack.id, include=["slot_id"])
+        assert retrieved.name.value == "rack-from-pool-template"
+        assert retrieved.slot_id.value is not None
+        assert 1 <= retrieved.slot_id.value <= 100
```
ajtmccarty, line 386: is there a way to check the source for the attribute?

backend/tests/functional/templates/test_template_resource_pool.py
```diff
@@ -333,3 +294,398 @@ async def test_template_cannot_add_direct_when_pool_exists(
...
+        rack = await client.create(
+            kind="InfraRack",
+            name="rack-with-explicit-slot",
+            object_template=template_with_pool_slot.id,
+            slot_id=999,
+        )
+        await rack.save()
+
+        retrieved = await client.get(kind="InfraRack", id=rack.id)
+        assert retrieved.slot_id.value == 999
```
ajtmccarty, line 402: and the source here should be the template, I believe

ajtmccarty (approved):

backend/tests/functional/templates/test_template_resource_pool.py
gmazoyer: Not here, because this is a test with a user provided value, but in another test, we can indeed make that check

ajtmccarty:
the code rabbit comment `ONE-cardinality path cannot load source metadata, unlike MANY.` is wrong b/c the LINKED_NODES metadata option includes source metadata

PR #8374, opened 2026-02-12
fix(docs): HA examples redis misconfig

fatih-acar:
This is to ensure the cache's service name doesn't conflict with Infrahub's env config keys.

When using "cache" as the fullnameOverride and "infrahub" as the Helm deployment name, the service name is created as "infrahub-cache".
Kubernetes then populates containers with env vars related to the services, such as "INFRAHUB_CACHE_PORT" which conflcits with Infrahub's own settings... thus preventing it from starting correctly.

## Summary by CodeRabbit

* **Chores**
  * Updated Redis infrastructure configuration to improve service naming consistency and deployment alignment across the system.

### DIFF
--- a/development/deploy.tf
+++ b/development/deploy.tf
@@ -408,7 +408,7 @@ resource "kubernetes_deployment_v1" "redis_sentinel_proxy_deployment" {
             "-listen",
             ":6379",
             "-sentinel",
-            "cache:26379",
+            "redis:26379",
           ]
           port {
             container_port = 6379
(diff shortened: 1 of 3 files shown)


minitriga (approved):
@fatih-acar People updating their deployments will this add another running container or will it know to just rename?

fatih-acar:
> @fatih-acar People updating their deployments will this add another running container or will it know to just rename?

Yes, it will rename the deployment automatically.

PR #8379, opened 2026-02-13
Prevent setting `generate_template: true` on generics

### LINKED ISSUE #8371: bug: using `generate_template: True` on a GenericSchema breaks the GraphQL schema (and UI)
### Component

API Server / GraphQL

### Infrahub version

1.7.4

### Current Behavior

setting `generate_template: True` on a GenericSchema will create an invalid GraphQL schema, which breaks the frontend and prevents any GraphQL queries/mutations from running

### Expected Behavior

we should just prevent setting `generate_template: True` on a GenericSchema

### Steps to Reproduce

1. load a schema that includes a Generic with `generate_template=True` and at least one Node that inherits from the Generic (generate_template can be either True or False)
2. the schema loads successfully, then the UI breaks

### Additional Information

_No response_

gmazoyer:
# Why

Setting `generate_template: true` on a `GenericSchema` creates an invalid GraphQL schema, which breaks the frontend and prevents any GraphQL queries or mutations from running. This PR removes the `generate_template` attribute from `GenericSchema` so that it cannot be set anymore.

> [!IMPORTANT]
> We should probably backport this to `stable` too.

Closes #8371

## What changed

- Schema load will now be rejected if a user tries to set `generate_template` on a generic since the attribute no longer exists
- Removed the integration test `TestSchemaTemplateGenericInheritance` that exercised the now-blocked scenario
- Adapted some component tests to avoid impossible code path

Auto-generated sub-templates for generics used as component peers (e.g. `InfraInterface` generic referenced via a component relationship) are **not affected**, that path relies already supported generics without requiring them to have `generate_template` set to `true`.

## How to review

Small diff. Focus on the validation method in `schema_branch.py` but it's mostly about adding `isinstance(node, ...)` for typing reasons and conditionals. There is also a migration to remove the attribute from the graph.

## How to test

```bash
# Run migration test
uv run pytest backend/tests/component/core/migrations/graph/test_062_remove_generic_generate_template.py -v

# Existing sub-template tests still pass (DEVICE + INTERFACE_HOLDER path)
uv run pytest backend/tests/component/core/schema_manager/test_manager_schema.py \
  -k "test_manage_object_templates_with_component_relationships or test_identify_object_templates_with_generics" -v
```

## Impact & rollout

- **Backward compatibility:** users with `generate_template: true` on a generic will get an error on next schema load but there should not be running instances out there using this, as they would be broken anyway.
- **Deployment notes:** safe to deploy independently

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the generate_template option from Generic schemas to prevent invalid GraphQL schemas and related frontend/backend errors

* **Migrations**
  * Added a migration to remove legacy generate_template attributes and clean up related schema entries

* **Documentation**
  * Clarified that generate_template is only available on node definitions

* **UI**
  * "Generate template" setting now appears only for node schemas

* **Tests**
  * Added migration tests and updated schema/weight tests; removed an obsolete integration test

* **Chores**
  * Bumped graph version and adjusted test artifacts to match changes


ogenstad:
Is this attribute ever set to `True` within the backend code when processing the schema? If not it might make more sense to remove the attribute? Alternatively convert it to a private attribute. One thing that I'd wonder with the current change is if we might have a situation if a user has set this to True and we'd prevent that schema from loading after an upgrade.

ogenstad:
FYI, this should target the `release-1.8` branch.

backend/infrahub/core/migrations/graph/m062_remove_generic_generate_template.py
```diff
@@ -0,0 +1,69 @@
...
+WITH sa, rel
+LIMIT 1
+
+// Delete the SchemaAttribute and its relationship node
+DETACH DELETE sa, rel
+        """
+        self.add_to_query(query)
+
+
+class Migration062(GraphMigration):
```
gmazoyer, line 66: Not sure if this migration is too brutal. I expected that I needed to use `NodeAttributeRemoveMigration` but that was unsuccessful when running `infrahub upgrade`.

backend/infrahub/core/migrations/graph/m062_remove_generic_generate_template.py
```diff
@@ -0,0 +1,77 @@
...
+    type: QueryType = QueryType.WRITE
+    insert_return = False
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
+        query = """
+MATCH (sg:SchemaGeneric)-[:HAS_ATTRIBUTE]->(attr:Attribute {name: "generate_template"})-[:HAS_VALUE]->(val)
+DETACH DELETE attr
+WITH val
+WHERE NOT EXISTS { MATCH (val)-[]-() }
+DELETE val
```
ajtmccarty: I think you can just do
```
MATCH (sg:SchemaGeneric)-[:HAS_ATTRIBUTE]->(attr:Attribute {name: "generate_template"})
WITH DISTINCT attr
DETACH DELETE attr
```
b/c the value would always just be either True or False and I think you can assume those `AttributeValue` vertices are used elsewhere

ajtmccarty (approved):

bilalabbad (approved):

muffizone:
so i just tried upgrading my staging environment to 1.8 and I am hitting this bug where it fails to load the schema cos of this fix.

The strangest part is I've never used templates in infrahub or have them defined in my schema.  What I can see from the logs is that the value of generate_template is false which matches the default before this release.

This is the error log which I get when the api-server/task-worker boots and crashes

````
15:19:12.735 | ERROR   | uvicorn.error - Traceback (most recent call last):
  File "/.venv/lib/python3.13/site-packages/starlette/routing.py", line 694, in lifespan
    async with self.lifespan_context(app) as maybe_state:
               ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 211, in merged_lifespan
    async with original_context(app) as maybe_original_state:
               ~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/.venv/lib/python3.13/site-packages/fastapi/routing.py", line 211, in merged_lifespan
    async with original_context(app) as maybe_original_state:
               ~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/source/community/backend/infrahub/server.py", line 109, in lifespan
    await app_initialization(application)
  File "/source/community/backend/infrahub/server.py", line 87, in app_initialization
    is_initial_setup = await initialization(db=db, add_database_indexes=True)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/community/backend/infrahub/core/initialization.py", line 190, in initialization
    await registry.schema.load_schema(db=db, branch=branch)
  File "/source/community/backend/infrahub/core/schema/manager.py", line 750, in load_schema
    branch_schema = await self.load_schema_from_db(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        db=db, branch=branch, schema=current_schema, schema_diff=schema_diff
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/source/community/backend/infrahub/core/schema/manager.py", line 823, in load_schema_from_db
    schema=await self.convert_generic_schema_to_schema(schema_node=schema_node, db=db),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/community/backend/infrahub/core/schema/manager.py", line 881, in convert_generic_schema_to_schema
    return GenericSchema(**node_data)
  File "/.venv/lib/python3.13/site-packages/pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for GenericSchema
generate_template
  Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
 ```

**Comment by gmazoyer** [comment]
The `generate_template` field was removed from generics, it could still have been stored as `false`, but a graph migration is supposed clean it up from the database. It could be that the migration missed something. Can you check if you still have this `generate_template` attribute in your database after? A Cypher query such as the following should do the trick:
```
MATCH (sg:SchemaGeneric)-[:HAS_ATTRIBUTE]->(attr:Attribute {name: "generate_template"})
RETURN sg.uuid AS generic_uuid, attr.uuid AS attr_uuid
```

muffizone:
so i can't run the migration as the server goes in a crash loop with the pydantic validation failure when booting in v1.8.0. I'll have to probably run this migration cleanup before i update to v1.8.0

PR #8381, opened 2026-02-13
Fix support for setting a pool/rel when unsetting its counterpart

gmazoyer:
# Why

When updating a template to swap between a direct relationship and a resource pool allocation (e.g. unsetting the pool and setting the direct IP, or vice versa), the exclusivity constraint incorrectly rejects the operation even though only one side ends up set.

The `TemplateResourcePoolExclusiveConstraint` checks whether the counterpart relationship has peers, but never checks whether the current*relationship actually has peers. When clearing a relationship and setting its counterpart in the same mutation, the cleared relationship's constraint still fires and sees the counterpart being set hence raising a false violation.

## What changed

- **Fix:** Added an early return in `TemplateResourcePoolExclusiveConstraint.check()` that skips validation when the current relationship has no peers (is being cleared or was never set). No exclusivity to enforce on an empty relationship.
- **Refactor:** Extracted the peer-detection logic into `_relationship_has_peers()` and reused it in `_check_counterpart_not_set()`, removing the duplicated code.
- **Tests:** Added two functional tests covering both swap directions (pool -> direct and direct -> pool).

## How to test

```bash
```

These tests have been heavily rewritten so they depend less on the Python SDK to create dataset to work with. We also use raw GraphQL queries to clearly show the intent and how we expect things to work.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for resource-pool vs direct address assignments by adding an early peer check and consolidating peer-detection to avoid incorrect conflict evaluations.

* **Tests**
  * Reworked tests to use strongly-typed schema fixtures and GraphQL helpers.
  * Added comprehensive functional tests for pool↔direct transitions, pool-backed allocations, explicit-address overrides, and related edge cases.

### DIFF
--- a/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
+++ b/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
@@ -33,6 +33,9 @@ class TemplateResourcePoolExclusiveConstraint(RelationshipManagerConstraintInter
         if not node_schema.is_template_schema:
             return
 
+        if not await self._relationship_has_peers(relm=relm):
+            return
+
         rel_name = relm.schema.name
 
         if rel_name.endswith(RESOURCE_POOL_REL_SUFFIX):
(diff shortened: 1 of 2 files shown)


ogenstad (approved):
The code looks good to me, but I wonder if it would be more explicit to have GraphQL mutations so it was clear what actually was sent in the mutation. This can also be a future iteration if you think it makes sense.

PR #8383, opened 2026-02-13
fix handling deleted branches for parent info in DiffTree

ajtmccarty:
IFC-2258

logic to retrieve parent relationships in `DiffTreeResolver` failed if a branch was deleted b/c it tried to get a schema on a deleted branch, which raises `BranchNotFoundError`

solution is 
- try to get the schema on the input branch
- if that fails, try to get it on the default branch
- if that fails, set the parent relationship name to `undefined` (this is a fallback that already existed)

updates an integration test to cover this

I did manual testing and verified the fix on a local instance of infrahub

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema and parent resolution across branches to make diff relationship detection more reliable in multi-branch scenarios.

* **Tests**
  * Expanded integration tests for diff freezing: added device and interface nodes, validated parent relationship serialization (relationship_name "interfaces"), and ensured frozen diffs remain stable.

### DIFF
--- a/backend/infrahub/graphql/queries/diff/tree.py
+++ b/backend/infrahub/graphql/queries/diff/tree.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


ogenstad (approved):
LGTM, if you rebase against `release-1.8` it should fix the frontend linting problem.

Issue #6906, opened 2025-07-28
bug: activities are lost after 7 days


fatih-acar:
### Component

API Server / GraphQL

### Infrahub version

1.3.4

### Current Behavior

Due to the default retention setting of the task manager (see https://docs.prefect.io/v3/advanced/database-maintenance#configure-event-retention).
We can only have 7 days of history regarding the activities log.

### Expected Behavior

Activities log should be kept forever?

### Steps to Reproduce

Run infrahub for at least 7 days with activities.

saltas888:
Status update, still open, nothing merged. The 7 day default is unchanged.

#8398 is linked as closing this, but it's built on the old event retention config and has gone stale, we're closing it as superseded.

Where we landed: Prefect ships a DB vacuum service in the version we pin, with a global event retention period plus per-event-type overrides. It doesn't give us what we need out of the box though. Overrides can only shorten retention, never extend it, and the global sweep deletes everything older than the global period regardless of overrides. So we can't keep a short 7 day global and hand `infrahub.*` events a long override, the global has to be set to the longest retention we want, and the shortening has to target Prefect's own internal events instead.

That matters because Prefect internals are roughly 70% of the events we store, flow-run state changes, heartbeats, task-run states, automations. Raising retention for the activity log today also multiplies storage for noise we never display.

The plan is an Infrahub-owned sweep: set the global period to whatever activity log retention we want, then run our own vacuum task that deletes anything not prefixed `infrahub.` past a much shorter internal retention. That way internal event types added by future Prefect upgrades get the short retention automatically, with no list to maintain. The alternative is contributing prefix/wildcard support for the overrides upstream, cleaner long term but slower to land.

Two things to settle alongside it: Prefect queries default to a 180 day `occurred` window, so retention above that is invisible unless `since` is passed explicitly, and we'd want a single "activity log retention" setting in config/helm/compose rather than exposing both knobs.

Tracked in IFC-2934, with IFC-2933 covering activity log query performance at scale.

PR #8398, opened 2026-02-17
IFC-1702: Events retention period

solababs:
# Why

Prefect's internal event store grows unbounded by default, which can lead to increased memory/disk usage on long-running Infrahub workers. There is currently no way to configure how long Prefect events are retained.

Goal: Expose a configurable retention period for Prefect events so anyone can control how long workflow event data is kept.

Non-goals: This does not change any other Prefect lifecycle or cleanup behavior.

Closes https://github.com/opsmill/infrahub/issues/6906

## What changed

- New config setting: Added worker_events_retention_period (integer, days, default 7, minimum 1) to WorkflowSettings.

- Wire it into the worker: The new setting is passed to Prefect via PREFECT_EVENTS_RETENTION_PERIOD in the worker's temporary_settings block, formatted as "<n>d" (e.g. "7d").

## How to review

Check new Field definition on WorkflowSettings
Check passing the setting to Prefect's temporary_settings

## How to test

```bash
# Unit tests (verify no regressions in config/worker)

# Manual: override the setting and confirm it takes effect
INFRAHUB_WORKFLOW_WORKER_EVENTS_RETENTION_PERIOD=14 uv run infrahub worker start
# Check Prefect logs/settings to confirm PREFECT_EVENTS_RETENTION_PERIOD=14d
```

## Impact & rollout

- **Backward compatibility:** 
No breaking changes. The new setting has a sensible default (7 days); existing deployments require no configuration changes.

- **Performance:** 
Should improve long-running worker performance by bounding event store growth.

- **Config/env changes:** 
New optional env var INFRAHUB_WORKFLOW_WORKER_EVENTS_RETENTION_PERIOD (integer, days).

- **Deployment notes:** 
Safe to deploy without coordination. The setting takes effect on worker restart.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a configurable worker event retention period for workflows (default 7 days, minimum 1 day).

* **Documentation**
  * Added configuration reference and environment variable to override the worker event retention period; updated local development setup to reflect the new setting.


backend/infrahub/config.py
```diff
@@ -422,6 +422,9 @@ class WorkflowSettings(BaseSettings):
         ge=0,
         description="Threshold for caching flow run counts (0 to always cache, higher values to disable)",
     )
+    worker_events_retention_period: int = Field(
+        default=7, ge=1, description="Specify the number of days to retain events for (days)"
```
ajtmccarty: I think  `(days)` can be removed
`days` is already specified earlier in the description

ajtmccarty (approved):

changelog/6906.added.md
```diff
@@ -0,0 +1 @@
+Added new optional env var INFRAHUB_WORKFLOW_WORKER_EVENTS_RETENTION_PERIOD (integer, days) with default of 7d.
```
ajtmccarty: I'd say this should be `with default of 7` b/c the user does not include the `d` in their env var, we add that before sending to Prefect, right?

development/Dockerfile
```diff
@@ -87,7 +87,8 @@ WORKDIR /source
 ENV PREFECT_WORKER_QUERY_SECONDS="2" \
     PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES=500 \
     PREFECT_API_DATABASE_TIMEOUT=60 \
-    PREFECT_FLOWS_HEARTBEAT_FREQUENCY=30
+    PREFECT_FLOWS_HEARTBEAT_FREQUENCY=30 \
+    PREFECT_EVENTS_RETENTION_PERIOD="7d"
```
BeArchiTek, line 91: I am surprised we have to add this to the Dockerfile.

BeArchiTek (changes requested):
Could you run the invoke task to update the env variable in the docker compose files so `INFRAHUB_WORKFLOW_WORKER_EVENTS_RETENTION_PERIOD` is present

saltas888:
Superseded, I'd suggest we close this.

This targets `PREFECT_EVENTS_RETENTION_PERIOD`, which is the old event retention config. The direction we settled on is Prefect's DB vacuum service, which already ships in the version we pin, with per-event-type retention overrides on top of the global period.

Worth flagging that the vacuum service doesn't solve #6906 on its own either. Overrides can only shorten retention, never extend it (effective retention is `min(override, global)`), and the global sweep deletes everything past the global period with no exclusion for overridden types. So a short global with a long `infrahub.*` override doesn't work, the global has to be the longest retention we want and the shortening has to target Prefect's internal events instead.

Tracked in IFC-2934. Thanks for the work here @solababs, the config surface is still roughly what we'll want, just applied at a different layer.

PR #8399, opened 2026-02-17
fix(speckit): change branch naming scheme

fatih-acar:
## Summary by CodeRabbit

* **New Features**
  * Optional initials prefix for feature branch names (e.g., initials-###-feature-name); initials can be auto-detected or provided manually.
  * Branch/feature discovery now recognizes both initials-prefixed and legacy numeric-only formats.

* **Documentation**
  * Added PR checklist item to review AI-generated content.
  * Updated templates and command docs to reflect the optional initials-based branch naming and examples.

### DIFF
--- a/.specify/scripts/bash/create-new-feature.sh
+++ b/.specify/scripts/bash/create-new-feature.sh
@@ -5,6 +5,7 @@ set -e
 JSON_MODE=false
 SHORT_NAME=""
 BRANCH_NUMBER=""
+INITIALS=""
 ARGS=()
 i=1
 while [ $i -le $# ]; do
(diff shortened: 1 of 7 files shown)


.specify/scripts/bash/create-new-feature.sh
```diff
@@ -80,6 +95,37 @@ find_repo_root() {
     return 1
 }
 
+# Function to extract initials from git user.name
+# Takes first letter of first name + first two letters of last name, lowercased
+# Example: "Fatih Acar" -> "fac", "John Doe" -> "jdo"
+get_git_initials() {
+    local full_name
+    full_name=$(git config user.name 2>/dev/null || echo "")
```
ogenstad, line 110: I think there are some different variations of this. For instance Aaron uses ajtm. Perhaps the git command could also be a fallback if an environment variable named something like OPSMILL_GIT_USER_SHORT isn't set.

.specify/scripts/bash/create-new-feature.sh
```diff
@@ -246,26 +299,43 @@ if [ -z "$BRANCH_NUMBER" ]; then
...
+        echo "Error: No git available and --initials not provided." >&2
+        echo "Use --initials <initials> to specify manually." >&2
+        exit 1
+    fi
+fi
+
 # Force base-10 interpretation to prevent octal conversion (e.g., 010 → 8 in octal, but should be 10 in decimal)
 FEATURE_NUM=$(printf "%03d" "$((10#$BRANCH_NUMBER))")
-BRANCH_NAME="${FEATURE_NUM}-${BRANCH_SUFFIX}"
+BRANCH_NAME="${INITIALS}-${FEATURE_NUM}-${BRANCH_SUFFIX}"
```
ogenstad, line 327: We could perhaps get to this later, but if we're looking at changing the existing files it could also be good to mention that we can add a suffix key to link any PR to a Jira issue. I.e. if the branch name ends with something like "-IFC-906" we can then see that the issue is linked to a PR from with Jira.

.specify/scripts/bash/create-new-feature.sh
fatih-acar: I added an optional `OPSMILL_GIT_USER_SHORT` env var for this case, thanks

.specify/scripts/bash/create-new-feature.sh
fatih-acar: Yeah I also thought including Jira IDs could be good, that would require asking the user for the ID during /speckit.specify I guess
We can come back to this later?

ogenstad: Yes

PR #8400, opened 2026-02-17
feat: local telemetry storage

fatih-acar:

As per the quickstart.md file:

## What This Feature Does

Infrahub now stores a daily telemetry snapshot locally in the database, regardless of whether remote telemetry reporting is enabled. This ensures all deployments — including air-gapped and opted-out — retain usage data for support, auditing, and license compliance.

## Key Changes

1. **Automatic daily storage**: The daily telemetry workflow now always stores a snapshot locally before optionally sending to the remote endpoint.
2. **CLI export**: `infrahubctl telemetry export` exports snapshots to a JSON file.
3. **CLI listing**: `infrahubctl telemetry list` shows a summary table of stored snapshots.
4. **REST API**: `GET /api/telemetry/snapshots` provides programmatic access.
5. **Permission**: Access requires `READ_TELEMETRY` global permission.


dev/specs/fac-001-local-telemetry-storage/research.md
```diff
@@ -0,0 +1,149 @@
...
+
+### R-003: Telemetry Payload Serialization
+
+**Decision**: Store the `TelemetryData` payload as a JSON string in a single `data` field on the `TelemetrySnapshot` StandardNode.
+
+**Rationale**:
+- `StandardNode.to_db()` automatically serializes Pydantic models to JSON via `model_dump_json()` and deserializes them in `from_db()` via `ujson.loads()`. This is proven by existing test fixtures (`PadanticStdNode` in test_node_standard.py).
+- The telemetry payload is ~3-5 KB of JSON. Storing as a single JSON string avoids fragmenting it across multiple Neo4j nodes/relationships.
+- The `TelemetryData` model is already well-typed with nested Pydantic models (`TelemetryWorkerData`, `TelemetryBranchData`, `TelemetrySchemaData`, `TelemetryDatabaseData`, `TelemetryPrefectData`).
+- Using the existing model type in the StandardNode field enables automatic Pydantic validation on read.
```
dgarros: I think we need to be careful with that, overtime the format of this data will change and we'll have old data in the database, we need to make sure that we can still read and export the old data, not just the latest one

dev/specs/infp-471-local-telemetry-storage/research.md
```diff
@@ -0,0 +1,149 @@
...
+- Commands will use `initialize_client()` for authenticated API access and Rich for formatted output (tables for `list`, file writing for `export`).
+- The CLI will call the REST API endpoint (FR-009) rather than directly accessing the database, ensuring permission checks are enforced and the CLI works remotely.
+
+**Alternatives considered**:
+- **Backend CLI (infrahub command)**: The `infrahub` CLI is for server-side operations (start server, manage DB). Telemetry export is a client-side administrative operation.
+- **Direct database access from CLI**: Would bypass permission checks and require database credentials. The SDK client pattern already handles authentication.
+
+---
+
+### R-006: REST API Design — Programmatic Interface
```
dgarros, line 92: it's not clearly mentioned but I think we should have a sane default limit for this API to avoid returning everything each time and potentially having a pagination system would be helpful to avoid returning 100s of responses at once

dev/specs/fac-001-local-telemetry-storage/research.md
fatih-acar: Good point... We should probably create a `TelemetryData` Pydantic class for each format version, so that it gets correctly (de)serialized. I will add to the plan.

dev/specs/infp-471-local-telemetry-storage/research.md
fatih-acar: Pagination system will be implemented as described in `contracts/rest-api.md‎`

dev/specs/fac-001-local-telemetry-storage/research.md
fatih-acar: Updated the spec: instead of versioned Pydantic classes, the `data` field is now stored as plain `dict[str, Any]` (JSON). The `TelemetryData` Pydantic model is still used at collection time for validation, then `model.model_dump()` converts it to a plain dict before storage. On read, `data` comes back as a raw dict — no deserialization needed. The `payload_format` field identifies which version of the schema produced the data, so consumers can interpret differences if needed. This keeps old snapshots always readable regardless of how the model evolves.

dev/specs/infp-471-local-telemetry-storage/contracts/rest-api.md
```diff
@@ -0,0 +1,172 @@
...
+
+```python
+class TelemetrySnapshotResponse(BaseModel):
+    id: str                          # UUID
+    created_at: str                  # ISO 8601 timestamp
+    kind: str                        # Telemetry kind identifier
+    payload_format: str              # Format version
+    deployment_id: str               # Deployment identifier
+    infrahub_version: str            # Product version
+    data: dict[str, Any]             # Full telemetry payload
```
ajtmccarty, line 79: ~might be nice for this to be structured data too, but not sure if it really matters in this case~
I see the research section explains why the data is unstructured: to be backwards compatible with old payload formats when updating the format going forward

dev/specs/fac-001-local-telemetry-storage/contracts/rest-api.md
```diff
@@ -0,0 +1,172 @@
...
+Export telemetry snapshots to a JSON file.
+
+```
+infrahubctl telemetry export [OPTIONS]
+
+Options:
+  --output PATH          Output file path [default: telemetry-export.json]
+  --start-date TEXT      Start date filter (ISO 8601, e.g. 2025-01-01)
+  --end-date TEXT        End date filter (ISO 8601, e.g. 2026-02-16)
+  --branch TEXT          Branch context (not used for filtering, required by SDK)
```
ajtmccarty: it seems like we should have some way to keep the `branch` option out of a sub-command

dev/specs/fac-001-local-telemetry-storage/data-model.md
```diff
@@ -0,0 +1,147 @@
...
+WHERE elementId(n) = $node_id OR n.uuid = $node_id
+RETURN n
+```
+
+### List Snapshots (with date-range filter)
+
+Custom filter via `TelemetrySnapshotGetListQuery`:
+```cypher
+MATCH (n:TelemetrySnapshot)
+WHERE n.created_at >= $start_date AND n.created_at <= $end_date
```
ajtmccarty: maybe want to support `null` date variables in here too

ajtmccarty (commented):
looks pretty good to me

fatih-acar:
Thanks for the review @ajtmccarty I've addressed your comments and pushed the actual implementation
I'll do some manual testing

dev/specs/fac-001-local-telemetry-storage/contracts/rest-api.md
```diff
@@ -0,0 +1,170 @@
...
+class TelemetrySnapshotResponse(BaseModel):
+    id: str                          # UUID
+    created_at: str                  # ISO 8601 timestamp
+    kind: str                        # Telemetry kind identifier
+    payload_format: str              # Format version
+    deployment_id: str               # Deployment identifier
+    infrahub_version: str            # Product version
+    data: dict[str, Any]             # Full telemetry payload
+    checksum: str                    # SHA-256 integrity hash
+    remote_send_status: str          # "sent" | "skipped" | "failed" | "pending"
```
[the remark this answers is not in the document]
polmichel: I was about to write it down as a review comment. This is not blocking and if there is some rush around this feature, I think we can go for merge. But this is cleaner.

dev/specs/infp-471-local-telemetry-storage/contracts/rest-api.md
```diff
@@ -0,0 +1,170 @@
...
+Export telemetry snapshots to a JSON file.
+
+```
+infrahubctl telemetry export [OPTIONS]
+
+Options:
+  --output PATH          Output file path [default: telemetry-export.json]
+  --start-date TEXT      Start date filter (ISO 8601, e.g. 2025-01-01)
+  --end-date TEXT        End date filter (ISO 8601, e.g. 2026-02-16)
+  --config-file TEXT     Path to infrahubctl config file
```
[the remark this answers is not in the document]
polmichel: Seems like a valid catch, the targeted personas in the specifications are listed as: _As a customer administrator or OpsMill support engineer_. For OpsMill support engineer I think an internal chat would help, but regarding the use case of customer administrator, this comment makes sense.

dev/specs/fac-001-local-telemetry-storage/contracts/rest-api.md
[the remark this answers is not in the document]
polmichel: (And I would tend to use a `StrEnum`)

backend/infrahub/telemetry/tasks.py
```diff
@@ -96,21 +104,64 @@ async def post_telemetry_data(url: str, payload: dict[str, Any]) -> None:
...
+        deployment_id=str(registry.id) if registry.id else "",
+        infrahub_version=__version__,
+        data=data_dict,
+        checksum=checksum,
+        remote_send_status=REMOTE_SEND_STATUS_PENDING,
+    )
+
+    # Always store locally
+    try:
+        database = await get_database()
```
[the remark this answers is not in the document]
polmichel: Seems like a legit comment, but the only impact that I would see is when the user has disabled anonymous usage and when the db is failing when get. Unlikely to happen, exception is caught anyway, but this is cleaner.

backend/tests/unit/telemetry/test_workflow.py
```diff
@@ -0,0 +1,123 @@
...
+from infrahub.telemetry.tasks import send_telemetry_push
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+LOGGER_NAME = "infrahub.telemetry.tasks"
+
+
+@asynccontextmanager
+async def _fake_session(read_only: bool = False) -> Any:
```
polmichel: is this used? Speaking about the `read_only` parameter, the tests seem to pass without it.
Once again, this is not blocking.

backend/infrahub/telemetry/tasks.py
```diff
@@ -96,21 +104,64 @@ async def post_telemetry_data(url: str, payload: dict[str, Any]) -> None:
...
+        log.info(f"Telemetry snapshot stored locally (uuid={snapshot.uuid}).")
+    except Exception as exc:
+        log.warning(f"Failed to store telemetry snapshot locally: {exc}")
+
+    # Conditionally send remotely
+    if config.SETTINGS.main.telemetry_optout:
+        log.info("User opted out of remote telemetry. Marking snapshot as skipped.")
+        try:
+            snapshot.remote_send_status = REMOTE_SEND_STATUS_SKIPPED
+            if snapshot.id:
```
polmichel: In a sense, this is related to the comment about `get_database()` failing, but if the save did not happen, the logger will log the message _"User opted out of remote telemetry. Marking snapshot as skipped"_. However, as the save failed, the node has no id, right?

-> `self.id = node.element_id` in `infrahub.core.node.standard.StandardNode.create`
 


This means that the message saying the snapshot is skipped will appear, but the if statement will just ignore the code and return without providing any additional information, which could be confusing. 

Perhaps it would be better not to enter there if the save did not happen beforehand (<=> if snapshot.id is None), or to add another log if snapshot.id is None.
This is based on the assumption that id is None if the node is not saved.
Does this make sense?

backend/infrahub/telemetry/snapshot.py
```diff
@@ -0,0 +1,130 @@
...
+        offset: int = 0,
+    ) -> list[TelemetrySnapshot]:
+        filters: list[str] = []
+        extra_params: dict[str, Any] = {}
+
+        if start_date:
+            filters.append("n.created_at >= $start_date")
+            extra_params["start_date"] = start_date
+        if end_date:
+            filters.append("n.created_at <= $end_date")
```
polmichel: Does it work when `end_date` is the same date than the creation date? 

created_at seems to be set during this method using Timestamp  
```
    @field_validator("created_at", mode="before")
    @classmethod
    def set_created_at(cls, value: str) -> str:
        return Timestamp(value).to_string()

```

This code executed "as of today" does not return anything for instance

    _snapshot = _make_snapshot()
    await snapshot.save(db=db)

    results = await TelemetrySnapshot.get_list_filtered(db=db, end_date="2026-04-10")_

And the description state

```
 end_date: str | None = Query(                                                                                                                                         
      default=None, description="Include snapshots created on or before this date (ISO 8601)"                                                                         
  )      
```                                                                                                                                                               
 
If Cypher compares the string this case will fail

polmichel (commented):
This looks good! Just a comment worth really looking into regarding the end_date bound in the query when querying "today's data

backend/infrahub/telemetry/constants.py
```diff
@@ -3,6 +3,13 @@
 TELEMETRY_KIND: str = "community"
 TELEMETRY_VERSION: str = "20250318"
 
+REMOTE_SEND_STATUS_PENDING: str = "pending"
+REMOTE_SEND_STATUS_SENT: str = "sent"
+REMOTE_SEND_STATUS_SKIPPED: str = "skipped"
+REMOTE_SEND_STATUS_FAILED: str = "failed"
```
ajtmccarty: looks like an enum to me

backend/infrahub/telemetry/snapshot.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class TelemetrySnapshot(StandardNode):
+    kind: str
+    payload_format: str
+    deployment_id: str
+    infrahub_version: str
+    data: dict[str, Any]
+    checksum: str
+    remote_send_status: str = Field(default=REMOTE_SEND_STATUS_PENDING)
```
ajtmccarty: also looks like an enum

backend/infrahub/telemetry/snapshot.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @field_validator("data")
+    @classmethod
+    def data_must_be_non_empty(cls, v: dict[str, Any]) -> dict[str, Any]:
+        if not v:
+            raise ValueError("data must be a non-empty dict")
+        return v
+
+    @classmethod
+    async def get_list_filtered(
```
ajtmccarty: this method should be in a separate class (`TelemetrySnapshotRepository` or something to indicate it is for CRUD `Snapshot` operations on the database)

my preferred organization would be 
- `TelemetrySnapshot` in `models.py`
- `TelemetrySnapshotGetListQuery` in `queries.py`
- `TelemetrySnapshotRepository` in `repository.py`

backend/infrahub/telemetry/snapshot.py
ajtmccarty: I'd say this logic belongs in the query b/c it is cypher-specific

backend/infrahub/api/telemetry.py
```diff
@@ -0,0 +1,85 @@
...
+    snapshots = await TelemetrySnapshot.get_list_filtered(
+        db=db,
+        start_date=start_date,
+        end_date=end_date,
+        limit=limit,
+        offset=offset,
+    )
+
+    return TelemetrySnapshotListResponse(
+        count=len(snapshots),
```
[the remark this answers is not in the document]
ajtmccarty: @fatih-acar I think the robot is right about this. not sure if a separate query to get the correct count or updating the existing query to return the count is the best approach

backend/infrahub/telemetry/tasks.py
```diff
@@ -96,21 +104,64 @@ async def post_telemetry_data(url: str, payload: dict[str, Any]) -> None:
...
+                async with database.start_session() as db:
+                    await snapshot.save(db=db)
+        except Exception as exc:
+            log.warning(f"Failed to update snapshot status to skipped: {exc}")
```
ajtmccarty: does this need to be in a try-except if it is always going to return on the next line? could it just allow the error to be raised?

backend/infrahub/telemetry/tasks.py
```diff
@@ -96,21 +104,64 @@ async def post_telemetry_data(url: str, payload: dict[str, Any]) -> None:
 @flow(name="anonymous_telemetry_send", flow_run_name="Send anonymous telemetry")
 async def send_telemetry_push() -> None:
```
ajtmccarty, line 103: I was just on my dependency injection soapbox yesterday for a different PR making some changes to a flow and I'm going to do it again

to be more SOLID (particularly the D for dependency injection), this flow should just instantiate a new `TelemetrySnapshotService` that encapsulates all of the below logic. something like this

```
async def send_telemetry_push() -> None:
    telemetry_repository = TelemetryRepository(db=db)
    telemetry_service = TelemetrySnapshotService(
        db=db, telemetry_repository=telemetry_repository,
    )

    data = await gather_anonymous_telemetry_data()
    await telemetry_service.save_and_send(data)
```

and then in your tests, you don't have to patch a bunch of stuff. you can create mocks using `telemetry_repo = AsyncMock(spec=TelemetryRepository)` or `create_autospec(TelemetryRepository, spec_set=True)` and you can assert they are called with `telemetry_repo.assert_awaited_once_with()` or `telemetry_repo.assert_has_awaits()` or another method on the mock like this

I think it improves tests (accurate interfaces for the mocks instead of "anything goes") and makes them easier to read (no more `patch` context managers)

that said, this would probably require some more refactoring to turn `gather_anonymous_telemetry_data` into `TelemetryDataGatherer` and `post_telemetry_data` into `TelemetryDataPoster` to get the full benefits, but something to consider and an LLM can probably make pretty quick work of it

backend/infrahub/telemetry/snapshot.py
fatih-acar: I don't think I can achieve that without duplicating the query from StandardNodeGetListQuery?

backend/infrahub/telemetry/tasks.py
fatih-acar: Thanks I'll try to tackle that refactor in another PR (to not block 1.9)

backend/infrahub/telemetry/snapshot.py
ajtmccarty: can you pass `start_date` and `end_date` into `TelemetrySnapshotGetListQuery` and use `TelemetrySnapshotGetListQuery.query_init` to create and pass the `raw_filter` into `StandardNodeGetListQuery.query_init`?

fatih-acar: oh ok got it, that's exactly what Claude did in my latest changes... thanks!

backend/tests/unit/telemetry/test_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+        return_value=logging.getLogger(LOGGER_NAME),
+    ):
+        yield
+
+
+@pytest.fixture
+def telemetry_mocks() -> Iterator[dict[str, Any]]:
+    """Combined fixture providing all mocks needed for telemetry workflow tests."""
+    repo = create_autospec(TelemetrySnapshotRepository, spec_set=True, instance=True)
+    repo.save = AsyncMock(return_value=None)
```
ajtmccarty: `repo.save.return_value = None` would work too I think and skips creating a new `AsyncMock`

ajtmccarty (approved):

PR #8402, opened 2026-02-17
chore(backend): remove unrequired config toggle

fatih-acar:
## Summary by CodeRabbit

* **Chores**
  * Removed enable/disable configuration toggles from broker, cache, and workflow services
  * Simplified service initialization by removing conditional enable checks
  * Updated mutation event validation to include additional context requirements

* **Tests**
  * Removed test fixtures and parameters related to broker configuration


backend/infrahub/config.py
gmazoyer, line 1: I suppose that this change is tied to the fact we actually need these components for Infrahub to work properly, so there is no point to having them disabled?

gmazoyer (approved):

backend/infrahub/config.py
fatih-acar: Yes, we had a case where the user disabled the cache and was expecting Infrahub to work properly... (see Slack thread)

ogenstad:
Can we add a news fragment to this since it could in theory be a breaking change if someone were to use `infrahub.toml` files and have this option set either to true or false?

fatih-acar:
> Can we add a news fragment to this since it could in theory be a breaking change if someone were to use `infrahub.toml` files and have this option set either to true or false?

Done

ogenstad (approved):

PR #8403, opened 2026-02-17
Add specs for custom webhook headers

ogenstad:
# Why

Adds specifications for Spec Kit for custom webhook headers INFP-INFP-445

## What changed

* Added markdown files for webhook specification

## How to review

* Consider the filename of the spec folder, currently it uses an integer and short name of the feature. This is the Spec Kit default. Other options include using a date as we've done in the past. Alternatively we could reference the epic in Jira.
* I opted for using the key value pair for generic and nodes instead of the initial suggestion to have a generic for HTTP headers. I figured that this approach is more general and could possibly apply to other things as well
* Review and match with regards to expectation of the feature

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive specification for custom HTTP headers feature in webhooks, detailing support for authentication headers, environment-variable-based header injection, and reusable header configurations across multiple webhooks
  * Introduced quality assurance checklist documentation to ensure specification completeness and readiness for implementation planning

### DIFF
--- a/dev/specs/infp-445-webhook-headers/spec.md
+++ b/dev/specs/infp-445-webhook-headers/spec.md
@@ -1,0 +1,135 @@
+# Feature Specification: Custom HTTP Headers for Webhooks
+
+**Feature Branch**: `infp-445-webhook-headers`
+**Created**: 2026-02-17
+**Status**: Draft
+**Input**: INFP-445 - Add custom headers for webhooks and custom webhooks
+**Related**: INFP-470 - Customer feature request for custom HTTP headers on webhooks
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - Attach Authentication Headers to a Webhook (Priority: P1)
+
+An infrastructure administrator configures a webhook in Infrahub to notify an external system (e.g., Ansible Automation Platform) about infrastructure changes. The external system requires an authentication header such as `Authorization: Bearer <token>` to accept incoming requests. The administrator creates a key-value pair containing the sensitive header credentials, associates it with the webhook, and when events fire, the custom header is automatically included in the HTTP request.
+
+**Why this priority**: This is the core use case driving the feature. Without it, customers cannot integrate Infrahub webhooks with any system requiring header-based authentication, which is the majority of modern APIs and automation platforms.
+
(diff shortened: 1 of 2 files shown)


dgarros (approved):

dev/specs/infp-445-webhook-headers/spec.md
```diff
@@ -0,0 +1,135 @@
...
+- Q: When an environment variable header references a missing variable, should the entire webhook delivery fail or just that header? → A: Skip the unresolvable header, send remaining headers, log a warning.
+- Q: Should the key-value pair name be globally unique across all types? → A: Yes, globally unique across all key-value pair types, consistent with standard Infrahub generic behavior.
+- Q: Does the feature need special audit logging for key-value pair lifecycle events? → A: No; rely on existing Infrahub audit/change tracking for node mutations. No feature-specific observability required.
+
+## Success Criteria *(mandatory)*
+
+### Measurable Outcomes
+
+- **SC-001**: Users can successfully send webhook requests to external systems requiring custom authentication headers (e.g., Ansible Automation Platform) without any workarounds or intermediary proxies.
+- **SC-002**: Users can configure a webhook with custom headers and trigger a successful authenticated request to an external endpoint within 5 minutes of starting the configuration process.
```
wvandeun: Where does this 5 minute come from? Also I think we should clarify this a bit, what do we mean with starting the configuration process here? Shouldn't this be "within x time after a event triggers a webhook"?

dev/specs/infp-445-webhook-headers/spec.md
wvandeun, line 1: In multiple places we are stating that sensitive values should be stored securely. Given how this is done in Infrahub today, wouldn't it be better to remove that from the spec? Similarly in multiple places it is mentioned that sensitive values would be masked, including the API response. Assuming we mean the GraphQL API here, which doesn't mask attribute values of kind Password, we should probably remove that requirement for the API?

wvandeun (approved):

dev/specs/infp-445-webhook-headers/spec.md
ogenstad: Yes good point. I've updated this part.

dev/specs/infp-445-webhook-headers/spec.md
ogenstad: I've rephrased this a bit.

PR #8405, opened 2026-02-17
IFC-2271 Implement closest match prefix via search

gmazoyer:
﻿# Why

When network engineers need to find which prefix an IP address belongs to, they currently have to manually navigate the IPAM prefix hierarchy or calculate prefix containment themselves. This makes IPAM adoption painful for customers managing large IP spaces.

This PR adds parent prefix lookup to the Search Anywhere dialog (Cmd+K): typing a valid IP address or CIDR prefix returns the containing parent prefixes, ordered by specificity, across all IP namespaces.

## What changed

**Backend: new prefix lookup query + resolver integration**
- Added `IPParentPrefixLookupQuery` Cypher query that finds all prefixes containing a given IP/CIDR by computing all possible binary prefix representations and matching against the existing `binary_address` RANGE index on `AttributeIPNetwork` nodes
- Added `_try_parse_ip_or_prefix()` helper to detect valid IP addresses and CIDR prefixes
- Modified `search_resolver` to route valid IPs/CIDRs through the new prefix lookup path instead of text search, returning results ordered by prefix length DESC
- Added `is_prefix_lookup` boolean field to the `NodeEdges` GraphQL type so the frontend can distinguish prefix lookups from text searches

**Frontend: dedicated prefix results UI**
- Added `SearchPrefixes` component showing parent prefixes with prefix value, namespace badge, and "Exact match" indicator
- `SearchNodes` hides when `isPrefixLookup` is true (avoids duplicate/confusing results)
- Fixed `SearchDocs` crash when the docs search API returns non-array responses (dev-only bug where docs index isn't built)

**What stayed the same**
- Existing text search behavior is unchanged for non-IP queries
- Existing IPv6 format normalization (`_collapse_ipv6`) still runs before prefix detection
- GraphQL API contract is backward-compatible: `is_prefix_lookup` is an optional nullable field, existing queries that don't request it are unaffected

### Suggested review order
1. `backend/infrahub/core/query/ipam.py` — the Cypher query (binary prefix matching logic)
2. `backend/infrahub/graphql/queries/search.py` — resolver integration and `_try_parse_ip_or_prefix`
3. `frontend/app/src/entities/navigation/ui/search-anywhere/search-prefixes.tsx` — new UI component
4. `frontend/app/src/entities/navigation/ui/search-anywhere/search-nodes.tsx` — conditional hide
5. Tests in `backend/tests/component/graphql/queries/test_search.py`

## How to review

- **Core logic**: The binary prefix matching in `IPParentPrefixLookupQuery.query_init` — verify the `possible_prefix_map` computation correctly generates all candidate binary prefixes and that the Cypher `any()` predicate properly filters by prefix length
- **Edge cases**: `_try_parse_ip_or_prefix` must return `None` for partial IPs like `10.1.2` so they fall through to text search
- **Frontend**: `SearchPrefixes` fires a second GraphQL query per result to fetch prefix details (value, namespace) — acceptable for the small result set (≤10) but worth noting

## How to test

```bash
# Backend unit + component tests
uv run pytest backend/tests/component/graphql/queries/test_search.py -v
```

Manual testing:
1. Start local dev environment with IPAM data
2. Open Cmd+K, type a valid IP like `10.10.1.1` → should show "Parent Prefixes" group with containing prefixes ordered by specificity
3. Type a CIDR like `10.10.1.0/24` → should show exact match + parent prefixes (no children)
4. Type partial IP like `10.1.2` → should fall through to normal text search
5. Type non-IP text like `router-01` → normal text search, no prefix section

## Impact & rollout

- **Backward compatibility:** Fully backward-compatible. `is_prefix_lookup` is an optional nullable GraphQL field; existing clients that don't request it see no change.
- **Performance:** The Cypher query uses the existing `binary_address` RANGE index. For IPv4, at most 33 candidate prefixes are checked; for IPv6, at most 129. No new indexes needed.
- **Deployment notes:** Safe to deploy. No migrations, no config changes.

## Checklist

- [x] Tests added/updated


gmazoyer:
Closing as it was just a test

PR #8410, opened 2026-02-18
Add spec for local computed attribute recalculation

ogenstad:
# Why

Add Spec Kit specification for recomputation of local computed attributes after node update mutations

## What changed

Add markdown files for Spec Kit.

### DIFF
--- a/dev/specs/infp-460-local-computed-attributes/checklists/requirements.md
+++ b/dev/specs/infp-460-local-computed-attributes/checklists/requirements.md
@@ -1,0 +1,62 @@
+# Specification Quality Checklist: Local Execution of Jinja2 Computed Attributes
+
+**Purpose**: Validate specification completeness and quality before proceeding to planning
+**Created**: 2026-02-18
+**Feature**: [spec.md](../spec.md)
+
+## Content Quality
+
+- [x] No implementation details (languages, frameworks, APIs)
+- [x] Focused on user value and business needs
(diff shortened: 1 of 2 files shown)


dev/specs/003-local-computed-attributes/spec.md
```diff
@@ -0,0 +1,171 @@
...
+- **Resource Optimization**: Reduces excessive CPU usage, database load, and logging overhead from background task processing
+- **Improved User Experience**: Updates visible immediately; fewer confusing duplicate notifications
+- **Better Scalability**: Frees up background worker capacity for other operations
+- **Tech Debt Reduction**: Addresses highest priority item from tech debt list
+
+## Dependencies & Assumptions *(mandatory)*
+
+### Dependencies
+
+- **INFP-441**: Placeholder automation refactor must be completed before this feature
```
dgarros: it's not clear to me why this INFP-441 is a pre-requisite to this task and if so should we add a spec for it as well to review both at once ?

dgarros (approved):
LGTM

dev/specs/003-local-computed-attributes/spec.md
ogenstad: No the wording of that one is a bit harsh. It's more that they are related to each other. I'll reword that since the other task was downvoted with regards to priority. I'll also update the folder name and then merge this.

PR #8416, opened 2026-02-18
link PC to diff if an open PC exists for branch diff being updated

ajtmccarty:
Intermittently failing E2E test revealed that there is a race condition if you create a proposed change for a branch and then refresh the diff on that proposed change as fast as possible, it is possible to end up temporarily not linking the new proposed change to the diff

PRs adds a check for an opened proposed change on the branch for the diff being updated and links the diff to the PC, but only if the diff being updated is set to track the whole branch

And update some tests that now require the `CoreProposedChange` schema to be registered

## Summary by CodeRabbit

* **Improvements**
  * Branch diffs now automatically discover and link open proposed changes for the source branch, improving change-tracking accuracy during diffs and merges.
* **Tests**
  * Expanded test coverage and added fixtures to validate proposed-change discovery and linking behavior across diff and merge scenarios.

### DIFF
--- a/backend/tests/component/core/diff/test_coordinator.py
+++ b/backend/tests/component/core/diff/test_coordinator.py
@@ -3,6 +3,8 @@ from typing import Any
 from unittest.mock import AsyncMock
 from uuid import uuid4
 
+import pytest
+
 from infrahub.core import registry
 from infrahub.core.branch import Branch
 from infrahub.core.constants import DiffAction, InfrahubKind
@@ -16,7 +18,6 @@ from infrahub.core.diff.repository.repository import DiffRepository
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
-from infrahub.core.schema.schema_branch import SchemaBranch
 from infrahub.core.timestamp import Timestamp
 from infrahub.database import InfrahubDatabase
 from infrahub.dependencies.registry import get_component_registry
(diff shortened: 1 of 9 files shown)


backend/tests/component/core/diff/test_coordinator.py
```diff
@@ -20,9 +22,14 @@
 from infrahub.database import InfrahubDatabase
 from infrahub.dependencies.registry import get_component_registry
 from infrahub.exceptions import SchemaNotFoundError
+from infrahub.proposed_change.constants import ProposedChangeState
 
 
 class TestDiffCoordinator:
+    @pytest.fixture(autouse=True)
+    async def _setup_core_schema(self, register_core_models_schema) -> None:
```
ogenstad: Please add a typehint to register_core_models_schema as we're doing that in parallell across the component tests.

backend/tests/component/core/diff/test_coordinator_lock.py
```diff
@@ -15,15 +14,16 @@
...
-from infrahub.core.schema.definitions.core.repository import core_read_only_repository, core_repository
-from infrahub.core.schema.node_schema import NodeSchema
 from infrahub.core.timestamp import Timestamp
 from infrahub.database import InfrahubDatabase, get_db
 from infrahub.dependencies.registry import get_component_registry
 
 
 class TestDiffCoordinatorLocks:
+    @pytest.fixture(autouse=True)
+    async def _setup_core_schema(self, register_core_models_schema) -> None:
```
ogenstad: Same comment with type hints

backend/tests/component/core/diff/test_diff_and_merge.py
```diff
@@ -43,6 +43,10 @@
 
 
 class TestDiffAndMerge:
+    @pytest.fixture(autouse=True)
+    async def _setup_core_schema(self, register_core_models_schema) -> None:
```
ogenstad: Here too

ogenstad (approved):
LGTM. I note that a PC test is still failing but didn't check the reason and it could be due to something else..

PR #8417, opened 2026-02-18
IFC-2297 / Partial fix of issue #7036 : Identifier is not retrieved when queried from the Number Pools related requests

### LINKED ISSUE #7036: bug: Identifier is not stored when using Number Pools
### Component

API Server / GraphQL

### Infrahub version

v1.3.4

### Current Behavior

Currently on demo.infrahub.app If I have a number pool and want to assign a number pool using an identifier so that new numbers are identifiable when specifying the "Identifier" it is then blank in the UI and API. 

### Expected Behavior

I would expect the identifier to be maintained as it is with IP Address and Prefix pools so that the same number can be retrieved. 

### Steps to Reproduce

1. Login to demo.infrahub.app
2. Create a number pool for InfraVlan and the vlan_id field
3. Create a new vlan using the pool
```
mutation MyMutation {
  InfraVLANCreate(
    data: {name: {value: "test"}, vlan_id: {from_pool: {id: "185b9728-1b76-dda7-d13d-106529b1bcd9", identifier: "test-identifier"}}, role: {value: "user"}, status: {value: "active"}}
  ) {
    ok
    object {
      id
    }
  }
}
```
4. Query resource pool
```
query MyQuery($resource_id: String = "") {
  InfrahubResourcePoolAllocated(
    pool_id: "185b9728-1b76-dda7-d13d-106529b1bcd9"
    resource_id: $resource_id
  ) {
    count
    edges {
      node {
        branch
        display_label
        id
        identifier
        kind
      }
    }
  }
}
```
5. Response has identifier blank
```
{
  "data": {
    "InfrahubResourcePoolAllocated": {
      "count": 1,
      "edges": [
        {
          "node": {
            "branch": "test",
            "display_label": "100",
            "id": "185b972f-4c29-7b92-d13d-1065bc2cab68",
            "identifier": null,
            "kind": "InfraVLAN"
          }
        }
      ]
    }
  }
}
```

### Additional Information

_No response_

polmichel:
# Why
**Problems statement:** 

1. When querying attributes whose value has been allocated by a `NumberPool`, the identifier is not returned.

### Query

```graphql
query MyQuery($resource_id: String = "") {
  InfrahubResourcePoolAllocated(
    pool_id: "185b9728-1b76-dda7-d13d-106529b1bcd9"
    resource_id: $resource_id
  ) {
    count
    edges {
      node {
        branch
        display_label
        id
        identifier
        kind
      }
    }
  }
}
```

### Reponse

```
"edges": [
  {
    "node": 
    {
      "branch": "test",
      "display_label": "100",
      "id": "185b972f-4c29-7b92-d13d-1065bc2cab68",
      "identifier": null,
      "kind": "InfraVLAN"
    }
  }
]
```

This query `infrahub.core.query.resource_manager.NumberPoolSetReserved.query_init` sets an identifier into the IS_RESERVED relationship , which can be either a custom identifier or the uuid.

Currently, when one runs the `InfrahubResourcePoolAllocated` GraphQL query, no identifier is returned `infrahub.core.query.resource_manager.NumberPoolGetAllocated.query_init`.

**Goal:** Add the IS_RESERVED identifier to the return of the Cypher query used to retrieve information from the NumberPools allocations.

### What is not included

Custom identifiers management on the number pool allocations.

Closes [IFC-2297]
Partially fix https://github.com/opsmill/infrahub/issues/7036

## What changed

1. `infrahub.core.query.resource_manager.NumberPoolGetAllocated` Cypher query, which matchs an additional relationship in order to return the identifier attached to this relationship.

## How to review

- Cypher query: `infrahub.core.query.resource_manager.NumberPoolGetAllocated` alongside with performance consideration
- Impact on the resolver `infrahub.graphql.queries.resource_manager.PoolAllocated.resolve`

## How to test

```bash
uv run invoke backend.test-unit --path backend/tests/component/graphql/queries/test_resource_pool.py::test_number_pool_utilization                                                                           
   
uv run invoke backend.test-unit --path backend/tests/component/core/resource_manager/test_number_pool_query.py::test_NumberPoolGetAllocated_returns_identifier  
```

## Impact & rollout

- **Performance:** Additional relationships are matched in the query.

## Checklist

- [x] Tests added/updated

[IFC-2297]: https://opsmill.atlassian.net/browse/IFC-2297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * GraphQL number-pool allocation responses now include resource identifiers so allocated items display the correct identifier.

* **Tests**
  * Added comprehensive tests validating allocation queries, identifier correctness after delete/reallocate, empty-state behavior, and cross-branch scenarios.

* **Chore**
  * Added changelog entry noting the identifier fix.

### DIFF
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -19,14 +19,6 @@ class NumberPoolIdentifierData:
     value: int
     identifier: str
 
-    @classmethod
-    def from_db(cls, result: QueryResult) -> NumberPoolIdentifierData:
-        """Convert raw QueryResult to typed dataclass."""
-        return cls(
-            value=result.get_as_type("value", return_type=int),
-            identifier=result.get_as_type("identifier", return_type=str),
-        )
-
 
 @dataclass(frozen=True)
 class PoolIdentifierResult:
@@ -304,7 +296,13 @@ class NumberPoolGetReserved(Query):
         Returns:
             List of NumberPoolIdentifierData containing value and identifier.
         """
-        return [NumberPoolIdentifierData.from_db(result) for result in self.get_results()]
+        return [
+            NumberPoolIdentifierData(
+                value=result.get_as_type("value", return_type=int),
+                identifier=result.get_as_type("identifier", return_type=str),
+            )
+            for result in self.get_results()
+        ]
 
     def get_reservations(self) -> Generator[NumberPoolIdentifierData]:
         """Yield reservations as typed dataclass instances.
@@ -435,7 +433,10 @@ class NumberPoolGetUsed(Query):
             NumberPoolIdentifierData for each used value in the pool.
         """
         for result in self.get_results():
-            yield NumberPoolIdentifierData.from_db(result)
+            yield NumberPoolIdentifierData(
+                value=result.get_as_type("value", return_type=int),
+                identifier=result.get_as_type("identifier", return_type=str),
+            )
 
 
 class NumberPoolGetFree(Query):
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -205,6 +205,10 @@ class NumberPoolGetAllocated(Query):
             MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
             WHERE hs_int.status = "active"
                 AND hs_int.to IS NULL
+                AND NOT EXISTS {
+                    MATCH (a)-[hs_deleted:HAS_SOURCE {branch: hs_int.branch, status: "deleted"}]->(pool)
+                    WHERE hs_deleted.from > hs_int.from
+                }
             RETURN true AS hs_active
             LIMIT 1
         }
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -60,6 +60,9 @@ class NumberPoolAllocatedResult:
     value: int
     """The allocated number value."""
 
+    identifier: str
+    """Identifier used for the reservation."""
+
     @classmethod
     def from_db(cls, result: QueryResult) -> NumberPoolAllocatedResult:
         """Convert raw QueryResult to typed dataclass."""
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -198,24 +198,42 @@ class NumberPoolGetAllocated(Query):
         )
         self.params.update(branch_params)
 
+        hs_branch_filter, hs_branch_params = self.branch.get_query_filter_path(
+            at=self.at.to_string(), branch_agnostic=self.branch_agnostic, variable_name="hs_int"
+        )
+        self.params.update(hs_branch_params)
+
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
         MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
+        CALL (a) {
+            MATCH (a)-[hs_int:HAS_SOURCE]->(pool:%(number_pool_kind)s)
+            WHERE (%(hs_branch_filter)s) AND hs_int.status = "active"
+            RETURN true AS hs_active
+            LIMIT 1
+        }
+        WITH n, ha, a, hv, av, hs, pool, ir, hs_active
         WHERE
-            pool.uuid = $pool_id
+            hs_active = TRUE
+            AND pool.uuid = $pool_id
             AND av.value >= $start_range and av.value <= $end_range
             AND all(r in [ha, hv, hs] WHERE (%(branch_filter)s))
             AND ha.status = "active"
             AND hv.status = "active"
-            AND hs.status = "active"
         """ % {
             "node": self.pool.node.value,
             "number_pool_kind": InfrahubKind.NUMBERPOOL,
             "branch_filter": branch_filter,
+            "hs_branch_filter": hs_branch_filter,
         }
         self.add_to_query(query)
 
-        self.return_labels = ["n.uuid as id", "hv.branch as branch", "av.value as value", "ir.identifier as identifier"]
+        self.return_labels = [
+            "DISTINCT n.uuid as id",
+            "hv.branch as branch",
+            "av.value as value",
+            "ir.identifier as identifier",
+        ]
         self.order_by = ["av.value"]
 
     def get_data(self) -> list[NumberPoolAllocatedResult]:
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -198,17 +198,17 @@ class NumberPoolGetAllocated(Query):
         )
         self.params.update(branch_params)
 
-        hs_branch_filter, hs_branch_params = self.branch.get_query_filter_path(
-            at=self.at.to_string(), branch_agnostic=self.branch_agnostic, variable_name="hs_int"
-        )
-        self.params.update(hs_branch_params)
-
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
         MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
-        CALL (a) {
-            MATCH (a)-[hs_int:HAS_SOURCE]->(pool:%(number_pool_kind)s)
-            WHERE (%(hs_branch_filter)s) AND hs_int.status = "active"
+        CALL (a, pool) {
+            MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
+            WHERE hs_int.status = "active"
+                AND hs_int.to IS NULL
+                AND NOT EXISTS {
+                    MATCH (a)-[hs_deleted:HAS_SOURCE {branch: hs_int.branch, status: "deleted"}]->(pool)
+                    WHERE hs_deleted.from > hs_int.from
+                }
             RETURN true AS hs_active
             LIMIT 1
         }
@@ -224,7 +224,6 @@ class NumberPoolGetAllocated(Query):
             "node": self.pool.node.value,
             "number_pool_kind": InfrahubKind.NUMBERPOOL,
             "branch_filter": branch_filter,
-            "hs_branch_filter": hs_branch_filter,
         }
         self.add_to_query(query)
 
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -63,16 +63,6 @@ class NumberPoolAllocatedResult:
     identifier: str
     """Identifier used for the reservation."""
 
-    @classmethod
-    def from_db(cls, result: QueryResult) -> NumberPoolAllocatedResult:
-        """Convert raw QueryResult to typed dataclass."""
-        return cls(
-            id=result.get_as_type("id", str),
-            branch=result.get_as_type("branch", str),
-            value=result.get_as_type("value", int),
-            identifier=result.get_as_type("identifier", str),
-        )
-
 
 @dataclass(frozen=True)
 class NumberPoolFreeData:
@@ -242,7 +232,15 @@ class NumberPoolGetAllocated(Query):
         Returns:
             List of NumberPoolAllocatedResult containing allocated number info.
         """
-        return [NumberPoolAllocatedResult.from_db(result) for result in self.get_results()]
+        return [
+            NumberPoolAllocatedResult(
+                id=result.get_as_type("id", str),
+                branch=result.get_as_type("branch", str),
+                value=result.get_as_type("value", int),
+                identifier=result.get_as_type("identifier", str),
+            )
+            for result in self.get_results()
+        ]
 
 
 class NumberPoolGetReserved(Query):
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -70,6 +70,7 @@ class NumberPoolAllocatedResult:
             id=result.get_as_type("id", str),
             branch=result.get_as_type("branch", str),
             value=result.get_as_type("value", int),
+            identifier=result.get_as_type("identifier", str),
         )
 
 
@@ -217,7 +218,7 @@ class NumberPoolGetAllocated(Query):
 
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
-        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)
+        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
         WHERE
             pool.uuid = $pool_id
             AND av.value >= $start_range and av.value <= $end_range
@@ -232,7 +233,7 @@ class NumberPoolGetAllocated(Query):
         }
         self.add_to_query(query)
 
-        self.return_labels = ["n.uuid as id", "hv.branch as branch", "av.value as value"]
+        self.return_labels = ["n.uuid as id", "hv.branch as branch", "av.value as value", "ir.identifier as identifier"]
         self.order_by = ["av.value"]
 
     def get_data(self) -> list[NumberPoolAllocatedResult]:
--- a/backend/infrahub/core/query/resource_manager.py
+++ b/backend/infrahub/core/query/resource_manager.py
@@ -205,10 +205,6 @@ class NumberPoolGetAllocated(Query):
             MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
             WHERE hs_int.status = "active"
                 AND hs_int.to IS NULL
-                AND NOT EXISTS {
-                    MATCH (a)-[hs_deleted:HAS_SOURCE {branch: hs_int.branch, status: "deleted"}]->(pool)
-                    WHERE hs_deleted.from > hs_int.from
-                }
             RETURN true AS hs_active
             LIMIT 1
         }
--- a/backend/tests/component/core/resource_manager/test_number_pool_query.py
+++ b/backend/tests/component/core/resource_manager/test_number_pool_query.py
@@ -191,11 +191,7 @@ async def test_NumberPoolGetAllocated_includes_allocation_active_on_other_branch
     default_branch: Branch,
     run_number_pool_validation: None,
 ) -> None:
-    """When a node created on main is deleted on a branch, NumberPoolGetAllocated still returns it.
-
-    The allocation is still active on main (HAS_SOURCE edge remains active there),
-    so it should appear in the allocated list regardless of the branch-local deletion.
-    """
+    """When a node created on main is deleted on a branch, NumberPoolGetAllocated still returns it."""
     incident_schema = registry.schema.get_node_schema(name=INCIDENT.kind, branch=default_branch)
     incidents = await create_objects(db=db, schema=incident_schema, branch=default_branch.name, start=1, end=3)
     pools: list[CoreNumberPool] = await NodeManager.query(db=db, schema=InfrahubKind.NUMBERPOOL, branch=default_branch)
@@ -223,13 +219,7 @@ async def test_NumberPoolGetAllocated_includes_allocation_when_source_cleared_on
     default_branch: Branch,
     run_number_pool_validation: None,
 ) -> None:
-    """When HAS_SOURCE is cleared on a branch, the allocation still appears because it is active on main.
-
-    Setup: 3 incidents allocated on default branch → values [1, 2, 3].
-    Then on br1, clear the source of incident #2's number attribute.
-    The default branch still has an active HAS_SOURCE edge for incident #2,
-    so the allocation remains visible — it is still allocated from the pool's perspective.
-    """
+    """When HAS_SOURCE is cleared on a branch, the allocation still appears because it is active on main."""
     incident_schema = registry.schema.get_node_schema(name=INCIDENT.kind, branch=default_branch)
     incidents = await create_objects(db=db, schema=incident_schema, branch=default_branch.name, start=1, end=3)
     pools: list[CoreNumberPool] = await NodeManager.query(db=db, schema=InfrahubKind.NUMBERPOOL, branch=default_branch)
--- a/backend/tests/component/core/resource_manager/test_number_pool_query.py
+++ b/backend/tests/component/core/resource_manager/test_number_pool_query.py
@@ -1,8 +1,13 @@
+from unittest.mock import AsyncMock
+
 import pytest
 
 from infrahub.core import registry
 from infrahub.core.branch import Branch
 from infrahub.core.constants import InfrahubKind
+from infrahub.core.diff.coordinator import DiffCoordinator
+from infrahub.core.diff.data_check_synchronizer import DiffDataCheckSynchronizer
+from infrahub.core.diff.merger.merger import DiffMerger
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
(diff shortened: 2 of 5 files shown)


backend/tests/component/core/resource_manager/test_number_pool_query.py
```diff
@@ -115,6 +120,38 @@ async def test_NumberPoolGetUsed(
     assert await get_used_numbers_in_pool(db=db, pool=incident_pool, branch=default_branch) == [1, 3]
 
 
+async def test_NumberPoolGetAllocated_returns_identifier(
+    db: InfrahubDatabase, register_test_schema: SchemaBranch, default_branch: Branch
+) -> None:
+    """Test that NumberPoolGetAllocated returns the identifier (node UUID) from the IS_RESERVED relationship."""
+    # Arrange
```
gmazoyer: Not sure these kind of comments are really valuable

polmichel: Are you speaking about the Arrange/Act/Assert pattern or the comment under the test name?

gmazoyer: Yes the docstring's fine

polmichel: I usually use it to separate the different zones in my tests, but spaces and indentations can also work.
In my opinion, the most important thing is to clearly define these three zones, either through comments or another method -> I can remove them

backend/infrahub/core/node/__init__.py
```diff
@@ -350,7 +350,7 @@ async def handle_pool(
         """Evaluate if a resource has been requested from a pool and apply the resource
 
         This method only works on number pools, currently Integer is the only type that has the from_pool
-        within the create code.
+        within the created code.
```
ajtmccarty: I actually think `create` is correct here. I think this is trying to say that only Integer attributes support the from_pool argument within the payload used to create an new object

but maybe I'm wrong. and it doesn't really matter

backend/infrahub/core/query/resource_manager.py
```diff
@@ -60,13 +60,17 @@ class NumberPoolAllocatedResult:
...
+    """Identifier used for the reservation."""
+
     @classmethod
     def from_db(cls, result: QueryResult) -> NumberPoolAllocatedResult:
         """Convert raw QueryResult to typed dataclass."""
         return cls(
             id=result.get_as_type("id", str),
             branch=result.get_as_type("branch", str),
             value=result.get_as_type("value", int),
+            identifier=result.get_as_type("identifier", str),
```
ajtmccarty: you are correctly following the existing pattern here, but these `from_db` methods violate the Single Responsibility Principle b/c this data transfer class should not need knowledge of how a response from the database is structured. this `from_db` method should be removed and the logic moved into `NumberPoolGetAllocated.get_data`

again, obviously not your fault. and making this update is optional, but I wanted to make sure to point it out so that everyone knows it is wrong going forward

ajtmccarty (commented):
I think the original issue is that we want to support the user passing in a custom identifier for an attribute provisioned from a NumberPool (from the bug `{from_pool: {id: "185b9728-1b76-dda7-d13d-106529b1bcd9", identifier: "test-identifier"}}`) and I'm not sure these changes accomplish this, but there will need to be a component/graphql or integration test for it to make sure

polmichel:
> I think the original issue is that we want to support the user passing in a custom identifier for an attribute provisioned from a NumberPool (from the bug `{from_pool: {id: "185b9728-1b76-dda7-d13d-106529b1bcd9", identifier: "test-identifier"}}`) and I'm not sure these changes accomplish this, but there will need to be a component/graphql or integration test for it to make sure

You are right, I was focused on the identifier being retrieved by the query. I will check the custom identifier management and do a related integration test.

Thanks for the feedback

backend/infrahub/core/node/__init__.py
polmichel: My bad, better to not change it

backend/infrahub/core/query/resource_manager.py
polmichel: I'm always happy to receive this type of feedback. I'll try to split the responsibilities.

backend/tests/component/core/resource_manager/test_number_pool_query.py
polmichel: I'll leave the `# Act` comment if it seems ok to you @gmazoyer, I think this shows the reader where the zone is that actually performs the test.

polmichel:
As the custom identifier management seems to have a larger impact than initially expected, I will limit this PR to the resolution of identifier from GraphQL NumberPools queries, I created a Jira issue related to this new limited perimeter https://opsmill.atlassian.net/browse/IFC-2297. This is a partial fix of https://github.com/opsmill/infrahub/issues/7036.
Custom identifier management will be related to another PR. 

This fix will also unblock the fix regarding this issue on `infrahub-sdk` project, https://opsmill.atlassian.net/browse/IHS-156 / https://github.com/opsmill/infrahub-sdk-python/issues/497

changelog/+identifier-number-pool-queries.fixed.md
```diff
@@ -0,0 +1 @@
+Fix identifier not being returned when querying number pool allocations via GraphQL API.
```
polmichel, line 1: It is a partial fix of #7036. I am not sure if I should mention it into the changelog.

backend/infrahub/core/query/resource_manager.py
```diff
@@ -229,7 +215,7 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
         }
         self.add_to_query(query)
 
-        self.return_labels = ["n.uuid as id", "hv.branch as branch", "av.value as value"]
+        self.return_labels = ["n.uuid as id", "hv.branch as branch", "av.value as value", "ir.identifier as identifier"]
         self.order_by = ["av.value"]
 
     def get_data(self) -> list[NumberPoolAllocatedResult]:
```
ajtmccarty, line 239: this assumes that there is always one and only one `IS_RESERVED` edge for an attribute provisioned from a NumberPool and I am not sure that is the case. If there are multiple `IS_RESERVED` edges on a given value, then this will return multiple rows for a single value (I'm not sure this is possible, but I think it is if you provision an object using an attribute from a pool, delete the object, then provision another object).

it looks like an attribute with a pool source must have at least one reservation, so perhaps handling the 0 reservations case is not completely necessary, but it is better to be extra safe in these cypher queries

so I would say that getting the reservation should be an optional match after this MATCH to the pool and that it should only get the latest reservation. something like the below (untested and might not be correct)
```
CALL (pool, av) {
    OPTIONAL MATCH (pool)-[ir:IS_RESERVED]->(av)
    RETURN ir
    ORDER BY ir.from DESC, ir.status ASC
    LIMIT 1
}
WITH ...
WHERE ir.status = "active" OR ir IS NULL
```

polmichel: Thanks for the feedback @ajtmccarty 

Regarding the multiple match, do we agree that this is only possible because the `IS_RESERVED` relationship is not released when the reservation is released?
Is it true that, if we release it correctly, only one relationship related to this node could be active?

Regarding the zero reservation case, I will keep these corner cases in mind when dealing with Cypher queries.

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
```diff
@@ -48,3 +68,89 @@ async def test_returns_identifier(
         for idx, incident in enumerate(incidents, start=1):
             result = results_by_value[idx]
             assert result.identifier == incident.get_id()
+
+    async def test_identifier_after_delete_and_reallocation(
```
polmichel: This is testing the case where there is a value allocation deletion, which leaves an orphan node.
This is a valid test case anyway, but it will have less value once `IS_RESERVED` relationships deletion handled.

backend/infrahub/core/query/resource_manager.py
```diff
@@ -60,14 +52,8 @@ class NumberPoolAllocatedResult:
...
-    @classmethod
-    def from_db(cls, result: QueryResult) -> NumberPoolAllocatedResult:
-        """Convert raw QueryResult to typed dataclass."""
-        return cls(
-            id=result.get_as_type("id", str),
-            branch=result.get_as_type("branch", str),
-            value=result.get_as_type("value", int),
-        )
+    identifier: str
+    """Identifier used for the reservation."""
```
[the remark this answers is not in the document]
polmichel: The `OPTIONAL MATCH` is used to prevent errors during the Cypher query. From a data modeling perspective, this should not happen. I am unsure whether including this possibility in the response data class is relevant. It makes it "real."

backend/tests/component/core/resource_manager/number_pool_queries/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+    """Helper function to get used numbers in a pool."""
+    query = await NumberPoolGetUsed.init(db=db, branch=branch, pool=pool, branch_agnostic=True)
+    await query.execute(db=db)
+    return sorted([result.value for result in query.iter_results()])
+
+
+async def get_reservations(db: InfrahubDatabase, pool: CoreNumberPool, branch: Branch) -> dict[str, int]:
+    query1 = await NumberPoolGetReserved.init(db=db, pool_id=pool.get_id(), branch=branch)
+    await query1.execute(db=db)
+    return {item.identifier: item.value for item in query1.get_reservations()}
```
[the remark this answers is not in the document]
polmichel: Due to test files reorganization

backend/infrahub/core/query/resource_manager.py
ajtmccarty: we generally do not hard-delete edges from the database (the exception being when an entire branch is deleted), so it would be possible for there to be multiple `IS_RESERVED` edges linking a given `:CoreNumberPool` and a given `AttributeValue`. But there should always be 1 or 0 active (`status="active"`, `from` timestamp < right now, `to` timestamp NULL or > right now) `IS_RESERVED` edges for a given pool and value on particular branch at a particular time.

So yes, only one edge should be active, but we still need to filter it to make sure the inactive ones are ignore

backend/infrahub/core/query/resource_manager.py
```diff
@@ -212,24 +198,34 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
         MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)
+        CALL (pool, av) {
+            OPTIONAL MATCH (pool)-[ir:IS_RESERVED]->(av)
+            RETURN ir
+            ORDER BY ir.from DESC, ir.status ASC
+            LIMIT 1
+        }
+          WITH n, ha, hv, av, hs, pool, ir
```
ajtmccarty: small spacing error here

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
```diff
@@ -0,0 +1,160 @@
...
+        run_number_pool_validation: None,
+    ) -> None:
+        """Test that NumberPoolGetAllocated returns the identifier (node UUID) from the IS_RESERVED relationship."""
+        branch = await create_branch(db=db, branch_name="test_returns_identifier_branch")
+        incident_schema = registry.schema.get_node_schema(name=INCIDENT.kind, branch=branch)
+        incidents = await create_objects(db=db, schema=incident_schema, branch=branch.name, start=1, end=3)
+        pools: list[CoreNumberPool] = await registry.schema.query(
+            db=db, schema=InfrahubKind.NUMBERPOOL, branch=branch.name
+        )
+        incident_pool = next(pool for pool in pools if pool.node.value == INCIDENT.kind)
```
ajtmccarty: you can include `filter={"node__value": INCIDENT.kind}` in the query call above to handle this filtering and then assert that there is only one result

also, I think it's better to use `NodeManager` directly (ie `await NodeManager.query`) because I hope to remove a lot of these `registry.schema` calls some day

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
```diff
@@ -0,0 +1,160 @@
...
+        db: InfrahubDatabase,
+        register_test_schema: SchemaBranch,
+        run_number_pool_validation: None,
+    ) -> None:
+        """After deleting a node and reallocating the same number, the identifier should point to the new node.
+
+        When a node is deleted, its IS_RESERVED relationship on the pool is not cleaned up (tech debt).
+        When the same number is later reallocated, a second IS_RESERVED relationship is created.
+        NumberPoolGetAllocated handles this via ORDER BY ir.from DESC, ir.status ASC LIMIT 1,
+        which should pick the most recent reservation.
```
ajtmccarty: I don't think this kind of implementation detail belongs in the test description. maybe others disagree
I think the LLMs in general are too into comments

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
ajtmccarty: can you elaborate on what you mean by `orphan node` here? is there a vertex left on the database with no linked edges? if so, we'll need to fix it. it doesn't need to be in this PR, but those orphan nodes eventually come back to cause problems for us

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
```diff
@@ -0,0 +1,160 @@
...
+        # Build a lookup by allocated value
+        results_by_value = {r.value: r for r in results}
+
+        # Each allocated result should have the identifier set to the node UUID
+        for idx, incident in enumerate(incidents, start=1):
+            result = results_by_value[idx]
+            assert result.identifier == incident.get_id()
+
+        # Teardown
+        await branch.delete(db=db)
```
ajtmccarty: I'd be surprised if this is necessary, especially b/c the branches used in the tests have different names

ajtmccarty (commented):
looks pretty good
a lot of good tests in here, which is great

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
polmichel: Sorry for this comment, I missed that one.

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
polmichel: This was to avoid any potential side effects. However, you are right that this is unnecessary given the branch names

backend/tests/component/core/resource_manager/number_pool_queries/test_number_pool_get_allocated.py
polmichel: This may not be the correct term. I wanted to point out that relationships 'IS_RESERVED' relationships are never deleted. I will change the test in order to make it easier to understand

backend/infrahub/core/query/resource_manager.py
```diff
@@ -212,24 +198,34 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
         )
         self.params.update(branch_params)
 
+        # "OPTIONAL MATCH" handles 0 reservation corner case
+        # "LIMIT 1 & ORDER BY ir.from DESC, ir.status ASC" handles multiple active reservations corner case
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
         MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)
```
ajtmccarty: not your fault, but this should probably be checking if the HAS_SOURCE edge is actually active
```
CALL (a) {
    MATCH (a)-[hs:HAS_SOURCE]->(pool:%(number_pool_kind)s)
    ORDER BY hs.from DESC
    RETURN hs.status = "active" AS hs_active
    LIMIT 1
}
WITH n, ha, a, hv, av, hs
WHERE hs_active = TRUE
```

and I think you can remove the `hs.status = "active"` below
without this change, a deleted HAS_SOURCE attribute could still be treated as active

polmichel: Thanks for the feedback.
I do not understand why this would happen to the `HAS_SOURCE` relationship and not to other relationships, such as `HAS_VALUE` or `HAS_ATTRIBUTE`. 
From what I understand, we did a similar filter for `IS_RESERVED` relationships because there are a special case, which is mentioned in a dedicated comment over there
`infrahub.core.query.resource_manager.NumberPoolGetUsed`

polmichel: I reached out to Aaron -> _for a given branch at a given time, if an attribute exists on a schema, then there will be one and only one AttributeValue for every instance of that attribute. so we can assume that there is always one current active path on a branch b and time t_

Now, I am trying to create a test that demonstrates the proposed fix for the 'HAS_SOURCE' relationship triage is achieving the desired behavior, which was not possible before the fix.

polmichel: @ajtmccarty fyi I started working on this again this morning, and you can find the integration test illustrating the difference of behavior we were discussion there `tests.component.core.resource_manager.test_number_pool_query.test_NumberPoolGetAllocated_excludes_cleared_source_on_branch`
Commit -> 0a854d78be4ae8b54bc1d0c2187067b37dd9fc83

I will start a discussion in back-end channel in order to discuss what is the correct behavior to support

polmichel:
Last feedbacks are gathered in this squashed commit https://github.com/opsmill/infrahub/pull/8417/changes/b556b7f6da135b45602a7de86c1a37ad3241cab5

backend/infrahub/core/query/resource_manager.py
```diff
@@ -212,24 +198,42 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
-        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)
+        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
+        CALL (a) {
+            MATCH (a)-[hs_int:HAS_SOURCE]->(pool:%(number_pool_kind)s)
+            WHERE (%(hs_branch_filter)s) AND hs_int.status = "active"
+            RETURN true AS hs_active
+            LIMIT 1
+        }
```
ajtmccarty, line 214: determining if an active `HAS_SOURCE` edge exists that links `a` to `pool` on _any_ branch is actually kind of difficult. I think the real fix is probably to make this `HAS_SOURCE` relationship branch-agnostic, but that feels like a much larger change that this is supposed to be. I think the best fix for this case is to just check if any `HAS_SOURCE` edge links the two vertices and is (`status="active"` and `to IS NULL`).

you could add another `EXISTS {}` to the `WHERE` clause that checks for a `status="deleted"` edge on the same branch as the active one with a later from time and I think that would exclude all deleted `HAS_SOURCE` edges correctly.

there's also one more case that I don't think we can reasonably cover without making these `HAS_SOURCE` edges branch-agnostic. given the following scenario: I create a `HAS_SOURCE` edge on the default branch, I create a new branch `br1`, I delete the `HAS_SOURCE` edge on the default branch. in this case, the `HAS_SOURCE` edge is still active on `br1` and we don't have a good way to check for this case within this query without it getting way too big for its simple purpose

in conclusion, something like this 
```
  CALL (a, pool) {
    MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
    WHERE hs_int.status = "active"
    AND hs_int.to IS NULL
    AND NOT EXISTS {
        MATCH (a)-[hs_deleted:HAS_SOURCE {branch: hs_int.branch, status: "deleted"}]->(pool)
        WHERE hs_deleted.from > hs_int.from
    }
    RETURN true AS hs_active
    LIMIT 1
  }
```

polmichel: Thanks for the feedback.

About the `hs_int to IS NULL`, I do agree. This condition was a part of the agnostic branch filter. I have added an integration test regarding this: `tests.component.core.resource_manager.test_number_pool_query.test_NumberPoolGetAllocated_requires_to_is_null_after_cross_branch_and_same_branch_clear`.

About the `NOT EXISTS` I am not sure how such situation would happen. It would mean that we can actually create two relations which are 
- active
- on the same branch
- with a `to` value to NULL

- BUT which would have different `from` values?

polmichel: And would the `LIMIT 1 return True` would not be equivalent to an `EXISTS`?

```
 MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
        WHERE
            pool.uuid = $pool_id
            AND av.value >= $start_range and av.value <= $end_range
            AND all(r in [ha, hv, hs] WHERE (%(branch_filter)s))
            AND ha.status = "active"
            AND hv.status = "active"
            AND EXISTS {
                MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
                WHERE hs_int.status = "active"
                    AND hs_int.to IS NULL
                    AND NOT EXISTS {
                        MATCH (a)-[hs_deleted:HAS_SOURCE {branch: hs_int.branch, status: "deleted"}]->(pool)
                        WHERE hs_deleted.from > hs_int.from
                    }
            }
```

ajtmccarty: the scenario I am considering is 

1. attr.source = pool
2. attr.source = null
3. attr.source = pool again

so there could be an old "deleted" edge that should be ignored b/c the `HAS_SOURCE` edge is active again

I _think_ that your query would work, but putting an `EXISTS` subquery within an `EXISTS` subquery seems unnecessarily confusing

polmichel: Ok for the `EXISTS` part.

Regarding the scenario, If I understood correctly, I just tried to make it happen in an integration test, but the `NOT EXISTS` part does not change the test result -> see `tests.component.core.resource_manager.test_number_pool_query.TestNumberPoolGetAllocated.test_NumberPoolGetAllocated_includes_allocation_when_source_cleared_then_reassigned_on_branch`

BUT, I managed to trigger the `NOT EXISTS` condition in another test which forks a branch, clears the source on this branch and merges it back into `main` branch -> `tests.component.core.resource_manager.test_number_pool_query.TestNumberPoolGetAllocated.test_NumberPoolGetAllocated_excludes_allocation_after_merging_cleared_source`

backend/infrahub/core/query/resource_manager.py
```diff
@@ -214,22 +200,39 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
 
         query = """
         MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]-(a:Attribute {name: $node_attribute})-[hv:HAS_VALUE]-(av:AttributeValueIndexed)
-        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)
+        MATCH (a)-[hs:HAS_SOURCE]-(pool:%(number_pool_kind)s)-[ir:IS_RESERVED]->(av)
+        CALL (a, pool) {
+            MATCH (a)-[hs_int:HAS_SOURCE]->(pool)
+            WHERE hs_int.status = "active"
+                AND hs_int.to IS NULL
```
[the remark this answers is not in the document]
polmichel: The goal is to look for _active_ relationship on _any_ branch

backend/tests/component/core/resource_manager/test_number_pool_query.py
```diff
@@ -83,76 +96,371 @@ async def get_reservations(db: InfrahubDatabase, pool: CoreNumberPool, branch: B
...
+
+    async def test_NumberPoolGetAllocated_includes_allocation_when_source_cleared_then_reassigned_on_branch(
+        self,
+        db: InfrahubDatabase,
+        register_test_schema: SchemaBranch,
+        default_branch: Branch,
+        run_number_pool_validation: None,
+    ) -> None:
+        """Scenario: on a child branch, source=pool -> source=null -> source=pool again; then the
+        main-branch active edge is closed so that it cannot satisfy the query on its own."""
```
ajtmccarty, line 355: `owh` -> `own`

ajtmccarty (approved):
looks good
this might query might have the most tests per lines of cypher in the codebase now

backend/tests/component/core/resource_manager/test_number_pool_query.py
polmichel: I think this is a display issue, this is written `own`

PR #8420, opened 2026-02-19
Fix changelog missing `CREATED` events for pools and templates

### LINKED ISSUE #7268: bug: No Activity Logs When Creating IP or Prefix from Resource Pool
### Component

API Server / GraphQL

### Infrahub version

1.4.5

### Current Behavior

When creating an IP or a prefix from a resource pool, there are no activity logs generated on the resource itself or its parent. This makes it difficult to track changes or audit resource allocations.

### Expected Behavior

Activity logs should be generated on both the resource and its parent when an IP or prefix is created from a resource pool.

### Steps to Reproduce

Create an IP or prefix from a resource pool.
Check the activity logs for the created resource and its parent.
Observe that no entries are present for the creation action.

### Additional Information

_No response_

gmazoyer:
# Why

When objects are created as side effects during node creation, resources allocated from pools (IP addresses, prefixes, number pools) or child objects instantiated via templates, no `CREATED` changelog events are emitted. This means the changelog is incomplete: users see the parent object appear but have no record of the related objects that were created alongside it.

This PR ensures every side-effect object created during a node creation (pool allocations, template children, nested combinations of both) emits a correlated `NodeCreatedEvent`.

Closes #7268

## What changed

- New `NodeCreationContext` (contextvars-based tracker) that records every side-effect node created during a `create_node()` call. The context is set on the resulting `Node.creation_context` property so callers can inspect it.
- Side-effect recording in pool allocators and template relationship
  handling, they call `NodeCreationContext.record_if_active()` after creating objects.
- `generate_node_mutation_events()` accepts a `side_effect_nodes` list and emits a `NodeCreatedEvent` for each, with `EventMeta.from_parent()` to correlate them to the parent creation.
- Fixed `user_id` propagation for pool allocations, IP address and prefix pool resource managers now receive and forward `user_id` so allocated objects are attributed to the requesting user.

### Suggested review order
1. `core/creation_context.py` — the new context tracking primitive (not the best place, but could not find a proper location while avoiding circular dependency)
2. `core/node/create.py` — where the context is instantiated and side effects recorded
3. `pools/default_allocator.py` + `core/relationship/model.py` — recording at allocation sites
4. `events/generator.py` + `graphql/mutations/main.py` — event emission changes
5. Tests in `test_resource_manager.py` and `test_template_pool_allocation.py`

## How to review

- `core/creation_context.py` is the new module
- The context is always scoped (entered/exited via `with` in `create_node`).
- Calling `record_if_active()` at the right spots and threading `side_effect_nodes` through to event generation.

## How to test

```bash
# Run the component tests for pool allocation events
uv run pytest backend/tests/component/graphql/mutations/test_resource_manager.py -v

# Run the component tests for template + pool creation context
```

Key scenarios covered:
- Creating an object that draws an IP address/prefix from a pool → parent + child `CREATED` events
- Creating an object from a template with pool-allocated relationships → all children and pool resources recorded as side effects
- Idempotent re-allocation (same identifier) → no duplicate events
- Pool exhaustion during template creation → proper error, no partial events

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Ambient creation context that records nodes created as side effects and is attached to created objects.
  * User context threaded into resource allocation and relationship resolution so allocations respect caller scope.
  * GraphQL helper to emit mutation events, including side-effect creations, immediately after allocation.

* **Bug Fixes**
  * Fixed missing changelog/mutation events for side-effect nodes (pool allocations and template-generated children).

* **Tests**
  * Expanded tests to validate side-effect tracking, event emission, and background delivery.


backend/infrahub/core/node/__init__.py
```diff
@@ -108,6 +109,8 @@ def __init__(self, schema: NodeSchema | ProfileSchema | TemplateSchema, branch:
         self._attributes: list[str] = []
         self._relationships: list[str] = []
         self._node_changelog: NodeChangelog | None = None
+        self._active_user_id: str = SYSTEM_USER_ID
+        self.creation_context: NodeCreationContext | None = None
```
ogenstad: I'd use a private variable here as we're in the node class and anything public would introduce a new conflict with regards to field names (attributes and relationships) that people can use within their schemas.

backend/infrahub/core/node/__init__.py
```diff
@@ -150,6 +153,10 @@ def get_id(self) -> str:
 
         raise InitializationError("The node has not been saved yet and doesn't have an id")
 
+    def get_user_id(self) -> str:
+        """Return the active user ID associated with the node."""
+        return self._active_user_id
```
ogenstad: A bit similar to the thing above with the name. It is nice with public methods on a class when we call it from other classes though. Wondering if we should have some sort of internal name like "infrahub_node" which could contain other helper methods etc. I think the name "get_user_id" is probably fairly safe with regards to conflicts from user schemas but we never know..

backend/infrahub/core/node/create.py
```diff
@@ -286,38 +297,43 @@ async def create_node(
     lock_names = get_lock_names_on_object_mutation(node=preview_obj, schema_branch=schema_branch)
 
     obj: Node
+    creation_context = NodeCreationContext()
```
ajtmccarty, line 304: can this be passed into `_do_create_node`? might make this easier to read with the nested context managers

backend/infrahub/core/relationship/model.py
```diff
@@ -573,9 +574,13 @@ async def resolve(self, db: InfrahubDatabase, at: Timestamp | None = None, field
                 if hfid_str:
                     data_from_pool["identifier"] = f"hfid={hfid_str} rel={self.name}"
 
-            assigned_peer: Node = await pool.get_resource(db=db, branch=self.branch, at=at, **data_from_pool)  # type: ignore[attr-defined]
+            user_id = self._node._active_user_id if self._node else SYSTEM_USER_ID
```
ajtmccarty: is it at all possible to pass `user_id` down from everywhere `resolve()` is called?

backend/infrahub/core/creation_context.py
```diff
@@ -0,0 +1,44 @@
...
+        return _creation_context.get(None)
+
+    @classmethod
+    def record_if_active(cls, node: Node) -> None:
+        ctx = cls.get_current()
+        if ctx is not None:
+            ctx.record(node)
+
+    def __enter__(self) -> Self:
+        self._token = _creation_context.set(self)
```
ajtmccarty, line 41: I'm surprised that this doesn't need to be defined in the class vars somehow

backend/infrahub/core/node/__init__.py
```diff
@@ -108,6 +109,8 @@ def __init__(self, schema: NodeSchema | ProfileSchema | TemplateSchema, branch:
         self._attributes: list[str] = []
         self._relationships: list[str] = []
         self._node_changelog: NodeChangelog | None = None
+        self._active_user_id: str = SYSTEM_USER_ID
+        self._creation_context: NodeCreationContext | None = None
```
ajtmccarty: I can accept that this approach is the best way to fix the bug right now, but I can't approve this without letting everyone know that I hate it
it is more confusing side effects on top of the pile of confusing side effects. we really need to refactor this whole class

ajtmccarty (approved):

backend/infrahub/core/node/__init__.py
gmazoyer: I 100% agree, I can acceot this fix as it is because it is simply a pragmatic fix while a broader `Node` refactor is planned.

backend/infrahub/core/relationship/model.py
gmazoyer: It is possible. Found one place where I'm not sure if passing the `user_id` is relevant, that place is the call to `resolve_relationships` from `NodeConstraintRunner`. That would have required to change the `NodeConstraintRunner.check` method signature to pass a `user_id` and update all its calls, but not sure this is useful.

backend/infrahub/core/node/create.py
```diff
@@ -155,7 +161,8 @@ async def allocate_from_resource_pools(
         if not pool:
             continue
 
-        allocated_resource = await pool.get_resource(db=db, branch=branch, identifier=obj.id, at=at)  # type: ignore
+        allocated_resource = await pool.get_resource(db=db, branch=branch, identifier=obj.id, at=at, user_id=user_id)  # type: ignore
```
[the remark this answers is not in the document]
gmazoyer: You are missing the fact that here we deal with relationship members allocated from pool, NumberPools are only used for attributes, so they don't apply here.

ajtmccarty (approved):

PR #8427, opened 2026-02-20
make profiles relationship on schemas virtual

### LINKED ISSUE #8426: bug: profiles SchemaRelationship should not be saved to database
### Component

API Server / GraphQL

### Infrahub version

1.7.4

### Current Behavior

Schemas with `generate_profile=True` can end up saving the `SchemaRelationship` object for the `profiles` relationship to the database when it should always be a virtual relationship. So it should never be saved to the DB and we should generate it whenever loading a Node/GenericSchema with `generate_profile=True`

This would generally not be an issue, but it can cause schema integrity failures if the relationship from ` SchemaNode` or `SchemaGeneric` to a `profiles` `SchemaRelationship` is deleted when the `SchemaRelationship` itself remains orphaned and active.

### Expected Behavior

`profiles` `SchemaRelationship` object should never be saved and should also exist only in-memory

### Steps to Reproduce

1. load a schema with `generate_profile=True` (this is the default value) on the default branch
2. the `profiles` `SchemaRelationship` will exist on the database at this point, but that does not cause an error for any users
3. update that schema to be `generate_profile=False`
4. node the `profiles` `SchemaRelationship` is orphaned b/c its relationship to its `SchemaNode/Generic` is closed
5. trying to merge any branch or create a proposed change will raise errors during the schema integrity check b/c the `SchemaRelationship`'s `node` relationship is mandatory and cardinality=one, so it will fail min_count/optional validation

### Additional Information

_No response_

ajtmccarty:
# Why

Found orphaned and incorrect `profiles` SchemaRelationship instances while fixing a different issue on an Infrahub system. `profiles` schema relationships are "virtual", meaning we generate them when processing a schema after it is loaded and they should never be saved to the database. We will generate `profiles` and `object_template` relationships in memory every time. In certain circumstances, these virtual relationships could be saved to the database and then orphaned from their parent SchemaGeneric/Node, which would cause schema integrity errors when trying to merge a branch.

## Goals
- All `profiles`, `object_template` and orphaned SchemaRelationships are deleted from the database in a migration
- Schema loading/saving/processing logic is updated to never save any of these virtual relationships

## Non-goal
Other profile-related schema issues, such as #8445
We are _always_ creating the `profiles` Relationship for a schema, even if `generate_profiles` is False. This is not an issue for the end user b/c they don't have a way to link a profile to a schema with `generate_profiles=False` and, even if they did, it would not matter b/c it would not be used.
I tried to fix this, but the frontend sends queries that expect this `profiles` relationship to exist in some places where it would be deleted. So it is safer to leave it as is on `stable` and make the change on a different branch with a little more planning

Closes #8426  \ IFC-2293

## What changed

- updates to schema loading to add a new check for whether a relationship is virtual or not and, if it is, to not save it to the database
- updates to existing tests to make sure we do not save virtual relationships

## Checklist

- [x] Tests added/updated
- ~External docs updated (if user-facing or ops-facing change)~
- ~Internal .md docs updated (internal knowledge and AI code tools knowledge)~

## Summary by CodeRabbit

* **New Features**
  * Schema handling treats certain relationships as virtual (kept in-memory and not persisted).

* **Bug Fixes**
  * Generated "profiles" and "object_template" schema relationships are no longer saved to the database.
  * Added a migration to remove any previously persisted virtual relationships.

* **Tests**
  * Added validation helpers and test steps to assert no virtual schema relationships remain after operations.

* **Chores**
  * Updated changelog entry documenting the change and migration.


gmazoyer:
`OBJECT_TEMPLATE_RELATIONSHIP_NAME` which is intended for the relationship between an object created from a template and the template itself probably needs the same treatment.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -645,6 +644,8 @@ def process_validate(self) -> None:
     def process_post_validation(self) -> None:
         self.cleanup_inherited_elements()
         self.add_groups()
+        self.manage_object_template_relationships()
+        self.manage_profile_relationships()
```
ajtmccarty, line 648: moved down here so that their order weight is always last and they won't be included in inheritance
now that these relationship schemas are not saved to the database, we will always load them last to keep the field order from changing

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1743,6 +1744,9 @@ def process_inheritance(self) -> None:
                     generics_used_by[generic_kind].append(node.kind)
                 node.inherit_from_interface(interface=generic_kind_schema)
 
+                if generic_kind_schema.generate_profile and not node.generate_profile:
+                    node.generate_profile = True
```
ajtmccarty: if a node inherits from a generic with `generate_profile=True`, then it must also generate profiles or it breaks the graphql schema

backend/tests/component/core/migrations/graph/test_034.py
```diff
@@ -37,7 +37,7 @@ async def test_migration_034(
         # verify that attributes and relationships are still present on all branches for both car and person
         for branch in [branch1, branch2, default_branch]:
             for schema in [car_generic, person_node]:
-                schema_field_ids = [field.get_id() for field in schema.attributes + schema.relationships]
```
ajtmccarty, line 40: had to make this change because `profiles` and `object_template` relationships will never have an ID now

backend/infrahub/core/schema/manager.py
```diff
@@ -451,13 +472,56 @@ async def update_node_in_db_based_on_diff(
...
                 getattr(obj, key).value = value
 
-        attribute_schema = self.get_node_schema(name="SchemaAttribute", branch=branch)
-        relationship_schema = self.get_node_schema(name="SchemaRelationship", branch=branch)
-
         new_node = node.duplicate()
-
         diff_attributes = diff.changed.get("attributes")
         diff_relationships = diff.changed.get("relationships")
+
```
ajtmccarty, line 478: had to move this chunk around b/c the `update_node_in_db_based_on_diff` method ticked over our max allowed complexity after my changes to add a couple of `self._is_virtual_relationship` conditionals

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2468,7 +2469,7 @@ def manage_object_template_relationships(self) -> None:
                 self.set(name=node_name, schema=node_schema)
 
     def add_relationships_to_template(self, node: NodeSchema | GenericSchema) -> None:
-        template_schema = self.get(name=self._get_object_template_kind(node_kind=node.kind), duplicate=False)
+        template_schema = self.get(name=self._get_object_template_kind(node_kind=node.kind), duplicate=True)
```
ajtmccarty, line 2472: seems like the ordering change of the schema processing revealed that we might be changing a schema in the registry in place in this method

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2670,7 +2671,12 @@ def manage_object_template_schemas(self) -> None:
                 or node.state == HashableModelState.ABSENT
             ):
                 try:
-                    node.relationships = [r for r in node.relationships if r.name != OBJECT_TEMPLATE_RELATIONSHIP_NAME]
```
ajtmccarty, line 2673: seems like the ordering change of the schema processing revealed that we might be changing a schema in the registry in place in this method

gmazoyer (approved):

PR #8447, opened 2026-02-23
Do not provide diff refresh actions for deleted branches

pa-lem:
image

## Summary by CodeRabbit

* **New Features**
  * Detects branch existence across the UI and exposes it to diff and empty-state views.
  * New hook to query branch presence for conditional rendering.

* **Bug Fixes**
  * UI now updates immediately after single or bulk branch deletions so deleted branches disappear from lists and queries.
  * Diff view shows a specific "branch not found" state and hides refresh/rebase actions when branch is missing.

### DIFF
--- a/frontend/app/src/entities/branches/domain/use-branch-exists.ts
+++ b/frontend/app/src/entities/branches/domain/use-branch-exists.ts
@@ -1,0 +1,9 @@
+import { useAtomValue } from "jotai";
+
+import { branchesState } from "@/entities/branches/stores";
+
+export function useBranchExists(branchName?: string): boolean {
+  const branches = useAtomValue(branchesState);
+  if (!branchName) return false;
+  return branches.some((b) => b.name === branchName);
+}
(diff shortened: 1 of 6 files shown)


frontend/app/src/entities/branches/domain/use-branch-exists.ts
```diff
@@ -0,0 +1,9 @@
+import { useAtomValue } from "jotai";
+
+import { branchesState } from "@/entities/branches/stores";
+
+export function useBranchExists(branchName?: string): boolean {
+  const branches = useAtomValue(branchesState);
+  if (!branchName) return false;
+  return branches.some((b) => b.name === branchName);
+}
```
bilalabbad, line 9: if branch is not in 1st pagination, it will wrongly returns false right?

bilalabbad: Because of that, I'm wondering if we should use useGetBranchDetails instead, and check if that is falsy

pa-lem: it comes from getAllBranches which doesn't use pagination

bilalabbad (approved):

PR #8448, opened 2026-02-23
WIP: Improve search anywhere by adding a new page

BeArchiTek:
- Add offset parameter to GraphQL search resolver for server-side pagination
- Collect all matches before applying offset/limit to return an accurate count
- Create a dedicated /search results page with results grouped by node kind
- Add "View all results" footer to search-anywhere dialog
- Add search results query hooks, types, and UI components
- Include backend component tests and frontend e2e test
- Add feature spec documents and changelog entry

## Summary by CodeRabbit

* **New Features**
  * Enhanced search dropdown: scrollable list (up to 10 items), footer shows total match count and "View all results" link.
  * Dedicated full search results page (/search) with results grouped by type, sortable tables, and per-kind expansion.
  * Pagination: offset + limit support and true total-match reporting.
  * Permission-aware filtering: users see only allowed node types.

* **Tests**
  * New unit and end-to-end tests for pagination, counts, grouping, case sensitivity, and permission filtering.

* **Documentation**
  * Spec, plan, checklist, research, tasks, and changelogs for the enhanced search experience.


backend/infrahub/core/query/node.py
```diff
@@ -2204,57 +2204,69 @@ def __init__(
...
-                "(av.value CONTAINS $search_original OR av.value CONTAINS $search_lower "
-                "OR av.value CONTAINS $search_upper OR av.value CONTAINS $search_title)"
-            )
+        if self.case_insensitive:
+            # True case-insensitive matching using toLower/toString
+            self.params["search_value"] = self.search_value
+            if self.partial_match:
+                search_predicate = "toLower(toString(av.value)) CONTAINS toLower(toString($search_value))"
+            else:
+                search_predicate = "toLower(toString(av.value)) = toLower(toString($search_value))"
```
ajtmccarty: this will be slow on a large database b/c the `toLower(toString(...))` calls mean that we skip the index
the better, but more time-consuming, solution to this is a different indexed value on `AttributeValueIndexed` that used the lower-case string version of the real value
it would require a simple migration and some code changes, but nothing too difficult

backend/infrahub/core/migrations/graph/m064_backfill_value_lower.py
```diff
@@ -0,0 +1,31 @@
...
+    async def validate_migration(self, db: InfrahubDatabase) -> MigrationResult:  # noqa: ARG002
+        return MigrationResult()
+
+    async def execute(self, migration_input: MigrationInput) -> MigrationResult:
+        db = migration_input.db
+        await db.execute_query(
+            query="""
+            MATCH (av:AttributeValueIndexed)
+            WHERE av.value_lower IS NULL
+            SET av.value_lower = toLower(toString(av.value))
```
ajtmccarty: should do this one in transactions b/c there could be a lot of `AttributeValueIndexed` vertices
```
MATCH (av:AttributeValueIndexed)
WHERE av.value_lower IS NULL
CALL (av) {
    SET av.value_lower = toLower(toString(av.value))
} IN TRANSACTIONS
```

backend/infrahub/core/query/node.py
```diff
@@ -335,13 +337,20 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG
             )
         }""" % {"attr_edge": attr_edge_prop_str, "attr_vertex": attr_vertex_prop_str}
 
+        indexed_prop = {
+            "value": "attr.content.value",
+            "is_default": "attr.content.is_default",
+            "value_lower": "attr.content.value_lower",
+        }
+        indexed_prop_list = [f"{key}: {value}" for key, value in indexed_prop.items()]
```
ajtmccarty: I think it is easier to understand to just update the query text to include `value_lower: attr.content.value_lower` instead of using this dictionary and formatting it

backend/infrahub/core/query/node.py
```diff
@@ -2204,57 +2217,61 @@ def __init__(
...
-        # We avoid toLower/toString to allow TEXT index usage
-        if self.partial_match:
-            # Use CONTAINS with multiple case variations to leverage TEXT index
-            search_predicate = (
-                "(av.value CONTAINS $search_original OR av.value CONTAINS $search_lower "
-                "OR av.value CONTAINS $search_upper OR av.value CONTAINS $search_title)"
-            )
+        if self.case_insensitive:
+            # Case-insensitive matching using pre-computed value_lower property (TEXT indexed).
+            # Falls back to toLower(toString(av.value)) for nodes not yet migrated (value_lower IS NULL).
```
ajtmccarty: the fallback should not be necessary if we have the migration

backend/infrahub/core/attribute.py
```diff
@@ -308,13 +308,19 @@ def to_db(self) -> dict[str, Any]:
         data: dict[str, Any] = {"is_default": self.is_default}
         if self.value is None:
             data["value"] = NULL_VALUE
+            data["value_lower"] = NULL_VALUE
```
ajtmccarty: this should either be `NULL_VALUE.lower()` or the migration should be updated to skip `AttributeValueIndexed` where `value = "NULL"`

backend/infrahub/core/attribute.py
```diff
@@ -308,13 +308,19 @@ def to_db(self) -> dict[str, Any]:
...
             if isinstance(serialized_value, str) and not is_large_attribute_type(self.schema.kind):
                 # Perform validation here to avoid an extra serialization during validation step.
                 # Standard non-str attributes (integer, boolean) do not exceed limit size related to neo4j indexing.
                 validate_string_length(serialized_value)
             data["value"] = serialized_value
+            # Pre-compute lowercase value for indexed case-insensitive search
+            if isinstance(serialized_value, str):
+                data["value_lower"] = serialized_value.lower()
+            else:
+                data["value_lower"] = str(serialized_value).lower()
```
ajtmccarty: can probably just use this line and remove the conditional

backend/tests/component/core/test_node_get_list_by_attribute_value_query.py
ajtmccarty, line 1: could probably use more query-level testing for case-insensitive vs not on NodeGetListByAttributeValueQuery

backend/infrahub/core/query/attribute.py
```diff
@@ -428,9 +428,7 @@ async def default_attribute_query_filter(
         else:
             if partial_match:
                 query_filter.append(QueryNode(name="av", labels=[attribute_value_label]))
-                query_where.append(
-                    f"toLower(toString(av.{filter_name})) CONTAINS toLower(toString(${param_prefix}_{filter_name}))"
-                )
+                query_where.append(f"av.value_lower CONTAINS toLower(toString(${param_prefix}_{filter_name}))")
```
ajtmccarty: I'm not sure if this needs to support un-indexed `AttributeValue` vertices, in which case this would fail b/c `AttributeValue` vertices that do not have the `AttributeValueIndexed` label will not have a `value_lower` property

backend/tests/component/graphql/queries/test_search.py
```diff
@@ -414,12 +417,12 @@ async def test_search_anywhere_case_sensitive_enabled(
...
 ) -> None:
-    """Test search with case_sensitive=True uses the optimized query with case variations."""
+    """Test search with case_sensitive=True does exact case matching on av.value."""
     branch.update_schema_hash()
     gql_params = await prepare_graphql_params(db=db, branch=branch)
 
-    # Search with lowercase "john" when case_sensitive=True still finds "John"
-    # because the query uses case variations (lowercase, UPPERCASE, Title Case)
+    # Search with lowercase "john" when case_sensitive=True should NOT find "John"
+    # because case_sensitive=True uses exact match on av.value
```
ajtmccarty: `av.value` shouldn't be mentioned in this test b/c it is a database-level implementation details

backend/tests/component/graphql/queries/test_search.py
ajtmccarty, line 1: the read-only tests could be combined into a class that would run faster b/c it would only need to load the data one time and then it could run the individual read queries in methods on the test class
I'd consider this optional b/c most of the data loaded in the tests use `function` scoped fixtures, so the new test class would probably need to load the data in a different way

ajtmccarty (commented):
this will be a good improvement on the backend. thanks for working on it
the backend stuff looks pretty good to me
it would be good to get it into `develop` early in the release cycle so that it can get a lot of testing to make sure the `value_lower` stuff is working correctly

backend/infrahub/core/attribute.py
```diff
@@ -308,13 +308,16 @@ def to_db(self) -> dict[str, Any]:
...
+            data["value_lower"] = NULL_VALUE.lower()
         else:
             serialized_value = self.serialize_value()
             if isinstance(serialized_value, str) and not is_large_attribute_type(self.schema.kind):
                 # Perform validation here to avoid an extra serialization during validation step.
                 # Standard non-str attributes (integer, boolean) do not exceed limit size related to neo4j indexing.
                 validate_string_length(serialized_value)
             data["value"] = serialized_value
+            # Pre-compute lowercase value for indexed case-insensitive search
+            data["value_lower"] = str(serialized_value).lower()
```
ajtmccarty, line 320: I think we only want to do this for indexed attributes, which would be within the `if isinstance(serialized_value, str) and not is_large_attribute_type(self.schema.kind):` block a few lines up
it probably would not hurt us to save a lower-case value of a very long json blob, but I don't think we would get any value from it b/c it is not indexed to be searched

backend/tests/component/graphql/queries/test_search.py
```diff
@@ -471,3 +474,363 @@ async def test_search_anywhere_case_sensitive_enabled(
...
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"search": "j", "limit": 10, "offset": 1},
+    )
+
+    assert result_with_offset.errors is None
+    assert result_with_offset.data
+    # Count reflects results fetched (capped by db_limit = offset + limit)
+    assert result_with_offset.data["InfrahubSearchAnywhere"]["count"] == 2
+    assert len(result_with_offset.data["InfrahubSearchAnywhere"]["edges"]) == 1
```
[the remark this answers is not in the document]
ajtmccarty: I think this is accurate

backend/tests/query_benchmark/test_search_case_insensitive.py
ajtmccarty, line 1: this feels like a test to validate that the change improves performance and not a test that we need to prevent regressions, so not sure if we need to keep it

ajtmccarty (commented):
backend looks good minus a few comments and the merge conflicts

fatih-acar:
just glancing over there: did we measure how much the database grows after running the included db migration? its not mentioned anywhere, in the worst case the database size will probably double... I don't know if we want to go that way and if we do it's probably a good idea to mention it in the changelog...

PR #8456, opened 2026-02-23
set profiles relationship weight to 99,000

ajtmccarty:
undo a couple lines from #8427 and set a static value for the weight of the profile relationships to try to prevent the `order_weight` of the `profiles` relationship from changing whenever a relationship is added to or removed from a schema

## Summary by CodeRabbit

* **Chores**
  * Adjusted relationship weighting and validation ordering for more consistent schema processing.
  * Optimized branch cleanup to use existing in-memory keys, reducing redundant recomputation.
  * Enhanced merge diagnostics and error reporting to surface detailed schema/branch state on failures.
* **Tests**
  * Added test-time schema registration and deterministic dump helpers for better branch/state validation.
  * Limited integration test runs to a targeted subset for faster, focused troubleshooting.

### DIFF
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -163,12 +163,7 @@ class SchemaBranch:
         Since the object themselves are considered immutable we just need to use the hash from each object to calculate the global hash.
         """
         md5hash = hashlib.md5(usedforsecurity=False)
-        for key, value in sorted(
-            tuple(self.nodes.items())
-            + tuple(self.generics.items())
-            + tuple(self.profiles.items())
-            + tuple(self.templates.items())
-        ):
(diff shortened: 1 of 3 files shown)


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -82,6 +82,7 @@
     "kind": RelationshipKind.PROFILE,
     "cardinality": RelationshipCardinality.MANY,
     "branch": BranchSupportType.AWARE,
+    "order_weight": 99_000,
```
gmazoyer, line 85: Seems like a viable solution. My guess is that we do not have this issue with template because we hard-coded the template relationship order weight to `1`.

gmazoyer (approved):

PR #8464, opened 2026-02-24
IFC-1919  Fix optional-to-mandatory attribute transition

### LINKED ISSUE #7409: bug: cannot change optional attribute with a default value to mandatory
### Component

API Server / GraphQL

### Infrahub version

1.4.10

### Current Behavior

if a schema includes an optional attribute that has a default value, it is not possible to update that attribute to be mandatory

this is even if there are no objects using the schema

### Expected Behavior

it should be possible to update an optional attribute to mandatory if there are either
1. no objects using the schema
2. all objects using the schema have a value set

### Steps to Reproduce

1. load a schema that includes an optional attribute with a default value
2. try to update the attribute to be mandatory with no default value

### Additional Information

the issue seems to be a combination of two things
1. in `SchemaBranch.process_default_values`, we set any attribute with a default value defined to optional
2. we ignore a default_value set to `None` when updating a schema

so it's impossible to remove the default value and anything with a default value gets set to optional

gmazoyer:
# Why

When an attribute has `optional: true` with a `default_value`, it is impossible to later change it to `optional: false` (mandatory) while keeping the default. The `process_default_values()` method in `SchemaBranch` forces `optional=True` on **every** attribute that has a non-null `default_value`, making `optional=False` + `default_value` an impossible combination.

Closes #7409

## What changed

- Override `AttributeSchema.update()` to detect when `default_value` is explicitly set to `None` (via Pydantic's `model_fields_set`) and clear it on the existing attribute. This lets `process_default_values()` skip the attribute, preserving `optional: false`.
- Removed `process_default_values()` from `SchemaBranch.process_pre_validation()` entirely. This method was overriding explicit `optional=False` on any attribute with a `default_value`, preventing the mandatory+default combination.

## How to test

```bash
uv run pytest backend/tests/unit/core/schema/ -x -q
uv run pytest backend/tests/component/core/schema_manager/ -x -q
uv run pytest backend/tests/integration/schema_lifecycle/test_schema_validator_optional_to_mandatory.py -xvs
```

## Impact & rollout

- **Backward compatibility:** This might affect schemas that are already loaded in the database when they are being re-loaded from the API if these schemas use a combination of `optional: false` with a `default_value` set. They are stored as `optional: true` in the database and they'll get updated to `optional: false`.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed changing an optional attribute with a default to mandatory — updates can now explicitly clear preserved defaults.

* **Behavioral Changes**
  * Refined schema processing and cardinality/default handling so optional/default transitions behave predictably.

* **Tests**
  * Added extensive unit, component, and integration tests for optional/default ↔ mandatory/no-default transitions, merging, diffs, and lifecycle flows; minor test renames and fixtures updated.

* **Documentation**
  * Marked many GraphQL input fields as "(required)" for clearer client guidance.

### DIFF
--- a/backend/infrahub/core/schema/attribute_schema.py
+++ b/backend/infrahub/core/schema/attribute_schema.py
@@ -27,6 +27,7 @@ if TYPE_CHECKING:
     from infrahub.core.attribute import BaseAttribute
     from infrahub.core.branch import Branch
     from infrahub.core.constants import BranchSupportType
+    from infrahub.core.models import HashableModel
     from infrahub.core.query import QueryElement
     from infrahub.database import InfrahubDatabase
 
@@ -99,6 +100,18 @@ class AttributeSchema(GeneratedAttributeSchema):
             if isinstance(value, Enum):
                 data[field_name] = value.value
         return data
+
+    def update(self, other: HashableModel) -> Self:
+        super().update(other=other)
+
+        if isinstance(other, AttributeSchema):
+            # Allow explicit clearing of default_value (setting it to None).
+            # The base update() skips None values, but for default_value,
+            # None means "remove the default" which is semantically meaningful.
+            if "default_value" in other.model_fields_set and other.default_value is None:
+                self.default_value = None
+
+        return self
 
     @field_validator("kind")
     @classmethod
--- a/backend/tests/component/core/schema_manager/test_manager_schema.py
+++ b/backend/tests/component/core/schema_manager/test_manager_schema.py
@@ -374,16 +374,17 @@ async def test_schema_branch_process_branch_support(schema_all_in_one) -> None:
     assert criticality.get_attribute(name="description").branch == BranchSupportType.AGNOSTIC
 
 
-async def test_schema_branch_process_default_values(schema_all_in_one) -> None:
+async def test_schema_branch_mandatory_with_default_value(schema_all_in_one) -> None:
     schema = SchemaBranch(cache={}, name="test")
     schema.load_schema(schema=SchemaRoot(**schema_all_in_one))
+    schema.process()
 
-    schema.process_default_values()
-
+    # Attribute with default_value AND explicit optional=True stays optional
     generic = schema.get(name="InfraGenericInterface")
     assert generic.get_attribute(name="mybool").optional is True
     assert generic.get_attribute(name="my_generic_name").optional is False
 
+    # Attribute with default_value AND explicit optional=True stays optional
     criticality = schema.get(name="TestingCriticality")
     assert criticality.get_attribute(name="color").optional is True
 
@@ -392,7 +393,7 @@ async def test_schema_update_optional_with_default_to_mandatory(schema_criticali
     """Load a schema with optional+default, update to mandatory without default."""
     schema = SchemaBranch(cache={}, name="test")
     schema.load_schema(schema=SchemaRoot(**schema_criticality_tag))
-    schema.process_default_values()
+    schema.process()
 
     criticality = schema.get(name="TestingCriticality")
     color_attr = criticality.get_attribute(name="color")
@@ -412,7 +413,7 @@ async def test_schema_update_optional_with_default_to_mandatory(schema_criticali
 
     updated_branch = SchemaBranch(cache={}, name="test")
     updated_branch.load_schema(schema=SchemaRoot(**updated_data))
-    updated_branch.process_default_values()
+    updated_branch.process()
 
     # Merge the update into the original schema
     updated_criticality = updated_branch.get(name="TestingCriticality", duplicate=False)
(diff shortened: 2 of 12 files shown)


backend/infrahub/core/schema/attribute_schema.py
```diff
@@ -100,6 +101,18 @@ def to_dict(self) -> dict:
...
+        super().update(other=other)
+
+        if isinstance(other, AttributeSchema):
+            # Allow explicit clearing of default_value (setting it to None).
+            # The base update() skips None values, but for default_value,
+            # None means "remove the default" which is semantically meaningful.
+            if "default_value" in other.model_fields_set and other.default_value is None:
+                self.default_value = None
+
+        return self
```
ajtmccarty, line 114: looks like the best option given the current state of the code

... we really need to figure out a better way to handle schema loading/updating

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -374,20 +374,126 @@ async def test_schema_branch_process_branch_support(schema_all_in_one) -> None:
...
+    assert updated_color.default_value == "#444444"
+
+    # Verify the diff detects the optional change
+    original_criticality = schema.get(name="TestingCriticality")
+    original_color = original_criticality.get_attribute(name="color")
+    diff = original_color.diff(updated_color)
+    assert "optional" in diff.changed
+
+
+async def test_schema_roundtrip_mandatory_optional_mandatory(schema_criticality_tag) -> None:
```
ajtmccarty: I think this test just does `optional` -> `mandatory`, not roundtrip `mandatory` -> `optional` -> `mandatory`

backend/tests/integration/schema_lifecycle/test_schema_validator_optional_to_mandatory.py
ajtmccarty, line 1: would be good to verify that the height values for the people are updated as expected or not after loading each schema

backend/tests/unit/core/schema/test_attribute_schema.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_attribute_schema_update_default_value_type_distinction() -> None:
+    """Verify AttributeSchema.update() correctly distinguishes default_value=None (clears) from falsy values (preserves)."""
+    attr1 = AttributeSchema(name="val", kind="Text", default_value="hello")
+    attr1.update(AttributeSchema(name="val", kind="Text", default_value=None))
+    assert attr1.default_value is None
+
+    attr2 = AttributeSchema(name="val", kind="Text", default_value="hello")
+    attr2.update(AttributeSchema(name="val", kind="Text", default_value=""))
+    assert not attr2.default_value
```
ajtmccarty: should this assert `== ""`?

ajtmccarty (commented):
I know that we store `is_default` as a property on `AttributeValue` vertices in the database and I _think_ that we handle updating those correctly during schema migrations, but I am not totally sure. I'm also not completely sure if we need that `is_default` property or what it is used for, so it would be good to check if you haven't already

backend/tests/unit/core/schema/test_attribute_schema.py
gmazoyer: Yes, also ruff complains when doing `== ""`, so we could either do `isinstance(attr2.default_value, str) and not attr2.default_value` or add a ignore ruff comment

ajtmccarty: I think ruff is wrong in this case. and some other cases, but definitely this one

gmazoyer:
> I know that we store `is_default` as a property on `AttributeValue` vertices in the database and I _think_ that we handle updating those correctly during schema migrations, but I am not totally sure. I'm also not completely sure if we need that `is_default` property or what it is used for, so it would be good to check if you haven't already

From what I read, the `is_default` property will remain on `AttributeValue` vertices until we re-save the nodes. That means if we go from having a default value to not having one, querying the nodes that were saved with the default value, will still report `is_default=True` until we save those nodes once again. It's probably not great, but I'd say it is not critical too. So if depending on how hard it is to fix, we could address it in a different PR. It's also possible I missed something in the codebase, and we actually already handle this correctly.

I think we use the `is_default` thing in a few places, profiles and templates from what I recall.

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -374,20 +374,196 @@ async def test_schema_branch_process_branch_support(schema_all_in_one) -> None:
...
     schema = SchemaBranch(cache={}, name="test")
     schema.load_schema(schema=SchemaRoot(**schema_all_in_one))
+    schema.process()
 
-    schema.process_default_values()
-
+    # Attribute with default_value AND explicit optional=True stays optional
     generic = schema.get(name="InfraGenericInterface")
     assert generic.get_attribute(name="mybool").optional is True
     assert generic.get_attribute(name="my_generic_name").optional is False
```
polmichel, line 385: I think there should be a blank line or another comment indicating what is actually being tested in the second assertion. It seems to be linked to the first line. 
But, it is maybe to verify that mandatory attributes do not turn to `.optional is true`.

This is really a minor comment.

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -374,20 +374,196 @@ async def test_schema_branch_process_branch_support(schema_all_in_one) -> None:
...
+    updated_branch.load_schema(schema=SchemaRoot(**updated_data))
+    updated_branch.process()
+
+    updated_criticality = updated_branch.get(name="TestingCriticality")
+    updated_color = updated_criticality.get_attribute(name="color")
+    assert updated_color.optional is False
+    assert updated_color.default_value == "#444444"
+
+    # Verify the diff detects the optional change
+    original_criticality = schema.get(name="TestingCriticality")
```
polmichel, line 461: Why is this only verified for difference detection in this unit test?
Could it also be verified in relation to the change in the `default_value` and `optional` in the previous one?

polmichel: Neither do other tests seem to verify this aspect

backend/tests/integration/schema_lifecycle/test_schema_validator_optional_to_mandatory.py
```diff
@@ -0,0 +1,234 @@
...
+            db=db, id=initial_dataset["alice"], kind=TestingPerson, branch=branch.name, raise_on_error=True
+        )
+        assert alice.height.value == 170
+
+        bob = await registry.manager.get_one(
+            db=db, id=initial_dataset["bob"], kind=TestingPerson, branch=branch.name, raise_on_error=True
+        )
+        assert bob.height.value == 180
+
+    async def test_final_validate(self, db: InfrahubDatabase) -> None:
```
polmichel, line 126: Is this test a safeguard to avoid any unwanted state in the database after launching all the other tests in the same class?

backend/tests/unit/core/schema/test_attribute_schema.py
```diff
@@ -0,0 +1,105 @@
...
+
+    update_source = AttributeSchema(name="role", kind="Text", optional=False, default_value=None)
+    assert "default_value" in update_source.model_fields_set
+
+    existing.update(update_source)
+
+    assert existing.optional is False
+    assert existing.default_value is None
+
+    original = AttributeSchema(name="role", kind="Text", default_value="spine", optional=True)
```
polmichel, line 62: Perhaps it would be easier to understand if this line was removed and the following added at the top of the test:
`original = deep.copy(existing)`

backend/tests/unit/core/schema/test_attribute_schema.py
```diff
@@ -0,0 +1,105 @@
...
+    assert existing.optional is True
+    assert existing.default_value == 180
+
+    update_source = AttributeSchema(name="timeout", kind="Number", optional=False, default_value=180)
+    existing.update(update_source)
+
+    assert existing.optional is False
+    assert existing.default_value == 180
+
+    original = AttributeSchema(name="timeout", kind="Number", optional=True, default_value=180)
```
polmichel, line 102: Perhaps it would be easier to understand if this line was removed and the following added at the top of the test:
`original = deep.copy(existing)`

polmichel (commented):
This seems to be well covered! These are just minor (and perhaps subjective) comments, and questions.

ajtmccarty:
thinking some more about how this applies to correctly determining if a given attribute `is_default` or not
currently, we don't do any sort of retroactive handling when the `default_value` of an attribute is updated in a schema. so an attribute created with a default value of `abc` will still be treated as a default value if we later update the default value on the schema to `def`. basically, an attribute will be considered `is_default` if and only if it is created with the default value

SO I think it is okay for this change to ignore updating `is_default` b/c we don't update it when a `default_value` is updated

ajtmccarty (approved):

PR #8470, opened 2026-02-25
make SchemaBranch.process() idempotent

ajtmccarty:
# Problem

Schema hash errors of the type `Schema 'LocationContinent' on branch None has incorrect hash: '93f166b0364aba675b37f07ffe0c000c'` following merging #8427 that made sure to keep `profiles` and `object_template` generated relationship schemas out of the database. These errors have been observed in local and E2E testing

- on profiles schemas (corrected in #8468)
- on hierarchical schemas, specifically `LocationContinent` and `LocationSite` which are at the top and bottom of the hierarchy in our testing schema https://github.com/opsmill/infrahub/actions/runs/22370786493/job/64748971197?pr=8460
- on schemas with `generate_template=True` (`InfraDevice` and `InfraPatchPanel` in our testing schema)

# Fix 

The root of the problem is that calling `SchemaBranch.process()` was not idempotent between the first and second time, so if a schema is loaded and `.process()` is called on the schema two or more times on the API worker handling the schema update, but only one time on all the other workers that just load the schema from the database, then the schema hashes could diverge and cause the `Schema "Abc" has incorrect hash` errors

Unit testing and manual testing against a local Infrahub instance revealed two issues
- the `object_template` Relationship schema was being saved to the database with `max_count=0`, but if it is run through `.process()` two or more times, that `max_count` is set to `1`. this does not happen on the first run through `process()` b/c `process_cardinality_counts()` is called before `manage_object_template_relationship()`

- the original code only created `parent`/`children` relationships when `node.parent` or `node.children` were truthy. an empty string (used for hierarchy edge nodes: `LocationContinent` / `LocationRack`) was falsy and skipped. This caused `process()` to be non-idempotent: on the first call, edge nodes got no parent/children rel. But `add_hierarchy_generic()` added these rels to the generic, and on a second `process()` call, `process_inheritance()` copied them from the generic into the nodes. the fix uses `parent_peer = node.parent or node.hierarchy` (and similar for `children`), so edge nodes always get a parent/children relationship pointing to the hierarchy generic. this matches the behavior when loading a schema due to multiple `process()` calls
# Testing

a couple of new tests that load a schema with all of our schemas that generate other schemas/attributes/relationships during processing and makes sure that `process()` is idempotent when run in memory and when retrieving a schema from the database so hopefully we can avoid this in the future.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect hash errors by making schema validation and processing fully idempotent.

* **Improvements**
  * More consistent handling of parent/child relationships in schema hierarchies to avoid mismatches.
  * Explicit cardinality constraints added for object templates to enforce expected limits.

* **Tests**
  * Added comprehensive idempotency tests and strengthened hierarchical validation tests.

* **Changelog**
  * Added entry describing idempotency and schema-processing fixes.

### DIFF
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2444,6 +2444,8 @@ class SchemaBranch:
                 "cardinality": RelationshipCardinality.ONE,
                 "branch": BranchSupportType.AWARE,
                 "order_weight": 1,
+                "max_count": 1,
+                "min_count": 0,
             }
 
             # Add relationship between node and template
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2213,17 +2213,15 @@ class SchemaBranch:
                 if parent_rel.peer != parent_peer:
                     parent_rel.peer = parent_peer
 
-            if node.children:
-                if "children" not in node.relationship_names:
-                    node.relationships.append(
-                        self._get_hierarchy_child_rel(
-                            peer=node.children, hierarchical=node.hierarchy, read_only=read_only
-                        )
-                    )
-                else:
-                    children_rel = node.get_relationship(name="children")
-                    if children_rel.peer != node.children:
-                        children_rel.peer = node.children
+            children_peer = node.children or node.hierarchy
+            if "children" not in node.relationship_names:
+                node.relationships.append(
+                    self._get_hierarchy_child_rel(peer=children_peer, hierarchical=node.hierarchy, read_only=read_only)
+                )
+            else:
+                children_rel = node.get_relationship(name="children")
+                if children_rel.peer != children_peer:
+                    children_rel.peer = children_peer
 
             self.set(name=node_name, schema=node)
 
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2198,20 +2198,20 @@ class SchemaBranch:
             node = node.duplicate()
             read_only = InfrahubKind.IPPREFIX in node.inherit_from
 
-            if node.parent:
-                if "parent" not in node.relationship_names:
-                    node.relationships.append(
-                        self._get_hierarchy_parent_rel(
-                            peer=node.parent,
-                            hierarchical=node.hierarchy,
-                            read_only=read_only,
-                            optional=node.parent in [node_name] + self.generic_names,
-                        )
+            parent_peer = node.parent or node.hierarchy
+            if "parent" not in node.relationship_names:
+                node.relationships.append(
+                    self._get_hierarchy_parent_rel(
+                        peer=parent_peer,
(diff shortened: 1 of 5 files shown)


ajtmccarty:
can be merged if approved so that it can be released

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2444,6 +2442,8 @@ def manage_object_template_relationships(self) -> None:
                 "cardinality": RelationshipCardinality.ONE,
                 "branch": BranchSupportType.AWARE,
                 "order_weight": 1,
+                "max_count": 1,
+                "min_count": 0,
```
ogenstad, line 2446: While this is not wrong, it also feels a bit redundant for a cardinality=one relationship. Perhaps something we should look into for the schema in general at some point..

ogenstad (approved):

backend/tests/component/core/schema/schema_branch/test_process_idempotency.py
```diff
@@ -0,0 +1,255 @@
...
+    profiles_after_second = sorted(schema_branch.profile_names)
+    templates_after_second = sorted(schema_branch.template_names)
+
+    assert hash_after_first == hash_after_second, (
+        f"get_hash() changed after second process()\n{_describe_hash_diff(snapshot, schema_branch)}"
+    )
+    assert nodes_after_first == nodes_after_second, "node names changed after second process()"
+    assert generics_after_first == generics_after_second, "generic names changed after second process()"
+    assert profiles_after_first == profiles_after_second, "profile names changed after second process()"
+    assert templates_after_first == templates_after_second, "template names changed after second process()"
```
polmichel, line 222: This is not very important, nor is it blocking the merge. However, gathering all the evidence at once might be helpful if other issues arise regarding this test.

````
failures: list[str] = []
if hash_after_first != hash_after_second:
    failures.append(f"get_hash() changed after second process()\n{_describe_hash_diff(snapshot, schema_branch)}")
if nodes_after_first != nodes_after_second:
    failures.append(
        f"node names changed: {sorted(set(nodes_after_second) - set(nodes_after_first))} added, {sorted(set(nodes_after_first) - set(nodes_after_second))} removed"
    )
if generics_after_first != generics_after_second:
    failures.append(
        f"generic names changed: {sorted(set(generics_after_second) - set(generics_after_first))} added, {sorted(set(generics_after_first) - set(generics_after_second))} removed"
    )
if profiles_after_first != profiles_after_second:
    failures.append(
        f"profile names changed: {sorted(set(profiles_after_second) - set(profiles_after_first))} added, {sorted(set(profiles_after_first) - set(profiles_after_second))} removed"
    )
if templates_after_first != templates_after_second:
    failures.append(
        f"template names changed: {sorted(set(templates_after_second) - set(templates_after_first))} added, {sorted(set(templates_after_first) - set(templates_after_second))} removed"
    )
assert not failures, "process() is not idempotent:\n" + "\n".join(failures)
````

PR #8478, opened 2026-02-25
First draft for schema visualizer using infrahub-schema-graph

pa-lem:
image


pa-lem:
will be done in https://github.com/opsmill/infrahub/pull/8582

PR #8479, opened 2026-02-25
Add instructions for Jira reference

ogenstad:
# Why

Update instructions to Spec Kit specify command to better align with how we want to name our folders for Spec Kit specifications.

## What changed

* Instructions to ensure that we prefix the folder names with Jira epic or JPD card along with the short name of the feature
* Update bash script to account for the fact that we don't use numbers
* Also remove the initials for the username from the shell script.

## How to review

* Consider if this better suites our needs
* Try it out

Note that in order to keep this small I've not changed any of the instructions when it comes to creating a local branch. This is to keep the change small.

## Summary by CodeRabbit

* **Documentation**
  * Updated feature creation workflow to require explicit Jira/JPD ticket IDs instead of auto-scanning for branch numbers.
  * Clarified user input requirements for ticket references (JPD: `infp-<digits>`; Jira: `ifc-<digits>`).

* **Refactor**
  * Simplified branch naming scheme to use ticket ID only, removing user initials from branch names.

### DIFF
--- a/.specify/scripts/bash/create-new-feature.sh
+++ b/.specify/scripts/bash/create-new-feature.sh
@@ -306,33 +306,23 @@ if [ -z "$BRANCH_NUMBER" ]; then
     fi
 fi
 
-# Determine initials
-if [ -z "$INITIALS" ]; then
-    if [ "$HAS_GIT" = true ]; then
-        INITIALS=$(get_git_initials)
-        if [ -z "$INITIALS" ]; then
-            echo "Error: Could not determine initials." >&2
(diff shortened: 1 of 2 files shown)


fatih-acar (approved):

gmazoyer:
I hope speckit is going to support git worktree at some point. But it seems this is a known issue for them.

ogenstad:
> I hope speckit is going to support git worktree at some point. But it seems this is a known issue for them.

I think we can just update the markdown files to allow for this ourselves. But it could be that it should be an option, and we could also consider how we actually want branches to be created automatically or what should happen within each feature.

PR #8480, opened 2026-02-25
 #8237 (IFC-2235): Update queries on node using HFID to reference relationships 

### LINKED ISSUE #8237: bug: display_label shows None for relationship fields after upsert via object file
### Component

API Server / GraphQL

### Infrahub version

Latest (1.7.x)

### Current Behavior

When loading data using object files, `display_label` templates that reference relationship fields (e.g., `{{ manufacturer__name__value }}`) work correctly on the **first load** but show `None` for the relationship part on **subsequent loads** (upserts) of the same data.

**Video demonstration:** https://www.loom.com/share/3d5f9394c3bb4c22a92d47eaae1641fb

### Expected Behavior

The `display_label` should consistently render correctly with relationship values, regardless of whether the node is being created or updated via upsert.

### Steps to Reproduce

1. Create schemas with a relationship-based display_label:

**org.yml:**
```yaml
version: "1.0"

generics:
  - name: Generic
    namespace: Organization
    label: Organization
    description: An organization represents a legal entity, a company.
    human_friendly_id:
      - name__value
    order_by:
      - name__value
    icon: mdi:domain
    include_in_menu: true
    attributes:
      - name: name
        kind: Text
        unique: true
        order_weight: 1000

nodes:
  - name: Manufacturer
    namespace: Organization
    description: Device Manufacturer
    icon: mdi:factory
    include_in_menu: true
    inherit_from:
      - OrganizationGeneric
```

**dcim.yml (relevant parts):**
```yaml
version: "1.0"

nodes:
  - name: DeviceType
    namespace: Dcim
    description: A model of device
    label: Device Type
    icon: mdi:poll
    include_in_menu: true
    human_friendly_id:
      - name__value
    display_label: "{{ name__value }} (Manufacturer: {{ manufacturer__name__value }})"
    order_by:
      - manufacturer__name__value
      - name__value
    uniqueness_constraints:
      - [manufacturer, name__value]
    attributes:
      - name: name
        kind: Text
        unique: true
        order_weight: 1000
    relationships:
      - name: manufacturer
        peer: OrganizationManufacturer
        cardinality: one
        kind: Attribute
        order_weight: 1250
        optional: false
```

2. Load the schemas into Infrahub

3. Create an object file:

**objects.yml:**
```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: OrganizationManufacturer
  data:
    - name: Juniper
      description: Juniper Networks
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: DcimDeviceType
  data:
    - name: ex4400-48f
      manufacturer: Juniper
      description: Juniper EX4400-48F Switch
```

4. Load the objects file - **First load works correctly**: DeviceType displays as "ex4400-48f (Manufacturer: Juniper)"

5. Load the same objects file again (upsert) - **Bug**: DeviceType now displays as "ex4400-48f (Manufacturer: None)"

6. The actual relationship is still intact (manufacturer field shows Juniper), only the `display_label` is incorrect.

### Technical Analysis

Based on code investigation, the issue appears to be in how `display_label` is recomputed during upserts:

1. **Update path:** When a node is upserted (updated), the `mutate_update_object` method calls `obj.from_graphql(db=db, data=data)` then `obj.save(db=db, fields=fields)` (`backend/infrahub/graphql/mutations/main.py:254-270`)

2. **display_label recomputation:** During `_update()`, if `display_label` needs updating, it calls `self._display_label.compute(db=db, node=self)` (`backend/infrahub/core/node/__init__.py:971-978`)

3. **Relationship resolution:** The `compute()` method calls `node.get_path_value(db=db, path=variable)` for each template variable, which for relationship fields calls:
   ```python
   relm: RelationshipManager = getattr(self, schema_path.relationship_schema.name)
   await relm.resolve(db=db)
   node = await relm.get_peer(db=db)
   attr = getattr(node, schema_path.attribute_schema.name)
   ```
   (`backend/infrahub/core/node/__init__.py:249-254`)

4. **Potential issue:** The relationship manager's `get_peer()` method may return `None` if the relationship is in an inconsistent state during the update, or if `resolve()` fails to properly load the peer node.

**Key code locations:**
- `backend/infrahub/core/node/__init__.py` - Node update and display_label handling
- `backend/infrahub/core/node/node_property_attribute.py` - DisplayLabel.compute()
- `backend/infrahub/core/relationship/model.py` - RelationshipManager.update() and get_peer()
- `backend/infrahub/graphql/mutations/main.py` - Upsert mutation handling

### Additional Information

This issue also affects:
- Interfaces with display_label showing device: `display_label: "{{ name__value }} (Device: {{ device__name__value }})"`
- Any node type with relationship-based display_label templates

Related issues:
- #7704 (Display label trigger not setup for existing nodes) - similar symptom but different root cause
- #6228 (Feature request for using related node attributes in display_label) - the feature exists but has this bug

polmichel:
# Why

### Problem statement

Update queries recompute and provide incorrect values for computed attributes using related node attributes.
For example, after an update query on an object with the following schema:

```
version: "1.0"

nodes:
  - name: DeviceType
    namespace: Dcim
    description: A model of device
    label: Device Type
    icon: mdi:poll
    include_in_menu: true
    human_friendly_id:
      - name__value
    display_label: "{{ name__value }} (Manufacturer: {{ manufacturer__name__value }})"
    order_by:
      - manufacturer__name__value
      - name__value
    uniqueness_constraints:
      - [manufacturer, name__value]
    attributes:
      - name: name
        kind: Text
        unique: true
        order_weight: 1000
    relationships:
      - name: manufacturer
        peer: OrganizationManufacturer
        cardinality: one
        kind: Attribute
        order_weight: 1250
        optional: false
```
`display_label` will be evaluated as "`name__value`(None)", because `{{ manufacturer__name__value }}` evaluates to `None`.

This is due to the fact that `infrahub.core.node.Node.resolve_relationships` method does not retrieve the required attributes from the relationships when the node is not being created.
This `display_label` recompute is called in the `infrahub.core.node.Node._update` method due to a legitimate call to `infrahub.core.node.node_property_attribute.NodePropertyAttribute.needs_update`.

Closes #8237 / [IFC-2235]

## What changed

Condition that queries additional attributes from the database when resolving the `relationships` in `infrahub.core.node.Node.resolve_relationships`.

**Before this PR,** this was only triggered during node creation.

**After this PR**, this will be called:
- During node creation
- During node update, if it has a display label or a hfid

## How to test

```bash
uv run pytest -xvs backend/tests/component/graphql/test_mutation_upsert.py::test_upsert_preserves_relationship_display_label                                                                                 
```

## Impact & rollout

- **Performance:** We are querying additional attributes from relationships whenever a node is update if it has a display label or a HFID.

## Checklist

- [x] Tests added/updated

[IFC-2235]: https://opsmill.atlassian.net/browse/IFC-2235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed display labels incorrectly showing 'None' for relationship-based fields after upsert operations.

* **Refactor**
  * Streamlined relationship resolution to pre-collect related filters so related attributes are reliably loaded during updates, preserving display labels.

* **Documentation**
  * Added guides on display labels, human-friendly identifiers, and GraphQL mutations; updated Python type-hint/import guidance.

* **Tests**
  * Added test verifying display label preservation across successive upserts.

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -766,11 +766,12 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
 
     async def resolve_relationships(self, db: InfrahubDatabase, user_id: str = SYSTEM_USER_ID) -> None:
         extra_filters: dict[str, set[str]] = {}
+        schema_branch = db.schema.get_schema_branch(name=self.get_branch_based_on_support_type().name)
 
-        if not self._existing:
-            # If we are creating a new node, we need to resolve extra filters from HFID and Display Labels,
-            # if we don't do this the fields might be blank
-            schema_branch = db.schema.get_schema_branch(name=self.get_branch_based_on_support_type().name)
+        # If we are creating a new node, we need to resolve extra filters from HFID,
+        # if we don't do this the fields might be blank.
+        # We could also need it when we need to recompute the HFID
+        if not self._existing or self._human_friendly_id:
             try:
                 hfid_identifier = schema_branch.hfids.get_node_definition(kind=self._schema.kind)
                 for rel_name, attrs in hfid_identifier.relationship_fields.items():
@@ -778,6 +779,10 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
             except KeyError:
                 # No HFID defined for this kind
                 ...
+        # If we are creating a new node, we need to resolve extra filters from Display Labels,
+        # if we don't do this the fields might be blank.
+        # We could also need it when we need to recompute the Display Labels
+        if not self._existing or self._display_label:
             try:
                 display_label_identifier = schema_branch.display_labels.get_template_node(kind=self._schema.kind)
                 for rel_name, attrs in display_label_identifier.relationship_fields.items():
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -53,6 +53,9 @@ if TYPE_CHECKING:
 
     from infrahub.core.branch import Branch
     from infrahub.core.creation_context import NodeCreationContext
+    from infrahub.core.schema.schema_branch import SchemaBranch
+    from infrahub.core.schema.schema_branch_display import TemplateLabel
+    from infrahub.core.schema.schema_branch_hfid import HFIDDefinition
     from infrahub.database import InfrahubDatabase
 
 SchemaProtocol = TypeVar("SchemaProtocol")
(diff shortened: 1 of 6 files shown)


dev/knowledge/backend/display-labels-and-hfid.md
```diff
@@ -0,0 +1,109 @@
+# Display Labels and Human Friendly IDs
+
+> Part of: `dev/knowledge/backend/` | Related: [mutations.md](mutations.md), [architecture.md](architecture.md)
+
+Display labels and HFIDs are computed node properties backed by Jinja2 templates (display labels) or schema path lists (HFIDs). They depend on attributes and relationship peer attributes, requiring special handling during relationship resolution.
```
gmazoyer: This sentence can be misleading, HFID computation is not performed using Jinja2.

dev/knowledge/backend/mutations.md
```diff
@@ -0,0 +1,133 @@
...
+`InfrahubMutationMixin.mutate()` dispatches based on class name:
+
+| Suffix | Method | Flow |
+|--------|--------|------|
+| `Create` | `mutate_create()` | `create_node()` -> `Node.new()` -> `Node.save()` |
+| `Update` | `mutate_update()` | `Node.load()` -> `Node.from_graphql()` -> `Node.save()` |
+| `Upsert` | `mutate_upsert()` | Find-or-create, then create or update path |
+| `Delete` | `mutate_delete()` | Load -> delete |
+
+## Create Flow
```
gmazoyer, line 20: I have no opinion on this yet, but I'm puzzled about mentioning actual code in such files. This specific code, will evolve, I do hope so, and it means we will have to be careful updating these files too.

backend/tests/component/graphql/test_mutation_upsert.py
```diff
@@ -780,3 +782,74 @@ async def test_upsert_with_required_relationship_from_template(
...
+    await _basic_asserts(result_upsert)
+    assert result_upsert.data["TestingTShirtUpsert"]["object"]["id"] == tshirt_id
+    # display_label should still be "Classic Red"
+    assert result_upsert.data["TestingTShirtUpsert"]["object"]["display_label"] == "Classic Red"
+
+
+async def _basic_asserts(result_upsert: ExecutionResult):
+    assert result_upsert.errors is None
+    assert result_upsert.data
+    assert result_upsert.data["TestingTShirtUpsert"]["ok"] is True
```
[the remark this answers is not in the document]
gmazoyer: The bot is right here.

backend/tests/component/graphql/test_mutation_upsert.py
```diff
@@ -780,3 +782,74 @@ async def test_upsert_with_required_relationship_from_template(
     assert tshirt_obj["name"]["value"] == "My Tshirt"
     assert tshirt_obj["color"]["node"]["id"] == color_node.id
     assert tshirt_obj["color"]["node"]["name"]["value"] == "Red"
+
+
+async def test_upsert_preserves_relationship_display_label(
```
gmazoyer, line 787: Nice regression test 👍

backend/infrahub/core/node/__init__.py
```diff
@@ -764,27 +767,36 @@ async def new(self, db: InfrahubDatabase, id: str | None = None, process_pools:
...
-                ...
-            try:
-                display_label_identifier = schema_branch.display_labels.get_template_node(kind=self._schema.kind)
-                for rel_name, attrs in display_label_identifier.relationship_fields.items():
-                    extra_filters.setdefault(rel_name, set()).update(attrs)
-            except KeyError:
-                # No Display Label defined for this kind
-                ...
+    async def resolve_relationships(self, db: InfrahubDatabase, user_id: str = SYSTEM_USER_ID) -> None:
+        schema_branch = db.schema.get_schema_branch(name=self.get_branch_based_on_support_type().name)
```
gmazoyer, line 798: In the old code, `db.schema.get_schema_branch(...)` was only called when `not self._existing`. Now it's called unconditionally on every `resolve_relationships()` invocation. While this is a cheap dict lookup, you could push it inside `_collect_extra_filters` and return early if neither condition applies, avoiding even the dict lookup for nodes without HFID or display_label. But this is a very minor point, the current approach is fine for clarity.

dev/knowledge/backend/mutations.md
polmichel: Thank you for the feedbacks.

This is a subject I would like to discuss with you all. We now have:
- a `/feedback` Claude command that provides insight into where our internal documentation is misleading or lacking. 
- a reminder in the PR checklist to update internal .md documentation

These items make it easier for us to incorporate code into the documentation, as we can expect this command to address the gaps and evolutions.

backend/infrahub/core/node/__init__.py
polmichel: I see your point. I did not want to include the `schema_branch` in the `_collect_extra_filters` method because I did not want the method to depend directly on the database.
I believe this will help with refactoring in future steps.

gmazoyer (approved):

PR #8482, opened 2026-02-25
make sure input number pool is validated for template instances

ajtmccarty:
We were not validating the input `from_pool` ID when it was used on an attribute of a Template instance. End result was that we would link a Number attribute to a non-existent pool in an illegal manner in the database

Now we will validate the input pool ID regardless of whether it is on a Template instance or not

Also found that we were silently swallowing number pool failures when trying to update a Number attribute to use a NumberPool. Fixed by updating `handle_pool` to actually raise errors instead of mutating an `errors` list in place and added a test to cover it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and error handling for number-pool allocations; operations now report clear errors for invalid or missing pool IDs and better handle exhausted pools.
  * Tightened allocation gating to avoid incorrect allocations when pool info is absent or incompatible.

* **Refactor**
  * Internal API updated to clarify pool-allocation control flow.

* **Tests**
  * Added and expanded tests covering invalid pool-ID scenarios for templates and resource creation.

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -623,7 +623,7 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
         if "from_pool" in data:
             self.from_pool = data["from_pool"]
             if process_pools:
-                await self.node.handle_pool(db=db, attribute=self, errors=[])
+                await self.node.handle_pool(db=db, attribute=self)
             changed = True
 
         if changed and self.is_from_profile:
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -350,7 +350,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         self,
         db: InfrahubDatabase,
         attribute: BaseAttribute,
-        errors: list,
+        errors: list[ValidationError],
         allocate_resources: bool = True,
     ) -> None:
         """Evaluate if a resource has been requested from a pool and apply the resource
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -346,8 +346,12 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
 
         return cls(**attrs)
 
-    async def handle_pool(
-        self, db: InfrahubDatabase, attribute: BaseAttribute, errors: list, allocate_resources: bool = True
+    async def handle_pool(  # noqa: PLR0911
+        self,
+        db: InfrahubDatabase,
+        attribute: BaseAttribute,
+        errors: list,
+        allocate_resources: bool = True,
     ) -> None:
         """Evaluate if a resource has been requested from a pool and apply the resource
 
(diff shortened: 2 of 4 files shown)


backend/infrahub/core/node/__init__.py
```diff
@@ -346,8 +346,12 @@ async def init(
 
         return cls(**attrs)
 
-    async def handle_pool(
-        self, db: InfrahubDatabase, attribute: BaseAttribute, errors: list, allocate_resources: bool = True
+    async def handle_pool(  # noqa: PLR0911
```
ajtmccarty: did not want to refactor this for the too-many-returns error b/c hopefully we will move it somewhere else soon

backend/infrahub/core/node/__init__.py
gmazoyer, line 1: Looking at this thing once, more does this means that we currently do not support referencing pool by HFID?

backend/tests/component/graphql/mutations/test_resource_manager.py
```diff
@@ -136,6 +197,79 @@ async def test_create_object_and_assign_prefix_from_pool(
...
+    result = await graphql(
+        schema=gql_params.schema,
+        source=CREATE_TEMPLATE_TICKET_WITH_POOL,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"pool_id": FAKE_POOL_ID},
+    )
+
+    assert result.errors
+    assert FAKE_POOL_ID in str(result.errors[0])
```
polmichel: Why only check for the presence of `FAKE_POOL_ID` in the assertion and not check a larger part of the message as in the previous test? It seems to me that it's the same error message
```
The pool requested {'id': 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee'} was not found. at ticket_id.from_pool

GraphQL request:3:5
2 | mutation CreateTemplateTicketWithPool($pool_id: String!) {
3 |     TemplateTestingTicketCreate(data: {
  |     ^
4 |         template_name: { value: "bad-pool-template" }
```

backend/infrahub/core/node/__init__.py
ajtmccarty: yes that's right. HFIDs do not work as input to `from_pool` right now

backend/infrahub/core/node/__init__.py
```diff
@@ -358,19 +362,25 @@ async def handle_pool(
...
             if attribute.from_pool:
-                attribute.source = attribute.from_pool["id"]
+                number_pool_id = str(attribute.from_pool["id"])
+                attribute.source = number_pool_id
                 attribute.value = None
+                allocate_resources = False
             elif attribute.from_pool is None:
                 attribute.clear_source()
-            return
+                return
```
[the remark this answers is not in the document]
ajtmccarty: updated to catch KeyError and raise ValidationError from ti

backend/tests/component/graphql/mutations/test_resource_manager.py
ajtmccarty: good point
updated

backend/infrahub/core/node/__init__.py
```diff
@@ -358,50 +361,57 @@ async def handle_pool(
...
             number_pool_id = attribute.from_pool.get("id")
             if not number_pool_id:
-                errors.append(ValidationError({f"{attribute.name}.from_pool": "No pool ID specified in from_pool."}))
-                return
+                raise ValidationError({f"{attribute.name}.from_pool": "No pool ID specified in from_pool."})
 
-        if not allocate_resources or not number_pool_id:
+        if not number_pool_id:
             # no pool allocation necessary
             return
```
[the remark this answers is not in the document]
ajtmccarty: this should be blocked by validation at the graphql layer that prevents an empty `dict` in `from_pool`

backend/infrahub/core/attribute.py
```diff
@@ -623,7 +623,7 @@ async def from_graphql(self, data: dict, db: InfrahubDatabase, process_pools: bo
         if "from_pool" in data:
             self.from_pool = data["from_pool"]
             if process_pools:
-                await self.node.handle_pool(db=db, attribute=self, errors=[])
```
ogenstad, line 626: This looks like a hard to spot bug based on how the errors worked before.

ogenstad (approved):

PR #8483, opened 2026-02-25
INFP-403 set order weight on system controlled attributes of the CoreFileObject generic

wvandeun:
# Why

Initially we decided to not show system controlled attributes, for objects inheriting from the CoreFileObject generic, in the UI.

After reviewing and receiving feedback we decided to revert that change.

However we want to make sure that these attributes are de-prioritized in the UI (detailed + list view).

## What changed

set order_weight on system controlled attributes of the CoreFileObject generic.

We will need to revisit the displaying of these attributes after we implement a way to not show attributes like this by default.

## Summary by CodeRabbit

* **Chores**
  * Updated file object attribute ordering configuration for improved consistency in the interface.

### DIFF
--- a/backend/infrahub/core/schema/definitions/core/file_object.py
+++ b/backend/infrahub/core/schema/definitions/core/file_object.py
@@ -16,6 +16,7 @@ core_file_object = GenericSchema(
             description="The name of the file as uploaded by the user",
             read_only=True,
             optional=False,
+            order_weight=99990,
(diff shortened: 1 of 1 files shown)


wvandeun:
@pa-lem @minitriga @BaptisteGi FYI

gmazoyer (approved):

minitriga (approved):

PR #8487, opened 2026-02-26
Add specifications for syslog support

ogenstad:
# Why

Adds Spec Kit specification for syslog support (infp-384)

## What changed

* Add markdown files for spec kit

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive specification documentation for Syslog support, including detailed requirements checklist and feature specifications covering transport options, queue management, observability, and integration guidelines.


dev/specs/infp-384-syslog-support/spec.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Acceptance Scenarios**:
+
+1. **Given** two syslog destinations configured, **When** a user performs a tracked action, **Then** the event is delivered to both destinations.
+2. **Given** two syslog destinations where one becomes unavailable, **When** a tracked action occurs, **Then** events continue to be delivered to the available destination without interruption.
+3. **Given** two destinations with different format settings (one RFC 5424, one RFC 3164), **When** an event occurs, **Then** each destination receives the event formatted according to its own configuration.
+
+---
+
+### User Story 4 - Forward Application Logs to Syslog (Priority: P4)
```
dgarros, line 60: I think this is out of scope for this effort

dev/specs/infp-384-syslog-support/spec.md
```diff
@@ -0,0 +1,176 @@
...
+- Persistent buffering to disk for unavailable TCP destinations is deferred to a future release. The in-memory queue holds events during brief outages, but events are not recovered after a process restart or after the queue fills.
+- In-memory queues are not persistent across process restarts. Messages pending in a worker's queue at the time of a crash or restart are lost. This is a known trade-off accepted for v1.
+- Event ordering at the syslog destination is governed by the syslog timestamp on each message, not by Infrahub insertion order. Messages from concurrent requests served by different worker processes may arrive interleaved at the syslog server; the SIEM is responsible for ordering by timestamp.
+- Each Infrahub worker process maintains its own independent syslog connection(s). The syslog server should be expected to handle one persistent TCP connection per active Infrahub worker process.
```
dgarros, line 157: Not sure this is a desirable ...

dgarros (commented):
Looking at the spec, it's not clear this is an ent feature and it's not clear what will be implemented in the community version vs ent

ogenstad:
> Looking at the spec, it's not clear this is an ent feature and it's not clear what will be implemented in the community version vs ent

We have this part as a requirement:

- **FR-026**: Syslog export MUST be available as an enterprise edition feature; the underlying event pipeline architecture MUST be available in all editions. The enterprise gate MUST be enforced using Infrahub's existing license/edition check mechanism by marking the syslog configuration section as an enterprise feature in the configuration layer.

I think what code lives where might be more of the technical implementation. The way I saw it is that we'd have the core part with a generic interface within the community version but the actual adapter code within enterprise (not completely sure what this will look like from a Spec Kit perspective yet, if we end up just copying the specs or referencing them from the submodule).

Does this sound ok to you?

dev/specs/infp-384-syslog-support/spec.md
ogenstad: We can scratch the part of the maintained TCP connections and perhaps revisit that again if it seems to become a performance issue. It should mainly be a problem if we start to log a huge amount of entries which I don't expect us to do.

dev/specs/infp-384-syslog-support/spec.md
ogenstad: There are some specific things that we won't have as events but still want to log in this way such as when an action is denied. But could perhaps be clarified to state that we don't include the full application logs in this only selected "events" (don't want to reuse the phrase but don't have a better word) where desired?

wvandeun (approved):

PR #8492, opened 2026-02-27
allow updating pools by name

ajtmccarty:
IFC-2300

should follow #8482 

allow specifying resource pools (IP Prefix/Address and NumberPool) by name in the `from_pool` input. the `name` attribute is unique on the `CoreResourcePool` generic

for example 
```
mutation {
    TestingTicketUpdate(data: {
        id: "abc"
        ticket_id: {
            from_pool: {
                id: "name of my NumberPool"
            }
        }
    }) {
        ok
    }
}
```

purposefully chose not to use `NodeManager.get_by_id_or_default_filter` b/c we are trying to not use `default_filter` any longer. instead this uses `is_valid_uuid` and tries to get by name using `NodeManager.query` if the input is not an ID.

added graphql component tests to cover the changes
- creating and updating prefix and address pools using IDs and names on IP relationships
- creating and updating number pools using IDs and names on Number attributes
- negative tests for all the above cases with invalid names

also moved some of the test files around and removed string interpolation from some of the graphql query strings

## Summary by CodeRabbit

* **New Features**
  * Pools can be referenced by name or UUID in GraphQL mutations for creating/updating IP prefixes, IP addresses, and tickets; existing UUID behavior is preserved.
  * Clear user-facing errors returned when a referenced pool name does not exist.

* **Tests**
  * Added comprehensive end-to-end tests covering pool lookup by name for IP and number pools, plus a test update ensuring schema hash refresh after schema registration.

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -415,9 +415,6 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         if not allocate_resources:
             return
 
-        if not allocate_resources:
-            return
-
         if (
             number_pool.node.value in [self._schema.kind] + self._schema.inherit_from
             and number_pool.node_attribute.value == attribute.name
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -415,6 +415,9 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         if not allocate_resources:
             return
 
+        if not allocate_resources:
+            return
+
         if (
             number_pool.node.value in [self._schema.kind] + self._schema.inherit_from
             and number_pool.node_attribute.value == attribute.name
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -404,9 +404,17 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
             return
 
         try:
-            number_pool = await registry.manager.get_one(
-                db=db, id=number_pool_id, kind=CoreNumberPool, raise_on_error=True
-            )
+            if is_valid_uuid(number_pool_id):
+                number_pool = await registry.manager.get_one(
+                    db=db, id=number_pool_id, kind=CoreNumberPool, raise_on_error=True
+                )
+            else:
+                results = await registry.manager.query(
+                    db=db, schema=InfrahubKind.NUMBERPOOL, filters={"name__value": number_pool_id}
+                )
+                if not results:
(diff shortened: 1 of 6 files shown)


backend/tests/component/graphql/resource_manager/test_number_pool_lookup_by_name.py
```diff
@@ -0,0 +1,162 @@
...
+    """Tests for referencing number pools by name instead of UUID in from_pool."""
+
+    @pytest.fixture(scope="class")
+    async def number_pool(
+        self,
+        db: InfrahubDatabase,
+        default_branch_scope_class: Branch,
+        register_core_models_schema_scope_class: SchemaBranch,
+    ) -> CoreNumberPool:
+        await load_schema(db=db, schema=SchemaRoot(nodes=[TICKET]))
```
gmazoyer, line 63: 👍

gmazoyer (approved):

backend/infrahub/core/node/__init__.py
```diff
@@ -403,9 +403,17 @@ async def handle_pool(
...
-            number_pool = await registry.manager.get_one(
-                db=db, id=number_pool_id, kind=CoreNumberPool, raise_on_error=True
-            )
+            if is_valid_uuid(number_pool_id):
+                number_pool = await registry.manager.get_one(
+                    db=db, id=number_pool_id, kind=CoreNumberPool, raise_on_error=True
+                )
+            else:
+                results = await registry.manager.query(
+                    db=db, schema=InfrahubKind.NUMBERPOOL, filters={"name__value": number_pool_id}
```
ogenstad, line 413: Not a blocker to make progress but `number_pool` would get different types compared to above where we are querying for `CoreNumberPool` instead of `InfrahubKind.NUMBERPOOL`. But the type checkers probably don't pick this up due to all of the ignore lines in `pyproject.toml` :(

backend/infrahub/core/relationship/model.py
```diff
@@ -561,8 +562,16 @@ async def resolve(
...
+            pool: Node | None = None
             pool_id = str(self.from_pool.get("id"))
-            pool = await registry.manager.get_one(db=db, id=pool_id, branch=self.branch)
+            if is_valid_uuid(pool_id):
+                pool = await registry.manager.get_one(db=db, id=pool_id, branch=self.branch)
+            else:
+                results = await registry.manager.query(
+                    db=db, schema=InfrahubKind.RESOURCEPOOL, filters={"name__value": pool_id}, branch=self.branch
+                )
+                results = cast("list[Node]", results)
```
ogenstad, line 573: I'm guessing that the problem here is that the type of the overload for NodeManager.query is incorrect and returns an Any instead of a Node. Once we change that (if it happens before the refactoring of those classes) I think this should resurface as a redundant `cast`, so we probably don't need to track it.

ogenstad (approved):

PR #8495, opened 2026-02-27
IFC-1624 Remove deprecated GraphQL mutations/queries

gmazoyer:
# Why

The following GraphQL mutations and queries have been deprecated for a while, we can now safely remove them: `IPAddressGetNextAvailable`, `IPPrefixGetNextAvailable`, `IPPrefixPoolGetResource`, `IPAddressPoolGetResource`

## What changed

Simple code removal

## Impact & rollout

- **Backward compatibility:** If users are still using these queries/mutations dispite the fact that they have been deprecated for a while, they will observe failure when trying to use them again.
## Checklist

## Summary by CodeRabbit

* **Removed**
  * Deprecated GraphQL queries IPAddressGetNextAvailable and IPPrefixGetNextAvailable removed — migrate to InfrahubIPAddressGetNextAvailable and InfrahubIPPrefixGetNextAvailable.
  * Deprecated GraphQL mutations IPAddressPoolGetResource and IPPrefixPoolGetResource removed — migrate to Infrahub-prefixed equivalents.
* **Documentation**
  * Changelog entry added noting removal of these deprecated queries and mutations.


ogenstad:
Can you double check to verify that the SDK doesn't still use the old versions?

gmazoyer:
> Can you double check to verify that the SDK doesn't still use the old versions?

I did. The SDK appears to be using the correct versions of these mutations/queries since we deprecated them.

ogenstad (approved):
LGTM. Final question would be if you also checked the frontend code if any of the deprecated versions were still in use anywhere?

gmazoyer:
> LGTM. Final question would be if you also checked the frontend code if any of the deprecated versions were still in use anywhere?

I verified this too, the frontend is using the proper versions.

PR #8506, opened 2026-02-27
Show related nodes and templates in the profile details view

bilalabbad:
image

## Summary by CodeRabbit

* **New Features**
  * Profile details view now shows objects that reference or use the profile.

* **Tests**
  * Added tests confirming which profile relationships appear in the profile tab and that unrelated relationships remain hidden.

* **Documentation**
  * Added a changelog entry describing the profile relationships visibility update.

### DIFF
--- a/changelog/+profile-relationships-tab.added.md
+++ b/changelog/+profile-relationships-tab.added.md
@@ -1,0 +1,1 @@
+Profile details view now displays the objects where the profile is being used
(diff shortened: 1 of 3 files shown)


wvandeun:
Terminology wise I think we should rename the tab to Related objects.

dgarros:
@bilalabbad would be great to get this one across the finish line

PR #8515, opened 2026-03-02
Auto create SSO groups feature flag

AlexanderGrooff:
# Why

I'm missing the ability to automatically create groups coming from SSO. This PR adds feature flag `sso_generate_groups` to automatically create groups that are imported from SSO, and adds the ability to filter it based on a pattern defined in `sso_generate_groups_filter`.

## What changed

Adds a new block in the authentication code to import groups from the SSO provider when the feature flag is enabled.
I've also changed the documentation to reflect these changes.

The current behaviour doesn't change. It only gets picked up when the feature flag is enabled.

## How to test

I've added testcases, but you can also try importing SSO groups when the feature flag is enabled, that way you can see automatic groups being created.

## Impact & rollout

- **Config/env changes:** new flags `INFRAHUB_SECURITY_SSO_GENERATE_GROUPS` and `INFRAHUB_SECURITY_SSO_GENERATE_GROUPS_FILTER`
- **Deployment notes:** safe to deploy

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Automatic creation of account groups for SSO users from identity provider groups, preserving existing groups and attaching memberships.
  * Concurrency-safe provisioning to avoid duplicate group creation.
  * New config options to enable auto-generation and provide regex filter(s) for extracting group names; invalid patterns are rejected.
  * Auto-created groups have no permissions by default

* **Documentation**
  * Guides and reference updated with instructions, examples, and notes on auto-group creation and filters

* **Tests**
  * New unit and functional tests covering auto-creation, filtering, membership, and validation

### DIFF
--- a/backend/infrahub/auth.py
+++ b/backend/infrahub/auth.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 import uuid
 from datetime import UTC, datetime, timedelta
 from enum import StrEnum
@@ -154,15 +155,27 @@ async def signin_sso_account(db: InfrahubDatabase, account_name: str, sso_groups
             prefetch_relationships=True,
         )
 
+        existing_group_names = {group.name.value for group in infrahub_groups}
+
         if config.SETTINGS.security.sso_generate_groups:
-            existing_group_names = {group.name.value for group in infrahub_groups}
+            filter_pattern = config.SETTINGS.security.sso_generate_groups_filter
+            created_group_names: set[str] = set()
             for group_name in sso_groups:
                 if group_name not in existing_group_names:
-                    new_group = await Node.init(db=db, schema=CoreAccountGroup)
-                    await new_group.new(db=db, name=group_name)
-                    await new_group.save(db=db)
-                    infrahub_groups.append(new_group)
-                    log.info(f"Auto-created SSO group: {group_name}")
+                    effective_name = group_name
+                    if filter_pattern is not None:
+                        match = re.match(filter_pattern, group_name)
+                        if match and match.lastindex:
+                            effective_name = match.group(1)
+                        elif not match:
+                            continue
+                    if effective_name not in existing_group_names and effective_name not in created_group_names:
+                        new_group = await Node.init(db=db, schema=InfrahubKind.ACCOUNTGROUP)
+                        await new_group.new(db=db, name=effective_name)
+                        await new_group.save(db=db)
+                        infrahub_groups.append(new_group)
+                        created_group_names.add(effective_name)
+
 
         for group in infrahub_groups:
             members = await group.members.get_peers(db=db, branch_agnostic=True, peer_type=CoreAccount)
--- a/backend/infrahub/auth.py
+++ b/backend/infrahub/auth.py
@@ -139,6 +139,35 @@ async def create_fresh_access_token(
     return models.AccessTokenResponse(access_token=access_token)
 
 
+def _extract_effective_sso_group_names(sso_groups: list[str], filter_pattern: str | None) -> set[str]:
+    if filter_pattern is None:
+        return set(sso_groups)
+
+    try:
+        compiled_filter = re.compile(filter_pattern)
+    except re.error as exc:
+        log.warning(
+            "Invalid SSO group filter pattern, skipping auto-group generation",
+            filter_pattern=filter_pattern,
+            error=str(exc),
+        )
+        return set()
+
+    effective_group_names: set[str] = set()
+    for group_name in sso_groups:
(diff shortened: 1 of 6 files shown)


backend/tests/unit/test_sso_generate_groups.py
```diff
@@ -0,0 +1,367 @@
...
+            config.SETTINGS.security.sso_generate_groups = original_generate
+            config.SETTINGS.security.sso_generate_groups_filter = original_filter
+
+
+class TestSSOGenerateGroupsFilterValidation:
+    def test_invalid_regex_pattern_raises_error(self):
+        """Verify that invalid regex patterns raise a validation error"""
+        from pydantic import ValidationError
+
+        from infrahub.config import SecuritySettings
```
gmazoyer: If possible, we should avoid lazy imports. We prefer having them in the top-level import block, unless we have a real reason not to.

backend/infrahub/auth.py
```diff
@@ -138,6 +139,40 @@ async def create_fresh_access_token(
...
+    for group_name in sso_groups:
+        for compiled_filter in compiled_filters:
+            match = compiled_filter.match(group_name)
+            if not match:
+                continue
+
+            if match.lastindex:
+                effective_group_names.add(match.group(1))
+            else:
+                effective_group_names.add(group_name)
```
gmazoyer: I have not verified this entirely but it seems this could lead to unexpected results (which could probably be highlighted by running `test_filter_with_complex_pattern`). I think that as regexp such as `^(dev|prod)-.*-team$` would produce groups like `dev` and `prod` instead f groups like `dev-foo-team` and `prod-bar-team`.

backend/infrahub/auth.py
```diff
@@ -138,6 +139,40 @@ async def create_fresh_access_token(
...
+            "Invalid SSO group filter pattern, skipping auto-group generation",
+            filter_pattern=filter_pattern,
+            error=str(exc),
+        )
+        return set()
+
+    effective_group_names: set[str] = set()
+    for group_name in sso_groups:
+        for compiled_filter in compiled_filters:
+            match = compiled_filter.match(group_name)
```
gmazoyer, line 177: The code uses `compiled_filter.match(group_name)` which anchors at the start of the string. A filter like `groups/(\w+)` would* *not** match `ldap/groups/network` because the pattern is expected to match from position 0. This should either use `re.search` and document that patterns are searched within the string or document explicitly that patterns are anchored to the start (like adding "Patterns are matched from the start of the group name" to the docs).

gmazoyer (commented):
The PR is a pretty good start, clean separation of `_extract_effective_sso_group_names` as a testable helper, pragmatic concurrency handling for group creation, and thorough documentation with good coverage across unit and functional tests 👍 

Users are never removed from groups when the IdP stops sending a claim. With auto-generation this becomes more significant since groups proliferate automatically and stale memberships accumulate silently. Worth documenting as a known limitation or addressing in a follow-up PR.

backend/infrahub/auth.py
BaptisteGi:
```python
import re

path = "ldap/groups/admins"
match = re.search(r"ldap/groups/(\w+)", path)

match.group(0)  # "ldap/groups/admins" — the full match
match.group(1)  # "admins" — the captured part
```

Also the current capture would match pattern containing `-` e.g. `full-admin`

AlexanderGrooff: I think it makes sense to keep using `match` to prevent matching unwanted cases through `search`. I've updated [the docs](https://github.com/opsmill/infrahub/pull/8515/changes/119e4c1771fe001a1e460e96125a111b417fc17b#diff-1acd717ebb2f2b63a558163a2ee62249254fb87f4609734c24ccb65fd5a08e9aR521) to match this.

What do you think?

backend/infrahub/auth.py
AlexanderGrooff: That makes sense right? If you want to capture a part of a group name, using regex groups is the way to go.

The reason I do this is because syncing LDAP groups to SSO causes groupnames to appear like `ldap/groups/team_name_1`. In that case, I want to capture only `team_name_1` using a pattern like `^ldap\/groups\/(.*)`

wvandeun:
@AlexanderGrooff thanks for the contribution. We have not forgotten about this. We like the overall direction this is going. But before we can decide on the final implementation, we need to get additional requirements.

We'll get back to you ASAP.

BeArchiTek:
Closing this PR as the engineering team redid the feature as part of 1.10 delivery

PR #8527, opened 2026-03-05
IFC-2216: Update read only repository operational status

### LINKED ISSUE #8166: bug: Read-Only repositories remain in operational status = unknown
### Component

_No response_

### Infrahub version

1.7.1

### Current Behavior

When importing a read-only repository the operational status remains as "unknown".

### Expected Behavior

On a successful import the expectation would be that the state is "online".

### Steps to Reproduce

1. Start a new instance of infrahub
2. Import a read-only repository
3. Observe the operational state

### Additional Information

We've had this bug before in #5755, now it seems to work for managed repositories but not read-only repositories.

solababs:
# Why

Read-only repositories did not update their `operational_status` to `ONLINE` after a successful sync. The status remained stale regardless of whether the sync actually imported new objects or found the local state already current.

Goal: ensure `operational_status` is set to `ONLINE` on every successful `sync_from_remote` call, covering both the "already up to date" and "newly synced" paths.

Non-goals: this does not change the sync logic itself, error handling, or status transitions for failure cases.

Closes https://github.com/opsmill/infrahub/issues/8166

## What changed

- `InfrahubReadOnlyRepository.sync_from_remote` now returns a `bool` indicating whether a sync was performed, and calls `_update_operational_status(ONLINE)` on both the early-return (already current) path and the full sync path.
- `update_latest_commit` uses the returned bool to avoid a redundant `update_commit_value` call when the sync was skipped.

### Suggested review order
1. `backend/infrahub/git/repository.py` — the core logic change
2. `backend/tests/integration_docker/test_repositories.py` — new integration tests
3. Test fixtures (`repos/core-repo/` and `repos/read-only-repo/`) — minimal `.infrahub.yml` + GQL query added to support the integration tests

## How to review

Focus on the two code paths in `sync_from_remote`: the early return when `commit == local_branches[self.ref].commit`, and the normal sync path. Both must call `_update_operational_status(ONLINE)`.

The `update_latest_commit` change prevents a double `update_commit_value` call that would previously occur when `sync_from_remote` returned early — now it only calls it when sync was skipped (i.e., `synced_from_remote` is False).

## How to test

```bash
# Integration tests (requires running Infrahub instance via Docker)
uv run pytest backend/tests/integration_docker/test_repositories.py -v
```

Expected: both `test_read_only_repo_operational_status` and `test_core_repo_operational_status` pass, with `operational_status` == `ONLINE` after sync.

## Impact & rollout

- **Backward compatibility:** No breaking changes. `sync_from_remote` signature adds a return value; all internal callers updated.
- **Performance:** No impact.
- **Deployment notes:** Safe to deploy.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Repository sync now sets repositories to ONLINE after synchronization.

* **Bug Fixes**
  * Avoids unnecessary local updates when a repository is already current.
  * Sync now reports whether a remote synchronization actually occurred, preventing redundant commit updates.

* **Tests**
  * Added integration tests verifying repository operational status for different repository types.

* **Documentation**
  * Changelog entry recording the operational status fix.


backend/tests/integration_docker/test_repositories.py
```diff
@@ -0,0 +1,39 @@
...
+        repo_name = "core-repo"
+        repo_dir = fixture_dir / "repos" / repo_name / "initial__main"
+        repo = GitRepo(name=repo_name, src_directory=repo_dir, dst_directory=remote_repos_dir)
+        await repo.add_to_infrahub(client=client)
+        in_sync = await repo.wait_for_sync_to_complete(client=client)
+        assert in_sync
+
+        repos = await client.all(kind=CoreRepository)
+        assert len(repos) == 1
+        assert repos[0].operational_status.value == RepositoryOperationalStatus.ONLINE.value
```
gmazoyer, line 48: Could this be a single parametrised test with `GitRepoType` as a parameter? Do we need the two different fixtures, also could be re-use an existing fixture such as `backend/tests/fixtures/repos/test_base`?

gmazoyer:
Can you add a changelog entry?

python_sdk
ogenstad, line 1: Looks like this change might be causing some of the typing issues. Was this change intentional?

backend/tests/integration_docker/test_repositories.py
solababs: @gmazoyer the other fixtures seem to be used specifically for some intentional test failures, as i have tested them, I have recreated the fixtures to fix this

ogenstad (approved):
LGTM, if it would be nice if we could have these changes target the `stable` branch as it's a bug and something we could include in the next release instead of having to wait for 1.9 with the `develop` branch.

ogenstad (approved):

PR #8535, opened 2026-03-07
simpler create/update for Template relationship pools

ajtmccarty:
goes with https://github.com/opsmill/infrahub-sdk-python/pull/863 in the SDK

# Why

Current approach to supporting relationships that provision from resource pools on templates is to use a `xyz_from_resource_pool` relationship that links to the CoreIPAddress/PrefixPool to use when provisioning from the template and where `xyz` is the name of the relationship defined on the node schema's template

for example, I create the `InfraInterface` node schema that includes a `prefix` relationship with a peer of `BuiltinIPPrefix`. then I create a template for `InfraInterface` and I set the `prefix_from_resource_pool` relationship to point to the CoreIPPrefixPool that I want to provision from when creating an `InfraInterface` instance using this template.

there is a good reason for using this approach in the backend (we don't have a way to save metadata about a relationship that does not yet exist), but the end user shouldn't have to deal with our implementation details. this approach is made even more confusing b/c we support NumberPools on attributes in templates in the regular way (i.e. `{from_pool: {id: "..."}}`)

## What changed

- set the generated `_from_resource_pool` relationships on templates to be read-only so that the user cannot try to edit them
- update the create and update mutations to support special handling for Templates so that creating a Template with a `from_pool` on a relationship that includes a `_from_resource_pool` relationship will link the pool to the template using the `_from_resource_pool` relationship, which is what the backend template provisioning logic expects
- tests for this and updates to existing tests to remove direct use of the `_from_resource_pool` relationships

## Summary by CodeRabbit

* **New Features**
  * Template mutations now route and apply pool-backed relationships automatically, supporting swaps and clears between pool and direct references
  * Added optional hfid/id inputs for IP address relationships for finer identification

* **Bug Fixes**
  * Validation prevents specifying both a direct reference and a pool-origin reference simultaneously
  * Resource-pool relationships are now treated as read-only where applicable

* **Tests**
  * Added and expanded tests covering pool vs direct relationship create/update lifecycle

* **Chores**
  * Updated SDK submodule


backend/infrahub/graphql/mutations/main.py
```diff
@@ -246,6 +247,9 @@ async def mutate_create(
         create_data = dict(data)
         create_data.update(override_data or {})
 
+        if isinstance(schema, TemplateSchema):
+            create_data = TemplatePoolHandler.route_relationships(data=create_data, schema=schema)
+
```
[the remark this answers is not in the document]
ajtmccarty: the graphql schema generation prevents including inputs for read_only relationships

backend/tests/component/graphql/templates/test_template_pool_relationships.py
```diff
@@ -0,0 +1,428 @@
...
+
+        reloaded = await NodeManager.get_one(id=template.id, db=db, branch=default_branch_scope_class)
+        pool_peer = await reloaded.get_relationship("primary_ip_from_resource_pool").get_peer(db=db)
+        assert pool_peer is not None
+        assert pool_peer.id == ip_address_pool.id
+
+        direct_peer = await reloaded.get_relationship("primary_ip").get_peer(db=db)
+        assert direct_peer is None
+
+    async def test_null_clears_both_relationships(
```
polmichel: I feel this test could be misleading, as the 'primary_ip' relationship is never populated during the 'save' method.
I will modify this test in a commit so that it only explicitly checks both relationships emptiness while verifying the 'primary_ip_from_resource_pool' relationship is correctly populated during the save method.
I will add another one based on 'primary_ip'  template creation

polmichel: Please tell me if I am wrong about the intention behind this test

ajtmccarty:
we are not taking the `from_pool` approach any longer. it's going to be `_from_resource_pool` all the way

PR #8536, opened 2026-03-07
set source of pool rels on Templates

ajtmccarty:
should follow #8535 

adds an extra layer of logic to the single relationship graphql resolver to handle a special case for templates using a resource pool on a relationship

if a Template instance has a `xyz_from_resource_pool` relationship linked to a CoreIPAddress/PrefixPool, this really means that the source of the `xyz` relationship will be the CoreIPAddress/PrefixPool on the `xyz_from_resource_pool` relationship

the new graphql serialization logic handles this by setting the source of the `xyz` relationship to the linked resource pool in this format
```
xyz {
    properties {
        source {
            id
            ...
        }
    }
}
```

in combination with #8535, this allows the `_from_resource_pool` relationships to be completely ignored by users of the UI, SDK, object files, and graphql


polmichel (approved):

ajtmccarty:
we are not taking the `from_pool` approach any longer. it's going to be `_from_resource_pool` all the way

PR #8559, opened 2026-03-10
PR2: Guard Against Setting Both Attribute Value and Pool Relationship

polmichel:
# Why

Currently, a user can define a direct value for an attribute related to a template object and, at the same time, define a `_from_resource_pool` relationship, which aims to provide a value for this attribute when an object is created from the template.

## Goal

Add additional safeguards to prevent the direct attribute value and resource pool provisioning from being used concurrently when a node object is created from the template.

## What changed

- An additional check has been added to the current existing template & ressource pool exclusive constraint `TemplateResourcePoolExclusiveConstraint`. This check is also triggered during the node creation through the `NodeConstraintRunner`.

### Design 

- Tried to keep the `infrahub.core.constraint.node.runner.NodeConstraintRunner` as generic as possible. But I still had to add a new logic step into his algorithm.

## How to test

```bash
uv run pytest -xvs backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
```

## Impact & rollout

- **Backward compatibility:** Issue can occur if there is already existing concurrent relationships and attribute value defined on an existing template
- **Deployment notes:** requires coordinated release

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added a default filter-expansion hook for relationship constraints so constraints can extend field filters.

* **Bug Fixes**
  * Constraint checks now respect expanded filters and consistently use the node schema when evaluating relationships.
  * Template pool constraint enforces mutual exclusion against both relationship and attribute counterparts, rejecting invalid pool+attribute combinations.

* **Tests**
  * Expanded tests and fixtures for pool-related constraints, attribute interactions, and runner filter expansion.

### DIFF
--- a/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
+++ b/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
@@ -29,6 +29,17 @@ class TemplateResourcePoolExclusiveConstraint(RelationshipManagerConstraintInter
         self.db = db
         self.branch = branch
 
+    def expand_filters(self, field_filters: list[str], node_schema: MainSchemaTypes) -> list[str]:
+        if not node_schema.is_template_schema:
+            return field_filters
+
+        expanded = list(field_filters)
+        for name in field_filters:
+            pool_rel_name = f"{name}{RESOURCE_POOL_REL_SUFFIX}"
+            if pool_rel_name in node_schema.relationship_names and pool_rel_name not in expanded:
+                expanded.append(pool_rel_name)
+        return expanded
+
     async def check(self, relm: RelationshipManager, node_schema: MainSchemaTypes, node: Node) -> None:
         if not node_schema.is_template_schema:
             return
--- a/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
+++ b/backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
@@ -87,8 +87,7 @@ class TemplateResourcePoolExclusiveConstraint(RelationshipManagerConstraintInter
                 }
             )
 
-    @staticmethod
-    def _check_attribute_counterpart_not_set(node: Node, attribute_name: str, current_name: str) -> None:
+    def _check_attribute_counterpart_not_set(self, node: Node, attribute_name: str, current_name: str) -> None:
         """Check that the counterpart attribute does not have a user-set value."""
         try:
             attr = node.get_attribute(name=attribute_name)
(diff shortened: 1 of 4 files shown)


backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
```diff
@@ -23,62 +28,85 @@
...
+            "name": "IPAddress",
+            "namespace": "Ipam",
+            "default_filter": "address__value",
+            "order_by": ["address__value"],
+            "display_labels": ["address__value"],
+            "branch": BranchSupportType.AWARE.value,
+            "inherit_from": [InfrahubKind.IPADDRESS],
+        },
+    ],
+}
```
ajtmccarty: looks like there's an `ipam_schema` fixture that I think matches this one you can use
`backend/tests/component/conftest.py`
oh, looks like it was like this before. well, change it if you want to

backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
```diff
@@ -238,7 +266,7 @@ async def test_constraint_rejects_direct_when_pool_is_set(
         )
         await template.save(db=db)
 
-        loaded_template = await registry.manager.get_one(db=db, id=template.id, branch=default_branch_scope_class)
+        loaded_template = await NodeManager.get_one(db=db, id=template.id, branch=default_branch_scope_class)
```
ajtmccarty, line 240: 👍

backend/infrahub/core/relationship/constraints/template_resource_pool_exclusive.py
```diff
@@ -70,3 +86,21 @@ async def _check_counterpart_not_set(self, node: Node, counterpart_name: str, cu
                     )
                 }
             )
+
+    @staticmethod
+    def _check_attribute_counterpart_not_set(node: Node, attribute_name: str, current_name: str) -> None:
```
ajtmccarty: I think this should be a regular instance method (ie not a `staticmethod`)
it's true that this method does not use any instance variables, so it does not _need_ the `self` argument, but the method should not be used anywhere outside of an instance of this class. and it may need access to `self` at some point in the future

ajtmccarty (approved):
a couple of small comments, but otherwise looks great
I think this is the best approach for now. I'm pretty sure that we have not needed to use attributes and relationships together in constraint validation before, but this looks like a good way to do it

backend/tests/component/core/constraint_validators/test_template_resource_pool_exclusive.py
polmichel: Will do it anyway, thanks for the feedbacks

PR #8561, opened 2026-03-10
Add spec for user facing schema separation

ogenstad:
# Why

This is just a draft

## What changed

* TBD

### DIFF
--- a/dev/specs/infp-234-user-facing-schema-refactor/spec.md
+++ b/dev/specs/infp-234-user-facing-schema-refactor/spec.md
@@ -1,0 +1,132 @@
+# Feature Specification: Schema Load Model Refactor
+
+**Feature Branch**: `infp-234-user-facing-schema-refactor`
+**Created**: 2026-03-10
+**Status**: Draft
+**Input**: Jira INFP-234: Refactor user-facing schema models generated from Infrahub's internal schema
+
+## Overview
+
+Infrahub has two distinct interactions with schema data:
+
+- **Schema loading** (`/api/schema/load`): Users submit schema definitions to configure Infrahub. This is the *write path*.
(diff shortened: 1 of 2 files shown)


dev/specs/infp-234-user-facing-schema-refactor/spec.md
```diff
@@ -0,0 +1,132 @@
+# Feature Specification: Schema Load Model Refactor
```
dgarros, line 1: While it's not technically mandatory, I think it would be good to maintain a Pydantic version of both schemas in the SDK

dev/specs/infp-234-user-facing-schema-refactor/spec.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. Its value is computed or populated by the system during schema processing (e.g., inheritance resolution, reference tracking)
+2. A user setting it manually would either be a no-op or produce incorrect system behaviour
+3. It is never meaningfully authored in a schema definition file by a human
+
+A field is **user-configurable** (included in the Schema Load Model) if a user is expected to set it when authoring a schema — regardless of whether the value can be updated later. Fields that are set on initial load but cannot be changed after (e.g., `human_friendly_id`, `uniqueness_constraints`) are user-configurable and must remain in the Load Model.
+
+### Functional Requirements
+
+- **FR-001**: Before implementation begins, every field across nodes, attributes, relationships, and generics in the schema definition source of truth MUST be individually classified as internal-only or user-configurable using the criteria above, and the complete list MUST be reviewed and approved; no Schema Load Model may be generated until this classification is approved
```
dgarros, line 96: I think HOW we do that is important, my preference is to implement a pattern similar to what we are doing to indicate if a field requires a migration of not using `json_schema_extra={"update": "migration_required"}`

dgarros (approved):

PR #8563, opened 2026-03-10
Fix flaky test_event_query_prefect

polmichel:
# Why

`test_event_query_prefect` fails from time to time in CI with:

```
FAILED backend/tests/component/graphql/queries/test_event.py::test_event_query_prefect - assert == failed.

LHS vs RHS shown below

<Insert a number greater than 11>
11
```
-> see https://github.com/opsmill/infrahub/actions/runs/22899989741/job/66460736640?pr=8543

The root cause: the query at line 639 filters by `event_type=["infrahub.node.created"]` without using `filter_outofscope_events`, so the count includes events from the entire Prefect instance and not just the 11 created by this test. The default `limit=10` also meant only 10 edges were returned, masking a second mismatch.

The generic branch names (`"branch1"`, `"branch2"`, `"branch3"`) could also cause collisions with other tests sharing the same Prefect instance, I have observed it while creating a test using new branches.

## What changed

- Applied `filter_outofscope_events` to the `event_type`-only query that was missing it, with `limit: 50` to fetch all matching events
- Fixed edges assertion from `* 10` to `* 11` as the default `limit=10` was silently dropping the 11th event
- Made branch names unique per test run by appending a random suffix (`uuid4().hex[:8]`), stored in module-level constants `BRANCH1_NAME`, `BRANCH2_NAME`, `BRANCH3_NAME` in order to preventive measure against cross-test collisions

## How to test

```bash
uv run pytest -xvs backend/tests/component/graphql/queries/test_event.py::test_event_query_prefect
uv run ruff check backend/tests/component/graphql/queries/test_event.py
```

## Checklist

- [x] Tests added/updated

🤖 Created with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/tests/component/graphql/queries/test_event.py
+++ b/backend/tests/component/graphql/queries/test_event.py
@@ -142,6 +142,11 @@ query MutatedNodes($id: [String!]) {
 }
 """
 
+_TEST_ID = uuid.uuid4().hex[:8]
+BRANCH1_NAME = f"branch1-{_TEST_ID}"
+BRANCH2_NAME = f"branch2-{_TEST_ID}"
+BRANCH3_NAME = f"branch3-{_TEST_ID}"
+
 ACCOUNT1_ID = "33b15615-649e-4e9e-89b0-85e187251f1f"
 ACCOUNT2_ID = "518b434f-40bf-4b65-b700-04696535ca8e"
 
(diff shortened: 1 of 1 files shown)


ogenstad (approved):
Thank you for looking into this and fixing it!

PR #8567, opened 2026-03-10
diff refresh updates

ajtmccarty:
goes with #8568 
there's no dependency in the code changes, but they are both about handling frozen diffs correctly

the tasks to automatically update or recalculate diffs following a merge or rebase need to be updated to handle frozen diffs for merged or deleted branches

## After a Rebase
The `DIFF_REFRESH_ALL` / `refresh_diff_all()` task is executed following a rebase to recalculate any diffs on the rebased branch from scratch b/c the rebase operation can have basically rewritten history on the branch. this logic did not account for diffs on branches that have been merged or deleted, in which case the diff will be frozen

## After a Merge
the `BRANCH_MERGE_POST_PROCESS` / `post_process_branch_merge()` task is called after branch is merged, as the name suggests. one of the things that this task does is submit tasks to recalculate any branch-tracking diffs on any other active branches. a branch-tracking diff is one that tracks all of the changes on a given branch from creation to merge/deletion. this logic needed to be updated to exclude frozen diffs for branches that have been merged or deleted and then recreated with the same name.

also, added an env var that can be used to skip these diff updates b/c they can be expensive operations and we shouldn't force users to run them all at one time

## Tests
the component tests for these updates do a little more mocking than we usually do because generating diffs is a somewhat expensive operation and the diff generation is not really part of this logic. skipping it makes the tests run much faster

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to control whether diff updates are automatically triggered for active branches after branch merges. Enabled by default.

* **Bug Fixes**
  * Fixed diff update logic to properly ignore diffs for merged and deleted branches following merge and rebase operations.


backend/infrahub/config.py
```diff
@@ -201,6 +201,10 @@ class MainSettings(BaseSettings):
         description="Enable strict schema validation. When set to `False`, "
         "`human_friendly_id` schema fields should not necessarily target a unique combination of peer attributes.",
     )
+    diff_update_after_merge: bool = Field(
+        default=True,
+        description="When enabled, diff updates are triggered for active branches after a branch merge.",
+    )
```
ogenstad, line 207: Only comment here is that the main section becomes public as part of /api/config. It's not really a problem but I also don't think most of these settings are useful for the public to know. The frontend would pretty much need `allow_anonymous_access` and `anonymous_access_role`. This isn't strictly related to your PR, only mentioning it as we're adding more stuff to the main config section. Could be fine for now and that we should consider some other type or refactoring at some different time..

backend/tests/component/core/diff/test_refresh_diff.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+        unfrozen = EnrichedRootFactory.build(
+            base_branch_name="other",
+            diff_branch_name="main",
+            tracking_id=BranchTrackingId(name="main"),
+            is_frozen=False,
+        )
+        mock_workflow = await self._call_system_under_test(db, [frozen, unfrozen])
+        diff_update_calls = _filter_calls(mock_workflow, DIFF_UPDATE)
+        assert len(diff_update_calls) == 1
```
ogenstad, line 181: This test looks fine, but it has a lot of patching and mocking going on.

ogenstad (approved):

PR #8568, opened 2026-03-10
freeze diffs for merged/deleted branches

ajtmccarty:
goes with #8567 
there's no dependency in the code changes, but they are both about handling frozen diffs correctly

we need to freeze diffs when branches are merged/deleted outside of the proposed change pipeline so that various other diff-related operations know to ignore those diffs. we also need a migration to look at all previous diffs and freeze those for merged/deleted branches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Diffs from merged or deleted branches are now automatically frozen to preserve branch history and prevent unintended modifications.

* **Chores**
  * Updated internal schema version for data consistency.


backend/tests/component/core/migrations/graph/test_066_freeze_orphaned_branch_tracking_diffs.py
```diff
@@ -0,0 +1,280 @@
...
+4. Branch with different branched_from time (name reuse) - old diffs SHOULD be frozen, new diffs NOT
+5. Already-frozen diffs - should remain unchanged
+6. Branch merged, deleted, then new branch created with the same name - old diffs frozen, new diffs untouched
+"""
+
+from dataclasses import dataclass
+from uuid import uuid4
+
+import pytest
+from infrahub_sdk.timestamp import Timestamp
```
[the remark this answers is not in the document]
ajtmccarty: core version does not correctly support Timestamp.add() which is used in this test class. `.add()` is type-hinted as returning the SDK version of `Timestamp`

backend/infrahub/core/diff/query/freeze_by_branch.py
```diff
@@ -0,0 +1,45 @@
...
+        }
+        query = """
+MATCH (diff_root:DiffRoot)
+WHERE diff_root.tracking_id = $branch_tracking_id
+AND (diff_root.is_frozen IS NULL OR diff_root.is_frozen <> TRUE)
+OPTIONAL MATCH (diff_root)-[:DIFF_HAS_PARTNER]-(partner:DiffRoot)
+SET diff_root.is_frozen = TRUE
+SET diff_root.tracking_id = $frozen_tracking_id
+SET partner.is_frozen = TRUE
+SET partner.tracking_id = $frozen_tracking_id
```
ogenstad, line 43: Not for now but do you see a future use for having something in addition to the branch name to identify the diff root. With the `diff_root.is_frozen <> TRUE` we won't re-set these on DiffRoot objects when people start to re-use branch names. Wondering if we might want to add an actual ID for other reasons at some point. But looks good with what we have now.

backend/infrahub/core/migrations/graph/m066_freeze_orphaned_branch_tracking_diffs.py
```diff
@@ -0,0 +1,46 @@
...
+        MATCH (diff_root:DiffRoot)
+        WHERE diff_root.tracking_id STARTS WITH "branch."
+        AND (diff_root.is_frozen IS NULL OR diff_root.is_frozen <> TRUE)
+        WITH diff_root, substring(diff_root.tracking_id, 7) AS branch_name
+        OPTIONAL MATCH (branch:Branch {name: branch_name})
+            WHERE branch.status <> "DELETING"
+        WITH diff_root, branch, branch_name
+        WHERE branch IS NULL
+            OR branch.branched_from <> diff_root.from_time
+            OR branch.status = "MERGED"
```
ogenstad, line 30: I don't think we need this as we introduced the MERGED status in 1.8. But I guess it doesn't hurt to have if we for whatever reason want to run the migration again later.

ogenstad (approved):

backend/infrahub/core/diff/query/freeze_by_branch.py
ajtmccarty: all the diffs have unique UUIDs, but we have to have some way to go from a branch name to the diff for the branch, if it exists at all

backend/infrahub/core/migrations/graph/m066_freeze_orphaned_branch_tracking_diffs.py
ajtmccarty: good point
forgot that was a new status

PR #8571, opened 2026-03-11
support number pool _from_resource_pool in template forms and views

bilalabbad:
Why

Number pools have companion _from_resource_pool relationships on template schemas (same pattern as IP address/prefix pools), but the frontend didn't support them. Template creation/update forms sent pool values directly on the attribute instead of through the companion relationship, and the details/table views didn't display pool assignments.

  What changed

  - Form field metadata: Number attributes on template schemas now detect their companion _from_resource_pool relationship and attach
  fromPoolRelationshipName to the field's pool metadata
  - Default value resolution: When editing a template, the form checks the companion relationship for an existing pool assignment and displays it
  correctly
  - Mutation handling: Update mutations now correctly split number attribute pool values into the companion relationship (using { value: null } for
  the attribute, not bare null), and clear the pool relationship when switching to a direct value
  - Details view: ObjectAttributeRow receives the pool node from the companion relationship and renders it as a link
  - Table view: Attribute cells check for a pool node in the companion relationship and render RelationshipNodeDisplay instead of the raw value

  Existing relationship pool behavior (IP address/prefix) is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Resource-pool support: forms can derive and show defaults from associated pools (including number fields).
  * Table and detail views surface pool-sourced relationship data in cells and detail displays.
  * Pool selector toggles selection—choosing an already-selected pool clears it.

* **Behavior Change**
  * Attribute values in object details are shown consistently (no conditional linking).

* **Tests**
  * Added unit tests covering pool-derived defaults and create/update mutation handling.


frontend/app/src/shared/components/form/utils/getFormFieldFromAttribute.ts
```diff
@@ -152,10 +153,20 @@ export const getFormFieldFromAttribute = ({
...
+    const fromPoolName = `${attributeSchema.name}${FROM_RESOURCE_POOL_SUFFIX}`;
+    const hasFromPoolRelationship = schema.relationships?.some((r) => r.name === fromPoolName);
+
     const dropdownField: DynamicNumberFieldProps = {
       ...basicFormFieldProps,
       type: "Number",
       pools: numberPools,
+      pool: hasFromPoolRelationship
+        ? {
+            kind: "CoreNumberPool",
```
pa-lem: is it possible to use a constant?

pa-lem (approved):

frontend/app/src/shared/components/form/utils/getFormFieldFromAttribute.ts
bilalabbad: done

PR #8573, opened 2026-03-11
Use query analyzer fields for artifact definition regeneration

ogenstad:
# Why

Improve artifact definition regeneration within pipeline to avoid regenerating artifacts that doesn't need to be touched.

## Current state

Some background for clarity: Currently when you have setup an artifact definition to create or update artifacts within a pipeline we have three scenarios when an artifact is regenerated (if an artifact for whatever reason doesn't exist it should always be created)

* The source branch is synced with git and there are file changes in the repo compared to what's in the default branch. Currently we can't guarantee that a change to a file might have been used by a transport as there are so many ways you can import other files from Python. (in the future this will be more selective to only trigger if the transform code or query has been updated.)
* The GraphQL query is written so that it fullfills the `only_has_unique_targets` requirements: In this case we will only regenerate artifacts that has modifications to one of the members of its GraphQL query group
The `only_has_unique_targets` is true if the query and the query parameter is enough to always target a single known node.

As an example using this query:

```graphql
query TagQueryUnique($tag_name: String!) {
  BuiltinTag(name__value: $tag_name) {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
}
```

Since the name of a tag has a uniqueness constraint required along with the fact that the `$tag_name` parameter is required in the query we can be sure that this query would always capture the required initial node along with its relationships if any. The way to break `only_has_unique_targets` would be to have the variable be optional `TagQuery($tag_name: String)` without "!". 

* The GraphQL query is written so that it doesn't fullfil `only_has_unique_targets`. If this is the case all artifacts would be regenerated.

If we look at this query instead:

```graphql
query TagQuery($tag_name: String!) {
  BuiltinTag(name__value: $tag_name) {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
  IpamNamespace {
    edges {
      node {
        name {
          value
        }
      }
    }
  }
}
```
Here the tag part is still fine, however when we want to use all of the IpamNamespace in the query we don't have a current way in the pipeline to determine if a new namespace has been created or deleted and as such we default to the safer option of over re-generation instead of risking an invalid artifact. (Down the road additional improvements could be done here to specifically trigger regeneration if in this case a new IpamNamespace has been created)

## What currently gets regenerated

Even with `only_has_unique_targets`, the way this works is that whenever we create an artifact from an artifact definition the query that goes with the artifact will have a matching `CoreGraphQLQueryGroup` that contains the members of all of the nodes included in the query. For larger environments or queries there could be over 1000 members in such a group and the criteria to regenerate an artifact would be if anything on at least one of those nodes got modified within a branch. As an example looking at `TagQueryUnique` from above we can see that the query uses the "name" attribute from the BuiltinTag (which is assumed to be used to render the artifact), however the current implementation would regenerated an artifact for a given BuiltinTag even if it was just the "description" attribute that was modified.

If we consider a network environment with any type of mesh setup there's a big chance that there is a big overlap between multiple artifacts and even unrelated changes to nodes multiple relationships away from the initial node of the artifact could trigger the regeneration of an artifact.

## The goal of this PR

Within this PR we further utilize the InfrahubGraphQLQueryAnalyzer class and instead of only consulting `only_has_unique_targets` within the `GraphQLQueryReport` we also look at the `requested_read` property. This allows us to specifically look at what fields (i.e attributes and or relationships) that are being read by the GraphQL query and only trigger a regeneration if any of those has been changed on the node.

We still need to combine this with `only_has_unique_targets`, but we get a few new states:

* The source branch is synced with git and there are file changes in the repo compared to what's in the default branch: This works exactly like before as described above
* With `only_has_unique_targets`: We only regenerate artifacts if any of the members of its GraphQL query group has any updated fields that matches the `requested_read` property of the GraphQL query group.
* Without `only_has_unique_targets`: We can look at the `requested_read` property and combine this with the diff. If the modifications within the branch doesn't touch any of the fields we are targeting no artifacts get regenerated.
* Without `only_has_unique_targets`: We can look at the `requested_read` property and combine this with the diff. If the modifications within the branch matches any of the fields we are targeting all artifacts get regenerated.

# What changed

* Added a fields property to `ObjectAccess` within the GraphQL query analyzer for easier access to all relationships and attributes that are being read by a query.
* Refactor and remove `_populate_subscribers()`, replaced by `_get_subscribers_for_nodes()` and `__get_subscribers_from_diff()`. There were two reasons first we had different input types then `__populate_subscribers()` mutated the branch_diff param in a weird way that wasn't obvious to the caller. This has been cleaned up so that we instead return a list of subscribers that are added to the branch_diff where it's called
* Updates to logic to determine which artifacts should be regenerated based on the fields that are actually read as defined by the GraphQL query report.
* Added tests to check for different scenarios. These aren't end to end tests, we just want to validate what would happen with regards to which artifacts would be requested to be regenerated

## Summary by CodeRabbit

* **Improvements**
  * Artifact regeneration is more selective: only artifacts whose queries reference changed fields (including attributes and relationships) are triggered, reducing redundant runs.
* **Tests**
  * Added tests, helpers, and a workflow recorder to simulate diffs and capture workflow invocations, validating selective regeneration across unique/non-unique queries, managed-branch overrides, and file-change scenarios.
* **Changelog**
  * Added entry describing the updated artifact regeneration behavior.

### DIFF
--- a/backend/tests/adapters/workflow.py
+++ b/backend/tests/adapters/workflow.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+import uuid
+from typing import TYPE_CHECKING, Any
+
+from infrahub.core.constants import ValidatorConclusion
+from infrahub.services.adapters.workflow import InfrahubWorkflow
+from infrahub.workflows.models import WorkflowDefinition, WorkflowInfo
+
+if TYPE_CHECKING:
+    from infrahub.context import InfrahubContext
+
+
+class WorkflowRecorder(InfrahubWorkflow):
+    """Records workflow calls without executing them. Use for testing code that submits workflows."""
+
+    def __init__(self) -> None:
+        self.execute_calls: list[dict[str, Any]] = []
+        self.submit_calls: list[dict[str, Any]] = []
+
+    async def execute_workflow(
+        self,
+        workflow: WorkflowDefinition,
+        expected_return: type | None = None,
+        context: InfrahubContext | None = None,
+        parameters: dict[str, Any] | None = None,
+        tags: list[str] | None = None,
+    ) -> Any:
+        self.execute_calls.append({"workflow": workflow, "parameters": parameters or {}})
+        if expected_return is ValidatorConclusion:
+            return ValidatorConclusion.SUCCESS
+        return None
+
+    async def submit_workflow(
+        self,
+        workflow: WorkflowDefinition,
+        context: InfrahubContext | None = None,
+        parameters: dict[str, Any] | None = None,
+        tags: list[str] | None = None,
+    ) -> WorkflowInfo:
+        self.submit_calls.append({"workflow": workflow, "parameters": parameters or {}})
+        return WorkflowInfo(id=uuid.uuid4())
+
+    def get_execute_calls_for(self, workflow: WorkflowDefinition) -> list[dict[str, Any]]:
+        return [c for c in self.execute_calls if c["workflow"] == workflow]
+
+    def get_submit_calls_for(self, workflow: WorkflowDefinition) -> list[dict[str, Any]]:
+        return [c for c in self.submit_calls if c["workflow"] == workflow]
--- a/backend/infrahub/graphql/analyzer.py
+++ b/backend/infrahub/graphql/analyzer.py
@@ -132,6 +132,10 @@ class ObjectAccess:
     attributes: set[str] = field(default_factory=set)
     relationships: set[str] = field(default_factory=set)
 
+    @property
+    def fields(self) -> set[str]:
+        return self.attributes.union(self.relationships)
+
 
 @dataclass
 class GraphQLVariable:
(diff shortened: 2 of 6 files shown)


gmazoyer (approved):
That is a proper PR, you made it easy to review thanks to the thorough description 👍

backend/tests/adapters/workflow.py
```diff
@@ -0,0 +1,48 @@
...
+
+from infrahub.core.constants import ValidatorConclusion
+from infrahub.services.adapters.workflow import InfrahubWorkflow
+from infrahub.workflows.models import WorkflowDefinition, WorkflowInfo
+
+if TYPE_CHECKING:
+    from infrahub.context import InfrahubContext
+
+
+class WorkflowRecorder(InfrahubWorkflow):
```
ajtmccarty, line 14: well this looks useful

ajtmccarty (approved):

Issue #7646, opened 2025-11-13
bug: uniqueness constraints not properly validated when objects get created using a profile


wvandeun:
### Component

API Server / GraphQL

### Infrahub version

1.4.13 and 1.5.0

### Current Behavior

Related to #7644
When that issue is solved, we likely don't have a path to trigger this bug. Nonetheless we should validate if this needs to be fixed.

When you create an object using a profile, it seems that we don't properly validate the uniqueness constraints of the schema node properly.

At this stage we can only trigger this, by defining an attribute for a schema node as both optional and unique. Then create an object of that schema node, then you create a profile object with the same name as the object, then create an object using the profile.

in 1.4.13:
- and object exists with the same value for the attribute
- they both have the same hfid (if the optional/unqiue attribute is part of the hfid)

in 1.5.0
- and object exists with the same value for the attribute
- the object that was created 2nd (using the profile) has no hfid or display_label (if the optional/unique attribute is part of the hfid)

### Expected Behavior

Uniqueness constraints are properly validated when creating objects using profiles.

### Steps to Reproduce

- spin up instance of Infrahub
- load this schema
```yaml
---
version: "1.0"
nodes:
  - namespace: Network
    name: Router
    uniqueness_constraints:
      - [name__value]
    human_friendly_id:
      - name__value
    display_label: name__value
    attributes:
      - name: name
        kind: Text
        optional: true
    relationships:
      - name: tags
        kind: Generic
        peer: BuiltinTag
        optional: true
        cardinality: many
```
- create a NetworkRouter object
- create a profile for NetworkRouter with the same value for the name attribute as the object that you created in the previous step
- create an object using the profile

PR #8577, opened 2026-03-11
IFC-2048: Remove optional and unique attributes from profile schema

### LINKED ISSUE #7644: bug: profile schema node contains attribute that is defined as optional and unique
### Component

API Server / GraphQL

### Infrahub version

1.4.13 and 1.5.0

### Current Behavior

When you define a schema node with an attribute that is both optional and part of a uniqueness contraint, then the generated Profile schema node will contain that attribute.

### Expected Behavior

Attributes that are part of uniqueness constraint, should not be added to the generated profile schema node.
https://docs.infrahub.app/topics/profiles#profile-components

### Steps to Reproduce

- Spin up instance of Infrahub (v1.4.13 or v1.5.0)
- Load this schema, with an optional attribute that is part of a uniqueness constraint
```yaml
---
version: "1.0"
nodes:
  - namespace: Network
    name: Router
    uniqueness_constraints:
      - [name__value]
    human_friendly_id:
      - name__value
    display_label: name__value
    attributes:
      - name: name
        kind: Text
        optional: true
    relationships:
      - name: tags
        kind: Generic
        peer: BuiltinTag
        optional: true
        cardinality: many
```
- The schema node ProfileNetworkRouter will contain a `name` attribute
- You can now create a ProfileNetworkRouter profile with a value for the name attribute

### Additional Information

_No response_

solababs:
# Why

Optional attributes that participate in uniqueness constraints must not appear in profile schemas. When a profile sets such an attribute, every node using that profile would share the same value, violating the uniqueness constraint at scale.

Closes https://github.com/opsmill/infrahub/issues/7644 and https://github.com/opsmill/infrahub/issues/7646

## What changed

- Optional attributes that are part of any `uniqueness_constraints` entry are now excluded when generating a profile from a node schema.
- When `uniqueness_constraints` changes during a `schema.load()`, a new migration (`NodeUniquenessConstraintsUpdateMigration`) automatically adds or removes the affected attribute from all existing profile nodes in the database.

- Added `BaseNodeSchema._check_attr_in_uniqueness_constraint` — checks whether an attribute name appears in any uniqueness constraint path (handles both bare attribute names and `attr__field` paths).
- Added `BaseNodeSchema.check_if_attr_supports_profiles` as the canonical gate used by both schema generation and migration logic.
- `NodeUniquenessConstraintsUpdateMigration` compares the *previous* attribute schema (not the new one) against the previous schema's support flags to detect actual changes — necessary because `sync_uniqueness_constraints_and_unique_attributes` writes `unique=True` onto the new attribute when it joins a single-attr constraint, which would otherwise mask the transition.
- `assert_attribute_path_status` and `LATEST_ATTRIBUTE_PATH_STATUS_QUERY` were extracted from the component test file into `tests/helpers/db_validation.py` and reused across component and integration tests.

- Required attributes (non-optional) in uniqueness constraints are still included in profiles — no change to that behavior.
- No GraphQL or REST API contract changes.
- No database graph migrations.

### Suggested review order
1. `BaseNodeSchema._check_attr_in_uniqueness_constraint` and `check_if_attr_supports_profiles` — the shared gate (`basenode_schema.py`)
2. `NodeUniquenessConstraintsUpdateMigration.execute` — the migration that fires on schema load (`node_uniqueness_constraints_update.py`), pay attention to `prev_attr` vs `attr`
3. `test_profile_does_not_contain_optional_unique_attributes` — schema-level unit test
4. `test_node_uniqueness_constraints_update.py` — component tests for the migration
5. `test_migration_uniqueness_constraints.py` — end-to-end integration tests via `client.schema.load()`

## How to review

The key condition in `check_if_attr_supports_profiles`: an attribute is excluded from profiles when `support_profiles` is False (i.e. `unique=True` or `read_only=True`) **or** it appears in any uniqueness constraint.

In `NodeUniquenessConstraintsUpdateMigration.execute`, `prev_attr` must come from `previous_schema`, not `new_schema`. When the constraint is *added*, `sync_uniqueness_constraints_and_unique_attributes` sets `unique=True` on the new attribute before the migration runs. Using the new attribute for the previous check too would make both sides return `False` (supports=False) and no migration would fire.

## How to test

```bash
# Schema generation
uv run pytest backend/tests/component/core/schema_manager/test_manager_schema.py::test_profile_does_not_contain_optional_unique_attributes -v

# Migration unit tests
uv run pytest backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py -v

# End-to-end integration tests
uv run pytest backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py -v
```

## Impact & rollout

- **Backward compatibility:** Profiles generated from schemas with optional attributes in uniqueness constraints will no longer include those attributes. Existing stored profile nodes are unaffected at rest; the migration only runs when `uniqueness_constraints` changes via a schema load.
- **Performance:** No impact.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added a migration to adjust schemas when optional attributes are added to or removed from uniqueness constraints, keeping generated profile schemas consistent.

* **Bug Fixes**
  * Profile generation now omits optional attributes that are governed by uniqueness constraints, preventing unintended exposure of unique optional fields.

* **Documentation**
  * Schema reference updated to mark uniqueness-constraint changes as requiring migrations.


backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4662,3 +4662,21 @@ async def test_manage_object_templates_component_relationship_to_excluded_kind()
     template = schema_branch.get(name=f"Template{TestKind.CAR}", duplicate=False)
     pool_rel = template.get_relationship(name="number_pools")
     assert pool_rel.peer == InfrahubKind.NUMBERPOOL
+
+
+async def test_profile_does_not_contain_optional_unique_attributes() -> None:
```
ajtmccarty, line 4710: test should cover uniqueness_constraints with multiple values too

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2366,6 +2366,15 @@ def manage_profile_relationships(self) -> None:
...
         return f"Profile{node_kind}"
 
+    def check_attr_in_uniqueness_constraint(self, attr: str, node_schema: NodeSchema) -> bool:
+        if not node_schema.uniqueness_constraints:
+            return False
+        for constraint_paths in node_schema.uniqueness_constraints:
+            for constraint_path in constraint_paths:
+                if constraint_path.startswith(f"{attr}__") or constraint_path == attr:
+                    return True
+        return False
```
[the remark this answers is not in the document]
ajtmccarty: this might be right too. b/c profile generation happens before `sync_uniqueness_constraints_and_unique_attributes()`

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2417,7 +2426,9 @@ def generate_profile_from_node(self, node: NodeSchema) -> ProfileSchema:
             )
 
         for node_attr in node.attributes:
-            if not node_attr.support_profiles:
```
ajtmccarty, line 2360: we use `support_profiles` in schema migrations to determine when we need to add/remove attributes on profile schemas. if the real logic for determining whether an attribute supports a profile also need to check the uniqueness constraints of the node,  then we either need a way to handle that in schema migrations or an issue to track that we don't handle it correctly
I don't think we do any schema migrations for a change to uniqueness_constraints, but that seems like what would be required here

ajtmccarty:
I think the full fix for this requires keep profile schemas correctly updated when a schema's `uniqueness_constraints` changes. for example, if I remove an attribute from the `uniqueness_constraints` of a schema, then that attribute might need to be added to the profile schema associated with the node. we don't currently do this and I think it would require a schema migration for uniqueness_constraints, which would require updating the `uniqueness_constraints` schema attribute on `base_node_schema` to `UpdateSupport.MIGRATION_REQUIRED`
I'm not entirely sure if this is in scope for this change or not, but it is the most correct fix. and we've had to run migrations to clean up this kind of missing profile/template schema attributes issue before

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -687,6 +687,15 @@ def update(self, other: HashableModel) -> Self:
 
         return self
 
+    def check_attr_in_uniqueness_constraint(self, attr: str) -> bool:
```
ajtmccarty: I think this method is only used in conjunction with `Attribute/RelationshipSchema.support_profiles` to check if a given attribute or relationship should be included on a profile schema
instead of repeating this logic throughout the code 
```
attr_schema.support_profiles
and not attr_schema.optional
and not schema.check_attr_in_uniqueness_constraint(attr=attr_schema.name)
```
it would be better to have a method that does this complete check. something like `check_if_field_supports_profiles(field_name: str) -> bool` that runs the above logic. the new method could either replace this one or call it

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,71 @@
...
+                name=f"node.uniqueness_constraints.update.{attr.name}",
+                queries=[],
+                new_node_schema=self.new_node_schema,
+                previous_node_schema=self.previous_node_schema,
+                schema_path=SchemaPath(
+                    path_type=SchemaPathType.ATTRIBUTE,
+                    schema_kind=self.new_schema.kind,
+                    field_name=attr.name,
+                ),
+            )
```
ajtmccarty, line 60: I don't think you should need to instantiate a new migration within this migration just to call `execute` on it
I'm looking at `AttributeSupportsGeneratedSchemaMigration.execute` as an example and that one just updates or sets the `queries` parameter before passing it up to `super().execute` and I think the same approach should work here too

ajtmccarty (commented):
not sure if it is in scope for this change or if it should be tracked with another issue (if it doesn't already exist), but the docs say that the same constraint should apply to relationships used in uniqueness_constraints too.

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
solababs: This `NodeUniquenessConstraintsUpdateMigration` doesn't have the same base class as `AttributeSchemaMigration` so we can not run `super()` on it as it would be missing the `new_attribute_schema` property

solababs:
> not sure if it is in scope for this change or if it should be tracked with another issue (if it doesn't already exist), but the docs say that the same constraint should apply to relationships used in uniqueness_constraints too.

@ajtmccarty what docs? as the attributes of a relationship schema are already handled when checking the attributes

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
ajtmccarty: I see this the schema migration/query logic is more of a mess than I remembered
I've made a technical debt item for this whole system https://opsmill.atlassian.net/browse/IFC-2399

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,71 @@
...
+            return result
+        if not self.new_schema.generate_profile:
+            return result
+
+        for attr in self.new_schema.attributes:
+            if not attr.support_profiles or not attr.optional:
+                continue
+
+            previous_in_constraint = self.previous_schema.check_attr_in_uniqueness_constraint(attr=attr.name)
+            new_in_constraint = self.new_schema.check_attr_in_uniqueness_constraint(attr=attr.name)
```
ajtmccarty: I think this should use `check_if_attr_supports_profiles` instead of breaking the logic apart

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -687,6 +687,20 @@ def update(self, other: HashableModel) -> Self:
...
+        for constraint_paths in self.uniqueness_constraints:
+            for constraint_path in constraint_paths:
+                if constraint_path.startswith(f"{attr}__") or constraint_path == attr:
+                    return True
+        return False
+
+    def check_if_attr_supports_profiles(self, attribute_schema: AttributeSchema) -> bool:
+        return attribute_schema.support_profiles and not (
+            attribute_schema.optional and self.check_attr_in_uniqueness_constraint(attr=attribute_schema.name)
+        )
```
ajtmccarty: this looks like it returns True if an attribute is required and in the uniqueness constraint, which I don't think we want. Pretty sure we want to exclude any attributes that are in a uniqueness constraint

backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
ajtmccarty, line 143: I think we might need to do this same update for templates, b/c which attributes are included on a template schema are also affected by the uniqueness_constraints
this could be a separate follow-up issue. but the template problem should probably be fixed in the same release as this profile one

solababs: @wvandeun please confirm the same behavior should be used for template

backend/infrahub/core/schema/basenode_schema.py
solababs: Yes correct, we want to """Remove optional and unique attributes from profile schema""" it can be included if it is required

ajtmccarty: but if the attribute is in the uniqueness_constraint, then shouldn't it be excluded from profiles?

solababs: it is only excluded if it is optional

ajtmccarty: this is what the docs say
https://docs.infrahub.app/topics/profiles#profile-constraints-and-limitations
image

do the docs need to change? or is this PR changing this constraint?

solababs: Ok got it, @wvandeun please advise

backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
wvandeun: If we have the same issue with templates, then yes!

backend/infrahub/core/schema/basenode_schema.py
wvandeun: The documentation is correct. Attributes/relationships that are part of the uniqueness constraint should not be in the profile (independent of the fact that they are optional or required).

backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,437 @@
...
+    WHERE r2.branch = $branch_name
+    RETURN r2
+    ORDER BY r2.branch_level DESC, r2.from DESC
+    LIMIT 1
+}
+RETURN node.uuid AS node_id, r1.status AS has_attr_status, r2.status AS has_val_status
+"""
+
+
+async def assert_attribute_path_status(
```
ajtmccarty: 1. the cypher query for this function will skip Nodes that do not have an attribute with the matching name and I think it would want to catch those. the fix is to start with a MATCH on the vertex label and then update the MATCHes in the subqueries to be OPTIONAL MATCH
2. it should probably also include an `r1/2.status DESC` in the ORDER BY clauses in the subqueries to make sure that active edges are preferred when edges are added and deleted at the same time
3. it would also be good to confirm that the `to` time on the edge is `null` to confirm that it is still active
4. looking at this some more, it looks like we do something pretty similar in some other tests. this PR has already getting kind of big, so maybe best to leave those other tests for a different PR

backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+async def test_migration_attribute_added_to_uniqueness_constraint(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    car_person_schema: SchemaBranch,
+    car_profile1_main: Node,
+) -> None:
+    """Adding nbr_seats to a uniqueness constraint removes it from profile nodes."""
+    await assert_attribute_path_status(
```
ajtmccarty, line 74: I think a cleaner way to handle these assertions is to get the `ProfileTestCar` schema and verify that it does or does not have the `nbr_seats` attribute
then get any instances of the `ProfileTestCar` created for or during the test and check the same thing

ajtmccarty (commented):
just some test comments
not sure what is going wrong with the component test failures
it would be a good to add an integration test that confirms the new uniqueness constraint migration is triggered and runs correctly on a schema update that includes a uniqueness constraint change

backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py
ajtmccarty: thinking about this a third time, maybe it is all right as is if it is only checking that an attribute was added that did not exist before. if it is testing updates, then it is probably not sufficient, but if it is only testing create it is probably fine

backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py
ajtmccarty: actually you might need to test it at the database level if you need to confirm that a new attribute was actually added with a NULL value b/c I think attribute loading will fill in missing attributes if they exist on the schema
he best way to confirm is probably to set an actual value on the new attribute on the template and verify that can be saved and retrieved

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -696,13 +696,25 @@ def check_attr_in_uniqueness_constraint(self, attr: str) -> bool:
                     return True
         return False
 
+    def check_attr_in_single_uniqueness_constraint(self, attr: str) -> bool:
```
ajtmccarty: look like this could share a private method with the `check_attr_in_uniqueness_constraint` method above
something like `_check_field_in_uniqueness_constraint(self, field: str, max_fields_in_constraint: int | None = None) -> bool:`

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -696,13 +696,25 @@ def check_attr_in_uniqueness_constraint(self, attr: str) -> bool:
...
+        return False
+
     def check_if_attr_supports_profiles(self, attribute_schema: AttributeSchema) -> bool:
         return attribute_schema.support_profiles and not self.check_attr_in_uniqueness_constraint(
             attr=attribute_schema.name
         )
 
     def check_if_attr_supports_templates(self, attribute_schema: AttributeSchema) -> bool:
-        return attribute_schema.support_templates and not self.check_attr_in_uniqueness_constraint(
+        return attribute_schema.support_templates and not self.check_attr_in_single_uniqueness_constraint(
```
ajtmccarty: does this mean that only unique attributes are prevented from being added to object template schemas? if so, `SchemaBranch.sync_uniqueness_constraints_and_unique_attributes` already sets any attributes that are in single-item uniqueness constraints to be `unique=True`, so this check would not be necessary. but I am surprised that the requirement is different for templates vs profiles. can you please add a comment or update the description to describe why they are treated differently?

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
```diff
@@ -0,0 +1,502 @@
...
+from ..shared import load_schema
+from .shared import TestSchemaLifecycleBase
+
+if TYPE_CHECKING:
+    from infrahub_sdk import InfrahubClient
+
+    from infrahub.core.branch import Branch
+    from infrahub.database import InfrahubDatabase
+
+LATEST_ATTRIBUTE_PATH_STATUS_QUERY = """
```
ajtmccarty: this query and the logic to run it should be moved somewhere that it can be called without needing to copy the query across test files. maybe the `db_validation` file.

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
ajtmccarty, line 1: these integration tests are pretty slow and each requires a decent amount of setup and teardown. I think they would run faster if they were combined into 1 more complex test that can verify updating a uniqueness constraint causes profiles/templates to have attributes added/removed

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
ajtmccarty: the logic that uses the query should be consolidated to a single place too

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -4690,3 +4695,51 @@ async def test_manage_object_templates_component_relationship_to_excluded_kind()
...
```
ajtmccarty, line 4710: should probably be a corresponding test for generated templates

changelog/7644.added.md
```diff
@@ -0,0 +1 @@
+Added condition to restrict the addition of unique and optional attributes to generated profile schemas
```
ajtmccarty: templates are changed too and should be described here

backend/tests/component/core/migrations/schema/test_node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,420 @@
...
+    execution_result = await migration.execute(migration_input=MigrationInput(db=db, at=at), branch=default_branch)
+
+    assert not execution_result.errors
+    after_snapshot = await snapshotter.snapshot()
+    assert_edge_timestamps(before_snapshot, after_snapshot, at.to_string())
+
+
+# ---------------------------------------------------------------------------
+# Template fixtures
+# ---------------------------------------------------------------------------
```
ajtmccarty: it's worth considering combining these tests with the profile tests above b/c they run the same migrations and it would save time

backend/infrahub/core/schema/basenode_schema.py
ajtmccarty: I still don't understand why templates are different from profiles in this case. the PR description needs an update to cover all of the changes in here

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,96 @@
...
+                attribute_schema=prev_attr
+            )
+            new_supports_templates = self.new_schema.check_if_attr_supports_templates(attribute_schema=attr)
+
+            queries_to_run: list[type[MigrationBaseQuery]] = []
+
+            if self.new_schema.generate_profile and previous_supports_profiles != new_supports_profiles:
+                queries_to_run.append(
+                    ProfilesAttributeRemoveMigrationQuery
+                    if not new_supports_profiles
```
ajtmccarty: I think using the explicit Boolean comparison is more clear here `if new_supports_profiles is False`

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,96 @@
...
+                )
+
+            if (
+                isinstance(self.new_schema, NodeSchema)
+                and self.new_schema.generate_template
+                and previous_supports_templates != new_supports_templates
+            ):
+                queries_to_run.append(
+                    TemplatesAttributeRemoveMigrationQuery
+                    if not new_supports_templates
```
ajtmccarty: same here

backend/infrahub/core/migrations/schema/node_uniqueness_constraints_update.py
```diff
@@ -0,0 +1,96 @@
...
+                    else TemplatesAttributeAddMigrationQuery
+                )
+
+            if not queries_to_run:
+                continue
+
+            attr_migration = AttributeSchemaMigration(
+                name=f"node.uniqueness_constraints.update.{attr.name}",
+                queries=[],
```
ajtmccarty: can `queries_to_run` be passed in here and then remove the `queries=[query_class]` from the call to `attr_migration.execute` can just call it one time instead of in a loop down below?

backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -687,6 +687,42 @@ def update(self, other: HashableModel) -> Self:
...
+
+        ``max_fields_in_constraint=1`` is passed so that only single-attr constraints
+        block the attribute. A compound constraint such as ``[["name__value", "nbr_seats__value"]]``
+        does not uniquely identify a node by ``nbr_seats`` alone, so the attribute is still
+        safe to override in a template. A single-attr constraint like ``[["nbr_seats__value"]]``
+        does uniquely identify a node, so the attribute must be excluded from templates.
+        """
+        return attribute_schema.support_templates and not self._check_attr_in_uniqueness_constraint(
+            attr=attribute_schema.name, max_fields_in_constraint=1
+        )
```
ajtmccarty: upon further review of how templates are used during object creation, I don't think that we need this method, although you can keep it you want an interface matching the `check_if_attr_support_profiles` method.

you are right that attributes should only be excluded from Template schemas if the attribute is unique. the `SchemaBranch.sync_uniqueness_constraints_and_unique_attributes()` already handles ensuring that an attribute in a single-item uniqueness constraints is set to unique and vice versa, so I think that we can count on the `Attribute.support_templates` method to tell us whether an attribute should be included in Template Schemas or not.

however, there is currently no migration that runs for an update to an attribute's uniqueness and I think a new migration will be required to handle this case. I documented it in https://github.com/opsmill/infrahub/issues/8894

so I think that template-related logic can be removed from this PR and updated in a PR aimed at that new bug

ajtmccarty (commented):
sorry for all the back-and-forth on this one, but I think you can remove Template-related changes from this PR and they can be addressed in PRs targeting #8722 and #8894

Templates are simpler than Profiles b/c Templates only apply their values during object creation, but Profiles can apply their values whenever they are linked to an object or updated. So Profiles need more strict constraints to prevent them from violation uniqueness constraints than Templates do. There are still bugs to be addressed regarding the Templates and unique attributes, but they are different enough that they should be separate from this PR

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
```diff
@@ -0,0 +1,250 @@
...
+from .shared import TestSchemaLifecycleBase
+
+if TYPE_CHECKING:
+    from infrahub_sdk import InfrahubClient
+
+    from infrahub.core.branch import Branch
+    from infrahub.database import InfrahubDatabase
+
+
+async def assert_attribute_absent(
```
ajtmccarty, line 26: looks like it belongs in `db_validation.py` with the query string it uses

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
```diff
@@ -0,0 +1,250 @@
...
+    query = LATEST_ATTRIBUTE_PATH_STATUS_QUERY % {"label": node_label}
+    results = await db.execute_query(query=query, params={"attr_name": attr_name, "branch_name": branch_name})
+    assert len(results) == 0, f"Expected no active/deleted {node_label}.{attr_name} edges, found {len(results)}"
+
+
+CAR_KIND = "TestingCar"
+PROFILE_CAR_KIND = "ProfileTestingCar"
+
+
+class TestUniquenessConstraintMigrationAddToConstraint(TestSchemaLifecycleBase):
```
ajtmccarty, line 41: would be good to also retrieve the `Profile` instance in these tests and confirm that the expected attributes are accessible or not

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
```diff
@@ -0,0 +1,250 @@
...
+            "uniqueness_constraints": [["name__value", "nbr_seats__value"]],
+            "attributes": [
+                {"name": "name", "kind": "Text"},
+                {"name": "nbr_seats", "kind": "Number", "optional": True},
+                {"name": "color", "kind": "Text", "optional": True},
+            ],
+        }
+
+    @pytest.fixture(scope="class")
+    def schema_car_name_unique_no_compound(self) -> dict[str, Any]:
```
ajtmccarty, line 162: can use copy on the `schema_car_compound_constraint_only` fixture and then update it to reduce line count and make this simpler to update

backend/tests/integration/schema_lifecycle/test_migration_uniqueness_constraints.py
```diff
@@ -0,0 +1,250 @@
...
+        response = await client.schema.load(schemas=[schema_step_02])
+        assert not response.errors
+
+        await assert_attribute_path_status(
+            db=db,
+            node_label=PROFILE_CAR_KIND,
+            attr_name="nbr_seats",
+            branch_name=default_branch.name,
+            expected_status="active",
+        )
```
ajtmccarty, line 245: would be nice to also test updating the profile instance to include a `nbr_seats` value now and retrieve it to verify it is set

ajtmccarty (approved):
just a few little test update suggestions

ajtmccarty:
maybe should target `release-1.9`. or even `stable`.
also, worth considering a database migration to clean up possible incorrect Profile instances affected by the bug this PR fixes

PR #8590, opened 2026-03-12
IFC-2303: CoreKeyValue schema definitions for webhook custom headers

polmichel:
# Why

Webhooks currently have no way to send custom HTTP headers (authentication tokens, API keys, metadata). This blocks integrations that require auth headers or tenant identification.

This PR adds the schema layer: a `CoreKeyValue` generic with three node type implementations, and a `headers` relationship on `CoreWebhook`.

Non-goals: internal .md documentation resulting from /feedback session with Claude.

Closes [IFC-2303]

## What changed

- **New schema entities**: `CoreKeyValue` generic (name, key, description) with `CoreStaticKeyValue`, `CoreEnvironmentVariableKeyValue` node schemas — all branch-agnostic
- **Webhook relationship**: Added `headers` relationship (peer=CoreKeyValue, cardinality=MANY, kind=ATTRIBUTE) to `CoreWebhook` generic
- **Constants**: Added `InfrahubKind` entries for all three new types
- **Generated files**: Regenerated `protocols.py` and test protocols to reflect new schema

No behavioral changes — this is purely additive schema definitions. API contract unchanged for existing webhooks (headers relationship is optional).

## How to review

1. Start with `backend/infrahub/core/schema/definitions/core/key_value.py` — the new schema definitions
2. Then `backend/infrahub/core/schema/definitions/core/webhook.py` — the headers relationship addition

## How to test

```bash
uv run invoke backend.generate   # Should succeed with new schemas
uv run invoke format && uv run invoke lint   # Should pass
```

## Impact & rollout

- **Deployment notes:** Safe to deploy. Need synchronization with a PR for schema-update from SDK side.

## Checklist

- [x] Tests added/updated

[IFC-2303]: https://opsmill.atlassian.net/browse/IFC-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/core/schema/definitions/core/key_value.py
```diff
@@ -0,0 +1,81 @@
...
+    generate_profile=False,
+    attributes=[
+        Attr(
+            name="name",
+            kind="Text",
+            unique=True,
+            description="Unique name identifying this key-value pair",
+            order_weight=1000,
+        ),
+        Attr(name="key", kind="Text", description="The key portion of the key-value pair", order_weight=2000),
```
ogenstad, line 31: I think that it could make sense to store the "value" as an attribute in the generic as well. Even if we override it in some of the nodes. The only real reason to override it would be to assign another description to it though.

The reason to have it on the generic would be that we'd be able to have a GraphQL query that targets the generic only and just include `__typename` to find the specific type of node. If we don't have the "value" on the generic we'd always have to query for specific sub types in order to get the value.

backend/infrahub/core/constants/infrahubkind.py
```diff
@@ -19,6 +19,7 @@
 CREDENTIAL = "CoreCredential"
 DATACHECK = "CoreDataCheck"
 DATAVALIDATOR = "CoreDataValidator"
+ENVIRONMENTVARIABLEKEYVALUE = "CoreEnvironmentVariableKeyValue"
```
ogenstad: Not really a problem, but I'm wondering if the name is a bit long. This would mostly be relevant from a UI point of view, for example if you're creating one of these variables from the generic. Potentially the UI could look a bit weird, not sure.

While I like that the name is descriptive, I wonder if it would be as easy to understan if we just had `CoreEnvKeyValue`. What do you think?

backend/infrahub/core/schema/definitions/core/key_value.py
polmichel: Thanks for the feedback. Actually, this is the root cause of the bug I am troubleshooting right now after some application testing

backend/infrahub/core/constants/infrahubkind.py
polmichel: Sounds good to me!

ogenstad (approved):

ogenstad (approved):

PR #8593, opened 2026-03-12
Stable to release 1.8

ajtmccarty:
resolved conflicts

- deleted `docs/docs/getting-started`
- incremented migrations to insert migration 059 from stable

## Release 1.7.7

* **Bug Fixes**
  * Fixed display labels for relationship-based fields no longer showing "None" after upsert
  * Corrected human-friendly IDs to properly load during node updates
  * Included migration to fix objects with display labels or HFIDs that were incorrectly set to null

* **Documentation**
  * Reorganized documentation with new Overview section structure
  * Added comprehensive Quick Start guide for new users
  * Enhanced documentation navigation and layout


docs/docs/getting-started/concepts.mdx
ajtmccarty, line 1: @BaptisteGi looks like these were deleted on `stable`, so I think deleting the `getting-started` directory here is the right resolution, but want to be sure

ogenstad (approved):

PR #8594, opened 2026-03-12
IHS-114: Add diff data api endpoint

solababs:
# Why

The Python SDK's `client.branch.diff_data()` called `GET /api/diff/data` but that endpoint did not exist, returning a 404. Additionally, the SDK's URL builder was missing a `?` separator before query parameters, producing malformed URLs like `/api/diff/databranch=test-branch&branch_only=true`.

Goal: make `client.branch.diff_data()` work end-to-end by implementing the missing REST endpoint and fixing the URL generation bug.

SDK PR https://github.com/opsmill/infrahub-sdk-python/pull/865

Closes https://github.com/opsmill/infrahub-sdk-python/issues/325

## What changed

- **New endpoint `GET /api/diff/data`** — returns the enriched data-level diff for a branch as a JSON array. Uses `DiffRepository` (the same backend used by the GraphQL diff tree query) and `jsonable_encoder` with custom encoders for `Timestamp` and `TrackingId` to produce a fully serializable response.
- **Integration test** — added `TestDiffAPI.test_diff_data_api_endpoint` in `tests/integration_docker/api/test_diff.py` to confirm the endpoint responds and returns a list.

No schema changes. No existing endpoint behavior changed. The `/api/diff/files` and `/api/diff/artifacts` endpoints are untouched.

## How to review

1. `backend/infrahub/api/diff/diff.py` — the new `get_diff_data` endpoint and its `jsonable_encoder` usage with `response_model=None`
2. `backend/tests/integration_docker/api/test_diff.py` — the integration test

The `response_model=None` on the endpoint decorator is intentional: `EnrichedDiffRoot` is a dataclass with non-Pydantic types (`Timestamp`, `TrackingId`), so FastAPI's automatic Pydantic serialization is bypassed in favour of explicit `jsonable_encoder` pre-processing.

## How to test

```bash
# Run the integration test
uv run pytest backend/tests/integration_docker/api/test_diff.py -v
```

Expected: `test_diff_data_api_endpoint` returns a list (empty `[]` for a branch with no changes).

## Impact & rollout

- **Backward compatibility:** additive only — new endpoint, no existing behaviour changed
- **Performance:** no impact on existing endpoints
- **Config/env changes:** none
- **Deployment notes:** safe to deploy

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint to retrieve enriched diff data with optional branch and time-range query parameters.

* **Tests**
  * Added integration test coverage for the diff data API endpoint.

* **Documentation**
  * OpenAPI/schema and generated client types updated to include the new endpoint.
  * Changelog entry added documenting the new diff data API.

### DIFF
--- a/backend/infrahub/api/diff/diff.py
+++ b/backend/infrahub/api/diff/diff.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING
 
 from fastapi import APIRouter, Depends, Request
+from fastapi.encoders import jsonable_encoder
 
 from infrahub.api.dependencies import get_branch_dep, get_current_user, get_db
 from infrahub.core import registry
@@ -14,6 +15,10 @@ from infrahub.core.diff.model.diff import (
     BranchDiffFile,
     BranchDiffRepository,
 )
+from infrahub.core.diff.model.path import EnrichedDiffRoot, TrackingId
+from infrahub.core.diff.repository.repository import DiffRepository
+from infrahub.core.timestamp import Timestamp
+from infrahub.dependencies.registry import get_component_registry
 
 if TYPE_CHECKING:
     from infrahub.auth import AccountSession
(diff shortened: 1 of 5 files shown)


ogenstad:
Looking at this I think we want to solve this in a different way where the SDK would use the existing features to get access to the diff so that we don't have to introduce a new endpoint specifically for the SDK. I.e. use existing queries / mutations. If this is problematic with the current SDK method perhaps that one should simply be removed and refactored instead. I'd suggest we park this and make some progress on the 1.9 work. Then we can have a discussion around what the expectations would be from the SDK point of view, as it's been broken for quite some time (I don't know if it ever worked or if we at some point removed an old diff rest API (which would have been a loong time ago)

saltas888:
We reworked this following Patrick's suggestion, so no new REST endpoint. The SDK method never worked and got removed in opsmill/infrahub-sdk-python#1229, and `get_diff_tree()` was extended there with `include_properties=True` to cover the data-level diff through the existing DiffTree GraphQL query. The scale test caller is switched over in #10144. Closing this one.

PR #8597, opened 2026-03-13
IFC-2304: Runtime resolution of webhook custom headers

polmichel:
# Why

Webhook custom headers defined via `CoreKeyValue` schema entities (from [IFC-2303](#8590)) need to be resolved at send time. Static headers pass their value through directly, while environment variable headers must read from `os.environ` on the worker at dispatch time. Without this, the schema-level header definitions have no runtime effect.

Closes [IFC-2304]

## What changed

- **Header resolution at send time**: Static headers are passed through as-is; environment variable headers are resolved from `os.environ` on the worker when the webhook fires. Missing env vars are logged as warnings and skipped (no crash).
- **Custom headers override defaults**: Headers defined on a webhook can override built-in defaults like `Content-Type`, giving full control over the HTTP request.
- **Headers extracted from schema relationships**: When a webhook is loaded (from DB or cache), its `CoreKeyValue` peers are mapped to a lightweight runtime model that carries through the send pipeline.
- **Cache-aware**: Custom headers survive the `to_cache()`/`from_cache()` roundtrip so cached webhooks retain their header configuration.

## How to review

1. Start with `backend/infrahub/webhook/models.py` — the `WebhookHeader` model and changes to `_assign_headers()`
2. Then `backend/infrahub/webhook/tasks/process.py` — the extraction logic and plumbing into `convert_node_to_webhook`
3. Finally the tests in `backend/tests/unit/webhook/`

## How to test

```bash
uv run pytest backend/tests/unit/webhook/
```

## Impact & rollout

- **Deployment notes:** Requires the schema PR ([#8590]) to be merged first. Environment variables referenced by `CoreEnvironmentVariableKeyValue` headers must be set on the worker environment.

## Checklist

- [x] Tests added/updated

[IFC-2304]: https://opsmill.atlassian.net/browse/IFC-2304
[#8590]: https://github.com/opsmill/infrahub/pull/8590

[IFC-2303]: https://opsmill.atlassian.net/browse/IFC-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -4,6 +4,8 @@ import base64
 import hashlib
 import hmac
 import json
+import logging
+import os
 from typing import TYPE_CHECKING, Any, Literal
 from uuid import UUID, uuid4
 
@@ -18,6 +20,8 @@ from infrahub.git.repository import InfrahubReadOnlyRepository, InfrahubReposito
 from infrahub.trigger.constants import NAME_SEPARATOR
 from infrahub.trigger.models import EventTrigger, ExecuteWorkflow, TriggerDefinition, TriggerType
 from infrahub.workflows.catalogue import WEBHOOK_PROCESS
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from httpx import Response
@@ -118,6 +122,20 @@ class WebhookHeader(BaseModel):
     value: str
     kind: Literal["static", "environment"]
 
+    def resolve(self) -> str | None:
+        """Resolve the header value based on its kind.
+
+        Returns None if the value cannot be resolved (e.g. missing environment variable).
+        """
+        if self.kind == "static":
+            return self.value
+
+        resolved = os.environ.get(self.value)
+        if resolved is None:
+            logger.warning("Environment variable '%s' not found, skipping header '%s'", self.value, self.key)
+            return None
+        return resolved
+
 
 class Webhook(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
@@ -140,8 +158,9 @@ class Webhook(BaseModel):
         }
 
         for header in self.custom_headers:
-            if header.kind == "static":
-                self._headers[header.key] = header.value
+            resolved = header.resolve()
+            if resolved is not None:
+                self._headers[header.key] = resolved
 
         if self.shared_key:
             message_id = f"msg_{uuid.hex}" if uuid else f"msg_{uuid4().hex}"
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -4,7 +4,7 @@ import base64
 import hashlib
 import hmac
 import json
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 from uuid import UUID, uuid4
 
 from pydantic import BaseModel, ConfigDict, Field, computed_field
@@ -113,12 +113,19 @@ class EventContext(BaseModel):
         )
 
 
+class WebhookHeader(BaseModel):
+    key: str
+    value: str
+    kind: Literal["static", "environment"]
+
+
 class Webhook(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
     name: str = Field(...)
     url: str = Field(...)
     event_type: str = Field(...)
     validate_certificates: bool | None = Field(...)
+    custom_headers: list[WebhookHeader] = Field(default_factory=list)
     _payload: Any = None
     _headers: dict[str, Any] | None = None
     shared_key: str | None = Field(default=None, description="Shared key for signing the webhook requests")
@@ -131,6 +138,10 @@ class Webhook(BaseModel):
             "Accept": "application/json",
             "Content-Type": "application/json",
         }
+
+        for header in self.custom_headers:
+            if header.kind == "static":
+                self._headers[header.key] = header.value
 
         if self.shared_key:
             message_id = f"msg_{uuid.hex}" if uuid else f"msg_{uuid4().hex}"
@@ -184,25 +195,27 @@ class CustomWebhook(Webhook):
     """Custom webhook"""
 
     @classmethod
-    def from_object(cls, obj: CoreCustomWebhook) -> Self:
+    def from_object(cls, obj: CoreCustomWebhook, custom_headers: list[WebhookHeader] | None = None) -> Self:
         return cls(
             name=obj.name.value,
             url=obj.url.value,
             event_type=obj.event_type.value,
             validate_certificates=obj.validate_certificates.value or False,
             shared_key=obj.shared_key.value,
+            custom_headers=custom_headers or [],
         )
 
 
 class StandardWebhook(Webhook):
     @classmethod
-    def from_object(cls, obj: CoreStandardWebhook) -> Self:
+    def from_object(cls, obj: CoreStandardWebhook, custom_headers: list[WebhookHeader] | None = None) -> Self:
         return cls(
             name=obj.name.value,
             url=obj.url.value,
             event_type=obj.event_type.value,
             validate_certificates=obj.validate_certificates.value or False,
             shared_key=obj.shared_key.value,
+            custom_headers=custom_headers or [],
         )
 
 
@@ -238,7 +251,9 @@ class TransformWebhook(Webhook):
         )  # type: ignore[call-overload]
 
     @classmethod
-    def from_object(cls, obj: CoreCustomWebhook, transform: CoreTransformPython) -> Self:
+    def from_object(
+        cls, obj: CoreCustomWebhook, transform: CoreTransformPython, custom_headers: list[WebhookHeader] | None = None
+    ) -> Self:
         return cls(
             name=obj.name.value,
             url=obj.url.value,
@@ -253,4 +268,5 @@ class TransformWebhook(Webhook):
             transform_timeout=transform.timeout.value,
             convert_query_response=transform.convert_query_response.value or False,
             shared_key=obj.shared_key.value,
+            custom_headers=custom_headers or [],
         )
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -117,23 +117,26 @@ class EventContext(BaseModel):
         )
 
 
+class WebhookHeaderResolutionError(Exception):
+    pass
+
+
 class WebhookHeader(BaseModel):
     key: str
     value: str
     kind: Literal["static", "environment"]
 
-    def resolve(self) -> str | None:
+    def resolve(self) -> str:
         """Resolve the header value based on its kind.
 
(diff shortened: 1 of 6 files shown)


backend/infrahub/webhook/tasks/process.py
```diff
@@ -36,15 +36,36 @@ async def webhook_send(webhook: Webhook, context: EventContext, event_data: dict
...
+
+def _extract_custom_headers(webhook_node: CoreWebhook) -> list[WebhookHeader]:
+    """Extract WebhookHeader list from a webhook node's headers relationship."""
+    if not hasattr(webhook_node, "headers"):
+        return []
+    headers: list[WebhookHeader] = []
+    for peer in webhook_node.headers.peers:
+        kind = KIND_MAP.get(peer.get_kind())
+        if kind is None:
+            continue
```
ogenstad, line 54: Just wanted to highlight here that this is an example of why I wanted to have that ability to limit the namespace for a generic. Since without that feature we wouldn't be able to know if users were to add a new type of key-value pairs within their schema and in this case those additions would just silently fail which would be confusing. Once we've locked that down we shouldn't need to require this type of safe guard since we'd know all of the valid types.

backend/infrahub/webhook/tasks/process.py
```diff
@@ -36,15 +36,36 @@ async def webhook_send(webhook: Webhook, context: EventContext, event_data: dict
...
+        headers.append(WebhookHeader(key=peer.key.value, value=peer.value.value, kind=kind))
+    return headers
+
+
 @task(name="webhook-convert-node", task_run_name="Convert node to webhook", cache_policy=NONE)
 async def convert_node_to_webhook(webhook_node: CoreWebhook, client: InfrahubClient) -> Webhook:
     webhook_kind = webhook_node.get_kind()
 
     if webhook_kind not in ["CoreStandardWebhook", "CoreCustomWebhook"]:
         raise ValueError(f"Unsupported webhook kind: {webhook_kind}")
```
ogenstad, line 64: Outside of the diff, but another thing we could get rid of once we had that namespace restriction in place.

backend/infrahub/webhook/models.py
```diff
@@ -113,12 +117,33 @@ def from_event(cls, event_id: str, event_type: str, event_occured_at: str, event
...
+        Returns None if the value cannot be resolved (e.g. missing environment variable).
+        """
+        if self.kind == "static":
+            return self.value
+
+        resolved = os.environ.get(self.value)
+        if resolved is None:
+            logger.warning("Environment variable '%s' not found, skipping header '%s'", self.value, self.key)
+            return None
+        return resolved
```
ogenstad, line 140: I like the fact that the kind is a `Literal` instead of a plain string. However I think it could be clarified a bit more with having a string enum instead. Then we can also modify the logic to account for this. One thing that we've started to use in a few places is assert_never. An example is this (just search for it in the code base):

```python
        match self.node_ordering.order_by:
            case OrderByField.ID:
                self.order_by = [f"{db.get_id_function_name()}(n)"]
            case OrderByField.CREATED_AT:
                self.order_by = [f"n.created_at {self.node_ordering.direction.value}"]
            case OrderByField.UPDATED_AT:
                self.order_by = [f"n.updated_at {self.node_ordering.direction.value}"]
            case _:
                assert_never(self.node_ordering.order_by)
```

The reason for having this instead is that it could catch issues when in the future we add a new type we could raise warnings within the code. An example here would be if we added back that password based option anything after we return for matching the "static" option would at that point be invalid code and something that we could potentially miss.

backend/infrahub/webhook/models.py
```diff
@@ -113,12 +117,33 @@ def from_event(cls, event_id: str, event_type: str, event_occured_at: str, event
...
+        """Resolve the header value based on its kind.
+
+        Returns None if the value cannot be resolved (e.g. missing environment variable).
+        """
+        if self.kind == "static":
+            return self.value
+
+        resolved = os.environ.get(self.value)
+        if resolved is None:
+            logger.warning("Environment variable '%s' not found, skipping header '%s'", self.value, self.key)
```
ogenstad: Considering the context here, it might be easier to track where the error comes from if we log this from the  _assign_headers() method of the webhook. This might be overkill though as we at least can see the name of the variable here and there shouldn't be any overlap. I.e. either it's set or it isn't. But the key itself isn't enough to actually figure out where it's coming from or why it is set. Also since we don't have a relationship going back from the key value pairs to the webhook if we have some environment with a high number of these key value nodes it might take a while to track down the webhook where it's used.

An option could be that .resolve() would return some HeaderResolve object that could contain a value but also a log warning that could be used from ._assign_headers()

What do you think?

backend/infrahub/webhook/models.py
polmichel: It makes sense! This is better from an error management perspective instead of just skipping the node.

polmichel: In that case, should we not also change (at some point, not in this PR) `InfrahubKind` into a `StrEnum`?

It would allow us to correctly type the dictionary 
```
KIND_MAP: dict[InfrahubKind, WebhookHeaderKind] = {
  InfrahubKind.STATICKEYVALUE: WebhookHeaderKind.STATIC,
  InfrahubKind.ENVKEYVALUE: WebhookHeaderKind.ENVIRONMENT,
}
```

Or correctly type some method as `infrahub.core.node.Node.init`

```
@classmethod
    async def init(
        cls,
        schema: NodeSchema | ProfileSchema | TemplateSchema | InfrahubKind | type[SchemaProtocol],
        db: InfrahubDatabase,
        branch: Branch | str | None = None,
        at: Timestamp | str | None = None,
    ) -> Self | SchemaProtocol:
```

ogenstad: I think that we probably want to get rid of "InfrahubKind" all together and remove those constants and instead use the SchemaProtocol classes where we have "InfrahubKind" today.

backend/infrahub/webhook/models.py
polmichel: Before we jump into a solution, I don't think users should have to find the related webhook. As you said, and since the key/value are also branch-agnostic, users should not have multiple environment variable key/value objects with the same environment variable name (even though we are allowing them to do so) and the same key. 
If they want to troubleshoot, they just have to find the correct environment variable key/value object relating to their environment variable.
image

However, perhaps I am missing something about how these environment variables will ultimately be used.


Regarding the solution you propose, I think this is a better design, whether or not we want to resolve the problem we are discussing. I will design something to add these additional logging capabilities.

backend/infrahub/webhook/models.py
polmichel: I was wondering if I had done it elsewhere, and I have done it in this PR: https://github.com/opsmill/infrahub/pull/8602. 
In this commit: https://github.com/opsmill/infrahub/pull/8602/changes/312e21312ea352b3e98ba946045012c8ec40195c

However, I still cannot type the dictionary keys, so my last comment is still valid in this respect. (EDIT: I had not seen your comment when I wrote this last sentence, ok for me).

backend/infrahub/webhook/models.py
polmichel: I decided not to use an intermediate `HeaderResolve` object because it handled two different aspects: logging capabilities and resolving the value.
To avoid this duplication, I let the `resolve()` method throw a dedicated error when the environment variable is not found. It is then up to the caller of the `resolve()` method to decide what to do when an error is received. It allows separating the "controller" from the "functional domain".

What do you think about this?

Cf -> https://github.com/opsmill/infrahub/pull/8597/changes/c2b4f1fa26818e8ad08de71e0b093826231829e6

backend/infrahub/webhook/tasks/process.py
polmichel: For now, I have noted two enhancements related to web hooks and external initiatives: migration from `ujson` to `orjson`, and the application of namespace restrictions. I will wrote them into the epic Jira issue.

backend/infrahub/webhook/models.py
ogenstad: `InfrahubKind` is a bit of a hack now as it's just a module that we're importing with an alias:

```python
from infrahub.core.constants import infrahubkind as InfrahubKind  # noqa: N812
```
We could potentially consider changing that but I think we should have a larger discussion around it first.

The way I viewed repositories after we have the namespace restriction was that we'd have some enum for repository types. Perhaps it would be the same for these key value pairs.

For now perhaps KIND_MAP could be a TypedDict if we want to lock it down for now?

backend/infrahub/webhook/models.py
ogenstad: That works for me. In the real world I don't expect people to have a high number of webhooks. But let's say that they do, then they have a high number of these environment variable objects. If one of the environment variables are missing and we only get the key value of the environment variable in the logs then an admin will want to correct that behaviour. It's easy enough to find the key/value node that references the missing variable. However, in order to figure out the correct value the admin probably wants to check where this is being used and from the key/value node there's no relationship back to the webhook (I don't think we want one). So the admin would have to go through the various webhooks to check which one is using the key/value node with a missing environment variable. My point was that if the log message included this information from the start it would be easier to tie everything together. Not sure if this was clear from the start, but does that make more sense?

polmichel: If I understand correctly, the focus here is on verifying the correct behavior of the web hook rather than just finding the right key/value to fix the issue. I was focused on the latter and assumed that each environment variables with the same value would be affected if one of them was not resolved (which, after some thoughts, is probably not the case).

backend/infrahub/webhook/models.py
polmichel: I am puzzled as to whether it is necessary to lock down the typing of this dictionary for now.

- `TypedDict` seems a bit over-engineered to me, and I don't think it's meant to be used this way.
- Using `dict[Literal["CoreStaticKeyValue", "CoreEnvKeyValue"]]` breaks the dependency. 
 
I guess we would be made aware of any changes related to this feature, even if this is not ideal.

What do you think about keeping it typed as it is for now?

backend/infrahub/webhook/models.py
ogenstad: My reasoning is basically that if the value isn't set you'd probably need to know which webhook is supposed to use the environment variable in order to know what value to set in the environment variable. In most real world scenarios I wouldn't expect people to have a high number of variables or webhooks so might not be a big issue. Another consideration is if we start to use these key/value things for something else in the future. In that case it might not even be clear that the lack of an environment variable is impacting a webhook at all.

backend/infrahub/webhook/models.py
ogenstad: Sure, that works for me.

ogenstad (approved):
LGTM. Nice work! I think we can revisit some of the typing parts once we've locked down the namespace and consider our options at that point.

PR #8431, opened 2026-02-20
Add specs for branch delete after merge

ogenstad:
# Why

Adds Spec Kit specification for INFP-389

## What changed

Added markdown files by Spec Kit

## Summary by CodeRabbit

* **Documentation**
  * Added specification for "Delete Branch After Merge" feature enabling automatic deletion of branches following merge events, with global configuration options and built-in safeguards to protect default branches.

PR #8600, opened 2026-03-13
Add config settings

solababs:
# Why

There is no configuration surface for controlling automatic branch deletion after merge. Without these settings, the feature cannot be safely rolled out — behavior needs to default to off and be explicitly opt-in.

This PR adds the two config flags that gate the feature, along with a dependency validator between them. It is Phase 1 of the full implementation plan.

## What changed

- Added `delete_branch_after_merge: bool` (default `false`) to `MainSettings` — when enabled, the Infrahub branch is automatically deleted after a successful merge.
- Added `delete_git_branch_after_merge: bool` (default `false`) to `MainSettings` — when enabled, the corresponding Git branch is also deleted; requires `delete_branch_after_merge` to be enabled.
- Added a `@model_validator` that raises at startup if `delete_git_branch_after_merge=true` without `delete_branch_after_merge=true`, preventing a misconfigured state.
- Added 3 unit tests covering the defaults and the validation error.

No behavioral changes beyond the new config fields — no deletion logic is wired up yet.

### Suggested review order
1. [`backend/infrahub/config.py`](backend/infrahub/config.py) — the two new fields and the validator
2. [`backend/tests/unit/test_config.py`](backend/tests/unit/test_config.py) — unit tests

## How to review

Focus on the validator logic in `config.py` — specifically whether the error message and validation mode (`after`) are appropriate. The rest is straightforward field additions.

No schema, API, or database changes in this PR.

## How to test

```bash
# Run the new tests
uv run pytest backend/tests/unit/test_config.py -v -k "delete_branch"

# Full unit suite
```

Expected: all three new tests pass. Attempting to instantiate `MainSettings(delete_git_branch_after_merge=True, delete_branch_after_merge=False)` raises a `ValidationError` with the message `requires 'delete_branch_after_merge' to be enabled`.

## Impact & rollout

- **Backward compatibility:** No breaking changes. Both settings default to `false`, so existing deployments are unaffected.
- **Config/env changes:** Two new optional settings in `MainSettings`: `delete_branch_after_merge` and `delete_git_branch_after_merge`.
- **Deployment notes:** Safe to deploy independently. The deletion logic is not implemented yet.

## Checklist

- [x] Tests added/updated


backend/infrahub/config.py
```diff
@@ -205,6 +205,21 @@ class MainSettings(BaseSettings):
...
     )
+    delete_branch_after_merge: bool = Field(
+        default=False,
+        description="When enabled, the Infrahub branch is automatically deleted after a successful merge.",
+    )
+    delete_git_branch_after_merge: bool = Field(
+        default=False,
+        description="When enabled, the corresponding Git branch is deleted after the Infrahub branch is deleted. "
+        "Requires delete_branch_after_merge to be enabled.",
+    )
```
ajtmccarty: probably more of a design-level question, but it seems like we should be able to delete a branch on git regardless of whether the branch is being automatically deleted after a merge or manually deleted later

ajtmccarty (approved):

backend/infrahub/config.py
solababs: @ajtmccarty as seen here https://github.com/opsmill/infrahub/pull/8431#discussion_r2923035254, i think this was intentionally designed this way

ajtmccarty (approved):

backend/infrahub/config.py
ogenstad: For context the reasoning around this is around the fact that it could be misleading that the branch within Infrahub is setup to be synced with Git but then we don't have a matching branch in Git for those branches. We might of course run into the same issue if the branch is deleted on the Git server itself.

backend/infrahub/config.py
```diff
@@ -205,6 +205,21 @@ class MainSettings(BaseSettings):
         default=True,
         description="When enabled, diff updates are triggered for active branches after a branch merge.",
     )
+    delete_branch_after_merge: bool = Field(
+        default=False,
+        description="When enabled, the Infrahub branch is automatically deleted after a successful merge.",
+    )
+    delete_git_branch_after_merge: bool = Field(
```
ogenstad: Hm, we have the `GitSettings` section for git specific things. I think it would make more sense to have this setting under that section instead. Even if it would cause changes to the validator. Perhaps the validator could live in the `Settings` class?

backend/tests/unit/test_config.py
```diff
@@ -42,6 +52,21 @@ def test_invalid_git_settings__sync_branch_names() -> None:
         GitSettings(import_sync_branch_names=["main", "infrahub/.*", "release/.*", "a[b"])
 
 
+def test_delete_branch_after_merge_defaults_to_false() -> None:
+    assert MainSettings().delete_branch_after_merge is False
+
+
+def test_delete_git_branch_after_merge_defaults_to_false() -> None:
+    assert GitSettings().delete_git_branch_after_merge is False
```
ogenstad: I think we can get rid of test_delete_branch_after_merge_defaults_to_false and test_delete_git_branch_after_merge_defaults_to_false as they just validate the default setting in the class and we don't have that for any of the other config options.

backend/infrahub/config.py
```diff
@@ -1030,6 +1039,14 @@ class Settings(BaseSettings):
     trace: TraceSettings = TraceSettings()
     experimental_features: ExperimentalFeaturesSettings = ExperimentalFeaturesSettings()
 
+    @model_validator(mode="after")
+    def validate_git_branch_deletion_requires_branch_deletion(self) -> Self:
+        if self.git.delete_git_branch_after_merge and not self.main.delete_branch_after_merge:
+            raise ValueError(
+                "'git.delete_git_branch_after_merge' requires 'main.delete_branch_after_merge' to be enabled"
```
ogenstad: From a users point of view they might not know that the config is divided into sections like this so I think we can remove the "git." and "main." prefix in the error here.

ogenstad (approved):

PR #8601, opened 2026-03-13
allow removing properties from attributes in merge logic

### LINKED ISSUE #8583: bug: attribute source property is not properly updated in main branch after merging a branch that overrides a profile provided value
### Component

API Server / GraphQL

### Infrahub version

1.7.6, 1.8.0rc0

### Current Behavior

When you override a profile provided attribute value on an object in a branch, and then merge that branch into main, the source property of the attribute for the object on the main branch will still point a the profile.

### Expected Behavior

The attribute source property in the main branch, should be empty when a user has overridden that value in the branch that is merged.

It could be another value depending on how the attribute value was overridden.

### Steps to Reproduce

- create a ProfileBuiltinTag object setting a value for the description attribute
- in the main branch create a BuiltinTag object applying the profile and only providing a value for the name attribute
- verify that the source of the description attribute is the profile
- create a new branch
- in the branch change the attribute description value for the tag object
- verify that the source property of the description attribute is now empty in the branch
- merge the branch
- the value of the description attribute of the tag object will be correct, but the source property will still be pointing at the profile

### Additional Information

_No response_

ajtmccarty:
fixes #8583 

the logic that serializes a diff and passes it to a cypher query during a merge was excluding removed properties of attributes if the attribute was not completely removed. for example, if the source of an attribute is deleted on a branch (like if the attribute stops using a profile), then this change would not make it back to the default branch during a merge

fix is a small change to make sure we send the property removals on to the merge query

adds two component tests covering this situation
- one ensures that the diff is correctly including the removed properties when calculated, which it already was
- one ensures that the removed source is merged into the default branch during a merge, which it was not

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of property removals during branch diffs/merges so value-type property removals are suppressed unless the whole attribute is removed.

* **Tests**
  * Added coverage for single-property updates (with and without source propagation) in diff/merge flows.
  * Added tests for source-property changes in diff calculations to validate updated/removed actions.

### DIFF
--- a/backend/infrahub/core/diff/merger/serializer.py
+++ b/backend/infrahub/core/diff/merger/serializer.py
@@ -225,8 +225,13 @@ class DiffMergeSerializer:
                 property_diff=property_diff, python_value_type=python_type
             )
             for action, value in actions_and_values:
-                # we only delete attributes when the whole attribute is deleted
-                if action is DiffAction.REMOVED and attribute_diff.action is not DiffAction.REMOVED:
+                # we only delete attribute values when the whole attribute is deleted,
+                # but other properties can be removed independently
+                if (
+                    action is DiffAction.REMOVED
+                    and attribute_diff.action is not DiffAction.REMOVED
+                    and property_diff.property_type is DatabaseEdgeType.HAS_VALUE
+                ):
                     continue
                 prop_dicts.append(
                     PropertyMergeDict(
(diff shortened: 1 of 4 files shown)


ogenstad (approved):
Would be good with a news fragment for the resolved bug as well.

PR #8602, opened 2026-03-14
IFC-2334: Webhook cache invalidation on KeyValue changes

polmichel:
# Why

When a `CoreKeyValue` node (used as a webhook header) is created, updated, or deleted, the cached webhook configuration becomes stale. Without invalidation, webhooks continue using outdated header values until the cache naturally expires or the server restarts.

Based on #8590 and #8597.

Closes [IFC-2334]

## What changed

- **Cache invalidation flow**: New Prefect task (`invalidate_webhook_headers`) that resolves which webhooks reference a given KeyValue node and clears their cache entries
- **Cypher query**: `KeyValueGetWebhooksQuery` traverses `webhook__headers` relationships to find affected webhooks
- **Trigger registration** (bug fix): `TRIGGER_KEYVALUE_WEBHOOK_INVALIDATE` has been added to `catalogue.py`
- **Tests**: Unit tests for the invalidation task, component tests for the query + cache integration, functional test verifying end-to-end cache clearing

## How to review

Since this PR is based on other PRs, commits to be reviewed start from [Encapsulate raw strings into small detectable objects](https://github.com/opsmill/infrahub/pull/8602/changes/312e21312ea352b3e98ba946045012c8ec40195c)

1. `backend/infrahub/webhook/query.py` — Cypher query joining KeyValue → Relationship → Webhook
2. `backend/infrahub/webhook/tasks/invalidate.py` + `cache.py` — the Prefect flow and cache deletion task
3. `backend/infrahub/webhook/triggers.py` — trigger definition for KeyValue events
4. Tests: `unit/webhook/test_invalidate.py`, `component/webhook/test_query.py`, `functional/webhook/test_invalidate.py`

## How to test

```bash
# Unit tests
uv run pytest backend/tests/unit/webhook/test_invalidate.py -v
uv run pytest backend/tests/unit/trigger/test_catalogue.py -v

# Component tests (requires Neo4j)
uv run pytest backend/tests/component/webhook/test_query.py -v

# Functional tests (requires Prefect + Neo4j)
uv run pytest backend/tests/functional/webhook/test_invalidate.py -v
```

## Impact & rollout

- **Backward compatibility:** No breaking changes
- **Performance:** Invalidation is event-driven (no polling), scoped to affected webhooks only
- **Deployment notes:** Should be merged after PRs #8590 and #8597 

## Checklist

- [x] Tests added/updated

[IFC-2334]: https://opsmill.atlassian.net/browse/IFC-2334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/webhook/query.py
```diff
@@ -0,0 +1,43 @@
...
+        self.keyvalue_id = keyvalue_id
+        super().__init__(**kwargs)
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        self.params["keyvalue_id"] = self.keyvalue_id
+
+        query = """
+        MATCH (kv:CoreKeyValue {uuid: $keyvalue_id})
+              -[e1:IS_RELATED]->(rl:Relationship {name: "webhook__headers"})
+              <-[e2:IS_RELATED]-(webhook:CoreWebhook)
```
ajtmccarty: pretty sure that this needs to support handling `status="deleted"` edges correctly, but I could be wrong if a `CoreKeyValue` is only ever linked to a single `CoreWebHook`

polmichel: Thanks for the feedback, I took the status in account in the new commits.

ajtmccarty: I didn't realize that the CoreKeyValue -> CoreWebhook relationship is 1-to-many. this will actually have to work a little differently
I think this should do it for a branch-agnostic relationship
```
MATCH (kv:CoreKeyValue {uuid: $keyvalue_id})-[:IS_RELATED]->(rl:Relationship {name: "webhook__headers"})<-[:IS_RELATED]-(webhook:CoreWebhook)
WITH DISTINCT kv, webhook
CALL (kv, webhook) {
    MATCH (kv)-[e1:IS_RELATED]->(rl:Relationship {name: "webhook__headers"})<-[e2:IS_RELATED]-(webhook)
    ORDER BY e1.from DESC, e1.status ASC, e2.from DESC, e2.status ASC
    WITH e1.status = "active" AND e1.to IS NULL AND e2.status = "active" AND e2.to IS NULL AS is_active
    RETURN is_active
    LIMIT 1
}
WITH kv, webhook
WHERE is_active = TRUE
```

ogenstad: One other consideration would be if we should use the NodeManager for this query instead of using a custom query. Using a custom query is probably a bit faster but also a bit more fragile if we do any changes and I don't expect this query to be executed very often. What do you guys think of this?

backend/infrahub/webhook/triggers.py
```diff
@@ -23,3 +23,28 @@
         ),
     ],
 )
+
+TRIGGER_KEYVALUE_WEBHOOK_INVALIDATE = BuiltinTriggerDefinition(
+    name="webhook-keyvalue-invalidate",
+    trigger=EventTrigger(
+        events={"infrahub.node.created", "infrahub.node.updated", "infrahub.node.deleted"},
```
ogenstad: We should actually update the hardcoded strings for the event types here (same with the config option above. If we compare with the code for computed attributes it looks like this:

```python
TRIGGER_COMPUTED_ATTRIBUTE_PYTHON_SETUP_COMMIT = BuiltinTriggerDefinition(
    name="computed-attribute-python-setup-on-commit",
    trigger=EventTrigger(events={CommitUpdatedEvent.event_name}),
```

That way we know that we have the correct names for the events if we change something.

One other consideration regarding the events we have:

```python
events={"infrahub.node.created", "infrahub.node.updated", "infrahub.node.deleted"}
```

I think we can get away with only using `infrahub.node.updated`. The reason being that when a node is created it isn't tied to a webhook, however when we tie a new key-value pair to a webhook it would be registered as a `infrahub.node.updated` event for the webhook itself which should also trigger the invalidation of the cache. This should also be true for when we delete a key-value pair tied to a webhook. That it should be marked as a relationship change on the webhook node and trigger it's own `infrahub.node.updated` event. So I think that if we have anything other than key-value updates here that it would cause redundant workflows.

backend/infrahub/webhook/query.py
polmichel: On the one hand, I don't think we'll have any performance issues related to this part of the code.
On the other hand, this query will be tested and could form the basis of any future generalisation regarding key/value objects usage.

Overall, I think using an abstract layer is better - at least for now.

polmichel:
While doing the application testing, there is a corner case I am thinking about regarding cache management.
                                                                                                                                                                                                                                        
When a **KeyValue** update fires `NodeUpdatedEvent`, both `webhook-process` (if the webhook listens to **KeyValue** changes) and `webhook-invalidate-headers` are triggered by the same event. Unless Prefect guarantees ordering between independent automations triggered by the same event, `webhook-process` could execute before `webhook-invalidate-headers` and serve stale cached headers. 

One possible solution could be to have the `webhook-process` flow watch to see if the node that triggered the flow is a **KeyValue**. If so, it could force the data to be retrieved from the database.

This is a hacky solution. I am not sure if the use case I am describing could be a critical issue for customers, i.e. using old headers in the update request of a `KeyValue` node.

ogenstad:
> While doing the application testing, there is a corner case I am thinking about regarding cache management.

I'm not sure we need to care about that. The way I see it is that we don't guarantee that a config change takes place immediately. A similar situation would be if you create a webhook from the first place and have it trigger on node creation events then we won't fire a webhook for that specific event.

Even with the changes you are making here it could be that we add a header to a webhook and then immediately afterwards we update some other unrelated node. There's a chance that one webhook would fire with the old headers before the new config is applied. I think this is fine as long as the convergence time is short.

backend/infrahub/webhook/tasks/cache.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import TYPE_CHECKING
+
+from prefect import task
+from prefect.cache_policies import NONE
+from prefect.logging import get_run_logger
+
+from infrahub.workers.dependencies import get_cache
+
+if TYPE_CHECKING:
+    from collections.abc import Set as AbstractSet
```
ogenstad: Not sure we need this alias for `AbstractSet`.. Should be fine with just `Set` here?

backend/infrahub/webhook/tasks/invalidate.py
```diff
@@ -0,0 +1,69 @@
...
+
+    keyvalue_id = event_data["node_id"] if event_data else None
+    if not keyvalue_id:
+        log.warning("No KeyValue ID provided, skipping")
+        return
+
+    database = await get_database()
+
+    async with database.start_session(read_only=True) as db:
+        webhooks = await NodeManager.query(
```
ogenstad, line 58: FYI, just highlighting here that this would have been problematic if we would have triggered on key-valute delete events as they would already have been removed from the webhooks so we wouldn't have gotten any nodes back with this query.

backend/infrahub/webhook/models.py
```diff
@@ -117,21 +118,26 @@ def from_event(cls, event_id: str, event_type: str, event_occured_at: str, event
...
-    kind: Literal["static", "environment"]
+    kind: HeaderKind
 
     def resolve(self) -> str:
         """Resolve the header value based on its kind.
 
         Raises WebhookHeaderResolutionError if the value cannot be resolved.
         """
-        if self.kind == "static":
+        if self.kind == HeaderKind.STATIC:
```
ogenstad: I think it would be nice to use that match statement here instead of using the if-statement and then end with "assert_never".

ogenstad (approved):

backend/infrahub/webhook/tasks/cache.py
polmichel: Yes, I was wondering whether it was useful to specify the abstraction, given that it came from collections.abc, to allow any implementation of Set later on. However, I agree that it is not necessary and kind of Java-like.

polmichel: This is imported and named this way in some librairies like `pytest` -> `_pytest/main.py`

ogenstad: I noticed that the linter started to complain when you changed this. Let's just go with what you had. I'm not really sure I see the confusion as it would only show up if you imported the deprecated `Set` from the typing module. But I'd rather not have any linting exceptions and go with some form of current recommendations. Sorry about the extra noise.

polmichel: Actually, this is a ruff rule

```
PYI025 Use `from collections.abc import Set as AbstractSet` to avoid confusion with the `set` builtin
  --> backend/infrahub/webhook/tasks/cache.py:12:33
   |
11 | if TYPE_CHECKING:
12 |     from collections.abc import Set
   |                                 ^^^
   |
help: Alias `Set` to `AbstractSet`
```

polmichel: https://docs.astral.sh/ruff/rules/unaliased-collections-abc-set-import/

What it does
Checks for from collections.abc import Set imports that do not alias Set to AbstractSet.

Why is this bad?
The Set type in collections.abc is an abstract base class for set-like types. It is easily confused with, and not equivalent to, the set builtin.

To avoid confusion, Set should be aliased to AbstractSet when imported. This makes it clear that the imported type is an abstract base class, and not the set builtin.

ogenstad: Yeah like I said let's just go with what ruff recommends so we don't need to exclude any rules because of this.

polmichel: (Once again, I have not seen your comment before sending my last comment)

PR #8603, opened 2026-03-14
IFC-2337: Invalidate webhook cache on reconcile_all

polmichel:
# Why

During webhook reconciliation (`reconcile_all`), the webhook cache is never invalidated. Both `_configure_one()` and `_delete_automation()` correctly call `invalidate_webhook_cache`, but `_reconcile_all()` does not. This means stale cached webhook data can persist after a full reconciliation — including for deactivated or modified webhooks.

[IFC-2337](https://opsmill.atlassian.net/browse/IFC-2337)

Builds on #8590, #8597 & #8602.

## What changed

- **New `invalidate_all_webhook_caches` task** (`backend/infrahub/webhook/tasks/cache.py`): uses `cache.list_keys("webhook:*")` to find and delete all webhook cache entries, covering both active and inactive webhooks
- **`_reconcile_all()` now invalidates the cache** after `setup_triggers_specific` completes
- **Extracted `CACHE_KEY_PREFIX` constant** (`backend/infrahub/webhook/constants.py`): replaces hardcoded `"webhook"` prefix strings across `cache.py` and `process.py`

## How to review

First commit to review is: 353549ff794627b603c1cf2827d416d61800cc7f

1. `backend/infrahub/webhook/constants.py`: new `CACHE_KEY_PREFIX` constant
2. `backend/infrahub/webhook/tasks/cache.py`: new `invalidate_all_webhook_caches` task
3. `backend/infrahub/webhook/tasks/configure.py`: one-line addition in `_reconcile_all()`
4. `backend/infrahub/webhook/tasks/process.py`: mechanical: uses `CACHE_KEY_PREFIX` instead of hardcoded string
5. `backend/tests/functional/webhook/test_configure.py`: new test `test_configure_all_invalidates_cache`

## How to test

```bash
uv run pytest backend/tests/unit/webhook/ -xvs
uv run pytest backend/tests/functional/webhook/test_configure.py -xvs
```

## Impact & rollout

- **Deployment notes:** To be merged after the related PRs have been merged #8590, #8597 & #8602.

## Checklist

- [x] Tests added/updated

[IFC-2337]: https://opsmill.atlassian.net/browse/IFC-2337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/webhook/tasks/configure.py
+++ b/backend/infrahub/webhook/tasks/configure.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Set as AbstractSet
 
 from infrahub_sdk.protocols import CoreWebhook
 from prefect import flow
@@ -10,6 +10,7 @@ from prefect.client.orchestration import get_client as get_prefect_client
 from prefect.logging import get_run_logger
 from prefect.runtime import flow_run
 
+from infrahub.trigger.constants import NAME_SEPARATOR
 from infrahub.trigger.models import ExecuteWorkflow, TriggerType
 from infrahub.trigger.setup import gather_all_automations, setup_triggers_specific
 from infrahub.workers.dependencies import get_client, get_database
@@ -17,7 +18,7 @@ from infrahub.workers.dependencies import get_client, get_database
 from ..constants import EVENT_TO_ACTION, WebhookAction
 from ..gather import gather_trigger_webhook
 from ..models import WebhookTriggerDefinition
-from .cache import invalidate_all_webhook_caches, invalidate_webhook_cache
+from .cache import invalidate_webhook_cache
 
 if TYPE_CHECKING:
     from prefect import Flow, State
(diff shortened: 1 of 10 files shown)


backend/infrahub/webhook/tasks/configure.py
```diff
@@ -192,4 +192,5 @@ async def _reconcile_all() -> None:
         triggers = await gather_trigger_webhook(db=db)
 
     await setup_triggers_specific(gatherer=gather_trigger_webhook, db=database, trigger_type=TriggerType.WEBHOOK)  # type: ignore[arg-type]
+    await invalidate_all_webhook_caches()
```
ogenstad: As an alternative to this one other option would be that setup_triggers_specific returns a `TriggerSetupReport` object which methods like `.modified_triggers_with_type()` and `.triggers_with_type()` (you can so a search in the code to see how this is used elsewhere).  With that we could get a list of triggers that have been updated and only invalidate those webhooks. The idea would be that for things that are just created there shouldn't be anything in the cache and if we delete a webhook the cache will be cleared out in two hours anyway.

I think the current approach is fine as well though. Do you have any thoughts around this?

polmichel: I like this idea. I think that having to use a "dump_all" method generally indicates that we don't have full control over our processes or that we're not tracking the data correctly.
This seems to be the right object to use. I will therefore remove the `invalidate_all_webhook_caches` method. However, I would also remove **_deleted_** objects from the cache to prevent behavior based on the overall cache state from being inaccurate (for example, counting the total number of web hooks, although I do not think that's necessary - this is only an example). Moreover, I feel this would be "expected" from a developer or user perspective.

Regarding updated objects:
I noticed that `infrahub.trigger.models.TriggerComparison` has three different states.
```
class TriggerComparison(StrEnum):
    MATCH = "match"  # Expected trigger and actual trigger is identical
    REFRESH = "refresh"  # The branch parameters doesn't match, the hash does, refresh in Prefect but don't run triggers
    UPDATE = "update"  # Neither branch or other data points match, update in Prefect and run triggers
```

I think that we have only to dump data related to `TriggerComparison.UPDATE` and not `TriggerComparison.REFRESH` since the `TriggerType.WEBHOOK` is not branch-specific (I assume because these objects are branch agnostic)

```
    @property
    def is_branch_specific(self) -> bool:
        return self in {
            TriggerType.COMPUTED_ATTR_JINJA2,
            TriggerType.COMPUTED_ATTR_PYTHON,
            TriggerType.COMPUTED_ATTR_PYTHON_QUERY,
            TriggerType.DISPLAY_LABEL_JINJA2,
            TriggerType.HUMAN_FRIENDLY_ID,
        } 
```


(EDIT: I did some additional exploration, I will modify a bit the structure around these files. And I guess a Webhook TriggerType would never ends as REFRESHed status)

ogenstad: Yeah webhooks should never have the REFRESHED status. The report object will also contain entries for deleted Automations, this is a different object type but I think we collect the ID and name of the automations and as such you should be able to use the `.name` param to get to the ID of the webhook.

ogenstad (approved):

PR #8606, opened 2026-03-14
IFC-2257: add unit test for webhook signature validation

polmichel:
# Why

Webhook HMAC signature calculation has had two regressions in the past:
1. Signature was computed on `str(dict)` instead of `json.dumps(dict)`
2. `json.dumps` used default separators (`, ` and `: `) instead of compact (`,`, `:`) after HTTPX 0.28.x switch

The existing `test_standard_webhook_header` test only validates signing with an empty payload, which serializes identically regardless of format — so it catches neither regression.

Closes https://opsmill.atlassian.net/browse/IFC-2257

## What changed

- Added `test_webhook_signature_with_payload` to `backend/tests/unit/webhook/test_models.py`
- Tests HMAC signature against a hardcoded expected value using a non-trivial nested payload
- Any change to JSON serialization format (spaced vs compact) or signing input (str vs JSON) breaks the test

## How to review

Single test addition in `backend/tests/unit/webhook/test_models.py`. The hardcoded signature was verified by mutating the source to reintroduce both historical bugs and confirming the test fails in each case.

## How to test

```bash
uv run invoke backend.test-unit -- -k test_webhook_signature_with_payload
```

## Checklist

- [x] Tests added/updated

### DIFF
--- a/backend/tests/unit/external/__init__.py
+++ b/backend/tests/unit/external/__init__.py
@@ -1,0 +1,1 @@
+"""Unit tests for external and third-party library behavior that Infrahub depends on."""
--- a/backend/tests/unit/external/test_httpx_serialization.py
+++ b/backend/tests/unit/external/test_httpx_serialization.py
@@ -1,0 +1,30 @@
+import json
+
+import httpx
+
+
+def test_httpx_json_serialization_matches_signature_format() -> None:
+    """Guard against httpx changing its JSON serialization.
+
+    The webhook signature is computed over compact JSON (no spaces).
+    If httpx ever changes its serializer this test will break before.
+    """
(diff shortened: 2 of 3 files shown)


ogenstad (commented):
LGTM, it would be nice if we could also capture any changes within the httpx package though.

polmichel:
> LGTM, it would be nice if we could also capture any changes within the httpx package though.

I added a new unit test package named `external` which as the following description
`"""Unit tests for external and third-party library behavior that Infrahub depends on."""`

A new unit test has been added testing the `httpx` serialization directly.

ogenstad (approved):
Thanks!

PR #8611, opened 2026-03-16
IFC-2271 IPAM find closes matched prefix for IP/prefix

gmazoyer:
# Why

This PR adds a parent prefix lookup: when a valid IP address or CIDR prefix is entered, all containing parent prefixes are displayed in a dedicated "Parent Prefixes" section, ordered from most specific to least specific, across all namespaces.

Non-goals: this PR does not modify any existing search behavior. The feature is purely additive, non-IP queries follow the exact same code path as before.

## What changed

- New `IPParentPrefixLookupQuery` Cypher query class that adapts the binary prefix matching algorithm from `IPPrefixReconcileQuery`, queries across all namespaces, returns results ordered by `prefixlen DESC`
- New `_try_parse_ip_or_prefix()` helper that uses `ipaddress` module to detect valid IP/CIDR input
- `NodeEdges` GraphQL type extended with nullable `parent_prefixes: [NodeEdge!]` field
- `search_resolver()` extended: after IPv6 normalization, detects IP/CIDR, runs the parent prefix lookup, always runs existing text search regardless

### Suggested review order
1. `backend/infrahub/core/query/ipam.py` with the new `IPParentPrefixLookupQuery` and `IPParentPrefixResult`
2. `backend/infrahub/graphql/queries/search.py` with `_try_parse_ip_or_prefix()` helper and resolver changes
4. Tests: `backend/tests/unit/graphql/queries/test_search.py` and `backend/tests/component/graphql/queries/test_search.py`

## How to review

- `The Cypher query is the core logic, compare it with `IPPrefixReconcileQuery` for the algorithm it adapts
- V verify the IP detection runs after IPv6 normalization and before text search, and that `parent_prefixes` is only populated when the field is requested

**Risky areas:**
- The Cypher query's `prefixlen <= prefix_and_length[1]` condition (not `=`) is critical — multiple prefixes can share the same binary address with different prefix lengths

## How to test

```bash
uv run pytest backend/tests/unit/graphql/queries/test_search.py -v
uv run pytest backend/tests/component/graphql/queries/test_search.py -v
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. The `parent_prefixes` field is nullable and additive. Clients that don't request it see no difference.
- **Performance:** Non-IP queries follow the exact same code path. IP queries add one additional Cypher query using the existing `binary_address` index.
- **Deployment notes:** Safe to deploy. No migration needed.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/+ifc2271.added.md`)
- [x] Internal .md docs updated (spec, plan, research, data-model, contracts in `dev/specs/infp-431-ipam-closest-prefix/`)

## Summary by CodeRabbit

* **New Features**
  * Global search now detects IP addresses/CIDRs and shows a "Parent Prefixes" section (IPv4 & IPv6), ordered most-to-least specific; exact-match prefixes remain in main results. Supports multi-namespace and branch-aware behavior.

* **Schema**
  * GraphQL contract extended to include optional parent_prefixes on search results so clients can request parent-prefix data.

* **Tests**
  * Added unit and component tests covering parsing, lookup ordering, branch isolation, and edge cases.

* **Documentation**
  * Added changelog, specs, quickstart, plan, and research docs for the feature.


backend/tests/component/graphql/queries/test_search.py
```diff
@@ -474,3 +494,118 @@ async def test_search_anywhere_case_sensitive_enabled(
...
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"search": query},
+    )
+    assert result.errors is None
+    assert result.data
+    return result.data["InfrahubSearchAnywhere"]
+
+
+async def test_search_parent_prefix_ipv4(db: InfrahubDatabase, ip_dataset_01: dict[str, Any], branch: Branch) -> None:
```
gmazoyer, line 514: These tests are testing multiple things within the same test functions, but this helps in saving time during tests. I tried having more simple/individual tests but testing time increased by 25% for this file.

backend/infrahub/graphql/queries/search.py
```diff
@@ -123,6 +139,18 @@ async def search_resolver(
             # Convert any IPv6 address, network or partial address to collapsed format as it might be stored in db.
             q = _collapse_ipv6(q)
 
+        # Detect if the query is a valid IP address or CIDR prefix for parent prefix lookup
+        parsed_ip = _try_parse_ip_or_prefix(q)
```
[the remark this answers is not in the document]
gmazoyer: While you could think this is a good idea, I think it is actually not. If someone searches for 10.10.1.1/24 (copy-pasted interface address), the IP address object is stored as "10.10.1.1/24". Before normalization, text search matches it exactly. After normalization, q becomes "10.10.1.0/24" and the IP address object is lost from results.

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,108 @@ def _get_namespace_id(
...
+        )
+        WITH av
+        WHERE av IS NOT NULL
+        // Trace back to the BuiltinIPPrefix node
+        MATCH (parent_prefix:%(ip_prefix_kind)s)-[r1:HAS_ATTRIBUTE]->(a:Attribute {name: "prefix"})-[r2:HAS_VALUE]->(av)
+        WHERE all(r IN [r1, r2] WHERE (%(branch_filter)s))
+        WITH parent_prefix, av, r1, r2,
+            r1.status = "active" AND r2.status = "active" AS is_active
+        ORDER BY elementId(parent_prefix), r1.branch_level DESC, r1.from DESC, r1.status ASC,
+            r2.branch_level DESC, r2.from DESC, r2.status ASC
```
ajtmccarty: I think the ordering in this query will cause edges that have been deleted on branches to be skipped b/c the query filters out the deleted edges before trying to select the latest path for parent_prefix->:Attribute->:AttributeValue

to get the correct results, you probably have to do a `WITH DISTINCT parent_prefix,av` then a subquery to check if each `parent_prefix` is currently linked to `av` on this branch

to cover it with a test I think you need to update an parent ip prefix's prefix value on a branch and confirm that is correctly included/excluded from the results

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,108 @@ def _get_namespace_id(
...
+    async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG002
+        self._build_possible_parent_prefixes()
+
+        branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
+        self.params.update(branch_params)
+        self.params["ip_prefix_kind"] = InfrahubKind.IPPREFIX
+        self.params["ip_prefix_attribute_kind"] = PREFIX_ATTRIBUTE_LABEL
+
+        query = """
+        // Find AttributeIPNetwork nodes matching possible parent binary addresses
```
ajtmccarty: the comments are easier to find and read if you border them with `// --------------` before and after

backend/infrahub/core/query/ipam.py
gmazoyer: Reworked the query and added a test to watch out for regressions

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,135 @@ def _get_namespace_id(
...
+        }
+        // ------------------
+        // Verify the prefix node itself is active on this branch
+        // ------------------
+        CALL (maybe_parent) {
+            OPTIONAL MATCH (maybe_parent)-[r:IS_PART_OF]->(:Root)
+            WHERE %(branch_filter)s
+            WITH maybe_parent, r.status = "active" AS is_active
+            ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+            RETURN head(collect(is_active)) = TRUE AS node_is_active
```
ajtmccarty: should do the ordering first, then pick the latest one, then check if it is active

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,135 @@ def _get_namespace_id(
...
+        branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
+        self.params.update(branch_params)
+        self.params["ip_prefix_kind"] = InfrahubKind.IPPREFIX
+        self.params["ip_prefix_attribute_kind"] = PREFIX_ATTRIBUTE_LABEL
+
+        query = """
+        // ------------------
+        // Shortlist candidate AttributeIPNetwork nodes using the binary_address index
+        // ------------------
+        CALL () {
```
ajtmccarty: I think you can remove this subquery
I mean, just do the `OPTIONAL MATCH`s outside of a `CALL() {}`

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,135 @@ def _get_namespace_id(
...
+        // ------------------
+        // Resolve the branch-effective attribute value for each candidate
+        // ------------------
+        CALL (maybe_parent) {
+            OPTIONAL MATCH (maybe_parent)-[r1:HAS_ATTRIBUTE]->(:Attribute {name: "prefix"})-[r2:HAS_VALUE]->(av:AttributeValue)
+            WHERE all(r IN [r1, r2] WHERE (%(branch_filter)s))
+            WITH maybe_parent, av, r1, r2, r1.status = "active" AND r2.status = "active" AS is_active
+            ORDER BY elementId(maybe_parent), r1.branch_level DESC, r1.from DESC, r1.status ASC,
+                r2.branch_level DESC, r2.from DESC, r2.status ASC
+            WITH maybe_parent, head(collect([av, is_active])) AS av_is_active
```
ajtmccarty: I think this will work in this scenario because there should always be a `prefix` attribute with a value, but if this were an optional relationship where it could be active or deleted, then you would need to use the safest method of getting the latest path, which is ordering, then picking the latest path, then checking if it is active

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,135 @@ def _get_namespace_id(
...
+        self.return_labels = ["parent_prefix_uuid", "parent_prefix_kind", "prefixlen"]
+
+    def get_data(self) -> list[IPParentPrefixResult]:
+        return [
+            IPParentPrefixResult(
+                prefix_id=result.get_as_type("parent_prefix_uuid", return_type=str),
+                prefix_kind=result.get_as_type("parent_prefix_kind", return_type=str),
+            )
+            for result in self.get_results()
+        ]
```
ajtmccarty, line 234: 👍

backend/tests/component/graphql/queries/test_search.py
```diff
@@ -474,3 +496,152 @@ async def test_search_anywhere_case_sensitive_enabled(
...
+
+
+async def test_search_parent_prefix_branch_effective_value(
+    db: InfrahubDatabase, default_branch: Branch, ip_dataset_01: dict[str, Any]
+) -> None:
+    """A prefix whose value is changed on a branch must not appear as a parent for the old value on that branch."""
+    # ip_dataset_01 has net143 = 10.10.1.0/27 in NS1 on the default branch.
+    # On a new branch, change it to 10.10.2.0/27 (a completely different subnet).
+    # Searching for 10.10.1.1 on that branch should NOT return the changed prefix.
+    branch = await create_branch(db=db, branch_name="search-branch-effective")
```
ajtmccarty, line 625: might be good to add a case for searching on default branch returns `10.10.1.0/27` when expected still

ajtmccarty (approved):

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,140 @@ def _get_namespace_id(
...
+        // ------------------
+        // Verify the prefix node itself is active on this branch
+        // ------------------
+        CALL (maybe_parent) {
+            OPTIONAL MATCH (maybe_parent)-[r:IS_PART_OF]->(:Root)
+            WHERE %(branch_filter)s
+            WITH r
+            ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+            WITH head(collect(r)) AS latest_r
+            RETURN latest_r IS NOT NULL AND latest_r.status = "active" AS node_is_active
```
ajtmccarty: I think a `LIMIT 1` here would perform better than `head(collect(r))`

backend/infrahub/core/query/ipam.py
```diff
@@ -96,6 +102,140 @@ def _get_namespace_id(
...
+        // Resolve the branch-effective attribute value for each candidate
+        // ------------------
+        CALL (maybe_parent) {
+            OPTIONAL MATCH (maybe_parent)-[r1:HAS_ATTRIBUTE]->(:Attribute {name: "prefix"})-[r2:HAS_VALUE]->(av:AttributeValue)
+            WHERE all(r IN [r1, r2] WHERE (%(branch_filter)s))
+            WITH av, r1, r2
+            ORDER BY r1.branch_level DESC, r1.from DESC, r1.status ASC,
+                r2.branch_level DESC, r2.from DESC, r2.status ASC
+            WITH head(collect([av, r1, r2])) AS latest
+            WITH latest[0] AS av, latest[1] AS r1, latest[2] AS r2
```
ajtmccarty: `LIMIT 1` would probably also work better here

ajtmccarty (approved):

PR #8613, opened 2026-03-16
Add delete branch after merge

solababs:
# Why

After a branch is merged, it typically has no further use but remains in the branch list until manually deleted. There is no way to automatically clean it up today.

This PR wires up the `delete_branch_after_merge` config setting (added in Phase 1) to the actual merge paths — both direct `BranchMerge` and proposed change merge — so that the Infrahub branch is automatically deleted after a successful merge when the setting is enabled.

## What changed

- **`backend/infrahub/core/branch/tasks.py`** — after a successful `BranchMerge`, if `delete_branch_after_merge=True` and the branch is not the default branch and there is no associated proposed change, submits the `BRANCH_DELETE` workflow.
- **`backend/infrahub/proposed_change/tasks.py`** — after a proposed change is merged, if `delete_branch_after_merge=True`, checks that no other open proposed changes exist for the source branch before submitting `BRANCH_DELETE`.
- **`backend/tests/functional/branch/test_branch_delete_after_merge.py`** — 4 functional tests covering: config-enabled standard merge triggers delete, config-disabled standard merge does not, proposed change merge triggers delete, and a skipped test for the multi-PC guard (multiple PCs on the same branch are not currently allowed).

The default value of `delete_branch_after_merge` remains `False`, so existing deployments are unaffected.

### Suggested review order
1. [backend/infrahub/core/branch/tasks.py](backend/infrahub/core/branch/tasks.py) — `merge_branch` change
2. [backend/infrahub/proposed_change/tasks.py](backend/infrahub/proposed_change/tasks.py) — `merge_proposed_change` change
3. [backend/tests/functional/branch/test_branch_delete_after_merge.py](backend/tests/functional/branch/test_branch_delete_after_merge.py) — functional tests

## How to review

- In `proposed_change/tasks.py`: confirm the open-PC guard logic is correct — we query for open PCs on the source branch before deciding to delete. Worth checking whether the PC being merged is already in a non-open state at this point and would be excluded from the query.
- The `proposed_change_id is None` guard in `branch/tasks.py` prevents a double-delete when a proposed change merge internally calls `merge_branch`.

## How to test

```bash
uv run pytest backend/tests/functional/branch/test_branch_delete_after_merge.py -v
```

Manual flow:
1. Set `delete_branch_after_merge = true` in `infrahub.toml`, restart service.
2. Create a branch, make a change, merge via `BranchMerge` GraphQL mutation — branch should no longer appear in the branch list.
3. Repeat by merging a proposed change — same result.
4. Confirm with `delete_branch_after_merge = false` (default) — branch is retained after merge.

## Impact & rollout

- **Backward compatibility:** No breaking changes. `delete_branch_after_merge` defaults to `false`; existing behaviour is unchanged.
- **Config/env changes:** Behaviour is controlled by `delete_branch_after_merge` in `MainSettings` (introduced in Phase 1).
- **Deployment notes:** Safe to deploy. Feature is opt-in via config.

## Checklist

- [x] Tests added/updated


backend/infrahub/core/branch/tasks.py
```diff
@@ -371,6 +372,13 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
             parameters={"branch_name": obj.name},
         )
 
+        if config.SETTINGS.main.delete_branch_after_merge and not obj.is_default and proposed_change_id is None:
```
ajtmccarty: why does the `proposed_change_id` need to be `None` here?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -245,6 +246,22 @@ async def merge_proposed_change(
...
+                filters={
+                    "source_branch__value": source_branch.name,
+                    "state__value": ProposedChangeState.OPEN.value,
+                },
+            )
+            if not open_pcs:
+                await get_workflow().submit_workflow(
+                    workflow=BRANCH_DELETE,
+                    context=context,
+                    parameters={"branch": source_branch.name},
```
ajtmccarty: I'm guessing that the logic here is to prevent breaking open proposed changes, but any open PCs will be unmergeable once this PC finishes merging
maybe the right thing to do in this case is to set any open PCs on this branch to closed and then allow deleting it

backend/infrahub/proposed_change/tasks.py
```diff
@@ -245,6 +247,18 @@ async def merge_proposed_change(
             user_id=context.account.account_id,
         )
 
+        if config.SETTINGS.main.delete_branch_after_merge and not source_branch.is_default:
+            await get_workflow().submit_workflow(
+                workflow=BRANCH_CANCEL_PROPOSED_CHANGES,
+                context=context,
+                parameters={"branch_name": source_branch.name},
+            )
```
ajtmccarty: upon further review, this workflow is already run inside of the `merge_branch` flow called above on line 231, so I think you can assume that the open proposed changes are or will be canceled
I guess there is the possibility that the BRANCH_DELETE workflow _could_ run before the BRANCH_CANCEL_PROPOSED_CHANGES, but it looks like that should not be a problem b/c the `cancel_proposed_changes_branch` function doesn't appear to count on the branch actually existing on the database
it would be good to test and confirm this though

backend/tests/functional/branch/test_branch_delete_after_merge.py
```diff
@@ -0,0 +1,145 @@
...
+            query = Mutation(
+                mutation="BranchMerge",
+                input_data={"data": {"name": branch.name}},
+                query={"ok": None, "task": {"id": None}, "object": {"id": None}},
+            )
+            await client.execute_graphql(query=query.render())
+
+            mock_submit.assert_any_call(
+                workflow=BRANCH_DELETE,
+                context=ANY,
```
ajtmccarty, line 59: I did not know that you could do this. that is cool

backend/tests/functional/branch/test_branch_delete_after_merge.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+            await client.execute_graphql(query=update_query.render())
+
+            mock_submit.assert_any_call(
+                workflow=BRANCH_DELETE,
+                context=ANY,
+                parameters={"branch": branch.name},
+            )
+
+    @pytest.mark.skip("Multiple proposed changes are not allowed for the same branch")
```
ajtmccarty: I forgot we made this change
is it worth keeping the test if it is skipped?

ajtmccarty (approved):

backend/infrahub/proposed_change/tasks.py
ogenstad: Yeah, this shouldn't be needed. Once the branch is merged we'd cancel the proposed change (if it would still be in the open state). So we don't need an extra action for to cancel it here. Also the state would already have changed to "MERGED" so we don't need to cancel it.

ogenstad (approved):

PR #8616, opened 2026-03-16
Add delete branch from git

solababs:
# Why

When `delete_branch_after_merge` is enabled and a branch has `sync_with_git=True`, the Infrahub branch is cleaned up automatically but the corresponding branch in the Git repositories is left behind. There is no way to have both sides cleaned up automatically today.

This PR adds the `GIT_REPOSITORIES_DELETE_BRANCH` workflow and wires `delete_git_branch_after_merge` into the `delete_branch()` flow so that all synced Git repositories have the branch removed after the Infrahub branch is deleted. Per-repo failures are logged but never block the deletion.

## What changed

- **`backend/infrahub/git/base.py`** — two new methods on `InfrahubRepositoryBase`: `origin_has_branch(branch_name)` (checks whether a branch exists on the remote) and `delete_remote_branch(branch_name)` (pushes `--delete` to origin and removes the local tracking ref).
- **`backend/infrahub/git/tasks.py`** — new `delete_git_branch()` flow and `git_branch_delete` task. The flow fans out across all `CoreRepository` instances via a batch; the task skips repos that don't have the branch and logs (without re-raising) any deletion failures.
- **`backend/infrahub/workflows/catalogue.py`** — new `GIT_REPOSITORIES_DELETE_BRANCH` workflow definition registered in the `WORKFLOWS` list.
- **`backend/infrahub/core/branch/tasks.py`** — after `BranchDeletedEvent` is sent in `delete_branch()`, submits `GIT_REPOSITORIES_DELETE_BRANCH` when `delete_git_branch_after_merge=True` and the branch has `sync_with_git=True`.
- **`backend/tests/component/git/test_delete_git_branch.py`** — 4 component tests for `origin_has_branch` and `delete_remote_branch` against a real local Git repo.
- **`backend/tests/functional/branch/test_delete_git_branch.py`** — 3 functional tests: config enabled + `sync_with_git=True` triggers workflow; config disabled does not; `sync_with_git=False` does not.

No schema, API contract, or database changes. Default for `delete_git_branch_after_merge` remains `False`.

### Suggested review order
1. [backend/infrahub/git/base.py](backend/infrahub/git/base.py) — `origin_has_branch` + `delete_remote_branch` 
2. [backend/infrahub/git/tasks.py](backend/infrahub/git/tasks.py) — `delete_git_branch` flow + `git_branch_delete` task
3. [backend/infrahub/core/branch/tasks.py](backend/infrahub/core/branch/tasks.py) — trigger in `delete_branch()` 
4. [backend/infrahub/workflows/catalogue.py](backend/infrahub/workflows/catalogue.py) — new definition + WORKFLOWS registration
5. Tests

## How to review

- In `git_branch_delete` task: confirm that swallowing exceptions with `log.exception` is the right behaviour (FR-013 says per-repo failures must not block deletion). The `async with lock.registry.get(...)` mirrors `git_branch_create` and prevents concurrent git ops on the same repo.
- `origin_has_branch` is synchronous — do not add `await` when calling it in the task.
- The `should_delete_git` guard in `delete_branch()` checks both the config flag and `obj.sync_with_git` — branches not tracked in Git are silently skipped.

## How to test

```bash
uv run pytest backend/tests/component/git/test_delete_git_branch.py -v
uv run pytest backend/tests/functional/branch/test_delete_git_branch.py -v
```

Manual flow:
1. Set `delete_branch_after_merge = true` and `delete_git_branch_after_merge = true` in `infrahub.toml`, restart.
2. Create a branch with `sync_with_git=True`, push a commit, merge via `BranchMerge`.
3. Verify the Infrahub branch is gone and the branch no longer exists in the linked Git repository.
4. Confirm with `delete_git_branch_after_merge = false` — Git branch is retained.

## Impact & rollout

- **Backward compatibility:** No breaking changes. `delete_git_branch_after_merge` defaults to `false`; requires `delete_branch_after_merge=true` to be set (validated at startup).
- **Config/env changes:** Behaviour gated on `delete_git_branch_after_merge` in `MainSettings` (Phase 1).
- **Deployment notes:** Safe to deploy. Feature is fully opt-in. Git deletion failures are non-fatal.

## Checklist

- [x] Tests added/updated


backend/infrahub/git/base.py
```diff
@@ -536,6 +536,20 @@ def get_branches_from_local(self, include_worktree: bool = True) -> dict[str, Br
...
+    def origin_has_branch(self, branch_name: str) -> bool:
+        """Return True if branch_name exists as a remote branch on origin."""
+        return branch_name in self.get_branches_from_remote()
+
+    async def delete_remote_branch(self, branch_name: str) -> None:
+        """Delete branch_name from origin and remove the local tracking ref."""
+        if not self.has_origin:
+            return
+        repo = self.get_git_repo_main()
+        repo.git.push("origin", "--delete", branch_name)
```
ogenstad, line 548: Here we'd want to catch any errors and write a task log to the repository object if the operation fails perhaps because the token has expired, or there's some network error or the user we use to connect to the git server doesn't have the permissions to delete a branch.

gmazoyer, line 548: This reminds me how much I had gitpython

backend/infrahub/git/tasks.py
```diff
@@ -312,6 +331,31 @@ async def git_branch_create(
...
+    repository_id: str,
+    repository_name: str,
+    repository_location: str,
+) -> None:
+    log = get_run_logger()
+    repo = await InfrahubRepository.init(
+        id=repository_id, name=repository_name, location=repository_location, client=client
+    )
+    if not repo.origin_has_branch(branch):
+        return
```
gmazoyer: Should this be inside the lock too, to prevent two possible concurrent tasks to pass the conditional before one of them acquire the lock?

backend/tests/functional/branch/test_delete_git_branch.py
```diff
@@ -0,0 +1,93 @@
...
+        client: InfrahubClient,
+        bus_simulator: BusSimulator,
+        prefect_test_fixture: None,
+    ) -> None:
+        await load_schema(db, schema=CAR_SCHEMA)
+
+    async def test_git_deletion_triggered_when_config_enabled_and_sync_with_git(
+        self, initial_dataset: None, client: InfrahubClient
+    ) -> None:
+        config.SETTINGS.main.delete_git_branch_after_merge = True
```
gmazoyer: Each test sets `config.SETTINGS.main.delete_git_branch_after_merge` directly. If a test fails partway through, the global config might remain mutated for subsequent tests. We usually use a fixture to handle this situations:
```python
def delete_git_branch_after_merge_reset_config():
    original = config.SETTINGS.main.delete_git_branch_after_merge
    yield
    config.SETTINGS.main.delete_git_branch_after_merge = original
```

backend/infrahub/workflows/catalogue.py
```diff
@@ -192,6 +192,13 @@
     tags=[WorkflowTag.DATABASE_CHANGE],
 )
 
+GIT_REPOSITORIES_DELETE_BRANCH = WorkflowDefinition(
+    name="git-repositories-delete-branch",
+    type=WorkflowType.CORE,
+    module="infrahub.git.tasks",
+    function="delete_git_branch",
+)
```
gmazoyer, line 201: I wonder if we should add a tag like the workflow above this one.

backend/infrahub/git/base.py
ogenstad: > This reminds me how much I had gitpython

@gmazoyer, typo for "hate"? ;)

backend/tests/integration/git/test_delete_git_branch_gogs.py
```diff
@@ -0,0 +1,131 @@
...
+        gogs_server: dict,
+        delete_branch_after_merge_reset_config: None,
+        delete_git_branch_after_merge_reset_config: None,
+    ) -> None:
+        base_url = gogs_server["base_url"]
+        token = gogs_server["token"]
+        branch_name = "feature-beta"
+
+        config.SETTINGS.main.delete_branch_after_merge = True
+        config.SETTINGS.git.delete_git_branch_after_merge = True
```
ogenstad, line 118: It would be good if we could move these two config settings to a fixture that reverted the behaviour the test.

ogenstad (approved):

backend/tests/integration/git/test_delete_git_branch_gogs.py
solababs: Yes the reset config fixtures for both of them are added in the function signature

PR #8624, opened 2026-03-17
lock when merging

### LINKED ISSUE #6974: bug: Artifact Diff Showing spacing in diff even though there is no actual diff
h3\. Component

[| #component]

Frontend UI

h3\. Infrahub version

[| #infrahub-version]

1.3.x

h3\. Current Behavior

[| #current-behavior]

When viewing the artifact diffs within a Proposed Change, random spaces between words show up within the diff. The changes within the PC is only data and no template changes.

h3\. Expected Behavior

[| #expected-behavior]

Proper artifact diff without any extraneous spacing

h3\. Steps to Reproduce

[| #steps-to-reproduce]

We cannot reliably reproduce the issue.

h3\. Additional Information

[| #additional-information]

We pulled down the artifacts from the API using the storage IDs. When diffing both artifacts, there is no diff in VS Code.

ajtmccarty:
# Why

Running multiple merge operation at one time, either for the same branch or different branches, can cause unexpected results.

We should only allow one merge operation to proceed at a time. Before executing, the merge operation confirms that the branch being merged is still open. This prevents any branch from being merged multiple times.

Closes #6974 / IFC-1657

## What changed

Adds a new very simple `MergeLocker` class and uses it inside of the `merge_branch` flow. There's also some refactoring of the `merge_branch` flow to move much of the logic to a `_do_merge_branch` function to reduce indentation levels, but there is no functional difference.

# Testing

Added some component tests with more mocking than we usually do. I could make it a functional test, but it would be very expensive (diffing and merging are our two slowest operation) and kind of hard to verify.

Also did manual testing for trying to merge different branches and the same branch simultaneously.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a global synchronization guard so only one merge runs at a time; merges now short-circuit when a branch is not open and reliably finalize merged state and post-merge processing.
* **Tests**
  * Added tests validating concurrent merge behavior: same-branch deduplication and correct sequential handling for different branches.

### DIFF
--- a/backend/infrahub/core/merge/__init__.py
+++ b/backend/infrahub/core/merge/__init__.py
@@ -1,0 +1,3 @@
+from infrahub.core.merge.branch_merger import BranchMerger
+
+__all__ = ["BranchMerger"]
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any, Sequence
 from uuid import uuid4
 
 import pydantic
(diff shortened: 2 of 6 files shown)


backend/infrahub/core/merge/__init__.py
```diff
@@ -0,0 +1,3 @@
+from infrahub.core.merge.branch_merger import BranchMerger
+
+__all__ = ["BranchMerger"]
```
gmazoyer, line 3: Should this `__init__py` be empty or did you add it to avoid changing the existing imports?

ajtmccarty: the second one. I added it for backwards compatibility with existing imports

gmazoyer (approved):

PR #8625, opened 2026-03-17
feat: Add graph path traversal and dependency discovery

minitriga:
# Why

Operators investigating infrastructure today have no way to ask the graph "how are these two things connected?" or "what depends on this node?". Answering either question requires clicking through object detail pages one relationship at a time, which doesn't scale for impact analysis or troubleshooting.

**Goal:** Add a graph path traversal feature that lets users (1) discover the paths connecting any two nodes and (2) discover all nodes of given kinds reachable from a source node, exposed through GraphQL and a visual topology explorer in the UI.

**Non-goals:** No changes to the core graph schema or write paths. Traversal is strictly read-only. This does not replace the existing object-detail relationship views.

Closes INFP-1991

## What changed

**Behavioral changes (user/system observable):**

- New GraphQL query `InfrahubPathTraversal(data: PathTraversalInput!)` returns all paths that meet the filtering criteria between a `source_id` and `destination_id`, as ordered hops (node + relationship traversed). Supports `kind_filter`, `relationship_filter`, `excluded_kinds`, `excluded_namespaces`, `max_depth` (default 5, max 20), and `max_paths` (default 10, max 5000).
- New GraphQL query `InfrahubReachableNodes(data: ReachableNodesInput!)` returns all nodes of the requested `target_kinds` reachable from a `source_id`, each with its full path and depth. Supports `max_depth`, `max_results` (default 50, max 200), and `max_paths` (default 500, max 5000).
- New `/path-traversal` page with two modes — Path mode (two-node connection discovery) and Dependencies mode (single-source reachability) — rendered as an interactive React Flow graph with a results sidebar, kind/namespace filters, and keyboard navigation between discovered paths.
- New "Path Traversal" item under the Internal menu section, and an action on the object-details menu to launch traversal pre-seeded from a node.
- Default namespace exclusions (Core, Internal, Builtin, Lineage, Profile, Template) are applied automatically and cannot be opted out of; user-supplied `excluded_namespaces` are unioned with them.

**Implementation notes:**

- Traversal is branch- and time-aware, reusing existing branch/temporal edge filtering so only relationships active on the queried branch are followed.
- A schema-based planner (`backend/infrahub/graph_traversal/planning/`) computes the legal `(start_kind, rel_name, end_kind)` adjacency triples per hop *before* hitting the database, pruned by user filters and by the requester's view permissions. The generated Cypher then only traverses edges the user is allowed to see.
- The Cypher path renderer was rewritten for performance (replacing an earlier projection approach) and projects hops as lists of maps, consumed via a new `QueryResult.get_as_list_of_type` helper.
- Permissions are resolved per-kind through a `KindPermissionCache` so a single traversal doesn't re-resolve the same kind repeatedly.

**What stayed the same:**

- No core graph schema changes — traversal runs over the existing `Node → IS_RELATED → Relationship → IS_RELATED → Node` substrate.
- Read-only: no mutations, no migrations.

### Path traversal — paths between two nodes

```graphql
query GetPathTraversal {
  InfrahubPathTraversal(
    data: {
      source_id: "17d8e4f2-0000-0000-0000-000000000001"
      destination_id: "17d8e4f2-0000-0000-0000-000000000002"
      max_depth: 5
      max_paths: 10
      kind_filter: ["InfraDevice", "InfraInterface"]
      excluded_kinds: ["LocationContinent"]
    }
  ) {
    count
    source { id kind display_label }
    destination { id kind display_label }
    paths {
      depth
      hops {
        node { id kind label display_label hfid }
        relationship { from_rel from_label to_rel to_label kind }
      }
    }
  }
}
```

```json
{
  "data": {
    "InfrahubPathTraversal": {
      "count": 1,
      "source": { "id": "17d8...0001", "kind": "InfraDevice", "display_label": "atl1-edge1" },
      "destination": { "id": "17d8...0002", "kind": "InfraDevice", "display_label": "atl1-edge2" },
      "paths": [
        {
          "depth": 2,
          "hops": [
            {
              "node": { "id": "17d8...0001", "kind": "InfraDevice", "label": "Device", "display_label": "atl1-edge1", "hfid": ["atl1-edge1"] },
              "relationship": null
            },
            {
              "node": { "id": "17d8...00aa", "kind": "InfraInterface", "label": "Interface", "display_label": "atl1-edge1 Ethernet1", "hfid": ["atl1-edge1", "Ethernet1"] },
              "relationship": { "from_rel": "interfaces", "from_label": "Interfaces", "to_rel": "device", "to_label": "Device", "kind": "Component" }
            },
            {
              "node": { "id": "17d8...0002", "kind": "InfraDevice", "label": "Device", "display_label": "atl1-edge2", "hfid": ["atl1-edge2"] },
              "relationship": { "from_rel": "device", "from_label": "Device", "to_rel": "interfaces", "to_label": "Interfaces", "kind": "Component" }
            }
          ]
        }
      ]
    }
  }
}
```

The first hop is always the source node, so its `relationship` is `null`. When no path satisfies the filters, `count` is `0` and `paths` is `[]` (`source`/`destination` are still returned).

### Reachable nodes — dependency discovery from one source

```graphql
query GetReachableNodes {
  InfrahubReachableNodes(
    data: {
      source_id: "17d8e4f2-0000-0000-0000-000000000001"
      target_kinds: ["InfraInterface", "IpamIPAddress"]
      max_depth: 5
      max_results: 50
      max_paths: 500
    }
  ) {
    count
    source { id kind display_label }
    dependencies {
      depth
      node { id kind display_label hfid }
      path {
        depth
        hops {
          node { id kind display_label }
          relationship { from_rel to_rel kind }
        }
      }
    }
  }
}
```

```json
{
  "data": {
    "InfrahubReachableNodes": {
      "count": 2,
      "source": { "id": "17d8...0001", "kind": "InfraDevice", "display_label": "atl1-edge1" },
      "dependencies": [
        {
          "depth": 1,
          "node": { "id": "17d8...00aa", "kind": "InfraInterface", "display_label": "atl1-edge1 Ethernet1", "hfid": ["atl1-edge1", "Ethernet1"] },
          "path": {
            "depth": 1,
            "hops": [
              { "node": { "id": "17d8...0001", "kind": "InfraDevice", "display_label": "atl1-edge1" }, "relationship": null },
              { "node": { "id": "17d8...00aa", "kind": "InfraInterface", "display_label": "atl1-edge1 Ethernet1" }, "relationship": { "from_rel": "interfaces", "to_rel": "device", "kind": "Component" } }
            ]
          }
        },
        {
          "depth": 2,
          "node": { "id": "17d8...00bb", "kind": "IpamIPAddress", "display_label": "10.0.0.1/31", "hfid": ["10.0.0.1/31"] },
          "path": {
            "depth": 2,
            "hops": [
              { "node": { "id": "17d8...0001", "kind": "InfraDevice", "display_label": "atl1-edge1" }, "relationship": null },
              { "node": { "id": "17d8...00aa", "kind": "InfraInterface", "display_label": "atl1-edge1 Ethernet1" }, "relationship": { "from_rel": "interfaces", "to_rel": "device", "kind": "Component" } },
              { "node": { "id": "17d8...00bb", "kind": "IpamIPAddress", "display_label": "10.0.0.1/31" }, "relationship": { "from_rel": "ip_addresses", "to_rel": "interface", "kind": "Component" } }
            ]
          }
        }
      ]
    }
  }
}
```

`dependencies` holds one entry per reachable (node, path) pair, ordered shortest-first by `depth`. A missing `source_id` returns an error; a source with no reachable nodes of the target kinds returns `count: 0` with an empty `dependencies` list.

## Impact & rollout

- **Backward compatibility:** Additive only — two new GraphQL queries, a new menu item, a new page. No existing query, type, or route changes shape. No migration.
- **Performance:** Depth and path-count are bounded by enforced maxima. Schema-based planning keeps the generated Cypher narrow by pre-pruning illegal/forbidden edges. Querying for path traversal should always perform pretty well. Querying for dependencies can be slow depending on the size of the graph, how many possible dependency terminal objects exist, and the depth of the path. More filtering will improve performance.
- **Config/env changes:** None. No new feature flag.
- **Deployment notes:** Safe to deploy — no coordinated release required.

## Checklist

- [x] Tests added/updated


backend/infrahub/core/query/dependencies.py
```diff
@@ -0,0 +1,158 @@
...
+                "AND all(n IN nodes(path) WHERE "
+                "n.uuid = $source_uuid "
+                "OR NOT n:Node "
+                "OR n.kind IN $target_kinds "
+                "OR NOT n.namespace IN $excluded_namespaces) "
+            )
+
+        # Find all reachable nodes of the target kinds, returning the shortest
+        # path to each so the frontend can render the graph.
+        query = f"""
```
ajtmccarty, line 76: we generally don't use f-strings for building cypher queries b/c then you need to use double `{{` and `}}`, which makes it harder to copy a query string from a code editor into the neo4j browser or CLI

backend/infrahub/core/query/dependencies.py
```diff
@@ -0,0 +1,158 @@
...
+        MATCH (source:Node {{ uuid: $source_uuid }})
+        MATCH path = (source)-[:IS_RELATED*2..{max_edge_length}]-(target:Node)
+        WHERE target.kind IN $target_kinds
+        AND target.uuid <> $source_uuid
+        AND all(r IN relationships(path) WHERE ({branch_filter}))
+        {namespace_filter}
+        WITH DISTINCT target, path, length(path) / 2 AS depth
```
ajtmccarty, line 83: I think this would exclude multiple paths of the same length to the same target via different relationships

backend/infrahub/core/query/dependencies.py
```diff
@@ -0,0 +1,158 @@
...
+        WITH target, collect(path)[0] AS shortest_path, min(depth) AS min_depth
+        RETURN target, shortest_path, min_depth
+        ORDER BY min_depth ASC
+        LIMIT {self.max_results}
+        """
+
+        self.add_to_query(query)
+        self.return_labels = ["target", "shortest_path", "min_depth"]
+
+    @staticmethod
```
ajtmccarty, line 94: should just be a regular instance method b/c it is only going to be used on an instance of this query even if it doesn't use the `self` variable
actually, looks like this could be extracted from both query classes and deduplicated in its own function

backend/infrahub/core/query/path.py
```diff
@@ -0,0 +1,253 @@
...
+    kind: str
+    display_label: str
+    db_id: str
+
+
+@dataclass(frozen=True)
+class PathRelationshipData:
+    uuid: str
+    name: str
+    direction: str
```
ajtmccarty, line 24: we have a `RelationshipDirection` enum that can be used here

backend/infrahub/core/query/dependencies.py
```diff
@@ -0,0 +1,158 @@
...
+        path_relationships: list[PathRelationshipData] = []
+
+        for i, node in enumerate(raw_nodes):
+            if i % 2 == 0:
+                path_nodes.append(
+                    PathNodeData(
+                        uuid=node.get("uuid", ""),
+                        kind=node.get("kind", ""),
+                        display_label=node.get("display_label", node.get("kind", "")),
+                        db_id=str(node.element_id) if hasattr(node, "element_id") else "",
```
ajtmccarty, line 110: doesn't look like you actually need this information

backend/infrahub/graphql/queries/path.py
```diff
@@ -0,0 +1,331 @@
...
+        db=graphql_context.db,
+        branch=graphql_context.branch,
+        at=graphql_context.at,
+        id=source_id,
+    )
+    if not source_node:
+        raise GraphQLError(f"Source node not found: {source_id}")
+
+    # Import here to avoid circular
+    from infrahub.core.query.dependencies import DependencyQuery
```
ajtmccarty, line 252: this is bad. we have enough crutches in place to keep the circular dependency errors away. we should not add more unless absolutely necessary

backend/infrahub/graphql/queries/path.py
```diff
@@ -0,0 +1,331 @@
...
+InfrahubPathTraversal = Field(
+    PathTraversalResultType,
+    data=PathTraversalInput(required=True),
+    description="Find all shortest paths between two nodes in the graph",
+    resolver=path_traversal_resolver,
+    required=True,
+)
+
+
+# --- Dependency Discovery types ---
```
ajtmccarty, line 201: I'd say that these two resolvers could be split into two files unless they really depend on one another in some way

backend/infrahub/graphql/queries/path.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+class DependencyInput(InputObjectType):
+    source_id = String(required=True, description="UUID of the source node")
+    target_kinds = List(of_type=NonNull(String), required=True, description="Node kinds to search for")
+    max_depth = Int(required=False, default_value=5, description="Maximum traversal depth (default: 5, max: 20)")
+    max_results = Int(required=False, default_value=50, description="Maximum results (default: 50, max: 200)")
+
+
+async def dependency_resolver(
```
ajtmccarty, line 229: I don't think "dependency" is the right word for this. to me, dependency implies a requirement and this will traverse optional relationships and even relationships that are only defined on the peer schema and not the source schema.
I think the traversal query is more of a point-to-point and the dependency query is more of all-paths or all-peers
...but I am not the word-picker

backend/infrahub/core/query/path.py
```diff
@@ -0,0 +1,253 @@
...
+    insert_return = False
+    insert_limit = False
+
+    def __init__(
+        self,
+        source_id: str,
+        destination_id: str,
+        max_depth: int = 5,
+        max_paths: int = 10,
+        node_filter: list[str] | None = None,
```
ajtmccarty, line 58: `kind_filter` feels more accurate

backend/infrahub/core/query/path.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+
+        # Node kind filter: allow source/target to be any kind, filter intermediates
+        if self.node_filter:
+            self.params["node_filter"] = self.node_filter
+            where_clauses.append(
+                "all(n IN nodes(path) WHERE "
+                "n.uuid IN [$source_uuid, $target_uuid] "
+                "OR NOT n:Node "  # Allow Relationship vertices through
+                "OR n.kind IN $node_filter)"
```
ajtmccarty, line 128: if you want to correctly support generics, you need to the filter to be if any label on the node matches any label in the `kind_filter`. otherwise, passing a generic kind in will never work because no nodes have their `kind` property set to a generic schema

backend/infrahub/core/query/path.py
```diff
@@ -0,0 +1,253 @@
...
+                "all(n IN nodes(path) WHERE "
+                "NOT n:Relationship "  # Skip Node vertices for this check
+                "OR n.name IN $relationship_filter)"
+            )
+
+        where_str = " AND ".join(where_clauses)
+
+        if self.branch.is_default:
+            # On the default branch, there are no branch-deleted edges to worry
+            # about. We can filter directly for active status, which is much faster.
```
ajtmccarty, line 144: you still have to account for deleted edges on the default branch. there are some situations where a status="deleted" edge is added and the status="active" edge remains with `to=NULL`
you can still do an easier filter where you only need to verify that no `:IS_RELATED {branch=$default_branch, status="deleted"}]` edge exists for the source to target
although to be completely correct it also has to account for relationship direction

backend/infrahub/core/query/dependencies.py
```diff
@@ -0,0 +1,158 @@
...
+        WITH DISTINCT target, path, length(path) / 2 AS depth
+        ORDER BY depth ASC
+        WITH target, collect(path)[0] AS shortest_path, min(depth) AS min_depth
+        RETURN target, shortest_path, min_depth
+        ORDER BY min_depth ASC
+        LIMIT {self.max_results}
+        """
+
+        self.add_to_query(query)
+        self.return_labels = ["target", "shortest_path", "min_depth"]
```
ajtmccarty, line 92: it would be nice if the query would return the specific data that you need and you could avoid returning Node, Relationship, or Path objects
in this case it would probably have to be a list of lists, so not sure how difficult that would be.

backend/infrahub/core/query/path.py
```diff
@@ -0,0 +1,253 @@
...
+            # Pattern from attribute.py:
+            #   CALL (a, np) {
+            #       MATCH (a)-[r:REL]->(np) WHERE branch_filter
+            #       ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
+            #       LIMIT 1
+            #       RETURN r
+            #   }
+            #   WHERE r.status = "active"
+            #
+            # First find a limited set of candidate paths, then validate edges:
```
ajtmccarty, line 169: too much comment

backend/infrahub/core/query/dependencies.py
minitriga: Fixed in ab907e240. Queries now use `%`-formatted strings so the Cypher can be copied into the Neo4j browser without escaping `{{`/`}}`. Applied to both `path.py` and the renamed `reachable.py`.

backend/infrahub/core/query/dependencies.py
minitriga: Fixed in ab907e240. Removed the `collect(path)[0]` collapse — the query now returns up to `max_results` distinct `(target, path)` pairs, ordered by depth, so alternate same-length routes through different relationships are preserved.

backend/infrahub/core/query/dependencies.py
minitriga: Fixed in ab907e240. Extracted `extract_path_data` as a module-level function in `backend/infrahub/core/query/path.py`. Both `PathTraversalQuery.get_paths` and `ReachableNodesQuery.get_reachable_nodes` now call it — no more duplicated `_extract_path` staticmethod.

backend/infrahub/core/query/path.py
minitriga: Fixed in ab907e240. `PathRelationshipData.direction` is now `RelationshipDirection` (from `infrahub.core.constants`), and `extract_path_data` emits enum values. The GraphQL layer calls `.value` when flattening for the response.

backend/infrahub/core/query/dependencies.py
minitriga: Fixed in ab907e240. Dropped the `db_id` field from `PathNodeData` entirely — the GraphQL layer never consumed it, so `extract_path_data` no longer reaches for `element_id`.

backend/infrahub/graphql/queries/path.py
minitriga: Fixed in ab907e240. Split into two resolver files, so the reachable-nodes resolver imports `ReachableNodesQuery` from its own module at top level. The local-import crutch inside the function is gone.

backend/infrahub/graphql/queries/path.py
minitriga: Fixed in ab907e240. `backend/infrahub/graphql/queries/path.py` now contains only `PathTraversal`; the reachable-nodes resolver moved to `reachable.py`. Shared pieces (`PathNodeType`, `PathResultType`, `_path_data_to_result`) stay in `path.py` and are imported.

backend/infrahub/graphql/queries/path.py
minitriga: Fixed in ab907e240. Renamed to `ReachableNodes` across the stack: `DependencyQuery` → `ReachableNodesQuery` (`backend/infrahub/core/query/reachable.py`), `InfrahubDependencies` → `InfrahubReachableNodes`, GraphQL `Dependency*` types → `ReachableNodes*`, `dependency_nodes` field → `reachable_nodes`. Frontend domain/hook files renamed to match.

backend/infrahub/core/query/path.py
minitriga: Fixed in ab907e240. Renamed `node_filter` → `kind_filter` on `PathTraversalQuery`, `PathTraversalInput`, the frontend params, and the test suite.

backend/infrahub/core/query/path.py
minitriga: Fixed in ab907e240. The kind filter now matches any label on the node: `any(l IN labels(n) WHERE l IN $kind_filter)`. Same pattern applied to the reachable-nodes target filter (`any(l IN labels(target) WHERE l IN $target_kinds)`) so generic kinds work there too.

backend/infrahub/core/query/path.py
minitriga: Fixed in ab907e240. Default-branch query now rejects any path where a `status="deleted"` twin edge exists between the same pair of vertices:

```cypher
AND none(
    r IN relationships(path) WHERE exists(
        (startNode(r))-[:IS_RELATED {branch: $default_branch, status: "deleted"}]-(endNode(r))
    )
)
```

The edge-direction nuance you mentioned: the deleted-edge probe is undirected (`-[...]-`) so it catches twin deleted edges regardless of their direction relative to the active one.

backend/infrahub/core/query/dependencies.py
minitriga: Partially addressed in ab907e240. The reachable-nodes query now projects explicit scalar fields instead of returning a raw `Node`:

```cypher
RETURN target.uuid AS target_uuid,
       target.kind AS target_kind,
       coalesce(target.display_label, target.kind) AS target_display_label,
       path,
       depth,
       coalesce(nodes(path)[-2].name, "") AS relationship_name
```

Python now uses `result.get_as_str` / `get_as_optional_type` instead of `.get()` on a Node. The `path` is still a Neo4j `Path` object — fully projecting it would require turning the alternating Node/Relationship vertices into a list of maps in Cypher, which I opted to leave for a follow-up since (a) direction has to be computed from edge-vertex comparisons that are awkward in Cypher list comprehensions, and (b) you flagged it as uncertain difficulty. Happy to tackle that separately if you'd like.

backend/infrahub/core/query/path.py
minitriga: Fixed in ab907e240. Stripped the long narrative comments; kept only a short inline note where it explains the reason for the pattern (e.g., why the latest-edge CALL subquery is needed off the default branch).

dev/knowledge/frontend/shared-components.md
```diff
@@ -131,7 +131,7 @@ The `tab` argument on each helper is a string-literal union (e.g. `BranchDetails
 | Debounced value | `useDebounce` | `shared/hooks/useDebounce.ts` |
 | Pagination state | `usePagination` | `shared/hooks/usePagination.ts` |
 | Local-storage state | `useLocalStorage` | `shared/hooks/useLocalStorage.ts` |
-| Copy to clipboard | `useCopyToClipboard` | `shared/hooks/useCopyToClipboard.ts` |
+| Copy to clipboard (incl. "Copied!" feedback) | `useCopyToClipboard` — never call `navigator.clipboard.*` directly | `shared/hooks/useCopyToClipboard.ts` |
```
bilalabbad: `(incl. "Copied!" feedback)` is wrong no? useCopyToClipboard does not include it.

frontend/app/tests/e2e/path-traversal.spec.ts
```diff
@@ -0,0 +1,143 @@
+import { expect, test } from "@playwright/test";
+
+import { ACCOUNT_STATE_PATH } from "../constants";
+
+test.describe
+  .skip("/path-traversal", () => {
```
bilalabbad, line 6: skip is expected?
if yes: let's delete it
if no: let's remove it and fix

frontend/app/src/entities/nodes/object/ui/object-table/toolbar/actions/groups/toolbar-add-to-groups-action.test.tsx
```diff
@@ -59,7 +59,7 @@ describe("ToolbarAddToGroupsAction Component", () => {
 
     // THEN
     await expect.element(component.getByRole("dialog", { name: "Add to groups" })).toBeVisible();
-    await expect.element(component.getByPlaceholder("Filter...")).toBeVisible();
+    await expect.element(component.getByPlaceholder("Search by name or UUID...")).toBeVisible();
```
bilalabbad: it is not really true. Currently, it search trought all node values or exact UUID. Why not keep "filter..." or just "Search"

frontend/app/src/shared/components/inputs/kind-multi-select.tsx
```diff
@@ -0,0 +1,129 @@
...
+  value: string[];
+  onChange: (kinds: string[]) => void;
+  label?: string;
+  placeholder?: string;
+  filter?: (namespace: string) => boolean;
+  className?: string;
+  id?: string;
+}
+
+export function KindMultiSelect({
```
bilalabbad, line 31: Image
it's does not looks right when an option is selected.

frontend/app/src/pages/path-traversal/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { PathTraversalPage } from "@/entities/path-traversal/ui/path-traversal-page";
+
+export function Component() {
+  return (
+    <div className="flex-1 overflow-hidden rounded-lg border border-gray-200 bg-white">
```
bilalabbad: why not use <Content.Card /> ?

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+    activeClass: "bg-amber-100 text-amber-700",
+  },
+};
+
+const MODE_LABELS: Record<Mode, string> = {
+  path: "Path",
+  dependencies: "Dependencies",
+};
+
+export function PathTraversalPage() {
```
bilalabbad, line 35: you can probably inline this file directly in `pages/path-traversal/index.tsx`. it avoid unnecessary redirection

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [mode, setMode] = useQueryState(
+    "mode",
+    parseAsStringEnum<Mode>(MODES as unknown as Mode[]).withDefault("path")
```
bilalabbad: `MODES` is not an enum. you probably want to use `parseAsStringLiteral(MODES)`

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+        ) : (
+          <DependenciesModeMain
+            parametersOpen={parametersOpen}
+            onParametersClick={toggleParameters}
+          />
+        )}
+      </main>
+
+      {parametersOpen && (
+        <aside className="absolute top-4 right-4 bottom-4 z-10 flex w-80 flex-col overflow-hidden rounded-lg border border-gray-200 bg-white shadow-xl">
```
bilalabbad: just curious, why did we put it on right? I'm asking because I feel like the left side is more natural and follow readin order.

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+                <p className="mt-1 text-gray-500 text-sm">{meta.description}</p>
+              </div>
+              <button
+                type="button"
+                onClick={toggleParameters}
+                className="-mt-1 -mr-1 flex h-7 w-7 shrink-0 items-center justify-center rounded text-gray-400 hover:bg-gray-100 hover:text-gray-600"
+                title="Close panel"
+              >
+                <Icon icon="mdi:close" className="text-lg" />
+              </button>
```
bilalabbad: You can use Button variant ghost.

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
```diff
@@ -0,0 +1,98 @@
...
+                >
+                  {MODE_LABELS[m]}
+                </button>
+              ))}
+            </div>
+          </div>
+
+          <div className="flex-1 overflow-y-auto">
+            {mode === "path" ? <PathModeSidebar /> : <DependenciesModeSidebar />}
+          </div>
```
bilalabbad, line 101: you could have use react aria ListBox + listBoxItem to make it keyboard navigable

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
bilalabbad: Instead or creating a custom `<aside />`, you could use Card component

frontend/app/src/entities/nodes/object/ui/object-table/toolbar/actions/groups/toolbar-add-to-groups-action.test.tsx
pa-lem: the goal is to provide an explicit indication to the user that we can also search by uuid which was not possible before

we can try something like "Search by value or UUID"

it can also wait for SA feedbacks

frontend/app/src/shared/components/inputs/kind-multi-select.tsx
pa-lem: I could not reproduce it, it's using the combobox items, is it still an issue for you? is it the case only here?

image

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
pa-lem: it has been decided like that after some prototypes and comparisons to also match what we have in the schema viewer, more improvements should be done in both path traversal and schema viewer but the layout will remain like that

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
pa-lem: navigation with keyboard is enabled, not sure to understand what you mean? for the modes buttons?

frontend/app/src/entities/path-traversal/ui/path-traversal-page.tsx
pa-lem: not sure to understand, it's the current logic for most of the pages

bilalabbad (approved):

PR #8626, opened 2026-03-17
Stop using `raise_for_error` for SDK functions

gmazoyer:
This parameter has been deprecated for a while and will soon be removed.

## Summary by CodeRabbit

* **Refactor**
  * Test fixtures now use protocol constants instead of string literals for pool types.

* **Tests**
  * GraphQL tests updated to rely on exception behavior for clearer error assertions.

* **Chores**
  * Extended type-checker overrides for test fixtures to suppress a specific validation.
  * Updated Python SDK reference.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1854,6 +1854,7 @@ include = ["backend/tests/fixtures/**"]
 no-matching-overload = "ignore"
 possibly-missing-attribute = "ignore"
 unresolved-import = "ignore"
+invalid-argument-type = "ignore"
 
 [[tool.ty.overrides]]
 include = ["backend/tests/unit/**"]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -1854,6 +1854,7 @@ include = ["backend/tests/fixtures/**"]
 no-matching-overload = "ignore"
 possibly-missing-attribute = "ignore"
 unresolved-import = "ignore"
+invalid-argument-type = "ignore"
```
gmazoyer, line 1857: Added this here, because this seem to be ignored by mypy but not ty, and it kinda is tricky to fix this typing (for now).

python_sdk
ogenstad, line 1: Just a question around this commit. Was that from the `infrahub-develop` branch after you merged your PR or from your branch in the SDK?

gmazoyer: This is the commit from `infrahub-develop`

ogenstad (approved):
LGTM there's a merge conflict on the sdk commit now. Probably since 1.8 was released and brought back into `develop`.

PR #8644, opened 2026-03-19
Migrate frontend to pnpm workspaces

pa-lem:
## 
## Release Notes

* **Chores**
  * Migrated frontend dependency management from npm to pnpm for improved monorepo handling.
  * Updated CI/CD pipelines and development environment setup scripts.
  * Centralized workspace configuration for better package dependency resolution.

* **Documentation**
  * Updated developer setup guides and contributor instructions to reflect package manager changes.

* Migrate the frontend from npm to pnpm workspaces, enabling proper multi-package management for app, ui, schema-visualizer, and future plugins
* Local packages now use workspace:* protocol instead of file: links
* Remove --prefix hacks from scripts — pnpm workspace handles cross-package linking
* Centralize dependency overrides in the workspace root frontend/package.json
* Update CI workflows (ci.yml, chromatic.yml) to use pnpm/action-setup@v4 + pnpm install --frozen-lockfile
* Update Dockerfile to single workspace-level pnpm install instead of per-package npm install
* Update all documentation and developer scripts to reference pnpm

## Why

As we add more packages under frontend/packages/ (ui, schema-visualizer, plugins), the current approach with file: links and manual npm install --prefix doesn't scale. pnpm workspaces give us:

* Single install and lockfile for all frontend packages
* Strict dependency resolution (catches undeclared/phantom deps)
* Faster installs with content-addressable store
* Clean workspace:* protocol that makes local deps explicit

## Migration for developers

```
# Install pnpm (one-time)
npm install -g pnpm

# Install all frontend dependencies
cd frontend && pnpm install

# Everything else works the same from frontend/app/
pnpm run dev
pnpm run test
pnpm run build
```


pa-lem:
done in https://github.com/opsmill/infrahub/pull/8667

PR #8651, opened 2026-03-19
add syslog forwarding config options

ajtmccarty:
IFC-2349

- adds logging configuration as defined in phase 1 of the [syslog forwarding plan](https://github.com/opsmill/infrahub-private/pull/338/changes#diff-a04b5305508a134a7602e76c42c39bc53378ab5f66873fdbe2ed250e58e20220) 
- updates `task/docs.py` to handle nested complex objects when generating `configuration.mdx`
- includes a test to confirm that multiple destinations can be easily defined in `infrahub.toml`
- to configure destinations using env vars, user needs to set the `INFRAHUB_LOG_FORWARDING_DESTINATIONS` env var to a JSON blob that looks something like this
```
'[{"name": "siem1", "type": "syslog", "host": "host1.example.com", "port": 514}, {"name": "siem2", "type": "syslog", "host": "host2.example.com", "port": 1514, "protocol": "udp"}]'
```
- if necessary, we can update env var handling to support single env vars like the below, but that would be a separate PR later
```
INFRAHUB_LOG_FORWARDING_DESTINATION_1_NAME = "siem1"
INFRAHUB_LOG_FORWARDING_DESTINATION_1_host = "..."
...
INFRAHUB_LOG_FORWARDING_DESTINATION_2_NAME = "siem2"
INFRAHUB_LOG_FORWARDING_DESTINATION_2_host = "..."
```

## Summary by CodeRabbit

* **New Features**
  * Added enterprise log forwarding with syslog support: multiple destinations, TCP/UDP, formats/framing, TLS, queuing, and per-destination controls.

* **Behavior / Validation**
  * Destination names must be unique; TLS over UDP is rejected; adding destinations enables the log-forwarding enterprise feature.

* **Tests**
  * New unit tests for parsing, validation, env/TOML loading, and feature aggregation.

* **Documentation**
  * Updated configuration reference and env var types for log forwarding and array schemas.

### DIFF
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -902,7 +902,9 @@ class LogForwardingDestination(BaseModel):
         default=LogForwardingDestinationType.SYSLOG, description="Destination type."
     )
     host: str = Field(description="Destination host or IP address.")
-    port: int = Field(ge=1, le=65535, description="Destination port number.")
+    port: int | None = Field(
+        default=None, ge=1, le=65535, description="Destination port number. Defaults to 6514 for TLS, 514 otherwise."
+    )
     protocol: SyslogProtocol = Field(default=SyslogProtocol.TCP, description="Transport protocol (tcp or udp).")
     format: SyslogFormat = Field(default=SyslogFormat.RFC5424, description="Syslog format standard.")
     tcp_framing: TcpFraming = Field(
@@ -926,6 +928,14 @@ class LogForwardingDestination(BaseModel):
         default=ExtraLogLevel.WARNING,
         description="Minimum Python log severity to forward when application log forwarding is enabled.",
     )
+
+    @property
+    def service_port(self) -> int:
+        if self.port:
+            return self.port
+        if self.tls_enabled:
+            return 6514
+        return 514
 
     @model_validator(mode="after")
     def validate_tls_protocol(self) -> Self:
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -944,10 +944,13 @@ class LogForwardingSettings(BaseSettings):
     @field_validator("destinations")
     @classmethod
     def validate_unique_names(cls, v: list[LogForwardingDestination]) -> list[LogForwardingDestination]:
-        names = [d.name for d in v]
-        if len(names) != len(set(names)):
-            raise ValueError("Destination names must be unique")
-        return v
+        unique_names = {d.name for d in v}
+        if len(unique_names) == len(v):
+            return v
+        all_names = [d.name for d in v]
+        duplicate_names = {name for name in unique_names if all_names.count(name) > 1}
+        sorted_dupes = ", ".join(sorted(duplicate_names))
+        raise ValueError(f"Destination names must be unique; duplicates found: {sorted_dupes}")
 
     @property
     def enterprise_features(self) -> list[EnterpriseFeatures]:
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -53,7 +53,7 @@ def default_append_git_suffix_domains() -> list[str]:
 class EnterpriseFeatures(StrEnum):
     PROPOSED_CHANGE_REQUIRE_APPROVAL = "proposed_change_require_approval"
     REVOKE_PROPOSED_CHANGE_APPROVALS = "revoke_proposed_change_approvals"
-    LOG_FORWARDING_SYSLOG = "log_forwarding_syslog"
+    LOG_FORWARDING = "log_forwarding"
 
 
 class UserInfoMethod(StrEnum):
@@ -956,7 +956,7 @@ class LogForwardingSettings(BaseSettings):
     def enterprise_features(self) -> list[EnterpriseFeatures]:
         """Returns enterprise features enabled by log forwarding configuration."""
         if any(d.type == LogForwardingDestinationType.SYSLOG for d in self.destinations):
-            return [EnterpriseFeatures.LOG_FORWARDING_SYSLOG]
+            return [EnterpriseFeatures.LOG_FORWARDING]
         return []
 
 
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -8,7 +8,7 @@ import tomllib
 from dataclasses import dataclass
 from enum import StrEnum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 from infrahub_sdk.utils import generate_uuid
 from pydantic import (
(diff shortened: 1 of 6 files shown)


backend/infrahub/config.py
```diff
@@ -867,6 +868,80 @@ class TraceSettings(BaseSettings):
...
+
+
+class TcpFraming(StrEnum):
+    NEWLINE = "newline"
+    OCTET_COUNTING = "octet-counting"
+
+
+class LogForwardingDestination(BaseSettings):
+    name: str = Field(description="Unique name for the destination, used in all observability output.")
+    type: Literal["syslog"] = Field(description="Destination type.")
```
polmichel: Is the idea behind this `Literal["syslog"] `to evolve it into a `StrEnum` later?

backend/infrahub/config.py
```diff
@@ -867,6 +868,80 @@ class TraceSettings(BaseSettings):
...
+    host: str = Field(description="Destination host or IP address.")
+    port: int = Field(ge=1, le=65535, description="Destination port number.")
```
polmichel: I usually find it clearer to write numbers in groups of three digits separated by underscores (e.g. `65_535`) . But I'm just throwing out a thought there, this is obviously neither important nor a blocker.

backend/infrahub/config.py
ajtmccarty: that was the idea
I've updated it to StrEnum. it might be a single-item enum forever, but I think that's fine

backend/infrahub/config.py
```diff
@@ -867,6 +868,80 @@ class TraceSettings(BaseSettings):
...
+        default=False, description="Forward application log messages to this destination."
+    )
+    min_log_severity: ExtraLogLevel = Field(
+        default=ExtraLogLevel.WARNING,
+        description="Minimum Python log severity to forward when application log forwarding is enabled.",
+    )
+
+    @model_validator(mode="after")
+    def validate_tls_protocol(self) -> Self:
+        if self.tls_enabled and self.protocol == SyslogProtocol.UDP:
```
polmichel, line 942: Would it be interesting to add other parameters? Like `tcp_framing`?

backend/infrahub/config.py
```diff
@@ -867,6 +868,80 @@ class TraceSettings(BaseSettings):
...
+        default=60, ge=1, description="Maximum reconnection backoff interval in seconds."
+    )
+    shutdown_drain_timeout: int = Field(
+        default=10, ge=0, description="Seconds to wait for queue drain on graceful shutdown."
+    )
+    forward_application_logs: bool = Field(
+        default=False, description="Forward application log messages to this destination."
+    )
+    min_log_severity: ExtraLogLevel = Field(
+        default=ExtraLogLevel.WARNING,
```
polmichel, line 928: Not directly related to your code, but given that there is the concept of a minimum level, would it be worth to establish an order within the enumeration `ExtraLogLevel` ? Like implementing some custom enum class like
```
from functools import total_ordering

@total_ordering
class OrderedStrEnum(StrEnum):
    def __lt__(self, other):
        members = list(self.__class__)
        return members.index(self) > members.index(other)
    
    def __eq__(self, other):
        return self.value == other.value
```

backend/infrahub/config.py
```diff
@@ -867,6 +868,80 @@ class TraceSettings(BaseSettings):
...
+        default_factory=list,
+        description="List of log forwarding destinations. (Enterprise only: not available in the community version.)",
+    )
+
+    @field_validator("destinations")
+    @classmethod
+    def validate_unique_names(cls, v: list[LogForwardingDestination]) -> list[LogForwardingDestination]:
+        names = [d.name for d in v]
+        if len(names) != len(set(names)):
+            raise ValueError("Destination names must be unique")
```
polmichel: I am not sure what the use case is or if this message will be useful to anyone. If not, please ignore this comment.

Perhaps something like this could help users in the event of real duplication? 
```
dupes = {n for n in names if names.count(n) > 1}                                                                                                                      
  if dupes:                                                                                                                                                             
      raise ValueError(f"Duplicate destination names: {', '.join(sorted(dupes))}. Destination names must be unique")      
```

backend/infrahub/config.py
polmichel: (This is more of an idea for future developments than a review)

backend/infrahub/config.py
polmichel: It is less effective from a performance perspective, but I don't think that matters here.

backend/infrahub/config.py
ajtmccarty: cool idea. I don't think this would do much right now, but it might be valuable later in this project

backend/infrahub/config.py
ajtmccarty: I don't think any of the other settings are good candidates for this kind of validation.
the TLS validation is there b/c it is a security concern
we don't want validation for `tcp_framing` b/c we want to have a default value and that can just be ignored if using UDP
`tls_ca_bundle` is optional even if `tls_enabled` is True

backend/infrahub/config.py
ajtmccarty: good idea
added the dup names to the error message

polmichel (approved):
LGTM !

backend/infrahub/config.py
```diff
@@ -53,6 +53,7 @@ def default_append_git_suffix_domains() -> list[str]:
 class EnterpriseFeatures(StrEnum):
     PROPOSED_CHANGE_REQUIRE_APPROVAL = "proposed_change_require_approval"
     REVOKE_PROPOSED_CHANGE_APPROVALS = "revoke_proposed_change_approvals"
+    LOG_FORWARDING_SYSLOG = "log_forwarding_syslog"
```
ogenstad: As syslog should just be one adapter and we want to have the ability to send logs to different types of destinations in the future, i.e. RabbitMQ, Kafka, Redis, HTTP endpoints etc, I wonder if we should name the enterprise feature as `LOG_FORWARDING` and skip the syslog suffix.

backend/infrahub/config.py
ogenstad: I'd leave the port as optional and look at what we're doing for BrokerSettings with the `service_port` property. I.e. if we can have sensible defaults to the port and update those defaults when we select to use TLS or just when we use default UDP and we know that the port should be 514 by default.

backend/infrahub/config.py
```diff
@@ -867,6 +868,89 @@ class TraceSettings(BaseSettings):
...
+
+
+class LogForwardingDestination(BaseModel):
+    name: str = Field(description="Unique name for the destination, used in all observability output.")
+    type: LogForwardingDestinationType = Field(
+        default=LogForwardingDestinationType.SYSLOG, description="Destination type."
+    )
+    host: str = Field(description="Destination host or IP address.")
+    port: int = Field(ge=1, le=65535, description="Destination port number.")
+    protocol: SyslogProtocol = Field(default=SyslogProtocol.TCP, description="Transport protocol (tcp or udp).")
```
ogenstad: I'm not sure what the default protocol should be. I was thinking that UDP was more common, but I'm not completely sure.

backend/infrahub/config.py
ajtmccarty: removed _SYSLOG

backend/infrahub/config.py
ajtmccarty: updated to default UDP

backend/infrahub/config.py
ajtmccarty: added a `service_port` property

ogenstad (approved):

PR #8653, opened 2026-03-19
add log forwarding service stub

ajtmccarty:
IFC-2350

should follow #8651 

adds interface class, models, and no-op community subclass for LogForwardingService

completes phase 1 of the [syslog plan](https://github.com/opsmill/infrahub-private/pull/338/changes#diff-a04b5305508a134a7602e76c42c39bc53378ab5f66873fdbe2ed250e58e20220)

## Summary by CodeRabbit

* **New Features**
  * Introduced log forwarding service for audit events and application logs
  * Added RFC 5424-compliant severity level classification
  * Implemented extensible service architecture supporting log destination handling

### DIFF
--- a/backend/infrahub/log_forwarding/models.py
+++ b/backend/infrahub/log_forwarding/models.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
+from dataclasses import dataclass
 from datetime import datetime  # noqa: TC003
 from enum import IntEnum, StrEnum
-
-from pydantic import BaseModel, computed_field
 
 LOG_AUTH = 4
 LOG_LOCAL0 = 16
@@ -27,15 +26,15 @@ class SyslogSeverity(IntEnum):
     DEBUG = 7
 
 
-class SyslogMessage(BaseModel):
+@dataclass(slots=True)
+class SyslogMessage:
     message_type: MessageType
     timestamp: datetime
     payload: str  # JSON for audit events, text for app logs
-    event_type: str | None = None  # e.g. "infrahub.node.created" (audit only)
     severity: int  # RFC 5424 severity code
     process_id: str  # worker identity
+    event_type: str | None = None  # e.g. "infrahub.node.created" (audit only)
 
-    @computed_field  # type: ignore[prop-decorator]
     @property
     def facility(self) -> int:
         """RFC 5424 facility: LOG_AUTH (4) for audit events, LOG_LOCAL0 (16) for app logs."""
--- a/backend/infrahub/log_forwarding/models.py
+++ b/backend/infrahub/log_forwarding/models.py
@@ -1,13 +1,9 @@
 from __future__ import annotations
 
+from datetime import datetime  # noqa: TC003
 from enum import IntEnum, StrEnum
-from typing import TYPE_CHECKING
 
 from pydantic import BaseModel, computed_field
-
-if TYPE_CHECKING:
-    from datetime import datetime
-
 
 LOG_AUTH = 4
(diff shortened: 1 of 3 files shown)


gmazoyer (approved):

backend/infrahub/log_forwarding/models.py
```diff
@@ -0,0 +1,45 @@
...
+LOG_AUTH = 4
+LOG_LOCAL0 = 16
+
+
+class MessageType(StrEnum):
+    AUDIT_EVENT = "audit_event"
+    APPLICATION_LOG = "application_log"
+
+
+class SyslogSeverity(IntEnum):
```
polmichel, line 16: Not a review, just a thought, this is another valid way to do an "OrderedStrEnum" -> https://github.com/opsmill/infrahub/pull/8651#discussion_r2967893167

backend/infrahub/log_forwarding/models.py
```diff
@@ -0,0 +1,45 @@
...
+    ALERT = 1
+    CRITICAL = 2
+    ERROR = 3
+    WARNING = 4
+    NOTICE = 5
+    INFORMATIONAL = 6
+    DEBUG = 7
+
+
+class SyslogMessage(BaseModel):
```
ogenstad: I'm not sure if it makes a huge difference as other things we send will be based on Pydantic. But given the discussion around performance could it make sense to have this as a `dataclass` with slots instead? Not sure how much we'd lose with the facility property..

(not a show stopper, just a comment and question)

backend/infrahub/log_forwarding/models.py
```diff
@@ -0,0 +1,45 @@
...
+    message_type: MessageType
+    timestamp: datetime
+    payload: str  # JSON for audit events, text for app logs
+    event_type: str | None = None  # e.g. "infrahub.node.created" (audit only)
+    severity: int  # RFC 5424 severity code
+    process_id: str  # worker identity
```
ogenstad, line 35: If we end up keeping this as a Pydantic class I think it's cleaner to use `Field` and set these comments as descriptions to the field instead of inline comments.

backend/tests/unit/log_forwarding/test_models.py
```diff
@@ -0,0 +1,71 @@
...
+        timestamp=datetime(2026, 3, 19, 12, 0, 0, tzinfo=UTC),
+        payload='{"action": "node.created", "user": "admin"}',
+        event_type="infrahub.node.created",
+        severity=6,
+        process_id="worker-1",
+    )
+    assert msg.message_type == MessageType.AUDIT_EVENT
+    assert msg.event_type == "infrahub.node.created"
+    assert msg.facility == LOG_AUTH
+    assert msg.severity == 6  # Informational
```
ogenstad: Compare against SyslogSeverity instead of a `6` ?

ogenstad (approved):
LGTM. Added some comments for consideration.

backend/infrahub/log_forwarding/models.py
ajtmccarty: going with the dataclass for now

backend/infrahub/log_forwarding/models.py
ajtmccarty: I prefer a dataclass. I'll switch to that for now

backend/tests/unit/log_forwarding/test_models.py
ajtmccarty: good point. updated

PR #8641, opened 2026-03-18
IFC-2272: Webhook custom headers support

polmichel:
# Why

Webhooks currently lack support for custom headers, which are required by many webhook consumers for authentication, routing, and metadata. This PR adds the ability to define custom headers via `CoreKeyValue` entries and resolves them at runtime.

Closes IFC-2272

## What changed

This is a roll-up of the following PRs:

- #8590 — **IFC-2303**: CoreKeyValue schema definitions for webhook custom headers
- #8597 — **IFC-2304**: Runtime resolution of webhook custom headers
- #8602 — **IFC-2334**: Webhook cache invalidation on KeyValue changes
- #8603 — **IFC-2337**: Invalidate webhook cache on `reconcile_all`
- #8606 — **IFC-2257**: Unit tests for webhook signature validation

Key behavioral changes:
- Webhook definitions can now reference custom headers stored as `CoreKeyValue` entries
- Headers are resolved at dispatch time, ensuring up-to-date values
- Webhook cache is properly invalidated when underlying KeyValue entries change

## How to review

Each constituent PR was reviewed individually. This roll-up merges the full feature branch into develop.

### Unit tests
```bash
uv run pytest backend/tests/unit/webhook/test_webhook_header.py -v
```

### Component tests
```bash
uv run pytest backend/tests/component/graphql/mutations/test_webhook.py -v
```

### Functional tests
```bash
uv run pytest backend/tests/functional/webhook/ -v
```

Covers: `test_configure.py`, `test_invalidate.py`, `test_process.py`

### Frontend E2E
```bash
cd frontend/app && npx playwright test tests/e2e/webhook/webhook.spec.ts
```

## Impact & rollout

- **Backward compatibility:** No breaking changes — existing webhooks without custom headers are unaffected
- **Config/env changes:** None
- **Deployment notes:** Safe to deploy

## Checklist

- [x] Tests added/updated

## Release Notes

* **New Features**
  * Webhooks now support custom HTTP headers with static values and environment-variable resolution.
  * Introduced Key-Value node types for managing webhook headers and configuration pairs.

* **Documentation**
  * Added comprehensive guides covering code generation, schema definitions, webhook configuration, and testing patterns.

PR #8654, opened 2026-03-19
IFC-2348 Webhook headers: unresolved feedbacks and last thoughts

polmichel:
# Why

CodeRabbit left 7 findings on PR #8641 (webhook headers). After reviewing each against the current code, 4 warranted fixes (the others were either by-design or already handled elsewhere).

## What changed

- **Duplicate header key warning:** `_assign_headers()` now logs a warning when `custom_headers` contains duplicate keys instead of silently letting the last value win. Unit test added.
- **Safe `event_data` access:** `invalidate.py` helpers used `event_data["node_id"]` which raises `KeyError` on malformed payloads. Replaced with `.get()` so the fallback paths are reached cleanly.
- **Order-independent test assertion:** `test_convert_node_to_webhook_with_headers` compared `custom_headers` as an ordered list. Relationship query order from the database is non-deterministic, so the assertion now sorts by key.
- **Docs: SDK protocols note:** `code-generation.md` listed the files checked by `backend.validate-generated` but omitted `python_sdk/infrahub_sdk/protocols.py`. Added a note that it is generated but not validated (lives in a Git submodule).

**What stayed the same:** No schema changes, no trigger changes, API contract unchanged.

### Findings intentionally not addressed

| # | Finding | Reason |
|---|---------|--------|
| 1 | `key_value.py` — use secret-backed attribute for `value` | By design: `StaticKeyValue` is explicitly plain-text; `EnvKeyValue` stores env var names, not secrets |
| 4+7 | Add create/delete events to `TRIGGER_KEYVALUE_WEBHOOK_INVALIDATE` | Create and delete cases are already handled elsewhere |

## How to test

```bash
uv run pytest backend/tests/unit/webhook/test_models.py -x -q
```

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added detection and warning for duplicate webhook header keys to surface configuration issues.

* **Bug Fixes**
  * Made webhook header invalidation resilient to missing or malformed event data to avoid errors.

* **Tests**
  * Added unit test for duplicate header key detection.
  * Updated functional test assertions to avoid relying on header ordering.

* **Documentation**
  * Clarified note about generated SDK files being excluded from a validation step.


ajtmccarty (approved):

backend/infrahub/webhook/models.py
```diff
@@ -169,7 +169,15 @@ def _assign_headers(self, uuid: UUID | None = None, at: Timestamp | None = None)
...
 
+        seen_keys: set[str] = set()
         for header in self.custom_headers:
+            if header.key in seen_keys:
+                logger.warning(
+                    "Webhook '%s': duplicate header key '%s', later value will overwrite earlier one",
+                    self.name,
+                    header.key,
+                )
+            seen_keys.add(header.key)
```
ogenstad, line 180: Potentially we could have `seen_keys.add(header.key)` after `self._headers[header.key] = header.resolve()` the only reason would be if there's a lookup failure on the first variable but the second one is static. I thing it's also fine as it is just a small consideration.

ogenstad (approved):

PR #8665, opened 2026-03-20
add wiring for LogForwardingService

ajtmccarty:
IFC-2351

wiring the LogForwardingService into the app

should follow #8653 

## Summary by CodeRabbit

* **New Features**
  * Integrated log forwarding capability into the event system, enabling the application to forward events to external logging services asynchronously without blocking operations.

### DIFF
--- a/backend/infrahub/server.py
+++ b/backend/infrahub/server.py
@@ -39,6 +39,7 @@ from infrahub.workers.dependencies import (
     get_component,
     get_database,
     get_installation_type,
+    get_log_forwarding_service,
     get_message_bus,
     get_workflow,
     set_component_type,
@@ -73,6 +74,7 @@ async def app_initialization(application: FastAPI, enable_scheduler: bool = True
     message_bus = await get_message_bus()
     cache = await get_cache()
     component = await get_component()
+    log_forwarding = get_log_forwarding_service()
     service = await InfrahubServices.new(
         cache=cache,
         database=database,
@@ -80,7 +82,9 @@ async def app_initialization(application: FastAPI, enable_scheduler: bool = True
         workflow=workflow,
         component=component,
         component_type=component_type,
+        log_forwarding=log_forwarding,
     )
+    await log_forwarding.start()
     initialize_lock(service=service)
     # We must initialize DB after initialize lock and initialize lock depends on cache initialization
     async with application.state.db.start_session() as db:
--- a/backend/infrahub/services/adapters/event/__init__.py
+++ b/backend/infrahub/services/adapters/event/__init__.py
@@ -48,11 +48,7 @@ class InfrahubEventService:
         )
 
     def _send_log_forwarding(self, event: InfrahubEvent) -> None:
-        """Forward an event to the log forwarding service."""
         if self.log_forwarding is None:
             return
 
-        try:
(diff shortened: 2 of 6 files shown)


backend/infrahub/server.py
```diff
@@ -73,14 +74,17 @@ async def app_initialization(application: FastAPI, enable_scheduler: bool = True
     message_bus = await get_message_bus()
     cache = await get_cache()
     component = await get_component()
+    log_forwarding = get_log_forwarding_service()
     service = await InfrahubServices.new(
```
polmichel, line 78: Is this also required in the `InfrahubWorkerAsync` class? Maybe not at the moment?
`infrahub.workers.infrahub_async.InfrahubWorkerAsync._init_services`
```
        service = await InfrahubServices.new(
            cache=await get_cache(),
            client=client,
            database=await get_database(),
            message_bus=await get_message_bus(),
            workflow=get_workflow(),
            component=await get_component(),
            component_type=self.component_type,
        )
```

polmichel:
And is a test service required there? `backend/tests/adapters`
There are already `http`, `cache`, `event`, ...

Maybe is this part of a later task? Like the one about _internal: E2E integration tests_ (internal link: https://opsmill.atlassian.net/browse/IFC-2360)

backend/infrahub/server.py
polmichel: Maybe is this part of another task, if so, please ignore this comment
In the specification it states "_Each Infrahub worker process maintains its own independent syslog connection(s)._"

backend/infrahub/services/adapters/event/__init__.py
```diff
@@ -36,3 +46,13 @@ async def _send_prefect(self, event: InfrahubEvent) -> None:
...
         )
+
+    def _send_log_forwarding(self, event: InfrahubEvent) -> None:
+        """Forward an event to the log forwarding service."""
+        if self.log_forwarding is None:
+            return
+
+        try:
+            self.log_forwarding.forward_event(event)
+        except Exception:
```
ogenstad: If we expect this to fail I think it would be nice to be able to restrict the exception. It should basically just add it to the queue (which depending on how it's done could potentially fail if the queue is full). Other than that the other task would be to optionally convert the message to a different format depending on the driver. Provided that we've done our job this conversion shouldn't fail. Is this just a placeholder or are we expecting some other type of error that could come up here?

backend/infrahub/server.py
ajtmccarty: no you are correct. I've added this to the `InfrahubWorkerAsync` as well. thanks for the thorough review

backend/infrahub/services/adapters/event/__init__.py
ajtmccarty: good point. I've removed the exception handling here

ajtmccarty:
> And is a test service required there? `backend/tests/adapters` There are already `http`, `cache`, `event`, ...

one might be needed eventually, but I'll wait until I'm adding a test that requires it

polmichel (approved):
LGTM ! The design is clean. Moreover, having small and targeted PRs to review is great!

ogenstad (approved):

PR #8672, opened 2026-03-23
IFC-2380: Jinja2 local recomputation code documentation & code simplification

polmichel:
## Summary

- Simplify `register_computed_jinja2` in `schema_branch_computed.py`: consolidate attribute and relationship dependency registration through unified control flow
- Refactor `get_jinja2_trigger_nodes` with extracted helper and clearer structure
- Add method-level docstrings to `schema_branch_computed.py` (classes, properties, methods)
- Add docstrings to Prefect task functions in `computed_attribute/tasks.py`
- Remove implementation details from docstrings (keep them focused on behavior, not internals)
- Add unit tests for `register_computed_jinja2` and `get_jinja2_trigger_nodes`

## Files Changed

- `backend/infrahub/core/schema/schema_branch_computed.py` — code simplification, docstrings, refactor
- `backend/infrahub/computed_attribute/tasks.py` — docstrings
- `backend/tests/unit/core/test_schema_branch_computed.py` — new unit tests

Closes [IFC-2380]

[IFC-2380]: https://opsmill.atlassian.net/browse/IFC-2380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/core/schema/schema_branch_computed.py
+++ b/backend/infrahub/core/schema/schema_branch_computed.py
@@ -164,15 +164,14 @@ class ComputedAttributes:
     ) -> None:
         """Register a Jinja2 computed attribute and its dependency graph.
 
-        Populates the ``_computed_jinja2_attribute_map`` so the system knows which nodes/fields
-        trigger recomputation of which computed attributes. The map is keyed by the kind of node
-        whose mutation should trigger recomputation.
-
         Args:
             node: The schema owning the computed attribute (e.g. InfraDevice).
             attribute: The computed attribute definition (e.g. ``computed_name``).
             schema_path: Parsed Jinja2 template token identifying the dependency — either a
                          local attribute or a relationship + peer attribute.
+
+        Returns:
+            None
         """
         # Determine the trigger key: for local attributes it's the node itself,
         # for relationship attributes it's the peer kind.
--- a/backend/infrahub/core/schema/schema_branch_computed.py
+++ b/backend/infrahub/core/schema/schema_branch_computed.py
@@ -188,12 +188,10 @@ class ComputedAttributes:
         if schema_path.is_type_relationship:
             rel_name = schema_path.active_relationship_schema.name
 
-            # Record the relationship on the *peer* entry so get_targets() can
-            # build filter_keys (e.g. "site__ids") to locate affected owner nodes.
+            # Record the relationship on the *peer* entry
             trigger_node.relationships.setdefault(rel_name, []).append(deepcopy(source_attribute))
 
-            # Register on the *owner* entry so that a relationship re-assignment
-            # (e.g. changing which site a device points to) also triggers recomputation.
+            # Register on the *owner* entry so that a relationship re-assignment also triggers recomputation.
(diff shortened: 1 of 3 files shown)


backend/infrahub/core/schema/schema_branch_computed.py
```diff
@@ -128,44 +162,40 @@ def python_attributes_by_transform(self) -> dict[str, list[PythonDefinition]]:
...
+                         local attribute or a relationship + peer attribute.
+
+        Returns:
+            None
+        """
+        # Determine the trigger key: for local attributes it's the node itself,
+        # for relationship attributes it's the peer kind.
+        key = node.kind if not schema_path.is_type_relationship else schema_path.active_relationship_schema.peer
+
+        trigger_node = self._computed_jinja2_attribute_map.setdefault(key, RegisteredNodeComputedAttribute())
```
ogenstad, line 180: I like the fact that we're simplifying the code like this, a comment though is that It's somewhat expensive to create a Pydantic class. Having said that in this case it's not exactly in the hot path as we only see it when we are loading the schema and for any given run we'd only have a very small number of extra calls for this. Highlighting for consideration if we use the same approach elsewhere within the app where we have more traffic.

backend/tests/unit/core/test_schema_branch_computed.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def _fake_node(kind: str) -> Any:
+    """Create a minimal object with a .kind attribute, sufficient for register_computed_jinja2."""
+
+    class _Node:
+        pass
+
+    node = _Node()
+    node.kind = kind  # type: ignore[attr-defined]
```
ogenstad: I think we should avoid creating ignores for type exceptions where we don't need it. In this case I'd just create a dataclass object for Node within this file and have "kind" be an attribute that you define on Node creation. We can just call it DummyNode. Then we can remove the _fake_node function that returns an `Any` and just create an object from the dataclass directly wherever we use _fake_node()

ogenstad (approved):
LGTM, just think we should change the Node stuff around the comments about `_fake_node()`

backend/tests/unit/core/test_schema_branch_computed.py
polmichel: It will create a typing error due to the type expected by this method
`infrahub.core.schema.schema_branch_computed.ComputedAttributes.register_computed_jinja2`

I will create a real `NodeSchema`

backend/infrahub/core/schema/schema_branch_computed.py
polmichel: I did not realise that the default object was instantiated when using `setDefault()`. It seems that using a `defaultdict` is the only solution in this kind of situation to avoid unnecessary object instantiation. But it does not fit here

In Java there is the `requireNonNullElseGet` that takes a lambda as input vs `requireNonNull` which takes and evaluates the default value

PR #8677, opened 2026-03-23
Local test AI Bug pipeline #2: suppress Neo4j warnings for optional relationship types (closes #8620)

### LINKED ISSUE #8620: bug: noisy and irrelevant error messages in the neo4j logs
### Component

API Server / GraphQL

### Infrahub version

1.7.x

### Current Behavior

if an Infrahub instance does not include any "owner" metadata for relationships or attributes, then cypher queries that look for these metadata can raise errors about the relationship not existing on the database

the error can be safely ignored, but it will be raised every time an object is retrieved from the database

the error looks something like this
```
Received notification from DBMS server: <GqlStatusObject gql_status='01N51', status_description='warn: relationship type does not exist. The relationship type `HAS_OWNER` does not exist in database `neo4j`. Verify that the spelling is correct.', position=<SummaryInputPosition line=30, column=24, offset=1325>, raw_classification='UNRECOGNIZED', classification=<NotificationClassification.UNRECOGNIZED: 'UNRECOGNIZED'>, raw_severity='WARNING', severity=<NotificationSeverity.WARNING: 'WARNING'>, diagnostic_record={'_classification': 'UNRECOGNIZED', '_severity': 'WARNING', '_position': {'offset': 1325, 'line': 30, 'column': 24}, 'OPERATION': '', 'OPERATION_CODE': '0', 'CURRENT_SCHEMA': '/'}> for query: ...
```

### Expected Behavior

We should suppress this error for optional relationship types, such as HAS_OWNER and HAS_SOURCE

### Steps to Reproduce

the error might only appear when the neo4j database is clustered

### Additional Information

the example error message from the user report is raised when the `RelationshipGetPeerQuery` is executed

polmichel:
# Why

Neo4j emits `01N51` warnings ("relationship type does not exist") whenever Cypher queries reference relationship types like `HAS_OWNER` or `HAS_SOURCE` in explicit `OPTIONAL MATCH` patterns and those types have never been created in the database. In clustered deployments, these warnings surface prominently and fire on every read query that requests metadata, creating significant log noise.

**Goal:** Eliminate these warnings by avoiding explicit relationship type patterns in `OPTIONAL MATCH` clauses for optional metadata relationships.

**Non-goals:** This does not change query semantics, results, or performance characteristics. Required `MATCH` patterns (non-optional) are intentionally left unchanged.

Closes #8620

## What changed

- Replaced explicit relationship type patterns (e.g., `[r:HAS_OWNER]`, `[r:HAS_SOURCE]`) in `OPTIONAL MATCH` clauses with generic patterns (`[r]`) filtered by `type(r) = $param` in `WHERE` clauses
- Relationship type names are passed as query parameters rather than string literals, which prevents Neo4j from checking its schema catalog for non-existent types
- Affected 5 locations across 2 files: `node.py` (3 locations) and `relationship.py` (2 locations)

## Fix strategy

This is a shallow issue with a mechanical fix. The root cause is that Neo4j's query planner checks its internal schema catalog for relationship types referenced in explicit pattern syntax (`[r:TYPE]`). When the type doesn't exist, it emits a warning. The `type(r)` function approach in a `WHERE` clause evaluates at runtime against actual relationships found, so no schema catalog lookup occurs for non-existent types.

The fix uses parameterized queries (`$rel_type_src`, `$rel_type_own`, `$rel_prop_types`) rather than string literals to pass type names. This has the added benefit that the type name strings don't appear in the query text at all, which is cleaner and avoids any false-positive detection by static analysis of query patterns.

For the multi-type pattern in `RelationshipDeleteQuery`, the explicit `[edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE]` syntax was replaced with `type(edge) IN $rel_prop_types` where the list is passed as a parameter.

## How to review

Key files:
- `backend/infrahub/core/query/node.py` — `_add_source_to_query`, `_add_owner_to_query`, and the profile source check in `query_init`
- `backend/infrahub/core/query/relationship.py` — `_add_node_property_query` and `RelationshipDeleteQuery.query_init`

## How to test

```bash
uv run pytest backend/tests/component/core/test_neo4j_notification_suppression.py -v --timeout=120
```

## Impact & rollout

- **Backward compatibility:** Fully backward compatible. Query semantics are identical — only the Cypher syntax changes.
- **Performance:** No measurable impact. The `type()` function is evaluated on already-matched relationships, and `OPTIONAL MATCH` with a generic pattern + WHERE filter has the same execution plan as an explicit type pattern.
- **Deployment notes:** No special deployment steps required. The fix takes effect immediately on next query execution.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

### DIFF
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -730,8 +730,8 @@ class NodeListGetAttributeQuery(Query):
             return
         source_query = """
 CALL (a) {
-    OPTIONAL MATCH (a)-[rel_source:HAS_SOURCE]-(source)
-    WHERE all(r IN [rel_source] WHERE ( %(branch_filter)s ))
+    OPTIONAL MATCH (a)-[rel_source]-(source)
+    WHERE type(rel_source) = $rel_type_src AND all(r IN [rel_source] WHERE ( %(branch_filter)s ))
     RETURN source, rel_source
     ORDER BY rel_source.branch_level DESC, rel_source.from DESC, rel_source.status ASC
     LIMIT 1
@@ -740,6 +740,7 @@ WITH *,
     CASE WHEN rel_source.status = "active" THEN source ELSE NULL END AS source,
     CASE WHEN rel_source.status = "active" THEN rel_source ELSE NULL END AS rel_source
         """ % {"branch_filter": branch_filter_str}
+        self.params["rel_type_src"] = "HAS_SOURCE"
         self.add_to_query(source_query)
         self.return_labels.extend(["source", "rel_source"])
 
@@ -748,8 +749,8 @@ WITH *,
             return
         owner_query = """
 CALL (a) {
-    OPTIONAL MATCH (a)-[rel_owner:HAS_OWNER]-(owner)
-    WHERE all(r IN [rel_owner] WHERE ( %(branch_filter)s ))
+    OPTIONAL MATCH (a)-[rel_owner]-(owner)
+    WHERE type(rel_owner) = $rel_type_own AND all(r IN [rel_owner] WHERE ( %(branch_filter)s ))
     RETURN owner, rel_owner
     ORDER BY rel_owner.branch_level DESC, rel_owner.from DESC, rel_owner.status ASC
     LIMIT 1
@@ -758,6 +759,7 @@ WITH *,
     CASE WHEN rel_owner.status = "active" THEN owner ELSE NULL END AS owner,
     CASE WHEN rel_owner.status = "active" THEN rel_owner ELSE NULL END AS rel_owner
         """ % {"branch_filter": branch_filter_str}
+        self.params["rel_type_own"] = "HAS_OWNER"
         self.add_to_query(owner_query)
         self.return_labels.extend(["owner", "rel_owner"])
 
@@ -829,6 +831,7 @@ CALL (a) {
         self.params["profile_node_relationship_name"] = PROFILE_NODE_RELATIONSHIP_IDENTIFIER
         self.params["profile_template_relationship_name"] = PROFILE_TEMPLATE_RELATIONSHIP_IDENTIFIER
         self.params["field_names"] = list(self.fields.keys()) if self.fields else []
+        self.params["rel_type_src"] = "HAS_SOURCE"
 
         branch_filter, branch_params = self.branch.get_query_filter_path(
             at=self.at, branch_agnostic=self.branch_agnostic
@@ -859,8 +862,8 @@ WITH n, r1, a, might_use_profile
 WHERE r1.status = "active"
 WITH n, r1, a, might_use_profile
 CALL (a, might_use_profile) {
-    OPTIONAL MATCH (a)-[r:HAS_SOURCE]->(:CoreProfile)
-    WHERE might_use_profile = TRUE AND %(branch_filter)s
+    OPTIONAL MATCH (a)-[r]->(profile:CoreProfile)
+    WHERE type(r) = $rel_type_src AND might_use_profile = TRUE AND %(branch_filter)s
     RETURN r.status = "active" AS has_active_profile
     ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
     LIMIT 1
--- a/backend/infrahub/core/query/relationship.py
+++ b/backend/infrahub/core/query/relationship.py
@@ -642,8 +642,8 @@ class RelationshipDeleteQuery(RelationshipQuery):
         }
         WITH rl
 
-        OPTIONAL MATCH (rl)-[edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE]->(peer)
-        WHERE %(rel_filter)s
+        OPTIONAL MATCH (rl)-[edge]->(peer)
+        WHERE type(edge) IN $rel_prop_types AND %(rel_filter)s
         ORDER BY type(edge), edge.branch_level DESC, edge.from DESC, edge.status ASC
         WITH rl, type(edge) AS edge_type, head(collect(edge)) AS edge, head(collect(peer)) AS peer
 
@@ -667,6 +667,7 @@ class RelationshipDeleteQuery(RelationshipQuery):
         }
 
         self.params["at"] = self.at.to_string()
+        self.params["rel_prop_types"] = ["IS_PROTECTED", "HAS_OWNER", "HAS_SOURCE"]
         self.return_labels = ["rl"]
 
         self.add_to_query(query)
@@ -740,10 +741,12 @@ CALL (rl) {
         self.update_return_labels(["rel_is_protected", "is_protected"])
 
     def _add_node_property_query(self, node_prop: str, branch_filter: str) -> None:
+        prop_abbrev = {"owner": "own", "source": "src"}
+        param_name = f"rel_type_{prop_abbrev[node_prop]}"
         query = """
 CALL (rl) {
-    OPTIONAL MATCH (rl)-[r:HAS_%(node_prop_type)s]-(%(node_prop)s)
-    WHERE %(branch_filter)s
+    OPTIONAL MATCH (rl)-[r]-(%(node_prop)s)
+    WHERE type(r) = $%(param_name)s AND %(branch_filter)s
     WITH r, %(node_prop)s
     ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
     LIMIT 1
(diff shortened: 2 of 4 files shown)


polmichel (commented):
## Code Review: fix: suppress Neo4j warnings for optional relationship types

**Verdict: APPROVED WITH SUGGESTIONS**

---

### A. Correctness

The fix correctly addresses the root cause. Neo4j emits `01N51` warnings when Cypher queries reference relationship types in explicit pattern syntax (`[r:HAS_OWNER]`) that have never been created in the database. Replacing these with `OPTIONAL MATCH (a)-[r]-(b) WHERE type(r) = $param` avoids the schema catalog lookup for non-existent types.

All 5 locations are covered:
- `node.py`: `_add_source_to_query`, `_add_owner_to_query`, profile source check in `query_init`
- `relationship.py`: `_add_node_property_query`, `RelationshipDeleteQuery.query_init`

The non-optional `MATCH` patterns (`IS_PROTECTED` in `_add_is_protected_query`, `HAS_ATTRIBUTE`, `HAS_VALUE`) are correctly left unchanged -- these are required relationships where a warning would actually be meaningful.

The profile source check correctly preserves the `CoreProfile` node label constraint (moved from anonymous `:CoreProfile` to named `profile:CoreProfile`), maintaining semantic equivalence.

**Minor observation:** `self.params["rel_type_src"] = "HAS_SOURCE"` is set twice in `NodeListGetAttributeQuery` -- once in `query_init` (line 834, needed for the always-running profile check) and once in `_add_source_to_query` (line 743, gated by metadata options). Both set the same value so this is harmless, but the duplication could confuse future readers. Not blocking.

**Edge cases:** No uncovered edge cases. The `OPTIONAL MATCH (rl)-[edge]->(peer)` without an explicit type in `RelationshipDeleteQuery` is safe because `rl` is a Relationship node with a bounded set of outgoing edges (IS_PROTECTED, HAS_OWNER, HAS_SOURCE), and the `WHERE type(edge) IN $rel_prop_types` filter immediately narrows the match. The execution plan is equivalent.

### B. Code Quality

- Follows project conventions: parameterized queries via `self.params`, `%` string formatting for Cypher templates, consistent naming
- The `prop_abbrev` dict pattern in `_add_node_property_query` is clean and avoids magic strings
- No unnecessary refactoring -- the changes are surgical and mechanical
- No performance concerns: `type()` is evaluated on already-matched relationships, not as a full graph scan
- No security concerns: relationship type names are passed as query parameters, which is actually slightly better than the prior string literals

- The fix aligns with the Query Pattern documented in `dev/guidelines/backend/python.md`
- No public API or behavior change -- query results are identical
- Changelog entry is present and correctly worded
- No contradiction with internal guidelines

### D. Test Quality

The tests are well-structured:
- They test **behavior** (generated Cypher must not contain explicit type patterns in OPTIONAL MATCH) rather than implementation details
- The regex pattern `EXPLICIT_REL_TYPE_IN_MATCH_PATTERN` is carefully scoped: it catches `OPTIONAL MATCH ...[r:HAS_OWNER]...` but not `WHERE type(r) = "HAS_OWNER"`
- Tests use existing fixtures (`person_jack_tags_main`, `tag_blue_main`) from component conftest -- no custom schema duplication
- Tests are deterministic since they inspect generated query text, not runtime database behavior

**One concern:** These tests validate that the current code doesn't produce problematic patterns, but they're regression tests that inspect query strings. If someone adds a new query method with an explicit `OPTIONAL MATCH [r:HAS_OWNER]` pattern in a different file, these tests won't catch it. This is acceptable for now -- the tests cover all existing locations. A broader static analysis check could be added later if the pattern recurs.

**Nit on test placement:** The test file is in `backend/tests/component/core/` which is appropriate since it requires DB fixtures for query initialization. Naming follows conventions.

1. **Merge as-is** -- the fix is correct, minimal, and well-tested
2. **Optional cleanup** -- consider removing the redundant `self.params["rel_type_src"] = "HAS_SOURCE"` in `_add_source_to_query` since `query_init` already sets it unconditionally for the profile check. Low priority since it's harmless.
3. **Future consideration** -- if this pattern needs to be enforced project-wide, a linting rule or test that scans all `*.py` files in `core/query/` for `OPTIONAL MATCH.*\[:.*HAS_OWNER|HAS_SOURCE\]` would prevent regressions in new code.

polmichel:
## Root cause analysis
Root cause: Cypher queries use explicit relationship type patterns (e.g., [r:HAS_OWNER], [edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE]) in OPTIONAL MATCH clauses for optional metadata relationships. When a Neo4j instance (especially clustered) has never created these relationship types, it emits 01N51 warnings on every query execution.

## Affected files:

backend/infrahub/core/query/relationship.py — line 745: RelationshipGetPeerQuery._add_node_property_query() uses OPTIONAL MATCH (rl)-[r:HAS_OWNER]-(owner) and OPTIONAL MATCH (rl)-[r:HAS_SOURCE]-(source)
backend/infrahub/core/query/relationship.py — line 645: RelationshipDeleteQuery.query_init() uses OPTIONAL MATCH (rl)-[edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE]->(peer)
backend/infrahub/core/query/node.py — line 733: NodeListGetAttributeQuery._add_source_to_query() uses OPTIONAL MATCH (a)-[rel_source:HAS_SOURCE]-(source)
backend/infrahub/core/query/node.py — line 751: NodeListGetAttributeQuery._add_owner_to_query() uses OPTIONAL MATCH (a)-[rel_owner:HAS_OWNER]-(owner)
backend/infrahub/core/query/node.py — line 862: Profile source check uses OPTIONAL MATCH (a)-[r:HAS_SOURCE]->(:CoreProfile)
Explanation: Neo4j tracks relationship types in its schema catalog. When a Cypher query explicitly references a relationship type in a MATCH or OPTIONAL MATCH pattern (e.g., [r:HAS_OWNER]), and that type has never been created in the database, Neo4j emits a 01N51 warning: "relationship type does not exist". In standalone mode, this is typically logged silently, but in clustered deployments the notification surfaces more prominently. Since HAS_OWNER and HAS_SOURCE are optional metadata — many deployments never set owner/source on any object — these warnings fire on every read query that requests metadata, creating significant log noise.

The fix is to replace explicit relationship type patterns in OPTIONAL MATCH clauses with generic patterns that use type() filters in WHERE clauses. For example:

-- Before (triggers warning):
OPTIONAL MATCH (rl)-[r:HAS_OWNER]-(owner)
WHERE branch_filter

-- After (no warning):
OPTIONAL MATCH (rl)-[r]-(owner)
WHERE type(r) = "HAS_OWNER" AND branch_filter
For the multi-type pattern in RelationshipDeleteQuery:

-- Before:
OPTIONAL MATCH (rl)-[edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE]->(peer)

-- After:
OPTIONAL MATCH (rl)-[edge]->(peer)
WHERE type(edge) IN ["IS_PROTECTED", "HAS_OWNER", "HAS_SOURCE"] AND ...
Replication test
Test file added: backend/tests/component/core/test_neo4j_notification_suppression.py

Why it reproduces the bug: The tests instantiate each affected query class with metadata options enabled, extract the generated Cypher text, and assert that no OPTIONAL MATCH clause contains explicit HAS_OWNER or HAS_SOURCE relationship type patterns. All 4 tests fail on the current code, confirming the problematic patterns are present.

## Notes for the fix agent
The fix must only change OPTIONAL MATCH patterns. MATCH (non-optional) patterns like the one in NumberPoolGetAllocated (line 217 of resource_manager.py) use HAS_SOURCE as a required relationship and should NOT be changed — the warning there is correct (it means data is missing).
IS_PROTECTED in RelationshipGetPeerQuery._add_is_protected_query() (line 732) uses a non-optional MATCH (not OPTIONAL MATCH), so it won't trigger the warning for missing types — it simply returns no rows. However, the RelationshipDeleteQuery line 645 combines it with HAS_OWNER/HAS_SOURCE in an OPTIONAL MATCH, so that pattern does need updating.
The profile source check at node.py:862 uses OPTIONAL MATCH (a)-[r:HAS_SOURCE]->(:CoreProfile) — the type(r) filter approach must also include the :CoreProfile node label constraint.
When converting multi-type patterns like [edge:IS_PROTECTED|HAS_OWNER|HAS_SOURCE] to type(edge) IN [...], ensure the WHERE clause properly combines with the existing branch filter using AND.
The DOTALL regex flag in the test means it matches across newlines — the regex is designed to catch patterns spanning multiple lines of the generated query.

PR #8686, opened 2026-03-24
GraphQL fragment inlining

ogenstad:
# Why

Allow for imports using GraphQL fragments as defined by a `.infrahub.yml` file.

## What changed

* Small update to import logic in the git integrator class, the bulk of the changes have been made within the SDK repo
* Update the SDK commit to the latest `infrahub-develop` commit (might remove this if the SDK commit is updated elsewhere)
* New tests
* Updated specifications and plan
* Update reference documentation for .infrahub.yml files

## Other notes

* One of the changes to the spec is that we raise an error for any query that has a fragment missing, the reason for this was to ensure that we mark the import as having failed for the repository as it's the easiest way for users to note that something didn't go as planned.

## Summary by CodeRabbit

* **New Features**
  * Repositories can declare reusable GraphQL fragment files in config and have fragment spreads automatically inlined into stored queries during sync and when running queries locally.
  * Resync reflects fragment file updates; repos remain isolated (same fragment name can differ per repo).

* **Bug Fixes / Reliability**
  * Improved error handling: missing/duplicate/circular fragment issues are detected and reported, causing import failures when encountered.

* **Tests**
  * Added component tests covering inlining, transitive dependencies, error cases, resync, and repo isolation.


backend/tests/component/git/test_graphql_query_import.py
```diff
@@ -0,0 +1,276 @@
...
+        id=UUIDT.new(),
+        name="fragment_repo",
+        location=str(upstream_path),
+        client=InfrahubClient(config=Config(requester=dummy_async_request)),
+    )
+    _stub_sdk_client(repo)
+    return repo
+
+
+async def test_us1_two_fragment_files_stored_query_contains_both_definitions(
```
gmazoyer: I wonder if having a user-story reference in the name of tests is a good thing. If this reference would be to disappear one day, we would probably be surprise by the test name.

backend/tests/component/git/test_graphql_query_import.py
```diff
@@ -0,0 +1,276 @@
...
+
+def _stub_sdk_client(repo: InfrahubRepository) -> None:
+    """Stub the SDK client so sdk.filters() returns an empty list.
+
+    sdk.filters() calls schema.get() internally, which requires a live server.
+    An AsyncMock bypasses this.
+    """
+    mock_client = AsyncMock()
+    mock_client.filters = AsyncMock(return_value=[])
+    repo.client = mock_client
```
gmazoyer, line 51: Everything is fake here. I'm not sure it really matters given what we test, but just wanted to point it out.

backend/tests/component/git/test_graphql_query_import.py
```diff
@@ -0,0 +1,276 @@
...
+        client=InfrahubClient(config=Config(requester=dummy_async_request)),
+    )
+    _stub_sdk_client(repo)
+    return repo
+
+
+async def test_us1_two_fragment_files_stored_query_contains_both_definitions(
+    fragment_repo: InfrahubRepository,
+) -> None:
+    """US1: A query using fragments from two files must have both definitions inlined."""
```
gmazoyer: Same goes with the comments

backend/tests/component/git/test_graphql_query_import.py
ogenstad: Yeah good point, I remove any mention of user stories.

backend/tests/component/git/test_graphql_query_import.py
ogenstad: Yeah I know, basically it prevents us from having to setup everything within the server as we're not trying to test the interaction with the server just how the import methods behave. Having said that I think we should have a further discussion around how we can avoid these type of things in the future and in this case use a client where we've populated the default schema and then uses a custom transport.

gmazoyer (approved):

ogenstad:
Merging as it's only the unrelated codspeed report that's failing.

PR #8687, opened 2026-03-24
Add login/logut events spec

solababs:
# Why

6This PR emits `infrahub.account.logged_in` and `infrahub.account.logged_out` events on successful authentication and explicit logout, covering password, OAuth2, and OIDC flows. The events include account metadata (name, type, groups, roles), session ID, auth method, client IP, and user-agent, and are queryable via the activity event feed.

Non-goals: this does not add failed login events, token refresh events, or API key auth events.

## What changed

- Successful logins (password, OAuth2, OIDC) now emit `infrahub.account.logged_in` on the event bus
- Explicit logouts now emit `infrahub.account.logged_out` on the event bus
- Login events include: account_name, account_type, auth_method, session_id, groups, roles, sso_provider, client_ip, user_agent and timestamp.
- Logout events include: account_name, session_id, logout_type, client_ip, user_agent and timestamp.

- `authenticate_with_password` and `signin_sso_account` now return `AuthResult` instead of `UserToken` — `AuthResult` bundles the token with account metadata needed for event construction, avoiding a second DB lookup in the API layer
- `invalidate_refresh_token` now returns `bool` so logout only emits an event when a token was actually deleted (not on no-op logouts)
- `fetch_account_groups_and_roles` is a new shared helper that resolves group and role names for a given account at login time
- `backend/infrahub/api/event_builder.py` centralises event construction for login/logout, keeping the auth endpoints thin
- Event emission is wrapped in `try/except` — a failure to emit never blocks the auth response
- Administrators (accounts that have the `MANAGE_ACCOUNTS` permission) can retrieve all authentication events

## How to test

```bash
# Component tests for login/logout event emission
uv run pytest backend/tests/component/api/test_auth_events.py -v

# Full auth unit tests
uv run invoke backend.test-unit -- -k auth
```

Login and verify the event appears in the activity feed via the GraphQL event query.

## Impact & rollout

- **Backward compatibility:** No breaking changes.
- **Performance:** Two extra DB queries per login (groups + roles). Should be negligible given login frequency.
- **Deployment notes:** Safe to deploy independently

## Checklist

- [x] Tests added/updated


dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,122 @@
...
+**Acceptance Scenarios**:
+
+1. **Given** a user authenticates successfully via password, **When** the activity event feed is queried for that user, **Then** a login event appears with the account identifier, authentication method, session identifier, and timestamp.
+2. **Given** a user explicitly logs out, **When** the activity event feed is queried for that user, **Then** a logout event appears with the account identifier, session identifier, and timestamp.
+3. **Given** a 30-day time window, **When** an administrator queries all authentication events for a specific account, **Then** all login and logout events within that window are returned in chronological order.
+4. **Given** login and logout events exist, **When** the event feed is filtered by event type, **Then** only authentication events of the specified type are returned.
+
+---
+
+### User Story 2 - Failed Login Detection (Priority: P2)
```
ogenstad: I think we can skip this as a separate event for now. To me a failed login attempt would be a different kind of log entry and not an "Infrahub event". Later we'll want the same type of log messages for queries and mutations where permission is denied.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- **Login Event**: Records a single authentication attempt — outcome (success/failure), account identifier (if known), authentication method, session identifier (if authenticated), failure reason (if failed), timestamp.
+- **Logout Event**: Records a session termination — account identifier, session identifier, timestamp, and logout type (`user_initiated` or `admin_forced`).
+- **Session Reference**: A unique identifier that links a login event to its corresponding logout event, enabling session duration analysis.
+
+## Success Criteria *(mandatory)*
+
+### Measurable Outcomes
+
+- **SC-001**: Login and logout events appear in the activity event feed within 5 seconds of the authentication action under normal load.
+- **SC-002**: Administrators can retrieve all authentication events for a specific user across any 30-day window without missing entries.
```
ogenstad: Let's skip the timelines here. The assumption would be that these events act as any other event published within Infrahub.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- **SC-003**: Authentication events are queryable alongside all other activity events using the existing query interface with no additional configuration required.
```
ogenstad: One thing here is that we don't want regular users to be allowed to query for these account events, only admin users so we'll need to take this into consideration.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- **SC-004**: Existing webhook integrations can be configured to trigger on login, failed login, or logout events without new development work on the integration layer.
```
ogenstad: We might want to review this part a bit as well as we don't want non admin users to be able to create webhooks or update existing ones that use these type of events specifically. This could be handled as a follow up item..

dev/specs/infp-474-login-logout-events/data-model.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+
+## `EventType` Constants
+
+Added to `backend/infrahub/core/constants/__init__.py`:
+
+```python
+ACCOUNT_LOGGED_IN = f"{EVENT_NAMESPACE}.account.logged_in"
+ACCOUNT_LOGGED_OUT = f"{EVENT_NAMESPACE}.account.logged_out"
```
ogenstad: These shouldn't be constants, instead they would be defined as events under "/backend/infrahub/events"

dev/specs/infp-474-login-logout-events/plan.md
```diff
@@ -0,0 +1,83 @@
...
+**Input**: Feature specification from `/specs/infp-474-login-logout-events/spec.md`
+
+## Summary
+
+Adds two new Prefect-backed activity events (`infrahub.account.logged_in`, `infrahub.account.logged_out`) to the existing Infrahub event system. Events are emitted from the three authentication endpoints (password, OAuth2, OIDC) and the logout endpoint. A new `AuthResult` dataclass carries rich account metadata (name, type, groups, roles, session ID) from the auth functions to the endpoints, where it is combined with HTTP context (client IP, user agent) before emission. All events are fire-and-forget — failures are logged but never block authentication.
+
+## Technical Context
+
+**Language/Version**: Python 3.12, FastAPI 0.121.1
+**Primary Dependencies**: Prefect Events (existing), RabbitMQ via `InfrahubEventService` (existing), Pydantic 2.10
```
ogenstad: Just for clarity I don't think RabbitMQ would be involved in this.

dev/specs/infp-474-login-logout-events/plan.md
```diff
@@ -0,0 +1,83 @@
...
+
+```text
+backend/
+├── infrahub/
+│   ├── auth.py                          # + AuthResult, AuthenticationError, _fetch_account_groups_and_roles
+│   ├── events/
+│   │   ├── __init__.py                  # + export 2 new event classes
+│   │   └── auth_action.py               # NEW: AccountLoggedInEvent, AccountLoggedOutEvent
+│   ├── api/
+│   │   ├── auth.py                      # + event emission (login success, login failure, logout)
```
ogenstad: We can scratch the "login failure" from here for now.

dev/specs/infp-474-login-logout-events/quickstart.md
```diff
@@ -0,0 +1,96 @@
...
+**Branch**: `infp-474-login-logout-events` | **Date**: 2026-03-24
+
+## What Was Built
+
+Three new activity events are emitted whenever users authenticate or log out:
+
+| Event | Trigger |
+|-------|---------|
+| `infrahub.account.logged_in` | Successful login via password, OAuth2, or OIDC |
+| `infrahub.account.login_failed` | Failed login attempt (wrong password or unknown user) |
```
ogenstad: Remove

dev/specs/infp-474-login-logout-events/quickstart.md
```diff
@@ -0,0 +1,96 @@
...
+          client_ip
+          user_agent
+        }
+      }
+    }
+  }
+}
+```
+
+### 3. Trigger a failed login
```
ogenstad: We can remove this section.

dev/specs/infp-474-login-logout-events/research.md
```diff
@@ -0,0 +1,98 @@
...
+
+**Alternatives considered**:
+- Fetch groups/roles directly from account node relationships: requires loading account node with all peers, more expensive
+- Store groups/roles in JWT: already done for authorization, but the source values must be fetched at login time anyway
+
+---
+
+### Q5: What `EventMeta` context to use for auth events?
+
+**Decision**: Construct a minimal `InfrahubContext` with default branch + `AccountSession(auth_type=AuthType.NONE, authenticated=False)`.
```
ogenstad: This sounds wrong, i.e. we should be able to populate the context with the real data for account session.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+1. **Given** a user authenticates successfully via password, **When** the activity event feed is queried for that user, **Then** a login event appears with the account identifier, authentication method, session identifier, and timestamp.
+2. **Given** a user explicitly logs out, **When** the activity event feed is queried for that user, **Then** a logout event appears with the account identifier, session identifier, and timestamp.
+3. **Given** authentication events exist, **When** an administrator queries all authentication events for a specific account, **Then** all login and logout events within the platform's standard retention window are returned in chronological order.
+4. **Given** login and logout events exist, **When** the event feed is filtered by event type, **Then** only authentication events of the specified type are returned.
+
+---
+
+### Edge Cases
+
+- What happens when a user's session is administratively invalidated (forced logout)? A logout event is emitted attributed to the account being logged out, distinguishing it from a user-initiated logout.
```
ogenstad: We don't currently support a way to administratively invalidate a session. I think this would be a separate feature to introduce and not related to this one where we're just creating the events.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+- What happens when a user logs in and immediately logs out within the same second? Both events are recorded with accurate timestamps and the same session identifier.
+- How does the system handle a logout request for an already-expired or non-existent session? No logout event is emitted if no valid session exists to invalidate.
```
ogenstad, line 30: This would probably be within `invalidate_refresh_token()` if a refresh token doesn't exist we know that the user would have logged out before.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+- What happens if the same account logs in from multiple concurrent sessions? Each session generates its own independent login event with a unique session identifier.
+
+## Requirements *(mandatory)*
+
+### Functional Requirements
+
+- **FR-001**: System MUST emit a login event for every successful interactive authentication via password, OAuth2, or OIDC.
+- **FR-002**: Successful login events MUST include: account identifier, authentication method (password/OAuth2/OIDC), session identifier, and timestamp.
+- **FR-003**: System MUST emit a logout event when a user explicitly initiates a logout.
+- **FR-004**: Logout events MUST include: account identifier, session identifier, logout type (`user_initiated` or `admin_forced`), and timestamp.
```
ogenstad: We don't currently have a logout type. I think this might come later.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+- **FR-005**: Authentication events (`infrahub.account.*`) MUST be queryable via the existing activity event interface, filterable by account identifier, event type, and time range — and accessible only to users with admin role.
+- **FR-006**: API key authentication (per-request, non-interactive) MUST NOT generate login or logout events.
+- **FR-007**: Automatic session expiry (session times out without explicit user action) MUST NOT generate a logout event.
+- **FR-008**: System MUST emit a logout event with `logout_type="admin_forced"` when an administrator invalidates a user's session via the API or GraphQL interface.
```
ogenstad: We don't have this option today and should not introduce it as part of this feature.

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Key Entities *(include if feature involves data)*
+
+- **Login Event**: Records a successful authentication — account identifier, authentication method, session identifier, timestamp, and optional group/role context.
+- **Logout Event**: Records a session termination — account identifier, session identifier, timestamp, and logout type (`user_initiated` or `admin_forced`).
```
ogenstad: Same comment with the logout type

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,79 @@
...
+- **SC-001**: Login and logout events appear in the activity event feed within 5 seconds of the authentication action under normal load.
+- **SC-002**: Administrators can retrieve all authentication events for a specific user within the platform's standard event retention window without missing entries.
+- **SC-003**: Authentication events (`infrahub.account.*`) are accessible only to users with admin role via the activity event interface — non-admin users cannot query these events.
+- **SC-004**: Zero successful authentication events are silently dropped under normal operating conditions.
+
+## Clarifications
+
+### Session 2026-03-24
+
+- Q: Should the system emit a logout event when an administrator invalidates a user's session? → A: Yes — emit `AccountLoggedOutEvent` with `logout_type="admin_forced"` when an admin invalidates a session.
```
ogenstad: No we don't support this

dev/specs/infp-474-login-logout-events/data-model.md
solababs: @ogenstad This is actually needed for the test here backend/tests/unit/event/test_event.py to pass

ogenstad: Instead of constants we should use it in this way:

https://github.com/opsmill/infrahub/blob/infrahub-v1.8.2/backend/infrahub/computed_attribute/triggers.py#L12

As an example:

```python
from infrahub.events.repository_action import CommitUpdatedEvent

print(CommitUpdatedEvent.event_name)
```

ogenstad: Ah ok I see we have some duplicated parts specifically for the GraphQL query API for the events. Let's leave that as it is for now then and look at refactoring it later to have one location where everything gets defined..

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,76 @@
...
+- How does the system handle a logout request for an already-expired or non-existent session? No logout event is emitted if no valid session exists to invalidate.
+- What happens if event emission fails after a successful login? The authentication itself succeeds; event emission failure does not prevent login. The event may be absent from the audit trail.
+- What happens if the same account logs in from multiple concurrent sessions? Each session generates its own independent login event with a unique session identifier.
+
+## Requirements *(mandatory)*
+
+### Functional Requirements
+
+- **FR-001**: System MUST emit a login event for every successful interactive authentication via password, OAuth2, or OIDC.
+- **FR-002**: Successful login events MUST include: account identifier, authentication method (password/OAuth2/OIDC), session identifier, and timestamp.
```
wvandeun: There's a bit of a disconnect here. The data-model, for example, also has other data points. We should add them here and make it conssitent

dev/specs/infp-474-login-logout-events/spec.md
```diff
@@ -0,0 +1,76 @@
...
+- **Login Event**: Records a successful authentication — account identifier, authentication method, session identifier, timestamp, and optional group/role context.
+- **Logout Event**: Records a session termination — account identifier, session identifier, timestamp, and logout type
+- **Session Reference**: A unique identifier that links a login event to its corresponding logout event, enabling session duration analysis.
+
+## Success Criteria *(mandatory)*
+
+### Measurable Outcomes
+
+- **SC-001**: Login and logout events appear in the activity event feed within 5 seconds of the authentication action under normal load.
+- **SC-002**: Administrators can retrieve all authentication events for a specific user within the platform's standard event retention window without missing entries.
```
wvandeun: Just for clarity we should specify what an Admin user is.. A user with Super Adminitrator role?

wvandeun (changes requested):

dev/specs/infp-474-login-logout-events/spec.md
solababs: This has been updated to add the `MANAGE_ACCOUNTS` permission https://github.com/opsmill/infrahub/pull/8687/changes#diff-77feec211e96203ed4f7427b9ff89553286be049ea0abf43af98a0f7ee9d5e55R57

wvandeun (approved):

ogenstad (approved):

ogenstad (approved):

frontend/app/tests/e2e/activities/global-activities-filters.spec.ts
```diff
@@ -59,6 +59,7 @@ test.describe("Global Activities - List view and filter usage", () => {
     await page.getByRole("button", { name: "Account" }).click();
     await page.getByRole("option", { name: "Jack Bauer" }).click();
     await page.getByRole("button", { name: "Apply" }).click();
-    await expect(page.getByText("No activities found")).toBeVisible();
+    await expect(page.getByRole("button", { name: /Account.*Jack Bauer/ })).toBeVisible();
+    await expect(page.getByText("Jack Bauer").nth(1)).toBeVisible();
```
bilalabbad: These new assertions test nothing, please replace them with somehting more meaningful like:
```ts
  await expect(page.getByText("infrahub.account.logged_in").first()).toBeVisible();
```

bilalabbad (approved):

PR #8688, opened 2026-03-24
IFC-2274: Add new events and update auth results

solababs:
# Why

Infrahub had no visibility into authentication activity. Login and logout events were not emitted to the event system, making it impossible to audit or react to authentication events via the activity feed or external integrations.

This PR adds `infrahub.account.logged_in` and `infrahub.account.logged_out` events emitted on successful authentication and explicit logout across all auth methods (password, OAuth2, OIDC).

**Non-goals:** Token refresh events, failed login attempts, and session expiry events are out of scope for this phase.

## What changed

- **New event types:** `AccountLoggedInEvent` and `AccountLoggedOutEvent` in `backend/infrahub/events/account_action.py`, carrying account ID, name, type, auth method, session ID, groups, roles, SSO provider, client IP, and user agent
- **`EventType` constants:** Added `ACCOUNT_LOGGED_IN` and `ACCOUNT_LOGGED_OUT` to `backend/infrahub/core/constants/__init__.py`
- **`AuthResult` model:** Authentication functions (`authenticate_with_password`, `signin_sso_account`) now return a richer `AuthResult` instead of `UserToken` directly, bundling the token with account metadata (ID, name, type, groups, roles, session ID) needed to populate the event payload
- **`fetch_account_groups_and_roles`:** New helper in `backend/infrahub/auth.py` that fetches group and role names for an account at login time
- **API layer updated:** `api/auth.py`, `api/oauth2.py`, `api/oidc.py` updated to unwrap `auth_result.token` from the new return type — no change to API contracts or cookie behavior

No schema changes. No API contract changes. No frontend changes.

### Suggested review order
1. `backend/infrahub/events/account_action.py` — new event models
2. `backend/infrahub/auth.py` — `AuthResult`, `fetch_account_groups_and_roles`, updated `authenticate_with_password` and `signin_sso_account`
3. `backend/infrahub/api/auth.py`, `oauth2.py`, `oidc.py` — thin adapter changes
4. `backend/tests/component/core/test_account.py` — new tests for `fetch_account_groups_and_roles`

## How to review

Focus on `auth.py` — specifically whether `fetch_account_groups_and_roles` silently swallowing exceptions on group/role fetch is the right call (event emission should not break login).

The API files (`auth.py`, `oauth2.py`, `oidc.py`) are mechanical — just unwrapping `auth_result.token` instead of returning the raw token.

## How to test

```bash
uv run pytest backend/tests/component/core/test_account.py -v 
uv run invoke backend.test-unit -k test_account
uv run invoke backend.test-integration -k test_account
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. The `UserToken` response shape returned by all auth endpoints is unchanged.
- **Performance:** `fetch_account_groups_and_roles` adds two DB queries per login. Groups and roles queries are branch-agnostic.
- **Deployment notes:** Safe to deploy independently

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/+203c8d78.added.md`)


backend/infrahub/events/account_action.py
```diff
@@ -0,0 +1,83 @@
...
+class LogoutType(InfrahubStringEnum):
+    USER_INITIATED = "user_initiated"
+    ADMIN_REVOKED = "admin_revoked"
+    TOKEN_EXPIRED = "token_expired"
+
+
+class AccountLoggedInEvent(InfrahubEvent):
+    """Emitted when a user successfully authenticates."""
+
+    event_name: ClassVar[str] = f"{EVENT_NAMESPACE}.account.logged_in"
```
ajtmccarty, line 29: seems like this should be linked to the `EventType.ACCOUNT_LOGGED_IN` enum, but not completely sure

backend/infrahub/events/account_action.py
```diff
@@ -0,0 +1,83 @@
...
+
+    account_id: str = Field(..., description="UUID of the account")
+    account_name: str = Field(..., description="Username of the account")
+    account_type: str = Field(..., description="USER or SCRIPT")
```
ajtmccarty: should this be an enum?

backend/infrahub/auth.py
```diff
@@ -62,6 +76,26 @@ class SSOStateCache(BaseModel):
...
+        groups = await NodeManager.query(
+            schema=CoreAccountGroup,
+            db=db,
+            filters={"members__ids": [account_id]},
+        )
+        group_names = [g.name.value for g in groups if hasattr(g, "name")]
+        role_names: list[str] = []
+        for group in groups:
+            roles = await group.roles.get_peers(db=db, branch_agnostic=True, peer_type=CoreAccountRole)
+            role_names.extend(r.name.value for r in roles.values() if hasattr(r, "name"))
```
ajtmccarty: I would think that passing `CoreAccountRole` in above would mean that you don't need to do the `if hasattr(r, "name")` check here

backend/infrahub/auth.py
```diff
@@ -105,7 +139,18 @@ async def authenticate_with_password(
     access_token = generate_access_token(account_id=account.id, session_id=session_id)
     refresh_token = generate_refresh_token(account_id=account.id, session_id=session_id, expiration=refresh_expires)
 
-    return models.UserToken(access_token=access_token, refresh_token=refresh_token)
+    groups, roles = await fetch_account_groups_and_roles(db=db, account_id=account.id)
+    account_type = account.account_type.value if hasattr(account, "account_type") else "User"
```
ajtmccarty: same thing here. looks like `CoreGenericAccount` should always have an `account_type`, so I would think that the `hasattr` conditional would not be necessary

ajtmccarty (approved):

backend/infrahub/events/account_action.py
```diff
@@ -0,0 +1,83 @@
...
+    OIDC = "oidc"
+    API_KEY = "api_key"
+
+
+class SSOProvider(InfrahubStringEnum):
+    OAUTH2 = "oauth2"
+    OIDC = "oidc"
+
+
+class LogoutType(InfrahubStringEnum):
```
ogenstad: As I mentioned elsewhere I think we should remove the logout type until we actually have support for something else within Infrahub.

backend/infrahub/events/account_action.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .constants import EVENT_NAMESPACE
+from .models import InfrahubEvent
+
+
+class AuthMethod(InfrahubStringEnum):
+    PASSWORD = "password"
+    OAUTH2 = "oauth2"
+    OIDC = "oidc"
+    API_KEY = "api_key"
```
ogenstad: We can remove the API_KEY from here as we won't send any events based on API key alone.

backend/infrahub/auth.py
```diff
@@ -62,6 +76,26 @@ class SSOStateCache(BaseModel):
...
+            schema=CoreAccountGroup,
+            db=db,
+            filters={"members__ids": [account_id]},
+        )
+        group_names = [g.name.value for g in groups if hasattr(g, "name")]
+        role_names: list[str] = []
+        for group in groups:
+            roles = await group.roles.get_peers(db=db, branch_agnostic=True, peer_type=CoreAccountRole)
+            role_names.extend(r.name.value for r in roles.values() if hasattr(r, "name"))
+    except Exception:
```
ogenstad: I feel like this bare exception could potentially hide any errors here. I'd expect that that unless we introduce some code issues the main problem here should be database errors. Otherwise if the groups or roles aren't found we should just have empty lists coming back regardless. While it would be good to have better internal error handling I think we should look at that at a global level and for now skip this try/expect check.

backend/infrahub/auth.py
```diff
@@ -34,6 +34,20 @@
...
+
+    model_config = {"frozen": True}
+
+    token: models.UserToken
+    account_id: str
+    account_name: str
+    account_type: str
+    session_id: uuid.UUID
+    groups: list[str]
+    roles: list[str]
```
ogenstad, line 48: The initial specs in JPD lists groups and roles as having IDs as well as names.

ogenstad (commented):
LGTM so far with some minor comments. Regarding the event classes this could come later for the query API, but we'll also want some related_resources methods so that we can capture all of the fields for display in the API responses.

backend/infrahub/auth.py
ogenstad: Actually it's a bit unclear if it's IDs and names vs IDs or names...

solababs: @wvandeun is it ok to use the names as I have used the names here?

wvandeun: Yes, the names of the groups and roles make the most sense.

wvandeun (approved):

ogenstad (approved):

PR #8689, opened 2026-03-24
Fix m056 to use targeted writes

gmazoyer:
# Why

When upgrading from 1.7 to 1.8, migration 056 updates internal schema constraints on all branches. The problem: it used `update_node_in_db()` which dumps the **entire** schema node via `model_dump()` and writes every attribute value to each branch's timeline — not just the 3 properties it actually changes.

Then `update_core_schema()` comes along and updates those same attributes (like `namespace.parameters`) on the default branch with new 1.8 values. Rebase sees both sides wrote different values → conflict. Every existing branch is stuck.

This PR fixes migration 056 to be surgical, and adds a cleanup migration (068) to repair databases that already ran the old version.

### 1. Targeted writes in migration 056

Rewrote `update_schema_on_branch()` to bypass `update_node_in_db()` and instead:
- Set `human_friendly_id` and `uniqueness_constraints` directly on the DB node, save with `fields=[...]` so only those two get persisted
- Load the `name` SchemaAttribute by ID, set `unique=False`, save with `fields=["unique"]`

No other SchemaAttribute nodes (namespace, description, label, etc.) are loaded or saved. No stray `HAS_VALUE` edges end up on branch timelines.

### 2. Cleanup migration 068 for already-upgraded instances

New `Migration068` runs a Cypher query per open branch that:
- Finds `SchemaAttribute.parameters` `HAS_VALUE` edges on the branch with known all-null JSON shapes (`{"id":null,"state":"present","regex":null,...}`)
- Verifies the default branch value at the branch's fork point was `NULL` or also all-null
- Expires the spurious branch edge and creates a new one pointing back to the original default branch value

Only touches SchemaAttributes under SchemaNode and SchemaGeneric (the only schemas migration 056 targeted). Skips merged/terminal branches. Idempotent.

### Suggested review order
1. `m056_update_schema_node_generic_constraints.py`
2. `m068_cleanup_branch_schema_parameters.py`
3. `test_068_cleanup_branch_schema_parameters.py`

## How to review

**m056**: The key mechanism is `Node.save(fields=[...])` which makes `Node._update()` skip any attribute not in the list.

**m068**: The Cypher query is the core of this migration. It walks SchemaNode/SchemaGeneric → SchemaAttribute → parameters Attribute, matches branch-specific `HAS_VALUE` edges with all-null values, looks up the original default branch value at the fork point, and swaps the edge target.

## How to test

```bash
uv run pytest backend/tests/component/core/migrations/graph/test_056_update_schema_node_generic_constraints.py -x -v
uv run pytest backend/tests/component/core/migrations/graph/test_068_cleanup_branch_schema_parameters.py -x -v
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. Already-upgraded instances that hit rebase conflicts will be cleaned up by migration 068 on next upgrade.

## Checklist

- [x] Tests added/updated


ajtmccarty:
pushed commit with a new small Migration068 to clean up possible unnecessary changes made by the old version of Migration056

the problem is that the `parameters` attribute of a `SchemaAttribute` could have a value of `"NULL"` if the Infrahub instance predates the introduction of the `parameters` property to `SchemaAttribute`s. in this case, when the old version of Migration056 ran, it would update the `parameters` value from `"NULL"` to a JSON blob (`{id: null, state: "present", ...}`) where all the values of the JSON blob are `null`. This is because the `"NULL"` value is treated as a `None` and when the parameters are serialized during the save operation, then are serialized as JSON of the default parameters value. this will be one of three different JSON blobs depending on the `kind` of the Attribute
 - `{"id":null,"state":"present"}`
 - `{"id":null,"state":"present","regex":null}`
 - `{"id":null,"state":"present","regex":null,"min_length":null,"max_length":null}`

this would not really be an issue, if the `update-core-schema` portion of the upgrade task did not also update the `parameters` value on the default branch, which then leads to an unresolvable conflict that prevents any branch from being updated.

the solution is to find all the `SchemaAttribute`s of the `SchemaNode` or `SchemaGeneric` schemas and, if any have had their `parameters` attribute updated on user-defined branches such that they are now the JSON blob version of `null`, then reset the value to whatever it used to be on the default branch to avoid a conflict.

I've tested this on some customer backups that I have and it works. I've also tested upgrading the sandbox backup using the old version of Migration056 that causes the unresolvable conflict state and verified that the new migration fixes that.

ajtmccarty (approved):

PR #8690, opened 2026-03-24
IFC-2274: Add events to graphql query

solababs:
# Why

This PR exposes account events through the existing `InfrahubEvent` GraphQL query with typed response shapes, and gates access behind the `MANAGE_ACCOUNTS` global permission to prevent non-admin users from enumerating authentication activity.

## What changed

- **GraphQL types:** `AccountLoggedInEventType` and `AccountLoggedOutEventType` added to `backend/infrahub/graphql/types/event.py` and registered in `EVENT_TYPES`, giving the `InfrahubEvent` query typed, structured fields for account events instead of a raw payload
- **Permission guard:** Querying any `infrahub.account.*` event type now requires the `MANAGE_ACCOUNTS` global permission — enforced in the `Events.resolve` method in `backend/infrahub/graphql/queries/event.py`
- **GraphQL schema updated:** `schema/schema.graphql` regenerated to include `AccountLoggedInEventType` and `AccountLoggedOutEventType` with all fields

No changes to event emission, auth flows, API contracts, or database schema.

### Suggested review order
1. `backend/infrahub/graphql/types/event.py` — new GraphQL types and `EVENT_TYPES` registration
2. `backend/infrahub/graphql/queries/event.py` — permission guard logic
3. `backend/tests/component/graphql/queries/test_event_account_permissions.py` — permission test cases
4. `schema/schema.graphql` — generated, review for correctness only

## How to review

The permission check in `queries/event.py` fires when **any** of the requested `event_type` values starts with `infrahub.account.` — verify this is the right granularity (per-type vs. any account type in the list).

The guard uses `MANAGE_ACCOUNTS` global permission. Confirm this is the right permission level — alternatives would be a dedicated `VIEW_ACCOUNT_EVENTS` permission (doesn't exist yet) or `SUPER_ADMIN` only.

## How to test

```bash
uv run invoke backend.test-unit -k test_event_account_permissions
```

Manual verification:
1. Start Infrahub locally and log in as a non-admin account
2. Query `InfrahubEvent` with `event_type: ["infrahub.account.logged_in"]` — expect a permission error
3. Log in as an admin account and repeat — expect an empty result set (no error)
4. Query with a non-account event type as a non-admin — expect no permission error

## Impact & rollout

- **Backward compatibility:** No breaking changes. Existing `InfrahubEvent` queries without account event types are unaffected.
- **Performance:** No additional DB queries; permission check is in-memory against already-loaded permissions.
- **Deployment notes:** Safe to deploy. Requires phase 1 to be deployed first (event types must exist in the event system).

## Checklist

- [x] Tests added/updated


backend/infrahub/graphql/queries/event.py
```diff
@@ -48,6 +53,27 @@ async def resolve(
             # Prefect restricts this to 50
             raise ValidationError(input_value="The parameter 'limit' can't be above 50")
 
+        exclude_prefixes = None
```
ogenstad: We should add a type here with:

```python
exclude_prefixes: list[str] | None = None
```

ogenstad (approved):
Added one minor comment, looks good overall. We'll need to add some more glue later to be able to query for the actual fields within the GraphQL query for these events.

PR #8691, opened 2026-03-24
IFC-2274: Emit events at login and logout

solababs:
# Why

This PR wires the emission: `AccountLoggedInEvent` is sent on successful password, OAuth2, and OIDC authentication; `AccountLoggedOutEvent` is sent on user-initiated logout.

## What changed

- **Event emission at login:** `api/auth.py`, `api/oauth2.py`, `api/oidc.py` now call `emit_auth_event` after a successful authentication, passing `auth_method` (`password`, `oauth2`, `oidc`), `sso_provider` (for SSO paths), `client_ip`, `user_agent`, groups, and roles. Failures are caught and logged as warnings — a broken event bus cannot block login.
- **Event emission at logout:** `POST /api/auth/logout` emits `AccountLoggedOutEvent` with `logout_type=user_initiated` after invalidating the refresh token.
- **`emit_auth_event` helper:** Shared utility in `api/auth.py` that builds `EventMeta` from the request context and dispatches via `service.event.send`.

No changes to event models, GraphQL types, schema, or existing API response shapes.

### Suggested review order
1. `backend/infrahub/api/auth.py` — `emit_auth_event` helper, login/logout emission
2. `backend/infrahub/api/oauth2.py` and `oidc.py` — SSO emission wiring (uses the same helper)
3. `backend/tests/component/api/test_auth_events.py` — end-to-end emission tests
4. `backend/tests/unit/event/test_account_action.py` — unit tests for event model construction

## How to review

The `try/except` wrapping every `emit_auth_event` call is intentional — event emission must never break auth. Confirm the logging level (`warning`) is appropriate versus silently swallowing failures.

## How to test

```bash
uv run invoke backend.test-unit -k test_account_action
uv run invoke backend.test-integration -k test_auth_events
```

Manual end-to-end:
1. Start Infrahub locally
2. `POST /api/auth/login` with valid credentials → check activity feed for `infrahub.account.logged_in` with correct `auth_method`, `groups`, `roles`
3. `POST /api/auth/logout` with the access token → check activity feed for `infrahub.account.logged_out` with `logout_type=user_initiated`

## Impact & rollout

- **Backward compatibility:** No breaking changes. Existing login/logout endpoints return the same responses.
- **Performance:** Each login/logout adds one async event dispatch, wrapped in try/except. No DB queries beyond what already happens in the auth flow, except for the account name lookup on logout.
- **Deployment notes:** Requires phases 1 and 2 to be deployed first. Safe to deploy independently once those are in.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/+203c8d78.added.md`)


backend/infrahub/api/auth.py
```diff
@@ -2,26 +2,59 @@
...
+
     from infrahub.database import InfrahubDatabase
+    from infrahub.events.models import InfrahubEvent
+    from infrahub.services import InfrahubServices
 
+log = get_logger()
 router = APIRouter(prefix="/auth")
 
 
+async def emit_auth_event(
```
ajtmccarty: I don't think we use this pattern anywhere else and I don't think we should use it here. the dependency chain is looped back onto itself here b/c the `event_factory` has to have knowledge about how `emit_auth_event` works.

I think this should be a class that accepts a particular version of an `EventBuilder` interface and uses that to construct the `InfrahubEvent`

something like this 
```python
class EventBuilderInterface
    async def build(self, request: ..., meta: ..., auth_result: ...) -> InfrahubEvent: ...

class AccountLoginPasswordEventBuilder(EventBuilderInterface):
    async def build(self, request: ..., meta: ..., auth_result: ...) -> AccountLoggedInEvent: ...

class AuthEventEmitter:
    def __init__(self, db: InfrahubDatabase, service: InfrahubServices, event_builder: EventBuilderInterface): ...
    async def emit(self, request :..., auth_result: ..., account_session: AcountSession) -> None: ...


auth_event_emitter = AuthEventEmitter(db, request.app.state.service, AccountLoginPasswordEventBuilder())
await auth_event_emitter.emit(request, auth_result, account_session)
```

backend/infrahub/api/auth.py
```diff
@@ -55,10 +110,35 @@ async def refresh_jwt_token(
...
-    response: Response, db: InfrahubDatabase = Depends(get_db), user_session: AccountSession = Depends(get_access_token)
+    request: Request,
+    response: Response,
+    db: InfrahubDatabase = Depends(get_db),
+    user_session: AccountSession = Depends(get_access_token),
 ) -> None:
     if user_session.session_id:
         await invalidate_refresh_token(db=db, token_id=user_session.session_id)
 
+    try:
```
ogenstad: I think we should move this logic a bit and perhaps alternate the return code for `invalidate_refresh_token` so that we know if the refresh token was deleted or not. The idea being that if the refresh token doesn't exist the user would already have been logged out.

backend/infrahub/api/auth.py
ogenstad: I regarding the pattern, but for now I'd just have some code that creates the event object with all of the input that we have and then after that do.

```python
await service.event.send(event=event)
```

Then if we want to consider some event builder interface we can consider that later as a global refactoring of the events.

backend/infrahub/api/event_builder.py
ajtmccarty, line 1: please rename file to something that actually describes the content of the file. maybe "event_builder.py"

backend/infrahub/api/helper.py
```diff
@@ -0,0 +1,52 @@
...
+from infrahub.auth import AccountSession, AuthResult
+from infrahub.context import InfrahubContext
+from infrahub.core.registry import registry
+from infrahub.database import InfrahubDatabase
+from infrahub.events.account_action import AccountLoggedInEvent, AccountLoggedOutEvent, AuthMethod, SSOProvider
+from infrahub.events.models import EventMeta
+
+
+async def make_event_meta(db: InfrahubDatabase, account_session: AccountSession) -> EventMeta:
+    branch = await registry.get_branch(db=db)
```
ajtmccarty: I think it's more appropriate for the API layer to determine which branch to pass to `make_event_meta` than to use the registry inside of the function
some day we can hopefully remove the `registry`, or at least make it less of a kludge, and that will be easier if the registry stays at the top layers where it can be replaced by components using dependency injection

ajtmccarty (approved):

PR #8692, opened 2026-03-24
IFC-2382: Trigger local Jinja2 recomputation on mutation

polmichel:
# Why

Enabling local recomputation of local Jinja2 attributes when updating a node.

Closes https://opsmill.atlassian.net/browse/IFC-2382

## What changed

- **Inline local recomputation**: recomputes Jinja2 computed attributes whose dependencies were modified, as part of `Node._update()`
- **Cascading chains** to transitively resolve local chained dependencies in dependency order
- **HFID/display_label encapsulation**: Extracted shared `_merge_relationship_fields` and `_collect_extra_filters` to consolidate duplicate logic between HFID, display labels, and computed attributes
- **Graceful error handling**: Jinja2 render failures are logged but don't fail the mutation, this behavior will be discussed in another issue.

## How to review

- The core behavioral change is in `_recompute_local_jinja2()`: it queries the registry for affected targets and re-renders their templates inline. 
- The cascade logic in `get_local_targets()` handles chained dependencies. 

## How to test

```bash
uv run pytest backend/tests/unit/core/test_schema_branch_computed.py -xvs
uv run pytest backend/tests/component/computed_attribute/test_local_computation.py -xvs
```

## Checklist

- [x] Tests added/updated
- [x] Internal .md docs updated (WP02)

## Release Notes

* **New Features**
  * Enhanced computed attribute evaluation with graceful error handling; template errors no longer block updates
  * Added support for chained computed attribute dependencies, enabling derived fields to reference other computed fields

* **Improvements**
  * Computed attributes now intelligently recompute only when their direct dependencies change, improving efficiency

* **Tests**
  * Added comprehensive component tests validating computed attribute behavior and error handling

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -780,7 +780,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
     async def _recompute_hfid(
         self,
         db: InfrahubDatabase,
-        fields: list[str] | None,
+        fields: set[str] | None,
         node_changelog: NodeChangelog,
         update_at: Timestamp,
     ) -> None:
@@ -800,7 +800,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
     async def _recompute_display_label(
         self,
         db: InfrahubDatabase,
-        fields: list[str] | None,
+        fields: set[str] | None,
         node_changelog: NodeChangelog,
         update_at: Timestamp,
     ) -> None:
@@ -1040,7 +1040,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         await self._recompute_local_jinja2(
             db=db, fields=fields, node_changelog=node_changelog, update_at=update_at, user_id=user_id
         )
-        updated_fields = list(set(fields or []) | set(node_changelog.updated_fields))
+        updated_fields = set(fields or []) | set(node_changelog.updated_fields)
         # Recompute the human-friendly ID if one of its variables was updated
         await self._recompute_hfid(db=db, fields=updated_fields, node_changelog=node_changelog, update_at=update_at)
         # Recompute the display label if one of its variables was updated
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -737,13 +737,14 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
     ) -> None:
         """Recompute local Jinja2 computed attributes whose dependencies were modified.
 
-        Only targets computed attributes that live on this node
-        though their Jinja2 templates may reference relationship peers for input values.
+        Cascades through local chained dependencies: if computed attribute A on this node
+        depends on computed attribute B on the same node, and B was just recomputed, A will
+        also be recomputed. Cross-node cascading is handled by async Prefect automations.
         """
         schema_branch = db.schema.get_schema_branch(name=self.get_branch_based_on_support_type().name)
 
         targets = schema_branch.computed_attributes.get_local_jinja2_targets(
-            kind=self._schema.kind, updates=fields
+            kind=self._schema.kind, updates=fields, cascade=True
         )
         if not targets:
             return
@@ -1042,9 +1043,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         # Recompute the human-friendly ID if one of its variables was updated
         await self._recompute_hfid(db=db, fields=fields, node_changelog=node_changelog, update_at=update_at)
         # Recompute the display label if one of its variables was updated
-        await self._recompute_display_label(
-            db=db, fields=fields, node_changelog=node_changelog, update_at=update_at
-        )
+        await self._recompute_display_label(db=db, fields=fields, node_changelog=node_changelog, update_at=update_at)
 
         node_changelog.display_label = await self.get_display_label(db=db)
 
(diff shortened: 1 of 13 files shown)


backend/infrahub/core/node/__init__.py
```diff
@@ -722,6 +727,97 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
...
+            try:
+                new_value = await jinja_template.render(variables=variables)
+            except JinjaTemplateError:
+                log.warning(
+                    "Failed to recompute Jinja2 attribute",
+                    node_kind=self._schema.kind,
+                    attribute_name=target.attribute.name,
+                    exc_info=True,
+                )
+                continue
```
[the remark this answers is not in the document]
polmichel: I think this would a part of the Jira issue about warnings versus error on Jinja2 computations https://opsmill.atlassian.net/browse/IFC-2405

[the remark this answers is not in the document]
polmichel: But I can include skipping the computation in this case

backend/infrahub/core/node/__init__.py
```diff
@@ -722,6 +727,111 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
...
+                failed_attributes.add(target.attribute.name)
+                continue
+
+            attr: BaseAttribute = self.get_attribute(name=target.attribute.name)
+            if attr.value != new_value:
+                attr.value = new_value
+                updated_attribute = await attr.save(db=db, user_id=user_id, at=update_at)
+                if updated_attribute:
+                    node_changelog.add_attribute(attribute=updated_attribute)
+
```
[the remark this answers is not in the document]
polmichel: You are right about the validation step which I have not used yet. But I do not think the generation flow would be useful as the attribute is already created?

backend/infrahub/core/node/__init__.py
```diff
@@ -722,6 +727,111 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
...
+        """Recompute the human-friendly ID if one of its variables was updated."""
+        if not self._human_friendly_id:
+            return
+        if not ((fields and "human_friendly_id" in fields) or self._human_friendly_id.needs_update(fields=fields)):
+            return
+
+        await self._human_friendly_id.compute(db=db, node=self)
+        updated_attribute = await self._human_friendly_id.get_node_attribute(node=self, at=update_at).save(
+            at=update_at, db=db
+        )
```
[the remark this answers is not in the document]
polmichel: Seems to be true but unrelated to this task. I will add this comment to the Jira issue that will group all the last thoughts https://opsmill.atlassian.net/browse/IFC-2393

backend/infrahub/core/node/__init__.py
```diff
@@ -646,16 +647,47 @@ async def _process_fields_attributes(
 
         return errors
 
-    async def _process_macros(self, db: InfrahubDatabase) -> None:
-        schema_branch = db.schema.get_schema_branch(self._branch.name)
+    async def _resolve_jinja2_variables(
```
ogenstad, line 650: I know that this method is just doing some cleanup for the logic of some previous code, having said that I think it would be nice with some attention to the use of `getattr` at least for the node level attributes we should be able to use `Node.get_attribute()` instead to get back a `BaseAttribute` potentially we could also add a `.get_property()` on BaseAttribute that would do the same for the property (typically this will always be value but there's a chance that it's something else). It would be `.get_relationship()` for relationships.

This step is not required, just nice.

backend/infrahub/core/node/__init__.py
```diff
@@ -722,6 +727,122 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
...
+                log.warning(
+                    "Failed to recompute Jinja2 attribute",
+                    node_kind=self._schema.kind,
+                    attribute_name=target.attribute.name,
+                    exc_info=True,
+                )
+                failed_attributes.add(target.attribute.name)
+                continue
+
+            attr: BaseAttribute = self.get_attribute(name=target.attribute.name)
```
ogenstad: As .get_attribute returns a `BaseAttribute` we don't need to define the kind here.

backend/infrahub/core/node/__init__.py
```diff
@@ -722,6 +727,122 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
...
+        if not self._display_label:
+            return
+        if not ((fields and "display_label" in fields) or self._display_label.needs_update(fields=fields)):
+            return
+
+        await self._display_label.compute(db=db, node=self)
+        self._display_label.get_node_attribute(node=self, at=update_at).get_create_data(node_schema=self._schema)
+        updated_attribute = await self._display_label.get_node_attribute(node=self, at=update_at).save(
+            at=update_at, db=db
+        )
```
ogenstad, line 836: Just a comment here that we have the same issue as above that we don't use the user_id field when saving the display label. Would be good to include in the follow up as well. As it is now I don't think we can actually display this externally i.e. who updated a display_label or hfid, but would be nice to at least store the data if we can.

polmichel: I saved your additional comment into the user id section over there https://opsmill.atlassian.net/browse/IFC-2393.

backend/infrahub/core/node/__init__.py
polmichel: I will change that
Perhaps this could be added as a custom linting rule?

backend/infrahub/core/node/__init__.py
polmichel: I will add these considerations when I explore the code as this is the proper interface to use

backend/infrahub/core/node/__init__.py
ogenstad: I think it would depend on what it looks like. But I think any linting should be things that are integrated into `ruff`. It's not really critical if we have this in a few places either it's just that it's redundant.

ogenstad (approved):

PR #8704, opened 2026-03-26
Fix template pool on required relationships

### LINKED ISSUE #8701: bug: Cannot create an object using an object template with a pool assigned to a required relationship field
### Component

_No response_

### Infrahub version

1.8.2

### Current Behavior

When using an object template that assigns a resource pool to a **required** relationship field, object creation fails both via the UI and via `infrahubctl object load`. Validation fires before pool allocation can satisfy the mandatory field, causing the operation to be rejected.

-  **UI:** Mutation is rejected with `prefix is mandatory for LocationSite at prefix`.
- **`infrahubctl object load`:** Fails with `prefix is mandatory` **before** any mutation is dispatched — indicating the Python SDK performs pre-mutation validation that does not account for pool-backed template fields.

### Expected Behavior

Pool allocation via an object template should satisfy required field validation. Object creation should succeed both via the UI and `infrahubctl object load`.

## 1. Load the schema
```yaml
---
version: "1.0"

nodes:
  - name: Site
    namespace: Location
    label: Site
    icon: mdi:map-marker
    display_label: name__value
    generate_template: true
    human_friendly_id:
      - name__value
    attributes:
      - name: name
        kind: Text
        unique: true
    relationships:
      - name: prefix
        peer: IpamIPPrefix
        cardinality: one
        kind: Attribute
        optional: false

  - name: IPPrefix
    namespace: Ipam
    label: Prefix
    icon: mdi:ip-network
    display_label: prefix__value
    inherit_from:
      - BuiltinIPPrefix
```

### 2. Load the pool/template object file
```yaml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: CoreIPPrefixPool
  data:
    - name: "Site Prefix Pool"
      description: "Allocates /24 prefixes to sites"
      default_prefix_type: IpamIPPrefix
      default_prefix_length: 24
      default_member_type: address
      ip_namespace: default
      resources:
        kind: IpamIPPrefix
        data:
          - prefix: 10.0.0.0/8
            member_type: prefix
            is_pool: true
```

### 3a. Create via UI using the object template

Image

Mutation sent by the frontend:
```json
{
  "variables": {},
  "query": "mutation {\n  LocationSiteCreate(\n    data: {object_template: {id: \"18a0327c-6c15-69fb-292a-c51820cd0fbf\"}, name: {value: \"lon01\"}}\n  ) {\n    object {\n      id\n      display_label\n      hfid\n      __typename\n    }\n    __typename\n  }\n}"
}
```

Response:

```
{
    "data": {
        "LocationSiteCreate": null
    },
    "errors": [
        {
            "message": "prefix is mandatory for LocationSite at prefix",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "LocationSiteCreate"
            ]
        }
    ]
}
```

**Error:**
```
prefix is mandatory for LocationSite at prefix
```

### 3b. Create via `infrahubctl object load`
```bash
INFRAHUB_ECHO_GRAPHQL_QUERIES=true infrahubctl object load test_object_template_resource_pools/objects.yml
```

**Error (returned before any mutation is sent):**
```
File 'test_object_template_resource_pools/objects.yml' [7] is not valid!
 1.prefix: prefix is mandatory
 2.prefix: prefix is mandatory
 3.prefix: prefix is mandatory
```

### Additional Information

_No response_

gmazoyer:
When creating an object from a template with a resource pool assigned to a required relationship, the preview phase (`process_pools=False`) skipped pool allocation via `NoOpPoolAllocator`, leaving the field unset. Mandatory validation then rejected the missing field before the actual creation could allocate from the pool.

Track fields with pending pool allocations in `NodeTemplateApplier`. When allocation returns `None` but a pool peer exists, record the field name. `handle_object_template` returns this set so the validation can skip those mandatory fields.

Closes #8701 

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Template-based object creation no longer reports missing required fields when those fields are deferred due to resource pool allocations in preview mode.
* **Tests**
  * Template tests updated to treat primary address and slot ID as required.
* **Changelog**
  * Added entry documenting the fix for pool-backed template object creation.

### DIFF
--- a/backend/infrahub/templates/node_applier.py
+++ b/backend/infrahub/templates/node_applier.py
@@ -25,6 +25,7 @@ class NodeTemplateApplier:
         self.db = db
         self.branch = branch
         self.pool_allocator = pool_allocator
+        self.pool_pending_fields: set[str] = set()
 
     async def apply(
         self,
@@ -116,6 +117,8 @@ class NodeTemplateApplier:
             if allocated_value is not None:
                 pool_node = await pool_relationship.get_peer(db=self.db, raise_on_error=True)
                 fields[original_name] = {"value": allocated_value, "source": pool_node.id}
+            elif await pool_relationship.get_peer(db=self.db):
+                self.pool_pending_fields.add(original_name)
             return
         # IP pool: allocate a node for the relationship
         allocated = await self.pool_allocator.allocate_for_relationship(
@@ -124,3 +127,5 @@ class NodeTemplateApplier:
         if allocated:
             pool_node = await pool_relationship.get_peer(db=self.db, raise_on_error=True)
             fields[original_name] = {"peer": allocated, "_relation__source": pool_node.id}
+        elif await pool_relationship.get_peer(db=self.db):
+            self.pool_pending_fields.add(original_name)
--- a/backend/tests/functional/templates/test_template_resource_pool.py
+++ b/backend/tests/functional/templates/test_template_resource_pool.py
@@ -504,7 +504,7 @@ class TestTemplateNumberPoolAttributes(TestInfrahubApp):
                     attributes=[
                         AttributeSchema(name="name", kind="Text", unique=True),
                         AttributeSchema(name="location", kind="Text", optional=True),
-                        AttributeSchema(name="slot_id", kind="Number", optional=True),
+                        AttributeSchema(name="slot_id", kind="Number", optional=False),
                     ],
                 ),
             ],
--- a/backend/tests/functional/templates/test_template_resource_pool.py
+++ b/backend/tests/functional/templates/test_template_resource_pool.py
@@ -72,7 +72,7 @@ class TestTemplateResourcePoolCreation(TestInfrahubApp):
                             peer="IpamIPAddress",
                             label="Primary IP Address",
                             cardinality=RelationshipCardinality.ONE,
-                            optional=True,
+                            optional=False,
                         ),
                     ],
                 ),
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -577,7 +577,11 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         # Generate Attribute and Relationship and assign them
         # -------------------------------------------
         errors.extend(await self._process_fields_relationships(fields=fields, db=db))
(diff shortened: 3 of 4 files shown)


backend/tests/functional/templates/test_template_resource_pool.py
```diff
@@ -72,7 +72,7 @@ async def device_schema(self, db: InfrahubDatabase, initialize_registry: None) -
                             peer="IpamIPAddress",
                             label="Primary IP Address",
                             cardinality=RelationshipCardinality.ONE,
-                            optional=True,
+                            optional=False,
```
gmazoyer, line 75: This allows us to use existing tests to validate our fix.

backend/infrahub/templates/node_applier.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, db: InfrahubDatabase, branch: Branch, pool_allocator: PoolAll
         self.db = db
         self.branch = branch
         self.pool_allocator = pool_allocator
+        self.pool_pending_fields: set[str] = set()
```
ajtmccarty, line 28: I think of "fields" as being both attributes and relationships, but this looks like it's just relationships. do attributes need to be handled in here somehow?
I see that `pool_pending_fields` is passed into `Node._process_fields_attributes` but it looks like this will never actually include an attribute name
I could definitely be missing something. the Node class is still making it hard to trace the logic

gmazoyer: Agreed this is hard to read. The change below this one is the one that applies to the attribute (e.g. numbers from pool), [URL for reference](https://github.com/opsmill/infrahub/pull/8704/changes#diff-9f6edaffb461d3268132e0f04c3a5854eb3ead70cd7ba078266c2509d19822ecR120-R121) but not sure it's stable. The second change after that one takes care of the relationship (e.g. IP prefix/address from pool)

FragmentedPacket: This might be a valid concern for Number attributes that are required and assigned by an object template?

gmazoyer: > This might be a valid concern for Number attributes that are required and assigned by an object template?

Not sure I understand. This PR fixes the behaviour for both relationships and attributes.

ajtmccarty (approved):

PR #8710, opened 2026-03-26
Migrate Node.js to 24 LTS

bilalabbad:
## Summary by CodeRabbit

* **Chores**
  * Migrated Node.js runtime from 22 to 24 LTS across CI workflows, Docker builds, and package engine constraints; added Node 24+ engine declarations to frontend packages.
  * Switched one workflow’s install step from pnpm to npm.
  * Updated an internal SDK submodule to a newer commit.
  * Added a housekeeping changelog entry noting the Node 24 migration.

* **Documentation**
  * Updated frontend setup and prerequisite notes to require Node.js 24+.


ogenstad:
The failure in the CI should be fixed after a rebase. Having said that perhaps we want this to go into `stable` instead?

fatih-acar (approved):
I believe it's safe to this upgrade in `stable` directly but would like you to confirm
Also need this change in the enterprise repo.

PR #8712, opened 2026-03-26
fix(backend): add db backoff retry for resolvers

### LINKED ISSUE #8696: task: add database query retry for read queries (not only mutations)
### Component

API Server / GraphQL

### Task Description

When running concurrent GraphQL queries that fetches a cardinality many relationship with a lot of peers, we can end up in a situation where we exhaust Neo4j's client thread pool; thus throwing a TransientError to the user.

We already have the `retry_db_transaction` decorator that handles such transient errors but it's mostly applied to mutations. We should also use it for the many_relationship_resolver and make the retries smoother (by implementing exponential backoff retries?)

fatih-acar:
This PR improves the retry_db_transaction decorator by having a configurable backoff retry.
It also enables the decorator on the graphql query resolvers.

Adds component tests to check the retry behavior solves hard failures
when neo4j thread pool exhausts: we must start a new container with a
low thread pool size.

I had to duplicate a few fixtures to have the new neo4j container working... I'm not very satisfied of the tests, looking for feedback on those. There's also high chance they are flaky... so we probably want to disable them in CI for now.

Fixes https://github.com/opsmill/infrahub/issues/8696

## Summary by CodeRabbit

* **Bug Fixes**
  * GraphQL read operations now automatically retry with exponential backoff and jitter to reduce transient failures.

* **New Features**
  * Three new configurable database retry parameters: base delay, max delay, and jitter.

* **Documentation**
  * Configuration reference updated with the new retry environment variables and defaults.

* **Tests**
  * Added unit and component tests validating retry behavior and backoff timing under contention.

* **Chores**
  * Minor test fixture refactor and changelog entry added.

### DIFF
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -306,6 +306,15 @@ class DatabaseSettings(BaseSettings):
     retry_limit: int = Field(
         default=3, description="Maximum number of times a transient issue in a transaction should be retried."
     )
+    retry_base_delay: float = Field(
+        default=0.1, ge=0, description="Base delay in seconds for exponential backoff on transaction retries."
+    )
+    retry_max_delay: float = Field(
+        default=2.0, ge=0, description="Maximum delay in seconds for exponential backoff on transaction retries."
+    )
+    retry_jitter_max: float = Field(
+        default=0.1, ge=0, description="Maximum jitter in seconds added to retry delay to avoid thundering herd."
+    )
     max_concurrent_queries: int = Field(
         default=0, ge=0, description="Maximum number of concurrent queries that can run (0 means unlimited)."
     )
--- a/backend/tests/component/graphql/test_graphql_read_retry.py
+++ b/backend/tests/component/graphql/test_graphql_read_retry.py
@@ -1,0 +1,244 @@
+"""Component tests for GraphQL read resolver retry on transient Neo4j errors.
+
+These tests simulate Neo4j server-side Bolt thread pool exhaustion by spinning up
+a dedicated Neo4j container with server.bolt.thread_pool_max_size=10 and firing
+concurrent queries that hit many-cardinality relationship resolvers.
+
+All fixtures are module-scoped so xdist (--dist loadscope) runs them on a single
+worker and the dedicated container is stopped as soon as the module finishes.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from typing import TYPE_CHECKING, AsyncGenerator, Generator
+
+import pytest
+from neo4j.exceptions import TransientError
+
+from infrahub import config
+from infrahub.core import registry
+from infrahub.core.initialization import add_indexes
+from infrahub.core.node import Node
+from infrahub.database import InfrahubDatabase, get_db
+from infrahub.graphql.initialization import prepare_graphql_params
+from tests.conftest import (
+    do_car_person_schema_unregistered,
+    do_data_schema,
+    do_default_branch,
+    do_empty_database,
+    do_group_schema,
+    do_local_storage_dir,
+    do_reset_registry,
+)
+from tests.helpers.constants import INFRAHUB_USE_TEST_CONTAINERS, NEO4J_IMAGE, PORT_BOLT_NEO4J
+from tests.helpers.graphql import graphql
+from tests.helpers.utils import get_exposed_port, start_neo4j_container
+
+if TYPE_CHECKING:
+    from graphql.execution import ExecutionResult
+
+    from infrahub.core.branch import Branch
+    from infrahub.core.schema.schema_branch import SchemaBranch
+
+NUM_CARS = 10
+
+QUERY_MANY_RELATIONSHIP = """
+query {
+    TestPerson {
+        edges {
+            node {
+                name { value }
+                cars {
+                    edges {
+                        node {
+                            name { value }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+"""
+
+# ---------------------------------------------------------------------------
+# Module-scoped fixtures for the dedicated small-thread-pool Neo4j container
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def neo4j_small_thread_pool(request: pytest.FixtureRequest) -> int:
+    """Start a dedicated Neo4j container with a minimal Bolt thread pool (size=10)."""
+    if not INFRAHUB_USE_TEST_CONTAINERS:
+        pytest.skip("Test containers are disabled")
+
+    container = start_neo4j_container(
+        neo4j_image=NEO4J_IMAGE,
+        extra_env={"NEO4J_server_bolt_thread__pool__max__size": "10"},
+    )
+    request.addfinalizer(container.stop)
+
+    return get_exposed_port(container, PORT_BOLT_NEO4J)
+
+
+@pytest.fixture(scope="module")
+async def db_small_thread_pool(
+    neo4j_small_thread_pool: int,
+) -> AsyncGenerator[InfrahubDatabase, None]:
+    """Create an InfrahubDatabase connected to the small-thread-pool Neo4j."""
+    original_address = config.SETTINGS.database.address
+    original_port = config.SETTINGS.database.port
+
+    config.SETTINGS.database.address = "localhost"
+    config.SETTINGS.database.port = neo4j_small_thread_pool
+
+    driver = await get_db(retry=5)
+    db = InfrahubDatabase(driver=driver)
+    await add_indexes(db=db)
+
+    yield db
+
+    await db.close()
+    config.SETTINGS.database.address = original_address
+    config.SETTINGS.database.port = original_port
+
+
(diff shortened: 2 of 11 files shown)


backend/infrahub/config.py
```diff
@@ -306,6 +306,15 @@ class DatabaseSettings(BaseSettings):
...
         default=3, description="Maximum number of times a transient issue in a transaction should be retried."
     )
+    retry_base_delay: float = Field(
+        default=0.1, ge=0, description="Base delay in seconds for exponential backoff on transaction retries."
+    )
+    retry_max_delay: float = Field(
+        default=2.0, ge=0, description="Maximum delay in seconds for exponential backoff on transaction retries."
+    )
+    retry_jitter_max: float = Field(
+        default=0.1, ge=0, description="Maximum jitter in seconds added to retry delay to avoid thundering herd."
```
ajtmccarty, line 316: "thundering herd" 👍

backend/tests/component/graphql/test_graphql_read_retry.py
```diff
@@ -0,0 +1,369 @@
...
+
+async def test_concurrent_queries_fail_without_retries(
+    db_small_thread_pool: InfrahubDatabase,
+    default_branch_stp: Branch,
+    test_data_stp: None,
+) -> None:
+    """Test that without retries, concurrent queries on a small thread pool fail
+    with TransientError, proving the retry mechanism is needed."""
+    original_retry_limit = config.SETTINGS.database.retry_limit
+    config.SETTINGS.database.retry_limit = 1  # Only 1 attempt, no retries
```
ajtmccarty: should be able to use a fixture with the yield pattern for this and avoid the `try-finally` below

backend/tests/component/graphql/test_graphql_read_retry.py
```diff
@@ -0,0 +1,369 @@
...
+    registry.schema = SchemaManager()
+    return branch
+
+
+@pytest.fixture(scope="module")
+async def car_person_schema_stp(
+    db_small_thread_pool: InfrahubDatabase,
+    default_branch_stp: Branch,
+) -> SchemaBranch:
+    """Register data_schema + node_group_schema + car_person_schema on the small-thread-pool DB.
```
ajtmccarty: we have a pattern that we've used to rescope fixtures in other places that I think would let you remove all or almost all of this function. I'm pretty sure we've got fixtures for all of these (maybe not CoreNode) in other places with the default or `"class"` scope

this is one example
```
@pytest.fixture
async def default_branch(
    reset_registry: None, local_storage_dir: Path, empty_database: None, db: InfrahubDatabase
) -> Branch:
    return await do_default_branch(db=db)


@pytest.fixture(scope="class")
async def default_branch_scope_class(
    reset_registry_scope_class: None,
    local_storage_dir_scope_class: Path,
    empty_database_scope_class: None,
    db: InfrahubDatabase,
) -> Branch:
    return await do_default_branch(db=db)


async def do_default_branch(db: InfrahubDatabase) -> Branch:
    branch = await create_default_branch(db=db)
    await create_global_branch(db=db)
    registry.schema = SchemaManager()
    return branch

```

backend/tests/component/graphql/test_graphql_read_retry.py
```diff
@@ -0,0 +1,369 @@
...
+        f"Errors: {[r.errors if hasattr(r, 'errors') else str(r) for r in results if r not in successful]}"
+    )
+
+    for result in successful:
+        person_data = result.data["TestPerson"]["edges"][0]["node"]
+        assert person_data["name"]["value"] == "John"
+        assert len(person_data["cars"]["edges"]) == NUM_CARS
+
+
+async def test_concurrent_queries_fail_without_retries(
```
ajtmccarty, line 211: we don't usually commit this kind of test to stable or develop. maybe I am misunderstanding, but this looks like the test you write to reproduce an issue before fixing it.

fatih-acar: > this looks like the test you write to reproduce an issue before fixing it

yes it is, and I don't think it's a good idea to keep it enabled in CI since it can be flaky...
should I disable the test instead and keep the code in case we need to come back later? or completely remove the code

ajtmccarty: I think disabling it with a note is fine

fatih-acar: done

backend/tests/component/graphql/test_graphql_read_retry.py
fatih-acar: thanks I think there's less duplicate now

BeArchiTek:
@cubic review this PR

backend/tests/component/graphql/test_graphql_read_retry.py
```diff
@@ -0,0 +1,244 @@
+"""Component tests for GraphQL read resolver retry on transient Neo4j errors.
+
+These tests simulate Neo4j server-side Bolt thread pool exhaustion by spinning up
+a dedicated Neo4j container with server.bolt.thread_pool_max_size=1 and firing
```
ajtmccarty: I think this is actually set to 10 in the test

backend/tests/component/graphql/test_graphql_read_retry.py
ajtmccarty, line 1: not certain if this should be a component test
I'd think of spinning up a new container as more of an `integration` or `integration_docker` test... BUT it is just another database container, so I'm not sure. asking for team's input in slack, but does not need to block merging this PR

backend/tests/component/graphql/test_graphql_read_retry.py
```diff
@@ -0,0 +1,244 @@
...
+    if not INFRAHUB_USE_TEST_CONTAINERS:
+        pytest.skip("Test containers are disabled")
+
+    container = start_neo4j_container(
+        neo4j_image=NEO4J_IMAGE,
+        extra_env={"NEO4J_server_bolt_thread__pool__max__size": "10"},
+    )
+    request.addfinalizer(container.stop)
+
+    return {PORT_BOLT_NEO4J: get_exposed_port(container, PORT_BOLT_NEO4J)}
```
ajtmccarty: looks like it would be more straightforward to just return the `container` or the result of `get_exposed_port` here

backend/tests/unit/test_retry_db_transaction.py
ajtmccarty, line 1: this isn't really a graphql test. I think this could be moved to `tests/unit/test_retry_db_transaction.py`

ajtmccarty (approved):
just some test comments that don't need to block merging

backend/tests/unit/graphql/resolvers/test_retry_db_transaction.py
```diff
@@ -0,0 +1,154 @@
...
+                patch("infrahub.database.random.uniform", return_value=0.0),
+                pytest.raises(TransientError),
+            ):
+                await decorated()
+
+            delays = [call.args[0] for call in sleep_mock.call_args_list]
+            for delay in delays:
+                assert delay <= 2.0
+        finally:
+            config.SETTINGS.database.retry_limit = 3
```
ogenstad: I agree with the bot on this comment. We typically revert config changes by using a fixture.

backend/tests/unit/graphql/resolvers/test_retry_db_transaction.py
```diff
@@ -0,0 +1,154 @@
...
+
+from infrahub import config
+from infrahub.database import retry_db_transaction
+from infrahub.database.metrics import TRANSACTION_RETRIES
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+
+@pytest.fixture(autouse=True)
```
ogenstad: Do we need this to be an autouse fixture? Can't we tie it to one of the tests that will use it?

backend/tests/component/graphql/test_graphql_read_retry.py
polmichel: In its current state, based on the `dev/guidelines/backend/testing.md` file, I think this should be into `component` as this is only related to the database, but maybe in a separate folder that test specifically the database container and its settings. I do not find an existing `database` folder in `component` tests, we could create one if this is the first time we try to solve this kind of db specific issue.

In my opinion, if we want this test scenario to be related to similar other `integration` tests, it may be justified if we group this kind of test. In that case, we should perhaps send the requests via the "accessible" API `infrahub_sdk.client.InfrahubClient.execute_graphql` rather than directly from the inner `graphql` library?

ajtmccarty: I think either way is fine as long as the test exercises the resolvers that have the new decorator

fatih-acar: I understand the debate but this felt obvious to me: I just want to test an internal behavior (the graphql resolvers) against a custom, real database... component test is the right place in my opinion

ajtmccarty (approved):

PR #8713, opened 2026-03-26
unify TLS logic and caching in one place

ajtmccarty:
# Why

TLS context creation logic was duplicated in [this enterprise PR](https://github.com/opsmill/infrahub-private/pull/346) to add a new `TCPTransport` class for log forwarding. Consolidating the TLS context creation code in one place that can be reused is better than duplicating it. This also adds a TLS context caching component `TlsContextRegistry` so that any TLS context can be computed one time per worker and then cached

## What changed

- **`infrahub/tls/context_builder.py`** — new `TlsContextBuilder` class with a single `build(insecure, ca_bundle, force_verify)` static method containing the canonical CA bundle resolution logic (replaces both `HTTPSettings.get_tls_context` and `build_tls_context`)
- **`infrahub/tls/registry.py`** — new `TlsContextRegistry` class that wraps `TlsContextBuilder`, caches built contexts by `(insecure, ca_bundle, force_verify)` key, and exposes a `validate()` method for startup-time pre-warming
- **`workers/dependencies.py`** — `build_tls_registry()` / `get_tls_registry()` added following the existing `build_*` / `get_*` singleton pattern; `build_http_service` and `build_workflow` now inject `TlsContextRegistry` into their respective components
- **`HttpxAdapter`** — now accepts `TlsContextRegistry` as a constructor argument; `@cached_property tls_context / tls_context_verified` removed in favour of direct registry calls in `verify_tls()`
- **`WorkflowWorkerExecution`** — class-level `_http_adapter = HttpxAdapter()` workaround removed (the comment explicitly anticipated this); now accepts `TlsContextRegistry` directly
- **`HTTPSettings.get_tls_context`** — removed; the model validator calls `TlsContextBuilder.build()` directly for startup validation
- **`build_tls_context` in enterprise `transport/base.py`** — removed; `tcp.py` now calls `TlsContextBuilder.build()` directly
- `server.py`, `workers/infrahub_async.py`, `cli/__init__.py`, `cli/git_agent.py` — updated to pass a constructed `HttpxAdapter` into `InfrahubServices.new()`

No public API changes, no schema changes, no config/env changes.

### Suggested review order
1. `infrahub/tls/context_builder.py` and `infrahub/tls/registry.py` — the new primitives
2. `workers/dependencies.py` — DI wiring
3. `services/adapters/http/httpx.py` and `services/adapters/workflow/worker.py` — the two main consumers
4. `config.py` — removal of `get_tls_context`
5. `enterprise/transport/base.py` and `tcp.py` — removal of `build_tls_context`
6. Tests

## How to review

Focus on `tls/registry.py` — specifically that the cache key `(insecure, ca_bundle, force_verify)` is correct and that `validate()` surfaces SSL errors at startup rather than at first use.

## How to test

```bash
# Unit tests for TlsContextBuilder
uv run invoke backend.test-unit -- -k tls
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. `HTTPSettings.get_tls_context()` was internal; no public API removed.
- **Performance:** SSL contexts are now shared across all components rather than built per-instance or per-reconnect.

## Checklist

- [x] Tests added/updated

## Summary by CodeRabbit

* **Refactor**
  * Centralized TLS certificate/context handling into a registry and builder for more reliable and consistent TLS behavior across HTTP and workflow components.
  * Updated service startup to wire the new TLS/HTTP dependencies throughout the app, improving TLS validation and reuse.

* **Tests**
  * Added unit tests for TLS context construction and updated integration tests to exercise the new TLS wiring.

### DIFF
--- a/backend/infrahub/cli/git_agent.py
+++ b/backend/infrahub/cli/git_agent.py
@@ -24,6 +24,7 @@ from infrahub.services.adapters.message_bus import InfrahubMessageBus
 from infrahub.services.adapters.workflow.local import WorkflowLocalExecution
 from infrahub.services.adapters.workflow.worker import WorkflowWorkerExecution
 from infrahub.trace import configure_trace
+from infrahub.workers.dependencies import build_tls_registry, get_http
 
 if TYPE_CHECKING:
     from infrahub.cli.context import CliContext
@@ -112,7 +113,7 @@ async def start(
     database = await context.init_db(retry=1)
 
     workflow = config.OVERRIDE.workflow or (
-        WorkflowWorkerExecution()
+        WorkflowWorkerExecution(tls_registry=build_tls_registry())
         if config.SETTINGS.workflow.driver == config.WorkflowDriver.WORKER
         else WorkflowLocalExecution()
     )
@@ -130,6 +131,7 @@ async def start(
         workflow=workflow,
         message_bus=message_bus,
         component_type=component_type,
+        http=get_http(),
     )
 
     # Initialize the lock
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -539,7 +539,9 @@ class HTTPSettings(BaseSettings):
         try:
             # Validate that the context can be created, we want to raise this error during application start
             # instead of running into issues later when we first try to use the tls context.
-            TlsContextBuilder.build(insecure=self.tls_insecure, ca_bundle=self.tls_ca_bundle)
+            TlsContextBuilder.build(
+                insecure=self.tls_insecure, ca_bundle=self.tls_ca_bundle, force_verify=bool(self.tls_ca_bundle)
+            )
         except ssl.SSLError as exc:
             raise ValueError(f"Unable load CA bundle from {self.tls_ca_bundle}: {exc}") from exc
 
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -27,6 +27,7 @@ from typing_extensions import Self
 
 from infrahub.constants.database import DatabaseType
 from infrahub.exceptions import InitializationError, ProcessingError
+from infrahub.tls.context_builder import TlsContextBuilder
 
 if TYPE_CHECKING:
     from infrahub.services.adapters.cache import InfrahubCache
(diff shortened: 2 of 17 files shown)


backend/infrahub/cli/git_agent.py
```diff
@@ -112,7 +113,7 @@ async def start(
     database = await context.init_db(retry=1)
 
     workflow = config.OVERRIDE.workflow or (
-        WorkflowWorkerExecution()
+        WorkflowWorkerExecution(tls_registry=build_tls_registry())
```
ajtmccarty, line 116: `WorkflowWorkerExecution` had a confusing kludge where it instantiated `HttpxAdapter` as a class variable so that it could use `HttpxAdapeter.verify_tls()` in a context manager for running a deployment.
passing in the new `TlxContextRegistry` lets us remove that weird workaround

backend/infrahub/services/adapters/http/httpx.py
```diff
@@ -24,7 +26,9 @@ class HttpxAdapter(InfrahubHTTP):
     allow users to define configurations such as timeout, TLS options
     and eventually proxy settings in one location."""
 
-    _settings: config.HTTPSettings | None = None
+    def __init__(self, tls_registry: TlsContextRegistry) -> None:
+        self._tls_registry = tls_registry
```
ajtmccarty, line 30: new dependency that needs to be injected. allows removing the `@cached_property`s below and moves them to the `TlsContextRegistry` which is now responsible for caching TLS context instances

backend/infrahub/workers/dependencies.py
```diff
@@ -132,9 +133,20 @@ def get_workflow(workflow: InfrahubWorkflow = Depends(build_workflow)) -> Infrah
     return workflow
 
 
+def build_tls_registry() -> TlsContextRegistry:
+    if "tls_registry" not in _singletons:
+        _singletons["tls_registry"] = TlsContextRegistry()
+    return _singletons["tls_registry"]
```
ajtmccarty, line 144: new singleton so everywhere can use the same `TlsContextRegistry` in a given worker

backend/infrahub/config.py
```diff
@@ -538,36 +539,12 @@ def set_tls_context(self) -> Self:
...
             # Validate that the context can be created, we want to raise this error during application start
             # instead of running into issues later when we first try to use the tls context.
-            self.get_tls_context()
+            TlsContextBuilder.build(insecure=self.tls_insecure, ca_bundle=self.tls_ca_bundle)
         except ssl.SSLError as exc:
             raise ValueError(f"Unable load CA bundle from {self.tls_ca_bundle}: {exc}") from exc
 
         return self
 
-    def get_tls_context(self, force_verify: bool = False) -> ssl.SSLContext:
```
ajtmccarty, line 547: moved to `TlsContextBuilder.build`

backend/tests/unit/config/test_config.py
```diff
@@ -75,82 +70,3 @@ def test_storage_max_file_size_environment_variable() -> None:
     with patch.dict(os.environ, {"INFRAHUB_STORAGE_MAX_FILE_SIZE": "75"}):
         assert StorageSettings().max_file_size == 75
     assert isinstance(SETTINGS.storage.max_file_size, int)
-
```
ajtmccarty, line 78: moved to their own file b/c they are no longer part of `config`

backend/infrahub/cli/git_agent.py
ogenstad: That sounds good, having said that, we should probably remove this cli command `git_agent.py`... but should be done in another PR as it's not related to this change.

ogenstad (approved):
LGTM, great improvement!

One additional thing to highlight could potentially be this part:

```python
def build_client() -> InfrahubClient:
    client_config = Config(address=config.SETTINGS.main.internal_address, retry_on_failure=True)
    client_config.set_ssl_context(context=get_http().verify_tls())
    client = InfrahubClient(config=client_config)
    # Populate client schema cache using our internal schema cache
    if registry.schema:
        for branch in registry.schema.get_branches():
            client.schema.set_cache(schema=registry.schema.get_sdk_schema_branch(name=branch), branch=branch)

    return client
```

Reaching into `get_http().verify_tls()` here still feels somewhat hacky. On the other hand even if we were to change this we'd still be using the settings from the HTTP adapter. Have a look at this and see if you want to change anything, otherwise we can leave it as is for now. I think the next step after this is merged would be to introduce a common TLS setting across Infrahub which will be the default configuration option that could be overridden by a specific adapter if needed.

fatih-acar (approved):
Thanks for this long awaited refactor, LGTM!

PR #8716, opened 2026-03-27
IFC-2383: Replace self-targeting trigger fields with _trigger_placeholder

polmichel:
# Why

Self-targeting computed attribute triggers were listing every referenced field individually. This caused a bunch of background task firings on every attribute update. Now, the computation of Jinja2 local attributes is handled inline during the mutation. 
To avoid useless triggers, we will replace the local fields trigger by a global `_trigger_placeholder` holder that is fired when the schema is changing. This mimics what is currently in place for HFID and Display Labels.

Relates to [IFC-2383](https://opsmill.atlassian.net/browse/IFC-2383)

## What changed

- `gather_trigger_computed_attribute_jinja2` now emits `_trigger_placeholder` for self-targeting triggers instead of real field names. Remote triggers (peer attribute changes) still preserve real field names.
- Added `TRIGGER_PLACEHOLDER_FIELD` constant in `infrahub.trigger.constants`.
- Fixed `get_trigger_nodes()` to deduplicate field names in `attributes` and `relationships` lists, preventing duplicates when multiple template tokens reference the same peer relationship.
- Added functional tests validating self-targeting vs remote trigger field structure.

## How to test

```bash
uv run pytest backend/tests/functional/computed_attributes/test_local_computation.py -v
```

## Checklist

- [x] Tests added/updated

[IFC-2383]: https://opsmill.atlassian.net/browse/IFC-2383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate entries in computed attribute trigger node attributes and relationships that could occur during trigger discovery.

* **Refactor**
  * Consolidated trigger placeholder field handling across modules using a centralized constant for improved maintainability.

### DIFF
--- a/backend/infrahub/computed_attribute/tasks.py
+++ b/backend/infrahub/computed_attribute/tasks.py
@@ -341,7 +341,7 @@ async def computed_attribute_setup_jinja2(
 
         report: TriggerSetupReport = await setup_triggers_specific(
             gatherer=gather_trigger_computed_attribute_jinja2, trigger_type=TriggerType.COMPUTED_ATTR_JINJA2
-        )  # type: ignore[misc]
+        )
         # Configure all ComputedAttrJinja2Trigger in Prefect
 
         all_triggers = report.triggers_with_type(trigger_type=ComputedAttrJinja2TriggerDefinition)
--- a/backend/tests/component/display_labels/test_gather.py
+++ b/backend/tests/component/display_labels/test_gather.py
@@ -102,4 +102,4 @@ async def test_gather_trigger_gather_trigger_display_labels_jinja2_custom_schema
     assert test_car_trigger.trigger.match == {"infrahub.node.kind": "TestCar"}
     assert isinstance(test_car_trigger.trigger.match_related, dict)
     assert "infrahub.field.name" in test_car_trigger.trigger.match_related
-    assert sorted(test_car_trigger.trigger.match_related["infrahub.field.name"]) == [TRIGGER_PLACEHOLDER_FIELD]
+    assert test_car_trigger.trigger.match_related["infrahub.field.name"] == [TRIGGER_PLACEHOLDER_FIELD]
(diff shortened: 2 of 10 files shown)


backend/infrahub/computed_attribute/tasks.py
```diff
@@ -341,7 +341,7 @@ async def computed_attribute_setup_jinja2(
 
         report: TriggerSetupReport = await setup_triggers_specific(
             gatherer=gather_trigger_computed_attribute_jinja2, trigger_type=TriggerType.COMPUTED_ATTR_JINJA2
-        )  # type: ignore[misc]
```
ogenstad, line 344: I wonder if this was due to an update in later Prefect versions and if we can remove this elsewhere..

backend/tests/component/display_labels/test_gather.py
```diff
@@ -101,4 +102,4 @@ async def test_gather_trigger_gather_trigger_display_labels_jinja2_custom_schema
     assert test_car_trigger.trigger.match == {"infrahub.node.kind": "TestCar"}
     assert isinstance(test_car_trigger.trigger.match_related, dict)
     assert "infrahub.field.name" in test_car_trigger.trigger.match_related
-    assert sorted(test_car_trigger.trigger.match_related["infrahub.field.name"]) == ["_trigger_placeholder"]
+    assert sorted(test_car_trigger.trigger.match_related["infrahub.field.name"]) == [TRIGGER_PLACEHOLDER_FIELD]
```
ogenstad: Just a copied line, but looks like `sorted()` might be overkill here 😅

ogenstad (approved):

PR #8719, opened 2026-03-27
IFC-2385: Event consolidation & bulk validation tests

polmichel:
# Why

[IFC-2385](https://opsmill.atlassian.net/browse/IFC-2385) — Local Jinja2 computed attributes need integration tests proving that event consolidation and bulk update work correctly without background tasks.

Closes IFC-2385

## What changed

- **Event consolidation tests**: Verify that updating a local attribute that triggers Jinja2 recomputation produces a single `NodeChangelog` containing both the original and computed attribute changes; also verify no spurious changelog entry when the computed value is unchanged.
- **Bulk update test**: Create 55 nodes with Jinja2 computed attributes, bulk-update a local attribute on each, and confirm all computed values are correct (proves inline path handles everything without Prefect).

## How to review

1. `backend/tests/helpers/schema/server.py` — new test schema
2. `backend/tests/functional/computed_attributes/test_local_computation.py`: three new test classes (`TestEventConsolidation`, `TestBulkUpdateLocalComputation`)

## How to test

```bash
uv run pytest backend/tests/functional/computed_attributes/test_local_computation.py -v
```

## Checklist

- [x] Tests added/updated

[IFC-2385]: https://opsmill.atlassian.net/browse/IFC-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/tests/functional/computed_attributes/test_local_computation.py
```diff
@@ -184,3 +184,149 @@ async def test_null_relationship_renders_template(
...
+        # Since neither name nor description actually changed value, the changelog
+        # should NOT contain 'description' (the recomputation produced the same value)
+        changelog = tshirt.node_changelog
+        assert "description" not in changelog.attributes, (
+            f"Computed attribute 'description' should not be in changelog when value is unchanged, "
+            f"got updated_fields={changelog.updated_fields}"
+        )
+
+
+class TestBulkUpdateLocalComputation(TestInfrahubApp):
```
polmichel: Should this test be in a dedicated place related to 'performance tests'? I don't think creating a test with more than three computed attributes makes sense if we are not precisely measuring and managing the execution time.

ogenstad: Yeah I think we should treat this kind of test as a performance test. Since we don't have any time component here we don't really know if there's a change in performance between runs and from a functional point of view three tests vs 55 shouldn't make a difference. So I'd suggest that you at least modify the count before merging this, or remove the entire test provided that we're already testing the same elsewhere and then later we can consider recreating it as a separate performance test.

ogenstad (approved):

backend/tests/functional/computed_attributes/test_local_computation.py
polmichel: Yes after the meeting with Fatih on performance tests, I got a clearer view on which test suite can be used regarding this subject. I will remove this test, and merge the PR. Thx for the review

PR #8721, opened 2026-03-27
Generator documentation

wvandeun:


python_sdk
BaptisteGi, line 1: Should we have changes here?

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+      }
+    }
+  }
+}
+```
+
+For each member of the `widgets` group, Infrahub:
+
+1. Reads the parameter mapping: `name` → `"name__value"`
+2. Extracts the value from the target object using the double-underscore path
```
BaptisteGi:
```suggestion
2. Extracts the value from the target object using the defined path
```

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+| `widget2`     | `widget2.name.value`   | `"widget2"`     | `$name = "widget2"`     |
+
+Each run's GraphQL query only returns data for its specific target, keeping runs independent.
+
+### Double-underscore notation
+
+The double-underscore (`__`) in parameter values traverses the object hierarchy:
+
+- `name__value`: attribute `name`, property `value`
+- `location__name__value`: relationship `location` (cardinality-one), then attribute `name`, property `value`
```
BaptisteGi, line 126: Is it really something for generators?

BaptisteGi: Overall I'm not sure about that relationship use case for generators

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+
+## The relationship between `member_of_groups` and Generator targeting
+
+Objects become Generator targets by being members of the group specified in the Generator definition's `targets` field.
+
+### How objects become targets
+
+You add objects to a target group through the `member_of_groups` relationship, which can be set:
+
+- **In the UI**: when creating or editing an object, select the group in the "Member of groups" field
```
BaptisteGi: It's under "Manage" > "Groups" for edit

BaptisteGi: For what's it's worth we even have a guide for groups https://docs.infrahub.app/guides/groups

BaptisteGi: So that section could be "add the object to a group so it becomes a target > here is the guide"

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+- **Remove an object from the group**: it is no longer targeted (existing generated objects are not automatically cleaned up — see [known limitation #3289](https://github.com/opsmill/infrahub/issues/3289))
+
+### Standard groups vs. Generator groups
+
+These two group types serve different purposes and are often confused:
+
+| Group type            | Purpose                                                                                            | You manage it |
+|-----------------------|----------------------------------------------------------------------------------------------------|---------------|
+| `CoreStandardGroup`   | Defines which objects are **targeted by** a Generator. Listed in the `targets` field of the Generator definition. | Yes — you create it and control membership |
+| `CoreGeneratorGroup`  | Tracks which objects were **created by** a Generator. Managed automatically by the SDK tracking feature. | No — Infrahub manages this automatically |
```
BaptisteGi:
```suggestion
| `CoreGeneratorGroup`  | Tracks which objects were **created by** a Generator instance. Managed automatically by the SDK tracking feature. | No — Infrahub manages this automatically |
```

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+```python
+class DeviceGenerator(InfrahubGenerator):
+    async def generate(self, data: dict) -> None:
+        for edge in data["TestDevice"]["edges"]:
+            device_name = edge["node"]["name"]["value"]
+            device_role = edge["node"]["role"]["value"]
+            # ... create objects using self.client
+```
+
+This is the more lightweight mode — no conversion overhead, and you work with plain Python dictionaries. You can also use [Pydantic models](https://docs.pydantic.dev/) to parse the response into typed objects for better IDE support and validation.
```
BaptisteGi, line 241: Can't recall the status of Pydantic models here, not sure we want to recommend that here

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+
+### SDK object mode
+
+When `convert_query_response` is `true`, Infrahub converts the GraphQL response into `InfrahubNode` SDK objects. These are available via `self.nodes` (a list) and `self.store` (indexed by node ID). You do **not** use the `data` parameter in this mode:
```
BaptisteGi: "`self.store` (indexed by node ID)" > should we keep this?

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+```
+
+### When to use each
+
+| Use raw dict mode (`false`) when | Use SDK object mode (`true`) when |
+|---|---|
+| Dict access is sufficient for your use case | You need SDK features like `.save()`, `.delete()`, or relationship traversal on the queried objects |
+| You want minimal overhead | You prefer cleaner attribute access (`node.name.value` vs `node["name"]["value"]`) |
+| The query returns a flat structure | You want to use `self.store` to look up nodes by ID |
+| You want to use Pydantic models for type-safe parsing | You want the SDK to handle response parsing automatically |
```
BaptisteGi: This table isn't clear to me. I guess to me the main differentiator is wether the expected complexity for the generator is high or not. If yes then protocols + convert on. If no then raw dict.

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+When `false`, the Generator is skipped after merges. This is rare — mainly useful for Generators that exist solely for preview or validation purposes in proposed changes.
+
+### Common combinations
+
+| `execute_in_proposed_change` | `execute_after_merge` | Use case |
+|---|---|---|
+| `true` | `true` | **Default.** Full lifecycle — preview in proposed changes, apply on merge. |
+| `false` | `true` | Skip preview, only run on merge. Use for Generators with external side effects. |
+| `true` | `false` | Preview only — validate in proposed changes but do not apply on merge. Rare. |
+| `false` | `false` | Never runs automatically. Only manual execution via `infrahubctl` or event triggers. Use this for event-driven Generators in a [modular setup](modular-generators). |
```
BaptisteGi:
```suggestion
| `false` | `false` | Never runs automatically. Only manual execution (via `infrahubctl` / UI...) or event triggers. Use this for event-driven Generators in a [modular setup](modular-generators). |
```

docs/docs/topics/generator.mdx
```diff
@@ -45,6 +45,252 @@ Generators can be executed in several ways, depending on your workflow and where
...
+
+When `false`, the Generator is skipped after merges. This is rare — mainly useful for Generators that exist solely for preview or validation purposes in proposed changes.
+
+### Common combinations
+
+| `execute_in_proposed_change` | `execute_after_merge` | Use case |
+|---|---|---|
+| `true` | `true` | **Default.** Full lifecycle — preview in proposed changes, apply on merge. |
+| `false` | `true` | Skip preview, only run on merge. Use for Generators with external side effects. |
+| `true` | `false` | Preview only — validate in proposed changes but do not apply on merge. Rare. |
```
BaptisteGi: This section isn't clear to me, when a generator runs in a branch (through a proposed change or not) the data it generates aren't part of the merge? If yes then it goes beyond the preview

docs/docs/topics/modular-generators.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+Modular Generators follow three principles.
+
+### Each Generator owns one layer
+
+A Generator reads from a group of target objects and creates or modifies objects that belong to that layer. It does not reach into other layers. The fabric Generator creates fabric-level resources; it does not create pod-level or rack-level objects.
+
+### Each Generator validates its upstream dependencies
+
+Since there is no central orchestrator, each Generator must be self-protecting. Before doing its work, a Generator checks that the objects it depends on actually exist and are complete. For example, the pod Generator verifies that the expected number of super spine switches are present before creating spine switches and wiring them up. If the upstream layer isn't ready, the Generator fails with a clear error rather than producing incomplete data.
```
BaptisteGi, line 58: imo not only it validates, it also builds upstream connections (links, p2p ...)

docs/docs/topics/generator.mdx
wvandeun: It is not specific to generators, you are right. I can't come up with a use case either, but it is possible.

docs/docs/topics/generator.mdx
wvandeun: What we mean here is the Pydantic models you can generate from GraphQL responses. I'm not sure if it is documented atm

docs/docs/topics/generator.mdx
wvandeun: I think it's ok like this

docs/docs/topics/modular-generators.mdx
wvandeun: True, but this section just explains that it needs to validate the success of the upstream generator (if any)

dgarros (approved):
Pls check the SDK commit before merging, not sure what are the changes in the SDK related to this PR

wvandeun:
> Pls check the SDK commit before merging, not sure what are the changes in the SDK related to this PR

For reference: it's the changes in this PR https://github.com/opsmill/infrahub-sdk-python/pull/905

Updating some descriptions in the InfrahubGeneratorDefintionConfig class, which needed to be brought into the infrahub repository to be able to regenerate some of the documentation.

PR #8742, opened 2026-03-30
feat: add health check endpoint at /api/health for load balancer and K8s probes

BeArchiTek:
## Summary

Delivers JPD-117 (and JPD-145) in two parts: an unauthenticated `/api/health` endpoint for load balancers and Kubernetes probes, and — new in this revision — surfacing that same backing-service health inside Infrahub's **anonymous telemetry**, so OpsMill can spot a degraded deployment remotely.

### 1. `/api/health` endpoint

Unauthenticated `GET /api/health` that concurrently probes the services Infrahub needs to serve traffic — database, message bus, cache, task manager, and a new `task_manager_db` (the Prefect Postgres store, probed with a short-lived `SELECT 1` via `PREFECT_API_DATABASE_CONNECTION_URL`). Returns `200` when all are up and `503` when any fail, with a per-dependency breakdown and **categorized error reasons only** (no internal details). Per-check timeout via `INFRAHUB_HEALTH_CHECK_TIMEOUT` (default `3s`). Built around `HealthChecker` + a pluggable `HealthStatusEvaluator`, with `is_healthy()` added to the database/cache/message-bus/workflow adapters. Excluded from metrics; OpenAPI and frontend REST types regenerated.

### 2. Backing-service health in anonymous telemetry (new)

Every anonymous telemetry payload now carries a point-in-time `health` section — an overall status plus the same five per-dependency results the endpoint reports.

- A shared `gather_dependency_health` core is extracted in `health.py` and consumed by **both** the endpoint and the telemetry gather, so the dependency set and status semantics can never drift between them.
- New `TelemetryHealthData` model and an optional `health` field on `TelemetryData`; the payload-format version is bumped to flag the additive field.
- **Resilient by design:** a failure while gathering health omits the section and never blocks the rest of the telemetry payload or its send. Honors the existing telemetry opt-out (recorded locally, not transmitted when opted out).
- Reuses the endpoint's per-check timeout — no new config knob.
- Aggregated/historical health (per-dependency uptime, failure counts over an interval) is intentionally **out of scope** and left to a follow-up.

Spec-kit design artifacts live under `dev/specs/infp-117-health-telemetry/`.

## Testing

- Unit: shared probe core (up / down / error / timeout), the telemetry gather + builder, failure-path resilience (health failure → payload still sent), a drift guard (telemetry dependency set == endpoint set), and a no-internal-details serialization check. No mocks — adapter probes per the repo testing rules.
- Existing endpoint unit/functional tests unchanged and green; full telemetry suite green.
- `ruff` + `mypy` clean; `docs.validate` reports no generated-file drift.


fatih-acar:
Will do a more thorough review later but it would be good to also include the task manager's health status.
End goal would be to have a "status"/"healthcheck" page on the frontend with each dependencies' health. (this has been requested a few times)

backend/infrahub/api/health.py
```diff
@@ -0,0 +1,145 @@
...
+
+router = APIRouter()
+
+
+@router.get("/health", response_model=HealthResponse, responses={503: {"model": HealthResponse}})
+async def health(request: Request) -> Response:
+    try:
+        service: InfrahubServices = request.app.state.service
+        db: InfrahubDatabase = request.app.state.db
+    except AttributeError:
```
ogenstad: I don't think there's a scenario where we would reach this attribute error unless we introduced new code changes and a thing like that should be caught by tests. As such we can remove this try/except cause.

backend/infrahub/server.py
```diff
@@ -147,6 +148,7 @@ def server_request_hook(span: Span, scope: dict) -> None:  # noqa: ARG001
 logger.handlers = gunicorn_logger.handlers
 
 app.include_router(api)
+app.include_router(health_router)
```
ogenstad: We'd want the health router to be part of the main API instead so that the endpoint is `/api/health` instead of `/health`, where the latter might be a future page on the frontend.

backend/infrahub/api/health.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+    checks = await get_health_checks(service=service, db=db)
+    status = determine_status(checks)
+
+    response_data = HealthResponse(
+        status=status,
+        checks=checks,
+        timestamp=datetime.now(tz=UTC).isoformat(),
+    )
```
ogenstad: While we don't follow this everywhere as it is now we really do want the API layer to be very thin and instead of defining all of the logic in this file we should have a new module under `infrahub.health` where we have all of the logic for what goes into a health report, so in this file we might have something like:

```python
from infrahub.health import health_report
```

At the top of the file and then within this API endpoint we just have the call to that method.

backend/infrahub/api/health.py
```diff
@@ -0,0 +1,145 @@
...
+    TIMEOUT = "timeout"
+    CONNECTION_REFUSED = "connection_refused"
+    CONNECTION_CLOSED = "connection_closed"
+    NOT_INITIALIZED = "not_initialized"
+    UNKNOWN_ERROR = "unknown_error"
+
+
+class DependencyHealth(BaseModel):
+    name: DependencyName
+    status: str
```
ogenstad: I think `status` here should be a string enum instead.

backend/infrahub/api/health.py
```diff
@@ -0,0 +1,145 @@
...
+    error: ErrorCategory | None = None
+
+
+class HealthResponse(BaseModel):
+    status: str
```
ogenstad: Same as above.

backend/infrahub/api/health.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def determine_status(checks: list[DependencyHealth]) -> str:
+    """Determine the overall health status from individual dependency checks.
+
+    Returns "healthy" when all dependencies are up, "unhealthy" otherwise.
+    Enterprise deployments can override this to support "degraded" states."""
+    if all(check.status == "up" for check in checks):
+        return "healthy"
+    return "unhealthy"
```
ogenstad: Once moved to `infrahub.health` I think this function should be part of the health report we could perhaps have a class that does all of the calculation and the return value for something like this should be an enum with known values instead of a string.

backend/tests/unit/api/test_health.py
```diff
@@ -0,0 +1,274 @@
+from __future__ import annotations
```
ogenstad: Once the business logic for the health checks move away from the API we can also move these tests to match with the new location.

backend/tests/unit/api/test_health.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def test_unknown_error(self) -> None:
+        assert _classify_error(RuntimeError("something")) == ErrorCategory.UNKNOWN_ERROR
+
+    def test_value_error(self) -> None:
+        assert _classify_error(ValueError("bad")) == ErrorCategory.UNKNOWN_ERROR
+
+
+class TestCheckDatabase:
+    @pytest.mark.anyio
```
ogenstad: We shouldn't need any of these pytest markers.

backend/tests/unit/api/test_health.py
```diff
@@ -0,0 +1,274 @@
...
+        assert _classify_error(ConnectionResetError()) == ErrorCategory.CONNECTION_REFUSED
+
+    def test_os_error(self) -> None:
+        assert _classify_error(OSError("connection failed")) == ErrorCategory.CONNECTION_REFUSED
+
+    def test_unknown_error(self) -> None:
+        assert _classify_error(RuntimeError("something")) == ErrorCategory.UNKNOWN_ERROR
+
+    def test_value_error(self) -> None:
+        assert _classify_error(ValueError("bad")) == ErrorCategory.UNKNOWN_ERROR
```
ogenstad: In general in this file if we don't use anything from the classes it would be cleaner to have the tests as functions to avoid excessive indentation. If we because of that want to rename some of the tests to add some additional context that's fine.

For the tests in this class it looks like a good candidate to use pytest params and we can define test classes with a custom dataclass as we do in other similar tests.

BeArchiTek:
@fatih-acar I added the task-manager (not the task-worker)
@ogenstad I moved to /api/health instead of /health, we can implement a frontend page in another PR

backend/infrahub/health/__init__.py
```diff
@@ -0,0 +1,148 @@
+from __future__ import annotations
```
ogenstad: As we only have one file for everything as it is now, while it might grow in the future for now I'd suggest to ditch the folder `health` and just have a `health.py` in the root of the repo.

backend/infrahub/services/adapters/message_bus/rabbitmq.py
```diff
@@ -118,6 +118,12 @@ async def _initialize_connection(self) -> None:
     async def shutdown(self) -> None:
         await self.connection.close()
 
+    async def is_healthy(self) -> bool:
```
ogenstad, line 92: This check doesn't seem to work. Even if I kill the RabbitMQ container it reports that everything is fine.

backend/infrahub/health/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+class OverallStatus(StrEnum):
+    HEALTHY = "healthy"
+    UNHEALTHY = "unhealthy"
+
+
+class DependencyHealth(BaseModel):
+    name: DependencyName
+    status: DependencyStatus
+    error: ErrorCategory | None = None
```
ogenstad: Potentially we could add a NONE value to the `ErrorCategory` enum instead, that would enable a caller to do a switch/match statement on the entire object instead of first checking if there was a response.

ogenstad (commented):
The structure looks good overall to me now. An issue with the actual error messages though is that if I kill all of the other containers all of the checks report UNKNOWN_ERROR, it could be that we're missing something there. Not sure if TIMEOUT would have been correct there. Could be that this is fine as a first round and that the most important thing is that we can enable the checks in the first place.

The one thing I'm considering is if the "checks" attribute name under HelathResponse is the correct one. It's mostly of relevance with regards to what we might want to add later if anything. One consideration would be to have a proof of life check where we can see the last checkin status of all of the workers (API and Prefect), this would be information that comes from the cache server. Not something that needs to be done in this PR. I'm just considering what to name something like that and if that future name would make "checks" a bit vague.

backend/infrahub/health/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class HealthResponse(BaseModel):
+    status: OverallStatus
+    checks: list[DependencyHealth]
+    timestamp: str
```
ogenstad: I think we can change timestamp to be a `datetime` instead as it is a better indicator to a consumer of the API what to expect here.

backend/tests/unit/health/test_health.py
```diff
@@ -0,0 +1,323 @@
+from __future__ import annotations
+
+import asyncio
+from dataclasses import dataclass
+from unittest.mock import AsyncMock, MagicMock
```
ogenstad: There's a lot of mocking and magic mocking in this file. I'd suggest that we instead use adapters from https://github.com/opsmill/infrahub/tree/stable/backend/tests/adapters. It's fine to create new adapters there if we need them. When we use things like MagicMock it can be hard to tell what's actually being tested or if what we test will actually end up working in the real world.

backend/infrahub/database/__init__.py
```diff
@@ -311,6 +311,15 @@ async def __aexit__(
     async def close(self) -> None:
         await self._driver.close()
 
+    async def is_healthy(self) -> bool:
+        """Check if the database is reachable by executing a lightweight query."""
+        try:
+            async with self.start_session(read_only=True) as db:
```
fatih-acar: I think we also need to test with `read_only=False` and/or a transaction; this is because read_only=True can make the query end up on a neo4j read replica where we also want to test the leader/write node.

To keep it simple we can probably just check if the leader/write node is healthy (and not use read_only=True at all)

backend/infrahub/health/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+        healthy = await asyncio.wait_for(workflow.is_healthy(), timeout=config.SETTINGS.health.check_timeout)
+        if healthy:
+            return DependencyHealth(name=DependencyName.TASK_MANAGER, status=DependencyStatus.UP)
+        return DependencyHealth(
+            name=DependencyName.TASK_MANAGER, status=DependencyStatus.DOWN, error=ErrorCategory.UNKNOWN_ERROR
+        )
+    except Exception as exc:
+        return DependencyHealth(
+            name=DependencyName.TASK_MANAGER, status=DependencyStatus.DOWN, error=classify_error(exc)
+        )
```
fatih-acar: Anyway to refactor the check_* functions? They all seem similar

backend/infrahub/services/adapters/message_bus/rabbitmq.py
fatih-acar: Indeed, a better healthcheck would be to try to write to RMQ (like we do with neo4j)

ogenstad:
Some comments by Claude.

## Review: `/api/health` + backing-service health in telemetry

The design is clean. The shared `gather_dependency_health` core with a drift-guard test is a good way to keep the endpoint and telemetry in lock-step, and the per-dependency timeout/catch boundary is well thought out. Issues below, ranked.

### Notable

**1. The error-category taxonomy is largely dead in production** (`backend/infrahub/health.py:62`)

`classify_error` only runs when a probe *raises*. But every real adapter catches its own exceptions and returns `False`:

- `database/__init__.py` `is_healthy` → `except Exception: return False`
- `cache/redis.py`, `cache/nats.py`, `message_bus/rabbitmq.py`, `workflow/worker.py` — all the same pattern.

When a probe returns `False`, `check_dependency` reports `error=UNKNOWN_ERROR`. So in practice the only categories that can ever appear for these four dependencies are `unknown_error` (down) and `timeout` (from the `asyncio.wait_for` wrapper). A Redis `ConnectionError` or a refused RabbitMQ channel both surface as `unknown_error`, never `connection_refused`. Only `task_manager_db` (which raises) and timeouts exercise the taxonomy. The unit tests mask this because `FailingProbe` raises, which the real adapters never do.

Either let the adapters propagate so `classify_error` can do its job, or accept that categories are best-effort and trim the enum. As written, the categorization mostly collapses to `unknown_error`.

**2. Unauthenticated endpoint opens a fresh Postgres engine + connection per call** (`backend/infrahub/health.py:182`)

`probe_task_manager_db` does `create_async_engine(...)` → connect → `dispose()` on every invocation. The Neo4j probe reuses the pooled driver via `start_session()`, but this one builds a brand-new engine each time. `/api/health` is unauthenticated and is exactly the endpoint a load balancer + K8s liveness/readiness probes hit every few seconds across every node — a new PG connection per probe with no pooling, and an anonymous caller can amplify it into connection-pressure on the task-manager DB. Consider a module-level cached engine or a lighter reachability check.

### Maintainability

**3. Inconsistent abstract-base contract for `is_healthy`**

`cache/__init__.py` declares `is_healthy` as `@abstractmethod` (a missing override fails at instantiation), but `message_bus/__init__.py` and `workflow/__init__.py` provide a `return False` default. A future message-bus or workflow adapter that forgets to override silently reports the deployment permanently unhealthy with no signal. Pick one policy — making all three `@abstractmethod` is the safer, more consistent choice.

**4. `ErrorCategory.CONNECTION_CLOSED` is never produced; `ConnectionResetError`/`OSError` are mislabeled** (`backend/infrahub/health.py:62-69`)

`classify_error` never returns `CONNECTION_CLOSED` (it only appears hand-built in test fixtures) — dead enum value. And it maps `ConnectionResetError` and any generic `OSError` to `CONNECTION_REFUSED`, which is semantically wrong (a reset or a broken pipe is not a refusal).

**5. Over-engineered DI indirection for the task-manager-db probe** (`backend/infrahub/workers/dependencies.py:146`)

`build_task_manager_db_probe` + `@inject get_task_manager_db_probe` is two functions and a `fast_depends` injection whose only job is to hand back a reference to the module-level `probe_task_manager_db` — which has no dependencies and reads its config from `os.environ`. A direct import would be clearer.

### Minor

- **Changelog has the wrong path:** `changelog/8508.added.md` advertises the endpoint at `/health`, but it's mounted at `/api/health`. The three fragments also mix naming conventions (`8508.added.md` vs `+health-telemetry.added.md`).
- **Two spec directories committed:** both `dev/specs/001-postgres-health-check/` and `dev/specs/infp-117-health-telemetry/`. The `001-` one looks like a superseded scaffold — drop it if leftover.
- **`CLAUDE.md` now hard-points at a transient plan** (`dev/specs/infp-117-health-telemetry/plan.md`). That pointer goes stale the moment this merges and the next feature starts.

ogenstad (commented):
One thing that stands out is `probe_task_manager_db` which seems to assume that the webserver has direct access to the Prefect database and can authenticate to the Prefect database. I don't think we should assume that this is the case. Looking at it from a security perspective I wouldn't expect that the webserver has these credentials at hand. It might be better to delegate that check and do something through the Prefect API instead.

changelog/8508.added.md
```diff
@@ -0,0 +1 @@
+Added a `/api/health` endpoint for load balancer and Kubernetes probe integration. It reports a per-dependency status for the backing services Infrahub relies on — database, message bus, cache, task manager, and the task manager's backing store (`task_manager_db`, checked through the task manager's readiness endpoint) — so a backing-store outage is pinpointed instead of only surfacing as a degraded task manager. Anonymous telemetry now includes a point-in-time snapshot of the same health information, so a degraded deployment can be observed remotely; a failure while gathering health never prevents the rest of the telemetry from being sent.
```
ogenstad, line 1: The filename for this looks incorrect. It should be `5433`?

PR #8745, opened 2026-03-30
IFC-2274: Update login/logout event graphql fields

solababs:
# Why

This PR implements login and logout activity events, making authentication actions queryable via the existing event interface, accessible only to admins.

## How to test

Login then query events via GraphQL (requires admin credentials):

```graphql
query MyQuery {
  InfrahubEvent {
    edges {
      node {
        event
        ... on AccountLoggedInEventType {
          id
          account_id
          account_name
          account_type
          auth_method
          branch
          client_ip
          event
          has_children
          level
          occurred_at
          parent_id
          payload
          primary_node {
            id
            kind
          }
          related_nodes {
            id
            kind
          }
          session_id
          identity_source
          timestamp
          user_agent
          groups
          roles
          kind
        }
        ... on AccountLoggedOutEventType {
          id
          account_id
          account_name
          branch
          client_ip
          event
          has_children
          level
          logout_type
          occurred_at
          parent_id
          payload
          primary_node {
            id
            kind
          }
          related_nodes {
            id
            kind
          }
          session_id
          timestamp
          user_agent
          kind
        }
      }
    }
  }
}
```

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Standardized login/logout events for IFC-2274. Added `kind` and `identity_source`, enriched resources/related nodes, and ensured logout always clears cookies.

- **Refactors**
  - Events/GraphQL: added required `kind` to login/logout; replaced `sso_provider` with `identity_source` (provider name for OAuth2/OIDC); updated `schema.graphql`, docs/specs, and frontend typings.
  - Auth: `AuthResult` now includes `kind`; groups/roles are `{id: name}`; fetch helpers return the same.
  - Event payloads/resources: set account as primary node; added `infrahub.node.{id,kind}` and `infrahub.account.*` (incl. `timestamp`, `client_ip`, `user_agent`, `identity_source`); groups/roles emitted as related nodes using `InfrahubKind`.
  - Processing/Logout: task manager parses timestamps and flattens related groups/roles to names; logout always deletes cookies and emits logout event with account `kind`.

<sup>Written for commit 6b7bd160bd32108a5130d00970ee2c1cc7d598b7. Summary will update on new commits.</sup>


backend/infrahub/api/event_builder.py
```diff
@@ -38,10 +39,11 @@ def make_login_event(
 
 
 def make_logout_event(
```
ajtmccarty, line 41: looks like this one might be able to accept `account` as a parameter instead of `account_id`, `account_name`, and `kind`, but I'm not sure if there are other logout events that could be added in the future that couldn't include an `account` argument
if the current arguments stay, perhaps `kind` should be `account_kind`? I don't think it's obvious what `kind` would mean in this function

backend/tests/component/core/test_account.py
```diff
@@ -199,8 +199,8 @@ async def test_fetch_account_groups_and_roles(
 
     groups, roles = await fetch_account_groups_and_roles(db=db, account_id=account.id)
 
-    assert sorted(groups) == ["admins", "readers"]
-    assert sorted(roles) == ["admin-role", "read-role"]
+    assert groups == [{group2.get_id(): "admins"}, {group1.get_id(): "readers"}]
+    assert roles == [{role2.get_id(): "admin-role"}, {role1.get_id(): "read-role"}]
```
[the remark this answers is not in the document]
ajtmccarty: I think the robot has a point here

ajtmccarty (approved):

backend/infrahub/api/auth.py
```diff
@@ -99,7 +99,7 @@ async def logout(
             event = make_logout_event(
                 request=request,
                 event_meta=await make_event_meta(account_session=user_session, branch=branch),
-                kind=account.get_kind() if account else "Unknown",
+                acount_kind=account.get_kind() if account else "Unknown",
```
ajtmccarty: missing a `c` in `account`

backend/infrahub/events/account_action.py
```diff
@@ -28,13 +28,14 @@ class AccountLoggedInEvent(InfrahubEvent):
 
     event_name: ClassVar[str] = f"{EVENT_NAMESPACE}.account.logged_in"
 
+    kind: str = Field(..., description="The type of object")
```
ajtmccarty: I'm not the most familiar with our prefect integration, but wondering if this could also be `account_kind` b/c it seems like it will always be the kind of the associated account for an `AccountLoggedIn\OutEvent`
or maybe just the description could make that clear?

ajtmccarty (approved):
not an expert on our Prefect stuff, but looks good to me

backend/infrahub/api/auth.py
```diff
@@ -99,6 +99,7 @@ async def logout(
             event = make_logout_event(
                 request=request,
                 event_meta=await make_event_meta(account_session=user_session, branch=branch),
+                account_kind=account.get_kind() if account else "Unknown",
```
ogenstad: I wonder if we should use "raise_for_error" when looking for the account on line 93 so that we know that we'll always get back an account?

ogenstad (commented):
LGTM. One thing though is that the "sso_provider" field actually doesn't reflect the name of the sso_provider only the type of provider and this is just redundant information from "auth_method". I think what we'd want here is the actual name of the provider. This will typically be "provider1", "provider2" or "google". Also since we're looking to add LDAP support in the future and LDAP isn't strictly SSO I think it could make sense to rename this field to something else so instead of "sso_provider" it could perhaps be "identity_source" or something like that. What do you think?

backend/infrahub/api/auth.py
```diff
@@ -90,7 +90,9 @@ async def logout(
         invalidated = await invalidate_refresh_token(db=db, token_id=user_session.session_id)
 
     if invalidated:
-        account = await NodeManager.get_one(id=user_session.account_id, db=db, kind=CoreGenericAccount)
+        account = await NodeManager.get_one(
+            id=user_session.account_id, db=db, kind=CoreGenericAccount, raise_on_error=True
+        )
```
ogenstad: With this change we can also change line 96 as account will never be `None`

backend/infrahub/events/account_action.py
```diff
@@ -18,24 +18,20 @@ class AuthMethod(InfrahubStringEnum):
...
     account_name: str = Field(..., description="Username of the account")
     account_type: AccountType = Field(..., description="USER or SCRIPT")
     auth_method: AuthMethod = Field(..., description="How they authenticated")
     session_id: str = Field(..., description="UUID of the session")
-    groups: list[str] = Field(default_factory=list, description="List of group names/IDs")
-    roles: list[str] = Field(default_factory=list, description="List of role names/IDs")
-    sso_provider: SSOProvider | None = Field(default=None, description="SSO provider name (if applicable)")
+    groups: list[dict[str, str]] = Field(default_factory=list, description="List of group names/IDs")
+    roles: list[dict[str, str]] = Field(default_factory=list, description="List of role names/IDs")
+    sso_provider: str | None = Field(default=None, description="SSO provider name (if applicable)")
```
ogenstad: Sorry if this was unclear earlier but I meant that we should use another name than `sso_provider` here since the this attribute name might not work as good with LDAP. Instead I'd suggest that we replace "sso_provider" with "identity_source", then update the other references where sso_provider is used today.

ogenstad (approved):
Thank you for the updates!

PR #8754, opened 2026-03-31
chore: update gitignore for worktrees folders

BeArchiTek:
## Summary by cubic
Ignore Git worktree directories across the repo to prevent accidental commits. Replace the scoped `dev/spec-kitty/.worktrees/` rule with a single `**/.worktrees/` pattern, and remove the duplicate `.ruff_cache` entry in favor of `**/.ruff_cache`.

<sup>Written for commit bce2c6b2a678c074f1506e6ba9416784f5195aa6. Summary will update on new commits.</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,5 +49,6 @@ python_testcontainers/infrahub_testcontainers.egg-info/*
 
 schema/node_modules/*
 
-# Spec-kitty
-dev/spec-kitty/.worktrees/
+# AI and SDD
+.worktrees/
+**/.worktrees/
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ development/docker-compose.dev-override.yml
 .DS_Store
 .python-version
 .venv/*
-.ruff_cache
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -49,5 +49,6 @@ python_testcontainers/infrahub_testcontainers.egg-info/*
 
 schema/node_modules/*
 
-# Spec-kitty
-dev/spec-kitty/.worktrees/
+# AI and SDD
+.worktrees/
+**/.worktrees/
```
polmichel, line 52: `**/.worktrees` already includes `.worktrees/` as `**` stands for "0 or more depth" -> https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html

You can actually test it using the `git check-ignore .worktrees/` command once you removed the first one.

.gitignore
polmichel: This is not related to your PR, but if you agree with my comment, could you also remove this line?

polmichel (approved):
Thanks!

PR #8764, opened 2026-03-31
IFC-2406: node creation local computation

polmichel:
# Why

During the development of the Jinja2 local computation feature [IFC-2273], I worked on the node update path, making all local Jinja2 computed attributes recompute inline instead of dispatching events to Prefect for asynchronous computation. 
The goal is to reduce the volume of recomputation events that Prefect has to manage.
                                                                                                                                                                        
"Local" here means that the Jinja2 template's dependencies can be resolved from the node itself (its own attributes or its relationship references). If a peer node is updated and that change affects a computed attribute on another node, the recomputation still goes through the async Prefect path.                                   
                                                                                                                                                                        
The node creation path already computed mandatory Jinja2 attributes inline during `Node.new()`. This change extends that to also compute optional Jinja2 attributes inline at creation time and ensure correct dependency ordering for chained computations.

Closes [IFC-2420]

## What changed

During node creation, `_process_fields `now collects all Jinja2 computed attributes (mandatory and optional, local and remote-referencing) in dependency order. Previously, only mandatory Jinja2 attributes were collected and computed inline, optional ones were left for the async Prefect path.  
                                                                                                                                                                 
The validation logic in `_process_fields` has been extracted into dedicated methods:                                                                                    
- `_validate_mandatory_attributes`: validates mandatory attributes are provided, skipping Jinja2 computed and NumberPool attributes
- `_validate_mandatory_relationships`: validates mandatory relationships are provided                                                                                   
                                                                                   
`  _process_macros` now iterates _computed_jinja2_attributes directly, since the list is already dependency-ordered by the registry, removing the need for a second  registry lookup.        


## How to review

Commits are unitary.

## How to test

```bash
uv run pytest tests/component/computed_attribute/test_local_computation.py -xvs
```

## Checklist

- [x] Tests added/updated


[IFC-2420]: https://opsmill.atlassian.net/browse/IFC-2420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[IFC-2273]: https://opsmill.atlassian.net/browse/IFC-2273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -564,15 +564,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
                         ValidationError({mandatory_rel: f"{mandatory_rel} is mandatory for {self.get_kind()}"})
                     )
 
-            if self._schema.is_node_schema:
-                schema_branch = db.schema.get_schema_branch(self._branch.name)
-                local_targets = schema_branch.computed_attributes.get_local_jinja2_targets(kind=self._schema.kind)
-                for target in local_targets:
-                    if (
-                        target.attribute.name not in fields
-                        and target.attribute.name not in self._computed_jinja2_attributes
-                    ):
-                        self._computed_jinja2_attributes.append(target.attribute.name)
+            self._computed_jinja2_attributes.extend(self._get_optional_local_jinja2_attributes(db=db, fields=fields))
 
         if errors:
             raise ValidationError(errors)
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -564,6 +564,16 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
                         ValidationError({mandatory_rel: f"{mandatory_rel} is mandatory for {self.get_kind()}"})
                     )
 
+            if self._schema.is_node_schema:
+                schema_branch = db.schema.get_schema_branch(self._branch.name)
+                local_targets = schema_branch.computed_attributes.get_local_jinja2_targets(kind=self._schema.kind)
+                for target in local_targets:
+                    if (
+                        target.attribute.name not in fields
+                        and target.attribute.name not in self._computed_jinja2_attributes
+                    ):
+                        self._computed_jinja2_attributes.append(target.attribute.name)
+
         if errors:
             raise ValidationError(errors)
 
@@ -680,8 +690,14 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
 
     async def _process_macros(self, db: InfrahubDatabase) -> None:
         schema_branch = db.schema.get_schema_branch(self._branch.name)
+
+        local_targets = schema_branch.computed_attributes.get_local_jinja2_targets(kind=self._schema.kind)
+        ordered_macros = [
+            t.attribute.name for t in local_targets if t.attribute.name in self._computed_jinja2_attributes
+        ]
+
         errors = []
-        for macro in self._computed_jinja2_attributes:
+        for macro in ordered_macros:
             attr_schema = self._schema.get_attribute(name=macro)
             if not attr_schema.computed_attribute:
                 errors.append(
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -539,13 +539,16 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
                 self._profile_provided_rels,
             ) = await self._get_profile_provided_mandatory_fields(db=db, fields=fields)
 
-            self._validate_mandatory_relationships(fields=fields, errors=errors)
+            errors.extend(self._validate_mandatory_attributes(fields=fields))
+            errors.extend(self._validate_mandatory_relationships(fields=fields))
 
-            self._computed_jinja2_attributes.extend(self._get_mandatory_jinja2_attributes(fields=fields, errors=errors))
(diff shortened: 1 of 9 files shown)


backend/infrahub/core/node/__init__.py
```diff
@@ -540,29 +539,10 @@ async def _process_fields(self, fields: dict, db: InfrahubDatabase, process_pool
...
-                            self._computed_jinja2_attributes.append(mandatory_attr)
-                            continue
-
-                        if mandatory_attribute.kind == "NumberPool":
-                            continue
-
-                    errors.append(
-                        ValidationError({mandatory_attr: f"{mandatory_attr} is mandatory for {self.get_kind()}"})
-                    )
+            self._validate_mandatory_relationships(fields=fields, errors=errors)
```
ogenstad: We are doing this in a few other places but we'd want to avoid sending in an object that's supposed to be mutated within another method/function. I.e. in a case like this it would be cleaner if `_validate_mandatory_relationships()` instead returned a list of `ValidationError` and didn't have a param for incoming errors. Instead we'd have something like:

```python
errors: list[ValidationError] = []
....
errors.extend(self._validate_mandatory_relationships(fields=fields)
```

 This would also lead to a change in `_get_mandatory_jinja2_attributes()`

polmichel: This was exactly what I was locally changing, since I do not to return anything more !

polmichel: see https://github.com/opsmill/infrahub/pull/8764/changes/4b580d13fb071760cbe2cc726d3d60cd5caa5e4d

backend/infrahub/core/node/__init__.py
```diff
@@ -540,32 +539,16 @@ async def _process_fields(self, fields: dict, db: InfrahubDatabase, process_pool
...
-                            continue
-
-                        if mandatory_attribute.kind == "NumberPool":
-                            continue
-
-                    errors.append(
-                        ValidationError({mandatory_attr: f"{mandatory_attr} is mandatory for {self.get_kind()}"})
-                    )
+            errors.extend(self._validate_mandatory_attributes(fields=fields))
+            errors.extend(self._validate_mandatory_relationships(fields=fields))
```
ajtmccarty, line 543: good change. easier to read

backend/infrahub/core/node/__init__.py
```diff
@@ -540,32 +539,16 @@ async def _process_fields(self, fields: dict, db: InfrahubDatabase, process_pool
...
-                    )
+            if errors:
+                raise ValidationError(errors)
 
-        if errors:
-            raise ValidationError(errors)
+            if self._schema.is_node_schema:
+                schema_branch = db.schema.get_schema_branch(self._branch.name)
+                local_targets = schema_branch.computed_attributes.get_local_jinja2_targets(kind=self._schema.kind)
+                self._computed_jinja2_attributes = [t.attribute.name for t in local_targets]
```
ajtmccarty, line 551: is it possible for this to happen outside of the `Node` class? I understand if not b/c `Node` is a bit of a mess, but this is a dependency loop where `SchemaBranch` needs to have knowledge of the `Node` class and `Node` needs to use `SchemaBranch` to get these Jinja2 attribute names.

polmichel: EDIT: I had misunderstood your review, you can ignore this comment

To be honest, I don't know. At first, I thought it was unnecessary, but if you look at the `_validate_mandatory_attributes` method, you will see that there is already a similar check on line 573.
When I looked at this boolean, I noticed that other things that override this method are Profiles or Templates, not just nodes.

Templates seem not to reach this part of the code since 
`infrahub.core.schema.attribute_schema.AttributeSchema.support_templates` remove all read-only attributes, and computed attributes are forced to be read-only (`infrahub.core.schema.schema_branch.SchemaBranch._validate_computed_attribute`)

The remaining question would be about profiles. I will take a look at those.

I should have raised this as an open question in the PR, good catch.

polmichel: EDIT: the check at least makes the intent clear and avoid useless computation. But let's discuss it after the dependencies issues, maybe this would not be an issue anymore.

Same conditions around `support_profiles` method `infrahub.core.schema.attribute_schema.AttributeSchema.support_profiles`.
Read-only are excluded. I will remove this check and the check on line 573.

polmichel: So, let's get back at your initial feedback

I need to access
`infrahub.core.schema.schema_branch_computed.facade.ComputedAttributes.get_local_jinja2_targets` method to have the dependency-ordered collection of Jinja2 attributes.
`ComputedAttributes` is a registry in `SchemaBranch` class.

This registry is filled by `infrahub.core.schema.schema_branch_computed.facade.ComputedAttributes.register_computed_jinja2` during schema validation process`infrahub.core.schema.schema_branch.SchemaBranch._validate_computed_attribute`. 
Moreover, the `schema_branch` object is called in `infrahub.core.node.Node._process_macros` just after in the node creation process, so at least these additional lines are not making the dependency problem worst.

If we want to remove this dependency, I feel that it would have a wide impact.

ajtmccarty: yes I see it used in `_process_macros`
I guess this is just another item to handle with the Grand Node Refactoring when we get to it

ajtmccarty (approved):

Issue #1945, opened 2024-01-16
feature: Add support for Float kind of Attribute or edit Number 


BeArchiTek:
### Component

API Server / GraphQL

### Describe the Feature Request

Add a new kind of Attribute to support Float

### Describe the Use Case

Be able to store float type of number

Example 1: a device is 7,7 kg, I would like to be able to set 7,7 kg and not be force to either set 8 kg or 770 gr
Example 2: a device is 1.5 U of height, I would like to be able to have to put 2U. (on a 48U rack it's 32 devices instead of 24)

ryanmerolle:
This would be useful to better model data.  Number is not a type I have ever seen before in a system like this.

PR #8765, opened 2026-04-01
WIP: Add Float support

BeArchiTek:
Implements #1945 

---
## Summary by cubic
Adds a new `Float` attribute kind across backend, GraphQL, and frontend to store decimal numbers with min/max constraints and full validation, filtering, and UI support. Implements the schema and client changes required by INFP-175.

- **New Features**
  - Backend: `Float` attribute with strict validation (coerce int, reject bool/NaN/∞) and range checks via `FloatAttributeParameters`; filters accept floats (including JSON-list matching).
  - GraphQL: `FloatAttribute` type plus `FloatAttributeCreate`/`FloatAttributeUpdate`; filtering via Float; regenerated schema/introspection.
  - Types/REST: added `Float` to `infrahub` types and OpenAPI (`FloatAttributeParameters`).
  - Frontend: new `Float` form field and table rendering; dynamic filter uses `step="any"`; client-side min/max validation; number validators now treat empty optional values as `null` and correctly handle `0` for min/max.

- **Migration**
  - Declare attributes as `kind: "Float"` with optional `parameters.min_value` / `parameters.max_value`.
  - Changing `Number` (integer) to `Float` is safe; changing `Float` to `Number` will fail if any fractional values exist.
  - Update any queries/UI expecting integers to accept floats where relevant.
  - Optional `Number` inputs now return `null` when empty (was `0`); update any logic that relied on a default zero.

<sup>Written for commit 11f24ecaa180972e807a3856bf136849abb66d68. Summary will update on new commits.</sup>


petercrocker:
Would be good to get this across the line! When pulling in object templates that define hardware models and converting their weight from lbs to kg it creates huge rounding errors until we can have a float: https://github.com/opsmill/infrahub-skills/pull/80

iddocohen:
Another use case, from the optical side. We model DWDM/WDM in `opsmill/schema-library`, and there the missing float kind is a blocker rather than a rounding annoyance.

`Number` is exposed as GraphQL `BigInt`, so a fractional value is not rounded, it is rejected outright:

```
Expected value of type 'BigInt', found 193.1.
```

That specific number matters: the ITU-T G.694.1 grid is *defined* fractionally, `f = 193.1 + n x 0.1 THz`. 193.1 THz is not an awkward measurement someone took, it is the anchor frequency of the standard. Every datasheet, vendor CLI and optical engineer writes it that way.

To ship at all, we rescaled every optical quantity to an exact integer and encoded the unit in the attribute name:

| domain value | stored as |
| --- | --- |
| 193.1 THz | `frequency_ghz` = 193100 |
| 1560.61 nm | `wavelength_pm` = 1560610 |
| 17.5 dB gain | `gain_ddb` = 175 |
| 5.5 dB noise figure | `noise_figure_ddb` = 55 |

Wavelength is the clearest case: lambda = 299792.458 / f, so it is inherently fractional. And "tenths of a dB" (`ddb`) is a unit that appears in no datasheet anywhere. We invented it purely to get around the integer constraint.

What that costs us:

- The unit is a naming convention, not schema-enforced. Nothing stops someone writing nm into `wavelength_pm`.
- Every consumer divides. Transforms, checks, generators and any UI all have to know the scale factor.
- Operators read `192100` in the UI while the equipment in front of them says `192.1`.

The precision actually needed is modest. The 50 GHz grid means 0.05 THz steps, and flex-grid goes down to 6.25 GHz granularity, so float64 covers it with room to spare. The `min_value` / `max_value` parameters in this PR would also be genuinely useful here, for example constraining a frequency to the C-band.

One design note, offered as a caution rather than an objection. If `Float` becomes usable in `uniqueness_constraints` or as a join key, binary float equality is a trap: 193.1 has no exact float64 representation, so two code paths computing "the same" frequency can compare unequal. We sidestep this by joining on a shared grid-point node rather than on the number itself, but sooner or later someone will put a float in a uniqueness constraint. Worth deciding deliberately whether that is supported, and if so whether the comparison is normalised.

Context, including the rescale and the reasoning behind it: https://github.com/opsmill/schema-library/pull/77

PR #8774, opened 2026-04-01
Store external identity in internal nodes at login time for SSO

ogenstad:
# Why

Closes 

## What changed

-

## How to test

```bash
# Commands to validate
```

## Impact & rollout

- **Backward compatibility:** 
- **Performance:** 
- **Config/env changes:** 
- **Deployment notes:** 

## Checklist

- [ ] Tests added/updated

---
## Summary by cubic
Persist external SSO identities at login by creating `InternalExternalIdentity` nodes and linking them to accounts. This anchors account mapping to the provider `sub` and hardens login, linking, and group sync.

- New Features
  - Added `InternalExternalIdentity` and `external_identities` on accounts; exposed in GraphQL.
  - `oauth2`/`oidc` token handlers now require `sub`; build an `ExternalIdentity`; return a clear error if missing; use `ExternalAuthProtocol`.
  - Login flow resolves by identity; creates or links on first login; updates account label only when changed; falls back to email on name collision; concurrency-safe to prevent duplicate accounts.
  - Group sync matches `AccountGroup`, ignores unknowns, deduplicates, and applies the configured default group when none are provided by the IdP.

- Migration
  - Ensure IdP scopes include `openid` so `sub` is returned.
  - Existing accounts without identities are auto-linked on first login if the display name matches and the account is unclaimed.

<sup>Written for commit d8e19cbcb1ce603c281f41406dc9a2093b30fed9. Summary will update on new commits.</sup>


backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+            "protocol__value": external_identity.protocol,
+        },
+        prefetch_relationships=True,
+        limit=1,
+    )
+
+    if identity_nodes:
+        identity_node = identity_nodes[0]
+        account = await identity_node.account.get_peer(db=db)
+        if account.label.value != external_identity.display_name:
```
polmichel, line 174: Would `account.get_attribute("label").get_value()` be better ?

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+                    identity_node = await Node.init(db=db, schema=InfrahubKind.EXTERNALIDENTITY)
+                    await identity_node.new(
+                        db=db,
+                        sub=external_identity.sub,
+                        provider_name=external_identity.provider_name,
+                        protocol=external_identity.protocol,
+                        account=account.id,
+                    )
+                    await identity_node.save(db=db)
+                    if account.label.value != external_identity.display_name:
```
polmichel, line 206: Same -> Would `account.get_attribute("label").get_value()` be better ?

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
     return models.AccessTokenResponse(access_token=access_token)
 
 
-async def signin_sso_account(db: InfrahubDatabase, account_name: str, sso_groups: list[str]) -> models.UserToken:
-    account = await NodeManager.get_one_by_default_filter(db=db, id=account_name, kind=InfrahubKind.ACCOUNT)
+async def signin_sso_account(
```
polmichel, line 154: I feel that this method has become rather long and complex. The following suggestions aim to make this decision graph easier to read.
Splitting the method into sub-blocks would also be helpful if we want to remove some parts of it later on (once all identifications have been synchronised with existing accounts, for example).

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+                    if existing_by_email:
+                        raise ProcessingError(
+                            message=(
+                                f"Cannot create account: both '{external_identity.display_name}'"
+                                f" and '{external_identity.email}' are already in use as account names."
+                            )
+                        )
+                    account_name = external_identity.email
+                else:
+                    account_name = external_identity.display_name
```
polmichel, line 224: This could be `account_name = _get_available_account_name(db, external_identity)`

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+
+                account = await Node.init(db=db, schema=InfrahubKind.ACCOUNT)
+                await account.new(
+                    db=db,
+                    name=account_name,
+                    label=external_identity.display_name,
+                    account_type="User",
+                    password=str(uuid.uuid4()),
+                )
+                await account.save(db=db)
```
polmichel, line 234: These last lines could be a call to a `_create_account(db, account_name, external_identity.display_name)` method

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+
+                identity_node = await Node.init(db=db, schema=InfrahubKind.EXTERNALIDENTITY)
+                await identity_node.new(
+                    db=db,
+                    sub=external_identity.sub,
+                    provider_name=external_identity.provider_name,
+                    protocol=external_identity.protocol,
+                    account=account.id,
+                )
+                await identity_node.save(db=db)
```
polmichel, line 244: These last lines could be a call to a `_create_identity(db, account_id, external_identity)` method

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+        prefetch_relationships=True,
+        limit=1,
+    )
+
+    if identity_nodes:
+        identity_node = identity_nodes[0]
+        account = await identity_node.account.get_peer(db=db)
+        if account.label.value != external_identity.display_name:
+            account.label.value = external_identity.display_name
+            await account.save(db=db)
```
polmichel, line 176: This block looks like it could be refactored into a `sync_with_existing_account` with a docstring indicating this method is not meant to be there forever

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+                        db=db,
+                        sub=external_identity.sub,
+                        provider_name=external_identity.provider_name,
+                        protocol=external_identity.protocol,
+                        account=account.id,
+                    )
+                    await identity_node.save(db=db)
+                    if account.label.value != external_identity.display_name:
+                        account.label.value = external_identity.display_name
+                        await account.save(db=db)
```
polmichel, line 208: Same method than upper `_sync_with_existing_account`

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+                    account = account_by_name
+                    identity_node = await Node.init(db=db, schema=InfrahubKind.EXTERNALIDENTITY)
+                    await identity_node.new(
+                        db=db,
+                        sub=external_identity.sub,
+                        provider_name=external_identity.provider_name,
+                        protocol=external_identity.protocol,
+                        account=account.id,
+                    )
+                    await identity_node.save(db=db)
```
polmichel, line 205: `link_identity_to_account(db, account, external_identity)` ?

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -0,0 +1,366 @@
...
+
+    await signin_sso_account(db=db, external_identity=identity, sso_groups=[])
+
+    unchanged_account = await NodeManager.get_one(db=db, id=account.id)
+    assert unchanged_account.label.value == "Dave Wilson"
+
+    identity_nodes = await NodeManager.query(
+        db=db, schema=InfrahubKind.EXTERNALIDENTITY, filters={"sub__value": "sub-nolabel-001"}
+    )
+    assert len(identity_nodes) == 1
```
polmichel, line 148: Is this test checking something valuable ? I think it would need to assert an assertion on the `save` method of the `account` node in order to bring something to the table?

polmichel: Or checking the `updated_at` value of `account` node

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -0,0 +1,366 @@
...
+
+    refreshed_group = await NodeManager.get_one(db=db, id=group.id, prefetch_relationships=True)
+    members = await refreshed_group.members.get_peers(db=db, branch_agnostic=True, peer_type=CoreAccount)
+    assert len(members) == 1
+
+
+async def test_unknown_group_is_silently_ignored(
+    db: InfrahubDatabase, default_branch: Branch, register_core_models_schema: SchemaBranch
+) -> None:
+    """The IdP may send group names that have not yet been created in Infrahub. Failing the
```
polmichel, line 313: I don't know much about authentication methods, but would it be helpful to have a log for this test case to show the system "manager" or whoever tracks the logs that these groups don't exist?

backend/infrahub/auth.py
```diff
@@ -138,13 +151,97 @@ async def create_fresh_access_token(
...
+            if account_by_name:
+                existing_identities = await NodeManager.query(
+                    db=db,
+                    schema=InfrahubKind.EXTERNALIDENTITY,
+                    filters={"account__ids": [account_by_name.id]},
+                    limit=1,
+                )
+                if existing_identities:
+                    # Account belongs to a different SSO user — treat as name collision
+                    name_collision = True
```
polmichel, line 193: This is a personal preference, so obviously you can ignore it, but I think these nested conditions and boolean updates are hard to follow.
If you take the `existing_identities` query at the same indentation level than the `account_by_name` query, it may look simpler.

You can do something like:
```
  account_by_name = await NodeManager.get_one_by_default_filter(
                db=db, id=external_identity.display_name, kind=InfrahubKind.ACCOUNT
            )
  existing_identities =
      await NodeManager.query(
          db=db,
          schema=InfrahubKind.EXTERNALIDENTITY,
          filters={"account__ids": [account_by_name.id]},
          limit=1,
      ) if account_by_name else None

   if account_by_name and not existing_identities:
        # Unclaimed account: true transition case, safe to link
           [...]
   if not account_by_name or existing_identities:
                if existing_identities:
                    existing_by_email = await NodeManager.get_one_by_default_filter(
                        db=db, id=external_identity.email, kind=InfrahubKind.ACCOUNT
                    )
                    if existing_by_email:
                        raise ProcessingError(
                            message=(
                                f"Cannot create account: both '{external_identity.display_name}'"
                                f" and '{external_identity.email}' are already in use as account names."
                            )
                        )
                    account_name = external_identity.email
                else:
                    account_name = external_identity.display_name
```

polmichel (commented):
From a logic perspective, it looks good to me. I think this will do the trick of silently updating and synchronising the new identity objects with the existing accounts. I do not see issues regarding this.
My comments are more about code readability and test cases.

backend/infrahub/auth.py
ogenstad: I think it depends, there are a few issues with the current code around this. First of all NodeManager.query() returns `list[Any]` if we don't send in a protocol so linting tools aren't complaining about anything. If we send in a protocol instead we'd be able to correctly identify the returned nodes for that query but we'd instead run into issues after `get_peer()` as it only returns a node. If we made that change switching to `.get_attribute()` would pass the linters, but it's not really safer as it would raise an error here if we had an invalid attribute name. I.e. it's as unsafe as what we have now. I think the correct approach moving forward is to start to correct the issue in NodeManager.query() and anything else from NodeManager that returns `Any` instead of `Node`. Then we'd need to either rework the internals of how `get_peer()` works within our own schema perhaps by looking at generics (the Python version), so that account.get_peer() would return the correct type. For now i think we can just leave this as it is. Mainly because this is the convention that we typically use. We've replaced other calls like `getattr` with `get_attribute()` which is cleaner. I'm not sure it makes much of a difference here for now.

backend/infrahub/auth.py
ogenstad: I agree, my thought though was that I'd keep changes now to a minimum as the change was targeting the `stable` release directly. But I think we should both consider combining the logic from Oauth2 and OIDC since they share a lot of code as well as cleaning up the code. I'd also like more unittests and actual integration tests for these. Potentially part of the work for introducing LDAP support.

polmichel (approved):
LGTM, other comments could be looked into in a future release.

PR #8775, opened 2026-04-01
IFC-2423: use namespace restriction webhooks

polmichel:
## Why

Today, users can create custom schemas based on generic ones. However, there are some assumptions in the code relating to some of these inherited schemas. This applies to `CoreGenericRepository` and `CoreWebHook`, for example.

To avoid any issues with `CoreWebHook`, we prevent other types than `CoreStandardWebhook` and `CoreCustomWebhook` from being used through conditions in web hook-related processes.
For example, in `backend/infrahub/webhook/tasks/process.py`:

```    
if webhook_kind not in ["CoreStandardWebhook", "CoreCustomWebhook"]:
        raise ValueError(f"Unsupported webhook kind: {webhook_kind}")
```

### Problem

This is not an effective way to manage this limitation; it allows the user to define their own custom, inherited schema before the error message appears. Ideally, we would like to block this directly during the schema validation process. 

## Solution

The namespace restriction feature has been merged (https://github.com/opsmill/infrahub/pull/8319) and can now be used to prevent users from creating custom schemas from generic ones. This is achieved by limiting the allowed namespaces to `Core`. Since this namespace is restricted to the Core package, schemas containing user-customised new schemas inherited from a restricted package cannot be validated.

Closes [IFC-2423]

## What changed

- Adding a new namespace restriction to `core_webhook` generic schema
- Removing useless runtime checks against web hooks types.

### Out-of-scope
- Use `InfrahubKind` variable string instead of "CoreStandardWebhook", "CoreWebhook" and "CoreCustomWebhook"

## How to test

The functionality is tested through namespace restriction feature. This is just a specific implementation of it.
The namespace restriction feature related tests are:
- `backend/tests/component/core/schema_manager/namespace_restriction/test_namespace_restriction.py`
- `backend/tests/component/core/schema_manager/namespace_restriction/test_restricted_namespaces_validation.py`

Having that said, an additional specific test could be added, which would load a schema inheriting from the new restricted generic schema without the namespace "Core", but I do not think this brings any value in the long run.

## Impact & rollout

- **Backward compatibility:** Schema retro-compatibility will **not** be managed in this PR.

Along with this modification, a specific item regarding this behavior should be mentioned in the release note. If a schema already exists inheriting from the generic schema `CoreWebhook`, the command `infrahub upgrade` will return an error message due to the invalidated state of the schema.

If this occurs, the user will need:

1. Delete all the nodes implementing this custom CoreWebhook schema from the database.
2. Importing once again the invalid schema with the `state` parameter with the value `absent`.
3. Triggering once again the `infrahub upgrade` command to make sure everything is in order.

#### State absent example:
```
---
version: "1.0"
nodes:
  - name: MyCustomWebhook
    namespace: Test
    label: "My Custom Webhook"
    inherit_from:
      - CoreWebhook
    generate_profile: false
    state: absent
```

#### Schema-loading command

`infrahubctl schema load my_custom_webhook_schema.yml  `

[IFC-2423]: https://opsmill.atlassian.net/browse/IFC-2423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Enforces a Core-only namespace for webhook schemas to block custom inheritance and removes brittle runtime kind checks. Implements IFC-2423 by validating at the schema level and using `InfrahubKind` for kind handling.

- **New Features**
  - Added `restricted_namespaces=["Core"]` to the `CoreWebhook` generic schema.

- **Refactors**
  - Removed runtime kind whitelist in `webhook/tasks/process.py`; route using `InfrahubKind.STANDARDWEBHOOK`.
  - Updated tests to use `InfrahubKind` constants instead of string literals.

<sup>Written for commit c1466e74dae791ea9cd64d6df8270258c768a17d. Summary will update on new commits.</sup>


ajtmccarty (commented):
code looks good to me, but a couple questions

do we need to do anything to handle node schemas that users may have created based on `CoreWebhook` generic? for example, if I have a `CoolWebhook` node schema that inherits from `CoreWebhook` and I upgrade to the latest version
- would Infrahub fail to start?
- would the webhook logic pick up any instances of `CoolWebhook` and try to use them?

polmichel:
> code looks good to me, but a couple questions
> 
> do we need to do anything to handle node schemas that users may have created based on `CoreWebhook` generic? for example, if I have a `CoolWebhook` node schema that inherits from `CoreWebhook` and I upgrade to the latest version
> 
> * would Infrahub fail to start?
> * would the webhook logic pick up any instances of `CoolWebhook` and try to use them?

As mentioned in the "Impact & rollout" section of the PR description, I have conducted some tests that revealed the upgrade command would not work. However, we do not expect users to have implemented any custom web hooks. Following discussions at last Thursday's daily meeting, a note should be added to the release notes regarding invalid schemas and how to resolve them (which I have started to do in the PR description).

However, I have not considered what would happen if users could launch the application without going through the upgrade process. I will carry out this additional test, but if this is possible, I assume deleting these nodes from the database and then removing these custom types through a migration would be required?

ajtmccarty (approved):

ajtmccarty:
I don't think you need to worry about someone launching an updated version of the application without going through the upgrade process

Issue #8694, opened 2026-03-25
bug: Custom webhooks stop triggering after period of inactivity (~24h)


gbieli:
### Component

Not Sure

### Infrahub version

1.7.4 (community edition)

### Current Behavior

**Description**
Custom webhook for event type "all" cease to trigger automatically after a certain period (observed after approximately 24 hours), even when relevant events like node created or node deleted occur.
The system remains in this "stale" state until the specific webhook is manually disabled and re-enabled in the Web UI. Once toggled, the webhooks begin functioning correctly again immediately.

**Behavior**
- Initial State: Webhooks trigger correctly upon node creation/deletion.
- After ~24 Hours: No tasks are visible in the Task Manager for webhook execution despite node changes being committed.
- Workaround: Deactivating and reactivating the custom webhook via the UI restores functionality.
- Logs: A review of docker compose logs across all Infrahub containers shows no explicit errors, timeouts, or stack traces related to the failure.

**Task Manager Screenshot - success**
<img width="1618" height="545" alt="Image" src="https://github.com/user-attachments/assets/c53a8337-4a77-421e-9d75-d79819077f87" />

**Task Manager Screenshot - fail**
<img width="1613" height="546" alt="Image" src="https://github.com/user-attachments/assets/734ad93a-3d51-4340-a9e1-53956f00c91f" />

**Request for Troubleshooting Guidance**
Since the standard container logs (docker compose logs) are not yielding any errors, I would appreciate guidance on how to provide more diagnostic data. Specifically:

- Which specific service (e.g., tasks-worker, api, or redis) should I monitor more closely?
- What is the recommended method to increase the log verbosity (DEBUG level) for the task execution engine?

Are there specific internal states or queues we should inspect via the CLI when the system enters this "stale" state?

### Expected Behavior

Webhooks should remain active and trigger reliably regardless of the time elapsed since the last configuration change or container restart.

### Steps to Reproduce

**Steps to Reproduce in our infrastructure**
1. Configure a custom webhook in Infrahub for event type all.
2. Verify the webhook is working (Task Manager shows successful execution).
3. Wait for a period of ~24 hours without restarting the containers.
4. Create or delete a node.
5. Observe the Task Manager: No webhook task is triggered.
6. Toggle the webhook (Disable -> Enable).
7. Create another node: The webhook triggers as expected.

### Additional Information

**Environment**
Infrahub Version: 1.7.4 (community edition)
Deployment Method: Docker Compose

[...]

PR #8777, opened 2026-04-01
#8772 webhook match statements lost during scheduled reconfiguration

### LINKED ISSUE #8772: bug: Match statements lost during scheduled reconfiguration of webhooks
### Component

_No response_

### Infrahub version

v1.8.3

### Current Behavior

When you have setup a webhook that triggers on a specific node kind this information is lost during the scheduled reconfiguration of webhooks (typically happens once a day)

The problem here is that the filter disappears and the webhook instead triggers on all of the node events and not just the expected one.

### Expected Behavior

The reconfiguration should leave the webhook configuration within Prefect intact.

### Steps to Reproduce

1. Create a webhook and specify that is should trigger on node update events and target `BuiltinTag`
2. Examine the webhook in Prefect
It will look something like this:

```json
{
  "type": "event",
  "match": {
    "infrahub.node.kind": "BuiltinTag"
  },
  "match_related": {
    "prefect.resource.role": "infrahub.branch",
    "infrahub.resource.label": "main"
  },
  "after": [],
  "expect": [
    "infrahub.node.updated"
  ],
  "for_each": [],
  "posture": "Reactive",
  "threshold": 1,
  "within": 0
}
```

3. Mimic running the nightly task in Prefect by triggering the deployment "webhook-setup-automation-all" (note that this workflow gets removed in Infrahub 1.9
4. Examine the webhook again.

```json
{
  "type": "event",
  "match": {},
  "match_related": {
    "prefect.resource.role": "infrahub.branch",
    "infrahub.resource.label": "main"
  },
  "after": [],
  "expect": [
    "infrahub.node.updated"
  ],
  "for_each": [],
  "posture": "Reactive",
  "threshold": 1,
  "within": 0
}
```

Note that the match condition has been removed.

### Additional Information

_No response_

polmichel:
**EDIT:** While working to unify the data source from which we query the web hooks (always using the database via the `NodeManager` and no longer using the SDK client queries), I realised that I had confused the root cause in both sources.
_"The SDK client is not deserializing Enum object as Enum, but as strings. The database query does"_ **->** _"The SDK client is not deserializing `Enum` object as string, but as Python `Enum`. The database query does."_

# Why

When configuring web hooks Prefect trigger definitions, web hooks can be configured through two different flows in 1.8.X:
- `infrahub.webhook.gather.gather_trigger_webhook`: which is querying the web hook object from the database through: `await NodeManager.query(db=db, schema=CoreWebhook)`.
- `infrahub.webhook.tasks.configure_webhook_one`: which is querying the web hook object from the SDK client through `await get_client().get(kind=CoreWebhook, id=event_data["node_id"])`.

The SDK client is not deserializing `Enum` object as string, but as Python `Enum`. The database query does.
This is the source of the following bug https://github.com/opsmill/infrahub/issues/8694. 

## Problem
In the same method, `obj.event_type.value` is reused later on in this condition:
```
  if obj.node_kind.value and obj.event_type.value in get_all_infrahub_node_kind_events():
      event_trigger.match = {"infrahub.node.kind": obj.node_kind.value}
```

## What's next
The larger picture is that we should not, in my opinion, manage both data format and only use one of them. This will be another task for a future release materialized in this internal issue https://opsmill.atlassian.net/browse/IFC-2434.

Jira: https://opsmill.atlassian.net/browse/IFC-2430
Closes https://github.com/opsmill/infrahub/issues/8772

## What changed

- `obj.event_type.value` usage has been totally removed from the method.
- The existing unit test regarding the Web hook trigger definition generation through SDK client has been enhanced and test all the important attributes of a web hook trigger definition.
- Another unit test has been created, doing the same thing but from a web hook queried from the database.

## How to test

```bash
uv run pytest tests/functional/webhook/test_task.py -k "test_trigger_definition_node_kind_match" -v
```

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Fixes loss of webhook node-kind match statements during scheduled reconfiguration so triggers keep their filters when webhooks are reloaded from the DB or SDK.

- **Bug Fixes**
  - Normalized `event_type` in `WebhookTriggerDefinition.from_object` and used it for event checks, preserving node-kind `match={"infrahub.node.kind": <kind>}` in both SDK (`CoreStandardWebhook`) and DB (`NodeManager`/`CoreWebhook`) paths.
  - Added functional tests with full assertions for both paths and included a changelog fragment.

<sup>Written for commit e0eb2959e79897922af030e8345e1d3b6f706a76. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -44,7 +44,7 @@ class WebhookTriggerDefinition(TriggerDefinition):
     def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
         event_trigger = EventTrigger()
         event_type = obj.event_type.value
-        if isinstance(obj.event_type.value, Enum):
+        if isinstance(event_type, Enum):
             event_type = obj.event_type.value.value
         if event_type == "all":
             event_trigger.events.add("infrahub.*")
@@ -62,7 +62,7 @@ class WebhookTriggerDefinition(TriggerDefinition):
                 "infrahub.resource.label": f"!{registry.default_branch}",
             }
 
-        if obj.node_kind.value and obj.event_type.value in get_all_infrahub_node_kind_events():
+        if obj.node_kind.value and event_type in get_all_infrahub_node_kind_events():
             event_trigger.match = {"infrahub.node.kind": obj.node_kind.value}
 
         definition = cls(
--- a/backend/tests/functional/webhook/test_task.py
+++ b/backend/tests/functional/webhook/test_task.py
@@ -9,7 +9,10 @@ from prefect.client.orchestration import PrefectClient, get_client
 from prefect.events.actions import RunDeployment
 
 from infrahub.core.constants import InfrahubKind
+from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
+from infrahub.core.protocols import CoreWebhook as CoreWebhookNode
+from infrahub.trigger.models import ExecuteWorkflow, TriggerType
 from infrahub.trigger.setup import gather_all_automations
 from infrahub.webhook.gather import gather_trigger_webhook
 from infrahub.webhook.models import EventContext, WebhookTriggerDefinition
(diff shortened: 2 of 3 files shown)


ajtmccarty (approved):

ogenstad (approved):

polmichel:
While working to unify the data source from which we query the web hooks (always using the database via the `NodeManager` and no longer using the SDK client queries), I realised that I had confused the root cause in both sources.

PR description is updated with a **EDIT** comment.

_"The SDK client is not deserializing Enum object as Enum, but as strings. The database query does"_ **->** _"The SDK client is not deserializing `Enum` object as string, but as Python `Enum`. The database query does."_

PR #8790, opened 2026-04-02
Remove deprecated "_updated_at" field from GraphQL queries

ogenstad:
# Why

* The `_updated_at` field within GraphQL queries has been deprecated since a while back and it's now being removed.
* The proper way is to query for the information from the node_metadata.

This was a weird field from the start and typically indicated when the node was created and not when it was last updated.

## What changed

* Any reference to this `_updated_at` field.

---
## Summary by cubic
Removed the deprecated `_updated_at` field from the GraphQL schema and code. All queries and UI now use `node_metadata.updated_at` (and `created_at` where needed) per IFC-2150.

- **Migration**
  - Replace `_updated_at` selections with `node_metadata { updated_at }`.
  - For nested/edged results, read timestamps from `edges[].node_metadata`.
  - Regenerate GraphQL types/codegen and remove any `_updated_at` references.

<sup>Written for commit f1e5978b93998c1abe77ac3307abd46779f145cc. Summary will update on new commits.</sup>


ajtmccarty (approved):
backend looks good

bilalabbad (approved):

PR #8798, opened 2026-04-02
wire in log forwarding for PermissionDeniedErrors

ajtmccarty:
IFC-2359

adds handling for sending `PermissionDeniedError`s to `LogForwardingService`. handles if the error is raised from a REST API endpoint or during graphql execution. the actual log forwarding is still a no-op in the community repo.

---
## Summary by cubic
Forward `PermissionDeniedError` events to the configured log forwarding service with rich request/GraphQL context; responses remain 403. Adds a `ForwardableError` base to avoid duplicate forwarding. Implements IFC-2359.

- **New Features**
  - Added `ForwardableError` with `log_forwarded` flag and registered a handler that forwards then returns the normal error response.
  - REST: handler builds `LogForwardingContext` from `request.state.account_session`, `request.state.branch_name`, IP, and path; safe no-ops if service is missing.
  - GraphQL: forwards on permission checks and when execution errors contain `PermissionDeniedError`, including operation name/type and operations.
  - Added `LogForwardingService.forward_exception(exception, context)` and payload models (`ExceptionLogPayload`, `PermissionDeniedPayload`) with field docs; exposed `InfrahubServices.log_forwarding`; stored `account_session` and `branch_name` on `request.state`; tests added.

- **Migration**
  - If you maintain a custom `LogForwardingService`, implement `forward_exception(self, exception, context)` (no-op is fine).

<sup>Written for commit 32c5b1fa7e3db803633d8ef1e017e17e70d25bf3. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/api/dependencies.py
+++ b/backend/infrahub/api/dependencies.py
@@ -73,11 +73,13 @@ async def get_refresh_token(
 
 
 async def get_branch_params(
+    request: Request,
     db: InfrahubDatabase = Depends(get_db),
     branch_name: str | None = Query(None, alias="branch", description="Name of the branch to use for the query"),
     at: str | None = Query(None, description="Time to use for the query, in absolute or relative format"),
 ) -> BranchParams:
     branch = await registry.get_branch(db=db, branch=branch_name)
+    request.state.branch_name = branch.name
 
     return BranchParams(branch=branch, at=Timestamp(at))
 
@@ -110,6 +112,7 @@ async def get_current_user(
         or request.url.path.startswith("/graphql")
         or (config.SETTINGS.main.allow_anonymous_access and request.method.lower() in ["get", "options"])
     ):
+        request.state.account_session = account_session
         return account_session
 
     raise AuthorizationError("Authentication is required")
--- a/backend/infrahub/api/exception_handlers.py
+++ b/backend/infrahub/api/exception_handlers.py
@@ -7,14 +7,11 @@ from pydantic import ValidationError
 
 from infrahub.auth import AccountSession
 from infrahub.exceptions import Error, PermissionDeniedError
-from infrahub.log import get_logger
 from infrahub.log_forwarding.models import LogForwardingContext
 from infrahub.services import InfrahubServices
 
 if TYPE_CHECKING:
     from starlette.requests import Request
-
-logger = get_logger()
 
 
 async def generic_api_exception_handler(_: Request, exc: Exception, http_code: int = 500) -> JSONResponse:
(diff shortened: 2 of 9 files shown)


backend/infrahub/api/dependencies.py
```diff
@@ -73,11 +73,13 @@ async def get_refresh_token(
 
 
 async def get_branch_params(
+    request: Request,
     db: InfrahubDatabase = Depends(get_db),
     branch_name: str | None = Query(None, alias="branch", description="Name of the branch to use for the query"),
     at: str | None = Query(None, description="Time to use for the query, in absolute or relative format"),
 ) -> BranchParams:
     branch = await registry.get_branch(db=db, branch=branch_name)
+    request.state.branch_name = branch.name
```
ajtmccarty, line 82: added these items to `request.state` so that they can be used in the `permission_denied_exception_handler` to include in the syslog message we forward

`request.state` feels like a code smell to me, but this has to work within the confines of FastAPI without requiring adding code to every API endpoint. not sure if there is a better way to do that

backend/infrahub/graphql/app.py
```diff
@@ -265,10 +277,15 @@ async def _handle_http_request(
...
-                    self._log_error(error=error.original_error)
+            self._process_graphql_errors(
+                errors=result.errors,
+                labels=labels,
+                account_session=account_session,
+                branch=branch,
+                analyzed_query=analyzed_query,
+                request=request,
+                graphql_params=graphql_params,
+            )
```
ajtmccarty, line 288: the additional logic in this function exceeded the complexity limit in some manner, so it required a little refactoring

backend/tests/unit/api/test_api_exception_handler.py
ajtmccarty, line 1: the tests in here uses mocks, but they are fast. real functional tests for this exist in the enterprise repo

backend/infrahub/api/exception_handlers.py
```diff
@@ -19,9 +30,44 @@ async def generic_api_exception_handler(_: Any, exc: Exception, http_code: int =
...
+            branch_name=branch_name,
+            operation_name="",
+            query_type="",
+            ip_address=request.client.host if request.client else "",
+            graphql_operations=[],
+            request_path=request.url.path,
+        )
+        log_forwarding.forward_exception(exception=exc, context=context)
+    except Exception:
+        logger.debug("Failed to forward PermissionDeniedError to syslog", exc_info=True)
```
gmazoyer: I wonder if these will raise an error if we were to enable Ruff's `BLE` rule

backend/tests/unit/api/test_api_exception_handler.py
```diff
@@ -38,14 +51,63 @@ def __init__(self, message: str | None) -> None:
...
+            "client": (client_host, 12345),
+            "app": MagicMock(state=app_state or State()),
+        },
+    )
+
+
+@pytest.fixture
+def request_fixture() -> RequestFixture:
+    mock_lf = MagicMock(spec=LogForwardingService)
+    account_session = AccountSession(account_id="user-1", auth_type=AuthType.JWT)
```
gmazoyer, line 85: Do you think it will be worth to have an in-memory implementation of such service, like we have for our cache service, in the future, for testing purpose?

gmazoyer (approved):

backend/infrahub/api/exception_handlers.py
ajtmccarty: I hadn't considered that, but this did make me think that the exception handling should be moved to the actual LogForwardingService implementation in the enterprise repo, which I've done

backend/tests/unit/api/test_api_exception_handler.py
ajtmccarty: good point. I've added one in the enterprise PR. this one doesn't actually do anything, so I think it is fine as a mock that uses `spec=`

backend/infrahub/api/exception_handlers.py
```diff
@@ -1,12 +1,20 @@
-from typing import Any
+from __future__ import annotations
```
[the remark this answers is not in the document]
ajtmccarty: exception handling is moved to LogForwardingService.forward_exception

backend/infrahub/graphql/app.py
```diff
@@ -312,6 +329,55 @@ async def _evaluate_permissions(
...
+        context = LogForwardingContext(
+            account_session=account_session,
+            branch_name=branch.name,
+            operation_name=analyzed_query.operation_name or "",
+            query_type="mutation" if analyzed_query.contains_mutation else "query",
+            ip_address=request.client.host if request.client else "",
+            graphql_operations=[op.name for op in analyzed_query.operations if op.name],
+            request_path=request.url.path,
+        )
+        service.log_forwarding.forward_exception(exception=exception, context=context)
```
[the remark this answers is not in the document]
ajtmccarty: samesies

polmichel (approved):

PR #8803, opened 2026-04-03
Local test AI bug pipeline #4: prevent duplicate branch names with atomic Cypher create query (closes #8368)

### LINKED ISSUE #8368: bug: The lack of a lock around branch creation allows for multiple branches with the same name

polmichel:
# Why

Branch creation has a TOCTOU (time-of-check-time-of-use) race condition: the `Branch.get_by_name()` uniqueness check and the actual `CREATE` Cypher statement are not atomic. Concurrent requests can both pass the check before either branch is saved, resulting in duplicate branches with the same name in the database.

**Goal:** Make branch name creation atomic so duplicate branch names are impossible at the database query level.

**Non-goals:** This PR does not add a distributed lock around the `create_branch` workflow (that could be done as defense-in-depth later) or add a Neo4j uniqueness constraint (which would require a migration).

Closes #8368

## What changed

- **Behavioral change:** Attempting to create a branch with a name that already exists now fails atomically at the Cypher query level instead of silently creating a duplicate.

- **Implementation:** Introduced `BranchNodeCreateQuery` in `backend/infrahub/core/query/branch.py` — a Branch-specific create query that uses `OPTIONAL MATCH` + `WHERE existing IS NULL` before the `CREATE` statement. If a Branch with the same name already exists, the query returns zero rows, and `StandardNode.create()` raises an error. The `Branch` model now uses this query class via `_query = BranchNodeCreateQuery` instead of the generic `StandardNodeCreateQuery`.

- **What stayed the same:** No schema changes. No API contract changes. The `StandardNodeCreateQuery` is untouched — only `Branch` gets the specialized query. The existing `local_schema_lock()` in `tasks.py` is preserved.

## Fix strategy

The root cause is a shallow issue in the data layer: the generic `StandardNodeCreateQuery` uses a bare `CREATE` with no duplicate protection. Rather than adding a guard clause at the application layer (which would still be susceptible to races), the fix pushes the uniqueness enforcement into the Cypher query itself using an atomic check-then-create pattern. This is a targeted fix that only affects Branch creation without modifying shared infrastructure.

## How to review

Focus on `backend/infrahub/core/query/branch.py` — the new `BranchNodeCreateQuery` class. The Cypher pattern is the core of the fix. Then check `backend/infrahub/core/branch/models.py` for the `_query` override wiring.

## How to test

```bash
uv run pytest backend/tests/component/core/test_branch_create_race.py::test_duplicate_branch_name_rejected -x -v
```

Expected: test passes — only one branch with the duplicate name exists in the database after two save attempts.

## Impact & rollout

- **Backward compatibility:** No breaking changes. The only observable difference is that creating a branch with an existing name now raises an error instead of silently creating a duplicate.
- **Performance:** Negligible — adds one `OPTIONAL MATCH` to the branch create query.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added
- [ ] External docs updated (N/A — no user-facing doc change)
- [ ] Internal .md docs updated (N/A)

---
## Summary by cubic
Make branch creation atomic to prevent duplicate branch names caused by concurrent requests. Creating a branch with an existing name now fails at the Cypher query level.

- **Bug Fixes**
  - Added `BranchNodeCreateQuery` that uses `OPTIONAL MATCH` + `WHERE existing IS NULL` before `CREATE`; duplicate names return no rows so `StandardNode.create()` raises.
  - Wired `Branch` to use this query and added a test to ensure only one branch persists when two saves race.

<sup>Written for commit dc9f68821200788df49ab0b5a854c81d87bdab80. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -13,6 +13,7 @@ from infrahub.core.models import SchemaBranchHash  # noqa: TC001
 from infrahub.core.node.standard import StandardNode, StandardNodeOrdering
 from infrahub.core.query import Query, QueryType
 from infrahub.core.query.branch import (
+    BranchNodeCreateQuery,
     BranchNodeGetListQuery,
     DeleteBranchRelationshipsQuery,
     RebaseBranchQuery,
@@ -48,6 +49,7 @@ class Branch(StandardNode):
     schema_hash: Optional[SchemaBranchHash] = None
     graph_version: int | None = None
 
+    _query: type[BranchNodeCreateQuery] = BranchNodeCreateQuery
     _exclude_attrs: list[str] = ["id", "uuid", "owner"]
 
     @field_validator("name", mode="before")
--- a/backend/infrahub/core/query/branch.py
+++ b/backend/infrahub/core/query/branch.py
@@ -6,11 +6,43 @@ from infrahub.core.branch.enums import BranchStatus
 from infrahub.core.branch.filters import BranchListFilters
 from infrahub.core.constants import GLOBAL_BRANCH_NAME
 from infrahub.core.query import Query, QueryType
-from infrahub.core.query.standard_node import StandardNodeGetListQuery
+from infrahub.core.query.standard_node import StandardNodeGetListQuery, StandardNodeQuery
 from infrahub.core.timestamp import Timestamp
 
 if TYPE_CHECKING:
     from infrahub.database import InfrahubDatabase
+
+
+class BranchNodeCreateQuery(StandardNodeQuery):
+    """Create a Branch node atomically, rejecting duplicates by name.
+
+    Uses an atomic Cypher pattern: check for an existing Branch with the same
+    name and only CREATE when none is found. If a branch with that name already
+    exists, the query returns no rows, which causes ``StandardNode.create()``
+    to raise an error.
+    """
+
+    name = "branch_node_create"
+    type = QueryType.WRITE
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        node_type = self.node.get_type()
+        self.params["node_prop"] = self.node.to_db()
+        self.params["branch_name"] = self.node.to_db()["name"]
+
+        # Atomically ensure no Branch with this name exists before creating.
+        # If a concurrent transaction already inserted one, this query returns
+        # zero rows and the caller treats it as a creation failure.
+        query = """
+        OPTIONAL MATCH (existing:%(node_type)s {name: $branch_name})
+        WITH existing
+        WHERE existing IS NULL
+        MATCH (root:Root)
+        CREATE (n:%(node_type)s $node_prop)-[r:IS_PART_OF]->(root)
+        """ % {"node_type": node_type}
+
+        self.add_to_query(query=query)
+        self.return_labels = ["n"]
 
 
 class DeleteBranchRelationshipsQuery(Query):
--- a/backend/tests/component/core/test_branch_create_race.py
+++ b/backend/tests/component/core/test_branch_create_race.py
@@ -1,0 +1,72 @@
+from infrahub.core.branch import Branch
+from infrahub.core.branch.enums import BranchStatus
+from infrahub.core.schema.schema_branch import SchemaBranch
+from infrahub.database import InfrahubDatabase, QueryType
+
+
+async def test_duplicate_branch_name_rejected(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: SchemaBranch,
+) -> None:
+    """Saving two branches with the same name must result in only one branch in the database.
+
+    The production create_branch flow (in core/branch/tasks.py) uses a TOCTOU pattern:
+    it checks Branch.get_by_name() then creates the branch, with no atomicity guarantee.
+    Concurrent requests can both pass the check before either branch is saved.
+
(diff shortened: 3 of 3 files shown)


polmichel (commented):
## Code Review: fix: prevent duplicate branch names with atomic Cypher create query

---

### A. Correctness

**The fix partially addresses the root cause, but the atomicity claim is overstated.**

The Cypher pattern `OPTIONAL MATCH (existing) ... WHERE existing IS NULL ... CREATE` executes within a single Cypher statement, which in Neo4j runs inside a single transaction. However, Neo4j Community Edition uses `READ_COMMITTED` isolation by default. This means two concurrent transactions can both execute the `OPTIONAL MATCH`, both see `existing IS NULL`, and both proceed to `CREATE` — the exact TOCTOU race this PR aims to prevent.

True atomicity for this pattern requires one of:
1. A **Neo4j uniqueness constraint** (`CREATE CONSTRAINT ... FOR (b:Branch) REQUIRE b.name IS UNIQUE`) — the PR explicitly defers this as a non-goal, but it's the only approach that provides a hard guarantee.
2. **SERIALIZABLE isolation** or explicit locking (e.g., `MERGE` with a lock on the node).

That said, the fix **significantly narrows the race window** compared to the current code where the check and create are separate round-trips. Within a single Cypher statement, the window is microseconds vs. milliseconds. This is a meaningful improvement even if not a complete guarantee.

The `create_branch` task in `tasks.py` (line 491-494) still does a separate `Branch.get_by_name()` check before calling `save()`. This is now redundant defense-in-depth, which is fine, but worth noting.

**Error handling concern:** When the atomic query returns zero rows (duplicate detected), `StandardNode.create()` raises `Error("Unable to create the node Branch")`. This generic message does not indicate that the failure was due to a duplicate name. A user or operator seeing this error will have no idea what went wrong. Consider catching this in `Branch.create()` and re-raising with a descriptive message.

**Edge case:** If the `Root` node doesn't exist (broken database state), the `MATCH (root:Root)` clause will also cause zero rows, producing the same generic error. This is the same behavior as the existing `StandardNodeCreateQuery`, so it's not a regression — but worth documenting.

---

### B. Code Quality

**Positive:**
- The `_query` override pattern is already established by `Root` in `root.py`, so this follows existing conventions.
- The change is minimal and targeted — only `Branch` gets the specialized query.
- Import structure is correct (runtime import, not under `TYPE_CHECKING`, since `_query` is used at runtime).

**Concerns:**
- **String interpolation in Cypher** (`% {"node_type": node_type}`) — this is consistent with other queries in the codebase (`StandardNodeCreateQuery`, `StandardNodeDeleteQuery`), so not a regression. However, `node_type` comes from `cls.__name__` which is always `"Branch"`, so this is safe.
- The `_query` attribute declaration on `Branch` (line 51 in the diff) sits between `graph_version` and `_exclude_attrs`. Private attributes (`_`-prefixed) are typically grouped together. It should be placed next to `_exclude_attrs` for consistency. Minor nit.

---

- No ADRs or design docs appear to be affected.
- **Missing changelog fragment** — the PR checklist shows `[ ] Changelog entry added`. Per project conventions, a bug fix that changes user-observable behavior (duplicate branch creation now raises an error instead of silently succeeding) should have a changelog entry.
- No public API contract changes (the REST/GraphQL API already returns errors for duplicate names at the application layer in `tasks.py`).

---

### D. Test Quality

**The test is well-structured but has a design flaw in the error handling:**

```python
try:
    await branch2.save(db=db)
except Exception:
    pass
```

Catching and swallowing all exceptions means the test passes even if `save()` fails for an unrelated reason (e.g., database connection error, serialization bug). The test should catch the specific exception type (`Error` from `infrahub.exceptions`) and optionally assert on the message.

**The test does not test the actual race condition.** It tests sequential duplicate creation, which is valuable, but the PR title and description focus on concurrent requests. A true concurrency test (e.g., using `asyncio.gather` to fire two saves simultaneously) would be more convincing, though admittedly harder to make deterministic.

**The test is correctly placed** in `backend/tests/component/core/` alongside other branch tests.

**The test name and docstring are clear and descriptive.**

---

### Verdict: APPROVED WITH SUGGESTIONS

The fix is a meaningful improvement that narrows the race window from "two separate network round-trips" to "within a single Cypher statement." It follows established patterns in the codebase and is well-tested for the sequential case.

### Recommended next steps for the human reviewer

1. **Consider adding a Neo4j uniqueness constraint** as a follow-up (the PR explicitly defers this). The Cypher-level check is not a hard guarantee under `READ_COMMITTED` isolation. A constraint would be the definitive fix.
2. **Improve the error message** when the atomic query detects a duplicate. Override `Branch.create()` to catch the generic `Error` and re-raise as `ValidationError(f"A branch named '{self.name}' already exists")`.
3. **Tighten the test exception handling** — catch `Error` specifically instead of bare `Exception`.
4. **Add a changelog fragment** for this bug fix.
5. **Minor:** Group `_query` next to `_exclude_attrs` in the `Branch` class body.

PR #8804, opened 2026-04-03
Local test AI bug pipeline #5: prevent branch creation TOCTOU race condition (closes #8368)

### LINKED ISSUE #8368: bug: The lack of a lock around branch creation allows for multiple branches with the same name

polmichel:
# Why

The branch creation flow (`create_branch` in `backend/infrahub/core/branch/tasks.py`) has a TOCTOU (time-of-check-time-of-use) race condition. The existence check (`Branch.get_by_name`) and the branch creation (`obj.save`) are not atomic — two concurrent `create_branch` calls can both pass the existence check before either persists the branch, resulting in duplicate branches with the same name.

Closes #8368

## Fix strategy

This is a shallow concurrency bug, not a design flaw. The `local_schema_lock` already exists and serializes the critical section (schema duplication, save, registry update). The existence check simply needs to be repeated inside that lock scope.

The fix adds a second `Branch.get_by_name` check inside the `local_schema_lock` context manager. The original check outside the lock remains as a fast-fail optimization (avoids acquiring the lock when the branch already exists). The check inside the lock is the authoritative one that prevents the race: the second concurrent caller will block on the lock, and when it proceeds, its `get_by_name` call will find the branch already created by the first caller and raise `ValidationError`.

## What changed

- **Behavioral change:** Two concurrent `create_branch` calls with the same name now correctly result in exactly one branch being created. The second call receives a `ValidationError`.
- **Implementation:** Added a second `Branch.get_by_name` existence check inside the `local_schema_lock` scope in `create_branch()`.
- **What stayed the same:** The API contract is unchanged. The `local_schema_lock` scope and its existing contents are unchanged. The fast-fail check outside the lock is preserved.

## How to review

Focus on `backend/infrahub/core/branch/tasks.py` — the only production code change is 8 lines added inside the `local_schema_lock` scope (a duplicate existence check).

## How to test

```bash
uv run pytest backend/tests/component/core/test_branch_create_race.py::TestBranchCreateRaceCondition::test_concurrent_create_same_branch_only_one_succeeds -x -v
```

The test simulates the race window using an `asyncio.Event` barrier that forces both coroutines past the outer existence check before either proceeds. With the fix, the inner check (inside the lock) catches the duplicate.

## Impact & rollout

- **Backward compatibility:** No breaking changes.
- **Performance:** Negligible — adds one extra DB query inside an already-held lock, only during branch creation.
- **Deployment notes:** Safe to deploy. No migration required.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Fixes a TOCTOU race in branch creation by re-checking existence inside the schema lock so concurrent requests create only one branch. Satisfies #8368 without changing the API.

- **Bug Fixes**
  - Added a second `Branch.get_by_name` inside `lock.registry.local_schema_lock()` in `create_branch`; kept the outer check for fast-fail.
  - Under concurrent creates with the same name, one succeeds and the other raises `ValidationError`.
  - Added `backend/tests/component/core/test_branch_create_race.py` to reproduce the race and verify the fix.

<sup>Written for commit 3fb232bbb41f32e7434f23a878fc21901f2f6451. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -503,6 +503,14 @@ async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> N
             raise ValidationError("\n".join(error_msgs)) from exc
 
         async with lock.registry.local_schema_lock():
+            # Re-check existence inside the lock to prevent TOCTOU race conditions.
+            # The check above is a fast-fail optimization; this is the authoritative check.
+            try:
+                await Branch.get_by_name(db=db, name=model.name)
+                raise ValidationError(f"The branch {model.name} already exists")
+            except BranchNotFoundError:
+                pass
+
             # Copy the schema from the origin branch and set the hash and the schema_changed_at value
             origin_schema = registry.schema.get_schema_branch(name=obj.origin_branch)
             new_schema = origin_schema.duplicate(name=obj.name)
--- a/backend/tests/component/core/test_branch_create_race.py
+++ b/backend/tests/component/core/test_branch_create_race.py
@@ -1,0 +1,129 @@
+import asyncio
+from unittest.mock import AsyncMock, patch
+from uuid import uuid4
+
+import pytest
+from fast_depends import Provider
+
+from infrahub import lock
+from infrahub.auth import AccountSession, AuthType
+from infrahub.context import InfrahubContext
+from infrahub.core.branch import Branch
+from infrahub.core.branch.tasks import create_branch
+from infrahub.core.schema.schema_branch import SchemaBranch
+from infrahub.database import InfrahubDatabase
+from infrahub.exceptions import BranchNotFoundError, ValidationError
+from infrahub.graphql.mutations.models import BranchCreateModel
+from infrahub.services.adapters.workflow.local import WorkflowLocalExecution
+from infrahub.workers.dependencies import build_database
+
+
+class TestBranchCreateRaceCondition:
+    @pytest.fixture(autouse=True)
+    async def _setup_core_schema(self, register_core_models_schema: SchemaBranch) -> None:
+        return
+
+    @pytest.fixture(autouse=True)
+    async def _setup_lock(self, default_branch: Branch, car_person_schema: SchemaBranch) -> None:
+        lock.initialize_lock(local_only=True)
+
+    @pytest.fixture
+    def context(self, default_branch: Branch) -> InfrahubContext:
+        return InfrahubContext.init(
+            branch=default_branch,
+            account=AccountSession(account_id=str(uuid4()), auth_type=AuthType.NONE),
+        )
+
+    @pytest.fixture
+    def branch_model(self) -> BranchCreateModel:
+        return BranchCreateModel(
+            name="race-test-branch",
+            sync_with_git=False,
+            origin_branch="main",
+        )
+
+    async def test_concurrent_create_same_branch_only_one_succeeds(
+        self,
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        workflow_local: WorkflowLocalExecution,
+        dependency_provider: Provider,
+        context: InfrahubContext,
+        branch_model: BranchCreateModel,
+    ) -> None:
+        """Two concurrent create_branch calls with the same name must result in exactly one branch.
+
+        The second call should raise a ValidationError. This test simulates the TOCTOU race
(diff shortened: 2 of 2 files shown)


polmichel (commented):
## Bug Review: fix branch creation TOCTOU race condition

**Verdict: CHANGES REQUESTED**

---

### A. Correctness — INCOMPLETE FIX

The fix adds a second `Branch.get_by_name` check inside `local_schema_lock()`. This addresses the race condition **only within a single process** (single asyncio event loop).

However, the bug report (#8368) explicitly describes a **multi-worker** scenario:

> "Start a fresh instance of Infrahub that has at least more than one worker of each type"

The `create_branch` flow is dispatched via Prefect's `run_deployment()` (see `WorkflowWorkerExecution.execute_workflow`), which means concurrent branch creations can execute on **different worker processes**. Each worker has its own `local_schema_lock` — an `asyncio.Lock` initialized per-process in `lock.py`. The lock provides zero cross-process protection.

Additionally, `BRANCH_CREATE` in `catalogue.py` has **no `concurrency_limit`** set (unlike `BRANCH_MERGE` which has `concurrency_limit=1`). There is also **no Neo4j unique constraint** on the `Branch.name` property, so the database provides no last-resort protection either.

**The fix as written does not solve the reported bug.** It only prevents the race within a single worker process, which is a narrow subset of the actual failure mode.

To properly fix this, one or more of the following are needed:
1. Add `concurrency_limit=1` (or a suitable value) to `BRANCH_CREATE` in `catalogue.py` — this is the simplest approach and mirrors what `BRANCH_MERGE` already does
2. Use a distributed lock (e.g., `global_graph_lock()` or a dedicated Redis/NATS lock) instead of `local_schema_lock`
3. Add a Neo4j unique constraint on `Branch.name` as a database-level safety net

Option 1 is the least invasive and most consistent with existing patterns. Option 3 provides defense-in-depth regardless of application-level locking.

### B. Code Quality — ACCEPTABLE (conditional on fix)

The 8 lines of production code are clean and follow the project's style. The double-check pattern (fast-fail outside lock, authoritative check inside lock) is well-commented and easy to understand. No unnecessary refactoring.

If the fix is augmented with a distributed protection mechanism, the local re-check still has value as a fast-path optimization within a single worker and should be kept.

### C. Documentation Alignment — NO ISSUES

- No API changes, no migration needed.
- Missing changelog entry (noted in the PR checklist).

### D. Test Quality — PARTIALLY EFFECTIVE

The test correctly validates the single-process race condition using an `asyncio.Event` barrier to force interleaving. The synchronization technique is sound and the assertions are clear.

However, the test **only proves the fix works for concurrent coroutines within one event loop** — which is precisely the scenario the fix addresses. It does not validate the multi-worker scenario from the bug report, and cannot do so with a unit/component test alone.

The `interleaved_get_by_name` mock has a subtle issue: it patches `Branch.get_by_name` globally, meaning the re-check inside the lock also hits the mock. The mock does call `real_get_by_name` first and raises `ValidationError` if the branch exists, so functionally the test works — but the logic flow is convoluted. The second coroutine's inner check succeeds because `real_get_by_name` finds the branch saved by the first coroutine, and the mock translates that into a `ValidationError`. This works but is fragile and hard to follow.

Minor: the `check_passed_count` variable is not thread-safe, but since this is asyncio (single-threaded), it works correctly here.

---

### Recommended Next Steps

1. **Critical:** Add `concurrency_limit=1` to the `BRANCH_CREATE` workflow definition in `catalogue.py`. This is a one-line change that provides the distributed serialization needed to actually fix the reported bug.
2. **Recommended:** Consider adding a Neo4j unique constraint on `Branch.name` for defense-in-depth.
3. **Keep** the local re-check inside `local_schema_lock` — it remains useful as a fast-fail within a single worker even after adding distributed protection.
4. Add a changelog fragment (`uv run towncrier create ...`).
5. The test is fine for validating the local re-check behavior but should not be cited as proof that the multi-worker race is resolved.

PR #8806, opened 2026-04-03
Local test AI bug pipeline #6 add distributed lock to prevent branch creation race condition (closes #8368)

### LINKED ISSUE #8368: bug: The lack of a lock around branch creation allows for multiple branches with the same name

polmichel:
# Why

Concurrent `create_branch` calls with the same branch name can both succeed, creating duplicate branches. This is a TOCTOU (time-of-check-time-of-use) race condition: the existence check (`Branch.get_by_name`) and the branch creation (`obj.save`) are not atomic, and the only lock used (`local_schema_lock`) is process-local -- it does not protect across multiple worker processes.

Closes #8368

## Fix strategy

The root cause is a shallow TOCTOU issue, not a deeper design flaw. The existing architecture already supports distributed locks via `lock.registry.get()`. The fix applies the standard double-checked locking pattern:

1. The initial existence check outside the lock remains as a **fast-fail optimization** (avoids acquiring the lock for branches that already exist).
2. A **distributed lock** scoped to the branch name (`namespace="branch_create"`) serializes concurrent creation attempts for the same branch.
3. A **second existence check inside the lock** is the authoritative guard -- if two callers both passed the initial check, the second one will find the branch after the first creates it.
4. The `local_schema_lock` is preserved inside the distributed lock for its existing purpose (schema registry synchronization).

The lock uses `local=False` so it will be distributed (Redis/NATS) in production and falls back to a local asyncio lock in test environments (when `connection` is None).

## What changed

- **Behavioral change:** Two concurrent branch creation requests with the same name now result in exactly one success and one `ValidationError`, instead of both succeeding.
- **Implementation:** Added a distributed lock around the branch creation path in `create_branch()` in `backend/infrahub/core/branch/tasks.py`, with a re-check of branch existence inside the lock.
- **What stayed the same:** The GraphQL mutation fast-fail check in `branch.py` is unchanged. The `local_schema_lock` is unchanged. No schema changes, no API contract changes.

## How to review

Focus on `backend/infrahub/core/branch/tasks.py` lines 505-535. The key design choice is placing the distributed lock AFTER the initial existence check (so it's only acquired when the branch doesn't exist yet) and BEFORE the `local_schema_lock` (so it serializes the entire check-and-create sequence).

## How to test

```bash
uv run pytest backend/tests/component/core/test_branch_create_race.py::TestBranchCreateRaceCondition::test_concurrent_create_same_branch_only_one_succeeds -x -v
```

Expected: test passes (1 success, 1 ValidationError from concurrent create).

## Impact & rollout

- **Backward compatibility:** No breaking changes. Existing single-caller behavior is identical.
- **Performance:** Negligible. The lock is only acquired when creating a new branch (not on reads), and is scoped per branch name so different branches can still be created concurrently.
- **Deployment notes:** Safe to deploy. Uses the existing distributed lock infrastructure.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Fix race condition in branch creation by adding a per-branch distributed lock and a second existence check. Parallel creates for the same name now result in one success and one `ValidationError` (closes #8368).

- **Bug Fixes**
  - Added a per-branch distributed lock around creation in `BranchCreator` (`lock.registry.get(name=model.name, namespace="branch_create", local=False)`) with a second existence check under the lock; kept `local_schema_lock`. Different branch names still create in parallel.
  - `create_branch()` now delegates to `BranchCreator` (injects `db`, `lock_registry`, `component`, `event_service`, `workflow`); tests use `create_autospec(spec_set=True)` and assert exactly one success and one `ValidationError`; added changelog entry.

<sup>Written for commit 6198e19b8003c97bbeec5553b940bef3fb8f7481. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -502,18 +502,31 @@ async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> N
             error_msgs = [f"invalid field {error['loc'][0]}: {error['msg']}" for error in exc.errors()]
             raise ValidationError("\n".join(error_msgs)) from exc
 
-        async with lock.registry.local_schema_lock():
-            # Copy the schema from the origin branch and set the hash and the schema_changed_at value
-            origin_schema = registry.schema.get_schema_branch(name=obj.origin_branch)
-            new_schema = origin_schema.duplicate(name=obj.name)
-            registry.schema.set_schema_branch(name=obj.name, schema=new_schema)
-            obj.update_schema_hash()
-            await obj.save(db=db, user_id=context.account.account_id)
+        # Use a distributed lock to prevent TOCTOU race conditions when creating branches.
+        # The lock is scoped to the branch name so different branches can be created concurrently.
+        # The existence check above is a fast-fail optimization; the authoritative check is
+        # inside the lock to prevent the race where two callers both pass the check above
+        # before either creates the branch.
+        async with lock.registry.get(name=model.name, namespace="branch_create", local=False):
+            # Re-check existence under the lock to prevent TOCTOU race
+            try:
+                await Branch.get_by_name(db=db, name=model.name)
+                raise ValidationError(f"The branch {model.name} already exists")
+            except BranchNotFoundError:
+                pass
 
-            # Add Branch to registry
-            registry.branch[obj.name] = obj
-            component = await get_component()
-            await component.refresh_schema_hash(branches=[obj.name])
+            async with lock.registry.local_schema_lock():
+                # Copy the schema from the origin branch and set the hash and the schema_changed_at value
+                origin_schema = registry.schema.get_schema_branch(name=obj.origin_branch)
+                new_schema = origin_schema.duplicate(name=obj.name)
+                registry.schema.set_schema_branch(name=obj.name, schema=new_schema)
+                obj.update_schema_hash()
+                await obj.save(db=db, user_id=context.account.account_id)
+
+                # Add Branch to registry
+                registry.branch[obj.name] = obj
+                component = await get_component()
+                await component.refresh_schema_hash(branches=[obj.name])
 
         event = BranchCreatedEvent(
             branch_name=obj.name,
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -63,6 +63,10 @@ if TYPE_CHECKING:
 
     from infrahub.core.changelog.models import NodeChangelog
     from infrahub.database import InfrahubDatabase
+    from infrahub.lock import InfrahubLockRegistry
+    from infrahub.services.adapters.event import InfrahubEventService
+    from infrahub.services.adapters.workflow import InfrahubWorkflow
+    from infrahub.services.component import InfrahubComponent
 
 
 @flow(name="branch-migrate", flow_run_name="Apply migrations to branch {branch}")
@@ -481,14 +485,24 @@ async def validate_branch(branch: str) -> State:
         return Completed(message="branch is valid")
 
 
-@flow(name="create-branch", flow_run_name="Create branch {model.name}")
-async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> None:
-    await add_tags(branches=[model.name])
+class BranchCreator:
+    def __init__(
+        self,
+        db: InfrahubDatabase,
+        lock_registry: InfrahubLockRegistry,
+        component: InfrahubComponent,
+        event_service: InfrahubEventService,
+        workflow: InfrahubWorkflow,
+    ) -> None:
+        self.db = db
+        self.lock_registry = lock_registry
+        self.component = component
+        self.event_service = event_service
+        self.workflow = workflow
 
-    database = await get_database()
-    async with database.start_session() as db:
+    async def create(self, model: BranchCreateModel, context: InfrahubContext) -> None:
         try:
-            await Branch.get_by_name(db=db, name=model.name)
+            await Branch.get_by_name(db=self.db, name=model.name)
             raise ValidationError(f"The branch {model.name} already exists")
         except BranchNotFoundError:
             pass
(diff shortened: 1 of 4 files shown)


polmichel (commented):
## Code Review: Distributed Lock for Branch Creation Race Condition

### A. Correctness

**Verdict: The fix correctly addresses the root cause.**

The TOCTOU race condition is real: two workers can both pass `Branch.get_by_name` (not found), then both proceed to `obj.save`, creating duplicate branches. The `local_schema_lock` is process-local (`asyncio.Lock`) and provides zero protection across workers.

The fix applies double-checked locking with a distributed lock — the standard solution for this class of bug:

1. **Fast-fail check** (outside lock): avoids acquiring the lock for branches that already exist. Good optimization.
2. **Distributed lock** via `lock.registry.get(name=model.name, namespace="branch_create", local=False)`: serializes concurrent creation attempts for the same branch name. Different branch names get different locks — no unnecessary serialization.
3. **Re-check under lock**: the authoritative guard. After the first coroutine creates the branch, the second finds it and raises `ValidationError`.

**Lock mechanism verification:** I traced the code path through `lock.py`:
- `registry.get()` generates the name `"global.branch_create.<branch_name>"` via `LockNameGenerator`
- `InfrahubLock.__init__` receives `connection=self.connection` from the registry. Note that `local` is NOT forwarded from `get()` to the `InfrahubLock` constructor — the `local=False` parameter only affects name generation (adds `"global."` prefix).
- In production (Redis/NATS connection present): line 145 does NOT force `use_local=True` because `connection` is truthy and the name doesn't start with `"local."`. The lock is distributed. Correct.
- In tests (`local_only=True`, `connection=None`): line 145 forces `use_local=True`. Falls back to `asyncio.Lock`. Correct for test environments.

**One subtlety worth noting:** The `local=False` parameter on `registry.get()` does not actually force a distributed lock — it only affects the generated name. The actual local-vs-distributed decision is made by `InfrahubLock.__init__` based on whether `connection` is `None`. In practice this works correctly because `connection` is set based on `local_only` at registry initialization time. But it means the `local=False` argument is cosmetic/documentary rather than truly forcing distributed behavior.

**Edge cases covered:**
- Same branch name concurrent creation: handled by the distributed lock
- Different branch names concurrent creation: no contention (per-name lock scoping)
- Single-caller behavior: unchanged (lock acquire/release is transparent)

**Edge case NOT covered but acceptable:**
- If the distributed lock backend (Redis/NATS) is down, lock acquisition will block/fail. This is pre-existing behavior for other distributed locks in the codebase and is not introduced by this change.

### B. Code Quality

**Verdict: Clean, follows project conventions.**

- The change is minimal and focused. No unnecessary refactoring.
- Code style matches the surrounding context (indentation, comment style, import patterns).
- The comment block explaining the distributed lock purpose is helpful and not excessive.
- The `local_schema_lock` is correctly preserved inside the distributed lock — it still serves its original purpose (schema registry synchronization within the process).
- No performance concerns: the lock is only acquired during branch creation (write path), scoped per branch name, and the fast-fail check outside the lock avoids unnecessary lock acquisition.

**Minor observation:** The lock is created with `metrics=True` (default). This means every branch creation will emit lock metrics. This is appropriate — branch creation is infrequent enough that metric overhead is negligible, and the observability is valuable.

**Verdict: Aligned with existing architecture.**

- The fix reuses the existing `lock.registry.get()` infrastructure — no new lock mechanisms introduced.
- The pattern of distributed locks for cross-worker coordination is already established in the codebase (e.g., `GLOBAL_SCHEMA_LOCK`, `GLOBAL_GRAPH_LOCK`).
- No public API or behavior changes (same `ValidationError` for duplicate branch names, just now consistently raised instead of sometimes both succeeding).
- The PR checklist shows the changelog entry is not yet created. The `dev/guidelines/changelog.md` requires one for every bug fix: `uv run towncrier create -c "Fixed race condition in concurrent branch creation" 8368.fixed.md`.

### D. Test Quality

**Verdict: Adequate but with caveats.**

The test correctly validates the end-to-end behavior: two concurrent `create_branch` calls result in exactly one success and one `ValidationError`.

**Strengths:**
- Follows the existing test patterns in the codebase (compare `test_merge_task_lock.py` — same fixture structure, same mock/patch approach).
- Uses `asyncio.gather` with `return_exceptions=True` to capture both outcomes.
- The `interleaved_get_by_name` mock forces the race condition timing that would be non-deterministic in a real concurrent scenario.

**Caveats:**
- The mock patches `Branch.get_by_name` globally, which means it intercepts BOTH the initial check AND the re-check inside the lock. The test still works because the mock delegates to `real_get_by_name` first (which finds the branch after the first coroutine creates it), but this coupling to the two-check implementation structure makes the test somewhat fragile if the internal structure changes.
- The test runs with `local_only=True` locks (no Redis/NATS). This validates the logic but not the distributed lock behavior in production. This is standard for component tests in this codebase.
- The `check_passed_count` counter is not thread-safe, but this is fine because asyncio is single-threaded and the counter is only accessed within coroutines (no preemption between increment and comparison).

### Summary

| Dimension | Rating |
|-----------|--------|
| Correctness | Pass |
| Code Quality | Pass |
| Documentation | Missing changelog entry |
| Test Quality | Pass (with noted caveats) |

1. **Create changelog entry:** `uv run towncrier create -c "Fixed race condition allowing duplicate branches when multiple workers handle concurrent creation requests" 8368.fixed.md`
2. **Verify the test passes:** `uv run pytest backend/tests/component/core/test_branch_create_race.py -x -v`
3. **Consider:** Whether a database-level uniqueness constraint on branch name would be a valuable defense-in-depth (separate from this PR — the distributed lock fix is sufficient and correct on its own).
4. **Mark PR as ready for review** (currently in DRAFT state).

**Overall verdict: APPROVED WITH SUGGESTIONS** — the fix is correct and well-structured. The missing changelog entry should be added before merge.

polmichel:
This bug has been created by the 4-steps AI bug pipeline: analyst, test writer, fixer and reviewer.
The human-review starts from this commit c5416bbf74ad2d517ec163c5a5d540608996ff31

Internal testing details related to the different tests of AI bug pipeline steps are over there https://opsmill.atlassian.net/wiki/spaces/SD/pages/522584066/Testing+history#Issue-2

backend/infrahub/core/branch/tasks.py
```diff
@@ -502,18 +502,31 @@ async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> N
...
-
-            # Add Branch to registry
-            registry.branch[obj.name] = obj
-            component = await get_component()
-            await component.refresh_schema_hash(branches=[obj.name])
+        # Use a distributed lock to prevent TOCTOU race conditions when creating branches.
+        # The lock is scoped to the branch name so different branches can be created concurrently.
+        # The existence check above is a fast-fail optimization; the authoritative check is
+        # inside the lock to prevent the race where two callers both pass the check above
+        # before either creates the branch.
```
ajtmccarty: this is too much comment
`# lock to prevent creating multiple branches with the same name` is sufficient

backend/tests/component/core/test_branch_create_race.py
ajtmccarty, line 1: this test is confusing
have you tried an integration test that uses the client to try to create multiple branches with the same name at the same time? I'm not sure if that would work, but if it does, then it would be easier to understand

polmichel: This is not easy to understand, but it allows the issue to be reproduced with just two tasks waiting for each other, which is cleaner to some extent.
I'm not convinced that creating X number of branches and assuming they will conflict is better from a test suite maintainability perspective.

This is true that another potential downside of this test in its current state is that it is not resistant to refactoring.

I will try to make it clearer. If you are still not convinced, I will try the integration test if that works for you.

polmichel: Test updated, thanks for the review

backend/infrahub/core/branch/tasks.py
```diff
@@ -481,14 +485,24 @@ async def validate_branch(branch: str) -> State:
         return Completed(message="branch is valid")
 
 
-@flow(name="create-branch", flow_run_name="Create branch {model.name}")
-async def create_branch(model: BranchCreateModel, context: InfrahubContext) -> None:
-    await add_tags(branches=[model.name])
-
-    database = await get_database()
-    async with database.start_session() as db:
+class BranchCreator:
```
ajtmccarty: I think this can move to its own file `branch/creator.py` or something so that it can be used elsewhere if necessary without importing everything in this file

ajtmccarty (approved):
👏 👏 👏

PR #8822, opened 2026-04-07
IFC-2110 Index HFID values and simplify lookup

gmazoyer:
## Why & What

Add `IndexedListAttribute` for HFID storage so values get the `AttributeValueIndexed` label on both create and update paths. Oversized values gracefully fall back to non-indexed storage with a warning.

Rewrite `get_one_by_hfid` to use `NodeGetByHFIDQuery` instead of unpacking HFID components and building per-field filters.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Indexes and normalizes HFID values to meet IFC-2110, and switches HFID lookups to an index-backed query for faster, reliable matches. Oversized HFIDs are stored without an index and are not searchable; graph version is bumped to 70 with a migration to normalize and index existing values.

- **New Features**
  - Store HFIDs with `IndexedListAttribute`; index within limits, else store without index and log a warning.
  - Enforce HFIDs as `list[str]`; `compute()` casts all parts to `str()`, and `get_path_value()` returns `Any`.
  - Replace `get_one_by_hfid` with `NodeGetByHFIDQuery`: start from `AttributeValueIndexed`, traverse back with branch-aware checks (respects `branch_agnostic`) on `IS_PART_OF`, `HAS_ATTRIBUTE`, and `HAS_VALUE`; uses `DISTINCT`; only indexed HFIDs are searchable.

- **Migration**
  - `070_index_hfid_values`: normalize HFID JSON to all-strings by creating/merging a normalized `AttributeValue` and reattaching `HAS_VALUE` (idempotent).
  - Index HFIDs under size limits: add label in-place for exclusive values; for shared values create a new `AttributeValueIndexed` and transfer edges. Graph version set to 70 (minimum 69).

<sup>Written for commit bce8858988224699eb0f1d33268555ef20448e52. Summary will update on new commits.</sup>


backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
ajtmccarty, line 1: this migration could end up changing the value of or indexing a non-HFID attribute if it is linked to the same value as an HFID. the logic should either confirm that only HFID Attributes are using a given HFID value OR create a new AttributeValue and transfer the HFID HAS_VALUE edges to the new version

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,127 @@
...
+    async def execute(self, migration_input: MigrationInput) -> MigrationResult:
+        db = migration_input.db
+        try:
+            await self._normalize_and_index_hfid_values(db=db)
+        except Exception as exc:
+            return MigrationResult(errors=[str(exc)])
+
+        return MigrationResult()
+
+    async def _normalize_and_index_hfid_values(self, db: InfrahubDatabase) -> None:
```
ajtmccarty: I think the logic for normalizing HFIDs and indexing them should be split. not necessarily into 2 migrations, but at least two distinct functions/methods. the normalization logic is probably for just a few HFIDs (probably) and requires in-memory Python processing. the index-adding will touch every HFID and can probably be done completely in cypher

backend/infrahub/core/query/node.py
```diff
@@ -1453,34 +1454,59 @@ def __init__(self, node_kind: str, hfids: list[list[str]], **kwargs: Any) -> Non
...
+
         # The list is stored as a string in the database
         self.params["hfid_values"] = [ujson.dumps(hfid) for hfid in self.hfids]
 
         query = """
-        MATCH (n:%(node_kind)s)
+        // --------------------------
+        // Start from matching HFID values and traverse back to nodes
+        // --------------------------
+        MATCH (av:AttributeValue)<-[:HAS_VALUE]-(attr:Attribute {name: "human_friendly_id"})<-[:HAS_ATTRIBUTE]-(n:%(node_kind)s)
```
ajtmccarty: should this be `AttributeValueIndexed` now? or does this have to handle the possible case where an HFID has exceeded the maximum string length too? in that case, I think we should try to get the `AttributeValueIndexed` vertex first and then fall back to searching by node kind/`AttributeValue` only if necessary

another more involved approach would be to only index the first MAXIMUM_STRING_LENGTH characters using a new `value_indexed` property on `AttributeValueIndexed`, but that would be more work than even this approach

backend/infrahub/core/manager.py
```diff
@@ -768,48 +769,34 @@ async def get_one_by_hfid(
...
-            fields=fields,
-            limit=2,
-            filters=filters,
-            branch=branch,
-            at=at,
-            include_metadata=include_metadata,
-            prefetch_relationships=prefetch_relationships,
-            branch_agnostic=branch_agnostic,
-            order=OrderModel(disable=True),
-        )
```
ajtmccarty, line 797: so close to cleaning this up 👍

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,144 @@
...
+ORDER BY elementId(av)
+SKIP $offset LIMIT $limit
+"""
+
+NORMALIZE_VALUES_QUERY = """
+UNWIND $updates AS update
+CALL (update) {
+    MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av)
+    WHERE elementId(old_av) = update.element_id
+    CREATE (new_av:AttributeValue {value: update.new_value, is_default: old_av.is_default})
```
ajtmccarty: I think this should be a `MERGE` and then use a `LIMIT 1` to make sure you don't end up with multiple `new_av` variables and then multiple identical `HAS_VALUE` edges

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,144 @@
...
+MATCH (attr:Attribute {name: "human_friendly_id"})-[:HAS_VALUE]->(av:AttributeValue)
+WHERE NOT av:AttributeValueIndexed
+    AND size(av.value) < $max_length
+    AND NOT EXISTS {
+        MATCH (other:Attribute)-[:HAS_VALUE]->(av)
+        WHERE other.name <> "human_friendly_id"
+    }
+CALL (av) {
+    SET av:AttributeValueIndexed
+} IN TRANSACTIONS OF 1000 ROWS
```
ajtmccarty, line 49: nice subquery with transactions 👍

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,144 @@
...
+    MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av)
+    WHERE elementId(old_av) = update.element_id
+    CREATE (new_av:AttributeValue {value: update.new_value, is_default: old_av.is_default})
+    CREATE (attr)-[new_r:HAS_VALUE]->(new_av)
+    SET new_r = properties(old_r)
+    DELETE old_r
+} IN TRANSACTIONS OF 500 ROWS
+"""
+
+INDEX_HFID_VALUES_QUERY = """
```
ajtmccarty: I think this query is missing handling for the case where an `AttributeValue` is linked to both `(:Attribute {name: "human_friendly_id"})` and `(:Attribute {name: "something_else"})` vertices. I think in HFID and non-HFID attributes case we want to create a new `:AttributeValue|AttributeValueIndexed` vertex and then transfer the `HAS_VALUE` edges to the new `AttributeValueIndexed` vertex

backend/infrahub/core/query/node.py
```diff
@@ -1445,42 +1446,73 @@ class NodeGetByHFIDQuery(Query):
...
         }
-        WITH n, r_part_of
-        WHERE r_part_of.status = "active"
-        MATCH (n)-[:HAS_ATTRIBUTE]->(attr:Attribute {name: "human_friendly_id"})
-        CALL (attr) {
-            MATCH (attr)-[r:HAS_VALUE]->(av)
+        WITH n, attr, av
+        WHERE r.status = "active"
+        // --------------------------
+        // Filter HAS_VALUE edges to resolve the active value
```
ajtmccarty: the ordering if the subqueries seems a little weird to me `IS_PART_OF`, `HAS_VALUE`, then `HAS_ATTRIBUTE` where `HAS_ATTRIBUTE` feels like it should be between `IS_PART_OF` and `HAS_VALUE`. I don't think this matters functionally, so not a blocker
it makes me think of how Americans order dates in the worst possible way

backend/infrahub/core/manager.py
```diff
@@ -768,48 +769,44 @@ async def get_one_by_hfid(
...
-            branch_agnostic=branch_agnostic,
-            order=OrderModel(disable=True),
+        # Try indexed lookup first for performance, fall back to non-indexed for oversized HFIDs that couldn't be indexed.
+        query = await NodeGetByHFIDQuery.init(
+            db=db, branch=branch, at=at, node_kind=kind_str, hfids=[hfid], use_index=True
         )
+        await query.execute(db=db)
+        node_uuids = query.get_node_uuids()
+
+        if not node_uuids:
```
ajtmccarty: I am not sure what the Right approach is here and this might be it, but this does mean that every search miss will end up running the slow version of the query too
maybe we just don't support searching for an HFID over 4,000 characters or whatever the max is

ajtmccarty: also, if you want to keep this fallback, you should add a component test for it

gmazoyer: TBH I don't know about this approach either. Is this likely that a HFID will be longer than 4000 characters? And if so, do we want to support searching for it. My opinion is that a HFID that would go past 4000 characters would not deserve its name anymore (espcially the "human-friendly" part). So maybe not supporting searching very long HFIDs is something we can live with.

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
gmazoyer: You are right. The chance of a HFID sharing its value with another attribute is unlikely (I guess) but not zero. I split the work into two queries. One taking care of HFIDs for which values are not shared with any other attributes. And one taking care of HFIDs sharing values with other attributes.

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+    from neo4j import Record
+
+    from infrahub.database import InfrahubDatabase
+
+console = get_migration_console()
+
+FETCH_HFID_VALUES_QUERY = """
+MATCH (attr:Attribute {name: "human_friendly_id"})-[:HAS_VALUE]->(av:AttributeValue)
+WHERE av.value IS NOT NULL
+RETURN elementId(av) AS element_id, av.value AS value, av:AttributeValueIndexed AS is_indexed
```
ajtmccarty: I think there could be duplicates in the first `MATCH` when there are multiple `HAS_VALUE` edges from one `Attribute` to one `AttributeValue`. I think making this line `RETURN DISTINCT` will remove them

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+ORDER BY elementId(av)
+SKIP $offset LIMIT $limit
+"""
+
+NORMALIZE_VALUES_QUERY = """
+UNWIND $updates AS update
+CALL (update) {
+    MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av)
+    WHERE elementId(old_av) = update.element_id
```
ajtmccarty: I think it is best practice to do all of the `MATCH`ing before the `CALL ... IN TRANSACTIONS` subquery
so I think this should be 

```
MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av)
WHERE elementId(old_av) = update.element_id
CALL (attr, old_r, old_av) {
    ...
} IN TRANSACTIONS OF 500 ROWS
```

backend/infrahub/core/manager.py
ajtmccarty: I'd say that we shouldn't support searching for HFIDs over the max length, but we should confirm that is acceptable with the SA team

backend/tests/component/core/migrations/graph/test_070_index_hfid_values.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+async def test_migration_070_normalizes_and_indexes(db: InfrahubDatabase, default_branch: Branch) -> None:
+    """HFID values with non-string elements should be normalized to all-strings and indexed."""
+    at = current_timestamp()
+
+    create_test_data_query = """
+    CREATE (n1:Node {uuid: "node-1"})
+    CREATE (attr1:Attribute {name: "human_friendly_id", uuid: "attr-1"})
+    CREATE (av1:AttributeValue {value: '["*","*","view","4"]', is_default: false})
```
ajtmccarty: should this be `'["*","*","view",4]'` with an integer `4` to test normalizing?

ajtmccarty (commented):
minus the one outstanding question "do we need to support searching for HFIDs over the maximum character limit?", I think this is ready for official review and approval

backend/tests/component/core/migrations/graph/test_070_index_hfid_values.py
gmazoyer: Good catch

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+MATCH (attr:Attribute {name: "human_friendly_id"})-[:HAS_VALUE]->(av:AttributeValue)
+WHERE av.value IS NOT NULL
+RETURN DISTINCT elementId(av) AS element_id, av.value AS value, av:AttributeValueIndexed AS is_indexed
+ORDER BY element_id
+SKIP $offset LIMIT $limit
+"""
+
+NORMALIZE_VALUES_QUERY = """
+UNWIND $updates AS update
+MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av)
```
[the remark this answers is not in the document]
ajtmccarty: in this case, we want to move EVERY `HAS_VALUE` edge linked to a `human_friendly_id` `Attribute` to the new normalized `AttributeValue`, regardless of branch or timestamp, so we don't need to care about active-ness

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+WHERE elementId(old_av) = update.element_id
+CALL (update, attr, old_r, old_av) {
+    MERGE (new_av:AttributeValue {value: update.new_value, is_default: old_av.is_default})
```
ajtmccarty, line 30: making sure that this vertex is an `AttributeValue:!AttributeValueIndexed` might be the more correct way to do it, but I don't think it actually matters b/c the next migration will handle making sure all of the `AttributeValue` vertices for `human_friendly_id` Attributes are also `AttributeValueIndexed`

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+
+INDEX_EXCLUSIVE_HFID_VALUES_QUERY = """
+MATCH (attr:Attribute {name: "human_friendly_id"})-[:HAS_VALUE]->(av:AttributeValue)
+WHERE NOT av:AttributeValueIndexed
+    AND size(av.value) < $max_length
+    AND NOT EXISTS {
+        MATCH (other:Attribute)-[:HAS_VALUE]->(av)
+        WHERE other.name <> "human_friendly_id"
+    }
+CALL (av) {
```
ajtmccarty, line 47: `WITH DISTINCT av` above this might help reduce the memory required if there are many duplicate `av`s

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+    CREATE (attr)-[new_r:HAS_VALUE]->(new_av)
+    SET new_r = properties(old_r)
+    DELETE old_r
+} IN TRANSACTIONS OF 500 ROWS
+"""
+
+INDEX_EXCLUSIVE_HFID_VALUES_QUERY = """
+MATCH (attr:Attribute {name: "human_friendly_id"})-[:HAS_VALUE]->(av:AttributeValue)
+WHERE NOT av:AttributeValueIndexed
+    AND size(av.value) < $max_length
```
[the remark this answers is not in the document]
ajtmccarty: @gmazoyer this might be right

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
```diff
@@ -0,0 +1,167 @@
...
+            for record in results:
+                update = _needs_normalization(record=record)
+                if update:
+                    updates.append(update)
+
+            if updates:
+                await db.execute_query(query=NORMALIZE_VALUES_QUERY, params={"updates": updates})
+                total_normalized += len(updates)
+
+            offset += self.update_batch_size
```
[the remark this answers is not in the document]
ajtmccarty: the mutation won't affect the FETCH query's results

backend/infrahub/core/query/node.py
```diff
@@ -1453,34 +1454,64 @@ def __init__(self, node_kind: str, hfids: list[list[str]], **kwargs: Any) -> Non
...
         self.params["hfid_values"] = [ujson.dumps(hfid) for hfid in self.hfids]
 
         query = """
-        MATCH (n:%(node_kind)s)
+        // --------------------------
+        // Start from matching HFID values and traverse back to nodes
+        // Uses AttributeValueIndexed for index-backed lookups; HFIDs larger
+        // than MAX_STRING_LENGTH are not indexed and thus not searchable.
+        // --------------------------
+        MATCH (av:AttributeValueIndexed)<-[:HAS_VALUE]-(attr:Attribute {name: "human_friendly_id"})<-[:HAS_ATTRIBUTE]-(n:%(node_kind)s)
```
ajtmccarty, line 1467: it might be safest to start with just the `av:AttributeValueIndexed` match to make sure that index is used by the query planner. if not, the planner _could_ make a bad choice and try to use the index on `Attribute.name` first. not sure how likely that is, but easy to avoid by doing the best index `MATCH` first

backend/infrahub/core/manager.py
```diff
@@ -768,48 +769,34 @@ async def get_one_by_hfid(
...
-            limit=2,
-            filters=filters,
-            branch=branch,
-            at=at,
-            include_metadata=include_metadata,
-            prefetch_relationships=prefetch_relationships,
-            branch_agnostic=branch_agnostic,
-            order=OrderModel(disable=True),
-        )
+        query = await NodeGetByHFIDQuery.init(db=db, branch=branch, at=at, node_kind=kind_str, hfids=[hfid])
```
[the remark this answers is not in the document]
ajtmccarty: I think this is a good point. fixing it might be as simple as passing `branch_agnostic=self.branch_agnostic` into `self.branch.get_query_filter_path` in the updated `NodeGetByHFIDQuery`

dev/guides/backend/creating-migrations.md
```diff
@@ -0,0 +1,165 @@
...
+MATCH (n:SomeNode)
+WHERE <filter>
+CALL (n) {
+    SET n.new_prop = value
+} IN TRANSACTIONS OF 1000 ROWS
+```
+
+`CALL ... IN TRANSACTIONS` requires the `MATCH` to be outside the subquery — move all matching up front.
+
+### Step 4: Handle Shared `AttributeValue` Nodes
```
ajtmccarty, line 114: this seems like it is specific to just this one migration

dev/guides/backend/creating-migrations.md
```diff
@@ -0,0 +1,165 @@
...
+### Step 5: Idempotency
+
+Every migration must be safe to run twice. After a successful run, a second run should be a no-op.
+
+- Check preconditions inside the query (e.g., `WHERE NOT n:AlreadyProcessed`)
+- Skip Python transformations when `new_value == raw_value`
+
+### Step 6: Write a Component Test
+
+Create `backend/tests/component/core/migrations/graph/test_m{NNN}_{name}.py`. Test data is inserted with raw Cypher, the migration runs, and post-state is verified.
```
ajtmccarty, line 147: test data should be inserted with the usual `Node`, `Node.init`, `Node.save()` methods if possible and only raw cypher if necessary

ajtmccarty (approved):
some small comments, but this is a big improvement

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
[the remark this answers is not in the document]
gmazoyer: It is right. The fix is impacting though. We'd need to extract every HFID to process in order to compute the size of each of them from a Python perspective. So that means moving our cypher based normalisation to a python-based approach, e.g. fetch all unindexed HFID values, check byte length in Python, then apply both normalisation and indexing in one pass. I just wonder if this issue really worth it.

Not really related, but I imagine we could have the same issue with `AttributeLengthChecker` in which we use `size()` too.

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
gmazoyer: You mean doing something like `MATCH (attr:Attribute {name: "human_friendly_id"})-[old_r:HAS_VALUE]->(old_av:AttributeValue&!AttributeValueIndexed)`? Probably cannot hurt.

ajtmccarty: yes that's what I mean

backend/infrahub/core/migrations/graph/m070_index_hfid_values.py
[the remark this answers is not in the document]
ajtmccarty: it sounds like it is probably not worth it to me

ajtmccarty (approved):

PR #8753, opened 2026-03-31
IFC-2425 Rework upgrade/migrate UX

gmazoyer:
# Why

The migration/upgrade workflow has accumulated friction on both the
developer and operator side. Adding a new migration means manually
editing `__init__.py`, which is a constant source of merge conflicts.
Operators running `infrahub upgrade` or `infrahub db migrate` get a
flat wall of output without having any way to see what's pending, no
timing, no structure, and noisy internal logs interleaved with the
actual progress.

This PR takes inspiration from Django's migration framework to
improve the experience without how migrations work today.

## What changed

Migrations are now discovered by scanning the `graph/` directory for
`m{NNN}_*.py` files at import time. All the imports are replaced by a
single call to `discover_migrations()`. Duplicate migration numbers
are caught at startup. Adding a migration is now a single-file
operation.

`infrahub db showmigrations` lists all migrations with `[X]`/`[ ]`
applied/pending status. `infrahub db showmigration <N>` inspects a
specific migration (type, status, source file, Cypher queries for
GraphMigration types, docstring for others).
`infrahub db migrate --plan` previews what will execute without
running anything.

Per-migration timing (`Applying X... OK (2.34s)`) are now displayed,
a summary line with total count and elapsed time.
`detect_migration_to_run()` displays structured output: database
version, target version, and each pending migration with its base
type name.

Migration files that previously called `get_migration_console()` now
receive their console via `MigrationInput.console`. In non-verbose
mode, `Console(quiet=True)` is used so migration-internal output is
suppressed. In verbose mode, the real console is passed. This is
combined with `suppress_internal_logs()`, a shared context manager
that raises the `infrahub`/`prefect` loggers to `CRITICAL` to silence
structlog noise during migrations, schema updates, and branch rebase.

Before executing any migration, `_validate_migration_sequence()`
checks gaps in the migration sequence.

## How to review

The core design decisions are in `discovery.py` (auto-discovery),
`shared.py` (DI + log suppression), and `db.py` (CLI commands +
output formatting).

Testing is mostly manual because all this is about user experience
rather than actual code mechanics.

## Impact & rollout

This change should be fully backward-compatible. Apart from the
discovery phase, we do not actually change the graph migration internals.

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../dev/guidelines/changelog.md) added (`uv run towncrier create ...`)
- [x] External docs updated (if user-facing or ops-facing change)

## Output Example

```
% uv run infrahub upgrade --rebase-branches
[...]

PR #8834, opened 2026-04-07
IFC-2440: Add unittest that fails on wrong graph version

solababs:
# Why

Adds a safety net to catch version drift between `GRAPH_VERSION` and the graph migration list. Without this, it's possible to add a migration without bumping `GRAPH_VERSION` (or vice versa) and have no automated signal that the two are out of sync.

## How to test

```bash
uv run pytest backend/tests/unit/core/graph/test_graph_version.py -v
```

To verify the tests actually catch drift, temporarily set `GRAPH_VERSION` in `backend/infrahub/core/graph/__init__.py` and re-run — both tests should fail with descriptive messages.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Adds unit tests to catch graph-version drift (IFC-2440). The tests ensure `GRAPH_VERSION` matches the latest migration `number`, the last migration’s `minimum_version` is `GRAPH_VERSION - 1`, and there are no duplicate `(minimum_version, number)` pairs in the migration list.

<sup>Written for commit 3a15f9f42c30eceae98424552dd5fc7f12dada39. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/core/graph/test_graph_version.py
+++ b/backend/tests/unit/core/graph/test_graph_version.py
@@ -1,0 +1,16 @@
+from infrahub.core.graph import GRAPH_VERSION
+from infrahub.core.migrations.graph import MIGRATIONS
+
+
+def test_graph_version_matches_migration_count() -> None:
+    assert len(MIGRATIONS) == GRAPH_VERSION, (
+        f"GRAPH_VERSION ({GRAPH_VERSION}) must equal to the number of migrations "
+        f"({len(MIGRATIONS)}). Update GRAPH_VERSION or add/remove a migration."
+    )
+
+
+def test_last_migration_minimum_version() -> None:
+    last = MIGRATIONS[-1].init()
+    assert last.minimum_version == GRAPH_VERSION - 1, (
+        f"Last migration minimum_version ({last.minimum_version}) must be GRAPH_VERSION - 1 ({GRAPH_VERSION - 1})."
+    )
--- a/backend/tests/unit/core/graph/test_graph_version.py
+++ b/backend/tests/unit/core/graph/test_graph_version.py
@@ -14,3 +14,9 @@ def test_last_migration_minimum_version() -> None:
     assert last.minimum_version == GRAPH_VERSION - 1, (
         f"Last migration minimum_version ({last.minimum_version}) must be GRAPH_VERSION - 1 ({GRAPH_VERSION - 1})."
     )
+
+
+def test_no_duplicate_migration_minimum_versions() -> None:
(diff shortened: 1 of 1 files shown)


backend/tests/unit/core/graph/test_graph_version.py
```diff
@@ -0,0 +1,16 @@
+from infrahub.core.graph import GRAPH_VERSION
+from infrahub.core.migrations.graph import MIGRATIONS
+
+
+def test_graph_version_matches_migration_count() -> None:
+    assert len(MIGRATIONS) == GRAPH_VERSION, (
+        f"GRAPH_VERSION ({GRAPH_VERSION}) must equal to the number of migrations "
+        f"({len(MIGRATIONS)}). Update GRAPH_VERSION or add/remove a migration."
+    )
```
ogenstad, line 10: I think for this one we'll actually want to start to delete older migrations soon as we don't support upgrading from a too early version of Infrahub regardless.

However it could be nice with another test that validated that no two migrations had the same minimum_version and number.

backend/tests/unit/core/graph/test_graph_version.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_last_migration_minimum_version() -> None:
+    last = MIGRATIONS[-1].init()
+    assert last.minimum_version == GRAPH_VERSION - 1, (
```
ajtmccarty, line 15: #8753 adds a `number` attribute to all migrations, which I think is better to use for this comparison than `minimum_version`
this test might be best just added to @gmazoyer 's PR

backend/tests/unit/core/graph/test_graph_version.py
gmazoyer: Are you thinking about a test verifying that we don't have duplicates for tuples like `(minimum_version, number)`?

Just wanted to mention that, to me, having the same `minimum_version` in multiple migrations should not be a problem as long as those migrations have different `number`.

backend/tests/unit/core/graph/test_graph_version.py
```diff
@@ -0,0 +1,25 @@
...
+        f"Last migration minimum_version ({last.minimum_version}) must be GRAPH_VERSION - 1 ({GRAPH_VERSION - 1})."
+    )
+
+
+def test_no_duplicate_migration_minimum_versions_and_number() -> None:
```
solababs, line 20: @gmazoyer test updated here as per your comment here https://github.com/opsmill/infrahub/pull/8834/changes#r3050170344

gmazoyer (approved):

PR #6416, opened 2025-05-06
fix(backend): IPAM read/list performance improvements

fatih-acar:

PR #8836, opened 2026-04-08
Local test AI bug pipeline #9: filter deleted edges in hierarchy query after merge (closes #8395)

### LINKED ISSUE #8395: Hierarchical ancestry changes in one branch affecting other branches and causing duplicate parents
## Description

A bug has been reported where changing hierarchical ancestry in one branch appears to affect other branches, and in some cases results in nodes showing multiple parents simultaneously.

## Environment

- **Version**: v1.7.4
- **Affected Feature**: Hierarchical nodes with parent/child relationships

## Symptoms

1. **Cross-branch ancestry contamination**: When changing the parent of a hierarchical node in one branch, the change may leak into other branches
2. **Duplicate parents**: After changing ancestry and reverting it back, nodes can end up with two parents at the same time
3. **UI vs Data discrepancy**: The UI may show the correct single parent, but querying the data returns multiple parents
4. **Persistent corruption**: The only known workaround is to delete the affected node and recreate it

## Reproduction Attempt

Wim tested with a simplified schema on v1.7.4 and could **not** reproduce the issue:

```yaml
---
version: "1.0"
generics:
  - name: Generic
    namespace: Location
    hierarchical: true
    attributes:
      - name: name
        kind: Text
        optional: false
nodes:
  - name: Region
    namespace: Location
    inherit_from:
      - LocationGeneric
    parent: ""
    children: LocationCountry
  - name: Country
    namespace: Location
    inherit_from:
      - LocationGeneric
    parent: LocationRegion 
    children: ""
```

**Test steps that worked correctly:**
1. In main branch: Created regions "Europe" and "Americas"
2. In main branch: Created country "Belgium" with parent set to "Europe"
3. Created a new branch
4. In new branch: Changed parent of "Belgium" to "Americas"
5. Switched back to main branch
6. **Expected**: Belgium's parent should still be "Europe" in main
7. **Actual**: Belgium's parent was correctly still "Europe" in main ✓

However, the original reporter is experiencing issues with a more complex hierarchy:
`rack -> room -> site -> metro -> country -> region -> global`

## Additional Context

- **No branch-agnostic nodes** are used in the affected setup
- **Hierarchical tree UI issue**: Wim noted that the hierarchical tree does not properly update when changing hierarchy, which may be a separate but related issue
- **Possible relation to old bug**: Reporter suspects this might be a remnant of an old ancestry issue that was fixed in earlier releases
- **Data level issue confirmed**: This is not just a UI bug - querying the data shows multiple parents

## Workaround

Delete the affected node and recreate it with the correct parent relationship.

## Next Steps

- [ ] Obtain a backup from the affected environment for investigation
- [ ] Test with the complex hierarchy schema (7 levels: rack -> room -> site -> metro -> country -> region -> global)
- [ ] Test the specific sequence: change ancestry -> revert ancestry -> observe duplicate parents
- [ ] Investigate Neo4j relationships for affected nodes
- [ ] Check if this is related to previous ancestry bugs that were fixed

## Reporter

Discord user: overloadbit
Thread date: 2026-02-07

polmichel:
## Why

After merging a branch that changes a node's hierarchical parent, querying the hierarchy on the default branch returns both old and new parents. This causes duplicate parents in the hierarchy tree and data-level inconsistencies.
The reproduction scenario was not detailed, but Claude managed to find a way to reproduce the described issue.

## Issue

Since this commit c50df9bf834009e9d59ae0ff00bb91a45fd77d20, the `NodeGetHierarchyQuery` is not filtering out `DELETED` edges if the query is launched on the default branch.
This comes from this PR which aimed to improve the performance https://github.com/opsmill/infrahub/pull/6416. 

## Solution

Reverting this missing filter when the query is launched on the default branch.

Closes #8395

### What changed

- **Behavioral change:** `NodeGetHierarchyQuery` on the default branch now filters out `status="deleted"` IS_RELATED edges in the initial MATCH clause, so only active hierarchical relationships are traversed.

- **What stayed the same:** The non-default branch code path is unchanged.

### Out of scope
`backend/tests/component/core/diff/test_diff_and_merge.py` test file has been split.

## How to review

- The commits are unitary and would be easier to review one by one than all together.
- Performance considerations should be taken into account during the review, since this was the original purpose of the code before it was reverted by this PR.

## How to test

Run the regression test that reproduces the bug (it fails on `stable` and passes with this fix):

```bash
uv run pytest backend/tests/component/core/diff/diff_and_merge/test_hierarchy.py::TestDiffAndMergeHierarchy::test_hierarchy_parent_change_on_branch_after_merge -x -v
```

Or run the whole `TestDiffAndMergeHierarchy` class to also cover `test_hierarchy_preserved` (ensures the existing ancestor/descendant traversal still works after merge):

```bash
uv run pytest backend/tests/component/core/diff/diff_and_merge/test_hierarchy.py::TestDiffAndMergeHierarchy -v
```

## Impact & rollout

- **Performance:** It may causes impact as this condition has been removed for performance consideration.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`uv run towncrier create ...`)

### DIFF
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2354,6 +2354,10 @@ class NodeGetHierarchyQuery(Query):
         query = """
         MATCH path = (n:Node { uuid: $uuid } )%(filter)s(peer:Node)
         WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
+        """ % {"filter": filter_str, "branch_filter": branch_filter}
+
+        if not self.branch.is_default:
+            query += """
         WITH DISTINCT n, last(nodes(path)) AS peer
         CALL (n, peer) {
             MATCH path = (n)%(filter)s(peer)
@@ -2365,6 +2369,11 @@ class NodeGetHierarchyQuery(Query):
         }
         WITH peer, is_active
         """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
+        else:
+            query += """
+        WITH DISTINCT n, last(nodes(path)) AS peer, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
+        WITH peer, is_active
+            """
 
         self.add_to_query(query)
         where_clause = ["is_active = TRUE"]
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2354,8 +2354,7 @@ class NodeGetHierarchyQuery(Query):
         query = """
         MATCH path = (n:Node { uuid: $uuid } )%(filter)s(peer:Node)
         WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
-        WITH n, collect(DISTINCT last(nodes(path))) AS peers
-        UNWIND peers AS peer
+        WITH DISTINCT n, last(nodes(path)) AS peer
         CALL (n, peer) {
             MATCH path = (n)%(filter)s(peer)
             WHERE all(r IN relationships(path) WHERE (%(branch_filter)s))
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2327,7 +2327,7 @@ class NodeGetHierarchyQuery(Query):
 
         super().__init__(**kwargs)
 
-    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002,PLR0915
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
         hierarchy_schema = self.node_schema.get_hierarchy_schema(db=db, branch=self.branch)
         branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
         self.params.update(branch_params)
@@ -2356,28 +2356,19 @@ class NodeGetHierarchyQuery(Query):
         WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
         WITH n, collect(DISTINCT last(nodes(path))) AS peers
         UNWIND peers AS peer
-
-        """ % {"filter": filter_str, "branch_filter": branch_filter}
-
-        if not self.branch.is_default:
-            query += """
         CALL (n, peer) {
             MATCH path = (n)%(filter)s(peer)
             WHERE all(r IN relationships(path) WHERE (%(branch_filter)s))
             WITH %(with_clause)s
-            RETURN peer as peer1, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
+            RETURN peer AS peer1, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
             ORDER BY branch_level DESC, froms[-1] DESC, froms[-2] DESC, is_active DESC
             LIMIT 1
         }
-        WITH peer1 as peer, is_active
-            """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
-        else:
-            query += """
-        WITH peer
-            """
+        WITH peer1 AS peer, is_active
+        """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
 
         self.add_to_query(query)
-        where_clause = ["is_active = TRUE"] if not self.branch.is_default else []
+        where_clause = ["is_active = TRUE"]
 
         clean_filters = extract_field_filters(field_name=self.direction.value, filters=self.filters)
 
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2353,12 +2353,9 @@ class NodeGetHierarchyQuery(Query):
 
         query = """
         MATCH path = (n:Node { uuid: $uuid } )%(filter)s(peer:Node)
-        WHERE $hierarchy IN LABELS(peer) and all(r IN relationships(path) WHERE (%(branch_filter)s))
-        WITH n, collect(last(nodes(path))) AS peers_with_duplicates
-        CALL (peers_with_duplicates) {
-            UNWIND peers_with_duplicates AS pwd
-            RETURN DISTINCT pwd AS peer
-        }
+        WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
+        WITH n, collect(DISTINCT last(nodes(path))) AS peers
+        UNWIND peers AS peer
 
         """ % {"filter": filter_str, "branch_filter": branch_filter}
 
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2359,11 +2359,11 @@ class NodeGetHierarchyQuery(Query):
             MATCH path = (n)%(filter)s(peer)
             WHERE all(r IN relationships(path) WHERE (%(branch_filter)s))
             WITH %(with_clause)s
-            RETURN peer AS peer1, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
+            RETURN all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
             ORDER BY branch_level DESC, froms[-1] DESC, froms[-2] DESC, is_active DESC
             LIMIT 1
         }
-        WITH peer1 AS peer, is_active
+        WITH peer, is_active
         """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
 
         self.add_to_query(query)
--- a/backend/infrahub/core/query/node.py
+++ b/backend/infrahub/core/query/node.py
@@ -2339,10 +2339,9 @@ class NodeGetHierarchyQuery(Query):
         )
         self.params["hierarchy"] = hierarchy_schema.kind
 
-        if self.direction == RelationshipHierarchyDirection.ANCESTORS:
-            filter_str = f"-{filter_str}->"
-        else:
-            filter_str = f"<-{filter_str}-"
+        filter_str = (
+            f"-{filter_str}->" if self.direction == RelationshipHierarchyDirection.ANCESTORS else f"<-{filter_str}-"
+        )
 
         froms_var = db.render_list_comprehension(items="relationships(path)", item_name="from")
         with_clause = (
@@ -2382,12 +2381,10 @@ class NodeGetHierarchyQuery(Query):
 
         if (clean_filters and "id" in clean_filters) or "ids" in clean_filters:
             where_clause.append("peer.uuid IN $peer_ids")
-            self.params["peer_ids"] = clean_filters.get("ids", [])
-            if clean_filters.get("id", None):
-                self.params["peer_ids"].append(clean_filters.get("id"))
+            extra_id = [clean_filters["id"]] if clean_filters.get("id") else []
+            self.params["peer_ids"] = list(clean_filters.get("ids", [])) + extra_id
 
-        if where_clause:
-            self.add_to_query("WHERE " + " AND ".join(where_clause))
+        self.add_to_query("WHERE " + " AND ".join(where_clause))
 
         self.return_labels = ["peer"]
 
--- a/backend/tests/component/core/diff/test_diff_and_merge.py
+++ b/backend/tests/component/core/diff/test_diff_and_merge.py
@@ -2021,6 +2021,62 @@ class TestDiffAndMerge:
 
         await verify_no_duplicate_paths(db=db)
 
+    async def test_hierarchy_parent_change_on_branch_after_merge(
+        self,
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        hierarchical_location_schema_simple: SchemaRoot,
+    ) -> None:
+        """After merging a branch that changes a node's parent, the hierarchy query on the
+        default branch should only return the new parent.
+        """
+        # Create hierarchy data on the default branch: europe -> paris, north-america -> seattle
+        hierarchy_data = await _build_hierarchical_location_data(db=db, branch=default_branch)
+        europe = hierarchy_data["europe"]
+        north_america = hierarchy_data["north-america"]
+        paris = hierarchy_data["paris"]
+
+        # Verify initial state: paris has parent europe
+        site_schema = paris.get_schema()
+        ancestors_map = await NodeManager.query_hierarchy(
+            db=db,
+            branch=default_branch,
+            id=paris.id,
+            node_schema=site_schema,
+            direction=RelationshipHierarchyDirection.ANCESTORS,
+            filters={},
+        )
+        assert set(ancestors_map.keys()) == {europe.id}
+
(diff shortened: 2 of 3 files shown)


polmichel (commented):
## Test Review: APPROVED

### A. Test Realism

The test uses real production APIs and schema patterns that match actual Infrahub usage:

- `NodeManager.query_hierarchy` is the actual production entry point for hierarchy queries.
- `RelationshipHierarchyDirection.ANCESTORS` is a real enum value used in the codebase.
- The `hierarchical_location_schema_simple` fixture defines a realistic hierarchical schema (Region -> Site -> Rack) matching Infrahub's location hierarchy pattern.
- The `_build_hierarchical_location_data` helper creates nodes with parent relationships using the standard `Node.init` / `new` / `save` flow.
- The branch create / diff / merge flow (`create_branch`, `DiffCoordinator.update_branch_diff`, `DiffMerger.merge_graph`) matches the exact production merge pipeline.

No fictional values or unreachable code paths detected.

### B. Test Correctness

The test asserts the **correct expected behavior**: after merging a parent-change branch, `query_hierarchy(ANCESTORS)` should return only the new parent (north-america), not both old and new parents.

The assertion `set(ancestors_after_merge.keys()) == {north_america.id}` will:
- **FAIL** on current code because `NodeGetHierarchyQuery`'s default-branch path (line 2377-2380 of `node.py`) does not filter `status="deleted"` IS_RELATED edges, causing both europe and north-america to appear.
- **PASS** once the production code is fixed to filter deleted edges on the default branch.

The test exercises the exact code path identified by the analyst: `NodeGetHierarchyQuery.query_init()`, specifically the `else` block at line 2377 that handles the default branch case.

The test cannot pass without modifying the affected production code -- the default branch query path currently has no status filtering.

### C. Test Quality

- **Isolated and deterministic:** The test creates its own hierarchy data from scratch and uses a dedicated branch name ("change-parent"). No shared mutable state with other tests.
- **Follows project conventions:** Placed in the existing `TestDiffAndMerge` class alongside the related `test_hierarchy_preserved` test. Uses the same fixtures (`hierarchical_location_schema_simple`, `db`, `default_branch`) and helper methods (`_get_diff_coordinator`, `_get_diff_merger`).
- **Tests observable behavior:** Asserts on the returned ancestor map from `query_hierarchy`, not on internal query structure or database edge properties.
- **Good assertion message:** The failure message clearly identifies which IDs were expected vs returned, making debugging straightforward.

### D. Alignment with Analysis

The analyst identified:
- **Root cause:** `NodeGetHierarchyQuery` does not filter `status="deleted"` IS_RELATED edges on the default branch.
- **Affected code:** `backend/infrahub/core/query/node.py` lines 2367-2385, specifically the `else` block at line 2379.

The test directly exercises this code path by:
1. Creating a hierarchy on the default branch.
2. Changing a parent on a branch.
3. Merging (which creates `status="deleted"` edges for the old relationship on the default branch).
4. Querying hierarchy on the default branch (hitting the `else` block).
5. Asserting only the new parent is returned.

This is a well-scoped test that targets exactly the identified bug without testing unrelated behavior.

### Recommended Next Steps

Proceed to the fixer agent. The test is well-constructed and correctly reproduces the bug.

polmichel (commented):
## Fix Review: APPROVED

> Note: Cannot submit as formal approval because the authenticated GitHub user is the PR author. This review should be treated as an APPROVED verdict.

### A. Correctness

The fix correctly addresses the root cause. The analyst identified that `NodeGetHierarchyQuery` does not filter `status="deleted"` IS_RELATED edges on the default branch, and the fix adds exactly that filtering.

The approach matches the analyst's recommended strategy: add `status="active"` filtering to the initial MATCH for the default branch, rather than modifying the merge logic or the non-default branch path.

The Cypher logic is sound. On the default branch, the query now requires:
```
all(r IN relationships(path) WHERE r.status = "active")
```
This prevents traversal through deleted edges that represent old parent relationships after merge.

The non-default branch path is correctly left untouched — it already handles `is_active` in its CALL subquery, and adding early filtering there would break the priority resolution logic that needs to see all edges.

One minor observation: the generated Cypher contains two separate `all()` calls over `relationships(path)` — one for the branch filter, one for the status filter. These could be combined into a single `all(r IN relationships(path) WHERE (<branch_filter>) AND r.status = "active")` for marginal efficiency, but the current form is correct and arguably more readable.

**Edge cases covered:** The fix handles multiple deleted edges for the same relationship vertex (which can accumulate after repeated parent changes and merges) because `all()` requires every edge in the path to be active.

### B. Code Quality

The change is minimal and targeted: 3 lines modified in the query template. No unnecessary refactoring.

The string interpolation pattern (`%(active_filter)s`) follows the existing style used throughout this method for `%(filter)s` and `%(branch_filter)s`.

The conditional assignment on line 2354 is clean and readable. No performance concerns — adding a WHERE clause filter to Neo4j can only reduce the result set, potentially improving performance by pruning paths early.

The fix aligns with the existing codebase pattern: the non-default branch path already filters by `is_active`, and `RelationshipGetPeerQuery` (used for node detail views) also checks `is_active = TRUE`. This fix brings the default branch hierarchy query in line with these existing patterns.

No public API changes. The changelog entry is well-written from the user's perspective.

### D. Test Quality

The test is well-constructed:
- Creates a realistic hierarchical schema (Region -> Site)
- Establishes initial parent relationship on default branch
- Changes parent on a branch, merges back
- Asserts only the new parent is returned after merge
- Includes a descriptive assertion message with IDs for debugging

The test exercises the exact production code path (`NodeManager.query_hierarchy` -> `NodeGetHierarchyQuery`) and cannot pass without the fix.

**Suggestion (non-blocking):** A second test case verifying the DESCENDANTS direction after a parent change would strengthen coverage, since `NodeGetHierarchyQuery` handles both directions.

### Unrelated files

The PR includes 4 files under `.github/bug-agent-pipeline/` that are agent pipeline infrastructure, not part of the bug fix. These should ideally be in a separate PR but don't affect correctness.

### Recommended Next Steps

1. Merge as-is. The fix is correct, minimal, and well-tested.
2. Consider extracting `.github/bug-agent-pipeline/` files into a separate PR (non-blocking).
3. Consider adding a DESCENDANTS direction test in a follow-up (non-blocking).

ajtmccarty:
I'm actually already in the process of refactoring the `test_diff_and_merge.py` tests to be more thorough and use smaller files. so it might be best to wait on this one for a day or so until I can get that PR reviewed and merged

backend/infrahub/core/query/node.py
```diff
@@ -2353,34 +2353,21 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
-        if not self.branch.is_default:
-            query += """
+        WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
+        WITH DISTINCT n, last(nodes(path)) AS peer
         CALL (n, peer) {
             MATCH path = (n)%(filter)s(peer)
             WHERE all(r IN relationships(path) WHERE (%(branch_filter)s))
             WITH %(with_clause)s
-            RETURN peer as peer1, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
+            RETURN peer AS peer1, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
```
ajtmccarty: I think you can just skip returning the `peer AS peer1` here completely

backend/tests/component/core/diff/diff_and_merge/test_hierarchy.py
```diff
@@ -0,0 +1,136 @@
...
+            id=paris.id,
+            node_schema=site_schema,
+            direction=RelationshipHierarchyDirection.ANCESTORS,
+            filters={"id": None},
+        )
+        assert set(ancestors_after_merge.keys()) == {north_america.id}, (
+            f"Expected only north-america ({north_america.id}) as ancestor of paris after merge, "
+            f"but got: {set(ancestors_after_merge.keys())}. "
+            f"europe ({europe.id}) should no longer be an ancestor."
+        )
```
ajtmccarty: can add a call to `verify_graph` from `db_validation.py` here once I merge my merge-testing changes in

ajtmccarty (commented):
query update looks good. not sure what we were thinking with skipping the `is_active = TRUE` check
only not approving now, b/c the tests will have some changes after my PR to update the diff_and_merge tests

fatih-acar:
I will give a try to this change regarding the performance issue, I had dropped a message in slack explaining why... should be able to confirm if it introduces a regression or not

backend/infrahub/core/query/node.py
polmichel: I just have to keep the `peer` without aliasing, it is used later in the query in the `where_clause` and `with_clause` objects.

polmichel:
This PR will go to draft state waiting for the test refactoring PR to be merged

backend/infrahub/core/query/node.py
```diff
@@ -2353,34 +2353,21 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
-        else:
-            query += """
-        WITH peer
-            """
+        WITH peer, is_active
+        """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
 
         self.add_to_query(query)
-        where_clause = ["is_active = TRUE"] if not self.branch.is_default else []
+        where_clause = ["is_active = TRUE"]
```
fatih-acar, line 2378:
```suggestion
        WHERE $hierarchy IN LABELS(peer) AND all(r IN relationships(path) WHERE (%(branch_filter)s))
        """ % {"filter": filter_str, "branch_filter": branch_filter}

        if not self.branch.is_default:
            query += """
        WITH DISTINCT n, last(nodes(path)) AS peer
        CALL (n, peer) {
            MATCH path = (n)%(filter)s(peer)
            WHERE all(r IN relationships(path) WHERE (%(branch_filter)s))
            WITH %(with_clause)s
            RETURN all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
            ORDER BY branch_level DESC, froms[-1] DESC, froms[-2] DESC, is_active DESC
            LIMIT 1
        }
        WITH peer, is_active
        """ % {"filter": filter_str, "branch_filter": branch_filter, "with_clause": with_clause}
        else:
            query += """
        WITH DISTINCT n, last(nodes(path)) AS peer, all(r IN relationships(path) WHERE (r.status = "active")) AS is_active
        WITH peer, is_active
            """

        self.add_to_query(query)
        where_clause = ["is_active = TRUE"]
```

fatih-acar (commented):
So it looks like this introduces a performance regression when querying the default branch in large datasets, we should keep the "branch traversal bypass" trick.
The following graphql query takes more than 10s vs 0.1s (100x slower) in my testing:
This is one of the queries that are executed on the main IPAM page.
```
query GET_IPAM_TREE_NODES($isTopLevel: Boolean, $parentIds: [ID!], $search: String, $ipNamespaceIds: [ID!], $limit: Int, $offset: Int) {
  BuiltinIPPrefix(
    is_top_level__value: $isTopLevel
    parent__ids: $parentIds
    any__value: $search
    partial_match: true
    ip_namespace__ids: $ipNamespaceIds
    offset: $offset
    limit: $limit
  ) {
    edges {
      node {
        id
        display_label
        descendants {
          count
          __typename
        }
        __typename
      }
      __typename
    }
    __typename
  }
}


{
    "ipNamespaceIds": [
        "18a49e76-7b31-20d0-3c98-1065be19fd0e"
    ],
    "limit": 80,
    "offset": 0,
    "isTopLevel": true
}
```

backend/infrahub/core/query/node.py
fatih-acar: The snippet above restores original performance gains, but I'm not sure if this fixes the issue we are trying to solve...

polmichel: Thank you for the performance test and feedback. 
It actually does pass the new integration test, as the `is_active` filter is back.

I will encapsulate the `is_active` filter so that the code clearly shows that it is an key part of the `default_branch` condition tree.

polmichel:
> query update looks good. not sure what we were thinking with skipping the `is_active = TRUE` check only not approving now, b/c the tests will have some changes after my PR to update the diff_and_merge tests

I think we can merge this one without waiting for the refactoring, @ajtmccarty. The PR targets `stable` and is a bug fix. I will remove the commit about splitting the test file.
If there are any merge conflicts with the `develop` branch, I will resolve them manually and apply once again the test file split if necessary and your suggestion about the additional call . If the code is ported to `develop` before your pull request is merged, I will create an additional commit on your branch.
Does that sound like a good plan?

backend/infrahub/core/query/node.py
```diff
@@ -2339,59 +2341,58 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
+            f"-{filter_str}->" if self.direction == RelationshipHierarchyDirection.ANCESTORS else f"<-{filter_str}-"
+        )
 
         froms_var = db.render_list_comprehension(items="relationships(path)", item_name="from")
         with_clause = (
             "peer, path,"
             " reduce(br_lvl = 0, r in relationships(path) | CASE WHEN r.branch_level > br_lvl THEN r.branch_level ELSE br_lvl END) AS branch_level,"
             f" {froms_var} AS froms"
         )
+        is_active_clause = 'all(r IN relationships(path) WHERE (r.status = "active")) AS is_active'
```
ajtmccarty: duplicates `IS_ACTIVE_CLAUSE` on line 2313

also, I think it is more readable to just include this line in the query where required than to copy it in using string interpolation or an f-string. it makes it harder to read the query as a single entity if you have to go read other variables

ajtmccarty (approved):
one small suggested change

backend/infrahub/core/query/node.py
polmichel: The advantage of this encapsulation is that it makes it clear that this code represents the same logic and is important for both variations of the query. This avoids one part being removed in future without the other.
But it does heavily impact the readability, I am ok to remove it

polmichel: And sorry I have missed the duplication

ajtmccarty:
> > query update looks good. not sure what we were thinking with skipping the `is_active = TRUE` check only not approving now, b/c the tests will have some changes after my PR to update the diff_and_merge tests
> 
> I think we can merge this one without waiting for the refactoring, @ajtmccarty. The PR targets `stable` and is a bug fix. I will remove the commit about splitting the test file. If there are any merge conflicts with the `develop` branch, I will resolve them manually and apply once again the test file split if necessary and your suggestion about the additional call . If the code is ported to `develop` before your pull request is merged, I will create an additional commit on your branch. Does that sound like a good plan?

I made sure to add your new test to my updated merge tests in #9042 before merging

polmichel:
> > > query update looks good. not sure what we were thinking with skipping the `is_active = TRUE` check only not approving now, b/c the tests will have some changes after my PR to update the diff_and_merge tests
> > 
> > 
> > I think we can merge this one without waiting for the refactoring, @ajtmccarty. The PR targets `stable` and is a bug fix. I will remove the commit about splitting the test file. If there are any merge conflicts with the `develop` branch, I will resolve them manually and apply once again the test file split if necessary and your suggestion about the additional call . If the code is ported to `develop` before your pull request is merged, I will create an additional commit on your branch. Does that sound like a good plan?
> 
> I made sure to add your new test to my updated merge tests in #9042 before merging

Thank you !

PR #8842, opened 2026-04-08
IFC-2260: Replace display_labels with display_label on internal schemas

solababs:
# Why

Replace `display_labels` with `display_label` on internal schemas as `display_labels` is deprecated.

---
## Summary by cubic
Replaces deprecated display_labels with display_label across internal, core, and action schemas per IFC-2260. Adds a post-uniqueness validator to convert legacy labels, validate Jinja templates, and inherit labels safely.

- **Refactors**
  - Switched schema definitions and generated profiles/templates to display_label; legacy display_labels is optional (None) and migrated.
  - Added `DisplayLabelValidator` (called after uniqueness) to convert lists to a path or Jinja template, validate paths/templates, inherit from a single parent generic when missing, skip validation for inherited nodes, and update duplicate schemas to avoid cache issues.
  - Updated `NodeSchema.generate_fields_for_display_label` to support simple paths and Jinja2 variables (and legacy lists).
  - Increased CI test job timeout from 45 to 50 minutes.

<sup>Written for commit 47a4db5d5692ca0d7395de95877481e8d9edba47. Summary will update on new commits.</sup>


ajtmccarty (approved):

backend/infrahub/core/schema/definitions/internal.py
```diff
@@ -182,7 +182,7 @@ class SchemaNode(BaseModel):
     attributes: list[SchemaAttribute]
     relationships: list[SchemaRelationship]
     display_label: str | None = None
-    display_labels: list[str]
+    display_labels: list[str] = Field(default_factory=list)
```
ogenstad: I'm wondering about this one since we're moving away from "display_labels". Should this perhaps be changed to this?

```python
display_labels: list[str] | None = None
```

The reasoning is that this is what we have on the `GeneratedBaseNodeSchema` class.

Could you look into this

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -661,8 +662,9 @@ def process_validate(self) -> None:
         self.validate_identifiers()
         self.sync_uniqueness_constraints_and_unique_attributes()
         self.validate_uniqueness_constraints()
-        self.validate_display_labels()
-        self.validate_display_label()
+        # Cant move DisplayLabelValidator into the validators yet as the validation sequence would be broken
+        # validate_uniqueness_constraints needs to run before display label validation
+        DisplayLabelValidator().check(schema_branch=self)
```
ajtmccarty, line 667: 👍 one step closer to dependency injection

ajtmccarty (approved):

PR #8844, opened 2026-04-08
Move menu tests to unit

ogenstad:
# Why

Move tests that doesn't require network access into unit tests

## What changed

* Relocate tests

---
## Summary by cubic
Moved menu tests that don’t need network access into unit tests to speed up and stabilize CI.

- **Refactors**
  - Relocated menu tests to `backend/tests/unit/menu/`.
  - Added `conftest.py` with shared `MenuItemDefinition` fixtures.
  - Converted async tests in `test_models.py` to sync (no event loop, no I/O).
  - No functional changes to menu models or test coverage.

<sup>Written for commit a7825f402234878fcec791f43c214543d5be22d6. Summary will update on new commits.</sup>


backend/tests/unit/menu/conftest.py
ajtmccarty, line 1: not required, but wondering if this could this move up to the `tests/conftest` level instead of being duplicated? I don't really want to add more to `tests/conftest.py` because it is already pretty big, but there's got to be some more organized way to share the fixture between `component` and `unit` tests. unless this version is different than the `component` version

ajtmccarty (approved):

backend/tests/unit/menu/conftest.py
ogenstad: Ah yes, completely agree missed that they got duplicated. I'll look into that later but I agree it doesn't make sense to have the same thing twice.

PR #8847, opened 2026-04-09
move merge logic to database level mostly

ajtmccarty:
# Why

Merge operation takes kind of a long time to run. It involves loading an entire diff into memory (which can be very large), then serializing it in chunks and passing it into merge queries. It would be faster and simpler to do as much of the merge as possible in cypher queries and skip the in-memory piece.

This PR refactors the merge logic and adds new cypher queries that handle almost the entire merge using only cypher queries on the database. The exception is any Nodes with conflicts, which need to be loaded into memory and examined to determine how to handle them.

Closes IFC-2438

## What changed

The big changes are
- new merge cypher queries in `bulk_merge.py`
- refactored logic in `DiffMerger.merge_graph` to use the new queries
- a couple of new diff-related queries for retrieving the UUIDs of all nodes in the diff and the UUIDs of all nodes with conflicts in the diff
- new logic for handling conflicts that determines which Nodes, Relationships, and Properties should be excluded from the merge (because there was a conflict and the user selected the target/default/base branch option to resolve the conflict)
  - there is some extra special handling for a few cases on conflicted cardinality-one relationships. if there are conflicts on a cardinality-one relationship for both the peer and a property (source, owner, is_protected), then a user can select different branches for each of the conflicts and this requires some unique logic to arrive at the end state expected by the user on the default branch. fortunately, conflicts are unusual and cross-resolved conflicts even more rare

### Suggested review order
1. Probably start with `DiffMerger` to see how things fit together over all
2. Then the new cypher queries
3. Then everything else

## Testing

Unit/component/integration tests were recently refactored and updated to provide more coverage for the merge logic. This PR adds a few more tests.
I've done performance and scale testing on very large databases and observed significant improvement in execution time and no OOM crashes from neo4j locally.

## Checklist

- [x] Tests added/updated


backend/infrahub/core/branch/tasks.py
```diff
@@ -340,8 +340,10 @@ async def _do_merge_branch(
             diff_locker=DiffLocker(),
             workflow=workflow,
         )
-        branch_diff = await merger.merge(at=merge_at)
+        await merger.merge(at=merge_at)
 
+    log.info("Loading enriched diff for changelog collection")
+    branch_diff = await diff_repository.get_one(diff_branch_name=obj.name, tracking_id=BranchTrackingId(name=obj.name))
```
ajtmccarty: diff loading moved up a few layers to here b/c the merge logic no longer needs to load the entire diff

backend/infrahub/core/diff/query/affected_diff_nodes.py
ajtmccarty, line 1: one of 2 new queries to get exactly the information we need from the diff instead of loading the whole thing

backend/infrahub/core/diff/repository/repository.py
```diff
@@ -548,6 +553,38 @@ async def get_all_conflicts_for_diff(
         for conflict_path, conflict_node in query.get_conflict_paths_and_nodes():
             yield (conflict_path, self.deserializer.deserialize_conflict(diff_conflict_node=conflict_node))
 
+    async def get_conflicted_node_uuids(
```
ajtmccarty, line 556: integrating the new queries into diff repository to make them easier to call

backend/infrahub/core/merge/branch_merger.py
ajtmccarty, line 1: `DiffMerger.merge_graph` no longer returns a diff, so neither does `BranchMerger.merge`

backend/infrahub/core/diff/merger/merger.py
```diff
@@ -114,17 +118,81 @@ async def merge_graph(self, at: Timestamp) -> EnrichedDiffRoot:
...
+        at: Timestamp,
+        conflict_uuids: set[str],
+        tracking_id: BranchTrackingId,
+    ) -> None:
+        """Load conflicted nodes from the diff and merge them via the old serializer path."""
+        enriched_diff = await self.diff_repository.get_one(
+            diff_branch_name=self.source_branch.name,
+            tracking_id=tracking_id,
+            filters=EnrichedDiffQueryFilters(ids=list(conflict_uuids)),
+        )
```
ajtmccarty: get just the objects that we need from the diff instead of the whole thing. this will usually be a small fraction of the entire diff

backend/tests/component/core/diff/test_diff_merger.py
ajtmccarty, line 1: these tests were refactored to be significantly smaller by using the database instead of faked diffs in memory

backend/infrahub/core/diff/merger/exclusion_plan.py
ajtmccarty, line 1: logic in here is for going through a diff and determining how to handle different conflicts. if the base/target/default branch is selected for a given conflict, that (generally) means we should ignore it during merge operations b/c the user has selected to keep the state already present on the default branch. if the diff/source/merging branch is selected, then we (generally) proceed the same as we would have if there was no conflict, b/c we want to merge the source branch state into the default branch

there is some special handling required for cardinality-one relationships
- if there is a conflict and the source/merging branch option is selected, we need to make sure to set the peer on the default/target branch to deleted so that there are not multiple peers, which would break the schema
- if there are cross-branch selections (I want the peer from the base/target/default branch and the `source` from the source/merging branch), then we need to do some extra special handling to make sure that the right edges are copied to the right vertices

backend/infrahub/core/diff/query/bulk_merge.py
ajtmccarty, line 1: new cypher queries for doing as much of the merge as possible at the database level

backend/tests/component/core/diff/merge/test_manual_conflict_resolution.py
```diff
@@ -348,3 +349,110 @@ async def test_base_delete_with_added_branch_relationship(
...
+    else:  # base_action == "update"
+        person_main = await NodeManager.get_one(db=db, id=person_john_main.id)
+        person_main.height.value = 200
+        await person_main.save(db=db, user_id="main-user")
+    after_base_update = Timestamp()
+
+    if diff_action == "delete":
+        person_branch = await NodeManager.get_one(db=db, branch=branch2, id=person_john_main.id)
+        await person_branch.delete(db=db, user_id="branch-user")
+    else:  # diff_action == "update"
```
ogenstad, line 407: These looks a bit strange, would we loose anything by just having an elif diff_action == "update"?

backend/tests/component/core/diff/merge/test_mixed_cardinality_one_conflicts.py
```diff
@@ -0,0 +1,323 @@
...
+
+    paris_after = await NodeManager.get_one(db=db, id=paris.id)
+    parent_rel = await paris_after.get_relationship("parent").get(db=db)
+    assert parent_rel.peer_id == initial["asia"].id
+    parent_source = await parent_rel.get_source(db=db)
+    assert parent_source is not None
+    assert parent_source.id == initial["paris-r2"].id
+
+    assert await _children_ids(db=db, branch=default_branch, site=paris) == expected_children
+    # TODO assert children have no source metadata
```
ogenstad: Is this something for this iteration? From now on I think we should avoid adding further TODO entries inline in code and instead create issues that points to specific commits and lines if we want some follow up work.

ogenstad (approved):
Nice work! With these changes in mind. I'm thinking that we should create a card in JPD for "Improve branch merge performance", even if we just link it to this job that has already been done. But I want to highlight it in the release and also assign someone that tests it out with some real data before we go live.

backend/tests/component/core/diff/merge/test_mixed_cardinality_one_conflicts.py
ajtmccarty: yes it is. thanks for pointing it out. I am all for no more TODOs in the code

PR #8852, opened 2026-04-09
Forbid to null an attribute with a default_value in the schema

ogenstad:
# Why

Currently Infrahub will update the mandatory/optional properties of an attribute if that attribute has a default_value assigned in the schema. Because we do this it also means that we'd allow that attribute to be updated to have a `null` value. On creation it would still have a default value but after the fact we can update it to a `null` value which is not always what a user would expect.

With some additional validations we will instead look at fixing this behaviour within Infrahub. For now this PR will assume that the intention of these attributes is to disallow updates that would set a `null` value.

## What changed

* Extra validation to raise an error if a user tries to violate this
* Tests

---
## Summary by cubic
Prevent updates that set an attribute with a schema `default_value` to null, so these attributes can’t be nulled after creation. This avoids unexpected nulls and keeps updates aligned with schema intent.

- **Bug Fixes**
  - Added validation to reject `value: null` when the attribute has a `default_value`.
  - Added GraphQL tests for update and upsert to ensure errors are raised and existing values are preserved.

<sup>Written for commit 2cd8046d78e812624fd907e7a6e2befebf23f61c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -617,6 +617,8 @@ class BaseAttribute(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
                 value_to_set = self.schema.convert_value_to_enum(data["value"])
             else:
                 value_to_set = data["value"]
+            if value_to_set is None and self.schema.default_value is not None:
+                raise ValidationError({self.name: f"'{self.name}' has a default value and cannot be set to null"})
             if value_to_set != self.value:
                 self.value = value_to_set
                 changed = True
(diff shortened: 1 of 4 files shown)


backend/infrahub/core/attribute.py
```diff
@@ -617,6 +617,8 @@ async def from_graphql(self, data: dict, db: InfrahubDatabase, process_pools: bo
                 value_to_set = self.schema.convert_value_to_enum(data["value"])
             else:
                 value_to_set = data["value"]
+            if value_to_set is None and self.schema.default_value is not None:
+                raise ValidationError({self.name: f"'{self.name}' has a default value and cannot be set to null"})
```
ajtmccarty, line 621: is there a bug or some documentation related to this change? I'd expect a default value to be set at creation time and then be updatable to any legal value, `null` included

PR #8862, opened 2026-04-10
IFC-2434: Unify Web hooks queries data sources

polmichel:
# Why

This PR is a follow-up task to the bug fix at the following link: : https://github.com/opsmill/infrahub/pull/8777.

Maintaining two different data sources to query web hooks (SDK client & database through `NodeManager`) and inject them into Prefect-related code resulted in a bug due to the data formats being different.
`Enum` object are unserialized as strings from the database and as Python `Enum` from the SDK client.

I believe unifying these data sources and using only the database through NodeManager is the correct way to manage these queries.

Additionally, as @ogenstad mentioned, the `infrahub upgrade` command does not have access to the SDK client, and uses this workflow through `infrahub.trigger.tasks.trigger_configure_all` -> `infrahub.webhook.gather.gather_trigger_webhook`. Therefore, even if we wanted to use the SDK client, it would not be possible.

Closes [IFC-2434]

## What changed

- The `infrahub.webhook.tasks.configure._configure_one` of the `"webhook-configure"` Prefect flow no longer uses the SDK client to retrieve the configured web hook but goes through the `NodeManager` layer.
- New component and unit tests have been added to cover the different use cases.
- Class method `WebhookTriggerDefinition.from_object` was not following the SRP, I have broken down it in focused utility methods or classes.

## Checklist

- [x] Tests added/updated

[IFC-2434]: https://opsmill.atlassian.net/browse/IFC-2434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Unifies webhook trigger construction via the DB-backed `NodeManager` using `WebhookTriggerDefinitionBuilder`, and splits Prefect automation into `WebhookAutomation` and `WebhookAutomationPrefectSyncer`. Aligns with IFC-2434 and lets `infrahub upgrade` configure webhook automations without the SDK.

- **Bug Fixes**
  - Single DB-backed query path fixes enum serialization mismatches that broke automation setup.
  - Triggers honor `registry.default_branch` for branch scopes and apply `node_kind` filters, including when `event_type="all"`.

- **Refactors**
  - Added `WebhookTriggerDefinitionBuilder`; `gather_trigger_webhook` builds triggers with the default branch injected; removed `from_object`.
  - Introduced `WebhookAutomation` + `WebhookAutomationPrefectSyncer.apply()`; replaced `_configure_one` with `SingleWebhookConfigurer` using `NodeManager.get_one` (with awaited `get_database()`), and invalidating the webhook cache after sync.
  - Extracted `generate_webhook_automation_name`; used in `_delete_automation` and by `WebhookTriggerDefinition.generate_name`. Added unit and component tests for the builder and syncer; obsolete tests removed. Docs: `"all"` maps to `"infrahub.*"`.

<sup>Written for commit 5860212bb12dce54347437a50609c33fb7e649cb. Summary will update on new commits.</sup>


backend/infrahub/webhook/models.py
```diff
@@ -93,7 +91,30 @@ def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
             ],
         )
 
-        return definition
+
+def webhook_to_trigger(webhook: CoreWebhook, default_branch: str) -> WebhookTriggerDefinition:
```
ajtmccarty: this method and `WebhookTriggerData.to_trigger` feel like they should be a `WebhookTriggerDefinitionBuilder` component that requires the `default_branch` as a dependency

backend/infrahub/webhook/tasks/configure.py
```diff
@@ -117,8 +119,9 @@ async def _configure_one(
     """
     log = get_run_logger()
 
-    webhook = await get_client().get(kind=CoreWebhook, id=webhook_id)
-    trigger = WebhookTriggerDefinition.from_object(webhook)
+    database = await get_database()
+    webhook_node = await NodeManager.get_one(id=webhook_id, db=database, kind=CoreWebhookNode, raise_on_error=True)
+    trigger = webhook_to_trigger(webhook_node, registry.default_branch)
```
ajtmccarty: given the new dependency on the database and the suggested `WebhookTriggerDefinitionBuilder` component below, I think this method should be a `WebhookPrefectConfigurer` component that depends on database, `WebhookTriggerDefinitionBuilder`, and `prefect_client` (and probably `NodeManager` to be completely correct).

ajtmccarty (approved):
I've left some optional dependency injection suggestions

backend/infrahub/webhook/models.py
polmichel: This is definitely cleaner. 
While moving into this refactoring path, I usually prefer not to use Builder objects because I think they hide declarative code behind their 'OOP' appearance

I would rather give this responsibility to a `WebhookTrigger` object, which should be aware of how to produce its own definition using its `to_definition()` method. This would remove the Data object and allows us to inject the dependency through this new object constructor which will be a composition of `CoreWebHook` object. Does that sound correct to you?

polmichel: see 1e05bebc1ffe1df4927895012ff5a25836ccfdd1

backend/tests/unit/webhook/test_webhook_automation.py
```diff
@@ -0,0 +1,142 @@
...
+    async def test_active_no_existing_creates(self, prefect_client: PrefectClient) -> None:
+        trigger_def = _make_trigger_definition()
+        automation = WebhookAutomation(trigger_definition=trigger_def, active=True)
+
+        _stub_no_automations(prefect_client)
+        _stub_deployment(prefect_client)
+
+        await automation.apply(prefect_client)
+
+        prefect_client.create_automation.assert_called_once()  # type: ignore[attr-defined]
```
polmichel, line 99: We definitely need a Prefect client adapter for these use cases I think, I'll add this topic back again as a discussion
cc @ogenstad

backend/infrahub/webhook/tasks/configure.py
polmichel: I did create a `WebhookAutomation` that has a `apply` method taking as parameter a `PrefectClient` object -> https://github.com/opsmill/infrahub/pull/8862/changes/6471d010e89dffbf02ccf989ea5a6cf2f2a59466
What do you think?

backend/infrahub/webhook/models.py
ajtmccarty: the `WebhookTrigger` class is not really acting like a model. Ideally, a model should just be a data container that defines structure and (if it is simple enough) can be used to serialize itself into different formats. In this case `WebhookTrigger` is just used as an intermediate step to call `definition()` on the object, which is a sign that the "model" is really a data transformer class of some sort.

And a model shouldn't require a private variable to track the `default_branch` b/c most models should not require logic related to the default branch. In this case `default_branch` is really a dependency of the logic in `WebhookTrigger`. 

So this still looks like a `WebhookTriggerDefinitionBuilder` with a `build` method to me. something like this

```
# trigger_definition_builder.py
class WebhookTriggerDefinitionBuilder:
    def __init__(self, default_branch: str) -> None:
        self.default_branch = default_branch

    def build(self, webhook: CoreWebHook) -> WebhookTriggerDefinition:
        ...
```

backend/infrahub/webhook/tasks/configure.py
ajtmccarty: I think `WebhookAutomation` should accept `prefect_client` as a dependency and be constructed within `configure_webhook` and then passed into `_configure_one`. 

Really, `_configure_one` should be its own component that accepts `WebhookAutomation` (and `InfrahubDatabase) as a dependency within its constructor.

And `active` looks like it should be the only argument to `apply` instead of a dependency

backend/infrahub/webhook/models.py
polmichel: I agree this model does not make sense as an object, this would need a larger refactoring to break down these pieces into meaningful objects. Having it in a semi-refactoring state does not sound like a good idea.
Your solution is better.

backend/infrahub/webhook/tasks/configure.py
polmichel: Thanks I will try to think more about which objects should or should not be injected as parameter or as dependencies.

Regarding 
> Really, _configure_one should be its own component that accepts WebhookAutomation (and `InfrahubDatabase) as a dependency within its constructor.

I did a SingleWebhookConfigurer

polmichel: The `NodeManager` will stay outside of the `Configurer` class if that sounds ok. It allows to externalize the dependency -> inject directly the web hook node.

polmichel:
I will review once again the dependencies, not completely done yet

backend/infrahub/webhook/models.py
```diff
@@ -93,7 +94,77 @@ def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
...
+
+class WebhookTriggerDefinition(TriggerDefinition):
+    id: str
+    type: TriggerType = TriggerType.WEBHOOK
+
+    def generate_name(self) -> str:
+        return generate_webhook_automation_name(self.id)
+
+
+class WebhookAutomation:
```
ajtmccarty, line 110: `WebhookAutomation` doesn't quite describe what this class does. it seems more like a `WebhookAutomationPrefectUpdater` to me, but not totally sure

backend/infrahub/webhook/models.py
```diff
@@ -93,7 +94,77 @@ def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
...
+
+    async def _find_existing(self) -> Automation | None:
+        from infrahub.trigger.setup import gather_all_automations
+
+        all_automations = await gather_all_automations(client=self._client)
+        matches = [a for a in all_automations if a.name == self.name]
+        return matches[0] if matches else None
+
+    async def _as_prefect_automation(self) -> AutomationCore:
+        from prefect.automations import AutomationCore as _AutomationCore
```
ajtmccarty, line 153: imports at the top

backend/infrahub/webhook/models.py
```diff
@@ -93,7 +94,77 @@ def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
...
+    def generate_name(self) -> str:
+        return generate_webhook_automation_name(self.id)
+
+
+class WebhookAutomation:
+    """A webhook's desired automation state in Prefect."""
+
+    def __init__(self, trigger_definition: WebhookTriggerDefinition, prefect_client: PrefectClient) -> None:
+        self._trigger_definition = trigger_definition
+        self._client = prefect_client
```
ajtmccarty, line 115: I'd say that the dependency should just be the `prefect_client`, b/c the class depends on `prefect_client` to work. `WebhookTriggerDefinition` should be passed into `apply` b/c it is the entity that the class operates on. You should be able to create one instance of this class and then call it multiple times on different trigger_definitions if required

backend/infrahub/webhook/tasks/configure.py
```diff
@@ -85,6 +91,23 @@ async def _configure_webhook_on_failure(flow: Flow, flow_run: FlowRun, state: St
     )
 
 
+class SingleWebhookConfigurer:
+    """Configures a single webhook's Prefect automation."""
+
+    def __init__(
+        self,
+        webhook: CoreWebhookNode,
```
ajtmccarty, line 99: `webhook` should be passed into `configure` b/c `webhook` is the entity that the class operates on. It should be possible to create one instance of `SingleWebhookConfigurer` and then call `configure` on multiple webhook instances

ajtmccarty (commented):
looks much better. just a couple small changes then I think it is ready

backend/infrahub/webhook/models.py
polmichel: I think this class contains a combination of two objects/responsibilities. 
- A `WebhookAutomation`, which is exactly what it says: a value object with a name, ID and a method called to_prefect_automation(deployment_id), which allows transition to the Prefect layer.
- The second is a `WebhookAutomationPrefectSyncer`, which manages all the creation, updating and deletion processes.

I was wondering about including 'Prefect' in the name because this is an implementation detail; when the developer reads the code, they don't need to know that it's related to Prefect. From their perspective, the automation is synced at the end of the process. Tbh, in that case, we would need an interface so that there is no specificity regarding the 'prefect_client' object either, but that would be for another time.

backend/infrahub/webhook/models.py
polmichel: This has been done as separating the responsibilities between `infrahub.webhook.models.WebhookAutomationPrefectSyncer` which takes as parameter in its `apply` method a `WebhookAutomation` object.

backend/infrahub/webhook/tasks/configure.py
polmichel: You are right. I was thinking in terms of composition using other objects, but going into the '-er' objects path implies changing this behavior.

backend/infrahub/webhook/models.py
```diff
@@ -93,7 +94,91 @@ def from_object(cls, obj: CoreWebhook | CoreWebhookNode) -> Self:
...
+
+        if existing:
+            await self._client.update_automation(automation_id=existing.id, automation=prefect_automation)
+            logger.info("Automation %s updated", automation.name)
+        else:
+            await self._client.create_automation(automation=prefect_automation)
+            logger.info("Automation %s created", automation.name)
+
+    async def _find_existing(self, name: str) -> Automation | None:
+        from infrahub.trigger.setup import gather_all_automations
```
ajtmccarty: another import escaped its home

ajtmccarty (approved):
looks great. small, understandable component injected where needed and fast unit tests

PR #8874, opened 2026-04-10
Fix race condition: ensure generators complete before artifact refresh in proposed changes

minitriga:
## Why

When a proposed change pipeline runs with `CheckType.ALL`, generators and artifacts are both dispatched as fire-and-forget workflows via `submit_workflow()`. Artifact generation can begin before generators have finished creating objects, producing stale or incomplete artifacts.

The race only manifests in production (Prefect worker adapter uses `run_deployment(timeout=0)`). The local/test adapter serializes all calls, masking the issue.

## What changed

- **`run_generators()`** is now single-purpose: runs generator definition checks via `execute_workflow` + `asyncio.gather`, blocks until all complete, returns. No longer dispatches artifacts or repository checks.
- **`run_proposed_change_pipeline()`** owns the sequencing: dispatches independent checks first (fire-and-forget), blocks on generators via `execute_workflow`, then dispatches artifact refresh and repository checks.
- **`RequestProposedChangeRunGenerators`** simplified: removed `refresh_artifacts` and `do_repository_checks` fields.
- Independent checks (data integrity, schema, user tests) still run concurrently with generators — no throughput regression.
- No schema changes, no API changes. Generator/artifact logic itself untouched.

### Review order
1. `backend/infrahub/proposed_change/models.py` — field removal
2. `backend/infrahub/proposed_change/tasks.py` — `run_generators()` then `run_proposed_change_pipeline()`
3. `backend/tests/unit/proposed_change/test_run_generators.py` — new ordering tests
4. `backend/tests/adapters/workflow.py` and integration test updates

## How to test

```bash
uv run pytest backend/tests/unit/proposed_change/test_run_generators.py -v --no-cov
```

For production: create a proposed change triggering both generators and artifacts. In Prefect UI, confirm generator flows complete before artifact refresh starts.

## Impact

- **Backward compatibility:** No breaking changes
- **Performance:** Pipeline now blocks for generator duration (intentional). Monitor Prefect work pool — `execute_workflow` holds a slot while waiting.
- **Deployment:** Safe to deploy, no coordination needed.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added
- [ ] Docs updated

### DIFF
--- a/backend/infrahub/proposed_change/models.py
+++ b/backend/infrahub/proposed_change/models.py
@@ -25,12 +25,7 @@ class RequestProposedChangeDataIntegrity(BaseProposedChangeWithDiffMessage):
 
 
 class RequestProposedChangeRunGenerators(BaseProposedChangeWithDiffMessage):
-    """Sent trigger the generators that are impacted by the proposed change to run."""
-
-    refresh_artifacts: bool = Field(..., description="Whether to regenerate artifacts after the generators are run")
-    do_repository_checks: bool = Field(
-        ..., description="Whether to run repository and user checks after the generators are run"
-    )
+    """Sent to trigger the generators that are impacted by the proposed change to run."""
 
 
 class RequestProposedChangeRepositoryChecks(BaseProposedChangeWithDiffMessage):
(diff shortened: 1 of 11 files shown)


minitriga:
We need to ensure checks also run after generator and that a diff is recalculated after generator has run

PR #8876, opened 2026-04-11
schema remove migration left orphaned edges

ajtmccarty:
the `NodeRemoveMigration` that we run when deleting a schema, did not fully delete objects at the database level. the migration only deleted the edges touching the Nodes to delete (`HAS_ATTRIBUTE`, `IS_RELATED`, `IS_PART_OF` edges), but left the lower-level edges (`HAS_VALUE`, `IS_PROTECTED`, far-side `IS_RELATED`) un-deleted. this did not cause an issue for the end user b/c this partial delete was enough to prevent the objects from being returned by the regular database get queries we use for querying for and retrieving objects.

however, it could cause a deleted object to not be included in the diff. and the only reason that the object would be deleted on the default branch after a merge is because the `NodeRemoveMigration` would run again as part of the branch merge workflow.

only found the issue while working on improvements to the merge logic and troubleshooting a failing test led me to this bug.

the fix is a rewrite of the `NodeRemoveMigration` to make sure that it gets all of the edges "under" a given object when deleting it, which requires traversing one level deeper into the graph so the query is bigger. I also chose to rewrite the query as more of a static string instead of the Python-based construction it was using before. I find one large query easier to read than many smaller pieces constructed in memory

also adds some more cypher queries that validate the database to find illegal data

### DIFF
--- a/backend/infrahub/core/migrations/schema/node_remove.py
+++ b/backend/infrahub/core/migrations/schema/node_remove.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Sequence
 
-from infrahub.core.constants import RelationshipStatus
+from infrahub.core.constants import GLOBAL_BRANCH_NAME, RelationshipStatus
 
 from ..query import MigrationQuery
 from ..shared import SchemaMigration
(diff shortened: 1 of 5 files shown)


backend/infrahub/core/migrations/schema/node_remove.py
```diff
@@ -97,106 +53,271 @@ def get_nbr_migrations_executed(self) -> int:
...
+    WITH DISTINCT active_node, type(r) AS edge_type, peer
+    CALL (active_node, edge_type, peer) {
+        MATCH (active_node)<-[r:$(edge_type)]-(peer)
+        WHERE %(branch_filter)s
+        RETURN r AS rel_inbound, peer AS peer_node
+        ORDER BY r.branch_level DESC, r.from DESC
+        LIMIT 1
+    }
+    WITH active_node, rel_inbound, peer_node
+    WHERE rel_inbound.status = "active"
```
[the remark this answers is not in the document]
ajtmccarty: the `WHERE %(branch_filter)s` part of the subquery covers this

backend/infrahub/core/migrations/schema/node_remove.py
```diff
@@ -97,106 +53,271 @@ def get_nbr_migrations_executed(self) -> int:
...
+    MATCH (active_node)-[:HAS_ATTRIBUTE|IS_RELATED|IS_PART_OF]->(peer:Attribute|Relationship|Root)
+    CALL (active_node, peer) {
+        MATCH (active_node)-[r]->(peer)
+        WHERE %(branch_filter)s
+        RETURN r AS rel_outbound, peer AS peer_node
+        ORDER BY r.branch_level DESC, r.from DESC
+        LIMIT 1
+    }
+    WITH active_node, rel_outbound, peer_node
+    WHERE rel_outbound.status = "active"
```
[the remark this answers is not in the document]
ajtmccarty: the `WHERE %(branch_filter)s` part of the subquery covers this

backend/infrahub/core/migrations/schema/node_remove.py
```diff
@@ -97,106 +53,271 @@ def get_nbr_migrations_executed(self) -> int:
...
+        SET rel_outbound.to = $current_time, rel_outbound.to_user_id = $user_id
+    }
+
+    // ----------------------------------------------------------
+    // Close sub-edges hanging off the Attribute/Relationship peer vertex
+    // ----------------------------------------------------------
+    WITH DISTINCT active_node, peer_node
+    MATCH (peer_node:Attribute|Relationship)-[]-(sub_peer)
+    WHERE sub_peer <> active_node
+    WITH DISTINCT active_node, peer_node, sub_peer
```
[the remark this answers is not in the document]
ajtmccarty: fixed in latest commit

backend/tests/component/core/diff/test_diff_after_node_remove_migration.py
```diff
@@ -0,0 +1,120 @@
+"""Test that the diff includes nodes deleted by a NodeRemoveMigration.
```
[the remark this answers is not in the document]
ajtmccarty: this test will prevent a regression in this scenario

backend/infrahub/core/migrations/schema/node_remove.py
```diff
@@ -97,106 +53,271 @@ def get_nbr_migrations_executed(self) -> int:
...
-        return sub_query_out, sub_query_out_args
+        WHERE $set_metadata
+        SET active_node.updated_at = $current_time, active_node.updated_by = $user_id
+    }
+
+    // ----------------------------------------------------------
+    // For each outbound edge from active_node, find the latest active edge on the branch
+    // ----------------------------------------------------------
+    WITH active_node
+    MATCH (active_node)-[:HAS_ATTRIBUTE|IS_RELATED|IS_PART_OF]->(peer:Attribute|Relationship|Root)
```
gmazoyer: Above for the inbound part, you used `WITH DISTINCT active_node, type(r) AS edge_type, peer` to carry the edge type, but not here. It's a bit asymmetric, but I don't know if it's required. That's just something that caught my attention.

backend/tests/helpers/db_validation.py
```diff
@@ -133,6 +134,152 @@ async def verify_no_duplicate_paths(db: InfrahubDatabase) -> None:
...
+WHERE r2.status = "active" AND r2.to IS NULL
+RETURN DISTINCT
+    field.name AS field_name,
+    r1.branch AS branch,
+    labels(field)[0] AS field_type,
+    type(r2) AS child_type
+    """
+    params = {
+        "default_branch": registry.default_branch,
+        "global_branch": GLOBAL_BRANCH_NAME,
```
gmazoyer: Seems like it is not used in the query.

backend/tests/component/core/diff/test_diff_after_node_remove_migration.py
```diff
@@ -0,0 +1,120 @@
...
+from infrahub.core.timestamp import Timestamp
+from infrahub.database import InfrahubDatabase
+from infrahub.dependencies.registry import get_component_registry
+from tests.helpers.db_validation import verify_graph
+
+
+class TestDiffAfterNodeRemoveMigration:
+    @pytest.fixture(autouse=True)
+    async def _setup_core_schema(self, register_core_models_schema: SchemaBranch) -> None:
+        return
```
gmazoyer: An empty body with just `pass` or a docstring is more idiomatic, but this is really a trival comment.

gmazoyer (approved):

backend/infrahub/core/migrations/schema/node_remove.py
ajtmccarty: It's not necessary here b/c we know that a `HAS_ATTRIBUTE` edge will only point to a single `:Attribute`, and `IS_RELATED` edge will only point to a single `:Relationship`, and the `IS_PART_OF` edge will point to the one and only `:Root`. it's necessary up above b/c a single `:Relationship` could be linked to a single `:Node` by `IS_RELATED`, `HAS_SOURCE`, and `HAS_OWNER` edges, so we have to make sure to track all three possibilities separately.
that said, I'll add it just to be safe and consistent. thanks for the thorough review of just a whole bunch of cypher

backend/tests/component/core/diff/test_diff_after_node_remove_migration.py
ajtmccarty: just going to delete this fixture and add it to the one test in the class, seems more straightforward.

backend/tests/helpers/db_validation.py
ajtmccarty: you are correct. removed

PR #8878, opened 2026-04-13
Make Generic attribute 'optional' field changeable

### LINKED ISSUE #8877: bug: Migration of the optional property on a schema attribute fails silently on generic nodes
# Current Behavior

Updating the `optional` property on an attribute defined in a generic schema node is silently ignored. The schema load appears to succeed without error, but the UI and API continue to display the field with its original optionality on all inheriting child nodes. There is no way to relax a mandatory field on a generic without rebuilding the schema from scratch.

# Expected Behavior

Changing the `optional` field on an attribute in a generic schema node should propagate to all inheriting child node types, in both directions:

- **Mandatory-to-optional**: should succeed and propagate to all inheriting nodes.
- **Optional-to-mandatory**: should be supported, with validation that all existing instances have non-null values.
- The system should reject the change if the attribute is still referenced in `human_friendly_id` or `uniqueness_constraints`.
- Child nodes should be able to independently override the generic's `optional` setting via `allow_override`.

# Steps to Reproduce

1. Define a generic schema node with a mandatory attribute (`optional: false`) that is part of the `human_friendly_id` (hfid) and uniqueness constraints.
2. Create one or more concrete node types that inherit from the generic.
3. Realize the attribute should be optional. Remove the attribute from the uniqueness constraint and hfid specification on the generic.
4. Update the attribute on the generic to `optional: true` and load the schema.
5. Observe that the schema load succeeds without error.
6. Inspect the inheriting child nodes via the UI and API — the field is still displayed and enforced as mandatory.

# Additional Information

1. The root cause is in `AttributeOptionalChecker`, which validates against the generic schema directly (abstract, no instances) rather than expanding to all concrete child node types via `used_by`.
2. The schema persistence pipeline already handles propagation correctly through `process_inheritance()` and the diff mechanism.
3. A constraint guard can be added to prevent making an attribute optional when it is still referenced in `human_friendly_id` or `uniqueness_constraints`.
4. This is related to the original restriction added in [bug: Overriding `Optional` between Generic and Node crash server #4936](https://github.com/opsmill/infrahub/issues/4936).

PhillSimonds:
# Why

Changing the optional property on attributes defined in generic schema nodes is silently ignored. The schema update appears to succeed, but the UI and API continue to show the original optionality on all inheriting child nodes. This was intentionally restricted in #4936 to prevent breaking GraphQL schema generation, but it blocks legitimate workflows where a customer needs to relax a mandatory field on a generic.                  

This PR enables changing optional in both directions (mandatory-to-optional and optional-to-mandatory) on generic schema attributes, with correct propagation to all inheriting child node types and proper constraint validation.
                                                                                                                                
  Non-goals: This PR does not change how optional works on relationship schemas or on non-generic node schemas, those paths are unchanged.

Closes #8877

## What changed

  - Generic attribute optional changes now propagate to child nodes: AttributeOptionalChecker.check() detects when the target schema is a GenericSchema and iterates all concrete child node types via used_by to validate instances, rather than querying the abstract generic itself (which has no instances).                                                                         
  - hfid/uniqueness constraint guard: A new _attr_in_hfid_or_uniqueness() check rejects making an attribute optional if it is
  still referenced in human_friendly_id or uniqueness_constraints on the generic or any inheriting node. Combined operationsn(remove constraints + change optional in a single schema load) work correctly because the validator checks the new schema state.                                                                                                                        
  - Documentation updated: Removed the restriction line "Properties that can't be updated on an Attribute inherited from a Generic: optional" from docs/docs/topics/schema.mdx.                                                                          
  - What stayed the same: No changes to the schema persistence pipeline (manager.py), the migration handler (AttributeSupportsGeneratedSchemaMigration), the ConstraintValidatorDeterminer, or the inheritance processing (update_from_generic). The existing diff mechanism already propagated child node changes correctly — only the validator was scoped too narrowly.  
 
### Suggested review order
1. Start with ...
2. Then ...
3. Tests are ...
-->

## How to review

  Start with backend/infrahub/core/validators/attribute/optional.py — this is the only production code change. The check() method has two new blocks: generic expansion (iterate used_by) and hfid/uniqueness guard. The rest is tests and docs.
                                                                                                                                
  The test file adds 3 new test classes with 5 tests covering: basic propagation, reverse direction constraint generation, hfid/uniqueness blocking, combined constraint removal, and child override behavior.

### Manual validation

  1. Start Infrahub locally: uv run invoke dev.start
  2. Load an initial schema with a generic that has a mandatory attribute:

```yaml
  version: "1.0"
  generics:
    - name: Device
      namespace: Testing
      attributes:
        - name: name
          kind: Text
        - name: asset_tag
          kind: Text
          optional: false
  nodes:
    - name: Router
      namespace: Testing
      inherit_from:
        - TestingDevice
    - name: Switch
      namespace: Testing
      inherit_from:
        - TestingDevice
```

  3. Create a few TestingRouter and TestingSwitch instances, providing values for asset_tag.
  4. Load an updated schema that changes asset_tag to optional on the generic:

```yaml
  version: "1.0"
  generics:
    - name: Device
      namespace: Testing
      attributes:
        - name: name
          kind: Text
        - name: asset_tag
          kind: Text
          optional: true
  nodes:
    - name: Router
      namespace: Testing
      inherit_from:
        - TestingDevice
    - name: Switch
      namespace: Testing
      inherit_from:
        - TestingDevice
```

  5. Verify:
    - In the UI, navigate to TestingRouter and TestingSwitch — asset_tag should no longer be marked as required
    - Create a new TestingRouter without providing asset_tag — it should succeed
    - Existing instances should retain their asset_tag values

  6. Load an updated schema that changes asset_tag back to mandatory:

```yaml
  version: "1.0"
  generics:
    - name: Device
      namespace: Testing
      attributes:
        - name: name
          kind: Text
        - name: asset_tag
          kind: Text
          optional: false
  nodes:
    - name: Router
      namespace: Testing
      inherit_from:
        - TestingDevice
    - name: Switch
      namespace: Testing
      inherit_from:
        - TestingDevice
```
  7. Verify:
    - If you created any instances in step 5 without asset_tag, the schema load should be rejected with a validation error identifying which nodes are missing values
    - If all instances have asset_tag populated (delete or update the ones missing it), the schema load should succeed and the field should appear as required again in the UI for both TestingRouter and TestingSwitch

  8. To test the constraint guard, try making asset_tag optional while it is still in an hfid:

  version: "1.0"
  generics:
    - name: Device
      namespace: Testing
      human_friendly_id: ["name__value", "asset_tag__value"]
      attributes:
        - name: name
          kind: Text
        - name: asset_tag
          kind: Text
          optional: true

  This should be rejected with a validation error indicating the attribute is referenced in the hfid.

### Automated validation

  uv run pytest backend/tests/integration/schema_lifecycle/test_generic_migrations.py::TestSchemaLifecycleGenericOptionalChange
  backend/tests/integration/schema_lifecycle/test_generic_migrations.py::TestSchemaLifecycleGenericOptionalWithConstraints backend/tests/integration/schema_lifecycle/test_generic_migrations.py::TestSchemaLifecycleGenericOptionalOverride -xvs

## Impact & rollout

  - Backward compatibility: No breaking changes. Previously silently ignored schema updates will now take effect — this is the intended behavioral change. Customers who were working around this limitation can now use the direct approach.
  - Performance: No impact. The validator iterates used_by (typically a small list) and runs the existing query per child node type.                                                                                                                         
  - Config/env changes: None.
  - Deployment notes: Safe to deploy. No migration required.  

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Enables changing a generic attribute’s `optional` flag in both directions and correctly propagates the change to all child node types. Adds a strict‑mode schema check for optional attributes used in `human_friendly_id`/uniqueness and a graph migration that reports existing violations.

- **New Features**
  - Validates optional→mandatory on generics via a single Cypher over the generic label; excludes child kinds that override to optional and reports the concrete instance kind.
  - Blocks making an attribute optional if it’s used in `human_friendly_id`/uniqueness unless it has a `default_value`; enforced only in strict mode.

- **Migration**
  - Adds `m074_check_optional_hfid_uniqueness` to report branches/node kinds where optional attributes (without `default_value`) are used in `human_friendly_id`/uniqueness; runs regardless of strict mode and includes bypass guidance (`INFRAHUB_SCHEMA_STRICT_MODE=false`).
  - Bumps graph version to 74.

<sup>Written for commit 82a041e448e2ca3b851a6ec542240c9df6c26d0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/8878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/validators/attribute/optional.py
```diff
@@ -75,12 +77,89 @@ def name(self) -> str:
...
+                                path_type=PathType.ATTRIBUTE,
+                                node_id="",
+                                field_name=request.schema_path.field_name,
+                                kind=node_schema.kind,
+                                value=child_conflict,
+                            ),
+                        )
+                        grouped_data_paths_list.append(gdp)
+
+            return grouped_data_paths_list
```
ajtmccarty: this logic should move to somewhere in `SchemaBranch.process`, which is where we handle validating a new or updated schema as a whole before running constraint validators. constraint validators are a later step in schema evaluation when we make sure that the existing data conforms to the updated schema. this part of the new logic for checking optional attributes against uniqueness constraints and HFIDs doesn't require access to data to determine if the schema is valid or not, so it can go in the earlier, faster validation we run in `SchemaBranch.process`

backend/infrahub/core/validators/attribute/optional.py
```diff
@@ -75,12 +77,89 @@ def name(self) -> str:
...
+                node_attr = node_schema.get_attribute(name=request.schema_path.field_name)
+                # Skip child nodes that locally override this attribute to optional
+                if node_attr.optional is True:
+                    continue
+                schema_path = request.schema_path.model_copy(update={"schema_kind": node_schema.kind})
+                for query_class in self.query_classes:
+                    query = await query_class.init(
+                        db=self.db, branch=self.branch, node_schema=node_schema, schema_path=schema_path
+                    )
+                    await query.execute(db=self.db)
```
ajtmccarty: instead of running the same query multiple times, I think a better approach is to update the database query that this constraint validator uses (`AttributeOptionalUpdateValidatorQuery`) to accept an optional list of specific node kinds to ignore. this requires some cypher changes, but is a cleaner approach than running database queries in a loop.

this constraint validator could actually stand to be cleaned up and make the use of `AttributeOptionalUpdateValidatorQuery` explicit instead of running it in a loop of `for query_class in self.query_classes`, but that might be out of scope for this work

ajtmccarty (commented):
another thing to consider is whether there can be existing Infrahub instances out there that will violate the new optional attribute HFID/uniqueness constraints validation and how to handle them

backend/infrahub/core/validators/attribute/optional.py
PhillSimonds: Awesome. Just made updates to this as well. Making the AttributeOptionalUpdateValidatorQuery explicit instead of running it in a for loop was easy, so I made those changes just to do a little hosue-keeping as we go along. Happy to separate that out if you'd prefer it in a separate change.

PhillSimonds:
> another thing to consider is whether there can be existing Infrahub instances out there that will violate the new optional attribute HFID/uniqueness constraints validation and how to handle them

Good thought here. I've `config.SETTINGS.main.schema_strict_mode` so that when it's configured to false the validate_optional_against_hfid_and_uniqueness method skips. I've also added a migration diagnoses any violations. It doesn't auto fix because changing optional true -> false could break data, it just reports.

Happy to approach this a different way if you have other thoughts!

backend/tests/component/core/migrations/graph/test_070.py
```diff
@@ -0,0 +1,68 @@
...
+                    "namespace": "Testing",
+                    "inherit_from": ["TestingGadget"],
+                }
+            ],
+        }
+        response = await client.schema.load(schemas=[violating_schema])
+        assert len(response.errors) == 0, response.errors
+
+    async def test_migration_clean_and_with_violations(self, db: InfrahubDatabase, client: InfrahubClient) -> None:
+        strict_mode_original = config.SETTINGS.main.schema_strict_mode
```
ajtmccarty, line 43: the preferred way to accomplish a temporary override of a setting is using a fixture with a `yield`
example
```
@pytest.fixture
def query_limit_of_one() -> Generator[None, None, None]:
    original_query_size_limit = config.SETTINGS.database.query_size_limit
    config.SETTINGS.database.query_size_limit = 1
    yield
    config.SETTINGS.database.query_size_limit = original_query_size_limit
```

backend/tests/component/core/migrations/graph/test_070.py
```diff
@@ -0,0 +1,68 @@
...
+
+        try:
+            # Baseline: the default schema has no violations, migration should report nothing.
+            migration = Migration070()
+            baseline_result = await migration.execute(migration_input=MigrationInput(db=db))
+            assert not baseline_result.errors, f"Expected no errors on baseline schema, got: {baseline_result.errors}"
+
+            # Simulate a pre-existing deployment by loading a violating schema with strict mode off.
+            config.SETTINGS.main.schema_strict_mode = False
```
ajtmccarty, line 52: should this be explicitly set to `True` before the first migration runs?

backend/tests/conftest.py
```diff
@@ -539,6 +539,7 @@ def reload_settings_before_each_module(tmpdir_factory: pytest.TempdirFactory) ->
     # Other settings
     config.SETTINGS.storage.driver = config.StorageDriver.FileSystemStorage
     config.SETTINGS.workflow.driver = config.WorkflowDriver.LOCAL
+    config.SETTINGS.main.schema_strict_mode = False
```
ajtmccarty, line 542: use the fixture I described above instead of changing the settings for all tests

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -1792,6 +1805,383 @@ async def _validate_inherited_schema_fields(
...
+        """Changing optional from false to true on a generic attribute should propagate to all child nodes."""
+        updated_generic = deepcopy(schema_generic_with_mandatory)
+        updated_generic["attributes"][1]["optional"] = True
+
+        updated_schema = {
+            "version": "1.0",
+            "generics": [updated_generic],
+            "nodes": [schema_child_one_base, schema_child_two_base],
+        }
+        await load_schema(db=db, schema=updated_schema)
```
ajtmccarty, line 1889: in integration/functional tests, we try to load updated schema using `client.schema.load` (unless it is impossible for some reason) to make sure the actual loading code path is exercised. and then if there are changes to the schema loading code in the API endpoint, this test can catch a regression

there's a `client: InfrahubClient` fixture you can use for this

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -1792,6 +1805,383 @@ async def _validate_inherited_schema_fields(
...
+                "version": "1.0",
+                "generics": [reverted_generic],
+                "nodes": [schema_child_one_base, schema_child_two_base],
+            }
+        )
+
+        candidate = schema_branch.duplicate()
+        candidate.load_schema(schema=reverted_schema_root)
+        candidate.process()
+        diff = candidate.diff(schema_branch)
```
ajtmccarty, line 1942: I'd say that this should use `client.schema.load` and then assert there are errors and that the errors contain the expected object kind/ID/attribute

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
```diff
@@ -1792,6 +1805,383 @@ async def _validate_inherited_schema_fields(
         return errors
 
 
+class TestSchemaLifecycleGenericOptionalChange(TestSchemaLifecycleBase):
```
ajtmccarty, line 1808: these 3 tests should be combined into 1 more complex test and moved to their own file. each instance of `TestSchemaLifecyleBase` invokes overhead to prepare the test class, so we can save testing time (and get  more thorough test cases) if we put them together into one class that goes through all of these cases and verifies each case succeeds or fails as expected along the way.

the huge test case above is an example. it uses a particular schema for each "step" and they are executed in a particular order

ajtmccarty (commented):
this is looking good. I have some comments on the tests, but the functional code looks good

backend/tests/component/core/migrations/graph/test_070.py
dgarros: would be good to catch that in our dev documentation

backend/infrahub/core/validators/attribute/optional.py
PhillSimonds: Done — the hfid/uniqueness check now lives in `SchemaBranch.validate_optional_against_hfid_and_uniqueness()`, called from `process_validate()` (the schema-only validation phase) rather than the constraint validator, so it runs in the earlier/faster validation step as suggested.

backend/infrahub/core/validators/attribute/optional.py
PhillSimonds: Followed through on the cleaner approach as well: `AttributeOptionalChecker.check()` no longer loops over `query_classes` — it calls `AttributeOptionalUpdateValidatorQuery` explicitly, and the query now takes an `excluded_kinds` list so a single batched Cypher over the generic's kind label validates all inheriting instances at once, while skipping child nodes that override the attribute to optional.

backend/tests/component/core/migrations/graph/test_070.py
PhillSimonds: Done — replaced the try/finally with a `disable_schema_strict_mode` yield fixture (now in `backend/tests/conftest.py`, plus an `enable_schema_strict_mode_class` counterpart). The migration test (`test_074.py`) uses the yield fixture.

PhillSimonds: Documented the yield-fixture pattern for temporary `config.SETTINGS` overrides in `docs/docs/development/backend.mdx` (new "Temporarily overriding a setting in a test" subsection).

backend/tests/component/core/migrations/graph/test_070.py
PhillSimonds: The migration reports violations regardless of strict mode by design, so the baseline run no longer toggles the setting — it runs under the suite's configured value, and a comment in the test now makes that explicit.

backend/tests/conftest.py
PhillSimonds: I tried scoping this to per-test fixtures, but it turned out to be impractical: the new validation runs over the entire schema branch on every load, and several shared test schemas legitimately rely on the shapes it now rejects (e.g. the `TICKET` helper has an optional+unique `ticket_id` in its `human_friendly_id`, and the lock/uniqueness tests deliberately place optional attrs in uniqueness_constraints). Per-test scoping also runs into class-scoped registry accumulation and cross-file schema persistence within the shared pytest session, which produced cascading failures across unrelated tests.

So I kept the global disable in the test conftest, but added documented yield fixtures — `disable_schema_strict_mode` / `enable_schema_strict_mode_class` — so tests that specifically assert strict-mode behavior opt in explicitly (the new `test_generic_optional_lifecycle.py` uses `enable_schema_strict_mode_class`). Happy to revisit if you'd prefer a different split.

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
PhillSimonds: Done — the lifecycle tests now drive `client.schema.load` instead of the low-level helpers, so the real schema-loading API path is exercised. See `test_generic_optional_lifecycle.py`.

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
PhillSimonds: Done — the optional→mandatory-blocked step (`test_step03_optional_to_mandatory_blocked_with_nulls`) uses `client.schema.load`, asserts `response.errors`, and checks the error message contains the concrete child kind (`TestingRouter`), the failing instance's ID, and the `optional` constraint.

backend/tests/integration/schema_lifecycle/test_generic_migrations.py
PhillSimonds: Done — the three classes are now a single stepwise `TestSchemaLifecycleGenericOptional` in `backend/tests/integration/schema_lifecycle/test_generic_optional_lifecycle.py`. It prepares one schema/dataset and runs 7 ordered steps (mandatory→optional propagation, optional→mandatory blocked-then-allowed-once-populated, child override, hfid guard rejection, and default-value allowance). The old three classes were removed from `test_generic_migrations.py`.

PhillSimonds:
Picking this back up. Rebased onto current `develop` and collapsed the work into a **single commit** so it's easy to see exactly what changed on top of `develop`, and addressed the review feedback.

### Rebase / housekeeping
- Squashed the original commits into one and rebased onto current `develop` (was ~800 commits behind).
- `develop` has since reached `GRAPH_VERSION = 73` and already uses `m070`–`m073`, so the migration was renumbered **`m070` → `m074`** (`Migration074`, `minimum_version = 73`) and `GRAPH_VERSION` bumped to **74**.
- The schema docs were reorganized on `develop`, so the removed restriction line now comes out of `docs/docs/schema/migration.mdx` (the old `topics/schema.mdx` was archived).

### Review feedback (replied inline on each thread)
- **Validation moved to `SchemaBranch.process`** — the hfid/uniqueness check is now `validate_optional_against_hfid_and_uniqueness()` run in `process_validate()`, not the constraint validator.
- **Validator query made explicit** — no more `query_classes` loop; `AttributeOptionalUpdateValidatorQuery` takes an `excluded_kinds` list so one batched Cypher validates all inheriting child instances.
- **Tests** — the three generic-optional classes are consolidated into one stepwise `TestSchemaLifecycleGenericOptional` (`test_generic_optional_lifecycle.py`) driven through `client.schema.load`, asserting error kind/ID/constraint on the blocked case; `test_074` uses a `yield` fixture; the fixture pattern is documented in `docs/docs/development/backend.mdx`.
- **Strict mode in tests** — see the discussion on the conftest thread: scoping per-test proved impractical (the validation spans the whole schema branch and shared test schemas legitimately use optional attrs in hfid/uniqueness), so strict mode stays globally disabled in the test conftest with documented `disable_/enable_schema_strict_mode` fixtures for tests that assert strict-mode behavior. Open to a different split if preferred.

### Local verification
- `test_074` (migration) and the 7-step `test_generic_optional_lifecycle` both pass (mandatory→optional propagation, optional→mandatory blocked-then-allowed, child override, hfid guard, default-value allowance), plus the `m074` discovery/version check.

backend/infrahub/core/validators/attribute/optional.py
```diff
@@ -18,15 +19,21 @@
...
         branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
         self.params.update(branch_params)
 
         self.params["attr_name"] = self.attribute_schema.name
         self.params["null_value"] = NULL_VALUE
+        self.params["excluded_kinds"] = self.excluded_kinds
 
         query = """
         MATCH (n:%(node_kind)s)
+        WHERE NONE(kind IN $excluded_kinds WHERE kind IN labels(n))
```
ajtmccarty, line 36: I think it might be a little more clear to do `n.kind NOT IN $excluded_kinds` b/c n.kind is the concrete schema type of the object and `labels(n)` will include all generics that `n` inherits from

backend/tests/component/core/migrations/graph/test_074.py
```diff
@@ -0,0 +1,69 @@
...
+        execution_result = await migration.execute(migration_input=MigrationInput(db=db))
+
+        assert execution_result.errors, "Migration should return errors when violations exist"
+        header = execution_result.errors[0]
+        assert "INFRAHUB_SCHEMA_STRICT_MODE=false" in header, (
+            "Migration header should explain how to bypass the validator"
+        )
+
+        violation_reports = execution_result.errors[1:]
+        matching = [msg for msg in violation_reports if "'serial'" in msg and "TestingGadget" in msg]
```
ajtmccarty, line 68: better to assert the exact expected errors instead of just that one expected error exists

backend/tests/integration/schema_lifecycle/test_generic_optional_lifecycle.py
```diff
@@ -0,0 +1,266 @@
...
+    ) -> None:
+        """Relaxing the generic attribute to optional propagates to every inheriting child node."""
+        updated = self._schema_with_optional(schema_generic_device, schema_router, schema_switch, optional=True)
+        response = await client.schema.load(schemas=[updated])
+        assert not response.errors
+
+        schema_branch = registry.schema.get_schema_branch(name=registry.default_branch)
+        for kind in (GENERIC_KIND, ROUTER_KIND, SWITCH_KIND):
+            assert schema_branch.get(name=kind, duplicate=False).get_attribute("asset_tag").optional is True, kind
+
```
ajtmccarty, line 131: one extra step that it would be good to include in these tests for the cases where the migration succeeds is to retrieve the updated schemas from the database and verify that the changed attribute has optional set as expected. `registry.schema.get_schema_branch` will just return the `SchemaBranch` from memory for the given branch, which doesn't guarantee the update is persisted.

loading a whole SchemaBranch using `registry.schema.load_schema_from_db()` is quite slow, but filtering it using the `schema_diff` kwarg to just get the selected schemas will be faster (not fast, but faster)

could even be a little helper method that validates all three schemas all have the expected values for the given attribute

backend/tests/integration/schema_lifecycle/test_generic_optional_lifecycle.py
```diff
@@ -0,0 +1,266 @@
...
+        client: InfrahubClient,
+        initial_dataset: dict[str, str],
+        schema_generic_device: dict[str, Any],
+        schema_router: dict[str, Any],
+        schema_switch: dict[str, Any],
+    ) -> None:
+        """Once every instance has a value, the optional->mandatory change is accepted."""
+        router2 = await registry.manager.get_one(db=db, id=initial_dataset["router2"])
+        assert router2 is not None
+        router2.asset_tag.value = "RTR-002"  # type: ignore[attr-defined]
```
ajtmccarty, line 170: use `get_attribute("asset_tag")` to avoid the typing ignore

backend/tests/integration/schema_lifecycle/test_generic_optional_lifecycle.py
```diff
@@ -0,0 +1,266 @@
...
+        updated = {
+            "version": "1.0",
+            "generics": [generic],
+            "nodes": [deepcopy(schema_router), switch_override],
+        }
+        response = await client.schema.load(schemas=[updated])
+        assert response.errors, "Expected the hfid/uniqueness guard to reject the optional attribute"
+        message = response.errors["errors"][0]["message"]
+        assert "human_friendly_id" in message, message
+        assert "asset_tag" in message, message
```
ajtmccarty, line 245: asserting the exact error message is our preferred approach for now. or a big chunk of it that covers the important parts

backend/tests/integration/schema_lifecycle/test_generic_optional_lifecycle.py
```diff
@@ -0,0 +1,266 @@
...
+        """An optional attribute in a human_friendly_id is allowed when it has a default_value."""
+        tagged = {
+            "name": "Tagged",
+            "namespace": "Testing",
+            "human_friendly_id": ["name__value", "status__value"],
+            "attributes": [
+                {"name": "name", "kind": "Text"},
+                # status is optional (default_value implies optional) but always populated, so it is
+                # safe to use in the human_friendly_id.
+                {"name": "status", "kind": "Text", "default_value": "active"},
```
ajtmccarty, line 257: is there a reason `"optional": True` can't be included in the definition here?

changelog/+generic-optional-update.added.md
```diff
@@ -0,0 +1 @@
+Allow changing attribute optionality (mandatory/optional) on generic schema nodes. Changes correctly propagate to all inheriting child node types. Includes validation that prevents making an attribute optional when it is still referenced in human-friendly identifier (hfid) or uniqueness constraints. The reverse direction (optional to mandatory) validates that all existing instances have non-null values across all child node types.
```
ajtmccarty, line 1:
```suggestion
Allow changing attribute optionality (mandatory/optional) on generic schemas. Changes correctly propagate to all inheriting schemas. Includes validation that prevents making an attribute optional when it is still referenced in human-friendly identifier (hfid) or uniqueness constraints. The reverse direction (optional to mandatory) validates that all existing instances have non-null values across all child node types.
```
I think "node" was being used in two different ways

backend/tests/conftest.py
ajtmccarty: I think this is the best option given how many changes would be required otherwise. I'm going to bring it up with the team and see what we want to do about existing tests that violate this new constraint going forward.

also, I wonder how common this is in user's schemas (or even our own example schemas) and if we will be forcing people into non-strict mode

ajtmccarty (commented):
I think the PR looks great. a few small comments, but the new integration test is very thorough. the main outstanding question for me is how many users is this going to force into non-strict schema mode?

PR #8887, opened 2026-04-13
IFC-2312: Fix branch delete permission preventing users from deleting own branch

### LINKED ISSUE #8528: bug: EDIT_DEFAULT_BRANCH permission denial prevents users from deleting their own branches
## Component

- [x] API Server / GraphQL

## Infrahub Version

develop (post-1.8.0)

## Current Behavior

When a user account does **not** have the `EDIT_DEFAULT_BRANCH` global permission (set to `DENY` or not granted), they are unable to delete **any** branch — including non-default branches they created themselves.

The `BranchDelete` GraphQL mutation fails with a permission denied error because the `DefaultBranchPermissionChecker` incorrectly blocks it.

### Root Cause

In `backend/infrahub/graphql/auth/query_permission_checker/default_branch_checker.py`, the `exempt_operations` list does not include `"BranchDelete"`:

```python
exempt_operations = [
    "BranchCreate",
    "DiffUpdate",
    "InfrahubAccountSelfUpdate",
    "InfrahubAccountTokenCreate",
    "InfrahubAccountTokenDelete",
]
```

The permission check at line 47 triggers when:
1. `operates_on_default_branch` is `True` — which happens when `analyzed_query.branch is None` (the typical case for branch management mutations that don't specify a branch context)
2. `analyzed_query.contains_mutation` is `True` — BranchDelete is a mutation
3. `is_exempt_operation` is `False` — BranchDelete is not in the exempt list

Since `BranchDelete` takes the target branch name as a mutation **argument** (`data.name`), not as the GraphQL request branch context, the analyzer sees `branch=None` and interprets the operation as targeting the default branch. This means the `EDIT_DEFAULT_BRANCH` permission is enforced even when deleting a non-default branch.

### Inconsistency with Other Branch Operations

- `BranchCreate` **is** correctly exempted from this check
- `BranchDelete` is correctly treated as a special operation in `backend/infrahub/graphql/middleware.py` (lines 3-4), where it's allowed on both `NEED_REBASE` and `MERGED` branches
- `BranchMerge` has its own dedicated permission checker (`MergeBranchPermissionChecker`) rather than relying on `EDIT_DEFAULT_BRANCH`

## Expected Behavior

Users should be able to delete non-default branches without needing the `EDIT_DEFAULT_BRANCH` permission. Branch deletion is a branch management operation, not a data editing operation on the default branch.

The fix should add `"BranchDelete"` to the `exempt_operations` list in `DefaultBranchPermissionChecker`:

```python
exempt_operations = [
    "BranchCreate",
    "BranchDelete",
    "DiffUpdate",
    "InfrahubAccountSelfUpdate",
    "InfrahubAccountTokenCreate",
    "InfrahubAccountTokenDelete",
]
```

Note: Depending on the desired security model, a dedicated `DELETE_BRANCH` permission could also be considered (similar to how `BranchMerge` has `MERGE_BRANCH`), but at minimum, `EDIT_DEFAULT_BRANCH` should not block branch deletion.

## Steps to Reproduce

1. Set up Infrahub with RBAC enabled
2. Create a user account with a role that has `EDIT_DEFAULT_BRANCH` set to `DENY`
3. As that user, create a new branch (this works because `BranchCreate` is exempted)
4. Attempt to delete the branch the user just created via GraphQL:
   ```graphql
   mutation {
     BranchDelete(data: { name: "my-test-branch" }) {
       ok
     }
   }
   ```
5. Observe permission denied error, even though the user is deleting their own non-default branch

### Relevant Files

| File | Lines | Description |
|------|-------|-------------|
| `backend/infrahub/graphql/auth/query_permission_checker/default_branch_checker.py` | 20-26 | Missing `BranchDelete` in exempt_operations |
| `backend/infrahub/graphql/mutations/branch.py` | 118-149 | BranchDelete mutation (no inline permission checks) |
| `backend/infrahub/graphql/middleware.py` | 3-4 | Correctly allows BranchDelete on restricted branch states |
| `backend/infrahub/graphql/auth/query_permission_checker/merge_operation_checker.py` | 13-35 | Example of proper branch-operation-specific permission checker |

solababs:
# Why

Users with `EDIT_DEFAULT_BRANCH` set to `DENY` (or not granted) cannot delete any branch — including non-default branches they created themselves. The `BranchDelete` mutation was missing from the `exempt_operations` list in `DefaultBranchPermissionChecker`, so the checker incorrectly treated branch deletions as default-branch data mutations and blocked them.

`BranchCreate` is already correctly exempted from this check. `BranchDelete` should be too — it is a branch management operation, not a data edit on the default branch.

Closes https://github.com/opsmill/infrahub/issues/8528

## How to test

```bash
# Run the targeted permission checker tests
uv run pytest backend/tests/component/graphql/auth/query_permission_checker/test_default_branch_checker.py -v

# Run the full unit suite to confirm no regressions
```

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Allows users to delete their own non-default branches even if `EDIT_DEFAULT_BRANCH` is denied, and blocks deleting others’ branches unless they have `DELETE_BRANCH`. Branch deletion is treated as branch management, not a default-branch edit.

- **Bug Fixes**
  - Exempted `BranchDelete` in `DefaultBranchPermissionChecker`; test asserts `NEXT_CHECKER` without default-branch permission.
  - Added `GlobalPermissions.DELETE_BRANCH` with description and denial message; enforced in `BranchDelete` so owners can delete without it and others need it; added allow/deny tests.

<sup>Written for commit 37c44bacdd7bd550348ed4e4d42f890c5ae3d236. Summary will update on new commits.</sup>


backend/infrahub/graphql/auth/query_permission_checker/default_branch_checker.py
```diff
@@ -19,6 +19,7 @@ class DefaultBranchPermissionChecker(GraphQLQueryPermissionCheckerInterface):
     )
     exempt_operations = [
         "BranchCreate",
+        "BranchDelete",
```
gmazoyer, line 22: This will allow someone to delete someone else's branch. I think this issue is another example falling into our discussion about the behaviour of the `DefaultBranchPermissionChecker` (especially with branch-agnostic objects).

solababs: Nice catch, see updated

backend/infrahub/graphql/mutations/branch.py
```diff
@@ -137,6 +139,14 @@ async def mutate(
         obj = await Branch.get_by_name(db=graphql_context.db, name=str(data.name))
         await apply_external_context(graphql_context=graphql_context, context_input=context)
 
+        if obj.created_by != graphql_context.active_account_session.account_id:
+            graphql_context.active_permissions.raise_for_permission(
+                permission=GlobalPermission(
+                    action=GlobalPermissions.DELETE_BRANCH.value,
+                    decision=PermissionDecision.ALLOW_ALL.value,
+                )
+            )
```
gmazoyer, line 148: This can work for now. I think will have to rewrite this at some point to avoid this kind of patch that we might need to make in a few places. Any thoughts about this @ogenstad?

backend/infrahub/graphql/mutations/branch.py
```diff
@@ -137,6 +139,14 @@ async def mutate(
         obj = await Branch.get_by_name(db=graphql_context.db, name=str(data.name))
         await apply_external_context(graphql_context=graphql_context, context_input=context)
 
+        if obj.created_by != graphql_context.active_account_session.account_id:
```
polmichel, line 142: I guess this is the only way of doing it based on the internal refactoring thoughts created by Guillaume. The permissions system has no ownership concept.
(Internal link: https://opsmill.atlassian.net/wiki/spaces/SD/pages/539820033/Permissions+thoughts#The-permission-system-has-no-ownership-concept)

backend/tests/component/graphql/auth/query_permission_checker/test_default_branch_checker.py
```diff
@@ -175,3 +175,42 @@ async def test_account_without_permission(
...
+        permissions_helper: PermissionsHelper,
+    ) -> None:
+        checker = DefaultBranchPermissionChecker()
+        session = AccountSession(
+            authenticated=True, account_id=permissions_helper.second.id, session_id=str(uuid4()), auth_type=AuthType.JWT
+        )
+        permission_manager = PermissionManager(account_session=session)
+        await permission_manager.load_permissions(db=db, branch=permissions_helper.default_branch)
+
+        graphql_query = MagicMock(spec=InfrahubGraphQLQueryAnalyzer)
```
polmichel: ### Optional comment

`MagicMock(spec=Branch)` has some caveats:
| Raise error on | `MagicMock(spec=)` | `MagicMock(spec_set=)` | `create_autospec` | `create_autospec(..., spec_set=True)` |                                                                    
  |---|---|---|---|---|           
  | **Get fake attr** | yes | yes | yes | yes |                                                                                                                            
  | **Set fake attr** | no | yes | no | yes |
  | **Wrong method signature** | no | no | yes | yes |                                                                                                                          
  | **Nested attrs auto-spec'd** | no | no | yes | yes |


So I think that using a `create_autospec(spec_set=True)` may be safer in the long run. But this is not that important and can be done later.

polmichel (approved):
LGTM. Just an optional comment about the mocks .

backend/tests/component/graphql/auth/query_permission_checker/test_default_branch_checker.py
solababs: `create_autospec(..., spec_set=True)` only allows attributes already known at class level. Since `branch` is assigned as an instance attribute in `__init__`, `spec_set` doesn't see it.
I will update to use `create_autospec` but leave it `spec_set=False` which still works

PR #8904, opened 2026-04-14
Local AI bug fixing pipeline test #10 normalize IPHost/IPNetwork/MAC address values at input time (closes #8896)

### LINKED ISSUE #8896: bug: IPHost attribute values vary in when they require/return their netmask from API and UI
### Component

_No response_

### Infrahub version

1.8.3

### Current Behavior

When an instance of a schema node with an IPHost attribute is created and the IPHost attribute is populated, the representation of the value is inconsistent across Infrahub's interfaces:

Web UI: The first column (display label) does not include the CIDR mask on the IPHost value, while the dedicated IPHost attribute column displays a /32 (IPv4) or /128 (IPv6) mask even when no mask was entered at creation time.
GraphQL: Queries that look up an instance by its hfid require the mask to be included in the hfid input, but the hfid returned by the GraphQL query response does not include the mask — making the returned hfid unusable as a direct input for subsequent lookups.
Python SDK: When retrieving the instance via client.get(...), the returned IPHost value includes the /32 or /128 mask, diverging from what the user entered and from what the UI display label shows.

### Expected Behavior

The representation of an IPHost attribute should be consistent across the Web UI, GraphQL API, and Python SDK:

The display label, the attribute column in the UI, the hfid returned by GraphQL, and the value returned by the Python SDK should all agree on whether the mask is included.
The hfid returned by a GraphQL query should be directly usable as input for a subsequent hfid-based lookup without the caller needing to re-append /32 or /128.
If no mask was entered at creation time, implicit default masks (/32 for IPv4, /128 for IPv6) should either be applied consistently everywhere or omitted consistently everywhere — not a mix of both.

### Steps to Reproduce

- Define a schema node with an IPHost attribute (and include that attribute in the node's human_friendly_id).
- Create an instance of that node, populating the IPHost attribute with a bare address (no mask), e.g. 192.0.2.10.
- Open the instance in the Web UI and observe.
  - The first column (display label) shows the address without a mask.
  - The IPHost attribute column shows the address with a /32 mask appended.
- Run a GraphQL query to retrieve the instance and note the returned hfid value (mask is absent).
- Attempt to look up the same instance via GraphQL using the returned hfid as input — the query fails unless the mask is re-appended to the hfid.
- Retrieve the same instance using the Python SDK via client.get(...) and observe that the IPHost value returned includes the /32 mask.

### Additional Information

The inconsistency appears to stem from different layers applying or stripping the default host mask (/32 for IPv4, /128 for IPv6) independently, rather than a single canonical representation being used when constructing display labels, hfid values, and serialized attribute values

The most user-impacting symptom is the GraphQL hfid round-trip: the value returned by a query cannot be fed back into a lookup without transformation, which breaks common automation patterns.

IPv6 equivalents should be verified to confirm the same /128 behavior occurs in parallel with the /32 IPv4 case.

polmichel:
## Why

IPHost attribute values are inconsistent across Infrahub's interfaces when no netmask is provided at creation time. The HFID returned by GraphQL cannot be used for subsequent lookups, display labels disagree with stored values, and the Python SDK returns a different representation than the UI.

This PR is about unifying this value display across the different layers of the application.

This was only impacting after attribute **creation**. Once the attribute is reloaded, bug should have disappeared.

### Goals
Ensure `IPHost`/`IPNetwork` values are normalized to their canonical `with_prefixlen` form immediately on input, so all downstream consumers (HFID, display labels, GraphQL, SDK) see a consistent representation.

### Out-of-scope

MAC address attributes had the same issue, and this PR also fixes it.

Closes #8896

## What changed

- **Behavioral change:** IPHost and IPNetwork `attr.value` is now always in canonical `with_prefixlen` form (e.g., `"192.0.2.1"` becomes `"192.0.2.1/32"`) from the moment the attribute is initialized. This makes HFIDs, display labels, GraphQL responses, and SDK values all consistent.

- **Implementation:** Added `_canonicalize_value()` hook to `BaseAttribute` (no-op by default), overridden in `IPHost` and `IPNetwork` and `MACAddress`. Called in `BaseAttribute.__init__()` after validation. For these classes, `serialize_value` has been removed.

## How to test

```bash
uv run pytest backend/tests/component/core/test_manager_node.py::test_hfid_round_trip_with_iphost_attribute -x -v
```

## Impact & rollout

- **Backward compatibility:** IPHost values now always include the prefix in `attr.value`. Code that previously relied on the raw un-prefixed form (e.g., `"192.0.2.1"`) from `attr.value` will now see `"192.0.2.1/32"`. Display labels using `{{ ip_address__value }}` will include the mask; users wanting mask-free display can use `{{ ip_address__ip }}`.
- **Deployment notes:** Safe to deploy. Existing nodes with stale HFIDs (stored without mask) will be corrected when their HFID is next recomputed (e.g., on update). A separate data migration may be desirable for bulk correction.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`uv run towncrier create ...`)


polmichel:
## AI Root cause analysis for #8896

**Issue:** bug: IPHost attribute values vary in when they require/return their netmask from API and UI
**Branch:** `ai-bug-pipeline-8896-iphost-hfid-consistency`
**Based on:** `741208b722f0544275a69c4d0acf087d2cc3d736`
**Bug clarity:** CLEAR
**Code identification:** RESOLVED

### Root cause
IPHost values are only normalized to canonical `with_prefixlen` form during DB serialization (`serialize_value()`), but HFID and display labels are computed from the raw `attr.value` **before** serialization occurs, causing the stored HFID/display label to disagree with the stored attribute value.

### Affected files
- `backend/infrahub/core/attribute.py` -- line 1111-1114: `IPHost.serialize_value()` normalizes to `with_prefixlen` but only at DB write time
- `backend/infrahub/core/attribute.py` -- line 128-157: `BaseAttribute.__init__()` sets `self.value` from raw user input with no normalization
- `backend/infrahub/core/node/__init__.py` -- line 877-878: `_create()` computes HFID and display label **before** `NodeCreateAllQuery` (line 880) which calls `to_db()` → `serialize_value()`
- `backend/infrahub/core/node/__init__.py` -- line 248-250: `get_path_value()` returns `attr.value` directly (raw at creation, serialized after DB load)
- `backend/infrahub/core/attribute.py` -- line 975-978: `IPNetwork.serialize_value()` has the same pattern

### Explanation

The inconsistency is a timing issue between value normalization and HFID/display label computation during node creation:

**Creation flow (user enters "192.0.2.1" without mask):**
1. `Node.new()` → `_process_fields()` → `BaseAttribute.__init__()` sets `self.value = "192.0.2.1"` (raw input)
2. `Node._create()` line 877: `add_human_friendly_id()` → `compute()` → `get_path_value("ip_address__value")` → returns `attr.value` = `"192.0.2.1"` → HFID stored as `["192.0.2.1"]`
3. `Node._create()` line 878: `add_display_label()` → same flow → display label stored as `"192.0.2.1"`
4. `Node._create()` line 880: `NodeCreateAllQuery` → `to_db()` → `serialize_value()` → `"192.0.2.1/32"` stored as the attribute value in DB

**After creation, the DB contains:**
- IPHost attribute value: `"192.0.2.1/32"` (normalized)
- HFID list attribute: `["192.0.2.1"]` (NOT normalized)
- Display label: `"192.0.2.1"` (NOT normalized)

**HFID round-trip failure** (`get_one_by_hfid` at `core/manager.py:771-783`):
- GraphQL returns HFID from stored value: `["192.0.2.1"]` (no mask)
- Lookup builds filter `{"ip_address__value": "192.0.2.1"}` and queries against DB
- DB has `"192.0.2.1/32"` → no match → lookup fails
- User must manually append `/32` for the lookup to succeed

**After loading from DB**, `attr.value` becomes `"192.0.2.1/32"` (the serialized form from `deserialize_value()`), which is why the SDK returns the value with mask and why recomputed HFIDs would include the mask — but the **stored** HFID was set at creation time without it.

IPNetwork has the same structural issue (`serialize_value()` at line 975-978), though it's less visible because users typically provide a prefix for networks.

## Fix strategy

**Approach:** Normalize IPHost and IPNetwork values at input time rather than only at DB serialization time. Add a `_canonicalize_value` method to `BaseAttribute` (no-op by default) and override it in `IPHost` to return `ipaddress.ip_interface(value).with_prefixlen` and in `IPNetwork` to return `ipaddress.ip_network(value).with_prefixlen`. Call `self.value = self._canonicalize_value(self.value)` in `BaseAttribute.__init__()` after validation (after line 157) and also after `value_from_db` in `load_from_db` for consistency. This ensures `attr.value` is always in canonical form regardless of when it's accessed, so HFID/display label computation sees the same representation as what ends up in the DB.

**Scope:** 
- `backend/infrahub/core/attribute.py`: Add `_canonicalize_value` to `BaseAttribute`, override in `IPHost` and `IPNetwork`. Call it in `__init__` after validation. ~15 lines changed.
- `backend/tests/component/core/test_attribute.py`: Update test assertions where `IPHost(data="192.0.2.1").value` currently returns `"192.0.2.1"` — it should now return `"192.0.2.1/32"`.
- Integration/component tests for HFID round-trip with IPHost attributes.

**Do NOT:**
- Do NOT try to strip the prefix from the stored DB value or from HFID lookups — that would require changing the query layer and break existing data
- Do NOT add special-casing in `get_path_value()` or `HumanFriendlyIdentifier.compute()` to call `serialize_value()` — that couples the HFID layer to attribute serialization internals
- Do NOT change the `serialize_value()` method itself or its call site in `to_db()` — the fix is upstream of that, at input time

## Notes for downstream steps
- The `serialize_value()` call in `to_db()` becomes effectively idempotent for IPHost/IPNetwork after this fix (normalizing an already-normalized value). Verify it's safe: `ipaddress.ip_interface("192.0.2.1/32").with_prefixlen` → `"192.0.2.1/32"` (confirmed idempotent).
- Existing data in production databases has HFIDs stored without the mask. A data migration may be needed for existing nodes (recomputing HFIDs), but this is out of scope for the code fix. The code fix ensures new nodes are correct and that recomputed HFIDs (on update) will also be correct.
- IPv6 addresses have the same issue: `"2001:db8::"` normalizes to `"2001:db8::/128"` for IPHost. Tests should cover both IPv4 and IPv6.
- The `validate_format` on IPHost (line 1092-1109) accepts bare addresses, so this doesn't need changing.
- The existing unit test at `backend/tests/component/core/test_attribute.py:83-84` creates `IPHost(data="192.0.2.1")` — the assertion that `self.value` remains `"192.0.2.1"` will need updating to `"192.0.2.1/32"`.
- IPNetwork normalization also handles IPv6 collapse (e.g., `"2001:0db8:0000::/48"` → `"2001:db8::/48"`), which is already done by `serialize_value()`.
- Display labels in the UI that use Jinja2 templates referencing `{{ ip_address__value }}` will now include the mask. If mask-free display is desired, users can use `{{ ip_address__ip }}` in their template instead. This is a behavioral change worth documenting.

backend/tests/component/core/test_manager_node.py
```diff
@@ -280,6 +280,70 @@ async def test_get_by_hfid_with_invalid_hfid(db: InfrahubDatabase, branch: Branc
...
+
+    # The HFID must use the canonical EUI-48 form so it matches the DB-stored
+    # attribute value and can round-trip through get_one_by_hfid.
+    assert hfid == ["AA-BB-CC-DD-EE-FF"]
+
+    found = await NodeManager.get_one_by_hfid(db=db, hfid=hfid, kind="TestInterface")
+    assert found is not None
+    assert found.id == node.id
+
+
```
gmazoyer: Not sure we need HFID tests for this issue. I think tests validating that no matter the format of an IP address/network or a MAC address these are stored in a consistent way would be enough

backend/infrahub/core/attribute.py
```diff
@@ -155,6 +155,7 @@ def __init__(
 
         if self.value is not None:
             self.validate(value=self.value, name=self.name, schema=self.schema)
+            self.value = self._canonicalize_value(self.value)
```
ajtmccarty: is `canonicalize` a networking term? `normalize` seems more common

backend/infrahub/core/attribute.py
```diff
@@ -978,9 +986,8 @@ def validate_format(cls, value: Any, name: str, schema: AttributeSchema) -> None
             raise ValidationError({name: f"{value} is not a valid {schema.kind}"}) from exc
 
     def serialize_value(self) -> str:
-        """Serialize the value before storing it in the database. If network is an IPv6 network, it is converted to collapsed form."""
-
-        return ipaddress.ip_network(self.value).with_prefixlen
+        """Serialize the value before storing it in the database."""
+        return self._canonicalize_value(self.value)
```
ajtmccarty: it looks like this should be handled by the update to `BaseAttribute.__init__`, so maybe `serialize_value` can be removed from `IPNetwork` and `IPHost` and `MacAddress`, but maybe I don't understand the order of operations

backend/tests/component/core/test_attribute.py
```diff
@@ -274,7 +274,9 @@ async def test_validate_mac_address_returns(
         name="test", schema=schema, branch=default_branch, at=Timestamp(), node=None, data=mac_address
     )
 
-    assert test_mac.value == mac_address
+    # MacAddress canonicalizes the input to the standard EUI-48 form so the
+    # in-memory value matches the DB-stored representation.
+    assert test_mac.value == test_mac._canonicalize_value(mac_address)
```
ajtmccarty: this isn't really testing that the MAC address is in the expected format. it's just testing that the mac address is in the format returned by `_canonicalize_value`

ajtmccarty (commented):
does this need a migration for existing non-normalized IP networks/hosts and MAC addresses?

backend/tests/component/core/test_attribute.py
polmichel: You are right, I have left the comme and remove the call to `_canonicalize_value`. Value is now hardcoded

backend/tests/component/core/test_manager_node.py
polmichel: You are right, this covers the issue scenario but is weird from a system expectations perspective. Will just check the attribute value on the `Attribute` class after `Node.save()` method, thank you

backend/infrahub/core/attribute.py
polmichel: I have heard it in mathematics, but normalize seems more common, thanks

backend/infrahub/core/attribute.py
polmichel: You are right, initially I spotted that `serialize_value` was containing the same logic but the refactor was not good. This is logic should not be in database serialization step but value normalization step.

polmichel:
> does this need a migration for existing non-normalized IP networks/hosts and MAC addresses?

I do not think so. Because the value was serialized before being saved into the database, so the real value in the database should be correct. However, the initial value displayed in _display_label_ or _hfid_ fields was using the `Attribute.value` before refetching it from the database. This value was not serialized (yet).

I have updated the PR description with an additional element about this precision.

dev/knowledge/backend/display-labels-and-hfid.md
```diff
@@ -119,6 +119,17 @@ The trigger definitions and gathering logic live in `backend/infrahub/display_la
...
 `set_display_label(value)` and `set_human_friendly_id(value)` set `manually_assigned=True`, which causes `compute()` to no-op on future calls.
 
+### Value Normalization (HFID Round-Trip)
+
+For attribute kinds whose accepted input form differs from their normalized storage form (`IPHost`, `IPNetwork`, `MacAddress`), `BaseAttribute.__init__()` calls `_normalize_value()` after `validate()` so `attr.value` always matches what `serialize_value()` writes to the DB. Without this, HFID compute (which reads `attr.value` *before* serialization) would emit the raw input while the DB stored the normalized form, breaking `get_one_by_hfid()` round-trips. See #8896.
+
+| Kind | Normalized form |
+|------|-----------------|
+| `IPHost` | `ipaddress.ip_interface(value).with_prefixlen` (e.g. `192.0.2.1` → `192.0.2.1/32`) |
+| `IPNetwork` | `ipaddress.ip_network(value).with_prefixlen` |
```
ajtmccarty: no example for this one?

backend/tests/component/core/test_manager_node.py
```diff
@@ -280,6 +280,56 @@ async def test_get_by_hfid_with_invalid_hfid(db: InfrahubDatabase, branch: Branc
...
+    node = await Node.init(db=db, schema="TestInterface", branch=default_branch)
+    await node.new(db=db, mac="aa:bb:cc:dd:ee:ff")
+    await node.save(db=db)
+
+    assert node.mac.value == "AA-BB-CC-DD-EE-FF"
+
+    reloaded = await NodeManager.get_one(db=db, id=node.id, branch=default_branch)
+    assert reloaded is not None
+    assert reloaded.mac.value == "AA-BB-CC-DD-EE-FF"
+
```
ajtmccarty, line 331: should there be a test for retrieving a Node that was saved before this change was made? i.e. an object with an IP address of `192.168.0.0` in the database is retrieved as `192.168.0.0/32`. I believe it would require a cypher query to set the data to be incorrect

ajtmccarty (approved):

dev/knowledge/backend/display-labels-and-hfid.md
polmichel: Good catch, thank you

backend/infrahub/core/attribute.py
```diff
@@ -1142,6 +1143,9 @@ class MacAddress(BaseAttribute):
     type = str
     value: str
 
+    def _normalize_value(self, value: Any) -> str:
```
BaptisteGi: My 2cents here: recently opened https://github.com/opsmill/infrahub/issues/9015, from a user perspective the `:` notation is more natural. Doesn't mean we need to change the way it's stored in the database but perhaps only the way it's returned in the API so the `value` returns MAC address in `:` format.

backend/tests/component/core/test_manager_node.py
polmichel: The attribute value was already serialized before reaching the database. The HFID and display labels were using the value before the serialization, at least until the node is recomputed. So the test would be:

1. Creating a fake value through a Cypher query to set up a fake value on a HFID attribute
2. Update the node
3. Verify the HFID is correctly recomputed

I think this is something that is already tested somewhere and would not bring value to the test suite.
This means that node containing wrong HFID value will be updated to correct HFID value when the node is updated. Would this be a correct way of handling this in your opinion?

backend/infrahub/core/attribute.py
polmichel: In my opinion, a specific update to this value using a database migration would be a better long-term solution, as it would avoid to add another step of data transformation between the database and the UI. This kind of layer is the root cause of this bug.
In that case, I would handle the value format update in another PR.

backend/tests/component/core/test_manager_node.py
ajtmccarty: right. I forgot this is only for HFIDs. I know I already asked about a migration for the existing IP Network/Host data and I agree that we don't need that. But now I'm wondering if there should be a migration to updates HFIDs using IPNetwork/Host or MAC address attributes to make sure they all have the same format.

I guess I'm not certain if this fix addresses this particular part of the bug report
> GraphQL: Queries that look up an instance by its hfid require the mask to be included in the hfid input, but the hfid returned by the GraphQL query response does not include the mask — making the returned hfid unusable as a direct input for subsequent lookups.

polmichel: If the node update is not enough as a synchronization tool, let's do a migration for this. In that case, I will add another changelog and do the migration about @BaptisteGi comment in a chained PR targeting this branch.

polmichel: Regarding the second part, I think this PR resolves the issue as the graphQL query will return the value of the hfid through
`infrahub.core.attribute.BaseAttribute.to_graphql` which get the attribute directly through `get_attr()` if I am not wrong. 
I will add an integration test in order to verify that there is no additional behavior that would remove the normalization when going through the graphQL layer. However, I do not think there would be one.

polmichel: for the component test reproducing the HFID lookup query see 4da9aa7c707d70fb1c1fc8dda3d9700d57ff4d9f

backend/tests/component/graphql/queries/test_hfid.py
```diff
@@ -0,0 +1,89 @@
+from infrahub.core import registry
```
polmichel, line 1: GraphQL query-level test about the look up HFID query

backend/infrahub/core/migrations/graph/m071_recompute_hfid_for_ip_mac_attributes.py
```diff
@@ -0,0 +1,310 @@
+from __future__ import annotations
```
polmichel: This is reusing the logic from m044 and m047

backend/infrahub/core/migrations/query/path_details.py
```diff
@@ -0,0 +1,403 @@
+from __future__ import annotations
```
polmichel, line 1: This coming from existing code which has been moved

backend/infrahub/core/migrations/helpers/display_label.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
```
polmichel, line 1: This coming from existing code which has been moved

backend/tests/component/core/migrations/graph/test_070_recompute_hfid_for_ip_mac_attributes.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+class TestMigration070(TestInfrahubApp):
+    @pytest.fixture(scope="class", autouse=True)
+    async def normalized_kind_schema(
+        self, db: InfrahubDatabase, default_branch: Branch, register_core_schema: SchemaBranch
+    ) -> SchemaBranch:
+        return registry.schema.register_schema(schema=_schema_root(), branch=default_branch.name)
+
+    @pytest.mark.parametrize("case", CASES, ids=[c.attr_kind for c in CASES])
```
ajtmccarty: this test looks like it might be slow. is it possible to load all the cases, run the migration, then verify all the cases? I suspect that would be faster

backend/tests/component/core/migrations/graph/test_071_recompute_hfid_for_ip_mac_attributes.py
```diff
@@ -0,0 +1,246 @@
...
+    NormalizedKindCase(
+        kind="TestingInterface",
+        schema_name="Interface",
+        attr_name="mac",
+        attr_kind="MacAddress",
+        raw_value="aa:bb:cc:dd:ee:ff",
+        canonical_value="AA-BB-CC-DD-EE-FF",
+        node_name="eth0",
+    ),
+]
```
ajtmccarty: given that this a `MigrationRequiringRebase` it would be good to test that this migration correctly handles schemas and objects that have been updated on a branch

polmichel:
A related PR focusing on MAC address format has been created https://github.com/opsmill/infrahub/pull/9134 cc @BaptisteGi

backend/tests/component/core/migrations/graph/test_071_recompute_hfid_for_ip_mac_attributes.py
polmichel: Done in the new test (and tested through a real application update)

polmichel:
PRs https://github.com/opsmill/infrahub/pull/8904 and https://github.com/opsmill/infrahub/pull/9149 have now an inverted order.

backend/tests/component/core/migrations/graph/test_071_recompute_hfid_for_ip_attributes.py
```diff
@@ -0,0 +1,295 @@
+from __future__ import annotations
```
polmichel, line 1: Reviews from https://github.com/opsmill/infrahub/pull/9149/changes#r3201946813 have also been taken in account on this migration test

polmichel:
**EDIT**: looks good

I have to check one last thing on execute_on_branch method regarding branch awareness type

ajtmccarty (approved):

PR #8905, opened 2026-04-14
Bug-fixing agent pipeline Claude commands

polmichel:
## Why

Using the pipeline through the CI is one use case which is currently more complex than necessary. To enable a wider audience to use the bug-fixing tool, it would be useful to have a set of Claude commands, which would give users more control and reduce their reliance on CI-related security guidelines.

### Claude commands

This PR introduces three new commands: 
- `/bug-analyze <issue_number or issue_url>`: creates a local .md file `.bug-analysis-<issue_number>.md` containing the issue analysis.
- `/bug-tdd <issue_number or issue_url> [pr]`: reads the `.bug-analysis-<issue_number>.md`, creates a working branch, reproduces the issue with a failing test, and commits it. Optionally opens a draft PR when `pr` is passed.
- `/bug-fix <issue_number or issue_url>`: reads the `.bug-analysis-<issue_number>.md`, identifies a solution and validates it using the test created by the `/bug-tdd` command.

### Usage examples

Run all three commands sequentially in Claude Code to go from issue to fix:

```bash
# Step 1 — Analyze the bug (creates .bug-analysis-1794.md, no branch)
/bug-analyze 1794

# Step 2 — Write a failing test (creates branch, commits test)
/bug-tdd 1794

# Step 2 (alt) — Write a failing test and open a draft PR
/bug-tdd 1794 pr

# Step 3 — Implement the fix (pushes fix, updates PR)
/bug-fix 1794
```

You can also pass a full GitHub URL instead of a number:

```bash
/bug-analyze https://github.com/opsmill/infrahub/issues/1794
```

### Refactoring of shared prompts part

As the Claude commands rely on the same tool than the bug-fixing CI pipeline, a refactoring has been done to extract shared prompts into `.github/bug-agent-pipeline/shared` folder.

### Renamed `/bug-test` into `/bug-tdd`

As the command is generating a failing test, in my opinion it would be clearer to call it `/bug-tdd`.

### Removed "fix-bug" command

Removed the command which was not used and may be confusing to keep.


.github/bug-agent-pipeline/shared/investigation.md
```diff
@@ -0,0 +1,90 @@
...
+
+Create a working branch from `origin/stable`:
+
+```bash
+git fetch origin stable
+git checkout -b ai-bug-pipeline-<issue_number>-<short-slug> origin/stable
+```
+
+- Name: `ai-bug-pipeline-<issue_number>-<short-slug>` (lowercase, hyphens only, max 50 chars total).
+- If the branch already exists, check it out instead of creating a new one.
```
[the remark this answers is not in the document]
polmichel: Not important, for now.

dgarros:
> /bug-analyze <issue_number or issue_url>: creates a local .md file .bug_analysis.md containing the issue analysis, and creates a branch named ai-bug-pipeline-<issue_number>-<short_desc>.

I think it would be good to add the issue number in the name of the file as well, to allow multiple analysis to exist locally without overwriting each other. 
Also, it's not clear to me why `/bug-analyze` should create a branch, are we committing anything at this point ?

> /bug-tdd <issue_number or issue_url>: reads the .bug_analysis.md, reproduces the issue with a failing test, commits it to the branch and creates a draft PR with the test.

To start, I think it would be good to make the draft PR optional, and only do it if the user ask for it, it should be easy to add to the command `/bug-tdd <issue_number or issue_url> pr` or similar

polmichel:
> > /bug-analyze <issue_number or issue_url>: creates a local .md file .bug_analysis.md containing the issue analysis, and creates a branch named ai-bug-pipeline-<issue_number>-<short_desc>.
> 
> I think it would be good to add the issue number in the name of the file as well, to allow multiple analysis to exist locally without overwriting each other. Also, it's not clear to me why `/bug-analyze` should create a branch, are we committing anything at this point ?

Thanks for the feedback, I will add these naming details.
Regarding why the analyst agent is the one creating the branch, the idea was that they would be the only agent having the right to create or modify branches, and then we would not need to take care of these rights. Now that these commands are separate and not a whole, it might be better to give this responsibility to the test writer, otherwise we might end up with a lot of branches, you are right.


> > /bug-tdd <issue_number or issue_url>: reads the .bug_analysis.md, reproduces the issue with a failing test, commits it to the branch and creates a draft PR with the test.
> 
> To start, I think it would be good to make the draft PR optional, and only do it if the user ask for it, it should be easy to add to the command `/bug-tdd <issue_number or issue_url> pr` or similar

Yes, good idea, will do.

dev/commands/bug-analyze.md
```diff
@@ -0,0 +1,55 @@
...
+When the shared investigation instructions say to "STOP and escalate":
+- If the issue is **UNCLEAR**: inform the developer what information is missing and **STOP**.
+- If you **cannot identify a root cause**: inform the developer and **STOP**.
+
+### Output
+
+Write the analysis to `.bug-analysis-<issue_number>.md` in the repo root using the
+analysis template from the shared investigation file.
+
+Fetch the latest remote state and fill the **Based on** field with the current `origin/stable` SHA:
```
dgarros, line 46: Not a blocker for now but I don't think we should hardcode the name of the base branch, I'm sure we'll want to analyze / fix bugs that are part of other branches in the future.

.github/bug-agent-pipeline/test_pipeline_shell_logic.sh
```diff
@@ -1229,20 +1229,22 @@ _TEST_WRITER_COMMON = [
...
     # Denied -- exact match blocks chaining
     ("Bash(uv run invoke format && rm -rf /)",         False, "invoke format chaining blocked"),
     ("Bash(npx biome check --write . && curl evil)",   False, "biome chaining blocked"),
     ("Bash(git status --porcelain)",                   False, "git status with flags blocked"),
 ]
 
-# write-test can create PRs; revise-test cannot
+# write-test can create branches, fetch, and create PRs; revise-test cannot
 WRITE_TEST_SCENARIOS = _TEST_WRITER_COMMON + [
+    ("Bash(git checkout -b ai-bug-pipeline-1042)",     True,  "create pipeline branch"),
```
dgarros, line 1241: Looks like the name of the branch is hardcoded here

.github/bug-agent-pipeline/test_pipeline_shell_logic.sh
```diff
@@ -1229,20 +1229,22 @@ _TEST_WRITER_COMMON = [
...
+    ("Bash(git fetch origin stable)",                  True,  "git fetch"),
     ("Bash(gh pr create --title test-pr)",             True,  "create PR"),
 ]
 
 REVISE_TEST_SCENARIOS = _TEST_WRITER_COMMON + [
+    ("Bash(git checkout -b ai-bug-pipeline-1042)",     False, "create branch (revise uses existing)"),
```
dgarros, line 1247: +1

dgarros (approved):
Let's ship it, great work
thank you for adding the claude commands,

I think we should get to a point where we are using the commands in GHA directly, it would simplify a lot of things but that's fine for now, we'll continue to iterate on it.

.github/bug-agent-pipeline/test_pipeline_shell_logic.sh
polmichel: This is local testing scenarios to validate which commands are allowed or denied

.github/bug-agent-pipeline/test_pipeline_shell_logic.sh
polmichel: This is local testing scenarios to validate which commands are allowed or denied

dev/commands/bug-analyze.md
polmichel: Thank you for the feedback.
I agree, I will note it as enhancement in the internal section about next steps https://opsmill.atlassian.net/wiki/spaces/SD/pages/522878980/Next+steps+ideas

PR #8907, opened 2026-04-14
Fix display_label backfill after schema change

gmazoyer:
## Why & What

When a schema was updated to add a display_label, the async Prefect backfill ran but skipped the update. The on-the-fly fallback in `get_display_label` computed the correct value during the backfill's own GraphQL query, making it appear no update was needed. The DB value stayed null, and regular queries returned empty strings.

Fix `get_display_label` to return `""` when the stored value is empty for saved nodes, so the backfill detects the difference and writes the correct value. On-the-fly computation is preserved only for virtual nodes (IPAM) that are never persisted.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Fixes display_label backfill by stopping on-the-fly computation for persisted nodes so DB updates occur. Adds a Docker integration test and updates internal docs to clarify query-time behavior and backfill flow.

- **Bug Fixes**
  - Update `get_display_label`: if `_display_label` exists and has a value, return it; if it exists but is empty and the schema defines a display label, return "" instead of computing; compute only for virtual nodes.

- **Refactors**
  - Add Docker integration test for schema-update backfill using pytest fixtures; remove a duplicated import; expand internal knowledge docs on display labels and backfill.

<sup>Written for commit 017767fc092df5ebd3fc0fc5ee88a0330851a54e. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -218,15 +218,15 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         await self._human_friendly_id.compute(db=db, node=self)
 
     async def get_display_label(self, db: InfrahubDatabase) -> str:
-        if self._display_label and (value := self._display_label.get_value(node=self, at=self._at)):
-            return value
+        if self._display_label is not None:
+            if value := self._display_label.get_value(node=self, at=self._at):
+                return value
+            # Stored value is empty do not compute on the fly to avoid recomputation trigger issues
+            if self._schema.display_label:
+                return ""
 
         if not self._schema.display_label:
             return repr(self)
-
-        if self._display_label is not None:
-            # Stored value is empty do not compute on the fly to avoid recomputation trigger issues
-            return ""
 
         # This should only happens for "virtual" nodes (that never exist inside the db)
         display_label = DisplayLabel(node_schema=self._schema, template=self._schema.display_label)
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -224,6 +224,11 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
         if not self._schema.display_label:
             return repr(self)
 
+        if self._display_label is not None:
+            # Stored value is empty do not compute on the fly to avoid recomputation trigger issues
+            return ""
+
+        # This should only happens for "virtual" nodes (that never exist inside the db)
         display_label = DisplayLabel(node_schema=self._schema, template=self._schema.display_label)
         await display_label.compute(db=db, node=self)
         return display_label.get_value(node=self, at=self._at) or ""
--- a/backend/tests/integration_docker/test_display_label_backfill.py
+++ b/backend/tests/integration_docker/test_display_label_backfill.py
@@ -4,7 +4,6 @@ from asyncio import sleep
 from typing import TYPE_CHECKING
 
 import pytest
-from infrahub_sdk import InfrahubClient
 from infrahub_sdk.task.models import TaskFilter, TaskState
 from infrahub_sdk.testing.docker import TestInfrahubDockerClient
 
--- a/backend/tests/integration_docker/test_display_label_backfill.py
+++ b/backend/tests/integration_docker/test_display_label_backfill.py
@@ -51,39 +51,43 @@ class TestDisplayLabelBackfillOnSchemaChange(TestInfrahubDockerClient):
     def schema_with_display_label(self) -> dict:
         return {"version": "1.0", "nodes": [COLOR.model_dump()]}
 
+    @pytest.fixture(scope="class")
+    async def color_before(self, client: InfrahubClient, schema_without_display_label: dict) -> str:
+        """Load schema without display_label and create a node."""
+        response = await client.schema.load(schemas=[schema_without_display_label], wait_until_converged=True)
+        assert response.schema_updated
+
+        color = await client.create(kind="TestingColor", name="Red", description="A warm color")
+        await color.save()
+        return color.id
+
+    @pytest.fixture(scope="class")
+    async def color_after(self, client: InfrahubClient, color_before: str, schema_with_display_label: dict) -> str:
+        """Update schema to add display_label and create a second node."""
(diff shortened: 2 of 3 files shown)


backend/tests/integration_docker/test_display_label_backfill.py
```diff
@@ -0,0 +1,91 @@
...
+        # Display label is null, so defaults to repr()
+        assert labels[color.id] == f"TestingColor(ID: {color.id})"
+
+    async def test_step03_load_schema_with_display_label(
+        self, client: InfrahubClient, schema_with_display_label: dict
+    ) -> None:
+        response = await client.schema.load(schemas=[schema_with_display_label], wait_until_converged=True)
+        assert response.schema_updated
+
+    async def test_step04_create_second_node(self, client: InfrahubClient) -> None:
```
polmichel: This is not blocking at all and optional.
However, as tests should not be order dependent, I think steps 1 to 3 should be chained fixtures (in my opinion they look more like setup steps than real tests.). Steps 4 and 5 are, in fact, two independent tests with step 5 only verifying the "Red" part which is a different scenario (pre-existing node). What do you think?

backend/tests/integration_docker/test_display_label_backfill.py
```diff
@@ -0,0 +1,91 @@
...
+async def wait_for_all_tasks_to_be_completed(client: InfrahubClient) -> None:
+    while (  # noqa: ASYNC110
+        await client.task.count(filters=TaskFilter(state=[TaskState.PENDING, TaskState.RUNNING, TaskState.SCHEDULED]))
+        > 0
+    ):
+        await sleep(1)
+
+
+class TestDisplayLabelBackfillOnSchemaChange(TestInfrahubDockerClient):
+    """IFC-2459: existing nodes should get their display_label backfilled
```
polmichel: Ideally, I would prefer it if there were no Jira/GH issues or mention of the initial bug when reading a test, which, in my opinion, should specify how the system should behave.
But this is just my personal opinion that I bring up here, you can ignore it

dev/knowledge/backend/display-labels-and-hfid.md
```diff
@@ -92,6 +92,29 @@ if self._existing:
 2. `Node.save()` -> `resolve_relationships()` (extra_filters gated by `self._human_friendly_id` / `self._display_label` for those sources, plus always-on computed attribute extra filters via `computed_attributes.get_registered_jinja2_node`)
 3. `_update()` checks `needs_update(fields)` for each property; recomputes and persists if needed
 
+### Query-Time Resolution (`get_display_label`)
+
+`Node.get_display_label(db)` returns the display label for a node during GraphQL queries. It distinguishes between saved and virtual nodes:
+
+1. **Stored value exists** (`_display_label` set with a non-empty value): return it directly.
+2. **Stored attribute is empty** (`_display_label` set but value is null): return `""`. The async backfill workflow is responsible for populating stored values after schema changes. Computing on the fly here would cause the backfill to detect no difference and skip the update (IFC-2459).
```
polmichel: The same applies to the Jira issue in my opinion. I think this is irrelevant in the documentation, whether for the agent or for us. I think that when we read this documentation, we want to know what is actually happening in the code, and not the history about it. do you agree?

backend/infrahub/core/node/__init__.py
```diff
@@ -218,12 +218,17 @@ async def add_human_friendly_id(self, db: InfrahubDatabase) -> None:
...
+            if value := self._display_label.get_value(node=self, at=self._at):
+                return value
+            # Stored value is empty do not compute on the fly to avoid recomputation trigger issues
+            if self._schema.display_label:
+                return ""
 
         if not self._schema.display_label:
             return repr(self)
 
+        # This should only happens for "virtual" nodes (that never exist inside the db)
```
polmichel, line 231: Oh ok I hadn't understood this before. Nice comment!

dev/knowledge/backend/display-labels-and-hfid.md
```diff
@@ -92,6 +92,29 @@ if self._existing:
 2. `Node.save()` -> `resolve_relationships()` (extra_filters gated by `self._human_friendly_id` / `self._display_label` for those sources, plus always-on computed attribute extra filters via `computed_attributes.get_registered_jinja2_node`)
 3. `_update()` checks `needs_update(fields)` for each property; recomputes and persists if needed
 
+### Query-Time Resolution (`get_display_label`)
```
polmichel, line 95: This is cool that you update the internal documentation each time. I am curious, are you using the `feedback` or `audit-docs` commands or are you directly prompting?

backend/tests/integration_docker/test_display_label_backfill.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import annotations
+
+from asyncio import sleep
+from typing import TYPE_CHECKING
+
+import pytest
+from infrahub_sdk import InfrahubClient
```
polmichel: This import is not used

polmichel (approved):
Nice integration test, I have added small comments on issue mention in documentation and tests and one about fixtures versus test_step_X.
Part of them are just personal opinion, feel free to ignore them of course

backend/tests/integration_docker/test_display_label_backfill.py
gmazoyer: Nice catch

backend/tests/integration_docker/test_display_label_backfill.py
gmazoyer: Agreed, it slips past my eyes

dev/knowledge/backend/display-labels-and-hfid.md
gmazoyer: I did use `audit-docs` to identify the gaps and patch it.

backend/tests/integration_docker/test_display_label_backfill.py
gmazoyer: Makes sense.

dev/knowledge/backend/display-labels-and-hfid.md
polmichel: A few adjustments have to been made regarding issue mention then, good to know, thanks

polmichel (approved):
Thank you!

PR #8908, opened 2026-04-14
Convert binary test repo to FileRepo

ogenstad:
# Why

Replace the binary repo within the tests with a repo populated by the FileRepo class instead. The reason is that it's hard to figure out what's happening within the tar.gz file without extracting the content and viewing offline. Test failures and future changes to this binary file is also problematic.

## What changed

* Extract repository from tar.gz file and remove the original
* Replace repo content with something that works with FileRepo
* Small modifications to FileRepo to allow for a branch without additional commits
* Replace old references to static git commits within tests

---
## Summary by cubic
Replaced the binary git test repo tarball with a `FileRepo`-built fixture for transparent, maintainable tests. Tests now compute commit SHAs and branch heads at runtime and compare against remote refs, removing hard-coded hashes.

- **Refactors**
  - Removed `backend/tests/fixtures/infrahub-test-fixture-01-f165752.tar.gz` and added a file-based repo under `backend/tests/fixtures/repos/infrahub-test-fixture-01` with `main`, `branch01`, `branch02`, and `clean-branch`.
  - Updated `conftest.py` to build the upstream repo via `FileRepo`; tests read commits from the upstream repo, assert 40-char SHAs, ensure distinct branch heads, and compare created branches against remote refs (or `main`) at runtime.
  - Tweaked `tests/helpers/file_repo.py`: re-checkout the initial branch before creating new branches, commit only when the index is dirty, and allow branches without additional commits.

<sup>Written for commit 34f9e8081b9a5cbb58001a22b12b6c49eb65617b. Summary will update on new commits.</sup>


backend/tests/component/git/test_git_repository.py
```diff
@@ -295,9 +296,17 @@ async def test_get_branches_from_graph(
...
-    assert repo.get_commit_value(branch_name="branch01", remote=True) == "30e911e25ef9e4fad9f9d00fe05395031f90d460"
-    assert repo.get_commit_value(branch_name="branch02", remote=True) == "4e2fd98a5fd1fb61dc53150c778e22ee35f26191"
+    commit_main = repo.get_commit_value(branch_name="main", remote=True)
+    commit_branch01 = repo.get_commit_value(branch_name="branch01", remote=True)
+    commit_branch02 = repo.get_commit_value(branch_name="branch02", remote=True)
+
+    # Each value should be a full 40-character SHA
+    assert re.fullmatch(r"[0-9a-f]{40}", commit_main)
+    assert re.fullmatch(r"[0-9a-f]{40}", commit_branch01)
+    assert re.fullmatch(r"[0-9a-f]{40}", commit_branch02)
```
ajtmccarty, line 306: do we not know the full commit SHA any longer? or is it not important to test?

ajtmccarty (approved):

backend/tests/component/git/test_git_repository.py
ogenstad: No we don't know them any longer, the only reason we knew them before was that they were. contained in a `.tar.gz` file which is what I wanted to avoid with this. Because of the reason the commits will be unique when we hand craft them each time.

PR #8913, opened 2026-04-14
Add integration tests for git operations

ogenstad:
# Why

Add integration tests for git operations which are good to have in place for upcoming refactor work

## What changed

* Adds tests and slight refactoring of the gogs fixture
* Note that one of the added tests are expected to later fail once we've reworked how the code works.

---
## Summary by cubic
Adds live-remote Git integration tests against a Gogs server and refactors the test fixture to a typed, session-scoped `GogsServer` for more reliable and faster test runs.

- **New Features**
  - Added `backend/tests/integration/git/test_git_live_remote.py` to cover:
    - Clone with wrong credentials → raises `RepositoryCredentialsError` and sets `ERROR_CRED`.
    - Push rejection (non-fast-forward) currently returns True; documents current behavior and is expected to fail after push-rejection handling is added.
    - Local merge conflict → raises `RepositoryError` and leaves a clean repo state.
    - Read-only repo sync detects remote changes and no-op when up to date.

- **Refactors**
  - Introduced `backend/tests/helpers/git.py` with a `GogsServer` dataclass.
  - Converted Gogs fixture to session scope and returned `GogsServer` instead of a dict.
  - Added `bad_credentials_clone_url()` and `create_gogs_repo(..., private=True)` for auth tests.
  - Updated existing Gogs branch deletion tests to use the new typed fixture.

<sup>Written for commit 8fcd8d3386aca9d81ddd985c33c45ae9871ab49d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/git/test_git_live_remote.py
+++ b/backend/tests/integration/git/test_git_live_remote.py
@@ -1,0 +1,326 @@
+"""Integration tests for InfrahubRepository and InfrahubReadOnlyRepository against a live remote."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import pytest
+
+from infrahub.core.constants import InfrahubKind, RepositoryOperationalStatus
+from infrahub.core.manager import NodeManager
+from infrahub.core.node import Node
+from infrahub.exceptions import RepositoryCredentialsError, RepositoryError
+from infrahub.git.repository import InfrahubReadOnlyRepository, InfrahubRepository
+from tests.helpers.test_app import TestInfrahubApp
+from tests.integration.git.conftest import bad_credentials_clone_url, create_gogs_repo
+
+if TYPE_CHECKING:
+    from infrahub_sdk import InfrahubClient
+    from testcontainers.core.container import DockerContainer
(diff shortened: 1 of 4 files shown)


backend/tests/integration/git/test_git_live_remote.py
```diff
@@ -0,0 +1,332 @@
...
+
+    async def test_clone_with_wrong_credentials_raises_credentials_error(
+        self,
+        auth_failure_dataset: dict,
+        db: InfrahubDatabase,
+        client: InfrahubClient,
+    ) -> None:
+        """A real auth rejection on clone maps to RepositoryCredentialsError and sets ERROR_CRED status.
+
+        test_repository.py patches fetch() with crafted stderr strings to verify the
```
ajtmccarty: does the docstring need to reference a different test file? I'd think it can just reference what this test is actually doing

ajtmccarty (approved):

backend/tests/integration/git/test_git_live_remote.py
ogenstad: Yeah good point. I've updated this.

PR #8918, opened 2026-04-15
Allow builtin Jinja2 filter to run on workers

gmazoyer:
## Why & What

In a previous commit of the Python SDK we were preventing the use of builtin Jinja2 filters to run on worker. This was a side effect of a bug fix. These filters were previously marked as untrusted, though we were not properly validating their use before. Since they are part of Jinja2 itself, we dediced to allow them to be used in both the worker and the local context.

---
## Summary by cubic
Re-enabled built-in `Jinja2` filters on workers and locally via `python_sdk` update; tests assert `upper` is allowed and `fqdn_to_ip` is blocked.

<sup>Written for commit 3f8a587c79ca47d7bfb1873445e06f9aa274ce8f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/component/git/test_artifact_composition.py
+++ b/backend/tests/component/git/test_artifact_composition.py
@@ -31,9 +31,9 @@ async def git_repo_filter_tests(git_upstream_repo_02: dict[str, str | Path], git
 
     templates = {
         # Untrusted filter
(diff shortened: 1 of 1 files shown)


polmichel (approved):
PR seems consistent with https://github.com/opsmill/infrahub-sdk-python/pull/943 PR.
LGTM besides the environment variables but should be taken care of in this PR https://github.com/opsmill/infrahub/pull/8923

Issue #8857, opened 2026-04-09
bug: Copy ID and HFID does not work


jjeff07:
### Component

Frontend UI

### Infrahub version

1.8.4

### Current Behavior

<img width="228" height="140" alt="Image" src="https://github.com/user-attachments/assets/b8f325ae-4082-439c-ba8a-979c50655890" />

These buttons do not work and does not copy anything.

### Expected Behavior

ID and HFID should be copied to clipboard.

### Steps to Reproduce

Chrome: Version 146.0.7680.178 (Official Build) (64-bit)

jjeff07:
Copy Token also does not work just an FYI

<img width="449" height="250" alt="Image" src="https://github.com/user-attachments/assets/ba14b7f9-e6f7-473c-861f-c5a9fe1efe99" />

PR #8920, opened 2026-04-15
Fix broken :* permission patterns across all bug-agent workflows

polmichel:
### Why

Analyst worked alone for the first time: https://github.com/opsmill/infrahub/issues/8857.
This is not the case of the test-writer agent, which still has denied commands https://github.com/opsmill/infrahub/actions/runs/24419907267/job/71341212296.

#### Wrong delimiter

The delimiter `:*` in permission patterns across all bug-agent workflow files does not work as I was expecting https://code.claude.com/docs/en/permissions#wildcard-patterns. It silently fails to match some complex commands, meaning **Allow rules** like `Bash(git diff :*)` never matched, causing commands to fall through to manual approval

Affected files:
- `.github/workflows/bug-agent-test.yml`
- `.github/workflows/bug-agent-fix.yml`
- `.github/workflows/bug-agent-review.yml`
- `.github/workflows/bug-agent-analyst.yml`

Example commands that were being blocked despite being in the allow list:
```
npx --prefix .../frontend/app biome check --write ...
npm --prefix .../frontend/app run test -- --run ...
gh pr create --title "..." --base stable --draft --body-file ...
mkdir -p .../infrahub/.agent-tmp
git add frontend/app/src/shared/hooks/useCopyToClipboard.test.ts
```

#### Missing commands

Commands were missing (`mkdir`, subcommands of `npx` or `npm`).

#### Reviewer trigger

Reviewer trigger is only related to Claude bot, but a member or collaborator could need it through a local execution of the pipeline.

#### Wrong delimiter

Replace `:*` with `*` (space + wildcard) in all allow/deny permission patterns across the 4 workflow files. Also expanded allowed commands for test-writer to support `npm --prefix`/`npx --prefix` workflows, and updated test assertions accordingly.

#### Missing commands

Commands added.

#### Reviewer trigger

Trigger now accepts Claude bot, Members & reviewers


polmichel:
Local testing mimicing the permissions worked:  https://github.com/opsmill/infrahub/pull/8921

ogenstad (approved):
One thing to note is that the frontend replaced npm with pnpm in `develop`, though we also have npm for the docs folder. We can probably allow both options for now and then clean up after 1.9 has been released and the stable branch is inline with the release branch. Could be that we want to update the commands used for tests when this gets merged back into `release-1.9`, but we should verify that things work in stable first just to ensure that the process is correct.

polmichel:
> One thing to note is that the frontend replaced npm with pnpm in `develop`, though we also have npm for the docs folder. We can probably allow both options for now and then clean up after 1.9 has been released and the stable branch is inline with the release branch. Could be that we want to update the commands used for tests when this gets merged back into `release-1.9`, but we should verify that things work in stable first just to ensure that the process is correct.

Thank you for the feedback, this is indeed something I had in mind regarding `pnpm` commands.

PR #8931, opened 2026-04-15
Fix git sync after pod restart/storage recovery on workers

### LINKED ISSUE #8930: bug: Git sync recovery after missing local directory never imports objects or updates commit hash
### Component

Git Integration

### Infrahub version

1.8.4

### Current Behavior

When a task worker's local git directory is missing (e.g., pod restart with ephemeral storage, node eviction, or `git-repository-add` flow never ran), the `init()` method in `InfrahubRepositoryIntegrator` catches `RepositoryInvalidFileSystemError` and calls `create_locally(update_commit_value=False)`.

This performs a full `Repo.clone_from()`, so the local branch HEAD immediately matches the remote. When `sync()` then calls `compare_local_remote()`, it sees local == remote and returns no new/updated branches. As a result:

- `import_objects_from_files()` is **never called** — schemas, transforms, GraphQL queries, and other objects from the repo are never loaded
- The database commit value is **never updated** — the UI shows no commit hash for the repository
- The `sync_remote_repositories` fallback path (which does call `new()` + `import_objects_from_files()`) never triggers because `init()` catches the error internally and succeeds silently

This creates a permanent broken state where periodic sync runs every minute, clones the repo, fetches, finds no diff, and completes "successfully" — but the repository is effectively non-functional.

### Expected Behavior

When `init()` recovers from a missing local directory by re-cloning, it should also:

1. Import objects from files (schemas, transforms, queries, etc.)
2. Update the commit value in the database

Alternatively, `init()` should signal to the caller that a re-initialization occurred so the existing fallback path in `sync_remote_repositories` can trigger the full `new()` + `import_objects_from_files()` flow.

### Steps to Reproduce

1. Deploy Infrahub on Kubernetes via Helm with 2+ task worker replicas
2. Add a git repository (ReadOnly or ReadWrite) containing schemas
3. Wait for the initial `git-repository-add-*` flow to complete on one worker
4. Delete the local git directory on a task worker pod (simulating pod restart with ephemeral storage):
   ```bash
   kubectl exec -n infrahub <task-worker-pod> -- rm -rf /opt/infrahub/git/<repo-uuid>
   ```
5. Wait for the next `git_repositories_sync` cycle (~1 minute)
6. Observe in the worker logs:
   - `Initialized the local directory for <repo> because it was missing.`
   - `Fetching the latest updates from remote origin.`
   - `Finished in state Completed()`
   - But NO `New commit detected` and NO schema/object import
7. The database commit value remains stale or null, and the worker never processes objects from the repo

The issue also manifests when the `git-repository-add-*` flow never runs at all (see related issue about startup timing), in which case NO worker ever imports objects.

### Additional Information

In the customer's environment, even after manually forcing the initial import via GraphQL mutation, subsequent new commits pushed to the remote were not picked up. This is explained by the same root cause: when a worker without a local clone runs the sync, `init()` calls `create_locally()` which clones at the **latest** remote commit. Since the clone is already at the newest commit, `compare_local_remote()` sees no diff and returns immediately — the new commit is never processed.

Whether a new commit is detected depends on timing: if the init-clone happens to grab an older commit and then `fetch()` picks up the new one, the diff IS detected. But this is a race condition, not reliable behavior. In local reproduction (kind cluster, 2 task workers, RW `CoreRepository`), we confirmed:

- Worker A (which had the initial clone) did NOT process the new commit (it was not assigned the sync flow run)
- Worker B (empty git dir) ran the sync, cloned at the OLD commit, fetched the NEW commit, and detected the diff — but only because the push happened between its clone and fetch
- If the push had occurred before the sync cycle, Worker B's clone would have been at the latest commit and the diff would have been missed

## Relevant Code

- `backend/infrahub/git/integrator.py` — `InfrahubRepositoryIntegrator.init()` catches `RepositoryInvalidFileSystemError` and calls `create_locally(update_commit_value=False)` without importing
- `backend/infrahub/git/base.py` — `create_locally()` does a full clone; `compare_local_remote()` compares local vs remote branch commits
- `backend/infrahub/git/repository.py` — `sync()` relies on `compare_local_remote()` returning updated branches to trigger import
- The fallback in `sync_remote_repositories` only triggers on `RepositoryError`, which `init()` never raises after successful recovery

FragmentedPacket:
# Why

Closes #8930 

## What changed

- Pass through db commits to ensure when a repo is initially synced (in this case, the storage was lost on a pod on the task worker in a k8s deployment), it will re-sync, and also compare the local ref to the Db commit ref value and update accordingly.

## How to review

This was vibe coded with Claude while I was working on customer issues so it may not be optimal or the proper fix, but it exists for a review.

## How to test

This one is somewhat difficult as it requires a repository and potentially a k8s deployment to restart pods with no persistence or override the volumes used for workers in the docker compose route.

## Impact & rollout

- **Backward compatibility:** 
- **Performance:** 
- **Config/env changes:** 
- **Deployment notes:** 

## Checklist

- [ ] Tests added/updated

---
## Summary by cubic
Fixes git sync after worker pod restart or storage recovery by detecting stale DB commit values and re-importing branches to align the DB with local git HEAD. Prevents repos from staying out of sync after a fresh re-clone.

- **Bug Fixes**
  - `InfrahubRepository.sync()` now accepts `db_commits` and detects branches where the DB commit differs from local HEAD (when local==remote). It updates the DB commit and re-imports objects for those branches.
  - `sync_remote_repositories()` now passes the per-branch commit map to `repo.sync()` after the normal sync to enable stale-DB detection.
  - Added unit tests for stale DB commit handling, match/no-op paths, and preserving behavior when `db_commits` is not provided.

<sup>Written for commit c1094a4f9138fc6bbb2d81d7c3cb31c49667b5b5. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/git/repository.py
+++ b/backend/infrahub/git/repository.py
@@ -62,10 +62,12 @@ class InfrahubRepository(InfrahubRepositoryIntegrator):
 
         return response
 
-    async def sync(self, staging_branch: str | None = None) -> None:
+    async def sync(self, staging_branch: str | None = None, db_commits: dict[str, str] | None = None) -> None:
         """Synchronize the repository with its remote origin and with the database.
 
         By default the sync will focus only on the branches pulled from origin that have some differences with the local one.
+        If db_commits is provided, also detect branches where the local git HEAD differs from the DB commit value
+        and re-import objects for those branches (handles stale DB state after a re-clone).
         """
 
         log.info("Starting the synchronization.", repository=self.name)
@@ -74,7 +76,26 @@ class InfrahubRepository(InfrahubRepositoryIntegrator):
 
         new_branches, updated_branches = await self.compare_local_remote()
 
-        if not new_branches and not updated_branches:
+        # Detect branches where the DB commit is stale compared to local git HEAD
+        stale_db_branches: list[str] = []
+        if db_commits and self.internal_status == RepositoryInternalStatus.ACTIVE.value:
+            already_handled = set(new_branches) | set(updated_branches)
+            local_branches = self.get_branches_from_local(include_worktree=False)
+            for branch_name, branch_data in local_branches.items():
+                if branch_name in already_handled:
+                    continue
+                db_commit = db_commits.get(self._get_mapped_target_branch(branch_name=branch_name))
+                if db_commit is not None and str(branch_data.commit) != db_commit:
+                    log.info(
+                        "Stale DB commit detected",
+                        repository=self.name,
+                        branch=branch_name,
+                        local_commit=str(branch_data.commit),
+                        db_commit=db_commit,
+                    )
+                    stale_db_branches.append(branch_name)
+
+        if not new_branches and not updated_branches and not stale_db_branches:
             return
 
         log.debug(f"New Branches {new_branches}, Updated Branches {updated_branches}", repository=self.name)
@@ -119,6 +140,13 @@ class InfrahubRepository(InfrahubRepositoryIntegrator):
                         repository=self.name,
                         branch=branch_name,
                     )
+
+            for branch_name in stale_db_branches:
+                infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
+                commit = self.get_commit_value(branch_name=branch_name, remote=False)
+                self.create_commit_worktree(commit=commit)
+                await self.update_commit_value(branch_name=infrahub_branch, commit=commit)
+                await self.import_objects_from_files(infrahub_branch_name=infrahub_branch, commit=commit)
 
         await self._sync_staging(staging_branch=staging_branch, updated_branches=updated_branches)
 
--- a/backend/infrahub/git/tasks.py
+++ b/backend/infrahub/git/tasks.py
@@ -260,6 +260,12 @@ async def sync_remote_repositories() -> None:
                     staging_branch=staging_branch,
                     infrahub_branch=infrahub_branch,
                 )
+
+                # After sync, check if DB commits are stale (e.g., after a fresh re-clone).
+                # The subflow sync only compares local git vs remote git — it doesn't
+                # detect drift between DB and local git HEAD.
+                if not init_failed:
+                    await repo.sync(db_commits=repository_data.branches)
                 # Tell workers to fetch to stay in sync
                 message = messages.RefreshGitFetch(
                     meta=Meta(initiator_id=WORKER_IDENTITY, request_id=get_log_data().get("request_id", "")),
--- a/backend/tests/unit/git/test_sync_stale_db_commit.py
+++ b/backend/tests/unit/git/test_sync_stale_db_commit.py
@@ -1,0 +1,138 @@
+from __future__ import annotations
+
+from pathlib import Path
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from infrahub_sdk import Config, InfrahubClient
+from infrahub_sdk.uuidt import UUIDT
+
+from infrahub import config
+from infrahub.core.constants import RepositoryInternalStatus
+from infrahub.core.registry import registry
+from infrahub.git import InfrahubRepository
+from tests.helpers.file_repo import FileRepo
+from tests.helpers.test_client import dummy_async_request
+
+
+@pytest.fixture
+def git_sources_dir(tmp_path: Path) -> Path:
+    source_dir = tmp_path / "sources"
+    source_dir.mkdir()
+    return source_dir
+
+
+@pytest.fixture
+def git_repos_dir(tmp_path: Path) -> Path:
+    repos_dir = tmp_path / "repositories"
+    repos_dir.mkdir()
+    config.SETTINGS.git.repositories_directory = str(repos_dir)
+    return repos_dir
+
+
+@pytest.fixture
+def setup_registry() -> None:
+    registry._default_branch = "main"
+
+
+@pytest.fixture
+def upstream_repo(git_sources_dir: Path) -> FileRepo:
+    return FileRepo(name="car-dealership", sources_directory=git_sources_dir)
+
+
+@pytest.fixture
+async def infrahub_repo(
+    upstream_repo: FileRepo,
+    git_repos_dir: Path,
+    setup_registry: None,
+) -> InfrahubRepository:
+    repo = await InfrahubRepository.new(
+        id=UUIDT.new(),
+        name="car-dealership",
+        location=upstream_repo.path,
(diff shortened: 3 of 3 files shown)


PhillSimonds:
Hey all! I did some testing with this patch. Here were my results:

## Reproduction and verification on local kind cluster

I reproduced the bug and verified this fix on a 3-task-worker HA deployment using `infrahub-enterprise:1.8.3` via the `oci://registry.opsmill.io/opsmill/chart/infrahub-enterprise` chart (version `4.6.0-small`) on a local kind cluster. Deployment values adapted from a customer's integration environment (external Neo4j/RabbitMQ/Redis-Sentinel replaced with bundled equivalents).

### Reproduction — vanilla 1.8.3 (before the fix)

1. Add a RW `CoreRepository` with customer schemas, let it sync successfully.
2. Simulate pod restart / storage loss: `kubectl exec -n infrahub <worker> -- rm -rf /opt/infrahub/git/<repo-uuid>` on all 3 task workers.
3. Rewind the DB commit value to an old hash via `CoreRepositoryUpdate` mutation (this mirrors what happens when the initial add flow stores a commit value but a later event drifts the DB).
4. Wait for periodic sync.

Result:

| Location | Value |
|---|---|
| All 3 worker HEADs | `29cbf75ff23e` (latest via init-recovery clone) |
| DB `commit` | `06cb5bcd1a7a` (stale) |
| Remote HEAD | `29cbf75ff23e` |
| DB `sync_status` | `in-sync` ← misleading |

Vanilla `sync()` returned early because `compare_local_remote()` saw local == remote and produced empty lists — nothing triggered `import_objects_from_files()` or `update_commit_value()`. The DB stayed wrong and the system reported itself as healthy.

### What we did not reproduce

Observed behavior in a customer environment showed that **new pushes to the remote also fail to rescue** the stuck state. We did not reproduce that specific variant in our kind setup — in our run, a subsequent push did update the DB because our workers happened to still hold a local clone at the pre-push commit, so `compare_local_remote()` had a real diff to see.

For the "new pushes also ignored" pattern to happen deterministically, the workers would need to re-clone on (most) sync cycles — so the init-recovery clone always lands at current remote HEAD and `compare_local_remote()` always returns empty. Something in the customer environment (storage churn, worker pod restarts, or sync flow running on different workers each cycle due to the Prefect work-pool's distribution) is plausibly producing that. We haven't pinned down the exact trigger.

The fix in this PR handles both cases the same way: whenever local HEAD and DB commit disagree, re-import regardless of whether local == remote — so it resolves the customer's symptom whether or not we can reproduce that specific failure mode locally.

### Verification — 1.8.3 + this PR

Built a patched image:

```dockerfile
FROM registry.opsmill.io/opsmill/infrahub-enterprise:1.8.3
USER root
COPY pr-8931.patch /tmp/
RUN cd /source/community && patch -p1 < /tmp/pr-8931.patch
USER 1000
```

Loaded into kind (`kind load docker-image`), updated both `infrahub-task-worker` and `infrahub-server` Deployments to use it, and re-ran the same repro.

Result: within one sync cycle (~25 seconds), the DB recovered from `29cbf75ff23e` to `0969db1a7c22`. The new log message fired exactly as expected:

```json
{
  "repository": "*****-test-repo",
  "branch": "main",
  "local_commit": "0969db1a7c22d859025dbe1bd14f97ce8aa95e95",
  "db_commit":    "29cbf75ff23ec572e7bb3a8192f88a6193df0e0c",
  "event": "Stale DB commit detected",
  "level": "info"
}
```

Followed by the expected `update_commit_value()` → `import_objects_from_files()` path, ending with DB and local HEAD aligned.

### Scope notes

- Verified on enterprise chart `4.6.0-small` + enterprise image `1.8.3` with `InfrahubEntWorkerAsync` worker type.
- Reproduced across 3 task worker replicas to exercise the Prefect work-pool distribution (only one worker runs each `git_repositories_sync` flow run at a time).
- Did **not** reproduce the customer's "initial hash never populates" symptom locally — in our reproduction, the initial `git-repository-add-read-write` flow always succeeded on first try. That symptom is likely tied to their specific environment (external Neo4j/RabbitMQ timing, custom CA bundles, internal GitLab network, or pod restarts mid-flow). This PR does not target that initial-add failure mode — but it is the correct fix for the resulting stuck state, which is what the two manual mutations the customer runs are working around.

LGTM from a behavior-verification standpoint. Happy to rerun if you iterate on the patch.

gmazoyer (commented):
The main thing that bugs me is the double sync. `sync_git_repo_with_origin_and_tag_on_failure` already creates its own `InfrahubRepository.init()` and calls `sync()` inside, which does a full `fetch()` + `compare_local_remote()`. Then the new `repo.sync(db_commits=...)` does all of that again on a different instance. That's two round-trips to the remote for every repo, every minute, even when nothing is wrong. Could we pass `db_commits` through to avoid that?

In the stale branch loop, both `new_branches` and `updated_branches` go through `validate_remote_branch` and wrap processing in try/except so one bad branch doesn't kill the whole sync. The new loop doesn't do either:

```python
for branch_name in stale_db_branches:
    infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
    commit = self.get_commit_value(branch_name=branch_name, remote=False)
    self.create_commit_worktree(commit=commit)  # import_objects_from_files already does this internally
    await self.update_commit_value(branch_name=infrahub_branch, commit=commit)
    await self.import_objects_from_files(infrahub_branch_name=infrahub_branch, commit=commit)
```

Was that on purpose? `create_commit_worktree` is called here but `import_objects_from_files` calls it again so it's probably redundant.

I wonder if there is a simpler approach. The problem is that `init()` silently recovers and nobody knows about it. What if we add a flag on the instance:

```python
# in InfrahubRepositoryBase
reinitialized: bool = Field(
    False,
    description="A re-clone was needed because the local directory because it was missing.",
)

# then in integrator.py, in init() recovery path
except RepositoryInvalidFileSystemError:
    await self.ensure_location_is_defined()
    await self.create_locally(infrahub_branch_name=self.infrahub_branch_name, update_commit_value=False)
    self.reinitialized = True
```

Finally in `sync_remote_repositories`, check the flag and call import, same thing the `init_failed` path already does:

```python
if repo.reinitialized:
    try:
        await repo.import_objects_from_files(
            git_branch_name=registry.default_branch, infrahub_branch_name=infrahub_branch
        )
    except RepositoryError as exc:
        log.info(exc.message)
        continue
```

It avoids touching `sync()`, no extra fetch, no overhead on normal cycles. But maybe I'm missing a case where the `db_commits` approach catches something this doesn't?

For the tests, could we use real git repos like the rest of `test_git_repository.py` does with `FileRepo`? All the mocking makes it hard to tell if the actual integration works.

PR #8933, opened 2026-04-15
fix: make secret_key required in SecuritySettings (closes #8925)

### LINKED ISSUE #8925: security: make INFRAHUB_SECURITY_SECRET_KEY mandatory to avoid JWT mismatch across multi-pod deployments
## Problem

`INFRAHUB_SECURITY_SECRET_KEY` is defined with a `default_factory=generate_uuid`, so when it is not explicitly set each API-server process generates its own fresh UUID at startup:

https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/config.py#L749-L751

That key is used directly to sign and verify JWT access/refresh tokens with HS256:

https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/auth.py#L242
https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/auth.py#L275

### Impact

In any multi-replica deployment (notably k8s with multiple `infrahub-server` pods, but also Docker Compose with a scaled `infrahub-server` service) the current default causes intermittent authentication failures:

- Pod A mints a JWT signed with its locally-generated UUID.
- The next request from the same client lands on pod B through the Service/Ingress.
- Pod B rejects the token as invalid signature (`InvalidSignatureError`), because it signed its own JWTs with a *different* UUID.

From the user's perspective this looks like random \"session expired\" / 401 responses correlated with replica count. There is no log line flagging the misconfiguration — it just silently breaks.

The production-deployment guide already tells users to set a strong value, but this is advisory; the server boots fine without it.

## Proposed fix

Make `INFRAHUB_SECURITY_SECRET_KEY` **mandatory** at the config layer:

1. Drop `default_factory=generate_uuid` from `SecuritySettings.secret_key` in `backend/infrahub/config.py` and require the field (no default), so `Settings()` fails fast with a clear Pydantic validation error if it is unset or empty.

   The bundled `docker-compose.yml`, `development/docker-compose.yml`, `python_testcontainers/*.yml`, and the Helm charts (`helm/charts/infrahub/values.yaml`, `helm/charts/infrahub-enterprise/values.yaml`) already pass an explicit value, so making the field mandatory does not break any shipped install path — it only surfaces a hard error for deployments that were silently relying on the per-process UUID default.

2. Add a changelog fragment flagging this as a behavior change: the server will now refuse to start if `INFRAHUB_SECURITY_SECRET_KEY` is not provided.

## Acceptance criteria

- [ ] `Settings()` raises a validation error when `INFRAHUB_SECURITY_SECRET_KEY` is unset or empty.
- [ ] Bundled `docker-compose.yml` variants, test containers, and Helm charts continue to boot unchanged (they already set the variable explicitly — no changes to those files).
- [ ] Changelog entry added under \`changelog/\` noting the behavior change.

## Out of scope

- Changing any hardcoded defaults that currently ship in `docker-compose.yml` / Helm values — those stay as-is to avoid breaking existing installs.
- Rotating or key-versioning the JWT signing key (separate feature).
- Switching away from HS256 to asymmetric signing (separate discussion).

polmichel:
### Random 401 errors

`SecuritySettings.secret_key` used `default_factory=generate_uuid`, so each process/pod generated a unique JWT signing key at startup. In multi-replica deployments, tokens signed by one pod failed verification on others, causing random 401 errors with no indication of misconfiguration.

Goal: make `secret_key` a required field so the server fails fast at startup if `INFRAHUB_SECURITY_SECRET_KEY` is not explicitly set.

### Fix consequence

Making the key mandatory is only useful if a missing key actually stops the server. In the first cut of this PR it didn't: to keep `infrahub.config` importable by tooling that sets env vars post-import. The factory is only called during `Settings()` instantiation, which happened inside FastAPI's lifespan and middleware. Result: gunicorn's master stayed up and every request returned HTTP 500.

Closes #8925

## What changed

- `secret_key` has not default value anymore.
- `infrahub.server` now exposes `create_app() -> FastAPI` instead of a module-level `app`.
- Gunicorn fail-fast without `preload_app`. A `gunicorn_config.py` `on_starting` hook validates settings in the master.
-  CLI and deployment configs updated to the new targets.

### Fail / Start behviors

```bash
uv run pytest backend/tests/unit/config/ -v

# Fail-fast (must exit non-zero before binding)
env -u INFRAHUB_SECURITY_SECRET_KEY uv run gunicorn --config backend/infrahub/serve/gunicorn_config.py -w 1 infrahub.serve.app:app
env -u INFRAHUB_SECURITY_SECRET_KEY uv run uvicorn --factory infrahub.server:create_app
env -u INFRAHUB_SECURITY_SECRET_KEY uv run infrahub server start        # CLI wraps uvicorn with factory=True

# Positive: CLI commands that build the app should still work when the key is set or without
INFRAHUB_SECURITY_SECRET_KEY=abc uv run infrahub dev export-json-schema --out /tmp/openapi.json && head -c 100 /tmp/openapi.json
uv run infrahub dev export-json-schema --out /tmp/openapi.json && head -c 100 /tmp/openapi.json


# Full stack (all services: server, task-manager, task-worker, etc.)
uv run invoke dev.destroy && uv run invoke dev.build && uv run invoke dev.start --wait
```
### Impacted invoke commands

```bash
  uv run invoke backend.generate && uv run invoke backend.validate-generated
```

Expected failure on failing cases: 
```
Configuration not valid, found 1 error(s)
secret_key | Value error, secret_key must be provided via config or INFRAHUB_SECURITY_SECRET_KEY environment variable (value_error)
```

 ## Impact & rollout
  - `from infrahub.server import app` no longer exists. Call `create_app()` or target `infrahub.serve.app:app` (gunicorn) / `--factory infrahub.server:create_app` (uvicorn).
  - `INFRAHUB_SECURITY_SECRET_KEY` must be set for the API server and for Prefect in distributed mode. Docker-compose dev setups unaffected (default provided in compose).

### Breaking change
infrahub.server:app → infrahub.serve.app:app

#### Example
  Before thisPR :
```
gunicorn --config backend/infrahub/serve/gunicorn_config.py -w ${WEB_CONCURRENCY:-4}  --logger-class infrahub.serve.log.GunicornLogger infrahub.server:app
```

After this PR:
```
gunicorn --config backend/infrahub/serve/gunicorn_config.py  -w ${WEB_CONCURRENCY:-4}  --logger-class infrahub.serve.log.GunicornLogger  infrahub.serve.app:app
````
## Checklist

- [x] Tests added/updated

### CODE REVIEW

**Comment by polmichel** [comment]
## APPROVED

### A. Correctness

The fix directly addresses the analyst's root cause: `default_factory=generate_uuid` removed, `secret_key` is now a required `str` field with no default. Pydantic enforces presence automatically — no custom validator needed, exactly as the analyst prescribed.

Verified that all existing deployment paths supply the key:
- `docker-compose.yml` line 270, 301: services override with `${INFRAHUB_SECURITY_SECRET_KEY:-327f747f-...}`
- `backend/tests/conftest.py` line 547: test infra sets `config.SETTINGS.security.secret_key = "4e26b3d9-..."`
- The base config anchor (line 108) passes the env var through — services that need it get it from the override lines

The unused `generate_uuid` import is correctly removed.

No edge cases missed. An empty string would technically satisfy the `str` type, but the analyst explicitly said not to add validators for empty/UUID-shaped keys — Pydantic presence enforcement is sufficient.

### B. Code quality

Minimal and clean: 3 lines removed from production code (1 import, 2 field definition lines replaced by 1). Follows the analyst's "Do NOT" guardrails (no validator, no fallback default, no docker-compose changes). No unnecessary refactoring.

### C. Documentation alignment

- Changelog uses `changed` type (not `fixed`), correctly reflecting this is a deliberate behavior change per the analyst's notes.
- Changelog message is user-facing and clear: "The server now requires INFRAHUB_SECURITY_SECRET_KEY to be explicitly set and will refuse to start without it."
- PR body correctly flags this as a breaking change for bare-metal dev setups without the env var.

### D. Test quality

`test_security_settings_requires_secret_key` properly clears `INFRAHUB_SECURITY_SECRET_KEY` from the environment before asserting `ValidationError`. The `patch.dict(os.environ, env_clean, clear=True)` is necessary because `SecuritySettings` is a `BaseSettings` subclass that reads env vars — without it, the test would pass spuriously in environments where the var is set.

The test asserts behavior (required field raises on missing value), not implementation (doesn't check for the absence of `default_factory`). Correctly uses `match="secret_key"` to validate the error message.

### Recommended next steps

1. Merge.

File: backend/infrahub/config.py [hunk 1]
```diff
@@ -1034,7 +1031,7 @@ class Settings(BaseSettings):
     analytics: AnalyticsSettings = AnalyticsSettings()
     initial: InitialSettings = InitialSettings()
     policy: PolicySettings = PolicySettings()
-    security: SecuritySettings = SecuritySettings()
```
fatih-acar, line 1037: Why this change?

polmichel: fyi I had not yet reviewed the code, I am about do it.
This is related to a `ty` typing check, I will replace this with a comment to enable the check

polmichel:
```
error[missing-argument]: No argument provided for required parameter `secret_key`
    --> backend/infrahub/config.py:1034:34
     |
1032 |     initial: InitialSettings = InitialSettings()
1033 |     policy: PolicySettings = PolicySettings()
1034 |     security: SecuritySettings = SecuritySettings()
     |                                  ^^^^^^^^^^^^^^^^^^
1035 |     storage: StorageSettings = StorageSettings()
1036 |     trace: TraceSettings = TraceSettings()
     |
info: rule `missing-argument` is enabled by default
```

polmichel: Actually the bot is right about this beside the `ty` check (which is also legitimate): 
It allows to defer the evaluation of `SecuritySettings` default value at `Settings` instantiation time and not at direct/indirect `Settings` class import time. 
Thanks to this, the `SecuritySettings()` call (which will fail if no environment variable is set and no direct parameter is passed for _secret_key_) happens at object instantiation, not before.

I have extracted a dedicated unit test file in order to explicitly state the expected behavior: `backend/tests/unit/config/test_security_settings.py`


#### Sources
- https://docs.python.org/3/library/dataclasses.html#dataclasses.field -> _default_factory: If provided, it must be a zero-argument callable that will be called when a default value is needed for this field_
- https://pydantic.dev/docs/validation/latest/api/pydantic/fields/#pydantic.fields.Field(default_factory) -> _To do so, the [Field()](https://pydantic.dev/docs/validation/latest/api/pydantic/fields/#pydantic.fields.Field) function is used a lot, and behaves the same way as the standard library [field()](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) function for dataclasses – by assigning to the annotated attribute:_

ogenstad: @polmichel the problem is that the implementation of the `Settings` class isn't done in a sensible way. 

The correct approach would be something like this:

```python
class Settings(BaseSettings):
    """Main Settings Class for the project."""

    main: MainSettings = Field(default_factory=MainSettings)
    api: ApiSettings = Field(default_factory=ApiSettings)
```

I.e. we only want to evaluate these during the creation of a settings object and not when the application loads. I think it could potentially have an impact on how the API worker behaves when started with gunicorn and a bunch of workers. Currently the app would fail to load if anything was wrong. But I think we could run into issues with the above change combined with an invalid config. This could be verified by removing the required variable and starting with:

```bash
gunicorn --config backend/infrahub/serve/gunicorn_config.py --logger-class infrahub.serve.log.GunicornLogger infrahub.server:app
```

polmichel: This is what has been done already regarding `SecuritySettings` class attribute. Are you saying that I should add the same behavior for all other settings attributes of `infrahub.config.Settings` in this PR?

I will test the workers' behavior regarding this change

backend/infrahub/serve/app.py
```diff
@@ -0,0 +1,14 @@
...
+`sh -c "$COMMAND"` (as in `development/docker-compose.yml`). This thin module
+exposes `app` so gunicorn can load it with plain `module:attribute` syntax.
+
+Uvicorn consumers should target `infrahub.server:create_app` with `--factory`
+instead of importing this module.
+"""
+
+from infrahub.server import create_app
+
+app = create_app()
```
polmichel, line 14: This is a breaking change as this changes the module to call

`infrahub.server:app` → `infrahub.serve.app:app`

polmichel:
We discussed this change with the team. This may cause security issues regarding the key on worker servers, as code can be imported and executed on these servers.
We would therefore need a different system for the settings on worker servers. Some of these settings may need to be split between workers and API servers, while others should be kept. This would need an additional design work before continuing on this path.
cc @ogenstad @fatih-acar

PR #8934, opened 2026-04-15
updates to SchemaBranch.process to keep uniqueness updates idempotent

### LINKED ISSUE #8897: bug: schema hash drifts across workers when `kind: NumberPool` attribute coexists with a `generate_template: true` node in the same schema load
### Component

_No response_

### Infrahub version

1.8.3

### Current Behavior

After loading a schema bundle into a fresh Infrahub 1.8.3 stack, `InfrahubStatus` reports `schema_hash_synced: false`, with one `task-worker` replica reporting a different `schema_hash` than all other workers. The affected worker is **not aware it has drifted** — its own structured log continues to report `"Branch schema hash is the same"` with the consensus hash.

The drift reproduces minimally (69-line schema, below) when **both** of the following are present in the same load:

1. Any concrete node with `generate_template: true`
2. Any attribute with `kind: NumberPool` anywhere in the bundle — *even on a generic that the `generate_template: true` node does not inherit from*

Stripping either condition independently clears the drift:

- Swap `kind: NumberPool` → `kind: Text` on the offending attribute → synced ✓
- Remove `generate_template: true` from the concrete node → synced ✓

Only the drifted worker logs the following during the post-load schema reconcile flow (no other worker logs it):

```
/.venv/lib/python3.13/site-packages/fastapi/encoders.py:333:
  RuntimeWarning: coroutine 'Redis.execute_command' was never awaited
  data = dict(obj)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

This is the same warning previously reported in #4879 and dismissed as cosmetic; the present findings suggest it is load-bearing — it coincides exactly with the worker that ends up with a divergent `schema_hash`.

### Expected Behavior

After a schema load completes and the schema-update flows report `"Schema converged"` on all workers, `InfrahubStatus.summary.schema_hash_synced` should be `true` and every worker should report the same `schema_hash`, regardless of whether the loaded schema contains `NumberPool` attributes alongside nodes with `generate_template: true`.

### Steps to Reproduce

1. Spin up Infrahub 1.8.3 from scratch (no backup restore required):
   ```
   VERSION=1.8.3 docker compose up -d
   ```
   (standard compose from https://infrahub.opsmill.io, 2× task-worker replicas)

2. Confirm an empty baseline is synced:
   ```graphql
   query {
     InfrahubStatus {
       summary { schema_hash_synced }
       workers { edges { node { id active schema_hash } } }
     }
   }
   ```
   → expected: `schema_hash_synced: true`, all workers on the same hash.

3. Load the minimal drifting schema bundle below (attached as `firewall_MINIMAL_DRIFTS.yml`). To satisfy cross-file dependencies the full invocation used is:
   ```
   infrahubctl schema load \
       schemas/base/ \
       schemas/extensions/routing/ \
       schemas/extensions/extend_builtin.yml \
       schemas/extensions/location/ \
       firewall_MINIMAL_DRIFTS.yml
   ```
   For a yet smaller surface, the trigger also reproduces with *just* a generic holding a `NumberPool` attribute plus *any* concrete node with `generate_template: true` — the surrounding base/routing/location schemas are not part of the trigger (they were needed only to satisfy existing cross-file dependencies from the source repo).

4. Wait ≥30s for the post-load convergence flows to finish (watch for `"Schema converged after X seconds"` in the `task-worker` and `infrahub-server` logs).

5. Re-run the `InfrahubStatus` query from step 2 → **`schema_hash_synced: false`**, with one `task-worker` reporting a different `schema_hash` than the other five reporting workers (1× task-worker-2, 1× infrahub-server, plus 3 sub-workers inside the server container, plus the affected task-worker-1).

6. `docker logs <drifted-task-worker>` around the time of the reconcile flow contains the `coroutine 'Redis.execute_command' was never awaited` RuntimeWarning; logs from all other workers do not.

## Minimal drifting schema (`firewall_MINIMAL_DRIFTS.yml`, 69 lines)

```yaml
version: '1.0'
generics:
- name: GenericDevice
  namespace: Firewall
  description: Generic Firewall object.
  include_in_menu: false
  order_by:
  - name__value
  display_label: '{{ name__value }}'
  human_friendly_id:
  - name__value
  attributes:
  - name: name
    kind: Text
  relationships: []
- name: GenericRules
  namespace: Firewall
  description: Generic Generic Rules object.
  human_friendly_id:
  - rule_id__value
  display_label: '{{ rule_id__value }}'
  include_in_menu: false
  order_by:
  - rule_id__value
  attributes:
  - name: rule_id
    kind: NumberPool              # ← trigger #1
    read_only: true
    branch: agnostic
  relationships: []
nodes:
- name: Firewall
  namespace: Dcim
  label: Firewall
  include_in_menu: false
  generate_template: true         # ← trigger #2
  inherit_from:
  - FirewallGenericDevice
  - DcimGenericDevice
  - DcimPhysicalDevice
  attributes:
  - name: status
    kind: Dropdown
    optional: true
    order_weight: 1100
    choices:
    - {name: active, label: Active, color: '#7fbf7f'}
    - {name: provisioning, label: Provisioning, color: '#ffff7f'}
    - {name: maintenance, label: Maintenance, color: '#ffd27f'}
    - {name: drained, label: Drained, color: '#bfbfbf'}
```

Note: `DcimFirewall` inherits `FirewallGenericDevice` only; it does **not** inherit `FirewallGenericRules`. The `NumberPool` attribute on `FirewallGenericRules` is sufficient to trigger drift by its mere presence in the bundle being processed.

## Cleanup variant that is CLEAN (single-attribute diff)

Identical schema with `FirewallGenericRules.rule_id.kind: NumberPool` swapped to `kind: Text`. Synced after load. This is the single-byte-level delta that flips the behavior.

## Environment
- Infrahub server image: `registry.opsmill.io/opsmill/infrahub:1.8.3`
- Compose stack: https://infrahub.opsmill.io (fetched 2026-04-13), pinned via `VERSION=1.8.3`
- Stack: 2× task-worker, 1× infrahub-server (4 internal sub-workers visible in `InfrahubStatus`), 1× task-manager, neo4j 2025.10.1, redis 8.4.0, rabbitmq 4.2.1, postgres 18
- No backup restore required — reproduces on a fresh `docker compose up -v` stack
- Discovered while standing up a local 1.8.3 instance to investigate a production drift symptom

## Bisection evidence
11 full wipe-and-load cycles were run to isolate the trigger. Summary:

| Cycle | Variant | `generate_template` | `kind: NumberPool` | Drift? |
|---|---|---|---|---|
| A | base + routing | — | — | ✓ synced |
| B | + ext/location + extend_builtin | — | — | ✓ synced |
| C | full firewall.yml | ✓ | ✓ | ❌ drift |
| D | + lb.yml (no firewall) | ✓ (DcimLoadBalancer) | — | ✓ synced |
| E | firewall with `generate_template: true` stripped from `DcimFirewall` | ✗ | ✓ | ✓ synced |
| H | minimal firewall (`FirewallGenericDevice` + `DcimFirewall` only) | ✓ | — | ✓ synced |
| I | H + zones bundle | ✓ | — | ✓ synced |
| J | H + rules bundle | ✓ | ✓ | ❌ drift |
| K1 | H + `FirewallGenericRules` only (no concrete rule nodes) | ✓ | ✓ | ❌ drift |
| **L** | **K1 with `rule_id` kind changed from `NumberPool` to `Text`** | ✓ | ✗ | ✓ synced |
| M | K1 with `devices` back-ref to `FirewallGenericDevice` stripped | ✓ | ✓ | ❌ drift |

Every cycle included a full `docker compose down -v` wipe (all named volumes destroyed) and a `VERSION=1.8.3 docker compose up -d` restart. Each cycle's pre-load `InfrahubStatus` check was confirmed synced before loading.

`lb.yml` (Cycle D) has `generate_template: true` on two nodes (`DcimLoadBalancer`, `DcimVirtualLoadBalancer`) and does **not** reproduce the drift — because `lb.yml` has no `NumberPool` attributes. This rules out a generic "templates are fragile" explanation and pins the trigger specifically on the `NumberPool × generate_template` interaction.

## Worker self-reporting vs. InfrahubStatus divergence

The drifted worker's own periodic log entries continue to report `hash=<consensus-hash>` (same as all other workers) via the `"Branch schema hash is the same, no need to refresh the GraphQL schema within the registry"` log message, even while `InfrahubStatus` reports that worker with a *different* `schema_hash`. This is itself a diagnostic quirk: the worker has no way to detect that it's drifted by inspecting its own state. External observation via `InfrahubStatus` (or the `/api/schema` REST endpoint for the consensus value) is the only way to spot the condition.

## Hypothesis

The template-generation code path registers `NumberPool` resource state via a Redis call that is being invoked without `await` (fire-and-forget coroutine). On the worker that hits this race, the coroutine is garbage-collected before it executes, so the Redis state-update step is silently skipped on that worker, and its in-memory schema state diverges from the reconciled database state. Its subsequent self-check compares its own stale copy to itself and reports "in sync". Other workers (including sub-workers inside the same `infrahub-server` container) don't hit the same scheduler race and converge normally. The `RuntimeWarning: coroutine 'Redis.execute_command' was never awaited` that appears in exactly the drifted worker's logs is consistent with this mechanism.

This is a hypothesis based on external observation, not a read of the actual code path — treat as a starting lead.

## Relationship to prior reports

- **#7157 "bug: Templates might cause unpredictable schema hashes"** (closed 2025-10-17, v1.4.4) — symptom description is identical. Fixed by PR #8470 "make SchemaBranch.process() idempotent" (merged 2026-02-25), which is included in 1.8.3. That fix targeted idempotency of `SchemaBranch.process()` for `object_template`/`profiles` relationship shaping and hierarchy edge nodes — it does not appear to cover NumberPool interaction. Either this is a regression of #7157 (new surface) or a distinct root cause with the same symptom.
- **#4879 "bug: task-worker log contains RuntimeWarning for every completed prefect flow"** (open, v1.0.1) — the `coroutine 'Redis.execute_command' was never awaited` warning was characterized there as cosmetic. The present finding — that this warning appears on exactly the task-worker that ends up with a divergent `schema_hash`, and on no other worker — suggests it may be load-bearing rather than cosmetic in at least some schema-reconcile flows. Worth re-examining #4879 in light of this.
- **#5178 "task: Create a test to validate that the schema is in sync after initial creation"** (open) — the test coverage this is asking for would have caught this regression.

## Artifacts

A minimal reproduction bundle is available (zip, 11 KB) containing:
- `firewall_MINIMAL_DRIFTS.yml` — the 69-line triggering schema
- `firewall_MINIMAL_TEXT_CLEAN.yml` — same schema with `NumberPool` → `Text`, drift-free
- `InfrahubStatus_current.json` — the divergent-hash snapshot from the most recent minimal-cycle repro
- `task-worker-1.log` — full log from the drifted worker, including the `Redis.execute_command` RuntimeWarning in its native context at the post-load schema reconcile
- `README.md` — summary + hypothesis

Happy to attach this to the issue or share it via whichever channel the backend team prefers.

[infrahub-drift-minimal-20260413-180859.zip](https://github.com/user-attachments/files/26692723/infrahub-drift-minimal-20260413-180859.zip)

ajtmccarty:
# Why

`SchemaBranch.process()` was non-idempotent for any schema where uniqueness state was *derived* (HFID-implied uniqueness, attribute auto-marked unique from `uniqueness_constraints`, label auto-derived from name, etc). Multiple calls to `SchemaBranch.process()` would result in attributes flipping their `unique` properties and attributes being added to / removed from generated Template and Profile schemas

**Goal:** `process()` produces the same `SchemaBranch` no matter how many times it's called against the same input

**Non-goals:** changing inheritance precedence semantics. documentation implies that generic listed first should take precedence over later generics, but that's not actually how it works 
refactoring `process_pre_validation` / `process_post_validation` boundaries beyond what the bug requires

Closes  #8897

### Behavioral changes
- `SchemaBranch.process()` is now idempotent across repeated calls for HFID-driven, uniqueness-driven, label-driven, and weight-driven derivations.
- A schema that previously needed two or three `process()` passes to fully converge (e.g. a generic whose HFID forces `name.unique=True`, then propagates to inheriting nodes) now converges in one pass.
- One existing test assertion in `test_schema_branch_processes_node_template_schema_weight` was updated to reflect the corrected behavior: an attribute referenced by `human_friendly_id` is now correctly excluded from the generated template on the *first* pass (previously it appeared on pass 1 and only disappeared on pass 2).

### Implementation notes
The fix is a localized reordering inside `process_pre_validation`

1. After the early state-cleanup steps and *before* `process_inheritance`, run `process_human_friendly_id` and `sync_uniqueness_constraints_and_unique_attributes` scoped to **generics only** so any HFID-derived uniqueness on the generic is finalized before `process_inheritance` copies it down to nodes.
2. After `process_inheritance` / `process_hierarchy` / `process_branch_support`, re-run both unscoped to handle nodes (which may now have inherited HFIDs and constraints), plus add `process_labels()` and `generate_weight()` here so template/profile generators see finalized labels and weights.
3. Add a `kinds` parameter to both functions to allow scoping, and a `raise_parsing_errors=False` flag to `process_human_friendly_id` so the early generic-only call doesn't blow up on transient invalid HFID paths (which `validate_human_friendly_id` will surface later as a structured `ValueError`).

### What stayed the same
- `process_post_validation` is unchanged.
- All existing call sites of `process_human_friendly_id` and `sync_uniqueness_constraints_and_unique_attributes` continue to work — both new parameters default to existing behavior.
- No GraphQL schema diff (`schema/schema.graphql` is unchanged after `uv run invoke schema.generate-graphqlschema`).
- No DB / API / migration changes.

### Suggested review order
1. `backend/infrahub/core/schema/schema_branch.py` — small, targeted (~36 lines).
2. `backend/tests/component/core/schema/schema_branch/test_uniqueness_propagation.py` — new file, 9 component tests + a shared `_validate_process_idempotent` helper that checks hash + each template/profile hash across 5 `process()` calls.
3. `backend/tests/component/core/schema_manager/test_manager_schema.py` — single-line correction.

## How to review

**Focus areas:**
- The two new ordering blocks in `process_pre_validation` (`schema_branch.py:639-665`). The argument for *why* each call is in its specific position is the comment above the block.
- `_validate_process_idempotent` is the helper every new test uses — confirm it actually catches drift (it asserts both the SchemaBranch hash and per-template / per-profile hashes individually because `SchemaBranch.get_hash()` only covers nodes + generics).

**Risky / uncertain:**
- The "scoped to generics only" pre-inheritance call is required because node HFIDs can reference attributes that don't exist on the node yet (e.g. `CoreGroupAction.name` comes from a generic). The `raise_parsing_errors=False` flag handles transient parse failures without swallowing real validation errors — those get raised later in `process_validate` as structured `ValueError`s, which is what the proposed-change pipeline already knows how to record (verified by `test_schema_integrity_process_validation_error`).

## How to test

```bash
# Idempotency component tests (the new coverage)
uv run pytest -v backend/tests/component/core/schema/schema_branch/test_uniqueness_propagation.py

# Full schema component suite (no regressions)
uv run pytest backend/tests/component/core/schema/ backend/tests/component/core/schema_manager/

# GraphQL metadata tests (the original CI failure class — interface field conformance)
uv run pytest backend/tests/component/graphql/metadata/

# Schema-integrity proposed change test (the validation-error pathway test)
uv run pytest backend/tests/component/message_bus/operations/requests/test_proposed_change.py::test_schema_integrity_process_validation_error

# Verify GraphQL schema is unchanged
uv run invoke schema.generate-graphqlschema
git diff schema/schema.graphql  # must be empty
```

## Impact & rollout

- **Backward compatibility:** It is possible that this update can cause a schema to load differently that it did before if it has one of these uniqueness propagation issues. For example, it could cause an attribute to be removed from Profile or Template schemas unexpectedly. However, this is basically raising an issue that the instance would have encountered eventually when a schema update or two were loaded.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Makes HFID- and constraint-derived uniqueness idempotent so repeated `SchemaBranch.process()` no longer mutates nodes, generics, templates, or profiles. Finalizes HFIDs, uniqueness, labels, and weights before template/profile generation with safe, deferred HFID parsing.

- **Bug Fixes**
  - Reordered `process_pre_validation`: run `process_human_friendly_id(kinds=generics, raise_parsing_errors=False)` and `sync_uniqueness_constraints_and_unique_attributes(kinds=generics)` before inheritance; then run both unscoped, plus `process_labels()` and `generate_weight()`, before managing templates/profiles.
  - Added scoping and error controls: `process_human_friendly_id(kinds, raise_parsing_errors)` and `sync_uniqueness_constraints_and_unique_attributes(kinds)`; split into `_derive_human_friendly_id` and `_propagate_human_friendly_id_to_constraints` with guarded path parsing.
  - Tests: added idempotence suite (including inherited + local relationships and template peer resolution); integration test now asserts inherited `uniqueness_constraints` propagate to nodes; updated template weight test to exclude `name`.

<sup>Written for commit 4f40bcfacbe47c62c57507396957a90ba854df3a. Summary will update on new commits.</sup>


PhillSimonds:
I did some testing on this and it looks to be resolving the issue with schema nodes falling out of sync. TY for all the work here!

ogenstad (approved):

PR #8939, opened 2026-04-16
Add testing guidelines related to schemas

gmazoyer:
# Why

Test schema definitions are duplicated across many test files. We already have shared helpers in `tests/helpers/schema/`, but there's no documented guidance on when and how to use them. Similarly, GraphQL error assertions in tests are often too loose (substring checks or just `assert result.errors`), which lets regressions slip through undetected.

## What changed

- Added a **Test Schemas** section to `dev/guidelines/backend/testing.md` with guidelines for reusing shared schema helpers, deriving variants with `deepcopy`, and when to promote local schemas to shared helpers.
- Added a **GraphQL Result Assertions** section with examples showing why exact `==` checks on error messages are preferable to substring or existence checks.

## Checklist

---
## Summary by cubic
Adds testing guidelines to `dev/guidelines/backend/testing.md` on reusing schemas from `tests/helpers/schema/` via `load_schema`, deriving variants with `deepcopy`, and using exact `==` checks for GraphQL errors. Updates `.github/CODEOWNERS` to tag the backend team for changes in `dev/guidelines/backend`, `dev/knowledge/backend`, and `python_sdk`.

<sup>Written for commit 17f0988ed76dead1377630c889a591883f6fd43d. Summary will update on new commits.</sup>

### DIFF
--- a/dev/guidelines/backend/testing.md
+++ b/dev/guidelines/backend/testing.md
@@ -25,6 +25,38 @@ Tests are behavioral specifications. Write them to describe **what the system sh
 - **Do not reference issue numbers, GitHub URLs, or Jira tickets** in test code (names, comments, or docstrings). The git history links commits to issues.
 - **Do not describe which bug a test prevents.** Describe the expected behavior instead.
 
+## Test Schemas
+
+Many tests require schemas to be loaded before they can run. Over time this has led to duplicated schema definitions scattered across test files. To reduce duplication and ease maintenance, shared helper schemas are available in [`tests/helpers/schema/`](../../../backend/tests/helpers/schema/).
+
+### Available helpers
+
+The module provides individual node/generic schemas (`CAR`, `DEVICE`, `TAG`, `PERSON`, etc.) as well as pre-composed `SchemaRoot` bundles (`CAR_SCHEMA`, `DEVICE_SCHEMA`, `LOCATION_SCHEMA`, `SNOW_TICKET_SCHEMA`, etc.). A `load_schema` helper function handles registering a schema in the branch registry during tests.
+
+### Guidelines
+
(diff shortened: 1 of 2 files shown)


dev/guidelines/backend/testing.md
```diff
@@ -229,6 +261,28 @@ assert "no more addresses available" in exc_info.value.message
...
+# Bad - a substring check passes for any error that mentions the ID
+assert TEMPLATE_ID in str(result.errors[0])
+
+# Bad - slightly better but still a substring match, any error containing
+# this text passes even if the overall message changed
+assert f"The template requested {{'id': '{TEMPLATE_ID}'}} was not found." in str(result.errors[0])
+
+# Good - exact match on the error message
+assert result.errors
+assert str(result.errors[0].message) == f"The template requested {{'id': '{TEMPLATE_ID}'}} was not found."
```
polmichel, line 281: The question here is about balancing resistance to refactoring and regression protection. 

Wouldn't you agree that including the key elements in the error message is safer than asserting the exact message? If we assert the exact message, even slight modifications or additional details in the message will cause the test to fail for the "wrong" reason.
I think we should avoid these fake negative cases, as they reduce our confidence in the test suite and the "test failing" event.

In my opinion: the third case `assert f"The template requested {{'id': '{TEMPLATE_ID}'}} was not found." in str(result.errors[0])` is better, although this pattern could be arguably too explicit in certain circumstances.

This is a personal opinion, if you all think this is better, let's do this.

gmazoyer: Matter of taste probably indeed. Mine's expressed in this PR.

ajtmccarty: I think testing for full equality where possible is best b/c it ensures the error message is exactly as expected. It might cause an unexpected test failure in the CI after updating an error message, but I think that is acceptable

ajtmccarty (approved):

polmichel (approved):

PR #8940, opened 2026-04-16
install playwright browsers for bug-fixing pipeline

polmichel:
## Problem

The test-writer pipeline has failed for two reasons: 

- the branch was already existing
- because `vitest-browser-react` (used by all frontend tests) requires Playwright's headless Chromium binary, and it wasn't being installed in the bug-agent workflows. The regular CI already had it.

The pipeline posted a comment on the gh issue to explain this, I have copied/pasted it in this PR.

Related Github action run: https://github.com/opsmill/infrahub/actions/runs/24499569202.

## Solution

Add the same CI step about installing Playwright browsers.

---
## Summary by cubic
Install Playwright Chromium in bug-agent workflows so frontend tests can run headless in CI. This fixes test failures in the bug-fixing pipeline caused by missing browser binaries.

- **Bug Fixes**
  - Added `npx playwright install chromium` after `npm ci` in `bug-agent-fix.yml` and `bug-agent-test.yml`.
  - Steps respect existing `skip` conditions.
  - Aligns bug-agent pipelines with regular CI and unblocks `vitest-browser-react` tests.

<sup>Written for commit bf846ca5d2cd19571d348c5c881104e652d2fcf8. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/bug-agent-fix.yml
+++ b/.github/workflows/bug-agent-fix.yml
@@ -128,6 +128,9 @@ jobs:
       - name: Install frontend dependencies
         if: steps.resolve.outputs.skip != 'true'
         run: cd frontend/app && npm ci
+      - name: Install Playwright browsers
+        if: steps.resolve.outputs.skip != 'true'
+        run: cd frontend/app && npx playwright install chromium
 
       - name: Read prompt
         if: steps.resolve.outputs.skip != 'true'
@@ -337,6 +340,9 @@ jobs:
       - name: Install frontend dependencies
         if: steps.resolve.outputs.skip != 'true'
         run: cd frontend/app && npm ci
+      - name: Install Playwright browsers
+        if: steps.resolve.outputs.skip != 'true'
+        run: cd frontend/app && npx playwright install chromium
 
       - name: Read prompt
         if: steps.resolve.outputs.skip != 'true'
--- a/.github/workflows/bug-agent-test.yml
+++ b/.github/workflows/bug-agent-test.yml
@@ -109,6 +109,9 @@ jobs:
       - name: Install frontend dependencies
         if: steps.analyst.outputs.skip != 'true'
         run: cd frontend/app && npm ci
+      - name: Install Playwright browsers
+        if: steps.analyst.outputs.skip != 'true'
+        run: cd frontend/app && npx playwright install chromium
 
       - name: Read prompt
         if: steps.analyst.outputs.skip != 'true'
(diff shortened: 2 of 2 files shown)


polmichel:
Comment from Claude: 

## Test-writer escalation: cannot verify test due to missing Playwright browsers

**Branch:** `ai-bug-pipeline-8857-copy-id-hfid`

### What was found

A previous pipeline run already committed a test to this branch:
- `frontend/app/src/shared/hooks/useCopyToClipboard.test.ts`
- Commit `542069e29`: `test: add failing test for #8857`
- Commit `03afdb5d1`: `tests update`

The current run wrote an equivalent test with a more robust setup (avoiding a potential `TypeError` in `beforeEach` when `navigator.clipboard` is `undefined` in an HTTP context), but cannot be pushed because the remote branch has diverged and force-push is not permitted.

### Why verification is blocked

The Vitest test suite requires Playwright's Chromium browser (`/home/runner/.cache/ms-playwright/chromium_headless_shell-1194/chrome-linux/headless_shell`), which is **not installed** in this environment.

The command `npx playwright install chromium` is not in the agent's allowed permissions list, and no system Chromium binary was found.

Both `npm --prefix frontend/app run test -- useCopyToClipboard` attempts failed with:
```
Error: browserType.launch: Executable doesn't exist at .../headless_shell
```

### Test logic analysis (manual verification)

The test at `frontend/app/src/shared/hooks/useCopyToClipboard.test.ts` asserts:

**When `window.isSecureContext` is `false` and `navigator.clipboard.writeText` returns a rejected Promise (async, not a synchronous throw), calling `copyToClipboard(value)` WITHOUT awaiting must result in `document.execCommand("copy")` being called synchronously.**

**On current code this assertion WILL fail:**
1. `copyToClipboard` is `async` — it calls `await navigator.clipboard.writeText(value)` unconditionally.
2. Since `writeText` returns a rejected Promise (not throws synchronously), `await` suspends the function.
3. The `catch` block runs as a microtask — AFTER the calling code continues.
4. At the `expect(execCommandSpy).toHaveBeenCalledWith("copy")` assertion line, `execCommand` has NOT been called yet.
5. Assertion fails: `expected "copy" but received no calls`.

**On fixed code this assertion WILL pass:**
The fix adds a synchronous guard: `if (!isSecureContext || !navigator.clipboard) { oldSchoolCopy(value); return; }` at the top, before any `await`. Since `isSecureContext` is `false`, `execCommand` is called synchronously. The assertion passes.

### Human action needed

1. Install Playwright browsers in the environment: `cd frontend/app && npx playwright install chromium`
2. Run: `cd frontend/app && npm run test -- useCopyToClipboard`
3. Confirm the test fails with `AssertionError: expected spy to have been called with "copy"` (not a browser launch error)
4. The test is ready to merge once empirically verified

bilalabbad (approved):

PR #8942, opened 2026-04-16
fix: search anywhere crashes on SchemaNode UUID lookup

BeArchiTek:
## Summary

- **Backend**: The UUID search path in `InfrahubSearchAnywhere` returned any node regardless of namespace. When a user searched for a SchemaNode UUID (e.g. from a proposed change validation error), the backend returned `kind: "SchemaNode"` which the frontend couldn't resolve, causing a crash. Added namespace filtering to exclude `Schema` and `Internal` nodes — consistent with the string search path that already filters by `CoreNode`/`CoreGroup`.
- **Frontend**: `useGetObject` was called with a null-asserted schema (`schema!`) before the null check, crashing in `getObjectQueryOptions` when accessing `.kind` on null. Split `NodesOptions` into a schema-resolving wrapper and a details component so `useGetObject` is only called when schema is guaranteed non-null.

## Test plan

- [x] Backend component test: `test_search_anywhere_by_uuid_excludes_internal_nodes` — creates nodes with `Internal` and `Schema` namespaces, verifies UUID search returns 0 results for both
- [x] Frontend unit test: `search-nodes.test.tsx` — verifies `NodesOptions` does not call `useGetObject` when `useSchema` returns null
- [ ] Manual: search for a SchemaNode UUID in the search bar — should return no results instead of crashing

---
## Summary by cubic
Fixes Search Anywhere crashing on Schema/Internal UUIDs by returning those nodes with a `display_label` and a `target_kind`, then rendering a simple result that links to the actual schema entry. Links preserve branch/datetime query params; text search and regular nodes are unchanged.

- **New Features**
  - GraphQL: Added nullable `display_label` and `target_kind` on search `Node`; UUID search now returns Schema/Internal nodes with both fields; schema docs updated.
  - Frontend: Requests `display_label`/`target_kind` and uses `target_kind` for link and kind badge in a simplified row (no extra fetch).

- **Bug Fixes**
  - Split `NodesOptions` into a schema resolver and a details component so `useGetObject` is never called with a null schema, preventing the crash.
  - Schema links use `constructPath` to keep existing branch/datetime query params.
  - Resolved mypy errors in the UUID search path by renaming the result dict and marking dynamic SchemaNode attributes with `attr-defined` ignores.
  - Regenerated GraphQL schema and TS types; updated tests to assert exact `display_label` values and ensure text search still excludes Schema/Internal; updated `.betterer` snapshot.
  - Aligned `python_sdk` submodule with `stable` to fix type checks and link validation in CI.

<sup>Written for commit a03c651726df7ba271442ee2f7038b9285e4fbc6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/8942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/component/graphql/queries/test_search.py
```diff
@@ -67,6 +70,137 @@ async def test_search_anywhere_by_uuid(
...
+        variable_values={"search": internal_node.id},
+    )
+
+    assert result.errors is None
+    assert result.data
+    assert result.data["InfrahubSearchAnywhere"]["count"] == 1
+    node_data = result.data["InfrahubSearchAnywhere"]["edges"][0]["node"]
+    assert node_data["id"] == internal_node.id
+    assert node_data["kind"] == "InternalWidget"
+    assert node_data["display_label"] is not None
```
ajtmccarty, line 121: the test should be able to assert what the display label is

backend/tests/component/graphql/queries/test_search.py
```diff
@@ -67,6 +70,137 @@ async def test_search_anywhere_by_uuid(
...
+        root_value=None,
+        variable_values={"search": car_accord_main.id},
+    )
+
+    assert result.errors is None
+    assert result.data
+    node_data = result.data["InfrahubSearchAnywhere"]["edges"][0]["node"]
+    assert node_data["display_label"] is not None
+    assert isinstance(node_data["display_label"], str)
+    assert len(node_data["display_label"]) > 0
```
ajtmccarty, line 163: samesies

backend/tests/component/graphql/queries/test_search.py
```diff
@@ -67,6 +70,137 @@ async def test_search_anywhere_by_uuid(
...
+        source=SEARCH_QUERY,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"search": "findable-widget"},
+    )
+
+    assert result.errors is None
+    assert result.data
+    assert result.data["InfrahubSearchAnywhere"]["count"] == 0
+    assert result.data["InfrahubSearchAnywhere"]["edges"] == []
```
ajtmccarty, line 201: doesn't seem to be introduced in this PR, but it seems like an anti-pattern that `Internal` schema objects are only included in the search results if the search term is the ID of the internal object. I would expect searching for the name of `findable-widget` here to return it

backend/infrahub/graphql/queries/search.py
```diff
@@ -21,6 +21,16 @@
...
     id = Field(String, required=True)
     kind = Field(String, required=True, description="The node kind")
+    display_label = Field(String, required=False, description="Human-readable label for the node")
+    target_kind = Field(
+        String,
+        required=False,
+        description=(
+            "For Schema records (SchemaNode/SchemaGeneric), the kind of the schema they "
+            "describe so clients can link to that schema's page instead of the generic "
+            "SchemaNode/SchemaGeneric page."
```
ajtmccarty, line 31: this description of a GraphQL field shouldn't describe how the client works. maybe something like 
"If kind is SchemaNode/SchemaGeneric, target_kind is set to the name of the schema (eg BuiltinTag)"

backend/tests/component/graphql/queries/test_search.py
BeArchiTek: Done. The test now configures `display_labels: ["name__value"]` on the `Internal`/`Schema` test schemas and asserts the exact values (`display_label == "test-internal"` and `"test-schema"`) instead of a non-null check.

---
_Generated by [Claude Code](https://claude.ai/code)_

backend/tests/component/graphql/queries/test_search.py
BeArchiTek: Done. This test now asserts the exact value (`display_label == "accord #444444"`, from `TestCar`'s `["name__value", "color__value"]` display labels with the default color) instead of just checking length.

---
_Generated by [Claude Code](https://claude.ai/code)_

backend/infrahub/graphql/queries/search.py
BeArchiTek: Reworded the field description to document the contract rather than client behavior:

> If kind is SchemaNode/SchemaGeneric, target_kind is set to the name of the schema the record describes (eg BuiltinTag).

`schema/schema.graphql` regenerated accordingly.

---
_Generated by [Claude Code](https://claude.ai/code)_

backend/tests/component/graphql/queries/test_search.py
BeArchiTek: Agreed it's an existing quirk and not introduced here. Text search is deliberately scoped to `CoreNode`/`CoreGroup` kinds, so `Internal`/`Schema` objects only surface on an exact UUID match (which is the path this PR fixes). Broadening text search to return them by name would mean the frontend has to render arbitrary internal/schema kinds, which is exactly the class of crash this PR is guarding against — so I'd keep it out of scope here and track it as a follow-up. I've kept `test_search_anywhere_text_excludes_schema_internal_nodes` asserting the current exclusion so the behavior is at least pinned. Let me know if you'd prefer a ticket.

---
_Generated by [Claude Code](https://claude.ai/code)_

BeArchiTek:
Replaced by #9596

PR #8946, opened 2026-04-16
Use inline fragment for hierarchical relationship query fields

### LINKED ISSUE #8943: bug: Jinja2 based computed attributes in a hierarchy targeting none generic attribute fails
### Component

_No response_

### Infrahub version

v1.8.4

### Current Behavior

When you have a computed attribute on a node that's part of a hierarchy and you try to use a "remote" attribute from the parent and this attribute isn't defined on the generic that defines the hierarchy the generated GraphQL query is invalid.

### Expected Behavior

When processing computed attributes we have to ensure to render the GraphQL query correctly by using an inline fragment.

### Steps to Reproduce

1. Load this schema
```yaml
---
version: "1.0"

generics:
  - name: Generic
    namespace: Location
    hierarchical: true
    attributes:
      - name: name
        kind: Text
        unique: true
      - name: shortname
        kind: Text

nodes:
  - name: Continent
    namespace: Location
    inherit_from:
      - LocationGeneric
    human_friendly_id:
      - name__value
    parent: ""
    children: LocationCountry

  - name: Country
    namespace: Location
    inherit_from:
      - LocationGeneric
    human_friendly_id:
      - country_path__value
    display_label: "{{ country_path__value }}"
    parent: LocationContinent
    children: LocationCity
    attributes:
      - name: country_code
        kind: Text
        optional: false
      - name: country_path
        kind: Text
        optional: false
        read_only: true
        computed_attribute:
          kind: Jinja2
          jinja2_template: "{{ parent__name__value }}-{{ shortname__value }}"

  - name: City
    namespace: Location
    inherit_from:
      - LocationGeneric
    human_friendly_id:
      - location_code__value
    display_label: "{{ location_code__value }}"
    parent: LocationCountry
    attributes:
      - name: location_code
        kind: Text
        optional: false
        read_only: true
        computed_attribute:
          kind: Jinja2
          jinja2_template: "{{ parent__country_code__value }}-{{ shortname__value }}"
```

2. Create a node of each kind

Could be Europe, Sweden, Stockholm for each appropriate node kind and add any kind of information

3. Update the name of the continent

Note that after converging the `country_path` attribute on the `LocationCountry` node will have been correctly updated. This is because "name" is an attribute that exists on the generic `LocationGeneric`.

4. Update the `country_code` attribute on the `LocationCountry` node

Observe the logs of the Prefect workers and observe the failure which will look something like this:

```python
  File "/Users/patrick/Code/opsmill/infrahub/.venv/lib/python3.13/site-packages/prefect/flow_engine.py", line 1566, in run_context
    yield self
  File "/Users/patrick/Code/opsmill/infrahub/.venv/lib/python3.13/site-packages/prefect/flow_engine.py", line 1628, in run_flow_async
    await engine.call_flow_fn()
  File "/Users/patrick/Code/opsmill/infrahub/.venv/lib/python3.13/site-packages/prefect/flow_engine.py", line 1580, in call_flow_fn
    result = await call_with_parameters(self.flow.fn, self.parameters)
  File "/Users/patrick/Code/opsmill/infrahub/backend/infrahub/computed_attribute/tasks.py", line 255, in process_jinja2
    response = await client.execute_graphql(query=query, branch_name=branch_name)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/infrahub/python_sdk/infrahub_sdk/client.py", line 991, in execute_graphql
    raise GraphQLError(errors=response["errors"], query=query, variables=variables)
infrahub_sdk.exceptions.GraphQLError: An error occurred while executing the GraphQL Query
query ComputedAttributeFilter {
    LocationCity(parent__ids: "18a6cee7-e995-7608-58dd-1746161dcab1") {
        edges {
            node {
                parent {
                    node {
                        country_code {
                            value
                        }
                    }
                }
                shortname {
                    value
                }
                id
                location_code {
                    value
                }
            }
        }
    }
}
, [{'message': "Cannot query field 'country_code' on type 'LocationGeneric'. Did you mean to use an inline fragment on 'LocationCountry'?", 'locations': [{'line': 8, 'column': 25}]}]
```

### Additional Information

Looking closer at the query we can see that it's invalid since `country_code` is not an attribute that exists on the `LocationGeneric`.

The correct query in this case should be:

```graphql
query ComputedAttributeFilter {
  LocationCity(parent__ids: "18a6cee7-e995-7608-58dd-1746161dcab1") {
    edges {
      node {
        parent {
          node {
            ... on LocationCountry {
              country_code {
                value
              }
            }
          }
        }
        shortname {
          value
        }
        id
        location_code {
          value
        }
      }
    }
  }
}
```

In normal circumstances we can't use a fragment like this for relationships with computed attributes. However in this case we should be able to figure out from the schema that `LocationCountry` is a parent of `LocationCity` and since they are part of a hierarchy we know that a `LocationCity` will never have another node kind than `LocationCountry` so the use of this type of fragment is safe under those circumstances.

polmichel:
## Why

When a Jinja2 computed attribute references an attribute on a hierarchical relationship's concrete peer type (e.g., `{{ parent__slug__value }}` where `slug` exists on `LocationCountry` but not on the hierarchy generic `LocationGeneric`), the generated GraphQL query is rejected because it queries the attribute directly on the generic type instead of using an inline fragment.

**Goal:** Generate valid GraphQL queries that use inline fragments (`... on PeerType`) for hierarchical relationships where the peer differs from the generic.

**Non-goals:** This PR does not change response parsing — GraphQL flattens inline fragment results, so `to_node_response`/`parse_response` work unchanged.

Closes #8943

## What changed

- **Behavioral change:** `query_fields` in `ComputedAttrJinja2GraphQL`, `DisplayLabelJinja2GraphQL`, and `HFIDGraphQL` now wraps relationship attributes in an inline fragment (`... on <PeerType>`) when the relationship is hierarchical and `peer != hierarchical`. This produces valid GraphQL like `parent { node { ... on LocationCountry { slug { value } } } }` instead of the invalid `parent { node { slug { value } } }`.

- **Implementation:** When `relationship.hierarchical` is set and differs from `relationship.peer`, the attribute dict is nested under a `f"... on {relationship.peer}"` key. The `render_query_block` renderer in the SDK already handles arbitrary string keys, so no renderer changes were needed.

## How to review

Focus on the three `query_fields` property changes — they are identical across the three files:
1. `backend/infrahub/computed_attribute/models.py` (primary)
2. `backend/infrahub/display_labels/models.py` (same pattern)
3. `backend/infrahub/hfid/models.py` (same pattern)

Then review the component and unit tests.

## How to test

```
uv run pytest backend/tests/unit/computed_attribute/test_models.py backend/tests/unit/display_labels/test_models.py backend/tests/unit/hfid/test_models.py backend/tests/unit/core/schema/test_schema_branch.py backend/tests/component/graphql/test_inline_fragment_hierarchical.py -x -v
```
## Checklist

- [x] Tests added/updated
- [x] Internal .md docs updated (internal knowledge and AI code tools knowledge) — N/A

### DIFF
--- a/backend/infrahub/hfid/models.py
+++ b/backend/infrahub/hfid/models.py
@@ -203,7 +203,13 @@ class HFIDGraphQL(BaseModel):
                 if relationship.cardinality == RelationshipCardinality.ONE:
                     if field_name not in output:
                         output[field_name] = {"node": {}}
-                    output[field_name]["node"][related_attribute] = {related_value: None}
+                    if relationship.hierarchical and relationship.peer != relationship.hierarchical:
+                        fragment_key = f"... on {relationship.peer}"
+                        if fragment_key not in output[field_name]["node"]:
+                            output[field_name]["node"][fragment_key] = {}
+                        output[field_name]["node"][fragment_key][related_attribute] = {related_value: None}
+                    else:
+                        output[field_name]["node"][related_attribute] = {related_value: None}
         return output
 
     def parse_response(self, response: dict[str, Any]) -> list[HFIDGraphQLResponse]:
--- a/backend/tests/component/graphql/test_inline_fragment_hierarchical.py
+++ b/backend/tests/component/graphql/test_inline_fragment_hierarchical.py
@@ -1,0 +1,236 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from infrahub.computed_attribute.jinja2 import InfrahubJinja2Template
+from infrahub.computed_attribute.models import ComputedAttrJinja2GraphQL
+from infrahub.core import registry
+from infrahub.core.node import Node
+from infrahub.core.schema import AttributeSchema, GenericSchema, NodeSchema, SchemaRoot
+from infrahub.core.schema.computed_attribute import ComputedAttribute, ComputedAttributeKind
+from infrahub.display_labels.models import DisplayLabelJinja2GraphQL
+from infrahub.hfid.models import HFIDGraphQL
+from tests.helpers.graphql import graphql_query
+from tests.helpers.schema import load_schema
+
+if TYPE_CHECKING:
+    from infrahub.core.branch import Branch
+    from infrahub.core.schema.schema_branch import SchemaBranch
+    from infrahub.database import InfrahubDatabase
+
+
+# ---------------------------------------------------------------------------
+# Schema matching issue: country_code only exists on Country, not on
+# the LocationGeneric hierarchy generic.  City's computed location_code
+# references parent__country_code__value which must produce an inline
+# fragment in the generated GraphQL query.
+# ---------------------------------------------------------------------------
+
+LOCATION_GENERIC = GenericSchema(
+    name="Location",
+    namespace="Testing",
+    hierarchical=True,
+    label="Location",
+    default_filter="name__value",
+    attributes=[
+        AttributeSchema(name="name", kind="Text", unique=True, optional=False),
+        AttributeSchema(name="shortname", kind="Text", unique=True, optional=False),
+    ],
+)
+
+CONTINENT = NodeSchema(
+    name="Continent",
+    namespace="Testing",
+    label="Continent",
+    default_filter="name__value",
+    inherit_from=["TestingLocation"],
+    parent="",
+    children="TestingCountry",
+    generate_profile=False,
+)
+
+COUNTRY = NodeSchema(
+    name="Country",
+    namespace="Testing",
+    label="Country",
+    default_filter="name__value",
+    inherit_from=["TestingLocation"],
+    parent="TestingContinent",
+    children="TestingCity",
+    generate_profile=False,
+    attributes=[
+        AttributeSchema(name="country_code", kind="Text", optional=False),
+        AttributeSchema(
+            name="country_path",
+            kind="Text",
+            optional=False,
+            read_only=True,
+            computed_attribute=ComputedAttribute(
+                kind=ComputedAttributeKind.JINJA2,
+                jinja2_template="{{ parent__name__value }}-{{ shortname__value }}",
+            ),
+        ),
+    ],
+)
+
+CITY = NodeSchema(
+    name="City",
+    namespace="Testing",
+    label="City",
+    default_filter="name__value",
+    inherit_from=["TestingLocation"],
+    parent="TestingCountry",
+    children="",
+    generate_profile=False,
+    attributes=[
+        AttributeSchema(
+            name="location_code",
+            kind="Text",
+            optional=False,
+            read_only=True,
+            computed_attribute=ComputedAttribute(
+                kind=ComputedAttributeKind.JINJA2,
+                jinja2_template="{{ parent__country_code__value }}-{{ shortname__value }}",
+            ),
+        ),
+    ],
+)
+
+
+async def assert_correct_graphql_structure(
+    db: InfrahubDatabase, default_branch: Branch, rendered_query: str, expected_value: str
+) -> None:
+    assert "... on TestingCountry" in rendered_query
+
+    result = await graphql_query(query=rendered_query, db=db, branch=default_branch)
+
+    assert result.errors is None
+    edges = result.data["TestingCity"]["edges"]
+    assert len(edges) == 1
+
+    node = edges[0]["node"]
+    assert node["parent"]["node"]["country_code"]["value"] == expected_value
+
+
+@pytest.fixture
+async def hierarchical_schema(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: SchemaBranch,
+) -> SchemaBranch:
+    await load_schema(
+        db,
+        schema=SchemaRoot(generics=[LOCATION_GENERIC], nodes=[CONTINENT, COUNTRY, CITY]),
+    )
+    return registry.schema.get_schema_branch(name=default_branch.name)
+
+
+def _extract_variables(template: str) -> list[str]:
+    return InfrahubJinja2Template(template=template).get_variables()
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+async def test_computed_attr_graphql_query_executes_successfully(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    hierarchical_schema: SchemaBranch,
+) -> None:
+    """The GraphQL query generated for a computed attribute that references a
+    peer-only parent attribute must use an inline fragment and execute without
+    errors against the database."""
+    continent = await Node.init(db=db, schema="TestingContinent", branch=default_branch)
+    await continent.new(db=db, name="Europe", shortname="eu")
+    await continent.save(db=db)
+
+    country = await Node.init(db=db, schema="TestingCountry", branch=default_branch)
+    expected_country_code = "SWE"
+    await country.new(db=db, name="Sweden", shortname="se", country_code=expected_country_code, parent=continent)
+    await country.save(db=db)
+
+    city = await Node.init(db=db, schema="TestingCity", branch=default_branch)
+    await city.new(db=db, name="Stockholm", shortname="sth", parent=country)
+    await city.save(db=db)
+
+    city_schema = hierarchical_schema.get_node(name="TestingCity")
+    attr_schema = city_schema.get_attribute(name="location_code")
+    variables = _extract_variables(attr_schema.computed_attribute.jinja2_template)
+
+    graphql_obj = ComputedAttrJinja2GraphQL(
+        node_schema=city_schema,
+        attribute_schema=attr_schema,
+        variables=variables,
+    )
+
+    rendered_query = graphql_obj.render_graphql_query(query_filter="parent__ids", filter_id=country.id)
+    await assert_correct_graphql_structure(db, default_branch, rendered_query, expected_country_code)
+
+
+async def test_display_label_graphql_query_executes_successfully(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    hierarchical_schema: SchemaBranch,
+) -> None:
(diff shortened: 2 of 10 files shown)


polmichel:
## APPROVED WITH SUGGESTIONS

### A. Test realism

**Pass.** Test inputs directly mirror the real `SITE` schema in `backend/tests/helpers/schema/location.py`:
- `TestingCountry`/`TestingLocation` kinds match the existing test hierarchy
- Variables `parent__slug__value` and `shortname__value` match the SITE computed attribute template `{{ parent__slug__value }}-{{ shortname__value }}`
- The `RelationshipSchema` with `peer="TestingCountry"`, `hierarchical="TestingLocation"`, `cardinality=ONE` is exactly how hierarchy parent relationships are configured in production

### B. Test correctness

**Pass.** The test:
- Asserts the **correct** behavior (inline fragment must be present), not the buggy behavior
- Exercises `ComputedAttrJinja2GraphQL.query_fields` directly — the exact code path at `models.py:364-378`
- Cannot pass without modifying the production `query_fields` property — confirmed by the failing output showing `{'slug': {'value': None}}` instead of the expected `{'... on TestingCountry': {'slug': {'value': None}}}`
- Also asserts the rendered GraphQL string, providing end-to-end coverage of the query building pipeline

### C. Test quality

**Pass.**
- Pure unit test — no DB, no fixtures, no external dependencies. Fully isolated and deterministic.
- Placed in the correct file (`tests/unit/computed_attribute/test_models.py`) matching source structure
- Descriptive name and docstring describe expected behavior without referencing issue numbers (per `dev/guidelines/backend/testing.md`)
- Tests observable behavior (dict structure + rendered query string), not internal implementation details
- Custom assertion message provides clear diagnostic output on failure

### D. Alignment with analysis

**Pass.** The test precisely targets the root cause: `query_fields` places relationship attributes directly under `node {}` instead of wrapping them in `... on <PeerType>` when `relationship.hierarchical` differs from `relationship.peer`. The failure output confirms this is exactly what happens.

### Suggestion (non-blocking)

The analysis identifies the identical bug in `DisplayLabelJinja2GraphQL.query_fields` (`display_labels/models.py:194-207`) and `HFIDJinja2GraphQL.query_fields` (`hfid/models.py:194-207`). The fixer should add parallel tests for those two classes since all three have identical `query_fields` implementations that need the same fix. This is a note for the fix phase, not a blocker for this test PR.

### Recommended next steps

1. Proceed to the fix phase — the test is ready to validate the fix.
2. When fixing, apply the inline fragment logic to all three `query_fields` implementations and add corresponding tests for `DisplayLabelJinja2GraphQL` and `HFIDGraphQL`.

polmichel:
## APPROVED WITH SUGGESTIONS

### A. Correctness

**Pass.** The fix exactly follows the analyst's strategy:

- Condition `relationship.hierarchical and relationship.peer != relationship.hierarchical` correctly identifies hierarchical relationships where the GraphQL type resolves to the generic, not the concrete peer.
- Attributes are wrapped under `f"... on {relationship.peer}"` dict key, which `render_query_block` in `python_sdk/infrahub_sdk/graphql/renderers.py:163` renders verbatim as valid GraphQL inline fragment syntax.
- The `else` branch preserves existing behavior for non-hierarchical relationships — no regression risk.
- `to_node_response`/`parse_response` correctly left untouched — GraphQL flattens inline fragment results into the response, so the parsing path is unchanged.
- Multiple variables referencing the same hierarchical relationship are handled correctly: `if fragment_key not in output[field_name]["node"]` deduplicates the fragment container.
- Applied identically to all three affected classes (`ComputedAttrJinja2GraphQL`, `DisplayLabelJinja2GraphQL`, `HFIDGraphQL`) as the analysis required.

### B. Code quality

**Pass.**
- Minimal, targeted change — 6 new lines replacing 1 in each file, no unnecessary refactoring.
- Follows existing code style: indentation, naming, dict manipulation patterns are consistent with surrounding code.
- No performance concerns — dict key manipulation during query building is negligible.
- No security concerns — the `relationship.peer` value comes from the schema registry, not user input.

### C. Documentation alignment

**Pass.**
- Respects all analyst guardrails: does NOT modify `to_node_response`/`parse_response`, does NOT look up the schema registry to check attribute existence on the generic, does NOT change the renderer.
- PR body clearly explains the behavioral change, what stayed the same, and rollout impact.
- Changelog entry (`changelog/8943.fixed.md`) follows `dev/guidelines/changelog.md` conventions: past tense, user-facing, one sentence.

### D. Test quality

**Pass with suggestion.**
- The existing test from the test-writer is valid and passes after the fix.
- The fixer did NOT modify the test file (verified via `git diff` between test commit and fix commits).
- The test covers the core scenario: hierarchical relationship with `peer != hierarchical`, asserting both dict structure and rendered query output.

**Suggestion (non-blocking):** The fix touches `DisplayLabelJinja2GraphQL` and `HFIDGraphQL` in addition to `ComputedAttrJinja2GraphQL`, but only `ComputedAttrJinja2GraphQL` has a test. While the code is identical across all three classes, parallel tests for the other two would guard against future divergence if someone refactors one class independently. Consider adding them as follow-up.

### Recommended next steps

1. Merge — the fix is correct, minimal, and well-tested for the primary code path.
2. Consider adding unit tests for `DisplayLabelJinja2GraphQL.query_fields` and `HFIDGraphQL.query_fields` as follow-up to cover the identical fix in those classes.

backend/infrahub/hfid/models.py
```diff
@@ -195,15 +195,25 @@ def query_fields(self) -> dict[str, Any]:
...
-                    output[field_name]["node"][related_attribute] = {related_value: None}
+                    # Edge case, the referenced attribute is part of the peer node schema but not part of the
+                    # generic node schema
+                    if relationship.hierarchical and relationship.peer != relationship.hierarchical:
+                        fragment_key = f"... on {relationship.peer}"
+                        if fragment_key not in output[field_name]["node"]:
+                            output[field_name]["node"][fragment_key] = {}
+                        output[field_name]["node"][fragment_key][related_attribute] = {related_value: None}
+                    else:
+                        output[field_name]["node"][related_attribute] = {related_value: None}
```
polmichel, line 212: This is clearly code duplication, but let's refactor it another time. Component and unit tests are in place, the refactoring will be safe

backend/tests/component/graphql/test_inline_fragment_hierarchical.py
```diff
@@ -0,0 +1,236 @@
...
+def _extract_variables(template: str) -> list[str]:
+    return InfrahubJinja2Template(template=template).get_variables()
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+async def test_computed_attr_graphql_query_executes_successfully(
```
polmichel, line 139: The three tests produce the same error mentioned in the issue when one reverts the fix
`"Cannot query field 'country_code' on type 'LocationGeneric'. Did you mean to use an inline fragment on 'LocationCountry'?"`

backend/tests/unit/core/schema/test_schema_branch.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        branch = SchemaBranch(cache={}, name="test")
+        branch.load_schema(schema=schema_root)
+        branch.process_inheritance()
+        branch.process_hierarchy()
+        branch.add_hierarchy_generic()
+        branch.add_hierarchy_node()
+        return branch
+
+    def test_concrete_node_parent_has_peer_different_from_hierarchical(self, processed_schema: SchemaBranch) -> None:
```
polmichel, line 49: This is certainly tested in higher level tests, but this is quick to execute and allow to test this behavior from an unit test perspective.
I think there is some value to keep it.

backend/tests/component/graphql/test_inline_fragment_hierarchical.py
```diff
@@ -0,0 +1,236 @@
...
+    return registry.schema.get_schema_branch(name=default_branch.name)
+
+
+def _extract_variables(template: str) -> list[str]:
+    return InfrahubJinja2Template(template=template).get_variables()
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
```
ogenstad, line 136: We can get rid of these section comments. The "test_" prefix to the function names should clarify this.

backend/tests/unit/display_labels/test_models.py
```diff
@@ -0,0 +1,88 @@
...
+
+from infrahub.core.constants import RelationshipCardinality
+from infrahub.core.schema import AttributeSchema, NodeSchema
+from infrahub.core.schema.relationship_schema import RelationshipSchema
+from infrahub.display_labels.models import DisplayLabelJinja2GraphQL
+
+
+class TestQueryFieldsInlineFragment:
+    @pytest.fixture(scope="class")
+    def build_schema(self) -> Callable[..., NodeSchema]:
```
ogenstad, line 13: Similar comments as above.

ogenstad (approved):
LGTM

backend/tests/unit/computed_attribute/test_models.py
```diff
@@ -64,3 +69,89 @@ async def test_load_from_prefect() -> None:
     query_obj = ComputedAttributeAutomations.from_prefect(automations=automations, prefix=QUERY_AUTOMATION_NAME_PREFIX)
     assert not query_obj.has(identifier="AAAAA", scope="default")
     assert query_obj.has(identifier="CCCCC", scope="default")
+
+
+class TestQueryFieldsInlineFragment:
+    @pytest.fixture(scope="class")
+    def build_schema(self) -> Callable[..., NodeSchema]:
```
ogenstad, line 76: Not a blocker for this PR, but it feels like a fixture for this just adds extra noice. It could be cleaner with just either a class method or a function called `_build()` instead of having the tests to require this fixture which is just a callable method.

PR #8948, opened 2026-04-16
docs: document display property

BaptisteGi:
I kept the naming for the topic relatively generic so if we add further features (field grouping...) we won't need to rename the page

---
## Summary by cubic
Added a docs page on controlling field visibility via the schema `display` property, covering behavior across list/detail/forms, examples, and how it interacts with other schema settings. Updated the docs sidebar to include the new topic.

<sup>Written for commit 1be48d09eaa38b5989a4af5a85e91785c32adb39. Summary will update on new commits.</sup>

### DIFF
--- a/docs/docs/topics/schema-display.mdx
+++ b/docs/docs/topics/schema-display.mdx
@@ -60,7 +60,7 @@ nodes:
         peer: LocationSite
         cardinality: one
       - name: audit_source
-        peer: InternalSource
+        peer: AuditSource
         cardinality: one
         display: extra          # hidden from list and detail views
 ```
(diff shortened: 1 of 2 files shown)


docs/docs/topics/schema-display.mdx
```diff
@@ -0,0 +1,91 @@
...
+        kind: Text
+      - name: internal_tracking_id
+        kind: Text
+        display: extra          # hidden from list and detail views
+    relationships:
+      - name: primary_site
+        peer: LocationSite
+        cardinality: one
+      - name: audit_source
+        peer: InternalSource
```
ogenstad: This example is actually problematic as the `Internal` namespace is restricted and not something that users can use within their own schemas. So I'd rework the name here a bit.

BaptisteGi:
```suggestion
        peer: AuditSource
```

PR #8959, opened 2026-04-17
Link branch merge/delete events to proposed change as primary node

pa-lem:
- `BranchDeletedEvent` and `BranchMergedEvent` now expose the associated proposed change (when present) as the event's **primary node** via `infrahub.node.id` / `infrahub.node.kind`, instead of attaching it through `get_related()`.
- Proposed change activity feed now subscribes to `infrahub.branch.merged` and `infrahub.branch.deleted`, so branch lifecycle events show up under the proposed change.
- Added `BRANCH_MERGED_EVENT` / `BRANCH_DELETED_EVENT` constants in the frontend branches entity and wired them into `PROPOSED_CHANGE_EVENTS`.
- Added unit tests covering the new resource payload for both events.
- Added changelog fragment documenting the breaking change for consumers filtering on `relatedNodes`.

image

---
## Summary by cubic
Branch merge/delete events now use the associated proposed change as the primary node (`infrahub.node.id` / `infrahub.node.kind`). This makes branch lifecycle events appear in the proposed change activity feed.

- **New Features**
  - Backend: Set the proposed change as the primary node for `infrahub.branch.merged` and `infrahub.branch.deleted` (removed from `get_related`).
  - Frontend: Added `BRANCH_MERGED_EVENT` and `BRANCH_DELETED_EVENT` and included them in `PROPOSED_CHANGE_EVENTS`.
  - Tests/Docs: Added unit tests for the new resource payload, updated integration tests to assert `primary_node`, and added a changelog entry.

- **Migration**
  - If you filter by proposed change using `relatedNodes`, switch to `primaryNodeIds`.
  - These branch events no longer include the proposed change under `relatedNodes`.

<sup>Written for commit 8f264ccf1db7d4f80b0a07e8a66f6e1ee2865a0d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/event/test_branch_action.py
+++ b/backend/tests/unit/event/test_branch_action.py
@@ -11,7 +11,7 @@ from infrahub.events.models import EventMeta
 
 
 def _make_meta() -> EventMeta:
-    branch = Branch(name="main", uuid=uuid.uuid4())
+    branch = Branch(name="test-branch-action-events", uuid=uuid.uuid4())
     return EventMeta(
         branch=branch,
         context=InfrahubContext.init(
--- a/backend/tests/unit/event/test_branch_action.py
+++ b/backend/tests/unit/event/test_branch_action.py
@@ -1,0 +1,116 @@
+from __future__ import annotations
+
+import uuid
+from unittest.mock import MagicMock
+
+from infrahub.auth import AccountSession, AuthType
+from infrahub.context import InfrahubContext
+from infrahub.core.branch import Branch
+from infrahub.core.constants import InfrahubKind
+from infrahub.events.branch_action import BranchDeletedEvent, BranchMergedEvent
+from infrahub.events.models import EventMeta
+
+
+def _make_meta() -> EventMeta:
+    branch = MagicMock(spec=Branch)
+    branch.name = "main"
+    branch.uuid = uuid.uuid4()
+    branch.get_uuid.return_value = branch.uuid
+    return EventMeta(
+        branch=branch,
+        context=InfrahubContext.init(
+            branch=branch,
+            account=AccountSession(auth_type=AuthType.NONE, authenticated=False, account_id=""),
+        ),
+    )
+
+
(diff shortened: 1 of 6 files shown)


backend/tests/unit/event/test_branch_action.py
```diff
@@ -0,0 +1,116 @@
...
+from infrahub.auth import AccountSession, AuthType
+from infrahub.context import InfrahubContext
+from infrahub.core.branch import Branch
+from infrahub.core.constants import InfrahubKind
+from infrahub.events.branch_action import BranchDeletedEvent, BranchMergedEvent
+from infrahub.events.models import EventMeta
+
+
+def _make_meta() -> EventMeta:
+    branch = MagicMock(spec=Branch)
```
polmichel: Just a **non-blocking** comment regarding this mock: is this mock required? 
`branch = Branch(name="main", uuid=uuid.uuid4())` would work I think

Using `MagicMock(spec=Branch)` or other mock tools have some caveats:
| Raise error on | `MagicMock(spec=)` | `MagicMock(spec_set=)` | `create_autospec` | `create_autospec(..., spec_set=True)` |                                                                    
  |---|---|---|---|---|           
  | **Get fake attr** | yes | yes | yes | yes |                                                                                                                            
  | **Set fake attr** | no | yes | no | yes |
  | **Wrong method signature** | no | no | yes | yes |                                                                                                                          
  | **Nested attrs auto-spec'd** | no | no | yes | yes |

But this is not necessary if we can just use the real object if this does not create additional dependencies / long execution time to the test.

This is non-blocking, and we should update the internal backend documentation about it, so you can ignore it if you want / if this is urgent no worries.

solababs (approved):

backend/tests/unit/event/test_branch_action.py
pa-lem: it's a fair assessment, I'll update it 👍

polmichel (approved):
Thanks for the mock update!

bilalabbad (approved):

PR #8976, opened 2026-04-20
Filter picker options are aligned with visible table columns

bilalabbad:
After product review, the decision was made to not display every filter possible. Instead, we only display filter for visible columns.

---
## Summary by cubic
Filter picker now only shows filters for columns visible in the table (IFC-2427). Metadata filters remain and appear after table fields.

- **New Features**
  - Added `getFilterDefinitions` in `ui/filters` to build options from visible list columns; skips resource‑pool relationships and appends metadata.
  - Moved icon and count logic to `ui/filters`; added tests for `getFilterDefinitions`, `getFilterDefinitionIcon`, and `getFilterPickerCount`.
  - Updated role‑manager table configs to match visibility: include `name` for Accounts, Groups, Roles, and Global Permissions; Object Permissions show `action` and `decision` only.

<sup>Written for commit b38ec14e8a4ab3c13d11c96d4d00c88072abb2fd. Summary will update on new commits.</sup>


frontend/app/src/entities/nodes/object/ui/filters/get-filter-definitions.ts
```diff
@@ -0,0 +1,130 @@
...
+  const relationshipNames = new Set(relationshipsVisible);
+
+  return getOrderedFieldDefinitions(schema.kind, [
+    ...(schema.attributes ?? []).filter(({ name }) => attributeNames.has(name)),
+    ...(schema.relationships ?? []).filter(({ name }) => relationshipNames.has(name)),
+  ]);
+}
+
+export function getFilterDefinitions(schema: ModelSchema): FilterDefinition[] {
+  const fieldDefinitions = (() => {
```
pa-lem: is it possible to avoid IIFE?

pa-lem (changes requested):
also a changelog (and maybe a screenshot?) could help understand the changes

frontend/app/src/entities/nodes/object/ui/filters/get-filter-definitions.ts
bilalabbad: updated

pa-lem (approved):

PR #8986, opened 2026-04-20
Split permission system into Loader, Resolver, and Manager

gmazoyer:
# Why

The permission system mixed three concerns inside `PermissionManager`: loading permissions from backends, resolving decisions, and enforcing them. On top of that, `report.py` had its own copy of the decision priority chain that sometimes disagreed with what the GraphQL pipeline enforced (super admins could mutate objects on merged branches even though the UI report said "denied.")

This PR splits the permission system into three components and makes the resolver the single source of truth so the report and the pipeline always agree.

## What

The permission stack is now `PermissionLoader` → `PermissionResolver` → `PermissionManager`. The loader owns the `registry.permission_backends` dependency. The resolver is a stateless decision engine that takes loaded permissions and a default branch name and computes everything. The manager is a thin facade that holds the resolver and provides the enforcement methods (`raise_for_*`).

A `PermissionManager.load_for_account(db, branch, account_session)` factory wires the three together for callers that want the all-in-one behaviour.

```python
# 3 components, each with one job
loader = PermissionLoader(account_session=account_session)
permissions = await loader.load(db=db, branch=branch)
resolver = PermissionResolver(permissions=permissions, default_branch_name=registry.default_branch)
manager = PermissionManager(account_session=account_session, resolver=resolver)

# Or, the factory wires everything:
manager = await PermissionManager.load_for_account(
  db=db, branch=branch, account_session=account_session
)
```

`report.py` now calls the resolver directly instead of reimplementing the priority chain. The pipeline and the report compute identical decisions because they go through the same code path.

## What stayed the same

- All `PermissionManager` enforcement methods keep their signatures (`has_permission`, `raise_for_permission`, `raise_for_permissions`, etc.)
- The pipeline registration order in `dependencies.py` is unchanged
- The `MANAGE_*` global permissions are still enforcement gates (Layer 2 from the analysis is intentionally out of scope)
- No database, GraphQL schema, or REST API changes
- No frontend changes required

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Unifies permission decisions behind a stateless `PermissionResolver` used by both the GraphQL checker pipeline and the permission report, eliminating mismatches and keeping `PermissionManager` focused on loading and enforcement.

- **Refactors**
  - Added `PermissionResolver` as the single source of truth for global/object and branch-relative decisions; exported from `infrahub.permissions`; added resolver tests.
  - `PermissionManager` now delegates via a lazy `resolver` property and invalidates it on load; public APIs unchanged.
  - Replaced the report’s logic: `report_schema_permissions()` now takes a `resolver`, precomputes globals with `resolver.build_global_report()`, and calls `resolver.get_branch_decision()`; removed `get_permission_report()`.
  - GraphQL permissions query now passes `active_permissions.resolver`.
  - Documented the checker pipeline contract in the interface; added a checker guide and permissions knowledge doc.

- **Bug Fixes**
  - Fixed pipeline vs report inconsistencies (mutations on merged/need-rebase branches, `manage-*` global checks, super admin precedence); rewrote merged-branch tests.

<sup>Written for commit 512f8aef12b7d3ca067d2b01c084b0cb2c043d3d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/permissions/loader.py
+++ b/backend/infrahub/permissions/loader.py
@@ -1,0 +1,30 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from infrahub.core import registry
+
+if TYPE_CHECKING:
+    from infrahub.auth import AccountSession
+    from infrahub.core.branch import Branch
+    from infrahub.database import InfrahubDatabase
+    from infrahub.permissions.types import AssignedPermissions
+
+__all__ = ["PermissionLoader"]
+
+
+class PermissionLoader:
+    """Loads permissions for an account session by orchestrating registered backends."""
+
+    def __init__(self, account_session: AccountSession) -> None:
+        self.account_session = account_session
+
+    async def load(self, db: InfrahubDatabase, branch: Branch) -> AssignedPermissions:
+        permissions: AssignedPermissions = {"global_permissions": [], "object_permissions": []}
+        for permission_backend in registry.permission_backends:
+            backend_permissions = await permission_backend.load_permissions(
+                db=db, branch=branch, account_session=self.account_session
+            )
+            permissions["global_permissions"].extend(backend_permissions["global_permissions"])
+            permissions["object_permissions"].extend(backend_permissions["object_permissions"])
+        return permissions
--- a/backend/infrahub/permissions/manager.py
+++ b/backend/infrahub/permissions/manager.py
@@ -26,7 +26,6 @@ class PermissionManager:
 
     @property
     def resolver(self) -> PermissionResolver:
-        """Lazily create a PermissionResolver from the current loaded permissions."""
         if self._resolver is None:
             self._resolver = PermissionResolver(permissions=self.permissions)
         return self._resolver
--- a/backend/infrahub/permissions/manager.py
+++ b/backend/infrahub/permissions/manager.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Sequence
+from typing import TYPE_CHECKING, Self, Sequence
 
 from infrahub.core import registry
 from infrahub.core.account import GlobalPermission
 from infrahub.exceptions import PermissionDeniedError
 from infrahub.permissions.constants import GLOBAL_PERMISSION_DENIAL_MESSAGE, PermissionDecisionFlag
+from infrahub.permissions.loader import PermissionLoader
 from infrahub.permissions.resolver import PermissionResolver
 
 if TYPE_CHECKING:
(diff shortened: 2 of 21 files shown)


backend/infrahub/permissions/report.py
```diff
@@ -2,126 +2,39 @@
...
-            }
-        )
-
-    return permission_objects
+    """Build a permission report for a list of schema types.
+
+    Uses PermissionResolver.get_branch_decision() as the single source of truth,
+    ensuring the report always matches what the pipeline enforces.
+    """
+    global_report = resolver.build_global_report()
```
ajtmccarty: feels a little weird to have to call `PermissionResolver.build_global_report()` just to pass it back in to `PermissionResolver.get_branch_decision()`. maybe `PermissionResolver` can build and cache the global report behind the scenes?

backend/infrahub/permissions/manager.py
```diff
@@ -19,12 +19,16 @@
...
     def __init__(self, account_session: AccountSession) -> None:
         self.account_session = account_session
         self.permissions: AssignedPermissions = {"global_permissions": [], "object_permissions": []}
+        self._resolver: PermissionResolver | None = None
+
+    @property
+    def resolver(self) -> PermissionResolver:
+        if self._resolver is None:
+            self._resolver = PermissionResolver(permissions=self.permissions)
+        return self._resolver
```
ajtmccarty: is it possible to pass `PermissionResolver` into `PermissionManager` where it is instantiated instead of creating the resolver within `PermissionManager`?

that might require moving `load_permissions` to a separate `PermissionLoader` component. not completely sure

backend/tests/component/permissions/test_resolver.py
```diff
@@ -0,0 +1,550 @@
...
+    return PermissionResolver(permissions=permissions)
+
+
+@pytest.fixture
+def empty_resolver() -> PermissionResolver:
+    return _make_resolver()
+
+
+@pytest.fixture
+def node(register_core_models_schema: None) -> MainSchemaTypes:
```
ajtmccarty: maybe `node_schema`?

backend/tests/unit/permissions/test_resolver.py
ajtmccarty, line 1: could this be a `unit` test module instead of a `component` test module? it looks like it's just the `registry` holding things back and you could pass `default_branch` into the `PermissionResolver` constructor to remove that dependency, I think

ajtmccarty (approved):

backend/infrahub/permissions/manager.py
```diff
@@ -1,116 +1,57 @@
...
-            specificity += 1
-        return specificity
+        self.resolver = resolver
+
+    @classmethod
+    async def load_for_account(cls, db: InfrahubDatabase, branch: Branch, account_session: AccountSession) -> Self:
+        loader = PermissionLoader(account_session=account_session)
+        permissions = await loader.load(db=db, branch=branch)
+        resolver = PermissionResolver(permissions=permissions, default_branch_name=registry.default_branch)
+        return cls(account_session=account_session, resolver=resolver)
```
ajtmccarty: this is definitely an improvement.

I think the construction and functional logic are a little mixed together here. ideally the `PermissionManager` class could be constructed completely without touching the database or awaiting anything. so the construction would look something like this 

```
manager = PermissionManager(
    resolver=PermissionResolver(
        loader=PermissionLoader(db=db, branch=branch),
        default_branch_name=registry.default_branch
    ),
    account_session=account_session
)
```

then when you call `manager.resolve_something(permission_to_check)`, then `PermissionManager` can call `PermissionLoader.load` and cache the permissions returned

I could of course be missing something that prevents us from doing this

backend/infrahub/permissions/loader.py
```diff
@@ -0,0 +1,30 @@
...
+
+class PermissionLoader:
+    """Loads permissions for an account session by orchestrating registered backends."""
+
+    def __init__(self, account_session: AccountSession) -> None:
+        self.account_session = account_session
+
+    async def load(self, db: InfrahubDatabase, branch: Branch) -> AssignedPermissions:
+        permissions: AssignedPermissions = {"global_permissions": [], "object_permissions": []}
+        for permission_backend in registry.permission_backends:
```
ogenstad, line 24: Not a blocker for this PR, just highlighting that we should get rid of this `registry.permission_backends` and find a better place to store any backends than the registry.

backend/infrahub/permissions/manager.py
```diff
@@ -1,116 +1,57 @@
...
-        if not permission.decision & PermissionDecisionFlag.ALLOW_ALL:
-            specificity += 1
-        return specificity
+        self.resolver = resolver
+
+    @classmethod
+    async def load_for_account(cls, db: InfrahubDatabase, branch: Branch, account_session: AccountSession) -> Self:
+        loader = PermissionLoader(account_session=account_session)
+        permissions = await loader.load(db=db, branch=branch)
+        resolver = PermissionResolver(permissions=permissions, default_branch_name=registry.default_branch)
```
ogenstad: Similar comment as above, not something that has to be done here. But I think we should move away from using the registry directly within methods of our classes. On this case the PermissionManager class could have a default_branch parameter in `__init__()`, that way we can transition away from depending on the registry.

ogenstad (approved):
Nice!

backend/infrahub/permissions/manager.py
gmazoyer: The `PermissionManager` does not really use the `default_branch`, it just passes it to the `PermissionResolver`, so it could be misleading to have it as a parameter of the `__init__()`. However, we could add it as a parameter of `load_for_account()` and update the callers (there are only 2 of them).

backend/infrahub/permissions/loader.py
gmazoyer: Agreed

backend/infrahub/permissions/manager.py
gmazoyer: I agree that would be much more cleaner like so. I believe we can achieve this, there would be one drawback though. I think we'd have to make methods like `has_permission()`, `raise_for_permission()` (and more) async, as they would require  lazy loading. It's still possible, we'd have to update the callers.

Could be you have something else in mind?

ajtmccarty: yes I think you're right. It would have to be an async method. it should only need to actually wait the first time the permissions are loaded during the request though

gmazoyer: As it would touch a lot of code (not difficult task though). I created a task to track this and do it in a follow-up PR.

Issue #8984, opened 2026-04-20
bug: can't scroll artifact preview


BaptisteGi:
### Component

Frontend UI

### Infrahub version

1.8.5

### Current Behavior

<img width="2610" height="1522" alt="Image" src="https://github.com/user-attachments/assets/6989ad6a-a6d0-40e0-a959-745cfce990b9" />

I have a relatively long SVG artifact (rack elevation) and somehow the artifact preview feature doesn't allow me to scroll. Thus I can only see the first half of my artifact.

### Expected Behavior

I'm expecting a scroll bar in the frontend component

### Steps to Reproduce

Follow https://docs.infrahub.app/demo-dc/install and then https://docs.infrahub.app/demo-dc/user-walkthrough
This should populate a couple of rack, then by looking at the artifact attached to the rack

### Additional Information

Seems like only MD and SVG artifacts are impacted

PR #8989, opened 2026-04-20
Fix SVG artifact preview scroll and download 

pa-lem:
issue: https://github.com/opsmill/infrahub/issues/8984

image

## Summary
- Wrap oversized SVG artifact previews in a scroll area so content is no longer clipped
- Fix SVG/text artifact downloads that produced the Vite index.html instead of the raw content
- Allow blob: / data: URLs to pass through ReactAriaRouterProvider so native download works
- Use client-side blob URLs for text artifacts (SVG, JSON, YAML, etc.) so `download` triggers a save across dev and prod

---
## Summary by cubic
Make SVG artifact previews horizontally scrollable so oversized SVGs aren’t clipped and can be viewed at full size. Fix artifact downloads by allowing `blob:`/`data:` URLs in `ReactAriaRouterProvider` and using client-side Blob URLs for text files while keeping direct links for binaries; addresses IFC-2483.

<sup>Written for commit 815aa806dbfd579ac0b78d458dd33c25de872b40. Summary will update on new commits.</sup>

### DIFF
--- a/frontend/app/src/shared/components/data-viewer/data-viewer.tsx
+++ b/frontend/app/src/shared/components/data-viewer/data-viewer.tsx
@@ -51,7 +51,7 @@ function DataViewerContent({
 
     case "image/svg+xml": {
       return (
-        <ScrollArea scrollX className="bg-white" scrollBarClassName="bg-transparent">
+        <ScrollArea scrollX className="bg-white rounded-lg" scrollBarClassName="bg-transparent">
           <Svg value={content} className="mx-auto" />
         </ScrollArea>
       );
--- a/frontend/app/src/shared/components/data-viewer/data-viewer.tsx
+++ b/frontend/app/src/shared/components/data-viewer/data-viewer.tsx
@@ -51,11 +51,7 @@ function DataViewerContent({
 
     case "image/svg+xml": {
       return (
-        <ScrollArea
-          scrollX
-          className="grow rounded-lg border border-neutral-700 bg-white shadow-sm"
-          scrollBarClassName="bg-transparent"
-        >
+        <ScrollArea scrollX className="rounded-lg bg-white" scrollBarClassName="bg-transparent">
           <div className="p-4">
             <Svg value={content} />
           </div>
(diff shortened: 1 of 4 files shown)


frontend/app/src/shared/components/data-viewer/data-viewer.tsx
```diff
@@ -51,7 +51,15 @@ function DataViewerContent({
...
       return (
-        <Svg value={content} className="grow rounded-lg border border-neutral-700 shadow-sm" />
+        <ScrollArea
+          scrollX
+          className="grow rounded-lg border border-neutral-700 bg-white shadow-sm"
+          scrollBarClassName="bg-transparent"
+        >
+          <div className="p-4">
+            <Svg value={content} />
+          </div>
```
bilalabbad: we do not need div here right? `p-4` can be className is <Svg/>?

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
```diff
@@ -51,7 +51,15 @@ function DataViewerContent({
 
     case "image/svg+xml": {
       return (
-        <Svg value={content} className="grow rounded-lg border border-neutral-700 shadow-sm" />
+        <ScrollArea
+          scrollX
+          className="grow rounded-lg border border-neutral-700 bg-white shadow-sm"
```
bilalabbad: I'm not sure if we need all these classes. The only one questionable is `rounded-lg bg-white`. The DataViewer already has a dark (bg-neutral-800) background. A white background creates contrast against the rest of the viewer.

- White: safer if we consider most SVGs use dark strokes/fills
- Dark: (inherit container bg-neutral-800): more consistent with the viewer theme clearly making svg background transparent

bilalabbad (changes requested):

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
pa-lem: without background it feels off but it also depends on the generated svg

image

pa-lem: to me we should keep white backgrounds, svg assume most of the time that it will be light and the background could be transparent

minitriga (approved):

bilalabbad (approved):

PR #8991, opened 2026-04-20
fix: reject generic deletions referenced by inherit_from and roll back failed schema updates (closes #8988)

### LINKED ISSUE #8988: bug: Schema delete of a generic with live `inherit_from` references is accepted, partially commits to Neo4j, crashes in `process_inheritance`, and permanently wedges the branch.
### Component

_No response_

### Infrahub version

1.8.5

### Current Behavior

Triggered by the following two-step sequence (each step a separate `POST /api/schema/load` — either two `infrahubctl schema load` calls or two commits via git integration):

- **Step 1**: apply a schema that sets `state: absent` on every `peer`-kind relationship pointing at a generic. This clears all `peer` references to the generic. `inherit_from` references from other nodes are untouched. Apply succeeds.
- **Step 2**: apply a schema that sets `state: absent` on the generic itself.

On step 2:

1. The pre-write validator checks `peer` references. After step 1 there are none, so validation passes. **`inherit_from` references are not checked.**
2. The delete is persisted to Neo4j.
3. Post-write, `schema.process() → process_inheritance()` at `backend/infrahub/core/schema/schema_branch.py:1795` raises `ValueError: <Node> Unable to find the generic <Generic>` because the inheriting node still carries an `inherit_from` edge to the just-deleted generic.
4. The client receives HTTP **500 Internal Server Error**.
5. The delete is **not rolled back**. The schema graph is now corrupted.
6. Every subsequent call to `POST /api/schema/load` on the branch — including re-declaring the generic with or without the original UUID — fails with:

   ```
   Unable to find the Schema associated with <original-UUID>, <Generic>
   ```

The branch cannot be unwedged through the public API.

Note: if the same two mutations are applied **in a single `POST /api/schema/load`** (peer removal and generic deletion together in one file), the validator catches the inconsistency and rejects it cleanly with `<Node>: Relationship '<rel>' is referring an invalid peer '<Generic>'`. The bug only manifests when the steps are split across two separate applies — which is the normal case for git integration, where each commit with schema changes is its own apply.

### Expected Behavior

On step 2 above, the delete validator should symmetrically check `inherit_from` references alongside `peer` references and reject the operation cleanly before any DB write, with a message like:

```
<InheritingNode>: Generic '<Generic>' cannot be deleted while referenced via inherit_from
```

The schema graph should remain unchanged.

### Steps to Reproduce

A fresh Infrahub 1.8.5 instance with no prior data. Three YAML files, three `infrahubctl schema load` calls. A trivial three-kind schema is sufficient.

### 1. Load the baseline schema — `baseline.yml`

```yaml
---
version: "1.0"

generics:
  - name: Parent
    namespace: Bugrepo
    description: "Generic that will be deleted. Has inheritors."
    human_friendly_id: ["rule_id__value"]
    display_label: "{{ rule_id__value }}"
    include_in_menu: false
    order_by:
      - rule_id__value
    attributes:
      - name: rule_id
        kind: Number
        optional: true

  - name: Device
    namespace: Bugrepo
    description: "Generic holding a peer relationship to BugrepoParent."
    include_in_menu: false
    attributes:
      - name: name
        kind: Text
    relationships:
      - name: rules
        peer: BugrepoParent
        kind: Generic
        cardinality: many
        optional: true

nodes:
  - name: Child
    namespace: Bugrepo
    description: "Node that inherits from BugrepoParent. Will be orphaned."
    inherit_from:
      - BugrepoParent
    include_in_menu: false
```

```bash
infrahubctl schema load baseline.yml   # succeeds
```

### 2. Remove the `peer` reference from `BugrepoDevice` — `step1.yml`

```yaml
---
version: "1.0"
generics:
  - name: Device
    namespace: Bugrepo
    relationships:
      - name: rules
        peer: BugrepoParent
        state: absent
```

```bash
infrahubctl schema load step1.yml   # succeeds
```

After this, `BugrepoParent` has no `peer:` references. It still has one `inherit_from:` reference from `BugrepoChild`.

### 3. Delete the generic — `step2.yml`

```yaml
---
version: "1.0"
generics:
  - name: Parent
    namespace: Bugrepo
    state: absent
```

```bash
infrahubctl schema load step2.yml
```

**Returns HTTP 500 Internal Server Error.** Server traceback:

```
File "/source/backend/infrahub/api/schema.py", line 392, in load_schema
    updated_hash = await coordinator.execute(...)
File "/source/backend/infrahub/core/schema/update_coordinator.py", line 167, in execute
    updated_hash = await self._update_schema(...)
File "/source/backend/infrahub/core/schema/update_coordinator.py", line 236, in _update_schema
    await self.schema_manager.update_schema_branch(...)
File "/source/backend/infrahub/core/schema/manager.py", line 217, in update_schema_branch
    updated_schema = await self.load_schema_from_db(...)
File "/source/backend/infrahub/core/schema/manager.py", line 837, in load_schema_from_db
    schema.process(validate_schema=validate_schema)
File "/source/backend/infrahub/core/schema/schema_branch.py", line 624, in process
    self.process_pre_validation()
File "/source/backend/infrahub/core/schema/schema_branch.py", line 645, in process_pre_validation
    self.process_inheritance()
File "/source/backend/infrahub/core/schema/schema_branch.py", line 1795, in process_inheritance
    raise ValueError(f"{node.kind} Unable to find the generic {generic_kind}")
ValueError: BugrepoChild Unable to find the generic BugrepoParent
```

### 4. Observe persistent corruption

Re-running the baseline load (or any other schema operation):

```bash
infrahubctl schema load baseline.yml
```

```
Unable to load the schema:
  Unable to find the Schema associated with <original-UUID>, BugrepoParent
```

The branch is now unrecoverable through `infrahubctl` / `/api/schema/load`.

### Root cause analysis (from traces)

Two interacting defects:

1. **Incomplete reference check on delete.** The delete-path validator checks `peer` references but does not check `inherit_from` references. Once peer references are cleared (as in step 1), the generic appears deletable and passes pre-write validation.
2. **Non-atomic write + post-processing.** In `update_schema_branch → load_schema_from_db`, the Neo4j delete persists before `schema.process()` re-validates the fully-loaded graph. When `process_inheritance` raises, the delete is already committed. The error surfaces as a 500, but orphaned `inherit_from` edges remain in the graph referencing the deleted generic's UUID.

### Recovery attempts that all fail through the public API

Tried against the corrupted branch from a reproducible instance:

| Attempt | Result |
|---|---|
| Re-declare the generic with its original UUID via `id:` in YAML | `Unable to find the Schema associated with <UUID>, <Generic>` |
| Re-declare the generic without any UUID | Same error |
| `state: absent` on the orphaned inheriting node in isolation | Rejected if any peer still references the inheritor |
| `state: absent` on the orphaned inheritor + cleanup of peer relationships referencing it, all in one apply | `Node-level 'inherit_from' constraint violation on schema 'SchemaNode'` |

### Proposed fix direction

1. Extend the delete-path validator to enumerate `inherit_from` references symmetrically with `peer` references and reject deletes with live inheritors. Clear error, no DB touch.
2. Wrap the Neo4j write + `schema.process()` call inside `update_schema_branch` in a single transaction so validation failures in `process_inheritance` roll back the delete.
3. Optionally ship an `infrahubctl schema repair` or Cypher-level recovery path for databases already in this state.

### Originating customer scenario

A shared dev instance hit this while attempting a migration — changing an attribute's `kind` from `Number` → `NumberPool`, which requires delete-and-re-add since `kind` is not editable in place. The path:

1. Pushed a commit changing `kind: Number` → `kind: NumberPool`. Silently rejected as invalid (`kind` cannot change in place). **No UI indication that the schema push failed**, which is its own usability issue worth filing separately.
2. Pushed a commit reverting `kind: NumberPool` → `kind: Number`, and in the same commit adding `state: absent` on the peer relationship pointing to the generic. Applied cleanly.
3. Pushed a commit with `state: absent` on the generic. Hit this bug.

Git integration amplifies the exposure because every commit is a separate schema apply — users naturally produce the two-step sequence that triggers the bug across two commits.

### Environment

- Infrahub: **1.8.5**
- Helm chart: `opsmill/infrahub` 4.21.0 (appVersion overridden to 1.8.5 via `global.infrahubTag`)
- Dependencies: Neo4j 2025.10.1-4 community single-node, Redis 19.5.2, RabbitMQ 14.4.1, Prefect 3.6.8
- Cluster: KIND 0.31.0, Kubernetes 1.35.0
- Python runtime (from image): 3.13
- Trigger: two sequential `infrahubctl schema load` calls on `main` in a fresh deploy

Reproduction artifacts (values override, schemas, server log tail) available on request.

polmichel:
## Why

Deleting a generic schema while another node still declared `inherit_from` against it was silently accepted by `validate_node_deletions`. The deletion committed to Neo4j, then `schema.process()` crashed inside `process_inheritance` because the inheriting node's edge pointed to a missing generic. `SchemaUpdateCoordinator.execute()` had no rollback for that failure path, so the branch stayed wedged with every subsequent request hitting HTTP 500.

Goal: close the validator gap so the bad deletion is rejected before any DB write, and add a rollback guardrail so an exception during `_update_schema` cleans up the Neo4j write instead of leaving the branch corrupted.

Closes #8988

## What changed

- The validation process related to node deletions now also walks each node's `inherit_from` list and raises `ValueError` when a removed kind is still referenced.
- Any exception raised during schema update before running database migrations are not detected and creates a rollback process.

## How to review

Two tests for two fixes that can be reviewed separately.

## How to test

```bash
uv run pytest backend/tests/component/core/schema_manager/test_manager_schema.py::test_schema_branch_validate_node_deletion_inherit_from backend/tests/component/core/schema_manager/test_schema_rollback.py::TestSchemaUpdateAndRollback::test_schema_update_rolls_back_on_post_write_process_failure -v
```

## Impact & rollout

- **Deployment notes:** Safe to deploy. Branches corrupted by the pre-fix behavior still need manual Cypher cleanup (delete the orphaned `inherit_from` edge from the inheriting node to the deleted generic's UUID).

## Checklist

- [x] Tests added/updated
- [ ] Internal .md docs updated (internal knowledge and AI code tools knowledge) — N/A


backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -2820,6 +2820,150 @@ async def test_schema_branch_validate_node_deletion(
...
+    assert "TestingParent" in diff.removed
+
+    with pytest.raises(ValueError, match="'TestingParent' has been removed but is still referenced"):
+        schema_branch.validate_node_deletions(diff=diff)
+
+
+async def test_schema_branch_validate_node_deletion_iterates_generics_safely(
+    db: InfrahubDatabase, reset_registry: None, default_branch: Branch, register_internal_models_schema: SchemaBranch
+) -> None:
+    """validate_node_deletions must iterate every schema kind without touching inherit_from on generics."""
```
ajtmccarty, line 2869: I think these tests should set `state=ABSENT` in the schema definition and then call `SchemaBranch,process()`. yes it will take longer than just calling this one validation, but the schema updating is somewhat brittle, so running the whole `process()` pipeline is worth it

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -2820,6 +2820,150 @@ async def test_schema_branch_validate_node_deletion(
         schema_branch.validate_node_deletions(diff=diff)
 
 
+async def test_schema_branch_validate_node_deletion_inherit_from(
```
ajtmccarty, line 2823: can you move these tests to a new file? this test file is already way too big

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -2820,6 +2820,150 @@ async def test_schema_branch_validate_node_deletion(
...
+    schema_branch.validate_node_deletions(diff=diff)
+
+
+async def test_schema_branch_validate_node_deletion_removes_generic_and_child_together(
+    db: InfrahubDatabase, reset_registry: None, default_branch: Branch, register_internal_models_schema: SchemaBranch
+) -> None:
+    """Removing a generic AND the nodes that inherit from it in the same diff must be accepted.
+
+    The inherit_from reference is harmless because the referencing node is also being deleted.
+    Mirrors the three-schema drop in test_numberpool_assignment_from_generic.
```
ajtmccarty, line 2920: doesn't seem necessary to describe the implementation details or another test

ajtmccarty (approved):
just a few test comments

backend/tests/component/core/schema_manager/test_validate_node_deletion.py
```diff
@@ -48,18 +49,15 @@ async def test_schema_branch_validate_node_deletion(
...
-    broken_schema = SchemaRoot(**FULL_SCHEMA)
-    broken_schema_branch = SchemaBranch(cache={}, name="test-broken")
-    broken_schema_branch.load_schema(schema=broken_schema)
-
-    diff = schema_branch.diff(other=broken_schema_branch)
-    assert "TestingTag" in diff.removed
-
-    with pytest.raises(ValueError, match="'TestingTag' has been removed but is still referenced"):
-        schema_branch.validate_node_deletions(diff=diff)
+    with pytest.raises(ValueError, match="Relationship 'tags' is referring an invalid peer 'TestingTag'"):
```
polmichel, line 59: As the path has changes, so is the error message. But this path was not initially reachable due to `validate_kinds`

backend/tests/component/core/schema_manager/test_schema_rollback.py
```diff
@@ -204,3 +204,90 @@ async def test_schema_update_with_migrations_and_rollback(
...
+                    "state": "absent",
+                },
+            ],
+        }
+        candidate_schema = original_schema_branch.duplicate()
+        candidate_schema.load_schema(schema=SchemaRoot(**absent_generic_dict))
+        candidate_schema.process()
+
+        diff = original_schema_branch.diff(other=candidate_schema)
+        assert "TestingParent" in diff.removed
```
polmichel, line 261: This is not the same path than the raised issue. But as we have fixed the issue, this patch is not reachable anymore. So we are tweaking the path in order to assert that the same step will rollback if another failure occurs.

PR #9004, opened 2026-04-21
docs: restricted namespace topic

BaptisteGi:
## Summary by cubic
Adds a "Restricting inheritance by namespace" section to schema docs explaining the `restricted_namespaces` property on generics and how it enforces allowed namespaces during validation. Includes a YAML example, notes that the field is optional, and clarifies built-in generics (`CoreGenericRepository`, `CoreValidator`, `CoreCheck`, `CoreWebhook`, `CoreComment`, `CoreThread`) that require the `Core` namespace.

<sup>Written for commit 9d7bcdf6f6c296591ebfef75b50c114926f9f243. Summary will update on new commits.</sup>

### DIFF
--- a/docs/docs/topics/schema.mdx
+++ b/docs/docs/topics/schema.mdx
@@ -800,6 +800,52 @@ nodes:
       - "InfraAsset"
 ```
 
+#### Restricting inheritance by namespace
+
+The `restricted_namespaces` property on a generic limits which namespaces are allowed to define nodes that inherit from it. When set, any node with `inherit_from` referencing that generic must belong to one of the listed namespaces — schema validation fails otherwise.
+
+This is useful when a generic is designed to be implemented only by a specific, known set of nodes. Without the restriction, the generic appears available to anyone, which can mislead schema authors into creating node types that look valid but behave incorrectly at runtime.
+
+```yaml
+generics:
+  - name: GenericService
+    namespace: Customer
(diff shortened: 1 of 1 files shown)


docs/docs/topics/schema.mdx
```diff
@@ -800,6 +800,52 @@ nodes:
...
+    inherit_from:
+      - CustomerGenericService
+
+```
+
+:::info
+`restricted_namespaces` is optional. When not set on a generic, nodes from any namespace can inherit from it.
+:::
+
+Infrahub uses this same mechanism on several of its own built-in generics — for example `CoreGenericRepository`, `CoreValidator`, `CoreCheck`, `CoreWebhook`, `CoreComment`, and `CoreThread` — to prevent accidental extension of types that rely on specific internal code. Nodes inheriting from these generics must belong to the `Core` namespace.
```
polmichel: Currently, only `CoreWebhook` and `CoreGenericRepository` generic schemas have the namespace restriction.

BaptisteGi:
```suggestion
Infrahub uses this same mechanism on several of its own built-in generics — for example `CoreGenericRepository` and `CoreWebhook` — to prevent accidental extension of types that rely on specific internal code. Nodes inheriting from these generics must belong to the `Core` namespace.
```

PR #9023, opened 2026-04-22
test: add regression coverage for login/logout activity events [INFP-…

PhillSimonds:
# Why

INFP-474 added `infrahub.account.logged_in` / `infrahub.account.logged_out` activity events. The shipped implementation is covered by two happy-path component tests and several unit tests on the event class itself, but several contracts introduced by the ticket aren't directly guarded by the regression suite — notably the admin-only GraphQL visibility, the fire-and-forget emission contract, and the negative paths (API-key auth, refresh, failed login) that the spec explicitly calls out as not emitting events. This PR adds regression coverage for those contracts so future refactors can't silently regress them.

Related: INFP-474

## What changed

**Behavioral changes:** None — this PR is test-only.

**New tests**

- `backend/tests/unit/api/test_event_builder.py` (4 unit tests)
  - OAuth2 and OIDC login events populate `identity_source`.
  - `client_ip` on login and logout events reads `request.client.host` — `X-Forwarded-For` is not consulted (documents current behavior; if `ProxyHeadersMiddleware` is ever wired up, these tests will need updating alongside deployment docs).
- `backend/tests/component/api/test_auth_events.py` (5 new component tests alongside the existing two)
  - Session correlation: `logged_out.session_id` matches the preceding `logged_in.session_id`.
  - API-key authentication does not emit `AccountLoggedInEvent` (spec FR-008).
  - `/api/auth/refresh` does not emit `AccountLoggedInEvent`.
  - Failed login attempts emit nothing.
  - Login returns 200 even when the event service raises (fire-and-forget contract).
- `backend/tests/component/api/test_event_visibility.py` (3 new component tests)
  - Non-admin filtering on `infrahub.account.*` gets `PermissionDeniedError`.
  - Non-admin unfiltered queries have `infrahub.account.` silently added to `exclude_prefix`.
  - Admin can pass both checks through to the Prefect query layer.

**What stayed the same**

- No production code touched.
- No schema, API, or configuration changes.
- Existing `test_auth_events.py` happy-path tests are unchanged.

## How to review

Suggested order:

1. `backend/tests/component/api/test_event_visibility.py` — new file covering the `MANAGE_ACCOUNTS` gate in `backend/infrahub/graphql/queries/event.py`.
2. `backend/tests/component/api/test_auth_events.py` — additions to the existing file, each test maps to a single contract.
3. `backend/tests/unit/api/test_event_builder.py` — builder-level coverage reused by all three auth endpoints.

The XFF tests in `test_event_builder.py` document current behavior, not desired behavior — see the follow-up question on the ticket about whether `ProxyHeadersMiddleware` should be configured for accurate audit logging behind reverse proxies.

## How to test

```bash
uv run pytest \
  backend/tests/unit/api/test_event_builder.py \
  backend/tests/component/api/test_auth_events.py \
  backend/tests/component/api/test_event_visibility.py -v
```

Expected: 14 passed.

## Impact & rollout

- **Backward compatibility:** no production changes.
- **Performance:** component tests add ~30s to the `test-component` suite.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/+infp-474-regression-tests.housekeeping.md`)
- [ ] External docs updated (N/A — test-only)
- [ ] Internal .md docs updated (N/A — test-only)


dgarros:
Thanks @PhillSimonds 
I think this PR needs to target the branch `release-1.9` also for new features that haven't shipped into an official release, we don't need to add a newsfragment for each PR. The feature will be included in the release note

solababs (approved):
LGTM
The docs at the top of the tests files might be misleading overtime (to human and AI) as more tests are added, you may consider removing them as the tests clearly state what they do
Remember also to change the base branch to release-1.9

ajtmccarty (changes requested):
looks like some commits from develop have been pulled into this one

ogenstad:
@PhillSimonds, as it is now this should either target the `stable` or `develop` branch. Could also be nice with a rebase.

PR #9027, opened 2026-04-23
fix 4 diff bugs around HAS_SOURCE and kind migrations

ajtmccarty:
fixes three small bugs found while working on changes to the merge query
1. failure while calculating the diff if an object included a lone property update on a relationship of a node that had its kind/inheritance migrated on the user's branch
2. diff could pick the wrong peer for a property update on a relationship if one of the peers had its kind/inheritance migrated on the default branch after the user's branch forked
3. diff could fail to identify a conflict if the source of an attribute or relationship were updated to different values on the user's branch and the default branch after the user's branch forked
4. diff could fail to include a source updated on a user's branch if the owning object had it's kind/inheritance migrated on the default branch after the user's branch forked

fixes are small changes to one of the cypher queries used for diff calculation and the class that reads and parses the rows returned by the diff calculation queries


backend/tests/component/core/diff/test_diff_calculator.py
```diff
@@ -4655,3 +4656,180 @@ async def test_diff_attribute_single_source_property_change(
         assert source_prop.new_value == person_jane_main.get_id()
     else:
         assert source_prop.new_value is None
+
+
+async def test_relationship_property_added_on_source_branch_kind_migration(
```
polmichel: Just an optional comment, we can look into it another time: 
At some point would it be possible to split these tests? Either split into different files or at least into different classes so we can launch the different tests "by category" ? 
I have noticed that each of this test is individually quite long, but there are 47 of them there.

polmichel (approved):

backend/tests/component/core/diff/test_diff_calculator.py
ajtmccarty: yeah I was thinking the same thing while working on it. you convinced me to split it up into multiple test files within `diff/diff_calculator/`

PR #9028, opened 2026-04-23
Fix generics links in schema-visualizer and use default schema icon

pa-lem:
image


pa-lem:
done here https://github.com/opsmill/infrahub/pull/9035

PR #9037, opened 2026-04-23
fix(git): fetch from remote when worktree commit is missing locally

BeArchiTek:
## Why

Task runs (artifact generation, transforms, ...) fail with `CommitNotFoundError` whenever the commit SHA recorded in the database hasn't been fetched yet to the local git clone on the worker picking up the task. Other paths like `get_commit_value` and `update_latest_commit` already fetch before they touch a commit; `create_commit_worktree` did not. In multi-worker deployments where fetch timing varies per worker this can cause issues for recently pushed commits.

## What

`InfrahubRepositoryBase.create_commit_worktree` now treats a `fatal: invalid reference` from git worktree add as a cue to fetch from origin and try again, mirroring the pattern used elsewhere. If the repo has no origin configured we don't bother fetching and surface a `RepositoryError` that explains the configuration problem.

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -705,30 +705,18 @@ class InfrahubRepositoryBase(BaseModel, ABC):
                 raise RepositoryError(identifier=self.name, message=exc.stderr) from exc
 
             if not self.has_origin:
-                raise CommitNotFoundError(
+                raise RepositoryError(
                     identifier=self.name,
-                    commit=commit,
+                    message=f"Commit {commit} not found and no remote origin configured to fetch from.",
                 ) from exc
 
             # Commit may exist on the remote but hasn't been fetched to this worker yet
             log.info(f"Commit {commit} not found locally, fetching from remote.", repository=self.name)
-            try:
-                repo.remotes.origin.fetch()
-            except GitCommandError:
-                raise CommitNotFoundError(
-                    identifier=self.name,
-                    commit=commit,
-                ) from exc
+            repo.remotes.origin.fetch()
 
-            try:
-                repo.git.worktree("add", directory, commit)
-                log.debug(f"Commit worktree created {commit} after fetch", repository=self.name)
-                return worktree
-            except GitCommandError:
-                raise CommitNotFoundError(
-                    identifier=self.name,
-                    commit=commit,
-                ) from exc
+            repo.git.worktree("add", directory, commit)
+            log.debug(f"Commit worktree created {commit} after fetch", repository=self.name)
+            return worktree
 
     def create_branch_worktree(self, branch_name: str, branch_id: str) -> bool:
         """Create a new worktree for a given branch."""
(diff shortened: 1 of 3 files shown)


backend/infrahub/git/base.py
```diff
@@ -679,12 +679,34 @@ def create_commit_worktree(self, commit: str) -> bool | Worktree:
             log.debug(f"Commit worktree created {commit}", repository=self.name)
             return worktree
         except GitCommandError as exc:
-            if "invalid reference" in exc.stderr:
+            if "invalid reference" not in exc.stderr:
+                raise RepositoryError(identifier=self.name, message=exc.stderr) from exc
+
+            if not self.has_origin:
+                raise CommitNotFoundError(
```
ajtmccarty: is it really `CommitNotFoundError` if the repo has no origin defined?

backend/infrahub/git/base.py
```diff
@@ -679,12 +679,34 @@ def create_commit_worktree(self, commit: str) -> bool | Worktree:
...
+                    identifier=self.name,
+                    commit=commit,
+                ) from exc
+
+            # Commit may exist on the remote but hasn't been fetched to this worker yet
+            log.info(f"Commit {commit} not found locally, fetching from remote.", repository=self.name)
+            try:
+                repo.remotes.origin.fetch()
+            except GitCommandError:
+                raise CommitNotFoundError(
```
ajtmccarty: same question but for failing to fetch. can the try-except be removed and just allow the `fetch()` to raise its own error if it fails?

backend/infrahub/git/base.py
```diff
@@ -679,12 +679,34 @@ def create_commit_worktree(self, commit: str) -> bool | Worktree:
...
+                    identifier=self.name,
+                    commit=commit,
+                ) from exc
+
+            try:
+                repo.git.worktree("add", directory, commit)
+                log.debug(f"Commit worktree created {commit} after fetch", repository=self.name)
+                return worktree
+            except GitCommandError:
```
ajtmccarty: same error-raising issue here. can `worktree("add",...)` be allowed to raise its own error?

PhillSimonds:
Tested against a 1.9.0 task worker in a kind cluster

**Repro (unpatched):**
1. Pushed a new commit to gitea on a branch the worker hadn't fetched.
2. Confirmed the SHA was absent from the worker's local main clone (`git cat-file -e <sha>` → exit 1).
3. Called `InfrahubReadOnlyRepository.create_commit_worktree(<sha>)` directly on the on-disk clone.
4. Result: `CommitNotFoundError: Commit <sha> not found with GitRepository 'INFRAHUB-DEV'` — matches the customer stack trace.

**Patched:** hot-applied this PR's diff to `/source/backend/infrahub/git/base.py` in the worker pod and re-ran the reproducer with a fresh missing SHA. Logs show the new path executing end-to-end:

```
git worktree add ... <sha>     # fails: fatal: invalid reference
Commit <sha> not found locally, fetching from remote.
git fetch -v -- origin
git worktree add ... <sha>     # succeeds
Commit worktree created <sha> after fetch
RESULT: success, returned Worktree(...)
```

LGTM — fix resolves the reported failure mode.

ajtmccarty (approved):

PR #9041, opened 2026-04-23
Prefect client adapter: Spec-kit specifications

polmichel:
## Why

- Externalize Prefect related complexity in a closed package which will be the only one exposing Prefect-based functionality
- Prevent fragile mocking techniques from coder humans and agents
- Prevent flakiness as exposing dedicated waiting methods in the test adapter

### Technical goals

- All the prefect imports will be contained in the Prefect adapter module
- Ability to assert against a real-shaped Prefect surface without hand-rolled mocks
- Ability to assert deterministically wait for an emitted event to be observable

### Downsides

- Additional maintenance regarding Prefect updates
- Progressive refactoring

### Schemas
image
image

### Keys entities

    • EventNotObservedError
      Raised by wait_for_event at timeout. Fields:
        - name: str
        - correlation_id: UUID | None
        - elapsed_seconds: float
        - reason: EventNotObservedReason

    • EventNotObservedReason (StrEnum)
        - NOT_EMITTED       — the in-memory test adapter observed no matching
                              event in its visible window (definitive).
        - NOT_OBSERVABLE    — the real test adapter's poll timed out; the event
                              may have been emitted but not yet indexed by
                              Prefect's events pipeline.

    Inheritance & composition summary
    ─────────────────────────────────

      PrefectClientAdapter (ABC, production port)
       ├── PrefectClientTestAdapter (ABC, test port — adds wait_for_event)
       │    ├── RealPrefectClientTestAdapter   (also inherits RealPrefectClientAdapter)
       │    └── InMemoryPrefectClientTestAdapter
       └── RealPrefectClientAdapter

    Wiring matrix
    ─────────────

    ┌──────────────────────────────────────────────┬─────────────────────────────────────────┬───────────────────────────────┐
    │ Context                                      │ Concrete class                          │ Type-hinted as                │
    ├──────────────────────────────────────────────┼─────────────────────────────────────────┼───────────────────────────────┤
    │ Production code                              │ RealPrefectClientAdapter                │ PrefectClientAdapter          │
    │ Integration contract lane + functional tests │ RealPrefectClientTestAdapter            │ PrefectClientTestAdapter      │
    │ Unit contract lane + unit tests using fake   │ InMemoryPrefectClientTestAdapter        │ PrefectClientTestAdapter      │
    └──────────────────────────────────────────────┴─────────────────────────────────────────┴───────────────────────────────┘

    Invariant: production code hints against PrefectClientAdapter (the production
    port). The test port's wait_for_event is not reachable from production code
    by construction — the boundary is enforced by the type system, not by
    convention or lint rules.

#### Contract Test

A test parametrized over both adapter implementations that asserts observable behavior is equivalent. The mechanism that prevents fake/real drift.

#### ID

Match events by Prefect's **native `Event.id`** (UUID), not by a custom `correlation_id` token stamped into `payload`. The production port stays identical to Prefect's API; all test-only concepts live on the test port.
- Already caller-providable
- Server-side filterable: Prefect 3.6.13's `EventFilter` exposes a first-class `id: EventIDFilter` field, so the real adapter filters at the API rather than client-side.
- No production-port leak

#### Test port primitives (three, all on `PrefectClientTestAdapter`)

| Primitive | Purpose |
|---|---|
| `checkpoint() -> Checkpoint` | Sync watermark. Pair with `wait_for_event(since=...)` for the simple "next event of name X after my action" case. |
| `captured_emits()` | Async context manager. Records every emission inside its `async with` block; tests pull `Event.id`s out for strict per-emission matching. |
| `wait_for_event(*, name=None, event_id=None, since=None, timeout_seconds=10.0)` | Bounded async wait. At least one of `name` / `event_id` required. |

- `checkpoint()` keeps the simple 1:1 case as cheap as the legacy helper (one extra line, no indentation).
- `captured_emits()` handles multi-emit-per-action cases without forcing every test through an `async with` block.

Tests pick the lighter tool per case.


polmichel:
To be repoened later

PR #9054, opened 2026-04-25
add support for all metadata to get_many(prefetch_relationships=True)

ajtmccarty:
`NodeManager.get_many()` with `prefetch_relationships=True` did not support all the types of metadata we have on our Node objects. Specifically, it did not support timestamps (updated/created_at/by) or is_protected.

should not affect performance b/c the new metadata types have to be affirmatively requested

this PR adds support for those metadata options to the method.


polmichel (approved):
LGTM!

PR #9055, opened 2026-04-25
Add new GraphQL query InfrahubGraphQLQueryReport

ogenstad:
# Why

New GraphQL query to make it easier for end users to figure out how the GraphQLQueryAnalyzer views a query (for artifact regeneration).

This will later be used by [infrahubctl graphql query-report](https://github.com/opsmill/infrahub-sdk-python/pull/976) (note that PR works together with this mutation but the PR itself hasn't been reviewed by me it's just some generated code which I'll get back to once this is merged to stable. Then after the next patch release of Infrahub I'll get back to the SDK PR and create some integration tests for it as well and potentially clean up that code. Once that is done I'll also remove the SpecKit files generated here.

## What changed

* Relevant changes are in backend/infrahub/graphql/queries/graphql_query_report.py, where we add a new GraphQL query called InfrahubGraphQLQueryReport, it produces a report based on the contents of a GraphQL query, as it is now the only field is `targets_unique_nodes` which indicates if the query can be used for artifact definitions and able to target a single unique base node. (In the future we can add stuff like models being queried along with attributes and relationships) Perhaps the permissions that would be required to run the query. The types of relationships it traverses, if fragments etc are used and the complexity of the query. But for now it's only a single field.
* Added a number of tests for this new mutation.


backend/tests/component/graphql/queries/test_graphql_query_report.py
```diff
@@ -0,0 +1,141 @@
...
+    ),
+]
+
+
+async def test_targets_unique_nodes(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    car_person_schema: SchemaBranch,
+) -> None:
+    assert UNIQUE_TARGETS_TEST_CASES, "No test cases defined for unique targets test"
```
ogenstad, line 82: The reason we have this here is to avoid test setups between each test case.

backend/infrahub/graphql/queries/graphql_query_report.py
```diff
@@ -0,0 +1,58 @@
...
+            "True if every operation in the submitted query resolves to uniquely identifiable nodes "
+            "(via a required ids argument or a required field matching the model uniqueness constraints). "
+            "When true, Infrahub limits artifact regeneration to only the nodes that changed. "
+            "When false, all artifacts for the definition are regenerated on any relevant node change."
+        ),
+    )
+
+
+async def resolve_graphql_query_report(
+    _root: None,
```
ogenstad, line 30: Note I tried a different approach here to avoid having to have `# noqa: ARG003` comments. This is an argument that Graphene adds that we never and up using. So something we could perhaps transfer to other places as well..

ajtmccarty (approved):

PR #9056, opened 2026-04-25
Follow up docs (ifc-2504-graphql-query-report) - For discussion

ogenstad:
# Why

Documentation following #9055, I'm not sure we want to have these though as I think it will be better without mentioning this query specifically and instead add the docs once the SDK has support for it through infrahubctl.

## What changed

* Documentation

## Stale PR

This will be picked up once https://github.com/opsmill/infrahub-sdk-python/pull/976 is merged and we have a release out in the wild.


ogenstad:
Replaced by #10293.

PR #9057, opened 2026-04-27
Upgrade SpecKit to version 0.8.1

ogenstad:
# Why

Stay up to date with current versions of SpecKit

## What changed

* Previously our "installation" didn't have a version, now it's properly tracked
* A lot of the previous commands got converted to skills instead
* Removed the custom OpsMill/Infrahub integrations from being nested into SpecKit itself, we now have `.specify/extensions/infrahub` instead.
* Updated configuration for `yamllint` to ignore yaml files coming from SpecKit.


dev/skills/speckit-specify/SKILL.md
```diff
@@ -36,36 +72,47 @@ Given that feature description, do this:
      - "Create a dashboard for analytics" → "analytics-dashboard"
      - "Fix payment processing timeout bug" → "fix-payment-timeout"
 
-2. **Determine the Jira/JPD reference and create the feature branch**:
```
ogenstad, line 39: These were custom Infrahub tasks that were moved to the Infrahub extension.

CLAUDE.md
```diff
@@ -1 +1,6 @@
-@AGENTS.md
```
fatih-acar, line 1: Should the speckit footer move to AGENTS.md instead?

fatih-acar (approved):

CLAUDE.md
ogenstad: @fatih-acar I was thinking about that as well, but left it there. My reasoning is that the SpecKit upgrade command would expect to find it within CLAUDE.md the next time we upgrade as well and I don't expect this to cause any problems if we leave it in the CLAUDE.md file. Any thoughts around that?

fatih-acar: I think the issue with CLAUDE.md is that it's not "agent agnostic"... if we want to keep compat with other agents we must use AGENTS.md

If so we should make the speckit upgrade command agnostic too, I don't know if it's possible

ogenstad: I'm not sure there's an actual value of having that `CLAUDE.md` file in the first place or if we can just symlink it.

.specify/presets/infrahub/commands/speckit.infrahub.feature.md
```diff
@@ -0,0 +1,70 @@
+---
+description: "Validate Jira/JPD ticket reference and create a feature branch"
```
polmichel, line 2: I wanted to try out this new extension but locally my Claude Code does not detect this new tool. I have taken a look at  https://github.com/github/spec-kit/blob/main/extensions/EXTENSION-USER-GUIDE.md#install-from-local-directory-development and it seems that the classic workflow would be to:

1. Create these new extensions in a dedicated folder while they are under development or being used for internal purposes (`dev/spec-kit/extensions/` ?).
2. Use the command line `specify extension add --dev dev/spec-kit/extensions/infrahub` to keep these extensions local. This preserves the spec-kit "internal" folder of being manually modified. This process will automatically create the file that this comment points to, and will also make some additional modifications:

- Create a custom skill at `infrahub/dev/skills/speckit-infrahub-feature/SKILL.md`.
- Update the internal registry with a new local extension (`.specify/integrations/.registry`).
These steps allow Claude to access the new Speckit extension. 

Image


Assuming you agree that this is the correct way to set up a new extension, would you find it more convenient to complete these steps in this PR? Alternatively, we could do it later on.

polmichel: And as I have the changes locally I can push them ofc

ogenstad: Please go ahead and push an update if you already have the code. I started this some time ago and updated my work again to the latest. Looks like I missed this :(

polmichel: I have pushed a preset that overrides the initial git extension `speckit.git.feature`

image

polmichel (approved):

PR #9058, opened 2026-04-27
Fix typing errors for protocols

solababs:
# Why

`CoreNode` in `protocols_base.py` was defined as a `@runtime_checkable Protocol`. Every generated class in `protocols.py` inherits from it, making them Protocol subclasses too. mypy's `type-abstract` rule rejects Protocol classes as `type[X]` arguments — the pattern used throughout the codebase (`kind=CoreNumberPool`, `schema=CoreAccountGroup`, etc.)

The prior workaround was a blanket `disable_error_code = ["type-abstract"]` in `pyproject.toml`. This PR removes the suppression and fixes the root cause.

## How to test

```bash
uv run invoke backend.mypy        # must pass with 0 errors, no type-abstract suppression
```

## Checklist

### DIFF
--- a/backend/infrahub/core/protocols_base.py
+++ b/backend/infrahub/core/protocols_base.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Any, ClassVar, Protocol, runtime_checkable
 
 from typing_extensions import Self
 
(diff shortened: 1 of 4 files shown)


backend/infrahub/core/protocols_base.py
```diff
@@ -69,15 +69,24 @@ def get_id_function_name(self) -> str: ...
     def to_database_id(self, db_id: str | int) -> str | int: ...
 
 
-@runtime_checkable
-class CoreNode(Protocol):
+class CoreNode:
+    _is_runtime_protocol: ClassVar[bool] = True
```
ajtmccarty: does this line actually do anything?

solababs: This has been removed

ajtmccarty (approved):

ogenstad (approved):

PR #9069, opened 2026-04-27
Update relationship manager typing fields

solababs:
# Why

`RelationshipManager` fields on protocol stubs were all typed as bare `RelationshipManager` with no peer information, so callers had to pass a redundant `peer_type=` argument to `get_peer`/`get_peers` to get a typed return value. mypy had no way to infer the peer type from the field declaration alone.

The goal is to make `RelationshipManager` generic so that the peer type flows from the field declaration — `transform.repository.get_peer(db=db)` returns `CoreGenericRepository` without any extra argument.

## How to test

```bash
# Confirm clean mypy
uv run invoke backend.mypy

# Confirm ruff is happy
uv run invoke backend.ruff

# Run unit tests
```


backend/infrahub/core/integrity/object_conflict/conflict_recorder.py
ajtmccarty: could the type checking be updated in the signature to be `CoreValidatorCheck` or whatever the peer of `validations` is ?

backend/infrahub/events/generator.py
```diff
@@ -99,7 +103,7 @@ async def generate_node_mutation_events(
         InfrahubKind.ARTIFACTTHREAD,
         InfrahubKind.FILETHREAD,
     ]:
-        proposed_change: CoreProposedChange = await node.change.get_peer(db=db, peer_type=CoreProposedChange)  # type: ignore[attr-defined]
+        proposed_change: CoreProposedChange = await node.change.get_peer(db=db)  # type: ignore[attr-defined]
```
ajtmccarty: can the `# type: ignore[attr-defined]` be removed?

backend/infrahub/menu/generator.py
```diff
@@ -4,21 +4,21 @@
...
     from infrahub.core.branch import Branch
+    from infrahub.core.protocols import CoreMenu, CoreMenuItem
     from infrahub.database import InfrahubDatabase
     from infrahub.permissions import PermissionManager
 
 log = get_logger()
 
 
-def get_full_name(obj: CoreMenuItem) -> str:
+def get_full_name(obj: CoreMenu) -> str:
```
ajtmccarty, line 21: so this was just wrong before it looks like
I'm a little surprised it was working

ajtmccarty (approved):
cool improvement. just a few little comments

Issue #9067, opened 2026-04-27
bug: Task details page fires one GraphQL query per related node, crashing browser on tasks with thousands of related nodes


fatih-acar:
### Component

Frontend UI

### Infrahub version

1.9 (develop)

### Current Behavior

The task details / task list pages render related nodes through the `<Id>` component at [frontend/app/src/shared/components/ui/id.tsx](frontend/app/src/shared/components/ui/id.tsx#L21-L25). Each instance of `<Id>` fires its own `useQuery(getObjectDisplayLabel)` to resolve the display label for a single node:

```tsx
export const Id = ({ id, kind = NODE_OBJECT, preventCopy, branch, date }: tId) => {
  const { loading, error, data } = useQuery(gql(getObjectDisplayLabel({ kind })), {
    variables: { ids: [id] },
    context: { uri: CONFIG.GRAPHQL_URL(branch, date) },
  });
  ...
```

Callers include the task UI which renders one `<Id>` per related node:

- [frontend/app/src/entities/tasks/ui/task-item-details.tsx:113](frontend/app/src/entities/tasks/ui/task-item-details.tsx#L113)
- [frontend/app/src/entities/tasks/ui/task-items.tsx](frontend/app/src/entities/tasks/ui/task-items.tsx)
- [frontend/app/src/entities/tasks/ui/tasks-homepage/task-homepage-item.tsx](frontend/app/src/entities/tasks/ui/tasks-homepage/task-homepage-item.tsx)

When a task has thousands of related nodes (e.g. computed-attribute flows that touch every node in a kind), the page issues thousands of concurrent GraphQL requests for display labels. This produces severe browser instability — long-running JS, layout thrash, dev tools unresponsive — and on large datasets the tab crashes.

### Expected Behavior

Display labels for all related nodes on a task should be resolved through a single (or batched, kind-grouped) GraphQL query, not one query per node. A reasonable approach:

1. Collect the unique `(kind, id)` pairs from `related_nodes` once at the page level.
2. Issue one query per kind (or one combined query) using the existing `ids: [String]` filter — `getObjectDisplayLabel` already accepts an array.
3. Pass resolved labels down to children instead of having each `<Id>` fetch on its own.

### Steps to Reproduce

1. Trigger a workflow that produces a task with a large related-node set (e.g. a Python transform computed-attribute run over a kind with thousands of instances — see #9025).
2. Navigate to the task list (`/tasks`) or open the task details page.
[...]

PR #9070, opened 2026-04-27
Fix task pages firing one GraphQL query per related node (#9067)

fatih-acar:
Task list, task details, and tasks homepage rendered an `<Id>` or `<NodeLabel>` per related node, each fetching its display label through its own GraphQL query. With thousands of related nodes the fan-out froze or crashed the browser.

Resolve display labels through a single batched operation per page that aliases one root field per kind, and pass the resolved labels down to the row components.


ogenstad:
I won't comment about the actual frontend code, but just want to highlight that I think this could serve as a quickfix but that we should also consider adding a resolver in the backend. Similar to what we did for node metadata here:
https://github.com/opsmill/infrahub/blob/infrahub-v1.9.1/backend/infrahub/graphql/manager.py#L690

PR #9079, opened 2026-04-28
IFC-2408: Fix attribute migration for sub-template schemas

### LINKED ISSUE #8722: bug: sub-template instances not updated with new attribute
### Component

API Server / GraphQL

### Infrahub version

1.8.2

### Current Behavior

updating an attribute on a schema that is used as a sub-template (another schema with `generate_template=True` has a COMPONENT relationship to this schema) does not cause the attribute to be added to existing instances of templates for this schema if `generate_template` is False on this schema

### Expected Behavior

sub-template instances should be correctly updated when a new attribute is added to a schema or an attribute on the schema is updated to be added to the linked template schema

## Reproduction Steps

1. Define two node schemas: `TestDevice` with `generate_template=True` and a Component relationship to `TestPart`, and `TestPart` with `generate_template=False` (default).
2. Load and process the schema. Observe that `TemplateTestPart` is automatically generated as a sub-template because `TestPart` is a Component of `TestDevice`.
3. On `TestPart`, define an attribute (e.g., `serial_number`) with `read_only=True`. Because `support_templates` is `read_only is False and unique is False`, this attribute is excluded from `TemplateTestPart` nodes.
4. Create a `TemplateTestPart` instance in the database. Confirm it does not have a `serial_number` attribute.
5. Update the schema to change `TestPart.serial_number` from `read_only=True` to `read_only=False`. This flips `support_templates` from `False` to `True`.
6. The schema diff detects `attribute.read_only.update` and triggers `AttributeSupportsGeneratedSchemaMigration`.
7. **Expected:** The migration adds `serial_number` to all existing `TemplateTestPart` nodes.
8. **Actual:** The migration silently skips the template update. The condition at `backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py:139` checks `self.new_schema.generate_template`, which is `False` for `TestPart`. The migration only runs for schemas that explicitly set `generate_template=True`, missing all sub-templates that exist due to Component/Parent relationships from other schemas.

### Additional Information

_No response_

solababs:
# Why

When a node has `generate_template=False` but is a COMPONENT or PARENT peer of a `generate_template=True` node, Infrahub auto-generates a `TemplateXxx` schema for it ("sub-template"). When an attribute on that node later changed `read_only` or `optional`, the `AttributeSupportsGeneratedSchemaMigration` skipped the migration because it only checked `self.new_schema.generate_template` — which is `False` for sub-template nodes. Existing `TemplateXxx` instances were left with stale attribute paths, silently out of sync with the schema.

Closes https://github.com/opsmill/infrahub/issues/8722

## How to test

```bash
uv run pytest backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py -v
```

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/8722.fixed.md`)

### DIFF
--- a/backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
+++ b/backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Sequence
 
+from infrahub.core import registry
 from infrahub.core.migrations.query.attribute_remove import AttributeRemoveQuery
 from infrahub.core.schema.generic_schema import GenericSchema
 from infrahub.core.schema.node_schema import NodeSchema
@@ -141,9 +142,15 @@ class AttributeSupportsGeneratedSchemaMigration(AttributeSchemaMigration):
                 all_queries.append(ProfilesAttributeRemoveMigrationQuery)
 
         # Check template support changes
+        schema_branch = (
+            registry.schema.get_schema_branch(name=branch.name)
+            if registry.schema.has_schema_branch(name=branch.name)
+            else None
+        )
+        is_sub_template = schema_branch is not None and schema_branch.has(name=f"Template{self.new_schema.kind}")
         if (
             isinstance(self.new_schema, NodeSchema)
-            and self.new_schema.generate_template
+            and (self.new_schema.generate_template or is_sub_template)
             and self.previous_attribute_schema.support_templates != self.new_attribute_schema.support_templates
         ):
             if self.new_attribute_schema.support_templates:
--- a/backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
+++ b/backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
@@ -4,14 +4,14 @@ import pytest
 
 from infrahub.core import registry
 from infrahub.core.branch import Branch
-from infrahub.core.constants import SchemaPathType
+from infrahub.core.constants import RelationshipCardinality, RelationshipKind, SchemaPathType
 from infrahub.core.migrations.schema.attribute_supports_generated_schema import (
     AttributeSupportsGeneratedSchemaMigration,
 )
 from infrahub.core.migrations.shared import MigrationInput
 from infrahub.core.node import Node
 from infrahub.core.path import SchemaPath
-from infrahub.core.schema import SchemaRoot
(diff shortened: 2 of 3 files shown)


backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -280,3 +280,158 @@ async def test_migration_edge_timestamps(
 
     after_snapshot = await snapshotter.snapshot()
     assert_edge_timestamps(before_snapshot, after_snapshot, at_str)
+
+
+# ---------------------------------------------------------------------------
+# Sub-template fixtures & tests (generate_template=False COMPONENT nodes)
+# ---------------------------------------------------------------------------
+
```
gmazoyer: Wonder if we should keep those headers that the robot seems to like so much

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -280,3 +280,158 @@ async def test_migration_edge_timestamps(
...
+        "name": "Part",
+        "namespace": "Test",
+        "branch": "aware",
+        "generate_template": False,
+        "attributes": [
+            {"name": "name", "kind": "Text", "unique": True},
+            {"name": "serial_number", "kind": "Text", "optional": True, "read_only": True},
+        ],
+    },
+]
```
gmazoyer: Should we use one of the helper schema instead? `backend/tests/helpers/schema/`

gmazoyer (approved):
Added a couple of comments related to the tests

backend/infrahub/core/migrations/schema/attribute_supports_generated_schema.py
```diff
@@ -141,9 +142,15 @@ async def execute(
                 all_queries.append(ProfilesAttributeRemoveMigrationQuery)
 
         # Check template support changes
+        schema_branch = (
+            registry.schema.get_schema_branch(name=branch.name)
+            if registry.schema.has_schema_branch(name=branch.name)
+            else None
+        )
+        is_sub_template = schema_branch is not None and schema_branch.has(name=f"Template{self.new_schema.kind}")
```
ajtmccarty, line 150: I added a note to the `Schema Migration Technical Debt` IFC-2399 about including the full SchemaBranch in schema migrations so that we can avoid using the registry in scenarios like this
it's too big of a change for this little fix

ajtmccarty (approved):

PR #9091, opened 2026-04-29
Convert relative links to absolute in the README

ogenstad:
# Why

Relative links such as the one for the logo are broken on PyPI. For reference, check the image on https://pypi.org/project/infrahub-server/

## What changed

* Links and image sources in readme

## Note

In the future we could consider to dynamically generate the readme during releases to account for this, so this is a bit of a quickfix.


dgarros (approved):
TY

PR #9092, opened 2026-04-29
Convert Prefect Tasks to Use Native GraphQL Instead of SDK

solababs:
# Why

Prefect tasks that trigger fan-out workflows were using `client.all()` to fetch full node objects when only `node.id` was needed to dispatch child workflows. The fetched payload grew linearly with the number of attributes and relationships on the schema kind — at 1 000 nodes with 20+ attributes, each task was transferring 20–30× more data than necessary.

This PR replaces the three highest-severity overfetching sites (`display_labels`, `hfid`, `computed_attribute`) with targeted `execute_graphql()` calls that fetch only `id`. Delivers a measurable reduction in execution time and data volume per workflow trigger.

## How to test

```bash
# Unit tests for the new query models
uv run invoke backend.test-unit -- -k "display_label_query or hfid_node_query or computed_attribute_query"

# Benchmark (prints time and payload reduction %)
INFRAHUB_LOG_LEVEL=CRITICAL uv run pytest backend/tests/benchmark/test_query_optimization.py --dist=no -s
```

Expected benchmark output (200 nodes):
```
display_labels : ~80% time reduction  |  ~82% size reduction
hfid           : ~76% time reduction  |  ~82% size reduction
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. Task inputs, outputs, and workflow parameters are unchanged.
- **Performance:** ~76–80% execution time reduction and ~82% payload reduction on the three migrated tasks (measured at 200 nodes; improvement scales with node count).

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../dev/guidelines/changelog.md)
- [x] Internal .md docs updated (`dev/knowledge/backend/async-tasks.md` — added "Read Query Optimization" section)


backend/infrahub/core/graphql_query/node_id_query.py
```diff
@@ -0,0 +1,64 @@
...
+if TYPE_CHECKING:
+    from collections.abc import AsyncGenerator
+
+    from infrahub_sdk.client import InfrahubClient
+
+
+class NodeID(BaseModel):
+    model_config = ConfigDict(frozen=True)
+
+    id: str
```
ogenstad: I think this class can be removed now?

ogenstad (approved):
LGTM

PR #9097, opened 2026-04-29
add spec for INFP-556 - auto create groups

wvandeun:


dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+**Input**: User description: "I want to create a spec for Jira JPD INFP-556"
+
+## Summary
+
+Enterprise customers manage dozens-to-hundreds of access groups in their external identity providers (Azure AD, Okta, AD behind OIDC, Google Workspace, native LDAP). Today an Infrahub admin must manually pre-create a matching `CoreAccountGroup` for every external group before a logging-in user can be assigned permissions through it; otherwise the claim is silently dropped (or the user falls through to the default group from IFC-922). For customers with hundreds of LDAP-managed groups, this manual pre-provisioning is operationally painful, error-prone, and delays team-by-team onboarding.
+
+This feature delivers opt-in, filter-scoped auto-creation of local Infrahub account groups from claims emitted by external authentication providers (SSO/OIDC, OAuth2, native LDAP), so admins can onboard teams without a manual pre-provisioning step. The feature ships in 1.10 alongside native LDAP support (INFP-105) as one "enterprise identity" story. It hardens an existing community contribution (PR #8515 by Alexander Grooff / Adyen) with the product shaping captured in this spec.
+
+## Clarifications
```
ogenstad: I'd suggest that the clarifications come at the end instead of showing up before we. read about the feature.

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+
+**Acceptance Scenarios**:
+
+1. **Given** auto-creation is enabled and a filter `^LDAP/group/(?P<name>.+)$` is configured, **When** a user logs in whose claim contains `LDAP/group/network-engineering`, **Then** Infrahub creates a local group `network-engineering` with no permissions and adds the user as a member.
+2. **Given** the local group `network-engineering` already exists from a prior auto-creation, **When** a different user logs in carrying the same external group, **Then** Infrahub adds the new user to the existing group and does not create a duplicate.
+3. **Given** a filter without a named capture group such as `^network-.*$`, **When** a user logs in with claim `network-eng`, **Then** Infrahub creates a group named `network-eng` (full claim used as-is).
+4. **Given** an admin attaches a role to an auto-created group, **When** members of that group log in, **Then** they receive the permissions associated with that role through the standard AccountGroup → Role → Permission chain.
```
ogenstad, line 31: We should add that if a group of the same name but different type i.e. CoreStandardGroup already exists a warning should be logged. I.e. if we fail to create the account group due to uniqueness constraint limitations from `CoreGroup`.

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+**Independent Test**: Configure auto-creation with a filter and configure `sso_user_default_group`. Simulate a login whose claims do not match any filter pattern. Verify the user is added to the configured default group (and not to any auto-created group).
+
+**Acceptance Scenarios**:
+
+1. **Given** auto-creation is enabled, a filter is configured, and `sso_user_default_group` is set, **When** a user logs in with no matching claims, **Then** the user is added to the default group (existing IFC-922 behavior is preserved).
+2. **Given** the same configuration, **When** a user logs in with one matching claim and several non-matching claims, **Then** the user is added to the auto-created group from the matching claim; the default group is not added on top (the two paths are independent — matching takes precedence over default).
+
+---
+
+### User Story 4 - Distinguish auto-created groups from manually-created ones (Priority: P2)
```
ogenstad, line 66: Clarification, what is source?

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+**Independent Test**: With the activity event log open, trigger auto-creation. Verify an event appears containing the new local group name, the source pattern that matched, the IdP identifier, and the username of the user whose login triggered creation.
+
+**Acceptance Scenarios**:
+
+1. **Given** auto-creation is enabled and a login triggers a new group, **When** the event log is queried, **Then** a structured event is present with fields: local group name, source pattern, IdP, triggering user.
+2. **Given** the same external group claim arrives on a subsequent login (no creation occurs because the group already exists), **When** the event log is queried, **Then** no auto-creation event is emitted for that login (only the original creation is recorded).
+
+---
+
+### User Story 6 - Refuse to start with an enabled-but-unfiltered configuration (Priority: P3)
```
ogenstad: An option could be that the filter must have at least one entry before the setting is active.

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+- **FR-016**: System MUST honor the existing `sso_user_default_group` (IFC-922) when a user's claims produce no matches under the configured filter and a default group is configured. Auto-creation and the default-group fallback MUST be independent: if at least one claim matches, the matching path takes effect; the default group is not stacked on top.
+- **FR-017**: System MUST, when a captured/effective name fails Infrahub group identifier validation, skip that claim, complete the login, and emit a logged event recording the rejected claim. No exception MUST propagate to the end user.
+- **FR-018**: System MUST deduplicate effective names within a single login (multiple claims that resolve to the same name produce one membership operation, not multiple) and across providers (two providers contributing the same effective name produce one local group, not duplicates).
+- **FR-019**: Documentation MUST be updated (`docs/topics/security/sso.mdx`) with a worked example, an explicit safety note about filter scoping, the interaction with the IFC-922 default group, and credit to the contributing customer/author per the assumption below.
```
ogenstad: typo worked -> working

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+- **Seeded description on auto-created groups** *(resolved 2026-04-30 — see Clarifications)*: Auto-created groups do not receive a system-seeded description. Provenance is captured exclusively on the auto-creation event log entry (FR-015). The group's `description` field is left empty at creation time and is freely editable by admins like any other group description.
+- **Lifecycle / removal**: Removing a user from a local group when the IdP stops sending the claim is explicitly out of scope here and is tracked under INFP-536 (Account Lifecycle Management).
+- **Existing PR**: Implementation hardens PR #8515 (Alexander Grooff / Adyen). Release notes credit to Alexander Grooff, wording coordinated with Yvonne — to be finalized at release time.
+- **Configuration surface**: Two settings are introduced under the existing `SecuritySettings` configuration area. The exact shape is captured in FR-001 / FR-002 above; the implementation plan owns naming and any internal data-class design.
```
ogenstad: It seems like there is another setting for soft cap as well that would default to 50?

dev/specs/infp-556-auto-create-groups/spec.md
```diff
@@ -0,0 +1,196 @@
...
+- **FR-009**: System MUST, on first encounter of an effective name that has no corresponding local `CoreAccountGroup`, atomically create that group with zero attached roles and zero attached permissions.
+- **FR-010**: System MUST, on every encounter of an effective name (first or subsequent), ensure the logging-in user is a member of the corresponding local group.
+- **FR-011**: System MUST be safe under concurrency: simultaneous first-logins for the same brand-new effective name MUST result in exactly one local group, and every involved login MUST succeed.
+- **FR-012**: System MUST tag every auto-created group with a `source` value that reflects the originating auth flow: `sso` for groups created from an SSO/OIDC or OAuth2 login, `ldap` for groups created from a native LDAP login. The `source` property MUST be modeled as a regular attribute on `CoreAccountGroup` (not as bespoke metadata) so the existing schema-driven UI and API surface it without dedicated UI work — admins can view and filter groups by `source` through the standard group list mechanism. Allowed values: `manual`, `system`, `sso`, `ldap`.
```
ogenstad: The part of the `source` is very vague, often it's mentioned as a property which gets very confusing since we already have properties that are called `source`. In this place we see that it should be an attribute though so I assume that's the desire. The choice of the word "source" could lead to future confusion as well and I wonder if there's some other attribute name that would be better suited.

Is it relevant to separate sso and ldap for this source? What about just having:

* manual
* system
* identity_provider

For existing groups and migrations should we set all existing groups to be "manual"? Also do we want this to be a read_only attribute or something that an admin user could go in and override and change?

wvandeun: clarified

ogenstad (approved):

PR #9106, opened 2026-04-30
IFC-2168 create standard nodes to replace trigger placeholder automations

polmichel:
## Why

This is the data model that will replace the trigger placeholder used for HFID, Display Labels and Computed attributes computation.

The current data model uses `backend/infrahub/trigger/constants.py:TRIGGER_PLACEHOLDER_FIELD` constant string to reference a fake field that will allow these objects to be recomputed when their definition/signature changes.

From now on, the goal will be to use these standard nodes to allow to recompute these nodes when required.

Closes [IFC-2168]

### What's not included in this PR

This PR is just about creating the new data model, not using it in production code. Replacement will be done later in another PR.

[IFC-2168]: https://opsmill.atlassian.net/browse/IFC-2168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/computed_attribute/signature.py
+++ b/backend/infrahub/computed_attribute/signature.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+from typing import Any, ClassVar
+
+from infrahub.trigger.signature import TriggerSignatureBase, TriggerSignatureGetListQueryBase
+
+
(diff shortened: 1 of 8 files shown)


polmichel:
I'll get back to it later

PR #9022, opened 2026-04-22
Flaky tests: "Event loop is closed" in test_client -> additional logs to find out

polmichel:
## Why

An existing test exhibits a flaky behavior `backend/tests/component/proposed_change/test_validate_artifacts_generation.py::TestValidateArtifactsGeneration::test_unique_query_selective_field_change`
Evidence: https://github.com/opsmill/infrahub/actions/runs/23923336916/job/69774700376?pr=8799

Closes [IFC-2442](https://opsmill.atlassian.net/browse/IFC-2442)

### Stack analysis

The issue may **not** be directly related to this specific component test `test_unique_query_selective_field_change`.
The logs are showing the following error "_RuntimeError: Event loop is closed_".

Following the logs which you can find in the final section

- `backend/infrahub/server.py:118:` in lifespan `await shutdown(application)`
- `backend/infrahub/services/__init__.py:180`  is the call await `self._cache.close_connection()`
- `backend/infrahub/services/adapters/cache/redis.py:66 ` is the one-line implementation `await self.connection.aclose()`
- The four frames inside `.venv/.../redis/asyncio/ `pin the self.connection to `redis.asyncio.Redis`, and `self.connection_pool.disconnect()` to the `redis-py` pool implementation. The `asyncio.StreamWriter._writer.close()` belongs to the `Connection` object inside that pool.

#### Conclusion 

This means that the Redis client object is trying to close an event loop which is already closed. This happens in the `teardown` of the test, not specifically in the `TestValidateArtifactsGeneration::test_unique_query_selective_field_change`.

#### How the current loop is transmitted to Redis

The first Redis command is what opens the socket
Any Redis operation goes through `Redis.execute_command` → `ConnectionPool.get_connection()` → returns a pooled `Connection`. If that `Connection.is_connected` is `false`, the pool calls `await connection.connect()` → `connection._connect()`:
```
  # redis/asyncio/connection.py  (SocketConnection._connect, abridged)
  async def _connect(self):
      reader, writer = await asyncio.open_connection(
          host=self.host,
          port=self.port,
          ssl=self.ssl_context,
          ...
      )
      self._reader = reader
      self._writer = writer
```
  That await `asyncio.open_connection(...)` is the line that links the connection to a specific loop.

#### Asyncio `open_connection` method

`asyncio.open_connection` permanently binds the transport to the running loop:
```
  async def open_connection(host=None, port=None, *, limit=_DEFAULT_LIMIT, **kwds):
      loop = events.get_running_loop()                                # (A)
[...]

PR #9107, opened 2026-04-30
event loop flaky test suite: second fix tentative

polmichel:
### Why

CI occasionally fails with `RuntimeError: Event loop is closed` during `test_client` teardown ([example](https://github.com/opsmill/infrahub/actions/runs/25103323984/job/73558143027?pr=9084)).

#9022 added a stderr dump, but it runs *after* redis-py's `Connection.disconnect()` `finally:` clause nulls `_writer`/`_reader`. The dump shows `writer=None loop=None`. Therefore, the loop the writer was actually bound to is hidden. This is the one fact we need to identify the offender (Prefect worker loop? executor loop? `asyncio.run` site?).

### What

Test-only additions on top of #9022 and no production code changes.

#### Additional information
When redis-py creates a new connection, we now record a snapshot of context against the connection itself: an identifier and a short textual description of the event loop the writer was bound to, the name of the thread the connect ran on, and a small call stack pointing at whatever code triggered the connect.

#### Loop mismatch detection
When a pool is disconnected we also intercept the call. If any pooled connection was created on a different loop than the one currently running, we log a divergence line per connection.

#### Loop human friendly label registry
Finally, a small registry maps loop identifiers to human-readable labels, with the pytest-asyncio session loop tagged once at startup. Both dump paths consult the registry so a divergence between two loops shows up as named subsystems rather than two opaque integers.

Installation runs once per test session via an autouse fixture and is safe to call repeatedly.

## What the next CI hit will show

Two stderr blocks instead of one:

- **Pre-disconnect** — a divergence block listing every connection whose creation loop is not the current loop, with the loop id, the registry label if known, the loop's textual description, the thread the connect ran on, and an indented call-stack sub-block.
- **Post-mortem** — the existing dump, now annotated per connection with the same set of creation-time fields, plus a top-level line naming the current loop's id, label, and closed state.

What actually identifies the owner of an orphaned writer:

- The registry label, when the foreign loop matches a known subsystem.
- The thread name, which usually distinguishes a Prefect worker, an executor pool worker, and the main thread.
- The captured call stack, which points at the synchronous awaiter of the connect.
- The loop id and textual description

## Example pre-disconnect output (loop divergence case)

```
Redis pool disconnect loop divergence detected:
  _available_connections: conn=4391928112 creation_loop=4391851520 creation_loop_label=None creation_loop_repr='<_UnixSelectorEventLoop running=False closed=True debug=False>' creation_thread='ThreadPoolExecutor-0_0' current_loop=4399238400 current_loop_label='pytest-session'
    creation_stack:
      File ".../prefect/utilities/asyncutils.py", line 213, in run_coro_as_sync
        return call.result()
      File ".../prefect/_internal/concurrency/calls.py", line 312, in result
        return self.future.result(timeout=timeout)
      File ".../infrahub/services/adapters/cache/redis.py", line 41, in get
        result = await self._client.get(key)
      File ".../redis/asyncio/client.py", line 612, in execute_command
        conn = await self.connection_pool.get_connection(...)
```

If every pooled connection's creation loop matches the current loop, nothing is printed.

To delete once the flakiness is identified and fixed.

### DIFF
--- a/backend/tests/helpers/diagnostics.py
+++ b/backend/tests/helpers/diagnostics.py
@@ -2,6 +2,7 @@
 import asyncio
 from sys import stderr
 from traceback import format_exception
+from typing import Any, cast
 
 from redis.asyncio.connection import Connection, ConnectionPool
 
@@ -114,8 +115,8 @@ def install_redis_loop_diagnostics() -> None:
             loop = asyncio.get_running_loop()
         except RuntimeError:
             return
-        self._creation_loop_id = id(loop)
-        self._creation_loop_repr = repr(loop)
+        cast("Any", self)._creation_loop_id = id(loop)
+        cast("Any", self)._creation_loop_repr = repr(loop)
 
     original_disconnect = ConnectionPool.disconnect
 
@@ -128,5 +129,5 @@ def install_redis_loop_diagnostics() -> None:
 
     setattr(_instrumented_connect, _INSTALLED_MARKER, True)
     setattr(_instrumented_disconnect, _INSTALLED_MARKER, True)
-    Connection._connect = _instrumented_connect
-    ConnectionPool.disconnect = _instrumented_disconnect
+    cast("Any", Connection)._connect = _instrumented_connect
+    cast("Any", ConnectionPool).disconnect = _instrumented_disconnect
--- a/backend/tests/helpers/diagnostics.py
+++ b/backend/tests/helpers/diagnostics.py
@@ -1,5 +1,6 @@
 # This file should be deleted once the flakiness is gone
 import asyncio
+import threading
 from sys import stderr
 from traceback import format_exception
 from typing import Any, cast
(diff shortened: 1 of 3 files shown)


backend/tests/helpers/diagnostics.py
```diff
@@ -34,10 +55,79 @@ def dump_event_loop_closed_diagnostic(nodeid: str, exc: BaseException) -> None:
...
+        try:
+            _dump_pool_loop_divergence(self)
+        except Exception as diag_exc:
+            print(f"(redis loop divergence dump failed: {diag_exc!r})", file=stderr, flush=True)
+        await original_disconnect(self, inuse_connections=inuse_connections)
+
+    setattr(_instrumented_connect, _INSTALLED_MARKER, True)
+    setattr(_instrumented_disconnect, _INSTALLED_MARKER, True)
+    cast("Any", Connection)._connect = _instrumented_connect
+    cast("Any", ConnectionPool).disconnect = _instrumented_disconnect
```
polmichel, line 187: Not clean, but this code will be hopefully removed soon

ajtmccarty (approved):

PR #9113, opened 2026-04-30
Migrate branch selector to design system components

bilalabbad:


frontend/app/src/entities/branches/ui/branch-list-item/branch-default-badge.tsx
```diff
@@ -1,10 +1,16 @@
-import { Badge, type BadgeProps } from "@/shared/components/ui/badge";
+import type { BadgeProps } from "@/shared/components/ui/badge";
 import { classNames } from "@/shared/utils/common";
 
 export function BranchDefaultBadge({ className, ...props }: BadgeProps) {
   return (
-    <Badge className={classNames("rounded-full font-normal text-gray-700", className)} {...props}>
+    <span
```
pa-lem, line 6: why removing the badge ot use a span?

frontend/app/src/shared/hooks/useIsTruncated.ts
```diff
@@ -1,17 +0,0 @@
-import { type RefObject, useEffect, useState } from "react";
```
pa-lem, line 1: doc can be updated to remove that from knowledge

pa-lem (commented):
the UI changes a lot, has it been discussed before changing it?

frontend/app/src/entities/branches/ui/branch-list-item/branch-default-badge.tsx
bilalabbad: because none of Badge's variants match the look this needs, and previous we were overriding every Badge styles

PR #9116, opened 2026-05-01
greater rollback coverage of merge logic

### LINKED ISSUE #9115: bug: failed merge can leave database in partially-merged state
### Component

API Server / GraphQL

### Infrahub version

1.9.0

### Current Behavior

If a merge operation fails, it is possible for it to leave the branch partially merged, which will most likely leave data and/or schemas in an illegal state.

### Expected Behavior

If a merge fails, we should roll back any changes made during the merge

### Steps to Reproduce

not really reproducible in an easy way b/c it involves an unexpected failure

### Additional Information

we have rollback handling for the 2 biggest portions of the merge logic: the graph merge and the schema migrations following the graph merge. if either of these fails, then all the changes that are part of that particular step are rolled back.
for example, if the graph merge fails, then all changes made as part of the graph merge are rolled back. if the schema migrations fail, then all changes made as part of the schema schema migrations are rolled back.
but there is logic between these steps that remains uncovered by rollback protection. and the rollback executed following a schema migrations failure is not exactly the same as the graph merge rollback.
ideally, all of the merge logic should covered by the same rollback protection

ajtmccarty:
# Why

1. It was possible for a merge operation to fail in such a way that the merge changes had already been applied and were not rolled back. This PR adds greater rollback coverage to the merge logic.
2. Both SchemaUpdateCoordinator and DiffMerger had rollback logic, but they were slightly different. The rollback logic is now consolidated in a single query.
3. `DiffMerger.merge_graph()` advances the source branch's `branched_from` to `merge_at - 1µs` on success. When a downstream step in the wider merge flow then failed, the advance was never undone — making the next merge attempt's diff window exclude the branch's pre-`merge_at` changes. The retry would then "succeed" while silently merging nothing, and any schema migrations would fail with confusing errors (`Unable to find the attribute …`). Surfaced via manual testing.

**Goal:** any unexpected failure between the start of `_do_merge_branch` and the `MERGED` status save should leave the graph and registry as they were before the merge attempt. The branch should never be left in a half-merged state silently looking like `OPEN`.

**Non-goals:**
- Does not change the rebase flow at `tasks.py:rebase_branch` (same shape, same gap, deferred).
- Does not add any method for recovering from a failure during a rollback. This would probably require restoring from a backup or some sort of manual cypher intervention

Closes #9115

## What changed

**Behavioral changes**

- New `BranchStatus.MERGING`. A branch is set to `MERGING` for the duration of `_do_merge_branch`; transitions to `MERGED` on full success, back to `OPEN` if rollback succeeds, or stays at `MERGING` if rollback itself fails (signals "needs human intervention"). Mutations, rebases, re-merges, and proposed-change creation are blocked against a `MERGING` branch with the same semantics as `MERGED`.
- A failed merge now restores the destination branch's graph and registry, regardless of exactly where the failure occurs. Failures in `load_schema_from_db`, `calculate_migrations`, and `coordinator.execute()` are all wrapped.
- A failed merge now also restores the source branch's `branched_from` to its pre-merge value, so a retry sees the same diff window the first attempt did.
- `SchemaUpdateCoordinator.execute()` accepts `manage_rollback: bool = True`. The merge call site passes `False` so the outer wrapper is the sole rollback authority. All other call sites are unchanged.

**Implementation notes**

- `DiffMergeRollbackQuery` and `RollbackQuery` are consolidated into a single `RollbackQuery`. It now accepts an optional `node_uuids` list to restore  `updated_at`/`updated_by` metadata for Node, Attribute, and Relationship vertices using `previous_updated_at/by`. It always resets `to_user_id` along with `to`, and always deletes orphan vertices. All write subqueries use `CALL { ... } IN TRANSACTIONS`.
- New private helper `_rollback_merge` in `tasks.py` runs three steps: `merger.rollback()`, registry restore, then a single `branch.save()` that flips status back to `OPEN` and restores `branched_from`. Each step is independently `try/except`-wrapped; the helper does not raise. If rollback fails, the branch is left at `MERGING`. Ideally this would be reorganized to be more SOLID, but that would be a bigger change and this fix is for a specific bug.
- `_do_merge_branch` captures `pre_merge_branched_from = branch.branched_from` before the merge starts and threads it through to `_rollback_merge`. The advance itself remains inside `DiffMerger.merge_graph`. Moving it to the success path of `_do_merge_branch` would arguably be cleaner — `branched_from` and `status=MERGED` are paired "this merge committed" facts — but that's a wider refactor (other direct callers of `merge_graph` would have to change). Capture + restore on the rollback path is the smaller, more targeted fix for the bug at hand.

**What stayed the same**

- API contract for `/schema/load` is unchanged (`manage_rollback` defaults to `True`).
- Rebase flow in `tasks.py` is unchanged.
- `BranchMerger.merge()` and its inner rollback are unchanged.
- Schema-only rollback path (`/schema/load`, CLI init) keeps `IN TRANSACTIONS` batching; behavior is identical except that `to_user_id` is now reset alongside `to` (a strictly more-correct fix).

### Suggested review order

1. `backend/infrahub/core/query/rollback.py` — the consolidated query.
2. `backend/infrahub/core/branch/tasks.py` — `_rollback_merge` and the restructured `_do_merge_branch`.
3. `backend/infrahub/core/schema/update_coordinator.py` — `manage_rollback` plumbing.
4. `backend/infrahub/core/branch/enums.py` and the audit changes (`status_checker.py`, `permissions/report.py`, GraphQL mutations).
5. Tests last.

## How to review

**Focus areas**

- The order of operations in `_rollback_merge`: DB rollback first, then registry restore, then the final `branch.save()` that resets `status` and `branched_from` together.
- The decision to keep the `branched_from` advance in `DiffMerger.merge_graph` (versus moving it to the success path of `_do_merge_branch`). Both options are correct; the move would tie the advance to the success path more directly, at the cost of a wider refactor of `merge_graph`'s direct callers. Worth a second opinion on whether the larger change is preferable.
- The `manage_rollback=False` path in `SchemaUpdateCoordinator.execute()` confirm it propagates the original exception (or `MigrationError` from `error_msgs`) without invoking `_handle_failure_and_rollback`.
- The `BaseException` (not `Exception`) catch in `_do_merge_branch` so cancellation also triggers cleanup.

## How to test

I did manual testing using a local Infrahub instance. I inserted an error into the merge branch workflow; verified all changes were rolled back and branch was in the correct state; removed the error; ran the merge again; verified that it succeeded and merged branch was updated correctly. The retry-after-rollback scenario surfaced the `branched_from` bug noted in **Why** above; the fix is covered by an extension to `test_merge_branch_rollback` that asserts both `status` and `branched_from` are restored after a failed merge.

```bash
# Targeted component / integration tests for rollback query and merge flow:
uv run pytest -x backend/tests/component/core/diff/query/test_rollback.py
uv run pytest -x backend/tests/component/core/diff/test_diff_merger.py
uv run pytest -x backend/tests/component/core/diff/test_diff_and_merge.py
uv run pytest -x backend/tests/component/core/schema_manager/test_schema_rollback.py
uv run pytest -x backend/tests/integration/diff/test_merge_rollback.py

# Unit tests for the BranchStatus.MERGING audit:
uv run pytest backend/tests/unit/core/branch/test_merged_status.py
```

Manual: trigger a merge through the UI on a branch with schema changes, confirm the success path is unchanged. To exercise rollback, the existing `BrokenBranchMerger` pattern in `test_merge_rollback.py` injects a `ValueError` after `merge_graph` returns; the branch returns to `OPEN` and `main` data is restored.

## Impact & rollout

- **Backward compatibility:** `BranchStatus` enum gains `MERGING`. Any consumer that exhaustively switches on `BranchStatus` (clients, dashboards, reports) needs a case for it. `is_terminal` does not include `MERGING`. The new `manage_rollback` parameter on `SchemaUpdateCoordinator.execute()` is keyword-only with a default that preserves existing behavior, so all current call sites are unaffected.
- **Performance:** Adds one extra `Branch.save()` at the start of every merge (for the `MERGING` write) and one at the end of the rollback path. Negligible. The consolidated `RollbackQuery` does the same DB work as the two queries it replaces, plus orphan vertex cleanup that the merge-rollback path was missing.
- **Config / env changes:** None.
- **Deployment notes:** Safe to deploy. No data migration needed. Branches currently at `OPEN` continue to behave as before; the new `MERGING` state only appears for newly initiated merges.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/9115.fixed.md`)


backend/infrahub/core/branch/tasks.py
```diff
@@ -297,7 +298,7 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
             node_events = await _do_merge_branch(
                 db=db,
                 log=log,
-                obj=obj,
+                branch=obj,
```
ogenstad, line 301: 👍

backend/infrahub/core/branch/tasks.py
```diff
@@ -321,110 +322,165 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
...
+        # reset destination branch's schema
+        registry.schema.set_schema_branch(name=merger.destination_branch.name, schema=pre_merge_schema)
+        merger.destination_branch.update_schema_hash()
+        await merger.destination_branch.save(db=db)
+    except Exception:
+        log.exception("Registry restore failed during merge rollback")
+
+    branch.branched_from = pre_merge_branched_from
+    branch.status = BranchStatus.OPEN
+    await branch.save(db=db)
```
ogenstad: Just highlighting here that we don't set a user_id here, could be that we don't have it in this context, and it could technically be true that it was the system user who did this

backend/infrahub/core/branch/tasks.py
```diff
@@ -321,110 +322,165 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
...
+        diff_repository=diff_repository,
+        source_branch=branch,
+        diff_locker=DiffLocker(),
+        workflow=workflow,
+    )
+    try:
+        async with lock.registry.global_graph_lock():
+            # Set to MERGING to lock the branch while merge proceeds
+            branch.status = BranchStatus.MERGING
+            await branch.save(db=db)
```
ogenstad: Here we would have the user from the context and could inject it to .save().

backend/infrahub/core/branch/tasks.py
```diff
@@ -321,110 +322,165 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
...
 
     # -------------------------------------------------------------
-    # Set branch status to MERGED to make it read-only
+    # Point of no return: merge fully succeeded. Advance to MERGED.
     # -------------------------------------------------------------
-    obj.status = BranchStatus.MERGED
-    await obj.save(db=db)
-    registry.branch[obj.name] = obj
+    branch.status = BranchStatus.MERGED
+    await branch.save(db=db)
```
ogenstad: Same as above as we have the user we might as well add it to branch.save here.

backend/infrahub/core/branch/tasks.py
```diff
@@ -321,110 +322,165 @@ async def merge_branch(branch: str, context: InfrahubContext, proposed_change_id
...
     # -------------------------------------------------------------
     await workflow.submit_workflow(
         workflow=BRANCH_CANCEL_PROPOSED_CHANGES,
         context=context,
-        parameters={"branch_name": obj.name},
+        parameters={"branch_name": branch.name},
     )
 
-    if config.SETTINGS.main.delete_branch_after_merge and not obj.is_default:
+    if config.SETTINGS.main.delete_branch_after_merge and not branch.is_default:
```
ogenstad, line 482: Not related to this PR, just highlighting that here we need a global config object to determine how the code should behave. I think we should consider to transition away from that. Basically when we initialize Infrahub we'd create a config object and instead tie it to the app and then we pass along the active config to other classes and functions. I think it could make various tests easier if we didn't have to go and modify a shared global object in various places.

ogenstad (approved):

backend/infrahub/core/branch/tasks.py
ajtmccarty: yes I agree. ideally, `_do_merge_branch` would be encapsulated in a component into which we can inject dependencies and settings, but I'm going to save that for a different PR

PR #9118, opened 2026-05-01
unify Branch filtering

ajtmccarty:
consolidate the logic for filtering Branches: is_default, is_global, is in a terminal state


backend/infrahub/cli/db.py
```diff
@@ -439,11 +438,7 @@ async def migrate_database(
 
 
 async def mark_branches_needing_rebase(db: InfrahubDatabase) -> list[Branch]:
-    branches = [
-        b
-        for b in await Branch.get_list(db=db)
-        if b.name not in [registry.default_branch, GLOBAL_BRANCH_NAME] and not b.is_terminal
-    ]
+    branches = await Branch.get_list(db=db, exclude_global=True, exclude_default=True, exclude_terminal=True)
```
ogenstad, line 441: I would expect Branch.get_list to always return a list and as such part of this function can be removed. Instead it becomes a one line function `return await Branch.get_list(...`

backend/infrahub/core/convert_object_type/object_conversion.py
```diff
@@ -91,8 +91,7 @@ async def get_unidirectional_rels_peers_ids(
 
 
 async def _get_other_active_branches(db: InfrahubDatabase) -> list[Branch]:
-    branches = await Branch.get_list(db=db)
-    return [branch for branch in branches if not (branch.is_global or branch.is_default or branch.is_terminal)]
+    return await Branch.get_list(db=db, exclude_global=True, exclude_default=True, exclude_terminal=True)
```
ogenstad: We could potentially consider this for later but this function is exactly the same as `mark_branches_needing_rebase()` from above. Since both should be one line functions I guess there's no huge harm. Just wondering how often they end up getting called.

backend/infrahub/core/migrations/graph/m068_cleanup_branch_schema_parameters.py
```diff
@@ -130,8 +130,7 @@ async def execute(self, migration_input: MigrationInput) -> MigrationResult:
 
         console.log("[bold]Cleaning up spurious branch schema attribute parameters from Migration056[/bold]")
 
-        branches = await Branch.get_list(db=db)
-        branches = [b for b in branches if not b.is_default and not b.is_global and not b.is_terminal]
+        branches = await Branch.get_list(db=db, exclude_global=True, exclude_default=True, exclude_terminal=True)
```
ogenstad, line 133: As an example here we didn't choose to call one of those functions, which makes me wonder if there's a huge need for those functions.

ogenstad (approved):
LGTM, added some minor comments.

backend/infrahub/cli/db.py
ajtmccarty: this one actually does some extra stuff under the diff, so I think we need to keep it

backend/infrahub/core/convert_object_type/object_conversion.py
ajtmccarty: removed this function

backend/infrahub/cli/db.py
ogenstad: Ah, I see now I missed that part.

PR #9119, opened 2026-05-02
unify Prefix <-> IP Pool relationships

### LINKED ISSUE #7070: bug: resource pool tab for an IP Prefix does not list related IP Prefix pools
### Component

API Server / GraphQL

### Infrahub version

1.3.7

### Current Behavior

When a prefix is a resource for a CoreIPPrefixPool, then the prefix pool is not listed in the Resource Pool tab of the detail view of the Prefix.

### Expected Behavior

The IP prefix pools for which this IP prefix is a resource should be listed on the Resource Pool tab of the detail view of the prefix.

### Steps to Reproduce

On the Infrahub sandbox
- Create a CoreIPPrefixPool for the 10.0.0.0/8 prefix
- Create a CoreIPAddressPool for the 10.0.0.0/8 prefix
- Navigate to the detailed view of the 10.0.0.0/8 IPAM prefix and open the `Resource Pool` tab
- Notice that only the CoreIPAddressPool will be listed

### Additional Information

It seems that definition of the BuiltinIPPrefix generic has a resource_pool relationship defined to CoreIPAddressPool, which is not correct.

It should be CoreIPAddressPool and CoreIPPrefixPool instead, but there doesn't seem to be a generic defined that only these two inherit from.

https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/core/schema/definitions/core/ipam.py#L129

ajtmccarty:
# Why

The "Resource Pool" tab on the IP prefix detail view only listed `CoreIPAddressPool` instances. If a user had a `CoreIPPrefixPool` that allocated from the same IP prefix, the IP prefix did not show it. The `BuiltinIPPrefix` schema and had no relationship to `CoreIPPrefixPool`, although the reverse relationship from `CoreIPPrefixPool` did exist.

**Goal:** make the Resource Pool tab list both pool kinds, and do it for pre-existing data without re-creating relationships.

**Non-goals:** no behavior change for `CoreIPAddressPool.resources` or `CoreIPPrefixPool.resources` allocation flows; the resource manager still resolves pool→prefix relationships through the same code paths. No changes to user-defined schemas that inherit from `BuiltinIPPrefix` beyond the schema-graph rewrite the migration performs.

Closes #7070

## What changed

**Behavioral changes:**
- A new (entirely empty) `CoreIPPool` generic has been introduced. Both `CoreIPPrefixPool` and `CoreIPAddressPool` now inherit from it.
- `BuiltinIPPrefix.resource_pool` now points at `CoreIPPool` (instead of `CoreIPAddressPool` directly).
- The `resources` relationship identifier on both pools is unified under `ippool__resource` (was `prefixpool__resource` and `ipaddresspool__resource` respectively).
- A graph migration (`m071_unify_ip_pool_resource_identifier`) updates all data regardless of branch or time so the new shape is consistent at every queryable timestamp. Effectively, once the migration runs, it is as though the schema has always been shaped this way. This is simpler than only updating the schema from the tiem of the migration, which would require this to be a migration that supports rebasing.
  - Renames the `Relationship` vertex `name` property from the two old identifiers to `ippool__resource` (constrained to `Pool` ↔ `BuiltinIPPrefix` peers so user schemas that happen to reuse those strings aren't touched).
  - Adds the `:CoreIPPool` label to existing pool vertices.
  - Bootstraps the `CoreIPPool` SchemaGeneric in the schema graph with `from` backdated to the original `CoreResourcePool` schema's `from` time.
  - Appends `CoreIPPool` to `inherit_from` on both pool SchemaNodes by hard-deleting the old `HAS_VALUE` edge and creating a new one that copies the old edge's properties.
  - Rewrites `identifier` on the three relevant SchemaRelationships and `peer` on `BuiltinIPPrefix.resource_pool`, using the same hard-delete-and-rebuild pattern.

**Implementation notes:**
- The migration is an `ArbitraryMigration` (we need to read schema-graph state, decide what to write, and run several writes atomically). All Cypher lives in dedicated `Query` subclasses with typed `parsed_*` accessors, no inline `db.execute_query` calls in the orchestration code.
- A single `BulkRewriteSchemaAttributeQuery` covers both the SchemaNode (`inherit_from`) and SchemaRelationship (`identifier`/`peer`) rewrites — the path-to-attribute branches in Cypher based on whether the rewrite payload's `relationship_name` is set.
- Backdating: the data-level rewrites (`Relationship.name` and the `:CoreIPPool` label) are non-temporal in Neo4j, so they apply at every past and future timestamp automatically. The schema-graph rewrites preserve the old edge's `from` so historical schema queries see the new shape too.
- The new `CoreIPPool` SchemaGeneric is added using `SchemaManager.load_node_to_db` with a backdated `at`.

**What stayed the same:**
- `CoreIPAddressPool.resources` and `CoreIPPrefixPool.resources` still target `BuiltinIPPrefix` with `cardinality=many`. Allocation paths (`get_resource`, `IS_RESERVED` edges, the resource manager queries) are untouched.
- `BuiltinIPAddress` is unchanged.
- No frontend changes — the existing "Resource Pool" tab logic already handles any `Generic` cardinality-many relationship.

## How to review

Suggested order:
1. **Schema definition deltas** — small and self-contained.
   - `backend/infrahub/core/constants/infrahubkind.py` — adds `IPPOOL`.
   - `backend/infrahub/core/schema/definitions/core/resource_pool.py` — adds the `core_ip_pool` generic, updates both pools' `inherit_from`, swaps the `resources` identifier to `ippool__resource`.
   - `backend/infrahub/core/schema/definitions/core/ipam.py` — `BuiltinIPPrefix.resource_pool` now points at `IPPOOL` with the unified identifier.
2. **Migration** — `backend/infrahub/core/migrations/graph/m071_unify_ip_pool_resource_identifier.py`. Walk the `Migration071.execute` method top-to-bottom; each step has its own Query class higher up.
3. **Test** — `backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py`. The test downgrades the in-memory schema *before registration* so the DB starts in true pre-migration shape, then exercises end-to-end with `NodeManager`.

## How to test

```bash
# Component test that drives the full pre→migrate→post flow end-to-end via NodeManager
uv run pytest -x -v backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py

# Sanity check that the existing template/resource-pool flows still pass
uv run pytest backend/tests/functional/schemas/test_template_resource_pool_relationships.py
uv run pytest backend/tests/integration/templates/test_template_resource_pool_lifecycle.py
uv run pytest backend/tests/functional/ipam/test_ipam_utilization.py
```

Completed manual end-to-end testing:

1. Start an instance running `v1.8.4` with a "large" dataset from `infrastructure_edge.py`
2. Run `infrahub upgrade` — `Migration071` applies.
3. Verify all schema updates are applied:
  a. `CoreIPPool` exists
  b. `CoreIPAddressPool` and `CoreIPPrefixPool` inherit from `CoreIPPool`
  c. `CoreIPAddressPool.resources` has new identifier: "ippool__resource"
  d. `CoreIPPrefixPool.resources` has new identifier: "ippool__resource"
  e. `BuiltinIPPrefix` / `IpamIPPrefix` relationship `resource_pool` has new peer (`CoreIPPool`) and new identifier "ippool__resource"
4. In the UI, verify that the existing prefix and address pools are still linked to their IP Prefixes
5. Verify that the address and prefix pools both show in the `Resource Pool` tab of the UI for the IP Prefixes.

## Impact & rollout

- **Backward compatibility:**
  - **GraphQL clients:** breaking. `BuiltinIPPrefix.resource_pool` now returns the abstract `CoreIPPool`, so any query that selected pool-specific fields directly under `resource_pool` needs inline fragments (`... on CoreIPAddressPool { ... }` / `... on CoreIPPrefixPool { ... }`). See the changelog entry for a before/after example.
  - **Existing pool data:** handled — `Migration071` performs the in-place rewrite as part of `infrahub upgrade`.
  - **Python SDK:** the existing `BuiltinIPPrefix.resource_pool` `RelationshipManager` keeps working; consumers get the broader `CoreIPPool` peer kind back.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/+ip_pool_resource_pool.fixed.md`, includes the GraphQL fragment migration note)


backend/infrahub/core/schema/definitions/core/resource_pool.py
```diff
@@ -53,6 +53,16 @@
...
+core_ip_pool = GenericSchema(
+    name="IPPool",
+    namespace="Core",
+    label="IP Pool",
+    description="A pool of IP resources (prefixes or addresses).",
+    include_in_menu=False,
+    branch=BranchSupportType.AGNOSTIC,
+    generate_profile=False,
+)
+
```
polmichel, line 65: Would `core_ip_pool` be a good candidate to implement the `restricted_namepsaces` generic schema attribute? 
This would look like `restricted_namespaces=["Core"]`

This would prevent users from creating their own sub-schemas from CoreIPPool. This would still allow the two existing schemas inheriting from this generic as they belong to the Core namespace: `CoreIPPrefixPool` and `CoreIPAddressPool`.

Another example can be found there `backend/infrahub/core/schema/definitions/core/webhook.py:18`.

backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py
```diff
@@ -0,0 +1,250 @@
...
+    assert loaded_address_pool.get_kind() == InfrahubKind.IPADDRESSPOOL
+    address_pool_resource_ids = {
+        r.peer_id for r in await loaded_address_pool.get_relationship(name="resources").get_relationships(db=db)
+    }
+    assert address_pool_resource_ids == {ip_prefix.id}
+
+    # ---------------------------------------------------------------------
+    # 6. Run the graph integrity checks.
+    # ---------------------------------------------------------------------
+    await verify_graph(db=db)
```
polmichel, line 230: Is this something you would recommend to add for every migration that modifies other things than a Node "functional" value ? (the migrations about MAC address values and related HFID, Display label typically seem not to be a good candidate)

backend/infrahub/core/migrations/graph/m071_unify_ip_pool_resource_identifier.py
```diff
@@ -0,0 +1,584 @@
...
+// --------------------
+WITH parent, rw
+WHERE name_value = rw.parent_name AND namespace_value = rw.parent_namespace
+// --------------------
+// pick the vertex whose attribute we'll rewrite: either the parent itself
+// (relationship_name is null), or the active SchemaRelationship with the
+// given name/identifier. If relationship_name is set but no active path is
+// found, vertex_to_update resolves to NULL and the row is dropped.
+// --------------------
+CALL (parent, rw) {
```
polmichel: This is executed each time no matter `rw.relationship_name` IS NULL or not right? I assume this is not important as this is a one shot migration but a `WITH parent, rw WHERE rw.relationship_name IS NOT NULL` at the start of the CALL would not be better if this was repeatable production code?

backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py
```diff
@@ -0,0 +1,250 @@
...
+@pytest.fixture
+async def pre_migration_schema_db(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_internal_models_schema: SchemaBranch,
+    ipam_schema: SchemaRoot,
+) -> SchemaBranch:
+    """Persist a downgraded core schema (+ concrete IPAM nodes) to the database.
+
+    Mirrors :func:`do_register_core_models_schema` but runs the in-memory downgrade
```
gmazoyer, line 70: Seems the bot learnt how to make reference in docstrings now \o/

gmazoyer (approved):

polmichel (approved):
Just a few questions as this is, from my perspective, an interesting and complex migration.
Thanks for all the different comments that clearly helped to review

polmichel:
I constantly hit API error while trying to post additional comments so I just add them there. 

1. `BulkRewriteSchemaAttributeQuery` -> could be helpful for future migrations
2. Does the `WHERE NOT` condition is there to prevent against any deletion error or for a real use case? `backend/infrahub/core/migrations/graph/m071_unify_ip_pool_resource_identifier.py` line 340
```
CALL (av_old) {
    WITH av_old
    WHERE NOT exists((av_old)<-[]-())
    DETACH DELETE av_old
}
```

backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py
ajtmccarty: You're right!

backend/tests/component/core/migrations/graph/test_071_unify_ip_pool_resource_identifier.py
ajtmccarty: yes. I would add this to the end of any migration test to catch duplicated edges and orphaned vertices that won't cause any problems in the ORM, but can lead to unexpected behavior at the database level during schema/database migrations

backend/infrahub/core/schema/definitions/core/resource_pool.py
ajtmccarty: Yes I think we want this. I will confirm and handle in a little follow-up PR

backend/infrahub/core/migrations/graph/m071_unify_ip_pool_resource_identifier.py
ajtmccarty: good question. it comes down to "unit" vs "returning" subqueries as described [here](https://neo4j.com/docs/cypher-manual/current/subqueries/call-subquery/#call-returning-unit). this subquery is a "returning" subquery.

- a "unit" subquery returns no rows and does NOT influence the rows in the result set
- a "returning" subquery returns rows and, if it does not return a row for a given input, then it effectively removes that row from the outer result set.

in this case, if we only returned for rows where `rw.relationship_name IS NOT NULL`, then we would silently drop all the rows for which `rw.relationship_name IS NULL`

backend/infrahub/core/migrations/graph/m071_unify_ip_pool_resource_identifier.py
```diff
@@ -0,0 +1,584 @@
...
+                console.log(f"  Skipping {node_name}.inherit_from: existing value is not valid JSON.")
+                continue
+            if not isinstance(current, list):
+                console.log(f"  Skipping {node_name}.inherit_from: existing value is not a list.")
+                continue
+            rows.append(InheritFromRow(node_name=node_name, current=current))
+        return rows
+
+
+class BulkRewriteSchemaAttributeQuery(Query):
```
ajtmccarty: as suggested by @polmichel , `BulkRewriteSchemaAttributeQuery` is moved to its own file

backend/infrahub/core/migrations/query/bulk_rewrite_schema_attribute.py
```diff
@@ -0,0 +1,150 @@
...
+DELETE hv_old
+// --------------------
+// detach delete orphaned AttributeValue vertices
+// --------------------
+WITH DISTINCT av_old
+CALL (av_old) {
+    WITH av_old
+    WHERE NOT exists((av_old)<-[]-())
+    DETACH DELETE av_old
+}
```
ajtmccarty, line 148: @polmichel regarding your question on this section. this subquery is looking at all the AttributeValue vertices that we removed edges from in the query above and deleting the vertex if it no longer has any relationships. the `WHERE NOT` check is required so that we don't accidentally delete a vertex that still has `HAS_VALUE` edges linked to it.

PR #9123, opened 2026-05-04
IFC-2536: Migrate display_labels, hfid and computed_attribute to use new query

solababs:
# Why

Prefect tasks that trigger fan-out workflows were using `client.all()` to fetch full node objects when only `node.id` was needed to dispatch child workflows. The fetched payload grew linearly with the number of attributes and relationships on the schema kind — at 1 000 nodes with 20+ attributes, each task was transferring 20–30× more data than necessary.

This PR replaces the three highest-severity overfetching sites (`display_labels`, `hfid`, `computed_attribute`) with targeted `execute_graphql()` calls that fetch only `id`. Delivers a measurable reduction in execution time and data volume per workflow trigger.

## How to test

```bash
# Unit tests for the new query models
uv run invoke backend.test-unit -- -k "display_label_query or hfid_node_query or computed_attribute_query"

# Functional output-equivalence tests
uv run pytest backend/tests/functional/display_labels/test_display_label_task_optimization.py -v
uv run pytest backend/tests/functional/hfid/test_hfid_task_optimization.py -v
uv run pytest backend/tests/functional/computed_attributes/test_computed_attribute_task_optimization.py -v

# Benchmark (prints time and payload reduction %)
INFRAHUB_LOG_LEVEL=CRITICAL uv run pytest backend/tests/benchmark/test_query_optimization.py --dist=no -s
```

Expected benchmark output (200 nodes):
```
display_labels : ~80% time reduction  |  ~82% size reduction
hfid           : ~76% time reduction  |  ~82% size reduction
```

## Impact & rollout

- **Backward compatibility:** No breaking changes. Task inputs, outputs, and workflow parameters are unchanged.
- **Performance:** ~76–80% execution time reduction and ~82% payload reduction on the three migrated tasks (measured at 200 nodes; improvement scales with node count).

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../dev/guidelines/changelog.md)
- [x] Internal .md docs updated (`dev/knowledge/backend/async-tasks.md` — added "Read Query Optimization" section)


backend/infrahub/display_labels/tasks.py
```diff
@@ -197,14 +202,9 @@ async def trigger_update_display_labels(
...
-    # NOTE we only need the id of the nodes, this query will still query for the HFID
-    node_schema = registry.schema.get_node_schema(name=kind, branch=branch_name)
-    nodes = await client.all(
-        kind=kind,
-        branch=branch_name,
-        exclude=node_schema.attribute_names + node_schema.relationship_names,
-        populate_store=False,
-    )
+    node_query = DisplayLabelNodeIDQuery(kind=kind)
+    response = await client.execute_graphql(query=node_query.render_query(), branch_name=branch_name)
```
ogenstad: One consideration here that differs compared to running `await client.all()` is that the SDK by default has a pagination size of 50 (https://github.com/opsmill/infrahub-sdk-python/blob/v1.20.0/infrahub_sdk/config.py#L58). While the above query should work for larger datasets, if I remember correctly Infrahub will paginate internally and eventually respond to all if you send in a query that is expected to return a high number of nodes say 8000. Potentially this could be fine if we only query for the ID, but still we should probably have a different pagination scheme here and run the above query in a loop so that each iteration doesn't have to wait too long for the server to complete the query. Similar to the comment on the other query the pagination variables needs to be sent in as variables to execute_graphql and not as part of the rendered query.

ogenstad (commented):
LGTM overall, but would be nice to address the issue of pagination. Also not sure if we need all of the optimization tests after this has been merged but we could keep them for now. It doesn't look like there's any overlap between this and the other PR. So perhaps this could target `develop` directly?

backend/infrahub/core/graphql_query/node_id_query.py
ajtmccarty, line 1: the other files in this directory are all for database queries, so I think this should move to a different dir. I'm not sure where exactly. there is `backend/infrahub/graphql/queries/`. or maybe this should just be a new `backend/infrahub/core/graphql_query/` dir if there's going to be more of this going forward

backend/infrahub/computed_attribute/graphql_queries.py
ajtmccarty, line 1: maybe this is just b/c I spend more time working with database queries, but I generally assume that anything called `query` outside of a graphql query directory is a cypher query. so I think this should be `graphql_queries.py`

backend/infrahub/display_labels/models.py
```diff
@@ -163,6 +164,10 @@ def new(
         )
 
 
+class DisplayLabelNodeIDQuery(NodeIDQuery):
```
ajtmccarty: should this be in its own separate file to match the computed attributes pattern?

solababs:
> LGTM overall, but would be nice to address the issue of pagination. Also not sure if we need all of the optimization tests after this has been merged but we could keep them for now. It doesn't look like there's any overlap between this and the other PR. So perhaps this could target `develop` directly?

It will be merged into this https://github.com/opsmill/infrahub/pull/9092 PR that contains the spec then to develop

backend/infrahub/hfid/models.py
```diff
@@ -163,6 +164,10 @@ def new(
         )
 
 
+class HFIDNodeIDQuery(NodeIDQuery):
+    query_name: ClassVar[str] = "HFIDFetchNodeIDs"
```
ajtmccarty: I don't think this belongs in `models.py`

backend/infrahub/core/graphql_query/node_id_query.py
ajtmccarty, line 1: I think `node_id_query.py` is a better name for this

backend/infrahub/core/graphql_query/node_query.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import TYPE_CHECKING, Any, ClassVar
+
+from infrahub_sdk.graphql import Query
+from pydantic import BaseModel, ConfigDict
+
+if TYPE_CHECKING:
+    from infrahub_sdk.client import InfrahubClient
+
+_PAGE_SIZE = 50
```
ajtmccarty: this limit seems low to me, but I could be wrong

backend/infrahub/display_labels/tasks.py
```diff
@@ -197,14 +202,8 @@ async def trigger_update_display_labels(
...
 
-    # NOTE we only need the id of the nodes, this query will still query for the HFID
-    node_schema = registry.schema.get_node_schema(name=kind, branch=branch_name)
-    nodes = await client.all(
-        kind=kind,
-        branch=branch_name,
-        exclude=node_schema.attribute_names + node_schema.relationship_names,
-        populate_store=False,
-    )
+    node_query = DisplayLabelNodeIDQuery(kind=kind)
```
ajtmccarty, line 205: I think that @ogenstad (please correct me if I am wrong) was suggesting that the pagination can happen at this level so that each iteration of the loop to get a batch of node IDs can submit the display label process workflow for each of those node IDs before continuing.

something like this, where fetch_all_paginated is probably a generator of some sort
```
workflow = get_workflow()
for node_batch in node_query.fetch_all_paginated(client=client, branch_name=branch_name, page_size=50):
    for node in node_batch:
        await workflow.submit_workflow(...)
```

backend/infrahub/core/graphql_query/node_id_query.py
```diff
@@ -0,0 +1,65 @@
...
+        return query.render()
+
+    def parse_response(self, response: dict[str, Any]) -> list[NodeID]:
+        result: list[NodeID] = []
+        if kind_payload := response.get(self.kind):
+            for edge in kind_payload.get("edges", []):
+                if node := edge.get("node"):
+                    if node_id := node.get("id"):
+                        result.append(NodeID(id=node_id))
+        return result
```
ogenstad, line 46: In general I always prefer to have known objects in the returns. However in this case it might be better to return `list[str]`. As `NodeID` only has a single attribute `.id` I don't think there's any extra value of having it as an object, it might be missing something but looks like a list of strings would do just as fine. The reason I'm mentioning it is that there's a small performance cost to creating a Pydantic object (a bit more costly than a `@dataclass`). I think it should be fine to remove the `NodeID` class and replace with strings?

backend/infrahub/display_labels/tasks.py
ogenstad: Correct, but it looks like that's what we have here but might be that the comment above referred to an older commit?

ogenstad (approved):
LGTM! Left a comment around a small optimization. I don't think that the optimization tests are relevant after this has been merged, but something that was useful to measure during development should those be removed?

Finally since this should work in isolation from the other work can we have this target `develop` instead so that we can treat this as a piece of the work that's already completed?

backend/infrahub/core/graphql_query/node_id_query.py
```diff
@@ -0,0 +1,65 @@
...
+
+    async def fetch_all_paginated(
+        self, client: InfrahubClient, branch_name: str, page_size: int = PAGE_SIZE
+    ) -> AsyncGenerator[list[NodeID], None]:
+        rendered_query = self.render_query()
```
ogenstad, line 52: Small change here can we change it to:

```python
    async def fetch_all_paginated(
        self, client: InfrahubClient, branch_name: str, page_size: int | None = None
    ) -> AsyncGenerator[list[NodeID], None]:
        page_size = page_size or client.config.pagination_size
        rendered_query = self.render_query()
```

That way we keep the pagination size that's configured for the SDK within this environment.

PR #9124, opened 2026-05-04
feat(permissions): add rebase_branch permission

### LINKED ISSUE #8050: bug: Add rebase_branch global permission for non-admin users
### Component

Backend

### Infrahub version

1.6.2

### Current Behavior

Currently, non-admin users cannot perform branch rebase operations even when they have been granted `merge_branch` and `merge_proposed_change` permissions. This is because there is no dedicated `rebase_branch` global permission - rebase operations are restricted to users with `super_admin` only.

  - Users with `global:merge_branch:allow_all` permission can merge branches
  - Users with `global:merge_proposed_change:allow_all` permission can merge proposed changes
  - **Rebase operations require `super_admin` permission** - there is no granular permission for rebase

### Expected Behavior

  A new global permission `rebase_branch` should be available, allowing administrators to grant rebase capabilities to non-admin users without requiring full `super_admin` access.

### Steps to Reproduce

  1. **Create a non-admin user account**
     - Navigate to Admin > Role Management > Accounts
     - Create a new user (e.g., `testuser`)

  2. **Create a role with branch permissions**
     - Navigate to Admin > Role Management > Roles
     - Create a new role (e.g., `branch-operator`)
     - Assign the following global permissions:
       - `global:merge_branch:allow_all`
       - `global:merge_proposed_change:allow_all`

  3. **Assign the role to the user**
     - Create a group and add the user to it
     - Assign the `branch-operator` role to the group

  4. **Login as the non-admin user**
     - Logout from admin account
     - Login with the `testuser` credentials

  5. **Create a branch and attempt to rebase**
     - Create a new branch (e.g., `test-branch`)
     - Make some changes on the branch
     - Attempt to rebase the branch onto `main`

  6. **Observe the error**
     - The rebase operation fails with a permission denied error
     - The user cannot rebase despite having merge permissions

### Additional Information

_No response_

gmazoyer:
## Why & What

Until now rebasing a branch required super_admin, even when the user already had merge_branch and merge_proposed_change. That left admins with no way to delegate rebase to a regular user.

Add a rebase_branch global permission alongside the existing merge permissions, register it during first time initialization, and add a checker that gates the BranchRebase mutation.

Closes #8050

In a follow-up PR, that will target `develop`, I will probably try to get rid of checkers for branch merge and branch rebase. Having `raise_for_permission` call in the mutation code should suffice. This PR targets `stable` that why I went with another checker, to be consistent with what already exists.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/backend/infrahub/graphql/auth/query_permission_checker/rebase_operation_checker.py
+++ b/backend/infrahub/graphql/auth/query_permission_checker/rebase_operation_checker.py
@@ -1,0 +1,35 @@
+from infrahub import config
+from infrahub.auth import AccountSession
+from infrahub.core.account import GlobalPermission
+from infrahub.core.branch import Branch
+from infrahub.core.constants import GlobalPermissions, PermissionDecision
+from infrahub.database import InfrahubDatabase
+from infrahub.graphql.analyzer import InfrahubGraphQLQueryAnalyzer
+from infrahub.graphql.initialization import GraphqlParams
+
+from .interface import CheckerResolution, GraphQLQueryPermissionCheckerInterface
+
+
+class RebaseBranchPermissionChecker(GraphQLQueryPermissionCheckerInterface):
+    """Checker that makes sure a user account can rebase a branch."""
+
+    permission_required = GlobalPermission(
+        action=GlobalPermissions.REBASE_BRANCH.value, decision=PermissionDecision.ALLOW_ALL.value
+    )
+
+    async def supports(self, db: InfrahubDatabase, account_session: AccountSession, branch: Branch) -> bool:  # noqa: ARG002
+        return config.SETTINGS.main.allow_anonymous_access or account_session.authenticated
+
+    async def check(
+        self,
+        db: InfrahubDatabase,  # noqa: ARG002
+        account_session: AccountSession,  # noqa: ARG002
+        analyzed_query: InfrahubGraphQLQueryAnalyzer,
+        query_parameters: GraphqlParams,
+        branch: Branch,  # noqa: ARG002
+    ) -> CheckerResolution:
+        if "BranchRebase" in [operation.name for operation in analyzed_query.operations]:
+            query_parameters.context.active_permissions.raise_for_permission(permission=self.permission_required)
+            return CheckerResolution.TERMINATE
+
+        return CheckerResolution.NEXT_CHECKER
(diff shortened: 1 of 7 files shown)


backend/infrahub/graphql/auth/query_permission_checker/rebase_operation_checker.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class RebaseBranchPermissionChecker(GraphQLQueryPermissionCheckerInterface):
+    """Checker that makes sure a user account can rebase a branch."""
+
+    permission_required = GlobalPermission(
+        action=GlobalPermissions.REBASE_BRANCH.value, decision=PermissionDecision.ALLOW_ALL.value
+    )
+
+    async def supports(self, db: InfrahubDatabase, account_session: AccountSession, branch: Branch) -> bool:  # noqa: ARG002
```
polmichel, line 20: This is an optional comment:

It seems that all checkers are using this exact logic, except `backend/infrahub/graphql/auth/query_permission_checker/anonymous_checker.py`.

Would promote this behavior as default behavior of the interface be a good idea? The down side of this solution would be that persons that would implement future other classes could miss customizing this behavior.

Otherwise, creating another abstract class named `AuthenticatedQueryPermissionCheckerBase` which overrides the `support` method could be another option?

backend/tests/component/graphql/auth/query_permission_checker/test_rebase_operation_checker.py
polmichel, line 1: Nice coverage

polmichel (approved):
LGTM! I have just written an optional comment

backend/infrahub/graphql/auth/query_permission_checker/rebase_operation_checker.py
ajtmccarty: I think we should keep it as is. it's a simple conditional that explicitly links the the actual `check` logic to its expected pre-conditions. and, like you said, it forces anyone adding a new subclass of `GraphQLQueryPermissionCheckerInterface` to consider when the `check` method should apply
I like DRY code too, but, in this case, I think forcing each subclass to write its own `supports` method is the better option

ajtmccarty (approved):

Issue #8945, opened 2026-04-16
task: Use GraphQL variables instead of inlining filter values in query strings


ogenstad:
## Summary

Six locations build GraphQL queries or mutations using `infrahub_sdk.graphql.Query` or `infrahub_sdk.graphql.Mutation` and embed runtime values as literals directly in the rendered string, rather than passing them as GraphQL variables. This bypasses query parameterization.

Identified at 86b583af (`infrahub-v1.8.4`).

## Reason for the change

The server can cache GraphQL queries after rendering them, but if we send in GraphQL queries with static values instead of variables we cache a new version of the same query instead and doing so has a slight hit to performance.

As part of this change we should update the knowledge base under ./dev to ensure that we don't introduce new queries that doesn't use variables for everything.

## Affected Locations

| File | Function | Line(s) | Hardcoded value(s) |
|------|----------|---------|---------------------|
| `backend/infrahub/actions/tasks.py` | `get_generator_run_query()` | 37, 75 | `definition_id`, `target_ids` |
| `backend/infrahub/actions/tasks.py` | `add_node_to_group()` | 128 | `group_id`, `node_id` |
| `backend/infrahub/actions/tasks.py` | `remove_node_from_group()` | 150 | `group_id`, `node_id` |
| `backend/infrahub/computed_attribute/models.py` | `ComputedAttrJinja2GraphQL.render_graphql_query()` | 356 | `filter_id` |
| `backend/infrahub/display_labels/models.py` | `DisplayLabelJinja2GraphQL.render_graphql_query()` | 185 | `filter_id` |
| `backend/infrahub/hfid/models.py` | `HFIDGraphQL.render_graphql_query()` | 185 | `filter_id` |

## Current (incorrect) patterns

Query with inlined filter value:

```python
query = Query(
    query={Kind: {"@filters": {filter_key: filter_id}, ...}}
).render()
client.execute_graphql(query=query, branch_name=branch_name)
```

Mutation with inlined input values:

```python
mutation = Mutation(
    mutation="RelationshipAdd",
    input_data={"data": {"id": group_id, "name": "members", "nodes": [{"id": node_id}]}},
    query={"ok": None},
)
client.execute_graphql(query=mutation.render(), branch_name=branch_name)
```

## Expected (correct) patterns

Query using GraphQL variables:

```python
query = Query(
    query={Kind: {"@filters": {filter_key: "$filterId"}, ...}},
    variables={"filterId": "String!"},
).render()
[...]

PR #9126, opened 2026-05-04
IFC-2536: Migrate InfrahubClient to graphql query for get and filters methods

solababs:
# Why

Prefect task files were making overfetching calls to `client.all()`, `client.filters()`, and `client.get()` that returned every attribute, relationship, and metadata field — but the tasks only consumed `id` (or a handful of fields). This caused unnecessary Neo4j reads, large GraphQL payloads, and slower task execution on every background worker cycle.

This PR replaces those calls with targeted `execute_graphql()` calls backed by purpose-built query model classes that request only the fields each task actually uses.

## How to test

```bash
# Unit tests for every query model class
uv run pytest backend/tests/unit/display_labels/ backend/tests/unit/hfid/ \
    backend/tests/unit/git/test_git_repository_query.py \
    backend/tests/unit/generators/test_generator_instance_query.py \
    backend/tests/unit/computed_attribute/ -v

# Benchmark (200 nodes, 5 iterations — requires Neo4j testcontainer)
INFRAHUB_LOG_LEVEL=CRITICAL uv run pytest -v \
    backend/tests/benchmark/test_query_optimization.py --dist=no -s

# Full functional suite
uv run invoke backend.test-functional
```

Benchmark expected output (≥30% time reduction, ≥50% size reduction per task):

| Task | Full fetch | Optimized | Time ↓ | Payload ↓ |
|------|-----------|-----------|--------|-----------|
| `display_labels` | 485 ms / 65 KB | 97 ms / 12 KB | **80%** | **82%** |
| `hfid` | 120 ms / 65 KB | 28 ms / 12 KB | **76%** | **82%** |
| `git` | 263 ms / 164 KB | 90 ms / 31 KB | **66%** | **81%** |

## Impact & rollout

- **Backward compatibility:** no breaking changes — task outputs are identical, only the internal fetch mechanism changes
- **Performance:** 65–80% execution time reduction and 80–82% payload size reduction for the six migrated call sites
- **Config/env changes:** none
- **Deployment notes:** safe to deploy; no migrations, no schema changes, no flag required

## Checklist

- [x] Tests added/updated


backend/infrahub/generators/models.py
```diff
@@ -1,8 +1,55 @@
...
 
 
+class GeneratorInstanceNode(BaseModel):
+    model_config = ConfigDict(frozen=True)
+
+    id: str
+    status: str
+
+
+class GeneratorInstanceQuery(BaseModel):
```
dgarros: are you using the Pydantic model generator from the SDK or have you generated this class using Claude directly ?

solababs: This is generated using claude

backend/infrahub/computed_attribute/graphql_queries.py
```diff
@@ -1,9 +1,116 @@
...
+
+
+class ComputedAttributeTransformQuery(BaseModel):
+    query_name: ClassVar[str] = "ComputedAttributeFetchTransform"
+    transform_id: str
+
+    def render_query(self) -> str:
+        filters: dict[str, str | list[str]] = (
+            {"ids": [self.transform_id]} if _is_uuid(self.transform_id) else {"name__value": self.transform_id}
+        )
```
ogenstad: We will need a small change with regards to how this query would be executed. If you check this issue #8945. In short there's a caching feature of the backend that would cache the rendering and processing of the GraphQL queries, i.e. to convert them into a GraphQL document. But this only happens if the raw text of the query looks exactly the same. As such we don't want to have `self.transform_id` be hardcoded into the body of the query but instead we'd need to send it as variables in the SDK call to `await client.execute_graphql()`

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -82,47 +82,56 @@ async def process_transform(
...
-            kind=CoreTransformPython,
-            branch=branch_name,
-            id=transform_attribute.transform,
-            prefetch_relationships=True,
-            populate_store=True,
-        )
+        if not transform_attribute.transform:
+            continue
+        transform_query = ComputedAttributeTransformQuery(transform_id=transform_attribute.transform)
+        transform_response = await client.execute_graphql(query=transform_query.render_query(), branch_name=branch_name)
```
ogenstad: I think a call like this should be fine with regards to pagination as we'd only expect to get a single result back, i.e. regarding the comments on pagination on the other PR.

backend/infrahub/generators/models.py
ogenstad: What we can do with the Pydantic model generator (https://docs.infrahub.app/infrahubctl/infrahubctl-graphql) is to have a reference to the GraphQL query in text and then by having it there we can run checks in the CI to validate that the query we've written is up to date based on the current core GraphQL schema shipped with Infrahub, i.e. the one stored here: https://github.com/opsmill/infrahub/blob/infrahub-v1.9.3/schema/schema.graphql. Basically together with a GraphQL query and the GraphQL schema we can generate Python bindings and validate that they are up to date in the CI. One of the thoughts around this JPD card was that we'd use this approach but it seems like this information was lost in the creation of that card. So the primary goal of the JPD card is to improve the performance by avoiding over fetching data, the secondary goal was to have queries defined as text and be able to both validate that they are up to date and that we have generated classes that match the queries. The intention is to start using this approach in multiple places within the backend but also get started in the frontend. This approach would not work for the changes in #9123 though since in that work we're directly querying for user defined nodes that doesn't exist in the core schema. But for the work in this PR it should be possible.

.github/workflows/ci.yml
```diff
@@ -280,9 +280,17 @@ jobs:
...
+            --exclude python_sdk \
+            --exclude backend/infrahub/computed_attribute/graphql_queries/computed_attribute_fetch_transform.py \
+            --exclude backend/infrahub/generators/graphql_queries/generator_instance_fetch.py
       - name: "Linting: ruff format"
-        run: "uv run ruff format --check --diff --exclude python_sdk ."
+        run: |
+          uv run ruff format --check --diff . \
+            --exclude python_sdk \
+            --exclude backend/infrahub/computed_attribute/graphql_queries/computed_attribute_fetch_transform.py \
+            --exclude backend/infrahub/generators/graphql_queries/generator_instance_fetch.py
```
ogenstad: These exclude lines are a bit problematic for anyone just running "ruff check ." which is a lot faster than using the invoke commands. I'd wonder what's failing in these generated files. Might it just be a matter of us having to run `ruff check [filenames] --fix` and `ruff format [filenames]`? I.e. when we generate them we should be able to format them correctly provided that there are no other violations in these files. For reference:

https://github.com/opsmill/infrahub/blob/infrahub-v1.9.3/tasks/backend.py#L260-L261

backend/infrahub/computed_attribute/graphql_queries/computed_attribute_fetch_transform.py
ogenstad, line 1: Note that we have a file called `backend/infrahub/computed_attribute/graphql_queries.py` in #9123 which would conflict with the module directory here `backend/infrahub/computed_attribute/graphql_queries/`. So could be that this PR should be rebased and reworked once the other one lives in `develop`

Actually this is probably the merge conflict for this PR.

backend/infrahub/computed_attribute/graphql_queries/queries.py
```diff
@@ -0,0 +1,139 @@
...
+                        name {
+                            value
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+"""
```
ogenstad: I think we'd want the same approach for this as you have a bit further down where we have `generator_instance_fetch.gql` as a separate file. It make it clearer that we'd expect to the bindings for this be generated from the actual file. As it is inline here it looks like it's just part of the code.

Also we can probably combine the two queries TRANSFORM_QUERY_BY_NAME and TRANSFORM_QUERY_BY_IDS into one, then we just have the queries accept both of the parameters and it would use whatever we send in with with the variables.

backend/infrahub/computed_attribute/graphql_queries/queries.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def get_variables(self) -> dict[str, Any]:
+        if _is_uuid(self.transform_id):
+            return {"transform_ids": [self.transform_id]}
+        return {"transform_name": self.transform_id}
+
+    def render_query(self) -> str:
+        if _is_uuid(self.transform_id):
+            return TRANSFORM_QUERY_BY_IDS
+        return TRANSFORM_QUERY_BY_NAME
```
ogenstad: Combining this into one query as per above would remove the if-statement here and we'd only have the `get_variables()` one.

backend/infrahub/generators/tasks.py
```diff
@@ -109,16 +110,25 @@ async def _define_instance(model: RequestGeneratorRun, client: InfrahubClient) -
...
-                definition__ids=[model.generator_definition.definition_id],
-                object__ids=[model.target_id],
-                branch=model.branch_name,
+            gen_query = GeneratorInstanceQuery(
+                definition_id=model.generator_definition.definition_id,
+                object_id=model.target_id,
             )
-            if instances:
-                instance = instances[0]
+            response = await client.execute_graphql(
```
ogenstad: I think that we might want to have pagination for this query if there's a high number of generator instances.

solababs: this request is for one instance, uses `instance_nodes[0].id` below

ogenstad: Hey, looking at this I'm wondering if we for how should revert the changes in this `_define_instance()`, function. The reason is that looking at the new file we issue the custom query at line 117, but we still need to issue a new query at line 124. So even if the initial query might be a bit more optimized we still end up with the new client.get() one. Probably the reason here is that we're expected to return as `CoreGeneratorInstance` instance and the call to client returns such an object. In the future we could provide a better way to create those objects or perhaps do a larger refactoring. But for now I think we should stick to `client.filters` as we had it so that we don't have to query the backend twice. We'd need to do it twice anyway if the generator instance didn't exist, but that's a separate issue and would be expected.

ogenstad (approved):

Issue #8871, opened 2026-04-10
bug: branch delete task stuck in e2e tests


fatih-acar:
### Component

API Server / GraphQL

### Infrahub version

1.8.4

### Current Behavior

# Race condition: concurrent neo4j driver session usage in Prefect task workers
 
## Summary
 
A race condition in how the neo4j database session is shared across concurrent Prefect flow/task handlers causes intermittent `RuntimeError: read() called while another coroutine is already waiting for incoming data`. This manifests as flaky CI failures, most visibly during the `validate-schema-number-pools` subflow that runs as part of post-merge branch cleanup.
 
## Symptoms
 
- The `validate-schema-number-pools` subflow fails with a `RuntimeError` from the async neo4j driver's socket layer.
- In some cases the branch-delete event handler gets stuck for several minutes (7+ min observed) before failing.
- The error is intermittent and has been reproduced across multiple CI runs over several days.
 
## Error
 
```
RuntimeError: read() called while another coroutine is already waiting for incoming data
```
 
Full traceback (from CI logs):
 
```
File "/source/backend/infrahub/pools/tasks.py", line 31, in validate_schema_number_pools
    updated_branches = await synchronizer.run(user_id=context.account.account_id)
  ...
  File "/source/backend/infrahub/core/manager.py", line 187, in query
    await query.execute(db=db)
  ...
  File "/source/backend/infrahub/database/__init__.py", line 408, in run_query
    response = await execution_method.run(query=_query, parameters=params)
  ...
  File "neo4j/_async/io/_common.py", line 53, in _buffer_one_chunk
    await receive_into_buffer(self._socket, self._buffer, 2)
  ...
RuntimeError: read() called while another coroutine is already waiting for incoming data
```
 
## Root cause analysis
 
`InfrahubWorkerAsync.run` injects `self.service` (and by extension `self.service.database`) into flow function parameters:
 
```python
inject_service_parameter(func=flow_func, parameters=flow_run.parameters, service=self.service)
```
 
When multiple flows or subflows execute concurrently, they all share the same `service.database` instance. The neo4j async driver does not support concurrent reads on the same connection/session, so two coroutines trying to read from the same socket at the same time triggers the `RuntimeError`.
 
[...]

PR #9127, opened 2026-05-04
Database sessions reused through Prefect flows signature and through get_database()

### LINKED ISSUE #9121: bug: proposed change stuck in 'merging' state when task worker crashes — no recovery path
### Component

API Server / GraphQL

### Infrahub version

1.9.2

### Current Behavior

When a task worker crashes during a proposed change merge, the proposed change is left in the `merging` state with no way to recover. All subsequent mutations on the proposed change fail with:

```json
{
  "data": {
    "CoreProposedChangeUpdate": null
  },
  "errors": [
    {
      "message": "A proposed change in the merging state is not allowed to be updated",
      "locations": [{"line": 2, "column": 3}],
      "path": ["CoreProposedChangeUpdate"]
    }
  ]
}
```

The validation comes from `ProposedChangeState.validate_updatable()` in [backend/infrahub/proposed_change/constants.py:32-36](backend/infrahub/proposed_change/constants.py#L32-L36), which rejects updates whenever the state is anything other than `open`. Since `merging` is treated as a "completed" state by `is_completed`, the proposed change is permanently blocked from any further action — it cannot be re-merged, closed, canceled, or reset.

### Expected Behavior

There should be a recovery path when the merge worker crashes. At minimum, an admin user should be able to reset the proposed change state back to `open` (or another safe state) so the merge can be retried or the proposed change can be closed.

Possible approaches:

- A dedicated GraphQL mutation (e.g. `CoreProposedChangeResetState`) restricted to admin users that forces the state back to `open`.
- Allow state transitions out of `merging` for admin users via the existing update mutation.
- Detect crashed merge tasks (e.g. on worker restart or via Prefect) and automatically reset the proposed change state.

### Steps to Reproduce

1. Trigger a proposed change merge.
2. Kill the task worker mid-merge (e.g. SIGKILL the Prefect worker handling the merge flow).
3. Observe that the proposed change remains in the `merging` state indefinitely.
4. Attempt any mutation on the proposed change (update, close, cancel, re-merge) — all are rejected.

### Additional Information

Related: #9115 (failed merge can leave database in partially-merged state) — that issue covers the data-side rollback; this issue is specifically about the proposed change being unrecoverable from the user/API perspective even after the underlying data is sorted out.

The merge flow sets the state to `merging` in [backend/infrahub/graphql/mutations/proposed_change.py:471](backend/infrahub/graphql/mutations/proposed_change.py#L471) and is expected to flip it to `merged` on success or back to `open` on failure, but a worker crash bypasses both paths.

polmichel:
## Why 

Related to the E2E tests flakiness issue https://github.com/opsmill/infrahub/issues/8871 or no recovery path issue https://github.com/opsmill/infrahub/issues/9121, it seems that database sessions or transactions are not correctly managed during the application lifecycle.

Closes https://github.com/opsmill/infrahub/issues/9121 and [[IFC-2551]](https://opsmill.atlassian.net/browse/IFC-2551)

## What this PR is about

In a specific Prefect worker, same database instance/session is reused through Prefect flows signature when service is injected through `infrahub.workers.utils.inject_service_parameter`.
This PR aims to make sure that when the database is injected through this `service` parameter, that a new database session is created if necessary.

#### Touch service.database directly (race-condition candidates):

- `validate_schema_number_pools`
- `branch_merged`
- `schema_updated`
- `configure_action_rules`
 
#### Don't touch service.database directly (only service.client HTTP / cache / message_bus):

- `add_node_to_group`
- `remove_node_from_group`
- `run_generator`
- `run_generator_group_event`
- `clean_up_deadlocks`

### Solution 

When the flow use `service.database`:

- make sure that a new database session is created. 
- check other potential callers
- put at the top level the database session creation if it is nested without any value added

When the flow is not using `service.database`:
- remove service parameter from signature if it is unused

### Testing part

A new introspective test has been created in order to ensure that, in each Prefect flow method, we do NOT use `service.database` without calling it in a `start_session()` or `start_transaction()` bloc.

### Additional callers fixed outside Prefect flow scope

The introspective test only scans Prefect flows that take a `service: InfrahubServices` parameter. A manual audit of all `service.database` and `get_database()` callers surfaced four more sites that were running queries against the DRIVER-mode singleton:

- `graphql/initialization.py` — `prepare_graphql_params` was loading permissions via `permissions.load_permissions(db=db, ...)` directly on the passed-in db. Now wrapped in
`start_session(read_only=True)`. (de21cabc)
- `proposed_change/tasks.py` — `merge_proposed_change` was running `registry.manager.get_one()` against the DRIVER db before entering the existing `start_session()` block. Moved the lookup inside the session. (4c2f1907)
- `cli/__init__.py` — the `infrahub shell` command had `async with service.database as db:`, which is a silent no-op because `__aenter__` on a DRIVER-mode `InfrahubDatabase`does not open a neo4j session. Changed to `start_session()`. (68e682c131423ded3f32b853958417ce63f38949)
- `services/component.py` — `InfrahubComponent.refresh_schema_hash` calls `registry.get_branch(db=self.db)` for every branch, where `self.db` is the DRIVER singleton stored at construction. Wrapped the loop in `start_session(read_only=True)`. (74d9f1b4)

[IFC-2551]: https://opsmill.atlassian.net/browse/IFC-2551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/cli/__init__.py
+++ b/backend/infrahub/cli/__init__.py
@@ -92,8 +92,8 @@ def shell() -> None:
         )
         initialize_lock(service=service)
 
-        async with service.database as db:
-            await initialization(db=db)
+        async with service.database.start_session() as safe_db:
+            await initialization(db=safe_db)
         await service.component.refresh_schema_hash()
 
         return service
--- a/backend/infrahub/cli/__init__.py
+++ b/backend/infrahub/cli/__init__.py
@@ -92,8 +92,8 @@ def shell() -> None:
         )
         initialize_lock(service=service)
 
-        async with service.database.start_session() as safe_db:
-            await initialization(db=safe_db)
+        async with service.database.start_session() as db:
+            await initialization(db=db)
         await service.component.refresh_schema_hash()
 
         return service
--- a/changelog/8871.fixed.md
+++ b/changelog/8871.fixed.md
@@ -1,1 +1,0 @@
-Fix intermittent `RuntimeError: read() called while another coroutine is already waiting for incoming data` that could occur during concurrent Prefect flow execution.
--- a/changelog/8871.fixed.md
+++ b/changelog/8871.fixed.md
@@ -1,0 +1,1 @@
+Fix intermittent `RuntimeError: read() called while another coroutine is already waiting for incoming data` that could occur during concurrent Prefect flow execution.
--- a/backend/infrahub/actions/tasks.py
+++ b/backend/infrahub/actions/tasks.py
@@ -163,7 +163,6 @@ async def run_generator(
     node_ids: list[str],
     generator_definition_id: str,
     context: InfrahubContext,
-    service: InfrahubServices,  # noqa: ARG001
 ) -> None:
     await add_tags(branches=[branch_name], nodes=node_ids + [generator_definition_id])
 
--- a/backend/infrahub/actions/tasks.py
+++ b/backend/infrahub/actions/tasks.py
@@ -208,11 +208,12 @@ async def run_generator_group_event(
 async def configure_action_rules(
     service: InfrahubServices,
 ) -> None:
-    await setup_triggers_specific(
-        gatherer=gather_trigger_action_rules,  # type: ignore[arg-type]
-        trigger_type=TriggerType.ACTION_TRIGGER_RULE,
-        db=service.database,
-    )
+    async with service.database.start_session(read_only=True) as db:
+        await setup_triggers_specific(
+            gatherer=gather_trigger_action_rules,  # type: ignore[arg-type]
+            trigger_type=TriggerType.ACTION_TRIGGER_RULE,
+            db=db,
+        )
 
 
 async def _get_targets(
--- a/backend/infrahub/actions/tasks.py
+++ b/backend/infrahub/actions/tasks.py
@@ -186,7 +186,6 @@ async def run_generator_group_event(
     members: list[EventGroupMember],
     generator_definition_id: str,
     context: InfrahubContext,
-    service: InfrahubServices,  # noqa: ARG001
 ) -> None:
     node_ids = [node.id for node in members]
     await add_tags(branches=[branch_name], nodes=node_ids + [generator_definition_id])
--- a/backend/infrahub/branch/tasks.py
+++ b/backend/infrahub/branch/tasks.py
@@ -22,13 +22,12 @@ async def branch_merged(
 ) -> None:
     target_branch = target_branch or registry.default_branch
     log = get_run_logger()
-    await wait_for_schema_to_converge(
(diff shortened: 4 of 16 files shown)


fatih-acar (changes requested):
I think this is not enough
We still have places where we use `get_database()` with no dedicated sessions, like https://github.com/opsmill/infrahub/blob/infrahub-v1.9.2/backend/infrahub/proposed_change/tasks.py#L172

changelog/8871.fixed.md
```diff
@@ -0,0 +1 @@
+Fix intermittent `RuntimeError: read() called while another coroutine is already waiting for incoming data` that could occur during concurrent Prefect flow execution.
```
fatih-acar: This also fix non-deterministic critical issues such as infinite loops within API and/or task workers due to race conditions across coroutines. The impact is wide and critical, so the changelog should reflect this since we are going to do an immediate release.

polmichel: You are right, I will change that

polmichel:
> I think this is not enough We still have places where we use `get_database()` with no dedicated sessions, like https://github.com/opsmill/infrahub/blob/infrahub-v1.9.2/backend/infrahub/proposed_change/tasks.py#L172

This PR was meant to be about Prefect flows signature issue only, I did not thought we had to release right away a new version. I will take care of these additional `get_database()` calls and will look for other similar issues.

Additionally, I will not reference https://github.com/opsmill/infrahub/issues/8871 but https://github.com/opsmill/infrahub/issues/9121 as the resolved issue. https://github.com/opsmill/infrahub/issues/8871 will be closed as duplicate.

polmichel:
> https://github.com/opsmill/infrahub/blob/infrahub-v1.9.2/backend/infrahub/proposed_change/tasks.py#L172

This was a Prefect flow, and I missed it.

backend/infrahub/cli/__init__.py
```diff
@@ -92,8 +92,8 @@ async def initialize_service() -> InfrahubServices:
         )
         initialize_lock(service=service)
 
-        async with service.database as db:
-            await initialization(db=db)
+        async with service.database.start_session() as safe_db:
```
ajtmccarty: I think you can leave the variable name as `db` here

ajtmccarty (approved):

backend/infrahub/cli/__init__.py
polmichel: There was an IDE warning about outer scope variable having the same name, but I do not think we really care, I do not find the variable. Thanks

fatih-acar (approved):

PR #9133, opened 2026-05-05
chore: prep release 1.9.3

polmichel:
PR for release 1.9.3 preparation


ogenstad (approved):
LGTM, see that the integration test is failing but looks like it's just network errors so I'd assume that it will succeed after a retry.

PR #9142, opened 2026-05-05
Add per-instance group membership fields on Object Templates

### LINKED ISSUE #9094: bug: member_of_groups on an Object Template applies to the template itself and is not inherited by instances
### Component

API Server / GraphQL

### Infrahub version

1.8.6 and 1.9.0

### Current Behavior

When a `member_of_groups` relationship is configured on an Object Template, two unexpected things happen:

1. The template node itself becomes a member of the referenced group as soon as the template is saved.
2. Objects subsequently created from that template do **not** inherit the group membership — the group relationship is silently dropped.

Effectively, `member_of_groups` set on a template applies to the template, never to its instances.

Confirmed in source:

- `add_groups()` in `backend/infrahub/core/schema/schema_branch.py` injects a `member_of_groups` relationship into nearly every node kind. Templates are not excluded, so setting `member_of_groups` on a template literally adds the template node to that real group.
- Template instantiation in `backend/infrahub/templates/node_applier.py` only propagates relationships whose kind is `ATTRIBUTE`, `GENERIC`, or `PROFILE`:

  ```python
  if rel_schema.kind not in [RelationshipKind.ATTRIBUTE, RelationshipKind.GENERIC, RelationshipKind.PROFILE]:
      continue
  ```

  `member_of_groups` is registered with `kind=RelationshipKind.GROUP`, so it is skipped at instantiation time.

### Expected Behavior

Configuring `member_of_groups` on an Object Template should express intent for the *resulting instances*, not for the template itself. Specifically:

- Setting `member_of_groups` on a template should not add the template node to the group.
- Each object created from the template should be added as a member of the configured group(s) automatically at creation time.

This is structurally the same problem as #6551 (selecting a resource pool on a template allocated a resource to the template instead of to instances). Resolving it consistently with that precedent — e.g. a template-side field that drives per-instance group membership — would be ideal.

### Steps to Reproduce

1. Spin up an Infrahub instance.
2. Load a schema with a node that has `generate_template: true`, e.g.:

   ```yaml
   version: "1.0"
   nodes:
     - name: Device
       namespace: Infra
       generate_template: true
       attributes:
         - name: name
           kind: Text
           unique: true
   ```

3. Create a `CoreStandardGroup` (e.g. `staged_devices`).
4. Create an object template for `InfraDevice` and add `staged_devices` to its `member_of_groups`.
5. Inspect the group: the template node is listed as a member. *(Behavior 1)*
6. Create a new `InfraDevice` from that template.
7. Inspect the new device's `member_of_groups` and the `staged_devices` group: the new device is not a member. *(Behavior 2)*

### Additional Information

- Related/precedent: #6551 (resource pool on template allocated to template, not instance — same shape, fixed via a template-side `<rel>_from_resource_pool` field).
- Source references in current `develop`:
  - `backend/infrahub/core/schema/schema_branch.py` — `add_groups()` (~line 2195)
  - `backend/infrahub/templates/node_applier.py` — `_apply_relationships()` (~line 60), specifically the relationship-kind filter (~line 80)
- Practical impact: event-driven workflows that rely on a `CoreGroupTriggerRule` firing when a template-instantiated object is added to a group cannot use `member_of_groups` on the template to drive that membership today; a separate `CoreNodeTriggerRule` + `CoreGroupAction` is required as a workaround.

gmazoyer:
# Why

When you set `member_of_groups` on an Object Template today, the template node itself becomes a member of the group, but objects created from that template don't inherit the membership. So a template can express "things like this" for attributes and most relationships, but not for groups.

This change adds a way to express per-instance group membership on a template, mirroring the `_from_resource_pool` sister-field pattern we already use for IPs and number pools.

Closes #9094

## What changed

Templates now expose two extra relationships: `member_of_groups_for_instances` and `subscriber_of_groups_for_instances`. They have kind `GENERIC`, so adding peers there does **not** put the template in those groups.

At template application time, `NodeTemplateApplier` reads the peers from those fields and writes them under `member_of_groups` / `subscriber_of_groups` on the new object. User-supplied groups on the create call still take precedence.

The existing `member_of_groups` and `subscriber_of_groups` on a template keep their original meaning: they describe the template's own membership and are not propagated to instances.

## How to test

```bash
uv run pytest backend/tests/unit/core/schema/test_template_group_relationships.py
uv run pytest backend/tests/component/templates/test_template_applier.py
```

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/9094.added.md`)
- [x] External docs updated (`docs/docs/topics/object-template.mdx`, `docs/docs/guides/object-template.mdx`, `docs/docs/topics/groups.mdx`)
- [x] Internal .md docs updated (`dev/knowledge/backend/templates.md` added; `architecture.md` and `schema-definitions.md` cross-link to it)


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2141,9 +2141,43 @@ def add_groups(self) -> None:
...
+        self, schema: MainSchemaTypes, changed: bool
+    ) -> tuple[MainSchemaTypes, bool]:
+        """Append `member_of_groups_for_instances` / `subscriber_of_groups_for_instances` on a template.
+
+        These are generic relationships using distinct identifiers so adding peers does not put the template node into
+        a group.
+        """
+        for rel_name, identifier in (
+            ("member_of_groups_for_instances", "template_group_member_for_instances"),
+            ("subscriber_of_groups_for_instances", "template_group_subscriber_for_instances"),
```
gmazoyer, line 2158: Names are up for debate.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2141,9 +2141,43 @@ def add_groups(self) -> None:
...
 
+            # On auto-generated templates, also expose template-side fields whose peers drive instance group membership at template application time
+            if isinstance(schema, TemplateSchema):
+                schema, changed = self._add_template_group_for_instances_relationships(schema=schema, changed=changed)
+
             if changed:
                 self.set(name=node_name, schema=schema)
 
+    def _add_template_group_for_instances_relationships(
+        self, schema: MainSchemaTypes, changed: bool
```
ajtmccarty: `changed` seems like it should be inverted and called `duplicate_before_update` or something similar

backend/tests/unit/core/schema/test_template_group_relationships.py
ajtmccarty, line 1: this one be a `unit` test

docs/docs/guides/object-template.mdx
```diff
@@ -588,3 +588,88 @@ For more information about resource pools, see:
 
 - [Managing resource pools](../guides/resource-manager.mdx) - Guide for creating and using pools
 - [Resource Manager concepts](../topics/resource-manager.mdx) - Deep dive into allocation behaviour
+
+## Assigning groups via templates
+
+Object templates expose a dedicated `member_of_groups_for_instances` relationship (and a `subscriber_of_groups_for_instances` counterpart). Groups assigned through this field are propagated to every object created from the template, while the template node itself is **not** added to those groups. This is useful when every instance produced from a given template should belong to the same operational group — for example, "all devices created from `staged-spine` go into the `staged_devices` group" — without having to wire up a `CoreNodeTriggerRule` and `CoreGroupAction`.
```
ajtmccarty: - `template node` -> `template object`
- seems unnecessary to me `without having to wire up a CoreNodeTriggerRule and CoreGroupAction.`

docs/docs/guides/object-template.mdx
ajtmccarty, line 1: - I think we're still in some sort of docs freeze, so this may need to change once that is unfrozen
- personally, I'd say we should try to remove the em-dashes that the LLMs love so much b/c it makes it look like our docs are written by robots and that we are lazy

docs/docs/guides/object-template.mdx
```diff
@@ -588,3 +588,88 @@ For more information about resource pools, see:
...
+
+:::info
+
+If you want the template node itself to belong to a group (for example, to track templates as a set), use the regular `member_of_groups` field on the template instead — it behaves like on any other node and is independent from the `_for_instances` field.
```
ajtmccarty: `node` - > `object`

docs/docs/topics/object-template.mdx
```diff
@@ -47,6 +47,27 @@ Without pool integration, templates can only set static values. Any attribute th
 
 For step-by-step instructions, see [Allocating resources from pools via templates](../guides/object-template.mdx#allocating-resources-from-pools-via-templates).
 
+## Integration with Groups
+
+Templates expose two distinct group fields, mirroring the [resource-pool sister-field pattern](#integration-with-resource-pools):
+
+- `member_of_groups` / `subscriber_of_groups` — the regular group relationships. Peers added here put the **template itself** in the group, just like on any other node.
+- `member_of_groups_for_instances` / `subscriber_of_groups_for_instances` — template-only fields whose peers are propagated to **each object created from the template** at creation time. The template node itself is not added to these groups.
```
ajtmccarty: `node` -> `object`

docs/docs/topics/object-template.mdx
```diff
@@ -47,6 +47,27 @@ Without pool integration, templates can only set static values. Any attribute th
...
+
+### How group propagation works
+
+When you create an object from a template, Infrahub reads the template's `member_of_groups_for_instances` and `subscriber_of_groups_for_instances` peers and writes them onto the new object's `member_of_groups` and `subscriber_of_groups`. Groups supplied directly on the create call take precedence over the template's `_for_instances` peers.
+
+The template's own `member_of_groups` (and `subscriber_of_groups`) are never propagated to instances — those describe the template's own membership, not its instances'.
+
+### When to use it
+
+Use `member_of_groups_for_instances` when you want to express "every device created from `staged-spine` should belong to the `staged_devices` group" without writing a separate `CoreNodeTriggerRule` plus `CoreGroupAction`. Workflows driven by `CoreGroupTriggerRule` fire naturally as each instance is added to the configured groups.
```
ajtmccarty: instead of "without writing a separate `CoreNodeTriggerRule` plus `CoreGroupAction`", should this be "without using a generator"? I'm truly not sure which we prefer in the docs

ajtmccarty (approved):
functionally looks great
all comments are little naming/docs things

backend/infrahub/core/schema/schema_branch.py
gmazoyer: I changed the var name but did not call it `duplicate_before_update` because that would imply to do a bunch of `not` swap to make it behaves as its name would imply; and that would have made this hard to follow (nobody wants to read a var that is the result of a "not of not of not of …").

ajtmccarty: good point. name is better now

FragmentedPacket (commented):
Tested locally and using object templates and it successfully added the devices to the group when creating devices from that template.

```
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: CoreStandardGroup
  data:
    - name: test-group
# Device template — management_address allocated from "Device and Interface IP Pool"
# Includes nested interface template — primary_ip also allocated from the same pool
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: TemplateDcimDevice
  data:
    - template_name: "Standard Device"
      management_address_from_resource_pool: "Device and Interface IP Pool"
      member_of_groups_for_instances: ["test-group"]
      interfaces:
        kind: TemplateDcimInterface
        data:
          - template_name: "Standard Device eth0"
            name: eth0
            primary_ip_from_resource_pool: "Device and Interface IP Pool"
```

PR #9164, opened 2026-05-07
reset proposed change status on unexpected merge failure

ajtmccarty:
IFC-2564

broaden the exception caught in `merge_proposed_change` to ensure that the merging Proposed Change's state is always reset to OPEN when a merge fails

### DIFF
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -224,22 +224,27 @@ async def merge_proposed_change(
                         )
 
         log.info("Proposed change is eligible to be merged")
+
+        merge_succeeded = False
         try:
             await merge_branch(branch=source_branch.name, context=context, proposed_change_id=proposed_change_id)
+            merge_succeeded = True
+        except MergeFailedError as exc:
+            return Failed(message=exc.message)
+        except Exception as exc:
+            log.exception("Unexpected failure during proposed change merge")
+            return Failed(message=f"Merge failure when trying to merge {source_branch.name}: {exc}")
         except BaseException as exc:
-            await _proposed_change_transition_state(
-                proposed_change=proposed_change,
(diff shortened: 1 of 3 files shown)


backend/infrahub/proposed_change/tasks.py
```diff
@@ -226,14 +226,20 @@ async def merge_proposed_change(
...
                 user_id=context.account.account_id,
             )
-            return Failed(message=f"Merge failure when trying to merge {exc.message}")
+            if isinstance(exc, MergeFailedError):
+                return Failed(message=f"Merge failure when trying to merge {exc.message}")
+            if isinstance(exc, Exception):
+                log.exception("Unexpected failure during proposed change merge")
+                return Failed(message=f"Merge failure when trying to merge {source_branch.name}: {exc}")
+            # propagate a BaseException
+            raise
```
gmazoyer: Could we rewrite this using a try/except/finally block? I'm thinking about something like:
```python
merge_succeeded = False
try:
    await merge_branch(...)
    merge_succeeded = True
except MergeFailedError as exc:
    return Failed(message=f"Merge failure when trying to merge {exc.message}")
except Exception as exc:
    log.exception("Unexpected failure during proposed change merge")
    return Failed(message=f"Merge failure when trying to merge {source_branch.name}: {exc}")
finally:
    if not merge_succeeded:
        await _proposed_change_transition_state(...)
```
I think that it would be easier to read.

Also the error message that would be displayed by `MergeFailedError` will be awkward because `MergeFailedError` will look lke `Failed to merge branch '{branch_name}'`. so we would read `"Merge failure when trying to merge Failed to merge branch '{branch_name}'"`

ajtmccarty: great suggestion. applied all

gmazoyer (approved):

PR #9184, opened 2026-05-07
add restrict_namespaces to core generics with special handling

ajtmccarty:
add `restricted_namespaces` to some of our internal generic schemas to prevent users from making node schemas that inherit from these generics. the restricted generics are ones that we have specific internal logic for that should not be extensible by users.

these are the generic schemas restricted in this PR. please review carefully to check if any are missing or should be removed
- `CoreCredential`
- `CoreGenericAccount`
- `CoreResourcePool`
- `CoreIPPool`
- `CoreTransformation`
- `CoreBasePermission`
- `CoreMenu`
- `CoreComment`
- `CoreThread`
- `CoreValidator`
- `CoreKeyValue`
- `CoreTriggerRule`
- `CoreAction`
- `CoreNodeTriggerMatch`

### DIFF
--- a/backend/infrahub/core/schema/definitions/core/account.py
+++ b/backend/infrahub/core/schema/definitions/core/account.py
@@ -113,6 +113,7 @@ core_credential = GenericSchema(
     branch=BranchSupportType.AGNOSTIC,
     uniqueness_constraints=[["name__value"]],
     documentation="/topics/auth",
+    restricted_namespaces=["Core"],
     attributes=[
         Attr(name="name", kind="Text", unique=True, order_weight=1000),
         Attr(name="label", kind="Text", optional=True, order_weight=2000),
@@ -134,6 +135,7 @@ core_generic_account = GenericSchema(
     branch=BranchSupportType.AGNOSTIC,
     documentation="/topics/auth",
     uniqueness_constraints=[["name__value"]],
+    restricted_namespaces=["Core"],
     attributes=[
         Attr(name="name", kind="Text", unique=True),
         Attr(name="password", kind="HashedPassword", unique=False),
--- a/backend/infrahub/core/schema/definitions/core/key_value.py
+++ b/backend/infrahub/core/schema/definitions/core/key_value.py
@@ -20,6 +20,7 @@ core_key_value = GenericSchema(
     icon="mdi:key-variant",
     branch=BranchSupportType.AGNOSTIC,
     generate_profile=False,
+    restricted_namespaces=["Core"],
     attributes=[
         Attr(
             name="name",
--- a/backend/infrahub/actions/schema.py
+++ b/backend/infrahub/actions/schema.py
@@ -29,6 +29,7 @@ core_trigger_rule = GenericSchema(
     branch=BranchSupportType.AGNOSTIC,
     uniqueness_constraints=[["name__value"]],
     generate_profile=False,
+    restricted_namespaces=["Core"],
     attributes=[
         Attr(
             name="name",
@@ -90,6 +91,7 @@ core_action = GenericSchema(
     branch=BranchSupportType.AGNOSTIC,
     uniqueness_constraints=[["name__value"]],
     generate_profile=False,
+    restricted_namespaces=["Core"],
     attributes=[
         Attr(
             name="name",
@@ -129,6 +131,7 @@ core_node_trigger_match = GenericSchema(
     label="Node Trigger Match",
     branch=BranchSupportType.AGNOSTIC,
     generate_profile=False,
+    restricted_namespaces=["Core"],
     attributes=[],
     relationships=[
         Rel(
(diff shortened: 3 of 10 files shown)


backend/infrahub/core/schema/definitions/core/key_value.py
```diff
@@ -20,6 +20,7 @@
     icon="mdi:key-variant",
     branch=BranchSupportType.AGNOSTIC,
     generate_profile=False,
+    restricted_namespaces=["Core"],
```
polmichel, line 23: `CoreKeyValue` is meant to be a generic object, not just a Webhook header. It's true that for the webhook case it would make sense to restrict the `CoreKeyValue` objects (see `infrahub.webhook.models.WebhookHeader.resolve`), but I'm not sure doing so wouldn't impact the intended general purpose of this kind of object.
Maybe the solution would be to add another layer with an intermediate object to restrict the webhook header object type, since the run-time limitation is already hard-coded in the code.

Any thoughts regarding this @ogenstad ?

ogenstad: I think I would rather restrict it now and later if we change our mind or see a need for user defined types from `CoreKeyValue` it would be easy for us to remove this restriction. As it is now I'm not sure if there  is a scenario where customers can do anything useful when creating custom nodes from this generic since I think it would require custom code in most cases that end users wouldn't have access to.

If we go the other way and implement the restrictions later there's a growing risk of anyone having one of these within their schemas.

backend/infrahub/core/schema/definitions/core/account.py
```diff
@@ -134,6 +135,7 @@
     branch=BranchSupportType.AGNOSTIC,
     documentation="/topics/auth",
     uniqueness_constraints=[["name__value"]],
+    restricted_namespaces=["Core"],
```
ogenstad, line 138: This one I'm a bit unsure of. Given SSO and LDAP I'm not sure that it was the correct choice to have a password field in the generic itself, potentially we should even set the password attribute to not be required.

However as we have name and password and everything else on the generic there really isn't anything special about `CoreAccount` and someone would be able to create an `ExpandedAccount` and add things like `address`, `hairstyle`, `employment_number` and I would assume that the login and overall access would still work. Possibly since the frontend uses a custom forms for users/permissions there could be some things that would need to be managed there. I'm not sure that it's really useful to do this. It's not like Infrahub is an HR system.. Also this would only really work for local users as anything created by SSO or LDAP would be a `CoreAccount`.

ajtmccarty: this sounds like it would be an unexpected behavior for the future. if we're making changes to CoreGenericAccount at some point, I'd probably assume that all NodeSchemas using `CoreGenericAccount` are also in the `Core` namespace and fully under our control. but I suppose it could be the case that someone has already made their own `ExpandedAccount` and it is mostly working, so maybe we have to leave it unrestricted

ogenstad: That's also a good point. If we ever add an attribute to the generic that someone else already have in their schema.

ajtmccarty: so that's another point in favor of restricting it now, right?

ogenstad: Yes, this is pretty much why I wanted to add namespaces in the first place. Anything that gets extended from the core modules can be problematic sometime in the future. So I think the best option would be to not allow this for most of the things coming from our own namespaces, the alternative is for things where we only use the generic as a marker where it isn't expected to have a growing list of attributes. Even the BuiltinIPAddress might at some point be problematic if we wanted to add some helper methods etc, but since those are so specific we could solve them in different ways in the code. Others like "artifact target", I'd expect that we don't change or add relationships and attributes on those.

BaptisteGi: Never had the use case where I had to extend the `CoreAccount` generic. I guess there are valid ones as Patrick mentioned. IMO, we keep it restricted for the moment, and if we have a strong use case that would require extending it, we unlock it.

ajtmccarty: then I think we all agree that `CoreGenericAccount` is correctly restricted in this PR

ogenstad: Yes, let's move forward with this. Before merging can you create an issue in that release activities epic I created in Jira and mention that we should list these as breaking changes that would need to be corrected prior to the upgrade?

ajtmccarty: new subtask on the issue for breaking change release notes
https://opsmill.atlassian.net/browse/IFC-2615

ogenstad (approved):

PR #9190, opened 2026-05-08
[INFP-530]  enable order_by schema attribute to support node metadata fields with explicit sort direction spec

wvandeun:


dev/specs/infp-530-order-by-metadata-direction/spec.md
```diff
@@ -0,0 +1,109 @@
...
+**Status**: Draft
+**Input**: User description: "Enable order_by schema attribute to support node metadata fields with explicit sort direction"
+**Source ticket**: [INFP-530](https://opsmill.atlassian.net/browse/INFP-530)
+**Customer**: University of Melbourne
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - Default newest-first ordering on a relationship list (Priority: P1)
+
+A schema designer maintains a `Notes` schema that is attached to many parent objects through a many-cardinality relationship. They want every list view of `Notes` — whether on the parent's detail page, in the API, or in any UI tab — to default to "newest first" without each query having to specify the order. Today they must specify the metadata-based ordering at query time on every read; the UI does not, so the notes appear in an unhelpful order.
```
ogenstad: The use of `Notes` here is a bit confusing at first I thought it was a typo for `Nodes`. If we are talking about a node kind in the schema we should use a properly supported name that includes a namespace.

On another note I think I might have missed the part with the cardinality=many in the JPD card. Given that this is defined on the node level and doesn't control the ordering of a related node I'm wondering if that comment is related to this feature. Having said that I'd assume that it would just work regardless of if you're querying a node directly or querying through a relationship. It might be good to keep as a sanity check though.

dev/specs/infp-530-order-by-metadata-direction/spec.md
```diff
@@ -0,0 +1,109 @@
...
+
+**Why this priority**: This is the customer's real, blocking pain. Users cannot tell at a glance which note is most recent, which makes the feature unusable for its intended purpose. The fix lives entirely at the schema-definition level, so it's a single edit that propagates to every consumer.
+
+**Independent Test**: Define a `Notes` schema whose `order_by` declares newest-first ordering by creation timestamp. Attach two notes to a parent object at different times. Open the parent's detail page; verify the most-recent note appears first. Verify the same ordering holds when fetching the parent's notes through the API without any per-query ordering argument.
```
ogenstad: A clarification here is that for the we're talking about node metadata at the node level. This above could read as we're referring to the relationship metadata which would be a different feature. I.e. the ordering would be based on the node creation or update timestamp, not on the relationship creation or update timestamp.

dev/specs/infp-530-order-by-metadata-direction/spec.md
```diff
@@ -0,0 +1,109 @@
...
+**Feature Branch**: `wvd-20260516-infp-530-spec`
+**Created**: 2026-05-08
+**Status**: Draft
+**Input**: User description: "Enable order_by schema attribute to support node metadata fields with explicit sort direction"
+**Source ticket**: [INFP-530](https://opsmill.atlassian.net/browse/INFP-530)
+**Customer**: University of Melbourne
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - Default newest-first ordering on a relationship list (Priority: P1)
```
ogenstad, line 24: There's no mention of the `node_metadata` until user story 3, I think we should include it here as well as it's not possible to complete user story 1 without it. We should mention that both `__created_at` and `__updated_at` should be supported.

dev/specs/infp-530-order-by-metadata-direction/spec.md
```diff
@@ -0,0 +1,109 @@
...
+A schema author makes a typo in their `order_by` declaration — for example, references a metadata field that doesn't exist, or uses a malformed direction suffix. They want the schema load to fail fast with a clear, actionable error message that names the offending node and field.
+
+**Why this priority**: Without strict validation, malformed entries silently degrade to default ordering at runtime, leaving authors confused about why their declared ordering "doesn't work". Catching the error at schema load is cheap and prevents production confusion.
+
+**Independent Test**: Attempt to load a schema with each of the rejection cases below; verify each one raises a descriptive error at schema-load time citing the offending node and entry.
+
+**Acceptance Scenarios**:
+
+1. **Given** an `order_by` entry references an unsupported metadata field, **When** the schema is loaded, **Then** loading fails with an error naming the entry and listing supported fields.
+2. **Given** an `order_by` entry uses a malformed direction token (e.g., `__descending`), **When** the schema is loaded, **Then** loading fails with an error stating only `asc` and `desc` are valid.
```
ogenstad, line 78: Given that we already have the loading order for attributes, the `asc` would be the default and currently doesn't exist. We should also mention that adding `__desc` would be the correct approach since I don't think the spec mentions what this should look like. It might be that we want to add `__asc` for clarity.

Also when looking at this I think we may have a schema bug related to these names as we seem to allow double underscores in attribute (and probably relationship) names which should be problematic.. I will verify if we have a bug around this.

dev/specs/infp-530-order-by-metadata-direction/spec.md
wvandeun: It does control the ordering of the related nodes. IE for the related nodes the `order_by` property on the schema node of the related node is used. There seems to be a gap when there is a tie though, which is handled for a node, but not for related nodes. Clarifying that in the spec.

dev/specs/infp-530-order-by-metadata-direction/spec.md
```diff
@@ -0,0 +1,135 @@
...
+**Status**: Draft
+**Input**: User description: "Enable order_by schema attribute to support node metadata fields with explicit sort direction"
+**Source ticket**: [INFP-530](https://opsmill.atlassian.net/browse/INFP-530)
+**Customer**: University of Melbourne
+
+## Clarifications
+
+### Session 2026-05-11
+
+- Q: When two items share the same value on the schema-level `order_by` field (e.g., two documentation notes created in the same millisecond), what is their relative order? → A: Always append `node.uuid` ascending as an implicit final tiebreaker on all three list paths (top-level, relationship-peer, hierarchy) whenever `order_by` is in effect; this also standardizes today's inconsistency where only the top-level path appends UUID.
```
ogenstad, line 14: This is already done by the backend by default (without it the pagination completely falls apart when querying by name and there are hundreds of nodes called Ethernet1. But could be good to have as a reference for the node_metadata specifically as those queries might look slightly different.

ogenstad (approved):

PR #9194, opened 2026-05-08
IFC-2458: Update profile and template on changing unique attribute

### LINKED ISSUE #8894: bug: updating attribute unique does not add/remove from Template schemas
### Component

API Server / GraphQL

### Infrahub version

1.8.4

### Current Behavior

only attributes with unique=False and read-only=False should be included in the Template schema generated for a given Node schema. but if I update an attribute from unique=True to unique=False, then the attribute is not added to existing Template instances for the Node

### Expected Behavior

update the uniqueness of an attribute should correctly update any existing instances of Template schemas for the node to add/remove the attribute as appropriate

### Steps to Reproduce

1. load a schema that includes a unique attribute and has `generate_template=True`
2. create an instance of the Template schema for your schema from step 1
3. load an updated version of the schema from step 1 that sets the unique attribute to be non-unique
4. the attribute should be added to the Template schema and be updateable on the Template instance from step 2, but it is not

### Additional Information

the fix requires adding a new `attribute.unique.update` migration that does something similar to the `AttributeSupportsGeneratedSchemaMigration` migration

solababs:
# Why

This PR ensures that removing `unique: True` from a schema attribute correctly triggers the `AttributeSupportsGeneratedSchemaMigration`, which adds the attribute to all existing profile and template node instances.

Closes https://github.com/opsmill/infrahub/issues/8894

## How to test

```bash
uv run pytest backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py -v
```

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/8894.fixed.md`)


generated/infrahub/schema/develop.json
ajtmccarty, line 1: what are these files?

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -425,3 +438,219 @@ async def test_migration_disable_support_sub_template(
...
+    # When a user removes unique=True from their schema definition, the uniqueness_constraints
+    # are rebuilt from scratch (not inherited). Replicate that by removing the stale entry.
+    # Constraint paths use the "attr__value" format (e.g. "nbr_seats__value"), so we must
+    # check startswith rather than an exact string match.
+    if new_car_schema.uniqueness_constraints:
+        new_car_schema.uniqueness_constraints = [
+            paths
+            for paths in new_car_schema.uniqueness_constraints
+            if not any(p == "nbr_seats" or p.startswith("nbr_seats__") for p in paths)
+        ] or None
```
ajtmccarty: can you just set `new_car_schema.uniqueness_constraints` to whatever it should be instead of doing it programatically? it seems like this should be a known value

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -425,3 +438,219 @@ async def test_migration_disable_support_sub_template(
...
+    )
+
+
+@pytest.fixture
+async def device_part_schema_unique_serial(
+    db: InfrahubDatabase, default_branch: Branch, node_group_schema: None, data_schema: None
+) -> SchemaBranch:
+    """TestDevice (generate_template=True) has COMPONENT TestPart (generate_template=False).
+    serial_number starts unique=True so it is NOT on template instances."""
+    part = _PART.model_copy(deep=True)
```
ajtmccarty: I believe our `duplicate()` method would work here

backend/tests/component/core/migrations/schema/test_attribute_supports_generated_schema.py
```diff
@@ -425,3 +438,219 @@ async def test_migration_disable_support_sub_template(
...
+    part.get_attribute("serial_number").unique = True
+    registry.schema.register_schema(
+        schema=SchemaRoot(generics=[core_object_template, core_object_component_template]),
+        branch=default_branch.name,
+    )
+    return registry.schema.register_schema(schema=SchemaRoot(nodes=[_DEVICE, part]), branch=default_branch.name)
+
+
+async def test_migration_unique_enable_support_sub_template(
```
ajtmccarty, line 526: I think this case deserves to be in an integration test, either a new one or an existing one. it should ensure that the template schema and instance correctly have the expected attribute added/removed when the attribute's uniqueness is updated

generated/infrahub/schema/develop.json
solababs: they were automatically generated, and the `generated/` is not in `.gitignore`

ajtmccarty: what are they? should they be included?

ajtmccarty (commented):
this looks good. I think it is ready to merge once those `generated/` files are addressed

changelog/8894.fixed.md
```diff
@@ -0,0 +1 @@
+Changing `unique` to `False` on a schema attribute now correctly triggers a migration that adds the attribute to existing profile and template nodes.
```
ajtmccarty: we try to say `objects` instead of `nodes`

ajtmccarty (approved):

PR #9199, opened 2026-05-09
Break apart exceptions for ruff documentation rules + auto fixes

ogenstad:
# Why

Previously all documentation rules in ruff were ignored, split these apart into smaller chunks so that we can consider each rule in isolation. The primary goal will be to add DOC501 (`docstring-missing-exception`) where it's missing today and then to enable the rules to require docstrings in general.

## What changed

* Mainly shift around rules in pyproject.toml
* One of the violating rules had to do with whitespaces in existing docstrings, these were auto fixable with ruff as a formatting issue so I reformatted the files violating that rule.
* Minor docstring change in backend/infrahub/core/branch/models.py due to line length after reformatting.

### DIFF
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -244,7 +244,6 @@ class Branch(StandardNode):
 
     def get_branches_and_times_to_query(self, at: Optional[Timestamp] = None) -> dict[frozenset, str]:
         """Return all the names of the branches that are constituing this branch with the associated times excluding the global branch"""
-
         at = Timestamp(at)
 
         if self.is_default:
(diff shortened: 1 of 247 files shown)


backend/infrahub/core/branch/models.py
```diff
@@ -335,10 +332,7 @@ async def delete(self, db: InfrahubDatabase) -> None:
     def get_query_filter_relationships(
         self, rel_labels: list, at: Optional[Timestamp] = None, include_outside_parentheses: bool = False
     ) -> tuple[list, dict]:
-        """
-        Generate a CYPHER Query filter based on a list of relationships to query a part of the graph at a specific time and on a specific branch.
-        """
-
+        """Generate a CYPHER Query filter from a list of relationships to query a part of the graph at a specific time and on a specific branch."""
```
ogenstad, line 335: Slight rewording "based on" -> "from" due to line length issues.

dgarros (approved):

PR #9203, opened 2026-05-10
fix: validate negative offset and limit in graphql resolver

### LINKED ISSUE #7474: bug: Late errors for negative offset or limit query parameters

DharmaBytesX:
Closes #7474

## What

Adds input validation in `default_paginated_list_resolver` to reject negative `limit` and `offset` values with a user-friendly `GraphQLError` before reaching the database layer.

## Before

Negative values were passed through to Neo4j, resulting in a `CypherSyntaxError` with an unhelpful stack trace in the server logs.

## After

The resolver immediately raises a `GraphQLError` with a clear message (`limit must be a non-negative integer` / `offset must be a non-negative integer`).

## Tests

Added unit tests in `backend/tests/unit/graphql/resolvers/test_resolver.py` covering both `limit` and `offset` validation.

### DIFF
--- a/backend/infrahub/graphql/resolvers/resolver.py
+++ b/backend/infrahub/graphql/resolvers/resolver.py
@@ -183,6 +183,13 @@ def _transform_metadata_day_filters(filters: dict[str, Any]) -> dict[str, Any]:
     return result
 
 
+def validate_offset_and_limit(offset: int | None, limit: int | None) -> None:
+    if limit is not None and limit < 0:
+        raise GraphQLError("limit must be a non-negative integer")
+    if offset is not None and offset < 0:
+        raise GraphQLError("offset must be a non-negative integer")
+
+
 @trace.get_tracer(__name__).start_as_current_span("default_paginated_list_resolver")
 @retry_db_transaction(name="default_paginated_list_resolver")
 async def default_paginated_list_resolver(
@@ -194,6 +201,7 @@ async def default_paginated_list_resolver(
     partial_match: bool = False,
     **kwargs: dict[str, Any],
 ) -> dict[str, Any]:
+    validate_offset_and_limit(offset, limit)
     schema: MainSchemaTypes = (
         info.return_type.of_type.graphene_type._meta.schema
         if isinstance(info.return_type, GraphQLNonNull)
@@ -235,11 +243,6 @@ async def default_paginated_list_resolver(
             for edge in permissions["edges"]:
                 if edge["node"]["kind"] == schema.kind:
                     permission_set = edge["node"]
-
-        if limit is not None and limit < 0:
-            raise GraphQLError("limit must be a non-negative integer")
-        if offset is not None and offset < 0:
-            raise GraphQLError("offset must be a non-negative integer")
 
         objs = []
         if edges or "hfid" in filters:
--- a/backend/infrahub/graphql/resolvers/resolver.py
+++ b/backend/infrahub/graphql/resolvers/resolver.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any
 
+from graphql import GraphQLError
 from graphql.type.definition import GraphQLNonNull
 from opentelemetry import trace
 
@@ -234,6 +235,11 @@ async def default_paginated_list_resolver(
             for edge in permissions["edges"]:
                 if edge["node"]["kind"] == schema.kind:
                     permission_set = edge["node"]
+
+        if limit is not None and limit < 0:
+            raise GraphQLError("limit must be a non-negative integer")
+        if offset is not None and offset < 0:
+            raise GraphQLError("offset must be a non-negative integer")
 
         objs = []
         if edges or "hfid" in filters:
--- a/backend/tests/unit/graphql/resolvers/test_resolver.py
+++ b/backend/tests/unit/graphql/resolvers/test_resolver.py
@@ -1,0 +1,20 @@
+from unittest.mock import MagicMock
+
+import pytest
+from graphql import GraphQLError
+
+from infrahub.graphql.resolvers.resolver import default_paginated_list_resolver
+
+
+@pytest.mark.anyio
+async def test_negative_limit_raises() -> None:
+    info = MagicMock()
+    with pytest.raises(GraphQLError, match="limit"):
+        await default_paginated_list_resolver(root=None, info=info, limit=-1)
+
+
+@pytest.mark.anyio
+async def test_negative_offset_raises() -> None:
+    info = MagicMock()
+    with pytest.raises(GraphQLError, match="offset"):
+        await default_paginated_list_resolver(root=None, info=info, offset=-1)
--- a/backend/tests/unit/graphql/resolvers/test_resolver.py
+++ b/backend/tests/unit/graphql/resolvers/test_resolver.py
@@ -1,20 +1,14 @@
-from unittest.mock import MagicMock
-
 import pytest
 from graphql import GraphQLError
 
-from infrahub.graphql.resolvers.resolver import default_paginated_list_resolver
+from infrahub.graphql.resolvers.resolver import validate_offset_and_limit
 
 
-@pytest.mark.anyio
-async def test_negative_limit_raises() -> None:
-    info = MagicMock()
+def test_negative_limit_raises() -> None:
     with pytest.raises(GraphQLError, match="limit"):
-        await default_paginated_list_resolver(root=None, info=info, limit=-1)
+        validate_offset_and_limit(offset=None, limit=-1)
 
 
-@pytest.mark.anyio
-async def test_negative_offset_raises() -> None:
-    info = MagicMock()
+def test_negative_offset_raises() -> None:
     with pytest.raises(GraphQLError, match="offset"):
-        await default_paginated_list_resolver(root=None, info=info, offset=-1)
+        validate_offset_and_limit(offset=-1, limit=None)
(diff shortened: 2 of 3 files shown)


dgarros (approved):
Looks good, thank you
Ideally please can you add an entry in the changelog in https://github.com/opsmill/infrahub/tree/stable/changelog to include this fix in the release note
the format is describe here https://github.com/opsmill/infrahub/blob/stable/dev/guidelines/changelog.md

DharmaBytesX:
Hey, I added the changelog entry in `changelog/7474.fixed.md`.

backend/infrahub/graphql/resolvers/resolver.py
```diff
@@ -235,6 +236,11 @@ async def default_paginated_list_resolver(
                 if edge["node"]["kind"] == schema.kind:
                     permission_set = edge["node"]
 
+        if limit is not None and limit < 0:
+            raise GraphQLError("limit must be a non-negative integer")
+        if offset is not None and offset < 0:
+            raise GraphQLError("offset must be a non-negative integer")
```
ogenstad: If we look at the rest of the function this check comes a bit to late. For instance on line 208 we start a new database session and on line 226 there's a query to the database regarding permissions. So a check like this should be at the start of this function and fail early.

backend/tests/unit/graphql/resolvers/test_resolver.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.mark.anyio
+async def test_negative_limit_raises() -> None:
+    info = MagicMock()
+    with pytest.raises(GraphQLError, match="limit"):
+        await default_paginated_list_resolver(root=None, info=info, limit=-1)
+
+
+@pytest.mark.anyio
```
ogenstad: These pytest markers should not be needed.

backend/infrahub/graphql/resolvers/resolver.py
gmazoyer, line 1: I suppose other resolvers would be affected by this too, the first coming to my mind is `backend/infrahub/graphql/resolvers/ipam.py`. Considering that, I would extract that tiny logic into a function that resolvers could call. That would also make the logic itself easier to test.

backend/tests/unit/graphql/resolvers/test_resolver.py
```diff
@@ -0,0 +1,20 @@
+from unittest.mock import MagicMock
+
+import pytest
+from graphql import GraphQLError
+
+from infrahub.graphql.resolvers.resolver import default_paginated_list_resolver
+
+
+@pytest.mark.anyio
```
gmazoyer: No need for those `pytest.mark.anyio`.

backend/tests/unit/graphql/resolvers/test_resolver.py
```diff
@@ -0,0 +1,20 @@
...
+async def test_negative_limit_raises() -> None:
+    info = MagicMock()
```
gmazoyer: We tend to avoid mocks but that seems about right here in a unit test.

ogenstad: While we have mocking in this repo we're trying to phase it out, the reason being that it becomes quite muddy with regards to what is actually being testing when some interfaces doesn't behave as you would expect and with future changes bugs might be introduced when we thought that we had tests to cover a given scenario. If we move the check up as described above we can send in anything to the info parameter. If it's just about making the type checker happy a `cast()` of None into the correct type could do the trick here. Ideally we should not use `cast()` within the code either, however in this case it could be fine short of creating a dummy object.

After the change the main problem with having a MagicMock for something like this is that AI agents looking at other tests to figure out the structure can think that it's ok to continue using mocking.

changelog/7474.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed GraphQL resolver returning an unhelpful database error when negative values were provided for limit or offset query arguments.
```
ogenstad: Unfortunately this doesn't fix all of the resolvers, could you please update this to say that it fixes the standard resolver for schema based objects?

ogenstad (changes requested):

ogenstad:
Ah see that both me and @gmazoyer was reviewing at the same type. 😅

DharmaBytesX:
Thx for the input. I've refacto the logic into a tiny validation function, moved the check earlier, cleaned up the tests, and updated the changelog.

DharmaBytesX:
The CI issue is not code related

ogenstad (approved):
Thank you for your contribution and the changes!

Approving, but also as an internal note as the release notes for Infrahub 1.9.4 has already been generated we should wait for that version to be released before merging this PR so that it gets included in Infrahub 1.9.5 instead.

PR #9204, opened 2026-05-10
Add missing raises section to existing docstrings

ogenstad:
# Why

* Gives the callers a better view of what might happen when calling this code
* Can be used by AI agents to better understand when an error could be raised.

## What changed

* Add missing raises section to current docstrings
* Enable ruff rule DOC501 in pyproject.toml

### DIFF
--- a/backend/infrahub/core/schema/basenode_schema.py
+++ b/backend/infrahub/core/schema/basenode_schema.py
@@ -74,7 +74,12 @@ class BaseNodeSchema(GeneratedBaseNodeSchema):
 
     @staticmethod
     def _enhance_attribute_validation_error(exc: ValidationError, attr_name: str, idx: int) -> NoReturn:
-        """Enhance validation error with attribute name and correct index in location path."""
+        """Enhance validation error with attribute name and correct index in location path.
+
+        Raises:
+            from_exception_data: ValidationError rebuilt with enriched location and input metadata derived from the original error.
+
+        """
         errors = []
         for error in exc.errors():
             error_copy = error.copy()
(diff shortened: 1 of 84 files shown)


backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -74,7 +74,12 @@ class BaseNodeSchema(GeneratedBaseNodeSchema):
 
     @staticmethod
     def _enhance_attribute_validation_error(exc: ValidationError, attr_name: str, idx: int) -> NoReturn:
-        """Enhance validation error with attribute name and correct index in location path."""
+        """Enhance validation error with attribute name and correct index in location path.
+
+        Raises:
+            from_exception_data: ValidationError rebuilt with enriched location and input metadata derived from the original error.
```
ajtmccarty, line 80: I don't think this one is quite right

ajtmccarty (approved):

backend/infrahub/core/schema/basenode_schema.py
ogenstad: I agree that it looks strange but it's actually what `ruff` expects when the method looks like this:

```python

    @staticmethod
    def _enhance_attribute_validation_error(exc: ValidationError, attr_name: str, idx: int) -> NoReturn:
        """Enhance validation error with attribute name and correct index in location path.

        Raises:
            from_exception_data: ValidationError rebuilt with enriched location and input metadata derived from the original error.

        """
        errors = []
        for error in exc.errors():
            error_copy = error.copy()
            # Insert the correct list index at the start of the location path
            error_copy["loc"] = (idx,) + error["loc"]
            nested_path = ".".join(str(item) for item in error["loc"])
            error_copy["input"] = f"[{attr_name}.{nested_path}]: {error.get('input', 'N/A')}"
            errors.append(error_copy)
        raise ValidationError.from_exception_data(title=exc.title, line_errors=errors) from exc
```

If we have ValidationError up front ruff will fail.

I'll merge this now then we can consider if we should refactor this method a bit instead.

PR #9214, opened 2026-05-11
IFC-1197: Fix CoreNumberPool not allowing upsert

### LINKED ISSUE #5636: bug: `CoreNumberPool` doesn't allow upsert
### Component

API Server / GraphQL

### Infrahub version

1.1.6

### Current Behavior

I have a script loading some data, it's designed in a way that I can run it multiple times. However when I run it a second time I hit an exception related with `CoreNumberPool`

```console
['CoreNumberPoolUpsert'] The fields 'node' or 'node_attribute' can't be changed.
```

### Expected Behavior

I expect the script to run especially since I'm not changing the value of that particular field.

### Steps to Reproduce

- Spin demo
- Load base schema (https://github.com/opsmill/schema-library/tree/main/base) and vlan schema (https://github.com/opsmill/schema-library/tree/main/extensions/vlan)
- Run the following piece of code (infrahubctl run xxx)

```python
async def run(
    client: InfrahubClient, log: logging.Logger, branch: str, **kwargs
) -> None:
    log.info("Generate all prefixes related data...")
    await create_prefixes(client=client, branch=branch, log=log)


async def create_prefixes(
    client: InfrahubClient, log: logging.Logger, branch: str
) -> None:
    # Create l2 domain
    l2_domain = await client.create(
        kind="IpamL2Domain",
        name="default",
    )
    await l2_domain.save(allow_upsert=True)

    # Create public prefix
    public_prefix = await client.create(
        kind="IpamPrefix",
        status="active",
        prefix="203.0.113.0/24",
        member_type="prefix",
        role="public",
    )
    await public_prefix.save(allow_upsert=True)

    # Create management prefix
    management_prefix = await client.create(
        kind="IpamPrefix",
        status="active",
        prefix="172.16.1.0/24",
        member_type="address",
        role="management",
    )
    await management_prefix.save(allow_upsert=True)

    # Create management ip pool
    management_pool = await client.create(
        kind="CoreIPAddressPool",
        name="Management IP pool",
        default_prefix_type="IpamIPPrefix",
        default_prefix_length=24,
        default_address_type="IpamIPAddress",
        default_member_type="address",
        ip_namespace="default",
        resources=[management_prefix],
    )
    await management_pool.save(allow_upsert=True)

    customer_prefix_pool = await client.create(
        kind="CoreIPPrefixPool",
        name="Customer prefixes pool",
        default_prefix_type="IpamPrefix",
        default_prefix_length=29,
        default_member_type="address",
        ip_namespace="default",
        resources=[public_prefix],
    )
    await customer_prefix_pool.save(allow_upsert=True)

    customer_vlan_pool = await client.create(
        kind="CoreNumberPool",
        name="Customer vlan pool",
        node="IpamVLAN",
        node_attribute="vlan_id",
        start_range=1000,
        end_range=2000,
    )
    await customer_vlan_pool.save(allow_upsert=True)
```

- First time it should work, second time it should fail with exception message

### Additional Information

_No response_

solababs:
# Why

This PR changes the guard from a **presence check** to a **value-diff check**: the mutation now loads the existing node and only rejects the request when the incoming value genuinely differs from the stored one.

Closes https://github.com/opsmill/infrahub/issues/5636

## How to test

```bash
# Run the five new tests
uv run pytest backend/tests/component/graphql/resource_manager/test_resource_manager.py \
  -v -k "upsert or update_with_unchanged or node_change or node_attribute"

uv run pytest backend/tests/component/graphql/resource_manager/ -v
```

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/5636.fixed.md`)


backend/infrahub/graphql/mutations/resource_manager.py
```diff
@@ -231,11 +231,21 @@ async def mutate_update(
...
     ) -> tuple[Node, Self]:
-        if (data.get("node") and data.get("node").value) or (
-            data.get("node_attribute") and data.get("node_attribute").value
-        ):
-            raise ValidationError(input_value="The fields 'node' or 'node_attribute' can't be changed.")
         graphql_context: GraphqlContext = info.context
+        new_node_value = data.get("node") and data.get("node").value
+        new_node_attr_value = data.get("node_attribute") and data.get("node_attribute").value
+        if new_node_value or new_node_attr_value:
+            existing = await NodeManager.find_object(
```
ajtmccarty: should this method be able to accept `node` and assume that it is the latest version of this CoreNumberPool? I don't think that is possible right now, but if a graphql mutation class inherits from CoreNumberPool in the future then it would be

backend/infrahub/graphql/mutations/resource_manager.py
```diff
@@ -231,11 +231,21 @@ async def mutate_update(
...
+                hfid=data.get("hfid"),
+                branch=branch,
+            )
+            if new_node_value and new_node_value != existing.node.value:  # type: ignore[attr-defined]
+                raise ValidationError(input_value="The fields 'node' or 'node_attribute' can't be changed.")
+            if new_node_attr_value and new_node_attr_value != existing.node_attribute.value:  # type: ignore[attr-defined]
+                raise ValidationError(input_value="The fields 'node' or 'node_attribute' can't be changed.")
 
         async with graphql_context.db.start_transaction() as dbt:
             number_pool, result = await super().mutate_update(
```
ajtmccarty, line 252: it looks like `exisiting`, if it is retrieved, should be passed in to the `node` parameter here so that the parent class doesn't also need to  retrieve the same object from the database. `existing` could probably also be called `node` to go with the current pattern.

backend/infrahub/graphql/mutations/resource_manager.py
```diff
@@ -231,11 +231,21 @@ async def mutate_update(
...
+            existing = await NodeManager.find_object(
+                db=graphql_context.db,
+                kind=protocols.CoreNumberPool,
+                id=data.get("id"),
+                hfid=data.get("hfid"),
+                branch=branch,
+            )
+            if new_node_value and new_node_value != existing.node.value:  # type: ignore[attr-defined]
+                raise ValidationError(input_value="The fields 'node' or 'node_attribute' can't be changed.")
+            if new_node_attr_value and new_node_attr_value != existing.node_attribute.value:  # type: ignore[attr-defined]
```
ajtmccarty: could you use `get_attribute()` here to remove the typing exception?

backend/tests/component/graphql/resource_manager/test_resource_manager.py
```diff
@@ -1183,3 +1209,208 @@ async def test_update_schema_number_pool_range(
...
+            "name": "pool-upsert-2",
+            "node": "TestingTicket",
+            "node_attribute": "ticket_id",
+            "start_range": 10,
+            "end_range": 30,
+        },
+    )
+    assert not upsert_mutable.errors
+    assert upsert_mutable.data
+    assert upsert_mutable.data["CoreNumberPoolUpsert"]["object"]["id"] == pool_id
```
ajtmccarty: the test should assert that the `end_range` is correctly updated in the response and that the end range is correct on the object retrieved from the database

backend/tests/component/graphql/resource_manager/test_resource_manager.py
```diff
@@ -1183,3 +1209,208 @@ async def test_update_schema_number_pool_range(
...
+
+
+async def test_number_pool_update_with_unchanged_node_fields(
+    db: InfrahubDatabase, default_branch: Branch, register_core_models_schema: SchemaBranch
+) -> None:
```
ajtmccarty: this looks like it tests the same thing as the test above, but maybe I am missing something

backend/tests/component/graphql/resource_manager/test_resource_manager.py
```diff
@@ -1183,3 +1209,208 @@ async def test_update_schema_number_pool_range(
         "start_range or end_range can't be updated on schema defined pools, update the schema in the default branch instead"
         in str(update_forbidden.errors)
     )
+
+
+async def test_number_pool_upsert_identical_fields(
```
ajtmccarty: all these tests could be combined into class so that the schema only has to be loaded once and the data only has to be loaded when necessary

backend/infrahub/graphql/mutations/resource_manager.py
ajtmccarty: now it looks like `existing` and `node` are unnecessarily swapping places. is it possible to just use `node` all the way through? you might have to do `find_object(..., kind="CoreNumberPool")` for the typing to be right

backend/tests/component/graphql/resource_manager/test_resource_manager.py
```diff
@@ -1183,3 +1211,231 @@ async def test_update_schema_number_pool_range(
...
+            },
+        )
+        assert not upsert_mutable.errors
+        assert upsert_mutable.data
+        assert upsert_mutable.data["CoreNumberPoolUpsert"]["object"]["id"] == pool_id
+        assert upsert_mutable.data["CoreNumberPoolUpsert"]["object"]["end_range"]["value"] == 30
+
+        pool = await NodeManager.get_one(id=pool_id, db=db, branch=default_branch_scope_class)
+        assert pool is not None
+        assert pool.get_attribute("end_range").value == 30  # type: ignore[union-attr]
```
ajtmccarty: is the type ignore really required here? I would think `value` would have a type of `Any`

backend/tests/component/graphql/resource_manager/test_resource_manager.py
```diff
@@ -1183,3 +1211,231 @@ async def test_update_schema_number_pool_range(
...
+            variable_values={
+                "id": pool_id,
+                "node": "TestingTicket",
+                "node_attribute": "ticket_id",
+                "start_range": 10,
+                "end_range": 25,
+            },
+        )
+        assert not update_same_values.errors
+        assert update_same_values.data
```
ajtmccarty: should confirm the expected value is returned and that the update actually persists on a retrieved instance of the object

backend/infrahub/graphql/mutations/resource_manager.py
```diff
@@ -231,11 +231,22 @@ async def mutate_update(
...
+                node = await NodeManager.find_object(
+                    db=graphql_context.db,
+                    kind=InfrahubKind.NUMBERPOOL,
+                    id=data.get("id"),
+                    hfid=data.get("hfid"),
+                    branch=branch,
+                )
+            if new_node_value and new_node_value != node.get_attribute("node").value:  # type: ignore[union-attr]
+                raise ValidationError(input_value="The fields 'node' or 'node_attribute' can't be changed.")
+            if new_node_attr_value and new_node_attr_value != node.get_attribute("node_attribute").value:  # type: ignore[union-attr]
```
ajtmccarty: I think you can remove these `type: ignore`s

ajtmccarty (approved):

PR #9218, opened 2026-05-11
Honor INFRAHUB_CACHE_TLS_* in Prefect result storage block

### LINKED ISSUE #9217: bug: Background flows hang against TLS-only Redis (INFRAHUB_CACHE_TLS_* not honored by Prefect result-storage block)
### Component

API Server / GraphQL

### Infrahub version

1.9.3 (present in earlier versions too — defect was introduced when the Prefect Redis result-storage block was added).

### Current Behavior

When Infrahub is configured to use an external TLS-only Redis (`INFRAHUB_CACHE_TLS_ENABLED=true`, plus optionally `INFRAHUB_CACHE_TLS_INSECURE` and `INFRAHUB_CACHE_TLS_CA_FILE`), background flows that persist results through Prefect — for example `git_repositories_sync`, `clean-up-deadlocks`, and `schema_validate_migrations` — start, run their internal task runs, and then hang in `RUNNING` indefinitely. The flow never reaches a terminal state.

Startup itself looks healthy, because `RedisStorageContainer.asave(...)` only writes the block document to Prefect's metadata DB; it does not connect to Redis. The failure only surfaces later, when a flow first tries to persist a result through the block.

Cache (`InfrahubCache`) and distributed-lock (`InfrahubLock`) traffic continue to work, because those code paths construct `redis.Redis(..., ssl=..., ssl_cert_reqs=..., ssl_check_hostname=..., ssl_ca_certs=...)` directly and honor `INFRAHUB_CACHE_TLS_*`. The Prefect result-storage block is the only Redis client in the codebase that does not.

### Expected Behavior

`INFRAHUB_CACHE_TLS_ENABLED`, `INFRAHUB_CACHE_TLS_INSECURE`, and `INFRAHUB_CACHE_TLS_CA_FILE` should apply uniformly to every Redis client Infrahub creates — including the Prefect result-storage block — so that flows complete successfully against a TLS-only Redis.

### Steps to Reproduce

1. Provision a TLS-only Redis reachable to Infrahub (a managed TLS-only Redis, or a local `stunnel`/testcontainer wrapper in front of a plaintext Redis).
2. Set on the Infrahub backend processes:
   - `INFRAHUB_CACHE_ADDRESS=<tls-redis-host>`
   - `INFRAHUB_CACHE_TLS_ENABLED=true`
   - `PREFECT_RESULTS_PERSIST_BY_DEFAULT=true`
3. Start Infrahub and trigger any flow that persists results (e.g. wait for `clean-up-deadlocks` on its schedule, or run `git_repositories_sync`).
4. Observe in Prefect: the flow's task runs complete, but the parent flow run stays in `RUNNING` indefinitely.

Workaround that confirms the diagnosis: patch the saved block document in Prefect's DB to use `connection_string: rediss://...` instead of host/port/db/username/password. All affected flows complete immediately afterwards.

### Additional Information

**Root cause.** `backend/infrahub/workflows/initialization.py::setup_blocks()` builds the block via `RedisStorageContainer.from_host(...)`. Prefect's `from_host` has no TLS parameter, and the internal client path used in that branch (`_client()`) calls `aioredis.Redis(host=..., port=..., username=..., password=..., db=...)` with no `ssl=` argument. The only TLS-capable path through `RedisStorageContainer` is `connection_string`, which is then resolved via `redis.Redis.from_url(...)`.

**Sites that already honor the setting** (for parity reference):

- `backend/infrahub/lock.py` — `redis.Redis(..., ssl=config.SETTINGS.cache.tls_enabled, ssl_cert_reqs=..., ssl_check_hostname=..., ssl_ca_certs=...)`
- `backend/infrahub/services/adapters/cache/redis.py` — same pattern

**Proposed fix.** Always construct a `connection_string` URL in `setup_blocks()`:

- Scheme switches between `redis://` and `rediss://` based on `tls_enabled`.
- `tls_insecure=true` encodes as `?ssl_cert_reqs=none&ssl_check_hostname=False` query params.
- `tls_ca_file` encodes as `?ssl_ca_certs=<path>` query param.

`redis.Redis.from_url` parses these query params and applies them to the underlying `SSLConnection`, so all three TLS knobs become deterministic across every Redis client in the codebase — no upstream Prefect change needed and no separate "this flag doesn't apply here" warning.

PR to follow.

PhillSimonds:
## Summary

- Background flows hung in `RUNNING` against TLS-only Redis because `setup_blocks()` built the Prefect `RedisStorageContainer` via `from_host()`, whose internal client path constructs `redis.Redis(...)` with no `ssl=` argument. `INFRAHUB_CACHE_TLS_ENABLED` / `INFRAHUB_CACHE_TLS_INSECURE` / `INFRAHUB_CACHE_TLS_CA_FILE` were silently ignored for this one client, while `lock.py` and the cache adapter honored them.
- `setup_blocks()` now always builds a `connection_string` URL from `CacheSettings`. Scheme switches between `redis://` and `rediss://` based on `tls_enabled`; `tls_insecure` encodes as `ssl_cert_reqs=none&ssl_check_hostname=False`; `tls_ca_file` encodes as `ssl_ca_certs=<path>`. `redis.Redis.from_url` parses these query params and applies them to the underlying `SSLConnection`, so all three TLS knobs become deterministic across every Redis client in the codebase.
- Single code path — no `if/else` between two constructors, no Prefect upstream change, no follow-on "doesn't apply here" warning needed.

## Behavior change

Previously, setting `INFRAHUB_CACHE_USERNAME` without `INFRAHUB_CACHE_PASSWORD` produced a Prefect block-construction error. The new helper raises `ValueError` with an explicit message before reaching Prefect — same outcome (startup fails), clearer diagnostic.

## Test plan

- [x] `uv run pytest backend/tests/unit/workflows/` — 276 passed (18 new tests in `test_initialization.py`)
- [x] `uv run mypy backend/infrahub/workflows/initialization.py` — clean
- [x] `uv run ruff check backend/infrahub/workflows/initialization.py backend/tests/unit/workflows/test_initialization.py` — clean
- [ ] CI: backend unit + integration suites
- [ ] Customer-side verification against external TLS-only Redis

### New test coverage

`backend/tests/unit/workflows/test_initialization.py` covers:

1. **URL composition** (11 parametrized cases): plain defaults, custom port/db, user+password, password-only, special-character escaping, TLS defaults, TLS+credentials, `tls_insecure`, `tls_ca_file`, both TLS flags combined, and the negative case that `tls_insecure` / `tls_ca_file` are ignored when `tls_enabled=False`.
2. **Username-without-password validation**: raises `ValueError` with explicit message.
3. **Round-trip through `redis.ConnectionPool.from_url`** (6 parametrized cases): proves the URL we build resolves to the expected `connection_class` (`Connection` vs `SSLConnection`) and propagates host / port / db / username / password / `ssl_cert_reqs` / `ssl_check_hostname` / `ssl_ca_certs` correctly. This guards against silent regressions in redis-py's URL parser.

Closes #9217

---
## Summary by cubic
Fixes Prefect result storage against TLS-only Redis by honoring `INFRAHUB_CACHE_TLS_*` settings via a `redis://`/`rediss://` connection string. Unifies TLS behavior across all Redis clients and prevents background flows from hanging.

- Bug Fixes
  - Build the connection string from `CacheSettings`; use `rediss://` when `INFRAHUB_CACHE_TLS_ENABLED` is true.
  - Encode TLS options: `INFRAHUB_CACHE_TLS_INSECURE` → `ssl_cert_reqs=none&ssl_check_hostname=False`; `INFRAHUB_CACHE_TLS_CA_FILE` → `ssl_ca_certs=<path>`. Parsed by `redis.Redis.from_url` and applied to `SSLConnection`.
  - Replace `RedisStorageContainer.from_host` with `RedisStorageContainer(connection_string=...)` for a single, consistent path.
  - Clearer startup error: setting `INFRAHUB_CACHE_USERNAME` without `INFRAHUB_CACHE_PASSWORD` now raises a `ValueError` with an explicit message.

<sup>Written for commit 2ae4c393ae49830fa67acb9c86881b7fbdea7fb1. Summary will update on new commits.</sup>


backend/infrahub/workflows/initialization.py
```diff
@@ -17,6 +21,38 @@
...
+    """Build a redis:// or rediss:// URL from cache settings.
+
+    All TLS knobs propagate through redis.Redis.from_url: the scheme selects
+    SSLConnection, and ssl_cert_reqs / ssl_check_hostname / ssl_ca_certs query
+    params are passed through to the connection. This keeps the Prefect result
+    storage block in parity with lock.py and the cache adapter, which read the
+    same INFRAHUB_CACHE_TLS_* settings directly.
+    """
+    if cache.username and not cache.password:
+        raise ValueError("INFRAHUB_CACHE_USERNAME is set but INFRAHUB_CACHE_PASSWORD is not. Both are required.")
```
ogenstad, line 34: I would expect something like this to be a validator on https://github.com/opsmill/infrahub/blob/infrahub-v1.9.6/backend/infrahub/config.py#L398 instead.

ogenstad (approved):

PR #9230, opened 2026-05-12
Fix IPPrefix utilisation on new branch

### LINKED ISSUE #9220: bug: IPPrefix utilization drops to 0 on a newly created branch
### Component

API Server / GraphQL

### Infrahub version

1.9.3

### Current Behavior

When an `IpamIPPrefix` reports non-zero `utilization` on `main`, creating a new branch from `main` causes the same prefix to report `utilization: 0` on the branch — even though no data has been modified.

The effect shows up on every prefix that had non-zero utilization on main: a fully populated `/32` sub-prefix (with its host `IpamIPAddress` assigned) and its `/24` parent both flip to `0` on the branch. The exact percentages aren't the point — utilization values are not preserved on branch creation.

### Expected Behavior

A freshly created branch should report the same `utilization` as `main` until data on the branch is modified.

### Steps to Reproduce

1. Load the `models/base/ipam.yml` schema, which provides `IpamIPPrefix` and `IpamIPAddress` (inheriting from `BuiltinIPPrefix` / `BuiltinIPAddress`):

   ```bash
   infrahubctl schema load models/base/ipam.yml
   ```

2. Load the following objects on `main`:

   ```yaml
   ---
   apiVersion: infrahub.app/v1
   kind: Object
   spec:
     kind: IpamIPPrefix
     data:
       - prefix: 10.100.0.0/24
         description: "Parent /24 — container for /32 sub-prefixes"
         member_type: prefix

   ---
   apiVersion: infrahub.app/v1
   kind: Object
   spec:
     kind: IpamIPPrefix
     data:
       - prefix: 10.100.0.1/32
         member_type: address
       - prefix: 10.100.0.2/32
         member_type: address
       - prefix: 10.100.0.3/32
         member_type: address
       - prefix: 10.100.0.4/32
         member_type: address

   ---
   apiVersion: infrahub.app/v1
   kind: Object
   spec:
     kind: IpamIPAddress
     data:
       - address: 10.100.0.1/32
       - address: 10.100.0.2/32
       - address: 10.100.0.3/32
       - address: 10.100.0.4/32
   ```

   ```bash
   infrahubctl object load ipam_utilization_bug_repro.yml
   ```

3. Open the IPAM view at <http://localhost:8000/ipam> on `main` — the `/24` and each `/32` sub-prefix show non-zero utilization.
4. Create a new branch:

   ```bash
   infrahubctl branch create test-1
   ```

5. In the same IPAM view, switch the branch selector to `test-1`. Every prefix that reported non-zero utilization on `main` now reports `0` on `test-1`, while `main` is unchanged.

### Additional Information

- `IPPrefix.utilization`, `IPPrefix.ip_addresses`, and `IPAddress.ip_prefix` are all declared `read_only=True` with `allow_override=NONE` in `backend/infrahub/core/schema/definitions/core/ipam.py` (lines 116-122, 166-175, 219-228). The hierarchical prefix→address association is computed, not stored.
- The sub-prefix and address objects themselves remain visible on the branch — only the `utilization` value is affected.

gmazoyer:
## Why & What

`IpamIPPrefix.utilization` returned wrong values on every non-default branch:

1. A freshly created branch reported `0` for every prefix that had non-zero utilisation on its origin (the issue this PR was opened for).
2. Deleting an address on a branch zeroed out the same prefix on the origin branch as well.
3. Changes made on the origin branch after a feature branch was created leaked into the branch's view, even when the branch was isolated.

The cause was that prefix utilisation had its own bespoke way of reading the graph that did not match how the rest of the platform handles branches. Both per-branch visibility and cross-branch attribution were squeezed into a single query, which mixed the two semantics and routinely produced wrong results when modifications existed on both sides.

Utilisation now reads from one branch at a time using the same branch scoping that powers every other read in Infrahub: an isolated branch sees its origin frozen at its creation point and adds its own changes on top. The pool view, which still needs both the current branch view and the default branch view, derives them by asking the same question twice and subtracting.

Closes #9220

## Checklist

- [x] Tests added/updated

### DIFF
--- a/backend/infrahub/core/ipam/utilization.py
+++ b/backend/infrahub/core/ipam/utilization.py
@@ -19,9 +19,10 @@ class PrefixChildDetails:
 
 
 class PrefixUtilizationGetter:
-    """Single-branch utilization: counts allocations visible from ``branch``.
+    """Counts allocations visible from a single branch.
 
-    To compare two branches, e.g. main against a feature branch for a pool view, instantiate one getter per branch and subtract.
+    Use one instance per branch and combine the results externally if a
+    cross-branch comparison is required.
     """
 
     def __init__(
--- a/backend/infrahub/core/ipam/utilization.py
+++ b/backend/infrahub/core/ipam/utilization.py
@@ -19,27 +19,24 @@ class PrefixChildDetails:
 
 
 class PrefixUtilizationGetter:
+    """Single-branch utilization: counts allocations visible from ``branch``.
+
+    To compare two branches, e.g. main against a feature branch for a pool view, instantiate one getter per branch and subtract.
+    """
+
     def __init__(
-        self,
-        db: InfrahubDatabase,
-        ip_prefixes: list[Node],
-        at: Timestamp | str | None = None,
-        branch: Branch | None = None,
+        self, db: InfrahubDatabase, ip_prefixes: list[Node], branch: Branch, at: Timestamp | str | None = None
     ) -> None:
         self.db = db
         self.ip_prefixes = ip_prefixes
-        self.at = at
         self.branch = branch
+        self.at = at
         self._has_data = False
-        self._results_by_prefix_id: dict[str, dict[str, list[PrefixChildDetails]]] = {}
+        self._results_by_prefix_id: dict[str, list[PrefixChildDetails]] = {}
 
     async def _fetch_data(self) -> None:
-        if self._has_data is False:
-            await self._run_and_parse_query()
-        self._has_data = True
-
-    async def _run_and_parse_query(self) -> None:
-        self._results_by_prefix_id = {}
+        if self._has_data:
+            return
         query = await IPPrefixUtilization.init(
             db=self.db,
             at=self.at,
@@ -49,93 +46,60 @@ class PrefixUtilizationGetter:
         )
         await query.execute(db=self.db)
 
+        results: dict[str, list[PrefixChildDetails]] = {}
         for item in query.get_data():
-            if InfrahubKind.IPADDRESS in item.child_labels:
-                child_type = PrefixMemberType.ADDRESS
-            else:
-                child_type = PrefixMemberType.PREFIX
-
-            if item.prefix_uuid not in self._results_by_prefix_id:
-                self._results_by_prefix_id[item.prefix_uuid] = {}
-            if item.branch not in self._results_by_prefix_id[item.prefix_uuid]:
-                self._results_by_prefix_id[item.prefix_uuid][item.branch] = []
-            self._results_by_prefix_id[item.prefix_uuid][item.branch].append(
+            child_type = (
+                PrefixMemberType.ADDRESS if InfrahubKind.IPADDRESS in item.child_labels else PrefixMemberType.PREFIX
+            )
+            results.setdefault(item.prefix_uuid, []).append(
                 PrefixChildDetails(child_type=child_type, prefixlen=item.prefixlen, ip_value=item.ip_value)
             )
+        self._results_by_prefix_id = results
+        self._has_data = True
 
     async def get_children(
         self,
         ip_prefixes: list[Node] | None = None,
         prefix_member_type: PrefixMemberType | None = None,
-        branch_names: list[str] | None = None,
     ) -> list[PrefixChildDetails]:
-        if branch_names is not None and self.branch is not None:
-            raise ValueError(
-                "branch_names is incompatible with a branch-scoped PrefixUtilizationGetter; "
-                "the query is already restricted to the branch's visible set"
-            )
         await self._fetch_data()
-        prefix_child_details_list: list[PrefixChildDetails] = []
         if ip_prefixes is None:
             ip_prefixes = self.ip_prefixes
-        prefix_ids = {prefix.get_id() for prefix in ip_prefixes}
-        prefix_ids &= set(self._results_by_prefix_id.keys())
-        if not prefix_ids:
-            return prefix_child_details_list
-        for prefix_id in prefix_ids:
-            child_details_by_branch = self._results_by_prefix_id[prefix_id]
-            branch_names_to_check = branch_names or list(child_details_by_branch.keys())
-            for branch_name in branch_names_to_check:
-                for child_details in child_details_by_branch.get(branch_name, []):
-                    if prefix_member_type and child_details.child_type != prefix_member_type:
-                        continue
-                    prefix_child_details_list.append(child_details)
-        return prefix_child_details_list
+        result: list[PrefixChildDetails] = []
+        for prefix in ip_prefixes:
+            for child in self._results_by_prefix_id.get(prefix.get_id(), []):
+                if prefix_member_type and child.child_type != prefix_member_type:
+                    continue
+                result.append(child)
+        return result
 
     async def get_num_children_in_use(
         self,
         ip_prefixes: list[Node] | None = None,
         prefix_member_type: PrefixMemberType | None = None,
-        branch_names: list[str] | None = None,
     ) -> int:
-        children = await self.get_children(
-            ip_prefixes=ip_prefixes, prefix_member_type=prefix_member_type, branch_names=branch_names
-        )
+        children = await self.get_children(ip_prefixes=ip_prefixes, prefix_member_type=prefix_member_type)
         return len(children)
 
-    async def _get_prefix_use_fraction(
-        self, ip_prefixes: list[Node] | None = None, branch_names: list[str] | None = None
-    ) -> tuple[int, int]:
+    async def _get_prefix_use_fraction(self, ip_prefixes: list[Node]) -> tuple[int, int]:
         total_prefix_space = 0
         total_used_space = 0
-        if ip_prefixes is None:
-            ip_prefixes = self.ip_prefixes
         for ip_prefix in ip_prefixes:
             total_prefix_space += get_prefix_space(ip_prefix=ip_prefix)
             max_prefixlen = ip_prefix.prefix.obj.max_prefixlen  # type: ignore[attr-defined]
-            children = await self.get_children(
-                ip_prefixes=[ip_prefix], prefix_member_type=PrefixMemberType.PREFIX, branch_names=branch_names
-            )
+            children = await self.get_children(ip_prefixes=[ip_prefix], prefix_member_type=PrefixMemberType.PREFIX)
             for child in children:
                 total_used_space += 2 ** (max_prefixlen - child.prefixlen)
         return total_used_space, total_prefix_space
 
-    async def _get_address_use_fraction(
-        self, ip_prefixes: list[Node] | None = None, branch_names: list[str] | None = None
-    ) -> tuple[int, int]:
-        total_prefix_space = 0
-        if ip_prefixes is None:
-            ip_prefixes = self.ip_prefixes
-        for ip_prefix in ip_prefixes:
-            total_prefix_space += get_prefix_space(ip_prefix=ip_prefix)
+    async def _get_address_use_fraction(self, ip_prefixes: list[Node]) -> tuple[int, int]:
+        total_prefix_space = sum(get_prefix_space(ip_prefix=ip_prefix) for ip_prefix in ip_prefixes)
         total_used_space = await self.get_num_children_in_use(
-            ip_prefixes=ip_prefixes, prefix_member_type=PrefixMemberType.ADDRESS, branch_names=branch_names
+            ip_prefixes=ip_prefixes, prefix_member_type=PrefixMemberType.ADDRESS
         )
         return total_used_space, total_prefix_space
 
-    async def get_use_percentage(
-        self, ip_prefixes: list[Node] | None = None, branch_names: list[str] | None = None
-    ) -> float:
+    async def get_use_percentage(self, ip_prefixes: list[Node] | None = None) -> float:
         grand_total_used, grand_total_space = 0, 0
         address_prefixes, prefix_prefixes = [], []
         if ip_prefixes is None:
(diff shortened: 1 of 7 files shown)


backend/tests/component/core/ipam/test_ipam_utilization.py
ajtmccarty, line 1: is there a test that covers the "until data is modified" part of  `show the same value as their origin branch until data is modified`?

ajtmccarty (approved):

backend/infrahub/core/node/ipam.py
```diff
@@ -44,7 +44,7 @@ async def to_graphql(
                     self.prefix = retrieved.prefix  # type: ignore[attr-defined,union-attr]
                 utilization_getter = PrefixUtilizationGetter(db=db, ip_prefixes=[self])
```
ajtmccarty: thinking about this some more. is prefix utilization branch-agnostic? or does this class need to support correct branch-based filtering? for example, if I create a branch `br1`, then make some changes on the default branch, would this class include the default branch changes when querying on `br1`?

gmazoyer: I do not think that prefix utilisation is branch-agnostic, and I don't think it should be. A simple use case is as simple as wanting to remove some IP addresses from some IP prefix in a branch, the utilisation should reflect the branch's and not affect any other branches. That said, I think this PR needs more tests to verify the behaviour in other scenarios (such as the one just mentionned), it could be the fix is actually more complex that this one-liner.

backend/infrahub/core/ipam/utilization.py
```diff
@@ -60,6 +69,11 @@ async def get_children(
         prefix_member_type: PrefixMemberType | None = None,
         branch_names: list[str] | None = None,
     ) -> list[PrefixChildDetails]:
+        if branch_names is not None and self.branch is not None:
+            raise ValueError(
+                "branch_names is incompatible with a branch-scoped PrefixUtilizationGetter; "
+                "the query is already restricted to the branch's visible set"
+            )
```
ajtmccarty: this error feels like a foot-gun. if we need to prevent retrieving results using `branch_names` when the class is instantiated using a `branch`, then we should probably enforce this at construction time in some manner.

it seems like this class is trying to do 2 things using a single cypher query that we don't really do anywhere else. it wants to both 
- get the utilization for a particular branch
- get the latest utilization for the default branch
so this class and the query it calls either need to be very special snowflakes. or maybe we can just call the query twice if necessary, once for `main` and once for the user branch

it looks like this class is only used in 2 places, so we should be in a pretty good place to make whatever changes are necessary for it

backend/infrahub/core/query/ipam.py
```diff
@@ -1030,10 +1032,27 @@ def __init__(self, ip_prefixes: list[str], allocated_kinds: list[str], **kwargs)
 
     async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:  # noqa: ARG002
         self.params["ids"] = [p.get_id() for p in self.ip_prefixes]
-        self.params["time_at"] = self.at.to_string()
 
-        def rel_filter(rel_name: str) -> str:
-            return f"{rel_name}.from <= $time_at AND ({rel_name}.to IS NULL OR {rel_name}.to >= $time_at)"
+        branch = self.branch
```
ajtmccarty: I spent longer than I should have trying to clean this query up to be more in line with how we do cypher now (replace the `%(id_func)s` with `elementId`, use string-interpolation instead of an f-string) and to support getting both default branch and user branch utilization at once, but I think it is probably better to just rewrite this thing to work for a single branch and then call it twice.

I can look into that later this week or next week if you don't have the time for it now

gmazoyer: Tried my luck a rewriting the query with your comment in mind. It can now only be used with a single branch.

backend/infrahub/core/query/ipam.py
```diff
@@ -1005,9 +1006,6 @@ class IPPrefixUtilizationResult:
     prefixlen: int
     """Prefix length of the child IP value."""
 
-    branch: str
-    """Branch name where this allocation exists."""
-
     @classmethod
     def from_db(cls, result: QueryResult) -> IPPrefixUtilizationResult:
```
ajtmccarty: not required, but this method's logic belongs in `IPPrefixUtilization.get_data()`
coupling this thin data model to the cypher used in the query class is not SOLID

backend/infrahub/core/query/ipam.py
```diff
@@ -1021,81 +1019,84 @@ def from_db(cls, result: QueryResult) -> IPPrefixUtilizationResult:
...
-            deepest_branch_details[0] AS branch_level,
-            deepest_branch_details[1] AS branch,
             head(collect(is_active)) AS is_latest_active
         WHERE is_latest_active = TRUE
         """ % {
-            "id_func": db.get_id_function_name(),
+            "rel1_filter": rel_filter("r_rel1"),
+            "rel2_filter": rel_filter("r_rel2"),
+            "attr_filter": rel_filter("r_attr"),
+            "attr_val_filter": rel_filter("r_attr_val"),
```
ajtmccarty: also optional, but these queries could be changed to a single `branch_filter` like we do in most other queries if the cypher was written to just use one edge with variable name `r` in each CALL subquery. and you could get rid of the `rel_filter` function
so it would be something like
```
        CALL (pfx) {
            MATCH (pfx)-[r:IS_RELATED]-(rl:Relationship)
            WHERE rl.name IN $allocated_kinds_rel
            AND %(branch_filter)s
            RETURN r AS r_rel1, rl
        }}
        CALL (rl) {
            MATCH (rl:Relationship)<-[r:IS_RELATED]-(child:Node)
            WHERE any(l IN labels(child) WHERE l IN $allocated_kinds)
            AND %(branch_filter)s
            RETURN r AS r_rel2, child
        }}
```
and so on

just to be clear. definitely optional if you don't want to mess with this cypher any further, but this query is built in an unusual way different from most of our other cypher queries

backend/infrahub/core/query/ipam.py
```diff
@@ -1021,81 +1019,84 @@ def from_db(cls, result: QueryResult) -> IPPrefixUtilizationResult:
...
             -[r_attr:HAS_ATTRIBUTE]->(attr:Attribute)
-            -[r_attr_val:HAS_VALUE]->(av:{PREFIX_ATTRIBUTE_LABEL}|{ADDRESS_ATTRIBUTE_LABEL})
+            -[r_attr_val:HAS_VALUE]->(av:%(prefix_label)s|%(address_label)s)
         )
-        WHERE %(id_func)s(r_1) = %(id_func)s(r_rel1)
-        AND %(id_func)s(r_2) = %(id_func)s(r_rel2)
-        AND ({rel_filter("r_attr")})
-        AND ({rel_filter("r_attr_val")})
+        WHERE elementId(r_1) = elementId(r_rel1)
+        AND elementId(r_2) = elementId(r_rel2)
```
ajtmccarty: another weird piece of this query. I think you might be able to drop these and just use `r_rel1` and `r_rel2` in the `MATCH`

backend/infrahub/graphql/queries/resource_manager.py
```diff
@@ -211,21 +221,23 @@ async def resolve(
...
+                total_utilization if total_utilization is not None else await branch_getter.get_use_percentage()
             )
             default_branch_utilization = (
                 default_branch_utilization
                 if default_branch_utilization is not None
-                else await utilization_getter.get_use_percentage(branch_names=[registry.default_branch])
+                else await default_branch_getter.get_use_percentage()
             )
-            response["utilization_branches"] = total_utilization - default_branch_utilization
+            # Clamp at 0: deletions on the branch can make its view smaller than the default
```
ajtmccarty: sometimes the robot picks a weird word. I've never heard a lower bound described as a "clamp" before

Image

backend/tests/functional/ipam/test_ipam_utilization.py
```diff
@@ -545,21 +539,21 @@ async def test_step03_utilization_with_deletes(
...
+        # isolated branch, which still counts prefix + prefix2 plus its own live prefix_branch.
+        assert await main_getter.get_use_percentage(ip_prefixes=[container]) == 100 / 16
+        assert await branch_getter.get_use_percentage(ip_prefixes=[container]) == 18.75
+        # prefix_branch only exists on branch2; one of its 14 addresses was deleted on branch2.
+        assert await main_getter.get_use_percentage(ip_prefixes=[prefix_branch]) == 0
+        assert await branch_getter.get_use_percentage(ip_prefixes=[prefix_branch]) == (13 / 14) * 100
+        # prefix on main lost one of its 7 addresses; branch2 still sees those 7 (isolation)
+        # plus its own 7 minus the one deleted on branch2.
+        assert await main_getter.get_use_percentage(ip_prefixes=[prefix]) == (6 / 14) * 100
+        assert await branch_getter.get_use_percentage(ip_prefixes=[prefix]) == (13 / 14) * 100
```
ajtmccarty: pre-existing problem, but I don't think this is a functional test if it uses `PrefixUtilizationGetter` directly

backend/tests/functional/ipam/test_ipam_utilization.py
```diff
@@ -586,18 +580,21 @@ async def test_step03_graphql_prefix_pool_delete_utilization(
         assert result.data
         pool_data = result.data["InfrahubResourcePoolUtilization"]
         assert pool_data["count"] == 2
-        assert pool_data["utilization"] == 100 / 16
-        assert pool_data["utilization_default_branch"] == 100 / 32
-        assert pool_data["utilization_branches"] == 100 / 32
+        # Pool space = 2 * 256 = 512. On branch2 (isolated) container shows prefix + prefix2
```
ajtmccarty: I'm not sure if `(isolated)` means anything here

ajtmccarty (approved):

backend/tests/functional/ipam/test_ipam_utilization.py
gmazoyer: Took the chance to clean this up while the file was open. The three tests that called `PrefixUtilizationGetter` directly are gone. Their coverage of `container`, `prefix` and `prefix_branch` on both branches across all three steps is already exercised through the GraphQL pool and per-prefix queries in the adjacent tests. The only thing uniquely tested was a prefix-with-no-children returning 0, which is already covered by the component-level tests.

gmazoyer:
Summary of the changes to the Cypher:

- The query now has the same shape as `IPPrefixSubnetFetch` and `IPPrefixReconcileQuery`
- One `MATCH path = (pfx)-[r_rel1:IS_RELATED]-(rl)<-[r_rel2:IS_RELATED]-(child)-[r_attr:HAS_ATTRIBUTE]->(attr)-[r_attr_val:HAS_VALUE]->(av)` covers all four hops.
- `all(r IN relationships(path) WHERE %(branch_filter)s)` is the single branch and time scope filter.
- The deepest-active dedup is inside a scoped `CALL (pfx) { ... ORDER BY ... head(collect(av)), head(collect(is_active)) ... }`, so the outer pipeline never sees the expansion.

Both `pfx` and `child` get an `IS_PART_OF` aliveness gate. A prefix or a child that has been deleted on the branch (or any prefix passed in by a caller that loaded it branch-agnostic) is filtered out before its edges are considered.

The query never counts the parent prefix itself as one of its own children. The `child <> pfx` predicate handles this at the node level, which is robust to multiple historical edges between the same node and the relationship node.

The deepest-active grouping is keyed on `(pfx, child)`, not `(pfx, child, av)`. If a child's value (the IP prefix or address attribute) has been edited on the branch, only the deepest active value is kept and the child contributes exactly one row.


Some polish:

- `elementId(...)` instead of `%(id_func)s(...)`.
- Allocated kinds passed as Cypher parameters, the child node uses a label expression `(child:BuiltinIPAddress|BuiltinIPPrefix)`.
- `attr.name`, `rl.name` and `child <> pfx` live in the path's `WHERE`.
- `sum_branch_level` is written as an explicit four-term sum and the `from` tiebreakers reference each edge directly in the `ORDER BY`, no list-of-timestamps or scalar aliases.

backend/infrahub/core/query/ipam.py
```diff
@@ -1005,107 +1006,101 @@ class IPPrefixUtilizationResult:
...
+            MATCH (pfx)-[r:IS_PART_OF]-(:Root)
+            WHERE %(branch_filter)s
+            RETURN r AS pfx_root
+            ORDER BY r.branch_level DESC, r.from DESC
+            LIMIT 1
+        }
+        WITH pfx, pfx_root
+        WHERE pfx_root.status = "active"
+        WITH pfx
+        CALL (pfx) {
```
ajtmccarty: I think that this subquery works correctly, but it would probably perform better to break it up into 2 or 4 subqueries that filtered out deleted/inactive edges along the way. so it would be something like
```
MATCH (pfx)-[:IS_RELATED]-(rl:Relationship)
WHERE rl.name IN $allocated_kinds_rel
WITH DISTINCT pfx, rl
CALL (pfx, rl) {
    MATCH (pfx)-[r:IS_RELATED]-(rl)
    WHERE %(branch_filter)
    RETURN r AS r_rel1
    ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
    LIMIT 1
}
WITH pfx, r_rel1, rl
WHERE r_rel1.status = "active"
CALL (rl) {
    MATCH (rl:Relationship)<-[r:IS_RELATED]-(child:%(allocated_labels)s)
    WHERE %(branch_filter)
    RETURN r AS r_rel2, child
    ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
    LIMIT 1
}
WITH pfx, r_rel1, rl, r_rel2, child
WHERE r_rel2.status = "active"
CALL (child) {
    MATCH (child)-[r:HAS_ATTRIBUTE]->(attr:Attribute)
    WHERE attr.name IN ["prefix", "address"]
    AND %(branch_filter)
    RETURN r AS r_attr, attr
    ORDER BY r.branch_level DESC, r.from DESC, r.status ASC
    LIMIT 1
}
WITH pfx, r_rel1, rl, r_rel2, child, r_attr, attr
WHERE r_attr.status = "active"

// ... and then one more for the HAS_VALUE edge
// and you might be able to drop the r_rel1, r_rel2, etc. along the way
// then you know that edge row in the result set is a single "child" and its active prefix/address value
```

backend/infrahub/core/query/ipam.py
```diff
@@ -1005,107 +1006,101 @@ class IPPrefixUtilizationResult:
...
+                av,
+                r_rel1.branch_level + r_rel2.branch_level + r_attr.branch_level + r_attr_val.branch_level AS sum_branch_level,
+                all(r IN relationships(path) WHERE r.status = "active") AS is_active,
+                r_rel1, r_rel2, r_attr, r_attr_val
+            ORDER BY child.uuid, sum_branch_level DESC, r_rel1.from DESC, r_rel2.from DESC, r_attr.from DESC, r_attr_val.from DESC
+            WITH child, head(collect(av)) AS av, head(collect(is_active)) AS is_latest_active
+            WHERE is_latest_active
+            RETURN child, av
+        }
+        CALL (child) {
```
ajtmccarty, line 1068: I don't think you need this part. if there's an active relationship to the `Node` we can assume that the `Node` itself is active

ajtmccarty (approved):
the suggested cypher changes are optional. this is a big improvement as is

PR #9242, opened 2026-05-13
fix(frontend): copy to clipboard fails over HTTP (non-secure context)

### LINKED ISSUE #8857: bug: Copy ID and HFID does not work

DharmaBytesX:
## Summary

- `navigator.clipboard` is unavailable in non-secure contexts (HTTP, non-localhost); previous code attempted it unconditionally, causing silent failures.
- Added an explicit `window.isSecureContext` guard: when the Clipboard API is unavailable, a synchronous `document.createRange()` + `window.getSelection()` + `execCommand("copy")` fallback is used instead.
- Replaced the previous `textarea`-based fallback, which was broken by React Aria's Popover focus trap — the textarea could not receive focus while the menu was open, so `execCommand("copy")` had no selection to act on.

Fixes #8857

## Test plan

- [ ] Open Infrahub over HTTP (non-localhost IP)
- [ ] Navigate to any object detail page → Actions → **Copy ID**: verify clipboard contains the UUID
- [ ] Actions → **Copy HFID**: verify clipboard contains the human-friendly ID
- [ ] Profile → Tokens → create a new token → click copy: verify clipboard contains the token value
- [ ] Verify all three copy actions still work over HTTPS / localhost (secure context, Clipboard API path)

---
## Summary by cubic
Fixes copy actions over HTTP so users can copy IDs and tokens in non-secure contexts, without breaking HTTPS behavior. Restores the async contract of `useCopyToClipboard` to preserve await ordering. Fixes #8857.

- **Bug Fixes**
  - Guard on `window.isSecureContext` and `navigator.clipboard`; fall back to selection-based `document.createRange()` + `window.getSelection()` + `document.execCommand("copy")` compatible with the React Aria popover focus trap.
  - Make `copyToClipboard` async; add tests for secure/non-secure, missing clipboard, and rejected writes; fix Biome lint errors in the tests.

<sup>Written for commit 399e42855e17c1aca8541326a56b2f88f1273430. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9242?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/changelog/8857.fixed.md
+++ b/changelog/8857.fixed.md
@@ -1,0 +1,1 @@
+Fixed "Copy ID", "Copy HFID", and "Copy Token" actions not working when accessing Infrahub over HTTP (non-secure context).
--- a/frontend/app/src/shared/hooks/useCopyToClipboard.test.tsx
+++ b/frontend/app/src/shared/hooks/useCopyToClipboard.test.tsx
@@ -1,4 +1,3 @@
-import React from "react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { render } from "vitest-browser-react";
 
@@ -7,7 +6,11 @@ import { useCopyToClipboard } from "@/shared/hooks/useCopyToClipboard";
 function CopyButton({ value }: { value: string }) {
   const { isCopied, copyToClipboard } = useCopyToClipboard();
   return (
-    <button data-testid="copy-btn" data-copied={String(isCopied)} onClick={() => copyToClipboard(value)}>
+    <button
+      data-testid="copy-btn"
+      data-copied={String(isCopied)}
+      onClick={() => copyToClipboard(value)}
+    >
       {isCopied ? "copied" : "copy"}
     </button>
   );
--- a/frontend/app/src/shared/hooks/useCopyToClipboard.test.tsx
+++ b/frontend/app/src/shared/hooks/useCopyToClipboard.test.tsx
@@ -1,0 +1,93 @@
+import React from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { render } from "vitest-browser-react";
+
+import { useCopyToClipboard } from "@/shared/hooks/useCopyToClipboard";
+
+function CopyButton({ value }: { value: string }) {
+  const { isCopied, copyToClipboard } = useCopyToClipboard();
+  return (
+    <button data-testid="copy-btn" data-copied={String(isCopied)} onClick={() => copyToClipboard(value)}>
+      {isCopied ? "copied" : "copy"}
+    </button>
+  );
+}
+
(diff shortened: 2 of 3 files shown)


DharmaBytesX:
Tested manually on HTTP (non-localhost IP). Copy ID, Copy HFID, and Copy Token all work correctly after the fix.

DharmaBytesX:
Hi! This fix for the clipboard issue over HTTP has been tested manually and is ready for review — could someone take a look when you have a moment? Thanks!

pa-lem (changes requested):
Thanks for the fix, the secure-context detection and selection-based fallback are the right approach. One small follow-up to keep the public API consistent:

The hook signature changed from `async (value) => Promise<void>` to `(value) => void`. That breaks the contract for
`frontend/app/src/shared/components/buttons/clipboard.tsx:21-24`, which still does:

```javascript
const handleCopy = async () => {
  await copyToClipboard(value);   // now awaits a non-Promise → resolves immediately
  toast(<Alert message={alert} type={ALERT_TYPES.INFO} />);
};
```

The toast now fires before the async navigator.clipboard.writeText() settles, so on the secure-context path the success toast can show even if the write rejects.

Suggested change would be that we keep the hook awaitable:

```javascript
const copyToClipboard = React.useCallback(async (value: string) => {
  function confirmCopied() {
    setIsCopied(true);
    setTimeout(() => setIsCopied(false), COPIED_FEEDBACK_DURATION);
  }

  if (!window.isSecureContext || !navigator.clipboard) {
    oldSchoolCopy(value);
    confirmCopied();
    return;
  }

  try {
    await navigator.clipboard.writeText(value);
    confirmCopied();
  } catch {
    oldSchoolCopy(value);
    confirmCopied();
  }
}, []);
```

This preserves the original Promise<void> return type, so await copyToClipboard(...) in clipboard.tsx keeps its original ordering guarantee, and the other four
call sites (which don't await) remain unaffected.

Nice-to-have: a small unit test for useCopyToClipboard covering both paths (mock window.isSecureContext = false and navigator.clipboard = undefined) would
protect this dual-path logic from future regressions.

DharmaBytesX (commented):
Thx for the review @pa-lem. Made the fix with Claude Code, restored the async contract and added unit tests covering both paths. Then, I manually verified the fix by booting up a dev instance and testing on Firefox myself.

pa-lem (approved):

Issue #2592, opened 2024-03-12
feature: enhance constraint validator determination


ajtmccarty:
### Component

API Server / GraphQL

### Describe the Feature Request

follow-up to #2573 

the `ConstraintValidatorDeterminer` class should be enhanced to only select those validators that are actually necessary.

### Describe the Use Case

running fewer validators will reduce the resources and amount of time required to validate a given branch

### Additional Information

this is the current implementation

1. if there are no diffs, then there are no changes and we don't need to run any constraints
2. always do the node property validators for everything in the schema, (node.uniqueness_constraints.update, node.parent.update, node.children.update). we always do all of them b/c they can reference other models in the schema
3. for each schema kind in the diff
    a.  if a given attribute is in the diff (added, updated, or removed) and a given property of the given attribute is not None, do the validator (attribute.regex.update, etc.) for the property on that attribute
    b. if a given relationship is in the diff (added, updated, or removed) and a given property of the given relations is not None, do the relationship validator (relationship.peer.update, etc.) for the property on that relationship

some potential improvements:
1. we can be smarter about which constraint validators we include based on the DiffSummary (list of `NodeDiff`s). for example, if the number of members in a relationship has not changed, then we don't need to run the `relationship.min_count.update` validator or the `relationship.max_count.update` validator
2. if we include more information about schema changes in the arguments we pass to `ConstraintValidatorDeterminer`, then we can check when we need to run the `relationship.peer.update` validator
3. the `node.generate_profile.update` constraint is run for every node in the schema for every change right now b/c the logic to determine schema properties constraints will run a constraint for every property that is not `None` and `generate_profile` is a Boolean (new in #3463)

ajtmccarty:
completed in #9943

PR #9247, opened 2026-05-13
fix: data-only proposed change spuriously triggers schema-migration validators

### LINKED ISSUE #9246: bug: Data-only proposed changes trigger schema-migration validators across all kinds
### Description

When a Proposed Change contains only data changes (no schema changes), the `Process schema integrity` flow still triggers `schema_validate_migrations` with a large number of constraints — including `node.uniqueness_constraints.update` validators that scan the entire database for every kind in the branch.

At small scale this is wasteful; at large scale it is expensive enough to materially slow down the PC pipeline.

### Reproduction

1. Create a branch with only data changes (no schema modifications).
2. Open a Proposed Change against the base branch.
3. Observe the `Process schema integrity` task running and dispatching `schema_validate_migrations` with constraints covering uniqueness checks for many kinds — most of which have no data diff in this PC.

### What we expect

A data-only PC should either skip migration-style validators entirely or only run validators that are strictly necessary for the kinds whose data actually changed. In particular, full-database uniqueness scans should not be invoked for kinds with no diff.

### Notes

- Debug logs added inside `run_proposed_change_schema_integrity_check` showed a non-empty `schema diff` for what should have been an unchanged schema, and a large `constraints_from_data_diff` containing constraints unrelated to the changed kinds.
- Likely areas to investigate:
  - `backend/infrahub/proposed_change/tasks.py::run_proposed_change_schema_integrity_check`
  - `backend/infrahub/core/schema/schema_branch.py` (schema processing and diff)
  - `backend/infrahub/core/schema/basenode_schema.py` (schema diff logic / `update()`)
  - `backend/infrahub/core/validators/determiner.py::ConstraintValidatorDeterminer.get_constraints`

### Impact

High on large deployments where the uniqueness validators scan many nodes per kind: each unnecessary constraint adds full-DB scan time to every data-only PC.

### Related

Related to #2592 (enhance constraint validator determination).

fatih-acar:
## Summary

Fixes #9246.

A data-only Proposed Change was running `schema_validate_migrations` over a large set of constraints — including full-DB uniqueness scans for kinds with no actual data diff. Root cause was a chain of three issues that combined to produce a non-empty schema diff and an over-broad constraint set:

- **Virtual relationships had no id.** `manage_profile_relationships` and `manage_object_template_relationships` appended `RelationshipSchema(...)` without an `id`. The diff logic in `_diff_element` then flagged these as `added` on every node when comparing two identical processed schemas. Fix: assign a stable per-node id `__virtual__/<kind>/<rel_name>` at injection time. The persistence layer already filters virtual relationships by name, so the synthetic id never reaches the DB.

- **`BaseNodeSchema.update()` over-cleared empty-string fields.** Empty-string sentinels in YAML (e.g. `LocationContinent.parent: ""`, `LocationRack.children: ""`) got unconditionally converted to `None` on the candidate schema during `update()` — even when the local already held `""`. That produced a phantom `changed={'parent': None}` against dest_schema. Fix: only clear when local actually has a value to clear.

- **Constraint determiner was over-broad.** `ConstraintValidatorDeterminer._get_all_property_constraints` iterated every schema in the branch and emitted node-level update constraints (uniqueness, hierarchy, inherit_from, generate_profile) for kinds with no data diff. Fix: scope to schemas whose kind appears in `self._node_kinds`. The fixture-count assertions in `test_determiner` and `test_proposed_change` were updated to match the now-scoped behaviour (e.g. the `_get_proposed_change_schema_integrity_constraints` test drops from 248 to 14 constraints — unrelated kinds are no longer revalidated).

Result: on a data-only PC, the schema diff is now empty and `constraints_from_data_diff` only contains constraints for kinds that actually changed.

## Test plan

- [x] `backend/tests/component/core/constraint_validators/test_determiner.py` (4 tests)
- [x] `backend/tests/component/core/schema_manager/` (177 tests)
- [x] `backend/tests/component/message_bus/operations/requests/test_proposed_change.py` (3 tests)
- [x] `backend/tests/component/core/schema_manager/test_template_resource_pool_relationships.py` (2 tests)
- [ ] Manual: reproduce on a branch with only data changes — `schema diff:` log should print `added={} changed={} removed={}` and `constraints_from_data_diff` should be scoped to changed kinds only.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

---
## Summary by cubic
Prevents schema-migration validators from running on data-only Proposed Changes by eliminating phantom schema diffs and scoping integrity checks to changed kinds. Data-only PCs now show an empty schema diff and avoid costly full-DB uniqueness scans.

- **Bug Fixes**
  - Assign deterministic ids to virtual relationships (`__virtual__/<kind>/<rel_name>`) so identical processed schemas don’t diff as added; persistence ignores these by name.
  - In `BaseNodeSchema.update()`, only clear optional text fields when the local value is set, avoiding "" → None phantom changes.
  - Scope `ConstraintValidatorDeterminer._get_all_property_constraints` to kinds present in the data diff (including `Schema*` kinds only when they’re in the diff).

<sup>Written for commit d846e73287537fdb23866858e51001fb539a8644. Summary will update on new commits.</sup>

### DIFF
--- a/backend/infrahub/core/schema/basenode_schema.py
+++ b/backend/infrahub/core/schema/basenode_schema.py
@@ -675,9 +675,12 @@ class BaseNodeSchema(GeneratedBaseNodeSchema):
 
         super().update(other=other)
 
-        # Allow to specify empty string to remove existing fields values
+        # Allow to specify empty string to remove existing fields values.
+        # Only clear when self actually has a value to clear — otherwise we synthesize a
+        # phantom diff (e.g. "" → None) when both sides start identical, which falsely
+        # triggers schema migration validation in the proposed-change integrity flow.
         for field_name in OPTIONAL_TEXT_FIELDS:
-            if getattr(other, field_name, None) == "":  # noqa: PLC1901
+            if getattr(other, field_name, None) == "" and getattr(self, field_name, None) not in (None, ""):  # noqa: PLC1901
                 setattr(self, field_name, None)
 
         return self
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2432,15 +2432,22 @@ class SchemaBranch:
             ):
                 continue
 
+            # Synthetic, deterministic id so diff matching is stable across processed schemas.
+            # See _diff_element in basenode_schema.py: two synthetic relationships with id=None
+            # on either side are reported as `added` even when structurally identical.
+            synthetic_id = f"__virtual__/{node_name}/{PROFILES_RELATIONSHIP_NAME}"
+
             # Add relationship between node and profile
             if PROFILES_RELATIONSHIP_NAME not in node.relationship_names:
                 node_schema = self.get(name=node_name, duplicate=True)
 
-                node_schema.relationships.append(RelationshipSchema(**profiles_rel_settings))
+                node_schema.relationships.append(RelationshipSchema(id=synthetic_id, **profiles_rel_settings))
                 self.set(name=node_name, schema=node_schema)
             else:
                 has_changes: bool = False
                 rel_profiles = node.get_relationship(name=PROFILES_RELATIONSHIP_NAME)
+                if rel_profiles.id != synthetic_id:
+                    has_changes = True
                 for name, value in profiles_rel_settings.items():
                     if getattr(rel_profiles, name) != value:
                         has_changes = True
@@ -2450,6 +2457,8 @@ class SchemaBranch:
 
                 node_schema = self.get(name=node_name, duplicate=True)
                 rel_profiles = node_schema.get_relationship(name=PROFILES_RELATIONSHIP_NAME)
+                if rel_profiles.id != synthetic_id:
+                    rel_profiles.id = synthetic_id
                 for name, value in profiles_rel_settings.items():
                     if getattr(rel_profiles, name) != value:
                         setattr(rel_profiles, name, value)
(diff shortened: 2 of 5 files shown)


backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -675,9 +675,12 @@ def update(self, other: HashableModel) -> Self:
 
         super().update(other=other)
 
-        # Allow to specify empty string to remove existing fields values
+        # Allow to specify empty string to remove existing fields values.
+        # Only clear when self actually has a value to clear — otherwise we synthesize a
+        # phantom diff (e.g. "" → None) when both sides start identical, which falsely
+        # triggers schema migration validation in the proposed-change integrity flow.
```
ajtmccarty, line 681: I don't think this should mention proposed changes or schema migration/validation

backend/infrahub/core/schema/schema_branch.py
ajtmccarty, line 1: the synthetic ID fix seems like a bit of a hack to me. we're setting a pretend ID that isn't really part of the schema processing logic to influence a method on a different class. If the issue is in `BaseNodeSchema._to_diff_element()`, then I think that the change should be in that method, perhaps ignoring missing IDs on relationships with particular names.

Ideally, refactoring of how we process schemas in general and extracting `BaseNodeSchema.diff` to its own component would fix this problem, but that is way out of scope for what you're working on here

backend/infrahub/core/validators/determiner.py
```diff
@@ -83,16 +83,17 @@ async def _get_constraints_for_one_schema(self, schema: MainSchemaTypes) -> list
 
     async def _get_all_property_constraints(self) -> list[SchemaUpdateConstraintInfo]:
         constraints: list[SchemaUpdateConstraintInfo] = []
-        schemas = list(self.schema_branch.get_all(duplicate=False).values())
+        # Scope to kinds with actual data diffs — otherwise we revalidate node-level
+        # constraints (uniqueness, hierarchy, inherit_from, generate_profile) for every
+        # schema in the branch even when only one kind's data changed.
+        schemas = [
+            schema for schema in self.schema_branch.get_all(duplicate=False).values() if schema.kind in self._node_kinds
+        ]
```
ajtmccarty, line 91: I think you may have also addressed #2592 (at least partially) with this one

ajtmccarty (commented):
this is a big improvement. I have a comment on the synthetic IDs bit, but this is a great update

ajtmccarty:
replaced by #9896

PR #9249, opened 2026-05-13
prevent duplicate schemas in case of stale registry

### LINKED ISSUE #9250: bug: schema load/update logic does not explicitly prevent duplicate Node, Generic, Attribute, or Relationship Schemas
### Component

API Server / GraphQL

### Infrahub version

1.9.4

### Current Behavior

Infrahub defines uniqueness constraints and our various internal schema data structures
- `NodeSchema` and `GenericSchema` are unique on `name` and `namespace`
- `AttributeSchema` and `RelationshipSchema` are unique on `name` and parent schema's kind (namespace + name)
but we don't actually enforce these where we load the schemas to the database. Infrahub _does_ enforce these uniqueness constraints in the schema loading API endpoint and that covers most cases, but not all of them

### Expected Behavior

It should be completely impossible to load duplicated Node, Generic, Attribute, or Relationship Schemas to the database. any Schema object that would be loaded as a duplicate should instead be an update to the existing object, basically an "upsert"

### Steps to Reproduce

there is no known method for reliably reproducing this right now b/c we have addressed these bugs in other places, but never at the database update level

the most recent appearance of this type of bug was during a branch merge operation that included a new schema where the new schema was merged at the database level, but not all workers had refreshed their registries with the newest version of the schema. so when a worker handled an updated CoreReadOnlyRepository that included the new schema, it treated the new schema as an object to be created even though it already existed at the database level

### Additional Information

_No response_

ajtmccarty:
# Why

When a schema is being updated, there is always a period of time when the workers not doing the update will have a stale in-memory `SchemaBranch` cache relative to the database. This is generally fixed when all the workers receive and act on the `RefreshRegistryBranches` message. But if something goes wrong and a worker with a stale cache begins processing a schema update that includes an added schema, attribute, or relationship, then the stale worker will end up creating a duplicated schema, attribute, or relationship at the database level. The next time a user creates a proposed change or tries to merge a branch, then they will get an error about uniqueness constraint violations on SchemaNode/Generic/Attribute/Relationship and this can be difficult to remedy if data exists for the duplicated schema/attribute/relationship.

Goal: Prevent duplicated schema entities from being created by checking the database for existing schemas (by namespace+name or UUID) before creating any schemas and attributes/relationships (by name+parent schema or UUID) before creating any attributes or relationships.

Closes #9250

### Behavioral
- `load_schema_to_db` and `update_schema_to_db` are now batch-aware and idempotent: a single `SchemaSummaryQuery` resolves every input schema's existence on the branch via `(namespace, name) UNION (uuid)`, then `_upsert_node_to_db` dispatches to insert or update based on the resolved DB Node.
- Stale-registry race on children: re-loading a schema where another worker concurrently added an attribute or relationship no longer creates a duplicate `SchemaAttribute` / `SchemaRelationship` for that name.
- `load_node_to_db` / `update_node_in_db` are removed; `create_node_in_db` is the only remaining public single-item entry point (insert-only, used by migrations).

### Implementation notes
- New `SchemaSummaryQuery` (`backend/infrahub/core/schema/queries.py`) new Cypher query that returns each matched parent's uuid, `is_generic` flag, and `name → uuid` maps for its active attributes and relationships. `kind_filter` and `uuid_filter` are UNION'd in the prefilter so schema kind renames are covered.
- updates to retrieving schema, attribute, and relationship Nodes from the database to use bulk `get-many()` queries instead of single `get_one()`s

### What stayed the same
- GraphQL / REST schema contracts unchanged.
- `SchemaDiff` semantics unchanged.

### Suggested review order
1. `backend/infrahub/core/schema/queries.py` — new `SchemaSummary`, `SchemaSummaryIndex`, and `SchemaSummaryQuery`.
2. `backend/infrahub/core/schema/manager.py` — `_get_existing_schema_summary_map`, `_prefetch_existing_children`, `_upsert_node_to_db`, `_create_node_in_db`, `_update_existing_node_in_db`, and the refactored `load_schema_to_db` / `update_schema_to_db`.
3. Tests:
   - `backend/tests/component/core/schema_manager/test_schema_summary_query.py` — class-scoped coverage of the new query (filter semantics, child-name filters, edge cases).
   - `backend/tests/component/core/schema_manager/test_manager_schema.py` — idempotency, stale-registry dedup, rename + add-id-less-child, type-mismatch rejection.
   - `backend/tests/integration/schema_lifecycle/test_schema_load_fake_id.py` — end-to-end via `client.schema.load(...)`.

## How to review

Focus areas:
- The Cypher in `SchemaSummaryQuery.query_init` — branch filtering, `OPTIONAL MATCH` placeholder semantics inside `CALL` subqueries (each child phase preserves the parent row when its name filter rejects all candidates), and the post-prefilter `WHERE` that handles the three states (`$filter_kinds IS NULL AND $filter_uuids IS NULL` → all; either filter set → only matching).
- The DB-first ordering in `_prefetch_existing_children` and `_update_existing_node_in_db` — the reasoning is that `(parent, child name)` is the authoritative uniqueness key on the branch, so the row currently linked under that name wins.
- `_get_existing_schema_summary_map`'s type-mismatch guard (raises `ValueError` when a NodeSchema input collides with an existing GenericSchema row of the same `(namespace, name)`, or vice versa).

## Impact & rollout

- **Backward compatibility:** Public method removals: `SchemaManager.load_node_to_db` and `SchemaManager.update_node_in_db` are gone. In-tree callers (one migration, one benchmark, three component tests) updated. 
- **Performance:** One additional Cypher query per `load_schema_to_db` / `update_schema_to_db` invocation (the batch `SchemaSummaryQuery`), independent of input size. Per-parent child fetches collapse into a single `get_many`. No measurable regression on full internal-schema reload in benchmarks; the registry-stale dedup path actually saves writes by reusing rows.
- **Deployment notes:** Safe to deploy. No data migration required — existing rows are still keyed by uuid and `(namespace, name)` as before; the change is purely in how new writes are routed. The next schema reload on each branch self-heals any in-memory cache drift.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Make schema loads idempotent and prevent duplicate schema nodes, attributes, and relationships when the in-memory registry is stale. Adds a fast summary query, bulk child prefetch, and a pure-insert `create_node_in_db`, and cleans up rename and type-mismatch cases.

- **Bug Fixes**
  - `load_schema_to_db` and `update_schema_to_db` upsert using a precomputed summary map: match by UUID first (rename-safe), then by `(namespace, name)`.
  - Reuse existing `SchemaAttribute`/`SchemaRelationship` by field name when ids are missing to avoid duplicates.
  - Reject Node/Generic type mismatches with a clear error, and never let input ids overwrite existing DB ids.

- **Refactors**
  - Added `SchemaSummaryQuery` with indexed results, used by `_get_existing_schema_summary_map` and `_prefetch_existing_children` for fast bulk lookups.
  - Introduced `create_node_in_db` (pure insert) and internal `_upsert_node_to_db`/`_create_node_in_db`/`_update_existing_node_in_db`; updated `m017_add_core_profile`, `m073_unify_ip_pool_resource_identifier`, benchmarks, and tests (including idempotent loads, rename reuse, and fake-id rejection).

<sup>Written for commit 3d7b6f3338d269eab84e0ab39c80384f01946dd5. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9249?utm_source=github">Review in cubic</a></sup>


backend/infrahub/core/schema/queries.py
ajtmccarty, line 1: new query to get a quick summary of the current name/kind and UUID of the node/generic schemas on this branch, filtered by kind, UUIDs, attribute names, and relationship names.

returns kind and uuid for every matching SchemaNode/Generic along with maps of their attribute and relationship names to UUIDs filtered by schema kind, schema UUID, attribute names, and relationship names

backend/infrahub/core/schema/manager.py
ajtmccarty, line 1: somewhat extensive refactoring to use the new `SchemaSummaryQuery` to check the state of the database before making any changes. also some renaming.

ideally, I think the schema loading/updating logic should be moved out of SchemaManager and into its own component, but this PR has already grown too large and taken longer than I wanted, so that's for a different day

backend/infrahub/core/schema/manager.py
```diff
@@ -240,43 +243,82 @@ async def update_schema_to_db(
...
+        for kind in upsert_kinds:
+            one_schema = schema.get(name=kind, duplicate=False)
+            # SchemaDiff excludes Profile/Template, so only NodeSchema/GenericSchema can appear here
+            # we also don't save Profile/Template schemas to the database, they're always generated
+            if isinstance(one_schema, (NodeSchema, GenericSchema)):
+                upsert_schemas.append(one_schema)
+        schema_summary_map = await self._get_existing_schema_summary_map(
+            schemas=upsert_schemas, branch=branch, db=db, at=at
+        )
+        existing_children = await self._prefetch_existing_children(
```
ajtmccarty, line 257: new for bulk-fetching all the attributes and relationships required for the schemas in question

backend/tests/component/core/schema_manager/test_manager_schema.py
ajtmccarty, line 1: mostly new tests for the "schema added on the database, but registry is stale" scenario

backend/tests/integration/schema_lifecycle/test_schema_load_fake_id.py
ajtmccarty, line 1: pretty sure that we don't have a test for this case and wanted to be sure it was prevented. we shouldn't let users set an ID on a schema before loading it

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -3100,6 +3111,218 @@ async def test_update_node_in_db_node_schema(db: InfrahubDatabase, default_branc
     assert results[new_node.attributes[0].id].unique.value is False
 
 
+async def test_load_schema_to_db_idempotent(db: InfrahubDatabase, default_branch: Branch) -> None:
```
polmichel: Since this file is already quite long, all the following tests (except maybe the last one, `test_load_schema_to_db_rejects_type_mismatch`) could belong to the same test class (or to a separate test file).

backend/tests/component/core/schema_manager/test_schema_summary_query.py
```diff
@@ -0,0 +1,341 @@
...
+        """kind_filter with multiple pairs returns each of them."""
+        query = await SchemaSummaryQuery.init(
+            db=db,
+            branch=default_branch_scope_class,
+            kind_filter=[("Querytest", "AttrsOnly"), ("Querytest", "Both")],
+        )
+        await query.execute(db=db)
+        assert query.get_summaries().get_kinds() == {"QuerytestAttrsOnly", "QuerytestBoth"}
+
+    async def test_kind_filter_nonexistent(
```
polmichel, line 109: Would a test exercising a path where the kind filter contains both a non-existent and an existent kind could be useful? It would describe a new expected behavior of the system and, I suspect, exercise some extra logic in the query.

polmichel: The same could be applied regarding non-existent/existent attributes and relationships

backend/tests/integration/schema_lifecycle/test_schema_load_fake_id.py
```diff
@@ -0,0 +1,64 @@
...
+                }
+            ],
+        }
+
+    @pytest.fixture(scope="class")
+    def widget_schema_with_fake_id(self, widget_schema: dict[str, Any]) -> dict[str, Any]:
+        widget_node = {**widget_schema["nodes"][0], "id": str(uuid.uuid4())}
+        return {"version": "1.0", "nodes": [widget_node]}
+
+    async def test_step_01_load_initial(self, client: InfrahubClient, widget_schema: dict[str, Any]) -> None:
```
polmichel: Is this a real test or a fixture? The following ones look like real tests, but I don't think this one is actually bearing the responsibility to verify that a widget schema can be loaded.

backend/tests/integration/schema_lifecycle/test_schema_load_fake_id.py
```diff
@@ -0,0 +1,64 @@
...
+        response = await client.schema.load(schemas=[widget_schema])
+        assert not response.errors
+
+    async def test_step_02_only_one_db_row_after_initial_load(self, db: InfrahubDatabase) -> None:
+        schema_nodes = await NodeManager.query(db=db, schema="SchemaNode", filters={"name__value": "Widget"})
+        assert len(schema_nodes) == 1
```
polmichel: I guess the same argument could be discussed on `step_02`

backend/infrahub/core/schema/manager.py
```diff
@@ -345,101 +559,149 @@ async def load_node_to_db(
...
+        for attribute in schema.local_attributes:
+            resolved_id = attr_resolved_ids.get(attribute.name)
+            if resolved_id and resolved_id in existing_fields:
+                await self.update_attribute_in_db(
+                    item=attribute, attr=existing_fields[resolved_id], db=db, at=at, user_id=user_id
                 )
-                new_node.attributes.append(new_attr)
+                new_attr_by_name[attribute.name].id = resolved_id
+            elif not resolved_id:
+                new_db_attr = await self.create_attribute_in_db(
```
polmichel, line 679: I was thinking about the case where a stale worker schema is missing a deletion that is already in the database, I don't think we can differentiate between a re-addition that is intentional and one that is unwanted due to a stale schema. I do not think either that this could be important since another deletion could be done.

polmichel (approved):
LGTM ! A few optional comments and there are some merge conflicts on the current state of the PR

backend/infrahub/core/schema/manager.py
ajtmccarty: yes I think deleting an entity twice is fine. ideally, we wouldn't try to do that, but it's not actually going to make a difference if we try to delete a deleted thing

backend/tests/component/core/schema_manager/test_manager_schema.py
ajtmccarty: good point. moved these `test_prevent_loading_duplicates.py`

backend/tests/component/core/schema_manager/test_schema_summary_query.py
ajtmccarty: good idea. added the tests

backend/tests/integration/schema_lifecycle/test_schema_load_fake_id.py
ajtmccarty: yes this test is not quite right. I've updated it so it is a fixture and one test case

ajtmccarty (commented):
thanks for the review of another too-big PR, Pol

PR #9257, opened 2026-05-15
User groups auto-creation MVP

polmichel:
## Why

Today, Infrahub can only associate users with a `CoreAccountGroup` if a group with the **exact same name** as the external IdP claim already exists locally. Administrators onboarding a new team via SSO must therefore pre-create every group out-of-band — a manual, error-prone step that defeats the point of identity-driven onboarding.

**Goal**: when an SSO login carries a group claim that matches a configured regex filter, Infrahub creates the corresponding automatically (zero roles / zero permissions), records the originating provider name on the group, and adds the user as a member. When the filter is active but a login produces no matches, the legacy `sso_user_default_group` fallback still fires so existing customers don't regress.

**Non-goals**:

- No event emission for auto-creation (deferred, US5 / IFC-2592).
- No per-login soft cap enforcement (US2 / IFC-2589).

Refs Jira epic: [IFC-2521](https://opsmill.atlassian.net/browse/IFC-2521).

## Epic coverage

| Phase | Jira | Scope | This PR |
|---|---|---|---|
| P1 Setup | [IFC-2586](https://opsmill.atlassian.net/browse/IFC-2586) | `auth_groups/` package + test dirs | ✅ Complete |
| P2 Foundational | [IFC-2587](https://opsmill.atlassian.net/browse/IFC-2587) | `origin` schema attribute + `auto_create_groups_*` settings | ✅ Complete |
| P3 US1 MVP | [IFC-2588](https://opsmill.atlassian.net/browse/IFC-2588) | Filter + find-or-create + auth hook | ✅ Complete |
| P5 US3 | [IFC-2590](https://opsmill.atlassian.net/browse/IFC-2590) | Honor `sso_user_default_group` when filter yields no matches | ✅ Complete |
| P4 US2 | [IFC-2589](https://opsmill.atlassian.net/browse/IFC-2589) | Skip non-matching + per-login cap | ⚠️ Scoping yes, cap enforcement pending |
| P6 US4 | [IFC-2591](https://opsmill.atlassian.net/browse/IFC-2591) | Audit `origin` provenance + read-only writes | ⚠️ Attribute landed, write-path enforcement pending |
| P7 US5 | [IFC-2592](https://opsmill.atlassian.net/browse/IFC-2592) | Auditable events | ❌ Follow-up PR |
| P8 Polish | [IFC-2593](https://opsmill.atlassian.net/browse/IFC-2593) | Changelog + `sso.mdx` + quickstart | ❌ Follow-up PR |
| Validation | [IFC-2522](https://opsmill.atlassian.net/browse/IFC-2522) | Validation instructions | ❌ Follow-up PR |

## What changed

- New `security.auto_create_groups_filter` setting; matching SSO group claims auto-create a `CoreAccountGroup` and the user joins it.
- New `origin` Text attribute on `CoreAccountGroup` records the provider name (unset on manual / pre-existing groups).
- `sso_user_default_group` still fires when the filter yields no matches; unset filter keeps the legacy exact-name lookup.
- Concurrent first-logins for the same claim produce one group (distributed lock).
- `signin_sso_account` split into focused helpers (no behavior change).

## How to test

```bash
# Backend unit tests
uv run pytest backend/tests/unit/auth_groups/ -v

# Backend component tests (require Neo4j test container)
uv run pytest backend/tests/component/auth_groups/ -v

# Lint + type check
uv run invoke format
uv run invoke lint
uv run mypy backend/infrahub/auth.py backend/infrahub/auth_groups/
```

End-to-end smoke (against a running dev instance):

```bash
# 1. Set the filter in the local env / .env
export INFRAHUB_SECURITY_AUTO_CREATE_GROUPS_FILTER='^LDAP/group/(?P<name>.+)$'

# 2. Restart the API so SecuritySettings re-loads
# 3. Trigger an SSO login carrying sso_groups=["LDAP/group/network-engineering"]
# 4. Query CoreAccountGroup; verify:
#    - name == "network-engineering"
#    - origin == "<your provider's configured name>"
#    - the logging-in user is a member
# 5. Repeat with the same claim from a second user; verify the group is reused, not duplicated.
# 6. With sso_user_default_group also set, log in a user whose claims do NOT match the filter;
#    verify they join the default group.
```

## Impact & rollout

- **Backward compatibility**: fully backward compatible. The feature is off by default (no filter). When off, the legacy SSO group-attachment path is preserved bit-for-bit. The default-group behavior is preserved both when the filter is off and when it is on but yields no matches.
- **Config/env changes**: two new optional fields under `SecuritySettings`. Documented via Pydantic descriptions; `infrahub.toml` documentation update is pending (US3 polish, IFC-2593).

## Checklist

- [x] Tests added/updated (unit + component; concurrent-login coverage included)
- [ ] [Changelog entry](../dev/guidelines/changelog.md) added — pending (IFC-2593)
- [ ] External docs updated (`docs/topics/security/sso.mdx`) — pending (IFC-2593)
- [ ] Internal `.md` docs updated — pending (IFC-2593)

[IFC-2586]: https://opsmill.atlassian.net/browse/IFC-2586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/auth_groups/filter.py
```diff
@@ -0,0 +1,54 @@
+"""A `ClaimFilter` decides which external IdP group claims become local groups, and how to
+name them.
+
+Patterns are tried in declared order. The first match per claim wins; subsequent patterns are
+not consulted for that claim. A pattern with a named capture group `(?P<name>...)` yields the
+captured value as the local group name; otherwise the full claim string is used. Positional
+capture groups are not used for name extraction.
+
+Compiled `re.Pattern` objects are produced by
+`SecuritySettings._compile_auto_create_groups_filter_patterns` at config load time, so by the
```
ajtmccarty, line 10: documentation for a different class belongs on the different class

backend/infrahub/auth_groups/service.py
```diff
@@ -0,0 +1,124 @@
...
+
+if TYPE_CHECKING:
+    from infrahub.database import InfrahubDatabase
+
+    from .filter import ClaimFilter
+
+log = get_logger()
+
+
+class AutoCreatedGroups:
```
ajtmccarty, line 32: I think `AutoCreatedGroupsService` makes it more clear that this object is used for auto-creating groups and is not an object that represent auto-created groups

backend/infrahub/auth_groups/service.py
```diff
@@ -0,0 +1,124 @@
...
+        self._db = db
+        self._account = account
+        self._provider_name = provider_name
+
+    async def assign(self, claims: Iterable[str], claim_filter: ClaimFilter) -> tuple[str, ...]:
+        """Find-or-create groups for `claims` under `claim_filter` and add the account as member.
+
+        Returns the effective names that resulted in a successful membership, in matching order.
+        An empty tuple means auto-creation produced no memberships — the caller may then fall
+        through to the SSO default-group path.
```
ajtmccarty, line 45: docstring shouldn't reference it's caller. the caller should document this if it is necessary

backend/infrahub/auth_groups/service.py
```diff
@@ -0,0 +1,124 @@
...
+        Serialized through the distributed lock registry under the `auto-create-group` namespace.
+        Returns `None` when a non-`CoreAccountGroup` `CoreGroup`-derived row already exists under
+        that name.
+        """
+        existing = await self._lookup_by_name(name)
+        if existing is not None:
+            return self._reuse_or_skip(name, existing)
+
+        lock_key = f"auto-create-group:{name}"
+        async with lock.registry.get(name=lock_key, namespace="auto-create-group"):
```
ajtmccarty, line 81: I think passing `lock` into the constructor as a dependency is more in line with dependency injection b/c this class _does_ depend on `InfrahubLockRegistry` even if it can be imported instead of injected

backend/infrahub/auth_groups/service.py
```diff
@@ -0,0 +1,124 @@
...
+                return self._reuse_or_skip(name, existing)
+
+            group = await Node.init(db=self._db, schema=InfrahubKind.ACCOUNTGROUP)
+            await group.new(db=self._db, name=name, origin=self._provider_name)
+            await group.save(db=self._db)
+            return group
+
+    async def _lookup_by_name(self, name: str) -> Node | None:
+        """Return any `CoreGroup`-derived row named `name`, or None."""
+        results = await NodeManager.query(
```
ajtmccarty, line 93: `NodeManager` is also a dependency of this service. We import and use it everywhere, but it should really be injected in most places. making this change is optional, but you've got me thinking about dependency injection

backend/infrahub/auth_groups/service.py
ajtmccarty, line 1: nice SOLID class

backend/infrahub/auth.py
```diff
@@ -215,12 +217,24 @@ async def create_fresh_access_token(
     return models.AccessTokenResponse(access_token=access_token)
 
 
-async def signin_sso_account(  # noqa: PLR0915
+async def signin_sso_account(
```
ajtmccarty, line 220: probably for a different PR if you want to do it, but this function looks like it should be a class that takes `db` as a dependency

the separate methods are a big improvement that make this much easier to read though 👍

backend/infrahub/auth.py
```diff
@@ -231,97 +245,185 @@ async def signin_sso_account(  # noqa: PLR0915
...
+        db=db,
+        sub=external_identity.sub,
+        provider_name=external_identity.provider_name,
+        protocol=external_identity.protocol,
+        account=account.id,
+    )
+    await identity_node.save(db=db)
+
+
+async def _refresh_label_if_stale(*, db: InfrahubDatabase, account: Node, display_name: str) -> None:
```
ajtmccarty, line 358: to me, "refresh" implies retrieving from the database. this seems more like `set_label_if_stale`

backend/infrahub/config.py
```diff
@@ -767,6 +767,79 @@ class SecuritySettings(BaseSettings):
...
         description="Name of the group to which users authenticated via SSO will belong if not provided by identity provider",
     )
+    auto_create_groups_filter: str | list[str] | None = Field(
+        default=None,
+        description="Regex(es) that decide which external identity-provider group claims become "
+        "Infrahub groups. Accepts one regex or a list; the first matching pattern wins. "
+        "Use a named capture group `(?P<name>...)` to set the group name; otherwise the "
+        "full claim is used. Leave empty to disable auto-creation.",
+    )
+    auto_create_groups_max_per_login: int = Field(
```
ajtmccarty, line 777: should this setting be added in this PR if it is not actually used yet?

backend/infrahub/config.py
```diff
@@ -767,6 +767,79 @@ class SecuritySettings(BaseSettings):
...
+        Raises:
+            ValueError: When a configured regex pattern fails to compile. Pydantic surfaces the
+                error attached to the setting name.
+
+        """
+        if value is None:
+            return value
+
+        raw_patterns: list[str] = [value] if isinstance(value, str) else list(value)
+        raw_patterns = [p for p in raw_patterns if p and p.strip()]
```
ajtmccarty, line 802: should it be `[p.strip() for p...]` to make sure whitespace is removed?

backend/tests/component/auth_groups/test_autocreate_flow.py
```diff
@@ -0,0 +1,428 @@
+"""Functional tests for the auto-creation flow under `signin_sso_account`.
+
+Drives the real `signin_sso_account` (and therefore the real `autocreate_groups_for_login`
+service) against the test database — no mocking of either. Fixtures follow the pattern in
+`backend/tests/component/core/test_signin_sso_account.py` (`db`, `default_branch`,
+`register_core_models_schema`).
```
ajtmccarty, line 6: I don't think you need to specifically say that there is no mocking or reference another test pattern in here

backend/tests/component/auth_groups/test_autocreate_flow.py
```diff
@@ -0,0 +1,428 @@
...
+        register_core_models_schema: SchemaBranch,
+        autocreate_filter_enabled: None,
+        sso_user_default_group_configured: str,
+    ) -> None:
+        """Filter active + empty claims + default configured: user lands in the default group.
+
+        Pre-refactor, the `if not sso_groups: sso_groups = [default]` swap-in lived upstream in
+        oidc.py / oauth2.py. Post-refactor it is owned by `_assign_group_memberships`; this test
+        invokes that function directly with an empty claim list, which is the path the upstream
+        callers now always take.
```
ajtmccarty, line 322: don't think this is necessary

ajtmccarty (approved):

backend/infrahub/auth_groups/filter.py
polmichel: You are right, thanks.
I have noted a pain point regarding code documentation generation in the internal documentation https://opsmill.atlassian.net/wiki/spaces/SD/pages/579829762/AI+code+generation+-+Pain+points

backend/infrahub/auth_groups/service.py
polmichel: Yes !

backend/infrahub/auth_groups/service.py
polmichel: Good idea thanks

backend/infrahub/auth_groups/service.py
polmichel: Correct

backend/infrahub/auth.py
polmichel: Good idea

backend/infrahub/config.py
polmichel: Yes this is better

backend/infrahub/config.py
polmichel: Yes, I would go for this one
`raw_patterns = [stripped for p in raw_patterns if (stripped := p.strip())]`

backend/tests/component/auth_groups/test_autocreate_flow.py
polmichel: Removed

backend/tests/component/auth_groups/test_autocreate_flow.py
polmichel: Removed

backend/infrahub/auth.py
polmichel: Good idea, I will add this as a follow-up task under the same Epic

polmichel: See internal task https://opsmill.atlassian.net/browse/IFC-2616

backend/infrahub/auth_groups/service.py
polmichel: I have created a PR to introduce new rules for AI assistants https://github.com/opsmill/infrahub/pull/9265

PR #9261, opened 2026-05-16
Add exclude_path configuration to lychee.toml

ajtmccarty:
skip specs in the link checker

specs can refer to files that do not exist yet, so I think they should be excluded from the link checker

---
## Summary by cubic
Skip link checking in spec directories to avoid false positives from WIP references. Adds `exclude_path` in `lychee.toml` to exclude `dev/specs` and `specs`.

<sup>Written for commit 72c1e5a3ada9dde7ae11d170dd6bcf2fe95978c9. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9261?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/dev/lychee.toml
+++ b/dev/lychee.toml
@@ -1,1 +1,5 @@
 offline = true
+exclude_path = [
+  "dev/specs",
+  "specs",
+]


ogenstad (commented):
I don't have a strong objection to this, however I'm wondering if there's a point of creating a link to something that doesn't exist. Why not just reference the file without the link if it doesn't yet exist?

ajtmccarty:
> I don't have a strong objection to this, however I'm wondering if there's a point of creating a link to something that doesn't exist. Why not just reference the file without the link if it doesn't yet exist?

very good point. I thought that the lychee was just picking up the names of the files and erroring out, but they were markdown links. un-linking those strings fixed the issue. seems obvious in hindsight

PR #9265, opened 2026-05-16
docs: add code-doc-style rule for AI assistants

polmichel:
# Why

AI assistants occasionally embed brittle references in code documentation — naming callers/other classes that get renamed, or pinning docstrings to ticket/task IDs that go stale once the work closes. A rule file makes the expectation explicit so it loads into every session.

## What changed

- New `dev/rules/code-doc-style.md`
- Two prohibitions: (1) no references to other code (callers, classes, methods) in docstrings/comments; (2) no Jira keys, GitHub issues, requirement IDs, or task IDs in source.

## Checklist

### DIFF
--- a/dev/rules/code-doc-style.md
+++ b/dev/rules/code-doc-style.md
@@ -1,6 +1,11 @@
+---
+paths:
+  - "backend/infrahub/**/*.py"
+---
+
 # Code documentation style
 
-Applies to docstrings, comments, and any inline documentation in source files (Python, TypeScript, TSX).
+Applies to docstrings, comments, and any inline documentation in Python source files.
 
 ## No references to other code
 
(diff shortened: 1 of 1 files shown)


dev/rules/code-doc-style.md
```diff
@@ -0,0 +1,42 @@
+# Code documentation style
```
ogenstad, line 6: I think we have to add some front matter here to highlight the match criteria, example from another file:

```yaml
---
paths:
  - "backend/tests/unit/**/*.py"
  - "backend/tests/component/**/*.py"
  - "backend/tests/functional/**/*.py"
  - "backend/tests/integration/**/*.py"
  - "backend/tests/integration_docker/**/*.py"
---
```

For this I think we could perhaps have `backend/infrahub/**/*.py` ? I'd suggest to start with just Python files and then we can check with the frontend team if they also want to have this rule in place.

polmichel: Done !

ajtmccarty (approved):

ogenstad (approved):

PR #9268, opened 2026-05-17
IFC-2589: cap auto-created groups per login

polmichel:
# Why

Implements [IFC-2589](https://opsmill.atlassian.net/browse/IFC-2589): cap the number of groups auto-created during a single SSO login to prevent unbounded growth from over-broad provider claims.

Part of [IFC-2521]

## What changed

- New `auto_create_groups_max_per_login` setting.
- `auth_groups/service.py` enforces the per-login cap: existing-group memberships do not consume budget; surplus matching claims are silently dropped; login still succeeds.

## How to test

```bash
uv run pytest backend/tests/component/auth_groups/test_autocreate_flow.py::TestPerLoginCap
```

[IFC-2589]: https://opsmill.atlassian.net/browse/IFC-2589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Caps auto-created SSO groups per login to prevent runaway group creation. Implements IFC-2589 with a configurable limit; existing-group reuse is uncapped and extra new groups are skipped while login still succeeds.

- **New Features**
  - Added `auto_create_groups_max_per_login` (default 50) and env `INFRAHUB_SECURITY_AUTO_CREATE_GROUPS_MAX_PER_LOGIN`.
  - Enforced per-login cap in `auth_groups` service: only brand-new groups count; reuse of existing groups is unlimited; surplus new groups are skipped and logged.
  - Refactored `_find_or_create` to return a result with `was_created`; pass the cap from `auth.py` into the `AutoCreatedGroupsService` constructor (policy set per instance).
  - Added component tests for cap behavior (including existing-group reuse); updated docs and `docker-compose.yml`.

<sup>Written for commit 77ea3cf61299a1b1ec619a81204f4abd31a9bdbc. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9268?utm_source=github">Review in cubic</a></sup>


backend/tests/component/auth_groups/test_autocreate_flow.py
```diff
@@ -414,3 +414,105 @@ async def test_filter_disabled_empty_claims_fall_through_to_default_group(
         accounts = await NodeManager.query(db=db, schema=InfrahubKind.ACCOUNT, filters={"name__value": "Mira Legacy"})
         assert len(accounts) == 1
         assert accounts[0].id in members
+
+
+class TestPerLoginCap:
+    """Per-login soft cap on new-group creation."""
```
ogenstad, line 420: I typically try to avoid to use classes for tests when the class itself doesn't provide any additional logic. I think Claude does this to group the tests. However it also introduced an extra level of indentation. My preference would be to use the name of the class and integrated it with the test methods below (in snake case), and then move these to be root level functions instead.

One place where I do see a value for a test class is if we have fixtures that are defined with the class scope so that there can be some reuse, which might be the case here. I haven't checked that.

backend/infrahub/auth_groups/service.py
```diff
@@ -46,56 +50,102 @@ def __init__(
...
+                    continue
 
             await self._add_member(group)
             granted.append(name)
 
+        self.dropped_claims = tuple(dropped)
         return tuple(granted)
 
-    async def _find_or_create(self, name: str) -> CoreAccountGroup | Node | None:
+    async def _find_or_create(self, name: str) -> tuple[CoreAccountGroup | Node | None, bool]:
```
ogenstad: This might be fine since it's a private method, but the return types looks a bit confusing and the caller pretty much needs to read the docstrings or code to figure out what's going on. Since it's private I mainly wanted to highlight it here, as I don't think we should do things like this for public functions/methods.

ogenstad (approved):

backend/tests/component/auth_groups/test_autocreate_flow.py
polmichel: Actually, I asked Claude to do it, and I was about to bring this subject as the backend meeting as an idea for a testing guideline. 

My take on this is that it also helps categorize tests under the same category from a functional perspective which allows us to:
- easily run the test class without having to run the whole file when we only modify the underlying functionality.
- easily refactor this class into a dedicated test file if the number of tests is growing too much

Another way of doing this would be to create separated test files and a test folder, or just leave them all at the root level.

What's your thoughts on this?

backend/infrahub/auth_groups/service.py
polmichel: The behavior is explained in the docstring with various examples. But maybe a dedicated object is better. I will encapsulate this result in an object.

backend/infrahub/auth_groups/service.py
```diff
@@ -46,56 +64,97 @@ def __init__(
...
         self._lock_registry = lock_registry
         self._node_manager = node_manager
+        self.dropped_claims: tuple[str, ...] = ()
 
-    async def assign(self, claims: Iterable[str], claim_filter: ClaimFilter) -> tuple[str, ...]:
+    async def assign(
+        self,
+        claims: Iterable[str],
+        claim_filter: ClaimFilter,
+        max_per_login: int,
```
ajtmccarty: this look like it should be set at the class level during construction

backend/infrahub/auth_groups/service.py
```diff
@@ -46,56 +64,97 @@ def __init__(
...
+                    dropped.append(claim)
+                    continue
+                group = self._reuse_or_skip(name, existing)
+                if group is None:
+                    continue
 
             await self._add_member(group)
             granted.append(name)
 
+        self.dropped_claims = tuple(dropped)
```
ajtmccarty: this looks like it is only written to and never used

backend/tests/component/auth_groups/test_autocreate_flow.py
```diff
@@ -414,3 +414,105 @@ async def test_filter_disabled_empty_claims_fall_through_to_default_group(
...
+        await preexisting.new(db=db, name="cap-test-preexisting")
+        await preexisting.save(db=db)
+
+        identity = _make_identity(sub="sub-cap-002", display_name="Oscar Cap")
+
+        await signin_sso_account(
+            db=db,
+            external_identity=identity,
+            sso_groups=[
+                "LDAP/group/cap-test-preexisting",
```
ajtmccarty, line 491: test should probably verify that Oscar ends up added to this existing group

backend/infrahub/auth_groups/service.py
polmichel: Good idea!

backend/infrahub/auth_groups/service.py
polmichel: Good catch. It is meant to be used in the emission of the cap breached event, but this is not required in this PR, I'll remove it and re add it in the event emission related item

backend/tests/component/auth_groups/test_autocreate_flow.py
polmichel: You are right, I'll add the assertion

backend/infrahub/auth_groups/service.py
```diff
@@ -39,63 +57,99 @@ def __init__(
...
+            if new_creations < self._max_per_login:
+                result = await self._find_or_create(name)
+                group = result.group
+                if group is None:
+                    continue
+                if result.was_created:
+                    new_creations += 1
+            else:
+                # Cap exhausted: only allow reuse of an existing row. A claim that would
+                # require a fresh creation is dropped.
```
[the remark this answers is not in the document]
polmichel: This will be added in the dedicated task using the value

ajtmccarty (approved):

PR #9271, opened 2026-05-18
ci: skip bug-agent-review conclusion job on fork PRs

### LINKED ISSUE #32991: `conclusion` job's App-token mint step fails hard when `GH_AW_APP_ID` isn't available and previous step are canceled
## Summary

The auto-generated `conclusion` job is reachable on workflow runs where the configured GitHub App secrets (`GH_AW_APP_ID` / `GH_AW_APP_PRIVATE_KEY`) are not available to the workflow context, most commonly on `pull_request` events from forks, where GitHub does not pass repo/org secrets to the runner, even after manual approval. The conclusion job's first step (`actions/create-github-app-token`) then fails with:

```
Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty string.
```

This puts a red ❌ on every PR opened from a fork (where repo/org secrets are not exposed to the workflow context). The bug is that the App-token mint step doesn't tolerate the secret being empty.

## Environment

- gh-aw compiler: `v0.71.5`
- Workflow triggers on `pull_request` with `safe-outputs:` configured (status comments, missing-tool, etc.)
- Real failing job: https://github.com/opsmill/infrahub/actions/runs/25793836171/job/76472898570?pr=9242

## Reproduction

1. Author a workflow `.md` with:

   ```yaml
   on:
     pull_request:
       branches: [main]
   safe-outputs:
     github-app:
       client-id: ${{ secrets.GH_AW_APP_ID }}
       private-key: ${{ secrets.GH_AW_APP_PRIVATE_KEY }}
     add-comment:
       max: 3
   ```

2. `gh aw compile`.
3. Trigger the workflow in a context where the App secrets are unavailable — easiest repro is opening a PR from a fork. (Same symptom on any run if the org just hasn't set up `GH_AW_APP_ID` yet.)
4. Observe: when the run reaches the `conclusion` job (after a cancellation, failure, or any non-skipped agent outcome), the `Generate GitHub App token` step fails with the `client-id required` error above.

## Why is this a problem

Specifically: a **first contribution from an external contributor**, where GitHub's first-time-contributor approval gate holds the workflow as `action_required`, and **multiple events queue up** while waiting (e.g. the PR is opened, then a comment is posted, both events trigger the workflow). When a maintainer eventually approves, both queued runs fire at once, `cancel-in-progress: true` on the shared concurrency group cancels one, and the `conclusion` job then trips on `create-github-app-token` because fork PRs aren't passed repo/org secrets.

## Root cause

This section is AI-generated, if you want to not consider this, feel free to ignore this section.

In [`pkg/workflow/notify_comment.go:493-545`](https://github.com/github/gh-aw/blob/main/pkg/workflow/notify_comment.go#L493-L545), the App-token step is added unconditionally:

```yaml
- name: Generate GitHub App token
  id: safe-outputs-app-token
  uses: actions/create-github-app-token@...
  with:
    client-id: ${{ secrets.GH_AW_APP_ID }}
    private-key: ${{ secrets.GH_AW_APP_PRIVATE_KEY }}
    ...
```

There's no `if:` guard on this step, so when `secrets.GH_AW_APP_ID` evaluates to empty string, `actions/create-github-app-token` errors out and fails the whole job.

## Proposed fix

This section is AI-generated, if you want to not consider this, feel free to ignore this section.

Guard the App-token step (and the downstream steps that use its output) so the conclusion job degrades gracefully when secrets aren't available.

1. Add an `if:` to the `Generate GitHub App token` step:

   ```yaml
   - name: Generate GitHub App token
     id: safe-outputs-app-token
     if: ${{ env.GH_AW_APP_ID != '' }}
     env:
       GH_AW_APP_ID: ${{ secrets.GH_AW_APP_ID }}
     uses: actions/create-github-app-token@...
   ```

   (Using `env:` is necessary because `secrets.*` isn't allowed in step-level `if:` directly.)

polmichel:
## Context

A red ❌ has been showing up on some external-contributor PRs on the `Bug reviewer agent / conclusion` check (e.g. PR #9242, [failed job](https://github.com/opsmill/infrahub/actions/runs/25793836171/job/76472898570?pr=9242)). The error — `Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty string.` — comes from a `Generate GitHub App token` step inside the auto-generated `conclusion` job and is a false negative: nothing about the PR itself is wrong.

## When it occurs

Specifically: a **first contribution from an external contributor**, where GitHub's first-time-contributor approval gate holds the workflow as `action_required`, and **multiple events queue up** while waiting (e.g. the PR is opened, then a review bot edits the body — both events trigger the workflow). When a maintainer eventually approves, both queued runs fire at once, `cancel-in-progress: true` on the shared concurrency group cancels one, and the `conclusion` job (which is designed to run via `always()` to report failures) then trips on `create-github-app-token` because fork PRs aren't passed repo/org secrets.

**Workaround for affected PRs:** re-trigger the workflow run (a fresh sync, push, or rerun). Without a second event queuing behind the approval, there is no concurrency cancellation, all upstream jobs cleanly skip, and `conclusion` skips with them.

## What this PR changes

- `.github/workflows/bug-agent-review.lock.yml` — add the cross-repo gate gh-aw already applies to `pre_activation`/`activation` to the `conclusion` job's `if:`. On fork PRs, `conclusion` is now skipped instead of failing.
- `.github/workflows/bug-agent-review.md` — header note that the lock file carries a manual patch that `gh aw compile` will strip.

Stop-gap until upstream gh-aw fixes https://github.com/github/gh-aw/issues/32991.

### DIFF
--- a/.github/workflows/bug-agent-review.lock.yml
+++ b/.github/workflows/bug-agent-review.lock.yml
@@ -1008,8 +1008,16 @@ jobs:
       - agent
       - detection
       - safe_outputs
+    # MANUAL PATCH: cross-repo (fork) PRs can't access GH_AW_APP_ID/PRIVATE_KEY
+    # secrets, which makes the create-github-app-token step in this job fail.
+    # `pre_activation`/`activation` already gate on cross-repo, but gh-aw does
+    # not propagate that gate to the `conclusion` job, so it ran and failed.
+    # The added clause below mirrors that gate. `gh aw compile` will strip
+    # this patch — re-apply it after any re-compile. Remove once
+    # https://github.com/github/gh-aw/issues/32991 is fixed.
     if: >
-      always() && (needs.agent.result != 'skipped' || needs.activation.outputs.lockdown_check_failed == 'true' ||
+      always() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.id == github.repository_id) &&
+      (needs.agent.result != 'skipped' || needs.activation.outputs.lockdown_check_failed == 'true' ||
       needs.activation.outputs.stale_lock_file_failed == 'true')
     runs-on: ubuntu-slim
     permissions:
--- a/.github/workflows/bug-agent-review.md
+++ b/.github/workflows/bug-agent-review.md
@@ -1,4 +1,9 @@
 ---
+# IMPORTANT: bug-agent-review.lock.yml carries a hand patch that adds a
+# cross-repo gate to the `conclusion` job's `if:`. Running `gh aw compile`
+# will strip it — re-apply the patch from the lock file comment after any
+# re-compile. Remove the workaround once
+# https://github.com/github/gh-aw/issues/32991 is fixed.
 description: Review bug pipeline test and fix PRs (triggered on PR open/synchronize)
 on:
   pull_request:


polmichel:
let's wait for the issue to be fixed

PR #9273, opened 2026-05-18
IFC-2025: Fix artifact diff migrated kind

### LINKED ISSUE #7490: bug: artifact diff needs update for migrated kind nodes
### Component

API Server / GraphQL

### Infrahub version

1.4.12

### Current Behavior

the artifact diff can produce unexpected results if the targets of artifacts have had their kind or inheritance updated
at the database-level, when a schema `name`, `namespace`, or `inherit_from` property is updated, we duplicate all nodes using that schema, updating the labels on the new versions and soft-deleting the old versions
the `ArtifactDiffCalculator` and `ArtifactDiffQuery` do not seem to handle this situation quite right

### Expected Behavior

the artifact diff endpoint should only include artifacts that were actually updated on a given branch

### Steps to Reproduce

1. update the `name`, `namespace`, and/or `inherit_from` property of a schema with existing nodes (either on a branch or on main)
2. generate artifacts that target nodes from the schema in step 1 on a branch
3. the artifact diff for the branch might include incorrect artifacts

### Additional Information

this branch includes a unit test that reproduces at least one issue, but there may be more
the problem in this unit test seems to be that an artifact is improperly included in the diff b/c it had one side of its `Artifact`->`CoreArtifactTarget` relationship updated during the node kind migration on the branch
the test should also be updated to include a node kind migration on main

solababs:
# Why

`ArtifactDiffQuery` anchors its target-branch artifact lookup on the Neo4j node identity of `target_node`. After `NodeKindUpdateMigration` runs, two physical Neo4j nodes share the same UUID — the original (soft-deleted on the migration branch) and the new-kind duplicate (active). The query starts from the new-kind node, which has no `IS_RELATED` edges on the target branch, so the OPTIONAL MATCH returns NULL and the artifact appears as `ADDED`.

Closes https://github.com/opsmill/infrahub/issues/7490

## How to test

```bash
uv run pytest backend/tests/component/core/diff/test_calculate_artifact_diff.py -v
```

## Impact & rollout

- **Backward compatibility:** no breaking changes; the fix only suppresses false positives. Operators who were working around the bug by ignoring ADDED entries in schema-rename scenarios will now see correct empty diffs.
- **Performance:** the UUID-based OPTIONAL MATCH (`WHERE tn_target.uuid = target_node.uuid`) relies on the existing `uuid` index on `:Node`; no regression expected.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`backend/changelog/7490.fixed.md`)

---
## Summary by cubic
Fixes artifact diffs for nodes re-keyed by a kind migration so we don’t show false ADDED/UPDATED entries. Aligns with IFC-2025 by matching migrated targets by UUID and only reporting real content changes.

- **Bug Fixes**
  - Cypher now matches the target by UUID (`tn_target.uuid = target_node.uuid`) to find the correct artifact across migrated nodes.
  - Diff calculator skips entries when the artifact storage is unchanged (same `storage_id` and `checksum`).
  - Tests cover branch-side migration, main-side migration, and a genuine content change after migration.

<sup>Written for commit 5d1b2ec7f92747fef2f0d1eb370b7198d62dba40. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9273?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/changelog/7490.fixed.md
+++ b/backend/changelog/7490.fixed.md
@@ -1,0 +1,1 @@
+ArtifactDiffCalculator no longer produces false-positive diff entries for artifacts whose target node was re-keyed by a node kind migration (schema name or namespace change).
--- a/backend/changelog/7490.fixed.md
+++ b/backend/changelog/7490.fixed.md
@@ -1,1 +1,1 @@
-ArtifactDiffCalculator no longer produces false-positive diff entries for artifacts whose target node was re-keyed by a node kind migration (schema name or namespace change).
+Artifact diffs no longer show spurious changes for artifacts whose schema kind was renamed or moved to a different namespace on a branch.
--- a/backend/infrahub/core/diff/query/artifact.py
+++ b/backend/infrahub/core/diff/query/artifact.py
@@ -138,9 +138,10 @@ CALL (target_node, definition_node){
     // get the corresponding artifact on the target branch, if it exists
     // -----------------------
     CALL (target_node, definition_node) {
-        OPTIONAL MATCH path = (target_node)-[trel1:IS_RELATED]-(trel_node:Relationship)-[trel2:IS_RELATED]-
+        OPTIONAL MATCH path = (tn_target:Node)-[trel1:IS_RELATED]-(trel_node:Relationship)-[trel2:IS_RELATED]-
         (target_artifact:%(artifact_kind)s)-[drel1:IS_RELATED]-(drel_node:Relationship)-[drel2:IS_RELATED]-(definition_node)
-        WHERE trel_node.name = $target_rel_identifier
+        WHERE tn_target.uuid = target_node.uuid
+        AND trel_node.name = $target_rel_identifier
         AND drel_node.name = $definition_rel_identifier
         AND all(
             r IN relationships(path)
--- a/backend/tests/component/core/diff/test_calculate_artifact_diff.py
+++ b/backend/tests/component/core/diff/test_calculate_artifact_diff.py
@@ -2,13 +2,17 @@ from uuid import uuid4
 
 import pytest
 
+from infrahub.core import registry
 from infrahub.core.branch.models import Branch
-from infrahub.core.constants import DiffAction, InfrahubKind
+from infrahub.core.constants import DiffAction, InfrahubKind, SchemaPathType
 from infrahub.core.diff.artifacts.calculator import ArtifactDiffCalculator
 from infrahub.core.diff.model.diff import ArtifactTarget, BranchDiffArtifact, BranchDiffArtifactStorage
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager
+from infrahub.core.migrations.schema.node_kind_update import NodeKindUpdateMigration
+from infrahub.core.migrations.shared import MigrationInput
 from infrahub.core.node import Node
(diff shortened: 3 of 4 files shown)


backend/changelog/7490.fixed.md
```diff
@@ -0,0 +1 @@
+ArtifactDiffCalculator no longer produces false-positive diff entries for artifacts whose target node was re-keyed by a node kind migration (schema name or namespace change).
```
ajtmccarty: users aren't going to know what `ArtifactDiffCalculator` is

backend/tests/component/core/diff/test_calculate_artifact_diff.py
```diff
@@ -289,3 +293,186 @@ async def test_calculate_artifact_diff(
...
+    assert not execution_result.errors
+
+    return {
+        "branch": branch,
+        "art_main": art_main,
+        "c1": car_person_data_generic["c1"],
+    }
+
+
+async def test_calculate_artifact_diff_migrated_kind(
```
ajtmccarty, line 364: if you combine this test with the one below into a class or single test case, then the migration only has to run one time instead of two

backend/infrahub/core/diff/query/artifact.py
```diff
@@ -138,9 +138,10 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
     // get the corresponding artifact on the target branch, if it exists
     // -----------------------
     CALL (target_node, definition_node) {
-        OPTIONAL MATCH path = (target_node)-[trel1:IS_RELATED]-(trel_node:Relationship)-[trel2:IS_RELATED]-
+        OPTIONAL MATCH path = (tn_target:Node)-[trel1:IS_RELATED]-(trel_node:Relationship)-[trel2:IS_RELATED]-
         (target_artifact:%(artifact_kind)s)-[drel1:IS_RELATED]-(drel_node:Relationship)-[drel2:IS_RELATED]-(definition_node)
-        WHERE trel_node.name = $target_rel_identifier
+        WHERE tn_target.uuid = target_node.uuid
+        AND trel_node.name = $target_rel_identifier
```
ajtmccarty, line 144: much smaller change than I would have expected this to be. very cool

backend/tests/component/core/diff/test_calculate_artifact_diff.py
```diff
@@ -289,3 +293,186 @@ async def test_calculate_artifact_diff(
...
+
+    assert len(artifact_diffs) == 1
+    assert artifact_diffs[0].action == DiffAction.UPDATED
+    assert artifact_diffs[0].item_new is not None
+    assert artifact_diffs[0].item_new.checksum == "new_checksum_after_genuine_content_change"
+    assert artifact_diffs[0].item_previous is not None
+    assert artifact_diffs[0].item_previous.checksum == "aabbccdd11223344556677889900aabb"
+
+
+async def test_calculate_artifact_diff_main_side_migration(
```
ajtmccarty, line 390: would be worth checking the `with_genuine_change` case for the `main` branch like the tests above too

ajtmccarty (approved):

PR #9276, opened 2026-05-18
Scope schema-constraint validators to diffed nodes

fatih-acar:
## Summary

Stacked on top of `fac-fix-pc-validators` (which scoped property-constraint *selection* to diffed kinds via `d846e73`). This branch narrows the *scan within* each touched kind, and closes one cross-kind correctness gap surfaced along the way.

Four commits:

- `86eafcd6` **scope uniqueness query to diffed node values** — pre-fetch diffed nodes' uniqueness-relevant values via `NodeManager.get_many` and route through the existing `_with_value` Cypher subqueries. No new Cypher. Value-anchored, so collisions with untouched peers still surface.
- `691e6dc1` **scope attribute checkers to diffed node uuids** — plumb `node_uuids` through `SchemaConstraintValidatorRequest` and the shared `SchemaValidatorQuery` base; gate each `MATCH (n:Kind)` with `WHERE $node_uuids IS NULL OR n.uuid IN $node_uuids` across kind, regex, length, min_max, optional, enum, choices, number_pool.
- `90927aa0` **trigger uniqueness re-check on cross-kind peer changes** — when a `rel__peer_attr` path's peer attribute changes, the dependent kind itself isn't in the diff. New `ConstraintValidatorDeterminer.find_cross_kind_uniqueness_dependents` plus a `min_count_required=0` reuse of the existing query resolves affected dependent UUIDs and merges them into the constraint.
- `62696cb7` **scope relationship checkers to diffed node uuids** — same UUID-scoping pattern for `relationship.optional`, `relationship.count`, `relationship.peer`, `relationship.common_parent`.

## Why

On a 400k-node kind, opening a PC that changes one attribute on one node previously triggered kind-wide label scans across every constraint validator that fired for the kind. With this stack the scan is bounded by the diffed node set (or, for the cross-kind case, the set of dependent nodes pointing to diffed peers). Schema-diff-origin constraints (e.g. flipping `optional` or adding `unique`) keep `node_uuids=None` and continue to full-scan — correctness is preserved.

## Test plan

- [x] `uv run pytest backend/tests/component/core/constraint_validators/` — 320 passed, 6 pre-existing skips (verified after each commit).
- [x] `uv run ruff check` and `uv run mypy` — clean on every modified file.
- [ ] End-to-end on the 400k-`DcimInterfaceL2` DB: open a PC that touches one `name`; capture the executed Cypher and `PROFILE` it to confirm the planner anchors on `:AttributeValueIndexed(value)` instead of a kind label scan. Time the validator — target sub-second.
- [ ] End-to-end cross-kind: rename a `DcimDevice` referenced by a `DcimInterface` `rel__peer_attr` uniqueness constraint; verify the PC reports any resulting collision.

## Out of scope (follow-ups)

- `attribute.unique.update` — group-by-value query; needs the same value pre-fetch as the main uniqueness checker.
- Cross-kind `relationship.common_parent.update` — when a peer's parent changes, the start kind isn't in the diff. Same shape as the uniqueness gap closed in `90927aa0`.

---
## Summary by cubic
Scopes all schema-constraint validators to only the nodes changed in a proposed change and fixes cross-kind uniqueness rechecks. This removes kind-wide scans on large datasets while preserving full scans when the schema itself changes.

- **Refactors**
  - Added optional node UUID scoping to validator requests and queries; gated each initial MATCH with WHERE $node_uuids IS NULL OR n.uuid IN $node_uuids across attribute and relationship validators.
  - Uniqueness checker now prefetches constraint-relevant values for diffed nodes and issues value-anchored queries to surface collisions with untouched peers.

- **Bug Fixes**
  - Re-check uniqueness on peer attribute changes across kinds (e.g., rel__peer_attr paths) by resolving affected dependent node UUIDs and merging them into the constraint scope; schema-diff constraints still full-scan.
  - Fixed value checks so falsy-but-set values (0, "", false) are treated as set in value-scoped queries.

<sup>Written for commit 62696cb7e5153ef5ac9e62bb18121d70f9c07101. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9276?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/infrahub/core/diff/model/path.py
+++ b/backend/infrahub/core/diff/model/path.py
@@ -105,6 +105,7 @@ class NodeDiffFieldSummary:
     kind: str
     attribute_names: set[str] = field(default_factory=set)
     relationship_names: set[str] = field(default_factory=set)
+    node_uuids: set[str] = field(default_factory=set)
 
 
 @dataclass
--- a/backend/infrahub/core/models.py
+++ b/backend/infrahub/core/models.py
@@ -156,6 +156,7 @@ class SchemaUpdateConstraintInfo(BaseModel):
     model_config = ConfigDict(extra="forbid")
     path: SchemaPath
     constraint_name: str
+    node_uuids: set[str] | None = None
 
     @property
     def routing_key(self) -> str:
@@ -163,6 +164,13 @@ class SchemaUpdateConstraintInfo(BaseModel):
 
     def __hash__(self) -> int:
         return hash((type(self),) + tuple(self.constraint_name + self.path.get_path()))
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, SchemaUpdateConstraintInfo):
+            return NotImplemented
+        # node_uuids is an optional scoping payload — not part of identity.
+        # Dedup must match __hash__ which only considers constraint_name + path.
+        return self.constraint_name == other.constraint_name and self.path == other.path
 
 
 class SchemaUpdateValidationResult(BaseModel):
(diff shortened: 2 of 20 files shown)


ajtmccarty:
superseded by #10019

PR #9285, opened 2026-05-18
fix merge with delete of migrated kind object

### LINKED ISSUE #9283: bug: merging object with migrated kind that is deleted on branch leaves object un-deleted
### Component

API Server / GraphQL

### Infrahub version

1.9.5

### Current Behavior

if I merge a branch that deletes an object that had it's kind/inheritance migrated on the default branch after my branch was created, then that migrated object remains undeleted.

### Expected Behavior

deleting an object on a branch and merging it to the default branch should delete the object on the default branch, regardless of how the object was changed on the default branch.

### Steps to Reproduce

1. create an object
2. create branch `br1`
3. update the schema of the object from step 1 to change its name, namespace, or inheritance
4. merge branch `br1`
5. the object from step 1, with its new kind/inheritance, will still exist

### Additional Information

_No response_

ajtmccarty:
closes #9283 

Fix a couple of bugs in the merge logic related to object that their kind/inheritance migrated on the default branch and then are deleted on the feature branch. The order of operations to produce the bug(s) has to be

1. create user branch
2. updated name, namepace, or inheritance of schema on the default branch
3. delete an instance of the schema from step 2 on the user branch
4. merge the user branch

the two bugs are 
1. object with their kind/inheritance migrated on the default branch and a delete merged on the user branch were not actually deleted
2. the updated_at/by metadata of an object "deleted" using the reproduction steps was not correctly updated. this is not really visible to the user in any way today once the object is deleted, but it could be important for auditing in the future so should be correct

---
## Summary by cubic
Fixes merge behavior so a delete from a feature branch actually removes an object when its kind/inheritance was migrated on the default branch after the branch fork. Also fixes `updated_at`/`updated_by` so deletes reflect the branch user and time.

- **Bug Fixes**
  - Resolve the active target-branch node (`tgt_n`) for migrated UUIDs and run the delete cascade from it (`IS_PART_OF`, `HAS_ATTRIBUTE`, `IS_RELATED`, `HAS_OWNER`, `HAS_SOURCE`).
  - Refresh node metadata on adds/deletes using `IS_PART_OF.from_user_id`/`to_user_id`, keep `previous_updated_*`, and avoid touching pre-migration vertices.
  - Skip metadata updates for target-branch nodes already deleted by a migration.
  - Add a component test for fork → migrate-on-target → delete-on-branch that verifies full cascade and correct metadata.

<sup>Written for commit ee45105c7a756ddf4275403d37dcfdf5d64e5a7d. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9285?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/tests/component/core/diff/test_target_branch_migration_after_fork.py
+++ b/backend/tests/component/core/diff/test_target_branch_migration_after_fork.py
@@ -1,0 +1,187 @@
+"""Merging a delete from a user branch when the same UUID was kind-migrated on
+the target branch *after* the user branch was forked must propagate the delete
+(and its metadata) to the post-migration Node vertex on the target branch.
+"""
+
+from unittest.mock import AsyncMock
+
+from infrahub.core import registry
+from infrahub.core.branch import Branch
+from infrahub.core.constants import SchemaPathType
+from infrahub.core.diff.coordinator import DiffCoordinator
+from infrahub.core.diff.data_check_synchronizer import DiffDataCheckSynchronizer
+from infrahub.core.diff.merger.merger import DiffMerger
+from infrahub.core.initialization import create_branch
+from infrahub.core.manager import NodeManager
+from infrahub.core.migrations.schema.node_kind_update import NodeKindUpdateMigration
+from infrahub.core.migrations.shared import MigrationInput
+from infrahub.core.node import Node
+from infrahub.core.path import SchemaPath
+from infrahub.core.schema.schema_branch import SchemaBranch
+from infrahub.core.timestamp import Timestamp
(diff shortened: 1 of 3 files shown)


backend/tests/component/core/diff/test_target_branch_migration_after_fork.py
polmichel, line 1: Nice test! Not really feedback, more of a thought: introducing additional test packages to separate the test files that are starting to pile up could be interesting, I think.

polmichel (approved):

PR #9287, opened 2026-05-19
add schema-based planning logic for graph traversal

ajtmccarty:
# Why

Phase 2D of schema-based graph traversal planning: introduce the synchronous `SchemaPlanner` that enumerates schema-derived routes between kinds, prunes them by user filters and view permissions, and returns a deterministic `Plan`. 

Non-goals:

- Cypher rendering from a `Plan` (Phase 2F).
- Wiring the planner into `PathTraversalQuery` / `ReachableNodesQuery` (Phases 3 and 4).

### Behavioral changes

- New internal module `backend/infrahub/graph_traversal/planning/planner.py` exposing `SchemaPlanner.plan(...)` — fully synchronous, deterministic, schema-only enumeration. Not yet called by any production code path.
- `UserFilters` gains an `allow_schema_revisits: bool = False` field. `from_graphql_input(...)` reads it tolerantly so the field is optional on the GraphQL side.
- No user-visible GraphQL change. The `allow_schema_revisits` input field will be added to the GraphQL `PathTraversalInput` / `ReachableNodesInput` models and wired in later.

### Implementation notes

- The planner is constructed with `(schema_branch, branch, permission_resolver)`. No async lifecycle: the caller is responsible for awaiting `PermissionLoader.load(...)` and passing in a populated `PermissionResolver`. This keeps `plan()` cheap to invoke repeatedly and statically checkable.
- Route enumeration is iterative BFS. Each frontier entry carries `visited_intermediates: frozenset[str]` containing every kind on its path *except the source*. When `allow_schema_revisits=False`, any peer whose kind is already in that set is pruned before the hop is materialized, so the BFS never explores cyclic subtrees. The source-as-terminal exception is enforced separately so `A→A` (a 1-hop self-relationship where source == terminal) is still emitted, while `A→B→A` is allowed only when `A` is also the terminal kind.
- Permission gating uses `KindPermissionCache` (added in p1), memoizing per-kind `can_view` lookups so re-querying the source across routes is free.
- Surviving routes are sorted lexicographically by `(length, kinds, identifiers, directions)` for determinism.

### What stayed the same

- No changes to `PathTraversalQuery` or `ReachableNodesQuery` — both still use their existing raw-Cypher implementations.
- No changes to the GraphQL schema (`PathTraversalInput`, `ReachableNodesInput`, returned types).
- No DB migrations, no new dependencies, no config or env additions.

### Suggested review order

1. `backend/infrahub/graph_traversal/planning/planner.py` — start at `plan()`, then `_enumerate_routes` / `_step`, then the helper methods.
2. `backend/tests/unit/graph_traversal/planning/test_planner.py` — the test classes mirror the planner's responsibilities (route enumeration, revisit policy, generic expansion, direction preservation, user filters, determinism, terminal predicates, kind filter).
3. `backend/infrahub/graph_traversal/planning/models.py` — small `UserFilters.allow_schema_revisits` addition.

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../dev/guidelines/changelog.md) added (`uv run towncrier create ...`) — deferred to the User Story phases where the change becomes user-visible.
- [ ] External docs updated — N/A, no user-facing change in this PR.
- [x] Internal .md docs updated (`contracts/planner.md`, `data-model-schema-planning.md`, `tasks-schema-planning.md`).
- [x] I have reviewed AI generated content.

---
## Summary by cubic
Adds a synchronous, schema-driven path planner with deterministic route enumeration and per-hop pruning, plus a new `infrahub.graph_traversal` package for traversal code. Injects a loaded `PermissionResolver` at construction; no production wiring yet. Supports INFP-1991 Phase 2D.

- New Features
  - `SchemaPlanner.plan(...)`: BFS over schema with depth bounds (1..20), deterministic sorting, and per-hop pruning by `UserFilters`, terminal predicates (`TerminalById`, `TerminalByKinds`), and kind view permissions via `KindPermissionCache`.
  - Revisit policy off by default (`allow_schema_revisits=False`), including blocking immediate self-loop hops; can be toggled per request.
  - `UserFilters`: adds `allow_schema_revisits` and defaults `excluded_namespaces` via `DEFAULT_EXCLUDED_NAMESPACES` moved to `planning.constants` (`Core`, `Internal`, `Builtin`, `Lineage`, `Profile`, `Template`).
  - Plan output simplified to only `routes`; pruned-route diagnostics removed.
  - Fully synchronous lifecycle: caller provides `schema_branch`, `branch`, and a loaded `PermissionResolver`. Tests and internal docs added; queries and GraphQL inputs unchanged.

<sup>Written for commit a8336bc2451f8958ba3b663cbdc07e743f25a6e4. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9287?utm_source=github">Review in cubic</a></sup>


backend/infrahub/graph_traversal/planning/planner.py
```diff
@@ -0,0 +1,365 @@
...
+            raise ValueError(f"source_kind {source_kind!r} not in schema")
+        if isinstance(terminal_predicate, TerminalById):
+            if not self._kind_exists(terminal_predicate.kind):
+                raise ValueError(f"terminal kind {terminal_predicate.kind!r} not in schema")
+        else:
+            for k in terminal_predicate.kinds:
+                if not self._kind_exists(k):
+                    raise ValueError(f"terminal kind {k!r} not in schema")
+
+        if not self._permission_cache.can_view(source_kind):
```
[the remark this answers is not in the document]
ajtmccarty: these are enforced within `_step()` now

backend/infrahub/graph_traversal/planning/planner.py
```diff
@@ -0,0 +1,365 @@
...
+}
+
+
+def _hop_direction_from_schema(direction: RelationshipDirection) -> HopDirection:
+    return _DIRECTION_FROM_SCHEMA[direction]
+
+
+def _invert_direction(direction: HopDirection) -> HopDirection:
+    """Flip a Hop's direction for reverse traversal."""
+    return _DIRECTION_INVERTED[direction]
```
ogenstad: For these two helper functions while it's a small thing I think it might be better if we skip the dictionaries and instead have a match statement on `direction` along with a return for each case ending with an assert_never. That way we'd get type safety for anything like this if we'd ever want to add some different type of `HopDirection` without having to worry that these dictionaries has to be updated to prevent a `KeyError`.

backend/tests/unit/graph_traversal/planning/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+from tests.helpers.graph_traversal.builders import build_schema_branch
+
+if TYPE_CHECKING:
+    from infrahub.core.schema.schema_branch import SchemaBranch
+
+
+def _node(
+    name: str, *, namespace: str = "Testing", relationships: list[RelationshipSchema] | None = None
+) -> NodeSchema:
+    return NodeSchema(name=name, namespace=namespace, relationships=relationships or [])
```
ogenstad, line 26: I'm not sure these functions are helpful or not, they make it a bit less verbose within the tests but also provides a layer of abstraction that might just hide what we're doing within the tests. Could be more useful when there's a larger list of parameters that get set.

ogenstad (approved):

backend/infrahub/graph_traversal/planning/planner.py
ajtmccarty: yes good point. I totally forgot about this better structure. updated

backend/tests/unit/graph_traversal/planning/conftest.py
ajtmccarty: I'm not completely convinced on the shape of these, but I do think that it makes reading the schema definitions below and in some other test files easier

PR #9290, opened 2026-05-19
Update NodeManager.query return type

ogenstad:
# Why

NodeManager.query was returning `list[Any]` when we didn't use SchemaProtocol and as such it messed up the typing as everything is permitted on an `Any`

## What changed

* Swap the return type so that we instead return `list[Node]` if a SchemaProtocol isn't used.
* Swap to protocols in a few places
* The change highlighted some other issues for instance the enum, i haven't checked but would assume that we have  either these inline type ignores in other places, alternatively that we have the entire file exclude with regards th `attr-defined` with mypy. Will track the enum issue separately
* Some new type-ignores in the test

## Notes

* Basically the errors that surfaced were there before we just ignored them all because we were returning `Any`.
* There will be some follow up work to correct these issues which will allow us to improve typing overall

### DIFF
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -102,7 +102,7 @@ class NodeManager:
         partial_match: bool = ...,
         branch_agnostic: bool = ...,
         order: OrderModel | None = ...,
-    ) -> list[Any]: ...
+    ) -> list[Node]: ...
 
     @overload
     @classmethod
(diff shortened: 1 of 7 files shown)


backend/infrahub/core/manager.py
```diff
@@ -139,7 +139,7 @@ async def query(
         partial_match: bool = False,
         branch_agnostic: bool = False,
         order: OrderModel | None = None,
-    ) -> list[Any]:
+    ) -> list[Node] | list[SchemaProtocol]:
```
ogenstad, line 142: Since we were previously returning `Any` here we couldn't trust the returned value if a `SchemaProtocol` wasn't used and as such all type checking was ignored.

ajtmccarty (approved):

PR #9295, opened 2026-05-19
Stable to develop

ajtmccarty:
merge stable to develop with a few extra updates see the second and later commits for the individual updates

- clean up the `NodeDuplicateQuery` to set the `updated_by/at` metadata on Node vertices being deleted on the default/global branch. also generally update to be in the more modern style
- updates to the `DiffMergeMetadataQuery` to cover a fix from `stable`. The diff merge logic has been completely rewritten between `stable` and `develop`, so `develop` required a different update to actually be fixed.

## Summary by cubic
Fixes merge behavior so deletes from a branch propagate to the correct node when the target branch migrated its kind/inheritance after the fork. The post‑migration vertex (and its child edges) is closed at merge time and node metadata is updated to the merge user/time.

- **Bug Fixes**
  - Merge query now handles post‑fork target‑branch migrations: includes migrated siblings, closes `IS_PART_OF` and child edges at `$at`, and refreshes `updated_at/updated_by` without clobbering `previous_updated_*`.
  - Includes fields added/closed at `$at` and skips vertices already deleted on the target branch.
  - `NodeMetadataDefaultBranchQuery` accepts `allowed_kinds` to filter returned nodes.
  - Added tests for delete propagation to post‑migration vertices and updated validators; changelog entry added.

- **Refactors**
  - Simplified `NodeDuplicateQuery`: streamlined subqueries and consistently close existing edges on branch/global while stamping `to_user_id`; set metadata when duplicating on default/global.

<sup>Written for commit b941db0902b5c5bd42bef550c93aae3b519dc7d3. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9295?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/infrahub/core/diff/query/merge.py
+++ b/backend/infrahub/core/diff/query/merge.py
@@ -78,16 +78,18 @@ WHERE NOT EXISTS {
     WHERE migrated_out.from < $at AND migrated_out.to IS NULL
 }
 CALL (n) {
-    MATCH (n)-[e:IS_PART_OF]->(:Root)
+    // get add or delete IS_PART_OF edge if it exists
+    OPTIONAL MATCH (n)-[e:IS_PART_OF]->(:Root)
     WHERE e.branch IN [$target_branch, $global_branch]
     AND (
         // pre-fork edge, edge created by the merge, or edge closed by the merge
         (e.branch = $target_branch AND (e.from <= $branched_from OR e.from = $at OR e.to = $at))
         OR (e.branch = $global_branch AND e.from <= $at)
     )
-    RETURN e AS is_part_of_e
+    WITH e
     ORDER BY e.from DESC, e.status ASC
     LIMIT 1
+    RETURN e AS is_part_of_e
 }
 // --------------------
 // Node-level metadata refresh for added and deleted Node vertices
@@ -118,6 +120,10 @@ WHERE e.branch IN [$source_branch, $target_branch, $global_branch]
 AND (
     (e.branch = $target_branch AND (e.from <= $branched_from OR e.from = $at OR e.to = $at))
     OR (e.branch IN [$source_branch, $global_branch] AND e.from <= $at)
+    // include any fields updated as part of this merge to cover the case where
+    // a Node had its kind migrated on the default branch after the user branch forked
+    OR exists((field)-[{branch: $target_branch, from: $at}]-())
+    OR exists((field)-[{branch: $target_branch, to: $at}]-())
 )
 
 // --------------------
--- a/backend/infrahub/core/diff/query/merge.py
+++ b/backend/infrahub/core/diff/query/merge.py
@@ -59,37 +59,64 @@ class DiffMergeMetadataQuery(Query):
         }
         query = """
 // --------------------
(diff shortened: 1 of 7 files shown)


backend/infrahub/core/migrations/query/node_duplicate.py
```diff
@@ -166,63 +165,68 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
...
             MATCH (root:Root)<-[r:IS_PART_OF]-(node)
             WHERE %(branch_filter)s
-            RETURN node as n1, r as r1
+            RETURN node as active_node, r.status = "active" AS is_active
             ORDER BY r.branch_level DESC, r.from DESC
             LIMIT 1
         }
-        WITH n1 as active_node, r1 as rb
-        WHERE rb.status = "active"
+        WITH active_node
```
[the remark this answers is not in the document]
ajtmccarty: `is_active` is included in this context and then removed following the `WITH...WHERE`

backend/infrahub/core/diff/query/merge.py
```diff
@@ -59,38 +59,71 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
 MATCH (n)-[e:HAS_ATTRIBUTE|IS_RELATED]-(field:Attribute|Relationship)
 WHERE e.branch IN [$source_branch, $target_branch, $global_branch]
 AND (
-    (e.branch = $target_branch AND e.from <= $branched_from)
+    (e.branch = $target_branch AND (e.from <= $branched_from OR e.from = $at OR e.to = $at))
     OR (e.branch IN [$source_branch, $global_branch] AND e.from <= $at)
+    // include any fields updated as part of this merge to cover the case where
+    // a Node had its kind migrated on the default branch after the user branch forked
+    OR exists((field)-[{branch: $target_branch, from: $at}]-())
+    OR exists((field)-[{branch: $target_branch, to: $at}]-())
```
polmichel, line 126: This is the part which is different right?

polmichel (approved):

backend/infrahub/core/diff/query/merge.py
polmichel: If I get it correctly, one of the reason this is there is for instance is this part of the `infrahub.core.diff.query.bulk_merge.BulkMergeRelationshipEdgesQuery  `  
```
 CREATE (tgt_n)<-[:IS_RELATED {
        branch: $target_branch, branch_level: $branch_level, from: $at,
        status: status, hierarchy: hierarchy, from_user_id: from_user_id
    }]-(rel)
```

ajtmccarty: yes that's right. there's a special case for this scenario
1. create branch
2. update schema kind/inheritance on default branch
3. update/delete object that is an instance of the schema from step 2 on the branch
4. merge the branch

during the merge, at the database level, we end up updating an object (from step 3) on the default branch that was created _after_ the branch forked, which is normally impossible. but we can pretty safely say that any update on the default branch at the exact microsecond of the merge was part of the merge

PR #9314, opened 2026-05-21
simplify schema-based route planner

ajtmccarty:
the schema-based route planning was producing a more complex data structure than was actually usable by the first version of the faster cypher query

the actual cypher query (next PR) will be mostly filtered by all possible starting kinds, all possible relationship names, and all possible ending kinds for a given step along the path from the `source` object to the `destination` object/kind(s). so it really only needs a list of all possible step start kinds, all possible step end kinds, and all relationship identifiers that could link one of the start kinds to one of the end kinds.

the previous version of the schema-based planner was producing a list of every possible route from the `source` kind to the `destination` kind where a `route` was a list of `hop`s and each `hop` was 1 start kind, 1 relationship identifier, and 1 end kind. so it was just much more information than the cypher query will need. 

also, the cypher query was not capable of enforcing the `allow_schema_revisit` boolean in its current state, so that has also been removed

---
## Summary by cubic
Simplified the schema-based planner to emit a compact per-hop adjacency map consumed by a single quantified-path-pattern Cypher, and removed unsupported options. Aligns with INFP-1991 by reducing planner output to only what the query needs for both default and user branches.

- **Refactors**
  - Replaced `Route`/`Hop`/`HopDirection` with `Plan.adjacency: Mapping[str, Mapping[str, frozenset[str]]]` (start kind → relationship name → end kinds).
  - Planner builds adjacency via bounded BFS (source-forward ∩ terminal-backward) with permission filtering; preserves `MIN_DEPTH`/`MAX_DEPTH`.
  - Removed `allow_schema_revisit` (not enforced by the Cypher).
  - Consolidated `render_plan_to_cypher` to a single QPP-based implementation (no branch dispatch); updated tests and specs to the new contract.

- **Migration**
  - Replace any `Plan.routes`/`Route`/`Hop` usage with reads from `Plan.adjacency`.
  - Drop any references to `allow_schema_revisit`.

<sup>Written for commit 1f20a90f7eb59ab6a58a3ecd653d87f54c44cc99. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9314?utm_source=github">Review in cubic</a></sup>


backend/infrahub/graph_traversal/planning/planner.py
```diff
@@ -188,168 +146,131 @@ def plan(
...
-          no exemption. If the peer (or this hop's identifier) violates one
-          of these, the entire subtree is dropped.
-        - ``kind_filter``: source and terminal are exempt. If the peer is not
-          in ``kind_filter`` AND does not match the terminal predicate, drop
-          the subtree. If it does match the terminal, emit the route but do
-          not extend.
+    ) -> bool:
+        """Apply per-hop filters; record the hop if it survives.
+
+        Returns ``True`` iff the peer is extendable (i.e. allowed as an
```
ogenstad, line 254: Typo iff -> if

ogenstad (approved):

backend/infrahub/graph_traversal/planning/planner.py
ajtmccarty: I think this is actually `if and only if`. or less likely but it could be Identify Friend Foe

PR #9316, opened 2026-05-21
cypher query for schema-based planning

ajtmccarty:
should follow #9314 

adds the actual cypher building and updates `PathTraversalQuery` to use it. also updates how `PathTraveralQuery` is built and called from the graphql resolver

more tests for the cypher query and graphql endpoint will be forthcoming

---
## Summary by cubic
Implements a schema-planned, single-QPP Cypher path query and wires it into the GraphQL resolver. Enforces schema routes, handles branch/version races, and returns a smaller, structured payload with a clear start node and hops for INFP-1991.

- **New Features**
  - Added `backend/infrahub/graph_traversal/_cypher.py` to render one QPP Cypher from a non-empty `Plan`, with default/user-branch handling and user-branch deletion checks.
  - Cypher now projects `start_node_uuid`, `start_node_kind`, `hops` as `[ {relationship_identifier, uuid, kind} ... ]`, and `depth` instead of returning a Neo4j `Path`.
  - Introduced `backend/infrahub/graph_traversal/results.py` with `PathData`, `PathHopData`, `PathNodeData` (`{uuid, kind}`).
  - Added `backend/infrahub/graph_traversal/_extract.py` to build `PathData` from the projected columns.
  - Added `QueryResult.get_as_list_of_type(...)` to parse Cypher-projected lists.

- **Refactors**
  - `PathTraversalQuery` now requires a non-empty `Plan`, accepts `default_branch_name`, renders via the new Cypher builder, and parses rows via `_extract.py`.
  - GraphQL `path` resolver builds a `Plan` via `SchemaPlanner`, short-circuits on empty plans, uses a wildcard-allow `PermissionResolver`, and defaults `excluded_namespaces` to `planning.DEFAULT_EXCLUDED_NAMESPACES`.
  - `ReachableNodesQuery` uses shared result types with a local extractor; behavior unchanged.
  - Tests and specs updated to the adjacency-based plan, new projection shape, and `default_branch_name` plumbing.

<sup>Written for commit 867dc16b217ef03e6812c87d212cef671952f557. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9316?utm_source=github">Review in cubic</a></sup>


backend/infrahub/graph_traversal/_cypher.py
```diff
@@ -0,0 +1,191 @@
...
+    allowed_path_maps: dict[str, dict[str, list[str]]] = {
+        start_kind: {rel_name: sorted(end_kinds) for rel_name, end_kinds in rels.items()}
+        for start_kind, rels in plan.adjacency.items()
+    }
+    all_rel_names = sorted({rn for rels in allowed_path_maps.values() for rn in rels})
+    start_kinds = sorted(allowed_path_maps)
+    end_kinds = sorted({ek for rels in allowed_path_maps.values() for eks in rels.values() for ek in eks})
+
+    if branch.is_default:
+        valid_branches = [registry.default_branch, GLOBAL_BRANCH_NAME]
```
ogenstad: Not a blocker, but I think we should try to avoid introducing direct access to the registry in new functions and classes. It feels like the registry is so deeply nested all over the place and it would be nice with a solid refactoring of that one as well and I have a feeling that we'll need to rip all of these out later.

backend/infrahub/graph_traversal/path.py
```diff
@@ -1,214 +1,82 @@
...
 
     def get_paths(self) -> list[PathData]:
         paths: list[PathData] = []
         for result in self.get_results():
-            path_obj = result.get_path(label="path")
-            if path_obj is None:
+            # ``Query.get`` is typed for node/relationship returns; the QPP
+            # projects a list[dict] (per the ``path_data`` projection in
+            # ``_cypher.py``), so cast at the boundary.
+            raw = cast("list[dict[str, Any]]", result.get(label="path_data"))
```
ogenstad: Could we have a method on the result object that returns a list of dicts instead of casting it here? I.e. similar to `.get_as_str()`

backend/infrahub/graphql/queries/path.py
```diff
@@ -238,25 +246,44 @@ async def path_traversal_resolver(
...
+            terminal_predicate=TerminalById(node_id=destination_id, kind=destination_node.get_kind()),
+            max_depth=int(max_depth),
+            user_filters=user_filters,
+        )
+    except ValueError as exc:
+        raise GraphQLError(str(exc)) from exc
+
+    if plan.is_empty:
+        # No schema route survives planning — return an empty result without
+        # executing any Cypher (FR-004).
```
ogenstad: Mentions of FR-004 will be confusing later.

ogenstad (approved):

backend/infrahub/graph_traversal/_cypher.py
ajtmccarty: completely agree. removed

backend/infrahub/graph_traversal/path.py
ajtmccarty: added a new `get_as_list_of_type` that can accept a dataclass or TypedDict or other dict-like thing

backend/infrahub/graphql/queries/path.py
ajtmccarty: removed

PR #9317, opened 2026-05-21
fix: emit one log line per user check log entry

### LINKED ISSUE #8224: bug: Calling log_error from a check emits one character per line
### Component

Python SDK, Git Integration

### Infrahub version

1.7.1 (2413b7c4c8495908f78bba731f4afd1fbf6d972c)

### Current Behavior

The following code in a check:

```python
    class MyCheck(InfrahubCheck):
        def validate(self, data: dict) -> None:
            [...redacted...]

            message = (
                f"Duplicate serial '{serial}' for manufacturer '{manufacturer_name}'. "
                f"Assets: '{first['tag']}' and '{asset_tag}'."
            )

            self.log_error(message=message, object_id=asset_id, object_type="DcimDeviceAsset")

```

Results in this (truncated, as it goes for quite a while) log message:

Image

It appears to be caused by InfrahubCheck.log_entries @ https://github.com/opsmill/infrahub-sdk-python/blob/d116aa39c2f739d1fc96af26679bc2e0087b43d9/infrahub_sdk/checks.py#L120-L131

...which returns a string that [backend.infrahub.git.tasks:run_user_check](https://github.com/opsmill/infrahub/blob/stable/backend/infrahub/git/tasks.py#L959-L961) then iterates over, emitting one character per line.

### Expected Behavior

Emit one log entry per line.

### Steps to Reproduce

1. Create a Check which emits a log entry
2. Trigger the check; in my testing, I submitted a proposed change (Draft)

### Additional Information

_No response_

DharmaBytesX:
## Summary

- Fixes #8224
- `run_user_check` iterated over `check_run.log_entries`, which is a single formatted string rather than the structured list of log records. Iterating a string in Python yields one character at a time, so the Prefect flow logger received one warning event per character instead of one per log entry.
- The loop now iterates `check_run.logs` (the structured `list[dict[str, Any]]`) and renders each entry as a single line via a new `format_check_log_entry` helper. The string assignment to `check.message.value` (stored in Neo4j and shown in the check message panel) is unchanged.

## Scope

Backend only. The SDK property `log_entries` is left intact: even with its current naming, the calling code is wrong to iterate it as a list. Any SDK clarification (return type annotation, naming) is orthogonal and can be addressed separately if desired.

## Test plan

- [x] New unit tests in `backend/tests/unit/git/test_tasks.py` covering the formatter on five shapes of log entries, including a regression assertion that no `\n` is produced per record.
- [x] `uv run pytest backend/tests/unit/git/` runs 13/13 green locally.
- [x] `uv run invoke format` and `uv run invoke lint` (ruff, ty, mypy) all pass.
- [x] Manual reproduction on `stable` confirmed the bug end to end (a custom check repo plus a proposed change produced one `prefect.flow_runs warning` per character in `docker logs infrahub-task-worker-1`).

---
## Summary by cubic
Fixes user-check logs emitting one character per line; each check log record now logs as a single warning line. Fixes #8224.

- **Bug Fixes**
  - Iterate `check_run.logs` (structured list) instead of `check_run.log_entries` (string) and use `format_check_log_entry` to render one line per entry.
  - Added unit tests for the formatter, including a regression ensuring no newline per record.

<sup>Written for commit f8570d361401bb94a2882c5430799490c214ccd6. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9317?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/infrahub/git/tasks.py
+++ b/backend/infrahub/git/tasks.py
@@ -64,6 +64,21 @@ from .repository import InfrahubReadOnlyRepository, InfrahubRepository, get_init
 from .utils import fetch_artifact_definition_targets, fetch_check_definition_targets, get_repositories_commit_per_branch
 
 
+def format_check_log_entry(entry: dict[str, Any]) -> str:
+    """Render one user-check log record as a single line for the Prefect flow logger."""
+    parts = [f"[{entry['level']}] {entry['message']}"]
+    object_type = entry.get("object_type")
+    object_id = entry.get("object_id")
+    if object_type or object_id:
+        details = []
+        if object_type:
+            details.append(f"object_type={object_type}")
+        if object_id:
+            details.append(f"object_id={object_id}")
+        parts.append(f"({', '.join(details)})")
+    return " ".join(parts)
+
+
 @flow(
     name="git-repository-add-read-write",
     flow_run_name="Adding repository {model.repository_name} in branch {model.infrahub_branch_name}",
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
LGTM, thanks!

DharmaBytesX:
Thx! Glad to contribute

PR #9325, opened 2026-05-21
Auto-creation groups feature related events

polmichel:
# Why

Implements of [IFC-2592](https://opsmill.atlassian.net/browse/IFC-2592) (Auto-creation of groups): the existing auto-create flow has no audit trail, which blocks investigations of "who was added to which group via which IdP claim, when, by which provider". This PR emits a structured event per auto-create outcome (success / rejected-claim / per-login cap breach) so the activity log preserves provenance for compliance.

Closes [IFC-2592](https://opsmill.atlassian.net/browse/IFC-2592)

## What changed

**Behavioral:**

- Three new audit event types emitted during the auto-create flow:
  - `infrahub.group.auto_create.created` — exactly one per successful new group
  - `infrahub.group.auto_create.rejected_claim` — when a claim's effective name fails identifier validation
  - `infrahub.group.auto_create.cap_breach` — at most one per login when the per-login cap is hit
- **No event** is emitted when a login reuses an existing auto-created group.

**What stayed the same:** API contract unchanged. No schema changes. No new config or env vars. Auto-create events fire only when `signin_sso_account` is called with `event_service=...` (production OAuth2/OIDC paths already wire it).

## How to test

```bash
uv run pytest \
  backend/tests/component/auth/auth_groups/test_event_created.py \
  backend/tests/component/auth/auth_groups/test_event_rejected.py \
  backend/tests/component/auth/auth_groups/test_event_cap_breach.py
```

The three new files cover the three outcomes end-to-end (matching claim → event; bad claim → rejected event with verbatim/truncated value; cap exceeded → single cap-breach event carrying the dropped claims).

## Checklist

- [x] Tests added/updated — 3 new component test files
- [x] External docs updated — `docs/docs/reference/infrahub-events/group.mdx` regenerated
- [x] Internal .md docs updated — `dev/specs/infp-556-auto-create-groups/tasks.md` Phase 7 status flipped to Complete

[IFC-2592]: https://opsmill.atlassian.net/browse/IFC-2592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/auth/auth_groups/service.py
```diff
@@ -187,6 +202,6 @@ async def _add_member(self, group: CoreAccountGroup | Node) -> bool:
         members = await members_rel.get_peers(db=self._db, branch_agnostic=True, peer_type=CoreAccount)
         if self._account.id in members:
             return True
-        await members_rel.add(db=self._db, data=self._account)
+        await members_rel.add(db=self._db, data={"id": self._account.id})
```
polmichel, line 203: This is not ideal, but I got a typing warning otherwise as the object is a `CoreAccount` and not a `Node`

backend/infrahub/auth/__init__.py
```diff
@@ -16,14 +16,12 @@
 )
 from infrahub.auth.session import AccountSession, AnonymousSession
 from infrahub.auth.types import AuthType
-from infrahub.external_protocols import ExternalAuthProtocol
```
polmichel, line 19: This was a missing element from the previous refactoring

backend/infrahub/graphql/types/event.py
polmichel, line 1: New "types" for graphql queries

backend/tests/unit/event/test_group_action.py
```diff
@@ -0,0 +1,135 @@
+from __future__ import annotations
```
polmichel, line 1: Unit tests that could also work like a "specification" of what the serialized events look like

backend/infrahub/api/oauth2.py
```diff
@@ -171,7 +171,9 @@ async def token(
     with trace.get_tracer(__name__).start_as_current_span("signin_sso_account") as span:
         span.set_attribute("account_name", ujson.dumps(userinfo_response.json()))
         span.set_attribute("sso_groups", sso_groups)
-        auth_result = await signin_sso_account(db=db, external_identity=external_identity, sso_groups=sso_groups)
+        auth_result = await signin_sso_account(
+            db=db, external_identity=external_identity, sso_groups=sso_groups, event_service=service.event
```
ogenstad, line 175: I don't know what the best approach is here, we do have the `get_event_service()` as well that could be used within `singin_sso_account()` instead, though in a way I like the fact that we can send in an argument that could be a test adapter from the start instead of having to modify the singleton where service adapters are registered. You don't need to change anything. It's more to highlight.

backend/infrahub/auth/auth_groups/emitter.py
```diff
@@ -0,0 +1,131 @@
...
+                dropped_count=len(dropped_claims),
+            )
+        )
+
+    async def _send(self, event: InfrahubEvent) -> None:
+        """Send an event, swallowing send failures so they cannot abort the login."""
+        try:
+            await self._deps.event_service.send(event=event)
+        except Exception:
+            log.exception("auth_groups.event_emission_failed", event_name=event.event_name)
```
ogenstad, line 114: Regarding your comment in my LDAP PR. I think that once this is in place, I'll probably update my code to point to these functions. I'm not certain that we'll ever hit this line and the Exception is a bit broad, but we can keep it for now and validate later.

backend/infrahub/auth/auth_groups/emitter.py
```diff
@@ -0,0 +1,131 @@
...
+    return value[:MAX_CLAIM_VALUE_LENGTH]
+
+
+@dataclass(frozen=True, slots=True)
+class EmissionDeps:
+    """Runtime dependencies required to emit auto-create events."""
+
+    event_service: InfrahubEventService
+    event_meta_factory: Callable[[], EventMeta]
+    protocol: ExternalAuthProtocol
```
ajtmccarty: I'm not sure this class is worth it. it looks like it's just adding a very thin layer between `AutoCreateEventEmitter`'s constructor and its dependencies. not a strong objection, just doesn't seem worth it to me

backend/infrahub/auth/auth_groups/emitter.py
```diff
@@ -0,0 +1,131 @@
...
+
+    async def _send(self, event: InfrahubEvent) -> None:
+        """Send an event, swallowing send failures so they cannot abort the login."""
+        try:
+            await self._deps.event_service.send(event=event)
+        except Exception:
+            log.exception("auth_groups.event_emission_failed", event_name=event.event_name)
+
+
+class _NoopAutoCreateEventEmitter(AutoCreateEventEmitter):
```
ajtmccarty: it would be more correct for `_Noop..` and `AutoCreate...` to share an interface that they both implement instead of one overriding the methods on the other

backend/infrahub/auth/auth.py
```diff
@@ -368,37 +392,63 @@ async def _assign_group_memberships(
...
+    event_service: InfrahubEventService | None,
+    protocol: ExternalAuthProtocol,
+    provider_name: str,
+) -> AutoCreateEventEmitter:
+    """Build the emitter the auto-create service uses to emit audit events.
+
+    Returns the disabled no-op variant when no event service is supplied.
+    """
+    if event_service is None:
+        return AutoCreateEventEmitter.disabled()
```
ajtmccarty: this is where the `Noop` version should be instantiated instead of passing through the `AutoCreate` class

backend/infrahub/auth/auth_groups/emitter.py
```diff
@@ -0,0 +1,131 @@
...
+@dataclass(frozen=True, slots=True)
+class EmissionDeps:
+    """Runtime dependencies required to emit auto-create events."""
+
+    event_service: InfrahubEventService
+    event_meta_factory: Callable[[], EventMeta]
+    protocol: ExternalAuthProtocol
+
+
+class AutoCreateEventEmitter:
```
ajtmccarty: just wondering if this class should be in the enterprise repo and the interface/no-op version should be in this repo

backend/infrahub/auth/auth_groups/service.py
ogenstad: I'd say that it's fine even though it might have been a bit prettier with just self._account. We could consider to add a SchemaProtocol entry as accepted parameters to `.add()`, not sure if we need to do more fixes after that if that's the case.

backend/infrahub/auth/auth.py
```diff
@@ -179,7 +182,10 @@ async def create_fresh_access_token(
 
 
 async def signin_sso_account(
-    db: InfrahubDatabase, external_identity: ExternalIdentity, sso_groups: list[str]
+    db: InfrahubDatabase,
```
ogenstad, line 191: Regarding what we talked about I think this signin_sso_account looks a lot better with the refactoring compared to what we talked about with what we had in the `stable` release. Well done!

backend/infrahub/auth/auth.py
```diff
@@ -368,37 +392,63 @@ async def _assign_group_memberships(
...
     )
     for group in infrahub_groups:
         members_rel = group.get_relationship(name="members")
-        members = await members_rel.get_peers(db=db, branch_agnostic=True, peer_type=CoreAccount)
-        if account.id in members:
-            continue
-        await members_rel.add(db=db, data=account)
-        await members_rel.save(db=db)
+        if await members_rel.add(db=db, data={"id": account.id}):
+            await members_rel.save(db=db)
```
ogenstad, line 412: This looks a lot cleaner and I like it. I did have to dig through the .add code to see if we are making a db query multiple times though and that part looks fine as we're using the cache. If that were to change at some point we might miss it here. I don't know that there's an action to be taken here for now, but something to consider looking to the future of what these nodes and relationship manager classes could look like after a refactoring.

backend/infrahub/core/constants/__init__.py
```diff
@@ -69,6 +69,9 @@ class EventType(InfrahubStringEnum):
 
     GROUP_MEMBER_ADDED = f"{EVENT_NAMESPACE}.group.member_added"
     GROUP_MEMBER_REMOVED = f"{EVENT_NAMESPACE}.group.member_removed"
+    GROUP_AUTO_CREATED = f"{EVENT_NAMESPACE}.group.auto_create.created"
+    GROUP_AUTO_CREATE_REJECTED_CLAIM = f"{EVENT_NAMESPACE}.group.auto_create.rejected_claim"
+    GROUP_AUTO_CREATE_CAP_BREACH = f"{EVENT_NAMESPACE}.group.auto_create.cap_breach"
```
ogenstad: I think we want to keep the events as three components for consistency. `name.name.name` i.e only two dots. I'd expect f"{EVENT_NAMESPACE}.group.[something]".

backend/infrahub/events/group_action.py
```diff
@@ -107,3 +109,83 @@ class GroupMemberRemovedEvent(GroupMutatedEvent):
...
+    def get_related(self) -> list[dict[str, str]]:
+        related = super().get_related()
+        for idx, claim in enumerate(self.dropped_claims):
+            related.append(
+                {
+                    "prefect.resource.id": f"infrahub.security.dropped_claim.{idx}",
+                    "prefect.resource.role": "infrahub.security.dropped_claim",
+                    "infrahub.security.dropped_claim.value": claim,
+                }
+            )
```
ogenstad, line 190: One thing to flag here is this:
https://github.com/opsmill/infrahub/blob/infrahub-v1.9.6/development/Dockerfile#L94

There's a limit to the number of related resources that are allowed in Prefect for events. We've bumped it up but can still run into issues in a few places. I don't think it should be a problem here as 500 is a very high number of groups to be a member of.

backend/infrahub/graphql/types/event.py
```diff
@@ -68,13 +68,31 @@ class BranchEventTypeFilter(InputObjectType):
...
 
 
+class GroupAutoCreateEventTypeFilter(InputObjectType):
+    idp = List(
+        NonNull(String),
+        required=False,
+        description="Filter by the configured identity-provider name (matches the event's `idp` field)",
+    )
+    protocol = List(
+        NonNull(String),
```
ogenstad, line 78: In the GraphQL schema I think this sets the protocol so that the input is required to be an array, but I think this would still be allowed `[null]` i.e that the String isn't required in this case just the array.

backend/infrahub/task_manager/models.py
```diff
@@ -158,6 +158,23 @@ def add_event_type_filter(
             if branches:
                 self.resource = EventResourceFilter(labels=ResourceSpecification({"infrahub.branch.name": branches}))
 
+        if (group_auto_create := event_type_filter.get("group_auto_create")) is not None:
+            auto_create_event_names = [
+                f"{EVENT_NAMESPACE}.group.auto_create.created",
+                f"{EVENT_NAMESPACE}.group.auto_create.rejected_claim",
+                f"{EVENT_NAMESPACE}.group.auto_create.cap_breach",
```
ogenstad: Check how we refer to the event names here:
https://github.com/opsmill/infrahub/blob/infrahub-v1.9.6/backend/infrahub/computed_attribute/triggers.py#L12

I think that's a safer option as there would otherwise be a disconnect between the event itself and the strings above. I.e. if we change it on one side things could start to silently fail.

ogenstad (commented):
LGTM, but I think we should change the names of the events with regards to the number of dots. Left a few other minor comments.

backend/infrahub/auth/auth.py
```diff
@@ -368,37 +392,63 @@ async def _assign_group_memberships(
...
+    branch = await registry.get_branch(db=db)
+    account_session = AccountSession(
+        auth_type=AuthType.JWT,
+        authenticated=True,
+        account_id=account.id,
+    )
+    event_context = InfrahubContext.init(branch=branch, account=account_session).to_event_context()
+
+    def factory() -> EventMeta:
+        return EventMeta(branch=branch, context=event_context, account_id=account_session.account_id)
```
ajtmccarty, line 439: we generally don't pass functions around like this. adding a component (and maybe interface) for this little object feels like overkill though. does it need to be a factory? can it just be passed into the `AutoCreateEventEmitter` constructor without being a function?

backend/infrahub/auth/auth_groups/service.py
```diff
@@ -53,17 +59,19 @@ def __init__(
         self,
         *,
         db: InfrahubDatabase,
-        account: Node,
+        account: CoreAccount,
         provider_name: str,
         lock_registry: InfrahubLockRegistry,
         max_per_login: int,
+        emitter: AutoCreateEventEmitter,
```
ajtmccarty, line 64: nice dependency injection 👏 👏 👏

ajtmccarty (commented):
looking good. just some dependency injection, component construction thoughts.

backend/infrahub/auth/auth_groups/emitter.py
polmichel: I tried to limit the scope, but I didn't really have a basis to choose which `Exception` subtypes could occur there.

If this sounds good I'll add an internal task just as a follow-up to this question https://opsmill.atlassian.net/browse/IFC-2640

backend/infrahub/auth/auth_groups/service.py
polmichel: I thought about it, but it felt invasive enough that we should discuss it first. If this is an acceptable way to handle the case, I'll do it, since it would silence a bunch of typing warnings.

backend/infrahub/api/oauth2.py
polmichel: My preference would be to inject the service through the constructor so all the object's dependencies are visible in one place. But it's not an object yet. Based on feedback from Aaron, this is a follow-up task I should take care of (internal: https://opsmill.atlassian.net/browse/IFC-2616).

backend/infrahub/events/group_action.py
polmichel: I hadn't realized that. Thanks for the heads-up.

backend/infrahub/graphql/types/event.py
polmichel: I've generated a test locally which gives me these results:
| Input | Accepted? |
  |---|---|
  | *(omitted)* | ✅ |
  | `null` | ✅ |
  | `{ group_auto_create: null }` | ✅ |
  | `{ group_auto_create: { protocol: [] } }` | ✅ |
  | `{ group_auto_create: { protocol: ["oauth2"] } }` | ✅ |
  | `{ group_auto_create: { protocol: [null] } }` (variable) | ❌ |
  | `{ group_auto_create: { protocol: [null, "oauth2"] } }` | ❌ |
  | `{ group_auto_create: { idp: [null] } }` | ❌ |
  | `{ group_auto_create: { protocol: [null] } }` (inline literal) | ❌ |
  
<details>

<summary> The test </summary>



```
from dataclasses import dataclass, field
from typing import Any

import graphene
import pytest

from infrahub.graphql.types.event import EventTypeFilter


@pytest.fixture(scope="module")
def schema() -> graphene.Schema:
    class Query(graphene.ObjectType):
        echo = graphene.String(event_type_filter=EventTypeFilter())

        def resolve_echo(  # type: ignore[no-untyped-def]
            root,  # noqa: ARG002
            info,  # noqa: ARG002
            event_type_filter: dict[str, Any] | None = None,
        ) -> str:
            if event_type_filter is None:
                return "None"
            return repr(event_type_filter.get("group_auto_create"))

    return graphene.Schema(query=Query, auto_camelcase=False)


QUERY = "query ($f: EventTypeFilter) { echo(event_type_filter: $f) }"


@dataclass(kw_only=True)
class FilterCase:
    name: str
    variables: dict[str, Any] = field(default_factory=dict)
    expected_data: str | None = None
    expected_error: str | None = None


ACCEPTED_CASES = [
    FilterCase(
        name="filter_omitted",
        variables={},
        expected_data="None",
    ),
    FilterCase(
        name="filter_null",
        variables={"f": None},
        expected_data="None",
    ),
    FilterCase(
        name="group_auto_create_null",
        variables={"f": {"group_auto_create": None}},
        expected_data="None",
    ),
    FilterCase(
        name="protocol_empty_list",
        variables={"f": {"group_auto_create": {"protocol": []}}},
        expected_data="{'protocol': []}",
    ),
    FilterCase(
        name="protocol_valid_value",
        variables={"f": {"group_auto_create": {"protocol": ["oauth2"]}}},
        expected_data="{'protocol': ['oauth2']}",
    ),
]


REJECTED_CASES = [
    FilterCase(
        name="protocol_list_of_null",
        variables={"f": {"group_auto_create": {"protocol": [None]}}},
        expected_error=(
            "Variable '$f' got invalid value None at 'f.group_auto_create.protocol[0]'; "
            "Expected non-nullable type 'String!' not to be None."
        ),
    ),
    FilterCase(
        name="protocol_null_mixed_with_value",
        variables={"f": {"group_auto_create": {"protocol": [None, "oauth2"]}}},
        expected_error=(
            "Variable '$f' got invalid value None at 'f.group_auto_create.protocol[0]'; "
            "Expected non-nullable type 'String!' not to be None."
        ),
    ),
    FilterCase(
        name="idp_list_of_null",
        variables={"f": {"group_auto_create": {"idp": [None]}}},
        expected_error=(
            "Variable '$f' got invalid value None at 'f.group_auto_create.idp[0]'; "
            "Expected non-nullable type 'String!' not to be None."
        ),
    ),
]


@pytest.mark.parametrize("case", ACCEPTED_CASES, ids=lambda c: c.name)
def test_group_auto_create_filter_accepts_valid_input(
    schema: graphene.Schema,
    case: FilterCase,
) -> None:
    result = schema.execute(QUERY, variables=case.variables)

    assert result.errors is None
    assert result.data == {"echo": case.expected_data}


@pytest.mark.parametrize("case", REJECTED_CASES, ids=lambda c: c.name)
def test_group_auto_create_filter_rejects_null_items(
    schema: graphene.Schema,
    case: FilterCase,
) -> None:
    result = schema.execute(QUERY, variables=case.variables)

    assert result.errors is not None
    assert len(result.errors) == 1
    assert result.errors[0].message == case.expected_error


def test_group_auto_create_filter_rejects_null_in_inline_literal(
    schema: graphene.Schema,
) -> None:
    inline_query = (
        "query { echo(event_type_filter: {group_auto_create: {protocol: [null]}}) }"
    )

    result = schema.execute(inline_query)

    assert result.errors is not None
    assert len(result.errors) == 1
    assert result.errors[0].message == "Expected value of type 'String!', found null."

```
</details>

backend/infrahub/task_manager/models.py
polmichel: You are right! Will do it

backend/infrahub/auth/auth.py
polmichel: The reason is that the `id` field of the `EventMeta` class (`infrahub.events.models`) has a default factory that regenerates a random UUID each time the class is instantiated (when no `id` is given). I wanted to preserve that behavior.

backend/infrahub/auth/auth_groups/emitter.py
polmichel: I think I drafted this class before creating the `AutoCreateEventEmitter` when the methods were using a larger number of parameter and were not using this new class. The goal at this point was too reduce cognitive complexity when one reads the code. But you are right this is better without it after the EventEmitter class. thanks

backend/infrahub/auth/auth_groups/emitter.py
polmichel: I thought it may be overkilled to have an abstract class only for an object and its noop version. But if this sounds good to you, happy to do it.

backend/infrahub/auth/auth_groups/emitter.py
polmichel: LDAP authentication is an enterprise feature, as well as the Log Forwarding Service. But `infrahub.services.adapters.event.InfrahubEventService` and related `infrahub.events.models.InfrahubEvent` are community features right?

ogenstad: Yes

backend/infrahub/auth/auth_groups/service.py
ogenstad: I'd suggest that we go what we have here now, then if you could create a follow up task? We could perhaps add it as a tech debt item for 1.10/1.11?

ogenstad (approved):

backend/infrahub/auth/auth_groups/service.py
polmichel: Good idea, here is the internal task: https://opsmill.atlassian.net/browse/IFC-2641

PR #9327, opened 2026-05-21
fix(workflows): bug-tdd gate uses pulls endpoint instead of issues

polmichel:
## Summary

- Comment gates in `bug-agent-{review,test,fix}`: drop `--paginate`, use `?per_page=100` instead. `gh api --paginate --jq` on the bare-array `/issues/{n}/comments` endpoint fails through the gh-aw proxy with `expected an object but got: array` (the gate-failure symptom on every bug-agent run). The same call works against `api.github.com` directly, so the breakage is in the `--paginate` + proxy interaction; bypassing pagination sidesteps it. 100 comments is plenty — bug-pipeline PRs cap at 3 review iterations.
- `bug-agent-{analyst,review,fix,test}` (f801febcf): `report-failure-as-issue: false` to stop gate failures spamming issues. This is also why the regression went unnoticed.
- `bug-tdd` gate (3716e4445): collapse `issues/{n}` + `pulls/{n}` probe into one `pulls/{n}` + 404 fallback. Saves an API call but does **not** fix the gate failures — the original root-cause diagnosis was wrong (`issues/{n}` returns an object, not an array, so it can't produce this error).


polmichel:
Was originally a bug on GH AW firewall side, which has been resolved

PR #9331, opened 2026-05-22
tests and wiring for PathTraversal and ReachableNodes queries

ajtmccarty:
# Why

`InfrahubPathTraversal` and `InfrahubReachableNodes` are updated to use the new schema-based cypher query

**Goal**: both GraphQL queries are plan-driven, share one Cypher renderer, and short-circuit when no schema route survives planning

## What changed

**Behavioral**

- `path_traversal_resolver` and `reachable_nodes_resolver` now build a `SchemaPlanner` from `graphql_context.active_permissions.resolver`, call `plan()`, short-circuit when `plan.is_empty`, and only otherwise construct and execute the Query.
- `ReachableNodesQuery` becomes a thin `Plan` consumer mirroring `PathTraversalQuery`: `__init__(*, plan, source_id, default_branch_name, max_results)` raises on empty plan; `query_init` delegates to `render_plan_to_cypher`; `get_reachable_nodes()` extracts targets via `extract_path_from_result`.
- `reachable_nodes_resolver` validates target kinds against the schema upfront and raises `GraphQLError("Unknown target kind: …")`.
- Planner now expands generic kinds passed in `TerminalByKinds.kinds`, `UserFilters.kind_filter`, and `UserFilters.excluded_kinds` to their concrete implementors via `_concrete_kinds_for`. Without this, e.g. `target_kinds=["TestAnimal"]` silently produced an empty plan because the adjacency only contains concrete kinds. Also added existence validation for `kind_filter` / `excluded_kinds` entries to surface typos at plan time.

**What stayed the same**

- GraphQL types (`PathTraversalResultType`, `ReachableNodeType`, etc.) output shapes unchanged.
- `PathTraversalQuery` surface.
- Error messages for missing source/destination and identical source==destination.

### Suggested review order

1. `backend/infrahub/graphql/queries/{path,reachable}.py`: the resolver rewires.
2. `backend/infrahub/graph_traversal/reachable.py`: Query class slimmed down.
3. `backend/infrahub/graph_traversal/planning/planner.py`: the three `_expand_*` helpers and their validation.
4. Tests: split between Query-class behavior (`backend/tests/component/graph_traversal/`) and resolver behavior (`backend/tests/component/graphql/queries/test_{path_traversal,reachable_nodes}.py`).

## How to review

- **Resolver-level tests** (`tests/component/graphql/queries/`) exercise the full GraphQL pipeline via `prepare_graphql_params` + `graphql()`. Use them to verify input handling, error messages, and the empty-plan short-circuit.
- **Query-class tests** (`tests/component/graph_traversal/`) construct a plan directly and execute it. Use them to verify branch isolation (default-only / user-only / deletion / untouched), filter semantics that need to inspect rendered behavior (relationship_filter selecting one of two parallel relationships; excluded_kinds dropping one concrete implementor of a generic while keeping another), and generic-input handling.

## Checklist

- [x] Tests added/updated (38 unit + 33 component; covers all four `UserFilters` fields plus generic-input expansion)
- [ ] Changelog entry — relying on the existing `changelog/+graph-path-traversal.added.md` umbrella entry per spec note in `tasks-schema-planning.md`. Add a new fragment if that entry has already shipped.
- [ ] External docs updated — none user-facing changed; the GraphQL field description tweak is in-API and self-documenting.
- [x] Internal .md docs updated (`spec-schema-planning.md` line 134, `tasks-schema-planning.md` task status / deviations)


backend/infrahub/graphql/queries/path.py
```diff
@@ -83,7 +79,11 @@ class PathTraversalInput(InputObjectType):
     excluded_namespaces = List(
         of_type=NonNull(String),
         required=False,
-        description="Namespaces to exclude from traversal. Pass empty list to include all.",
+        description=(
+            "Additional namespaces to exclude from traversal. Unioned with the default "
+            "excluded set (Core, Internal, Builtin, Lineage, Profile, Template); the "
+            "defaults cannot be opted out of from this input."
+        ),
```
polmichel, line 86: Nice, this is clearer. This is related to Patrick's feedback I assume

backend/tests/component/graphql/queries/test_path_traversal.py
```diff
@@ -0,0 +1,331 @@
...
+
+    car_a = await Node.init(db=db, schema="TestCar", branch=default_branch)
+    await car_a.new(db=db, name="Roadster", is_electric=True, nbr_seats=2, color="#ff0000", owner=person)
+    await car_a.save(db=db)
+
+    car_b = await Node.init(db=db, schema="TestCar", branch=default_branch)
+    await car_b.new(db=db, name="Sedan", is_electric=False, nbr_seats=4, color="#0000ff", owner=person)
+    await car_b.save(db=db)
+
+    return car_a, car_b, person
```
polmichel: Optional comment: a dedicated sub package with a common fixture between this test and `backend/tests/component/graphql/queries/test_reachable_nodes.py` maybe could be more maintainable?

backend/tests/helpers/graph_traversal/builders.py
```diff
@@ -22,6 +22,13 @@
     from collections.abc import Iterable
 
     from infrahub.core.schema import GenericSchema, NodeSchema
+    from infrahub.database import InfrahubDatabase
+
+
+def identifier_of(*, db: InfrahubDatabase, branch: Branch, kind: str, relationship: str) -> str:
```
polmichel, line 28: I keep this one in mind, this is clean and could be useful somewhere else. I'll move it another `helpers` folder if I found any usage later

polmichel (approved):
LGTM! Clean. An optional comment posted

backend/tests/component/graphql/queries/test_path_traversal.py
ajtmccarty: yes good point. moved to the nearest `conftest.py` and deduplicated

PR #9342, opened 2026-05-22
feat(graphql): wire enriched error catalogue into GraphQL responses

ogenstad:
# Why

Introduce enriched GraphQL errors by using errors from a catalog.

As an example of how errors will change would be for a mutation that looks like this:

```graphql
mutation CreateTag {
  BuiltinTagCreate(data: {description: {value: "the red tag"}}) {
    ok
    object {
      hfid
      display_label
      id
    }
  }
}
````

Where the current response would look like this:

```json
{
  "data": {
    "BuiltinTagCreate": null
  },
  "errors": [
    {
      "message": "name is mandatory for BuiltinTag at name",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "path": [
        "BuiltinTagCreate"
      ]
    }
  ]
}
```

After this change we'll instead have:

```json
{
  "data": {
    "BuiltinTagCreate": null
  },
  "errors": [
    {
      "message": "name is mandatory for BuiltinTag",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "path": [
        "BuiltinTagCreate"
      ],
      "extensions": {
        "code": "ATTRIBUTE_REQUIRED",
        "http_status": 422,
        "data": {
          "node_kind": "BuiltinTag",
          "field_name": "name"
        }
      }
    }
  ]
}
```

The goal is that for each string in the code we'll have an error catalogue that defines what the data payload could look like. I.e. what fields exist and what the type is of those fields. Then we'll create bindings for the frontend along with the ability to generate bindings for the SDK and make the error catalogue available for inclusion in other tools as well.

## What changed

* Add mappings to present the enriched errors within the GraphQL layer
* Code to parse `ValidationError` this exception was very generic and will need to be broken down into smaller parts, but I didn't want to change the internals of the code for this round and that part will come later in this work.
* Since the error[].extensions.code changed from an integer to a string some changes to the frontend were required, and need some extra care in terms of review.

---
## Summary by cubic
Adds a catalogue-backed GraphQL error formatter and routes both resolver errors and auth short-circuits on `/graphql` through a structured envelope with string codes, HTTP status, and typed payloads for programmatic handling (INFP-468). REST `/api/...` responses are unchanged.

- **New Features**
  - GraphQL errors now include `extensions.code` (string), `extensions.http_status` (int), and typed `extensions.data`; uncatalogued errors use `UNDEFINED_ERROR` with HTTP from `Error.HTTP_CODE` or 500.
  - Validation errors fan out per field with precise `path` and codes: `ATTRIBUTE_REQUIRED`, `ATTRIBUTE_INVALID_TYPE`, `ATTRIBUTE_CONSTRAINT_VIOLATION`; legacy `ValidationError` is auto-classified without refactoring call sites.
  - Auth errors split into `AUTHENTICATION_REQUIRED` vs `TOKEN_EXPIRED`; `/graphql` short-circuits return the catalogue envelope while `/api/...` keeps the old integer code shape.
  - Security: `PERMISSION_DENIED` payload exposes only `action` and `resource_kind`; structlog logs the emitted `code`.

- **Migration**
  - Switch consumers from integer `extensions.code` to string codes; use `extensions.http_status` for numeric checks.
  - Frontend updated: `graphqlClientApollo.tsx` matches on `AUTHENTICATION_REQUIRED`/`TOKEN_EXPIRED` and suppresses toasts for `PERMISSION_DENIED`; `login.tsx` displays string codes.
  - No changes needed for REST `/api/...` consumers.

<sup>Written for commit 471b2424e1abb95e302d4723e5d3704b2f23f86e. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9342?utm_source=github">Review in cubic</a></sup>


backend/infrahub/graphql/mutations/main.py
```diff
@@ -150,7 +151,7 @@ async def _process_file(
         return True
 
     @classmethod
-    async def mutate(
+    async def mutate(  # noqa: PLR0915
```
ogenstad: This method was already too complex and catching the new exception pushed it over the top to require the too-many-statements (PLR0915) noqa line. I didn't want to make larger changes then needed for this iteration, and catching `ValidationError` as we now do below is temporary. But I think we'll have to go back and simplify this method in general later.

backend/infrahub/errors/validation.py
```diff
@@ -0,0 +1,148 @@
...
+        raise original
+
+    raise MultiFieldValidationError(graphql_errors)
+
+
+def flatten_validation_input(exc: ValidationError) -> dict[str, str] | None:
+    """Flatten a ``ValidationError.input_value`` into a single ``{field: reason}`` dict.
+
+    Returns ``None`` when the input cannot be structurally interpreted as per-field reasons,
+    in which case the caller should let the original error propagate as ``UNDEFINED_ERROR``.
```
ajtmccarty: I'd say that this line can go

backend/infrahub/graphql/mutations/main.py
ajtmccarty: would it be better to add it to the `pyproject.toml` exclusions instead of inline?

docs/superpowers/plans/2026-05-27-graphql-error-link-catalogue.md
ajtmccarty, line 1: should this be included?

ajtmccarty (approved):

docs/superpowers/plans/2026-05-27-graphql-error-link-catalogue.md
pa-lem: it can be removed, sorry for the noise

pa-lem (approved):

PR #9343, opened 2026-05-22
remove arg-type exclusion from mypy rules on `schema_branch.py`

ajtmccarty:
## Summary

  Removes `arg-type` from the mypy override for `infrahub.core.schema.schema_branch` and fixes all 20 hidden type errors

  ## Changes

  ### `SchemaBranch.name` is now required
  - `__init__` signature changed from `name: str | None = None` to `name: str` — every production call site already passed a name; only test fixtures were relying on
  the default.
  - `to_dict_schema_object()` now carries `"name"` in the payload; `from_dict_schema_object()` reads it. This keeps round-tripping honest across the Pydantic
  validator/migration pipelines (`validators/model.py`, `validators/models/validate_migration.py`, `migrations/schema/models.py`).
  - `duplicate(name=None)` now falls back to `self.name` instead of producing a nameless copy.
  - 34 test instantiation sites across 16 files updated to pass `name="test"`.

  ### Type narrowing in `SchemaBranch` methods
  - Replaced overly broad `self.get()` calls with the existing typed helpers `get_node()` / `get_generic()` at 6 sites where the iteration source guarantees the kind.
  - Extracted `_diff_node_or_generic()` helper in `SchemaBranch.diff()` to dispatch by kind so each subclass's `diff(other: Self)` contract is satisfied without `cast`.
  - Added `assert isinstance(node, NodeSchema | GenericSchema)` at two loops that iterate `node_names + generic_names_without_templates` — the union return type was
  strictly wider than the iteration source.
  - Widened `generate_profile_from_node(node: NodeSchema)` to `NodeSchema | GenericSchema`. The body only touches `BaseNodeSchema` attributes, and the calling loop
  legitimately invokes it for both kinds (both have `generate_profile`).

  ### Guard against `None` in `add_hierarchy_node`
  Wrapped the parent- and children-relationship blocks in `if parent_peer is not None:` / `if children_peer is not None:` guards. Schema procession would make reaching this point impossible, but the typing does not make that clear. 

---
## Summary by cubic
Make `SchemaBranch.name` required and include it in schema serialization. Tightens typing across `schema_branch.py`, adds safe dispatch in `diff()`, and removes the `arg-type` mypy override.

- **Refactors**
  - `SchemaBranch.__init__` now requires `name: str`; tests updated to pass `name="test"`.
  - `to_dict_schema_object()` returns `SchemaBranchDict` and writes `"name"`; `from_dict_schema_object()` now requires it.
  - Introduced `SchemaBranchDict` `TypedDict`.
  - Added `get_node_or_generic_schema()`; replaced broad `get()` with typed getters (`get_node`/`get_generic`/`get_profile`/`get_template`) and `_diff_node_or_generic()` for type-safe `diff()`.
  - `generate_profile_from_node()` accepts `NodeSchema | GenericSchema`; replaced assertions with conditional checks.
  - `duplicate(name=None)` falls back to `self.name`; removed `arg-type` from mypy overrides in `pyproject.toml`.

- **Bug Fixes**
  - Guarded `add_hierarchy_node` to skip creating parent/children relationships when peers are `None`.

<sup>Written for commit 71b0f018196ce1cd0ffdaa0b81c14d2f9c37a085. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9343?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2484,13 +2484,7 @@ class SchemaBranch:
 
         profile_schema_kinds = set()
         for node_name in self.node_names + self.generic_names_without_templates:
-<<<<<<< HEAD
-            node = self.get(name=node_name, duplicate=False)
-            if not isinstance(node, NodeSchema | GenericSchema):
-                continue
-=======
             node = self.get_node_or_generic_schema(name=node_name, duplicate=False)
->>>>>>> d6ea36073 (add and use get_node_or_generic_schema())
             if (
                 (node.namespace in RESTRICTED_NAMESPACES and node.namespace != "Builtin")
                 or not node.generate_profile
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -2467,7 +2467,7 @@ class SchemaBranch:
         profile_schema_kinds = set()
         for node_name in self.node_names + self.generic_names_without_templates:
             node = self.get(name=node_name, duplicate=False)
-            if not isinstance(node_schema, NodeSchema | GenericSchema):
+            if not isinstance(node, NodeSchema | GenericSchema):
                 continue
             if (
                 (node.namespace in RESTRICTED_NAMESPACES and node.namespace != "Builtin")
--- a/backend/infrahub/core/schema/schema_branch.py
+++ b/backend/infrahub/core/schema/schema_branch.py
@@ -246,9 +246,7 @@ class SchemaBranch:
 
         # Process of the one that have been updated to identify the list of impacted fields
         for key in present_both:
-            local_node = self.get(name=key, duplicate=False)
-            other_node = other.get(name=key, duplicate=False)
-            diff_node = other_node.diff(other=local_node)
+            diff_node = self._diff_node_or_generic(other_schema_branch=other, local_key=key, other_key=key)
             if diff_node.has_diff:
                 schema_diff.changed[key] = diff_node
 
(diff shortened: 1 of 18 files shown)


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1144,6 +1158,7 @@ def validate_parent_component(self) -> None:
         dependency_map: dict[str, set[str]] = defaultdict(set)
         for name in self.generic_names_without_templates + self.node_names:
             node_schema = self.get(name=name, duplicate=False)
+            assert isinstance(node_schema, NodeSchema | GenericSchema)
```
ajtmccarty: not sure if `assert` is better or a `if not isinstance(...): continue`

ogenstad: I'd avoid assert in production code. We have it in a few places, but it can be excluded/ignored at runtime with environment variables.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2457,6 +2467,8 @@ def manage_profile_schemas(self) -> None:
         profile_schema_kinds = set()
         for node_name in self.node_names + self.generic_names_without_templates:
             node = self.get(name=node_name, duplicate=False)
+            if not isinstance(node_schema, NodeSchema | GenericSchema):
```
ajtmccarty:
```suggestion
            if not isinstance(node, NodeSchema | GenericSchema):
```

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -188,8 +188,9 @@ def get_hash_full(self) -> SchemaBranchHash:
     def to_dict(self) -> dict[str, Any]:
         return {"nodes": self.nodes, "generics": self.generics, "profiles": self.profiles, "templates": self.templates}
 
-    def to_dict_schema_object(self, duplicate: bool = False) -> dict[str, dict[str, MainSchemaTypes]]:
+    def to_dict_schema_object(self, duplicate: bool = False) -> dict[str, Any]:
```
ogenstad: I think this is fine as a way to correct things. But Looking at this method I also think that we could do a bit better. It looks like this should return a TypedDict where "name" is just a string and "nodes" is a dict[str, NodeSchema] etc. Would have to change the method so that we call something other than "self.get" for that to work.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -224,7 +225,7 @@ def from_dict_schema_object(cls, data: dict) -> Self:
 
                 cache[node_hash] = node
 
-        return cls(cache=cache, data=nodes)
+        return cls(cache=cache, data=nodes, name=data.get("name", "Unknown"))
```
ogenstad: While I think this should be fine in general there's also the fact that "Unknown" is a perfectly fine name for a branch. I'm assuming that the data dict from here would match the one that's coming from `.to_dict_schema_object()` if that method returned a typeddict instead we should be fine to just access "name" directly.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -1144,6 +1158,8 @@ def validate_parent_component(self) -> None:
         dependency_map: dict[str, set[str]] = defaultdict(set)
         for name in self.generic_names_without_templates + self.node_names:
             node_schema = self.get(name=name, duplicate=False)
+            if not isinstance(node_schema, NodeSchema | GenericSchema):
+                continue
```
ogenstad: Was this only for typing? Wondering if we should have some other method aside from self.get() which would return the expected types instead.

backend/infrahub/core/schema/schema_branch.py
```diff
@@ -2457,6 +2467,8 @@ def manage_profile_schemas(self) -> None:
         profile_schema_kinds = set()
         for node_name in self.node_names + self.generic_names_without_templates:
             node = self.get(name=node_name, duplicate=False)
+            if not isinstance(node, NodeSchema | GenericSchema):
+                continue
```
ogenstad: Same comment as above.

ogenstad (approved):
Nice!

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: oh yeah. way better. added it.

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: added

backend/infrahub/core/schema/schema_branch.py
ajtmccarty: yes. I added a `get_node_or_generic_schema()` method to cover this case and replaced the existing `isinstance` and `assert`s for those types with the new method

PR #9345, opened 2026-05-22
LLM rule for SOLID basics

ajtmccarty:
## Summary by cubic
Adds a backend component design rule for SOLID and dependency injection in `backend/infrahub/**/*.py` to standardize structure and improve testability. Clarifies multiple-implementation guidance and fixes the link to the Python testing rule.

- **New Features**
  - Prefer constructor injection over internal instantiation.
  - Wire components at the app entry point; use a builder/factory if needed.
  - Expose a single entry method; pass work items per-call. Put long-lived deps in the constructor (`db` always; `branch` depends on lifetime).
  - When a second implementation exists, define a `Protocol`/ABC and choose it at the wiring layer; the second impl can be no-op or test, with an updated link to `testing-python.md`.
  - Avoid drive-by refactors of nearby legacy code; propose separately.

<sup>Written for commit 6bd9134668373b16fde535a3bca3fc3d33f1b053. Summary will update on new commits. <a href="https://cubic.dev/pr/opsmill/infrahub/pull/9345?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/dev/rules/backend-component-design.md
+++ b/dev/rules/backend-component-design.md
@@ -41,7 +41,7 @@ in-memory version of a component typically backed by the database).
 
 ## Why this matters
 
-Constructor-injected long-lived dependencies plus method-passed transient entities is the boundary that lets components be reused across requests/operations and mocked with adapter implementations instead of `unittest.mock`. The [testing rule](`dev/rules/testing-python.md`) requires adapter/protocol patterns for tests — that requirement is only practical when production code follows this design.
+Constructor-injected long-lived dependencies plus method-passed transient entities is the boundary that lets components be reused across requests/operations and mocked with adapter implementations instead of `unittest.mock`. The [testing rule](./testing-python.md) requires adapter/protocol patterns for tests — that requirement is only practical when production code follows this design.
 
 ## Existing code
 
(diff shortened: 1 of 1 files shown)


dev/rules/backend-component-design.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Single Responsibility Principle
+
+Each component should have one reason to change. If a class is doing two unrelated things, split it. Prefer composition of small components over large multi-purpose ones.
+
+## Interfaces for multiple implementations
+
+When more than one implementation of a component is required (e.g. real vs. in-memory adapter, different backends, A/B variants), define a `Protocol` or abstract base class. The correct implementation is selected at the wiring layer and injected to the constructor — the consumer codes against the interface, not a concrete class.
+
+A single implementation does not need an interface yet; introduce one when the second implementation arrives.
```
ogenstad: Potentially note that a second implementation could be if we need to use it for a test.

dev/rules/backend-component-design.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Why this matters
+
+Constructor-injected long-lived dependencies plus method-passed transient entities is the boundary that lets components be reused across requests/operations and mocked with adapter implementations instead of `unittest.mock`. The testing rule (`dev/rules/testing-python.md`) requires adapter/protocol patterns for tests — that requirement is only practical when production code follows this design.
```
ogenstad: Regarding the links with lychee we talked about here we could optionally have the dev/rules/testing-python.md be defined as a markdown link after pointing to the file.

ogenstad (approved):
Will be interesting to see how this changes things.

dev/rules/backend-component-design.md
polmichel: Not sure if this is too specific, but the No-op pattern could also be a use case for the abstraction layer.

gmazoyer (approved):

PR #9350, opened 2026-05-26
IFC-2113: Fix failure when removing a profile with a many relationship and overriding with same peers

solababs:
# Why

### Issue 1

When a node's cardinality-many relationship is sourced from a profile (peers A, B, C) and the user saves an override containing a peer that also appears in the profile (e.g. [C, D, E]), peer C is silently dropped. Only D and E remain.

The root cause: `RelationshipManager.update()` reuses the existing `Relationship` object for C without clearing its `is_from_profile=True` flag. When `NodeProfilesApplier.apply_profiles()` runs immediately after (before `save()`), it sees C as profile-sourced and removes it because C is not in the set of user-set peers.

This PR clears `is_from_profile` and the source metadata on any reused relationship peer when the user explicitly includes it in their override.

### Issue 2

Updating a relationship via the `RelationshipAdd` or `RelationshipRemove` mutations left profile-sourcing untouched on the updated relationship. It was possible to add or remove a peer such that 
- the peers on the relationship no longer matched those defined on the profile
- some peers on the relationship had a source of a profile and some did not
both of these situations were unexpected

The fix is to update the two `RelationshipAdd/Remove` mutations so that they still add or remove a single peer AND also clear the profile sourcing on any remaining peers. This way we maintain both the invariants (Claude uses this word all the time and I think I am using it correctly here)
- either all peers on a given relationship have the same profile source or none of them have a profile source
- if any peer on a relationship has a profile source then the peers on the profile's relationship will match the peers on the local relationship

Closes IFC-2113 / IFC-2693

### Issue 1 - overriding a profile-sourced relationship keeps the peers

- **`RelationshipManager.update()` now clears the profile source on reused peers.** When a peer from `previous_relationships` is included in the new override list, `is_from_profile` is set to `False` and `clear_source()` is called before it is re-appended to `_relationships` (applied in all reuse branches: node-object, string peer-ID, and dict). Previously the overlapping peer was wrongly removed; now it is kept and becomes a node-defined value. The per-item logic was extracted into `_process_update_item()` / `_detach_profile_source()` to keep `update()` readable.
- **`RelationshipManager.update_relationships()` returns early when there is nothing to add or remove**, so a no-op reconciliation no longer reports a spurious change.

### Issue 2 - `RelationshipAdd` / `RelationshipRemove` mutations

- **A user modification to a profile-sourced relationship now detaches the whole relationship from the profile.** `RelationshipAdd` and `RelationshipRemove` clear the profile source from every surviving peer when the relationship was profile-sourced and was actually modified. Peers are kept (not deleted) and the profile assignment itself is left in place; only the sourcing is removed. This makes the additive mutations consistent with the object `Update` override and preserves the invariant that a relationship's peers are either all profile-sourced or all user-defined.
- The detach runs **inside the mutation transaction and after validation**, so a failed or no-op mutation leaves the sourcing untouched.
- **`_collect_current_peers()` now requests `SOURCE` metadata** so callers can identify profile-sourced peers (and clear their source) from the already-loaded peer data, without re-reading the relationships.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -1210,7 +1210,7 @@ class RelationshipManager:
                 return rel
         return None
 
-    async def update(
+    async def update(  # noqa: PLR0915
         self,
         data: list[str | Node | dict[str, Any] | PeerWithRelationshipMetadata]
         | dict[str, Any]
@@ -1251,7 +1251,11 @@ class RelationshipManager:
             if hasattr(item, "_schema"):
                 item_id = getattr(item, "id", None)
                 if item_id and item_id in previous_relationships:
-                    self._relationships.append(previous_relationships[str(item_id)])
+                    rel = previous_relationships[str(item_id)]
+                    if rel.is_from_profile:
+                        rel.is_from_profile = False
+                        rel.clear_source()
+                    self._relationships.append(rel)
                     continue
 
             if item is None:
@@ -1263,7 +1267,11 @@ class RelationshipManager:
                 continue
 
             if isinstance(item, str) and item in previous_relationships:
-                self._relationships.append(previous_relationships[item])
+                rel = previous_relationships[item]
+                if rel.is_from_profile:
+                    rel.is_from_profile = False
+                    rel.clear_source()
+                self._relationships.append(rel)
                 continue
 
             if isinstance(item, dict) and item.get("id", None) in previous_relationships:
@@ -1272,6 +1280,9 @@ class RelationshipManager:
                 rel.load(data=item, db=db)
                 if hash(rel) != hash_before:
                     changed = True
+                if rel.is_from_profile:
+                    rel.is_from_profile = False
+                    rel.clear_source()
                 self._relationships.append(rel)
                 continue
 
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -1210,7 +1210,7 @@ class RelationshipManager:
                 return rel
         return None
 
-    async def update(  # noqa: PLR0915
+    async def update(
         self,
         data: list[str | Node | dict[str, Any] | PeerWithRelationshipMetadata]
         | dict[str, Any]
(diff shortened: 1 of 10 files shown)


backend/infrahub/core/relationship/model.py
```diff
@@ -1210,7 +1210,7 @@ async def get_relationship(self, db: InfrahubDatabase, peer_id: str) -> Relation
                 return rel
         return None
 
-    async def update(
+    async def update(  # noqa: PLR0915
```
ajtmccarty: should either add this to the exceptions in `pyproject.toml` or refactor the method so the error isn't raised

backend/tests/component/core/profiles/test_node_applier.py
ajtmccarty, line 698: not caused by this PR, but it looks like this function does not correctly validate the case where peers are defined and `source_uuid=""`, which is what we need to test for this case. it looks like it correctly tests that the source of the peer(s) is empty if no `ExpectedProfileRelationship` for the relationship name, but does not verify that if the `ExpectedProfileRealtionship` is defined with a source set to the empty string

backend/infrahub/core/relationship/model.py
ajtmccarty, line 1: `_detach_profile_source()` is the only new functional piece added in `update()`, the rest is just refactoring b/c the new logic surpassed the complexity limit

backend/infrahub/core/relationship/model.py
```diff
@@ -1296,6 +1261,77 @@ async def update(
 
         return changed
 
+    def _detach_profile_source(self, rel: Relationship) -> None:
```
ajtmccarty, line 1264: this means that any time a peer is passed into `RelationshipManager.update()` and that peer is already part of this relationship and has a profile source, then the peer will have its profile source removed. I think this is the expected behavior. Verified that this works with the UI b/c relationships from profiles are only included in the graphql mutation if actually being updated

backend/infrahub/core/relationship/model.py
```diff
@@ -1485,6 +1521,9 @@ async def update_relationships(
         len_rels_to_remove = len(relationships_to_remove)
         len_rels_to_insert = len(relationships_to_insert)
 
+        if not len_rels_to_remove and not len_rels_to_insert:
```
ajtmccarty, line 1524: this missing check was causing a relationship to appear as changed in `NodeProfilesApplier.apply()` even if no changes were made to the relationship

backend/tests/component/core/profiles/test_node_applier.py
```diff
@@ -257,8 +257,10 @@ async def _validate_node_profile_relationships(
 
         if expected_profile_relationship:
             assert {p.id for p in updated_peers} == {p.id for p in expected_profile_relationship.peers}
-            if expected_profile_relationship.source_uuid or updated_source:
+            if expected_profile_relationship.source_uuid:
                 assert updated_source == {expected_profile_relationship.source_uuid}
+            else:
+                assert updated_source == set()
```
ajtmccarty, line 263: test did not correctly verify the "has no source" case before

backend/tests/component/graphql/profiles/test_mutation_relationship_override.py
ajtmccarty, line 1: new graphql layer test for updating profile-sourced relationships to update with an existing peer

ajtmccarty:
follow-up issue for xfail test added in this PR: https://opsmill.atlassian.net/browse/IFC-2693

backend/infrahub/graphql/mutations/relationship.py
ajtmccarty, line 1: changes for clearing profile-sourcing during a `RelationshipAdd` or `RelationshipRemove` graphql mutation are in here

backend/tests/component/graphql/profiles/test_mutation_relationship_override.py
ajtmccarty: and to test RelationshipAdd/Remove correctly clear profile source from all peers

polmichel (approved):
LGTM, nice test suite ! `backend/tests/component/graphql/profiles/test_mutation_relationship_override.py`

PR #9359, opened 2026-05-26
fix: enforce uniqueness on computed attributes (closes #7924)

### LINKED ISSUE #7924: bug: Unique constraint not applied on computed attribute leading to duplicated HFID
### Component

API Server / GraphQL

### Infrahub version

1.5.4

### Current Behavior

I use the name of my object as a HFID. It's unique and computed as jinja2 attribute.

However under certain circumstances it seems that this name could be duplicated and so my hfid as well, leading to various issues down the line.

Image

### Expected Behavior

I expect the application to refuse the data before it reaches the database.

### Steps to Reproduce

- Load this schema

```yml
---
version: "1.0"

nodes:
  - name: Stuff
    namespace: Random
    display_label: name__value
    human_friendly_id:
      - name__value
    attributes:
      - name: name
        kind: Text
        computed_attribute:
          kind: Jinja2
          jinja2_template: "{{ description__value | upper }}-STUFF"
        read_only: true
        unique: true
        optional: false
      - name: description
        kind: Text

```

- Then load the following object file

```yml
---
apiVersion: infrahub.app/v1
kind: Object
spec:
  kind: RandomStuff
  data:
    - description: DESCRIPTION
    - description: DESCRIPTION
    - description: DESCRIPTION
    - description: DESCRIPTION

```

-> This will still create 4 objects, all with the same name even tho it should be unique

### Additional Information

_No response_

polmichel:
## Why

Loading objects whose human-friendly id (or any `unique: true` field) is a **computed** Jinja2 attribute could create duplicates. 

### Schema used
```
---
version: "1.0"

nodes:
  - name: Stuff
    namespace: Random
    display_label: name__value
    human_friendly_id:
      - name__value
    attributes:
      - name: name
        kind: Text
        computed_attribute:
          kind: Jinja2
          jinja2_template: "{{ description__value | upper }}-STUFF"
        read_only: true
        unique: true
        optional: false
      - name: description
        kind: Text
```

Submitting four `RandomStuff` objects with the same `description` produced four nodes sharing the same computed `name` / HFID, because the uniqueness check silently skipped the computed field.

Root cause: the constraint widened its check filter only for *optional* unique attributes. A required computed attribute is never in the user-supplied data, and the macro populates its value only after the filter was captured, so the check was bypassed.

Closes #7924

## What changed

- A required computed unique attribute is now enforced on create. Duplicates are rejected before reaching the database.
- The violation message names the input attributes that drive the computed value, e.g. `Violates uniqueness constraint 'name' (computed from: model)` (or `owner.name` for relationship-derived inputs), so the user knows which field to change rather than seeing an opaque reference to a read-only field.

**What stayed the same**
- `NodeAttributeUniquenessConstraint` deliberately left alone: it is registered as a dependency but not wired into the production `NodeConstraintRunner`.

## How to test

```bash
uv run pytest backend/tests/unit/core/node/constraints/test_uniqueness_violation_message.py -v
uv run pytest backend/tests/component/core/node/test_create_node_computed_uniqueness.py -v
uv run pytest backend/tests/component/core/constraint_validators/ -v
```

Reproduction from the issue: load the `RandomStuff` schema (computed unique `name`), then load four objects with identical `description`. Before: four duplicates created. After: rejected with `Violates uniqueness constraint 'name' (computed from: description)`.

## Follow-up: complementary SDK-side guard

This PR fixes enforcement at the server boundary, which is the correct authoritative place. A complementary improvement could live in the Python SDK / `infrahubctl object load` to catch the common case *before* anything is sent to the server.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (n/a — bug fix, no user-facing doc change)
- [ ] Internal .md docs updated (n/a)


backend/tests/component/core/constraint_validators/test_node_uniqueness.py
```diff
@@ -25,7 +32,13 @@ async def test_node_validate_constraint_node_uniqueness_failure(
 async def test_node_validate_constraint_node_uniqueness_success(
     db: InfrahubDatabase, default_branch: Branch, car_accord_main: Node, car_volt_main: Node, person_john_main: Node
 ) -> None:
-    constraint = NodeGroupedUniquenessConstraint(db=db, branch=default_branch)
+    constraint = NodeGroupedUniquenessConstraint(
+        db=db,
+        branch=default_branch,
+        message_builder=UniquenessViolationMessageBuilder(
```
polmichel, line 38: I was wondering if a Noop class could be useful to avoid instantiating this class every time we instantiate `NodeGroupedUniquenessConstraint` or if this is over-engineered. Not sure of the answer, so I let the code in this state for this iteration.

backend/infrahub/core/node/constraints/grouped_uniqueness.py
```diff
@@ -208,9 +217,23 @@ def _frozen_constraints(schema: MainSchemaTypes) -> frozenset[frozenset[str]]:
...
             raise ValidationError(error_msg)
 
         if hfid_violation:
-            error_msg = f"Violates uniqueness constraint '{'-'.join(hfid_violation.fields)}'"
+            error_msg = self._message_builder.build(node_schema=node.get_schema(), fields=hfid_violation.fields)
             raise HFIDViolatedError(error_msg, matching_nodes_ids=hfid_violation.nodes_ids)
+
+    @staticmethod
+    def _should_implicitly_check_uniqueness(attr_schema: AttributeSchema, schema_filters: list[str]) -> bool:
+        """A unique attribute whose value is set by the system rather than supplied by the user.
```
ajtmccarty: this doesn't actually say what this method does. the details below describe the method, but this first line says nothing

backend/infrahub/core/node/constraints/grouped_uniqueness.py
```diff
@@ -208,9 +217,23 @@ def _frozen_constraints(schema: MainSchemaTypes) -> frozenset[frozenset[str]]:
...
-            error_msg = f"Violates uniqueness constraint '{'-'.join(violation.fields)}'"
+            error_msg = self._message_builder.build(node_schema=node.get_schema(), fields=violation.fields)
             raise ValidationError(error_msg)
 
         if hfid_violation:
-            error_msg = f"Violates uniqueness constraint '{'-'.join(hfid_violation.fields)}'"
+            error_msg = self._message_builder.build(node_schema=node.get_schema(), fields=hfid_violation.fields)
             raise HFIDViolatedError(error_msg, matching_nodes_ids=hfid_violation.nodes_ids)
+
+    @staticmethod
```
ajtmccarty: I think this should be an instance method. it is not going to be used anywhere else and it could benefit from using `self` at some point in the future.
I think the ruff or mypy or ty or whatever rule that mandates any method that does not use `self` be a `staticmethod` is wrong

backend/infrahub/core/node/constraints/uniqueness_violation_message.py
```diff
@@ -0,0 +1,78 @@
...
+
+class UniquenessViolationMessageBuilder:
+    """Format a uniqueness-violation message.
+
+    Names the input attributes that drive any computed field involved, so the user knows
+    which value to change.
+    """
+
+    def __init__(self, schema_branch: SchemaBranch) -> None:
+        self.schema_branch = schema_branch
```
ajtmccarty, line 29: excellent 👍

ajtmccarty (approved):
great work

backend/infrahub/core/node/constraints/grouped_uniqueness.py
polmichel: You are right I'll fix the docstring text style

backend/infrahub/core/node/constraints/grouped_uniqueness.py
polmichel: I think you've already given me this kind of feedback. I couldn't remember why, but when I was writing this code, I had the feeling that a non-static method was a better idea, and at the same time there was the warning from the tool. Thanks for the feedback (once again)

PR #9360, opened 2026-05-26
fix: prevent git task workers from diverging during sync fan-out (#9349)

### LINKED ISSUE #9349: bug: task workers diverge when upstream commit lands between sync start and fan-out git pulls
### Component

Git Integration

### Infrahub version

1.9.2

### Current Behavior

In a multi-worker `infrahub-worker` pool, the git sync flow exhibits a race condition when a new upstream commit lands during the fan-out phase of the sync. Specifically:

1. Worker A picks up the sync job and resolves the upstream HEAD to commit `X`.
2. Worker A fans out follow-up work to the other workers in the pool.
3. **A new commit `X+1` is pushed upstream** between fan-out and the moment the other workers perform their own `git pull`.
4. Workers B and C pull, landing on `X+1`, while Worker A continues operating against `X`.
5. The pool ends in a divergent state: different workers have different `HEAD` commits for the same repo, and `Repository.commit` in the database is **not** updated (neither to `X` nor `X+1`).

The result is silent drift — the DB pointer never advances, and downstream consumers (generators, checks, transforms, artifacts) end up reading whichever local clone the worker handling their flow happens to have.

### Expected Behavior

The sync flow should pin a single commit SHA at the start of the operation and have all fan-out workers fetch and check out **that exact SHA**, regardless of what has since landed upstream. Concretely:

- Worker A resolves upstream HEAD once and records the SHA.
- Fan-out work carries the pinned SHA, and other workers `git fetch` + check out that SHA (not `pull` to the latest).
- On success, `Repository.commit` is updated to the pinned SHA.
- A separate, subsequent sync job picks up `X+1` cleanly.

### Steps to Reproduce

Intermittent / race-condition-like; the trigger pattern is:

1. Deploy Infrahub 1.9.2 on Kubernetes via the `infrahub-helm` chart with 2–3 replicas in the task worker pool.
2. Workers started via `prefect worker start --type infrahubasync --pool infrahub-worker --with-healthcheck`.
3. Connect a git repository to Infrahub.
4. Trigger a sync job (webhook or manual).
5. Push a new commit upstream **after** the sync flow has started but **before** the fan-out workers reach their `git pull` step.
6. Inspect each worker pod's local clone (`git log -1`) — different HEADs across pods.
7. Inspect the `Repository` object in Infrahub — `commit` is older than both the original and the late-arriving upstream commit.

### Additional Information

- Deployment: Kubernetes via `opsmill/infrahub-helm` chart.
- Worker pool size: 2–3 replicas.
- Worker command: `prefect worker start --type infrahubasync --pool infrahub-worker --with-healthcheck`.
- **Workaround / self-healing condition**: pushing a new commit upstream *after* the divergent sync has finished on all workers typically resolves the issue. The subsequent sync cycle converges all workers and updates `Repository.commit`. This suggests the divergent state is recoverable on the next clean sync, but a stuck repo will not self-heal until a fresh commit triggers another flow.
- Suspected root cause: fan-out workers `git pull` (resolve-then-fetch) instead of fetching a SHA pinned at flow start. A commit landing in this window makes the workers' resolutions diverge from Worker A's, and the DB update — likely conditioned on consensus or on Worker A's view — never completes.
- Open questions for maintainers:
  - Does the sync flow pin a SHA at job start and pass it to fan-out tasks, or do fan-out workers re-resolve `HEAD`?
  - Under what conditions is the DB `Repository.commit` write skipped? Is there a guard that suppresses the write when workers disagree?

polmichel:
## Why

In a multi-worker task pool, the git sync flow fans work out across workers. When a new upstream commit landed between the orchestrator resolving HEAD and the other workers running their own `git pull`, workers ended up on different commits for the same repo.

Goal: make every fan-out worker converge on the exact commit the orchestrator pinned, regardless of what lands upstream during fan-out.

This PR also carries a second, independent git fix (see below).

Closes #9349

## What changed

**Behavioral**

- The sync orchestrator now records the commit SHA it just synced and broadcasts it in `RefreshGitFetch`. Receiving workers check out that exact SHA instead of fast-forwarding to whatever upstream HEAD currently is.
- **Separate fix:** pulling a repository that had to create a *missing* local branch crashed when recording its commit value (an unbound `infrahub_branch`). The branch is now bound before the commit is written.

## How to review

A commit by commit approach would be relevant since there is:
- One commit for the initial bug fix 93e86ebfab0ef3b82b51be608cfe5c928a41b014 and one for the related test 7e2f13975564a1f6e54b4ef0a6f67f7965701eb7
- One commit for the collateral bug and related test 308110c5e8a738d7922a270214a4f2ee4bcbdba8
- One commit for the refactor 419ef910aa2e7c457d072416154a9f37955834a9

Following commits are unitary fix or similar issue using the new pinned commit, or documentation/lint/format matter.
Extra scrutiny welcome on the pin source per flow.

## How to test

```bash
# Divergence reproduction (fails before the fix, passes after)
uv run pytest backend/tests/component/message_bus/operations/git/test_refresh_git_fetch.py -x -v

# Missing-branch pull regression
uv run pytest backend/tests/component/git/test_git_repository.py -x -v
```

## Known follow-ups (intentionally out of scope)

- That read-only flow broadcasts `repository_kind=REPOSITORY` for a read-only repo. Worth confirming independently.

## Checklist

- [x] Tests added/updated
- [x] Changelog entries added (`9349.fixed.md`, `+pull-new-branch-unbound-commit.fixed.md`)
- [ ] External docs updated (no user-facing/ops-facing surface change)
- [ ] Internal .md docs updated (no knowledge-doc change needed)

---
## Summary by cubic
Prevents git task workers from diverging during sync fan-out by pinning the orchestrator commit and hard-resetting to that SHA under the repo lock. Also fixes a crash when pulling a new branch, records commit values, broadcasts the correct kind for read-only adds, and improves errors when a pinned commit is unreachable; addresses IFC-2642.

- **Bug Fixes**
  - Added optional `commit` to `RefreshGitFetch`; workers fetch then hard-reset via `reset_to_commit` under the repo lock, falling back to pull when the pin is missing or can’t be resolved.
  - Applied pinning across periodic sync, repository add (read/write and read-only), branch create, merge, and read-only pull; ref-only read-only flows resolve a concrete SHA once and broadcast it.
  - Read-only add now broadcasts `repository_kind=READONLYREPOSITORY` so workers materialize the correct repo type.
  - Fixed missing-branch pull crash by binding the branch before writing its commit; commit value updates when a new local branch is created.
  - Clearer error when a pinned commit isn’t reachable after fetch; raises a descriptive `RepositoryError` and leaves the worktree unchanged; narrowed pin-resolution exceptions; added fan-out convergence and sad-path tests; updated RPC mocks and message-bus docs for the new `commit` field.

- **Refactors**
  - Removed unused `_update_commit_value_if_requested` helper.

<sup>Written for commit 472ad025467e1b18d7c14e4ee9f55781fd7729e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9360?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -969,6 +969,7 @@ class InfrahubRepositoryBase(BaseModel, ABC):
             await self.create_branch_in_git(branch_name=branch_name, branch_id=branch_id)
             repo = self.get_git_repo_worktree(identifier=branch_name)
             commit_after = str(repo.head.commit)
+            infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
         else:
             raise ValueError(
                 f"Unable to identify the worktree for the branch : {branch_name} "
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -986,9 +986,9 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         else:
             raise ValueError(f"Unable to identify the worktree for the branch : {branch_name}")
 
-        await self._update_commit_value_if_requested(
-            branch_name=branch_name, commit=commit_after, update_commit_value=update_commit_value
-        )
+        if update_commit_value:
+            infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
+            await self.update_commit_value(branch_name=infrahub_branch, commit=commit_after)
         return commit_after
 
     async def reset_to_commit(
@@ -998,7 +998,7 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         branch_id: str | None = None,
         create_if_missing: bool = False,
         update_commit_value: bool = True,
-    ) -> str:
+    ) -> None:
         """Hard-reset a branch worktree to a specific commit already present locally.
 
         The caller must have fetched the commit beforehand; this method does not
@@ -1008,9 +1008,6 @@ class InfrahubRepositoryBase(BaseModel, ABC):
             ValueError: When no worktree exists for the branch and ``branch_id`` is not provided to create one.
 
         """
-        if not self.has_origin:
-            return commit
-
         repo = self._get_branch_worktree(branch_name)
         if repo is None:
             if not create_if_missing or not branch_id:
@@ -1026,10 +1023,10 @@ class InfrahubRepositoryBase(BaseModel, ABC):
             await self._raise_enriched_error(error=exc, branch_name=branch_name)
 
         self.create_commit_worktree(commit=commit)
-        await self._update_commit_value_if_requested(
-            branch_name=branch_name, commit=commit, update_commit_value=update_commit_value
-        )
-        return commit
+
+        if update_commit_value:
+            infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
+            await self.update_commit_value(branch_name=infrahub_branch, commit=commit)
 
     async def get_conflicts(self, source_branch: str, dest_branch: str) -> list[str]:
         repo = self.get_git_repo_worktree(identifier=dest_branch)
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -980,6 +980,54 @@ class InfrahubRepositoryBase(BaseModel, ABC):
 
         return commit_after
 
+    async def reset_to_commit(
+        self,
+        branch_name: str,
+        commit: str,
+        branch_id: str | None = None,
+        create_if_missing: bool = False,
+        update_commit_value: bool = True,
+    ) -> str:
+        """Hard-reset a branch worktree to a specific commit already present locally.
+
+        The caller must have fetched the commit beforehand; this method does not
+        contact the remote.
+
+        Raises:
+            ValueError: When no worktree exists for the branch and ``branch_id`` is not provided to create one.
+
+        """
+        if not self.has_origin:
+            return commit
+
+        identifier = branch_name
+        if branch_name == self.default_branch and branch_name != registry.default_branch:
+            identifier = "main"
+
+        try:
+            repo = self.get_git_repo_worktree(identifier=identifier)
+        except RepositoryError as exc:
+            if not create_if_missing or not branch_id:
+                raise ValueError(f"Unable to identify the worktree for the branch : {branch_name}") from exc
+            await self.create_branch_in_git(branch_name=branch_name, branch_id=branch_id)
+            repo = self.get_git_repo_worktree(identifier=branch_name)
+
+        # Hard reset, not merge: the worktree is a disposable mirror of the remote, so we
+        # force it onto the requested commit and intentionally drop any local divergence.
+        # Convergence on this exact SHA is the goal; a fast-forward could land elsewhere.
+        try:
+            repo.git.reset("--hard", commit)
+        except GitCommandError as exc:
+            await self._raise_enriched_error(error=exc, branch_name=branch_name)
+
+        self.create_commit_worktree(commit=commit)
+        infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
+
+        if update_commit_value:
+            await self.update_commit_value(branch_name=infrahub_branch, commit=commit)
+
+        return commit
+
     async def get_conflicts(self, source_branch: str, dest_branch: str) -> list[str]:
         repo = self.get_git_repo_worktree(identifier=dest_branch)
         if not repo:
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -1133,6 +1133,15 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         if "error: pathspec" in error.stderr:
             raise RepositoryInvalidBranchError(identifier=name, branch_name=branch_name, location=location) from error
 
+        if "reset" in (error.command or []) and "Could not parse object" in error.stderr:
+            raise RepositoryError(
+                identifier=name,
+                message=(
+                    f"Commit not found in the local clone of repository {name}; "
+                    "it may have been force-pushed or pruned upstream."
+                ),
+            ) from error
+
         if "SSL certificate problem" in error.stderr or "server certificate verification failed" in error.stderr:
             raise RepositoryConnectionError(
                 identifier=name, message=f"SSL verification failed for {name}, please validate the certificate chain."
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -924,6 +924,31 @@ class InfrahubRepositoryBase(BaseModel, ABC):
 
         return True
 
+    def _resolve_worktree_identifier(self, branch_name: str) -> str:
+        """Map a branch name to the identifier used to locate its worktree on disk."""
+        if branch_name == self.default_branch and branch_name != registry.default_branch:
+            return "main"
+        return branch_name
+
+    def _get_branch_worktree(self, branch_name: str) -> Repo | None:
+        """Return the existing worktree for a branch, or None when it has none yet."""
+        identifier = self._resolve_worktree_identifier(branch_name)
+        try:
+            return self.get_git_repo_worktree(identifier=identifier)
+        except RepositoryError:
+            return None
+
+    async def _create_branch_worktree(self, branch_name: str, branch_id: str) -> Repo:
+        """Create the branch in git and return its freshly created worktree."""
+        await self.create_branch_in_git(branch_name=branch_name, branch_id=branch_id)
+        return self.get_git_repo_worktree(identifier=branch_name)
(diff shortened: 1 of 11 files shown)


backend/infrahub/git/tasks.py
polmichel, line 152: I've not pinned a commit to this fetch because this is a read-only repository and `get_commit_value` re-resolves the repository. Maybe it would be better from a standardization perspective.

backend/infrahub/git/tasks.py
```diff
@@ -145,6 +157,7 @@ async def add_git_repository_read_only(model: GitRepositoryAddReadOnly) -> None:
                 repository_kind=InfrahubKind.REPOSITORY,
```
polmichel: Is this an issue that this is tagged as `REPOSITORY` and not a read-only repository type?

docs/docs/reference/message-bus-events.mdx
```diff
@@ -79,6 +79,7 @@ For more detailed explanations on how these events are used within Infrahub, see
 | **repository_kind** | The type of repository | string | None |
 | **infrahub_branch_name** | Infrahub branch on which to sync the remote repository | string | None |
 | **infrahub_branch_id** | Id of the Infrahub branch on which to sync the remote repository | string | None |
+| **commit** | Commit SHA to check out, pinned by the sync orchestrator instead of pulling the latest upstream HEAD | N/A | None |
```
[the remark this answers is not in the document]
polmichel: This documentation is auto-generated. This is surely due to the fact that the field as a default value at `None`, but I am not totally sure.

backend/infrahub/git/base.py
```diff
@@ -924,6 +924,31 @@ def validate_remote_branch(self, branch_name: str) -> bool:
...
+            return self.get_git_repo_worktree(identifier=identifier)
+        except RepositoryError:
+            return None
+
+    async def _create_branch_worktree(self, branch_name: str, branch_id: str) -> Repo:
+        """Create the branch in git and return its freshly created worktree."""
+        await self.create_branch_in_git(branch_name=branch_name, branch_id=branch_id)
+        return self.get_git_repo_worktree(identifier=branch_name)
+
+    async def _update_commit_value_if_requested(self, branch_name: str, commit: str, update_commit_value: bool) -> None:
```
ajtmccarty: this one feels like it should be a conditional at the caller-level. 
`if update_commit_value: await self._update_commit_value(...)`
maybe it needs a different name to distinguish it from `self.update_commit_value`

backend/infrahub/git/base.py
```diff
@@ -939,47 +964,73 @@ async def pull(
...
         return commit_after
 
+    async def reset_to_commit(
+        self,
+        branch_name: str,
+        commit: str,
+        branch_id: str | None = None,
+        create_if_missing: bool = False,
+        update_commit_value: bool = True,
+    ) -> str:
```
ajtmccarty: does this really need to return anything if it only ever returns the `commit` passed in?

backend/infrahub/git/base.py
```diff
@@ -939,47 +964,73 @@ async def pull(
...
+
+        The caller must have fetched the commit beforehand; this method does not
+        contact the remote.
+
+        Raises:
+            ValueError: When no worktree exists for the branch and ``branch_id`` is not provided to create one.
+
+        """
+        if not self.has_origin:
+            return commit
```
ajtmccarty: is this check necessary if this method doesn't contact the remote?

backend/infrahub/message_bus/operations/git/repository.py
```diff
@@ -46,12 +47,24 @@ async def fetch(message: messages.RefreshGitFetch) -> None:
...
-    await repo.pull(
-        branch_name=message.infrahub_branch_name,
-        branch_id=message.infrahub_branch_id,
-        create_if_missing=True,
-        update_commit_value=False,
-    )
+    if message.commit:
+        # reset_to_commit hard-resets the worktree, which would silently discard a concurrent
+        # local change. Serialize it against the merges and syncs holding the same lock.
+        async with lock.registry.get(name=message.repository_name, namespace="repository"):
```
ajtmccarty: is this counting on the lock to ensure there are no local changes that will be discarded by the reset? it seems like it would still be possible to clobber local changes even after the lock is released, but I might not entirely understand this workflow

backend/infrahub/git/base.py
polmichel: This method only exists to encapsulate the following shared code block:
```
        if update_commit_value:
            infrahub_branch = self._get_mapped_target_branch(branch_name=branch_name)
            await self.update_commit_value(branch_name=infrahub_branch, commit=commit)
```

which appears in both `infrahub.git.base.InfrahubRepositoryBase.pull` and `infrahub.git.base.InfrahubRepositoryBase.reset_to_commit`. Having said that, this is true that encapsulating this condition might actually hurt readability, and wrapping just two lines isn't really worth it. I'll remove it.

backend/infrahub/git/base.py
polmichel: You are right, good catch. I'll type the return as `None`. The return value is not even used.

backend/infrahub/git/base.py
polmichel: Yes, you're right. At some point the method was called `pull_to_commit` or something like that, and it called a `fetch` method that took a specific commit. But Claude warned me that some git servers don't allow fetching a specific commit directly, since there's no guarantee that commit exists in the remote's history. So I renamed it to `reset_to_commit` and... forgot this one. I'll remove it.

backend/infrahub/message_bus/operations/git/repository.py
polmichel: No, this lock is to make sure that there is no parallel operation that impacts the repository while we are resetting the local worktree on each worker.

This lock cannot protect against local changes, and I have not found a way to prevent local changes to be overwritten by this reset.

The comment is confusing, I'll change it.

polmichel: Thinking about it more, this lock may be too broad, I think a worker-internal lock would be better. As it stands, all workers will end up locking the repository from any operation one after another.

polmichel: But this would require to verify that operations related to worker-internal state are all using this scope of lock I guess.

backend/infrahub/git/base.py
ajtmccarty: I think this is dead now

backend/infrahub/git/tasks.py
ajtmccarty: it certainly looks wrong

ajtmccarty (approved):
looks good to me. but maybe someone more familiar with our git stuff wants to review it too (@ogenstad or @gmazoyer)

backend/infrahub/git/base.py
polmichel: :shame:
Thank you.

backend/infrahub/git/tasks.py
polmichel: i'll push a commit changing that, but will also ask a review from @ogenstad or @gmazoyer tomorrow before merging

gmazoyer (approved):
Looks good to me.

You could add a couple of tests:
* one for the read-only repository_kind correction (REPOSITORY → READONLYREPOSITORY). A short assertion on the broadcast `repository_kind` from `add_git_repository_read_only` would lock that in. Maybe an existing test could be slightly changed to integrate this assertion.
* one for the failure case where the pinned SHA isn't reachable after fetch (e.g., broadcasted SHA that was force-pushed away). Worth a sad-path test confirming the worker raises rather than silently fast-forwards.

polmichel:
> Looks good to me.
> 
> You could add a couple of tests:
> 
> * one for the read-only repository_kind correction (REPOSITORY → READONLYREPOSITORY). A short assertion on the broadcast `repository_kind` from `add_git_repository_read_only` would lock that in. Maybe an existing test could be slightly changed to integrate this assertion.
> * one for the failure case where the pinned SHA isn't reachable after fetch (e.g., broadcasted SHA that was force-pushed away). Worth a sad-path test confirming the worker raises rather than silently fast-forwards.

Thanks for the feedbacks @gmazoyer. I've added the two test. I also added some logic to generate a clean error message when the `reset --hard` operation fails. This is tested through the second test related to the sad path.

PR #9361, opened 2026-05-27
Drop LDAP email attribute

gmazoyer:
## Why

Two community-side LDAP loose ends. First, LDAPSettings carries an attribute_email field that doesn't pay rent: the community account schema has no email field, and the LDAP runtime never reaches the place where ExternalIdentity.email matters (SSO uses it as a fallback name on display-name collision, but the LDAP path refuses same-name collisions outright before that fallback would fire). Operators get a knob they don't need on an already-busy config block. Second, the SSO model requires email to be a non-empty string, which means LDAP either fabricates a value or carries one through purely to keep the schema happy. Making email optional removes the fabrication.

## What changed

`LDAPSettings.attribute_email` is removed.

`ExternalIdentity.email` becomes `str | None`, defaulting to `None`. OIDC and OAuth2 callers still pass it; the LDAP runtime can now omit it.

---
## Summary by cubic
Removes the unused LDAP email attribute and updates docs and docker-compose envs. Makes `ExternalIdentity.email` optional and raises a clear error when a display-name collision needs an email but the provider didn’t supply one.

- **New Features**
  - Clear error when a display-name collision occurs and no email is available for fallback.

- **Migration**
  - Remove `LDAPSettings.attribute_email` and any `INFRAHUB_LDAP_ATTRIBUTE_EMAIL` env var from configs.
  - Handle `None` for `ExternalIdentity.email` in any custom code; OIDC/OAuth2 still provide it.

<sup>Written for commit b8c1baf7a6d604ae1ea153f9dd6e30657acd4a54. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9361?utm_source=github">Review in cubic</a>


ogenstad:
Hey @gmazoyer, I'm wondering if this is actually needed if you check this validation: https://github.com/opsmill/infrahub/blob/develop/backend/infrahub/server.py#L54

The idea would be that if you try to enable LDAP in the community version the server would fail to start.

ajtmccarty (approved):

ogenstad (approved):

PR #9365, opened 2026-05-27
Surface group auto-create events in the activity feed

polmichel:
Frontend for the SSO group auto-creation audit events added on the backend in #9325.

## Changes
- Render the three auto-create audit events (`auto_created` / `auto_create_rejected` / `auto_create_capped`) in the activity feed and event detail view.
- Add the three event types to the event-type filter dropdown.
- Emit the auto-created group as a related node so the event appears in that group's own activity timeline.

The idp/protocol event filter was implemented and then dropped — INFP-556 specifies displaying these events, not filtering them.

### Filtering on Group auto-created event
image
#### Details
image
#### Accessible from groups
image

### Filtering on Group auto-create rejected
image
#### Details
image

### Filtering on Group auto-create capped
image
#### Details
image

Jira: [IFC-2646](https://opsmill.atlassian.net/browse/IFC-2646) / Part of [IFC-2521](https://opsmill.atlassian.net/browse/IFC-2521)

[IFC-2646]: https://opsmill.atlassian.net/browse/IFC-2646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/app/src/entities/events/ui/group-auto-create-events/group-auto-create-event-title.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <AutoCreateLayout idp={props.idp}>
+          claim <Highlight>{props.rejected_claim_value}</Highlight> rejected (invalid group name)
+        </AutoCreateLayout>
+      );
+    }
+    case "GroupAutoCreateCappedEventType": {
+      return (
+        <AutoCreateLayout idp={props.idp}>
+          auto-create cap of <Highlight>{props.cap_value}</Highlight> reached, {props.dropped_count}{" "}
+          claim{props.dropped_count === 1 ? "" : "s"} dropped
```
pa-lem: very small nitpicking, maybe we could use `<= 1` instead of `=== 1` so if (for any reason) there is a bug and the api returns 0, it will be fine

pa-lem (approved):
screenshots could help document this PR so we can quickly see how it looks, but otherwise LGTM 👍

frontend/app/src/entities/events/ui/group-auto-create-events/group-auto-create-event-title.tsx
polmichel: You are right, thanks

PR #9368, opened 2026-05-27
rewrite path traversal query renderer to improve performance

ajtmccarty:
# Why

PathTraversalQuery and ReachableNodesQuery share logic for writing cypher. They both
do similar thing
- PathTraversalQuery finds all the paths from one object to another
- ReachableNodesQuery finds all paths from one object to all objects of a given kind (or kinds)

Unfortunately, the ReachableNodesQuery cypher did not perform very well b/c it was checking too many
paths. The PathTraversalQuery works fine b/c it has one start vertex and one end vertex, both
identifiable by their UUIDs, which are indexed.

Improving the performance of the ReachableNodesQuery required adding some new generated cypher to
find all the possible terminal objects for the query, filtering and limiting them as the paths get
deeper to prevent an exponential path explosion. Then, we use the more-or-less existing cypher
query to find all the possible paths from source to the terminals. This has the added benefit of
ensuring that we show all paths (according to the filters) to every terminal object (up to a maximum)
instead of potentially only showing some of the paths to each terminal after ordering and limiting. 

The cypher generation used to have just one parameter for setting a maximum result size, but now there
are two
- max_targets: maximum number of terminal objects (always 1 for PathTraveral, defaults to 50 for ReachableNodesQuery)
- max_paths: total number of maximum paths returned for all terminal objects (defaults to 500)

The new cypher generation uses a two-phase approach:
- Phase 1: find the source and target(s). easy if there is only 1 target
- Phase 2: find all the paths from source to target(s)

# Performance Comparison

From the local performance testing I had Claude do on the original version of these queries vs the updated versions in this PR. Keep in mind that the original versions also returned fewer results.

### PathTraversalQuery

| depth | old median ms | new median ms | speedup | old paths | new paths |
|---|---|---|---|---|---|
| 3 | 188 | 142 | 1.3x | 1 | 1 |
| 4 | 2113 | 957 | 2.2x | 10 | 2 |
| 5 | 140030 | 80850 | 1.7x | 10 | 10 |

NOTE: the old version of the query would include paths that returned to the source object, which is why the old version finds 10 paths at depth 4 and the new version finds only 2. 8 of those paths returned to the source object.

### ReachableNodesQuery

| depth | old median ms | new median ms | speedup | old paths | new paths | old terminals | new terminals |
|---|---|---|---|---|---|---|---|
| 3 | 208 | 171 | 1.2x | 50 | 111 | 22 | **50** |
| 4 | 2112 | 1608 | 1.3x | 50 | 286 | 22 | **50** |
| 5 | 213509 | 80464 | 2.7x | 50 | 500 | 22 | **50** |

## Headline results

- **RNQ at depth=5**: new is **~2.7x faster** (80s vs 214s) and delivers all **50** distinct
  terminals (vs old's **22**). The old impl saturated its 50-path budget on alternate
  paths to popular terminals, missing 28 of the 50 targets entirely.
- **RNQ at every depth**: new delivers full 50-terminal coverage; old caps at 22 terminals.
  This is the core semantic win, independent of wall time.
- **PTQ at depth=5**: new is **~1.7x faster** (81s vs 140s) at the same 10-path cap.
- **At shallow depths (3, 4)**: new is also faster (10–55% across both queries),
  ruling out a "new shape only wins on the slow case" concern.
- **Path-row counts differ** between old and new because the cap semantics differ:
  new RNQ returns up to `max_paths=500` rows distributed across `max_targets=50`
  terminals; old RNQ returns up to `max_results=50` total rows. The terminal-coverage
  column is the apples-to-apples comparison.

### Behavioral changes

- `ReachableNodesQuery` now returns paths to up to `max_targets` *distinct
  terminal nodes* (Phase 1 discovery cap), capped overall at `max_paths` rows
  (Phase 2 path-safety cap). Previously a single `max_results` capped raw path
  rows, which silently dropped terminals when popular ones had many alternate
  paths.
- New `max_paths` field on `ReachableNodesInput` (default 500, max 5,000) — gives
  callers control over the path-safety cap.
- `ReachableNodesInput.max_results` keeps its name but its semantics are now
  "maximum distinct terminal nodes to discover" (it maps to `max_targets`
  internally). The field description is updated to reflect that.
- `PathTraversalQuery` is unchanged from a caller perspective: its `max_paths`
  field already meant "max paths returned"; the new renderer just skips Phase 1
  and anchors `terminal_uuids = [target.uuid]`.

### Implementation notes

- `PathTraversalCypherRenderer` is rewritten around a two-phase rendered shape:
  - **Phase 1** (ReachableNodes only): per-depth chained `CALL (var) { ... RETURN DISTINCT ... }`
    subqueries, each separated by `WITH ... ORDER BY kind, uuid LIMIT $max_targets`
    to keep intermediate cardinality bounded; results UNION-ed and capped at
    `$max_targets` distinct terminals before `collect(target.uuid) AS terminal_uuids`.
  - **Phase 2** (both query types): per-depth fixed-length `MATCH` UNION ALL
    branches anchored on `source` and `target.uuid IN terminal_uuids`, capped
    at `$max_paths`. Phase 2 is byte-identical between ReachableNodes and PathTraversal.
  - `TerminalById` (PathTraversal) skips Phase 1 entirely: anchors the target by uuid,
    builds the single-element `terminal_uuids` list inline, and runs only Phase 2.
- Renderer API split between long-lived deps and per-request inputs:
  `__init__(*, branch, default_branch_name, at, max_targets, max_paths)` now
  validates bounds (`max_targets ∈ [1, 200]`, `max_paths ∈ [1, 5_000]`).
  `render(*, plan, source_id) -> RenderedCypher` is the per-call entry point and
  dispatches on `isinstance(plan.terminal_predicate, TerminalById)`.
- Branch envelopes (`_REACHABLE_NODES_ENVELOPE`, `_PATH_TRAVERSAL_ENVELOPE`) and
  hop predicates (`_EDGE_ACTIVE_PREDICATE`, `_HOP_TUPLE_PREDICATE`,
  `_DELETION_SHADOW_PREDICATE`) are template constants; per-hop rendering
  uses `_render_phase_one_branch` and `_render_phase_two_branch`.
- Phase 1 source-excludes the destination on every hop (including the terminal)
  so `$source_id` can never enter `terminal_uuids`. Phase 2 source-excludes
  only intermediates and adds an `intermediate != target` predicate, since the
  terminal is already constrained by `target.uuid IN terminal_uuids`.
- Test helpers (`build_reachable_query`, `build_path_traversal_query`) updated
  to construct the renderer with the new `max_targets`/`max_paths` parameters;
  PathTraversal helper passes `max_targets=1` internally (the param is emitted but unused
  in PathTraversal Cypher — kept for a uniform param surface).
- New unit-test module `backend/tests/unit/graph_traversal/test_cypher.py`
  covers bound validation, empty-plan rejection on `render()`, and dispatch
  shape (ReachableNodes emits two `CALL` blocks; PathTraversal skips Phase 1, anchors `target` by
  uuid, builds `[target.uuid] AS terminal_uuids`).

### What stayed the same

- No schema or migration changes.
- `Plan`, `SchemaPlanner`, `TerminalByKinds`/`TerminalById` contracts unchanged.
- `ReachableNodesQuery.init()` / `PathTraversalQuery.init()` surface unchanged
  except that the renderer is now constructed by the caller and injected
  (`renderer=` kwarg). The Query subclasses just call `renderer.render(plan,
  source_id)` from `query_init`.
- The `max_results` GraphQL field name on `ReachableNodesInput` is preserved
  (semantics updated; description rewritten).
- Component tests in `backend/tests/component/graph_traversal/` use small
  fixtures with `>=1` / `==1` assertions and need no modification — the new
  query produces the same correctness on simple inputs.

### Suggested review order

1. `backend/infrahub/graph_traversal/_cypher.py` — the rewrite; start with the
   module docstring and `render()`, then read `_render_phase_one_branch` and
   `_render_phase_two_branch`.
2. `backend/infrahub/graphql/queries/reachable.py` — the only caller with
   semantic changes (new `max_paths` input field; `max_results → max_targets`
   wiring).
3. `backend/infrahub/graphql/queries/path.py` — minimal: just constructs the
   renderer with `max_targets=1`.
4. `backend/tests/unit/graph_traversal/test_cypher.py` and
   `backend/tests/helpers/graph_traversal/builders.py` — confirm the new
   surface is exercised.
5. The query module changes (`reachable.py`, `path.py` under
   `graph_traversal/`) are mechanical: drop the inline renderer construction
   and accept it via `__init__`.

## How to review

Focus areas:

- The two-phase Cypher shape and its branch-conditional pieces. In particular,
  the source-exclude / intermediate-exclude predicate flags (`source_exclude_on_target`,
  `intermediate_exclude_target`) differ between Phase 1 and Phase 2 — there's a
  comment explaining why, but worth sanity-checking against the rendered text
  in the dispatch tests.
- The semantic shift on `ReachableNodesInput.max_results`: existing callers
  that pass `max_results=200` will now receive paths to up to 200 distinct
  terminals (potentially many more total path rows than before, up to
  `max_paths`). The field name is preserved, but the response shape grows for
  callers that don't set `max_paths` explicitly.
- The PathTraversal-side emission of `$max_targets=1` even though the parameter is not
  referenced in the rendered PathTraversal Cypher — kept for a uniform param surface;
  flagged in a comment.

Risks:

- Plan-time complexity at small fixtures: the new query has more clauses (two
  nested `CALL` subqueries). Spike measurements showed depth=2 took ~5.5s wall
  against tiny inputs, mostly Cypher planning, not execution (~3460 dbHits).
- Inner `$max_targets` cap can still drop intermediate kinds from a depth-`d`
  Phase 1 branch when the chained DISTINCT cardinality exceeds the cap. The
  strict "every path to a discovered terminal is returned" guarantee holds
  modulo this cap; documented in the renderer module docstring.

---
## Summary by cubic
Rewrote the path traversal into a two‑phase Cypher renderer to speed up terminal discovery and return full paths without path explosion. Adds loop prevention, separate caps for terminal discovery and total paths, and updates GraphQL inputs.

- **New Features**
  - Added `PathTraversalCypherRenderer` with two phases: discover up to `max_targets` terminals, then enumerate paths up to `max_paths`.
  - Prevents looped paths in Phase 2 by excluding repeats of earlier intermediates.
  - `ReachableNodesQuery` returns paths to up to `max_targets` terminals and up to `max_paths` total paths. `ReachableNodesInput` adds `max_paths` (default 500, max 5000); `max_results` now controls terminal discovery.
  - `PathTraversalInput.max_paths` now allows up to 5000.
  - Queries convert renderer init errors to GraphQL errors for clearer client feedback.
  - Consistent branch-aware deletion-shadow checks across all hops and phases.

- **Refactors**
  - Renderer API: constructor now takes branch info only; `render(plan, source_id, at, max_targets, max_paths)` validates bounds and dispatches. `PathTraversalQuery` uses `max_targets=1`.
  - Planner simplifies plan building; `Plan` accumulates hops and exposes per-kind min-depths via accessors.
  - GraphQL resolvers instantiate and inject the renderer; tests updated with builders and a renderer unit test.

<sup>Written for commit 893c51126d918ef311c8feb4609b5b2cfe85c1d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9368?utm_source=github">Review in cubic</a>


backend/infrahub/graph_traversal/planning/models.py
```diff
@@ -60,25 +57,70 @@ def from_graphql_input(
...
     terminal_predicate: TerminalPredicate
     max_depth: int
+    _adjacency: dict[str, dict[str, set[str]]] = field(default_factory=dict, init=False, repr=False)
+    _min_depth_to_terminal: dict[str, int] = field(default_factory=dict, init=False, repr=False)
 
     def __post_init__(self) -> None:
         if not MIN_DEPTH <= self.max_depth <= MAX_DEPTH:
             raise ValueError(f"Plan.max_depth must be in [{MIN_DEPTH}, {MAX_DEPTH}], got {self.max_depth}")
 
+    def add_hop(
```
ajtmccarty, line 86: I thought methods to set and get these data were more understandable than the raw `_adjacency` and `_min_depth_to_terminal` dicts

backend/infrahub/graph_traversal/_cypher.py
ajtmccarty, line 1: sadly, I cannot find a way to write this dynamic cypher generation that makes it any easier to understand than this. the path and filtering construction has to be almost completely dynamic based on the shape of the schema, so all the cypher is broken up into little chunks that then get added together

backend/infrahub/graphql/queries/reachable.py
```diff
@@ -47,7 +48,16 @@ class ReachableNodesInput(InputObjectType):
...
+    max_results = Int(
+        required=False,
+        default_value=50,
+        description="Maximum number of distinct terminal nodes to discover (default: 50, max: 200)",
+    )
+    max_paths = Int(
+        required=False,
+        default_value=500,
+        description="Maximum total paths returned across all discovered terminals (default: 500, max: 5000)",
+    )
```
ajtmccarty, line 60: @pa-lem FYI. I've updated the description of the `max_results` param on the `InfrahubReachableNodes` query and added a new `max_paths` param. not sure if that requires any changes on the front end

backend/infrahub/graph_traversal/planning/models.py
```diff
@@ -60,25 +57,70 @@ def from_graphql_input(
...
+        return sorted(kind for kind, dist in self._min_depth_to_terminal.items() if dist <= max_hops)
+
+    def get_all_source_kinds(self) -> list[str]:
+        """Sorted list of kinds that have at least one outgoing relationship recorded in this plan."""
+        return sorted(self._adjacency)
+
     @property
     def is_empty(self) -> bool:
-        return not self.adjacency
+        return not self._adjacency or not self._min_depth_to_terminal
```
polmichel, line 126: This is just an extra safety related to future developments right? Since currently the only way to add an element to the `_min_depth_to_terminal` is to call `infrahub.graph_traversal.planning.models.Plan.add_hop` which fills `_adjacency`

backend/infrahub/graph_traversal/planning/planner.py
```diff
@@ -157,23 +157,26 @@ def plan(
...
+        for start_kind, rels in adjacency.items():
+            for rel_name, end_kinds in rels.items():
+                for end_kind in end_kinds:
+                    new_plan.add_hop(
+                        source_kind=start_kind,
+                        relationship_identifier=rel_name,
+                        end_kind=end_kind,
+                        min_hops_to_terminal=min_depth_to_terminal[end_kind],
+                    )
+        return new_plan
```
polmichel: This is an optional comment/thought:

The `add_hop` method is a clean way of specifying how the internal collection should be built, and I think it's clearer this way. 

That said, I feel this specific part of the code becomes harder to understand: a already filtered `min_depth_to_terminal` collection is built, then passed element by element to a method that filters it based on the `adjacency` collection, which will be finally assigned to internal `Plan` values that are named almost the same. Moreover, this makes the `Plan` object somewhat mutable, and it can't be used as-is right after instantiation.

What do you think about adding an additional object which would contain these shrank collections, and would be given to the `Plan`? It could be a `PlanGraph` which would filters the non-minimal paths, or a `MinimalPlanGraph` which would verify that the state is correct or throw an error. This object would be built from these collections and then passed to the `Plan` object constructor that would store it as an attribute. If `not self._permission_cache.can_view(source_kind):` we can return it empty. 

This may be over engineered and duplicate some logic, and I may have missed something, so feel free to ignore this comment

backend/infrahub/graph_traversal/_cypher.py
```diff
@@ -95,98 +109,423 @@ class RenderedCypher:
...
+                return None
+            tuples_per_hop.append(hop_tuples)
+        return tuples_per_hop
+
+    def _render_phase_one_branch(self, branch_data: _DepthBranchData, *, terminal_label_union: str) -> str:
+        """Phase 1 per-depth branch: chained DISTINCT-capped CALLs returning ``target``.
+
+        Each hop is a separate ``CALL (var) { ... RETURN DISTINCT bN }``
+        subquery. Between hops, a ``WITH … ORDER BY bN.kind, bN.uuid LIMIT $max_targets``
+        clause caps intermediate cardinality so the chain can't multiply out.
```
polmichel, line 311: I was writing some comments regarding the edge case around `max_level` limit when I read once again the PR description, and you have explicitly written some text regarding this in the risks section of the PR.
So from what I understand, the ordering logic guarantees that the query is deterministic, returns the elements and always cut it at the `max_level` limit. 

Maybe at some point, a future enhancement could be interesting to notify the user of the maximum limit has been reached, with a dedicated boolean which we can switch to `True` whenever we hit a number of possible targets > the maximum limit.

backend/infrahub/graph_traversal/_cypher.py
```diff
@@ -95,98 +109,423 @@ class RenderedCypher:
...
+                          -[r2_s]-(rel2:Relationship)-[r2_e]-(b2:K1|...)
+                          -[r3_s]-(rel3:Relationship)-[r3_e]-(target:TerminalKinds)
+            WHERE target.uuid IN terminal_uuids
+              AND <edge active for r1_s, r1_e>
+              AND <hop1 triple>
+              AND b1.uuid <> $source_id AND b1 <> target
+              AND <deletion shadow per hop, on user branches>
+              AND <edge active for r2_s, r2_e> AND <hop2 triple>
+              AND b2.uuid <> $source_id AND b2 <> target
+              AND <edge active for r3_s, r3_e> AND <hop3 triple>
```
polmichel, line 412: If we have schemas allowing an object to be related to itself, or where one of the two objects is the same, couldn't that create a loop in this query?
If so, storing the discovered UUIDs and asserting `b1.uuid not in discovered_uuids` instead of` b1.uuid <> $source_id `would save some computation time? Or chaining the `b3 <> source AND b3 <> b2 AND b3 <> b1` ...

backend/infrahub/graphql/queries/reachable.py
```diff
@@ -61,6 +71,7 @@ async def reachable_nodes_resolver(
     target_kinds = list(data.target_kinds) if data.target_kinds else []
     max_depth = data.max_depth or 5
     max_results = data.max_results or 50
+    max_paths = data.max_paths or 500
```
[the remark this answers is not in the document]
polmichel: I guess `max_paths = 0` would not make any sense, but sending 0 and having the full maximal limit is maybe dangerous in weird use case.

backend/infrahub/graphql/queries/reachable.py
```diff
@@ -99,14 +110,20 @@ async def reachable_nodes_resolver(
 
     reachable_data: list[ReachableNodeData] = []
     if not plan.is_empty:
+        renderer = PathTraversalCypherRenderer(
+            branch=graphql_context.branch,
+            default_branch_name=registry.default_branch,
+            at=graphql_context.at,
+            max_targets=max_results,
+            max_paths=max_paths,
+        )
```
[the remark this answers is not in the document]
polmichel: I think the bot is right here: the preceding lines are wrapped in a try/except block that validates the `max_depth` parameter and raises a `ValueError` when it's outside the allowed range.
```
MIN_DEPTH = 1
MAX_DEPTH = 20
```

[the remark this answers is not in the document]
polmichel: Just a wild thought, nothing needs to be done here: 
Instead of checking it only at runtime, maybe, in some specific cases, having a new GraphQL type that's a capped and floored `Int` could help. Not sure if this idea has already been explored.

polmichel (commented):
The analysis is helpful and interesting.
One question on the PTQ case: the results show that at depth 4, the number of paths was reduced from 10 to 2. I'm not sure why this changed since it is not impacted by the `max_targets` risk/limitation.

I've left a few comments on the PR, nothing blocking. Just trying to understand the whole stuff.

backend/infrahub/graph_traversal/planning/models.py
ajtmccarty: yes that's right. if either of these is falsy, then no paths exist. so mostly a guard in case they are calculated differently later

backend/infrahub/graph_traversal/_cypher.py
ajtmccarty: yes that is correct. and I think some sort of "maximum reached" message would be a good future enhancement for this, but I'm going to leave it out for now

backend/infrahub/graph_traversal/_cypher.py
ajtmccarty: good point. I've added another predicate that prevents returning to the same uuid so that no loops are possible in the paths

backend/infrahub/graphql/queries/reachable.py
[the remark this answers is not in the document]
ajtmccarty: I've moved the `max_paths`, `max_targets`, and `at` into the `render()` method of `PathTraversalCypherRenderer` where they belong and wrapped the PathTraversal/ReachableNodes query inits in try-except blocks to catch the error

backend/infrahub/graph_traversal/planning/planner.py
ajtmccarty: I agree that this method is not easy to read. I don't want to make another intermediate object for this logic though. I want the `Plan` to be responsible for recording and organizing its data with methods for adding and retrieving so that we can be more sure that each piece of data added to the plan is correct. I've made a few changes to simplify the plan construction so that now it is built in place instead of generating `adjacency` and then passing it into `Plan` one item at a time

ajtmccarty (commented):
@polmichel thanks for the thorough review so far

backend/infrahub/graph_traversal/planning/models.py
polmichel: It makes sense!

backend/infrahub/graphql/queries/reachable.py
```diff
@@ -71,7 +71,7 @@ async def reachable_nodes_resolver(
     target_kinds = list(data.target_kinds) if data.target_kinds else []
     max_depth = data.max_depth or 5
     max_results = data.max_results or 50
```
polmichel, line 73: Optional comment, and out-of-scope regarding this PR: I think these variables would deserve the same modification than `max_paths`

polmichel (approved):
This looks great! Just an optional and out-of-scope comment. Thanks for the analysis and the thorough PR description and comments that helped with the review.

PR #9369, opened 2026-05-28
parser for new `order_by` grammar

ajtmccarty:
new parsing class and tests to support the new `order_by` options in the schema
- <attribute_name>[__<asc_or_desc>]
- <relationship_name>__<attribute_name>[__<asc_or_desc>]
- node_metadata__<created/updated__at>[__<asc_or_desc>]

existing logic in `SchemaBranch.validate_order_by` handles verifying that attributes on relationships actually exist in the schema

---
## Summary by cubic
Adds a new parser for schema `order_by` with support for metadata fields and explicit sort direction, addressing INFP-530. Also reserves `node_metadata` and tightens validation errors.

- **New Features**
  - Parse `order_by` entries:
    - `<attribute>__<property>[__asc|desc]`
    - `<relationship>__<attribute>__<property>[__asc|desc]`
    - `node_metadata__<created_at|updated_at>[__asc|desc]`
  - Defaults to ascending when direction is omitted.
  - Clear errors for unknown fields, bad direction tokens, undefined attributes, or empty segments.

- **Migration**
  - `node_metadata` is now a reserved name for attributes and relationships; rename if used.
  - Validation errors now state the reserved name and whether it was used on an attribute or relationship.

<sup>Written for commit 5e7025dac04f3039d2d065a0fd094997abbab738. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9369?utm_source=github">Review in cubic</a>

### DIFF
--- a/backend/infrahub/core/schema/order_by.py
+++ b/backend/infrahub/core/schema/order_by.py
@@ -67,9 +67,7 @@ class ParsedMetadataOrderBy(_ParsedOrderByBase):
         return (self.kind.value, self.metadata_field.value)
 
 
-type ParsedOrderByEntry = (
-    ParsedAttributeOrderBy | ParsedRelationshipAttributeOrderBy | ParsedMetadataOrderBy
-)
+type ParsedOrderByEntry = ParsedAttributeOrderBy | ParsedRelationshipAttributeOrderBy | ParsedMetadataOrderBy
 
 
 def parse_order_by_entry(entry: str, node_schema: BaseNodeSchema) -> ParsedOrderByEntry:
@@ -91,12 +89,8 @@ def _consume_direction(entry: str, parts: list[str], path_length: int) -> OrderD
         token = parts[-1]
         if token in _DIRECTION_TOKENS:
             return OrderDirection(token.upper())
-        raise ValueError(
-            f"invalid direction (entry: {entry!r}). Direction must be 'asc' or 'desc'."
-        )
-    raise ValueError(
-        f"invalid entry (entry: {entry!r}). Unexpected number of segments."
-    )
+        raise ValueError(f"invalid direction (entry: {entry!r}). Direction must be 'asc' or 'desc'.")
+    raise ValueError(f"invalid entry (entry: {entry!r}). Unexpected number of segments.")
 
 
 def _parse_metadata(entry: str, parts: list[str]) -> ParsedMetadataOrderBy:
@@ -112,9 +106,7 @@ def _parse_metadata(entry: str, parts: list[str]) -> ParsedMetadataOrderBy:
         metadata_field = OrderByMetadataField(field_token)
     except ValueError as exc:
         supported = ", ".join(field.value for field in OrderByMetadataField)
-        raise ValueError(
-            f"unknown metadata field (entry: {entry!r}). Supported metadata fields: {supported}."
-        ) from exc
+        raise ValueError(f"unknown metadata field (entry: {entry!r}). Supported metadata fields: {supported}.") from exc
 
     direction = _consume_direction(entry=entry, parts=parts, path_length=2)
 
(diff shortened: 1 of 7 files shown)


backend/infrahub/core/schema/order_by.py
```diff
@@ -0,0 +1,161 @@
...
+            relationship_name=parts[0],
+            attribute_name=parts[1],
+            property_name=parts[2],
+        )
+
+    if first in node_schema.attribute_names:
+        if len(parts) < 2:
+            raise ValueError(
+                f"invalid attribute path (entry: {entry!r}). "
+                f"Expected '<attribute>__<property>' with an optional direction suffix."
```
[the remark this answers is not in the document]
ajtmccarty: this is wrong. schema validation requires a property suffix for the order_by definition, which, at this point, is always `__value`

backend/infrahub/core/schema/order_by.py
```diff
@@ -0,0 +1,151 @@
...
+    def target_key(self) -> tuple[str, ...]:
+        return (self.kind.value, self.metadata_field.value)
+
+
+type ParsedOrderByEntry = ParsedAttributeOrderBy | ParsedRelationshipAttributeOrderBy | ParsedMetadataOrderBy
+
+
+def parse_order_by_entry(entry: str, node_schema: BaseNodeSchema) -> ParsedOrderByEntry:
+    if not entry:
+        raise ValueError(f"order_by entries must be non-empty strings (entry: {entry!r}).")
```
polmichel, line 73: does the `(entry: {entry!r})` is useful since it is empty?

backend/infrahub/core/schema/order_by.py
```diff
@@ -0,0 +1,151 @@
...
+
+    direction = _consume_direction(entry=entry, parts=parts, path_length=2)
+
+    return ParsedMetadataOrderBy(raw=entry, direction=direction, metadata_field=metadata_field)
+
+
+def _parse_path(
+    entry: str, parts: list[str], node_schema: BaseNodeSchema
+) -> ParsedAttributeOrderBy | ParsedRelationshipAttributeOrderBy:
+    first = parts[0]
```
polmichel, line 120: There's an implicit assumption here that `parts` contains at least one element, which relies on the state of the caller's data. That said, I think it's fine since this is a private method that probably won't be reused elsewhere.

backend/infrahub/core/schema/order_by.py
```diff
@@ -0,0 +1,151 @@
...
+    def target_key(self) -> tuple[str, ...]:
+        return (self.kind.value, self.attribute_name, self.property_name)
+
+
+@dataclass(frozen=True, slots=True)
+class ParsedRelationshipAttributeOrderBy(_ParsedOrderByBase):
+    relationship_name: str
+    attribute_name: str
+    property_name: str
+    schema_path: SchemaAttributePath | None = None
```
polmichel: I guess this `schema_path` will be used and filled later during the feature development

polmichel (approved):
LGTM! Just one optional comment and a few thoughts

backend/infrahub/core/schema/order_by.py
ajtmccarty: yes... although it should just be an empty string. but better to be sure, I suppose

backend/infrahub/core/schema/order_by.py
ajtmccarty: thanks for pointing this out. I actually don't think it will be. and it's definitely not used now. removing it

backend/infrahub/core/schema/order_by.py
ajtmccarty: true. I think this is okay b/c we know the caller and that the list will be at least length one

PR #9292, opened 2026-05-19
IFC-2620 refactor: restructure auth package and break dependency cycle

polmichel:
## Why
When implementing the next item of the authentication group auto-creation feature, which is about audit events (internal task: [IFC-2592](https://opsmill.atlassian.net/browse/IFC-2592)), a dependency cycle between the `auth` and `event` packages surfaced.
Part of [IFC-2521](https://opsmill.atlassian.net/browse/IFC-2521) and closes [IFC-2620](https://opsmill.atlassian.net/browse/IFC-2620).

## What
This is due to a specific object, `infrahub.context.InfrahubContext`, imported in `backend/infrahub/events/models.py`, which brings functional domain dependencies with it.

### Thoughts

- `events` should not rely on domain-specific package as it is used in domain-specific package to send the events.
- `infrahub.context.InfrahubContext` is a "god" class retrieving a lot of dependencies inside. It should be not be imported into `events` package which should keep minimal dependencies towards other packages.

[IFC-2592]: https://opsmill.atlassian.net/browse/IFC-2592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-2521]: https://opsmill.atlassian.net/browse/IFC-2521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-2620]: https://opsmill.atlassian.net/browse/IFC-2620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#### Split `infrahub.auth` 
<details>
  <summary>Folder structure: before vs after</summary>

  ```text
  BEFORE                                  AFTER
  ──────                                  ─────
  backend/infrahub/                       backend/infrahub/
  ├── auth.py        (~25 KB)             ├── auth/
  │   ├── AuthType                        │   ├── __init__.py     (re-exports)
  │   ├── AccountSession                  │   ├── auth.py         (remaining logic)
  │   ├── AnonymousSession                │   ├── session.py      (AccountSession,
  │   ├── ExternalAuthProtocol            │   │                    AnonymousSession)
[...]

PR #9372, opened 2026-05-28
Drop the auth package re-exports that reopened the import cycle

gmazoyer:
TL;DR Diff is big but it's only about moving to absolute imports.

PR #9325 added a top-level `from infrahub.context import InfrahubContext` in `auth/auth.py`. The earlier #9292 had restructured the auth package precisely to break a cycle of this shape, and #9325 reintroduced it. The cycle stays dormant in most entry points because they happen to load `auth.auth` to completion before anything touches `infrahub.context`, but it fires hard in a fresh process whose first auth-touching import is `infrahub.context`.

We caught it in the entreprise code for which `PREFECT_API_BLOCKS_REGISTER_ON_START=false` and `PREFECT_API_DATABASE_MIGRATE_ON_START=false` are set. This triggers `asyncio.run(_init_prefect())`, whose `from infrahub.services import InfrahubServices` is the first thing that reaches `infrahub.context`, and the import chain trips:

```
infrahub.context line 5
  from infrahub.auth.session import AccountSession
    triggers auth/init.py
      from infrahub.auth.auth import (...)
        auth.auth line 27: from infrahub.context import InfrahubContext
           context is mid-init at line 5
           ImportError: cannot import name 'InfrahubContext' from partially initialized module 'infrahub.context'
```

Community CI never set those envs so the branch stayed dark there; the bug ships green (-ish we all love flakies heh) from community and detonates the first time anyone runs a distributed-mode worker.

A tiny fix would have been to drop `InfrahubContext.init(...).to_event_context()` from `_build_auto_create_event_emitter` (its only caller in `auth.auth`) and build the same `EventContext` directly. This PR instead takes the structural route , so the next person who adds `from infrahub.context import X` to `auth.auth` does not reopen the cycle.

---
## Summary by cubic
Stops re-exporting from `infrahub.auth` to remove an import cycle with `infrahub.context` that could crash startup in some entry paths. Call sites now import directly from `infrahub.auth.session`, `infrahub.auth.auth`, and `infrahub.auth.types`.

- **Bug Fixes**
  - Fixed ImportError caused by `infrahub.context` -> `infrahub.auth.__init__` -> `infrahub.auth.auth` -> `infrahub.context` cycle.
  - Prevents indirect `infrahub.context` imports when loading auth types and helpers.

- **Migration**
  - Use `from infrahub.auth.session import AccountSession` (was `from infrahub.auth import AccountSession`).
  - Use `from infrahub.auth.types import AuthType` (was `from infrahub.auth import AuthType`).
  - Import auth helpers and types (e.g., `authentication_token`, `authenticate_with_password`, `ExternalIdentity`, `SSOStateCache`, `signin_sso_account`, `validate_auth_response`) from `infrahub.auth.auth`.

<sup>Written for commit 2e83ecea8fedf9b1cc6f0955ecb80acbea0a202b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9372?utm_source=github">Review in cubic</a>


backend/infrahub/auth/__init__.py
ajtmccarty, line 1: Image

ajtmccarty (commented):
looks good to me, but the tests remain sad

backend/infrahub/auth/__init__.py
gmazoyer: image

PR #9382, opened 2026-05-29
add metadata order by to GET cypher queries

ajtmccarty:
# Why

Today, schema authors can declare `order_by: ["name__value"]` on a schema, but they cannot ask for descending order or for ordering by node-level metadata (`created_at`, `updated_at`). For example, a `DocumentationNote` relationship that should always render newest-first on the parent detail page requires per-query ordering arguments at every callsite.

**Goal:** make `order_by: ["node_metadata__created_at__desc"]` (and the rest of the new grammar) honored consistently across the three list paths — top-level node list, relationship-peer list, and hierarchy list — including direction suffixes on regular attributes (`name__value__desc`).

**Non-goals:** stricter rejection messages for malformed `order_by` entries (US3, deferred) and US2 verification beyond what falls out of US1's direction propagation.

Closes IFC-2650

### Behavioral

- New `order_by` grammar accepted by the schema validator: `<attr>__<prop>[__<dir>]`, `<rel>__<attr>__<prop>[__<dir>]`, `node_metadata__<field>[__<dir>]` where `<field>` ∈ `{created_at, updated_at}` and `<dir>` ∈ `{asc, desc}`. Direction defaults to ascending when no suffix is present. Existing schemas continue to behave as ascending without modification.
- Direction on `<attr>` paths is now applied at query time. Affects `NodeGetListQuery`, `RelationshipGetPeerQuery`, and `NodeGetHierarchyQuery`.
- Schema-level `order_by: ["node_metadata__created_at__desc"]` is honored across all three list paths and through the matching GraphQL fields (top-level + nested relationship-peer).
- **Precedence change**: when a caller passes an explicit query-time `order` argument (REST/GraphQL/SDK), the schema's `order_by` is now ignored entirely. The UUID tiebreaker is appended once per query in all cases for deterministic ordering.

### Implementation notes

- `validate_order_by` routes metadata-prefixed entries (`node_metadata__…`) through the new parser and leaves attribute/relationship-attribute path entries to `validate_schema_path` so existing error messages ("`X` is not a valid attribute of `Y`", "cannot use `X` relationship") are preserved.
- The parser is strict on its own responsibilities (empty entry, unknown metadata field, malformed direction token, direction token in property slot like `account__name__desc`) and lenient on path existence that requires knowledge of other schemas (remains delegated to `validate_schema_path`).
- New shared helper `build_subquery_order_metadata(node_alias, metadata_field, …)` in `core/query/subquery.py` produces the inner Cypher for `CALL (<alias>) { … }` ordering by `created_at` / `updated_at`. Handles both default/global-branch shortcuts and the branch-aware traversal. Used by both `RelationshipGetPeerQuery` and `NodeGetHierarchyQuery`.
- `RelationshipGetPeerQuery` and `NodeGetHierarchyQuery` each gained a `_add_peer_order_by(...)` private method; their `query_init` ordering sections collapse to one helper call plus an unconditional `peer.uuid ASC` tiebreaker.
- `NodeMetaOrder` gained `__bool__` so an empty input (`order: {node_metadata: {}}`) doesn't override the schema's `order_by`.
- `NodeInheritanceHandler._update_order_by_for_renamed_attributes` skips entries that are metadata (via `is_metadata_order_by_entry`) so renames can't corrupt `node_metadata__*` references on the inheriting node.

### What stayed the same

- Wire shape: `order_by` is still `list[str] | None`.
- An empty list or an absent `order_by` behave identically to today — UUID-only sort, unchanged.
- `validate_schema_path` error messages for malformed paths are unchanged; the existing `test_validate_order_by_error` parametrized cases all still pass with their exact expected messages.

## Impact & rollout

- **Backward compatibility:** existing `order_by` declarations are unchanged in behavior (ascending by default, same paths accepted). The one observable change is the new replace-not-stack precedence when a caller passes a query-time `order` argument — previously the schema's entries would still apply as secondary sort, now they're dropped. Surfaces only for callers that set both schema `order_by` and a query-time `order`; the customary pattern (set one or the other) is unaffected.
- **Performance:** each new direction suffix adds at most one Cypher keyword (`ASC`/`DESC`) to the outer `ORDER BY`. Each new metadata entry on the relationship-peer / hierarchy paths adds one subquery comparable in cost to the existing top-level metadata subquery. UUID tiebreaker is appended once per query on a projected property; no extra traversal.

## Checklist

- [x] Tests added/updated (parser unit, schema-validator component, three query-layer component suites, GraphQL precedence, inheritance unit)

---
## Summary by cubic
Adds schema-level order_by for node metadata (`created_at`, `updated_at`) with ASC/DESC and consistent, directional ordering across node lists, relationship peers, and hierarchies. Query-time order now replaces schema order for top-level lists (REST/GraphQL); when no query-time order is given, schema order is honored (INFP-530).

- New Features
  - New syntax: `node_metadata__created_at[__asc|__desc]`, `node_metadata__updated_at[__asc|__desc]`.
  - Attribute/relationship `order_by` supports `__asc/__desc`; property segment required; non-`value` properties allowed.
  - Implemented in `NodeGetListQuery`, `RelationshipGetPeerQuery`, and `NodeGetHierarchyQuery` with per-entry direction and a UUID ASC tiebreaker; added `build_subquery_order_metadata` for branch-aware `created_at`/`updated_at`.
  - Parser-backed validation in schema; rejects direction tokens used as properties (e.g. `name__desc`); inheritance keeps metadata entries on rename. GraphQL top-level lists replace schema order when an `order` argument is provided; when absent, schema order is honored for top-level and relationship-peer lists.

- Bug Fixes
  - Fixed `NodeMetaOrder` truthiness so empty query-time `node_metadata` no longer suppresses schema `order_by`.

<sup>Written for commit e0da4e0ae1bc6971798ed3b34c92affa1221bf5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/schema/schema_branch.py
```diff
@@ -954,20 +955,26 @@ def validate_uniqueness_constraints(self) -> None:
...
             node_schema = self.get(name=name, duplicate=False)
 
             if not node_schema.order_by:
                 continue
 
-            allowed_types = SchemaElementPathType.ATTR_WITH_PROP | SchemaElementPathType.REL_ONE_ATTR_WITH_PROP
-            for order_by_path in node_schema.order_by:
-                element_name = "order_by"
+            for order_by_entry in node_schema.order_by:
+                if is_metadata_order_by_entry(order_by_entry):
```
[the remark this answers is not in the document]
ajtmccarty: I do not think this is right. the method attempts parsing metadata order_by entries within the conditional `if is_metadata_order_by_entry(order_by_entry):`, but non-metadata entries pass through the existing `self.validate_schema_path() call`

backend/infrahub/core/query/node.py
```diff
@@ -55,6 +60,7 @@
 # Grouped constants for validation/iteration
 METADATA_CREATED_FIELDS = (METADATA_CREATED_AT, METADATA_CREATED_BY)
 METADATA_UPDATED_FIELDS = (METADATA_UPDATED_AT, METADATA_UPDATED_BY)
+# TODO: import from constants and append the "__"
```
polmichel: Since this isn't explicitly stated in the _Non-goals_ section of the PR description, I'm just bringing this line up in case it's not meant to be left in this PR.

backend/tests/component/core/schema_manager/test_manager_schema.py
```diff
@@ -1145,7 +1145,7 @@ class SchemaBranchValidateNamesTestCaseData:
                         {"name": "name", "kind": "Text"},
                     ],
                     "relationships": [
-                        {"name": "node_metadata", "peer": "TestCriticality", "cardinality": "one"},
+                        {"name": "node_metadata", "peer": "Criticality", "cardinality": "one"},
```
polmichel, line 1148: I am not sure why this has changed since the test seems to pass with or without the change.

Moreover, that is true that the other upper schemas using this relationship have the same `peer` but it seems that the classic convention is to concatenate the `name` and the `namespace` (e.g: `TestCriticality`).

But I guess this does not really matter.

backend/infrahub/core/query/node.py
gmazoyer: Is that something missing in the PR, or just a note for later cleanup?

backend/infrahub/core/schema/node_inheritance_handler.py
```diff
@@ -104,7 +105,13 @@ def _update_default_filters_for_renamed_attributes(self, node: NodeSchema, renam
     def _update_order_by_for_renamed_attributes(self, node: NodeSchema, renamed_attrs: dict[str, str]) -> None:
         if not node.order_by:
             return
-        node.order_by = self._get_updated_renamed_attrs_data(attr_data=node.order_by, renamed_attrs=renamed_attrs)
+        updated_entries: list[str] = []
+        for entry in node.order_by:
+            if is_metadata_order_by_entry(entry):
+                updated_entries.append(entry)
+                continue
+            updated_entries.extend(self._get_updated_renamed_attrs_data(attr_data=[entry], renamed_attrs=renamed_attrs))
```
gmazoyer: `_get_updated_renamed_attrs_data` accepts a list, could we avoid multiple calls to it by going through the `for` loop then sending a list of entries to `_get_updated_renamed_attrs_data`?

polmichel (commented):
Review is not finished yet on my side, but still posting intermediate comments instead of keeping them locally.

backend/infrahub/core/query/node.py
ajtmccarty: yes this was leftover, I've addressed it now

ajtmccarty: yes this was leftover, I've addressed it now

backend/infrahub/core/schema/node_inheritance_handler.py
ajtmccarty: good point. moved the `is_metadata_order_by_entry` check down into the lower loop so that order of the entries is preserved

backend/tests/component/core/schema_manager/test_manager_schema.py
ajtmccarty: I think the name didn't matter b/c this is actually a unit test, but better to actually use the correct name

polmichel (approved):
LGTM!

PR #9385, opened 2026-05-29
fix(ci): replace chromium netlink binary patch with LD_PRELOAD shim, upgrade Playwright to 1.60.0

fatih-acar:
## Summary

Two commits:

1. **`ci:` replace the chromium netlink binary patch with an `LD_PRELOAD` shim.** The E2E job worked around `ERR_NETWORK_CHANGED` flake (Docker/veth netlink churn on the runners) by patching the installed chrome binary: `objdump` to locate `net::internal::AddressTrackerLinux::ReadMessages(...)`, then `dd` an `0xc3` (x86 `RET`) over its first instruction. That offset detection depends on an exact unstripped mangled symbol and the `--no-shell` binary layout — both of which newer Playwright/Chromium builds change, so the patch breaks. Replaced with an `LD_PRELOAD` shim (`frontend/app/tests/e2e/fixtures/no_netlink.c`, ported from styrmin) that fails `NETLINK_ROUTE` sockets with `EAFNOSUPPORT` at the libc level, so `AddressTrackerLinux::Init` takes its graceful "assume always online" path. Version-independent — it never inspects the chromium binary.

2. **`test:` upgrade Playwright `1.56.1 → 1.60.0`** (latest stable): `@playwright/test` + the `playwright`/`playwright-core` pnpm overrides, refreshed lockfile, and the AGENTS.md tech-stack note.

Ordered shim-first so the version bump doesn't ride on the already-broken `dd` patch.

## Implementation notes

- The shim is compiled in CI (`gcc -shared -fPIC … -ldl`) and exported as `NO_NETLINK_SO`. `playwright.config.ts` injects it via `launchOptions.env` scoped to the **browser process only**, so it doesn't leak into node/pnpm subprocesses (whose libuv interface enumeration also uses `NETLINK_ROUTE`).
- The lockfile change is surgical — only the Playwright version + 3 integrity hashes (20 lines). A plain `pnpm install` additionally re-tidied unrelated `schema-visualizer`/`@xyflow` entries (pnpm 10.33 cosmetic reformatting), which `--frozen-lockfile` does not require; that churn was kept out.

## Verification

- Shim compiles clean and exports `socket`.
- `playwright.config.ts` loads and enumerates 245 tests with and without `NO_NETLINK_SO`; Biome clean.
- `pnpm install --frozen-lockfile` passes and resolves `@playwright/test 1.60.0`; `pnpm exec playwright install chromium --no-shell` downloads the 1.60.0 build (Chrome for Testing 148).
- Local checks ran on arm64; CI runs x86_64 (the swap is arch-independent). Authoritative check is the `E2E-testing-playwright` job.

---
## Summary by cubic
Replaced the brittle Chromium binary patch with an `LD_PRELOAD` shim to stop netlink churn from causing E2E flakes, and upgraded Playwright to 1.60.0.

- **Bug Fixes**
  - Removed the `objdump`/`dd` Chrome patch; build a `no_netlink.so` shim in CI and export `NO_NETLINK_SO`.
  - Inject the shim via `launchOptions.env.LD_PRELOAD` in `playwright.config.ts`, scoped to the browser process only.
  - The shim blocks `NETLINK_ROUTE` sockets so Chromium assumes “always online,” eliminating `ERR_NETWORK_CHANGED` flakes.

- **Dependencies**
  - Upgraded `@playwright/test` to `1.60.0` with `playwright`/`playwright-core` overrides.
  - Updated the lockfile and the testing version in AGENTS.md.

<sup>Written for commit 2a321bbbd0886d5ba2c88a1c5e79093e51d90bbc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9385?utm_source=github">Review in cubic</a>

### DIFF
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -115,7 +115,7 @@
     "@graphql-codegen/cli": "^6.3.1",
     "@graphql-codegen/typescript": "^5.0.10",
     "@graphql-codegen/typescript-operations": "^5.1.0",
-    "@playwright/test": "1.56.1",
+    "@playwright/test": "1.60.0",
     "@types/apollo-upload-client": "18.0.1",
     "@types/node": "^25.6.0",
     "@types/prismjs": "^1.26.6",
@@ -140,8 +140,8 @@
       "playwright-core"
     ],
     "overrides": {
-      "playwright": "1.56.1",
-      "playwright-core": "1.56.1"
+      "playwright": "1.60.0",
+      "playwright-core": "1.60.0"
     },
     "peerDependencyRules": {
       "allowedVersions": {
--- a/frontend/app/tests/e2e/fixtures/no_netlink.c
+++ b/frontend/app/tests/e2e/fixtures/no_netlink.c
@@ -1,0 +1,28 @@
+// LD_PRELOAD shim that fails NETLINK_ROUTE sockets with EAFNOSUPPORT, so
+// Chromium's AddressTrackerLinux::Init takes its graceful "assume always
+// online" path (the same path FreeBSD's Linuxulator triggers). This avoids
+// netlink flake on CI runners where Docker/veth churn floods host netlink
+// and causes ERR_NETWORK_CHANGED mid-navigation.
+//
+// Build: gcc -shared -fPIC -o no_netlink.so no_netlink.c -ldl
+// Inject via Playwright launch env: {"LD_PRELOAD": "/path/to/no_netlink.so"}.
+
+#define _GNU_SOURCE
+
+#include <dlfcn.h>
+#include <errno.h>
+#include <linux/netlink.h>
+#include <stddef.h>
+#include <sys/socket.h>
+
+int socket(int domain, int type, int protocol) {
+    if (domain == AF_NETLINK && protocol == NETLINK_ROUTE) {
+        errno = EAFNOSUPPORT;
+        return -1;
+    }
+    static int (*real_socket)(int, int, int) = NULL;
+    if (!real_socket) {
+        real_socket = dlsym(RTLD_NEXT, "socket");
+    }
+    return real_socket(domain, type, protocol);
+}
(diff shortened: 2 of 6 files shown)


frontend/app/package.json
```diff
@@ -140,8 +140,8 @@
       "playwright-core"
     ],
     "overrides": {
-      "playwright": "1.56.1",
-      "playwright-core": "1.56.1"
+      "playwright": "1.60.0",
+      "playwright-core": "1.60.0"
```
bilalabbad, line 144: no need to overrides anymore since we can be on latest version I think

frontend/app/tests/e2e/fixtures/no_netlink.c
```diff
@@ -0,0 +1,28 @@
...
+    if (domain == AF_NETLINK && protocol == NETLINK_ROUTE) {
+        errno = EAFNOSUPPORT;
+        return -1;
+    }
+    static int (*real_socket)(int, int, int) = NULL;
+    if (!real_socket) {
+        real_socket = dlsym(RTLD_NEXT, "socket");
+    }
+    return real_socket(domain, type, protocol);
+}
```
bilalabbad, line 28: just curious if it should be part of frontend folder or if it's forgotten file

frontend/app/package.json
fatih-acar: I think it's still good to have control on the playwright version used to avoid unknown flakiness, etc

frontend/app/tests/e2e/fixtures/no_netlink.c
fatih-acar: This file is currently used for the chromium workaround on our CI runners, it's only used during e2e tests that's why it's in the frontend folder, but we can move it somewhere else if you want?

bilalabbad (approved):

PR #9391, opened 2026-05-29
IFC-2664 lock password changes for externally authenticated accounts

gmazoyer:
## Why

Any account provisioned through an external identity provider (LDAP, OIDC, OAuth2) could overwrite its local password through the self-update mutation, then sign in via the local-auth endpoint with the chosen password. The bypass survived directory-side revocation. This PR closes that gap on both the security side (backend) and the UX side (the frontend never lets the user reach the form).

Found while exercising LDAP against a live Active Directory; the bug applies to every account linked to an external identity, not just LDAP ones.

## What changed

The self-update mutation now refuses to write the password when the account is linked to an external identity, while continuing to accept description-only updates. The current account profile gains a boolean flag so the frontend can act before any submission. The synthesized field is added on the account interface.

Three alternative shapes for the signal can be considered (whitelisting the relationship in the schema-gen filter, moving the underlying type out of the Internal namespace, encoding the source in JWT claims).

The profile page reads the new flag, hides the Password tab for externally managed accounts, and shows an "externally managed" panel on the password page for direct navigation. The earlier reactive error-string workaround is gone.

image

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Blocks local password changes for accounts linked to external identity providers and hides the password UI to prevent SSO bypass. Adds an `is_externally_managed` flag to GraphQL account types and `AccountProfile` so the frontend can gate the form (IFC-2664).

- **Bug Fixes**
  - Backend: `InfrahubAccountSelfUpdate` rejects `password` updates when an `ExternalIdentity` exists; `description` remains allowed. Adds `is_externally_managed: Boolean!` on `CoreGenericAccount`/`CoreAccount` (resolver checks for a linked external identity) and exposes it on `AccountProfile`.
  - Frontend: Reads `is_externally_managed` to hide the Password tab; shows a “Password managed externally” panel on direct navigation; adds a loading state while fetching the profile.
  - Tests & Changelog: Adds tests for password rejection, allowed description updates, and UI gating/loading. Adds a security changelog entry.

<sup>Written for commit dd96d564c8d6a701439361e827fb10aff4e40856. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9391?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graphql/manager.py
```diff
@@ -507,8 +508,24 @@ def generate_query_mixin(self) -> type[object]:
...
+                    ),
+                    resolver=is_externally_managed_resolver,
+                )
+                node_type._meta.fields["is_externally_managed"] = is_externally_managed_field
+                # Mirror the field on every concrete schema that inherits from
+                # CoreGenericAccount so graphene's interface contract holds.
+                for implementer_kind in node_schema.used_by:
+                    self.get_type(name=implementer_kind)._meta.fields["is_externally_managed"] = (
+                        is_externally_managed_field
+                    )
```
gmazoyer, line 528: Not sure this is the best way. But this shows a possible fix. This can evolve though depending on what we decide next.

ajtmccarty: I wonder if it is more Infrahub-ish (it's like Pythonic, but for Infrahub) to make this a read-only field on the GenericAccount schema that we manage during user create/update. I'm really not sure what the best approach is. maybe we need a better framework for this kind of API-only virtual field. or maybe I just don't like how our Graphql queries/mutations are built

backend/tests/component/graphql/test_core_account.py
```diff
@@ -55,6 +55,99 @@ async def test_everyone_can_update_password(db: InfrahubDatabase, default_branch
...
+    query = (
+        """
+    mutation {
+        InfrahubAccountSelfUpdate(data: {password: "%s"}) {
+            ok
+        }
+    }
+    """
+        % rejected_password
+    )
```
ajtmccarty: you could use a graphql variable for this

backend/tests/component/graphql/test_core_account.py
```diff
@@ -55,6 +55,99 @@ async def test_everyone_can_update_password(db: InfrahubDatabase, default_branch
...
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={},
+    )
+
+    assert result.errors is not None
+    assert len(result.errors) == 1
+    message = result.errors[0].message.lower()
+    assert "password" in message
+    assert "external" in message
```
ajtmccarty: I think we agreed to have a more precise match on error strings, but maybe I'm wrong

ajtmccarty (approved):
backend part looks good, I think. actually, I am not sure if we have a good way to add this kind of virtual field right now, but this seems reasonable. it seems like we're constrained by how our graphql library works.
also, could/should this target `stable`? not sure if it is an issue there too

bilalabbad (approved):

PR #9396, opened 2026-05-29
Add feature flag to gate SSO display-name account fallback (IFC-2662)

polmichel:
## Summary

Adds the `security.sso_account_name_fallback` setting (env `INFRAHUB_SECURITY_SSO_ACCOUNT_NAME_FALLBACK`, default **enabled**) to gate whether an SSO login with no linked identity may claim a pre-existing account that matches by display name.

- **Enabled (default):** a first SSO login adopts a same-named account, as long as it has not already been linked to another identity — preserving the upgrade/migration path.
- **Disabled:** such a login always provisions a separate account (named by email, or a hard error if the email is also taken) instead of reusing an existing one.

This lets operators turn off the transitional name-based adoption once all SSO users have completed their first login. Full removal of the fallback is tracked separately for 1.11.

Part of [IFC-2662](https://opsmill.atlassian.net/browse/IFC-2662)

---
## Summary by cubic
Adds `security.sso_account_name_fallback` (`INFRAHUB_SECURITY_SSO_ACCOUNT_NAME_FALLBACK`, default true) to gate display‑name account adoption on first SSO login, registers it in `docker-compose.yml` and docs, and tightens the identity lookup to use `account__id`. When disabled, we create a separate account using the email; if both display name and email are taken the login fails, reducing spoofing risk per IFC-2662 ahead of removal in 1.11.

- **Refactors**
  - Use scalar `account__id` filter in SSO signin and tests; add test to ensure unrelated identities don’t block adoption.

- **Migration**
  - Keep enabled until all SSO users complete first login.
  - Then set `INFRAHUB_SECURITY_SSO_ACCOUNT_NAME_FALLBACK=false` to harden; with the flag off we create by email, and fail if both names are taken.

<sup>Written for commit 366fbeb30dd292e4e15c0a0ed2e5c15049a6cf0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2662]: https://opsmill.atlassian.net/browse/IFC-2662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/auth.py
```diff
@@ -244,35 +244,39 @@ async def signin_sso_account(  # noqa: PLR0915
...
+            name_fallback_enabled = config.SETTINGS.security.sso_account_name_fallback
+
+            # A pre-existing same-named account may be adopted on a first login, but only as a
+            # transitional convenience and only while it has never been linked to an SSO identity.
+            adopt_existing_account = False
+            if account_by_name and name_fallback_enabled:
                 existing_identities = await NodeManager.query(
                     db=db,
                     schema=InfrahubKind.EXTERNALIDENTITY,
                     filters={"account__ids": [account_by_name.id]},
```
ajtmccarty: should this be `"account_id": account_by_name.id`?

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -238,6 +248,88 @@ async def test_name_and_email_collision_raises_processing_error(
...
+    assert len(new_identities) == 1
+
+
+async def test_name_fallback_disabled_both_names_taken_raises(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: SchemaBranch,
+    sso_account_name_fallback_disabled: None,
+) -> None:
+    """With the fallback disabled a display-name match is not adopted, so a separate account is.
```
ajtmccarty: this is just half a sentence with a period in the middle

ajtmccarty (approved):

ogenstad:
Small comment here is that we'd also want to backport this to 1.8.x and release a 1.8.7. It could be fine to have this PR target stable and then create a new branch from the 1.8.6 tag and cherry pick.

backend/tests/component/core/test_signin_sso_account.py
polmichel: Thank you, fixed.

backend/infrahub/auth.py
polmichel: Yes it should, good point. 
I think it still works as this is covered by the tests: the tests fail when `account_ids` is replaced by `account__id`.
But better to explicitly use that since this is a one-to-one relationship

PR #9397, opened 2026-05-30
support ordering in GraphQL many relationships, respect order of ordering clauses

ajtmccarty:
# Why

Adds tests for various ordering combinations for NodeGetListQuery, RelationshipGetPeerQuery, and NodeGetHierarchyQuery.

Adds support for ordering to many relationships within the graphql resolver. See the new `test_graphql_relationship_peer_query_order_replaces_schema_order` for an example.

Fixes a bug in `NodeGetListQuery` that caused metadata ordering to always be applied first regardless of the ordering defined on the schema or in the graphql request.

Non-goals: no changes to the wire shape of `order_by`, no changes to the parser grammar, no new metadata fields beyond `created_at` / `updated_at`.

Closes IFC-2561

## What changed

Behavioral:

- `NodeGetListQuery` now emits its outer `ORDER BY` clauses in the schema-declared order, even when entries mix metadata and attributes. Pure-attribute and pure-metadata schemas are unchanged.
- The GraphQL many-relationship `order` argument is now honored. When provided, it fully replaces the peer schema's `order_by` (no stacking).

Implementation:

- Added a `schema_order_position` field to `FieldAttributeRequirement` so ORDER FARs can be sorted back into declared order after both metadata and attribute subqueries have been emitted. The subquery helpers (`_add_metadata_subqueries`, `_add_node_order_attributes`) no longer touch `self.order_by`; a new `_emit_schema_order_by` sorts ORDER FARs by `schema_order_position` and emits the outer clauses.
- Refactored `_get_order_requirements` into two small helpers (`_upsert_metadata_order_requirement`, `_upsert_attribute_order_requirement`) that either promote an existing filter FAR to also act as an ORDER or return a fresh ORDER-only FAR. The nested `requirements_map` dict is gone; lookup is now a direct iteration over `filter_requirements`.
- New FAR.index assignment is `len(filter_requirements) + 1 + position`. Unique by construction, no threaded counter.
- Dropped the dead `_get_metadata_order_fields` helper. The `has_any_order` check now leans on the new `OrderModel.__bool__` (true when `disable` or `node_metadata` carries non-default content) plus the existing `NodeMetaOrder.__bool__`.
- `_query_time_order_overrides_schema` in both `NodeGetListQuery` and `RelationshipGetPeerQuery` collapses to `return bool(self.requested_order)`.
- Added a `requested_order: OrderModel | None` parameter to `RelationshipGetPeerQuery`, threaded through `NodeManager.query_peers`, the `PeerRelationshipsDataLoader` cache key, and the many-relationship resolver. `_add_peer_order_by` drops the peer schema's `order_by` when `requested_order` carries non-default content.

What stayed the same:

- Wire shape of `order_by` is unchanged.
- Parser grammar is unchanged.
- Existing pure-attribute and pure-metadata schemas produce byte-identical ORDER BY clauses.
- Relationship and hierarchy query paths already iterated the schema's `order_by` in declared order, so they needed no fix.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Adds ordering to GraphQL many relationships and fixes ORDER BY precedence so mixed metadata/attribute entries follow the schema’s order. Query-time `order` fully replaces schema `order_by`, cache keys vary by `order`, and the `order_by` wire shape and grammar are unchanged.

- **New Features**
  - Many-relationship `order` is honored end-to-end (threaded through `NodeManager.query_peers`, `PeerRelationshipsDataLoader`, and `RelationshipGetPeerQuery`).
  - Supports `created_at`/`updated_at` ASC/DESC; query-time overrides replace schema defaults. Aligns with INFP-530.

- **Bug Fixes**
  - Node list and hierarchy queries emit ORDER BY clauses in schema-declared order when mixing metadata and attributes.
  - `OrderModel` is now frozen and hashable; peer loader cache keys include `order` to prevent stale results.

<sup>Written for commit f0c35236c4a3f35c299941a2fc1b860a110db897. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/query/node.py
```diff
@@ -2221,105 +2202,127 @@ def _get_filter_requirements(self, start_index: int) -> list[FieldAttributeRequi
...
+        index: int,
+    ) -> FieldAttributeRequirement | None:
+        """Promote an existing filter FAR for the same metadata field, or build a new ORDER-only FAR.
+
+        Metadata filter FARs use operator-specific `field_attr_name` ("before", "after",
+        "value", ...), so we match on `field_name` alone — any FAR for the same metadata
+        field is fine because the outer ORDER BY references the field name directly.
+        """
+        for req in filter_requirements:
+            if req.field_name == metadata_field:
```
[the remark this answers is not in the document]
ajtmccarty: this is only called for metadata fields

backend/infrahub/core/query/node.py
ajtmccarty, line 1: the problem with NodeGetListQuery was that it handled all of the metadata ordering (including adding them to the ORDER BY clause), then handled all of the other ordering, so the metadata ordering would always take precedence over the other ordering. fixing this was more involved than I would have expected and it required weaving the ordering parameters throughout the query generation logic

backend/infrahub/core/query/relationship.py
ajtmccarty, line 1: this is to support ordering override at the many-relationship level, so for a graphql query that looks like this
```
    query($owner_id: ID!) {
        TestPerson(ids: [$owner_id]) {
            edges {
                node {
--->                cars(order: {node_metadata: {created_at: ASC}}) {
                        edges { node { name { value } } }
                    }
                }
            }
        }
    }
```

backend/infrahub/core/order.py
```diff
@@ -15,6 +15,8 @@
 
 
 class NodeMetaOrder(BaseModel):
+    model_config = ConfigDict(frozen=True)
```
ajtmccarty, line 18: freezing so that these can accurately be used for hashing in `ManyRelationshipResolver`. we group many relationship queries by their parameters so we only need to run them one time if they are duplicated within a graphql query

ogenstad (approved):
LGTM. Within the Query classes I'm wondering if at some point the ordering should always be present instead of having to have those if-conditions and perhaps combined with order statements within the schema. But I don't know what that would look like just now, mainly thinking out loud.

PR #9403, opened 2026-06-01
fix(security): verify OIDC id_token signature, audience and issuer by default

polmichel:
# Why

The OIDC `id_token` group-claim path decoded tokens with `verify_signature`, `verify_aud` and `verify_iss` all hardcoded to `False`, so any token presented to the callback was trusted regardless of who signed it or which client/issuer it was minted for. This enables verification by default and documents the related SSO account-correlation hardening.

Jira: [IFC-2662](https://opsmill.atlassian.net/browse/IFC-2662)

Related: #9396 introduced `INFRAHUB_SECURITY_SSO_ACCOUNT_NAME_FALLBACK` (merged to `stable`); this PR adds its user-facing documentation.

## What changed

**Behavioral**

- OIDC `id_token` **signature, audience and issuer are now verified by default** before its group claims are trusted.
- The "verification disabled" warning is emitted **once at config-load** per provider (in a settings validator).

**Documentation**

- `docs/docs/reference/sso.mdx`: added the `ID_TOKEN_VERIFY_SIGNATURE` parameter, plus an **Account name fallback** section documenting `INFRAHUB_SECURITY_SSO_ACCOUNT_NAME_FALLBACK` (introduced in #9396).
- `docs/docs/.../sso/advanced-sso.mdx`: note in Group mapping that the `id_token` is verified by default.
- `misconfigured` added to documentation spell check exceptions.

## How to test

```bash
uv run pytest backend/tests/unit/api/test_oidc.py backend/tests/unit/config/test_config.py
```

## Impact & rollout

- **Backward compatibility:** verification is now on by default; a provider returning improperly signed/scoped `id_token`s will be rejected. Opt out per-provider via the new env var if required.
- **Config/env changes:** new `INFRAHUB_OIDC_<SLOT>_ID_TOKEN_VERIFY_SIGNATURE` (default `true`), registered in `docker-compose.yml`.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [x] External docs updated

[IFC-2662]: https://opsmill.atlassian.net/browse/IFC-2662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/tests/unit/api/test_oidc.py
```diff
@@ -35,52 +47,138 @@ async def test_get_id_token_groups_for_oidc() -> None:
...
         service=service,
         payload=token_response,
-        client_id=client_id,
+        provider=_make_provider(verify_signature=False),
     )
 
     assert groups == ["operators"]
 
 
+async def test_get_id_token_groups_rejects_forged_signature() -> None:
```
polmichel, line 142: A note there about the fact that Test classes could help organizing the different tests in a clustered manner

backend/infrahub/api/oidc.py
```diff
@@ -257,7 +255,10 @@ async def token(
 
 
 async def _get_id_token_groups(
-    oidc_config: OIDCDiscoveryConfig, service: InfrahubServices, payload: dict[str, Any], client_id: str
+    oidc_config: OIDCDiscoveryConfig,
+    service: InfrahubServices,
+    payload: dict[str, Any],
+    provider: config.SecurityOIDCSettings,
```
ajtmccarty: with a name like "provider", I would expect this to be an instance of a class that actually does OIDC provider stuff. I think "provider_settings" or something might be a better name

ajtmccarty (approved):

backend/infrahub/api/oidc.py
polmichel: You're right. I'll do it

gmazoyer (approved):
Always careful when AI is involved in fixing a security issue. Those robots are good at finding them, but they are also good at creating new vulnerabilities. But it looks it has done a decent job here with your supervision.

polmichel:
> Always careful when AI is involved in fixing a security issue. Those robots are good at finding them, but they are also good at creating new vulnerabilities. But it looks it has done a decent job here with your supervision.

Duly noted, thanks for the feedback

PR #9423, opened 2026-06-02
Account for changes to artifact definitions and GraphQL queries (infp-409)

ogenstad:
# Why

Today any commit on a PC's linked repository triggers regeneration of every artifact in that PC, regardless of whether the commit actually affects them.
Editing a README has the same blast radius as editing a transform. Users also can't tell why artifacts regenerated from the task log.

Goal: introduce precise per-definition predicates so unrelated data changes stop forcing artifact regeneration, and lay the foundation (canonicalizer +
closure-builder protocol) the file-change path will consume in Stage 2.

Non-goals: this PR does not touch the file-change selection path. Edits inside a repository still over-regenerate as they do today (residual
has_file_modifications short-circuit retained). The closure builder introduced here is foundational only; its consumers and the _transform_changed predicate
that closes the file-change gap land in Stage 2.

Closes INFP-409 (Stage 1 of a staged rollout).

### Behavioral

- refresh_artifacts selection gate now ORs in two per-definition predicates: _query_changed fires when the definition's CoreGraphQLQuery node is modified
(covers fragment edits, since the SDK inlines fragments before persisting); _definition_changed fires when the CoreArtifactDefinition node itself is modified
(attribute change or targets/transformation/query relationship repoint).
- validate_artifacts_generation flips managed_branch=True on the same predicates rather than unconditionally on any-file-modified.
- Net effect: a PC whose only data-graph change is unrelated to a definition's query/definition node no longer over-selects that definition. File-change
behavior is unchanged in this PR.

### Implementation

- Two pure predicates in backend/infrahub/proposed_change/tasks.py doing id-equality lookups against the cached diff_summary.
- New DefinitionSelect.DEFINITION_OR_QUERY_CHANGED flag; existing MODIFIED_KINDS and FILE_CHANGES flags untouched.
- New backend/infrahub/git/closure_builder/ package containing the path canonicalizer, the ClosureResult/UnresolvedRef result types, and the ClosureBuilder
Protocol. Foundational - no production caller yet; consumers land in Stage 2.
- Unit tests

### DIFF
--- a/backend/infrahub/git/closure_builder/canonicalizer.py
+++ b/backend/infrahub/git/closure_builder/canonicalizer.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+
+
+def canonicalize_path(path: str) -> str:
+    """Return the canonical repo-relative form of a path entering a transform's dependency closure.
+
+    Properties enforced:
+
+    - POSIX forward-slash separator (backslashes are converted).
+    - Leading ``/`` is stripped; a leading slash is interpreted as the repository
+      root, matching the ``.gitignore`` convention.
+    - Leading ``./`` segments are stripped.
+    - Trailing ``/`` is stripped.
+    - Symlinks are not resolved; the canonical form is what git sees.
+    - Case is preserved.
+    - Idempotent: ``canonicalize_path(canonicalize_path(p)) == canonicalize_path(p)``.
+
+    Empty strings and inputs that collapse to the repository root are rejected because
+    they do not name a dependency and would silently match every entry in a diff.
+
+    Raises:
+        ValueError: If ``path`` is empty or collapses to the repository root.
+
+    """
+    if not path:
+        raise ValueError("Path must not be empty")
+
+    normalized = path.replace("\\", "/")
+
+    while True:
+        previous = normalized
+        normalized = normalized.lstrip("/")
+        normalized = normalized.removeprefix("./")
+        normalized = normalized.rstrip("/")
+        if normalized == previous:
+            break
+
+    if normalized in ("", "."):
+        raise ValueError(f"Path resolves to the repository root and is not a valid dependency: {path!r}")
+
+    return normalized
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -808,7 +808,7 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
         or (model.source_branch_sync_with_git and model.branch_diff.has_file_modifications)
     )
     if managed_branch:
-        log.info("Definition-level change or repository file modification detected, all artifacts will be processed")
+        log.info("Query, definition or repository file change detected, all artifacts will be processed")
 
     checks = []
 
@@ -1205,7 +1205,8 @@ class DefinitionSelect(IntFlag):
     NONE = 0
     MODIFIED_KINDS = 1
     FILE_CHANGES = 2
-    DEFINITION_OR_QUERY_CHANGED = 4
+    QUERY_CHANGED = 4
+    DEFINITION_CHANGED = 8
 
     @staticmethod
     def add_flag(current: DefinitionSelect, flag: DefinitionSelect, condition: bool) -> DefinitionSelect:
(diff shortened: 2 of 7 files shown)


backend/infrahub/git/closure_builder/canonicalizer.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+
+
+def canonicalize_path(path: str) -> str:
```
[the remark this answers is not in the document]
ogenstad: In this PR we're just setting up this for a future step.

backend/infrahub/proposed_change/tasks.py
```diff
@@ -1169,10 +1173,39 @@ def _run_generator(instance_id: str | None, managed_branch: bool, impacted_insta
...
+    change uniformly.
+    """
+    return any(entry["id"] == definition.definition_id for entry in diff_summary)
+
+
 class DefinitionSelect(IntFlag):
     NONE = 0
     MODIFIED_KINDS = 1
     FILE_CHANGES = 2
+    DEFINITION_OR_QUERY_CHANGED = 4
```
gmazoyer: This is probably a silly question: why not having two separate flags such as `DEFINITION_CHANGED` and `QUERY_CHANGED`?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -802,9 +802,13 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
...
 
-    managed_branch = model.source_branch_sync_with_git and model.branch_diff.has_file_modifications
+    managed_branch = (
+        _query_changed(definition=model.artifact_definition, diff_summary=diff_summary)
+        or _definition_changed(definition=model.artifact_definition, diff_summary=diff_summary)
+        or (model.source_branch_sync_with_git and model.branch_diff.has_file_modifications)
+    )
     if managed_branch:
-        log.info("Source branch is synced with Git repositories with updates, all artifacts will be processed")
+        log.info("Definition-level change or repository file modification detected, all artifacts will be processed")
```
gmazoyer: I'd go with a phrasing that follows the conditional like: `"Query, definition or repository file change detected, all artifacts will be processed"`

gmazoyer (approved):

backend/infrahub/proposed_change/tasks.py
ogenstad: Yes, that's a good point. Will change, thanks!

backend/infrahub/proposed_change/tasks.py
```diff
@@ -1169,10 +1173,39 @@ def _run_generator(instance_id: str | None, managed_branch: bool, impacted_insta
...
+    diff_summary: list[NodeDiff],
+) -> bool:
+    """Return True when the definition's GraphQL query node is modified in the diff.
+
+    The SDK inlines every fragment body into the stored query text before persisting,
+    so any edit to the primary ``.gql`` file or any transitively referenced fragment
+    surfaces as a single ``CoreGraphQLQuery`` node modification. A node-id match is
+    therefore sufficient.
+    """
+    return any(entry["id"] == definition.query_id for entry in diff_summary)
```
ajtmccarty: should this take NodeDiff.action into account in case the NodeDiff is UNCHANGED... or maybe DELETED?

backend/tests/unit/proposed_change/test_predicates.py
```diff
@@ -0,0 +1,261 @@
...
+        expected=False,
+    ),
+    DefinitionChangedCase(
+        name="attribute_change_on_definition",
+        diff=[
+            _node_diff(
+                node_id=DEFINITION_ID,
+                kind="CoreArtifactDefinition",
+                action="updated",
+                elements=[
```
ajtmccarty: are the nested attribute/relationship test cases really doing anything right now? the check is just on whether the DEFINITION_ID is in the diff at all right?

ajtmccarty (approved):

backend/infrahub/proposed_change/tasks.py
ogenstad: Yeah, perhaps for clarity we can change to added or updated. If it's deleted we should never end up here and I don't think unchanged would come into play as it's only really for hierarchy objects, but we don't need to trigger on it regardless.

backend/tests/unit/proposed_change/test_predicates.py
ogenstad: Fair point, I guess those could be changed.

PR #9430, opened 2026-06-02
fix: resolve S3 credentials via AWS default chain when static keys are unset (#9429)

### LINKED ISSUE #9429: bug: S3 storage breaks with IRSA / instance-profile credentials (empty access key passed to boto3)
## Component

API Server / GraphQL

## Infrahub version

4.7.3 (Enterprise, medium), deployed on EKS

## Current Behavior

When deploying with S3 artifact storage on EKS using IRSA (IAM Roles for
Service Accounts) — the standard EKS pattern where pods assume an IAM role via
a projected service account token, with no static `AWS_ACCESS_KEY_ID` /
`AWS_SECRET_ACCESS_KEY` set — S3 access fails:

    botocore.exceptions.ClientError: AuthorizationHeaderMalformed —
    a non-empty Access Key (AKID) must be provided

Root cause: `S3StorageSettings` in `backend/infrahub/config.py` defines
`access_key_id` and `secret_access_key` with `default=""`, so they can never be
`None`. `InfrahubS3ObjectStorage` (`backend/infrahub/storage.py`) inherits from
`fastapi_storages.S3Storage`, whose `__init__` passes these straight through:

    self._s3 = boto3.resource(
        "s3",
        endpoint_url=self._url,
        aws_access_key_id=self.AWS_ACCESS_KEY_ID,        # ""
        aws_secret_access_key=self.AWS_SECRET_ACCESS_KEY, # ""
    )

When an empty string is passed explicitly, boto3 uses it verbatim instead of
falling back to the default credential provider chain (IRSA, instance profiles,
ECS task roles, etc.). This restricts S3 storage to static keys only.

## Expected Behavior

When no static AWS credentials are configured, Infrahub should let boto3 fall
back to the default AWS credential provider chain, so S3 storage works with
IRSA, EC2 instance profiles, ECS task roles, and any standard AWS credential
method — not only static access keys.

## Steps to Reproduce

1. Deploy Infrahub on EKS with S3 artifact storage configured (bucket,
   endpoint), but **without** `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`.
2. Attach an IAM role to the pod via IRSA (`AWS_WEB_IDENTITY_TOKEN_FILE` +
   `AWS_ROLE_ARN` present in the environment).
3. Trigger any S3 read/write (e.g. generate or fetch an artifact).
4. Observe `AuthorizationHeaderMalformed` — the empty access key is sent
   instead of credentials being resolved via IRSA.

## Additional Information

Verified that passing `None` (vs `""`) restores the credential-chain fallback:

- `aws_access_key_id=""` → sends empty AKID → fails
- `aws_access_key_id=None` → falls back to credential chain → works

Suggested direction (for the implementer to evaluate): when the credential
fields are empty, omit them from the `boto3.resource(...)` call rather than
passing empty strings — e.g. override the resource construction in
`InfrahubS3ObjectStorage` so the kwargs are only included when an access key is
set. This would cover IRSA, instance profiles, and any standard AWS credential
method. The fix likely needs to live in Infrahub's `InfrahubS3ObjectStorage`
subclass rather than relying on the upstream `fastapi_storages` library.

### Working workaround (no static keys, no custom image)

We unblocked ourselves by injecting a `sitecustomize.py` into site-packages via
a Kubernetes ConfigMap volume mount. Python loads `sitecustomize.py` at
interpreter startup, before application code, and it flips the empty-string
credentials to `None`:

    import fastapi_storages.s3

    _orig_init = fastapi_storages.s3.S3Storage.__init__

    def _patched_init(self):
        if not self.AWS_ACCESS_KEY_ID:
            self.AWS_ACCESS_KEY_ID = None
        if not self.AWS_SECRET_ACCESS_KEY:
            self.AWS_SECRET_ACCESS_KEY = None
        _orig_init(self)

    fastapi_storages.s3.S3Storage.__init__ = _patched_init

Deployment:

1. ConfigMap `infrahub-irsa-patch` holds the patch script.
2. A Flux HelmRelease `postRenderer` mounts it into the server and task-worker
   pods at `/.venv/lib/python3.13/site-packages/sitecustomize.py` via `subPath`.
3. On startup Python loads the patch → empty credentials become `None` → boto3
   falls back to the default credential provider chain → picks up the IRSA
   token from `AWS_WEB_IDENTITY_TOKEN_FILE`.

Result: artifact generation flows complete successfully with S3 storage via
IRSA — no static keys, no custom images, no init containers. Once the upstream
fix ships, the ConfigMap and volume mount are simply removed.

minitriga:
## Summary

Fixes #9429. S3 object storage failed with `AuthorizationHeaderMalformed: a non-empty Access Key (AKID) must be provided` on deployments that rely on AWS role-based credentials (IRSA, EC2 instance profiles, ECS task roles) instead of static keys.

`S3StorageSettings` defaults `access_key_id` / `secret_access_key` to empty strings, and the storage layer forwarded those empty strings straight to `boto3.resource(...)`. boto3 treats an empty string as an explicit credential and skips the default credential provider chain, so role-based credential resolution never happened.

## Change

`InfrahubS3ObjectStorage.__init__` now treats blank, whitespace-only, or incomplete static credentials as "not configured" and passes `None`, letting boto3 fall back to the default AWS credential provider chain. Complete static credentials are forwarded unchanged, so existing static-key and S3-compatible (e.g. MinIO) deployments are unaffected.

The fix lives entirely in Infrahub's subclass — the upstream `fastapi_storages` library is not modified.

## Behavior

| Access key | Secret key | Result |
|---|---|---|
| set | set | forwarded verbatim (unchanged) |
| blank | blank | `None` → default credential chain |
| set | blank (or reverse) | incomplete → `None` → default credential chain |
| whitespace-only | whitespace-only | treated as blank → default credential chain |

## Tests

New `backend/tests/unit/test_storage.py` stubs `boto3.resource` and asserts the credential kwargs across all four states (5 tests, all passing). No external services required.

## Verification not covered by CI

The end-to-end IRSA path (removing the `sitecustomize.py` workaround on a live EKS deployment) can only be confirmed on a role-based deployment — see `dev/specs/ifc-2683-fix-s3-irsa-credentials/quickstart.md` §4.

## Notes

- Two `# type: ignore[assignment]` are introduced (one per credential attribute): `fastapi_storages.S3Storage` types them as `str`, but boto3 requires `None` to trigger credential-chain fallback. Justified inline.
- Changelog: `changelog/9429.fixed.md`.
- Spec/plan/tasks artifacts under `dev/specs/ifc-2683-fix-s3-irsa-credentials/`.

---
## Summary by cubic
Fix S3 storage auth for AWS role-based credentials by letting `boto3` use the default credential chain when both static keys are unset. Partial static credentials now raise a clear error. Addresses Linear IFC-2683.

- **Bug Fixes**
  - In `InfrahubS3ObjectStorage.__init__`, treat blank/whitespace-only keys as not configured and set them to `None` so `boto3.resource(...)` resolves IRSA/instance-profile/task-role credentials.
  - Reject partial credentials with a `ValueError` explaining both or neither must be set.
  - Forward complete static credentials unchanged (S3-compatible deployments unaffected).

<sup>Written for commit 732e394f5e78d725adf94e4b8a42b8172ca5aa4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9430?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/storage.py
```diff
@@ -19,6 +19,18 @@ def __init__(self, **kwargs: Any) -> None:
         for key, value in kwargs.items():
             if hasattr(self, key):
                 setattr(self, key, value)
+
+        access_key = kwargs.get("AWS_ACCESS_KEY_ID")
+        secret_key = kwargs.get("AWS_SECRET_ACCESS_KEY")
+        if not (access_key and str(access_key).strip()) or not (secret_key and str(secret_key).strip()):
```
ajtmccarty: I find this very hard to read.

I think this is a little better.
```
if access_key is None or (access_key is not None and str(access_key).strip() == ""):
    self.AWS_ACCESS_KEY_ID = None
# repeat for secret_key
```
It also sounds like it is an error to only set either ACCESS_KEY or SECRET_KEY without the other. if that is the case, then I think this should raise an error instead of silently dropping something halfway configured

.specify/feature.json
ajtmccarty, line 1: accidentally included?

CLAUDE.md
```diff
@@ -2,5 +2,6 @@
 
 <!-- SPECKIT START -->
 For additional context about technologies to be used, project structure,
-shell commands, and other important information, read the current plan
+shell commands, and other important information, read the current plan:
+`specs/ifc-2683-fix-s3-irsa-credentials/plan.md`
```
ajtmccarty: whoopsidaisies

ajtmccarty (commented):
I'd say that the ~500 lines of spec/plan/etc for this pretty small code change is excessive and can be left out. But I don't know if we've really agreed on a limit or border for including these specs or not

backend/infrahub/storage.py
minitriga: Reworked per your comment:

- **Readability**: each value is normalized up front — `access_key = str(kwargs.get("AWS_ACCESS_KEY_ID") or "").strip()` (same for the secret) — so the logic reads as two plain branches: `if bool(access_key) != bool(secret_key):` (exactly one set) and `if not access_key:` (both unset).
- **Partial config now raises** a `ValueError` instead of silently dropping a half-configured pair, as you suggested. Both unset → default AWS credential provider chain; both set → static credentials used verbatim. Added a unit test asserting the raise for each one-sided case.

.specify/feature.json
minitriga: Yes — accidental. Removed `.specify/feature.json` along with all the other spec artifacts; the PR diff is now just the three code files.

CLAUDE.md
minitriga: Reverted — the `CLAUDE.md` change is no longer in the PR.

.specify/feature.json
```diff
@@ -0,0 +1,3 @@
+{
+  "feature_directory": "specs/ifc-2683-fix-s3-irsa-credentials"
```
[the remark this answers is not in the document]
minitriga: Removed `.specify/feature.json` and all spec artifacts from the PR, so this path issue no longer applies.

dev/specs/ifc-2683-fix-s3-irsa-credentials/plan.md
```diff
@@ -0,0 +1,77 @@
...
+## Summary
+
+S3 object storage cannot use AWS role-based credentials (IRSA, EC2 instance profiles, ECS task roles). The S3 credential settings default to empty strings, and the storage layer passes those empty strings straight to the AWS client, which uses them verbatim instead of falling back to the AWS default credential provider chain — producing an `AuthorizationHeaderMalformed` error.
+
+**Technical approach**: In the Infrahub S3 storage subclass, treat blank access-key / secret-key values as "not configured" and let the AWS client resolve credentials through its default provider chain when static credentials are absent or incomplete. When complete static credentials are present, behavior is unchanged. The fix lives entirely in Infrahub's `InfrahubS3ObjectStorage` subclass — the upstream `fastapi_storages` library is not modified.
+
+## Technical Context
+
+**Language/Version**: Python 3.12 (backend)  
+**Primary Dependencies**: `fastapi_storages` (S3 backend), `boto3` / `botocore` (AWS client), Pydantic 2.10 (`BaseSettings`)
```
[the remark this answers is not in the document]
minitriga: Removed the spec/plan artifacts from the PR, so this no longer applies.

minitriga:
Pushed an update addressing the review:

- **Removed all spec/plan/tasks artifacts** (`dev/specs/...`, `.specify/feature.json`) and reverted the accidental `CLAUDE.md` change. The PR diff is now just the three code files: `backend/infrahub/storage.py`, `backend/tests/unit/test_storage.py`, `changelog/9429.fixed.md`. Agreed they were excessive for a change this size.
- **`storage.py`**: simplified the credential check, and partial config (only one of `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`) now raises a `ValueError` instead of silently falling back — per @ajtmccarty's suggestion.
- Updated the changelog and the unit tests accordingly (partial config now asserts the raise). ruff, mypy, and the 5 unit tests pass.

Force-pushed to rewrite the single commit so the spec files are absent from history, not just deleted in a follow-up.

ajtmccarty (approved):

PR #9439, opened 2026-06-03
test: reusable lock recording adapter

polmichel:
## Why

Testing lock behavior in the backend previously required `unittest.mock`/`patch`, which the testing guidelines forbid. This adds a reusable, no-mock lock recording adapter so a test can observe exactly when a lock is acquired/released and which locks are held at any point. It injects a `RecordingLockRegistry` in place of the global one.

## What

- `RecordingLockRegistry` / `RecordingLock` backed by a shared `LockTimeline` that records acquire/release events (re-entrancy aware) plus arbitrary checkpoints.
- Query/assert helpers: `held_at`, `currently_held`, `acquire_sequence`, `assert_held_at_checkpoint`, `assert_never_overlap`.
- A `recording_lock_timeline` pytest fixture and unit tests covering the adapter itself.

## Context

Subtask [IFC-2672](https://opsmill.atlassian.net/browse/IFC-2672) of [IFC-1566](https://opsmill.atlassian.net/browse/IFC-1566) (git-sync long lock durations). The stacked follow-up PR uses this adapter to prove the repository lock is no longer held during the object import. Kept standalone because the adapter is reusable for any lock ordering / contention / scope test.

[IFC-2672]: https://opsmill.atlassian.net/browse/IFC-2672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-1566]: https://opsmill.atlassian.net/browse/IFC-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Add a reusable lock recording adapter for tests to assert acquire/release order and detect overlap without mocks. Now supports checking overlap across any number of locks and adds explicit checkpoint assertions.

- **New Features**
  - `RecordingLockRegistry` and `RecordingLock` with a shared `LockTimeline` that logs acquire/release and checkpoints; re-entrancy aware, and the registry requires an injected timeline.
  - Query/assert helpers: `held_at`, `currently_held`, `acquire_sequence`, `assert_held_at_checkpoint`, `assert_not_held_at_checkpoint`, `assert_never_overlap` (accepts a collection of lock names).
  - `pytest` fixture `recording_lock_timeline` and helper `install_recording_lock_registry` to install the recording registry per test and restore the original.
  - Unit tests for order, checkpoints, re-entrancy, multi-lock membership, and fixture swap.

- **Bug Fixes**
  - Explicit `RecordingLock` constructor mirroring `InfrahubLock` to satisfy CI type checking.

<sup>Written for commit 678b4b146054bb93d5cdedd244729a2801ccd3aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/adapters/lock.py
+++ b/backend/tests/adapters/lock.py
@@ -3,9 +3,15 @@ from __future__ import annotations
 import itertools
 from dataclasses import dataclass
 from enum import StrEnum
+from typing import TYPE_CHECKING
 
 from infrahub import lock
 from infrahub.lock import InfrahubLock, InfrahubLockRegistry
+
+if TYPE_CHECKING:
+    import redis.asyncio as redis
+
+    from infrahub.services import InfrahubServices
 
 
 class LockAction(StrEnum):
@@ -100,8 +106,16 @@ class LockTimeline:
 class RecordingLock(InfrahubLock):
     """A local lock that logs its real (non-re-entrant) acquire/release boundaries to a timeline."""
 
-    def __init__(self, *args: object, timeline: LockTimeline, **kwargs: object) -> None:
-        super().__init__(*args, **kwargs)  # type: ignore[arg-type]
+    def __init__(
+        self,
+        name: str,
+        connection: redis.Redis | InfrahubServices | None = None,
+        in_multi: bool = False,
+        metrics: bool = True,
+        *,
+        timeline: LockTimeline,
+    ) -> None:
+        super().__init__(name=name, connection=connection, in_multi=in_multi, metrics=metrics)
         self._timeline = timeline
 
     async def acquire(self) -> None:
--- a/backend/tests/adapters/lock.py
+++ b/backend/tests/adapters/lock.py
@@ -9,6 +9,8 @@ from infrahub import lock
 from infrahub.lock import InfrahubLock, InfrahubLockRegistry
 
 if TYPE_CHECKING:
+    from collections.abc import Collection
+
     import redis.asyncio as redis
 
     from infrahub.services import InfrahubServices
@@ -92,13 +94,21 @@ class LockTimeline:
                     f"Held at that point: {sorted(self.held_at(seq))}"
                 )
 
-    def assert_never_overlap(self, lock_a: str, lock_b: str) -> None:
+    def assert_never_overlap(self, lock_names: Collection[str]) -> None:
+        """Assert that no two of ``lock_names`` were ever held at the same time.
+
+        Raises:
+            AssertionError: if two or more of ``lock_names`` were held simultaneously.
+
+        """
+        watched = set(lock_names)
         held: set[str] = set()
         for event in self.events:
             if event.action == LockAction.ACQUIRE:
                 held.add(event.name)
-                if lock_a in held and lock_b in held:
-                    raise AssertionError(f"Locks {lock_a!r} and {lock_b!r} were held simultaneously")
+                overlap = held & watched
+                if len(overlap) > 1:
+                    raise AssertionError(f"Locks {sorted(overlap)} were held simultaneously")
             elif event.action == LockAction.RELEASE:
                 held.discard(event.name)
 
--- a/backend/tests/adapters/lock.py
+++ b/backend/tests/adapters/lock.py
@@ -1,0 +1,156 @@
+from __future__ import annotations
+
+import itertools
+from dataclasses import dataclass
+from enum import StrEnum
+
+from infrahub import lock
+from infrahub.lock import InfrahubLock, InfrahubLockRegistry
+
+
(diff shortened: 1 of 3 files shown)


backend/tests/unit/adapters/test_lock.py
```diff
@@ -0,0 +1,62 @@
+from infrahub import lock
```
polmichel, line 1: Tests of the test adapter, illustrating use cases and correct behavior.

backend/tests/adapters/lock.py
```diff
@@ -0,0 +1,165 @@
...
+            raise AssertionError(f"No checkpoint named {label!r} was recorded")
+        for seq in seqs:
+            actually_held = lock_name in self.held_at(seq)
+            if actually_held is not expected:
+                raise AssertionError(
+                    f"At checkpoint {label!r}, expected lock {lock_name!r} held={expected} but held={actually_held}. "
+                    f"Held at that point: {sorted(self.held_at(seq))}"
+                )
+
+    def assert_never_overlap(self, lock_a: str, lock_b: str) -> None:
```
ajtmccarty: could this accept a `list` or `set` parameter instead of 2 strings?

backend/tests/adapters/lock.py
```diff
@@ -0,0 +1,165 @@
...
+            for event in self.events
+            if event.action == LockAction.ACQUIRE and (prefix is None or event.name.startswith(prefix))
+        ]
+
+    def checkpoint_seqs(self, label: str) -> list[int]:
+        return [event.seq for event in self.events if event.action == LockAction.CHECKPOINT and event.label == label]
+
+    # --- assertion helpers ---
+
+    def assert_held_at_checkpoint(self, lock_name: str, label: str, *, expected: bool) -> None:
```
ajtmccarty: I think the `expected` parameter is a confusing name. this might be better as two methods `assert_held...` and `assert_not_held...` that call a shared private method

backend/tests/adapters/lock.py
```diff
@@ -0,0 +1,165 @@
...
+        will_release = self._recursion_var.get() == 1
+        await super().release()
+        if will_release:
+            self._timeline.record(self.name, LockAction.RELEASE)
+
+
+class RecordingLockRegistry(InfrahubLockRegistry):
+    """Local-only lock registry that hands out ``RecordingLock`` instances backed by a shared timeline."""
+
+    def __init__(self, timeline: LockTimeline | None = None) -> None:
```
ajtmccarty: I think you should make `timeline` required here. initializing an object within an `__init__` does not follow dependency injection. we do it in `InfrahubLockRegistry`, but we should not

backend/tests/adapters/lock.py
```diff
@@ -0,0 +1,165 @@
...
+        metrics: bool = True,
+    ) -> InfrahubLock:
+        lock_name = self.name_generator.generate_name(name=name, namespace=namespace, local=local)
+        if lock_name not in self.locks:
+            self.locks[lock_name] = RecordingLock(
+                name=lock_name,
+                connection=self.connection,
+                in_multi=in_multi,
+                metrics=metrics,
+                timeline=self.timeline,
```
ajtmccarty, line 172: I don't like passing `self.timeline` down into the instantiation of this object b/c it couples the recording registry to the recording lock implementation, but it is for testing and it is a reflection of how our actual lock logic should be updated, so I think it is all right in this case

ajtmccarty (approved):

backend/tests/adapters/lock.py
polmichel: Great idea, I'll do it.

backend/tests/adapters/lock.py
polmichel: Makes sense, I'll also remove the `# --- assertion helpers ---` section

backend/tests/adapters/lock.py
polmichel: I agree this is better.

ajtmccarty (approved):

PR #9440, opened 2026-06-03
fix(git): run object import outside the repository lock during sync

polmichel:
## Why

During periodic git sync, the per-repository lock was held for the whole sync, including the object import. When a branch carries conflicting data the import/reconcile phase is slow, the lock stayed reserved during that time and degraded the platform. See [IFC-2673](https://opsmill.atlassian.net/browse/IFC-2673) (subtask of [IFC-1566](https://opsmill.atlassian.net/browse/IFC-1566)).

## What

- Split `repo.sync()` into `collect_pending_imports()` and the import loop.
- Narrow the lock: a `RepositorySyncer` component holds the repository lock only for the git phase and runs the imports after releasing it.
- `RepositoryImporter` (ABC) + `RepositoryFileImporter` make the import an injectable collaborator..
- Component test asserts the repository lock is not held when the import runs.

## Context

Stacked on top of #9439 (lock recording adapter), which this PR uses to assert the lock scope.

## Notes / residual risk
- `infrahub.git.base.InfrahubRepositoryBase.create_commit_worktree` currently performs a remote `fetch` inside the local worktree primitive; this will be removed in a dedicated PR targeting this feature branch.
- `add_git_repository` still imports under its own lock; it will be tackled in a dedicated PR targeting this feature branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[IFC-2673]: https://opsmill.atlassian.net/browse/IFC-2673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-1566]: https://opsmill.atlassian.net/browse/IFC-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Move object import out of the repository lock during git sync by collecting pending imports first, then importing after the lock is released. This reduces long lock holds during slow reconciles and improves throughput (IFC-2673, IFC-1566).

- **Bug Fixes**
  - Split `InfrahubRepository.sync()` into `collect_pending_imports()` (fetch/branch/pull/worktree pinning) and a follow-up import loop.
  - Added `PendingObjectImport` and `_collect_staging_imports()` to return all imports, including staging/default-branch updates.
[...]

PR #9441, opened 2026-06-03
fix(git): run object import outside the repository lock when adding a repository

polmichel:
## Why

`add_git_repository` held the repository lock across the whole flow, including the default-branch object import. The object import reads from the immutable per-commit worktree and does not mutate the shared clone, so holding the lock across it needlessly blocks other repository operations for the duration of the import. See [IFC-2686](https://opsmill.atlassian.net/browse/IFC-2686) (subtask of [IFC-1566](https://opsmill.atlassian.net/browse/IFC-1566)).

## What

- Narrow the lock to the git working-copy mutations: the local clone and pinning the default-branch commit worktree.
- Run the default-branch object import after the lock is released; it reads the pinned per-commit worktree and does not need the lock.
- Restore `create_commit_worktree` as a local-only primitive, `InfrahubRepository.init(commit=…)` now fetches once under the repository lock and retries before raising.
- Add a no-mock test asserting the repository lock is not held when the import runs, and update the existing create test for the new structure.

## Context

Stacked on top of #9440 (sync-path lock-scope fix), which introduced `RepositoryImporter` / `RepositorySyncer`. Target this PR's base to #9440 so the diff stays scoped to `add_git_repository`.

## Notes / residual risk

Same semantics shift as the sync-path fix: the lock is released between the git working-copy phase and the object import rather than held across both. The origin sync (fetch/collect) stays inside the lock.

[IFC-2686]: https://opsmill.atlassian.net/browse/IFC-2686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-1566]: https://opsmill.atlassian.net/browse/IFC-1566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Narrow the repository lock in add and sync flows to only cloning and pinning the default-branch commit. Run object imports and origin sync after releasing the lock; restore worktree creation as local-only and fetch missing commits in `init()`. Meets IFC-2686.

- **Bug Fixes**
  - Introduced `RepositoryAdder`; `add_git_repository` uses it and `RepositoryFileImporter` by default. Lock now covers only `InfrahubRepository.new`, resolving the default branch, and creating its worktree (no importer param on the task).
  - Default-branch import runs outside the lock via `RepositoryImporter.import_branch` with `PendingObjectImport(commit=…)`.
  - Repository sync runs outside the lock via `RepositorySyncer`; only git working-copy mutations (clone/init/reinit and collecting pending imports) are locked. Still publishes `RefreshGitFetch` with the pinned commit.
  - `create_commit_worktree` raises `CommitNotFoundError` when the commit is missing; `InfrahubRepository.init(commit=…)` fetches once under the repository lock (if origin exists) and retries before raising.
  - Tests: add a no-mock add-repo lock-scope test with a recording lock/`RecordingImporter`; move recording adapters to `tests.adapters.lock` and expose a `recording_lock_timeline` fixture; tighten error-message assertions and update the RPC create test for the new import path.

<sup>Written for commit 5a9b38aad978d50d20894f7ef2be0985507d44a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/adapters/lock/registry.py
polmichel, line 1: Moved here due to duplication in two different test files

backend/tests/component/git/test_git_rpc.py
```diff
@@ -92,33 +95,35 @@ async def test_git_rpc_create_successful(
...
+        self.mock_repo.collect_pending_imports = AsyncMock(return_value=[])
+
+        original_registry = lock.registry
+        timeline = install_recording_lock_registry()
+        try:
+            with patch("infrahub.git.tasks.InfrahubRepository", spec=InfrahubRepository) as mock_repo_class:
+                mock_repo_class.new.return_value = self.mock_repo
+                await add_git_repository(model=model)
+
+                assert f"repository.{git_upstream_repo_01['name']}" in timeline.acquire_sequence()
```
polmichel: This is new assertion which replaces the assertions against the older mock of Lock class

backend/infrahub/git/tasks.py
```diff
@@ -85,7 +85,8 @@ def format_check_log_entry(entry: dict[str, Any]) -> str:
     name="git-repository-add-read-write",
     flow_run_name="Adding repository {model.repository_name} in branch {model.infrahub_branch_name}",
 )
-async def add_git_repository(model: GitRepositoryAdd) -> None:
+async def add_git_repository(model: GitRepositoryAdd, importer: RepositoryImporter | None = None) -> None:
```
ajtmccarty: doesn't look like the `importer` parameter is ever used, so might be cleaner to just remove it and always instantiate `RepositoryImporter`

backend/infrahub/git/tasks.py
```diff
@@ -98,30 +99,39 @@ async def add_git_repository(model: GitRepositoryAdd) -> None:
...
+    await effective_importer.import_branch(
+        repo,
+        PendingObjectImport(
+            infrahub_branch_name=model.infrahub_branch_name,
+            git_branch_name=repo.default_branch,
+            commit=default_commit,
+        ),
+    )
+
+    if model.internal_status == RepositoryInternalStatus.ACTIVE.value:
```
ajtmccarty: you could invert this conditional and return early here

ajtmccarty (approved):

backend/tests/component/git/test_git_rpc.py
```diff
@@ -380,3 +383,29 @@ async def test_new_repository(self, setup: None, prefect_test_fixture: None) ->
...
+        repository_name=git_upstream_repo_01["name"],
+        location=str(git_upstream_repo_01["path"]),
+        default_branch_name="main",
+        infrahub_branch_name="main",
+        infrahub_branch_id=str(UUIDT()),
+        internal_status=RepositoryInternalStatus.INACTIVE.value,
+    )
+
+    with dependency_provider.scope(build_client, lambda: client):
+        await add_git_repository(model=model, importer=importer)
```
polmichel, line 409: `importer` parameter is used here

backend/infrahub/git/tasks.py
polmichel: The parameter is used in a new test from this PR introduced here 3dffbf0e9130f7a54859e6b0ef1e22e726ae1207
`tests.component.git.test_git_rpc.test_add_git_repository_releases_lock_before_import`.
While checking it I've created another component next to `RepositorySyncer` which is `RepositoryAdder`. We remove the injection through the Prefect flow to inject it directly at the `RepositoryAdder` level.

backend/infrahub/git/tasks.py
polmichel: Thanks! I'll do it!

backend/infrahub/git/tasks.py
ajtmccarty: excellent

backend/tests/component/git/test_git_rpc.py
polmichel: (Old version of a test, which now uses `RepositoryAdder` component)

ajtmccarty (approved):

PR #9442, opened 2026-06-03
feat(artifacts): precise per-transform regeneration gate (INFP-409)

ogenstad:
# Why

Today, any commit on a PC's linked repository regenerates every artifact in
that repository, regardless of whether the commit touched a file the transform
actually reads. Editing a README has the same blast radius as editing a
transform source.

This PR introduces the mechanism that makes the file-change gate precise:
each transform's dependency closure is computed at git-import time, persisted
on the CoreTransformation node, and used at PC-pipeline time to intersect
against the per-repo file diff. Transforms that have not been re-imported
yet (closure null or marked incomplete) fall back to the existing
"any file changed" check, so pre-feature transforms keep working until their
next natural re-import.

Non-goals (each lands in a follow-up PR):
- `watch.files` SDK opt-in for cross-package dependencies (US3)
- Diagnostic logging of which file triggered which regeneration (US2)
- End-to-end functional + integration_docker tests (T032-T034)
- Stage 2 changelog fragment (T067)

### Behavioral

- `refresh_artifacts` and `validate_artifacts_generation` now also consult
  `_transform_changed(definition, repo_diff)`. On the trusted path
  (`dependencies` non-null + `dependencies_complete=True`), regeneration fires
  only when the stored closure intersects the per-repo file diff. On the
  fallback path (closure null or incomplete), it returns
  `repo_diff.has_modifications` so pre-feature transforms keep working until
  their next natural re-import.
- The residual legacy `has_file_modifications` short-circuit is removed; the
  precise check fully owns the file-change selection path.

### Implementation

- Schema: `CoreTransformation` gains `dependencies: List` (canonical
  repo-relative paths) and `dependencies_complete: Boolean` (trust bit). Both
  optional, populated by the integrator at import time.
- `backend/infrahub/git/closure_builder/`:
  - `Jinja2Closure` walks `{% include %}` / `{% import %}` / `{% extends %}`
    transitively via `jinja2.meta.find_referenced_templates`. Dynamic refs,
    missing templates, unreadable templates, and references escaping the
    worktree root are each recorded as `UnresolvedRef` and flip
    `complete=False`; the walk continues so one import surfaces every problem.
  - `PythonClosure` uses the package-directory floor via `git ls-files`,
    excludes `.pyc` / `__pycache__/`. Root-level transforms collapse to the
    entry file. Git enumeration failure produces `complete=False` so the gate
    falls back to the coarser path.
  - `build_transform_closure` dispatches per config kind via exhaustive `match`
    + `assert_never`; isolated failure set produces a fallback `ClosureResult`
    instead of aborting the repository import.
  - `append_manifest_path` unions `.infrahub.yml` into every closure.
  - `ClosureResult` is a frozen dataclass; sorted / canonical / deduplicated
    invariants enforced in `__post_init__`.
- `_transform_changed` in `proposed_change/tasks.py`: canonicalize both sides,
  set-intersect; trusted path explained above. `_repo_diff_or_none` helper
  dedupes the per-call-site `NodeNotFoundError` handling.
- Integrator (`import_jinja2_transforms`, `import_python_transforms`) calls the
  dispatcher after parsing each transform from `.infrahub.yml`, persists the
  result on the `CoreTransformation` node, error-isolates per-transform so one
  broken transform does not abort the repository import.
- Unit tests


backend/infrahub/core/schema/definitions/core/transform.py
```diff
@@ -31,6 +31,18 @@
...
+            name="dependencies",
+            kind="List",
+            description="Canonical repo-relative paths feeding this transform's output. Null falls back to legacy file gate.",
+            optional=True,
+        ),
+        Attr(
+            name="dependencies_complete",
+            kind="Boolean",
+            description="True when the dependency closure can be trusted. False when auto-detection found unresolved references.",
+            optional=True,
```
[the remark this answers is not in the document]
ogenstad: The way it currently works is that the Prefect workers use the public API using the SDK to set these values, so this would be the approach for now.

backend/infrahub/core/schema/definitions/core/transform.py
```diff
@@ -31,6 +31,18 @@
         Attr(name="label", kind="Text", optional=True),
         Attr(name="description", kind="Text", optional=True),
         Attr(name="timeout", kind="Number", description="Maximum execution time in seconds", default_value=60),
+        Attr(
+            name="dependencies",
+            kind="List",
+            description="Canonical repo-relative paths feeding this transform's output. Null falls back to legacy file gate.",
+            optional=True,
```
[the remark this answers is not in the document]
ogenstad: Same comment as above.

backend/infrahub/git/closure_builder/dispatcher.py
```diff
@@ -0,0 +1,57 @@
...
+            case InfrahubPythonTransformConfig():
+                raw = PythonClosure().build(transform_config=transform_config, worktree_root=worktree_root)
+            case _:
+                assert_never(transform_config)
+    except ISOLATED_FAILURES:
+        if logger is not None:
+            logger.exception(f"Closure builder failed for transform {transform_config.name!r}")
+        return ClosureResult(dependencies=(), complete=False, unresolved=())
+
+    return append_manifest_path(result=raw)
```
polmichel: Optional comment:
This method (which looks like an object `ClosureResultBuilder` but this is not objictively cleaner to use an object) seems to have two different responsibilities: matching the correct `Closure` and build the related result.
This is just a thought but would it make sense to create common abstract object on top of `Closure` objects and create a `match_closure(transform_config: TransformConfig) -> Closure` method which would make the matching? 
It would enforce the signature of the `build` method and generalize the code. Maybe this is not the best way to design this if the signature will change regarding other `Closure` types

backend/infrahub/git/closure_builder/jinja2_closure.py
```diff
@@ -0,0 +1,98 @@
...
+    Parses each reached template, collects referenced templates via
+    ``jinja2.meta.find_referenced_templates``, and walks transitively. Each
+    individual reference is handed to a `Jinja2ReferenceResolver` which decides
+    whether the reference can be enqueued or must be recorded as unresolved.
+
+    The returned `ClosureResult` is sorted, canonicalized, and deduplicated.
+    """
+
+    def __init__(self, reference_resolver: Jinja2ReferenceResolver | None = None) -> None:
+        self._reference_resolver = reference_resolver or Jinja2ReferenceResolver()
```
polmichel: This is maybe just a personal opinion:
This is a nice dependency injection.
But having it implicit kind of freezes the intended usage of the object and makes it unclear that there is a dependency, for the developer who uses this object. 

Of course, this is lighter to read at object instantiation, but I do not think we gain that much with only one object that has no constructor parameters.

polmichel: Maybe a convention we would need to agree on, since this kind of DI review will hopefully happen a lot from now on

backend/infrahub/git/closure_builder/jinja2_closure.py
```diff
@@ -0,0 +1,98 @@
...
+                    current=current,
+                    worktree_root=worktree_root,
+                    worktree_real=worktree_real,
+                    loader=loader,
+                    env=env,
+                )
+                if isinstance(resolved, UnresolvedRef):
+                    complete = False
+                    unresolved.append(resolved)
+                    continue
```
polmichel, line 93: An optional thought: likewise, there could be a tiny object named `Reference` that would contain only the content and a method `.resolved()` that returns `False` when called on an `UnresolvedRef` object. It would save us from relying on an `isinstance()` check, which I think creates quite fragile code and wrongly placed responsibility in the long run.

backend/infrahub/git/closure_builder/protocols.py
```diff
@@ -18,10 +18,8 @@
 class ClosureBuilder(Protocol):
```
polmichel: Is this used somewhere? Maybe in a later PR?

backend/infrahub/message_bus/types.py
```diff
@@ -89,7 +89,7 @@ class ProposedChangeArtifactDefinition(BaseModel):
     definition_id: str
     definition_name: str
     artifact_name: str
-    query_name: str  # Deprecated
+    query_name: str
```
polmichel, line 92: This is not Deprecated anymore? It seems to come from a commit from Fatih in 2025

backend/tests/unit/git/closure_builder/test_dispatcher.py
```diff
@@ -0,0 +1,111 @@
...
+    with caplog.at_level(logging.ERROR, logger="test_dispatcher"):
+        result = build_transform_closure(
+            transform_config=config,
+            worktree_root=tmp_path,
+            logger=logger,
+        )
+
+    assert result.complete is False
+    assert result.dependencies == ()
+    assert any("missing-entry" in record.message for record in caplog.records)
```
polmichel: We have discussed in a precedent discussion with @ajtmccarty and @gmazoyer that asserting the whole error message is better in the long run. What's your thoughts on that? Maybe this is different since there seems to be multiple message in this test.

polmichel: Maybe something to discuss with the backend team, and if true to add to the Claude tests rules?

polmichel (commented):
Besides the comments on deprecated element removed and code that do not seem to be used, these are only optional comments, and maybe most of them are subjective, so feel free to ignore or postpone those.

The test coverage seems to be super good and all this logic well encapsulated in a dedicated package.

backend/infrahub/git/closure_builder/jinja2_closure.py
ajtmccarty: I'd say make the `reference_resolver` required. [PEP 20](https://peps.python.org/pep-0020/) "Explicit is better than implicit." just always require the dependencies for the class to be passed into the constructor for new code. in old code we have to do this to minimize the size of a change in some cases, but it is like optional dependency injection.

or maybe you've got a good reason for this that I am missing

backend/infrahub/git/closure_builder/dispatcher.py
```diff
@@ -0,0 +1,57 @@
...
+
+ISOLATED_FAILURES: tuple[type[BaseException], ...] = (
+    ValueError,
+    OSError,
+    TemplateError,
+    GitCommandError,
+)
+
+
+def build_transform_closure(
```
ajtmccarty: this looks like an `AggregatedTransformClosureBuilder` to me that accepts a list of implementations of a `TransformClosureBuilderInterface` class

```python
class TransformClosureBuilderInterface:
    def supports(self, transform_config) -> bool: ...
    def build(self, transform_config, worktree_root) -> ClosureResult: ...

class Jinja2ClosureBuilder(TransformClosureBuilderInterface):
    def support(self, transform_config):
        return isinstance(transform_config, InfrahubJinja2TransformConfig)

class PythonClosureBuilder(TransformClosureBuilderInterface):
    def support(self, transform_config):
        return isinstance(transform_config, InfrahubPythonTransformConfig)

class AggregatedTransformClosureBuilder:
    def __init__(self, closure_builders: list[TransformClosureBuilderInterface]): ...

    def build(self, transform_config, worktree_root) -> ClosureBuilder:
        for closure_builder in self.closure_builders:
            if closure_builder.supports(transform_config):
                return closure_builder.build(transform_config, worktree_root)
        raise NoClosureBuilderError(...)
```
not saying you have to do it this exact way, but the components that depend on other components should be classes with their dependencies injected.

there might be some way to type this with Generics to handle the same `match`, `case _: assert_never()` logic you've got here too

UPDATE: your `ClosureBuilder` protocol is basically what I'm suggesting with `TransformClosureBuilderInterface`

backend/infrahub/git/closure_builder/result.py
```diff
@@ -14,3 +16,20 @@ class ClosureResult:
     dependencies: tuple[str, ...]
     complete: bool
     unresolved: tuple[UnresolvedRef, ...]
+
+    def __post_init__(self) -> None:
```
ajtmccarty, line 20: could `dependencies` be a `set` to avoid this check?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -1220,6 +1227,42 @@ def _definition_changed(
...
+
+    closure = {canonicalize_path(entry) for entry in definition.dependencies}
+    changed_files: set[str] = set()
+    for raw in (*repo_diff.files_added, *repo_diff.files_changed, *repo_diff.files_removed):
+        try:
+            changed_files.add(canonicalize_path(raw))
+        except ValueError:
+            continue
+
+    return not closure.isdisjoint(changed_files)
```
ajtmccarty: I find the double negative hard to read
can it be `bool(closure & change_files)`?

backend/tests/unit/git/closure_builder/test_jinja2_closure.py
```diff
@@ -0,0 +1,262 @@
...
+def test_entry_path_escaping_the_worktree_is_rejected(tmp_path: Path) -> None:
+    """A `template_path` that resolves outside the worktree is rejected before any read.
+
+    The entry path is user-controlled via `.infrahub.yml`. Without an early
+    boundary check, a `template_path` like ``../../../etc/passwd`` would resolve
+    outside the worktree and be read by the closure walker. The boundary check
+    on the entry path mirrors the one applied to transitively-discovered
+    references and short-circuits the walk before any I/O.
+    """
+    outside_dir = tmp_path.parent / "outside_worktree_entry"
```
ajtmccarty: could going outside of the `tmp_path` in a test be an issue?

ajtmccarty (approved):
approving b/c it looks good functionally. I just have some SOLID comments, which are everyone's favorite

backend/infrahub/message_bus/types.py
ogenstad: Yes, this was reintroduced again in order to have the name for logging. Though that entry will actually come later.

backend/infrahub/git/closure_builder/result.py
ogenstad: I want them to be in a known order so keeping as is.

PR #9453, opened 2026-06-03
support all order_by options in graphql ordering overrides

ajtmccarty:
this was not technically part of the spec/plan for INFP-530 / IFC-2513, but it is not too big of a change and I think it makes the order_by work feel much more complete.

before this change, users would only have the option to order by metadata (update/created_at__asc/desc) within a graphql query, which feels like kind of a waste b/c the cypher queries behind these graphql queries support the full range of order by options: attribute values, attribute values on relationships, object metadata, and ascending/descending. this PR make a few changes so that all of the schema `order_by` options are also supported at different levels of our graphql queries: root level, many relationships, and hierarchical ancestors/descendants relationships.

---
## Summary by cubic
Add full GraphQL ordering overrides with `order_by` across root, many-relationship, and hierarchy fields. Branch queries are restricted to metadata-only ordering; query-time `order` fully replaces schema defaults. Aligns with INFP-530.

- **New Features**
  - `OrderInput` adds `order_by: [String!]` using schema grammar (e.g. `name__value__desc`).
  - Supported on root fields, many relationships, and hierarchy (`ancestors`/`descendants`).
  - Branch uses `MetadataOrderInput` only; no `order_by` or `disable`.
  - Validation: cannot combine `node_metadata` with `order_by`; cannot use both `created_at` and `updated_at`; parser now raises `ValidationError` for unknown fields/bad directions.

- **Migration**
  - Query-time `order` replaces schema `order_by` defaults—include all desired keys in `order_by` or omit `order`.
  - `node_metadata` is reserved—rename any attribute/relationship using that name.
  - Update Branch queries to `MetadataOrderInput` and remove any `order_by` or `disable` usage.

<sup>Written for commit 744832130429ce5e38122bf883ec517dc1a7a045. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/order.py
```diff
@@ -29,12 +29,23 @@ class OrderModel(BaseModel):
...
+    order_by: tuple[str, ...] | None = None
 
     def __bool__(self) -> bool:
-        return bool(self.disable) or bool(self.node_metadata)
+        return bool(self.disable) or self.has_explicit_entries
+
+    @property
+    def has_explicit_entries(self) -> bool:
+        """True when the model carries override ordering entries (metadata or order_by)."""
+        return bool(self.node_metadata) or bool(self.order_by)
```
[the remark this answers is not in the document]
ajtmccarty: I don't think this matters. this is what OrderInput.disable is for

backend/infrahub/core/query/node.py
ajtmccarty, line 1: pass the ordering override into `NodeGetListQuery` and `NodeGetHierarchyQuery`

backend/infrahub/core/order.py
```diff
@@ -29,12 +29,23 @@ class OrderModel(BaseModel):
...
 
     @model_validator(mode="after")
     def validate_metadata(self) -> Self:
         if self.node_metadata and self.node_metadata.created_at and self.node_metadata.updated_at:
             raise ValidationError("Cannot order by both created_at and updated_at simultaneously.")
+        if self.node_metadata and self.order_by:
```
ajtmccarty, line 46: continue allowing setting `node_metadata` for backwards compatibility, but it is also supported in the new `order_by` list of strings

backend/infrahub/graphql/types/metadata.py
```diff
@@ -10,9 +10,31 @@ class InfrahubNodeMetadataOrder(InputObjectType):
...
+    "Ordering overrides support attributes (`name__value__desc`), "
+    "relationship attributes (`owner__name__value`), or "
+    "node metadata (`node_metadata__created_at__desc`). The trailing "
+    "`__asc`/`__desc` is optional (default is ascending). When provided, "
+    "fully replaces the schema's `order_by`. Cannot be combined with "
+    "`node_metadata` in the same input."
+)
+
+
+class MetadataOrderInput(InputObjectType):
```
ajtmccarty, line 23: new input type for the `Branch` query that does not include `disable` or the new `order_by` b/c neither is supported in `Branch`s get list cypher query

backend/infrahub/core/query/node.py
```diff
@@ -2580,12 +2590,41 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
     async def _add_peer_order_by(
         self, db: InfrahubDatabase, hierarchy_schema: NodeSchema | GenericSchema, branch_filter: str
     ) -> None:
-        if not (hasattr(hierarchy_schema, "order_by") and hierarchy_schema.order_by):
+        if self.requested_order and self.requested_order.disable:
```
ogenstad, line 2593: Not something for this PR but looking forward it feels like we could make this easier where we'd have some self.ordering option that was populated by default and this line would be:

```python
if self.ordering.disable:
```

The next one would be:

```python
if self.ordering.requested_order_by:
...
elif self.ordering.schema_order_by:
```

ogenstad (commented):
Nice addition! 

Would be nice to have some features for this in the frontend as well.

It would be good if you could validate from the SDK as well, while it would be good to add the features now as well we would need to just validate that the old metadata ordering from the SDK still works as intended. It looks like nothing should have changed with that but could be nice with a sanity check.

One change that would be good though is that now we expose some internals from a previous PR that were never exposed earlier and we need all errors that can be raised by GraphQL to have it's root in our own infrahub.exceptions.Error class.

If we look at a query like this:

```graphql
query MyQuery {
  BuiltinTag(order: {order_by: ["name_stuff"]}) {
    edges {
      node {
        display_label
      }
    }
  }
}
```

The error we get back from the API looks like this:

```json
{
  "data": null,
  "errors": [
    {
      "message": "attribute 'name_stuff' not defined on this schema (entry: 'name_stuff').",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "BuiltinTag"
      ],
      "extensions": {
        "code": "UNDEFINED_ERROR",
        "http_status": 500,
        "data": {}
      }
    }
  ]
}
```

We also see a traceback in the server logs:

```python
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/graphql/execution/execute.py", line 530, in await_result
    return_type, field_nodes, info, path, await result
                                          ^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/opentelemetry/util/_decorator.py", line 71, in async_wrapper
    return await func(*args, **kwargs)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/database/__init__.py", line 524, in wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/graphql/resolvers/resolver.py", line 252, in default_paginated_list_resolver
    objs = await NodeManager.query(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/manager.py", line 177, in query
    query = await NodeGetListQuery.init(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/query/__init__.py", line 403, in init
    await query.query_init(db=db, **kwargs)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/query/node.py", line 1936, in query_init
    field_attribute_requirements = self._get_field_requirements(disable_order=disable_order)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/query/node.py", line 2328, in _get_field_requirements
    return filter_requirements + self._get_order_requirements(filter_requirements)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/query/node.py", line 2305, in _get_order_requirements
    for position, parsed in enumerate(self._get_parsed_schema_order_by()):
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/query/node.py", line 1710, in _get_parsed_schema_order_by
    parse_order_by_entry(entry=entry, node_schema=self.schema) for entry in self.requested_order.order_by
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/schema/order_by.py", line 94, in parse_order_by_entry
    return _parse_path(entry=entry, parts=parts, node_schema=node_schema)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/core/schema/order_by.py", line 175, in _parse_path
    raise ValueError(f"attribute {first!r} not defined on this schema (entry: {entry!r}).")
ValueError: attribute 'name_stuff' not defined on this schema (entry: 'name_stuff').
```

The change we'd need here is that `order_by.py` should not raise ValueError but instead raise a ValidationError (later we actually want a more specific error to be used based on the work with the enrichment of GraphQL errors but I think ValidationError would be good for now.

ogenstad (approved):

ajtmccarty:
updated `parse_order_by_entry()` to raise `ValidationError` instead of `ValueError` b/c it can make it to the end user now

PR #9456, opened 2026-06-04
fix(proposed-change): scope generator dispatch to fields the query reads

### LINKED ISSUE #9378: bug: PC pipeline dispatches generator instances that share a relationship target, plus unrelated generators that walk through the same object
### Component

API Server / GraphQL

### Infrahub version

1.9.4 (also reproduces on `stable` at 1.9.6)

### Current Behavior

When a Proposed Change touches an object that is referenced as a relationship endpoint by many generator instances (a "hub" object — e.g. `site=london`, a tenant, a hall), the PC pipeline dispatches generator runs against:

1. **Every generator instance that shares that relationship target** — even when the specific instance's own inputs were not changed by the PC.
2. **Generator instances of *other* generator definitions whose queries also walk through the same hub object** — even when those generators do not read any attribute on the hub at all and were not semantically affected.

### Expected Behavior

A Proposed Change should only dispatch a generator instance when the change actually affects that instance's inputs. Specifically:

- Adding a child object whose only relationship to a hub is "I point at it" should not be treated as modifying the hub for the purpose of selecting impacted generators.
- A generator whose query walks through a hub object only to reach a child's attribute should not be considered subscribed to changes on the hub itself. (This may have more to do with the structure of the GraphQL query so the primary issue is the one above)

### Steps to Reproduce

> This is reproducible with the [infrahub-bundle-dc](https://github.com/opsmill/infrahub-demo-dc) and I will provide a backup, etc. in corresponding internal issue.

1. Define two unrelated generator definitions, A and B, whose GraphQL queries both traverse a shared hub node (e.g. both walk `foo.tenant.id` or `foo.site.id`, but neither reads any hub attribute).
2. On `main`, run both generators so each has instances created against multiple targets sharing the same hub.
3. Cut a fresh branch from `main`.
4. Add a single new child object pointing at the hub (e.g. a new VLAN that points at the tenant).
5. Open a Proposed Change.
6. Observe that both generator A and B dispatch runs against every one of their instances, despite the new child being relevant to at most one instance of one generator.

### Additional Information

We've investigated this against our local server checkout and believe there are two reinforcing causes worth flagging for whoever picks this up:

1. **Query-group membership is too broad.** `Node.to_graphql()` adds every visited node to `related_node_ids` (the membership of the per-query `CoreGraphQLQueryGroup`), including transit-only nodes that the query walks through but never reads attributes from. Result: hub objects end up as members of every generator's query group, even generators that don't consume them.

2. **The branch-diff "modified" list is too broad.** `get_modified_node_ids()` in `backend/infrahub/proposed_change/branch_diff.py` does not filter on `entry["action"]` or on whether `entry["elements"]` contains any actual attribute change. UNCHANGED entries and UPDATED entries whose only `elements` are `RELATIONSHIP_MANY` endpoint flips both pass through. A node whose only "change" is "a peer started pointing at me" is treated as modified.

The fan-out comes from these two combining: hub object appears in the modified list because of (2), and then it impacts every generator's query group because of (1).

#### Local mitigations we applied

- Rewrote each affected generator's GraphQL query to drop unused traversals through hub nodes, refetching the few fields needed via `client.get` inside the generator's Python entry point (SDK reads don't update the query group). This breaks (1) for the rewritten queries without a server change.
- Added an in-container patch to `get_modified_node_ids()` that filters out UNCHANGED entries and UPDATED entries with no `ATTRIBUTE` element. With this patch alone, a re-trigger of an existing PC that previously fired all 11 `hall_deployment` instances now fires only 2 (the ADDED HallDeployment plus 1 unrelated one whose `status` legitimately changed).

We don't have a strong view on the right shape of the upstream fix and have not opened a PR. Either tightening the query-group composition or tightening the modified-node filter would help; doing both would resolve the cascade we observe.

ogenstad:
# Why

Within a proposed change, generator instances were dispatched for every instance sharing a relationship target with a changed node, even when the change touched no field the generator's GraphQL query actually reads. This ports the field-level filter that artifacts already use to the generator path.

A shared helper resolves, from a query payload and the branch diff, which subscribers are impacted: the specific instances whose read fields changed (unique-target queries), all instances (non-unique query with a relevant change), or none. The artifact path is refactored onto the same helper with no change in behaviour.

Removes the now-unused subscriber plumbing whose only consumer was the generator instance check: get_subscribers_ids, _get_subscribers_from_diff, get_modified_node_ids, and the ProposedChangeBranchDiff.subscribers field.

Fixes #9378

Note that while the issue was marked as a bug this is also a new change of behavior that could use some more testing and as such might be better served as a performance improvement for 1.10.

The implementation follows what https://github.com/opsmill/infrahub/pull/8573 did for artifacts along with some improvements with the introduction of the `ImpactedSubscribers` dataclass to make it easier to understand when something should be regenerated as the previous method returned an optional list so `None` or `[]` would mean very different things which wasn't clear in the code, while we had code comments using a dedicated class should make this easier to understand.

## What changed

**Behavioral changes:**
- Generator instances in a proposed change are now only re-run when the diff
  modifies a field (attribute or relationship) that the generator's query
  reads. A relationship endpoint flip on a relationship the query doesn't
  read no longer triggers a run.
- Non-unique-target queries still fall back to running all instances when a
  read field changes (safe over-generation), and git-synced branches with
  file changes still run everything. Both unchanged.

**Implementation notes:**
- New shared helper `get_field_level_impacted_subscribers` resolves, from a
  query payload plus branch diff, an `ImpactedSubscribers` result with scope
  `SPECIFIC` (unique-target queries map to exact instances), `ALL`
  (non-unique query with a relevant change), or `NONE`. Used by both the
  artifact and generator paths.
- The field-matching core is extracted as a pure function
  (`_relevant_node_changes`) and unit-tested without infrastructure.
- `query_payload` added to `ProposedChangeGeneratorDefinition`, populated at
  all construction sites.
- Removed the now-unused subscriber plumbing whose only consumer was the
  generator instance check: `get_subscribers_ids`,
  `_get_subscribers_from_diff`, `get_modified_node_ids`, and the
  `ProposedChangeBranchDiff.subscribers` field.

**What stayed the same:**
- Artifact regeneration behavior is byte-for-byte equivalent to before
  (dispatched set and log lines). The refactor onto the shared helper is
  behavior-preserving, verified against the existing artifact test suite.
- No schema changes; no GraphQL/REST contract changes.

### Suggested review order
1. `get_field_level_impacted_subscribers` plus `_relevant_node_changes` in
   `backend/infrahub/proposed_change/tasks.py` (the core logic).
2. The two callers (`validate_artifacts_generation`,
   `request_generator_definition_check`). Note they map onto the same
   three-scope result.
3. The dead-code removal across `tasks.py`, `branch_diff.py`, `types.py`.
4. Tests.

## How to review

Focus area: the field-level matching and the unique/non-unique branching.
That's where a wrong call would either skip a needed run or over-generate.

Risk to scrutinize: behavior equivalence of the refactored artifact path.
It's guarded by the existing `test_validate_artifacts_generation.py` suite
(unchanged assertions) plus a diff against the original inline logic.

---
## Summary by cubic
Scopes generator instance dispatch to only changes that affect fields read by the generator’s GraphQL query, sharply reducing fan-out in Proposed Changes. Reuses the artifact field-level filter for generators and adds tests. Fixes #9378 and addresses IFC-2657.

- **Bug Fixes**
  - Generators now dispatch only when a queried field changes; unique-target queries run matching instances, non-unique run all only if a relevant field changed.
  - Added `query_payload` to `ProposedChangeGeneratorDefinition` and wired it through `run_generators` and generator mutations.
  - Updated PC pipeline to compute impacted subscribers from the query + diff; added unit and component tests; added changelog.

- **Refactors**
  - Introduced a shared helper to map branch diffs to impacted subscribers; artifacts now use it with no behavior change.
  - Removed unused subscriber plumbing: `ProposedChangeBranchDiff.subscribers`, `get_subscribers_ids`, `_get_subscribers_from_diff`, `get_modified_node_ids`.

<sup>Written for commit fb0a381407d34f1a19139bc7314dabad1acb9141. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9456?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import asyncio
 import os
 import sys
-from dataclasses import dataclass
-from enum import IntFlag
+from dataclasses import dataclass, field
+from enum import Enum, IntFlag
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -76,7 +76,6 @@ from infrahub.message_bus.types import (
 from infrahub.proposed_change.branch_diff import (
     GitRepositoryFileDiffer,
     RepositoryFileDiffPopulator,
-    get_modified_node_ids,
     has_data_changes,
     has_node_changes,
     set_diff_summary_cache,
@@ -118,6 +117,7 @@ from .checker import verify_proposed_change_is_mergeable
 
 if TYPE_CHECKING:
     import logging
+    from uuid import UUID
 
     from infrahub_sdk.client import InfrahubClient
     from infrahub_sdk.diff import NodeDiff
@@ -358,6 +358,7 @@ async def run_generators(model: RequestProposedChangeRunGenerators, context: Inf
             file_path=generator.file_path.value,
             query_name=generator.query.peer.name.value,
             query_models=generator.query.peer.models.value,
+            query_payload=generator.query.peer.query.value,
             repository_id=generator.repository.peer.id,
             parameters=generator.parameters.value,
             group_id=generator.targets.peer.id,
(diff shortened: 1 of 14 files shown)


backend/infrahub/proposed_change/tasks.py
```diff
@@ -683,6 +684,103 @@ def _execute(
...
+    readable_fields_by_kind = {kind: access.fields for kind, access in query_report.requested_read.items()}
+    changed_node_ids = _relevant_node_changes(
+        diff_summary=diff_summary, source_branch=source_branch, readable_fields_by_kind=readable_fields_by_kind
+    )
+
+    # only_has_unique_targets is True when the query is guaranteed to return results for a
+    # specific set of nodes -- e.g. it uses an `ids` argument or a uniqueness constraint. When
+    # False, the query may return any number of nodes and we cannot map a changed node back to a
+    # specific subscriber without re-processing every target.
+    if query_report.only_has_unique_targets:
```
polmichel, line 766: If I understand this correctly, it means that a new good practice would be to always ask for the nodes `id` in generator-related queries (even if the `id` is not used afterwards)? If true, would it be relevant to write some documentation about this somewhere?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -683,6 +684,103 @@ def _execute(
...
+        # which subscribers are linked to the changed nodes and limit processing to only those.
+        subscribers = await _get_subscribers_for_nodes(node_ids=changed_node_ids, branch=source_branch, client=client)
+        ids = [subscriber.subscriber_id for subscriber in subscribers if subscriber.kind == subscriber_kind]
+        return ImpactedSubscribers(scope=ImpactScope.SPECIFIC, ids=ids)
+
+    if changed_node_ids:
+        # The query does not guarantee unique targets, so we cannot determine which specific
+        # subscribers are affected. At least one relevant field changed, so the caller must fall
+        # back to processing all targets to be safe.
+        return ImpactedSubscribers(scope=ImpactScope.ALL)
```
polmichel, line 777: If the precedent comment about the good practice is true, maybe highlighting the absence of `id` or `uniqueness_constraints` there would help the user to enhance the performance of the application?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -683,6 +684,103 @@ def _execute(
...
+    # False, the query may return any number of nodes and we cannot map a changed node back to a
+    # specific subscriber without re-processing every target.
+    if query_report.only_has_unique_targets:
+        # The query targets specific nodes by id or unique constraint, so we can look up exactly
+        # which subscribers are linked to the changed nodes and limit processing to only those.
+        subscribers = await _get_subscribers_for_nodes(node_ids=changed_node_ids, branch=source_branch, client=client)
+        ids = [subscriber.subscriber_id for subscriber in subscribers if subscriber.kind == subscriber_kind]
+        return ImpactedSubscribers(scope=ImpactScope.SPECIFIC, ids=ids)
+
+    if changed_node_ids:
```
polmichel, line 773: Optional/Nitpick and maybe subjective comment: as the precedent condition block is ending with an unconditional `return` maybe a `elif` would make the code more readable: these cases are exclusive.

backend/infrahub/proposed_change/tasks.py
polmichel: I just saw you've done it in the caller's code.

backend/tests/unit/proposed_change/test_tasks.py
polmichel, line 1: Really like this kind of unit test suite. This is at the same time an help for the review, a specification and a protection against regressions if a refactoring occurs (even if no refactor should occur as this component seems cohesive)

polmichel (approved):
LGTM! Only nitpicking comments have been posted. Curious about the scale of the performance enhancement in real-life scenarios.

backend/infrahub/proposed_change/tasks.py
ogenstad: Introducing an `elif` causes a warning in the linter based on https://docs.astral.sh/ruff/rules/superfluous-else-return/, so I guess it's a matter of taste but leaving it as it is now.

backend/infrahub/proposed_change/tasks.py
ogenstad: We don't need to ask for ids specifically but they can be useful for the filter, but there are other options as well.

It's covered here: https://docs.infrahub.app/development-resources/graphql/single-target-queries

backend/infrahub/proposed_change/tasks.py
ogenstad: As mentioned earlier I think this should be covered in the docs and the goal in the code is more to highlight the options without going into the features too much.

PR #9459, opened 2026-06-04
feat(artifacts): regeneration diagnostics and safe per-transform rollout (INFP-409)

ogenstad:
# Why

When a proposed change includes commits to a linked repository, the regeneration
gate already decided which artifacts to rebuild, but gave no visibility into why. A
user seeing a definition's artifacts regenerate had to cross-reference the branch diff
and the database to understand the cause. Separately, transforms imported before the
precise-gate feature, or whose dependency closure cannot be fully resolved, needed a
path that keeps working without operator intervention.

Goal: make every regeneration decision explainable from the proposed-change task log,
and guarantee that pre-feature and incomplete-closure transforms keep working and
self-heal.

## What changed

Behavioral changes:
- The proposed-change task log now records, per definition, why its artifacts
  regenerate: the GraphQL query (name and id), the changed definition field(s), or the
  specific file inside the transform's dependency closure.
- The Jinja2 closure builder logs every unresolved reference it encounters at import.
- A transform whose closure cannot be trusted (`dependencies=null` for pre-feature
  transforms, or `dependencies_complete=False`) falls back to regenerate-on-any-change,
  and the log states it will self-heal on the transform's next import.
- A closure-builder failure on one transform no longer affects sibling transforms in
  the same repository; each still imports with a full closure.

Implementation notes:
- The three selection predicates are now pure functions returning a
  `PredicateOutcome(matched, reason)` instead of a bare bool; the refresh gate owns
  logging and emits `reason`. This separates the decision from the explanation, and
  addresses the SRP and explicit-DI review feedback from the precise-gate PR.
- `logger` is now a required dependency across the closure builders; the previous
  test-only `None` path was removed.


ajtmccarty (approved):
looks like the artifact_regen test files could maybe have some components extracted, but not completely sure and doesn't need to prevent merging

PR #9445, opened 2026-06-03
update changelog guidance for AGENTS.md

ajtmccarty:
Claude and friends seem to have been adding changelogs by hand in the wrong place (fixed in #9436)
these updates are hopefully enough to prevent it going forward

---
## Summary by cubic
Updated changelog guidance to require `towncrier create` and clarify that fragments always live in the repo-root `changelog/` directory. Added reminders in `backend/AGENTS.md` and expanded `dev/guidelines/changelog.md` to prevent manual fragment files and incorrect locations.

<sup>Written for commit 70188e14edfa6420f9a26eaaf19c467cc72ad3f9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9445?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #9460, opened 2026-06-04
fix(git): register transforms before objects during repository import [#9406]

### LINKED ISSUE #9406: Git import loads objects before transforms, breaking same-repo refs
### Component

Backend / Git integration (repository import)

### Infrahub version

1.9.6

### Current Behavior

During repository (git) import, `objects/` are imported **before** the `.infrahub.yml` `python_transforms` are registered. An object that references a `CoreTransformPython` defined in the **same** repository therefore fails to resolve. For example, a `CoreCustomWebhook` with `transformation: "MyTransform"` aborts the import with:

```
Unable to find the node MyTransform / CoreTransformPython in the database.
```

The whole repository import aborts and the repository goes to `sync_status: error-import`. Because the failure happens during object import, the transforms/checks/generators are never registered either, so a retry can't converge.

The order is fixed in `backend/infrahub/git/integrator.py` (`import_all`, ~lines 201–216):

1. `import_schema_files`
2. `import_all_graphql_query`
3. `import_objects` ← objects referencing a `CoreTransformPython` fail here
4. `import_all_python_files` → `import_python_transforms` (transforms registered here)

So any object referencing a repo-defined Python artifact (transforms, and presumably checks/generators registered in step 4) cannot be imported.

### Expected Behavior

Either objects that reference repo-defined Python artifacts should resolve (e.g. register Python files before importing objects, or resolve such references in a dependency-aware / two-pass manner); **or**, at minimum, the importer should emit a clear, actionable error naming the offending object file and the unresolved transform, instead of a generic "Unable to find the node …" that aborts the entire import.

### Steps to Reproduce

1. Define a transform in `.infrahub.yml`:

   ```yaml
   python_transforms:
     - name: MyTransform
       class_name: MyTransform
       file_path: transforms/my_transform.py
   ```

2. Add an object file (e.g. `objects/00_webhooks.yml`) referencing it:

   ```yaml
   apiVersion: infrahub.app/v1
   kind: Object
   spec:
     kind: CoreCustomWebhook
     data:
       - name: my-webhook
         event_type: infrahub.node.created
         node_kind: SomeNode
         url: https://example.com/webhook
         transformation: "MyTransform"
   ```

3. Add the repository to Infrahub and let it sync.
4. Observe `sync_status: error-import` with "Unable to find the node MyTransform / CoreTransformPython in the database." No transforms or webhooks are created.

### Additional Information

- Reproduced via `infrahub-testcontainers` (Infrahub server 1.9.6) on macOS arm64.
- Workaround: don't reference repo-defined Python transforms from object files in the same repository; create such webhooks out-of-band (UI/API) after the transform exists, or omit the `transformation` link.
- Related but distinct root cause: #9248 (also "repository import fails", but caused by the group peer-kind allowlist).

lancamat1:
## Summary

During repository import, `objects/` were imported **before** the `.infrahub.yml` Python transforms were registered. An object referencing a `CoreTransformPython` defined in the **same** repository (e.g. a `CoreCustomWebhook` with `transformation: "MyTransform"`) therefore aborted the entire import with:

```
Unable to find the node MyTransform / CoreTransformPython in the database.
```

…leaving the repository in `sync_status: error-import`. Because the failure happened during object import, transforms/checks/generators were never registered either, so a retry could not converge.

Fixes #9406 (IFC-2669).

## Change

Reorder `InfrahubRepositoryIntegrator.import_objects_from_files` so the end-to-end order is:

1. Schemas
2. GraphQL queries
3. **Transforms** (Python + Jinja2) — moved ahead of objects
4. Objects (+ menus)
5. Checks, generator definitions, artifact definitions

Transforms move *before* objects so same-repo references resolve. Generators and artifact definitions stay *after* objects because their `targets` reference groups created from the repository's own object files — moving them up would trade one failure for another. This splits the previously-bundled `import_all_python_files` call accordingly (the method is retained for its existing integration-test caller).

No schema, GraphQL, or database changes.

## Tests

Extends the existing full-stack `integration_docker` `car-dealership` fixture and `test_propose_change_repository.py` (no new fixture repo):

- **US1** — a `CoreCustomWebhook` referencing the repo's existing `WebhookTransformer` Python transform; asserts `transformation` resolves after import.
- **US2** — a `CoreStandardGroup` defined as a repo object plus a generator definition targeting it; asserts the generator's `targets` resolves to the same-repo group (guards the objects-before-generators side of the ordering).

Validation:
- ✅ Against a source build (`:local`): `test_propose_change_repository.py` 4/4 pass.
- ❌ Against the released `1.9.6` image (no fix): `test_load_initial_data` fails with the exact issue error `Unable to find the node WebhookTransformer / CoreTransformPython` — confirming the tests are genuine regression guards.

## Notes

- Also adds `package-lock.json` to `frontend/app/.gitignore` (this project uses pnpm; an stray npm lockfile kept appearing).
- `dev/specs/ifc-2669-fix-import-order/` contains the spec/plan/tasks design record for the ticket.

---
## Summary by cubic
Fixes repository import ordering so Python and Jinja2 transforms register before objects, allowing same-repo references (e.g., `CoreCustomWebhook.transformation`) to resolve. Keeps checks, generator definitions, and artifact definitions after objects so their `targets` to repo-defined groups still resolve. Aligns with IFC-2669.

- **Bug Fixes**
  - Reordered `InfrahubRepositoryIntegrator.import_objects_from_files` to: Schemas → GraphQL queries → Transforms (Python + Jinja2) → Objects → Checks/Generators/Artifacts.
  - Removed `import_all_python_files`; tests now call `import_python_check_definitions` and `import_python_transforms` explicitly.
  - Added a dedicated `same-repo-references` fixture and full-stack `integration_docker` test (`TestSameRepoReferences`) that verifies a webhook resolves its same-repo transform and a generator resolves a same-repo group.

<sup>Written for commit 94f7bd840f3a2d9802818194fbf0621cb1fbee47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9460?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/integrator.py
+++ b/backend/infrahub/git/integrator.py
@@ -204,13 +204,25 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
             await self.import_all_graphql_query(
                 branch_name=infrahub_branch_name, commit=commit, config_file=config_file
             )  # type: ignore[call-overload]
+            # Transforms must be registered before objects so that an object referencing a transform
+            # defined in the same repository resolves during import.
+            await self.import_python_transforms(
+                branch_name=infrahub_branch_name, commit=commit, config_file=config_file
+            )  # type: ignore[call-overload]
+            await self.import_jinja2_transforms(
+                branch_name=infrahub_branch_name, commit=commit, config_file=config_file
+            )  # type: ignore[call-overload]
             await self.import_objects(
                 branch_name=infrahub_branch_name,
                 commit=commit,
                 config_file=config_file,
             )  # type: ignore[call-overload]
-            await self.import_all_python_files(branch_name=infrahub_branch_name, commit=commit, config_file=config_file)  # type: ignore[call-overload]
-            await self.import_jinja2_transforms(
+            # Checks, generators and artifact definitions are imported after objects because their
+            # targets reference groups that are defined as objects in the repository.
+            await self.import_python_check_definitions(
+                branch_name=infrahub_branch_name, commit=commit, config_file=config_file
+            )  # type: ignore[call-overload]
+            await self.import_generator_definitions(
                 branch_name=infrahub_branch_name, commit=commit, config_file=config_file
             )  # type: ignore[call-overload]
             await self.import_artifact_definitions(
@@ -1217,16 +1229,6 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
         ):
             return False
         return True
-
-    @flow(name="import-python-files", flow_run_name="Import Python file")
-    async def import_all_python_files(
-        self, branch_name: str, commit: str, config_file: InfrahubRepositoryConfig
-    ) -> None:
-        await add_tags(branches=[branch_name], nodes=[str(self.id)])
-
-        await self.import_python_check_definitions(branch_name=branch_name, commit=commit, config_file=config_file)  # type: ignore[call-overload]
-        await self.import_python_transforms(branch_name=branch_name, commit=commit, config_file=config_file)  # type: ignore[call-overload]
-        await self.import_generator_definitions(branch_name=branch_name, commit=commit, config_file=config_file)  # type: ignore[call-overload]
 
     @task(name="jinja2-template-render", task_run_name="Render Jinja2 template", cache_policy=NONE)
     async def render_jinja2_template(self, commit: str, location: str, data: dict) -> str:
--- a/backend/tests/fixtures/repos/same-repo-references/initial__main/.infrahub.yml
+++ b/backend/tests/fixtures/repos/same-repo-references/initial__main/.infrahub.yml
@@ -1,0 +1,23 @@
+---
+# An object references a Python transform defined in this repository, and a generator definition
+# targets a group defined as an object in this repository.
+queries:
+  - name: tags_query
+    file_path: "tags_query.gql"
+
+python_transforms:
+  - name: TagsTransform
+    class_name: TagsTransform
+    file_path: "transforms/tags_transform.py"
+
+generator_definitions:
+  - name: tags_generator
+    file_path: "generators/tags_generator.py"
+    targets: repo_defined_group
+    query: tags_query
+    execute_in_proposed_change: false
+    execute_after_merge: false
+
+objects:
+  - objects/groups.yml
+  - objects/webhooks.yml
--- a/backend/tests/integration/git/test_git_repository.py
+++ b/backend/tests/integration/git/test_git_repository.py
@@ -192,22 +192,15 @@ class TestInfrahubClient:
         with pytest.raises(NodeNotFoundError):
             await client.get(kind=CoreGraphQLQuery, id=obj.id)
 
-    async def test_import_all_python_files(
+    async def test_import_python_definitions(
         self, db: InfrahubDatabase, client: InfrahubClient, repo: InfrahubRepository, query_99: Node
     ) -> None:
-        for group in ["backbone_services", "maintenance_circuits", "provisioning_circuits", "upstream_interfaces"]:
-            obj = await Node.init(schema=InfrahubKind.STANDARDGROUP, db=db)
-            await obj.new(
-                db=db,
-                name=group,
-            )
-            await obj.save(db=db)
-
         commit = repo.get_commit_value(branch_name="main")
         config_file = await repo.get_repository_config(branch_name="main", commit=commit)  # type: ignore[call-overload]
         assert config_file
 
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
 
         check_definitions = await client.all(kind=CoreCheckDefinition)
         assert len(check_definitions) >= 1
@@ -217,7 +210,8 @@ class TestInfrahubClient:
 
         # Validate if the function is idempotent, another import just after the first one shouldn't change anything
         nbr_relationships_before = await count_relationships(db=db)
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
         assert await count_relationships(db=db) == nbr_relationships_before
 
         # 1. Modify an object to validate if its being properly updated
@@ -259,7 +253,8 @@ class TestInfrahubClient:
         )
         await obj2.save(db=db)
 
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
 
         modified_check0 = await client.get(kind=CoreCheckDefinition, id=check_definitions[0].id)
         assert modified_check0.timeout.value == check_timeout_value_before_change
--- a/backend/tests/integration_docker/test_repositories.py
+++ b/backend/tests/integration_docker/test_repositories.py
@@ -46,3 +46,36 @@ class TestRepositoryOperationalStatus(TestInfrahubDockerClient):
         repos = await client.all(kind=repo_kind)
         assert len(repos) == 1
         assert repos[0].operational_status.value == RepositoryOperationalStatus.ONLINE.value
+
+
+class TestSameRepoReferences(TestInfrahubDockerClient):
+    async def test_same_repo_references_resolve_on_import(
+        self,
+        client: InfrahubClient,
+        remote_repos_dir: Path,
+    ) -> None:
+        fixture_dir = get_fixtures_dir()
+        repo_dir = fixture_dir / "repos" / "same-repo-references" / "initial__main"
+        repo = GitRepo(
+            type=GitRepoType.INTEGRATED,
+            name="same-repo-references",
+            src_directory=repo_dir,
+            dst_directory=remote_repos_dir,
+        )
+        await repo.add_to_infrahub(client=client)
+        in_sync = await repo.wait_for_sync_to_complete(client=client)
+        assert in_sync
+
+        # An object that references a Python transform defined in the same repository should be
+        # correctly resolved.
+        webhook = await client.get(kind="CoreCustomWebhook", name__value="tags-webhook")
+        assert webhook.transformation.id is not None
+        transform = await client.get(kind="CoreTransformPython", id=webhook.transformation.id)
+        assert transform.name.value == "TagsTransform"
+
+        # A generator definition whose target group is defined as an object in the same repository
+        # should also be resolved correctly.
+        generator = await client.get(kind="CoreGeneratorDefinition", name__value="tags_generator")
+        assert generator.targets.id is not None
+        group = await client.get(kind="CoreStandardGroup", name__value="repo_defined_group")
+        assert generator.targets.id == group.id
--- a/backend/tests/fixtures/repos/same-repo-references/initial__main/generators/tags_generator.py
+++ b/backend/tests/fixtures/repos/same-repo-references/initial__main/generators/tags_generator.py
@@ -1,0 +1,6 @@
+from infrahub_sdk.generator import InfrahubGenerator
+
+
+class Generator(InfrahubGenerator):
+    async def generate(self, data: dict) -> None:
+        return
--- a/backend/tests/fixtures/repos/same-repo-references/initial__main/objects/groups.yml
+++ b/backend/tests/fixtures/repos/same-repo-references/initial__main/objects/groups.yml
@@ -1,0 +1,7 @@
+---
+apiVersion: infrahub.app/v1
+kind: Object
+spec:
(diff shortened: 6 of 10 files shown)


backend/tests/fixtures/repos/same-repo-references/initial__main/.infrahub.yml
```diff
@@ -0,0 +1,24 @@
+---
+# Fixture exercising same-repository references during import (issue #9406):
+# - an object (CoreCustomWebhook) references a Python transform defined in this repo (US1)
+# - a generator definition targets a group defined as an object in this repo (US2)
```
polmichel: Usually we tend to not mention GitHub issues, spec-kit elements or Jira items inside the comments or descriptions. The reason is that we want this documentation to describe the expected behavior of the code, not to explain why this code is there or which older problem it solves.

This would flood the code with information that is not what we (or the agents) are looking for when the code is scanned, consuming intellectual bandwidth (or tokens).
If we want to dig into such information, we could use the `git blame` capabilities and the related PRs/GH issues/Jira items, which are supposed to describe the problem and the intent.

We've created some Claude rules for this matter (`.claude/rules/code-doc-style.md`) but we've been applying them only to Python files. Maybe this is something to change, I'll note it for the next backend meeting.

backend/tests/integration_docker/test_repositories.py
```diff
@@ -46,3 +46,38 @@ async def test_repo_operational_status(
...
+            src_directory=repo_dir,
+            dst_directory=remote_repos_dir,
+        )
+        await repo.add_to_infrahub(client=client)
+        in_sync = await repo.wait_for_sync_to_complete(client=client)
+        assert in_sync
+
+        # An object (CoreCustomWebhook) that references a Python transform defined in the same
+        # repository resolves only because transforms are imported before objects. Before the fix,
+        # importing the object first aborted the whole import with a NodeNotFoundError.
```
polmichel: I would also tend to trim this comment and only mentioning which test case are we verifying there

_An object that references a Python transform defined in the same repository should be correctly resolved._

(Feel free to enhance the comment if you have a better solution ofc)

backend/tests/integration_docker/test_repositories.py
```diff
@@ -46,3 +46,38 @@ async def test_repo_operational_status(
...
+        webhook = await client.get(kind="CoreCustomWebhook", name__value="tags-webhook")
+        assert webhook.transformation.id is not None
+        transform = await client.get(kind="CoreTransformPython", id=webhook.transformation.id)
+        assert transform.name.value == "TagsTransform"
+
+        # A generator definition whose target group is defined as an object in the same repository
+        # resolves only because objects are imported before generator definitions. This guards the
+        # other side of the ordering: transforms move before objects, generators stay after them.
```
polmichel: Likewise maybe more something like:

_A generator definition whose target group is defined as an object in the same repository should also be resolved correctly_

(Feel free to enhance the comment if you have a better solution ofc)

changelog/+ifc-2669-git-import-order.fixed.md
polmichel, line 1: The file naming doesn't follow the guidelines in `dev/guidelines/changelog.md`: 

1. the name should mention only the GH issue number
2. therefore no "+"
3. consist of a single sentence

An attempt at enforcing this rule has been made to guide AI agents into using them. However, I don't think it's present yet on your branch: https://github.com/opsmill/infrahub/pull/9445/files

dev/specs/ifc-2669-fix-import-order/checklists/requirements.md
polmichel, line 1: I'm not sure we want to keep all these Spec-kit documents inside the project for every bug fix we do. Maybe it would be better to remove them.

Out of curiosity, have you used [/tiny-spec ](https://tinyspec.dev/) or similar tool to do it or have you used the heavy native Spec-kit workflow?

backend/infrahub/git/integrator.py
```diff
@@ -206,13 +206,25 @@ async def import_objects_from_files(
...
+            )  # type: ignore[call-overload]
+            await self.import_jinja2_transforms(
+                branch_name=infrahub_branch_name, commit=commit, config_file=config_file
+            )  # type: ignore[call-overload]
             await self.import_objects(
                 branch_name=infrahub_branch_name,
                 commit=commit,
                 config_file=config_file,
             )  # type: ignore[call-overload]
-            await self.import_all_python_files(branch_name=infrahub_branch_name, commit=commit, config_file=config_file)  # type: ignore[call-overload]
```
polmichel, line 212: `import_all_python_files` is then unused (besides in some tests, see `backend/tests/integration/git/test_git_repository.py`). This is a Prefect flow, so we also have to verify that there's no workflow attached to it (`backend/infrahub/workflows/catalogue.py`.), but it seems there isn't.
I assume these methods were grouped because it made sense to import all of them together. This was wrong, and your work and analysis fixed that false assumption.

But then:
- `import_all_python_files` should be removed, and the related tests should call other primitives (`import_python_transforms`, `import_jinja2_transforms`, `import_python_check_definitions`, `import_generator_definitions`, or several of them depending on the test case)
- One thing to consider is that some of these methods may need to be grouped because it only makes sense to use them together, and decoupling them could lead to future issues. Not knowing very well these objects, I am not the best to make that call.

polmichel (commented):
Thanks for your work and the fix on this issue, @lancamat1!
I've left comments that are mainly about code conventions and maintainability. We can also take over this work if you'd prefer.
One additional thing to note: we prefer to target the `stable` branch for bug fixes.
Of course, this makes sense only if this is a bug. However, the `fixed.md` part of the changelog seems to indicate that. In that case adding the "type/bug" label to the Github issue would also be required. 
We want to target `stable` branch so customers don't have to upgrade to the next minor version to get the fix and only upgrade to the next patch version which is less impacting.

lancamat1:
Thanks so much for the thorough and kind review, @polmichel — really appreciate the time you took, and every pointer was spot-on. Pushed an update addressing all of them:

- **Comments referencing the issue / spec-kit / US tags** (the fixture `.infrahub.yml` and the two test comments): trimmed to describe just the expected behavior, in line with `dev/rules/code-doc-style.md`. The two test comments now read as you suggested.
- **Changelog**: renamed to `9406.fixed.md` (issue number only, no `+`) and reduced to a single user-facing sentence, per `dev/guidelines/changelog.md`.
- **Spec-kit documents**: removed from the PR — agreed they don't belong in the repo for a bug fix.
- **`import_all_python_files`**: removed. Confirmed it has no workflow in `catalogue.py` and was only used by one integration test, which now calls the primitives directly and is renamed accordingly.
- **Target branch**: retargeted to `stable` and added the `type/bug` label to the issue — thanks for explaining the patch-vs-minor rationale, that makes total sense. The branch was develop-based, so I replayed the single commit onto `stable`; it applied cleanly since the import code is identical there.

Thanks again — happy to iterate on anything else!

dev/specs/ifc-2669-fix-import-order/checklists/requirements.md
lancamat1: Removed these from the PR. To answer your curiosity: this was the native spec-kit workflow rather than /tiny-spec — agreed it's heavier than needed for a bug fix, so it makes sense to keep it out of the repo.

backend/infrahub/git/integrator.py
lancamat1: Removed `import_all_python_files`. Checked `catalogue.py` — no workflow was attached, and the only caller was `test_import_all_python_files`, which now calls the primitives directly (and is renamed to `test_import_python_definitions`). On the grouping concern: the three methods share no state and the import flow already invokes them independently after the reorder, so decoupling them is safe here. Good call on flagging it though.

frontend/app/.gitignore
```diff
@@ -3,6 +3,9 @@
 /.pnp
 .pnp.js
 
+# this project uses pnpm; npm's lockfile must not be committed
+package-lock.json
+
```
polmichel: Are you ok with this change @pa-lem / @bilalabbad ?

ajtmccarty (commented):
@ogenstad left a comment on the bug this PR fixes that is worth considering (https://github.com/opsmill/infrahub/issues/9406#issuecomment-4592355359). it looks like we probably need a more robust importing system that allows objects to reference transforms and vice versa

lancamat1:
Thanks @ajtmccarty! The fixed-ordering approach here was agreed as part of IFC-2669. It covers both reference directions in the current schema — objects → transforms (the reported case) and generators/artifacts → groups, which are objects and so import after. A more flexible / dependency-aware importer would be a larger piece of work; happy to open a follow-up issue to track it if we'd like to pursue it.

ogenstad:
> @ogenstad left a comment on the bug this PR fixes that is worth considering ([#9406 (comment)](https://github.com/opsmill/infrahub/issues/9406#issuecomment-4592355359)). it looks like we probably need a more robust importing system that allows objects to reference transforms and vice versa

I was actually thinking about artifact definitions that refer to a group (same with generator definitions) but transforms should be fine to import earlier than we do now.

backend/infrahub/git/integrator.py
```diff
@@ -204,13 +204,25 @@ async def import_objects_from_files(
             await self.import_all_graphql_query(
                 branch_name=infrahub_branch_name, commit=commit, config_file=config_file
             )  # type: ignore[call-overload]
+            # Transforms must be registered before objects so that an object referencing a
+            # repo-defined transform (e.g. CoreCustomWebhook.transformation) resolves during import.
+            # Checks, generators and artifact definitions stay after objects because their targets
+            # reference groups created from the repository's object files.
```
polmichel: Optional nitpick comment: the second part of this comment should be as close as possible to the code it explains, so just before `import_generator_definitions` & `import_python_check_definitions` calls.

backend/tests/integration/git/test_git_repository.py
```diff
@@ -207,7 +207,9 @@ async def test_import_all_python_files(
         config_file = await repo.get_repository_config(branch_name="main", commit=commit)  # type: ignore[call-overload]
         assert config_file
 
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_generator_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
```
polmichel: Are all these instructions used? The test passes without `import_generator_definitions`, if I'm not mistaken. If that's the case, it's better to remove it from all the different steps in this file.

The reasoning behind this is that it looks like this step is required in this test, which could be confusing for people (or agents) reading the code.

backend/tests/integration/git/test_git_repository.py
```diff
@@ -217,7 +219,9 @@ async def test_import_all_python_files(
 
         # Validate if the function is idempotent, another import just after the first one shouldn't change anything
         nbr_relationships_before = await count_relationships(db=db)
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_generator_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
```
polmichel: same

backend/tests/integration/git/test_git_repository.py
```diff
@@ -259,7 +263,9 @@ async def test_import_all_python_files(
         )
         await obj2.save(db=db)
 
-        await repo.import_all_python_files(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_check_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_generator_definitions(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
```
polmichel: same

polmichel (approved):
LGTM on the backend side, I'll just ask the front-end team to validate the `.gitignore` modification before you merge. Fyi, I usually "squash" the commits (this is one of the option of the merge button) so that only one commit remains once the PR is merged. This gives a cleaner git history, I think. Thank you for the fix.

I've posted two **optional** comments but it would be great if you can:
- move the comment closer to the code it explains
- remove useless instructions from the test

backend/infrahub/git/integrator.py
lancamat1: Good call — split the comment so the second half now sits right before the checks/generators/artifacts calls, and dropped the specific class reference while at it.

backend/tests/integration/git/test_git_repository.py
lancamat1: You're right — the test only asserts on checks and transforms, so I removed import_generator_definitions from all three steps (and the StandardGroup setup that only existed to satisfy it). Confirmed the class still passes (5 passed).

frontend/app/.gitignore
bilalabbad: Hi! this is not the right fix. I think we should go the other way here, block `npm install` outright instead of hiding its output.

For now, you can remove this, I'll handle it later. (just always use pnpm, never use npm for now)

lancamat1: Removed the .gitignore entry as suggested — happy to leave the npm-blocking approach to @bilalabbad. Thanks both!

polmichel (approved):
LGTM

PR #9464, opened 2026-06-04
feat(artifacts): compute file diffs for read-only repositories (INFP-409)

ogenstad:
# Why

Artifact regeneration's per-repository file diff was gated on the source
branch's `sync_with_git` flag and on `ProposedChangeRepository.has_diff`, which
returns `False` for read-only repositories. As a result a `CoreReadOnlyRepository`
never contributed a file diff, and a proposed change whose only change was a
pinned-commit bump on a read-only repo could fail to regenerate the affected
artifacts.

## What changed

Behavioral changes:
- Per-repo file diffs are computed for both managed and read-only repositories,
  regardless of `source_branch.sync_with_git`. A repository whose source and
  destination commits match yields an empty diff and triggers no regeneration.
- Read-only repositories now diff their per-branch pinned commits; managed
  repositories diff their tracked Git branch tips.

Implementation notes:
- Added `ProposedChangeRepository.has_file_diff` (both commits present and
  differing, for any repo kind). `has_diff` is left unchanged - it is still used
  by merge-conflict validation, where the managed-only semantics are correct.
- The file-diff populator moved from `proposed_change/tasks.py::_gather_repository_repository_diffs`
  into `branch_diff.py::populate_repository_file_diffs`. Git access sits behind a
  `RepositoryFileDiffer` protocol with a production `GitRepositoryFileDiffer`
  (client injected via constructor) and an in-memory test adapter, per the
  backend-component-design and no-mock testing rules. Repository-kind dispatch
  reuses the existing `get_initialized_repo`.
- Removed the unreachable `else: list_all_files` branch (the old gate already
  required `destination_commit`); `list_all_files` itself is still used elsewhere.


backend/infrahub/proposed_change/branch_diff.py
```diff
@@ -2,22 +2,109 @@
...
+        *,
+        repository_id: str,
+        repository_name: str,
+        repository_kind: str,
+        source_commit: str,
+        destination_commit: str,
+    ) -> RepositoryFileDiff: ...
+
+
+class GitRepositoryFileDiffer:
```
ajtmccarty, line 52: should this inherit from the protocol?
UPDATE: I guess you don't have to with a Protocol... I don't like that. feel un-Pythonic. "explicit is better than implicit" unless typing with a Protocol

backend/infrahub/proposed_change/branch_diff.py
```diff
@@ -2,22 +2,109 @@
...
+            name=repository_name,
+            repository_kind=repository_kind,
+        )
+        files_changed, files_added, files_removed = await git_repo.calculate_diff_between_commits(
+            first_commit=source_commit, second_commit=destination_commit
+        )
+        return RepositoryFileDiff(files_changed=files_changed, files_added=files_added, files_removed=files_removed)
+
+
+async def populate_repository_file_diffs(
```
ajtmccarty: this could be a class that accepts the `differ` dependency in its constructor. looks like it's only used in `run_proposed_change_pipeline()`, which would be a good place to build it b/c it's the start of a workflow and this mostly an application entry point

ajtmccarty (approved):

backend/infrahub/proposed_change/branch_diff.py
ogenstad: It's very similar to how interfaces work in Go. It should be quite clear in terms of where you'd be able to use a class like this. I.e. if the target expects a protocol then anything that would satisfy that protocol would be accepted.

Issue #9465, opened 2026-06-04
feature: surface git repository branch sync errors to users


minitriga:
## Need

When a repository branch fails to sync — for example a divergent remote branch after a force-push, or an import error — the failure detail only exists in task-worker logs. In Infrahub itself the user sees:

- `operational_status = error` on the repository, with no indication of *which branch* failed or *why*
- `sync_status = in-sync` on branches that are actually stale (they were never reached, so the status was never updated)
- no error message, no failing-branch name, no timestamp of the last successful sync

An operator looking at the UI or API cannot tell that a branch is broken, which one, what the git error was, or that other branches are stale as a result. Diagnosing currently requires shell access to the worker logs. Repository sync problems are silent failures from the user's point of view, which matters because they leave data in Infrahub quietly outdated.

## Context

- The periodic sync runs as the `git_repositories_sync` flow, which is declared `WorkflowType.INTERNAL` (`backend/infrahub/workflows/catalogue.py`). Internal workflows are not tagged into the Infrahub task namespace (`WorkflowDefinition.get_tags`, `backend/infrahub/workflows/models.py`), so their runs — and the errors they raise — never appear in Infrahub's task list. This is the main reason these failures are invisible today.
- Open question to resolve as part of this issue: should the sync task runs (or just their failures) be surfaced in Infrahub's task view, or should sync errors be recorded somewhere else user-visible (e.g. on the repository/branch itself)? Showing every 1-minute cron run would be noisy; losing the errors entirely is the current problem.
- Existing signals today: repository `operational_status` / per-branch `sync_status` attributes, and failure tagging via `sync_git_repo_with_origin_and_tag_on_failure`.
- Observed while reproducing #9463 on 1.9.7 (a divergent branch wedging sync for all later branches, with no user-visible explanation).

## References

- Related issue: #9463

PR #9466, opened 2026-06-04
fix: continue git repository sync when a single branch fails (closes #9463)

### LINKED ISSUE #9463: bug: git repository sync aborts remaining branches when one branch fails
### Component

Git Integration

### Infrahub version

1.9.7 (reproduced against the released image via testcontainers)

### Current Behavior

When the git worker syncs a repository and the pull of one branch fails (e.g. the remote branch was force-pushed and now diverges from the worker's local copy), the exception propagates out of the per-branch loop in `InfrahubRepository.sync()` and the remaining branches are never processed.

Because the loop iterates branches in sorted order, every subsequent sync cycle (1-minute cron) fails on the same branch first — branches sorting after it stay stale indefinitely until the remote branch is repaired.

Additionally, the repository's `operational_status` goes to `error` but `sync_status` on the stale branch still reports `in-sync`.

### Expected Behavior

A failure on one branch should be recorded and the worker should continue syncing the remaining branches; one broken branch should not block updates to every other branch of the repository.

### Steps to Reproduce

1. Add a repository with branches `main`, `branch-a-conflict`, `branch-b-good`; wait for all branches to be imported.
2. Rewrite `branch-a-conflict` on the remote (force-push / divergent history).
3. Push a normal fast-forward commit to `branch-b-good`.
4. Wait several sync cycles.

Result: every cycle the worker logs:

```
git pull -v -- origin branch-a-conflict
fatal: Need to specify how to reconcile divergent branches.
```

followed by a `RepositoryError`, and `branch-b-good`'s commit in Infrahub never advances (observed stale for 3+ consecutive sync cycles).

### Additional Information

- Per-branch loops without exception handling: `backend/infrahub/git/repository.py` (`InfrahubRepository.sync()`, lines 85-122). The new-branches loop only guards graph branch creation; the updated-branches loop has no per-branch handling at all.
- `sync_remote_repositories()` (`backend/infrahub/git/tasks.py`) catches the error at repository level, so other repositories proceed — but the failed repository's remaining branches are skipped.
- Reproduced with an integration test using `infrahub-testcontainers` against 1.9.7; test available on request.

minitriga:
# Why

When the git worker syncs a repository and one branch fails (e.g. the remote branch was force-pushed and now diverges from the worker's local copy), the exception escapes the per-branch loop in `InfrahubRepository.sync()` and all remaining branches are skipped. Because branch iteration is sorted, every 1-minute sync cycle fails on the same branch first, so branches sorting after it stay stale indefinitely. Reproduced against the released 1.9.7 image via testcontainers.

Goal: a failure scoped to one branch must not prevent the synchronization of the other branches.

Non-goals: surfacing branch sync errors to users in the UI (#9465), auto-resolving divergent branches, fixing `sync_status` reporting for starved branches.

Closes #9463

## What changed

Behavioral changes:

- A pull or import failure on one branch no longer aborts the sync of the remaining branches; each branch is now synced inside its own error boundary.
- After all branches are processed, `sync()` raises a single consolidated `RepositoryError` listing the branches that failed, so the existing repository-level failure tagging and operational-status handling still fire.
- `RepositoryConnectionError` and `RepositoryCredentialsError` still abort immediately — if the remote itself is unreachable or unauthorized, iterating the remaining branches is pointless.

Implementation notes:

- The bodies of the new-branches and updated-branches loops moved to two private helpers (`_sync_new_branch`, `_sync_updated_branch`); the loops wrap them in try/except, accumulate failed branch names, and continue.
- The boundary catches `Exception` (not just `RepositoryError`) because object imports re-raise arbitrary exceptions after recording the per-branch `error-import` status.

What stayed the same: no schema changes, no API contract changes, branch processing order unchanged, branches synced before a failure were already updated and still are.

## How to review

- `backend/infrahub/git/repository.py` is the entire production change — review `sync()` and the two new helpers.
- Extra scrutiny: the broad `except Exception` per branch and the connection/credential re-raise — confirm no repo-wide failure mode is silently converted into a per-branch failure.
- `backend/tests/component/git/conftest.py` adds the `git_repo_07` fixture (one conflicting branch + one fast-forward branch); `test_git_repository.py` adds the regression test.

## How to test

```bash
uv run pytest backend/tests/component/git/test_git_repository.py::test_sync_continues_after_branch_pull_failure -x -v
uv run pytest backend/tests/component/git/ -k "sync"
```

The new test fails on `stable` (the fast-forward branch stays on its old commit) and passes with this fix. Full git component suite: 102 passed, 1 pre-existing unrelated failure (`test_extract_repo_file_information`, also failing on clean `stable`).

## Impact & rollout

- **Backward compatibility:** `sync()` now raises a consolidated `RepositoryError` after processing all branches instead of the first branch's error; both callers already handle `RepositoryError`.
- **Performance:** negligible — same operations, plus continued iteration after a failure.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (if user-facing or ops-facing change) — N/A
- [ ] Internal .md docs updated (internal knowledge and AI code tools knowledge) — N/A

---
## Summary by cubic
Keep syncing git repositories even if one branch fails during pull or import. After all branches are processed, raise one repository-level error listing the failed branches to prevent starvation in #9463.

- **Bug Fixes**
  - Isolate per-branch failures in git collection and import; continue syncing other branches. Abort immediately on `RepositoryConnectionError` and `RepositoryCredentialsError`.
  - Added `CollectedImports` (imports + `failed_imports`) and `FailedImport` with `ImportStep` and reason; `RepositorySyncer.sync` aggregates failures and calls `raise_if_branches_failed()` which logs step/reason per branch and raises a `RepositoryError` listing only failed branch names.
  - Added regression test `test_sync_continues_after_branch_pull_failure` with fixture `git_repo_07`; updated RPC/integration tests to use `CollectedImports` and `RepositorySyncer`.

- **Refactors**
  - Removed `InfrahubRepository.sync`; all sync paths now use `RepositorySyncer.sync`.

<sup>Written for commit 3f30e30b1cbe8eabbe188872bae30ae87d967161. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/git/repository.py
```diff
@@ -33,6 +38,19 @@ class PendingObjectImport:
     git_branch_name: str | None = None
 
 
+@dataclass
+class CollectedImports:
+    """Outcome of the git/branch-setup phase of a sync.
+
+    ``imports`` are the branches ready to have their objects imported. ``failed_branches`` are the
+    branches whose git or branch setup failed and were skipped so the others could still proceed;
+    the caller surfaces them once every branch has been handled.
```
polmichel: There should be no mention of the caller's behavior inside this class description, which should be limited to the class itself.

backend/infrahub/git/repository.py
```diff
@@ -176,10 +244,10 @@ async def collect_pending_imports(self, staging_branch: str | None = None) -> li
                         branch=branch_name,
                     )
 
-        pending_imports.extend(
+        collected.imports.extend(
```
polmichel: Usually it's better to make an object as immutable as possible. Here there are modifications to the object's state in several places, through additions to `failed_branches` and `imports`. A better approach would be to construct the lists inside this method, then construct the object once it's finalized.

backend/infrahub/git/repository.py
```diff
@@ -75,26 +93,68 @@ async def sync(self, staging_branch: str | None = None) -> None:
...
+                raise
+            except Exception as exc:
+                # The import already records its own per-branch error status before re-raising, so
+                # isolate the failure here to keep importing the remaining branches.
+                log.warning(
+                    "Failed to import branch objects, skipping it.",
+                    repository=self.name,
+                    branch=pending.infrahub_branch_name,
+                    exc_info=exc,
+                )
```
polmichel: There is a log related to this error which happens here, and another one in the `raise_if_branches_failed` method. I would suggest unifying this behavior. Keeping them separate can lead to future issues.
One way of doing that would be to collect a `FailedImport` dataclass instead of a `list[str]`. This object would also contain the reason why the import failed.
You can then create a proper global error message inside the `raise_if_branches_failed` method.

backend/infrahub/git/sync.py
```diff
@@ -86,6 +91,23 @@ def __init__(self, lock_registry: InfrahubLockRegistry, importer: RepositoryImpo
 
     async def sync(self, repo: InfrahubRepository, staging_branch: str | None = None) -> None:
```
polmichel, line 89: I do not think this method is used anymore. I should have removed it in my last PR and replaced the tests that call it to use the new component `RepositorySyncer.sync()`. I can push a commit doing that if you want.

backend/infrahub/git/repository.py
```diff
@@ -75,26 +93,68 @@ async def sync(self, staging_branch: str | None = None) -> None:
...
+                failed_branches.append(pending.infrahub_branch_name)
```
polmichel: This should not mutate the existing object either. A cleaner design would be to create a separate list of `FailedImport` here, and pass an aggregation of the two `FailedImport` lists to `raise_if_branches_failed`.

Based on this comment and the other one, you could for example attach a specific failing step to these `FailedImport` objects (modeled through a `StrEnum` with `ImportStep.COLLECTION` and `ImportStep.IMPORT`, for instance), and then modify your aggregated message depending on which step failed for the branch.

backend/infrahub/git/sync.py
minitriga: Thanks — I'll take the removal in this PR since it touches the same methods/tests as the `FailedImport` refactor I'm doing for your other comments. Deleting `InfrahubRepository.sync` and migrating the test call sites to `RepositorySyncer.sync()`. No need to push your commit.

backend/infrahub/git/repository.py
minitriga: Done in 0ddb8461d — trimmed the docstring to describe only `imports`/`failed_imports`, dropped the caller-behavior sentence.

backend/infrahub/git/repository.py
minitriga: Done in 0ddb8461d — `collect_pending_imports` now builds local `imports`/`failed_imports` lists and constructs `CollectedImports` once at the end; no more in-place mutation.

backend/infrahub/git/repository.py
minitriga: Done in 0ddb8461d — went with your `FailedImport` suggestion. The loops now collect `FailedImport(branch, step, reason)` with no in-loop logging; `raise_if_branches_failed` is the single place that logs each failure and builds the consolidated message.

backend/infrahub/git/repository.py
minitriga: Done in 0ddb8461d — added `ImportStep` (StrEnum: `COLLECTION`/`IMPORT`) and `FailedImport(branch_name, step, reason)`. The import phase in `RepositorySyncer.sync` collects its own `FailedImport(step=IMPORT)` and aggregates with the collection failures; the error message is annotated per step, e.g. `branch01 (collection: ...), branch07 (import: ...)`.

backend/infrahub/git/repository.py
```diff
@@ -70,31 +104,42 @@ async def create_branch_in_git(
...
+            log.warning(
+                "Failed to synchronize branch, skipping it.",
+                repository=self.name,
+                branch=failed.branch_name,
+                step=failed.step.value,
+                reason=failed.reason,
             )
 
-    async def collect_pending_imports(self, staging_branch: str | None = None) -> list[PendingObjectImport]:
+        details = ", ".join(f"{failed.branch_name} ({failed.step}: {failed.reason})" for failed in failed_imports)
```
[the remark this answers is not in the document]
polmichel: There are already dedicated logs with warning severity in the same method. Is it useful to raise a `RepositoryError` giving the exact same details?

In my opinion we should keep only one of both the log and the error. Or maybe keep the `RepositoryError` but have it only contain the branches in its details (as Cubic is suggesting). Personally I would lean toward the first solution.

backend/tests/component/git/test_git_repository.py
```diff
@@ -590,12 +598,32 @@ async def test_sync_updated_branch(prefect_test_fixture: None, git_repo_04: Infr
...
+    for branch_name in ["branch01", "branch02"]:
+        branch = Branch(name=branch_name, uuid=uuid4())
+        registry.branch[branch.name] = branch
+
+    remote_commit_branch02 = repo.get_commit_value(branch_name="branch02", remote=True)
+    assert repo.get_commit_value(branch_name="branch02", remote=False) != str(remote_commit_branch02)
+
+    # A branch failure may surface as an error once all branches have been processed.
+    with contextlib.suppress(RepositoryError):
+        await _sync(repo)
```
polmichel: A better solution would be to use:
`with pytest.raises(RepositoryError, match=re.escape(expected_message)):`

with `expected_message` defined as `f"Unable to synchronize the following branches of repository {repo.name}: branch01 (collection: Unable to pull the branch branch01 for repository {repo.name}, there are conflicts that must be resolved.)"`

First, because we want to assert exact error message: it gives a clear specification of what we should expect from the system.
Second, we want to assert an error is raised here, `pytest.raises` checks it.

polmichel (commented):
One comment answered in the Cubic comment thread.
One comment about the test. 

Thanks for fixing the dead-code issue aswell.

backend/infrahub/git/repository.py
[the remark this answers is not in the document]
minitriga: Went with keeping the structured per-branch warning logs as the observability surface and reducing the `RepositoryError` message to just the failed branch names — the option you and Cubic both flagged. The reason details are no longer duplicated; they live in the warning logs (queryable per branch), and the error stays a clean aggregate failure signal. Done in the latest push.

backend/tests/component/git/test_git_repository.py
minitriga: Switched to `pytest.raises(RepositoryError, match=re.escape(expected_message))` as suggested. Since the error message is now terse (branch names only), `expected_message` is `f"Unable to synchronize the following branches of repository {repo.name}: branch01"`. Pushed.

polmichel (approved):
LGTM ! Nice improvements

PR #9467, opened 2026-06-04
MVP: Scope recompute to changed schema elements

### LINKED ISSUE #9415: bug: any schema change recomputes all computed attributes, even unrelated ones
**Component:** API Server / GraphQL

**Infrahub version:** 1.9.6

## Current Behavior

When any schema change is applied on a branch, Infrahub recomputes computed attributes that have nothing to do with the change.

This is most severe for **Python-transform** computed attributes: on every `SchemaUpdatedEvent`, *every* Python computed attribute on the branch is fully recomputed (one recompute job per object of its kind), regardless of whether the changed schema node owns or feeds that attribute. On large datasets this floods background workers with redundant recompute work.

**Jinja2** computed attributes have the inverse problem: they are gated by a template hash, so they don't over-recompute — but a data-affecting migration to a field their template reads does **not** trigger a recompute, so the stored value can silently go stale.

Root cause (for maintainers): `computed_attribute_setup_python` builds its work list from *all* gathered Python triggers and fans out a full recompute for each, ignoring which schema elements changed (`backend/infrahub/computed_attribute/tasks.py`); the in-code comment notes it "can't tell right now if a given computed attribute has changed." `SchemaUpdatedEvent` (`backend/infrahub/events/schema_action.py`) carries only `branch_name` + `schema_hash`, so the setup flows never receive the diff — even though a `SchemaDiff` is already computed at the emission sites (`graphql/mutations/schema.py`, `api/schema.py`).

## Expected Behavior

A schema change should recompute a computed attribute **only** when the change affects a schema element (kind / attribute / relationship) the attribute's value actually depends on — including dependencies reached "further away" through relationships (a Python transform's query read-set; a Jinja2 template's referenced fields). The scoping should apply consistently to both Jinja2 and Python attributes. Recompute remains asynchronous / eventually consistent. When the changed-element set can't be determined (e.g. branch deletion, commit-driven setup), fall back to full recompute so nothing regresses. The recompute-vs-skip decision should be observable.

## Steps to Reproduce

1. Define two unrelated models: `Widget` (with a Python-transform computed attribute reading only `Widget` fields) and `Gadget` (no relationship to `Widget`).
2. Create several `Widget` objects so the computed attribute is populated.
3. Apply a schema change that touches only `Gadget` (e.g. add an attribute).
4. Observe that the computed attribute on **every** `Widget` is recomputed (recompute jobs queued for all `Widget` objects) even though nothing `Widget` reads changed.

Inverse case (Jinja2 staleness): define a Jinja2 computed attribute on `Widget` rendering `{{ owner__name__value }}`; apply a migration changing `owner.name`; observe the computed attribute is **not** recomputed and keeps its old value.

## Additional Information

- **Proposed direction:** scope recompute for both Jinja2 and Python computed attributes by intersecting each attribute's dependency set (a Python transform's query read-set; a Jinja2 template's referenced fields, including those reached through relationships) with the set of `(kind, field)` pairs the schema change actually touched. This requires propagating the `SchemaDiff` — already computed at the schema-update sites — into the computed-attribute setup flows (e.g. via the `SchemaUpdatedEvent` payload). When the changed-element set is unavailable (branch deletion, commit-driven setup), fall back to full recompute so nothing regresses.
- Possibly related downstream symptoms: #9067 and #9068 (tasks with thousands of related nodes overloading the browser / Prefect), #8872 (computed-attribute recalculation KeyError on branch delete).
- Out of scope: recompute triggered by transform *source-code* changes via repository sync (a different code path).

BeArchiTek:
## Summary

Fixes #9415. Previously, any schema change recomputed **every** computed attribute on the branch — and for Python-transform attributes, one recompute job per object of its kind — regardless of relevance. On large datasets this was slow and wasteful.

Recompute is now **scoped** to the schema elements a change actually touches: a computed attribute is recomputed only when the change affects an element its value depends on (including elements reached through relationships). When the impact cannot be determined precisely, it conservatively falls back to a full recompute so nothing goes stale.

### How it works
- `SchemaUpdatedEvent` now carries a `ChangedElementsPayload` (added/removed kinds + per-kind changed fields), built from the `SchemaDiff` by the two schema-update emitters and threaded through the Prefect triggers into the setup flows.
- `RecomputeScoper` (constructor-injected, single entry point) intersects each attribute's dependency set against the changed elements. Per-kind `ComputedAttributeDependencyDeriver` implementations cover Jinja2 (template dependency graph) and Python transforms (GraphQL query analysis).
- Conservative fallbacks: no change set → full recompute; unanalyzable or derived reads (display label, human-friendly id) → recompute that attribute on every change, without escalating others.

### Scope notes
- Branch-aware: a change on one branch never broadens recompute onto another.
- Merge/rebase emit branch/node events (not `SchemaUpdatedEvent`), so this scoping does not run on the merge path — unchanged from before; merged data changes recompute via the existing data-change path.

## Testing
- **Unit (38):** scoping decision matrix, both derivers, payload flattening, trigger rendering.
- **Component (8, TestContainers):** Jinja2 + Python scoped recompute end-to-end.
- **integration_docker (8, full stack):** unrelated change → zero recompute; read-field change (incl. across a relationship) → scoped recompute; branch isolation; merge characterization.
- `ruff` + `mypy` clean.

Also adds `backend/tests/integration_docker/conftest.py` disabling commit signing for the test session so repo-based integration tests run regardless of a developer's `commit.gpgsign`.

---
## Summary by cubic
Scopes computed-attribute recompute to only the schema elements that changed, instead of recomputing everything on each schema update. Fixes #9415 and reduces unnecessary jobs on large datasets while keeping correctness with conservative fallbacks.

- **Bug Fixes**
  - `SchemaUpdatedEvent` now includes `changed_elements` built by `build_changed_elements_payload`; schema load and registry update both emit it, and triggers pass it to setup flows.
  - Added `RecomputeScoper` with `Jinja2DependencyDeriver` and `PythonTransformDependencyDeriver` using `TransformReadSet`; imprecise reads (`display_label`, `hfid`) mark an attribute as “depends on everything,” so it always recomputes.
  - Fallbacks: missing change set → full recompute; branch isolation preserved; merge/rebase path unchanged.

- **Refactors**
  - Setup flows now accept `ChangedElementsPayload | None` (dropped the `dict` union).
  - Extracted SchemaDiff→payload mapping into `build_changed_elements_payload`; `ChangedElementsPayload` is data-only.
  - Moved query-report parsing out of `TransformReadSet` into tasks; `TransformReadSet` now only holds read data.

<sup>Written for commit 5f1f259ba05bbe42e2c68486ec7cbf13035034a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ogenstad:
I wasn't able to do a proper review of this this week, but can look again when I'm back if it's not progressed until then. I'm wondering if it would be good with some additional testing in general before merging it to `stable`.

Overall I think the structure looks good, but we could consider some names.

One thing that stands out is:
```python
changed_elements: ChangedElementsPayload | dict | None = None
```

We shouldn't have a bare `dict` as a parameter that could have any shape. I'm wondering if we could just t have `ChangedElementsPayload | None = None` and have Prefect convert the input correctly.

backend/infrahub/events/schema_action.py
```diff
@@ -1,13 +1,65 @@
...
+
+    added_kinds: list[str] = Field(default_factory=list, description="Object-type kinds newly added")
+    removed_kinds: list[str] = Field(default_factory=list, description="Object-type kinds removed")
+    changed_fields: dict[str, list[str]] = Field(
+        default_factory=dict,
+        description="Kind to the attribute/relationship names changed on that kind",
+    )
+
+    @classmethod
+    def from_schema_diff(cls, diff: "SchemaDiff") -> "ChangedElementsPayload":
```
ajtmccarty: I'd say that this should be its own component (or function, it's pretty simple). it shouldn't couple this model to the shape of the SchemaDiff. the model should just be responsible for holding and organizing its data. also, the type-hinting makes it look like this is struggling not to cause a circular import error

backend/infrahub/core/schema/schema_branch_computed/python_transform.py
```diff
@@ -22,6 +30,48 @@ def key_name(self) -> str:
...
+        read_fields: dict[str, frozenset[str]] = {}
+        for kind, names in read_fields_by_kind.items():
+            fields = frozenset(names)
+            if not fields or fields & IMPRECISE_READ_FIELDS:
+                return cls.imprecise()
+            read_fields[kind] = fields
+
+        return cls(read_kinds=frozenset(read_fields), read_fields=read_fields)
+
+    @classmethod
```
ajtmccarty: the first 2 classmethods can stay if you really want them, but this one should move elsewhere. this model should not be responsible for parsing a `GraphQLQueryReport`

backend/tests/component/computed_attribute/test_scoped_recompute_jinja2.py
```diff
@@ -0,0 +1,228 @@
...
+        client: InfrahubClient,
+    ) -> None:
+        """A change to a peer field read across the owner relationship recomputes the dependent attribute."""
+        await self._run(
+            default_branch,
+            admin_account,
+            ChangedElementsPayload(changed_fields={"TestComputeOwner": ["name"]}),
+        )
+
+        assert "remote_label" in self._submitted_attribute_names(workflow_recorder)
```
ajtmccarty: is it possible to assert exactly what the submitted attribute names should be? applies to some other tests too
you can maybe set this up using a parametrized case with named dataclasses like we do elsewhere too

backend/tests/component/computed_attribute/test_scoped_recompute_python.py
```diff
@@ -0,0 +1,237 @@
...
+            attributes=[AttributeSchema(name="name", kind="Text", unique=True)],
+            relationships=[
+                RelationshipSchema(name="cars", peer="TestCar", cardinality=RelationshipCardinality.MANY),
+            ],
+        ),
+    ]
+)
+
+
+class TestScopedRecomputePython(TestInfrahubAppBase):
```
ajtmccarty: could maybe share a base class or some fixtures with the other test to reduce this boilerplate duplication

backend/tests/component/computed_attribute/test_scoped_recompute_python.py
```diff
@@ -0,0 +1,237 @@
...
+        context = self._context(admin_account, default_branch)
+        changed_elements = ChangedElementsPayload(changed_fields={"TestCar": ["nbr_seats"]})
+
+        await computed_attribute_setup_python(
+            context=context,
+            branch_name=default_branch.name,
+            changed_elements=changed_elements,
+        )
+
+        assert "computed_desc_python" not in self._submitted_attribute_names(workflow_recorder)
```
ajtmccarty: same thing about asserting precise expectations and dataclasses for parametrization

backend/tests/integration_docker/test_computed_attributes.py
```diff
@@ -273,3 +277,177 @@ async def test_update_schema_not_related_to_computed_attribute(
...
+        TestingTShirt.description reads color__name__value, so a schema change to TestingColor.name
+        recomputes it, while a change to a field no template reads recomputes nothing — the template
+        itself is untouched in both cases (this is the read-field path, not a definition edit).
+        """
+        jinja2_before = await client.task.count(
+            filters=TaskFilter(workflow=[COMPUTED_ATTRIBUTE_JINJA2_UPDATE_VALUE.name])
+        )
+
+        # Unrelated: re-order a LocationSite field that no template reads -> no recompute.
+        schema_computed_tshirt["nodes"][4]["attributes"][1]["order_weight"] = 8500  # LocationSite.address
```
ajtmccarty: should this increment the current value by some amount to ensure it is a new value?

backend/tests/integration_docker/test_computed_attributes.py
```diff
@@ -273,3 +277,177 @@ async def test_update_schema_not_related_to_computed_attribute(
...
+
+        await sleep(1)
+        await wait_for_all_tasks_to_be_completed(client)
+        jinja2_after_unrelated = await client.task.count(
+            filters=TaskFilter(workflow=[COMPUTED_ATTRIBUTE_JINJA2_UPDATE_VALUE.name])
+        )
+        assert jinja2_after_unrelated == jinja2_before
+
+        # Related: re-order TestingColor.name, which TestingTShirt.description reads via the color relationship.
+        schema_computed_tshirt["nodes"][0]["attributes"][0]["order_weight"] = 8800  # TestingColor.name
```
ajtmccarty: I could definitely be missing something, but this seems mostly unnecessary b/c almost all schema changes to an attribute do not affect its value. the only one that I can think of is `default_value`, but I haven't gone through all the options. maybe we are just not capable of detecting schema changes with enough granularity and need to be conservative when determining which attributes require recalculation

backend/tests/integration_docker/test_computed_attributes.py
```diff
@@ -273,3 +277,177 @@ async def test_update_schema_not_related_to_computed_attribute(
...
+        branch_before = await client.task.count(
+            filters=TaskFilter(workflow=[COMPUTED_ATTRIBUTE_JINJA2_UPDATE_VALUE.name], branch=branch.name)
+        )
+
+        # Change the LocationSite.slug template only on the isolated branch.
+        branch_schema = deepcopy(schema_computed_tshirt)
+        branch_schema["nodes"][4]["attributes"][3]["computed_attribute"]["jinja2_template"] = (
+            "Isolated branch: {{ name__value }}"
+        )
+        loaded = await client.schema.load(schemas=[branch_schema], branch=branch.name, wait_until_converged=True)
```
ajtmccarty: schema loading and waiting could maybe be extracted to its own method/function for these tests

backend/tests/integration_docker/test_computed_attributes.py
```diff
@@ -273,3 +277,177 @@ async def test_update_schema_not_related_to_computed_attribute(
...
+        self, client: InfrahubClient, schema_computed_tshirt: dict
+    ) -> None:
+        """Merging a branch's schema change does not run this feature's schema-scoped recompute.
+
+        Merge and rebase emit branch and node events, not a schema-update event, so the
+        computed-attribute setup flow is not triggered on merge. A schema-only change applied by a
+        merge (no object-data change) therefore does not recompute on the target branch; merged
+        data changes are recomputed by the separate data-change path. This characterizes the
+        boundary and confirms a merge never broadens recompute onto the default branch.
+        """
```
ajtmccarty, line 431: very thorough testing

ajtmccarty (approved):

backend/tests/integration_docker/test_computed_attributes.py
gmazoyer: Looks like FR-004 in the spec covers exactly this: any schema edit to a read element counts as a change for recompute, including cosmetic ones like ordering. The test exercises that conservative path on purpose, matching your second hypothesis. A finer-grained per-property impact classification could be a follow-up if you think it is worth scoping separately.

ajtmccarty (approved):

PR #9473, opened 2026-06-05
feat: add error catalogue docs generation and freshness tooling [INFP-468]

ogenstad:
# Why

The enriched GraphQL error catalogue (string extensions.code + http_status + typed data) landed in earlier INFP-468 work, but there was no
human-readable reference for the codes, no generator to keep such a page in sync with the catalogue, and nothing guarding that the committed
schema/error-catalogue.json stays in sync with the backend Python catalogue. The export also stamped a wall-clock generated_at, producing diff churn on
every regeneration.

Goal: generate a reference docs page from the catalogue, add invoke tooling to regenerate/validate the catalogue artifacts, and make the committed JSON
deterministic and guarded.

### Behavioral changes
- New generated reference page docs/docs/reference/error-catalogue/index.md — per-code description, stability, HTTP status, data-shape table,
worked-example envelope, plus "For SDK and third-party consumers", "Third-party consumers", and "Known gaps" sections. Linked under Reference in the
Docusaurus sidebar.
- The catalogue export no longer emits generated_at; the committed schema/error-catalogue.json drops that field (one line).

### Implementation notes
- docs.generate-error-catalogue invoke task renders the page from schema/error-catalogue.json via a Jinja2 template
(docs/_templates/error-catalogue.j2), wired into the _generate() chain so the existing docs.validate CI job guards its freshness — consistent with every
other generated reference page.
- New tasks/frontend.py: regenerate-error-bindings (export JSON → pnpm generate:error-bindings → docs page) and check-error-bindings (regenerate + git
diff --exit-code, printing the regen command on drift); frontend collection registered in tasks/__init__.py.
- test_export_matches_committed_file asserts export_catalogue() equals the committed JSON exactly (deterministic now that generated_at is gone), gating
JSON<-->backend drift inside backend-tests-unit; added a get_repository_dir() test helper.

---
## Summary by cubic
Generates a deterministic, machine-readable GraphQL Error Catalogue and an auto-built reference docs page, with tooling to keep backend JSON, frontend bindings, and docs in sync. Supports INFP-468 by enabling SDK/frontend to rely on stable error codes and fields.

- **New Features**
  - Added `docs.generate-error-catalogue` to render the docs page from `schema/error-catalogue.json` via a Jinja template; new page at `docs/docs/reference/error-catalogue/index.md` and sidebar entry.
  - Added `frontend.regenerate-error-bindings` and `frontend.check-error-bindings` tasks to regenerate `schema/error-catalogue.json`, run `pnpm generate:error-bindings`, and rebuild docs; fails on drift via `git diff`.

- **Refactors**
  - Removed `generated_at` from the exported catalogue and schema for deterministic output.
  - Added unit test to assert `export_catalogue()` matches the committed `schema/error-catalogue.json`; updated helpers to locate the repo root.
  - Updated specs to document consumer guidance and CI responsibilities without adding new workflow steps.

<sup>Written for commit b68cbd5b3f59d3f239e2f8975a5ba1d7a2584dc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9473?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/errors/export.py
```diff
@@ -32,7 +31,6 @@ def export_catalogue() -> dict[str, Any]:
     return {
         "$schema": JSON_SCHEMA_VERSION,
         "infrahub_catalogue_version": CATALOGUE_VERSION,
-        "generated_at": datetime.now(tz=UTC).strftime("%Y-%m-%dT%H:%M:%SZ"),
```
ogenstad, line 35: The timestamp here just caused problems when comparing different files and it doesn't really serve a purpose.

docs/_templates/error-catalogue.j2
```diff
@@ -0,0 +1,58 @@
+---
+title: Error Catalogue
+---
+
+{# Generated from schema/error-catalogue.json — do not edit by hand. #}
+{# Regenerate with: uv run invoke docs.generate-error-catalogue #}
```
ajtmccarty: looks like this doesn't make it to the rendered catalogue

tasks/docs.py
```diff
@@ -859,3 +866,103 @@ def group_events_by_category(event_classes: list[type]) -> dict[str, list[dict[s
...
+
+def _generate_error_catalogue_documentation() -> None:
+    """Render the GraphQL error catalogue reference page from the committed JSON Schema artefact.
+
+    The output is fully data-driven and deterministic (codes sorted, no timestamps) so it can be
+    validated with `git diff --exit-code` the same way the other generated reference pages are.
+    """
+    import json
+
+    import jinja2
```
ajtmccarty: should be at the top?

ajtmccarty (approved):

docs/_templates/error-catalogue.j2
ogenstad: I removed them from the template instead as we don't want it to show up on the public website regardless.

tasks/docs.py
ogenstad: I've moved the json one up top but kept jinja2 as it's not part of the standard library and we still have that situation with `invoke demo.start` that where we only want to import toml. But hopefully we can get rid of those in the not to distant future..

PR #9482, opened 2026-06-05
fix(graphql): reject `at` query times before branch creation

### LINKED ISSUE #4076: bug: Rewinding time to before an instance is spun up gives unactionable error message
### Component

Not Sure

### Infrahub version

0.16.0

### Current Behavior

Spin up an instance, load data, and login. Then rewind time to a time before the instance was spun up. You get an error message, but it's misleading as to what the actual cause is:
image

### Expected Behavior

A relevant error message about the time frame being out of bounds, or not allowing the user to choose a time that's before the instance was spun up.

### Steps to Reproduce

Listed above

### Additional Information

_No response_

gmazoyer:
## Why

When a user picked a time-travel timestamp earlier than the Infrahub instance was created, the backend silently loaded an empty schema at that time and surfaced a generic, unactionable error (typically a deep schema-lookup failure from a resolver).

This PR makes the backend reject out-of-range `at` values at the request boundary with a clear, user-facing 422 that names the offending timestamp and the branch's effective lifetime.

Closes #4076

## What changed

- Requests with `?at=<time earlier than the branch's effective creation>` now return HTTP 422 with a message like `Requested time '2000-01-01T00:00:00.000000Z' is before branch 'main' was created at '2026-06-05T...'.`.
- On a user branch, the floor is the *origin* branch's `created_at`, not the user branch's own `branched_from`, so legitimate historical-on-origin queries between `origin.created_at` and `self.branched_from` continue to work.
- Mutations are unaffected: the existing override forces `at = now` before the validator can run.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Rejects GraphQL and REST requests where `at` is earlier than the branch’s effective creation time. Returns HTTP 422 with a clear message, fixing #4076 and preventing empty-schema resolver errors.

- **Bug Fixes**
  - Validate `at` at REST and GraphQL request boundaries; on failure, raise `ValidationError` and let the global handler return 422 with the requested time and the boundary branch’s creation time.
  - For user branches, use the origin branch’s `created_at` as the floor; mutations still force `at=now`. Added REST and GraphQL tests and a changelog entry.

- **Refactors**
  - Extracted `BranchQueryTimeValidator` (registry-injected) to encapsulate branch lifetime checks and replace previous branch-level validation.
  - Moved the `PLR0915` ignore for `_handle_http_request` to `pyproject.toml`.

<sup>Written for commit bdd4488de74e6d33f9b62c6db2b9d57f939db6d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/api/dependencies.py
+++ b/backend/infrahub/api/dependencies.py
@@ -8,6 +8,7 @@ from pydantic import BaseModel, ConfigDict
 
 from infrahub import config
 from infrahub.auth import AccountSession, authentication_token, validate_jwt_access_token, validate_jwt_refresh_token
+from infrahub.branch.query_time_validator import BranchQueryTimeValidator
 from infrahub.context import InfrahubContext
 from infrahub.core.branch import Branch  # noqa: TC001
 from infrahub.core.registry import registry
@@ -83,7 +84,7 @@ async def get_branch_params(
 
     at_ts = Timestamp(at)
     if at is not None:
-        branch.validate_query_time(at_ts)
+        BranchQueryTimeValidator(registry=registry).validate(branch=branch, at=at_ts)
 
     return BranchParams(branch=branch, at=at_ts)
 
--- a/backend/infrahub/api/dependencies.py
+++ b/backend/infrahub/api/dependencies.py
@@ -81,7 +81,11 @@ async def get_branch_params(
     branch = await registry.get_branch(db=db, branch=branch_name)
     request.state.branch_name = branch.name
 
-    return BranchParams(branch=branch, at=Timestamp(at))
+    at_ts = Timestamp(at)
+    if at is not None:
+        branch.validate_query_time(at_ts)
+
+    return BranchParams(branch=branch, at=at_ts)
 
 
 async def get_branch_dep(
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -96,6 +96,27 @@ class Branch(StandardNode):
             raise RuntimeError(f"created_at not set for branch {self.name}")
         return self.created_at
 
+    def validate_query_time(self, at: Timestamp) -> None:
+        """Validate that `at` falls within this branch's effective lifetime.
+
+        Raises:
+            ValidationError: When `at` is earlier than the branch's effective creation time.
+
+        """
+        if self.is_default or self.is_global or self.origin_branch == self.name:
+            boundary_branch_name = self.name
+            boundary_created_at = self.get_created_at()
+        else:
+            origin = registry.get_branch_from_registry(branch=self.origin_branch)
+            boundary_branch_name = origin.name
+            boundary_created_at = origin.get_created_at()
+
+        if at < Timestamp(boundary_created_at):
+            raise ValidationError(
+                f"Requested time '{at.to_string()}' is before "
+                f"branch '{boundary_branch_name}' was created at '{boundary_created_at}'."
+            )
+
     @property
     def is_terminal(self) -> bool:
         return self.status in TERMINAL_BRANCH_STATUSES
--- a/backend/infrahub/core/branch/models.py
+++ b/backend/infrahub/core/branch/models.py
@@ -96,27 +96,6 @@ class Branch(StandardNode):
             raise RuntimeError(f"created_at not set for branch {self.name}")
         return self.created_at
 
-    def validate_query_time(self, at: Timestamp) -> None:
-        """Validate that `at` falls within this branch's effective lifetime.
-
-        Raises:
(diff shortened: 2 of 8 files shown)


backend/infrahub/api/dependencies.py
```diff
@@ -81,7 +81,11 @@ async def get_branch_params(
     branch = await registry.get_branch(db=db, branch=branch_name)
     request.state.branch_name = branch.name
 
-    return BranchParams(branch=branch, at=Timestamp(at))
+    at_ts = Timestamp(at)
+    if at is not None:
+        branch.validate_query_time(at_ts)
```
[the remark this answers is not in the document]
gmazoyer: The `test_query_endpoint_at_before_branch_creation` test covers this path and shows that a 422 is surfaced.

backend/infrahub/graphql/app.py
```diff
@@ -216,18 +216,24 @@ async def _handle_http_request(
...
-                operation_name=operation_name,
-                branch=branch,
-                document=cached_parse(query),
-            )
+        elif at:
+            at_ts = Timestamp(at)
+            try:
+                branch.validate_query_time(at_ts)
+            except ValidationError as exc:
+                return JSONResponse(exc.api_response(), status_code=exc.HTTP_CODE)
```
ogenstad: I'm wondering if we need to raise a JSONResponse error here or if it would get managed automatically regardless?

gmazoyer: I think you're right, it was redundant. The global `Error` handler already converts `ValidationError` to 422 with a JSON shape, so this `try/except` is redundant. I dropped it, and tests still pass, confirming your assumption.

backend/infrahub/core/branch/models.py
```diff
@@ -96,6 +96,27 @@ def get_created_at(self) -> str:
...
+
+        Raises:
+            ValidationError: When `at` is earlier than the branch's effective creation time.
+
+        """
+        if self.is_default or self.is_global or self.origin_branch == self.name:
+            boundary_branch_name = self.name
+            boundary_created_at = self.get_created_at()
+        else:
+            origin = registry.get_branch_from_registry(branch=self.origin_branch)
```
ajtmccarty: I think this method would be acceptable on the `Branch` model if it didn't need to use the registry. so if the comparison was just `self.get_created_at() < at` then that would be all right. but if it needs to retrieve a branch from the registry, then it should be a separate component that gets its dependencies passed in during construction

gmazoyer: Good idea. I extracted the logic into `BranchQueryTimeValidator`, with registry injected through the constructor and branch/at passed to the `validate` method.

backend/infrahub/graphql/app.py
```diff
@@ -173,7 +174,7 @@ async def _get_on_get(self, request: Request) -> Response | None:
 
         return response
 
-    async def _handle_http_request(
+    async def _handle_http_request(  # noqa: PLR0915
```
ajtmccarty: can you add this exclusion to pyproject.toml so there is a chance that we will come back and fix it please?

backend/tests/component/api/test_20_graphql.py
```diff
@@ -70,6 +77,72 @@ async def test_graphql_endpoint_with_timestamp(
...
+    client: TestClient,
+    admin_headers: dict[str, str],
+    default_branch: Branch,
+    create_test_admin: Node,
+    car_person_data: dict[str, Node],
+) -> None:
+    """Querying with an `at` earlier than the branch's effective lifetime must produce a clear, user-facing error.
+
+    The error always references the floor branch — the default branch itself for default-branch queries, and the
+    origin branch (`main`) for user-branch queries — never the user branch's own `created_at` or `branched_from`.
```
ajtmccarty: it's really just always the default branch with the possible exception of the global branch, right?

ajtmccarty (approved):
it does seem weird that we allow querying before a user branch was created, but I guess we've been supporting that, so we have to explicitly deprecate it if we want to remove it

gmazoyer:
> it does seem weird that we allow querying before a user branch was created, but I guess we've been supporting that, so we have to explicitly deprecate it if we want to remove it

Good point, I logged it as a follow-up task.

PR #9487, opened 2026-06-07
fix: scope artifact/generator regeneration for complex uniqueness constraints

### LINKED ISSUE #9486: Artifact/generator regeneration is not scoped to changed targets for composite, multiple, relationship-based, or HFID uniqueness constraints
### Summary

When a proposed change runs, Infrahub decides which artifacts (or generator instances) need to be re-rendered. If the artifact/generator definition's GraphQL query targets a single object, only the artifacts for the actually-changed objects are regenerated. Otherwise Infrahub falls back to regenerating **every** artifact in the definition's target group and logs:

> `Artifact definition <name> query does not guarantee unique targets. All targets will be processed.`

This fallback is triggered even when the query *does* unambiguously target a single object, as long as it does so via anything other than the simplest possible uniqueness key. Specifically, the query is treated as non-unique whenever it filters by:

- a **composite** uniqueness constraint (more than one field in the constraint, e.g. `namespace__value` + `name__value`),
- a node that has **more than one** uniqueness constraint defined (even when the query fully satisfies one of them by a single field),
- a uniqueness constraint that includes a **relationship** component (e.g. uniqueness on `name__value` + `site`), or
- the node's **human-friendly ID** (`hfid:` filter).

Only two cases are currently recognized as targeting a unique object: a filter by `ids:` with a single required ID, or a filter by a node whose schema has exactly one uniqueness constraint consisting of exactly one attribute, where the filter argument matches that single attribute.

### Conditions to reproduce

1. Define a schema node with a composite or multiple uniqueness constraint (e.g. `uniqueness_constraints: [["namespace__value", "name__value"]]`, or `[["name__value"], ["identifier__value"]]`), or one that includes a relationship.
2. Create an artifact definition (or generator definition) whose GraphQL query filters that node by the uniqueness key (or by `hfid`).
3. Open a proposed change that modifies a single object of that kind.
4. Observe that the artifact validator logs the "does not guarantee unique targets" warning and regenerates artifacts for **all** members of the target group, not just the changed object.

### Impact on users

- **Slow proposed changes on large groups.** A one-object change forces regeneration and re-validation of every artifact in the definition's group. For definitions targeting hundreds or thousands of objects, a trivial edit triggers a full re-render.
- **Worsens silently when a schema gains a second uniqueness constraint.** A definition whose query was previously scoped correctly (single single-field constraint) starts regenerating all targets the moment a second uniqueness constraint is added to the node's schema — with no change to the query or the definition itself.
- **No way to opt into scoped regeneration for common modeling patterns.** Composite natural keys, relationship-scoped uniqueness (e.g. "name unique within a site"), and HFID-based filtering are standard, recommended schema patterns, yet none of them benefit from targeted regeneration.
- **Increased load and longer feedback loops** on every affected proposed change, proportional to group size rather than to the size of the actual change.

### Additional information

The uniqueness determination is the `only_has_unique_targets` check used when validating artifact generation during a proposed change. It currently recognizes a query as single-target only when:

- the query filters by a required, single-valued `ids:` argument, or
- the targeted node's schema has exactly one uniqueness constraint, that constraint has exactly one component, and a required filter argument's name equals that component.

It does not account for uniqueness constraints made of multiple components, nodes carrying more than one uniqueness constraint, constraints that reference relationships, or filtering by `hfid`. The fallback path (regenerate all targets) is always functionally correct — the issue is scope/performance, not incorrect output.

FragmentedPacket:
## Summary

During a proposed change, the artifact/generator validator decides whether it can limit regeneration to the objects that actually changed, or whether it must regenerate every target in the definition's group. That decision is driven by `only_has_unique_targets` on the GraphQL query analyzer.

Previously a query was recognized as targeting a single object only when it filtered by a required `ids` argument, or by a node whose schema had **exactly one** uniqueness constraint made of **exactly one** attribute (with the filter argument name matching that attribute). Any other shape fell back to regenerating all targets and logged:

> `Artifact definition <name> query does not guarantee unique targets. All targets will be processed.`

This penalized several standard, recommended schema patterns:

- **composite** uniqueness constraints (more than one component, e.g. `namespace__value` + `name__value`),
- nodes carrying **more than one** uniqueness constraint (even when the query fully satisfies one of them),
- constraints that include a **relationship** (e.g. uniqueness on `name__value` + `site`),
- filtering by the node's **human-friendly id** (`hfid:`).

It also degraded silently: a definition whose query was scoped correctly would start regenerating all targets the moment a second uniqueness constraint was added to the node's schema, with no change to the query or definition.

## Change

`only_has_unique_targets` now applies the same model the uniqueness checker uses. A root query resolves to a single object when:

- it filters by a required, single-valued `ids` or `hfid` argument, **or**
- every component of **at least one** of the node's uniqueness constraints is pinned by a required, single-valued filter argument.

Component pinning:

- **attribute** component `attr__value` → a required, single-valued `attr__value` argument,
- **relationship** component → a required, single-valued `<rel>__ids` or `<rel>__hfid` argument, and only when the relationship is single-cardinality.

The fallback path was always functionally correct; this is a scope/performance improvement — more changes now regenerate only the affected artifacts instead of the whole group.

## Notes

- No schema or API changes. The uniqueness-constraint metadata is unchanged; only the analyzer's interpretation of it is broadened.
- Existing behavior is preserved (including a required `ids: $ids` list variable counting as unique, as the artifact framework drives it per target).
- The `targets_unique_nodes` field description on `InfrahubGraphQLQueryReport` was updated to match.

## Tests

- Extended `test_query_report_single_target` coverage with a new `test_query_report_single_target_complex_constraints` exercising composite, multiple, relationship-based, and hfid constraints (both pinned and unpinned cases).
- Verified the downstream proposed-change suite (`test_validate_artifacts_generation.py`) and the `InfrahubGraphQLQueryReport` resolver tests still pass.

Closes #9486

---
## Summary by cubic
Improve artifact/generator regeneration by correctly detecting single-target GraphQL queries using composite, multiple, overlapping, relationship-based uniqueness constraints, or `hfid`. Also fixes a misclassification when required list variables were used to pin uniqueness components, avoiding unnecessary full-group rebuilds.

- **Bug Fixes**
  - Recognize a query as unique when it filters by a required single `ids` or `hfid`, or pins every component of at least one uniqueness constraint (including composites, overlapping/multiple groups, and single-cardinality relationships via `<rel>__ids`/`<rel>__hfid`).
  - Do not treat a required list-typed variable as a single-valued pin for uniqueness components; only single-element list literals, static values, or required scalar variables count. Still accept a required list variable for top-level `ids`/`hfid`.
  - Updated the `GraphQLQueryReport.targets_unique_nodes` description and regenerated the GraphQL schema. Added tests for composite, overlapping and multi-constraint, relationship, `hfid`, and list-variable cases.

<sup>Written for commit bd452feb1590c5c7c6564492aa76b1cd236ad13d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9487?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


FragmentedPacket:
Addressed in b71208ba6 — thanks, this was a valid catch identified by cubic.

The flagged line treated any required variable (including a list-typed one) as single-valued. For the **top-level `ids`/`hfid`** target selector that is intentional and preserved — the artifact framework drives that filter per target member, so a required list resolves to one object (this is the existing `QUERY_UNIQUE_TARGETS` contract).

The real risk was in the **new uniqueness-constraint pinning path**: a relationship component filtered by a required list variable (e.g. `color__ids: $colors` with `$colors: [ID!]!`) could carry several ids and match several objects, yet was being counted as a single target. That direction causes *under*-regeneration, so it's the dangerous one.

Fix: constraint-component pinning now rejects required list variables (only a single-element list literal, a static value, or a required scalar variable pins a component); the top-level selector keeps the per-member leniency via an explicit `allow_required_list` flag. Added regression coverage:
- `color__ids: $colors` with `$colors: [ID!]!` → not unique
- `TestingTShirtComposite(ids: $ids)` with `$ids: [ID!]!` → unique (documents the intentional top-level divergence)

ogenstad (approved):

PR #9507, opened 2026-06-08
docs: improve advanced sso page

wvandeun:
## Summary by cubic
Clarifies and restructures Advanced SSO docs for group mapping: map to existing groups, auto-create from claims (regex filters incl. first‑match multi‑pattern via JSON‑array env var; per‑login cap emits one capped event; audit events; provenance; invalid regex halts startup with index), a default‑group fallback (SSO logins only), and additive membership (no auto‑removal).

Adds IdP group‑claim setup, a log example, and a membership‑resolution overview (with/without auto‑creation, noting unmatched claims don’t fall through when a filter is enabled), plus `toc_max_heading_level: 4`.

<sup>Written for commit 77505677a13c4a89cf5d768ad1bbfc0ca62ca870. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/docs/deploy-manage/user-management/sso/advanced-sso.mdx
+++ b/docs/docs/deploy-manage/user-management/sso/advanced-sso.mdx
@@ -191,7 +191,7 @@ See the [group events reference](../../../reference/infrahub-events/group.mdx) f
 
 ### Set a default group
 
-When your identity provider sends no group information, assign a default group to SSO users.
+When your identity provider sends no group information, assign a default group to SSO users. The default group applies to SSO logins only; accounts that sign in with a local username and password are unaffected and keep the groups they already belong to.
 
 :::warning
 You must create this default group in Infrahub before configuring it here.
--- a/docs/docs/deploy-manage/user-management/sso/advanced-sso.mdx
+++ b/docs/docs/deploy-manage/user-management/sso/advanced-sso.mdx
@@ -1,5 +1,6 @@
 ---
 title: Advanced SSO configuration
+toc_max_heading_level: 4
 ---
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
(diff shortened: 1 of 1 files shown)


yjouffrault:
Reviewed against the source — the auto-create logic in `auth/auth_groups/` and `auth/auth.py` matches the docs precisely, including the membership-resolution order and the "filter disables existing-name fallthrough" behavior. Two small notes; I've pushed the first, the second is a question for you.

**Pushed** (`by hand` → `manually`, plus an additive-membership note in "How Infrahub resolves membership"): group membership is additive — `_assign_group_memberships` only ever calls `members_rel.add(...)`, so a user is never removed from a group when a claim stops being sent. Comparable SSO docs (GitLab, Grafana) call this out; added a short `:::info`.

**Question — worth a troubleshooting line?** In `service.py`, a matching claim whose effective name collides with an existing **non-`CoreAccountGroup`** node (`_reuse_or_skip`) is silently skipped — the claim matched, but no membership is granted and no `Rejected` event fires (that path is only for empty/whitespace names). An operator could see "claim matched, no group" with nothing obvious in the events. Might be worth a one-liner in a troubleshooting note, or it may be too rare to bother. Your call.

wvandeun:
LGTM

> **Question — worth a troubleshooting line?** In `service.py`, a matching claim whose effective name collides with an existing **non-`CoreAccountGroup`** node (`_reuse_or_skip`) is silently skipped — the claim matched, but no membership is granted and no `Rejected` event fires (that path is only for empty/whitespace names). An operator could see "claim matched, no group" with nothing obvious in the events. Might be worth a one-liner in a troubleshooting note, or it may be too rare to bother. Your call.

I don't think this a scenario we need to call out. Group mapping should only work for CoreAccountGroup, so the current behavior is correct.

yjouffrault (approved):

PR #9513, opened 2026-06-09
test(git): cover closure-builder failure isolation at import

ogenstad:
# Why

The closure-builder failure-isolation path, where a malformed transform must not block import of its siblings, was only covered at the unit level (the dispatcher in test_dispatcher.py). There was no test proving the contract end to end through the real integrator import seam against persisted CoreTransformation nodes. This is INFP-409 task T057.

## What changed

* New integration test backend/tests/integration/git/test_closure_failure_isolation.py (mirrors quickstart Scenario 11) driving import_jinja2_transforms on TestInfrahubApp plus FileRepo and asserting, against the imported CoreTransformJinja2 nodes:
  * both transforms are persisted, so the malformed one is flagged rather than dropped;
  * the malformed transform ends up with dependencies_complete = False;
  * the well-formed sibling imports with a complete closure (.infrahub.yml, templates/report.j2, templates/partial.j2);
  * the closure-builder failure is logged naming the offending transform only.
* New fixture repo backend/tests/fixtures/repos/closure-failure-isolation/ (initial__main: one well-formed transform with a transitive include plus one with a syntax-error template).

**Implementation notes**:

The malformed {% include "missing is a TemplateSyntaxError caught inside Jinja2Closure, which yields complete=False rather than escaping to the dispatcher's ISOLATED_FAILURES catch. The observable contract is the same either way, and the dispatcher-exception path stays unit-covered by T053. The test lives in integration/git/ alongside the existing import tests rather than the integration_docker/ path the task originally named, because it is import-only with no second commit and no proposed-change diff.

---
## Summary by cubic
Adds an integration test to prove Jinja2 closure-builder failures are isolated during Git import. A malformed template flags only its own transform and does not block a well‑formed sibling; the import log names only the offending transform.

- **New Features**
  - Fixture repo `closure-failure-isolation` with `well_formed_report` (transitive include) and `broken_report` (syntax error).
  - Test `backend/tests/integration/git/test_closure_failure_isolation.py` verifies both `CoreTransformJinja2` nodes import; `well_formed_report` has a complete closure (`.infrahub.yml`, `templates/partial.j2`, `templates/report.j2`), `broken_report` sets `dependencies_complete=False`, and logs include the unresolved-reference message for `broken_report` only. Updates `dev/specs/infp-409-artifact-regen-triggers/tasks.md` to mark T057 complete.

<sup>Written for commit c093094776c72e1d682c4727805de50c438d60f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9513?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/integration/git/test_closure_failure_isolation.py
```diff
@@ -0,0 +1,108 @@
...
+            ".infrahub.yml",
+            "templates/partial.j2",
+            "templates/report.j2",
+        }
+
+        broken = transforms["broken_report"]
+        assert broken.dependencies_complete.value is False
+
+        # The closure-builder failure is reported against the offending transform only.
+        assert "Closure builder for transform 'broken_report'" in caplog.text
```
ajtmccarty: should this assert that it says something about being broken too?

ajtmccarty (approved):

PR #9523, opened 2026-06-10
feat(database): support comma-separated cluster members in INFRAHUB_DB_ADDRESS

fatih-acar:
### Summary

- `INFRAHUB_DB_ADDRESS` now accepts a comma-separated list of cluster members in `host[:port]` format (e.g. `neo4j-core-1,neo4j-core-2:7777,neo4j-core-3`); members without an explicit port fall back to `INFRAHUB_DB_PORT`
- A [custom driver address resolver](https://neo4j.com/docs/python-manual/current/connect/#cluster) expands the initial connection address into all configured members, so the initial connection fails over across members when some are unreachable
- The resolver only expands the initial address and passes any other address through unchanged — the driver invokes the resolver for every connection it opens, including ones targeting specific servers from a routing table, so `neo4j://` routing is not corrupted
- A warning is logged when multiple members are configured with the `bolt` protocol, which pins all traffic to the first reachable member without routing awareness; `neo4j://` is recommended for clusters
- Single-address behavior is unchanged (no resolver is installed)
- Docs: new "Connecting Infrahub to a Neo4j cluster" section in the HA guide presenting DNS routing (headless service) and the member list as the two bootstrap options

### Verification

- Unit tests for address parsing (`DatabaseSettings.address_members` / `database_uri`, including per-member ports and bracketed IPv6) and for the resolver (expansion + pass-through)
- Verified end-to-end against real TCP listeners with driver debug logging: the driver connects to member 1 at its explicit port and fails over to member 2 at a different port after a handshake failure

---
## Summary by cubic
Adds cluster-aware DB bootstrap: `INFRAHUB_DB_ADDRESS` now accepts a comma-separated list of `host[:port]` members, and a resolver provides failover across them. Single-address setups are unchanged.

- **New Features**
  - `INFRAHUB_DB_ADDRESS` accepts multiple members; missing ports default to `INFRAHUB_DB_PORT` and bracketed IPv6 is supported.
  - Only the first member is used in the URI; a driver resolver expands it into all members and passes through other addresses, preserving `neo4j://` routing.
  - Logs a warning when multiple members are used with `bolt`; `neo4j` is recommended for clusters.
  - Docs updated: HA guide adds cluster connection options (DNS routing and member list); config reference documents the `INFRAHUB_DB_ADDRESS` format. Tests cover parsing (incl. IPv6), resolver behavior, and URI generation.

- **Refactors**
  - Moved database unit tests to `backend/tests/unit/database` (resolver tests and retry transaction tests).

<sup>Written for commit 18ae6825b353e70d8506aa3e9a7bc8b914a22691. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/config.py
```diff
@@ -336,14 +340,40 @@ class DatabaseSettings(BaseSettings):
...
             return f"{base_uri}?policy={self.policy}"
         return base_uri
 
+    @staticmethod
+    def _split_member(member: str) -> tuple[str, int | None]:
+        """Split a 'host[:port]' member into host and optional port, supporting bracketed IPv6 hosts."""
+        if member.startswith("["):
+            host, _, rest = member.partition("]")
+            host += "]"
+            if rest.startswith(":") and rest[1:].isdigit():
```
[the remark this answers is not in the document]
fatih-acar: I will ignore this for now

backend/infrahub/database/__init__.py
```diff
@@ -507,18 +508,50 @@ async def validate_database(
...
+    every connection it opens, including ones targeting specific servers discovered through
+    a routing table, so any address other than the initial one is passed through unchanged.
+    """
+    addresses = [Address.parse(member, default_port=default_port) for member in members]
+    initial_address = addresses[0]
+
+    def resolver(address: Address) -> list[Address]:
+        if (address.host, address.port) == (initial_address.host, initial_address.port):
+            return addresses
+        return [address]
```
ajtmccarty, line 525: I'm assuming you've tested this, but this format does not look like the example from the neo4j docs
```
def custom_resolver(socket_address):
    yield "server01.example.com", 7687
    yield "server02.example.com", 7687
    yield "server03.example.com", 7687
    yield "backup.example.org", 7687
```
it looks like each yield/return is a tuple[str,int]

I also just find this kind of confusing. maybe this is how it has to be, but it is hard to understand

ajtmccarty (approved):

backend/infrahub/database/__init__.py
fatih-acar: Good catch on the format question — I dug into the driver to be sure.

The driver normalizes whatever the resolver returns via `map(Address, resolver(address))` (`neo4j/_async_compat/network/_util.py`), and `Address` is a `tuple` subclass. So returning `list[Address]` and the docs' `yield (host, port)` tuples are equivalent — each element gets passed through `Address(...)` either way. I went with `Address.parse(member, default_port=...)` rather than raw tuples because it also handles the `host[:port]` / bracketed-IPv6 parsing that a bare tuple wouldn't. I'll expand the docstring to make that explicit since it reads as non-obvious.

On "have you tested this" — yes, now against a real 3-node Neo4j Enterprise cluster via `infrahub-testcontainers` cluster mode, with `INFRAHUB_DB_ADDRESS=database,database-core2,database-core3`:

- Full stack (server + workers) boots healthy and routes reads/writes through the cluster via the resolver.
- Bootstrap failover: with a dead-but-still-resolvable first seed, the server fails over to a surviving member and boots; a single dead seed (no resolver built) fails as expected.

One caveat that testing surfaced and is worth documenting: the driver runs DNS on each resolver-returned address and raises on the first that fails to resolve (`_util.py`, `_dns_resolver`) rather than skipping it. So the failover guarantee covers members that are **down but still DNS-resolvable** (the VM/IP case and the k8s-headless case this feature targets). If a listed member's name stops resolving entirely (NXDOMAIN), the driver aborts before trying the rest — so it's worth listing an always-resolvable member first. I'll add a note to the docs.

PR #9527, opened 2026-06-10
chore(ci): introduce pytest-playwright based E2E tests

fatih-acar:
## Summary by cubic
Introduces a fully async Python E2E suite using `pytest-playwright-asyncio` + `infrahub-testcontainers`, sharded across 4 CI jobs for faster runs. Adds a runtime API to change API response delay via message bus without restarts; GraphQL reads the delay per request.

- **New Features**
  - `tests/e2e/`: async Playwright suite with SDK‑backed, composable data fixtures and a parity tool; tutorials moved to `tests/e2e/tutorial/` with `save_screenshot_for_docs`.
  - CI: 4‑way sharded job with strict markers and a separate tutorials invocation; faster compose healthchecks; file filters include `tests/e2e/**` and `python_testcontainers/**`.
  - Runtime response delay: POST `/api/response-delay` (super‑admin) broadcasts `refresh.settings.response_delay`; all API workers subscribe to `refresh.settings.*`; GraphQL reads `config.SETTINGS.miscellaneous.response_delay` per request for live updates. Event documented.

- **Bug Fixes**
  - Data loaders/models: save symmetric `connected_endpoint` pairs sequentially; batched saves use `save_with_retry` to absorb transient create anomalies.
  - Frontend/Tests: use per‑node success toast IDs and prefix‑match selectors; bound reload‑until loops; 500‑response guards fail tests.

<sup>Written for commit f03ceafe1e8ca047774524d17c2a019ca54e23ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9527?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.github/workflows/ci.yml
```diff
@@ -1165,7 +1166,9 @@ jobs:
         run: pnpm exec playwright test artifact
 
       - name: Add response delay if required
-        if: needs.files-changed.outputs.e2e_tests == 'true'
+        # TEMP: disabled to measure real wall time; revert to
```
ajtmccarty: remember to turn back on

backend/infrahub/api/internal.py
```diff
@@ -59,6 +64,32 @@ async def get_info(request: Request, _: AccountSession = Depends(get_current_use
     return InfoAPI(deployment_id=str(registry.id), version=request.app.version)
 
 
+class ResponseDelayAPI(BaseModel):
+    response_delay: int = Field(ge=0, description="Delay in seconds added to each GraphQL request")
+
+
+@router.post("/response-delay", include_in_schema=False)
```
ajtmccarty, line 71: this seems like the kind of thing we don't want in a regular Infrahub app. is there some way to only add this endpoint if the app is being tested?

backend/infrahub/graphql/app.py
```diff
@@ -176,10 +177,12 @@ async def _get_on_get(self, request: Request) -> Response | None:
     async def _handle_http_request(
         self, request: Request, db: InfrahubDatabase, branch: Branch, account_session: AccountSession
     ) -> JSONResponse:
-        if request.app.state.response_delay:
-            self.logger.info(f"Adding response delay of {request.app.state.response_delay} seconds")
+        # Read live so the delay can be updated at runtime (see POST /api/response-delay).
+        response_delay = config.SETTINGS.miscellaneous.response_delay
+        if response_delay:
```
ajtmccarty, line 183: and maybe only apply the delay if the app is in testing mode too

models/infrastructure_edge.py
```diff
@@ -861,11 +879,21 @@ async def find_and_connect_interfaces(
...
     second_interface.description.value = f"Connected to {first_device_name}::{first_interface.name.value}"
-    batch.add(task=second_interface.save, node=second_interface)
+    second_interface.connected_endpoint = first_interface
+
+    batch.add(
+        task=save_connected_interfaces,
+        node=first_interface,
+        first_interface=first_interface,
+        second_interface=second_interface,
+    )
```
ajtmccarty, line 896: nice fix

AGENTS.md
```diff
@@ -55,9 +55,51 @@ cd frontend/app && pnpm install       # Install frontend dependencies
 uv run invoke backend.test-unit       # Backend unit tests
 uv run invoke backend.test-integration # Backend integration tests
 cd frontend/app && pnpm test          # Frontend unit tests
-cd frontend/app && pnpm test:e2e      # Frontend E2E tests
+cd frontend/app && pnpm test:e2e      # Frontend E2E tests (legacy TS suite)
+uv run pytest -c tests/e2e/pytest.ini tests/e2e  # E2E tests (pytest, testcontainers)
 ```
 
+#### Debugging e2e tests with `--pdb`
```
ajtmccarty, line 62: should this be in a skill or moved to its own separate file and linked to?

AGENTS.md
```diff
@@ -55,9 +55,51 @@ cd frontend/app && pnpm install       # Install frontend dependencies
...
+```
+
+Pytest then stays paused indefinitely on failure (`-s` keeps pdb's output
+unbuffered through the pipe). Inspect with docker/curl from a separate shell,
+then kill the process once done — pdb cannot receive commands over the dummy
+stdin. Match the node id, not the literal shell line (a bare
+`pkill -f <pattern>` kills your own shell if the pattern appears in its
+command line). Killing pytest skips the fixture teardown, so sweep the
+leftover stack afterwards: `docker compose -p $PROJECT down -v
+--remove-orphans`.
```
ajtmccarty, line 99: I don't really understand how it is possible to interact with pdb from a different shell, but I guess I don't have to

ajtmccarty (approved):

backend/infrahub/api/internal.py
fatih-acar: yes we could gate this behind a (disabled by default) setting, but this delay thing is handy in production while debugging timeout issues (when executing long running graphql queries within generators for example)

I'd like to keep this enabled by default, it's still behind a super admin permission check so should be safe?

AGENTS.md
fatih-acar: Yes I think this will move into a separate file dedicated to e2e test based development? Can do in a follow-up PR.

AGENTS.md
fatih-acar: > Inspect with docker/curl from a separate shell,
then kill the process once done — pdb cannot receive commands over the dummy
stdin

That's actually the point, it can't interact with pdb from a different shell... 
Here it means inspect the infrahub instance from a different shell. `--pdb` is used to not teardown the stack when a failure happens, so the agent can directly troubleshoot using docker commands and/or curl and/or infrahub sdk for example.

backend/infrahub/api/internal.py
```diff
@@ -59,6 +64,32 @@ async def get_info(request: Request, _: AccountSession = Depends(get_current_use
...
+    """Update the API response delay at runtime (testing facility).
+
+    The new value is broadcast over the message bus so every API worker process
+    across all server replicas applies it without a restart.
+    """
+    permission_manager.raise_for_permission(
+        permission=GlobalPermission(
+            action=GlobalPermissions.SUPER_ADMIN.value, decision=PermissionDecision.ALLOW_ALL.value
+        )
+    )
```
gmazoyer, line 86: Nice use of a permission

backend/tests/component/api/test_50_internals.py
```diff
@@ -141,3 +148,55 @@ async def test_no_search_docs(client: TestClient, no_search_index_path: None) ->
...
+) -> None:
+    with client:
+        response = client.post("/api/response-delay", headers=client_headers, json={"response_delay": 1})
+
+    assert response.status_code in (401, 403)
+
+
+async def test_response_delay_message_updates_settings() -> None:
+    original = config.SETTINGS.miscellaneous.response_delay
+    try:
```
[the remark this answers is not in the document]
gmazoyer: I think the bot is right, we usually `yield` in existing code

gmazoyer (approved):

backend/infrahub/api/internal.py
ajtmccarty: yes that sounds good. I didn't realize this had value in production as well

PR #9534, opened 2026-06-10
fix(config): trim whitespace from groups_claim setting

polmichel:
## Summary

The OAuth2/OIDC `groups_claim` provider setting validator rejected whitespace-only values but returned the raw value otherwise, so a padded value (e.g. `"groups "`) was used verbatim as a claim key and failed to resolve the group claim. The validator now returns the stripped value. Applies to both the OIDC and OAuth2 base settings.

No changelog fragment: the `groups_claim` feature is new and unreleased in 1.10, so this bugfix is covered by the feature's own changelog entry.

Surfaced by cubic's review of #9525.

### DIFF
--- a/backend/infrahub/config.py
+++ b/backend/infrahub/config.py
@@ -632,7 +632,7 @@ class SecurityOIDCBaseSettings(BaseSettings):
     def _validate_groups_claim(cls, value: str) -> str:
         if not value.strip():
             raise ValueError("groups_claim must not be empty or whitespace-only")
-        return value
+        return value.strip()
 
 
 class SecurityOIDCSettings(SecurityOIDCBaseSettings):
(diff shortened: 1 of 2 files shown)


changelog/+fix-groups-claim-whitespace.fixed.md
```diff
@@ -0,0 +1 @@
+The OAuth2/OIDC `groups_claim` provider setting now trims surrounding whitespace, so values configured with leading or trailing spaces resolve the group claim correctly.
```
ogenstad: Since this feature is set to be introduced in 1.10 we don't need a separate newsfragment to indicate that this was fixed. I.e. we'd want this newsfragment if the PR was a bug fix after 1.10 was released but as it is now we don't need to highlight it.

polmichel: Thanks for the review! I agree with that.
About the other one, I think the best would be to redirect the related commit to `stable` branch and keeping the changelog since it has been released in 1.9.8. I'll keep only the second fix into this PR.

For the context, I asked Claude to fix the relevant Cubic feedbacks and generate the PR in a draft state. I have not reviewed the code yet.

ajtmccarty (approved):
could this be covered with a unit test somewhere?

PR #9535, opened 2026-06-10
fix: floor computed-attribute chunk size at 1; dedup ipam utilization children

polmichel:
## Summary

Two small, independent backend fixes plus a docs path correction.

### 1. Computed-attribute chunk size floored at 1 (primary)

When `PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES` is set to `1`, the computed-attribute submission chunk size (that value integer-divided by 2) became `0`, and `range(0, len(ids), 0)` raises `ValueError: range() arg 3 must not be zero`, breaking batch submission.

The chunk-size computation is extracted into `_get_submission_chunk_size()` in `backend/infrahub/computed_attribute/tasks.py`, which floors the result at 1 via `max(1, ...)`. Both submission call sites now route through it.

Surfaced by cubic's review of #9525.

### 2. Dedup prefixes in IPAM utilization child lookup

`get_children` in `backend/infrahub/core/ipam/utilization.py` iterated the input prefix list directly, so passing the same prefix more than once double-counted its children and inflated utilization counts/percentages. Prefix ids are now collected into a set before lookup. No current caller passes duplicates; this is defensive hardening.

### 3. Docs

Corrected a stale `configuration.mdx` path in `dev/knowledge/backend/code-generation.md`.

### DIFF
--- a/backend/tests/component/core/ipam/test_ipam_utilization.py
+++ b/backend/tests/component/core/ipam/test_ipam_utilization.py
@@ -192,3 +192,16 @@ async def test_graphql_utilization_main_addition_after_branch_creation(
     assert isinstance(branch_prefix, BuiltinIPPrefix)
     branch_response = await branch_prefix.to_graphql(db=db, fields={"utilization": None})
     assert branch_response["utilization"] == {"value": 3}
+
+
+async def test_get_children_deduplicates_repeated_prefixes(
+    db: InfrahubDatabase, default_branch: Branch, ip_dataset_01: dict[str, Node]
+) -> None:
+    net143 = ip_dataset_01["net143"]
+    utilization = PrefixUtilizationGetter(db=db, ip_prefixes=[net143], branch=default_branch)
+
+    single = await utilization.get_children(ip_prefixes=[net143])
+    duplicated = await utilization.get_children(ip_prefixes=[net143, net143])
+
+    assert len(single) > 0
+    assert len(duplicated) == len(single)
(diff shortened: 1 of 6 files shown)


backend/tests/component/core/ipam/test_ipam_utilization.py
```diff
@@ -192,3 +192,16 @@ async def test_graphql_utilization_main_addition_after_branch_creation(
     assert isinstance(branch_prefix, BuiltinIPPrefix)
     branch_response = await branch_prefix.to_graphql(db=db, fields={"utilization": None})
     assert branch_response["utilization"] == {"value": 3}
+
+
+async def test_get_children_deduplicates_repeated_prefixes(
```
polmichel, line 197: As this a defensive fix which can not be reached in the current state of the code (based on the method's caller analysis), no changelog has been added.

backend/tests/unit/computed_attribute/test_tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.mark.parametrize(
+    ("configured_max", "expected"),
+    [
+        ("1", 1),  # 1 // 2 == 0 without the floor, which would break batching
+        ("2", 1),
+        ("10", 5),
+        ("500", 250),
+    ],
+    indirect=["configured_max"],
```
ajtmccarty, line 40: I did not know you could do this

ajtmccarty (approved):

backend/tests/unit/computed_attribute/test_tasks.py
polmichel: Me neither, this was a Claude generation

PR #9540, opened 2026-06-10
fix(git): correct added/removed file action in repository diffs

### LINKED ISSUE #9530: Diff: added file reported as "removed" in /api/diff/files
**Component:** Backend — Git integration / diff

**Infrahub version:** 1.9.6

## Current Behavior

When a branch adds a new file, the file diff **renders correctly in the UI** (the file's added content displays as expected), but the **`action` field is wrong** — the newly added file is labeled `removed` instead of `added`. The same wrong action comes straight from the backend `GET /api/diff/files`, so this is a backend classification bug, not a rendering issue. A *modified* file in the same diff is classified correctly as `updated` — only added (and, by symmetry, likely deleted) files are mislabeled.

API response (sanitized) for a branch that adds `transforms/test_transform.py` and edits `.infrahub.yml`:

```json
{ "<branch>": { "<repo-id>": {
    "commit_from": "<target_commit>",
    "commit_to":   "<source_commit>",
    "files": [
      { "location": ".infrahub.yml",                "action": "updated" },
      { "location": "transforms/test_transform.py", "action": "removed" }
    ] } } }
```

`commit_from`/`commit_to` are correct (target → source), and the modified file is right — only the **added** file's `action` is wrong.

## Expected Behavior

A file added on the source branch should have action `added` (so the Proposed Change Files tab badges it as added).

## Steps to Reproduce

1. Connect a read-write repository.
2. Create a git-sync-enabled branch; add a new file (e.g. a new transform) and commit/push it.
3. `GET /api/diff/files?branch=<branch>` (or open a Proposed Change → Files tab).
4. The added file's action is `removed`; a modified file in the same change is correctly `updated`.

## Additional Information

- Git ground truth: `git diff --name-status <commit_from> <commit_to>` shows the file as `A` (added).
- Likely root cause: the per-file *action* is computed in the **reverse** direction relative to the reported `commit_from`→`commit_to` — `added`↔`removed` swapped, while `updated` is symmetric so unaffected. A deleted file would likely show as `added`.
- Screenshot (to be attached): Proposed Change **Files** tab showing the added transform labeled as *removed* — diff content itself renders correctly.
- Environment: macOS arm64.

Image

lancamat1:
## Summary

Fixes #9530 (IFC-2723).

`calculate_diff_between_commits` inverted the per-file action: a file present only in the new commit was labelled `removed`, and one present only in the old commit `added`. This surfaced as added files showing **removed** in `GET /api/diff/files` and the Proposed Change → Files tab. `updated` is symmetric and was unaffected.

## Root cause

The `a_blob`/`b_blob` → action mapping was backwards relative to the function's own git operation (`first.diff(second)` = `git diff first second`, where `first` is the old/base side). Two callers passed commits in opposite order, so one path was correct (proposed-change gathering, by accident) and one was wrong (`/api/diff/files`).

## Changes

- **`git/base.py`** — corrected the mapping so a file present only in `second_commit` (new) is `added` and only in `first_commit` (old) is `removed`; documented the `first`=old / `second`=new direction.
- **`proposed_change/tasks.py`** — corrected the caller's argument order so the proposed-change repo-diff path stays correct after the function fix (it previously relied on the inverted behaviour via reversed arguments).
- **`test_git_repository.py`** — the component test now calls in the API consumption direction and asserts git-ground-truth-correct actions.

## Testing

- `test_calculate_diff_between_commits` updated to call in the real API direction; confirmed it fails before the fix (added file reported as `removed`) and passes after.
- Full git component suite green: 61 passed, 1 xfailed (pre-existing).

---
## Summary by cubic
Fixes inverted added/removed actions in repository diffs. Added files now show added and deleted files removed in `GET /api/diff/files` and the Proposed Change Files tab (IFC-2723).

- **Bug Fixes**
  - Fixed `calculate_diff_between_commits` to follow `git diff first second` (first=old, second=new); updated the docstring.
  - Corrected commit order in `proposed_change/tasks.py` for proposed-change diffs.
  - Updated the component test to assert actions in the API direction.

<sup>Written for commit 9267d6cb03aaf7be1ce2434428025f9c1c47a8b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -749,7 +749,13 @@ class InfrahubRepositoryBase(BaseModel, ABC):
     async def calculate_diff_between_commits(
         self, first_commit: str, second_commit: str
     ) -> tuple[list[str], list[str], list[str]]:
-        """TODO need to refactor this function to return more information.
+        """Return the (changed, added, removed) files going from first_commit to second_commit.
+
+        Direction follows `git diff first_commit second_commit`: first_commit is the old/base side
+        and second_commit is the new/target side. A file present only in second_commit is "added";
+        a file present only in first_commit is "removed".
+
+        TODO need to refactor this function to return more information.
 
         Like :
           - What has changed inside the files
@@ -765,12 +771,12 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         added_files = []
 
         for x in commit_in_branch.diff(commit_to_compare, create_patch=True):
-            if x.a_blob and not x.b_blob and x.a_blob.path not in added_files:
-                added_files.append(x.a_blob.path)
+            if x.a_blob and not x.b_blob and x.a_blob.path not in removed_files:
+                removed_files.append(x.a_blob.path)
             elif x.a_blob and x.b_blob and x.a_blob.path not in changed_files:
                 changed_files.append(x.a_blob.path)
-            elif not x.a_blob and x.b_blob and x.b_blob.path not in removed_files:
-                removed_files.append(x.b_blob.path)
+            elif not x.a_blob and x.b_blob and x.b_blob.path not in added_files:
+                added_files.append(x.b_blob.path)
 
         return changed_files, added_files, removed_files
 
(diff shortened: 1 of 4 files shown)


ajtmccarty (approved):
nice quick fix

PR #9541, opened 2026-06-10
prevent writes to default/merging branch during a merge

ajtmccarty:
# Why

During a branch merge into the default branch, the default branch (and the branch being merged) stay open to API writes. A write that lands mid-merge can interleave with the merge's graph writes and leave the result inconsistent, and clients get no signal that a merge is in progress.

Goal: block API writes to the default branch and to the branch being merged for the entire merge window, refuse new merges/rebases while one is in progress, and lift the block automatically when the merge completes or rolls back.

A couple exceptions to the write block: `BranchCreate` is always allowed and `BranchDelete` is allowed for any branches other than the one being merged.

Non-goals: failed-merge detection, the `MERGE_FAILED` status, and the `infrahub recover` command (later increments US2/US3); repository (git) merges.

Closes IFC-2562

## What changed

**Behavioral changes:**

- While a merge is in progress, writes to the **default branch** are rejected with a transient message (`"A merge is currently in progress; writes are temporarily blocked. Please retry shortly."`). It becomes writable again once the merge completes.
- Writes to the **branch being merged** are rejected as read-only, mirroring a `MERGED` branch b/c it is heading to `MERGED`.
- New **merges and rebases** are refused while any branch is being merged.
- The block is **lifted automatically** on a successful merge and on a rollback.
- The block is **immediately consistent across workers** (a shared `merge:protected` cache key, read once per mutation). If the cache is unreachable, the gate logs and falls back to the durable DB branch status, so a cache outage blocks writes only when a merge is genuinely in progress.

**Implementation notes:**

- moved `merge_repositories()` out of `BranchMerger` and into `_do_merge_branch` so that it can be placed outside of the `MERGING` write-block b/c `merge_repositories()` calls a workflow that sends a graphql mutation to update a repository's commit and this would be blocked while the branch is in its `MERGING` status
- New `MergeWriteBlocker` DI component (`backend/infrahub/core/merge/write_blocker.py`) owns the `merge:protected` cache key (`set`/`get`/`delete`, value `"{branch}::{state}"`).
- `Branch.merge_started_at` persisted on the `:Branch` node, set once at `OPEN→MERGING` (records the merge start for the future recovery rollback). Stored as `Optional[str]` (ISO string) — a `Timestamp` field can't serialize through `StandardNode`, so it mirrors the existing `branched_from` pattern.
- `_do_merge_branch` sets the key + `merge_started_at` before any graph write and deletes the key at `MERGED` / in `_rollback_merge`; `merge_branch` and `rebase_branch` reject when another branch is protected.
- `BranchStatusChecker` is now `async` and takes `db` + `MergeWriteBlocker`. It enforces the source/target gates and the cache-unreachable DB fallback. Wired into the GraphQL mutation middleware (now async) and the REST `schema`/`artifact` handlers.

**What stayed the same:**

- No database schema or migration changes.
- The `MERGED` read-only behavior and message are unchanged.
- No new external GraphQL/REST contract; no new config or feature flags.

## Impact & rollout

- **Backward compatibility:** No breaking API/schema changes; the behavior is additive write-gating. Clients may now receive a transient or read-only error during a merge (the intended new behavior).
- **Performance:** One extra cache `GET` per mutation (sub-ms, before any transaction; gated to the top-level mutation field). A branch-status DB query is issued only when the cache is unreachable.
- **Config/env changes:** None. Relies on the existing distributed cache (Redis or NATS).
- **Deployment notes:** Safe to deploy; no coordinated release required.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../dev/guidelines/changelog.md) added (`changelog/+merge-write-protection.added.md`)
- [ ] External docs updated (if user-facing or ops-facing change) — n/a (no doc-site surface; only error-message wording)

---
## Summary by cubic
Blocks writes to the default branch and the merging source branch for the full merge window, and refuses new merges/rebases until the merge completes or rolls back. Protection is immediate across workers via a shared `merge:protected` key with a durable DB fallback so cache outages don’t freeze writes (IFC-2437).

- New Features
  - Default-branch writes return a transient “merge in progress, retry shortly”; writes to the source branch are rejected as read‑only. The block lifts on success or a clean rollback; if rollback fails mid‑way, protection remains for recovery.
  - New merges and rebases are refused while protection is active. `BranchCreate` and `BranchDelete` are allowed, but deleting the branch being merged is blocked.
  - Cross-worker gate via `MergeWriteBlocker`; persists `Branch.merge_started_at` (`Optional[str]`, ISO) at `OPEN→MERGING` for recovery. If the cache is unreachable, the gate logs and falls back to the durable DB `MERGING` status.

- Refactors
  - `BranchStatusChecker` is async with `db` + `MergeWriteBlocker` and powers the GraphQL mutation middleware (top-level field gate) and REST `schema`/`artifact` endpoints.
  - Repository (git) merge runs after the graph merge commits and after protection lifts, so it doesn’t self‑block.
  - Tests cover the cache‑unreachable fallback and end‑to‑end merge‑in‑progress blocking; GraphQL tests use an in‑memory cache.

<sup>Written for commit af6d18ea68436d5999cb6657bb9aca951fc09861. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9541?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ajtmccarty:
@claude analyze the E2E test failures on the latest CI run. I suspect that some tests are running while a merge is in progress and are then blocked by the new write-guard.

ajtmccarty:
@claude I added exemptions for BranchCreate and BranchDelete for the merge-write guard in my latest commit, but it looks like one of the E2E tests still failed with the "A merge is currently in progress" error message. Investigate why this error was still raised during a BranchCreate mutation

backend/infrahub/branch/status_checker.py
```diff
@@ -1,27 +1,97 @@
...
+                identifier=branch.name, message=_merging_branch_message(branch.name)
+            ) from None
+
+        if branch.is_default and merging_branch_names:
+            raise BranchAlreadyMergedError(identifier=branch.name, message=MERGE_IN_PROGRESS_MESSAGE) from None
+
+    async def check(self, branch: Branch) -> None:
         self.check_needs_rebase_status(branch)
         self.check_merge_status(branch)
+        await self.check_merging_status(branch)
```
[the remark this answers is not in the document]
ajtmccarty: this should be addressed as the cache key is now set before the status is updated

backend/tests/functional/merge/test_merge_in_progress_block.py
```diff
@@ -0,0 +1,114 @@
...
+        await client.branch.create(branch_name=merging_name)
+        await client.branch.create(branch_name=blocked_name)
+
+        merge_write_blocker = MergeWriteBlocker(cache=service.cache)
+        await merge_write_blocker.set(branch=merging_name, state=MergeProtectionState.MERGING)
+
+        try:
+            with pytest.raises(GraphQLError) as merge_exc:
+                await client.branch.merge(branch_name=blocked_name)
+            assert merge_exc.value.errors[0]["message"] == MERGE_IN_PROGRESS_MESSAGE
```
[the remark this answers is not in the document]
ajtmccarty: this has been addressed

backend/tests/functional/merge/test_merge_in_progress_block.py
```diff
@@ -0,0 +1,114 @@
...
+
+            with pytest.raises(GraphQLError) as rebase_exc:
+                await client.branch.rebase(branch_name=blocked_name)
+            assert rebase_exc.value.errors[0]["message"] == MERGE_IN_PROGRESS_MESSAGE
```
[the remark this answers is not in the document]
ajtmccarty: this has been addressed

backend/infrahub/branch/status_checker.py
```diff
@@ -1,27 +1,97 @@
...
+
+        Raises:
+            BranchAlreadyMergedError: if the branch is blocked by a merge in progress.
+
+        """
+        try:
+            protection = await self.merge_write_blocker.get()
+        except Exception:
+            log.exception("merge-protection cache unreachable; falling back to branch status in the database")
+            await self._check_merging_status_from_db(branch)
```
ajtmccarty, line 64: if we fail to read the "merge-in-progress" key from the cache, then we fall back to getting any MERGING branches from the database. this should be very rare

backend/infrahub/core/branch/models.py
```diff
@@ -26,6 +26,8 @@
 
 
 class Branch(StandardNode):
+    # Persisted nullable fields must use `Optional[X]`` rather than `X | None` until we move to
+    # Python 3.14 b/c of how StandardNode.guess_field_type works
```
ajtmccarty, line 30: I didn't really dig into it but it something about how typing resolves `Optional[X]` vs `X | None` that will change in `3.14`

backend/infrahub/core/branch/tasks.py
```diff
@@ -355,12 +369,18 @@ async def _rollback_merge(
         merger.destination_branch.update_schema_hash()
         await merger.destination_branch.save(db=db, user_id=user_id)
     except Exception:
+        # Hold the write protection (leave the branch MERGING + key set) rather than reopening on a
+        # partially-recovered state, so the failure is handled by recovery instead of allowing writes.
         log.exception("Registry restore failed during merge rollback")
+        return False
```
ajtmccarty, line 375: the failure recovery tool (`infrahub recover` CLI) command will be able to handle this failure to roll back the schema in the registry. this failure state should be extremely rare. also, restarting the workers would fix it.

backend/infrahub/core/merge/branch_merger.py
```diff
@@ -192,7 +192,6 @@ async def merge(
                 log.exception("Merge failed, beginning rollback")
                 await self.rollback()
                 raise MergeFailedError(branch_name=self.source_branch.name) from exc
-        await self.merge_repositories()
```
ajtmccarty, line 195: moved to the new little `_merge_repositories_after_merge()` in `branch/tasks.py` above

backend/tests/unit/branch/test_status_checker.py
ajtmccarty, line 1: these two deleted test files overlapped with new tests added at the component level, so just kept the component level ones b/c they are required

frontend/app/tests/e2e/utils/graphql.ts
ajtmccarty, line 1: @opsmill/frontend could you please review the change in this testing file. it should be a retry loop for when the merge-is-in-progress error is raised. the point of this PR is to block writes to the default branch while any merge is in progress, so adding this little wait should allow tests waiting on a merge to still succeed... hopefully. later PR will match on the error more precisely using the new graphql error catalogue, but this PR has already crept out of its scope somewhat

backend/infrahub/branch/status_checker.py
```diff
@@ -1,27 +1,97 @@
...
+        a merge is genuinely in progress rather than freezing every default-branch write.
+
+        Raises:
+            BranchAlreadyMergedError: if the branch is blocked by a merge in progress.
+
+        """
+        try:
+            protection = await self.merge_write_blocker.get()
+        except Exception:
+            log.exception("merge-protection cache unreachable; falling back to branch status in the database")
```
polmichel, line 63: I was thinking about the scope of exceptions caught here, but I assume we can't narrow it. Would logging the exception inside the message provide useful information in case this happens?

polmichel, line 63: I was thinking about the scope of caught Exceptions here but I assume we cannot reduce this scope. Would logging the Exception provide useful information in case this happens?

backend/tests/adapters/cache.py
```diff
@@ -26,3 +26,24 @@ async def set(self, key: str, value: str, expires: int | None = None, not_exists
         return True
 
     async def close_connection(self) -> None: ...
+
+
+class UnreachableCache(InfrahubCache):
```
polmichel, line 31: Nice adapter

backend/infrahub/core/branch/models.py
polmichel: Optional comment:
Out of curiosity I looked into it: https://docs.python.org/3/library/typing.html

_Changed in version 3.14: [types.UnionType](https://docs.python.org/3/library/types.html#types.UnionType) is now an alias for Union, and both Union[int, str] and int | str create instances of the same class. To check whether an object is a Union at runtime, use isinstance(obj, Union). For compatibility with earlier versions of Python, use get_origin(obj) is typing.Union or get_origin(obj) is types.UnionType._

So I guess an option could be to add inside the `infrahub.core.node.standard.StandardNode.guess_field_type` code
```
        if (annotation_origin == Union and len(annotation_args) == 2 and type(None) in annotation_args) or (
            annotation_origin is list and len(annotation_args)
        ):
            return get_origin(annotation_args[0]) or annotation_args[0]

```

the following "fix"

```
        if (annotation_origin in (Union, UnionType) and len(annotation_args) == 2 and type(None) in annotation_args) or (
            annotation_origin is list and len(annotation_args)
        ):
            return get_origin(annotation_args[0]) or annotation_args[0]

```

It allows `StandardNode.guess_field_type` to correctly resolve `int | None` type to `int`.
Not sure this is required though.

polmichel (approved):

PR #9550, opened 2026-06-11
feat(backend): add user and global preferences (IFC-2719, IFC-2722)

pa-lem:
# Why

UI settings like date format and timezone live in component state or the browser locale today: an organisation cannot set defaults centrally and a user's choice doesn't survive a device switch. This PR adds the backend foundation — two schema-backed preference surfaces and one merged read query — so the frontend (follow-up PRs) can render dates per user preference.

Non-goals: frontend tabs/hooks (IFC-2720), DateDisplay integration (IFC-2721), filtering generic *reads* of `CoreUserPreference` (mutation-only enforcement in V1, deliberate).

Part of epic IFC-2718 · Spec: `dev/specs/2026-04-user-preferences.md` (branch `ple-user-preferences`) · Idea: INFP-512

## What changed

- **Schema**: `CoreGlobalPreference` (singleton, branch-agnostic) and `CoreUserPreference` (one per account, cascade-deleted, unique on `account`), both with optional `date_format`/`timezone` Text attributes.
- **Permissions**: new `manage_global_preferences` global permission, wired into `get_global_permission_for_kind()` + a dedicated checker (mirrors `RepositoryManagerPermissionChecker`). Without it, the default General Access wildcard would let every user edit org defaults.
- **Effective query**: `InfrahubEffectivePreferences` returns scalar `date_format`/`timezone` merged user > global > null for the calling account.
- **Owner scoping**: `CoreUserPreference` writes are restricted to the owning account (admins bypass); id-less upserts are idempotent (relationship-based uniqueness derives no hfid — see `dev/knowledge/backend/mutations.md`); singleton creation is lock-guarded.
- **Seeding**: `first_time_initialization()` + migration m074 for existing installs (permission node + empty singleton; role assignment beyond super-admins is a documented manual step).
- Knowledge docs + towncrier fragments included.

## Testing

93 targeted tests green (merge matrix, ownership attack paths, permission gating admin/wildcard/role-granted, migration idempotency, schema integrity) + ruff/mypy clean.

---
## Summary by cubic
Adds backend support for per-user and org-wide UI preferences (`date_format`, `timezone`) via an owner-scoped `Preference` `StandardNode`, three typed GraphQL reads, and a single write mutation. Global edits require `manage_global_preferences`; identity-bound queries require auth; writes are serialized per owner for safe concurrency (IFC-2719, IFC-2722).

- New Features
  - Model: `Preference` keyed by `owner_id` (account id or `GLOBAL_OWNER_ID`); reads never create; per-owner lock serializes writes; no seeding; account deletion can orphan rows (IFC-2867).
  - Repository: `get_for_owner`, `get_for_owners`, `get_all`, `save` with deterministic ordering and first-row wins if duplicates exist.
  - GraphQL reads: `InfrahubEffectivePreferences` returns `date_format`/`timezone` with `source` = `USER` | `GLOBAL` | `DEFAULT`; `InfrahubUserPreferences` and `InfrahubGlobalPreferences` return raw values. All require auth; `GLOBAL` read/write gated by `manage_global_preferences`.
  - GraphQL write: `InfrahubSetPreferences(scope: USER|GLOBAL, date_format?, timezone?)`; omitted args are ignored; explicit `null` resets.
  - Date formats: `DateFormat` enum (`ISO_8601`, `ISO_DATETIME`, `ISO_DATETIME_SECONDS`, `EU_DATETIME`, `US_12H`) persisted and validated by the GraphQL enum.

- Refactors
  - Persistence centralized in `PreferenceRepository`; read queries return only required fields; `GLOBAL_OWNER_ID` replaces Root-id coupling; lock namespace moved to constants.
  - Auth: anonymous sessions are denied for identity-bound operations via `QUERIES_REQUIRING_AUTHENTICATION`; `InfrahubSetPreferences` is exempt from the default-branch write gate; `GLOBAL` scope enforced via `manage_global_preferences`. 
  - Query layer: copy shared return-labels to avoid cross-instance mutation; `timezone` remains free-form (no backend validation).

<sup>Written for commit c85d09f7d4d6bf5e66ae4c6be42cd5045b4f8426. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9550?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pa-lem:
## 🧪 Manual testing (GraphiQL)

Paste the whole block below into the Infrahub **GraphiQL** sandbox (it uses your logged-in session). Every operation is **named** and uses **inline arguments**, so no Variables pane is needed — just pick each operation from the ▶ run dropdown and run them top-to-bottom. Each query carries inline `#` comments explaining the logic and the expected result.

> `date_format` is the **`DateFormat` enum** — valid keys: `ISO_8601`, `ISO_DATETIME`, `ISO_DATETIME_SECONDS`, `EU_DATETIME`, `US_12H`.
> Steps **3, 7, 8** exercise the `manage_global_preferences` gate — run them once as an account **with** the permission (or super-admin) and once **without** to see both the allow and the fail-closed deny paths.

```graphql
# ============================================================================
# READS — InfrahubPreferences(scope). Auth required for every scope.
# ============================================================================

# 1) EFFECTIVE (default scope) = the caller's resolved view: user value, else the
#    org global, else the built-in default. Read-only projection.
#    EXPECT: one entry per key {key, value, source}. On a fresh account with no
#    global set, source = DEFAULT and value = null. `can_edit_global_preferences`
#    reflects whether you hold manage_global_preferences.
query ReadEffective {
  InfrahubPreferences {
    preferences { key value source }
    can_edit_global_preferences
  }
}

# 2) USER = only the caller's OWN raw overrides (bound to your account; there is no
#    account argument, so you can never read another user's row).
#    EXPECT: your values, or value=null where you have no override; source always USER.
query ReadUser {
  InfrahubPreferences(scope: USER) { preferences { key value source } }
}

# 3) GLOBAL = the org-wide singleton's raw values. Fail-closed gate.
#    EXPECT (with permission): the org values, source GLOBAL.
#    EXPECT (without permission): an errors[] entry, NO data — the gate raises before any read.
query ReadGlobal {
  InfrahubPreferences(scope: GLOBAL) { preferences { key value source } }
}

# ============================================================================
# WRITES — InfrahubSetPreferences(scope, date_format, timezone)
# Omitted arg = leave unchanged; explicit null = reset that field.
# ============================================================================

# 4) Set MY override (date_format is a semantic key, not a rendering pattern).
#    EXPECT: ok=true, echoes date_format=EU_DATETIME, timezone="Europe/Paris".
#    Then re-run #2 (source USER) and #1 (EFFECTIVE now shows these, source USER).
mutation SetUser {
  InfrahubSetPreferences(scope: USER, date_format: EU_DATETIME, timezone: "Europe/Paris") {
    ok date_format timezone
  }
}

# 5) Reset ONE of my fields — explicit null clears date_format; timezone omitted stays.
#    EXPECT: ok=true, date_format=null, timezone still "Europe/Paris".
#    Then re-run #1: date_format falls back to GLOBAL (or DEFAULT if global unset),
#    while timezone is still your USER override.
mutation ResetUserDateFormat {
  InfrahubSetPreferences(scope: USER, date_format: null) { ok date_format timezone }
}

# 6) Idempotency / no duplicate rows: run #4 several times, then #2.
#    EXPECT: still exactly one set of USER values (no duplicate user row is created).

# 7) Set the ORG default. Fail-closed gate.
#    EXPECT (with permission): ok=true, echoes ISO_8601 / "UTC".
#    EXPECT (without permission): errors[], and NOTHING is written (re-run #3 as an
#    admin to confirm the singleton was untouched).
mutation SetGlobal {
  InfrahubSetPreferences(scope: GLOBAL, date_format: ISO_8601, timezone: "UTC") {
    ok date_format timezone
  }
}

# 8) EFFECTIVE is a read-only projection — it is never writable.
#    EXPECT: errors[] containing "read-only"; no write occurs.
mutation EffectiveIsReadOnly {
  InfrahubSetPreferences(scope: EFFECTIVE, timezone: "UTC") { ok }
}

# 9) Enum validation: an unknown date_format key is rejected at the GraphQL layer,
#    BEFORE any write.
#    EXPECT: errors[] naming the `DateFormat` enum / the bad value; no row created.
mutation InvalidDateFormat {
  InfrahubSetPreferences(scope: USER, date_format: NOT_A_FORMAT) { ok }
}
```

### Privacy check (two accounts)
Log in as **account A**, run #4 to set an override. Log in as **account B**, run #2 (USER) and #1 (EFFECTIVE) — B must **never** see A's values (each read is bound to the session's own account; there is no account argument to target another user).

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class GlobalPreference(StandardNode):
+    # Persisted nullable fields must use `Optional[X]` rather than `X | None` until we move to
+    # Python 3.14 b/c of how StandardNode.guess_field_type works (mirrors Branch).
+    date_format: Optional[str] = None
+    timezone: Optional[str] = None
+
+    @classmethod
+    async def get_global(cls, db: InfrahubDatabase) -> Self:
```
ogenstad: We should avoid having a method called `get_xyz()` that can end up writing to the database. In a distributed environment we'd want any read call to land on any Neo4j server but calls that write to the database must land on the primary one. Just looking at the method name this is not clear. I'd suggest that we have a resolver that would just return some "not set" data if the global preferences doesn't exist.

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,86 @@
...
+            return obj
+
+
+class UserPreference(StandardNode):
+    # Orphan-on-account-deletion is accepted for V1: a StandardNode cannot declare a schema
+    # relationship with `on_delete: cascade` (that is a schema-Node feature), so deleting an
+    # account leaves its UserPreference row behind. Account ids are UUIDs (UUIDT) and are never
+    # reused, so such a row is permanently unreachable dead data — benign. Cleanup is out of scope
+    # for V1; there is no cascade mechanism for StandardNode and we deliberately do not convert this
+    # to a schema node.
```
ogenstad: This way of leaving these nodes behind should be fine, however it would be good with an issue in Jira to track it. Later we can have different types of cascade options.

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,86 @@
...
+            existing = await cls.get_list(db=db, limit=1)
+            if existing:
+                return existing[0]
+
+            obj = cls()
+            await obj.create(db=db)
+            return obj
+
+
+class UserPreference(StandardNode):
```
ogenstad: I think we have a problem here with regards to having multiple classes where the `UserPreference` type might differ from the `GlobalPreference` one. I see two options here. Either we just have the UserPreference one and instead of account_id we have some `owner_id` field that can either refer to the account_id of a user or the `Root` id of the system then this would indicate that it's a global preference object.

Alternatively the UserPreference class would be based on GlobalPreference so that we could share common attributes such as date_format and timezone.

backend/infrahub/core/preferences/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    GlobalPreference,
+    UserPreference,
+)
+
+# Permission required to manage the org-wide GlobalPreference singleton. Defined once here and
+# shared by both the preferences query and mutation modules to keep the gate consistent.
+MANAGE_GLOBAL_PREFERENCES_PERMISSION = GlobalPermission(
+    action=GlobalPermissions.MANAGE_GLOBAL_PREFERENCES.value,
+    decision=PermissionDecision.ALLOW_ALL.value,
+)
```
ogenstad: We should avoid having code directly in a `__init__.py` file. We do it in a lot of other places but try to avoid it for new code.

backend/infrahub/core/preferences/formats.py
```diff
@@ -0,0 +1,46 @@
...
+# Applied when neither the user nor the global singleton has set date_format. Kept in sync with the
+# frontend default (same key) so a date the backend renders matches what the UI would show.
+DEFAULT_DATE_FORMAT = "ISO_DATETIME"
+
+# Valid semantic keys in display order. The GraphQL `DateFormat` enum is built from this tuple so
+# the enum and the render map can never drift apart.
+DATE_FORMAT_KEYS: tuple[str, ...] = tuple(DATE_FORMAT_STRFTIME)
+
+
+def render_datetime(value: datetime, date_format: str | None) -> str:
```
ogenstad: Is this function used anywhere outside of tests?

backend/infrahub/core/preferences/formats.py
```diff
@@ -0,0 +1,46 @@
...
+# format is one entry here plus the frontend's key->date-fns map; StandardNode stores the key
+# string, so there is no schema or migration change.
+#
+# The set is deliberately limited to formats that need no locale LIBRARY and no localized month/day
+# names (which would render differently per client and require babel server-side) and no relative
+# mode — those were left out for exactly that reason (see dev/specs/2026-04-user-preferences.md).
+# The one built-in locale touch is US_12H's `%p` (AM/PM): under the backend's default C/POSIX
+# locale — which Infrahub does not override with locale.setlocale(LC_TIME, ...) — it renders English
+# "AM"/"PM", matching the web client's date-fns `a` token in its default (en-US) locale.
+DATE_FORMAT_STRFTIME: dict[str, str] = {
```
ogenstad: I think it would be better if the date format was an enum, then we could have the other part as values. It should also allow us to have the `date_format: Optional[str] = None` part as an enum instead of having it look like we support any string. There are other places where we convert a Python enum into a GraphQL enum, so we could then re-use that.

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,161 @@
...
+            return await cls._set_user(
+                graphql_context, account_id=account_id, date_format=date_format, timezone=timezone
+            )
+        if scope == SCOPE_GLOBAL:
+            return await cls._set_global(
+                graphql_context, account_id=account_id, date_format=date_format, timezone=timezone
+            )
+        if scope == SCOPE_EFFECTIVE:
+            # The resolved view is a read-only projection; there is nothing concrete to write.
+            raise ValidationError("The EFFECTIVE scope is read-only; write to USER or GLOBAL instead")
```
ogenstad: Here it looks like we have a guard against a value in an enum that should only ever be used for reading. I'd suggest to remove this so that the mutation to set values points to an enum with only two possible entries. I'm not sure we need the enum for the query we could instead have different fields where one is "user" another is "global" and then the third is "effective".

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,161 @@
...
+    GLOBAL_PREFERENCE_LOCK_NAME,
+    GLOBAL_PREFERENCE_LOCK_NAMESPACE,
+    MANAGE_GLOBAL_PREFERENCES_PERMISSION,
+    GlobalPreference,
+    UserPreference,
+)
+from infrahub.database import retry_db_transaction
+from infrahub.exceptions import PermissionDeniedError, ValidationError
+
+from ..queries.preferences import SCOPE_EFFECTIVE, SCOPE_GLOBAL, SCOPE_USER, DateFormat, PreferenceScope
```
ogenstad: The mutation shouldn't import from a query if these are shared resources they should be defined elsewhere.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+class PreferencesType(ObjectType):
+    """Per-key preference entries for the requested scope plus the org-defaults edit gate.
+
+    Privacy: EFFECTIVE and USER entries are account-bound (the resolver reads only
+    account_session.account_id), so no account ever sees another account's user preferences.
+    GLOBAL entries are org-wide and returned only after the manage_global_preferences gate.
+    """
+
+    preferences = Field(List(of_type=NonNull(PreferenceEntryType), required=True), required=True)
+    can_edit_global_preferences = Field(Boolean, required=True)
```
ogenstad: It feels weird that this query should return permission information. Is this because the frontend needs to know if the user is allowed to set the global preferences? I think we should use a generic query to gather that information instead of sprinkling permission information in other queries.

backend/infrahub/core/initialization.py
```diff
@@ -554,6 +556,11 @@ async def first_time_initialization(db: InfrahubDatabase) -> None:
     default_branch = await create_default_branch(db=db)
     await create_global_branch(db=db)
 
+    # Seed the GlobalPreference singleton so new installs have it and the effective-preferences
+    # read path (GlobalPreference.get_global) stays lock-free. Existing installs rely on that
+    # method's lazy create-with-lock instead.
+    await GlobalPreference.get_global(db=db)
```
ogenstad: I don't think we should set any global preference during init and instead allow it to be empty.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+    UserPreference,
+)
+from infrahub.exceptions import PermissionDeniedError
+
+if TYPE_CHECKING:
+    from graphql import GraphQLResolveInfo
+
+    from infrahub.graphql.initialization import GraphqlContext
+
+PREFERENCE_ATTRIBUTES = ("date_format", "timezone")
```
ogenstad: This gets disconnected from the values we have in the classes. We should have a single reference to the attributes so that they aren't spread out in multiple places.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+        preferences = []
+        for attribute_name in PREFERENCE_ATTRIBUTES:
+            user_value: str | None = getattr(user_preference, attribute_name) if user_preference else None
+            global_value: str | None = getattr(global_preference, attribute_name)
+            if user_value is not None:
+                preferences.append(_entry(attribute_name, user_value, PreferenceSource.USER))
+            elif global_value is not None:
+                preferences.append(_entry(attribute_name, global_value, PreferenceSource.GLOBAL))
+            else:
+                preferences.append(_entry(attribute_name, None, PreferenceSource.DEFAULT))
```
ogenstad: I think we should do this in a different way and have the a function that would combine the two objects, with some logic in the classes we could get away from having to iterate over a disconnected list and instead have the StandardNode classes determine which attributes it is we want to expose and export.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+    preferences: list[dict]
+
+    if scope == SCOPE_EFFECTIVE:
+        # Any authenticated caller. The global singleton is read INTERNALLY (no permission gate):
+        # the caller only ever gets their own resolved view, never the raw org values as such.
+        # StandardNode reads carry no branch filter, so these lookups are branch-agnostic.
+        global_preference = await GlobalPreference.get_global(db=db)
+        user_preference = await UserPreference.get_for_account(db=db, account_id=account_id)
```
ogenstad: I don't know if we ever want to have different global preferences vs user preferences. But as per the comment above if we store everything in one class we should be able to get all of the preferences in one go where we query for anything connected to the current user along with that of the root ID.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+
+class PreferencesType(ObjectType):
+    """Per-key preference entries for the requested scope plus the org-defaults edit gate.
+
+    Privacy: EFFECTIVE and USER entries are account-bound (the resolver reads only
+    account_session.account_id), so no account ever sees another account's user preferences.
+    GLOBAL entries are org-wide and returned only after the manage_global_preferences gate.
+    """
+
+    preferences = Field(List(of_type=NonNull(PreferenceEntryType), required=True), required=True)
```
ogenstad: I'm not sure that the preferences should be a list. First of all it hides the available preferences from the schema. So you'd have to keep a matching list of all of the names in the frontend. Secondly it restricts all values to be strings. I'm not sure what we want to store here in the future but even with only the two we have at least the date_format we know that's an enum and if we keep it as an enum you could use the same enum in the frontend with the bindings. Shouldn't we have proper fields in the response as well?

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+# Well-known distributed-lock coordinates for the GlobalPreference singleton. A fixed name in a
+# dedicated namespace lets every worker serialise the lazy-create through the same global lock.
+GLOBAL_PREFERENCE_LOCK_NAME = "singleton"
+GLOBAL_PREFERENCE_LOCK_NAMESPACE = "global_preference"
+
+
+class GlobalPreference(StandardNode):
+    # Persisted nullable fields must use `Optional[X]` rather than `X | None` until we move to
+    # Python 3.14 b/c of how StandardNode.guess_field_type works (mirrors Branch).
```
polmichel: Not blocking but it reminds me this review I've made https://github.com/opsmill/infrahub/pull/9541#discussion_r3403421067

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,161 @@
...
+    async def _set_global(
+        cls,
+        graphql_context: GraphqlContext,
+        account_id: str,
+        date_format: str | _Unset | None,
+        timezone: str | _Unset | None,
+    ) -> Self:
+        # Gated: raise BEFORE any read-modify-write (fail-closed). Super admins bypass via the
+        # permission manager.
+        graphql_context.active_permissions.raise_for_permission(permission=MANAGE_GLOBAL_PREFERENCES_PERMISSION)
```
polmichel: Permission should not be managed inside a method which is supposed to set the value I think this method should remain a pure value set and permissions should be managed outside

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,174 @@
...
+
+# Scope string values. Kept as plain constants (not read off the graphene Enum members via
+# `.value`) because graphene's Enum metaclass makes static type-checkers treat members as `str`,
+# which breaks `.value`. Graphene passes these string values to the resolver/mutation at runtime.
+SCOPE_EFFECTIVE = "effective"
+SCOPE_GLOBAL = "global"
+SCOPE_USER = "user"
+
+
+class PreferenceScope(Enum):
```
polmichel: Can you use directly a `StrEnum` directly?

pa-lem:
## ✅ Review follow-up — preferences redesigned

Thanks for the thorough review. The design comments pointed at a cleaner shape, so we've reworked the feature accordingly. Backend changes are in this PR; the frontend counterpart is in #9776. Summary of how each open comment was addressed:

| Comment | Resolution |
|---|---|
| **`get_global()` writes on read** | Reads **never create** now — `Preference.get_for_owner` / `get_for_owners` only read; the row is created only on the first write. |
| **Two classes (`Global`/`User`) may drift** | Collapsed into **one `Preference` `StandardNode` keyed by `owner_id`** — an account id, or the Root id (`registry.id`) for global. |
| **Two queries for effective / merge in resolver / attr list** | The effective read fetches both owners in **one query** and returns **typed fields**; the hand-maintained `PREFERENCE_ATTRIBUTES` tuple is gone. |
| **`date_format` should be an enum** | New `DateFormat` **Python enum** is the single source of truth: the GraphQL enum is `Enum.from_enum(DateFormat)` and the model validates against it. |
| **`render_datetime` unused outside tests** | **Removed** the server-side renderer + strftime map (clients render). This also makes the `%p`-locale comment moot. |
| **Write enum incl. read-only EFFECTIVE** | Write mutation now takes a **write-only `PreferenceWriteScope` (`USER`/`GLOBAL`)** — `EFFECTIVE` is unrepresentable; the runtime guard is gone. |
| **Reads should be separate fields, not a scope arg** | Three typed root fields: `InfrahubEffectivePreferences` / `InfrahubUserPreferences` / `InfrahubGlobalPreferences`. |
| **Response is a stringly `{key,value,source}` list** | Now a **typed** response (`date_format: DateFormat`, per-field `{ value, source }` for effective); the schema is self-describing. |
| **Permission info in a data query** | `can_edit_global_preferences` **removed**; the frontend gates on the generic **`InfrahubPermissions`** query (#9776). |
| **Don't seed a global preference at init** | The init seed is **removed** — a missing row means "nothing set". |
| **Mutation imports from the query module / code in `__init__`** | Shared GraphQL vocab moved to a neutral `graphql/types/preferences.py`; the permission constant to `permissions.py`; `__init__` is re-exports only (+ a `dev/knowledge/backend` guideline). |
| **Track orphaned rows with a ticket** | Filed **IFC-2867** and referenced it from the `Preference` docstring. |
| **`account_id`/`owner_id` plain string — no relationship/cascade** | Inherent to `StandardNode` (no `on_delete: cascade`); orphans are benign (UUIDs never reused) and tracked in **IFC-2867**. |

Backend gate: ruff + ty + gql-schema regen locally; component tests run in CI (Docker unavailable locally). Spec updated (`dev/specs/2026-04-user-preferences.md`).

backend/infrahub/core/preferences/constants.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from infrahub.utils import InfrahubStringEnum
+
+
+class DateFormat(InfrahubStringEnum):
+    """Semantic date-format keys (INFP-512).
```
polmichel: This is not useful to mention Jira items or Spec-kit items inside the code. Let's just describe how the system is behaving and potentially why if it's not obvious or coming from a design descision.

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,97 @@
...
+    They share the same fields; the only difference is the owner, identified by `owner_id` — an
+    account id for a user's preferences, or the Root node id (registry.id) for the organisation-wide
+    (global) preferences. Reads NEVER create a row: a missing row means "nothing set" and the caller
+    falls back (user → global → the client's built-in default).
+
+    `owner_id` is a plain string, not a graph relationship: a StandardNode cannot declare a schema
+    relationship with `on_delete: cascade` (that is a schema-Node feature), so deleting an account
+    leaves its Preference row behind as unreachable dead data. Account ids are UUIDs and never reused,
+    so such a row is permanently unreachable and benign. Cleanup is out of scope for V1 and tracked in
+    Jira (IFC-2867).
```
polmichel: If we take a step back and we read these comments in a few months, is it useful to have all these information?
For instance, I think we can remove the following sentence as it seems to come from previous iterations
`Reads NEVER create a row: a missing row means "nothing set" and the caller falls back (user → global → the client's built-in default).`

I've created the following skills https://github.com/opsmill/opsmill-skills/pull/33 which trims text content (comments, PR description, documentation) from all these iterations residues

backend/infrahub/core/initialization.py
```diff
@@ -554,6 +555,10 @@ async def first_time_initialization(db: InfrahubDatabase) -> None:
     default_branch = await create_default_branch(db=db)
     await create_global_branch(db=db)
 
+    # No global-preference row is seeded: preferences reads never create, and a missing row means
+    # "nothing set" (the client falls back to its default). The row is created lazily on the first
+    # global write. See backend/infrahub/core/preferences/models.py.
```
polmichel: Is this a behavior someone would expect at this point in the code and so deserves a comment?

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,121 @@
...
+
+    scope=USER   → the calling account's OWN Preference row (owner_id = account_session.account_id;
+                   no account argument, so there is no path to write another user's preferences).
+    scope=GLOBAL → the organisation-wide row (owner_id = the Root id), gated on
+                   manage_global_preferences (super admins bypass) checked BEFORE any read.
+
+    EFFECTIVE is not a member of PreferenceWriteScope, so writing the resolved view is unrepresentable
+    (no runtime guard needed). The _UNSET sentinel leaves an omitted field unchanged while an explicit
+    `null` resets it. The row is lazily created on first write under a per-owner lock — the only path
+    that ever creates a Preference row.
```
polmichel: Same comments around this EFFECTIVE part

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,98 @@
...
+def _require_authenticated(graphql_context: GraphqlContext) -> str:
+    """Reject anonymous/unauthenticated sessions (fail-closed) and return the caller's account id.
+
+    Raises:
+        PermissionDeniedError: if there is no authenticated account session.
+
+    """
+    if not graphql_context.account_session or not graphql_context.account_session.authenticated:
+        raise PermissionDeniedError("This operation requires an authenticated account")
+    return graphql_context.account_session.account_id
```
polmichel: Is there something specific we would like to add as authentication layer for this query specifically? Otherwise, could we not use something already existing? 

`infrahub.graphql.mutations.preferences.InfrahubSetPreferences.mutate`
`infrahub.graphql.auth.query_permission_checker.default_branch_checker.DefaultBranchPermissionChecker.supports`

polmichel: Usually we'd rather having methods that do something and return nothing, or method that return something but "do" nothing.
And this method seems to be generic, I think we should put it next to other similar methods inside `infrahub.graphql.initialization.GraphqlContext.active_account_session`

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,98 @@
...
+    client applies its built-in default). Open to any authenticated caller; the global row is read
+    internally, never exposed as raw org values.
+    """
+    graphql_context: GraphqlContext = info.context
+    account_id = _require_authenticated(graphql_context)
+    global_id = global_owner_id()
+
+    # Account row + global row in ONE query; a missing row simply isn't in the map (reads never
+    # create). StandardNode reads carry no branch filter, so this is branch-agnostic.
+    preferences = await Preference.get_for_owners(db=graphql_context.db, owner_ids=[account_id, global_id])
```
polmichel: if the result is a dict (same ids end up collided in the same key block), let's use a set as an input. Faster and prevents duplicate

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -0,0 +1,98 @@
...
+    def resolve(attribute_name: str) -> dict:
+        user_value = getattr(user, attribute_name) if user else None
+        if user_value is not None:
+            return {"value": user_value, "source": PreferenceSource.USER}
+        global_value = getattr(global_, attribute_name) if global_ else None
+        if global_value is not None:
+            return {"value": global_value, "source": PreferenceSource.GLOBAL}
+        return {"value": None, "source": PreferenceSource.DEFAULT}
+
+    return {"date_format": resolve("date_format"), "timezone": resolve("timezone")}
```
polmichel: This block prevents us to correctly analyze typing on these objects I think we should do something to avoid these `getattr()` calls

polmichel: idea `PreferenceResolver` tiny object which takes as input a `UserPreference` and `GlobalPreference`

backend/tests/component/core/test_initialization.py
```diff
@@ -12,6 +14,22 @@ async def test_first_time_initialization(db: InfrahubDatabase, default_branch: B
     assert True
 
 
+async def test_first_time_initialization_does_not_seed_preferences(
```
polmichel, line 16: Either we want to clean the db with a fixture first, or just avoid that others tests impact this one

polmichel (commented):
A mix of old and recent comments, I'll hide those who are outdated.

backend/infrahub/graphql/queries/preferences.py
polmichel: Still valid for `PreferenceWriteScope`

backend/infrahub/core/preferences/models.py
polmichel: tackled in a separate PR https://github.com/opsmill/infrahub/pull/9821

backend/infrahub/core/preferences/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from .permissions import MANAGE_GLOBAL_PREFERENCES_PERMISSION
+
+__all__ = [
+    "DEFAULT_DATE_FORMAT",
+    "MANAGE_GLOBAL_PREFERENCES_PERMISSION",
+    "PREFERENCE_LOCK_NAMESPACE",
+    "DateFormat",
+    "Preference",
+    "global_owner_id",
+]
```
ajtmccarty: I don't think you should import all of these in this `__init__`. this means that anytime one object from this module is imported, all of the dependencies of every object in this module must be imported, so it makes circular import errors more likely.

backend/infrahub/core/preferences/constants.py
ajtmccarty: I think these classes should be covered by 2 one-line docstrings at the most. `DateFormat` seems to pretty clearly cover what this enum is: a list of date format options. same for `DEFAULT_DATE_FORMAT`, it is the default date format. all the other stuff is not the responsibility of this class

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    owner_id: str
+    # Persisted nullable fields must use `Optional[X]` (not `X | None`) until Python 3.14, because of
+    # how StandardNode.guess_field_type works (mirrors Branch).
+    date_format: Optional[str] = None
```
ajtmccarty: I'd say that this should be `date_format: DateFormat` and `from_db` and `to_db` can be overridden to handle the enum [de]serialization. then the `_validate_date_format` can be removed. although I am assuming that the database serialization is the only thing preventing this.

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,97 @@
...
+    @field_validator("date_format")
+    @classmethod
+    def _validate_date_format(cls, value: Optional[str]) -> Optional[str]:
+        # date_format is stored as a plain string (StandardNode has no enum field type) but must be
+        # one of the DateFormat semantic keys — reject anything else so it is never "any string".
+        if value is not None:
+            DateFormat(value)  # raises ValueError for an unknown key
+        return value
+
+    @classmethod
```
ajtmccarty: these should not be on the model. I know that's how we do it in the `Branch` class, but that class is old and wrong. interactions with the database should be controlled by a `PreferenceRepository` class that looks something like this. it should live in `preferences/repository.py`

```
class PreferenceRepository
    def __init(self, db: InfrahubDatabase):
        self.db = db

    async def get_for_owner(self, owner_id: str) -> Preference | None:
        ...

    async def get_for_owners(self, owner_ids: list[str]) -> dict[str, Preference]:
        ...
```

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,97 @@
...
+    async def get_for_owner(cls, db: InfrahubDatabase, owner_id: str) -> Self | None:
+        """Return the Preference owned by `owner_id`, or None. Never creates a row."""
+        query = await PreferenceGetByOwnerQuery.init(db=db, owner_ids=[owner_id], node_type=cls.get_type())
+        await query.execute(db=db)
+
+        result = query.get_result()
+        if not result:
+            return None
+
+        return cls.from_db(result.get_node("n"))
```
ajtmccarty: we shouldn't have `from_db` methods on models, this is also the old wrong way. database [de]serialization is not the model's responsibility. this logic belongs in the `PreferenceGetByOwnerQuery` class. it should be explicit, meaning that it should precisely define exactly the fields to [de]serialize and how to [de]serialize them. the query itself should only return exactly the fields that are required to build instances of the model. `PreferenceGetByOwnerQuery` should have a method on it that is something like `get_owners(self) -> list[Preference]`

backend/infrahub/core/query/preference.py
```diff
@@ -0,0 +1,40 @@
...
+
+    There is at most one row per owner (guaranteed by the per-owner upsert lock).
+    """
+
+    name = "preference_get_by_owner"
+    type = QueryType.READ
+
+    def __init__(self, owner_ids: list[str], node_type: str, **kwargs: Any) -> None:
+        self.owner_ids = owner_ids
+        self.node_type = node_type
```
ajtmccarty: I think you can drop `node_type`, this doesn't seem to do anything that a plain string cannot

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,97 @@
...
+    from infrahub.database import InfrahubDatabase
+
+# Distributed-lock namespace for Preference upserts, keyed on `owner_id`: concurrent upserts for the
+# SAME owner serialise (preventing a duplicate first row or a lost update) while different owners
+# never contend. The global row locks on the Root id, a user's on the account id. Reads are lock-free
+# (they never write).
+PREFERENCE_LOCK_NAMESPACE = "preference"
+
+
+def global_owner_id() -> str:
```
ajtmccarty: does this need to be the UUID of the `Root`? can it just be some sentinel value, like "GLOBAL_OWNER" that would never correspond to an owner's ID? it seems like this unnecessarily links the global_owner_id to the database, but I could be missing something

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,121 @@
...
+        timezone: str | _Unset | None,
+    ) -> Self:
+        # Per-owner distributed lock: the read-create-or-update-save block below is not atomic on its
+        # own, so concurrent first-upserts for the same owner could otherwise each see "no row" and
+        # create a duplicate (and concurrent updates could lose writes). Keyed on owner_id so distinct
+        # owners never contend; the global row locks on the Root id, a user's on the account id. Reads
+        # stay lock-free.
+        async with lock.registry.get(name=owner_id, namespace=PREFERENCE_LOCK_NAMESPACE, local=False):
+            async with graphql_context.db.start_transaction() as db:
+                obj = await Preference.get_for_owner(db=db, owner_id=owner_id)
```
ajtmccarty: `preference` is a better name than `obj` here

backend/infrahub/graphql/queries/preferences.py
polmichel: Resolved in 45fe6faa439d5f68aa75a62467612bb53bd6f9d9

backend/infrahub/graphql/queries/preferences.py
polmichel: Resolved in c2d2f482a44051489e3900030244e00eb5ce63bc

backend/infrahub/graphql/mutations/preferences.py
polmichel: Trimmed in a04b216b8dd70c2aea5ed553d83dd1dab0b34ddd and e68c706d307a9ab97f8960fbbb686b73a429b578

backend/infrahub/core/preferences/models.py
polmichel: Trimmed in a04b216b8dd70c2aea5ed553d83dd1dab0b34ddd and e68c706d307a9ab97f8960fbbb686b73a429b578

backend/infrahub/graphql/queries/preferences.py
polmichel: Resolved in 096d07a50e9bbde5b77e2817043f0419ff614cdb

backend/infrahub/core/preferences/__init__.py
polmichel: Resolved in 65171e0fbde08967218776cf59c68e6889804465

backend/tests/component/core/test_initialization.py
polmichel: Resolved in 4d52191d96a8f3c707da3d57e95d72edbffaa058

backend/infrahub/graphql/queries/preferences.py
polmichel: Resolved in d78082cc995c97212a3af1fdb22fbe4f6ac86492

backend/infrahub/core/preferences/models.py
polmichel: I think I've already tackled this inside 0c1bdffc1

backend/infrahub/core/preferences/models.py
polmichel: Resolved in 366743053d602a0f37fff5b8943ec591441d6531

backend/infrahub/core/preferences/models.py
polmichel: Resolved in 3d5d8b7816cb08450b00a7789aa3e1b6ee19b42d

backend/infrahub/core/query/preference.py
polmichel: Resolved in ae83ec501cc1064c9b7eb7c7d2d9007a29bed8bd

backend/infrahub/graphql/mutations/preferences.py
polmichel: Resolved in f4d184cb049c87f2c41124963e717261af0b4268

backend/infrahub/core/preferences/models.py
```diff
@@ -4,11 +4,9 @@
 
 from typing import TYPE_CHECKING, Optional, Self
 
-from pydantic import field_validator
-
 from infrahub.core import registry
 from infrahub.core.node.standard import StandardNode
-from infrahub.core.preferences.constants import DateFormat
+from infrahub.core.preferences.constants import DateFormat  # noqa: TC001  pydantic resolves the annotation at runtime
```
polmichel: This comment is removed in a later commit

dev/knowledge/backend/permissions.md
```diff
@@ -29,6 +29,7 @@ Gate workflow actions and operational scope. Defined in `GlobalPermissions` enum
 | `MANAGE_ACCOUNTS` | Allows mutations on Account, AccountGroup, AccountRole |
 | `MANAGE_PERMISSIONS` | Allows mutations on permission objects and reading role-permission relationships |
 | `MANAGE_REPOSITORIES` | Allows mutations on Repository and ReadOnlyRepository |
+| `MANAGE_GLOBAL_PREFERENCES` | Allows reading and writing the organisation-wide preference defaults |
```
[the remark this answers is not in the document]
polmichel: Fixed -> _Allows reading and writing the organisation-wide preference defaults; enforced imperatively via `raise_for_permission()` in the preference query/mutation resolvers, not through the kind-based checker pipeline_

backend/infrahub/core/preferences/repository.py
```diff
@@ -0,0 +1,53 @@
...
+
+if TYPE_CHECKING:
+    from infrahub.core.preferences.models import Preference
+    from infrahub.database import InfrahubDatabase
+
+# Distributed-lock namespace for Preference upserts, keyed on `owner_id`: concurrent upserts for the
+# SAME owner serialise (preventing a duplicate first row or a lost update) while different owners
+# never contend. The global row locks on the Root id, a user's on the account id. Reads are lock-free
+# (they never write).
+PREFERENCE_LOCK_NAMESPACE = "preference"
```
ajtmccarty: seems like this belongs elsewhere. it isn't used in here and is not directly affect database access

backend/infrahub/core/preferences/repository.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class PreferenceRepository:
```
ajtmccarty, line 13: that's the good stuff

backend/infrahub/core/query/preference.py
```diff
@@ -0,0 +1,82 @@
...
+    from infrahub.database import InfrahubDatabase
+
+
+def _nullable_str(result: QueryResult, label: str) -> str | None:
+    """Read an optional property column, mapping the stored NULL sentinel to None."""
+    value = result.get_as_str(label)
+    return None if value is None or value == NULL_VALUE else value
+
+
+class PreferenceGetByOwnerQuery(StandardNodeQuery):
```
ajtmccarty: also the good stuff

backend/infrahub/graphql/mutations/preferences.py
```diff
@@ -0,0 +1,112 @@
...
+        graphql_context: GraphqlContext,
+        owner_id: str,
+        actor_id: str,
+        date_format: str | _Unset | None,
+        timezone: str | _Unset | None,
+    ) -> Self:
+        # Per-owner distributed lock: the read-create-or-update-save block below is not atomic on its
+        # own, so concurrent first-upserts for the same owner could otherwise each see "no row" and
+        # create a duplicate (and concurrent updates could lose writes). Keyed on owner_id so distinct
+        # owners never contend. Reads stay lock-free.
```
ajtmccarty: the read appears to be within the lock, which I think is correct, but I wouldn't call it "lock-free"

backend/infrahub/core/preferences/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+    A field is taken from the user layer if set, else the global layer, else DEFAULT with a null
+    value (a missing layer counts as "nothing set"). DEFAULT carries a None value: the backend does
+    not know the clients' built-in defaults, it only reports that neither layer sets the field.
+    """
+
+    user: Preference | None
+    global_: Preference | None
+
+    def resolve_date_format(self) -> ResolvedPreference[DateFormat]:
```
ajtmccarty: this looks like it might be graphql-specific behavior that doesn't belong on the model, but maybe I'm just thinking that b/c it is called `resolve_...`

backend/infrahub/graphql/types/preferences.py
```diff
@@ -0,0 +1,87 @@
...
+    ),
+)
+
+
+class PreferenceWriteScope(StrEnum):
+    """The WRITABLE axes of the preferences store.
+
+    EFFECTIVE is intentionally absent: the resolved view is read-only, so it is unrepresentable as a
+    write target. USER writes the caller's own preferences; GLOBAL writes the organisation-wide ones
+    (gated on manage_global_preferences).
```
ajtmccarty: this is the only part of the docstring that belongs here "USER writes the caller's own preferences; GLOBAL writes the organisation-wide ones"

backend/infrahub/graphql/types/preferences.py
```diff
@@ -0,0 +1,87 @@
...
+    GLOBAL = "global"
+
+
+# StrEnum so the value Graphene hands the resolver compares equal to the member.
+PreferenceWriteScopeType = Enum.from_enum(
+    PreferenceWriteScope,
+    description=(
+        "The writable axes of the preferences store: USER writes the caller's own preferences, "
+        "GLOBAL writes the organisation-wide ones (gated on manage_global_preferences). EFFECTIVE is "
+        "intentionally absent — the resolved view is read-only."
```
ajtmccarty: this is impossible to understand if you are not a robot

backend/tests/unit/core/test_preference_models.py
ajtmccarty, line 1: looks like a unit test to me

backend/tests/component/graphql/mutations/test_preferences.py
```diff
@@ -0,0 +1,300 @@
...
+        db=db,
+        branch=default_branch,
+        account_session=session_first_account,
+        variables={"scope": "USER", "timezone": "UTC"},
+    )
+    assert result.errors is None
+    assert result.data is not None
+    assert result.data["InfrahubSetPreferences"]["timezone"] == "UTC"
+
+    updated = await PreferenceRepository(db=db).get_for_owner(owner_id=first_account.id)
```
ajtmccarty, line 87: does this really test that there's no second object created? I think `get_for_owner` would always only return 1 object, but not completely sure

backend/tests/component/graphql/mutations/test_preferences.py
```diff
@@ -0,0 +1,300 @@
...
+        variables={"scope": "USER", "date_format": "NOT_A_FORMAT"},
+    )
+    assert result.errors is not None
+    # Specifically the DateFormat enum-coercion error for the bad value.
+    messages = " ".join(str(error.message) for error in result.errors)
+    assert "NOT_A_FORMAT" in messages or "DateFormat" in messages, messages
+    assert await PreferenceRepository(db=db).get_for_owner(owner_id=first_account.id) is None
+
+
+async def test_rejects_effective_scope_enum_value(
```
ajtmccarty: this seems like a weird one, but maybe I am missing something

backend/tests/component/graphql/mutations/test_preferences.py
```diff
@@ -0,0 +1,300 @@
...
+    first_account: Node,
+    session_first_account: AccountSession,
+) -> None:
+    result = await run_mutation(
+        db=db,
+        branch=default_branch,
+        account_session=session_first_account,
+        variables={"scope": "GLOBAL", "date_format": "ISO_DATETIME"},
+    )
+    assert result.errors is not None
```
ajtmccarty, line 209: worth checking what the error is?

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -0,0 +1,280 @@
...
+) -> None:
+    result = await run_query(db=db, branch=default_branch, query=EFFECTIVE_QUERY, account_session=session_first_account)
+    assert result.errors is None
+    assert result.data is not None
+    prefs = result.data["InfrahubEffectivePreferences"]
+    # Nothing defined anywhere: value null, source DEFAULT.
+    assert prefs["date_format"] == {"value": None, "source": "DEFAULT"}
+    assert prefs["timezone"] == {"value": None, "source": "DEFAULT"}
+    # A read never fabricates a row.
+    assert await PreferenceRepository(db=db).get_for_owner(owner_id=first_account.id) is None
```
ajtmccarty: I guess it's good to be thorough, but this seems like something we should not need to test

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+async def test_effective_mixed_per_attribute_sources(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: None,
+    first_account: Node,
+    session_first_account: AccountSession,
+) -> None:
+    """Per-attribute resolution: one USER, one GLOBAL, one DEFAULT in a single read."""
```
ajtmccarty: I don't see the DEFAULT in here

backend/tests/unit/core/test_preference_constants.py
ajtmccarty, line 1: I don't think this is necessary. it seems to be testing how an Enum works

changelog/+ifc-2719-user-global-preferences.added.md
```diff
@@ -0,0 +1 @@
+Added user and global preferences as internal `StandardNode` objects (not schema nodes). An owner-keyed `Preference` model holds both layers: a row owned by an account id stores that user's overrides, and a row owned by the Root id stores the organisation-wide defaults (`date_format`, `timezone`). Rows are lazily created on first write; a missing row means nothing is set. The `InfrahubEffectivePreferences` GraphQL query returns the merged effective values for the calling user (user override → global default → the client's built-in default), each with the layer it was resolved from. User preferences are private by construction — there is no generic query, and the custom query and mutation only ever touch the caller's own row.
```
ajtmccarty: does a user care about `StandardNode`s? or the "Root id"? and I think the "private by construction" bit is expected when we say it is per user

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+async def test_effective_global_only_source_global(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: None,
+    first_account: Node,
+    session_first_account: AccountSession,
+) -> None:
+    await Preference(owner_id=registry.id, date_format="ISO_DATETIME", timezone="UTC").create(db=db)
```
ajtmccarty: this should use the repository based `save()` method

backend/tests/component/core/test_initialization.py
```diff
@@ -12,6 +13,21 @@ async def test_first_time_initialization(db: InfrahubDatabase, default_branch: B
...
+    """A fresh install seeds NO preference row.
+
+    Preferences reads never create, and there is no init seed — a Preference row exists only after
+    the first write.
+    """
+    # delete_all_nodes_in_db leaves a truly empty graph, so first_time_initialization runs against
+    # a fresh install and builds its own Root.
+    await first_time_initialization(db=db)
+
+    assert await Preference.get_list(db=db) == []
```
ajtmccarty: this is only used in tests, but maybe it is worth adding a `get_list` method to the repository

ajtmccarty (commented):
I think this looks pretty good. I'm still not sure about using `registry.id` as the identifier for a global preference. also, someone else should check on how permissions are handled in here b/c I am not a permissions expert

backend/infrahub/core/preferences/models.py
polmichel: I agree, I do not think this value is used somewhere, I'll remove it

backend/infrahub/core/preferences/repository.py
polmichel: True that. I'll move it to `constants.py`

backend/infrahub/core/preferences/models.py
polmichel: Indeed, the naming is misleading. This is just internal domain preference resolution, and I don't think it belongs in the GraphQL layer. I'll change the name.

backend/infrahub/graphql/mutations/preferences.py
polmichel: you're right, I'll change the comment

backend/infrahub/graphql/types/preferences.py
polmichel: Something I and my `pruning-residues` skill missed.

backend/infrahub/graphql/types/preferences.py
polmichel: That's true, I'll change it

backend/tests/component/graphql/mutations/test_preferences.py
polmichel: EFFECTIVE is an old value of the enumeration, I'll delete this one.

changelog/+ifc-2719-user-global-preferences.added.md
polmichel: @pa-lem if you can take a look at how you'd want to phrase the feature changelog, I also agree this is kind of complex from a user standpoint

backend/tests/component/core/test_initialization.py
polmichel: I think this is kind of expected to have such method from a `Repository`, I'll add a `get_all()`

dev/knowledge/backend/preferences.md
```diff
@@ -0,0 +1,114 @@
...
+## Data model
+
+A single `Preference` model holds both layers. The layer is identified only by `owner_id`:
+
+| `owner_id` | Layer | Meaning |
+|------------|-------|---------|
+| an account id (UUID) | USER | that account's own overrides |
+| `GLOBAL_OWNER_ID` (`"__global__"`) | GLOBAL | the organisation-wide defaults |
+
+The global owner id is a fixed sentinel string rather than a real node id. Accounts are UUID-keyed,
```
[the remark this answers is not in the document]
polmichel: There are no `CoreGlobalPreference` and `CoreUserPreference` schemas

gmazoyer (commented):
Looks good to me from a permission code perspective.

ajtmccarty (approved):

polmichel (approved):

PR #9559, opened 2026-06-12
test(proposed-change): artifact-reg selection against real closures

ogenstad:
# Why

The artifact-regeneration selection gate decides, per artifact definition, whether a proposed change actually needs to regenerate. Its correctness depends on a dependency closure that the integrator builds at import time from a real repository (transitive Jinja2 includes, the Python package-directory floor, the repo manifest, and watch.files declarations) and stores on the transform node.

Until now that integrator-to-gate seam had no end-to-end coverage:

The component selection test (test_artifact_regen_selection.py) hand-sets the closures and synthesizes the diff, so it proves the gate logic but never that the integrator actually produces those closures or that real canonicalized file paths line up with them.
The closure builders are unit-tested in isolation, and test_closure_failure_isolation.py covers the Jinja2 import path - but nothing exercised the Python closure through import, and nothing proved a watch: block survives import or that a watched file drives regeneration.
The highest-risk gap was the watch import seam: watch is an SDK-config-only field that must be stripped before the node payload. If that exclusion regresses, the unknown key makes schema validation reject the transform and it is silently dropped at import with no error.

## What changed

Test-only change (no production code, no generated files, no schema/migration).

* **TestArtifactRegenE2E** imports a real FileRepo through the integrator so the stored Jinja2 and Python closures are genuinely auto-detected, asserts those closures, then edits real repository files and checks the gate selects only the owning definition: transform-source edit, package-sibling edit (package-directory floor), transitive Jinja2 partial edit, and a README edit that selects nothing.
* **TestWatchConfigImport** guards the watch-exclusion import seam: a watch.files-declared transform must survive import with the watched directory expanded into its stored dependencies and dependencies_complete = True.
* **TestWatchConfigRegen** drives the real gate so a file inside the watched directory selects the definition and an unrelated file does not.

---
## Summary by cubic
Adds an integration harness and end-to-end tests to validate artifact regeneration selection using real repository closures and to prove `watch.files` survives import and drives selection. Progress toward INFP-409.

- **New Features**
  - Added `ArtifactRegenGateHarness` to record `refresh_artifacts` submissions and assert selected definitions.
  - Added `backend/tests/integration/proposed_change/test_artifact_regen_e2e.py` to verify selection on real closures: transform source, package helper, Jinja partial, and unrelated README.
  - Added `backend/tests/integration/proposed_change/test_artifact_regen_watch.py` to assert `watch.files` expands into `dependencies` and selects on watched-file edits.
  - Added fixture repos `artifact-regen-e2e` and `watch-config` with real `.infrahub.yml`, queries, templates, and transforms.
  - Updated component test to include `transforms/foo/__init__.py` in the Python closure for fidelity.

<sup>Written for commit c2be2cd81419570f605f813a7fb1e6ea694219da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/integration/proposed_change/test_artifact_regen_e2e.py
```diff
@@ -0,0 +1,244 @@
...
+            client=client,
+        )
+
+        commit = repo.get_commit_value(branch_name="main")
+        config_file = await repo.get_repository_config(branch_name="main", commit=commit)  # type: ignore[call-overload]
+        assert config_file
+
+        await repo.import_all_graphql_query(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_jinja2_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
+        await repo.import_python_transforms(branch_name="main", commit=commit, config_file=config_file)  # type: ignore[call-overload]
```
ajtmccarty, line 97: what's wrong with the typing for these?

ajtmccarty (approved):

backend/tests/integration/proposed_change/test_artifact_regen_e2e.py
ogenstad: @ajtmccarty, the typing was related to prefect issues for the decorator that Prefect uses. :(

PR #9565, opened 2026-06-13
improve path traversal performance

ajtmccarty:
- run path traversal (point-to-point) query once at each depth until max depth is reached or max paths are found
  - instead of running the point-to-point cypher query at every depth included in the path traversal graphql query (1 to 5, by default), run the query at each max depth ascending until the maximum number of paths is reached. depending on the shape of the graph and the schema, every extra layer of depth can cause an exponential increase in cypher search space. finding the lower-depth paths first, if they exist, allows us to avoid deeper paths that take much more time to find.
  - another PR is in-progress to increase the performance of the reachable nodes / dependencies version of the graphql query, but it would be too much to include in this PR
- exclude IP namespace and inheriting schemas by default
  - schemas that act like hubs within the graph (so 1 object linked to 1,000s of object, like an IP namespace) drastically reduce performance of both versions of the path traversal query b/c they add so many possible paths. excluding them by default will help improve performance. 
  - users can still add the excluded schema(s) back in using `included_kinds` parameter if they want, but they have to affirmatively opt in to including these kinds

---
## Summary by cubic
Runs path traversal depth-by-depth and stops once enough paths are found to cut latency and DB load, while defaulting to exclude IP-namespace kinds and reporting effective exclusions in GraphQL.

- **New Features**
  - Added `PathTraversalExecutor` to query one feasible depth at a time with the remaining `max_paths`; preserves shortest-first order.
  - `PathTraversalCypherRenderer` now supports a `depths` filter and exposes `feasible_depths()`.
  - Default exclusions include Builtin IP namespace implementers (`IpamNamespace`); `included_kinds` lets callers re-include these defaults without overriding explicit `excluded_kinds`.
  - GraphQL: input adds `included_kinds`, result returns concrete `excluded_kinds`, resolver uses the executor; schema and generated TS types updated.

- **Performance**
  - Stops evaluating deeper branches once `max_paths` is filled.
  - Per-depth queries avoid unnecessary planning/execution while keeping results identical to an all-depths query.

<sup>Written for commit da54ee2b05c98aa0d558436f39074c11991ba06a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9565?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graph_traversal/planning/planner.py
ajtmccarty, line 1: `included_kinds` acts like an override for our default excluded_kinds. so it's not an allow-list, just a way to add kinds that are excluded by default (just IP namespace and its inheritors for now) back into the search space

backend/infrahub/graphql/queries/path.py
```diff
@@ -65,6 +65,14 @@ class PathTraversalResultType(ObjectType):
     source = Field(PathNodeType, required=True, description="The start node")
     destination = Field(PathNodeType, required=True, description="The end node")
     count = Field(Int, required=True, description="Total number of paths discovered")
+    excluded_kinds = Field(
```
ajtmccarty, line 68: returned to the user to explicitly show which kinds were excluded due to generic expansion

gmazoyer (approved):

PR #9576, opened 2026-06-15
fix(git): pin sync commit via the repository git default branch

polmichel:
## Why

Feedbacks brought up by Cubic in this PR: https://github.com/opsmill/infrahub/pull/9564.

### Git module related problems

`sync_repository_from_origin` and `bootstrap_local_repository` resolved the commit to pin/import using an **Infrahub** branch name (`infrahub_branch` / `registry.default_branch`) where a **git** branch name is required. `get_commit_value` resolves against local git branches, which are named after the repository's git default branch. The Infrahub branch name only coincides with a local git branch in the common case (git default == Infrahub default, no staging). In the other cases the lookup misses:

- **Broadcast** (`sync_repository_from_origin`): `RefreshGitFetch.commit` is dropped to `None`, so the worker pool can't converge on the pinned SHA and falls back to pulling latest.
- **Bootstrap** (`bootstrap_local_repository`): `get_commit_value` raises `ValueError`, crashing the sync for repos created on the init-failed path.

Affected setups: repositories whose git default branch differs from Infrahub's default, and staging-branch syncs (where the Infrahub branch has no local git branch at all).

#### Fix

Resolve both lookups via `repo.default_branch` — the git branch the local clone actually checks out.

#### Fix

Assert against a real error code.

---
## Summary by cubic
Fixes commit pinning during repository sync and bootstrap by resolving from each repo’s Git default branch. Workers now converge on the correct SHA; initial imports run only for fresh or re-cloned repos.

- **Bug Fixes**
  - In sync, resolve the pinned commit from `repo.default_branch`; the `RefreshGitFetch` broadcast now carries the correct SHA across mismatched defaults and staging (covered by a component test).
  - In bootstrap, seed graph data from the Git default branch only on fresh or reinitialized clones.

- **Refactors**
  - Extract initial-import branch selection into `resolve_initial_import_branch` with unit tests.
  - Tightened a read-only API test to assert `== 200`.

<sup>Written for commit 0d5733b9a5d9940e2e12c995637f3529d0f69c90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/tasks.py
+++ b/backend/infrahub/git/tasks.py
@@ -278,7 +278,7 @@ async def bootstrap_local_repository(
             except RepositoryError as exc:
                 log.info(exc.message)
                 return None
-            default_import_git_branch = registry.default_branch
+            default_import_git_branch = repo.default_branch
 
         if repo.reinitialized:
             default_import_git_branch = repo.default_branch
@@ -326,7 +326,7 @@ async def sync_repository_from_origin(
             infrahub_branch=infrahub_branch,
         )
         try:
-            pinned_commit: str | None = repo.get_commit_value(branch_name=infrahub_branch, remote=False)
+            pinned_commit: str | None = repo.get_commit_value(branch_name=repo.default_branch, remote=False)
         except (ValueError, InvalidGitRepositoryError) as exc:
             log.debug(
                 f"Could not resolve pinned commit for {repository.name.value}, workers will fall back to pull: {exc}"
--- a/backend/infrahub/git/tasks.py
+++ b/backend/infrahub/git/tasks.py
@@ -234,6 +234,19 @@ async def sync_git_repo_with_origin_and_tag_on_failure(
         raise
 
 
+def resolve_initial_import_branch(repo: InfrahubRepository, init_failed: bool) -> str | None:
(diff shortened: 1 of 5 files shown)


backend/infrahub/git/tasks.py
```diff
@@ -234,6 +234,19 @@ async def sync_git_repo_with_origin_and_tag_on_failure(
         raise
 
 
+def resolve_initial_import_branch(repo: InfrahubRepository, init_failed: bool) -> str | None:
```
polmichel, line 237: I guess a dedicated component like `RepositorySyncer` and `RepositoryAdder` would be useful here to manage which branch should be matched in the different use cases inside this git module. But this would be part of a future PR.

ajtmccarty (approved):

PR #9584, opened 2026-06-16
improve reachable nodes performance (shortest paths by default)

ajtmccarty:
Must follow #9565 

# Why

Improve `InfrahubReachableNodes` performance. Using a local 520-device / 1,920-interface dataset for testing, a default "interfaces reachable from a device" query took **~58–90 s**. The cost came entirely from Phase 2 *path enumeration* in which the query finds every path below a maximum length to every discovered target. The deeper the path, the longer it takes to identify. Paths of depth 4 or 5 take exponentially longer than shorter paths, so avoiding them is best if they are not necessary. Discovering the reachable target objects takes almost no time: ~100 ms.

**Goal:** make ReachableNodes faster by default and skip redundant deep paths if not required

## What changed

**Behavioral changes**

- **ReachableNodes returns the shortest path(s) per target by default.** New input `shortest_paths_only: Boolean = true`. When `true`, each reachable target returns only its minimum-depth path(s) via Cypher `ALL SHORTEST`. When `false`, it returns every path to each target within `max_depth` (the previous semantics). This is a deliberate default change — the common "what's reachable" question wants nearest paths, not the full depth-5 fan-out.
- New GraphQL input field on `ReachableNodesInput` (regenerated `schema/schema.graphql` + frontend introspection types).

**Implementation notes**

- **Two-phase split executor.** `ReachableNodesExecutor` now discovers the terminal set once (Phase 1, ~100 ms), then runs Phase 2 separately:
  - **shortest mode (default):** one per-target `UNWIND … ALL SHORTEST` quantified-path-pattern search, bounded to `max_depth` and the plan's schema-legal hop triples. The shortest-path engine avoids the combinatorial fixed-length enumeration entirely.
  - **all-paths mode:** iterative deepening — loop feasible depths shallowest-first with a shrinking `max_paths` budget, stopping once the budget fills (so deep branches aren't evaluated when shallow paths already satisfy the limit).
- **Renderer renamed** `PathTraversalCypherRenderer` → `GraphTraversalCypherRenderer` (it serves both PathTraversal and ReachableNodes), with internal cleanups: `_DepthRenderData` (was `_DepthBranchData` — "branch" collided with Infrahub's git-branch concept), envelope-builder helpers, and shared `_validate` / `_base_params`.
- **`QueryResult.get_as_list_of_type()` now supports scalar lists** (e.g. `collect(target.uuid)` → `list[str]`), dispatching per element: maps are keyword-unpacked (unchanged), scalars are constructed positionally.

## How to review

Suggested order:
1. `backend/infrahub/graph_traversal/_cypher.py` — the new `render_reachable_targets` / `render_paths_to_targets` / `render_shortest_paths_to_targets`, and the rename/cleanup.
2. `backend/infrahub/graph_traversal/reachable.py` — `ReachableNodesExecutor` (shortest vs all-paths split) and the query classes.
3. `backend/infrahub/graphql/queries/reachable.py` — the `shortest_paths_only` input wiring.
4. Tests: `backend/tests/component/graph_traversal/test_reachable_nodes_query.py`.
5. Mechanical/generated: `schema/schema.graphql`, `graphql-env.d.ts`, the `get_as_list_of_type` one-liner.

Extra scrutiny: the **deletion-shadow predicate inside the `ALL SHORTEST` quantified pattern** on user branches — this is validated by `test_shortest_mode_reroutes_when_branch_deletes_shortest_edge` (deletes the depth-1 edge on a branch and asserts shortest reroutes to the depth-3 path through intermediate hops).

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — N/A, graph traversal is an unreleased feature
- [ ] External docs updated — N/A (feature not yet user-documented)
- [x] Internal .md docs updated (n/a — no `dev/knowledge` changes required)

---
## Summary by cubic
Make InfrahubReachableNodes faster by default by returning only the shortest path(s) per target, splitting traversal into two phases with a reusable terminal set, and wiring a `shortest_paths_only` toggle end-to-end (GraphQL and UI).

- **New Features**
  - Default: shortest path(s) per reachable node; GraphQL input `shortest_paths_only` (default true) controls this.
  - Two-phase execution via `ReachableNodesExecutor`: discover terminal uuids once, then enumerate paths. Shortest mode runs per-target ALL SHORTEST; all-paths mode iterates depths shallowest-first with a shrinking `max_paths` budget.
  - Cypher split for TerminalByKinds: `render_reachable_targets`, `render_paths_to_targets`, `render_shortest_paths_to_targets`; TerminalById uses `render`.
  - Shortest-path search honors branch deletion shadow, rerouting when the shallowest edge is deleted on a user branch.
  - Frontend Dependencies Mode: adds a “Shortest paths only” checkbox (checked by default), sends `shortest_paths_only` to the API, renames “Max Results” to “Max Targets,” and updates the results count label to “paths found.”

- **Migration**
  - Behavior change: clients now get shortest paths only. Set `shortest_paths_only: false` to restore the previous “all paths within max_depth” behavior.
  - The Cypher renderer was renamed from `PathTraversalCypherRenderer` to `GraphTraversalCypherRenderer`. Update imports if referenced directly.
  - `QueryResult.get_as_list_of_type()` now supports scalar lists (e.g., `list[str]`).

<sup>Written for commit 43926db6188d2988a3465d1ddb9a6805c1fff367. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graph_traversal/reachable.py
```diff
@@ -20,55 +24,231 @@ class ReachableNodeData:
...
+        )
+        self.add_to_query(rendered.text)
+        self.params.update(rendered.params)
+        self.return_labels = list(rendered.return_labels)
+
+    def get_terminal_uuids(self) -> list[str]:
+        for result in self.get_results():
+            uuids: list[str] = result.get_as_list_of_type("terminal_uuids", return_type=str)
+            return uuids
+        return []
```
gmazoyer, line 79: The for + return on first iteration reads like a bug. `collect(target.uuid)` always emits one row, so if the intent is "take row 0". I'd write it as:

```python
results = list(self.get_results())
return results[0].get_as_list_of_type("terminal_uuids", return_type=str) if results else []
```

or `next(iter(self.get_results()), None)`. Pure style, but it kept tripping me up when I read it.

backend/infrahub/graph_traversal/_cypher.py
```diff
@@ -192,102 +290,185 @@ def render(
...
+            plan=plan, terminal_kinds=_terminal_kinds_for_plan(plan), depths=depths
+        )
+        text = _path_traversal_text(phase_two_inner=self._phase_two_inner(plan=plan, depth_renders=depth_renders))
+        params: dict[str, Any] = {
+            **self._base_params(source_id=source_id, at=at),
+            "max_targets": max_targets,
+            "max_paths": max_paths,
+            "target_id": plan.terminal_predicate.node_id,
+            **self._hop_tuple_params(depth_renders),
+        }
```
gmazoyer, line 322: I might have missed something so following comment might be wrong: `_PATH_TRAVERSAL_ENVELOPE` doesn't reference `max_targets` anywhere. The param is dead in this code path. Neo4j won't complain, but the next person reading this will go hunting for where it's used. So we could drop it from the params dict (and consider whether the function still needs to take `max_targets` at all).

backend/infrahub/graph_traversal/reachable.py
```diff
@@ -20,55 +24,231 @@ class ReachableNodeData:
...
+        )
+        self.add_to_query(rendered.text)
+        self.params.update(rendered.params)
+        self.return_labels = list(rendered.return_labels)
+
+    def get_reachable_nodes(self) -> list[ReachableNodeData]:
+        return _reachable_from_results(self.get_results())
+
+
+class ReachableNodesExecutor:
```
gmazoyer, line 167: I would have expected to see this one in the `executor.py` file.

gmazoyer (approved):

PR #9591, opened 2026-06-16
fix(permissions): unblock proposed change workflow for non-admin roles

### LINKED ISSUE #8755: bug: Default "General Access" role lacks permission to create Proposed Changes
## Description

Users assigned the default **General Access** role cannot create a Proposed Change. The UI returns a permission denied error ("does not have permission to access this view").

This breaks the fundamental contributor workflow: create branch → make changes → create proposed change → merge.

## Root Cause Analysis

In `backend/infrahub/core/initialization.py` (lines 349–424), the **General Access** role is initialized with:

| Permission | Scope | Decision |
|---|---|---|
| `MANAGE_REPOSITORIES` | global | `ALLOW_ALL` |
| `MANAGE_SCHEMA` | global | `ALLOW_ALL` |
| `MERGE_PROPOSED_CHANGE` | global | `ALLOW_ALL` |
| `VIEW` | `*/*` | `ALLOW_ALL` |
| `ANY` | `*/*` | `ALLOW_OTHER` |

The `ANY */* → ALLOW_OTHER` permission grants modifications only in **non-default branches**. However, `CoreProposedChange` is created against the **default branch**, so the permission checker in `object_permission_checker.py` evaluates it with `ALLOW_DEFAULT` — which is not granted.

The role can **merge** proposed changes (`MERGE_PROPOSED_CHANGE`) but cannot **create** them — a contradiction in the intended workflow.

## Steps to Reproduce

1. Set up a default Infrahub instance
2. Create a user account assigned to a group with the "General Access" role
3. Create a branch and make changes
4. Attempt to create a Proposed Change
5. Observe: permission denied error

## Expected Behavior

Users with the General Access role should be able to:
- [x] Create a branch
- [x] Make changes in non-default branches
- [ ] **Create a Proposed Change** ← currently blocked
- [x] Merge a Proposed Change

## Acceptance Criteria

- [ ] Users with the default "General Access" role can create Proposed Changes without permission errors
- [ ] Existing permission behavior for other operations is not affected
- [ ] Default role initialization in `backend/infrahub/core/initialization.py` is updated
- [ ] Unit/integration tests cover Proposed Change creation with the General Access role

## References

- `backend/infrahub/core/initialization.py` — default role definitions
- `backend/infrahub/graphql/auth/query_permission_checker/object_permission_checker.py` — permission enforcement
- `backend/infrahub/graphql/mutations/proposed_change.py` — ProposedChange mutation

iddocohen:
## Summary

- Exempt proposed change comment, thread, review, and lifecycle mutations (create/update/upsert/delete) from `DefaultBranchPermissionChecker` so they fall through to the existing object-level and PC-specific permission checks. `CoreProposedChangeMerge` stays gated since it has no inline guard; `CoreProposedChangeUpdate` is safe because its `state → merged` path has its own inline `MERGE_PROPOSED_CHANGE` check.
- Add object-level permissions (`Core/ProposedChange`, `Core/ChangeComment`, `Core/ChangeThread`, `Core/ThreadComment` with `any`/`ALLOW_ALL`) to the default "General Access" role so users assigned to it can use the PC workflow out of the box. `ALLOW_ALL` is required to keep specificity below the wildcard `*/*/any/ALLOW_OTHER` already on the role; `ALLOW_DEFAULT` would shadow it and strip non-default-branch access.
- Backfill the same permissions onto existing installations via graph migration `m072` (`ArbitraryMigration`, idempotent, no-ops when the role was renamed or removed).

Fixes #8755 (default General Access role couldn't create proposed changes) and the related operator/reviewer case where accounts with `Core/ProposedChange/any/ALLOW_ALL` + global `review_proposed_change` but no `edit_default_branch` were blocked from commenting, threading, approving, closing, or reopening PCs.

## Test plan

- [ ] `uv run pytest backend/tests/component/graphql/auth/query_permission_checker/test_default_branch_checker.py` — 30 tests, including parametrized coverage of all 18 newly-exempt operations and a regression guard that `CoreProposedChangeMerge` still raises `PermissionDeniedError`.
- [ ] `uv run pytest backend/tests/component/core/test_initialization.py` — asserts the four PC permissions land on the General Access role at first-time init.
- [ ] `uv run pytest backend/tests/component/core/migrations/graph/test_072_general_access_role_proposed_change_permissions.py` — three cases: backfill works after deleting the new permissions, idempotent re-run produces identical IDs (no duplicates), missing role returns success.
- [ ] Manual: stand up an instance with the demo `permissions.yml` (`*/*/any/ALLOW_OTHER`, `*/*/view/ALLOW_ALL`, `Core/ProposedChange/any/ALLOW_ALL`, global `review_proposed_change`, no `edit_default_branch`) and verify the operator account can comment, open a thread, approve, close, and reopen a PC, while merge stays denied.

---
## Summary by cubic
Unblocks the proposed change workflow for non-admins by exempting PC mutations from the default-branch gate and adding object-level PC permissions to the default “General Access” role. Also aligns helper names to use “proposed_change” for consistency. Fixes #8755.

- **Bug Fixes**
  - Exempted PC mutations (create/update/upsert/delete for `CoreProposedChange`, `Core/ChangeComment`, `Core/ChangeThread`, `Core/ThreadComment`, plus `CoreProposedChangeReview`) from the default-branch gate so object-level and inline checks apply; `CoreProposedChangeMerge` remains gated.
  - Added object-level permissions (`Core/ProposedChange`, `Core/ChangeComment`, `Core/ChangeThread`, `Core/ThreadComment` with `any/ALLOW_ALL`) to the “General Access” role.
  - Updated `docs/reference/permissions.mdx` to list the new General Access proposed change permissions.

- **Migration**
  - `m072` backfills the above permissions on existing installations; idempotent and typed against `CoreAccountRole`/`CoreObjectPermission`; no manual steps.

<sup>Written for commit 692ea4a3a721079538847d4a9edaeb678fe477e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graphql/auth/query_permission_checker/default_branch_checker.py
```diff
@@ -24,6 +24,24 @@ class DefaultBranchPermissionChecker(GraphQLQueryPermissionCheckerInterface):
...
+        "CoreChangeCommentUpsert",
+        "CoreChangeCommentDelete",
+        "CoreChangeThreadCreate",
+        "CoreChangeThreadUpdate",
+        "CoreChangeThreadUpsert",
+        "CoreChangeThreadDelete",
+        "CoreThreadCommentCreate",
+        "CoreThreadCommentUpdate",
+        "CoreThreadCommentUpsert",
+        "CoreThreadCommentDelete",
```
gmazoyer, line 44: This only exempts `CoreChangeThread*`. The other three thread kinds (`CoreFileThread`, `CoreObjectThread`, `CoreArtifactThread`) are also branch-agnostic, generate full CRUD mutations, and are probably what the UI actually creates for inline review comments.

Rather than enumerating field names, we can detect branch-agnostic kinds via schema lookup. It is a different fix, probably involving more work. Having said that, it would save us from adding a migration.

backend/infrahub/core/initialization.py
```diff
@@ -439,6 +439,26 @@ async def create_default_role(db: InfrahubDatabase) -> CoreAccountRole:
...
+        await permission.new(
+            db=db,
+            name=name,
+            namespace=namespace,
+            action=PermissionAction.ANY.value,
+            decision=PermissionDecision.ALLOW_ALL.value,
+            description=f"Allow a user to manage {label}",
+        )
+        await permission.save(db=db)
+        proposed_change_object_permissions.append(permission)
```
gmazoyer, line 460: `Core:ProposedChange:any:allow_all` plus the Delete/Update exemptions means any General Access user can edit or delete anyone's PCs, comments, and threads, not just their own. We have no instance-ownership concept (covered in our analysis), so kind-wide `allow_all` is the only lever, but this hands de-facto moderation rights to every standard user. I'm fine with it if we decide it explicitly, but it shouldn't ride in as a side effect of the bug fix.

backend/infrahub/core/migrations/graph/m072_general_access_role_proposed_change_permissions.py
```diff
@@ -0,0 +1,98 @@
...
+async def _existing_proposed_change_permissions(db: InfrahubDatabase, role: CoreAccountRole) -> set[tuple[str, str]]:
+    peers = await role.permissions.get_peers(db=db, peer_type=CoreObjectPermission)
+    existing: set[tuple[str, str]] = set()
+    for peer in peers.values():
+        if peer.get_kind() != InfrahubKind.OBJECTPERMISSION:
+            continue
+        action_value = peer.action.value
+        decision_value = peer.decision.value
+        action_str = action_value.value if hasattr(action_value, "value") else action_value
+        decision_int = decision_value.value if hasattr(decision_value, "value") else decision_value
```
gmazoyer, line 40: The `hasattr(action_value, "value")` branch is unnecessary. `peer.action.value` / `peer.decision.value` already return the raw string/int. Drop the defensive branch and follow that convention. The same block is duplicated in both test helpers, worth collapsing into one.

backend/infrahub/core/migrations/graph/m072_general_access_role_proposed_change_permissions.py
```diff
@@ -0,0 +1,98 @@
...
+                f"[green]'{GENERAL_ACCESS_ROLE_NAME}' role already has the expected proposed change permissions, nothing to do[/green]"
+            )
+            return result
+
+        for permission in new_permissions:
+            await role.permissions.add(db=db, data=permission)
+        await role.permissions.save(db=db)
+        console.log(
+            f"[green]Added {len(new_permissions)} proposed change permission(s) to '{GENERAL_ACCESS_ROLE_NAME}' role[/green]"
+        )
```
gmazoyer, line 96: Each permission is saved before being linked, with the role link saved separately at the end. If it dies in between you get orphaned `ObjectPermission` nodes, and on re-run `_existing_proposed_change_permissions` only checks this role's peers so it won't find them and will create duplicates. Building the nodes and adding them to the role before a single `role.permissions.save()`` (like `create_default_role` does) avoids the window.

backend/tests/component/graphql/auth/query_permission_checker/test_default_branch_checker.py
```diff
@@ -176,41 +177,88 @@ async def test_account_without_permission(
...
+        )
+
+        with pytest.raises(PermissionDeniedError, match=r"You are not allowed to change data in the default branch"):
+            await checker.check(
+                db=db,
+                account_session=session,
+                analyzed_query=analyzed_query,
+                query_parameters=gql_params,
+                branch=permissions_helper.default_branch,
+            )
```
gmazoyer, line 264: Good call adding this explicit guard, and the move off create_autospec/MagicMock to the real analyzer is the right direction.

gmazoyer (approved):

PR #9597, opened 2026-06-16
refactor merge branch flow to be more SOLID

ajtmccarty:
# Refactor: branch-merge flow components

## Summary

Breaks up the old `BranchMerger` god-object into small, constructor-injected components wired by a
builder near the flow entry point. The `merge_branch` flow holds the global merge lock for the whole
flow and delegates to an orchestrator composed of focused sub-components, with rollback consolidated to
a single owner. The decomposition itself is behavior-preserving; it also carries a set of review-driven
hardening fixes (see *Fixes & hardening*) that make a few deliberate, contained behavior changes.

## Components added

- **`BranchMergeOrchestrator`** (`core/merge/orchestrator.py`) — top-level driver, run under the
  caller-held global merge lock: pre-checks, graph merge, schema migrations, status transitions, and
  follow-up/event delegation. Takes `schema_manager` + `schema_update_coordinator` and runs the
  post-merge migration step (load post-merge schema → calculate migrations → run via the coordinator)
  where the analyzer already lives. It does not acquire the merge lock itself (the flow does).
- **`GraphMerger`** (`core/merge/graph_merger.py`) — graph-merge half of the removed `BranchMerger`;
  owns only `merge` / `rollback`. `merge()` does not self-roll-back: on failure it raises
  `MergeFailedError` and lets the single caller-side handler reverse the whole merge transaction.
- **`MergeSchemaAnalyzer`** (`core/merge/schema_analyzer.py`) — schema-analysis half of the removed
  `BranchMerger` (candidate schema, validations, migration calculation). Shared by merge, rebase, and
  the GraphQL pre-flight. Takes `schema_manager` by injection rather than reading the global `registry`.
- **`RepositoryMergeDispatcher`** (`core/merge/repository_merge_dispatcher.py`) — builds
  `GitRepositoryMerge` payloads and submits the `GIT_REPOSITORIES_MERGE` workflows (no graph work, so it
  did not belong on `GraphMerger`). Injected into `PostMergeDispatcher`.
- **`PostMergeDispatcher`** (`core/merge/post_merge.py`) — post-MERGED follow-ups (repo merge, IPAM
  reconcile, cancel-PCs, conditional delete-branch, post-process) and merge-event dispatch.
- **`MergeRollbackHandler`** (`core/merge/rollback_handler.py`) — in-process best-effort rollback of a
  failed merge; leaves the write-block key set on a partial rollback so out-of-process recovery takes over.
- **`build_branch_merge_orchestrator`** (`core/merge/builder.py`) — factory that constructs and wires
  all of the above near the flow entry point.

## Components updated

- **`SchemaUpdateCoordinator`** (`core/schema/update_coordinator.py`) — reshaped so only long-lived
  collaborators (`db`, `schema_manager`, `workflow`) are constructor args; per-operation values
  (`branch`, `origin_schema`, `candidate_schema`, `migration_executor`, `context`, `migrations`,
  `update_*`, `manage_rollback`, …) are keyword args to `execute()`. Built once at the entry point and
  stateless across calls. Stays migration-source-agnostic (callers pass `migrations`), so the merge
  flow, rebase flow, `api/schema.py`, and `cli/db.py` all reuse the same `execute()`.
- **`merge_branch` / `rebase_branch`** (`core/branch/tasks.py`) — `merge_branch` acquires the global
  merge lock for the whole flow, loads the branches *under* that lock, then builds the orchestrator via
  the factory and delegates; `rebase_branch` builds a `SchemaUpdateCoordinator` and runs its migration
  step inline (mirroring the orchestrator), and its local `obj` was renamed to `user_branch`.
- **`merge_branch_mutation`** (`graphql/mutations/tasks.py`) — uses `MergeSchemaAnalyzer` for the
  conflict/validation pre-flight.
- **`api/schema.py`, `cli/db.py`** — mechanical move of the per-operation args from the coordinator
  constructor to `execute()`; no behavior change.
- **`core/merge/__init__.py`** — dropped the `BranchMerger` export.
- **Tests** — patch seam repointed to the `GraphMerger` symbol in the builder module; component/lock-test
  constructions updated to the new signatures.

## Components removed

- **`BranchMerger`** (`core/merge/branch_merger.py`) — the god-object that conflated graph merge, schema
  analysis, and repository merge. Split into `GraphMerger`, `MergeSchemaAnalyzer`, and
  `RepositoryMergeDispatcher`; its migration-running logic now lives in the callers via
  `SchemaUpdateCoordinator`.

## Fixes & hardening

Review-driven fixes folded into this PR. Most address latent issues that pre-date the refactor (the
decomposition preserved them); each is a small, deliberate behavior change:

- **Merge lock held for the whole flow** (`merge_branch`, orchestrator) — the lock + a fresh
  under-lock branch load moved to the flow, so the orchestrator and its components never operate on a
  stale pre-lock branch snapshot. Event dispatch now runs under the merge lock.
- **Conflicts raise `ValidationError`, not `MergeFailedError`** (`GraphMerger.merge`) — the conflict
  check moved out of the `MergeFailedError`-wrapping `try`, so unresolved conflicts surface their
  specific message; only the graph merge maps to `MergeFailedError`.
- **Post-merge event dispatch is best-effort** (`PostMergeDispatcher.dispatch_events`) — a failed send
  is logged per-event and no longer fails an already-committed merge.
- **Diff finalization is best-effort** (orchestrator) — `mark_tracking_ids_merged` /
  `freeze_diffs_for_branch` failures are logged and skipped rather than failing a committed merge or
  rolling back correct data. (Known follow-up: a finalization failure can leave stale diff metadata on
  a merged branch that no mechanism — including `infrahub recover` — currently reconciles.)
- **In-process rollback never raises** (`MergeRollbackHandler`) — the final branch-state restore + key
  delete are exception-guarded, so a rollback hiccup returns `False` (holding the write block for
  recovery) instead of masking the original merge error.
- **`has_schema_changes` `SchemaGeneric` filter** — investigated and left as-is (a generic's
  migration-relevant changes already surface via attribute/relationship/inheriting-node diffs); rationale
  documented inline.


backend/infrahub/core/merge/rollback_handler.py
```diff
@@ -0,0 +1,85 @@
...
+        except Exception:
+            # Hold the write protection (leave the key set) rather than reporting a clean rollback on a
+            # partially-restored state, so recovery takes over instead of allowing writes. Never raise:
+            # raising here would mask the original merge error that triggered the rollback.
+            self.log.exception("Branch state restore failed during merge rollback")
+            return False
+
+        self.log.info(f"Merge rollback completed; branch '{branch.name}' returned to OPEN")
+
+        return True
```
polmichel, line 85: I assume we dot not want to reset `schema_changed_at` or `merge_started_at` values for audit and rollback analysis capabilities?

backend/infrahub/core/merge/schema_analyzer.py
```diff
@@ -0,0 +1,122 @@
...
+                name=self.destination_branch.name
+            ).duplicate()
+
+        return self._destination_schema
+
+    @property
+    def initial_source_schema(self) -> SchemaBranch:
+        if self._initial_source_schema:
+            return self._initial_source_schema
+        raise ValueError("_initial_source_schema hasn't been initialized")
```
polmichel, line 59: Is this to prevent anyone from using this attribute directly from external code locations? And is the attribute not loaded at first to avoid any performance issues since the attribute queries the full schema?

backend/infrahub/core/merge/schema_analyzer.py
```diff
@@ -0,0 +1,122 @@
...
+
+        diff_source = initial_source_schema.diff(other=self.source_schema)
+        diff_destination = initial_source_schema.diff(other=self.destination_schema)
+        return diff_source + diff_destination
+
+    async def calculate_migrations(self, target_schema: SchemaBranch) -> list[SchemaUpdateMigrationInfo]:
+        diff_3way = await self.get_3ways_diff_schema()
+        validation = SchemaUpdateValidationResult.init(diff=diff_3way, schema=target_schema)
+        self.migrations = validation.migrations
+        return self.migrations
```
polmichel: is this value stored as an attribute only for testing purpose? can this not be removed and only assert against the return of this method?

backend/infrahub/log.py
```diff
@@ -10,6 +10,10 @@
 if TYPE_CHECKING:
     from structlog.types import Processor
 
+# Any logger that may flow through the application: the structlog logger returned by get_logger, or a
+# stdlib logger / adapter (e.g. a Prefect flow-run logger). Used where a component accepts either.
+InfrahubLogger = logging.Logger | logging.LoggerAdapter[logging.Logger] | structlog.stdlib.BoundLogger
```
polmichel, line 15: Should we not keep only `infrahub.services.protocols.InfrahubLogger` protocol or this typing?

polmichel (approved):
This looks way cleaner! LGTM!

I've posted some questions and an optional comment just to be sure to understand the intent behind some of the changes.

backend/infrahub/core/merge/rollback_handler.py
ajtmccarty: `merge_started_at` is kept so that we know this branch has already tried to merge and failed. `schema_changed_at` is handled within `destination_branch.update_schema_hash()` above, so that timestamp will end up set to "now". I think it would be more correct for this method to accept `pre_merge_schema_changed_at` alongside `pre_merge_branched_from` and reset it correctly. I've added it to the tasks to make sure I come back for it.

backend/infrahub/core/merge/schema_analyzer.py
ajtmccarty: it's only loaded once it's needed b/c it is an expensive operation. and it need to be async, so it can't go in the constructor

backend/infrahub/core/merge/schema_analyzer.py
ajtmccarty: yes you're right about this, it can be removed

backend/infrahub/log.py
ajtmccarty: I find the different loggers that be used confusing. I've added a follow-up to the tasks to look into this

Issue #9568, opened 2026-06-13
bug: merge to non-main default_branch silently dropped on multi-worker pools


lancamat1:
## Component

Git Integration

## Infrahub Version

1.9.6

## Current Behavior

On a deployment with **more than one task worker**, when a git-remote-backed
read-write `CoreRepository` is configured with `default_branch` set to something
other than `main` (e.g. `develop`), merging an Infrahub branch into `main`
succeeds inside Infrahub but the git-side write-back to the remote `develop`
branch is **silently lost on a subset of merges**.

The merge flow's `git push origin develop` fails with:

```
error: src refspec develop does not match any
```

The remote `develop` never advances, yet the Infrahub merge reports success and
**no error is surfaced** to the user.

**Root cause** (confirmed by inspecting the worker clones): in a worker pool,
only the worker that performed the **initial import** of the repository checks
out a *local* `develop` branch. The other workers' clones contain only a local
`main` plus remote-tracking refs (`origin/develop`). This arises because the
selective-sync filter (`INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES`) must *exclude*
`develop` to avoid creating a redundant standalone Infrahub branch — so no
worker except the importer ever materialises a local `develop`.

When the asynchronous `git-repository-merge` flow is dispatched to a worker
whose clone has no local `develop`, the subsequent `git push origin develop`
fails on the missing refspec. Which worker the flow lands on determines success
vs failure, so the failure is **intermittent**.

Observed worker clone state after several merges (2-worker pool):

```
worker-1:  * main    [origin/main: ahead 2, behind 7]     (no local `develop`)
             remote-tracking: origin/develop, origin/main
worker-2:  * develop [origin/develop: ahead 2]            (has local `develop`)
             main
```

## Expected Behavior

Merging an Infrahub branch should reliably write the merge back to the
repository's configured `default_branch` on the remote, regardless of which
worker in the pool executes the merge flow and regardless of whether that worker
previously checked out the default branch. The merge flow should ensure the
destination/default branch exists locally on the executing worker (fetch +
create a local tracking branch) before merging/pushing — or fail loudly rather
than silently dropping the push.

## Steps to Reproduce

1. Deploy Infrahub with **more than one** task worker.
2. Add a git-remote-backed read-write `CoreRepository` and set `default_branch`
   to a non-`main` branch (e.g. `develop`) via `CoreRepositoryCreate`/`Update`.
[...]

Issue #8749, opened 2026-03-31
bug: Artifact generation fails for read-write repositories when default branch is not "main"


dgarros:
## Component

Git Integration

## Infrahub Version

1.8.2

## Current Behavior

When a repository is configured in read-write mode with a default branch other than `main` (e.g., `master`, `develop`), Infrahub fails to generate artifacts from that repository. The repository's operational state transitions to an error.

## Expected Behavior

Artifact generation should work regardless of the repository's default branch name. The system should respect the `default_branch` attribute configured on the repository object.

## Steps to Reproduce

1. Create an external Git repository with a default branch named something other than `main` (e.g., `master`)
2. Add it to Infrahub as a read-write repository, setting `default_branch` to match
3. Configure an artifact definition that pulls data from this repository
4. Trigger artifact generation
5. Observe that the repository operational state becomes an error

## Analysis

Several locations in the codebase have hardcoded assumptions about "main" as the branch name:

- **`backend/infrahub/git/base.py:204`** — `directory_default` property hardcodes `"main"` as the directory name for the primary branch worktree
- **`backend/infrahub/git/base.py:868-870`** — worktree identifier mapping uses `"main"` as a fallback identifier when the repository's default branch differs from the system default
- **`backend/infrahub/git/repository.py:194`** — `rebase()` defaults `source_branch` to `"main"`
- **`backend/infrahub/core/schema/definitions/core/repository.py:44`** — schema `default_value="main"` (this one is expected as a default, but the runtime code should honor the configured value)

The `default_branch` property in `base.py:187-189` correctly falls back from `default_branch_name` to `registry.default_branch`, but downstream code that hardcodes `"main"` as a directory or identifier bypasses this logic.

PR #9601, opened 2026-06-17
fix(git): import non-main default_branch commits in periodic sync

### LINKED ISSUE #9600: bug: periodic git sync ignores non-main default_branch, never imports default-branch updates
### Component

Git Integration

### Infrahub version

1.9.6

### Current Behavior

A read-write `CoreRepository` whose `default_branch` is **not `main`** (e.g. `default_branch=staging`) never imports new commits on its default branch via the periodic sync. The instance stays pinned to its last-imported commit while reporting `sync_status: in-sync`; new commits on the git default branch are silently never applied.

The periodic `git_repositories_sync` flow **fails on every cycle** with:

```
RepositoryError: Unable to get worktree : staging
ValueError: Unable to identify the worktree for the branch : staging
→ Flow run Finished in state Failed(...)
```

### Expected Behavior

The periodic sync should use the repository's configured `default_branch` and import new commits on it into the instance's `main` (advancing `CoreRepository.commit`). It must never silently report `in-sync` while pinned to a stale commit.

### Steps to Reproduce

1. Deploy Infrahub 1.9.6.
2. Add a read-write `CoreRepository` and set `default_branch` to a non-`main` branch (e.g. `staging`) via `CoreRepositoryCreate`.
3. Let it import (commit = C0).
4. Push a new commit to the git `staging` branch.
5. Observe: `git_repositories_sync` fails with `Unable to identify the worktree for the branch : staging`; `CoreRepository.commit` stays at C0; `sync_status` stays `in-sync`. The new commit is never imported.

### Additional Information

**Root cause**

`get_repositories_commit_per_branch` (`backend/infrahub/git/utils.py`, ~line 40) does not request the `default_branch` attribute when it loads repositories for the sync flow:

```python
repos = await NodeManager.query(
    db=db, branch=branch, schema=kind, order=OrderModel(disable=True),
    fields={"id": None, "name": None, "commit": None,
            "internal_status": None, "location": None, "ref": None},
    #                                                   ^^^ "default_branch" is missing
)
```

Because the attribute is not fetched, `repository.default_branch.value` in `sync_remote_repositories` (`backend/infrahub/git/tasks.py`, ~line 262/301) returns the **schema default `"main"`** (`backend/infrahub/core/schema/definitions/core/repository.py`, `default_value="main"`) instead of the configured value. The repository is initialised with `default_branch_name="main"`, so `self.default_branch == "main"` for the whole sync.

That breaks the default-branch → worktree mapping in `pull()` (`backend/infrahub/git/base.py`):

```python
identifier = branch_name                                   # "staging"
if branch_name == self.default_branch and branch_name != registry.default_branch:
    identifier = "main"                                    # ← skipped: self.default_branch is "main"
repo = self.get_git_repo_worktree(identifier="staging")    # ← no worktree keyed "staging" → RepositoryError
```

The default branch's worktree is keyed `main`; with the wrong `default_branch` the code looks up a worktree named after the git branch (`staging`), which does not exist, and the flow raises.

Corroborating evidence (all consistent with `self.default_branch == "main"`):

- `validate_remote_branch` logs `Merging staging into main has no conflicts` — it calls `has_conflicting_changes(target_branch=self.default_branch, source="staging")`, and the target is logged as `main`.
- `git_repositories_sync` flow runs are all `FAILED` with the `worktree : staging` `ValueError`.
- A manual `git pull origin staging` inside the worker's worktree fast-forwards cleanly — git, refspec, and credentials are all healthy; only the in-process `default_branch` is wrong.
- An instance with `default_branch=main` is **unaffected** (the accidental default is correct).

**Fix (verified)**

Add `default_branch` to the fetched fields in `get_repositories_commit_per_branch`:

```python
fields={"id": None, "name": None, "commit": None, "internal_status": None,
        "location": None, "ref": None, "default_branch": None},
```

Verified live on 1.9.6: with this single change, the previously-stuck instance imported a fresh commit on its non-`main` default branch within ~30 s, the `git_repositories_sync` flow returned `Completed`, and the `worktree : staging` error disappeared. Without it, the instance stayed stuck across many cycles with `FAILED` sync flows.

**Scope & relation to other issues**

- Affects only **non-`main` `default_branch`** read-write repos; `default_branch=main` is immune.
- Same non-`main` `default_branch` family as **#8749** (artifact generation) and **#9568** (multi-worker write-back/`push`), but a distinct mechanism: this is the **import/pull** path and the wrong value comes from an under-fetched attribute. **#9499** (merge skipped before sync) and **#9349** (closed; multi-worker divergence) are unrelated code paths.
- OS: macOS arm64. Single task worker (not load- or multi-worker-related).

lancamat1:
## Summary

Read-write `CoreRepository` whose `default_branch` is **not `main`** (e.g. `staging`) never imported new commits on its default branch via the periodic sync. The instance stayed pinned to its last-imported commit while reporting `sync_status: in-sync`, and `git_repositories_sync` failed every cycle with `Unable to identify the worktree : staging`.

## Root cause

`get_repositories_commit_per_branch` (`backend/infrahub/git/utils.py`) did not request the `default_branch` attribute when loading repositories for the sync flow. The attribute fell back to the schema default `"main"`, so `sync_remote_repositories` looked up a worktree keyed by the actual git branch name (`staging`) — which does not exist — and the flow raised. Repositories with `default_branch=main` were unaffected because the accidental default happened to be correct.

## Fix

Add `default_branch` to the fetched fields so the configured value reaches the pull path:

```python
fields={
    "id": None, "name": None, "commit": None, "internal_status": None,
    "location": None, "ref": None, "default_branch": None,
},
```

## Verification

Verified live on 1.9.6: with this change, a previously-stuck instance imported a fresh commit on its non-`main` default branch within ~30s, the `git_repositories_sync` flow returned `Completed`, and the worktree error disappeared. `default_branch=main` behavior is unchanged.

## Scope

Distinct mechanism from the related non-`main` `default_branch` issues #8749 (artifact generation) and #9568 (multi-worker write-back/push) — this is the import/pull path.

Fixes #9600

---
## Summary by cubic
Fixes periodic Git sync for repositories whose `default_branch` is not `main`. We now load `default_branch` during sync so pulls target the configured branch and import new commits (IFC-2757).

- **Bug Fixes**
  - Request `default_branch` in `get_repositories_commit_per_branch` to prevent fallback to `"main"` and worktree errors; enables importing commits on non-`main` default branches.
  - Add a regression test to ensure non-`main` `default_branch` is loaded; verified on 1.9.6, `main` behavior unchanged.

<sup>Written for commit fb8772d54db0c68bfe65bbd1a3ea07f3ecf81880. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/utils.py
+++ b/backend/infrahub/git/utils.py
@@ -40,7 +40,15 @@ async def get_repositories_commit_per_branch(
         repos: list[CoreRepository | CoreReadOnlyRepository] = await NodeManager.query(
             db=db,
             branch=branch,
-            fields={"id": None, "name": None, "commit": None, "internal_status": None, "location": None, "ref": None},
+            fields={
+                "id": None,
+                "name": None,
+                "commit": None,
+                "internal_status": None,
+                "location": None,
+                "ref": None,
+                "default_branch": None,
+            },
             schema=kind,
             order=OrderModel(disable=True),
         )
--- a/backend/tests/component/git/test_utils.py
+++ b/backend/tests/component/git/test_utils.py
@@ -52,6 +52,18 @@ async def test_get_repositories_commit_per_branch_main(
     }
 
 
+async def test_get_repositories_commit_per_branch_non_main_default_branch(
+    db: InfrahubDatabase, register_core_models_schema: SchemaBranch, default_branch: Branch
+) -> None:
+    repo = await Node.init(db=db, schema=InfrahubKind.REPOSITORY, branch=default_branch)
+    await repo.new(db=db, name="repo01", default_branch="staging", commit="commit01", location="location01")
+    await repo.save(db=db)
+
+    repositories = await get_repositories_commit_per_branch(db=db)
+
+    assert repositories["repo01"].repository.default_branch.value == "staging"
+
+
 async def test_get_repositories_commit_per_branch_branches(
     db: InfrahubDatabase, register_core_models_schema: SchemaBranch, repository_01: Node, repository_02: Node
 ) -> None:
--- a/changelog/9600.fixed.md
+++ b/changelog/9600.fixed.md
@@ -1,0 +1,1 @@
+Periodic Git synchronization now imports new commits on a repository's configured default branch when that branch is not `main`. Previously the sync loaded repositories without their `default_branch` attribute, fell back to the schema default of `main`, and failed to locate the worktree, leaving such repositories pinned to a stale commit while still reporting `in-sync`.


lancamat1:
### Note: flaky CI on first run (unrelated to this change)

The first CI run failed on `backend-docker-integration` — `test_artifact_composition.py::TestArtifactComposition::test_add_composite_repo` (`wait_for_sync_to_complete` returned `False` after 12 retries). Re-running the failed job passed clean (36 passed, 0 failed).

This is a flake, not a regression from this PR:

- **Root cause of the flake** is a multi-worker race during repo import. With two task workers, both race to create the same object and one loses the uniqueness check:
  ```
  task-worker-1: CoreGraphQLQueryCreate(name: "person_composite") → "Violates uniqueness constraint 'name'"
  ```
  This is the multi-worker write-back family (cf. #9568), a distinct code path.
- **This PR cannot affect that test.** The fix only changes behavior when `default_branch != "main"`. The failing test uses a repo with `default_branch=main`, for which the resolved value is `"main"` both before (schema-default fallback) and after (fetched attribute) — the exercised path is identical. The change is confined to `get_repositories_commit_per_branch`, which touches neither import object-creation nor worker coordination.

Flagging in case it's worth tracking the `test_add_composite_repo` flake separately.

gmazoyer (commented):
Could you add a regression test? The existing `test_get_repositories_commit_per_branch_main`/`_branches` won't catch this as they `model_dump(exclude=["repository"])` and the fixture sets `default_branch="main"` (== the buggy default). Something like: create a repo with `default_branch="staging"`, then `assert repositories["repo01"].repository.default_branch.value == "staging"`. That fails before this change and passes after.

lancamat1:
@gmazoyer missed this one — added a regression test in `test_get_repositories_commit_per_branch_non_main_default_branch`. Creates a repo with `default_branch="staging"` and asserts `repositories["repo01"].repository.default_branch.value == "staging"`. Fails before the fix (falls back to `main`), passes after.

gmazoyer (approved):

PR #9602, opened 2026-06-17
fix(computed-attribute): refresh derived values after branch merge [IFC-2758]

gmazoyer:
## Summary

A branch merge that applies schema changes updates the registry but emits no `SchemaUpdatedEvent`, so the display label, HFID, and computed-attribute backfills never run on the target branch. Stored derived values on destination nodes that were absent from the merge diff (for example nodes created on the target branch after the source branch forked) stay stale until the node is next mutated.

This emits a `SchemaUpdatedEvent` for the destination branch after a merge that applied schema changes, mirroring the interactive schema-load path, so the existing backfill recomputes the affected values. The event is sent after the merge fully succeeds so an event-send failure cannot trigger a rollback.

## Testing

Adds an integration_docker repro (`test_merge_derived_value_staleness.py`): a node created on the target branch after the source branch forked keeps its old computed value after a template change is merged, while an interactive-path control refreshes correctly. The test was red before the change and green after, verified against a locally built image.

## Out of scope

- The rebase path has the same gap but needs a deeper fix: emitting the event is not sufficient because the backfill deduplicates the rebased branch against the default branch and skips recompute when their schemas match. Tracked for follow-up.
- On large datasets the triggered backfill is currently a full per-kind sweep; scoping it to the changed elements is tracked separately.

---
## Summary by cubic
Emits a SchemaUpdatedEvent after schema-changing branch merges so the target branch recomputes display labels, HFIDs, and computed attributes. Addresses the merge half of IFC-2758; the rebase gap remains open.

- **Bug Fixes**
  - Emit SchemaUpdatedEvent on the destination branch only when schema changes were applied; queued with post-merge events after the merge commits, and send failures never abort the merge.
  - Add integration test backend/tests/integration_docker/test_merge_derived_value_refresh.py that verifies a main-only node refreshes after merging a template change.

- **Refactors**
  - _do_merge_branch now returns MergeBranchResult exposing schema_was_updated; component tests updated to use it.

<sup>Written for commit c3cab135e424f8e7cfa4b573a7fb32b4b449263e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/integration_docker/test_merge_derived_value_refresh.py
+++ b/backend/tests/integration_docker/test_merge_derived_value_refresh.py
@@ -1,0 +1,99 @@
+"""A branch merge that applies a computed-attribute schema change refreshes derived values on main-only nodes."""
+
+from __future__ import annotations
+
+from asyncio import sleep
+from copy import deepcopy
+from typing import TYPE_CHECKING
+
+import pytest
+from infrahub_sdk.testing.docker import TestInfrahubDockerClient
+
+from infrahub.core.constants import ComputedAttributeKind
+from infrahub.core.schema import AttributeSchema
+from infrahub.core.schema.computed_attribute import ComputedAttribute
+from tests.helpers.schema import WIDGET
+
+if TYPE_CHECKING:
+    from infrahub_sdk import InfrahubClient
+
+KIND = WIDGET.kind
+
+
+def _schema(template: str) -> dict:
+    widget = deepcopy(WIDGET)
+    widget.attributes.append(
+        AttributeSchema(
+            name="label",
+            kind="Text",
+            read_only=True,
+            optional=True,
+            computed_attribute=ComputedAttribute(kind=ComputedAttributeKind.JINJA2, jinja2_template=template),
+        )
+    )
+    return {"version": "1.0", "nodes": [widget.model_dump()]}
+
+
+async def _poll_label(client: InfrahubClient, node_id: str, expected: str, max_wait: int = 120) -> str:
+    """Poll up to ``max_wait`` seconds for the node's stored label to reach ``expected``.
+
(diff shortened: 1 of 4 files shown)


backend/tests/integration_docker/test_merge_derived_value_refresh.py
```diff
@@ -0,0 +1,99 @@
...
+        r1 = await client.schema.load(schemas=[_schema("{{ name__value }}")], wait_until_converged=True)
+        assert r1.schema_updated
+        assert await client.schema.in_sync()
+
+        alpha = await client.create(kind=KIND, data={"name": "alpha"})
+        await alpha.save()
+        alpha_initial = await client.get(kind=KIND, id=alpha.id)
+        assert alpha_initial.label.value == "alpha"
+
+        # sanity check: an interactive schema-load on main refreshes the stored value
```
ajtmccarty: what does "interactive" really mean here?

ajtmccarty (approved):

ajtmccarty:
FYI, the merge logic has been significantly refactored in `develop`, so merging this forward is going to cause a conflict. I think it shouldn't be too hard to resolve though

backend/tests/integration_docker/test_merge_derived_value_refresh.py
gmazoyer: It meant "applied without going through a merge" but I rephrased it to tell exactly this.

PR #9603, opened 2026-06-17
perf(IFC-2517): more performance improvements for graph traversal

ajtmccarty:
# Why

The path-traversal GraphQL queries — `InfrahubPathTraversal` (by-id) and `InfrahubReachableNodes` (by-kinds) — were slow on dense graphs and, at high `max_depth`, could run effectively unbounded server-side. A pathological request (deep cone + many paths, or sparse/deep targets) could tie up Neo4j with no abort, and the "shortest paths" mode scaled poorly with the number of targets.

**Goal:** make these queries fast and bounded, and guarantee a runaway query aborts cleanly instead of overloading the database.

**Non-goals:** no GraphQL schema or API-contract changes

### Behavioral changes

- **By-id path traversal** now runs as a single `SHORTEST k` quantified-path-pattern search anchored on both endpoints. It walks Neo4j's frontier and stops once `k` (`max_paths`) paths are found, so shallow targets return without exploring the full `max_depth` cone.
  - this is the query that could bring down the database before these changes. the main problem was that the "target" side of the query was unbounded within the MATCH and only filtered afterwards, so a deep query would have tried to walk a significant portion of the entire graph.
- **Reachable-nodes shortest mode** (the default, `shortest_paths_only: true`) is now resolved by **ascending depth bands** instead of one `ALL SHORTEST` search per discovered target. A target reached at depth *d* is dropped before band *d+1* runs, so every remaining target is resolved in one batched per-band query. Same result set and ordering as before, but cost scales with the number of depth bands, **not** the number of targets.
  - this was slow before because each identified target was basically running its own `MATCH ALL SHORTEST` query. now these are all run together for a single depth.
- **Server-side timeout for all graph-traversal queries.** Both queries now run under a Neo4j transaction timeout (new `INFRAHUB_DB_GRAPH_TRAVERSAL_QUERY_TIMEOUT`, default `75`s). When exceeded, the query is aborted and surfaces a clean error rather than overloading the DB — the GraphQL resolver returns an actionable message ("reduce max_depth, lower max_paths/max_results, or add excluded_kinds/namespaces").

### Implementation notes

- New `DatabaseSettings.graph_traversal_query_timeout`. `timeout_seconds` is threaded `Query.execute → InfrahubDatabase.execute_query[_with_metadata] → run_query`; a Neo4j `TransactionTimedOut` `ClientError` is translated into a new domain `QueryTimeoutError` (HTTP 504). `PathTraversalExecutor` / `ReachableNodesExecutor` receive `timeout_seconds` via constructor injection, wired from config at the resolver.
- The per-target `ALL SHORTEST` reachable implementation was **removed** (`ReachableShortestPathsQuery`, `render_shortest_paths_to_targets`, and its Cypher envelope) — the depth-banded path replaces it. Net deletion in `reachable.py` / `_cypher.py`.
- The reachable all-paths mode (`shortest_paths_only: false`) is unchanged: per-depth fixed-length queries with a shrinking `max_paths` budget that early-stops.

### What stayed the same

- **GraphQL schema / API contract unchanged.** `InfrahubReachableNodes` still exposes `shortest_paths_only` (default `true`); `InfrahubPathTraversal` inputs unchanged.
- **Shortest-mode results and ordering unchanged** — same `(depth, terminal.kind, terminal.uuid)` ordering and same path set; the swap is behavior-preserving (verified against the prior implementation and in component tests).

### Suggested review order

1. `backend/infrahub/graph_traversal/_cypher.py` + `executor.py` — the query strategies (SHORTEST k by-id, depth-banded shortest, all-paths).
2. `backend/infrahub/config.py`, `database/__init__.py`, `core/query/__init__.py`, `exceptions.py` — the timeout plumbing and `QueryTimeoutError`.
3. `backend/infrahub/graphql/queries/{path,reachable}.py` — resolver wiring (inject timeout, translate the error).
4. Tests: `backend/tests/component/graph_traversal/test_reachable_nodes_query.py`, `backend/tests/unit/graph_traversal/test_cypher.py`.

## How to review

Focus areas:

- The depth-banded shortest logic in `ReachableNodesExecutor._run_shortest` (per-band target exclusion) — the correctness-sensitive part of the swap.
- The `TransactionTimedOut → QueryTimeoutError` translation in `database/__init__.py` (error-code matching).

Lower scrutiny: resolver wiring is mechanical; `db_query_profiler.py` only gains the new `timeout_seconds` parameter to keep its override signature compatible (caught by `ty`).

## Testing

Timeout and performance have been tested against a local Infrahub instance. Verified that a long-running query reachable nodes or path traversal query does not crash the database and is canceled after passing the timeout and that the UI displays the error correctly.

## Impact & rollout

- **Backward compatibility:** No schema or API changes. Shortest-mode results/ordering are preserved. No migration.
- **Performance:** all numbers below are warm, single-process runs against a local Neo4j on a dense demo graph — illustrative, not a formal benchmark.

  **By-id path traversal** (`InfraDevice` source, `max_paths 10`). The single `SHORTEST k` search replaces the old fixed-length per-depth enumeration, which expanded the full neighborhood at each depth and only filtered to the target afterwards:

  | scenario (`max_depth` unless noted) | before — fixed-length per-depth | after — `SHORTEST k` |
  |---|--:|--:|
  | target 1 hop away (depth 5) | sub-second | ~0.3 s |
  | target 4–5 hops away, sparse (depth 5) | ~80 s | ~6 s |
  | depth 6, `max_paths 11` (deep, many paths) | unbounded — could exhaust the DB | ~63 s, and bounded by the timeout |

  Shallow targets were already fast; the win is on deep/sparse targets where the old approach walked a large fraction of the graph. `SHORTEST k` early-stops once `k` paths are found, so it only pays for the full `max_depth` cone when the budget genuinely requires the deepest bands (the depth-6 row).

  **Reachable shortest mode** no longer scales with target count (`IpamIPPrefix → IpamIPAddress`, depth 5):

  | targets | shortest — per-target `ALL SHORTEST` (old) | shortest — depth-banded (new) |
  |--------:|-------------------------------------------:|------------------------------:|
  | 10 | 449 ms | 24 ms |
  | 50 | 2442 ms | 51 ms |
  | 150 | 7303 ms | 51 ms |

  The banded shortest is flat in target count and now faster than the all-paths query (as expected, since shortest ⊆ all-paths). A deliberately pathological deep/dense query aborts cleanly at the 75s server-side timeout with the DB remaining responsive.
- **Config/env changes:** New `INFRAHUB_DB_GRAPH_TRAVERSAL_QUERY_TIMEOUT` (default `75`s).

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../dev/guidelines/changelog.md) added (`uv run towncrier create ...`) — existing `graph-path-traversal*.changed.md` fragments cover the feature; confirm whether this perf/timeout tuning warrants its own fragment
- [x] External docs updated (generated `docs/docs/reference/configuration.mdx` for the new setting)

---
## Summary by cubic
Shifted by-id path traversal to a single SHORTEST k search anchored on source and target, added a server-side timeout for traversal queries, and reworked reachable-nodes shortest mode to resolve per-depth bands. Faster results on dense/deep graphs, fail-fast on runaway queries, and a simpler Cypher API.

- **New Features**
  - Path traversal now uses one SHORTEST k search between bound endpoints; returns up to `max_paths`, shortest first, with stable ordering.
  - Added `INFRAHUB_DB_GRAPH_TRAVERSAL_QUERY_TIMEOUT` (`database.graph_traversal_query_timeout`, default 75s). Threaded through `Query.execute()` and the DB layer; timeouts raise `QueryTimeoutError`. GraphQL applies the timeout and surfaces a helpful error. `docker-compose.yml` exposes the setting. Docs regenerated.
  - Reachable-nodes shortest mode is depth-banded: resolves each target at its own minimum depth and returns all tied shortest paths.

- **Refactors**
  - Replaced TerminalById `render()` with `render_shortest_path_by_id()`; removed unused `max_targets`. Callers (`PathTraversalQuery`, executors, GraphQL) and tests updated.
  - `ReachableNodesExecutor` moved to `graph_traversal/executor.py` and now supports timeouts.
  - `ReachableTargetsQuery.get_terminal_uuids()` now uses `get_result()` for the single row.

<sup>Written for commit 58ffe926846c10f512522dd5927cc24e66e9d2ed. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graph_traversal/_cypher.py
```diff
@@ -168,25 +168,21 @@ def hop_tuple_param(self, hop: int) -> str:
...
+# paths between the two bound nodes. ``SHORTEST k`` walks Neo4j's frontier in a stable
+# traversal order and stops once ``k`` paths are found, so shallow targets return
+# without exploring the full ``max_depth`` cone. The outer ``ORDER BY`` gives a stable
+# presentation order (depth, then the intermediate relationship:uuid sequence). Binding
+# both ``source`` and ``target`` lets the planner anchor both ends.
+#
+# Determinism note: the result is fully reproducible whenever the number of connecting
+# paths is <= ``max_paths`` (no tier is cut). When more paths exist than ``max_paths``,
+# which paths fill the final tier is decided by Neo4j's traversal order and is stable for
+# unchanged data and a fixed Neo4j version/plan.
```
polmichel, line 182: Optional comment: Having a message that would warn the user that the `max_path` limit has been reached and that a change inside the data could lead to different results shown in the UI could be useful at some point in a follow-up PR.

backend/infrahub/graphql/queries/path.py
```diff
@@ -298,6 +299,7 @@ async def path_traversal_resolver(
                 branch=graphql_context.branch,
                 default_branch_name=registry.default_branch,
             ),
+            timeout_seconds=config.SETTINGS.database.graph_traversal_query_timeout,
         )
         try:
             path_data_list = await executor.run(
```
polmichel, line 305: I was wondering whether this was the correct location to add the warning message I am talking about in my upper optional comment. But I guess that since this a limitation of the internal engine, it should rather be implemented inside the `infrahub.graph_traversal.executor.PathTraversalExecutor` or lower.

backend/infrahub/graph_traversal/executor.py
```diff
@@ -4,47 +4,172 @@
...
+                branch=self._branch,
+                at=at,
+                renderer=self._renderer,
+                plan=plan,
+                source_id=source_id,
+                terminal_uuids=remaining_targets,
+                max_paths=budget,
+                depths={depth},
+            )
+            await paths_query.execute(db=self._db, timeout_seconds=self._timeout_seconds)
```
polmichel, line 170: In that case, the timeout used for each of these queries has the same value, right? Based on your performance testing, is this something we should worry about: potentially having N (N=max feasible depth) queries executed where the sum of their execution times exceeds the defined timeout?

backend/infrahub/config.py
```diff
@@ -336,6 +336,14 @@ class DatabaseSettings(BaseSettings):
     max_concurrent_queries_delay: float = Field(
         default=0.01, ge=0, description="Delay to add when max_concurrent_queries is reached."
     )
+    graph_traversal_query_timeout: float = Field(
+        default=75,
+        ge=1,
+        description=(
+            "Server-side transaction timeout in seconds for graph-traversal queries "
+            "(path traversal and reachable nodes); the query is aborted once it is exceeded."
```
polmichel, line 344: If my last comment about the same timeout being applied to each of the subqueries inside `_run_shortest` is relevant, maybe a concise additional note about it could be useful in the PR description.

polmichel (approved):
LGTM, some optional comment published: one about an optional message related to the query determinism and one about the timeout which seems to be reused for each subquery.

backend/infrahub/graph_traversal/executor.py
ajtmccarty: the timeout is really just to prevent a single query from hoarding database resources. we don't really care if 5 queries run and they all take `timeout_seconds - 1` as long as they complete. so the sum of the query time does not matter to us.

PR #9506, opened 2026-06-08
fix(git): narrow the repository lock to git working-copy mutations

polmichel:
## Intent

This branch narrows the **repository lock** so it only guards the git working-copy mutations (clone, fetch, worktree creation) and no longer wraps the object import into the graph.

The object import reads file content from a per-commit worktree that is pinned while the lock is held, so it does not need the lock itself. By moving the import outside the critical section, repositories can import concurrently while the on-disk git state stays correctly serialized.

A reusable, mock-free lock-recording test adapter underpins the series, so the "import runs outside the lock" guarantee is asserted directly from the recorded acquire/release boundaries.

Closes [IFC-1566](https://opsmill.atlassian.net/browse/IFC-1566) / https://github.com/opsmill/infrahub/issues/6639

## Intermediate PRs

Each PR below was reviewed and approved before being merged into this branch:

| PR | Title | Approved |
|----|-------|----------|
| [#9439](https://github.com/opsmill/infrahub/pull/9439) | test: reusable lock recording adapter | Yes, merged 2026-06-04 |
| [#9440](https://github.com/opsmill/infrahub/pull/9440) | fix(git): run object import outside the repository lock during sync | Yes, merged 2026-06-08 |
| [#9441](https://github.com/opsmill/infrahub/pull/9441) | fix(git): run object import outside the repository lock when adding a repository | Yes, merged 2026-06-08 |

## Commits made directly on this branch (not yet reviewed)

From cbbecf1a103cbe564e806ebc6bf88a46ff0fc52e to the last one.

**Review-driven fixes**

- **Pin the sync import commit while the repository lock is held.**
- **Isolate per-branch failures while collecting pending imports.**
- **Mark `git_repos_dir` as a `usefixtures` dependency**.

**Readability / cleanup (no behavior change)**

- **Split `sync_remote_repositories`** into `bootstrap_local_repository` and `sync_repository_from_origin` so the per-repository loop reads as two named steps.
- **Refresh the `collect_pending_imports` / `sync` docstrings**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Narrows the repository lock to only clone/fetch/worktree steps and runs imports after the lock is released. Import commits are pinned while the lock is held; init fetches missing commits under the lock and sync isolates per-branch failures so other imports still run, resolving IFC-1566.

- **Bug Fixes**
  - Run imports after releasing the repository lock in both sync and add-repo flows; git state mutations remain serialized.
[...]

PR #9618, opened 2026-06-18
fix(git): serialize repository object import under the repository lock

polmichel:
## Summary

The graph-mutation phase of a repository import was not serialized, so two concurrent imports of the same repository could both create the same globally-unique-named object (transform, query, check, …). The loser hit the `name` uniqueness constraint, the import failed, and the repository was left out of sync.

This splits the import into a lock-free **build** phase (reads the pinned commit worktree) and an **apply** phase (all graph mutations), and serializes the apply phase under the repository lock at every import entry point.

## Two distinct fixes

| Path | State before this PR | Origin | Changelog |
|---|---|---|---|
| Manual object re-import (`import_objects_from_git_repository`) | racy in **1.9.8 (released)** | shipped without the lock since 1.9.8, independent of #9506 | **yes** — `fixed` fragment |
| Add / sync / bootstrap (`RepositoryAdder`, `RepositorySyncer`, `bootstrap_local_repository`) | unreleased regression | #9506 narrowed the lock to git working-copy mutations and moved the import out, dropping serialization of the graph phase (a TOCTOU race) | **no** — #9506 is in no release tag (after `infrahub-v1.9.8` / `infrahub-v1.10.0b0`), so it never shipped |

## What was done

- Split the import into `build_import_plan` (no graph mutation) and `apply_import_plan` (every graph mutation), with GraphQL query-reference resolution moved into apply so build is graph-free.
- Serialize the apply phase under the repository lock at all four entry points (add, sync, bootstrap, manual re-import); the slow build phase stays lock-free.

## Validation

Each path was validated before and after the fix. "Before the fix" was produced by removing the apply lock and rebuilding/relaunching the same environment.

| Path | How it was validated | Before the fix | After the fix |
|---|---|---|---|
| Add / sync / bootstrap (commit 9) | `test_same_repo_references_resolve_on_import` — integration test on the real multi-worker stack (image built from source), run repeatedly | ❌ flaky — 2 of 9 runs fail with `Violates uniqueness constraint 'name'` | ✅ 12/12 runs green (0 uniqueness failures) |
| Manual re-import (commit 10) | Live multi-worker app (API server + 2 Prefect `infrahubasync` workers), burst of concurrent `InfrahubRepositoryProcess` re-imports | ❌ ~23 collisions (69 log lines) on `CoreTransformPythonCreate`; repository left `error-import` | ✅ 0 collisions across 32 concurrent imports; repository `in-sync` |

The add/sync race is hard to trigger from the application on demand: it needs two imports of the same repository to interleave within a brief check-create window, so it surfaces only intermittently (hence a flaky test rather than a reliably reproducible bug). The manual re-import (commit 89629a2c207f8ac84b65c464a4629f0d058181cc) can instead be fired as a deterministic concurrent burst, which is why it was used for the live before/after demonstration.

### DIFF
--- a/backend/infrahub/git/integrator.py
+++ b/backend/infrahub/git/integrator.py
@@ -269,16 +269,21 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
         commit: str,  # noqa: ARG002
         config_file: InfrahubRepositoryConfig,
     ) -> None:
+        local_transforms = await self._build_jinja2_transform_definitions(
+            branch_name=branch_name, config_file=config_file
+        )
+        await self._apply_jinja2_transform_definitions(branch_name=branch_name, local_transforms=local_transforms)
+
+    async def _build_jinja2_transform_definitions(
+        self, branch_name: str, config_file: InfrahubRepositoryConfig
+    ) -> dict[str, InfrahubRepositoryJinja2]:
+        """Build the desired Jinja2 transform definitions from the repository config.
+
+        Performs no graph mutation, so it does not need to be serialized against concurrent imports.
+        """
         log = get_run_logger()
 
         schema = await self.sdk.schema.get(kind=InfrahubKind.TRANSFORMJINJA2, branch=branch_name)
-
-        transforms_in_graph = {
-            transform.name.value: transform
-            for transform in await self.sdk.filters(
-                kind=CoreTransformJinja2, branch=branch_name, repository__ids=[str(self.id)]
-            )
-        }
 
         local_transforms: dict[str, InfrahubRepositoryJinja2] = {}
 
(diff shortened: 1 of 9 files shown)


backend/infrahub/git/integrator.py
```diff
@@ -195,6 +215,19 @@ async def ensure_location_is_defined(self) -> None:
...
+            infrahub_branch_name=infrahub_branch_name, git_branch_name=git_branch_name, commit=commit
+        )
+        await self.apply_import_plan(plan)
+
+    async def build_import_plan(
+        self, infrahub_branch_name: str, git_branch_name: str | None = None, commit: str | None = None
+    ) -> ObjectImportPlan:
+        """Build the desired state of a branch's objects by reading the pinned commit worktree.
+
+        Performs no graph mutation other than recording that a sync is in progress, so the expensive
```
polmichel, line 228: This is mentioning external code, but I feel this is important to note. This code has been split for an implicit reason, which is this one.

backend/infrahub/git/integrator.py
```diff
@@ -203,59 +236,100 @@ async def import_objects_from_files(
...
+            check_definitions=check_definitions,
+            generator_definitions=generator_definitions,
+            artifact_definitions=artifact_definitions,
+        )
+
+    async def apply_import_plan(self, plan: ObjectImportPlan) -> None:
+        """Apply a previously built plan to the graph: schema, queries, transforms, objects, definitions.
+
+        Performs every graph mutation of the import, so it must run serialized against any concurrent
+        import of the same repository.
```
polmichel, line 277: Same thought around this, not really resistant to future refactoring, but it can avoid people mixing these two methods.

backend/infrahub/git/tasks.py
```diff
@@ -765,7 +765,9 @@ async def import_objects_from_git_repository(model: GitRepositoryImportObjects)
         repository_kind=model.repository_kind,
         commit=model.commit,
     )
-    await repo.import_objects_from_files(infrahub_branch_name=model.infrahub_branch_name, commit=model.commit)  # type: ignore[call-overload]
+    plan = await repo.build_import_plan(infrahub_branch_name=model.infrahub_branch_name, commit=model.commit)
+    async with lock.registry.get(name=model.repository_name, namespace="repository"):
+        await repo.apply_import_plan(plan)
```
polmichel, line 780: This path was not guarded by a lock before this PR. This is now fixed.

backend/infrahub/git/integrator.py
```diff
@@ -144,6 +145,25 @@ class TransformPythonInformation(BaseModel):
     """Description of the Transform"""
 
 
+@dataclass
+class ObjectImportPlan:
```
gmazoyer, line 149: I like this 👍

gmazoyer (approved):

PR #9627, opened 2026-06-19
improve path traversal performance with meet in the middle query

ajtmccarty:
# Why

Path traversal between two specific objects (`InfrahubPathTraversal`) rendered a single
`SHORTEST k` quantified-path-pattern search. Reaching a *specific, deep* target forces the
engine to enumerate an exponential number of path-steps, so finding a path between distant
objects — especially when an intermediate kind is excluded, pushing the route deeper — blew
past the `graph_traversal_query_timeout` (75s) and the request failed.

**Goal:** make deep by-id path traversal fast and reliable, and let users search deeper.

**Non-goals:** the `InfrahubReachableNodes` algorithm is intentionally left as-is — warm-cache
measurement showed the meet-in-the-middle rewrite *regresses* its common (shallow) case, since it
already returns closest-first with early termination. No schema changes.

Closes 

## What changed

**Behavioral**
- By-id path traversal now uses a **bidirectional ("meet-in-the-middle") search**. Deep paths
  that previously timed out now return in well under a second; results are unchanged where the
  old query completed.
- **Maximum traversal depth raised 20 → 30** (backend cap, GraphQL inputs, and the UI forms),
  now that deep traversal is cheap. Default depth stays **5**.
- **Branch-isolation fix:** edge visibility now bounds default-branch edges by the queried
  branch's fork time (`branched_from`). A change made on the default branch *after* a branch
  forked no longer leaks into that branch, and a branch-local change/deletion stays local.
- Results are **deterministic and monotonic** in `max_paths` (raising the cap only appends
  paths) — stronger than the old `SHORTEST k`, which relied on Neo4j's traversal order at the
  cap boundary.

**Implementation**
- Replace the single `SHORTEST k` with: a node-bounded BFS frontier expanded inward from each
  anchor to `ceil(max_depth/2)`, intersected to find candidate middle nodes, then shortest paths
  reconstructed tier-by-tier via exact-length **canonical-split joins**. Cost scales with graph
  size (nodes), not path count (exponential).
- The BFS runs as **one chained-`CALL` query per direction** (vs a query per depth) — keeps the
  page cache warm across hops and cuts round-trips.
- Edge and anchor (`IS_PART_OF`) predicates use a branch/time-aware visibility filter mirroring
  `Branch.get_query_filter_path`.
- Anchor resolution selects the **active** same-UUID vertex (latest `IS_PART_OF` by branch/time,
  then status check) so a kind/namespace migration's stale duplicate can't anchor the search.

**What stayed the same**
- `InfrahubReachableNodes` algorithm — unchanged.
- The result/output contract (`PathData` shape, GraphQL response) — unchanged.
- No schema changes, no new config/feature flags.

### Suggested review order
1. `backend/infrahub/graph_traversal/_cypher.py` — the rendered Cypher (`_BRANCH_VISIBLE`, the
   anchor matches, `render_bfs`, `render_canonical_join`).
2. `backend/infrahub/graph_traversal/executor.py` — `PathTraversalExecutor.run` orchestration.
3. `backend/infrahub/graph_traversal/path.py` — the two query classes.
4. Tests + the `MAX_DEPTH` / frontend bump.

## How to review

Focus on the branch/time visibility filter and the canonical-split join in `_cypher.py` — the
correctness-sensitive parts. The regenerated `schema/schema.graphql` and frontend
`generated/types.ts` are mechanical (description-only). The depth bump is a one-line constant plus
GraphQL/UI strings.

## How to test

```bash
# Backend (unit, no DB)
uv run pytest backend/tests/unit/graph_traversal/ -q

# Backend (component + GraphQL, needs Neo4j)
uv run pytest backend/tests/component/graph_traversal/ \
  backend/tests/component/graphql/queries/test_path_traversal.py -q
```

The branch-isolation matrix is covered by four dedicated component tests (add/delete × branch vs
default, before/after fork). Frontend lint via `pnpm exec biome ci .`.

The three Cypher shapes below are what the renderer now emits (rendered against a plan for a
`DcimDevice → … → DcimDevice` chain; `$param` placeholders are bound at execution). Together they
replace the single retired `SHORTEST k` by-id query.

<details>
<summary><b>1. Branch/time edge-visibility filter + active-anchor resolution</b> (shared by every query; the correctness-sensitive part)</summary>

Every edge predicate uses this two-clause filter — an edge is visible if it is on the queried
branch (or global) at `$at`, **or** on the default branch (or global) at the branch's fork time
`$default_time`. Bounding default-branch edges by fork time is the branch-isolation fix:

```cypher
(
    (ri.branch IN $at_branches      AND ri.from <= $at           AND (ri.to IS NULL OR ri.to >= $at))
    OR
    (ri.branch IN $default_branches AND ri.from <= $default_time AND (ri.to IS NULL OR ri.to >= $default_time))
)
AND ri.status = "active"
```

Each anchor (source / destination / BFS seed) resolves to its **active** same-UUID vertex. A
kind/namespace migration can leave several `Node` vertices sharing one UUID, so per candidate we
take its latest `IS_PART_OF` edge *without* pre-filtering on status, keep the vertex only if that
latest edge is `active`, then deterministically pick the most-authoritative survivor (highest
branch, then most recent) so traversal always anchors on the same vertex:

```cypher
MATCH (source:Node {uuid: $source_id})
CALL (source) {
    MATCH (source)-[r:IS_PART_OF]->(:Root)
    WHERE <branch/time filter on r>
    RETURN r AS part_of
    ORDER BY r.branch_level DESC, r.from DESC
    LIMIT 1
}
WITH source, part_of
WHERE part_of.status = "active"
WITH source, part_of
ORDER BY part_of.branch_level DESC, part_of.from DESC
LIMIT 1
WITH source
```

On a forked branch each edge predicate additionally carries a deletion-shadow `NOT EXISTS { … }`
so an edge deleted on the user branch is hidden even though its default-branch original is still
visible (see the full predicate in the BFS query below).
</details>

<details>
<summary><b>2. Bidirectional BFS — one chained-<code>CALL</code> query per direction</b> (forward, <code>max_hops=3</code>, on a forked branch)</summary>

Expands the frontier inward from the anchor: hop 1 seeds a `CALL` from the resolved anchor, each
later hop `UNWIND`s the previous frontier and a carried `visited` list dedupes globally so a node
is recorded only at its first-seen (shortest) depth. Returns a single `frontiers` column whose
`i`-th element is the uuids first reached at depth `i+1`. (`forward` uses `$hop_triples` from the
source; `backward` uses the reversed triples from the destination.)

```cypher
MATCH (anchor:Node {uuid: $anchor_id})
CALL (anchor) {
    MATCH (anchor)-[r:IS_PART_OF]->(:Root)
    WHERE <branch/time filter on r>
    RETURN r AS part_of ORDER BY r.branch_level DESC, r.from DESC LIMIT 1
}
WITH anchor, part_of WHERE part_of.status = "active"
WITH anchor, part_of ORDER BY part_of.branch_level DESC, part_of.from DESC LIMIT 1
WITH anchor

CALL (anchor) {
    MATCH (anchor)-[ri:IS_RELATED]-(rel:Relationship)-[ro:IS_RELATED]-(b:Node)
    WHERE <branch/time filter on ri> AND ri.status = "active"
      AND <branch/time filter on ro> AND ro.status = "active"
      AND [anchor.kind, rel.name, b.kind] IN $hop_triples
      AND b.uuid <> $anchor_id
      AND NOT EXISTS { (anchor)-[del_a:IS_RELATED {status: "deleted", branch: $user_branch}]-(rel)
            WHERE del_a.from > ri.from AND del_a.from <= $at AND (del_a.to IS NULL OR del_a.to >= $at) }
      AND NOT EXISTS { (rel)-[del_b:IS_RELATED {status: "deleted", branch: $user_branch}]-(b)
            WHERE del_b.from > ro.from AND del_b.from <= $at AND (del_b.to IS NULL OR del_b.to >= $at) }
    RETURN collect(DISTINCT b.uuid) AS h1
}
WITH h1, [$anchor_id] + h1 AS visited
CALL (h1, visited) {
    UNWIND h1 AS fid
    MATCH (a:Node {uuid: fid})-[ri:IS_RELATED]-(rel:Relationship)-[ro:IS_RELATED]-(b:Node)
    WHERE <same edge predicates as above, with `a` as the from-node>
      AND [a.kind, rel.name, b.kind] IN $hop_triples
      AND NOT b.uuid IN visited
    RETURN collect(DISTINCT b.uuid) AS h2
}
WITH h1, h2, visited + h2 AS visited
CALL (h2, visited) {
    UNWIND h2 AS fid
    MATCH (a:Node {uuid: fid})-[ri:IS_RELATED]-(rel:Relationship)-[ro:IS_RELATED]-(b:Node)
    WHERE <same edge predicates> AND [a.kind, rel.name, b.kind] IN $hop_triples AND NOT b.uuid IN visited
    RETURN collect(DISTINCT b.uuid) AS h3
}
WITH h1, h2, h3, visited + h3 AS visited
RETURN [h1, h2, h3] AS frontiers
```
</details>

<details>
<summary><b>3. Canonical-split join — reconstruct one depth tier</b> (<code>left_len=2 / right_len=2</code>, plus the depth-1 direct case)</summary>

For tier depth `d` the executor pins the middle nodes (forward-distance `floor(d/2)`,
backward-distance `ceil(d/2)`) into `$tier_middles` and joins two exact-length quantified paths
through them. Both halves walk forward `$legal_triples` and exclude `$source_id`; the right half
ends pinned at `$target_id`. The `ORDER BY depth, <relationship>>uuid sequence>` + `LIMIT
$tier_limit` is what makes the output deterministic and monotonic in `max_paths`:

```cypher
-- (active-anchor resolution for $source_id and $target_id, as in panel 1) …
WITH source, target
MATCH lpath = (source) (
    (a)-[ri:IS_RELATED]-(relx:Relationship)-[ro:IS_RELATED]-(b:Node)
    WHERE <edge predicates on ri, ro> AND [a.kind, relx.name, b.kind] IN $legal_triples AND b.uuid <> $source_id
){2} (mid:Node)
WHERE mid.uuid IN $tier_middles
MATCH rpath = (mid) (
    (c)-[si:IS_RELATED]-(srel:Relationship)-[so:IS_RELATED]-(d:Node)
    WHERE <edge predicates on si, so> AND [c.kind, srel.name, d.kind] IN $legal_triples AND d.uuid <> $source_id
){2} (target)
WITH source, lpath, rpath
WITH source.uuid AS start_node_uuid, source.kind AS start_node_kind, 4 AS depth,
    [i IN range(0, 1) | {relationship_identifier: nodes(lpath)[i*2+1].name, uuid: nodes(lpath)[i*2+2].uuid, kind: nodes(lpath)[i*2+2].kind}]
  + [i IN range(0, 1) | {relationship_identifier: nodes(rpath)[i*2+1].name, uuid: nodes(rpath)[i*2+2].uuid, kind: nodes(rpath)[i*2+2].kind}] AS hops
ORDER BY depth ASC, reduce(ordering_key = "", h IN hops | ordering_key + h.relationship_identifier + ">" + h.uuid) ASC
LIMIT $tier_limit
RETURN start_node_uuid, start_node_kind, hops, depth
```

The depth-1 tier (`left_len=0`) is a single right-half match straight from the source — no middle
set, no left path:

```cypher
-- … WITH source, target
MATCH rpath = (source) (
    (c)-[si:IS_RELATED]-(srel:Relationship)-[so:IS_RELATED]-(d:Node)
    WHERE <edge predicates on si, so> AND [c.kind, srel.name, d.kind] IN $legal_triples AND d.uuid <> $source_id
){1} (target)
WITH source.uuid AS start_node_uuid, source.kind AS start_node_kind, 1 AS depth,
    [i IN range(0, 0) | {relationship_identifier: nodes(rpath)[i*2+1].name, uuid: nodes(rpath)[i*2+2].uuid, kind: nodes(rpath)[i*2+2].kind}] AS hops
ORDER BY depth ASC, reduce(ordering_key = "", h IN hops | ordering_key + h.relationship_identifier + ">" + h.uuid) ASC
LIMIT $tier_limit
RETURN start_node_uuid, start_node_kind, hops, depth
```
</details>

## Impact & rollout

- **Backward compatibility:** GraphQL request/response contract unchanged. Returned paths are
  identical where the old query completed; at the `max_paths` boundary the *set* of returned
  paths may differ (the new search is deterministic and monotonic, the old one was not).
- **Performance:** the headline win. Representative measurements:

  | Scenario | old `SHORTEST k` | new bidirectional |
  |---|---|---|
  | depth-18 leaf-interface pair (fabric dataset) | **75 s → timeout** | **~0.9 s** |
  | depth-10 interface pair (627k-node dataset) | — | 45 ms (`max_depth 10`), 1.7 s (`max_depth 30`) |
  | wide IPAM prefix fan-out | ~43 s (path enumeration) | **<1.5 s** |

  Default `max_depth` stays 5, so the deeper-BFS cost is only paid by callers that explicitly
  request a deep search.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — intentionally omitted; this is part of the still-unreleased
      path-traversal feature and folds into its pending changelog fragments.
- [x] External docs updated (GraphQL `max_depth` descriptions; UI cap)
- [x] Internal .md docs updated (code docstrings; n/a for `dev/` knowledge)


backend/infrahub/graph_traversal/executor.py
```diff
@@ -19,14 +20,20 @@
...
-    rather than overloading the database.
+    """Find paths between two specific nodes with a bidirectional ("meet-in-the-middle") search.
+
+    A node-bounded BFS frontier is expanded inward from each anchor to depth
+    ``ceil(max_depth / 2)``, building a per-node shortest-distance map from each end. Nodes
+    reached from both ends are candidate middles; the shortest connecting distance is the
+    minimum of ``forward + backward`` over them. Paths are then reconstructed tier by tier in
+    ascending depth: for tier ``d`` the canonical split is ``floor(d/2)`` / ``ceil(d/2)`` and
+    the join runs only over the middles at exactly those distances. This avoids the
+    exponential path enumeration of a single deep ``SHORTEST k`` search to a specific target.
```
polmichel, line 31: Nice technique, this should reduce a lot the complexity for long paths cases.

backend/infrahub/graph_traversal/executor.py
```diff
@@ -43,18 +50,101 @@ def __init__(
...
+        )
+        backward_depth_map = await self._bfs(
+            plan=plan,
+            source_id=source_id,
+            target_id=target_id,
+            seed_uuid=target_id,
+            direction="backward",
+            max_hops=half_depth,
+            at=at,
+        )
```
polmichel, line 77: Would it be more efficient to compute exactly the required depth for both directions? Meaning taking `max_depth // 2` for `forward_depth` and `(max_depth + 1) // 2` for `backward_depth` (or vice versa)?

backend/infrahub/graph_traversal/executor.py
```diff
@@ -43,18 +50,101 @@ def __init__(
...
+
+        common_uuids = forward_depth_map.keys() & backward_depth_map.keys()
+        if not common_uuids:
+            return []
+        min_join = min(forward_depth_map[node] + backward_depth_map[node] for node in common_uuids)
+
+        collected: list[PathData] = []
+        for depth in range(min_join, max_depth + 1):
+            if len(collected) >= max_paths:
```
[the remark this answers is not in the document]
polmichel: This case cannot be reached since I think the UI or GraphQL layer prevents it. But always better to make this component safe. That said, this is optional.

backend/infrahub/graph_traversal/_cypher.py
gmazoyer, line 1: This comment might not be correct, as I'm not a cypher expert and don't know much about this feature, it is about the "No loop-prevention predicate is needed" comment.

I don't think this holds for paths longer than the shortest one. Because the path is built in two separate halves, nothing stops the halves from sharing a node, so you can get a result that revisits a node. Of course this comment is invalid if this cannot happen in a real life scenario.

For example: 3 objects sharing one tag T, query P1→P2 with max_depth=4. I believe it returns P1 → T → P3 → T → P2 (T twice) on top of the real P1 → T → P2. Worth a test for this case, and a guard when joining the halves so they can't reuse the same link/node.

Again this comment could be wrong, as it is hard for me, at this time, to make sense of all this without understanding properly the global picture.

backend/infrahub/graph_traversal/executor.py
polmichel: for the first one, to manage the `max_depth = 1` use case, maybe `max(1, max_depth // 2)` would be better to avoid to send 0 inside `self._bfs` method

polmichel (approved):
LGTM, minor comments, one describing an idea that could avoid to query an additional frontier level in some cases.

PR #9631, opened 2026-06-19
feat(IFC-1869): custom prefix length for from-pool IP allocation

saltas888:
## What

Adds an optional **prefix-length override** to the from-pool allocation flow for **both IP address pools and IP prefix pools**. When a user allocates from a pool (e.g. a device's *Primary IP Address*, or a subnet carved from a prefix pool), they can now request a specific prefix length instead of always getting the pool default. The pool's default is surfaced as the input placeholder.

> 🔗 **Try it live:** [`/objects/InfraDevice` in the preview environment](https://infrahub-preview-9631.tailc018d.ts.net/objects/InfraDevice) — open a device, edit its *Primary IP Address*, allocate from a pool, and set a custom prefix length.

## Screenshots
Screenshot 2026-06-25 at 11 10 29
Screenshot 2026-06-25 at 11 10 38

image

## Idempotency & the conflict error (FR-014)

Allocation stays idempotent — the reserved resource never changes by re-allocating:

- Re-selecting a pool that already holds a reservation for the object **reuses** it.
- Supplying an explicit prefix length that **conflicts** with an existing reservation is now **rejected with a clear inline error** (`its prefix length cannot be changed, only /<n> can be used`) instead of being silently ignored. The guard (`validate_reserved_prefix_length`) is shared between `CoreIPAddressPool` and `CoreIPPrefixPool`. **No GraphQL schema change.**

## Scope

- **Both IP address and IP prefix pools.** Address pools take `prefixlen` (the new address's mask) and prefix pools take `size` (the carved-out subnet's prefix length) on their inline `from_pool` input — both fields already exist in the schema, so no GraphQL input change was needed.
- The IPAM create form (which allocates via the dedicated `GetResource` mutation rather than the generic `from_pool` path) threads the prefix length through as well.

## Deferred / alternatives considered (FR-014)

- Releasing the reservation on relationship re-point (too broad a behavior change, idempotency risk) and a read-only reserved-resource lookup endpoint (cleaner, larger — good follow-up).

## Testing

- **Backend**: `test_ipaddress_pool` and `test_prefix_pool` — a conflicting prefix length raises the anchored error; absent/matching stays idempotent.
- **Frontend**: unit tests for the mutation builders (prefix length emitted only when set, mapped to `prefixlen`/`size` per pool kind), `PoolPrefixLengthInput` (incl. placeholder), `updateFormFieldValue`, and `findErrorMessage`.
- **E2E (pytest-playwright)**: custom-prefix-length allocation covered for both an IP address pool and an IP prefix pool — asserting the pool default appears as the placeholder and the typed length is honoured. A shared `select_pool` helper de-duplicates the from-pool flow across the suite.
- **Manual**: verified end-to-end in the live preview (conflict error, zero-effort reuse, fresh custom allocation for both pool kinds).

## Spec

See [`dev/specs/ifc-1869-from-pool-prefix-mask/`](dev/specs/ifc-1869-from-pool-prefix-mask/spec.md) — reconciled to as-built.

Jira: IFC-1869


frontend/app/src/shared/components/form/utils/updateFormFieldValue.ts
```diff
@@ -28,6 +28,13 @@ export const updateAttributeFieldValue = (
   defaultValue?: FormAttributeValue
 ): FormAttributeValue => {
   if (newValue && "from_pool" in newValue) {
+    if (defaultValue?.source?.type === "pool" && defaultValue.source.id === newValue.from_pool.id) {
```
saltas888, line 32: TODO in a followup, avoid repetition and unify those 2 functions

frontend/app/src/shared/components/inputs/pool-select.tsx
```diff
@@ -1,25 +1,42 @@
...
   poolDefaultAllocatedObjectKind,
   poolKind,
   onChange,
   selectedPoolId,
+  value,
 }: PoolSelectProps) {
   const [isOpen, setIsOpen] = React.useState(false);
+  // Shown as the prefix-length placeholder so the user sees what the pool would
+  // allocate by default. The override is only sent when the user types a value.
+  const [defaultPrefixLength, setDefaultPrefixLength] = React.useState<number | null>(null);
```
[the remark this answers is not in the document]
saltas888: Addressed: removed the `useState` entirely. The pool-default placeholder was dropped — the input is now just an empty/optional integer field — so there's no per-pool default state left to initialise on edit/reopen.

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * and owns how its value is validated and persisted. Clearing emits `null` (rather
+ * than `undefined`) so react-hook-form actually writes the empty state.
+ */
+export function PoolPrefixLengthInput({
+  value,
+  placeholder,
+  invalid,
+  onChange,
+}: PoolPrefixLengthInputProps) {
+  const handleChange = (raw: string) => {
```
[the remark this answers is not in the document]
saltas888: Fixed: the input now rejects non-integer strings (only `/^\d+$/` is accepted, so `24.5`/`1e2` are dropped) and the field validation also rejects non-integers ("Prefix length must be a whole number").

frontend/app/src/shared/components/form/utils/updateFormFieldValue.ts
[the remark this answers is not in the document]
saltas888: Not reachable in this architecture, so no change: the prefixlen override is entered on the nested `from_pool.prefixlen` RHF field, not passed through `updateAttributeFieldValue`. The PoolValue from pool selection never carries a prefixlen, so the same-pool guard can't discard one — returning the original (idempotent) allocation here is intentional.

backend/infrahub/core/node/resource_manager/ip_address_pool.py
```diff
@@ -51,6 +51,20 @@ async def get_resource(
                     node = await registry.manager.get_one(db=db, id=address.get("uuid"), branch=branch)
 
                     if node:
+                        # Allocation is idempotent: a reserved address can't change its prefix
```
saltas888: This is handling an edge case, that would require a huge rework. The edge case, is when you are returning to a pool that was allocated before. In this case, Frontend doesn't ahve the information that there's an address that was allocated to this device before. Therefore we show the prefix. 

Instead of making a huge rework, I prefered to give the feedback to the user that's not allowed and limit them to select the only one available

gmazoyer: The UX reasoning makes sense, but note that this guard lives in `get_resource`, which is also the entry point for the `IPAddressPoolGetResource` mutation used by the SDK and generators. Those callers never hit the frontend flow, yet a re-allocation with the same `identifier` and a different `prefix_length` now raises where it used to return the existing allocation.

2 things follow from that: the changelog should call it a behaviour change for that mutation, not only added; `CoreIPPrefixPool` has the identical reserved-resource path (and `IPPrefixPoolGetResource` also threads `prefix_length`), so it should get the same treatment or we should say why the address pool is special. A shared helper would keep both consistent.

changelog/+ifc-1869-from-pool-prefix-length.added.md
```diff
@@ -0,0 +1,3 @@
+When allocating an IP address from a pool (the `from_pool` directive on an IP address pool), you can now request a specific prefix length instead of always using the pool's default. In the UI, selecting an IP address pool shows an optional prefix-length field next to the value input, with the pool default as a placeholder — leave it empty for the default, or enter a value to override it.
```
polmichel: Usually, we try to have shorter changelog content.

The guidelines regarding this part of the process are gathered inside the `dev/guidelines/changelog.md` file.

The message should be short, user-facing, and describe what was fixed or implemented without technical implementation details (one sentence is usually recommended).

One of the reasons behind this is that the content will be automatically added by towncrier to the release notes section (for instance: https://github.com/opsmill/infrahub/releases/tag/infrahub-v1.9.8), and this content can get quite long for a release notes format. The PR description and the commits should capture the intent and enough details if someone wants to dig in from an internal standpoint. From a user standpoint, it may be worth considering a dedicated documentation page, if need be.

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
```diff
@@ -49,6 +50,13 @@ async def get_resource(
                     # TODO add support for branch, if the node is reserved with this id in another branch we should return an error
                     node = await registry.manager.get_one(db=db, id=prefix.get("uuid"), branch=branch)
                     if node:
+                        validate_reserved_prefix_length(
+                            pool_kind="IPPrefixPool",
+                            pool_name=self.name.value,  # type: ignore[attr-defined]
+                            reserved_value=node.prefix.value,  # type: ignore[attr-defined]
```
polmichel: To avoid this kind of comment ignoring the typing, a solution could be to use the `infrahub.core.node.Node.get_attribute` API which will return a `BaseAttribute` object if the attribute exists, or raise a `ValueError` if it does not.

backend/infrahub/core/node/resource_manager/ip_address_pool.py
```diff
@@ -51,6 +52,13 @@ async def get_resource(
                     node = await registry.manager.get_one(db=db, id=address.get("uuid"), branch=branch)
 
                     if node:
+                        validate_reserved_prefix_length(
+                            pool_kind="IPAddressPool",
+                            pool_name=self.name.value,  # type: ignore[attr-defined]
+                            reserved_value=node.address.value,  # type: ignore[attr-defined]
```
polmichel: Same: To avoid this kind of comment ignoring the typing, a solution could be to use the `infrahub.core.node.Node.get_attribute` API which will return a `BaseAttribute` object if the attribute exists, or raise a `ValueError` if it does not.

backend/tests/component/core/resource_manager/test_ipaddress_pool.py
```diff
@@ -82,6 +82,48 @@ async def test_get_next_weighted(
...
+    ip_dataset_prefix_v4: dict,
+) -> None:
+    ns1 = ip_dataset_prefix_v4["ns1"]
+    net145 = ip_dataset_prefix_v4["net145"]
+
+    adress_pool_schema = registry.schema.get_node_schema(name=InfrahubKind.IPADDRESSPOOL, branch=default_branch)
+
+    pool = await CoreIPAddressPool.init(schema=adress_pool_schema, db=db)
+    await pool.new(db=db, name="pool1", resources=[net145], ip_namespace=ns1, default_address_type="IpamIPAddress")
+    await pool.save(db=db)
```
polmichel, line 101: Optional comment and not directly related to your PR: this block of code is duplicated several time in this file and just adds complexity to read this test. There is the same block inside `backend/tests/component/core/resource_manager/test_prefix_pool.py` for `InfrahubKind.IPPREFIXPOOL`.

A solution could be to add two different fixtures to a new file `backend/tests/component/core/resource_manager/conftest.py` which would be `address_pool_factory` and `prefix_pool_factory` and would look something like:

```
    async def _create(name: str, resources: list[Node], ip_namespace: Node) -> CoreIPAddressPool:
        schema = registry.schema.get_node_schema(name=InfrahubKind.IPADDRESSPOOL, branch=default_branch)
        pool = await CoreIPAddressPool.init(schema=schema, db=db)
        await pool.new(
            db=db, name=name, resources=resources, ip_namespace=ip_namespace, default_address_type="IpamIPAddress"
        )
        await pool.save(db=db)
        return pool

    return _create
```

Then, all of these tests would reduce their configuration part to 

```
pool = await address_pool_factory(name="pool1", resources=[net145], ip_namespace=ns1)
```

backend/tests/component/core/resource_manager/test_prefix_pool.py
```diff
@@ -43,6 +44,54 @@ async def test_get_next(
...
+) -> None:
+    ns1 = ip_dataset_prefix_v4["ns1"]
+    net140 = ip_dataset_prefix_v4["net140"]
+    net141 = ip_dataset_prefix_v4["net141"]
+
+    prefix_pool_schema = registry.schema.get_node_schema(name=InfrahubKind.IPPREFIXPOOL, branch=default_branch)
+
+    pool = await CoreIPPrefixPool.init(schema=prefix_pool_schema, db=db)
+    await pool.new(db=db, name="pool1", resources=[net140, net141], ip_namespace=ns1)
+    await pool.save(db=db)
```
polmichel, line 64: Same comment than the upper one, but related to `InfrahubKind.IPPREFIXPOOL`

backend/tests/component/core/resource_manager/test_ipaddress_pool.py
```diff
@@ -82,6 +82,48 @@ async def test_get_next_weighted(
...
+
+    # An explicit prefixlen that conflicts with the existing reservation is rejected
+    # rather than silently ignored.
+    with pytest.raises(
+        ValidationError,
+        match=r"already allocated as .*; its prefix length cannot be changed, only /30 can be used\.$",
+    ):
+        await pool.get_resource(
+            db=db, address_type="IpamIPAddress", identifier="item1", branch=default_branch, prefixlen=28
+        )
```
polmichel: After a discussion with @gmazoyer and @ajtmccarty we've decided to assert against the exact error message rather than regular expressions matching several of those.

One of the reasons in my opinion is that we want a clear description of what we expect from the system, not just verifying that some elements are present in the message. This test loses a bit of resistance to refactoring, but it also avoids having nested error messages become indigestible.

backend/tests/component/core/resource_manager/test_prefix_pool.py
```diff
@@ -43,6 +44,54 @@ async def test_get_next(
...
+        match=r"already allocated as .*; its prefix length cannot be changed, only /17 can be used\.$",
+    ):
+        await pool.get_resource(
+            db=db,
+            prefixlen=18,
+            prefix_type="IpamIPPrefix",
+            member_type="prefix",
+            identifier="item1",
+            branch=default_branch,
+        )
```
polmichel: Same comment regarding the assertions against an error message

changelog/+ifc-1869-from-pool-prefixlen-conflict.changed.md
```diff
@@ -0,0 +1,3 @@
+Allocating from an IP address or IP prefix pool with an explicit prefix length that conflicts with an existing reservation for the same identifier now raises a validation error instead of silently returning the existing allocation. A request with no prefix length, or one matching the reservation, is unchanged (the existing allocation is returned).
+
+This is a behaviour change for callers of `get_resource` beyond the UI — notably the `IPAddressPoolGetResource` / `IPPrefixPoolGetResource` mutations used by the SDK and generators, which thread `prefix_length` into the same path. Re-allocating a reservation with a different `prefix_length` that previously succeeded (returning the existing allocation unchanged) now errors.
```
polmichel: Same comment regarding the changelog. That said, it might make sense to have a long changelog in specific cases, so I guess this will be your call.

polmichel (commented):
The review is related to the backend part of the code. Nothing blocking from a functionality perspective, just a few comments regarding code maintainability.

backend/tests/component/core/resource_manager/test_ipaddress_pool.py
saltas888: Good point. I will take care of it in a follow up PR. I will need some related for improvements

backend/infrahub/core/node/resource_manager/ip_address_pool.py
saltas888: Both points are addressed: (1) the get-resource behaviour change is now called out in a separate `.changed.md` fragment — a conflicting prefix length returns a clear error instead of silently reusing the reservation, which also covers the SDK/generator mutation callers; (2) the same guard is applied to `CoreIPPrefixPool` through a shared helper `validate_reserved_prefix_length` in `resource_manager/reservation.py` (995750ae2), so both pools behave identically.

changelog/+ifc-1869-from-pool-prefix-length.added.md
saltas888: Shortened to a single sentence in 70d25d75e: "Added the ability to request a specific prefix length when allocating an IP address from a pool."

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
saltas888: The guard now uses the `get_attribute` API — `self.get_attribute("name").value` and `node.get_attribute("prefix").value` (995750ae2) — so there is no `# type: ignore` on that path. The remaining `# type: ignore[attr-defined]` lines are pre-existing pool code this PR does not touch, so I left them out of scope.

backend/infrahub/core/node/resource_manager/ip_address_pool.py
saltas888: Same here — the guard uses `self.get_attribute("name").value` and `node.get_attribute("address").value` (995750ae2), no `# type: ignore` on that path. The other `# type: ignore` lines are pre-existing and outside this change.

backend/tests/component/core/resource_manager/test_ipaddress_pool.py
saltas888: Agreed the setup is duplicated, but it spans pre-existing tests across both files and is unrelated to this change. To keep the PR scoped I would rather extract `address_pool_factory`/`prefix_pool_factory` into a `conftest.py` as a focused follow-up. Happy to open that separately.

backend/tests/component/core/resource_manager/test_prefix_pool.py
saltas888: Same as the address-pool thread — I would take the shared `conftest.py` factory extraction as a separate follow-up to keep this PR scoped.

backend/tests/component/core/resource_manager/test_ipaddress_pool.py
saltas888: The conflict test asserts the exact message via `assert str(exc_info.value) == (...)` (no regex / `match=`), so it is already in the exact-equality form you described.

backend/tests/component/core/resource_manager/test_prefix_pool.py
saltas888: Same here — the prefix-pool conflict test uses `assert str(exc_info.value) == (...)`, exact equality rather than a regex match.

changelog/+ifc-1869-from-pool-prefixlen-conflict.changed.md
saltas888: Tightened it in 70d25d75e while keeping the before/after context since it is a behaviour change: "Allocating from an IP address or prefix pool with a conflicting prefix length now returns a clear error instead of silently reusing the existing reservation."

backend/tests/component/core/resource_manager/test_prefix_pool.py
```diff
@@ -43,6 +44,55 @@ async def test_get_next(
...
+    assert str(exc_info.value) == (
+        f"IPPrefixPool: pool1 | This resource is already allocated as {first.get_attribute('prefix').value}; "
+        "its prefix length cannot be changed, only /17 can be used."
+    )
```
polmichel: I think these lines are why `pytest.raises(match="")` semantic exists, but this is really a nitpick comment

polmichel (approved):
Backend code LGTM ! Nice work ! A nitpick comment has been posted regarding `pytest.raises(match="")` usage, feel free to ignore it if you want.

Fyi have not reviewed the front-end part.

backend/tests/component/core/resource_manager/test_prefix_pool.py
saltas888: Actually went with your suggestion in 1a67abe7d — switched to `pytest.raises(ValidationError, match=rf"^{re.escape(expected_error)}$")` in both pool tests. Anchored + escaped keeps the exact-message guarantee while using the idiomatic `match` form (also what `dev/rules/testing-python.md` prescribes for `pytest.raises`).

changelog/+ifc-1869-from-pool-prefix-length.added.md
```diff
@@ -0,0 +1 @@
+Added the ability to request a specific prefix length when allocating from an IP address or IP prefix pool, instead of always using the pool's default.
```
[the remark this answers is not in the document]
saltas888: Not applying — the changelog is accurate. Custom prefix length for IP prefix pools is implemented in this PR (`buildFromPoolPayload` maps to the prefix pool's `size` input, with backend conflict guard in `test_prefix_pool` and an e2e test), with no GraphQL schema change. The 'deferred' wording was stale in the PR description, which has now been corrected.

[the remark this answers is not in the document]
saltas888: Agreed — no change needed. The changelog is correct and the PR description has been updated.

frontend/app/src/entities/resource-manager/api/allocate-resource-from-api.ts
```diff
@@ -6,24 +6,24 @@ import type { BranchContextParams } from "@/shared/api/types";
...
 export interface AllocateResourceFromApiParams extends BranchContextParams {
   poolGetResourceMutationName: string;
-  poolId: string;
-  data: Record<string, any>;
+  /**
+   * The pool's GetResource input, built by the caller: `{ id, prefix_length?, data? }`.
+   * Mirrors createObjectFromApi/updateObjectFromApi, where the api wraps a caller-built
+   * `data` rather than owning individual fields.
+   */
+  data: Record<string, unknown>;
```
saltas888: @bilalabbad curious for your opinion here

frontend/app/.betterer.results
saltas888, line 1: Heads-up on why `.betterer.results` changed in this PR (it's an auto-generated baseline, regenerated with `betterer --update`, not hand-edited):

1. **Merge with develop** regenerated it to resolve a conflict — betterer V2 results files conflict on tsc line-number shifts.
2. **The `RelationshipComboboxList` generic refactor** fixed 2 pre-existing `tsc` errors (the old `NodeCore`/`RelationshipNode` filter mismatches), so the issue count dropped **208 → 206**.

No new suppressed errors were added; the baseline only shrank.

frontend/app/src/entities/resource-manager/api/allocate-resource-from-api.ts
bilalabbad: Not a big fan tbh. `poolId` is mandatory, while `data` was optional customization.

Now the required `poolId` is hidden inside `data`, so the function signature doesn’t clearly show what’s mandatory anymore. It also makes it easier to miss id. For example, `allocateResourceFromApi({ data: {} })` compiles cleanly and then explodes in the backend because id is missing. Previously the type signature should prevent that.

saltas888: Good catch. Introduced a typed `AllocateResourceInput` ({ id: string; prefix_length?; data? }) so the mandatory pool `id` is enforced at the call site again — `allocateResourceFromApi({ data: {} })` no longer compiles — while keeping the allocation fields in one caller-built object rather than ad-hoc params.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/form/type.ts
```diff
@@ -49,7 +55,7 @@ export type AttributeValueFromProfile = {
 
 export type AttributeValueFromPool = {
   source: PoolSource;
-  value: { from_pool: { id: string } };
+  value: { from_pool: { id: string; prefixlen?: number } };
```
bilalabbad: WDYT about uniformizing the wording. `prefixlen` is not something we use clientside. I would suggest either `prefix_length` to match api shape or `prefixLength`

frontend/app/src/entities/ipam/ip-prefixes/ui/ipam-creation-form.tsx
```diff
@@ -98,11 +98,29 @@ function IpamCreationForm(props: IpamCreationFormProps) {
...
+              : null;
+
+          const allocationData: Record<string, unknown> = {
+            id: fieldDataForIpField.source.id,
+            data: getCreateMutationFromFormData(formFieldsWithoutIpField, formData),
+          };
+          // Only send a prefix length when one was entered; otherwise the pool default applies.
+          if (typeof pendingFromPool?.prefixlen === "number") {
+            allocationData.prefix_length = pendingFromPool.prefixlen;
+          }
```
bilalabbad: t feels off to me that we’re still assembling the mutation after calling getCreateMutationFromFormData.

I’d expect that function to return the fully constructed mutation data, so callers don’t need to do additional assembly afterward. Otherwise, this logic will likely be duplicated in other forms outside of IPAM too, for example on Device forms.

wdyt?

frontend/app/src/shared/components/form/type.ts
saltas888: Renamed the client value-model field to `prefixLength` (camelCase). `buildFromPoolPayload` stays the single boundary that maps it to the API names (`prefixlen` for address pools, `size` for prefix pools), so the API-shape naming no longer leaks into client code.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/inputs/pool-select.tsx
```diff
@@ -41,30 +63,88 @@ export function PoolSelect({
...
+  //
+  // Offered for both IP address pools (sets the new address's mask) and IP prefix pools
+  // (sets the size of the carved-out subnet); both map to the allocation's prefix length.
+  const pendingFromPool =
+    value.source?.type === "pool" &&
+    value.value &&
+    typeof value.value === "object" &&
+    "from_pool" in value.value
+      ? value.value.from_pool
+      : null;
```
bilalabbad, line 71: Maybe we can handle this later, but I think the root issue is that current typing of sourcePool isn't precise enough anymore with the introduction of customization.

What I have in mind is to fully update the PoolSource typing. Today, we know there can only be three pool types:

kind: CoreIPAddressPool | CoreIPPrefixPool | CoreNumberPool

With proper type narrowing, we should be able to get the exact expected shape for each one. And instead of hardcoding the customization shape, we could derive it from the GraphQL schema through type generation.

frontend/app/src/entities/ipam/ip-prefixes/ui/ipam-creation-form.tsx
saltas888: Went ahead and extracted it: the GetResource input assembly now lives in a tested `buildAllocateResourceInput` helper in the resource-manager entity, and the form just delegates to it. (Generic forms like Device still allocate through the from-pool create path, so this helper is only the IPAM/dedicated-mutation case — no duplication.)

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/inputs/pool-select.tsx
saltas888: Tackled it now rather than deferring.

frontend/app/src/shared/components/form/utils/make-pool-source.ts
```diff
@@ -0,0 +1,38 @@
...
+  defaultPrefixLength?: number | null;
+}): PoolSource => {
+  if (kind === NUMBER_POOL_KIND) {
+    const source: NumberPoolSource = { type: "pool", id, label, kind };
+    if (fromTemplate) source.fromTemplate = true;
+    return source;
+  }
+
+  // The from-pool list only ever surfaces pool kinds, so a non-number kind is an IP pool.
+  const source: IpPoolSource = { type: "pool", id, label, kind: kind as IpPoolSource["kind"] };
```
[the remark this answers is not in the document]
saltas888: Fixed — removed the blind cast. The IP kind is now chosen explicitly (`kind === IP_PREFIX_POOL ? IP_PREFIX_POOL : IP_ADDRESS_POOL`), so it's type-checked rather than asserted; an unexpected kind falls back to IP address explicitly (and the from-pool list only ever surfaces the three known kinds, so it isn't reachable in practice).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
```diff
@@ -0,0 +1,58 @@
...
+      className={classNames(
+        "h-10 w-16 shrink-0 items-center gap-1 rounded-lg border border-neutral-300 px-2",
+        invalid && inputErrorStyle
+      )}
+      title="Prefix length"
+    >
+      <span className="text-gray-500">/</span>
+      <Input
+        value={value ?? ""}
+        onChange={(event) => handleChange(event.target.value)}
```
bilalabbad: `inputMode` is only a keyboard hint. It does not make the value numeric.

Since this field is a real numeric value between 0 and 128, using `type="number"` would let us set `min`/`max` and read the value with `event.target.valueAsNumber`.

We should still handle `NaN`, because browsers may allow intermediate invalid values.

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
```diff
@@ -0,0 +1,58 @@
+import { Row } from "@/shared/components/container";
+import { Input } from "@/shared/components/ui/input";
+import { inputErrorStyle } from "@/shared/components/ui/style";
+import { classNames } from "@/shared/utils/common";
+
+export interface PoolPrefixLengthInputProps {
+  value: number | null | undefined;
+  invalid?: boolean;
+  /** Pool's default prefix length, shown as a hint of the mask used when left blank. */
+  placeholder?: number | null;
```
bilalabbad: placeholder is of type string

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
```diff
@@ -0,0 +1,58 @@
...
+    onChange(Number(raw));
+  };
+
+  return (
+    <Row
+      className={classNames(
+        "h-10 w-16 shrink-0 items-center gap-1 rounded-lg border border-neutral-300 px-2",
+        invalid && inputErrorStyle
+      )}
+      title="Prefix length"
```
bilalabbad: to make easier to have a consistant styling we have some utils class, you could use the default html input and do instead:
On Row:
```jsx
className={classNames(inputStyle, focusWithinStyle, "w-18", invalid && inputErrorStyle)}
```
then remove all default browser styling on html's input

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
saltas888: Done — switched to `type="number"` with `min`/`max`, reading `event.target.valueAsNumber` and treating `NaN` as cleared (null). Integer/range enforcement stays in the field's validate rule, so non-integers like 24.5 surface an inline error rather than being silently accepted.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
saltas888: Fixed — `placeholder` is now typed `string`; the caller stringifies the pool's default prefix length.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

frontend/app/src/shared/components/form/pool-prefix-length-input.tsx
saltas888: Done — the wrapper now uses `classNames(inputStyle, focusWithinStyle, "w-18", invalid && inputErrorStyle)` over a plain `<input>` with default browser styling removed, instead of the hand-rolled border classes.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

bilalabbad (approved):

BeArchiTek:
### Code review

Found 1 issue:

1. Overriding the prefix length on a **prefix pool** `from_pool` allocation fails with `TypeError: CoreIPPrefixPool.get_resource() got an unexpected keyword argument 'size'` (reproduced manually in the UI). The frontend correctly sends `size` per the published GraphQL contract, but the backend never accepts it: `IPPrefixPoolInput.size` was left behind when 29865d6fa renamed the `get_resource()` parameter to `prefixlen`, and the relationship resolver splats the raw payload into `get_resource(**data_from_pool)` with no field-name translation. Address pools only work because `IPAddressPoolInput.prefixlen` happens to match. The spec deferred prefix pools for exactly this hazard, but a later commit expanded the UI gate to prefix pools without reconciling the backend — so the new `validate_reserved_prefix_length()` guard is unreachable on this path, and the new e2e tests don't catch it (they cover the IPAM screens, which use the separate `IPPrefixPoolGetResource` mutation with `prefix_length`). Fix options: accept `size` as an alias in `CoreIPPrefixPool.get_resource()`, or restore the address-pool-only UI gate; renaming the public GraphQL field would be a breaking change.

   - Frontend sends `size` for prefix pools: https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/frontend/app/src/shared/components/form/utils/mutations/buildFromPoolMutationValue.ts#L16-L23
   - Stale GraphQL contract (`size` missed by the rename): https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/backend/infrahub/graphql/types/attribute.py#L33-L38
   - Payload splatted into `get_resource()` untranslated: https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/backend/infrahub/core/relationship/model.py#L598-L601
   - `get_resource()` accepts `prefixlen`, not `size`: https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/backend/infrahub/core/node/resource_manager/ip_prefix_pool.py#L28-L41
   - UI gate expanded to prefix pools: https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/frontend/app/src/shared/components/inputs/pool-select.tsx#L71-L74
   - Spec/QA checklist still document prefix pools as deferred for this exact reason: https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/dev/specs/ifc-1869-from-pool-prefix-mask/spec.md#L108-L110 and https://github.com/opsmill/infrahub/blob/9d91f44fcf7d285f6626f2061cfb0a6bf719ac45/dev/specs/ifc-1869-from-pool-prefix-mask/qa-checklist.md#L12-L15

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

saltas888:
Fixed in f4ba7d8a5 (option 1 from https://github.com/opsmill/infrahub/pull/9631#issuecomment-5070867424): `CoreIPPrefixPool.get_resource()` now accepts `size` as an alias mapped onto `prefixlen` before the reservation guard, with a component regression test covering allocation via `size`, idempotent reuse, and the conflict-guard path. The proper rename of the public field is tracked in [IFC-2945](https://opsmill.atlassian.net/browse/IFC-2945) and flagged with a TODO at the alias. Spec/QA docs updated to include prefix pools.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[IFC-2945]: https://opsmill.atlassian.net/browse/IFC-2945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9642, opened 2026-06-19
fix(proposed-change): match diff action case-insensitively for artifact regeneration

### LINKED ISSUE #9641: Artifact regeneration: query-change and definition-change triggers never fire (diff action case mismatch, INFP-409)
## Summary

The scoped artifact-regeneration feature (INFP-409 / IFC-2541, shipped in 1.10 via #9414) silently fails to regenerate artifacts on two of its three trigger types. A **GraphQL-query-only change** and an **artifact-definition-node change** never trigger regeneration, because the predicates compare the diff `action` against the wrong case.

This is an **under-regeneration / correctness** bug: a proposed change can merge with stale artifacts.

## Root cause

`backend/infrahub/proposed_change/tasks.py`:

```python
_TRIGGERING_DIFF_ACTIONS = {DiffAction.ADDED.value, DiffAction.UPDATED.value}  # == {"added", "updated"}
```

But `client.get_diff_summary()` serialises each entry's `action` as the **uppercase** GraphQL enum name (`"UPDATED"`, `"ADDED"`, ...). So:

```python
entry["action"] in _TRIGGERING_DIFF_ACTIONS   # "UPDATED" in {"added","updated"} -> always False
```

`_query_changed` (the query-node predicate) and `_definition_changed` (the definition-node predicate, which also uses the same constant for its changed-field detail) therefore never match a real diff.

The third trigger, `_transform_changed`, is unaffected — it intersects file paths, not diff actions. The data-change path (`_relevant_node_changes`) is also unaffected — it keys on node `kind` and field names and never inspects `action`. So the blast radius is exactly these two predicates.

## Reproduction (live, against 1.10)

1. Connect a repository with an artifact definition whose GraphQL query lives **outside** its transform's dependency closure (e.g. a Jinja2 transform whose query file is not the template). In the demo-edge repo: `Startup Config for Edge devices` / query `device_startup_info` / `templates/device_startup_info.gql`.
2. On a `sync_with_git` branch, edit that `.gql` file and commit; let the repo sync.
3. Open a proposed change. **Expected:** the definition's artifacts regenerate. **Actual:** 0 artifact validators created; nothing regenerates.

Minimal predicate-level proof:

```python
from infrahub.proposed_change.tasks import _query_changed
entry = {"branch": "b", "kind": "CoreGraphQLQuery", "id": QID, "action": "UPDATED", "elements": []}
defn  = ...  # query_id == QID
_query_changed(definition=defn, diff_summary=[entry]).matched   # -> False (bug)
# lowercasing the action to "updated" -> True
```

## Why CI is green

Both `backend/tests/unit/proposed_change/test_predicates.py` and the `make_node_diff` fixture in `backend/tests/component/proposed_change/conftest.py` hardcode `action="updated"` (lowercase), which matches the buggy lowercase constant. The fixtures diverge from production casing, so every test passes despite the defect.

## Field impact / masking

The bug is masked when the changed query file happens to sit **inside** a transform's directory closure (Python transforms store `transforms/*.gql` alongside the transform, so `_transform_changed` regenerates anyway). It is **exposed** for:
- Jinja2-transform queries (query file outside the template closure),
- any query stored outside its transform's directory,
- any pure artifact-definition-node change (re-targeting the group, repointing the transform/query).

## Fix

Compare case-insensitively, and correct the test fixtures to emit the real uppercase action so the regression is actually caught. PR linked below.

Found during QA of IFC-2541.

PhillSimonds:
## What

Fixes a correctness bug in the INFP-409 scoped artifact-regeneration feature (#9641): a proposed change that only modified a **GraphQL query** or an **artifact definition node** never regenerated the affected artifacts, so they could merge stale.

Closes #9641.

## Root cause

`_TRIGGERING_DIFF_ACTIONS = {DiffAction.ADDED.value, DiffAction.UPDATED.value}` is `{"added", "updated"}` (lowercase), but `client.get_diff_summary()` emits each entry's `action` as the uppercase GraphQL enum name (`"UPDATED"`/`"ADDED"`). So `entry["action"] in _TRIGGERING_DIFF_ACTIONS` is always `False`, and both `_query_changed` and `_definition_changed` silently never match.

`_transform_changed` (file-path intersection) and the data-change path (`_relevant_node_changes`, keyed on kind/fields) never inspect `action`, so they're unaffected — the blast radius is exactly these two predicates.

## Change

- Add a case-insensitive `_is_triggering_action()` helper and use it at all three comparison sites in `tasks.py`.
- Correct the unit (`test_predicates.py`) and component (`conftest.py make_node_diff`, `test_artifact_regen_edge_cases.py`) fixtures to emit the real **uppercase** action. These fixtures previously hardcoded lowercase, matching the buggy constant — which is why CI stayed green.

## Test evidence

With the fixtures corrected to production casing, the predicate unit tests **fail before** the fix and **pass after**:

```
# fixtures uppercased, fix reverted:
6 failed, 21 passed
  test_query_changed[updated_query_is_true] / [added_query_is_true] / [query_match_among_other_entries]
  test_definition_changed[updated_definition_is_true] / [added_definition_is_true] / [definition_match_among_other_entries]
# with the fix:
27 passed
```

Also reproduced and confirmed fixed end-to-end during QA: editing a Jinja2 transform's query file (`templates/device_startup_info.gql`, outside the template closure) on a proposed change produced 0 artifact validators before the fix.

## Notes for review

- Found during QA of IFC-2541; the epic is being bounced back from "Ready for Testing".
- Two related **over-regeneration** behaviors surfaced in the same QA (whole-file `.infrahub.yml` trigger; Python directory-level closure). They are safe (no missed regen) and documented, and are filed separately as enhancements — out of scope for this correctness fix.

---
## Summary by cubic
Fixes missed artifact regeneration when a proposed change updates a GraphQL query or an artifact definition. Diff actions are now matched case-insensitively (handles uppercase "UPDATED"/"ADDED") so affected artifacts regenerate and don’t merge stale, aligning with INFP-409.

- **Bug Fixes**
  - Added `_is_triggering_action()` and used it in `_query_changed`, `_definition_changed`, and element-level comparisons.
  - Updated unit/component fixtures to emit uppercase actions from `get_diff_summary` so tests catch the regression.

<sup>Written for commit dfa94bcc026d7c1688859620a58578085723c97f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -1272,6 +1272,17 @@ def _run_generator(instance_id: str | None, managed_branch: bool, impacted_insta
 _TRIGGERING_DIFF_ACTIONS = {DiffAction.ADDED.value, DiffAction.UPDATED.value}
 
 
+def _is_triggering_action(action: str) -> bool:
+    """Return True for diff actions that should trigger artifact regeneration.
+
+    ``get_diff_summary`` serialises a node/element action as the GraphQL enum *name*
+    (uppercase, e.g. ``"UPDATED"``/``"ADDED"``), whereas ``DiffAction.*.value`` is
+    lowercase. The comparison must therefore be case-insensitive: a literal
+    ``action in _TRIGGERING_DIFF_ACTIONS`` silently never matches real diff data.
+    """
+    return action.lower() in _TRIGGERING_DIFF_ACTIONS
+
+
 @dataclass(frozen=True, kw_only=True, slots=True)
 class PredicateOutcome:
     """The verdict of a regeneration predicate plus the diagnostic explaining it.
@@ -1304,7 +1315,7 @@ def _query_changed(
     leaves the definition broken and there is nothing to regenerate against.
     """
     matched = any(
-        entry["id"] == definition.query_id and entry["action"] in _TRIGGERING_DIFF_ACTIONS for entry in diff_summary
+        entry["id"] == definition.query_id and _is_triggering_action(entry["action"]) for entry in diff_summary
     )
     if not matched:
         return PredicateOutcome(matched=False)
@@ -1340,7 +1351,7 @@ def _definition_changed(
         (
             entry
             for entry in diff_summary
-            if entry["id"] == definition.definition_id and entry["action"] in _TRIGGERING_DIFF_ACTIONS
+            if entry["id"] == definition.definition_id and _is_triggering_action(entry["action"])
         ),
         None,
     )
@@ -1348,7 +1359,7 @@ def _definition_changed(
         return PredicateOutcome(matched=False)
 
     changed_fields = ", ".join(
-        element["name"] for element in matched_entry["elements"] if element["action"] in _TRIGGERING_DIFF_ACTIONS
+        element["name"] for element in matched_entry["elements"] if _is_triggering_action(element["action"])
     )
     detail = f"definition node was modified ({changed_fields})" if changed_fields else "definition node was modified"
     return PredicateOutcome(
--- a/backend/tests/component/proposed_change/conftest.py
+++ b/backend/tests/component/proposed_change/conftest.py
@@ -72,13 +72,16 @@ def make_node_diff(
     kind: str,
     branch: str,
     field_names: list[str],
-    action: str = "updated",
+    action: str = "UPDATED",
     element_type: str = DiffElementType.ATTRIBUTE.value,
 ) -> NodeDiff:
     """Build a NodeDiff for use in diff summary cache.
 
     `element_type` accepts the raw diff value (e.g. "RELATIONSHIP_ONE") so tests can reproduce a
     relationship endpoint flip exactly as the diff summary emits it.
+
+    `action` is the uppercase GraphQL enum name as emitted by `get_diff_summary`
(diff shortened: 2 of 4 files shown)


changelog/+infp-409-regen-action-case.fixed.md
ajtmccarty, line 1: do we need a changelog if this has not been released yet?

ajtmccarty (approved):

ogenstad (approved):
LGTM, as Aaron mentioned we should remove the `changelog/+infp-409-regen-action-case.fixed.md` file as these changes hadn't been included in a release yet, i.e. it's a bugfix for something that isn't live and as such not something that users have experienced in a published release.

changelog/+infp-409-regen-action-case.fixed.md
PhillSimonds: Good catch, removed.

PR #9648, opened 2026-06-20
feat(graph-traversal): exhaustive mode, truncation & lazy fast-mode path traversal

ajtmccarty:
# Why

PR #9627 (now in `release-1.10`) added bidirectional ("meet-in-the-middle") by-id path traversal.
A few issues remained:

- Issue that could be considered a bug. There was **no way to ask for all loopless paths up to `max_paths`** because the search only returned the shortest path through each intermediate, silently omitting longer simple (meaning no loops) routes.
- More of a user experience issue. The default search still **eagerly expanded the BFS frontier to `ceil(max_depth/2)`** before joining, and **hard-failed on a query timeout**. A deep search on a dense graph could spend a long time (tens of seconds locally) looking for an impossibly large number of "middle" vertices to traverse and then fail with a 504. 

**Goals:** add an opt-in exhaustive mode that returns all loopless paths up to `max_paths`, make the default search lazy (so expanding and search one layer of depth at a time), improve failure to still return some results if possible and give a better error message to help the user craft their search, and surface the new mode and the truncation signal through the GraphQL API

Closes IFC-2807

## Performance

Performance for most cases will remain about the same. It will degrade by <1s for some requests that could be handled in a single database call that now require several smaller database calls.

The default search no longer eagerly expands the frontier or hard-fails.

The most-improved case is the **unfiltered leaf→leaf query** (`max_depth=10`, `max_paths=10`) on a
  dense fabric dataset (~627k nodes) — a deep search with no `relationship_filter`, where almost
  every node is a candidate "middle". Measured locally:

  | | Before (`release-1.10`, eager bidirectional) | After (this PR: lazy + truncation + dedicated timeout) |
  |---|---|---|
  | BFS expansion | both frontiers expanded to `ceil(max_depth/2)` — tens of seconds, ~316k-node frontiers per side | expands one hop at a time; the shallow connection is found at depth 2–3 (a few seconds) |
  | Outcome | deep-tier joins ran under the shared 75 s query timeout → **request hard-failed with a 504, no results** | **returns 5 shortest paths (depths 2–3) + `truncated_at_depth=4`** instead of failing |
  | End-to-end wall time | timeout / 504 | **~35 s** — a few seconds to find the shortest paths, then the doomed depth-4 tier truncates after the dedicated 30 s point-to-point timeout |

  The end-to-end time is now dominated by the single doomed-tier timeout, which the **dedicated
  point-to-point timeout** lowers from the shared 75 s to a 30 s default (operator-tunable). The same
  query under the old 75 s timeout takes ~78 s; lowering it to 30 s yields the same paths and the same
  `truncated_at_depth=4` in less than half the time, and an operator who prefers snappier truncation
  over completeness can lower it further. A **filtered** leaf→leaf query is unaffected: it completes
  fully (`truncated_at_depth=null`) in ~3 s, far under the 30 s budget — no false truncation.

## What changed

**Behavioral**
- **`run()` returns `PathTraversalResult(paths, truncated_at_depth)`** instead of a bare list.
  `truncated_at_depth` is `None` on a completed search, and set to the depth at which a query
  exceeded its budget (the shallower paths are still returned).
- **Default (fast) mode is now lazy / early-terminating.** The BFS expands one hop at a time, each
  side only to the radius the current tier needs, stopping as soon as `max_paths` is reached or the
  frontiers are exhausted — no more eager expansion to `ceil(max_depth/2)`. Results are unchanged
  (same shortest-path-per-intermediate set).
- **New exhaustive mode (`shortest_paths_only=False`)** returns all loopless paths from source to
  destination up to `max_paths`, via per-tier half-enumeration joined on a shared middle (loopless
  enforced per pair). It recovers routes whose midpoint is not at its shortest distance — the ones
  the pinned-middle join omits. for example, say we have paths `A->B->M->Y->Z` and `A->M->Z`. `shortest_paths_only=True` would only return `A->M->Z`, but `shortest_paths_only=False` would return both.
- **Graceful degradation:** a per-tier query timeout (or, in exhaustive mode, a half-path set
  exceeding `exhaustive_half_cap`) truncates and returns the shallower paths + `truncated_at_depth`,
  instead of failing the whole request.
- **Dedicated point-to-point query timeout.** The single shared `graph_traversal_query_timeout`
  (75 s) is split into `path_traversal_query_timeout` (default **30 s**) and
  `reachable_nodes_query_timeout` (default **75 s**, unchanged). Point-to-point traversal now runs
  many small per-tier queries where a tier that doesn't finish quickly is doomed, so it gets the
  lower budget; reachable-nodes still runs larger queries and keeps the higher one. A doomed tier is
  aborted at the budget and the search truncates (above), rather than burning the full 75 s.
- **GraphQL surface (additive):** the query gains an optional `shortest_paths_only: Boolean`
  input (default `true`, so existing callers are unaffected) to opt into exhaustive mode, and the
  result type gains a nullable `truncated_at_depth: Int` field that is non-null only when the search
  was truncated. No existing field changed.

**Implementation**
- `render_bfs_hop` (resumable single-hop BFS; the seed hop resolves the active anchor) replaces the
  eager `render_bfs`, and `BfsHopQuery` replaces `BfsQuery`.
- Exhaustive half-enumeration: `render_half_from_source` / `render_half_to_target` +
  `LeftHalfPathsQuery` / `RightHalfPathsQuery`; the join and per-pair disjointness check live in
  `_run_exhaustive`.
- Query execution is routed through an injected `QueryRunner` (`runner.py`); production uses
  `DefaultQueryRunner`, wired at the resolver. This lets a test adapter simulate a per-query timeout
  without mocking the query or the database.
- The resolver reads the `shortest_paths_only` input and maps `result.truncated_at_depth` onto the
  result type.
- `graph_traversal_query_timeout` is replaced by `path_traversal_query_timeout` and
  `reachable_nodes_query_timeout`; each resolver passes its own. The generated config reference doc is
  regenerated. No new executor logic — the existing `QueryTimeoutError` → truncation path consumes
  the lower budget directly.

**What stayed the same**
- `InfrahubReachableNodes` behaviour is untouched — it keeps the 75 s budget (now via
  `reachable_nodes_query_timeout`). No schema or migration changes. The exhaustive half-cap remains a
  constructor default (no config).

## Front-end changes
- added `shortest_paths_only` checkbox and help text in this commit [a873c4f](https://github.com/opsmill/infrahub/pull/9648/changes/a873c4f144ead5fa8cc1ef0082660bcb2590e224)
- added depth truncation warning message in this commit [9237904](https://github.com/opsmill/infrahub/pull/9648/changes/9237904216e0055e4b045f1aca29a3dcb228a44f)
image

### Suggested review order
1. `backend/infrahub/graph_traversal/executor.py` — `run()` dispatch, `_run_shortest` (lazy),
   `_run_exhaustive`.
2. `backend/infrahub/graph_traversal/_cypher.py` — `render_bfs_hop`, `render_half_from_source` /
   `render_half_to_target`.
3. `runner.py` / `path.py` — the query classes and the runner injection.
4. `backend/infrahub/graphql/queries/path.py` — the `shortest_paths_only` input and
   `truncated_at_depth` output wiring (plus the regenerated schema/types).
5. Tests.

## How to review

Focus on `_run_shortest`'s lazy expansion + truncation and the exhaustive join's per-pair loopless
check (`_run_exhaustive`). The `QueryRunner` seam is what makes the timeout-truncation deterministically
testable.

## Impact & rollout

- **Backward compatibility:** GraphQL changes are additive only (a new optional input that defaults
  to today's behaviour, a new nullable output field); default behaviour preserved (fast mode, same
  paths). The internal `run()` return type changed to `PathTraversalResult` — only the resolver
  consumes it.
- **Performance:** described above
- **Config/env changes:** `graph_traversal_query_timeout` (env
  `INFRAHUB_DB_GRAPH_TRAVERSAL_QUERY_TIMEOUT`) is replaced by `path_traversal_query_timeout` (default
  30 s) and `reachable_nodes_query_timeout` (default 75 s). The replaced setting was added by the
  still-unreleased #9627, so no released deployment sets it; an operator who set it on `release-1.10`
  should switch to the new name(s).

## Checklist

- [x] Tests added/updated
- [x] Changelog 
- [x] External docs — the GraphQL change is additive and self-describing (field descriptions on the
      new input/output); the generated `schema/schema.graphql` and the config reference
      (`docs/docs/reference/configuration.mdx`) are regenerated and committed.
- [x] Internal .md docs — n/a


backend/infrahub/graph_traversal/runner.py
```diff
@@ -0,0 +1,27 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol
+
+if TYPE_CHECKING:
+    from infrahub.core.query import Query
+    from infrahub.database import InfrahubDatabase
+
+
+class QueryRunner(Protocol):
```
ajtmccarty, line 17: I don't feel great about this b/c it feels like writing the code to fit the test, but it _does_ provide a clean way to test a timeout on the neo4j side

ogenstad: I don't see anything wrong with this approach.

backend/infrahub/graphql/queries/path.py
```diff
@@ -299,15 +319,19 @@ async def path_traversal_resolver(
...
                 source_id=source_id,
                 max_paths=max_paths,
+                shortest_paths_only=shortest_paths_only,
                 at=graphql_context.at,
             )
+            path_data_list = result.paths
+            truncated_at_depth = result.truncated_at_depth
         except ValueError as exc:
             raise GraphQLError(str(exc)) from exc
         except QueryTimeoutError as exc:
```
ogenstad: Now we are catching `QueryTimeoutError` within `executor.run()`. As such are we still expected to ever catch this error here?

ogenstad (approved):

backend/infrahub/graphql/queries/path.py
ajtmccarty: we are not. it's removed

pa-lem (approved):

PR #9656, opened 2026-06-22
fix(auth): harden SSO auto-create account groups (claim-failure isolation + uniqueness)

BeArchiTek:
## Why

The SSO auto-create-groups find-or-create path had gaps that were not covered by tests:
- a single failing claim could abort the whole login
- group creation bypassed the name uniqueness constraint, so the same name created from another route (e.g. the GraphQL API) could produce a duplicate

This PR adds the failing-case coverage and the production fixes.

Closes [IFC-2771](https://opsmill.atlassian.net/browse/IFC-2771)

## Cases

| Case | Before | After (covering test) |
|------|--------|------------------------|
| One claim fails mid-assignment | exception aborted the login (500); other valid groups dropped | logged + rejected event, remaining claims still apply (`test_claim_failure_isolation.py`) |
| Auto-create skipped the uniqueness constraint | raw `save`, no constraint runner, no DB unique index; silent duplicate with the same name and HFID | created via `create_node()`; a create that loses the uniqueness race reuses the winner with no second event (`test_autocreate_race_reuse.py`) |
| Cross-path race (SSO vs GraphQL/API create) | bespoke `auto-create-group` lock used a different key than the mutation path, no serialization; silent duplicate | `create_node()` locks on the shared schema key, so all creation paths serialize; the loser reuses (`test_autocreate_race_reuse.py`) |

## How to test

```bash
uv run pytest backend/tests/component/auth/auth_groups/
```

## Manual verification (live stack)

Scenarios from `dev/specs/infp-556-auto-create-groups/test-plan.md` run against a live Keycloak + Infrahub stack on this branch (checked via GraphQL probes + server logs):

| Scenario | Result |
|---|---|
| A — groups created, `origin` set, created events | ✅ |
| B — filter exclusion is silent (no group, no event) | ✅ |
| C — idempotent reuse on re-login (stable UUID, no duplicate, no new event) | ✅ |
| D — per-login cap + capped event | ✅ |
| E — rejected-claim event | ✅ |
| F — `origin` read-only (mutation input parse errors) | ✅ |
| G — default-group fallback (first login) | ✅ |

H (docs) and I (activity-feed UI) not run.

---
## Summary by cubic
Route SSO auto-create of account groups through `create_node` to enforce name uniqueness under the shared object lock and reuse the winning row on races without emitting a second creation event; also isolate per-claim failures so a bad claim is logged, emits a rejected event, and the login continues.

Add component tests confirming a create race produces one `CoreAccountGroup` with no duplicate event and that a claim colliding with a non-AccountGroup is skipped while login succeeds; refactor test helpers to use a shared `RecordingAutoCreateEventEmitter` and expose the race-reuse flag publicly.

<sup>Written for commit 7d299d0b8f2d6f034a450cc7f3ae5958d611099e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9656?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2771]: https://opsmill.atlassian.net/browse/IFC-2771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/auth/auth_groups/service.py
```diff
@@ -111,32 +119,42 @@ async def assign(
...
-                    )
-                    dropped.append(claim)
-                    continue
-                group = self._reuse_or_skip(name, existing)
-                if group is None:
-                    continue
-
-            if await self._add_member(group):
-                granted.append(name)
+            try:
```
polmichel, line 119: One fail claim should not block all the login process and other following valid claims

backend/infrahub/auth/auth_groups/service.py
```diff
@@ -145,23 +160,35 @@ async def assign(
...
             return FindOrCreateResult(group=self._reuse_or_skip(name, existing), was_created=False)
 
-        lock_key = f"auto-create-group:{name}"
-        async with self._lock_registry.get(name=lock_key, namespace="auto-create-group"):
+        branch = await registry.get_branch(db=self._db)
+        try:
+            group = await create_node(
+                data={"name": name, "origin": self._provider_name},
+                db=self._db,
+                branch=branch,
```
polmichel, line 181: usage of `create_node` API to make sure a parallel GraphQL mutation does not create the same group at the same time. The initial lock is not required anymore since  `async with InfrahubMultiLock(lock_registry=lock.registry, locks=lock_names, metrics=False):` is used inside and `lock_names` comes from `infrahub.core.node.lock_utils.get_lock_names_on_object_mutation`

backend/infrahub/auth/auth_groups/service.py
```diff
@@ -145,23 +160,35 @@ async def assign(
...
-            if existing is not None:
-                return FindOrCreateResult(group=self._reuse_or_skip(name, existing), was_created=False)
-
-            group = await Node.init(db=self._db, schema=CoreAccountGroup)
-            await group.new(db=self._db, name=name, origin=self._provider_name)
-            await group.save(db=self._db)
-            await self._emitter.created(group=group, source_pattern=source_pattern)
-            return FindOrCreateResult(group=group, was_created=True)
+            if existing is None:
+                raise
```
polmichel, line 187: If this is another reason than a duplication, let's raise it. The caller catches and logs it.

backend/infrahub/auth/auth.py
```diff
@@ -400,7 +400,6 @@ async def _assign_group_memberships(
             db=db,
             account=account,
             provider_name=external_identity.provider_name,
-            lock_registry=lock.registry,
```
polmichel, line 403: not used anymore since `create_node()` api has its own lock injected through the registry internally

backend/infrahub/core/node/create.py
```diff
@@ -275,20 +277,50 @@ async def _do_create_node(
...
+) -> Node | SchemaProtocol:
     """Create a node in the database if constraint checks succeed.
 
+    A schema protocol class may be passed instead of a schema object; the return type is then
+    narrowed to that protocol.
+
     Raises:
-        ValueError: When the schema is a `GenericSchema` and cannot be instantiated.
+        ValueError: When the schema is a `GenericSchema` and cannot be instantiated, or when a
+            class that is not a node schema protocol is passed.
```
polmichel, line 318: new API which types the object returned by the `create_node` method

backend/tests/component/auth/auth_groups/test_autocreate_race_reuse.py
```diff
@@ -0,0 +1,107 @@
...
+
+if TYPE_CHECKING:
+    from infrahub.core.branch import Branch
+    from infrahub.core.schema.schema_branch import SchemaBranch
+    from infrahub.database import InfrahubDatabase
+
+CONTESTED_NAME = "ops-admins"
+
+
+class _LookupMissesOnceNodeManager(NodeManager):
```
polmichel, line 21: This test if a similar request (GraphQL API or another similar login process) is triggered between `_lookup_by_name` call and `create_node(...)` call

backend/tests/component/auth/auth_groups/test_autocreate_race_reuse.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    winner = await Node.init(db=db, schema=CoreAccountGroup)
+    await winner.new(db=db, name=CONTESTED_NAME, origin="OktaProd")
+    await winner.save(db=db)
+
+    account = await Node.init(db=db, schema=CoreAccount)
+    await account.new(db=db, name="Pat Auto", account_type="User", password="pat-password")
+    await account.save(db=db)
+
+    _LookupMissesOnceNodeManager._first_lookup_done = False
```
ajtmccarty: the instance variable should probably be public or use a setter. doesn't really matter b/c it is all in this one test file

backend/tests/component/auth/auth_groups/test_claim_failure_isolation.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @classmethod
+    async def query(cls, *args: Any, **kwargs: Any) -> Any:
+        filters = kwargs.get("filters")
+        if filters and filters.get("name__value") == POISON_NAME:
+            raise RuntimeError("simulated lookup failure for one claim")
+        return await super().query(*args, **kwargs)
+
+
+class _RecordingEmitter(AutoCreateEventEmitter):
```
ajtmccarty: can maybe move somewhere to be reused with the other new test

backend/tests/component/auth/auth_groups/test_claim_failure_isolation.py
```diff
@@ -0,0 +1,100 @@
...
+    await account.save(db=db)
+
+    emitter = _RecordingEmitter()
+    service = AutoCreatedGroupsService(
+        db=db,
+        account=account,
+        provider_name="AzureAD-corp",
+        max_per_login=10,
+        emitter=emitter,
+        node_manager=_FailLookupNodeManager,
```
ajtmccarty, line 59: excellent.

ajtmccarty (approved):

backend/tests/component/auth/auth_groups/test_autocreate_race_reuse.py
polmichel: You're right, will do it

backend/tests/component/auth/auth_groups/test_claim_failure_isolation.py
polmichel: Yep good call, I will move it elsewhere. It is also duplicated in another file so it's better that way

PR #9663, opened 2026-06-22
chore: convert changelog guideline into a skill

saltas888:
## What

Converts the `dev/guidelines/changelog.md` guideline into a discoverable Agent Skill at `dev/skills/creating-changelog-entries/`, auto-exposed through the existing `.claude/skills` → `dev/skills` symlink.

## Why

Skills are surfaced to AI agents by their description and loaded on demand, so changelog guidance now triggers automatically when an agent is about to record a change — instead of relying on someone opening a guideline file.

## Details

- New skill `creating-changelog-entries`, written project-agnostic with **Towncrier as the precondition**:
  - the `description` gates discovery (only fires for projects that use Towncrier),
  - the `compatibility` frontmatter field formally declares the requirement.
- Updated references to "Use the `creating-changelog-entries` skill" (agent-facing): `backend/AGENTS.md`, `.github/workflows/bug-agent-fix.md`, `dev/bug-pipeline/fix-implementation.md`, `dev/guidelines/git-workflow.md`. The human-facing PR-template checkbox keeps a clickable link.
- Removed `dev/guidelines/changelog.md` and its "example guideline" entry in `repository-organization.md`.
- Drive-by: pre-commit removed one pre-existing trailing space in `repository-organization.md`.

## Testing

- Verified the skill resolves through the `.claude/skills` symlink and that no references to the old guideline remain anywhere.
- Ran a fresh subagent with a realistic task and no hints: it discovered the skill and produced the correct `towncrier create` command, the repo-root `changelog/` location, the right change type, and a user-facing message.

No changelog fragment added: this is internal dev-docs/tooling only, not user-facing.

## Skill validation — RED→GREEN subagent test

The skill was validated by running fresh subagents on realistic tasks in an **isolated standalone Towncrier project** (real `[tool.towncrier]` config, a monorepo `backend/` package, and no other changelog docs), so the skill is the only variable. Each scenario was run **without the skill (RED)** and **with the skill (GREEN)**.

**Test A — message discipline** (bug fix, issue #4821) — 5 reps each:

| Metric | RED (no skill) | GREEN (with skill) |
|---|:--:|:--:|
| Message leaks implementation ("…by invalidating the cached auth token") | 5 / 5 ❌ | 0 / 5 ✅ |
| Wrong tense (imperative "Invalidate…") | 2 / 5 | 0 / 5 |

Every GREEN rep converged on one user-facing sentence — *"Fixed users being unexpectedly logged out after changing their password."* — while every RED rep described the internals.

**Test B — fragment location** (same fix, with a decoy `backend/changelog/` fragment as a tempting precedent and a "put it where backend keeps fragments" nudge) — 4 reps each:

| Metric | RED (no skill) | GREEN (with skill) |
|---|:--:|:--:|
| New fragment misplaced in `backend/changelog/` | 4 / 4 ❌ | 0 / 4 ✅ |
| Used `towncrier create` → repo-root `changelog/` | 0 / 4 | 4 / 4 |

Under pressure, RED reverted to hand-writing the fragment into the sub-package directory — exactly the two mistakes the skill's **Common Mistakes** call out. GREEN recognized the decoy as misplaced and used `towncrier create`, landing in the configured `changelog/`.

**Takeaway:** the skill produces a clean, reproducible RED→GREEN improvement on its two judgment rules — user-facing message and correct fragment location (perfect separation across reps). The mechanical conventions (`towncrier create`, change type, `+` orphan prefix) are handled correctly with or without the skill by a capable agent, so those sections act as guardrails rather than behavior-changers.

---
## Summary by cubic
Converted the changelog guideline into a discoverable `creating-changelog-entries` skill that auto-loads in `towncrier` projects. The skill lives under `.agents/skills`, examples use `uv run`, and docs now align on change types (dependency bumps → `changed`; `housekeeping` for maintenance/tooling).

- **New Features**
  - Added `.agents/skills/creating-changelog-entries/` skill, gated by `description`/`compatibility`, and auto-exposed via `.claude/skills` → `.agents/skills`.
  - Refined skill text: “Use when …” wording, removed redundant precondition line; Quick Reference and examples use `uv run`, with guidance for other runners.
  - Clarified conventions and caveats: fragment location rules, nested-package Towncrier caveat; updated `docs/docs/development/changelog.mdx` to include `housekeeping` and note deps → `changed`.

- **Migration**
  - Use the `creating-changelog-entries` skill instead of `dev/guidelines/changelog.md`; references and the PR template link now point to the skill.
  - No user-facing changes.

<sup>Written for commit fc37b9c939c902a680f0e6a09367673f982d3055. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9663?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.agents/skills/creating-changelog-entries/SKILL.md
saltas888, line 1: This is an almost exact copy-paste of the guidelines. The difference are:
- Compatibility section https://agentskills.io/specification#compatibility-field
- Minor adjustments in the section so it's not infrahub specific.

The aim of this skill is to be part of our opsmill-wide skills, project agnostic

dev/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- Focus on what changed, not how.
+- Use past tense ("Fixed", "Added", "Removed").
+- Keep it concise — one sentence.
+- Avoid technical jargon.
+
+## Examples
+
+```bash
+# Bug fix for issue #1234
+towncrier create -c "Fixed sidebar collapse issue" 1234.fixed.md
```
BeArchiTek: The examples use a bare `towncrier create`, but this repo runs Python tools through `uv` — every other reference updated in this PR (the PR template, `backend/AGENTS.md`, the bug-pipeline files) uses `uv run towncrier create`. The Quick Reference's runner note covers it in principle, but the examples are the part people copy verbatim, so the canonical project case is the one that breaks (bare `towncrier` won't be on PATH here). Suggest prefixing the examples with `uv run`, or adding a short marker that this repo needs the `uv run` prefix.

BeArchiTek (commented):
A few tightening notes on the new skill — all minor, none blocking. Mechanics look good: frontmatter is well-formed, the references are rewired completely (no dangling links to the old guideline), and the skill is discoverable via the existing `.claude/skills` symlink.

dev/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+compatibility: Requires the project to use Towncrier for changelog management — a configured Towncrier setup (e.g. `[tool.towncrier]` in pyproject.toml) and the `towncrier` CLI available, typically via a Python runner such as uv or poetry.
+---
+
+# Creating Changelog Entries
+
+## Overview
+
+[Towncrier](https://towncrier.readthedocs.io/) assembles the changelog from per-change "fragment" files — one small file per change, collected and rendered at release time. This skill applies to any project that uses Towncrier. Every issue fix or new feature should ship with a fragment, written short and user-facing: describe *what* changed, not how it was implemented.
+
+**Requires Towncrier.** If the project has no `[tool.towncrier]` config (or another Towncrier config file), this skill does not apply — follow that project's own changelog process instead.
```
BeArchiTek: The Towncrier precondition is now stated four times — the `description`, the `compatibility` field, this bold Overview line, and the "When NOT to use" bullet. The `compatibility` field is the canonical home for it; this Overview line is the most redundant of the four. Consider dropping one or two to keep the body lean.

dev/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,86 @@
+---
+name: creating-changelog-entries
+description: Use in a project that manages its changelog with Towncrier, when you've fixed a bug, added a feature, or made any user-facing change and need to record it — before committing or opening a PR, or whenever asked to add a changelog entry, towncrier fragment, or news fragment.
```
BeArchiTek: Minor: the description opens with "Use in a project that…". The common convention is to open with "Use when…" and keep it purely about triggering conditions — which this already does well (good keyword coverage, and it avoids restating the workflow). Only worth rewording if you're already touching the frontmatter, e.g. "Use when recording a user-facing change in a project that manages its changelog with Towncrier — …".

dev/skills/creating-changelog-entries/SKILL.md
saltas888: Dropped the bold "Requires Towncrier" Overview line. The `compatibility` field stays as the canonical declaration. The `description` keeps the precondition because it is the only field read at discovery time, so it has to gate there; and the "When NOT to use" bullet stays as the lightweight in-body signal. ca6f7708e.

dev/skills/creating-changelog-entries/SKILL.md
saltas888: Reworded to open with "Use when": the description now reads "Use when you've fixed a bug, added a feature, or made any user-facing change in a project that uses Towncrier and need to record it for the changelog — …". ca6f7708e.

dev/skills/creating-changelog-entries/SKILL.md
saltas888: Adopting this — the examples now use `uv run towncrier create ...` (the runner used across our Python projects), so the verbatim-copy case works here. Kept a Quick Reference note for non-uv projects: drop the prefix (`towncrier create ...`) or use the project runner (e.g. `poetry run`). 13e1f3dce.

dgarros:
@saltas888 looks good but ideally we should target stable for this PR not develop
also just this morning we started to refactor the organization of the repo and now the skills lives under `.agents/skills`
#9595

BeArchiTek (commented):
Reviewed the new skill against the live Towncrier setup (`[tool.towncrier]` in `pyproject.toml`), the existing fragments in `changelog/`, and the related docs/commands. The conversion is solid overall — the description, structure, and the change-type table all line up with the config, and the "`towncrier create`, never hand-write" framing is the right core rule.

A few things to reconcile, the first being the only real blocker:

1. **Dependency change-type conflicts with existing convention.** The skill files dependency bumps under `housekeeping`, but `dev/commands/upgrade-backend-dependencies.md` (Step 8) and the only existing upgrade fragment — `changelog/+upgrade-python-3.14-neo4j.changed.md` — both use `changed` with `+upgrade-{component}` naming. As written, two contributors will changelog the same dependency bump under different types. Pick one convention and make the skill and the command agree (details inline).

2. **`docs/docs/development/changelog.mdx` is now narrower than the skill.** It lists only 6 types and never mentions `housekeeping`, yet the skill's "See Also" sends readers there for project conventions. Worth syncing in this PR or a fast follow-up.

3. **SDK submodule edge case** in Fragment Location (inline).

Plus a couple of wording nits inline. Everything else checks out — all 7 types match `[tool.towncrier]`, the orphan `+` prefix matches `orphan_prefix`, and the referenced doc exists.

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+
+| Type | Use For |
+|------|---------|
+| `added` | New features |
+| `changed` | Changes in existing functionality |
+| `deprecated` | Soon-to-be removed features |
+| `removed` | Now removed features |
+| `fixed` | Bug fixes |
+| `security` | Security vulnerabilities |
+| `housekeeping` | Internal maintenance, dependencies, tooling |
```
BeArchiTek: **Type conflict to resolve.** This row lists `dependencies` under `housekeeping`, but `dev/commands/upgrade-backend-dependencies.md` (Step 8) changelogs dependency bumps as `changed`:

```bash
uv run towncrier create +upgrade-{component}.changed.md -c "Upgraded {component} from {old} to {new}"
```

and the only existing upgrade fragment agrees — `changelog/+upgrade-python-3.14-neo4j.changed.md` is `changed`. The repo's actual `housekeeping` fragments (`+permission-resolver`, `+pnpm-workspaces`) are tooling/workspace changes, not version bumps.

Suggest narrowing this row to "internal maintenance, tooling" and routing dependency bumps to `changed` (aligning the example on line 70), or — if the team genuinely wants deps -> `housekeeping` — updating that command and the existing fragment so they don't disagree.

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+
+```bash
+# Bug fix for issue #1234
+uv run towncrier create -c "Fixed sidebar collapse issue" 1234.fixed.md
+
+# New feature for issue #7549
+uv run towncrier create -c "Added breadcrumb navigation for hierarchical schemas" 7549.added.md
+
+# Housekeeping without an issue
+uv run towncrier create -c "Updated dependencies to latest versions" +deps-update.housekeeping.md
```
BeArchiTek: Same conflict, concrete: this teaches `+deps-update.housekeeping.md`, but the dependency-upgrade command and `changelog/+upgrade-python-3.14-neo4j.changed.md` use `changed` + `+upgrade-{component}` naming. Whichever type wins in the table above, make this example match it so the skill and the command tell contributors the same thing.

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+- **Hand-writing the fragment file.** Use `towncrier create` so the name and location are correct.
+- **Placing it in a sub-package directory** (e.g. `backend/changelog/`) instead of the configured fragments directory.
+- **Describing the implementation.** "Refactored the auth-token cache layer" → instead say what the user sees: "Fixed users being unexpectedly logged out".
+- **Wrong tense or multiple sentences.** One past-tense sentence.
+
+## See Also
+
+- [Towncrier documentation](https://towncrier.readthedocs.io/) — fragment types, configuration, and CLI options.
+- [Keep a Changelog](https://keepachangelog.com/) — changelog best practices.
+- Your project's contributing or changelog guide for project-specific conventions (in Infrahub: `docs/docs/development/changelog.mdx`).
```
BeArchiTek, line 86: Heads-up on this pointer: `docs/docs/development/changelog.mdx` currently lists only 6 types (Added/Changed/Deprecated/Removed/Fixed/Security) and never mentions `housekeeping`, even though it's a configured type in `[tool.towncrier]`. A reader who follows this link for "project-specific conventions" gets a narrower, partly-conflicting list. Updating that doc to include `housekeeping` (here or a follow-up) keeps the skill and the public doc from diverging.

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+| `security` | Security vulnerabilities |
+| `housekeeping` | Internal maintenance, dependencies, tooling |
+
+The available types come from the project's Towncrier config (`[tool.towncrier]`); the set above is the common default.
+
+## Fragment Location
+
+Fragments live in the directory Towncrier is configured to use (its `directory` setting — commonly the repo-root `changelog/` or `newsfragments/`). `towncrier create` puts them there automatically — don't move them or hand-place them.
+
+In a monorepo, they do **not** belong in a sub-package directory such as `backend/changelog/` or `frontend/changelog/`. A fragment found there was hand-written in the wrong place and must move to the configured directory.
```
BeArchiTek, line 50: Accurate for `backend/`/`frontend/` — those aren't separate Towncrier projects. One edge case: the `python_sdk/` submodule has its *own* `[tool.towncrier]` config and its own `changelog/`, and `docs/docs/development/changelog.mdx` routes SDK changes there. As written, someone fixing an SDK bug could read this and move a valid `python_sdk/changelog/` fragment to the repo root. A one-line caveat — "a nested package with its own towncrier config (e.g. the `python_sdk` submodule) keeps its own fragments" — closes that gap.

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+compatibility: Requires the project to use Towncrier for changelog management — a configured Towncrier setup (e.g. `[tool.towncrier]` in pyproject.toml) and the `towncrier` CLI available, typically via a Python runner such as uv or poetry.
+---
+
+# Creating Changelog Entries
+
+## Overview
+
+[Towncrier](https://towncrier.readthedocs.io/) assembles the changelog from per-change "fragment" files — one small file per change, collected and rendered at release time. This skill applies to any project that uses Towncrier. Every issue fix or new feature should ship with a fragment, written short and user-facing: describe *what* changed, not how it was implemented.
+
+**Core rule:** always create fragments with `towncrier create`, never hand-write them. The command derives the location and naming from the Towncrier config and places the fragment correctly no matter which directory you run it from.
```
BeArchiTek: Tiny precision nit: `towncrier create` still has to locate the config, so "no matter which directory you run it from" only holds *inside the repo*. "from anywhere in the repo" is the exact claim.

saltas888:
> @saltas888 looks good but ideally we should target stable for this PR not develop also just this morning we started to refactor the organization of the repo and now the skills lives under `.agents/skills` #9595

@dgarros good point. I added to `cubic.dev` this so it's captured by the review agent
image

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: Resolved in 3d8448074. Checked the convention against the repo: `.agents/commands/upgrade-backend-dependencies.md` (lines 418/423) creates `+upgrade-{component}.changed.md`, and historical fragments agree (`+upgrade-deps-20251206`, `+upgrade-python-3.14-neo4j`, `+upgrade-process` all `.changed.md`; only `+upgrade-neo4j-2026-04` was `housekeeping`). So deps → `changed`.

Since the skill is meant to be project-agnostic, I narrowed this row to `Internal maintenance and tooling (build, CI, dev scripts)` and added a line noting that deps-bump classification is project-specific (deferring to the project guide) rather than hard-coding deps → `housekeeping`. The Infrahub-specific `changed` rule now lives in `changelog.mdx` so the skill, the doc, and the command agree.

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: Fixed in 3d8448074 — the housekeeping example no longer uses a dependency bump. It is now `+pnpm-workspaces.housekeeping.md` ("Migrated the frontend build to pnpm workspaces"), a genuine tooling change that matches a real housekeeping fragment. Dependency bumps are documented as `changed` in `changelog.mdx`, consistent with the upgrade command.

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: `changelog.mdx` updated in 3d8448074: added `housekeeping` to both the narrative category list and the Newsfragment-suffix list, plus a note that dependency-version bumps are recorded as `changed` (pointing at the `upgrade-backend-dependencies` command), reserving `housekeeping` for tooling/build/CI. The doc and the skill no longer diverge.

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: Added in 3d8448074. Fragment Location now carries an exception: "a nested package with its *own* Towncrier config (e.g. a submodule like Infrahub's `python_sdk`) keeps its own fragments in that package's configured directory — those stay put and must not be moved to the root." This matches the SDK routing already documented in `changelog.mdx`.

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: Done in 3d8448074 — reworded to "places the fragment correctly from anywhere in the repo".

polmichel:
> > @saltas888 looks good but ideally we should target stable for this PR not develop also just this morning we started to refactor the organization of the repo and now the skills lives under `.agents/skills` #9595
> 
> @dgarros good point. I added to `cubic.dev` this so it's captured by the review agent image

Regarding this @saltas888, some bugs we're working on (at least I am working on) have a potential impact which is definitely not low (lock scope modification, git sync issues, ...). However, we still target `stable` for those as we want to avoid users to have to wait for a minor release to get bug fixes. If this is the correct way to manage those issues, based on the current definition inside Cubic, it might be that Cubic will tag everything as high-impact and flood the PR threads. Not sure this is important since we have the capability to resolve the thread.

saltas888:
> > > @saltas888 looks good but ideally we should target stable for this PR not develop also just this morning we started to refactor the organization of the repo and now the skills lives under `.agents/skills` #9595
> > 
> > 
> > @dgarros good point. I added to `cubic.dev` this so it's captured by the review agent image
> 
> Regarding this @saltas888, some bugs we're working on (at least I am working on) have a potential impact which is definitely not low (lock scope modification, git sync issues, ...). However, we still target `stable` for those as we want to avoid users to have to wait for a minor release to get bug fixes. If this is the correct way to manage those issues, based on the current definition inside Cubic, it might be that Cubic will tag everything as high-impact and flood the PR threads. Not sure this is important since we have the capability to resolve the thread.

@BeArchiTek Good point, adjusted to the following
```
Base branch is chosen by release vehicle, not impact level:

- develop: product changes on the normal release train (features, and fixes that can wait for the next minor).
- stable: changes that must ship before the next minor (hotfixes, regardless of impact) or that can't affect a running product (repo tooling, docs, skills, AGENTS.md, CI).

Urgency cannot be inferred from the diff, so do not flag a high-impact fix for targeting stable — that is the expected hotfix path. Only raise a base-branch concern when: (1) a pure tooling/docs/CI change targets develop, or (2) a product feature (not a fix) targets stable. Raise it as a low-severity note, never a blocking issue.
```

fatih-acar (approved):

Issue #6533, opened 2025-05-26
bug: tasks view does not show up


fatih-acar:
### Component

Not Sure

### Infrahub version

1.2.11

### Current Behavior

When opening the task view after a few actions, we get in a state where it errors with:

Cannot return null for non-nullable field TaskRelatedNode.kind.

![Image](https://github.com/user-attachments/assets/22c54de5-5884-4b9f-869a-7be4e158b1a3)

### Expected Behavior

Tasks view should open

### Steps to Reproduce

Using https://github.com/opsmill/poc-service-catalog : 

- create service request
- delete proposed changes manually on infrahub
- delete branch on infrahub

- open task views

dgarros:
is this bug still present in 1.3 ?

fatih-acar:
This bug is still present in `1.3.2` (version currently in use by the service catalog demo)

saltas888:
Closing as a duplicate of **#9662**, which carries the full diagnosis and a linked fix.

This is the oldest of three filings of one bug:

- **this one** (2025-05-26)
- #8342 (2026-02-09) — same failure reached via webhook configuration
- #9662 (2026-06-22) — names the exact `TypeError`

The defect: `TaskRelatedNode.kind` is declared `String(required=True)` in the GraphQL schema while the producer emits `{"id": ..., "kind": None}` for any id it cannot resolve. Resolution requires an active `IS_PART_OF` edge, so the reproduction in this report — delete a proposed change, delete its branch, open Tasks — leaves flow-run tags pointing at nodes that no longer resolve. The non-nullable field then fails the entire query, and Apollo's default error policy discards partial data, which is why the page renders blank rather than partially.

One correction for the record: **the lookup is not branch-scoped.** `NodeGetKindQuery` is called with no `branch`, so the "fall back to main" idea that comes up in these threads would be a no-op — the ids fail because they have no active edge at all, not because they are being looked up on the wrong branch.

PR #9664 declares "Closes #9662"; merging it resolves all three issues. Removing `state/need-more-info` since the diagnosis is now complete.

Issue #8342, opened 2026-02-09
bug: Task page fails to load after configuring webhooks


minitriga:
### Component

Frontend UI

### Infrahub version

1.7.4

### Current Behavior

<img width="1702" height="1327" alt="Image" src="https://github.com/user-attachments/assets/cc530b62-9ad3-43a3-8326-05211a2d7ada" />

```
index-DgAzZxpo.js:76 [GraphQL error]: Message: Cannot return null for non-nullable field TaskRelatedNode.kind., Location: [{"line":19,"column":11}], Path: InfrahubTask,edges,9,node,related_nodes,0,kind
```

All API requests return 200. 

```
{
    "data": {
        "InfrahubTask": {
            "count": 1399,
            "edges": [
                {
                    "node": {
                        "created_at": "2026-02-09T16:07:07.209299Z",
                        "id": "286b668d-a146-498b-b15b-c261d7d72bf2",
                        "branch": "main",
                        "related_node": null,
                        "related_nodes": [],
                        "title": "Trigger updates for computed attributes on branch main for NetworkInterface.description",
                        "updated_at": "2026-02-09T16:07:07.892572Z",
                        "state": "COMPLETED",
                        "progress": null,
                        "workflow": "trigger_update_python_computed_attributes",
                        "__typename": "TaskNode"
                    },
                    "__typename": "TaskNodes"
                },
                {
                    "node": {
                        "created_at": "2026-02-09T16:07:06.099606Z",
                        "id": "b861f67e-d751-41a7-a996-29afc2c1ea35",
                        "branch": "main",
                        "related_node": null,
                        "related_nodes": [],
                        "title": "Setup computed attributes for Python transforms in task-manager",
                        "updated_at": "2026-02-09T16:07:07.493475Z",
                        "state": "COMPLETED",
                        "progress": null,
                        "workflow": "computed-attribute-setup-python",
                        "__typename": "TaskNode"
                    },
                    "__typename": "TaskNodes"
                },
                {
                    "node": {
                        "created_at": "2026-02-09T16:07:03.149605Z",
                        "id": "a7e5c639-96c2-4f78-a5a7-f9ddc32ffedb",
                        "branch": "main",
                        "related_node": "18929d45-e491-707c-3021-c51214b4fbf7",
                        "related_nodes": [
                            {
                                "id": "18929d45-e491-707c-3021-c51214b4fbf7",
                                "kind": "CoreRepository",
                                "__typename": "TaskRelatedNode"
                            }
[...]

PR #9664, opened 2026-06-22
fix(backend): omit task related nodes with unresolvable kind

### LINKED ISSUE #9662: bug: `TypeError: Cannot return null for non-nullable field TaskRelatedNode.kind` when viewing Tasks on a non-default branch
### Component

API Server / GraphQL

### Infrahub version

 1.9.8

### Current Behavior

When the **Tasks** page is opened on any non-default branch, the server throws repeated critical GraphQL resolver errors and the page fails to load:

```
[critical] Unhandled exception occurred in resolvers [infrahub.graphql]
TypeError: Cannot return null for non-nullable field TaskRelatedNode.kind.
```

The error fires once per task whose `related_nodes` list contains a node ID that cannot be resolved on the current branch. With many tasks in the queue this produces dozens of critical log entries per page load.

The underlying sequence:

1. Generator and query-group workflows tag their Prefect flow runs with `infrahub.app/node/<id>`, where `<id>` is the UUID of a `CoreGeneratorDefinition` or `CoreGraphQLQuery` node.
2. These nodes exist only on `main` — they are never present in the DB view of a feature branch.
3. `_get_related_nodes` in `task_manager/task.py` runs `NodeGetKindQuery` using the **branch-scoped** `db` context of the requesting user. The query returns no results for those IDs.
4. `RelatedNodeInfo.kind` defaults to `None` and is never populated.
5. `TaskRelatedNode.kind` is declared `required=True` (non-nullable) in the GraphQL type, so the resolver crashes.

### Expected Behavior

The Tasks page loads successfully on any branch. When a related node ID cannot be resolved on the current branch — because it belongs to a schema-level object (`CoreGeneratorDefinition`, `CoreGraphQLQuery`) that exists only on `main` — `kind` should either be resolved via a fallback lookup against the default branch, or the field should be handled gracefully without crashing the resolver.

### Steps to Reproduce

1. Create any non-default branch.
2. On `main`, run a generator or any workflow that produces a Prefect flow run tagged with `infrahub.app/node/<id>` referencing a `CoreGeneratorDefinition` or `CoreGraphQLQuery` node.
3. Switch the UI to the non-default branch and open the **Tasks** page.
4. Observe repeated `TypeError: Cannot return null for non-nullable field TaskRelatedNode.kind.` in the server log.
5. Alternatively, reproduce directly with the following query against any non-default branch:

```graphql
{
  InfrahubTask {
    edges {
      node {
        id
        title
        state
        related_nodes {
          id
          kind
        }
      }
    }
  }
}
```

### Additional Information

Relevant code paths:

- **Branch-scoped lookup (root cause):** `backend/infrahub/task_manager/task.py` — `_get_related_nodes()` passes the request-scoped `db` to `NodeGetKindQuery`; nodes defined only on `main` are invisible from a feature branch.
- **Unset kind field:** `backend/infrahub/task_manager/models.py` — `RelatedNodeInfo.kind: str | None = None` stays `None` when the query returns no match.
- **Non-nullable GraphQL field:** `backend/infrahub/graphql/types/task.py` — `TaskRelatedNode.kind = String(required=True)` rejects `None` and crashes the resolver.

**Suggested fixes (in order of correctness):**

1. **Fall back to `main`** — when `NodeGetKindQuery` returns no result for an ID on the current branch, retry against the default branch. Matches the actual semantics: generator/query definitions always live on `main`.
2. **Make `kind` nullable** — change `required=True` to `required=False` in `TaskRelatedNode`. Minimal change; UI consumers must handle a null `kind`.
3. **Filter unresolvable nodes** — in `get_related_nodes_as_dict`, drop entries where `kind is None` before returning. Avoids the crash without changing the GraphQL contract.

lancamat1:
## Problem

Closes #9662.

Opening the **Tasks** page crashes the GraphQL resolver with:

```
TypeError: Cannot return null for non-nullable field TaskRelatedNode.kind.
```

A Prefect flow run can be tagged with `infrahub.app/node/<id>` where `<id>` no longer resolves to an active node (deleted definition, stale tag). `RelatedNodeInfo.kind` then stays `None`, and the non-nullable `TaskRelatedNode.kind` field rejects it — one critical log entry per unresolvable task, and the page fails to load.

## Correction to the issue's root-cause analysis

The issue's suggested fix #1 ("branch-scoped lookup → fall back to `main`") does **not** apply to the current code. `NodeGetKindQuery` only filters by branch when a branch is passed, and `_get_related_nodes` passes none, so the query is already branch-agnostic (verified identical in `stable` and the `infrahub-v1.9.8` tag). A node living only on `main` resolves fine from a feature branch — so a fallback would be dead code. The real trigger is an ID that resolves to **no active node at all**.

## Fix

Backend-only, no GraphQL contract change. Filter related-node entries whose `kind` could not be resolved out of the model's list accessors, so the task is returned without the unrenderable node instead of erroring. The deprecated singular `related_node`/`related_node_kind` fields stay consistent since they share the same accessor.

## Testing

- New component test `test_task_query_unresolvable_related_node` reproduces the crash (confirmed RED → GREEN).
- `backend/tests/component/graphql/queries/test_task.py` + `backend/tests/unit/task_manager/` — 18 passed.
- ruff + mypy clean.

---
## Summary by cubic
Fixes a Tasks page crash by omitting task-related nodes whose kind cannot be resolved so tasks load without GraphQL errors; backend-only with no schema changes. Fixes #9662.

- **Bug Fixes**
  - Filter out related nodes with unresolved kind in `get_related_nodes`; `get_related_nodes_as_dict` now delegates to it to keep deprecated singular fields consistent.
  - Added component tests for an unresolvable ID and a mixed case to ensure resolvable nodes and singular fields are returned correctly.

<sup>Written for commit b138ab99361a02228574d32c4200496ac4cb51e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9664?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/component/graphql/queries/test_task.py
+++ b/backend/tests/component/graphql/queries/test_task.py
@@ -803,14 +803,19 @@ async def test_task_query_unresolvable_related_node(
     default_branch: Branch,
     prefect_client: PrefectClient,
     register_core_models_schema: None,
+    tag_blue: Node,
 ) -> None:
-    """A related-node tag whose kind cannot be resolved is omitted instead of crashing the resolver."""
+    """A related-node tag whose kind cannot be resolved is omitted while resolvable ones are still returned."""
     unresolvable_id = "0177c800-0000-0000-0000-0000000000ff"
     flow = await prefect_client.create_flow_run(
         flow=dummy_flow,
         name="dummy-unresolvable-related-node",
         parameters={"firstname": "xxxx", "lastname": "yyy"},
-        tags=[TAG_NAMESPACE, WorkflowTag.RELATED_NODE.render(identifier=unresolvable_id)],
+        tags=[
+            TAG_NAMESPACE,
+            WorkflowTag.RELATED_NODE.render(identifier=tag_blue.get_id()),
+            WorkflowTag.RELATED_NODE.render(identifier=unresolvable_id),
+        ],
         state=State(type="RUNNING"),
     )
 
@@ -828,9 +833,9 @@ async def test_task_query_unresolvable_related_node(
     assert len(edges) == 1
     node = edges[0]["node"]
     assert node["id"] == str(flow.id)
-    assert node["related_nodes"] == []
-    assert node["related_node"] is None
-    assert node["related_node_kind"] is None
+    assert node["related_nodes"] == [{"id": tag_blue.get_id(), "kind": "BuiltinTag"}]
+    assert node["related_node"] == tag_blue.get_id()
+    assert node["related_node_kind"] == "BuiltinTag"
 
 
 async def test_task_no_count(
--- a/backend/tests/component/graphql/queries/test_task.py
+++ b/backend/tests/component/graphql/queries/test_task.py
@@ -798,6 +798,41 @@ async def test_task_query_progress(
     }
 
 
+async def test_task_query_unresolvable_related_node(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    prefect_client: PrefectClient,
+    register_core_models_schema: None,
+) -> None:
+    """A related-node tag whose kind cannot be resolved is omitted instead of crashing the resolver."""
+    unresolvable_id = "0177c800-0000-0000-0000-0000000000ff"
+    flow = await prefect_client.create_flow_run(
+        flow=dummy_flow,
+        name="dummy-unresolvable-related-node",
+        parameters={"firstname": "xxxx", "lastname": "yyy"},
+        tags=[TAG_NAMESPACE, WorkflowTag.RELATED_NODE.render(identifier=unresolvable_id)],
+        state=State(type="RUNNING"),
+    )
+
+    result = await run_query(
+        db=db,
+        branch=default_branch,
+        query=QUERY_TASK,
+        variables={"related_nodes": [unresolvable_id]},
+    )
+
+    assert result.errors is None
+    assert result.data
+
+    edges = result.data["InfrahubTask"]["edges"]
+    assert len(edges) == 1
+    node = edges[0]["node"]
+    assert node["id"] == str(flow.id)
+    assert node["related_nodes"] == []
+    assert node["related_node"] is None
+    assert node["related_node_kind"] is None
+
+
 async def test_task_no_count(
     db: InfrahubDatabase,
     default_branch: Branch,
--- a/backend/infrahub/task_manager/models.py
+++ b/backend/infrahub/task_manager/models.py
@@ -49,12 +49,10 @@ class RelatedNodesInfo(BaseModel):
     def get_related_nodes(self, flow_id: UUID) -> list[RelatedNodeInfo]:
         if flow_id not in self.flows or len(self.flows[flow_id].keys()) == 0:
             return []
-        return list(self.flows[flow_id].values())
+        return [item for item in self.flows[flow_id].values() if item.kind is not None]
 
     def get_related_nodes_as_dict(self, flow_id: UUID) -> list[dict[str, str | None]]:
-        if flow_id not in self.flows or len(self.flows[flow_id].keys()) == 0:
-            return []
-        return [item.model_dump() for item in list(self.flows[flow_id].values())]
+        return [item.model_dump() for item in self.get_related_nodes(flow_id=flow_id)]
 
     def get_first_related_node(self, flow_id: UUID) -> RelatedNodeInfo | None:
         if nodes := self.get_related_nodes(flow_id=flow_id):
--- a/changelog/9662.fixed.md
+++ b/changelog/9662.fixed.md
@@ -1,0 +1,1 @@
+Fixed a crash when loading the Tasks page where a task was tagged with a related node whose kind could no longer be resolved (for example a deleted definition or a stale tag). Such unresolvable related nodes are now omitted from the task instead of causing a GraphQL resolver error.


lancamat1:
### Flaky `python-lint` (`ty`) failure — not related to this PR

The `python-lint` job failed on the first run, then **passed on a rerun of the identical commit** with no code changes:

- ❌ Failed: https://github.com/opsmill/infrahub/actions/runs/27949728786/job/82703369131
- ✅ Passed (rerun): https://github.com/opsmill/infrahub/actions/runs/27949728786/job/82704931785

The failing step was `uv run ty check .` (not `ruff`). All 4 reported errors were `invalid-argument-type` at `backend/infrahub/message_bus/operations/__init__.py:33` (`await func(message=message)`), where `func = COMMAND_MAP[routing_key]` — a heterogeneous-handler dispatch this PR does not touch (the diff is confined to `task_manager`).

This is order-dependent non-determinism in `ty` 0.0.32 (pre-1.0): the union type of `func` sometimes resolves such that passing `InfrahubMessage` is flagged against the per-handler expected message types. Reproduction confirmed clean locally on the exact tree — fresh `uv sync --all-groups`, isolated `ty` cache — all report `All checks passed!`.

No action needed for this PR. Worth tracking separately as a `ty` flakiness / dispatch-typing issue.

backend/tests/component/graphql/queries/test_task.py
```diff
@@ -798,6 +798,41 @@ async def test_task_query_progress(
...
+    prefect_client: PrefectClient,
+    register_core_models_schema: None,
+) -> None:
+    """A related-node tag whose kind cannot be resolved is omitted instead of crashing the resolver."""
+    unresolvable_id = "0177c800-0000-0000-0000-0000000000ff"
+    flow = await prefect_client.create_flow_run(
+        flow=dummy_flow,
+        name="dummy-unresolvable-related-node",
+        parameters={"firstname": "xxxx", "lastname": "yyy"},
+        tags=[TAG_NAMESPACE, WorkflowTag.RELATED_NODE.render(identifier=unresolvable_id)],
```
ajtmccarty: might be worth including a valid related node in this test to confirm it is still returned

ajtmccarty (commented):
would a better fix be to allow retrieving a deleted Node so that we can correctly show its kind even if it is no longer active?

backend/tests/component/graphql/queries/test_task.py
lancamat1: Done in b138ab9 — the test now tags the flow with both a resolvable node (`tag_blue`) and the unresolvable id, and asserts the resolvable one is still returned (`related_nodes == [{id, kind: BuiltinTag}]`) while the unresolvable one is dropped. Confirms the filter is selective, not nuking the whole list.

lancamat1:
> would a better fix be to allow retrieving a deleted Node so that we can correctly show its kind even if it is no longer active?

Good instinct, but I checked this empirically and it turns out deleted nodes **already resolve** on this code path — so widening the query wouldn't change the Tasks page.

`_get_related_nodes` calls `NodeGetKindQuery` **without a branch**, so no branch/time filter is applied — it matches purely on `IS_PART_OF {status: "active"}`. And `Node.delete()` doesn't remove that active edge; it bounds it with a `to` timestamp and adds a separate `{status: "deleted"}` edge. With no time filter, the bounded active edge still matches.

Verified with a throwaway component test (create → query → `delete()` → query):

```
BEFORE delete, no branch:        {<id>: 'BuiltinTag'}
AFTER  delete, no branch:        {<id>: 'BuiltinTag'}   # still resolves
AFTER  delete, branch=default:   {}                     # branch path excludes it
random UUID,  no branch:         {}                     # no vertex at all
```

So on the branch-agnostic path a deleted node's kind comes back fine. The **only** input that yields `None` is a related-node tag whose UUID maps to no `Node` vertex at all — a stale/garbage Prefect tag (never-persisted id, hard-purged node). `include_deleted` returns nothing for that case too, since there's no vertex to read `kind` from.

That's why the fix lives at the model layer (drop entries with no resolvable kind) rather than in the query. Happy to add an `include_deleted` flag to `NodeGetKindQuery` as a separate correctness-in-depth change if you'd like the branch-scoped callers to opt in later, but it has no effect on this bug.

saltas888:
While going through the open `type/bug` backlog we found this same defect filed two more times, so merging this resolves three reports rather than one:

- #6533 (2025-05-26) "tasks view does not show up", the oldest filing, repro is delete a proposed change, delete its branch, then open Tasks
- #8342 (2026-02-09) "Task page fails to load after configuring webhooks", same crash reached by a different trigger
- #9662 (2026-06-22) this one, which names the exact `TypeError`

I've closed #6533 and #8342 as duplicates of #9662 so the `Closes` linkage here stays clean, both carry a pointer back to this PR.

Two things from the older reports worth having to hand when this gets reviewed:

- configuring webhooks is a separate way to produce the stale related-node tags, distinct from the delete-branch path, so both are worth covering in verification
- #6533's history independently supports your correction on the branch scoping, we traced `NodeGetKindQuery` the same way and reached the same conclusion, the ids fail because they have no active `IS_PART_OF` edge at all rather than because of the branch, so a fall back to `main` would be a no-op

Flagging that this is labelled `stale` and sitting at `mergeStateStatus: BLOCKED` while showing `MERGEABLE`, it's been ready since June and it closes a thread that has now been reported three times over 13 months.

ajtmccarty:
I think that the real fix for this issue is on `NodeGetKindQuery`. it looks like we want to include deleted objects at least some of the time when running this query on a branch and that seems to be the real fix instead of deleting related objects from the list. `NodeGetKindQuery` is used in a few places (5 on `stable` right now) and I am not sure if we _always_ want to include deleted objects or not, but the change to the query should be pretty simple either way

ajtmccarty (approved):

PR #9672, opened 2026-06-22
refactor(webhook): split the webhook send into its own flow

polmichel:
[IFC-2785](https://opsmill.atlassian.net/browse/IFC-2785)

Draft, stacked on `pmi-20260618-ifc-2119-resend-webhook`.

## Why make the send its own flow

The webhook send is currently an inline task inside the `webhook-process` flow, so it has no run of its own and cannot be surfaced, inspected, resent, or cancelled on its own. As its own flow, a delivery becomes a Prefect flow run with its own state, parameters, logs, and retries, which is what the resend, cancel, typing, and capture work builds on.

Roles of the two flows:

* `webhook-process` (internal): resolves the config and computes the payload once, then passes that payload to the send flow. Computing it here fixes the payload before the send flow runs, so every retry replays the same payload and a `TransformWebhook` script runs once per delivery instead of once per attempt. An internal flow is not shown in the tasks screen, so the orchestrator stays out of it.
* `webhook-send`: resolves the config and assigns the headers on each run, so the signature is rebuilt on every attempt. This is the run that retries and that gets shown and acted on.

## What changed

* `webhook_send` becomes a subflow called by `webhook_process` (was an inline task).
* `webhook_process` is marked INTERNAL, so it is not shown in the tasks screen.
* The payload is computed in the process flow and the headers in the send flow.
* The payload is passed explicitly; the transient `_payload`/`_headers` state is removed and header building is now pure.

The HTTP send itself is unchanged (same retries, same raise on HTTP error).

## Test plan

* `pytest backend/tests/functional/webhook/test_process.py`

[IFC-2785]: https://opsmill.atlassian.net/browse/IFC-2785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Split the webhook HTTP send into its own `webhook-send` subflow and made `webhook-process` internal. Each delivery is now a standalone flow run with its own state, retries, and logs while keeping payloads stable across retries.

- **Refactors**
  - `webhook-send` is now a `@flow` subflow invoked by `webhook-process`; it re-resolves config, builds headers, and posts the provided payload with the same retries/error handling; it is not registered in the workflow catalogue.
  - `webhook-process` (now INTERNAL) resolves config, computes the payload once, then calls the send subflow.
  - Added `_resolve_webhook` to fetch from cache or DB; caches for 2 hours.
  - Model API is now pure: `compute_payload(...)` returns the payload, `_build_headers(payload)` returns headers, `send_payload(payload, http_service)` posts; removed `_payload`/`_headers`, `prepare`, and `get_payload`.
  - Signatures are rebuilt on each send attempt; transforms run once per delivery.
  - Updated unit and functional tests to use the explicit payload/header API and assert on returned dicts.
  - Synced `dev/knowledge/backend/webhooks.md` to reflect the send subflow and INTERNAL workflow.

<sup>Written for commit 48817bf976f11d974ba257ff77d1099e3b86c9d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9672?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -304,12 +304,17 @@ class Webhook(BaseModel):
         await self._prepare_payload(data=data, context=context, client=client)
         self._assign_headers()
 
-    async def send(
-        self, data: dict[str, Any], context: EventContext, http_service: InfrahubHTTP, client: InfrahubClient
-    ) -> Response:
-        await self.prepare(data=data, context=context, client=client)
+    async def compute_payload(self, data: dict[str, Any], context: EventContext, client: InfrahubClient) -> Any:
+        """Build and return the payload to deliver, running any transform once."""
+        await self._prepare_payload(data=data, context=context, client=client)
+        return self._payload
+
+    async def send_payload(self, payload: Any, http_service: InfrahubHTTP) -> Response:
+        """Assign the headers for the given payload and POST it to the webhook endpoint."""
+        self._payload = payload
+        self._assign_headers()
         return await http_service.post(
-            url=self.url, json=self.get_payload(), headers=self._headers, verify=self.validate_certificates
+            url=self.url, json=payload, headers=self._headers, verify=self.validate_certificates
         )
 
     def get_payload(self) -> dict[str, Any]:
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -246,15 +246,14 @@ class Webhook(BaseModel):
     event_type: str = Field(...)
     validate_certificates: bool | None = Field(...)
     custom_headers: list[WebhookHeader] = Field(default_factory=list)
-    _payload: Any = None
-    _headers: dict[str, Any] | None = None
     shared_key: str | None = Field(default=None, description="Shared key for signing the webhook requests")
 
-    async def _prepare_payload(self, data: dict[str, Any], context: EventContext, client: InfrahubClient) -> None:  # noqa: ARG002
-        self._payload = {"data": data, **context.model_dump()}
+    async def compute_payload(self, data: dict[str, Any], context: EventContext, client: InfrahubClient) -> Any:  # noqa: ARG002
+        """Build and return the payload to deliver."""
+        return {"data": data, **context.model_dump()}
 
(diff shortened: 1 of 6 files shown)


backend/infrahub/webhook/models.py
```diff
@@ -300,20 +301,10 @@ def signing_key(self) -> str:
...
-        self, data: dict[str, Any], context: EventContext, http_service: InfrahubHTTP, client: InfrahubClient
-    ) -> Response:
-        await self.prepare(data=data, context=context, client=client)
-        return await http_service.post(
-            url=self.url, json=self.get_payload(), headers=self._headers, verify=self.validate_certificates
-        )
-
-    def get_payload(self) -> dict[str, Any]:
-        return self._payload
+    async def send_payload(self, payload: Any, http_service: InfrahubHTTP) -> Response:
```
ajtmccarty, line 304: I think the separation between `compute` and `send_payload` is a little unexpected, but I understand why you are doing it this way. maybe `async def send(..., force_recompute: bool = False)` is another option. not saying you need to change it, just something to consider

ajtmccarty (approved):

backend/infrahub/webhook/models.py
polmichel: Yes, the `self._payload` attribute was kind of making the object unusable if we did not call `compute_payload` before sending the webhook. I was not totally satisfied with that code either. I am thinking about properly splitting the responsibilities in these files. I think there is too much going on, and it's hard to know how to read or use these objects and methods.

My goal was to do it in a later PR once the retry policy works. So I'll keep it that way for this PR but I'll keep this change in mind.

PR #9676, opened 2026-06-23
feat(webhook): automatic retry policy for the webhook send

polmichel:
Closes [IFC-2779].

Draft, stacked on the webhook send/subflow split (#9672).

## What this adds

#9672 already makes `webhook-send` a flow that retries 3 times. On top of that, this PR adds a fixed back-off between those attempts: `retry_delay_seconds=120` (2 minutes), so retries are spaced 2 minutes apart instead of immediate. Each retry re-runs the flow body, which re-resolves the webhook config and re-signs the headers.

A fixed delay is used because Prefect flow retries take a single `retry_delay_seconds` value; a per-attempt schedule is only honoured for task retries.

The send itself is wrapped in a `webhook-post` task that the flow calls.

## Tests

The `immediate_webhook_retries` fixture overrides the flow delay to 0 so the failing-delivery test exhausts its retries within the suite timeout while still exercising the real retry path.

* `pytest backend/tests/functional/webhook/test_process.py`

[IFC-2779]: https://opsmill.atlassian.net/browse/IFC-2779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Adds an automatic retry policy for webhook deliveries by splitting the HTTP POST into a `webhook-post` task and retrying the `webhook-send` flow with a fixed 2-minute backoff (3 retries). Implements IFC-2779; each attempt re-resolves config and re-signs headers.

- **New Features**
  - Adds `webhook-post`; `webhook-send` now handles retries natively (retries=3, `retry_delay_seconds=120`).
  - Each retry re-invokes `webhook-post` to refresh config and signing.
  - Test fixture `immediate_webhook_retries` overrides the flow delay to 0 so failing-delivery tests exhaust retries quickly.

<sup>Written for commit 11bf64a5010f8c5652deb5a248aee2a2c93ca362. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/webhook/tasks/process.py
+++ b/backend/infrahub/webhook/tasks/process.py
@@ -28,8 +28,8 @@ WEBHOOK_MAP: dict[str, type[Webhook]] = {
 }
 
 
-# 0s, 1m, 30m, 6h
-WEBHOOK_SEND_RETRY_DELAYS: list[float] = [0, 60, 1800, 21600]
+WEBHOOK_SEND_RETRIES: int = 3
+WEBHOOK_SEND_RETRY_DELAY_SECONDS: float = 120  # fixed 2m delay between attempts
 
 
 @task(name="webhook-post", task_run_name="Send webhook {webhook_name}", cache_policy=NONE)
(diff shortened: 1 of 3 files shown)


backend/infrahub/webhook/tasks/process.py
```diff
@@ -28,14 +28,32 @@
 }
 
 
-@flow(name="webhook-send", flow_run_name="Send webhook {webhook_name}", retries=3)
-async def webhook_send(webhook_id: str, webhook_kind: str, webhook_name: str, payload: Any) -> Response:  # noqa: ARG001
-    """Resolve the webhook config, assign its headers, and POST the prepared payload. Retries up to 3 times."""
-    log = get_run_logger()
+WEBHOOK_SEND_RETRIES: int = 3
+WEBHOOK_SEND_RETRY_DELAY_SECONDS: float = 120  # fixed 2m delay between attempts
```
ajtmccarty, line 32: I don't know if this is something we'd want to expose to the user, but it we do then the test could use the config instead of monkeypatch

ajtmccarty (approved):

backend/infrahub/webhook/tasks/process.py
polmichel: The fixture doesn't patch the constants — it swaps `webhook_send` for a `.with_options(retry_delay_seconds=0)` variant, so the test still exercises all 3 retries but drops the 2-minute back-off (otherwise a failing-delivery test sleeps through ~4 minutes of real delay).

Reading the delay from config would be the cleaner path *if* we expose retry tuning to users, but IFC-2779 treats this as a fixed policy, so there's no config to read today. Happy to switch to a config-driven value if we decide retry count/delay should be user-configurable.

polmichel: This is a good point. I am not sure a customer would need it right now, so I would keep it that way for now. That said, I'll mention this to the product team to make sure I am not missing something here.

And sorry for the AI-generated comment, it was not meant to be posted, I've mixed up shells.

PR #9680, opened 2026-06-23
fix(auth): narrow SSO group create-race handling to uniqueness violations

polmichel:
Addresses cubic feedback on #9677 about the SSO auto-create group flow's exception handling.

## Behavior

The auto-create SSO groups flow reuses an existing group when a create loses the find-or-create race against a concurrent creation of the same name. The reuse path triggers on a uniqueness collision only; a creation error for any other reason propagates to the caller.

## Change

- Add `UniquenessViolationError(ValidationError)` in `exceptions.py`, with `HFIDViolatedError` reparented under it (an HFID violation is a uniqueness violation).
- The grouped-uniqueness constraint raises `UniquenessViolationError` on a standard-constraint violation; the HFID path raises the now-subclassed `HFIDViolatedError`.
- The auto-create flow catches `UniquenessViolationError`.

Because `UniquenessViolationError` subclasses `ValidationError`, every existing `except ValidationError` / `pytest.raises(ValidationError)` behaves identically. The uniqueness message is an unstructured string, so the GraphQL mutation error path — which reclassifies only a bare `ValidationError` — re-raises it unchanged, keeping the same 422 and message.

## Tests

Covered by existing component tests run in CI:

- `backend/tests/component/auth/auth_groups/test_autocreate_race_reuse.py` — exercises the create-race reuse path via the typed exception.
- `backend/tests/component/core/constraint_validators/test_node_uniqueness.py` — asserts `pytest.raises(ValidationError)`, satisfied by the subclass.

Not run locally (need a Neo4j stack); validated `ruff`, `ruff format`, and `ty` on the changed files.

### DIFF
--- a/backend/infrahub/auth/auth_groups/service.py
+++ b/backend/infrahub/auth/auth_groups/service.py
@@ -27,7 +27,7 @@ from infrahub.core.constants import InfrahubKind
 from infrahub.core.manager import NodeManager
 from infrahub.core.node.create import create_node
 from infrahub.core.protocols import CoreAccount, CoreAccountGroup
-from infrahub.exceptions import ValidationError
+from infrahub.exceptions import UniquenessViolationError
 from infrahub.log import get_logger
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


ajtmccarty (approved):
is it possible to add a test that would raise a non-uniqueness ValidationError during SSO group creation to validate that it is raised instead of caught? not sure if this is possible if everything is working correctly, so don't waste your time on it if it is not necessary

polmichel:
> is it possible to add a test that would raise a non-uniqueness ValidationError during SSO group creation to validate that it is raised instead of caught? not sure if this is possible if everything is working correctly, so don't waste your time on it if it is not necessary

With the current dependencies, it would be kind of difficult to create this test

PR #9705, opened 2026-06-24
fix: re-validate proposed changes at merge time

### LINKED ISSUE #9704: bug: can create duplicate objects via proposed change
### Component

API Server / GraphQL

### Infrahub version

1.10.0

### Current Behavior

it is possible to create duplicate objects that violate a schemas uniqueness constraint using 2 proposed changes.

### Expected Behavior

duplicate objects should never be created

### Steps to Reproduce

1. create branches `br1` and `br2`
2. create a new `BuiltinTag` with the same name on both branches
3. create a proposed change for each branch
4. merge one proposed change
5. merge the other proposed change

now you have 2 tags with the same name, which is illegal

### Additional Information

_No response_

ajtmccarty:
Closes #9704

## Problem

Merging two branches that each introduce the same unique value (e.g. a `BuiltinTag` named `orange`) can leave **duplicate values** on the destination branch.

A proposed change validates its schema and data integrity during its check pipeline, but those run once (point-in-time) and the merge relies on the **stored** validator conclusions. So:

1. PC for branch A and PC for branch B both run their checks while `main` is clean → both pass.
2. A is merged → `main` now has the value.
3. B is merged → it reuses its stale "passing" schema-integrity conclusion and merges anyway → duplicate.

This is a time-of-check/time-of-use gap. It exists for **constraints** (uniqueness, cardinality, …) specifically because a constraint is not a diff conflict — nothing re-detects it after another branch merges. (Data conflicts do not have this problem; see below.)

## Fix

Move constraint validation **into the shared merge flow** so it runs against the current state of the destination branch and gates **both** the proposed-change merge and the direct `BranchMerge` mutation.

1. **`MergeConstraintValidator`** (`backend/infrahub/core/merge/constraints.py`) — a component constructed with `(db, branch, diff_repository)` and injected into `BranchMerger`. Its single entry point validates the union of data-diff-derived constraints (the determiner over `get_node_field_summaries`, which is relationship-correct) and schema-diff-derived constraints, runs `schema_validate_migrations`, and returns the surviving violations.
2. **Conflict-aware suppression** — a violation on a `(node, field)` that has a *resolved* diff conflict is dropped. That field is reconciled by the conflict resolution applied during the merge, so flagging its pre-resolution cross-branch state would be a false positive (this is what made the earlier attempts misfire on cardinality-one conflicts — see below).
3. **`BranchMerger.merge()`** runs two gates *before* mutating the graph (so a rejection needs no rollback): unresolved conflicts → `MergeConflictsUnresolvedError`; constraint violations → `MergeConstraintsViolatedError`. Both subclass `ValidationError` (HTTP 422).
4. **`merge_proposed_change`** records a `MergeConstraintsViolatedError` on the **Schema Integrity** validator so the proposed change reflects the failure (instead of staying green while the merge is rejected), then returns a failed state. `merge_branch_mutation` drops its now-redundant inline determiner validation — the merge flow covers it.
5. **Relationship constraints in the pipeline check** — `_get_proposed_change_schema_integrity_constraints` compared the SDK diff-summary `element_type` (`"RELATIONSHIP_ONE"`) against `DiffElementType` **values** (`"RelationshipOne"`) instead of member **names**, so it silently dropped every relationship element. It now compares against member names. This keeps the proposed-change UI feedback accurate; the merge gate itself does not depend on it (it uses the backend diff summary).

## Data conflicts are already handled (and unaffected)

The symmetric data case — branch B's PC is green, branch A merges a conflicting change to the same node/field, B now conflicts — is already handled by the existing merge logic, which updates a diff to the latest timestamp before merging. The diff update will identify any new conflicts and mark them as data integrity violations on the linked proposed change, if it exists.

## Approaches considered and superseded

- **Inline validation in `BranchMerger.merge` without conflict-awareness** (first attempt, reverted): fixed the TOCTOU but false-positived on cardinality-one relationship conflicts — the relationship-`count` check counted both peers (`2 > max_count 1`) *before* the merge applied the resolution. The conflict-aware suppression in this PR is the correct version of that idea.
- **Re-running the proposed-change integrity checks at merge** (intermediate, superseded by this PR): re-executed two of the pipeline's checks out of band right before merging. It worked for uniqueness but was a bolt-on, only covered the proposed-change path (not the direct mutation), and its data-side half was both redundant (the post-merge diff refresh already handles data) and ineffective (the synchronizer short-circuits on a metadata diff with an existing validator). Validating in the shared merge flow is a single gate that covers both entry points.

## Testing

`backend/tests/integration/proposed_change/test_proposed_change_uniqueness_merge.py`:

- `TestProposedChangeUniquenessMerge` — the reported bug via the proposed-change path: the second merge is blocked, `main` keeps a single `orange` tag, and the Schema Integrity validator is `FAILURE`.
- `TestDirectMergeUniqueness` — the same collision via the direct `BranchMerge` mutation, proving the unified gate covers that path.
- `TestProposedChangeDataConflictMerge` — cross-branch data conflict (default config): blocked by the stored data-check gate, Data Integrity validator `FAILURE`, `main` unchanged.
- `TestProposedChangeDataConflictMergeBackstop` — same, with `diff_update_after_merge=False`: blocked by the merge-time conflict backstop.

Also covered: `test_proposed_change_cardinality_one_conflict` (a resolved cardinality-one conflict stays mergeable), `test_schema_integrity_relationship_constraints` (regression guard for the `element_type` fix), and `test_merge_relationship_count_validation` (cross-branch cardinality-one never leaves two peers on `main`).

Regression suites run locally (all green): `test_diff_update` (incl. the cardinality-one case that broke the first attempt), `test_branch_merge`, `test_coordinator_lock`.

## Notes / follow-ups

- A constraint violation surfaces the generic `Unable to merge proposed change containing failing checks`; the detailed violation lives on the Schema Integrity validator's check.
- **Separate, pre-existing gap (not addressed here):** constraint validation runs against the *merging branch's* data, but a data-conflict resolution can retain the *destination's* value, which is never re-validated against the newly-merged schema. So tightening a constraint on one branch while the value is resolved to the other side can let illegal data through. Worth a dedicated issue — the proper fix is validating the post-resolution state. [IFC-2838](https://opsmill.atlassian.net/browse/IFC-2838)
- Orphaning a mandatory cardinality-one relationship by deleting its peer node is not caught by the relationship-count constraint (the change isn't attributed to the dependent node in the diff). Pre-existing, separate gap. [IFC-2837](https://opsmill.atlassian.net/browse/IFC-2837)

[IFC-2837]: https://opsmill.atlassian.net/browse/IFC-2837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[IFC-2838]: https://opsmill.atlassian.net/browse/IFC-2838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/core/merge/constraints.py
```diff
@@ -0,0 +1,88 @@
...
+
+    async def _get_conflicted_fields(self) -> set[tuple[str, str]]:
+        conflicted_fields: set[tuple[str, str]] = set()
+        async for conflict_path, _conflict in self.diff_repository.get_all_conflicts_for_diff(
+            diff_branch_name=self.branch.name, tracking_id=BranchTrackingId(name=self.branch.name)
+        ):
+            # path_identifier is "data/<node_uuid>/<field>/...".
+            path_parts = conflict_path.split("/")
+            if len(path_parts) >= 3 and path_parts[0] == "data":
+                conflicted_fields.add((path_parts[1], path_parts[2]))
```
polmichel: Just a thought, not mandatory and should not block this PR.

This code looks like specific logic that should be encapsulated in a dedicated method that makes the intent explicit. And as it happens, this block is duplicated (see `infrahub.proposed_change.tasks.run_proposed_change_schema_integrity_check`). That could lead to asymmetrical behavior on both ends if we ever need to fix something or add capabilities to this block.

backend/tests/integration/diff/test_merge_relationship_count_validation.py
```diff
@@ -0,0 +1,91 @@
...
+        await car_on_main.get_relationship("previous_owner").update(db=db, data={"id": initial_dataset["sarah_id"]})
+        await car_on_main.save(db=db)
+
+        with contextlib.suppress(GraphQLError):
+            await client.branch.merge(branch_name=branch.name)
+
+        merged_car = await NodeManager.get_one(db=db, id=initial_dataset["car_id"], branch=default_branch)
+        assert merged_car
+        previous_owners = await merged_car.get_relationship("previous_owner").get_relationships(db=db)
+        # TODO: should assert the actual expected peer
```
polmichel: is this intentionally kept for a later?

backend/infrahub/proposed_change/tasks.py
```diff
@@ -504,10 +558,29 @@ async def run_proposed_change_schema_integrity_check(model: RequestProposedChang
         )
     )
 
+    # A node/field with a diff conflict is reconciled by the data-conflict resolution applied during
+    # the merge. Validating its pre-resolution cross-branch state here would report a spurious
+    # (and non-resolvable) schema violation, so those fields are skipped.
```
polmichel: Not blocking at all, but the comment fragment
`Validating its pre-resolution cross-branch state here would report a spurious (and non-resolvable) schema violation, so those fields are skipped.`
is fairly far from the condition check it refers to, which is lower down in the loop. I'd say it's more maintainable to keep comments as close as possible to the code they refer to. One possible effect of this is a desync between code and comments.

backend/tests/integration/proposed_change/test_proposed_change_cardinality_one_conflict.py
```diff
@@ -0,0 +1,108 @@
...
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        initial_dataset: dict[str, str],
+        client: InfrahubClient,
+    ) -> None:
+        branch = await client.branch.create(branch_name="rival_manufacturers")
+        car_on_branch = await NodeManager.get_one(db=db, id=initial_dataset["car_id"], branch=branch.name)
+        assert car_on_branch
+        # TODO: use get_relationship("manufacturer") to clear typing ignore
+        await car_on_branch.manufacturer.update(db=db, data={"id": initial_dataset["omnicorp_id"]})  # type: ignore[attr-defined]
```
polmichel: just highlighting in case this is not intentionally left here, not blocking

backend/tests/integration/proposed_change/test_proposed_change_uniqueness_merge.py
```diff
@@ -0,0 +1,388 @@
...
+        await asyncio.sleep(1)
+    raise AssertionError(f"Validators for proposed change {proposed_change_id} did not all succeed")
+
+
+async def _count_main_tags(db: InfrahubDatabase, branch: Branch, name: str) -> int:
+    tags = await NodeManager.query(db=db, schema=InfrahubKind.TAG, filters={"name__value": name}, branch=branch)
+    return len(tags)
+
+
+class TestProposedChangeUniquenessMerge(TestInfrahubApp):
```
polmichel, line 63: Nice test suite

backend/tests/integration/proposed_change/test_proposed_change_uniqueness_merge.py
```diff
@@ -0,0 +1,388 @@
...
+            kind=SdkCoreProposedChange,
+            data={"source_branch": branch_two, "destination_branch": "main", "name": "backstop-desc-two"},
+        )
+        await pc_two.save()
+        await _wait_for_validators_success(db=db, proposed_change_id=pc_two.id)
+
+        # Merge the first branch: main now holds "from_one". No diff refresh is triggered for pc_two.
+        pc_one.state.value = ProposedChangeState.MERGED.value
+        await pc_one.save()
+        for _ in range(PREFECT_EVENT_WAIT_SECONDS):
```
polmichel, line 375: Just a thought.
I totally understand why this is here, and why all the other identical blocks are too, but in your opinion, is there a possibility that we could (in the future) downgrade these tests to component tests and use a fake Prefect adapter? 
Or do you think we should keep a real Prefect client in the loop?

My take on this one, without being an expert on the merge subject, would be to keep one integration test scenario and downgrade all the others to component tests and faking the Prefect client.

backend/infrahub/core/merge/branch_merger.py
```diff
@@ -49,12 +56,14 @@ def __init__(
         self.diff_merger = diff_merger
         self.diff_repository = diff_repository
         self.diff_locker = diff_locker
+        self.constraint_validator = constraint_validator
```
polmichel, line 59: Very clean ! new architecture pays off

polmichel (approved):
LGTM, nice fix and it really pleasing to see how the `MergeConstraintValidator` just gets injected into the `BranchMerger` thanks to your new architecture around the merge process.

I've put optional comments and a few thoughts, nothing blocking from a correctness perspective

backend/infrahub/core/merge/constraints.py
ajtmccarty: very good point. this has been refactored to consolidate the conflict-parsing logic

backend/tests/integration/diff/test_merge_relationship_count_validation.py
ajtmccarty: it was not. it's been addressed

backend/infrahub/proposed_change/tasks.py
ajtmccarty: moved to where the conflict-parsing logic was consolidated

backend/tests/integration/proposed_change/test_proposed_change_cardinality_one_conflict.py
ajtmccarty: thanks. another leftover. addressed

backend/tests/integration/proposed_change/test_proposed_change_uniqueness_merge.py
ajtmccarty: yes I think this is a good approach

PR #9714, opened 2026-06-26
feat(generators): add dependency-closure foundation for precise regeneration

ogenstad:
# Why

Generators re-run on every file change in any linked repository today - the same
blunt gate INFP-409 already replaced for artifacts. An unrelated commit (e.g. a
README edit) re-runs every generator across large target groups.

**Goal:** lay the shared foundation so generator triggering can later be driven
by precise per-generator dependency closures, reusing the machinery already
shipped for artifacts.

## What changed

**Behavioral changes**
- Generator git-import now builds and persists a dependency closure on each
  `CoreGeneratorDefinition` (`dependencies` / `dependencies_complete`).
- A content change that alters a generator's closure now triggers a node update
  on re-import (previously such a change could be missed).
- Generator *run* triggering is otherwise unchanged.

**Implementation notes**
- Added nullable `dependencies` / `dependencies_complete` attributes to
  `CoreGeneratorDefinition`, mirroring `CoreTransformation`; regenerated
  protocols, schema, GraphQL/OpenAPI exports, frontend types, and the gql.tada
  cache.
- Widened the closure builder (`PythonClosure.supports` + the `TransformConfig`
  union) to accept `InfrahubGeneratorDefinitionConfig`.
- Introduced the `RegenerationDefinition` structural Protocol and retyped the
  three regeneration predicates to it; parametrized their diagnostic wording via
  `source_noun` / `instance_noun` so both artifact and generator callers read
  correctly.
- Added a required `query_id` to `ProposedChangeGeneratorDefinition`, populated
  at every construction site.

---
## Summary by cubic
Adds dependency-closure capture and storage for generator definitions to enable precise regeneration later. No behavior change yet; generators still use legacy gates and artifact selection/logs remain identical (FR-013), aligning with IFC-2809.

- **New Features**
  - Store `dependencies` and `dependencies_complete` on generator definitions; GraphQL schema and frontend types updated.
  - Build and persist generator dependency closures at import time; update nodes when closures change.
  - Extend the Python closure builder to support `InfrahubGeneratorDefinitionConfig`; introduce a shared `RegenerationDefinition` protocol with noun-aware predicate messages applied to artifacts and generators.
  - Carry `query_id`, `dependencies`, and `dependencies_complete` through generator pipeline models and message bus; pass `query_id` via generator tasks and GraphQL mutation.
  - Add integration test for generator closure import and unit tests for Python closure; confirm artifact regression suite remains unchanged (FR-013).

<sup>Written for commit 4a2696bdf420c15226a6788f8c25108e11a4dfc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/git/integrator.py
```diff
@@ -1079,6 +1112,8 @@ async def _generator_requires_update(
             or existing_generator.targets.id != generator.targets
             or existing_generator.execute_in_proposed_change.value != generator.execute_in_proposed_change
             or existing_generator.execute_after_merge.value != generator.execute_after_merge
+            or existing_generator.dependencies.value != list(closure.dependencies)
```
saltas888, line 1115: @ogenstad Does the order matter here? `[1, 2] == [2, 1] is False`. If not then we could do
`set(existing_generator.dependencies.value) != set(closure.dependencies)`

ogenstad: It doesn't matter today, but I'd still like to keep it. For the next phase of this (which might come in 1.11 depending on exact implementation of other efforts) is that we want to store a hash/fingerprint of all Transforms, Generators and Artifact Definitions (which will be the transform fingerprint + that of the artifact definition). In that scenario we'd want the dependencies to be in a consistent ordered list so that the final fingerprint is always the same. The fingerprint of a transform would be fingerprint/hash of the GraphQLQuery, the base transform code and any of the files within the dependencies.

gmazoyer:
One thing before merge: the new generator integrator paths (`_build_generator_definitions`, `_generator_requires_update`'s dependency comparison, the closure writes) have no direct test. That comparison is the load-bearing bit of this PR. Could T026/T027 come forward into this PR, or is there a reason to defer them?

ogenstad:
> One thing before merge: the new generator integrator paths (`_build_generator_definitions`, `_generator_requires_update`'s dependency comparison, the closure writes) have no direct test. That comparison is the load-bearing bit of this PR. Could T026/T027 come forward into this PR, or is there a reason to defer them?

No, mainly stuck to the phases in order to keep each PR smallish. But that makes sense, I've added some tests.

saltas888 (approved):

backend/infrahub/git/integrator.py
```diff
@@ -161,6 +162,18 @@ class TransformPythonInformation(BaseModel):
...
+@dataclass
+class GeneratorDefinitionInformation:
+    """A generator definition read from disk, paired with the dependency closure computed for it.
+
+    The closure is computed at build time while the worktree is in scope and travels with its
+    config so the downstream apply step never has to recompute it or look it up by name.
+    """
+
+    config: InfrahubGeneratorDefinitionConfig
+    closure: ClosureResult
```
gmazoyer, line 174: Very minor comment here: this is a `@dataclass` while its `*Information` siblings (`TransformPythonInformation`, `CheckDefinitionInformation`) are pydantic `BaseModel`s. Fine as-is given it's just a config+closure pairing, but a different name would signal the difference. What name though heh :D

gmazoyer (approved):

PR #9696, opened 2026-06-24
docs(webhook): webhook delivery operability specification

polmichel:
## Summary

Adds the spec-kit specification package for **Webhook Delivery Operability**, the design umbrella over the in-flight webhook send-flow work. Specification only — no production code in this PR.

**Phase:** specification & planning (spec-kit) — no production code; the implementation lands in follow-up PRs.

## Prerequisite work (out of specification scope)

The structural send-flow work this specification builds on is tracked in its own PRs and is **not** part of this spec's scope:

- opsmill/infrahub#9672 — phase: split the webhook send into its own `webhook_send` subflow (the structural foundation the operability layer sits on).
- opsmill/infrahub#9676 — phase: automatic fixed-delay retry policy for the send (this PR's base branch).
- https://github.com/opsmill/infrahub/pull/9613 — independent bug fix.

`spec.md` and `plan.md` now carry an **Implementation Sync — 2026-06-26** revision recording this landed foundation against the requirements it satisfies (FR-001, FR-012a, FR-027) and marking the operability layer and the still-outstanding foundation (`WEBHOOK_SEND` catalogue registration, transient-only retry gating) as not yet implemented. `tasks.md` is unchanged.

## Scope it covers

Builds on the webhook send-flow split to make a delivery a first-class, inspectable, recoverable object in the Tasks tab:

- Inspect a delivery's request/response + payload (secrets redacted at capture)
- Classify failures into clean reasons with remediation hints; fixed-delay transient-only retry
- Generic, task-id-addressable resend (any terminal state) and cancel (non-terminal)
- Polymorphic task typing mirroring the events GraphQL type derivation — no new Neo4j node, no migration

Related: IFC-2755, IFC-2119, IFC-2753, IFC-2754.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9718, opened 2026-06-26
feat(webhook): classify delivery failures with clean, user-facing reasons

polmichel:
Stacked on #9696 (`pmi-20260624-webhook-operability-spec`). Implements the failure-classification slice of the webhook delivery operability spec.

## What
- Header resolution now fails with a clear configuration error when a header value cannot be resolved.
- `WebhookFailureClassifier` maps a delivery failure to a stable class (`CONFIG`/`CONNECTION`/`TLS`/`TIMEOUT`/`HTTP_CLIENT_ERROR`/`HTTP_SERVER_ERROR`/`UNKNOWN`) with a clean message and a remediation hint.
- Classified failures settle the run into a clean failed state with no stacktrace; an `UNKNOWN` error keeps its traceback so it surfaces as a genuine crash.
- Independent fix (own commit + changelog): the HTTP adapter now recognizes a TLS verification error that httpx wraps as a transport error, so any outbound HTTPS caller reports it as a TLS error rather than a generic connection error.

## How to review
Each commit is atomic and self-contained and they are ordered by dependency, so reviewing commit-by-commit is the easiest path. The TLS fix (3) is independent of the feature.

| # | Commit | Intent / what to review |
|---|--------|-------------------------|
| 1 | [`eebfda8`](https://github.com/opsmill/infrahub/commit/eebfda8df866fb7744505ad625731121f9bc5848) feat(webhook): fail header resolution | A configured header that cannot be resolved now fails the delivery with a clear config error instead of being dropped silently. `webhook/models.py` + model/header unit tests. |
| 2 | [`3631abd`](https://github.com/opsmill/infrahub/commit/3631abd050b7dcd888a50b74b8b96ac54b19b045) feat(webhook): classify delivery failures | The pure `WebhookFailureClassifier`: failure → stable class + message + remediation + `transient`. Standalone, not yet wired. `webhook/classifier.py` + unit tests. |
| 3 | [`5990455`](https://github.com/opsmill/infrahub/commit/5990455ef0b0db43e6725b0f2b9b3bf641443099) fix(http): wrapped TLS error → SSL failure | Independent HTTP-adapter fix: detect a TLS verification error httpx wraps as a transport error. Benefits every HTTPS caller; ships its own changelog. `services/adapters/http/httpx.py` + unit/integration tests. |
| 4 | [`ba9c67c`](https://github.com/opsmill/infrahub/commit/ba9c67cdfa56f59ea5c7ec5c59b735435ae8fb1e) feat(webhook): classify and log the outcome | Wire the classifier into the send flow: classify, log the outcome, settle classified failures into a clean failed state (no stacktrace), `UNKNOWN` crashes. `webhook/tasks/process.py` + functional tests. |
| 5 | [`cf21b12`](https://github.com/opsmill/infrahub/commit/cf21b12343b42fcf48d75bd2d446172376cf30aa) docs(webhook): document failure handling | Knowledge-doc Failure handling section + spec task status (T025 done, T026 descoped, T001 pending). Docs only. |

Three follow-up commits address PR-review feedback.

## Notes
- Only the standalone TLS fix ships a changelog fragment.

## Testing
Unit, integration (real TLS / unresponsive / unreachable endpoints), and functional suites pass.

Verified end-to-end against a running Infrahub instance: six webhooks, each pointed at a controllable target failing in one specific way, were fired by a single node event. Every delivery ended in a clean failed task carrying the classified reason and remediation.

| Live scenario | Target | Class | Reason shown in the task |
|---------------|--------|-------|--------------------------|
| Unreachable endpoint | closed port | `CONNECTION` | Unknown http error when connecting to … Verify the target endpoint is reachable from Infrahub. |
| Untrusted certificate | self-signed HTTPS | `TLS` | Unable to validate TLS certificate for connection to … Check the endpoint's TLS certificate … |
| Slow target | endpoint that stalls past the 10s timeout | `TIMEOUT` | Connection to … timed out after 10. The target did not respond in time … |
| Client error | endpoint returning 404 | `HTTP_CLIENT_ERROR` | The target responded with HTTP 404. The target rejected the request; check the URL and authentication. |
| Server error | endpoint returning 500 | `HTTP_SERVER_ERROR` | The target responded with HTTP 500. The target returned a server error … |
| Misconfiguration | env-sourced header with an unset variable | `CONFIG` | Webhook '…': could not resolve header '…'. Check the webhook's headers … |

`TLS` was confirmed to classify as TLS end-to-end (not as a generic connection error), validating the adapter fix.


backend/infrahub/services/adapters/http/httpx.py
```diff
@@ -92,15 +112,20 @@ async def _request(
...
                 ) from exc
             except httpx.RequestError as exc:
+                ssl_error = self._ssl_error_extractor.extract(exc)
+                if ssl_error is not None:
+                    log.info(f"TLS error for connection to {url}")
+                    if isinstance(ssl_error, ssl.SSLCertVerificationError):
+                        message = f"Unable to validate TLS certificate for connection to {url}"
+                    else:
+                        message = f"TLS error when connecting to {url}"
+                    raise HTTPServerSSLError(message=message) from exc
```
ogenstad, line 125: Does this mean that there were situations where an SSL error was buried within a httpx.RequestError that wasn't previously caught as an SSL error?

backend/infrahub/webhook/tasks/process.py
```diff
@@ -49,12 +53,35 @@ async def webhook_post(webhook_id: str, webhook_kind: str, webhook_name: str, pa
...
-    response = await webhook_post(
-        webhook_id=webhook_id, webhook_kind=webhook_kind, webhook_name=webhook_name, payload=payload
-    )
-    log.info(f"Successfully sent webhook to {response.url} with status {response.status_code}")
+    started = time.monotonic()
+    try:
+        response = await webhook_post(
+            webhook_id=webhook_id, webhook_kind=webhook_kind, webhook_name=webhook_name, payload=payload
+        )
+    except Exception as cause:
```
ogenstad: At some point we should try to phase these out instead as it catches everything even `Ctr+C` or terminate requests when something wants to gracefully shutdown a docker container etc.

backend/infrahub/webhook/classifier.py
```diff
@@ -0,0 +1,89 @@
...
+    CONNECTION = "CONNECTION"
+    TLS = "TLS"
+    TIMEOUT = "TIMEOUT"
+    HTTP_CLIENT_ERROR = "HTTP_CLIENT_ERROR"
+    HTTP_SERVER_ERROR = "HTTP_SERVER_ERROR"
+    UNKNOWN = "UNKNOWN"
+
+    @property
+    def remediation(self) -> str:
+        return _REMEDIATION[self]
```
ogenstad: I'd suggest to change this part instead so that we have a match statement and end with an `assert_never` that is to avoid any future KeyError exceptions if we ever add a new `StatusClass` entry but forget to update the disconnected dictionary.

backend/infrahub/webhook/models.py
```diff
@@ -270,7 +282,9 @@ def _build_headers(self, payload: Any, uuid: UUID | None = None, at: Timestamp |
             try:
                 headers[header.key] = header.resolve()
             except WebhookHeaderResolutionError as exc:
-                logger.warning("Webhook '%s': %s, skipping header '%s'", self.name, exc, header.key)
+                raise WebhookHeaderResolutionError(
+                    f"Webhook '{self.name}': could not resolve header '{header.key}': {exc}"
+                ) from exc
```
ogenstad, line 287: It looks a bit weird that we'd just re-raise the same error but I guess we're enriching it with the webhook name and info that might isn't available from `header.resolve()`? Not saying that we have to change it, just wondering about the approach.

backend/infrahub/workers/dependencies.py
```diff
@@ -157,7 +157,9 @@ def get_tls_registry(tls_registry: TlsContextRegistry = Depends(build_tls_regist
 
 def build_http_service() -> InfrahubHTTP:
     if "http_service" not in _singletons:
-        _singletons["http_service"] = HttpxAdapter(tls_registry=build_tls_registry())
+        _singletons["http_service"] = HttpxAdapter(
+            tls_registry=build_tls_registry(), ssl_error_extractor=SSLErrorExtractor()
```
ogenstad: I'm wondering if this external `SSLErrorExtractor` adds anything of value or if the extractor should just be a private method on the base HttpAdapter?

backend/tests/helpers/http_server.py
```diff
@@ -0,0 +1,144 @@
...
+    sock.close()
+    return port
+
+
+def write_self_signed_cert(directory: Path) -> tuple[str, str]:
+    """Write a short-lived, self-signed certificate for 127.0.0.1 that no client should trust.
+
+    The address is set in the Subject Alternative Name so the only thing wrong with the
+    certificate is that it is self-signed, not a hostname mismatch. The validity starts a
+    minute in the past to tolerate clock skew.
```
ogenstad: If we don't expect the client to trust the cert perhaps we can use a static one that has expired as well? I.e to avoid having to spend the CPU cycles to generate the cert. At some point it would be nice to capture the reason for a failure though so that we can see that a cert isn't trusted at all vs if it's has expired..

ogenstad (approved):
Added some comments but overall LGTM.

backend/infrahub/workers/dependencies.py
polmichel: I think you're right.
The initial idea was that, if at some point we need to add other logic capabilities to this error extractor, we could reuse the existing object and extend the principle to other error extractors. But I guess this is maybe over-designed right now. If the need comes up again, I'll reconsider it. For now I'll remove it.

polmichel: Something else came once again to my mind, and this was actually the initial idea: creating a private method that does the same thing prevents us from writing a clean unit test for this specific behavior. The alternative is to test a private method, which is often a signal that we should extract the behavior into a dedicated public API.
I've encountered the same issue while trying to extract this in a private method, this is what made me remember this.

ogenstad: I think this is a case where the code gets more complex due to requirements of testing which can be fine in some situations. An alternative could be to make the method public, however this also doesn't make sense since there's no other point of exposing a method like that. Another alternative is to add a public method on whatever the unittest is using and have that method call this private method. That way the complexity lives at the test level that is if we ever need to overwrite it to be something else. Regardless this is not a big deal for me, it just feels like a workaround due to typing that we need to care about in several places in the production code..

backend/infrahub/services/adapters/http/httpx.py
polmichel: I was also surprised, I've made Claude create a test to reproduce the issue `tests.integration.services.adapters.http.test_httpx.test_failed_handshake_buries_the_ssl_error_in_a_transport_error`. Related commit will pushed once I've resolved all review feedback

backend/infrahub/webhook/tasks/process.py
polmichel: Thanks for the feedback, I'll make a list of expected errors. Other ones will just directly be raised.

backend/infrahub/webhook/models.py
polmichel: It adds additional information yes and it also categorizes all errors that aren't specifically tied to the `WebhookHeaderResolution` step under this process, since they occurred during webhook header resolution. This allows us to group all these different errors into one domain-related category. We can catch this new error type if we want to handle webhook header resolution errors.

backend/tests/helpers/http_server.py
polmichel: I'll push a file containing the frozen expired cert

backend/infrahub/webhook/models.py
polmichel: It not the cleanest design, I guess not sure how to manage it properly. Maybe something to discuss in depth at some backend meetings.

backend/infrahub/workers/dependencies.py
polmichel: Right, I'll test directly the private method

backend/infrahub/services/adapters/http/httpx.py
polmichel: pushed https://github.com/opsmill/infrahub/pull/9718/changes/c46a146ce87f856fac46b86cd193586dbac95400

backend/infrahub/webhook/tasks/process.py
ajtmccarty: [SystemExit](https://docs.python.org/3/library/exceptions.html#SystemExit) and [KeyboardInterrupt](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt) inherit from `BaseException`, so I think those should remain uncaught. unless `webhook_post` or something it calls is catching them and reraising as subclasses of plain `Exception`

PR #9721, opened 2026-06-26
[FIX] make branch names containing a slash clickable (closes #9716)

### LINKED ISSUE #9716: bug: branches with a slash in their name are not clickable on the /branches page
### Component

Frontend UI

### Infrahub version

1.10.0 (reproduced on https://sandbox.infrahub.app, latest `stable`)

### Current Behavior

On the **Branches** page (`/branches`), a branch whose name contains a forward slash (e.g. `feature/my-branch`) is effectively not clickable — clicking it does **not** open that branch's detail page. Instead of landing on the branch detail view for `feature/my-branch`, navigation silently fails / redirects to a non-existent branch.

This affects every place that links to a branch detail page via a slashed name: the branch list items, the branches table name cell, and the branch detail tabs.

### Expected Behavior

Clicking a branch whose name contains a slash should open that branch's detail page (`/branches/<branch name>`), exactly like branches without a slash.

### Steps to Reproduce

1. Create a branch with a `/` in its name, e.g. `feature/my-branch`.
2. Go to the **Branches** page (`/branches`).
3. Click the `feature/my-branch` entry in the list (or its name in the branches table).
4. Observe that the branch detail page for `feature/my-branch` does not load — you stay on / are redirected away from the intended branch instead of seeing its details.

### Additional Information

**Root cause**

The branch detail route uses a single dynamic segment, and the URL is built without encoding the branch name. React Router treats the unencoded `/` in the name as a path separator, so the segment never matches the full branch name.

1. **URL is built without encoding** — `getBranchDetailsUrl` interpolates the raw branch name straight into the path:

   `frontend/app/src/entities/branches/utils.ts:7-14`
   ```ts
   export function getBranchDetailsUrl(
     branchName: string,
     tab?: BranchDetailsTab,
     overrideParams?: overrideQueryParams[]
   ): string {
     const path = tab ? `/branches/${branchName}/${tab}` : `/branches/${branchName}`;
     return constructPath(path, overrideParams);
   }
   ```
   For `feature/my-branch` this produces `/branches/feature/my-branch` (the `/` is not encoded).

2. **Route matches a single segment** — the detail route is defined as `:branchName` with child tab routes:

   `frontend/app/src/app/router.tsx:58-89`
   ```tsx
   {
     path: ":branchName",
     lazy: () => import("@/pages/branches/details"),
     children: [
       { index: true, lazy: () => import("@/pages/branches/branch-details/details-tab") },
       { path: "data", /* ... */ },
       { path: "files", /* ... */ },
       { path: "artifacts", /* ... */ },
       { path: "schema", /* ... */ },
       {
         // Redirect /branches/:branchName/<unknown> back to the branch's index tab.
         path: "*",
         element: <Navigate to="." replace />,
       },
     ],
   }
   ```
   For `/branches/feature/my-branch`, React Router matches `:branchName` = `feature`, then tries to match `my-branch` against the children. It hits the `path: "*"` catch-all, which redirects to `.` (the parent, `/branches/feature`). So the user lands on a branch named `feature`, which does not exist — `details.tsx` does `branches.find((b) => b.name === branchName)` with `branchName === "feature"` and finds nothing.

3. **Link generation sites** (all use the unencoded `getBranchDetailsUrl`):
   - `frontend/app/src/entities/branches/ui/branch-list-item/branch-list-item.tsx:25` — `href={getBranchDetailsUrl(branch.name)}`
   - `frontend/app/src/entities/branches/ui/branches-table/cells/branch-name-cell.tsx:43` — `href={getBranchDetailsUrl(branch.name)}`
   - `frontend/app/src/entities/branches/ui/branch-tabs.tsx:13-17` — `to={getBranchDetailsUrl(branchName)}`

**Suggested fix**

Encode the branch name when building the URL (so `/` becomes `%2F` and the whole name stays in one path segment), and decode it when reading the route param:

- In `getBranchDetailsUrl`, wrap the branch name with `encodeURIComponent(branchName)` before interpolation.
- Where the param is read (`useRequiredParams("branchName")` in `frontend/app/src/pages/branches/details.tsx`), `decodeURIComponent` the value before comparing against `branch.name`.

Note: React Router / browsers can normalize an encoded `%2F` in the path back to `/` in some cases, so this should be verified end-to-end; if it proves unreliable, an alternative is a splat route (`branches/*`) for the branch name, or passing the branch via a query parameter.

bilalabbad:
# Why

On the **Branches** page, a branch whose name contains a `/` (e.g. `feature/my-branch`) was not clickable: clicking it failed to open that branch's detail page. The unencoded name was interpolated straight into the path, so React Router parsed the `/` as a separator and the single-segment `:branchName` route never matched the full name — the user was redirected to a non-existent branch instead.

Closes #9716

## What changed

- **Behavioral:** branches with a `/` in their name now open their detail page correctly from the branches list, the branches table, the breadcrumb selector, and the proposed-changes links.
- **Implementation:** `getBranchDetailsUrl` now `encodeURIComponent`s the branch name so the whole name stays in one path segment (`/branches/feature%2Fmy-branch`); React Router decodes the param back when reading it. This single helper is the chokepoint for every branch-detail link, so all call sites are fixed at once.
- **Unchanged:** no read-side `decodeURIComponent` was added — React Router v7 already decodes path params, and double-decoding would corrupt names containing a literal `%`. No route-structure or API changes.

## How to test

```bash
# Frontend e2e (legacy TS suite)
cd frontend/app && pnpm exec playwright test tests/e2e/branches/branch-details.spec.ts --grep "containing a slash"

# Python e2e suite
INFRAHUB_ADDRESS=http://localhost:8000 \
  uv run pytest -c tests/e2e/pytest.ini "tests/e2e/branches/test_branch_details.py::TestBranchDetailsSlashName"
```

Both suites create a branch whose name contains `/`, click it from the list, and assert the detail page opens (heading + tabs) and that path-based tab navigation works. Verified passing locally.

## Impact & rollout

- **Backward compatibility:** none affected — URLs for slash-free branch names are unchanged (encoding is a no-op for them).
- **Deployment notes:** safe to deploy; frontend-only.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (not user-facing doc change)

---
## Summary by cubic
Fixes branches with a “/” in their name so detail pages (and tabs) open correctly across the app. Keeps the full name in one path segment for routing.

- **Bug Fixes**
  - Encode `branchName` in `getBranchDetailsUrl` (e.g., `/branches/feature%2Fmy-branch`) and use it in branch-event titles; all branch-detail links now work.
  - No extra decoding added; React Router already decodes params. Added e2e tests for slash names and tab navigation.

<sup>Written for commit cff3485a1bcd1eb58b90a7df0ed3ec66ec6df813. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/app/src/entities/branches/utils.ts
```diff
@@ -9,7 +9,11 @@ export function getBranchDetailsUrl(
   tab?: BranchDetailsTab,
   overrideParams?: overrideQueryParams[]
 ): string {
-  const path = tab ? `/branches/${branchName}/${tab}` : `/branches/${branchName}`;
+  // Encode the branch name so a `/` in it (e.g. "feature/my-branch") stays inside
+  // a single path segment and is not parsed as a separator by the `:branchName`
+  // route. React Router decodes the param again when reading it via useParams.
+  const encodedBranchName = encodeURIComponent(branchName);
+  const path = tab ? `/branches/${encodedBranchName}/${tab}` : `/branches/${encodedBranchName}`;
```
saltas888, line 16: should we also do the same at https://github.com/opsmill/infrahub/blob/3026b905998b8aa20ebdae36b643be1842c175c8/frontend/app/src/entities/events/ui/branch-events/branch-event-title.tsx#L17, or even better use this util function?

bilalabbad: true (also it there is another bug in this line which is it truncate any url params)

saltas888 (approved):

PR #9726, opened 2026-06-26
docs(skills): add optimizing-neo4j-cypher-query skill

saltas888:
## What

Adds the **optimizing-neo4j-cypher-query** skill: a profile-driven workflow for diagnosing the slow Cypher behind a slow Infrahub endpoint — reproduce → pull the exact query + params from neo4j's `query.log` → `PROFILE` → read the plan against a curated operator-level anti-pattern reference → fix → re-PROFILE.

Ships `profile_slow_query.py` (the log parser / PROFILE runner) plus `test_profile_slow_query.py`, a stdlib test that pins neo4j's `query.log` line format so an upgrade-driven format drift fails loudly in the test instead of silently making `find-slow` report "nothing is slow".

## Relationship to `neo4j-cypher-guide`

The two are complementary, not overlapping — they sit on different axes, and this PR cross-links them both ways:

- **`neo4j-cypher-guide`** is the *write-time* counterpart: how to write correct modern Cypher (deprecated syntax, QPP, CALL subqueries, null-safe sorting). Generic, portable, used for query generation.
- **`optimizing-neo4j-cypher-query`** (this PR) is the *diagnose-time* counterpart: why a *specific* query is slow and how to fix it (PROFILE operators, db-hits selectivity, Infrahub-specific anti-patterns like the `index.py` rel indexes and the `$param IS NULL` UNION trap).

The anti-pattern reference deliberately lives with this skill rather than in the guide, so the guide stays generic and free of Infrahub-internal detail.

## Notes

- `ruff check` / `ruff format` / `ty check` clean; the 6 parser tests pass (`python scripts/test_profile_slow_query.py`).
- This is the first Python under `.agents/skills/`, which CI lints repo-wide (`ruff check .`, `select = ["ALL"]`). The scripts are kept **noqa-free**; instead `pyproject.toml` gains two scoped `per-file-ignores` entries for `.agents/skills/**/scripts` — `S404`/`S603` (the helper drives docker/cypher-shell by design) and `S101`/`SLF001` in its test — mirroring the existing `backend/tests/**` convention. The DB password now reads from `NEO4J_PASSWORD` (default `admin`) rather than a hardcoded literal.
- Otherwise pure docs + dev-tooling under `.agents/skills/`; no product code or changelog fragment.

### DIFF
--- a/.agents/skills/optimizing-neo4j-cypher-query/SKILL.md
+++ b/.agents/skills/optimizing-neo4j-cypher-query/SKILL.md
@@ -1,0 +1,155 @@
+---
+name: optimizing-neo4j-cypher-query
+description: Use when an Infrahub API call, branch diff, GraphQL query, or other backend operation is slow on a realistic-sized dataset, and the suspicion is a slow Cypher query. Walks through reproducing the slow path, pulling the exact query and parameters from neo4j's query.log, running PROFILE, and reading the plan against a curated anti-pattern reference.
+---
+
+# Optimize Neo4j Cypher Query
+
+## Overview
+
+Slow Infrahub endpoints are almost always slow Cypher queries underneath. This skill drives the loop: **reproduce → find → extract → PROFILE → analyze → fix → re-PROFILE**. It does not try to be clever about which query to optimize — it gives you the tools to find the slowest one against a realistic dataset and read the plan.
+
+The single source of truth for "which query ran with which parameters" is the neo4j container's `logs/query.log*` — every executed query is logged with its parameters, duration, and query name. Each entry also carries a `bolt_id` (the per-connection sequence number neo4j stamps on every query) which is the primary identifier the helper script uses. When the API server is started with `INFRAHUB_TRACE_ENABLE=true`, the log also carries the OTel span id as `infrahub_id`, which lets you cross-reference from Tempo/Grafana — but tracing is **not required** for any of the steps below.
+
+## When NOT to use
+
+- The slow path isn't database-bound (CPU profiling, network, file I/O). Profile the Python side first if you don't already know it's the DB.
+- The dataset is tiny. Optimizing against an empty DB just teaches you nothing about the production plan.
+- You're chasing a correctness bug, not a performance bug.
+
+## Prerequisites
+
+```dot
+digraph prereq {
+    "Start" [shape=doublecircle];
+    "Large dataset loaded?" [shape=diamond];
+    "Load it" [shape=box];
+    "DB container running?" [shape=diamond];
+    "Start dev.deps" [shape=box];
+    "Ready" [shape=doublecircle];
+
+    "Start" -> "DB container running?";
+    "DB container running?" -> "Start dev.deps" [label="no"];
+    "DB container running?" -> "Large dataset loaded?" [label="yes"];
+    "Start dev.deps" -> "Large dataset loaded?";
+    "Large dataset loaded?" -> "Load it" [label="no"];
+    "Large dataset loaded?" -> "Ready" [label="yes"];
+    "Load it" -> "Ready";
+}
+```
+
+Check large-dataset presence with:
+```bash
+docker exec infrahub-database-1 cypher-shell --format plain -u neo4j -p admin -d neo4j \
+  "MATCH (n) RETURN count(n);"
+```
+If the count is under ~100k, the dataset is not the realistic one. **REQUIRED SUB-SKILL:** Use `loading-infrahub-test-dataset` to restore a team snapshot before continuing — optimizing against an empty DB is a waste of cycles.
+
+The neo4j query log is on by default in the bundled image (`logs/query.log`, rotating to `query.log.01`, etc.); no extra config needed.
+
+**Optional — distributed tracing.** If the user wants to correlate Cypher queries to a higher-level trace (which HTTP request triggered them, which other queries ran in the same flow), set these in the API server's environment and restart it:
+
+```
+INFRAHUB_TRACE_ENABLE=true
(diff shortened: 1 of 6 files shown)


.agents/skills/optimizing-neo4j-cypher-query/references/cypher-profile-anti-patterns.md
```diff
@@ -0,0 +1,145 @@
...
+The planner couldn't find a label or index to start from. Fix: add a label on the start vertex, or filter by an indexed property in the same `MATCH`.
+
+### 2. `NodeByLabelScan` + downstream `Filter`
+
+```
+| NodeByLabelScan | Node     |  9M  |
+| Filter          | n.uuid = …|  1  |
+```
+
+Label scan then post-filter on a property. Add a property index for that label (`backend/infrahub/core/graph/index.py`) so the planner uses `NodeIndexSeek` instead.
```
ajtmccarty, line 33: I don't think we can add an index for everything. so the right solution could also be rewriting the query to filter by something that has an index first, if possible

.agents/skills/optimizing-neo4j-cypher-query/references/cypher-profile-anti-patterns.md
```diff
@@ -0,0 +1,145 @@
...
+
+`toString`, `toInteger`, `toLower`, `apoc.text.…`, or any function wrapping an indexed property will fall back to a scan. This is the exact bug PR #9225 hit on `node_get_kind_query`. Strip the conversion; convert the param in Python before passing.
+
+### 4. `Expand(All)` with high `db hits`
+
+```
+| Expand(All) | (n)-[:IS_RELATED]->(m) |  120k rows  |  4.5M db hits |
+```
+
+The expansion is reading many relationship properties to filter. If you're filtering `IS_RELATED` (or `IS_PART_OF`, `HAS_SOURCE`, `HAS_OWNER`, `IS_PROTECTED`) by `branch` / `from` / `status`, add a **relationship-property index** in `index.py`. PR #9225 added these specifically for the diff queries.
```
ajtmccarty, line 50: same here. first choice should be rewriting to use an existing index/faster filter if possible

.agents/skills/optimizing-neo4j-cypher-query/references/cypher-profile-anti-patterns.md
```diff
@@ -0,0 +1,145 @@
...
+
+- writes that depend on reads of the same labels/properties,
+- aggregation that the planner thinks must be globally consistent,
+- `DETACH DELETE` after a match it wrote into.
+
+Move the read above the write, split into multiple statements, or push the filter earlier so less work is materialised.
+
+### 8. `runtime=parallel` masking the plan
+
+The OTel-traced Cypher may have a `CYPHER runtime=parallel\n` prefix when running on Neo4j Enterprise. Parallel runtime can hide serial inefficiencies and the plan text is harder to read (per-pipeline rows multiply across workers). Always strip it for the first analysis (`profile_slow_query.py profile … --no-parallel`). Re-introduce parallel only after the single-thread plan is clean.
```
ajtmccarty, line 92: `OTel` ?

.agents/skills/optimizing-neo4j-cypher-query/references/cypher-profile-anti-patterns.md
```diff
@@ -0,0 +1,145 @@
...
+- `Expand(Into)` — both endpoints already pinned. Cheap.
+- `Apply` / `Optional` with small inputs.
+- `Distinct` with small input cardinality.
+
+## When to add an index vs restructure the query
+
+- **Add an index** when the same operator appears across multiple slow queries and the property/relationship has a clear access pattern (`branch`, `from`, `uuid`, `name`). Put it in `backend/infrahub/core/graph/index.py` so migrations carry it.
+- **Restructure the query** when the planner has the right indexes but is still picking a bad entry point (often because of an `IS NULL` guard, function on indexed prop, or unnecessary `UNION`).
+
+When in doubt: add the index, re-PROFILE, see if the planner uses it. If it does and the plan is clean, ship. If it doesn't, the issue is structural.
```
ajtmccarty, line 138: @fatih-acar might have some input on whether it is safe to add indexes like this. I would assume that there are space concerns for large databases

.agents/skills/optimizing-neo4j-cypher-query/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+- the change applied and why,
+- the new plan's worst operator + db hits.
+
+Skip the play-by-play of how you got the span id — the user cares about cause and fix.
+
+## Reference
+
+- **Anti-patterns:** `references/cypher-profile-anti-patterns.md` — operator-level PROFILE checklist (the focus of this skill).
+- **Modern Cypher syntax:** see the `neo4j-cypher-guide` skill — the write-time counterpart. Reach for it in step 6 when restructuring a query (QPP rewrites, CALL subqueries, null-safe sorting, replacing deprecated syntax). It covers *how to write* correct Cypher; this skill covers *how to diagnose* a slow one.
+- **Example PR:** opsmill/infrahub#9225 (slow branch diffs — fixed via entry-path restructure + rel indexes + removing `toString` on indexed property).
```
ajtmccarty, line 140: is this necessary? looked like the changes from that PR are captured in `cypher-profile-anti-patterns.md`

ajtmccarty (commented):
I think this looks pretty good. I didn't go through the log-reading code or test. loading a large dataset can take a long time so some option to run without the big dataset or as part of a component test might be good

saltas888:
Closing it in favor of https://github.com/opsmill/infrahub/pull/9967

PR #9747, opened 2026-06-29
fix: render trigger action parameters as plain strings (closes #9742)

### LINKED ISSUE #9742: bug: webhook automations crash during Prefect parameter render
### Component
API Server / GraphQL

### Infrahub version
1.10.0

### Current Behavior

Webhook automations fail at **Prefect parameter-render time** when an event fires, so the `webhook-process` deployment is never run and **no HTTP request is sent**. The failure happens inside Prefect's action parameter hydration, before any Infrahub webhook code executes.

`WebhookTriggerDefinitionBuilder.build()` (`backend/infrahub/webhook/models.py`) emits several action parameters as bare single-expression Jinja strings:

```python
"branch_name": "{{ event.resource['infrahub.branch.name'] }}",
"event_id": "{{ event.id }}",
"event_type": "{{ event.event }}",
"event_occured_at": "{{ event.occurred }}",
```

When the automation fires, Prefect's `RunDeployment.render_parameters()` runs `_upgrade_v1_templates()`, which auto-rewrites every bare `{{ … }}` parameter into `{{ … | tojson }}` (to preserve value types through a JSON round-trip). `tojson` then calls `json.dumps()` on values that are not JSON-native:

- `event_id` → `event.id` is a `UUID` → `TypeError: Object of type UUID is not JSON serializable` — **crashes on every webhook fire**
- `event_occured_at` → `event.occurred` is a `datetime` → `TypeError: Object of type datetime is not JSON serializable` — **crashes on every webhook fire**
- `branch_name` → resolves to `ChainableUndefined` whenever the firing event's resource has no `infrahub.branch.name` key (e.g. account/login events) → `TypeError`

The error surfaces from Prefect as:

```
prefect.utilities.schema_tools.hydration.InvalidJinja: Invalid jinja: Failed to render
template due to the following error: TypeError('Object of type UUID is not JSON serializable')
```

Note: the `event_payload` parameter (`{{ event.payload | tojson }}`) is correctly wrapped in a `json`+`jinja` structure and is **not** the cause — it renders fine.

**Why custom webhooks are the most visible:** standard and custom webhooks share `WebhookTriggerDefinitionBuilder`, so `event_id`/`event_occured_at` break both. Custom webhooks additionally default to `event_type="all"` (→ listen on `infrahub.*`), so they also receive events lacking `infrahub.branch.name` and hit the `branch_name` crash branch too.

### Expected Behavior

Webhook automations render their parameters successfully for any triggering event and dispatch the configured HTTP request. Scalar parameters (`event_id`, `event_occured_at`, `branch_name`) should be delivered as strings; events without a branch resource should yield an empty/absent branch rather than crashing.

### Steps to Reproduce

1. Create a Custom Webhook pointing at any reachable URL, leaving `event_type` at its default (`all`).
2. Trigger any Infrahub event (e.g. create/update a node, or any account/login event).
3. Observe the `webhook-process` automation fail on Prefect during parameter rendering — no HTTP request reaches the target URL.

Minimal mechanism reproduction (mirrors `RunDeployment.render_parameters`):

```python
import copy
from prefect.server.events.actions import RunDeployment
from prefect.utilities.schema_tools.hydration import hydrate, HydrationContext

params = {"event_id": "{{ event.id }}"}            # as emitted by the webhook builder
RunDeployment._upgrade_v1_templates(params)        # Prefect appends "| tojson"
hydrate(params, HydrationContext(raise_on_error=True, render_jinja=True,
        jinja_context={"event": received_event}))  # received_event.id is a UUID -> TypeError
```

### Additional Information

- **Affected code:** `backend/infrahub/webhook/models.py` → `WebhookTriggerDefinitionBuilder.build()`, the `event_id` / `event_occured_at` / `branch_name` action parameters.
- **Test gap:** `backend/tests/functional/webhook/test_process.py` calls `webhook_process()` directly with a hand-built `event_payload` dict, bypassing Prefect's parameter rendering; `test_trigger_definition_builder.py` only asserts the un-rendered parameter shape. No test exercises the render path where the crash occurs.
- **Suggested fix (two verified options):**
  - *Preferred:* wrap each scalar in an explicit plain-`jinja` kind, e.g. `{"__prefect_kind": "jinja", "template": "{{ event.id }}"}`. A dict already carrying `__prefect_kind` is skipped by `_upgrade_v1_templates`, so no `| tojson` is appended and the value renders as a string.
  - *Alternative:* add `| string`, e.g. `{{ event.id | string }}` — the auto-appended `| tojson` then serializes a `str`, which is safe.
- **Environment:** Infrahub 1.10.0 · Prefect 3.7.5 · Python 3.13 · observed on macOS arm64 (platform-independent — the defect is in server-side trigger-definition/parameter-render logic).

polmichel:
# Why

Webhook automations crashed at Prefect parameter-render time, so no HTTP request was ever sent. Trigger definitions emitted single-expression action parameters as bare Jinja strings; `RunDeployment._upgrade_v1_templates` appends `| tojson` to them, and `json.dumps()` fails on values that are not JSON-native (a `UUID`, a `datetime`) or that resolve to an absent resource key (`ChainableUndefined`).

The `| tojson` coercion was introduced in **Prefect 3.6.24** (PrefectHQ/prefect#21231); it is absent in 3.6.23 and earlier. The crash surfaced once we moved to `prefect==3.7.5` (Infrahub 1.10.0, up from `3.6.13`). Webhooks were the visible victim because custom webhooks default to `event_type="all"`, so they receive branch-less events (e.g. account events) where the branch key is missing. The same bare-string pattern existed in seven other trigger builders; they did not crash only because their narrow event triggers always carry the referenced keys.

Closes #9742

## What changed

- **Behavioral:** Webhooks now fire for every triggering event, including events with no branch. Scalars are delivered as strings; a missing branch resolves to an empty string.
- **Implementation:** Add a shared `jinja_parameter()` helper that wraps a template as an explicit `{"__prefect_kind": "jinja", "template": ...}` parameter — [Prefect's documented way](https://docs.prefect.io/v3/how-to-guides/automations/access-parameters-in-templates) to mark a templated parameter, which Prefect leaves untouched by the `tojson` upgrade. Apply it to every single-expression action parameter across all trigger builders.
- **Scope note:** The webhook builder is the live fix; the other seven are preventative hardening against the same failure. String-valued params render identically to before, so existing behavior is preserved.

## How to review

Start with `backend/infrahub/trigger/models.py` (`jinja_parameter` helper), then `backend/infrahub/webhook/models.py` (the fix). The remaining builder edits are mechanical applications of the same helper. Commits are atomic: helper → webhook fix+tests → systemic hardening → changelog.

## How to test

```bash
# Unit: builder output renders to strings through Prefect's render path
uv run pytest backend/tests/unit/webhook/test_trigger_definition_builder.py::TestWebhookParameterRender -v
# Live Prefect: a branch-less event fires the automation and dispatches webhook-process
uv run pytest backend/tests/functional/webhook/test_render.py -v
```

### Manual end-to-end verification (full Docker stack)

Verified on a live `dev.start` stack (real server, Prefect, two task-workers) with a custom webhook (`event_type=all`) pointing at a local capture server:

- **Node event** (`infrahub.node.created`): webhook delivered; payload carried the previously-crashing values as strings — `"id"` (UUID) and `"occured_at"` (datetime).
- **Branch-less event** (`infrahub.account.logged_in`, whose resource has no `infrahub.branch.name`): webhook also delivered — the `ChainableUndefined` path.
- **Negative control** (fix reverted in the running app): the same node event produced no delivery; the Prefect server logged `TypeError: Object of type UUID is not JSON serializable` → `Action failed: 'Unable to create flow run from deployment: InvalidJinja()'`, exactly the reported failure.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated — N/A
- [x] Internal .md docs updated — N/A


backend/tests/functional/webhook/test_render.py
```diff
@@ -0,0 +1,82 @@
...
+        runs_before = len(await prefect_client.read_flow_runs(deployment_filter=deployment_filter))
+
+        # A branch-less event: the resource carries no infrahub.branch.name, the id is a UUID and the
+        # occurred time a datetime -- all values the action parameters must render as plain strings.
+        event = Event(
+            id=uuid4(),
+            event="infrahub.node.created",
+            payload={"data": {"node_id": "abc"}, "context": {}},
+            resource=Resource({"prefect.resource.id": "infrahub.account.xyz"}),
+        )
```
ajtmccarty, line 76: I don't see a datetime in here, which is what the comment describes, but perhaps it happens behind-the-scenes

backend/tests/unit/webhook/test_trigger_definition_builder.py
```diff
@@ -98,6 +104,44 @@ def test_generate_name(self) -> None:
...
+            payload={"data": {"id": "abc"}, "context": {}},
+            resource=Resource({"prefect.resource.id": "infrahub.account.xyz"}),
+        )
+
+        # Mirror Prefect's server-side parameter render that runs when the automation fires.
+        RunDeployment._upgrade_v1_templates(parameters)
+        rendered = hydrate(
+            parameters,
+            HydrationContext(raise_on_error=True, render_jinja=True, jinja_context={"event": event}),
+        )
```
ajtmccarty: this looks like we are testing Prefect's code, which we don't really have control over. if this test covers something that the functional test does not, then it might be worth keeping. otherwise it seems like it will be too brittle to be valuable

ajtmccarty (approved):

backend/tests/functional/webhook/test_render.py
polmichel: It's defaulted to today's date, but you're right, explicit is better than implicit. And a fixed date is better than a moving one from a test perspective

backend/tests/unit/webhook/test_trigger_definition_builder.py
polmichel: Good point. This is what was executed behind the scenes during parameters serialization. Since we have the component test covering this, I think we can remove it. Useful at the review step but not further.

PR #9749, opened 2026-06-29
fix(computed-attribute): refresh derived values after a branch merge

gmazoyer:
# Why

On `develop`, a schema-changing merge updated the registry but did not emit a `SchemaUpdatedEvent`, so the display-label, HFID, and computed-attribute backfills never ran on the destination branch. Stored derived values on destination nodes that were absent from the merge diff stayed stale until the node was next mutated.

Goal: a merge that edits a derived-value definition (display label, human-friendly ID, or a Jinja2 computed-attribute template) refreshes the stored derived values of nodes that exist on the target branch but were absent from the source branch.

Non-goals: the rebase path. It has the same gap, but develop's scoped-recompute design only recomputes branches whose schema differs from the default branch, and a rebased branch's schema matches the default by definition (confirmed end to end: the post-rebase setup selects 0 attributes and deletes the branch trigger). Rebase recompute is left to the coalesced recompute work (IFC-2761). Tracked separately as IFC-2831.

Backport of IFC-2758 (resolved on stable, deferred from the stable-to-develop sync in #9712) as the scoped variant that fits develop's existing recompute machinery (#9415 / #9467). Tracked as IFC-2815.

## What changed

Behavioral:
- Merging a branch that changes a computed attribute, display label, or HFID template now refreshes those derived values on destination-only nodes.
- The refresh is scoped to the schema elements the merge changed, so unrelated attributes are not recomputed and a merge does not broaden recompute into a full sweep on the default branch.

Implementation:
- The merge orchestrator computes the destination schema diff (`pre_merge_schema.diff(candidate_schema)`) when a merge applies schema changes.
- The post-merge dispatcher emits a `SchemaUpdatedEvent` for the destination branch carrying the changed-element set, sent with the other post-merge events after the merge is committed so a send failure cannot roll back the merge.

What stayed the same:
- No database schema changes, no migration, no `GRAPH_VERSION` bump.
- The rebase path is unchanged.
- The scoped-recompute machinery is unchanged; this only feeds it a correctly scoped event on merge.
- No new changelog fragment. This backports a fix already on stable, whose `+merge-refresh-derived-values.fixed.md` fragment lands on develop via the stable-to-develop sync.

## How to review

- `backend/infrahub/core/merge/orchestrator.py` and `backend/infrahub/core/merge/post_merge.py` are the fix.
- `backend/tests/component/core/merge/test_post_merge.py` covers event emission and scope without a full stack.
- `backend/tests/integration_docker/test_merge_derived_value_refresh.py` is the end-to-end repro (a main-only node refreshes after a schema-changing merge).
- In `backend/tests/integration_docker/test_computed_attributes.py`, the former `test_merge_does_not_trigger_schema_scoped_recompute` is replaced by `test_merge_triggers_scoped_recompute`, which asserts the merge recomputes the changed attribute on the default branch.

## How to test

```bash
# Component (fast)
uv run pytest backend/tests/component/core/merge/test_post_merge.py --neo4j

# Integration (docker; builds the local image first)
uv run invoke dev.build
uv run pytest backend/tests/integration_docker/test_merge_derived_value_refresh.py
```

Verified locally against a freshly built image: the docker repro passes (a main-only node refreshes from `v2-beta` to `v3-beta` after the merge), and the component tests pass.

## Impact & rollout

- **Backward compatibility:** no breaking changes; no migration.
- **Performance:** recompute on merge is scoped to the changed schema elements, so it does not broaden into a full sweep.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry: not added here. This backports a stable fix whose changelog fragment already exists on stable and syncs to develop.
- [ ] External docs updated (no user-facing doc change)
- [ ] Internal .md docs updated (no internal doc made a stale claim about this path)

### DIFF
--- a/backend/infrahub/core/merge/post_merge.py
+++ b/backend/infrahub/core/merge/post_merge.py
@@ -4,9 +4,11 @@ from typing import TYPE_CHECKING, Any, Sequence
 
 from infrahub import config
 from infrahub.core.constants import MutationAction
+from infrahub.core.registry import registry
 from infrahub.events.branch_action import BranchMergedEvent
 from infrahub.events.models import EventMeta, InfrahubEvent
 from infrahub.events.node_action import get_node_event
+from infrahub.events.schema_action import SchemaUpdatedEvent, build_changed_elements_payload
 from infrahub.log import get_logger
 from infrahub.workflows.catalogue import (
     BRANCH_CANCEL_PROPOSED_CHANGES,
@@ -21,6 +23,7 @@ if TYPE_CHECKING:
     from infrahub.core.changelog.models import NodeChangelog
     from infrahub.core.constants import DiffAction
     from infrahub.core.diff.ipam_diff_parser import IpamNodeDetails
+    from infrahub.core.models import SchemaDiff
     from infrahub.log import InfrahubLogger
     from infrahub.services.adapters.event import InfrahubEventService
     from infrahub.services.adapters.workflow import InfrahubWorkflow
@@ -101,6 +104,7 @@ class PostMergeDispatcher:
         proposed_change_id: str | None,
         node_events: Sequence[tuple[DiffAction, NodeChangelog]],
         context: InfrahubContext,
+        schema_diff: SchemaDiff | None = None,
     ) -> None:
         event_context = context.to_event_context()
         merge_event = BranchMergedEvent(
@@ -111,6 +115,17 @@ class PostMergeDispatcher:
         )
 
         events: list[InfrahubEvent] = [merge_event]
+        if schema_diff is not None:
+            # Drive the display-label, HFID and computed-attribute backfills for destination-only nodes,
+            # scoped to the elements the merge changed so the recompute stays narrow.
+            events.append(
+                SchemaUpdatedEvent(
+                    branch_name=self.default_branch.name,
+                    schema_hash=registry.schema.get_schema_branch(name=self.default_branch.name).get_hash(),
+                    changed_elements=build_changed_elements_payload(schema_diff),
+                    meta=EventMeta.from_parent(parent=merge_event, branch=self.default_branch),
+                )
+            )
         for action, node_changelog in node_events:
             meta = EventMeta.from_parent(parent=merge_event, branch=self.default_branch)
             node_event_class = get_node_event(MutationAction.from_diff_action(diff_action=action))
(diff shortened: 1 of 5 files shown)


backend/infrahub/core/merge/post_merge.py
```diff
@@ -111,6 +115,17 @@ async def dispatch_events(
         )
 
         events: list[InfrahubEvent] = [merge_event]
+        if schema_diff is not None:
+            # Drive the display-label, HFID and computed-attribute backfills for destination-only nodes,
+            # scoped to the elements the merge changed so the recompute stays narrow.
+            events.append(
+                SchemaUpdatedEvent(
+                    branch_name=self.default_branch.name,
+                    schema_hash=registry.schema.get_schema_branch(name=self.default_branch.name).get_hash(),
```
ajtmccarty: instead of using the `registry` in here, could you pass `pre_merge_schema` in from `BranchMergeOrchestrator`? I'm not sure if `pre_merge_schema` is exactly what you want, but we should try to avoid using `registry` in new code anywhere other than an entry point (gql mutation, REST endpoint, Prefect flow, etc.)

backend/tests/integration_docker/test_computed_attributes.py
```diff
@@ -447,11 +443,16 @@ async def test_merge_does_not_trigger_schema_scoped_recompute(
...
+            await sleep(1)
+            await wait_for_all_tasks_to_be_completed(client)
+            main_after = await client.task.count(
+                filters=TaskFilter(workflow=[COMPUTED_ATTRIBUTE_JINJA2_UPDATE_VALUE.name], branch="main")
+            )
+            if main_after > main_before:
+                break
 
-        assert main_after == main_before
+        assert main_after > main_before
```
ajtmccarty, line 471: not sure if this is possible, but it would be nice for this to be a slightly more precise assertion. something about the shape and content of the dispatched task

ajtmccarty (approved):

backend/tests/integration_docker/test_computed_attributes.py
gmazoyer: It now creates a main-only node and polls until its description refreshes to the merged template's output, so it pins the recomputed value, not just the task count.

backend/infrahub/core/merge/post_merge.py
gmazoyer: Good call! Moved it to the orchestrator: it now passes the post-merge `candidate_schema.get_hash()` into `dispatch_events`, so the dispatcher no longer touches the registry. Used the post-merge schema rather than `pre_merge_schema`, since the event should carry the new hash.

PR #9751, opened 2026-06-29
Coalesce merge and rebase recompute

gmazoyer:
## Why

When a branch is merged or rebased, Infrahub has to refresh the values derived from the change: computed attributes, display labels, and human-friendly ids. Until now every changed node drove its own refresh, so the background work after the operation grew in step with the size of the change. A large merge or rebase could leave the instance under heavy recompute load for a long window before it settled, even though the number of values that genuinely changed was far smaller.

This PR makes that recompute scale with the number of derived values actually affected rather than with how many nodes were touched. The end state is unchanged; only the amount of work to reach it is reduced.

## How the parts fit together

The change introduces one new decision-making piece, a coordinator, and reuses the machinery that already exists to do the actual refresh.

```
  merge / rebase
       │  hands over the set of changed nodes
       ▼
  ┌──────────────┐   asks "what does each change affect?"
  │ coordinator  │ ───────────────►  per-value-type dependency derivers
  │              │ ◄───────────────  (computed attributes, display
  └──────────────┘   returns targets   labels, human-friendly ids)
       │
       │  deduplicates into one set of values to refresh
       ▼
  existing per-family refresh jobs
  (one grouped job per value type, over all affected nodes at once)

  in parallel:
  ┌─────────────────────────────────────────────────────────────┐
  │ each change a merge or rebase replays is tagged "replayed";  │
  │ the old per-node automations are set to act only on live     │
  │ edits, so they step aside for replayed changes               │
  └─────────────────────────────────────────────────────────────┘
```

The coordinator does not invent its own idea of what depends on what. It asks the same per-value-type dependency information the normal, live recompute already relies on, so its decisions stay consistent with everyday behavior and respect each value type's own rules. For example, an id built only from a node's own name is not refreshed when a node it is merely related to changes, while a display label that reads across a relationship is.

Once the coordinator has the deduplicated set, it hands the work to the per-family refresh jobs that already exist, asking each to refresh all of its affected nodes in one pass rather than one node at a time. Nothing in the refresh logic itself is rebuilt.

The suppression is what keeps the work from happening twice. Because the coordinator now owns the refresh for these value types on the merge and rebase path, the existing per-node automations are told to fire only for live edits and to step aside for changes that a merge or rebase replayed. This also closes a gap the per-node replay could miss: readers that live only on the branch the recompute runs on are still covered by the coordinator's pass. Merge and rebase share the same coordinator, each acting on the correct branch.

## Behavior

This is a performance change and is meant to be invisible in its results. The derived values an instance ends up with after a merge or rebase are the same as before.

## Validation

Correctness is checked end to end against a from-scratch recompute, including the case that only the coordinated pass can satisfy, and the reduction in work is measured with a profiling harness on a full stack.

## Scope

This increment covers Jinja2 computed attributes, display labels, and human-friendly ids. Python-transform computed attributes and the profile-refresh family keep their current path and are a follow-up.

Relates to IFC-2761.

---
## Summary by cubic
Coalesces merge/rebase recompute into one deduplicated pass for Jinja2 computed attributes, display labels, and HFIDs so work scales with affected values, not changed-node count. Adds an origin label on node events and pins recompute writes to `live`, meeting IFC-2761.

- **New Features**
  - Build one recompute plan from the diff; reuse the Jinja2 deriver and add display-label/HFID scopers from schema metadata; cover created, updated (cross-node), and deleted changes on the correct branch.
  - Submit one per-family union using `object_ids`; chunk via `get_submission_chunk_size` in `infrahub.events.limits` to stay under Prefect’s cap, and isolate each submission.
  - Integrated into merge (destination) and rebase (user) via `CoalescedRecomputeBuilder`, `CoalescedRecomputeSubmitter`, and `MergeRecomputeCoordinator` at the entry points; per-node events still emit for other consumers.
  - Events include `meta.origin` and resource label `infrahub.node.origin`; GraphQL recompute writes explicitly emit `live`.

- **Bug Fixes**
  - Per-node recompute automations for Jinja2/display/HFID now match `live` origin, preventing double-dispatch on merge/rebase replays.
  - Marked a strict xfail for the deleted-peer stale-reader gap (IFC-2852) and regenerated event reference docs with the neutral `meta.origin` description.

<sup>Written for commit 870a3b745edf4700faae669c3477efca21356c65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9751?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,364 @@
...
+            reads_across_relationship=self.reads_across_relationship,
+            reader_lookups=frozenset(
+                ReaderLookup(source_kind=source_kind, filter_key=filter_key, source_node_ids=frozenset(ids))
+                for (source_kind, filter_key), ids in self.ids_by_lookup.items()
+            ),
+            precise=self.precise,
+        )
+
+
+def build_coalesced_recompute(
```
saltas888: seems like this function does a lot of heavy lifting. Would it make sense to have a test in backend/tests/benchmark?

gmazoyer: Good idea. We are keeping this one benchmark-free here on purpose: build_coalesced_recompute runs once per merge or rebase (not per node), so its cost is fixed overhead rather than a hot path, and the signal that matters for this work is the end-to-end dispatch reduction and trailing recompute window at scale. We are moving the profiling and benchmark tests to a dedicated repo (opsmill/infrahub-private-tests), which is better suited to running against production-like datasets of various sizes, and that is where this will be tracked.

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,394 @@
...
+    context: EventContext,
+) -> list[CoalescedSubmission]:
+    """Submit the coalesced recompute by reusing the existing per-family process flows.
+
+    Each submission runs one process flow over a chunk of changed nodes, so the merge and rebase
+    path no longer dispatches one flow per changed node. A single submission failure is logged and
+    skipped rather than dropping the rest, since a missed submission leaves a stale stored value.
+    Returns the submissions that were dispatched.
+    """
+    submitted: list[CoalescedSubmission] = []
```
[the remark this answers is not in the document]
gmazoyer: This is by design on the post-commit path. By the time recompute is submitted the merge or rebase is already committed, so this follow-up is best-effort and must not fail the operation. It uses the same log-and-continue contract as the sibling event-send follow-up in `post_merge`. The per-submission `try/except` here is a strict improvement over the previous behavior: one failed submission no longer aborts the rest of the batch.

The dominant failure mode was a submission exceeding Prefect's 512 KB flow-run parameter limit and returning a 422, which is exactly what this change removes by chunking the union. What remains is transient Prefect API errors, which are rare and which retrying inside an already Prefect-hosted flow would not reliably recover.

A skipped submission is already surfaced: each one logs at exception level with its family and target kind, so a partial failure is greppable rather than silent. Guaranteeing eventual consistency after a transient submission failure would need a reconciliation or backfill job, which is a separate piece of work rather than something to bolt onto this best-effort path.

backend/infrahub/hfid/tasks.py
```diff
@@ -84,13 +84,21 @@ async def hfid_update_value(
 async def process_hfid(
     branch_name: str,
     node_kind: str,
-    object_id: str,
     target_kind: str,
     context: EventContext,
+    object_id: str | None = None,
```
saltas888, line 89: do we have a reason to keep both arguemnts? Why not just supporting list of strings? It makes the code easier later

backend/infrahub/hfid/tasks.py
```diff
@@ -106,7 +114,7 @@ async def process_hfid(
         node_schema=node_schema, variables=hfid_definition.hfid, filter_key=hfid_definition.filter_key
     )
 
-    query = hfid_graphql.render_graphql_query(filter_id=object_id)
+    query = hfid_graphql.render_graphql_query(filter_id=filter_id)
```
saltas888, line 115: Like we discussed, lets make sure this is scaling (e.g.  what if filter_id contains 1000 ids? How does neo4j handle an OR in this many)? Maybe @ajtmccarty can give some tips

ajtmccarty: if this is a list of UUIDs, it shouldn't be much different from getting a single item by ID. the cypher queries are batched to get all the necessary pieces for 1 page of data in a few queries (not sure exactly how many). but the number of database queries will stay about constant from 1 to the max page size

backend/infrahub/hfid/tasks.py
gmazoyer: Agreed it would be cleaner. Both exist because the two callers pass different shapes: the live per-node automation binds a single id from the event (`object_id`), while the coalesced merge/rebase path passes the whole union (`object_ids`). Collapsing to just a list means changing the live automation trigger template in all three families (computed attribute, display label, hfid) to emit a one-element list, which touches the existing live recompute path. To keep this PR focused I'll leave both here for now and open a follow-up to unify on a single list argument.

backend/infrahub/events/constants.py
```diff
@@ -3,6 +3,18 @@
 EVENT_NAMESPACE = "infrahub"
 ACCOUNT_EVENT_PREFIX = f"{EVENT_NAMESPACE}.account."
 
+# Recompute automations match LIVE positively, not a list of negations: Prefect ORs a multi-value
+# match, so "not merge, not rebase" would always be true and exclude nothing.
```
ajtmccarty: I have no idea what this means. this looks like it is just a namespace of some sort

backend/infrahub/events/models.py
```diff
@@ -75,6 +75,10 @@ class EventMeta(BaseModel):
 
     parent: UUID | None = Field(default=None, description="The UUID of the parent event if applicable")
     ancestors: list[ParentEvent] = Field(default_factory=list, description="Any event used to trigger this event")
+    origin: NodeMutationOrigin | None = Field(
+        default=None,
```
ajtmccarty: should the default be `NodeMutationOrigin.LIVE`? is that possible? seems more accurate

backend/infrahub/graphql/mutations/computed_attribute.py
```diff
@@ -107,6 +107,8 @@ async def mutate(
             log_data = get_log_data()
             request_id = log_data.get("request_id", "")
 
+            # A recompute write must stay live-origin: a merge/rebase origin here would be suppressed by
+            # the coalesced-recompute triggers, leaving dependent recomputes stale.
```
ajtmccarty: should this comment be down a few lines over `EventMeta(...`? I can't really tell what this means here. should there be an explicit origin=None/NodeMutationOrigin.LIVE in the `EventMeta` instantiation?

same for the other places with this comment

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,396 @@
...
+    """Build the deduplicated recompute for a merge or rebase change set.
+
+    Groups the changes by signature so derivation runs once per distinct change
+    shape, runs the three family derivers, and merges the results so each derived
+    target is recomputed once with its reader lookups unioned.
+
+    Targets come only from the changes in the set. A branch-originated merge carries every level the
+    branch already cascaded, so a transitive chain is covered in one pass; a level absent from the set
+    (a destination-only reader, or a rebase whose source branch had no readers) is reached instead by
+    the live recompute its predecessor's write triggers.
```
ajtmccarty: I've tried to read this a few times and have no idea what it means. not sure what a "branch-originated merge" is either

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,396 @@
...
+                for (source_kind, filter_key), ids in self.ids_by_lookup.items()
+            ),
+            precise=self.precise,
+        )
+
+
+def build_coalesced_recompute(
+    *,
+    changes: Iterable[MergeChange],
+    schema_branch: SchemaBranch,
```
ajtmccarty: this looks like a class where `SchemaBranch` is a dependency instead of a function that passes `schema_branch` down through several layers of functions

backend/infrahub/core/merge/recompute_coalescing.py
ajtmccarty, line 1: this looks like  it should be split into multiple classes that are injected into one "super" class that runs `CoalescedRecomputeBuilder` and  passes its output into `CoalescedRecomputeSubmitter` (or better names). the two look like they are always called together and depend on one another so that should be built into the structure of their classes

backend/tests/unit/core/merge/test_build_coalesced_recompute.py
ajtmccarty, line 1: I think this could be a unit test, but not entirely certain. I think the recomputing logic just needs a SchemaBranch, so I don't think the database is actually used for anything

this might apply to other test files in this directory too. I don't think database access is required for any of this logic

backend/tests/component/merge_recompute_coalescing/test_build_coalesced_recompute.py
```diff
@@ -0,0 +1,172 @@
...
+    default_branch: Branch,
+    register_core_models_schema: SchemaBranch,
+) -> None:
+    """An update with no recorded fields recomputes every reader and is marked imprecise."""
+    schema_branch = await _profile_schema_branch(db=db)
+    changes = [MergeChange(node_id="peer-0", kind=PROFILE_PEER_KIND, action="updated", changed_fields=frozenset())]
+
+    result = build_coalesced_recompute(changes=changes, schema_branch=schema_branch, branch="main")
+
+    assert result.targets != frozenset()
```
ajtmccarty: can you assert the expected targets? or something about them other than `!= frozenset()`?

ajtmccarty (commented):
I've reviewed some of this and added some comments. one higher-level question: does this account for the case when an attribute/relationship that affects a display label/HFID/computed attribute is updated on `main` after the branch forks? this would not be covered by the diff for the branch.

example:
1. add object `A` that includes a computed attribute based on the object's `name` and `description` attributes
2. create branch `br1`
3. on branch `br1`, `A.description` is updated
4. on `main`, `A.name` is updated
5. I merge `br1` into `main`

the computed attribute on `A` is correct on the branch, but will not include the update to the `name` on `main`. I did not read the spec, it is too big.

backend/infrahub/core/merge/recompute_coalescing.py
gmazoyer: Done, added a a constructor parameter for `CoalescedRecomputeBuilder`

backend/infrahub/events/models.py
gmazoyer: Good point, done.

backend/infrahub/graphql/mutations/computed_attribute.py
gmazoyer: True, I added an explicit `origin=NodeMutationOrigin.LIVE` and moved the comments on top of it.

backend/infrahub/core/merge/recompute_coalescing.py
gmazoyer: Good idea, I stole the names from you because I just don't know how to name things. So I added `CoalescedRecomputeBuilder` and `CoalescedRecomputeSubmitter`, run by `MergeRecomputeCoordinator.new(schema_branch=..., workflow=...)` at merge and rebase.

backend/infrahub/hfid/tasks.py
saltas888: @ajtmccarty So what you are saying is that the same query if it does an or in 1 or 1000 items would require the same (CPU) resources?

ajtmccarty: yes, mostly. there will be some difference for getting more results and processing them, but it should be about the same. still probably worth performance testing

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,430 @@
...
+    """Yield ``ids`` in contiguous slices of at most ``size``."""
+    for start in range(0, len(ids), size):
+        yield ids[start : start + size]
+
+
+class CoalescedRecomputeBuilder:
+    """Derive the deduplicated recompute for a merge or rebase change set from one schema branch.
+
+    The schema branch is the dependency the derivation reads, so it is held on the instance rather
+    than threaded through every call.
```
ajtmccarty: unnecessary

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,430 @@
...
+                    filter_key=filter_key,
+                    reads_across_relationship=not is_self,
+                    precise=precise,
+                )
+
+
+class CoalescedRecomputeSubmitter:
+    """Submit a coalesced recompute by reusing the existing per-family process flows.
+
+    The workflow adapter is the dependency the submission needs, so it is held on the instance.
```
ajtmccarty: yes, this is obvious

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -0,0 +1,430 @@
...
+    def __init__(self, builder: CoalescedRecomputeBuilder, submitter: CoalescedRecomputeSubmitter) -> None:
+        self.builder = builder
+        self.submitter = submitter
+
+    @classmethod
+    def new(cls, *, schema_branch: SchemaBranch, workflow: InfrahubWorkflow) -> Self:
+        return cls(
+            builder=CoalescedRecomputeBuilder(schema_branch=schema_branch),
+            submitter=CoalescedRecomputeSubmitter(workflow=workflow),
+        )
```
ajtmccarty: I don't really think this belongs on the class b/c then it links the dependencies of the dependencies to this class and `MergeRecomputeCoordinator` shouldn't need knowledge about how to build `CoalescedRecompute...`

but it is quite simple, so doesn't really matter

backend/tests/component/merge_recompute_coalescing/test_merge_self_attribute_both_sides.py
```diff
@@ -0,0 +1,122 @@
+"""A self computed attribute whose two inputs change on each side of a merge.
+
+Reproduces the question raised in review: the branch edits one input and the destination edits
```
ajtmccarty: no one cares where the question came from

backend/tests/component/merge_recompute_coalescing/test_merge_self_attribute_both_sides.py
```diff
@@ -0,0 +1,122 @@
...
+        account=AccountSession(account_id=str(uuid4()), auth_type=AuthType.NONE),
+    )
+    # Each side recomputed the summary to a different value, so the merge's conflict gate blocks it
+    # rather than committing a value that reflects only one of the two edits.
+    with (
+        dependency_provider.scope(build_database, lambda singleton=True: db),  # noqa: ARG005
+        dependency_provider.scope(build_event_service, lambda: event_recorder),
+        dependency_provider.scope(build_workflow, lambda: workflow_recorder),
+        dependency_provider.scope(build_cache, lambda: cache),
+        pytest.raises(MergeConflictsUnresolvedError, match=r"conflict resolution missing:.*summary/value"),
```
ajtmccarty: I think this is actually kind of a bug. definitely not to be addressed in this PR. but we should allow users to update different attributes on `main` and a branch and then handle updating the display label as part of a merge/rebase. we've got a couple of issues for this kind of thing that are becoming higher priority. added a comment for this issue to https://opsmill.atlassian.net/browse/IFC-1898

backend/infrahub/events/constants.py
```diff
@@ -3,6 +3,21 @@
 EVENT_NAMESPACE = "infrahub"
 ACCOUNT_EVENT_PREFIX = f"{EVENT_NAMESPACE}.account."
 
+# Resource label on a node mutation event carrying its origin ("live", "merge", or "rebase"). The
+# per-node recompute automations subscribe to it by matching "live" directly, so a merge or rebase
+# replay (a non-live origin) is skipped and handled by the coalesced pass instead. Matching by
+# exclusion would not work: Prefect ORs a multi-value match, so "not merge, not rebase" is always
+# true and would exclude nothing.
```
ajtmccarty: still have no idea what this comment is saying. this looks like it's just a constant namespace, no explanation required

ajtmccarty (approved):

gmazoyer:
Correction to what I first posted here, now with a clean test to back it.

The per value updates drain through an in process batch capped at 5 at a time. That cap is not new. 1.10 and the commit just before this branch do the same thing in the same place, so this branch did not introduce it. What this branch changes is the dispatch, one process flow per family instead of one per changed node. It does not touch the per value writes, like the commit message says.

To settle it I built the commit just before this branch and this branch the same way, and ran both on the same box back to back, same merge (one device type, 300 devices, same batch of 5). The window was 587s before this branch and 562s on this branch. So this branch is not slower, no regression. My earlier big number was a bad compare. 1.10 was the published image on a smaller device type, and the branch was a source build on a bigger one.

What still holds. The batch of 5 is a real throughput limit for a big fan out, but it is pre existing, it would help 1.10 too. The bulk write work is the better fix since it removes the per value writes.

PR #9754, opened 2026-06-29
feat(webhook): retry and cancel actions for deliveries

polmichel:
## Summary

Adds operator recovery actions for webhook deliveries. Each delivery runs as a user-facing `webhook_send` task; from a delivery's task detail page an operator can retry a settled delivery or cancel one still in flight.

Closes [IFC-2753](https://opsmill.atlassian.net/browse/IFC-2753) and [IFC-2119](https://opsmill.atlassian.net/browse/IFC-2119)

## What this adds

- **`webhook_send` as the user-facing delivery.** Each delivery is its own task run, surfaced in the webhook's Tasks tab.
- **Available actions on the GraphQL `Task` type.** `available_actions` reports, per delivery, whether retry and cancel apply and, when they do not, the reason. Retry applies once a delivery has settled; cancel applies while it is still in flight.
- **`InfrahubTaskRetry` and `InfrahubTaskCancel` mutations.** Retry replays the delivery's frozen payload as a new independent delivery, recomputing its signature and leaving the original row unchanged. Cancel stops remaining retries on an in-flight delivery without recalling an already-sent request. Both authorize against update permission on the webhook node and reject actions that do not apply to the delivery's current state.
- **Retry and cancel controls on the task detail page**, each behind a confirmation dialog, shown only when the action is available.

## How to review

This PR is best reviewed **commit by commit**. Each commit is a self-contained, logically scoped step, ordered so each builds on the previous — review them top to bottom.

| # | Commit | What it implements |
|---|--------|--------------------|
| 1 | [`c1b460a8d`](https://github.com/opsmill/infrahub/pull/9754/commits/c1b460a8dc20a03f99f10aba53056ab92ecd5834) | Registers `WEBHOOK_SEND` in the workflow catalogue so each delivery is its own run that can be re-submitted independently. |
| 2 | [`9acd5e490`](https://github.com/opsmill/infrahub/pull/9754/commits/9acd5e4904bda7adf4e3a1306db71b951f7c9e89) | Adds `Task.available_actions` and `TaskActionGenerator`, gating retry/cancel by the run's workflow and current state. |
| 3 | [`518cf0af4`](https://github.com/opsmill/infrahub/pull/9754/commits/518cf0af4e8e019b6081868b5496811cd2ca8945) | `InfrahubTaskRetry` / `InfrahubTaskCancel`, with the load-vs-authorize split (`DeliveryReader`, `DeliveryActionAuthorizer`) and backend tests. |
| 4 | [`124712bfd`](https://github.com/opsmill/infrahub/pull/9754/commits/124712bfdecc30fd5f28ae59d741f471fd8d33d6) | The `tasks` entity (api/domain/ui) and the Retry/Cancel buttons with confirmation dialogs, shown only when an action is available. |
| 5 | [`8dc7cf293`](https://github.com/opsmill/infrahub/pull/9754/commits/8dc7cf293e79f3e285ca2c8b49e873f4ab875b8a) | Reconciles the spec with what was built. |
| 6 | [`655309b91`](https://github.com/opsmill/infrahub/pull/9754/commits/655309b910af14c6b73ea881c7e93f4a37c82a87) | Looks up delivery runs by id and sorts the shared read newest-first so the tests stay deterministic as the Prefect server accumulates runs. |
| 7 | [`72c2855b1`](https://github.com/opsmill/infrahub/pull/9754/commits/72c2855b12d4f626d0fc3f7ec4416a9958d3d37b) | Documents the operability surface. |
| 8 | [`ee7ebfac4`](https://github.com/opsmill/infrahub/pull/9754/commits/ee7ebfac41102f7ccff0103834fd2b4a2ff48dc0) | Review feedback in one commit: a validation error for a malformed task id, a shared retry/cancel result mapper, and the cancel-effect docs aligned to CANCELLING. |
| 9 | [`bc1518b4c`](https://github.com/opsmill/infrahub/pull/9754/commits/bc1518b4cdc5d9c86c264f90fb49451cc4ab186b) | Renders the actions in the object's task-tab detail panel (not just the standalone task page), uses the `mdi:restore` retry icon, points the success toast's "view new run" link at the task view the retry was triggered from, and tags retried deliveries at submission so the task-tab count updates immediately. |

Commit 10 and + are front-end review feedbacks or file regeneration.

## Testing

- Backend unit and functional tests for available-action gating, retry resubmit, and cancel state transition.
- Frontend tests for the mutation hooks and action rendering.
- Verified against a locally built stack: the mutations, `available_actions` field, and frontend are present and served; per-webhook delivery automations are configured by the `webhook-configure` flow.

#### Button 
image

#### Confirmation pop-up
image

#### Notification once done
image

#### Button
image

#### Confirmation pop-up
image

#### Notification once done
image

### Task tab of web hooks panel
Same buttons when you open the task.
image

---
## Summary by cubic
Adds retry and cancel controls for webhook deliveries with state-aware GraphQL actions and UI buttons. Improves delivery logs and authorization, and fixes branchless event processing. Fulfills IFC-2119 (manual retry), IFC-2753 (manual cancel), and IFC-2755 (delivery operability via GraphQL + UI).

- **New Features**
  - User-facing deliveries run as `WEBHOOK_SEND` and carry webhook + branch tags.
  - GraphQL: `Task.available_actions` with reasons; `InfrahubTaskRetry` and `InfrahubTaskCancel` mutations returning `TaskInfo`.
  - Behavior: Retry replays the frozen payload as a new run; Cancel stops remaining retries without recalling a sent request; actions are state-gated and authorized against webhook update permission on the delivery’s branch.
  - UI: `TaskActions` on the task page and in the object's Tasks tab with confirmations, success link to the new run, immediate tab-count updates, and a shared layout.
  - Observability: Each delivery logs target URL, masked request headers, payload (truncated inline; full at debug), and attempt counters.

- **Bug Fixes**
  - Authorize retry/cancel using the delivery’s branch (branchless deliveries use the default branch), not the request branch.
  - Restored webhook processing for branchless events by emitting action parameters via `jinja_parameter`.
  - Cancel reports success only when a run reaches CANCELLING/CANCELLED; the Prefect client adapter now returns the resulting state after orchestration.
  - Malformed task ids return a user-facing validation error; wording standardized to “Retry,” and task-action helpers return null when no task id is present.

<sup>Written for commit 6493473d0ce97476f771cb6a0bae4d7980e1d073. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9754?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2753]: https://opsmill.atlassian.net/browse/IFC-2753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-2119]: https://opsmill.atlassian.net/browse/IFC-2119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/graphql/schema.py
```diff
@@ -111,6 +112,9 @@ class InfrahubBaseMutation(ObjectType):
 
     DiffUpdate = DiffUpdateMutation.Field()
 
+    InfrahubTaskRetry = InfrahubTaskRetry.Field()
```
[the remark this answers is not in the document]
polmichel: Good question, will be tackled later in the feature. Internal task: https://opsmill.atlassian.net/browse/IFC-2836

backend/infrahub/graphql/mutations/task.py
```diff
@@ -0,0 +1,162 @@
...
+    async def mutate(cls, root: dict, info: GraphQLResolveInfo, data: TaskActionInput) -> InfrahubTaskCancel:  # noqa: ARG003
+        graphql_context: GraphqlContext = info.context
+
+        async with get_client(sync_client=False) as client:
+            prefect = PrefectClientAdapter(client)
+            delivery = await DeliveryReader(prefect).read(str(data.id))
+            build_delivery_action_authorizer(graphql_context).authorize(delivery, TaskActionType.CANCEL)
+            await prefect.set_flow_run_state(
+                flow_run_id=UUID(str(data.id)), state=State(type=StateType.CANCELLING), force=False
+            )
```
polmichel: For now, a `SCHEDULED` flow (awaiting a retry) will stay in `CANCELLING` state until it's picked up by a worker (once the scheduled time is reached).

This may be confusing for the user: a `SCHEDULED` run while it's awaiting a retry, and then a `CANCELLING` run for as long as 2 minutes.

A dedicated internal task will look into this. https://opsmill.atlassian.net/browse/IFC-2839

backend/tests/functional/webhook/conftest.py
```diff
@@ -99,10 +110,29 @@ async def prefect_client(prefect_test_fixture: None) -> AsyncGenerator[PrefectCl
...
+async def read_send_runs(querier: FlowRunQuerying) -> list[FlowRun]:
+    return await querier.read_flow_runs(flow_filter=FlowFilter(name=FlowFilterName(any_=["webhook-send"])))
+
+
+def only_new_run(runs: list[FlowRun], seen: set[str]) -> FlowRun:
+    # Other tests in the session leave their own webhook-send runs in the shared Prefect server, so a
+    # test must identify the delivery it just created rather than assume a clean slate.
+    new = [run for run in runs if str(run.id) not in seen]
+    assert len(new) == 1, f"expected exactly one new delivery, found {len(new)}"
+    return new[0]
```
polmichel, line 135: Trying to reduce the flakiness of this Prefect test suite

backend/infrahub/graphql/mutations/task.py
```diff
@@ -50,10 +50,14 @@ async def read(self, task_id: str) -> DeliveryRun:
...
-            ValidationError: When the delivery has aged out of retention.
+            ValidationError: When the id is malformed or the delivery has aged out of retention.
 
         """
-        runs = await self.client.read_flow_runs(flow_run_filter=FlowRunFilter(id=FlowRunFilterId(any_=[UUID(task_id)])))
+        try:
+            flow_run_id = UUID(task_id)
+        except ValueError:
+            raise ValidationError(input_value="This delivery is no longer available.") from None
+        runs = await self.client.read_flow_runs(flow_run_filter=FlowRunFilter(id=FlowRunFilterId(any_=[flow_run_id])))
```
polmichel, line 62: This kind of situation can occur when the retention period (which is defaulted to 7 days today I think) has wiped the data from Prefect database

frontend/app/src/entities/tasks/domain/task-action-result.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!payload) {
+    throw new Error(failureMessage);
+  }
+
+  return {
+    ok: payload.ok ?? false,
+    taskId: payload.task?.id ?? undefined,
+  };
+};
```
bilalabbad: What's the point of returning `ok` here? It's never read, so a failed mutation still shows a success toast. I'd use ok to validate success (throw when it's false), then just return `taskId`

frontend/app/src/entities/tasks/domain/task-action-result.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+interface TaskActionMutationPayload {
+  ok?: boolean | null;
+  task?: { id?: string | null } | null;
+}
+
+// Shared mapping for task-action mutations (retry, cancel): surface a GraphQL error,
+// fail when the mutation returned no payload, otherwise normalise to a TaskActionResult.
+export const mapTaskActionResult = (
```
bilalabbad: Let's drop mapTaskActionResult. Each function should handle its own mapping. Just return like`data?.InfrahubTaskRetry?.task?.id` in each.

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
```diff
@@ -0,0 +1,16 @@
...
+  mapTaskActionResult,
+  type TaskActionResult,
+} from "@/entities/tasks/domain/task-action-result";
+
+export type RetryTaskParams = RetryTaskFromApiParams;
+export type RetryTaskResult = TaskActionResult;
+
+export const retryTask = async (params: RetryTaskParams): Promise<RetryTaskResult> => {
+  const { data, errors } = await retryTaskFromApi(params);
+  return mapTaskActionResult(data?.InfrahubTaskRetry, errors, "Failed to retry the task");
```
bilalabbad: using the pattern in frontend codebase is:
1. call api
2. check for error
3. data validation
4. optional mapping if needed

Here it is unusual for frontend codebase to have a function that wrap everything.

frontend/app/src/pages/objects/object-details/task-details.tsx
```diff
@@ -1,23 +1,35 @@
...
           className="flex items-center p-2"
         >
           <Icon icon="mdi:chevron-left" />
           All tasks
         </Link>
+
+        {taskData && (
+          <div className="ml-auto flex items-center gap-2 p-2">
+            <TaskActions task={taskData} />
+          </div>
```
bilalabbad: why not add a props className to TaskActions and give it to him, instead of wrapping in a div?

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,141 @@
...
+      const message = result.taskId ? (
+        <>
+          Task retried.{" "}
+          <Link to={newRunPath} className="inline-flex items-center gap-1 underline">
+            View new run <ArrowUpRightIcon className="size-3.5" />
+          </Link>
+        </>
+      ) : (
+        "Task retried."
+      );
```
bilalabbad: it is actually possible that taskId does not exist after calling retry?

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,141 @@
...
+      toast(<Alert type={ALERT_TYPES.SUCCESS} message={message} />);
+      await queryClient.invalidateQueries({ queryKey: tasksQueryKeys.all });
+    },
+    onError: (error: Error) => {
+      toast(<Alert type={ALERT_TYPES.ERROR} message={`Error: ${error.message}`} />);
+    },
+  });
+
+  const cancelMutation = useCancelTaskMutation({
+    onSuccess: async (_result: Awaited<ReturnType<typeof cancelTask>>) => {
```
bilalabbad: just remove `_result` since it's not used.

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,141 @@
...
+  (task.available_actions ?? []).some((entry) => entry?.action === action && entry?.available);
+
+export const TaskActions = ({ task }: { task: TaskActionsTask }) => {
+  const queryClient = useQueryClient();
+  const { pathname } = useLocation();
+  const [isRetryConfirmOpen, setIsRetryConfirmOpen] = useState(false);
+  const [isCancelConfirmOpen, setIsCancelConfirmOpen] = useState(false);
+
+  const retryMutation = useRetryTaskMutation({
+    onSuccess: async (result: Awaited<ReturnType<typeof retryTask>>) => {
```
bilalabbad: You do not need to type callback, it's automatically done by useMutation.

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,141 @@
...
+  available: boolean | null;
+  unavailability_reason?: string | null;
+};
+
+export type TaskActionsTask = {
+  id?: string | null;
+  title?: string | null;
+  state?: string | null;
+  available_actions?: (TaskAction | null)[] | null;
+};
```
bilalabbad, line 31: Another question about typing. Are really all theses optional and nullable? For example `id` could be mandatory. Then it'll simply the component by removing checks.

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,141 @@
...
+        description="This stops the task and skips its remaining retries. It will be marked cancelled and won't run again."
+        confirmLabel="Cancel task"
+        cancelLabel="Keep task"
+        confirmVariant="danger"
+        icon="mdi:alert"
+        iconClassName="text-red-600"
+        iconContainerClassName="bg-red-100"
+        isLoading={cancelMutation.isPending}
+        onConfirm={() => cancelMutation.mutate({ id: task.id as string })}
+      />
```
bilalabbad: ModalConfirm grew a bit much here ^^ icon/iconClassName/iconContainerClassName are raw style passthroughs, so the cancel modal hand-paints its danger look at the call site. That danger styling already exists using ModalDelete (same layout, red icon, danger button).

frontend/app/src/shared/components/modals/modal-confirm.tsx
```diff
@@ -21,13 +27,21 @@ export function ModalConfirm({
   description,
   onConfirm,
   isLoading,
+  confirmLabel = "Confirm",
+  cancelLabel = "Cancel",
+  confirmVariant = "primary",
+  icon = "mdi:alert-circle-outline",
+  iconClassName = "text-yellow-600",
+  iconContainerClassName = "bg-yellow-100",
```
bilalabbad: I think we should drop these 4 props and only keep label override. ModalConfirm has a consistent styling

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Since these tasks expose what's inside Prefect, I think the only situation is when Prefect's retention period (currently set to 7 days) has expired. It would mean the UI called Prefect's internal flows before the retention expired, and that the user kept the window open until after the retention expiration. Unlikely, but possible.

frontend/app/src/entities/tasks/domain/task-action-result.ts
polmichel: Good catch. There's no path in the backend that returns false today, I'll remove the `ok` from typing and manage the `ok` value in this code in case this changes one day. 

This also surfaced a backend case worth handling: a delivery can settle between the availability check and the state transition. The cancel mutation now confirms the run actually reaches a cancelling/cancelled state and raises otherwise, instead of reporting a no-op as success. I added a component test on the Prefect adapter verifying it reports the real resulting state (cancelled for an in-flight run, still-terminal for one that already settled), which is the signal the mutation keys off.

frontend/app/src/entities/tasks/domain/task-action-result.ts
polmichel: Ok will do!

polmichel: Unit tests have also been created covering the `cancel-task` and `retry-task` files

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
polmichel: Ok, I think this has been tackled thanks to your precedent feedback https://github.com/opsmill/infrahub/pull/9754#discussion_r3502046738

polmichel: I am not sure whether a _check for error_ part is required here or not since backend raises `ValidationError` with a clear message

frontend/app/src/pages/objects/object-details/task-details.tsx
polmichel: `className` is now a prop of `TaskActions`!

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
polmichel: I've added error management inside the `domain` layer and related unit tests

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: You're right I'll align with the backend nullability reality

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: In order to make the specificities of task cancellation modal, I would like to rename the component `ModalDanger` and we would be able to customize some elements inside:
  | Customizable | Fixed (the "danger" identity) |
  |---|---|
  | `title` | red icon container (`bg-red-100`) |
  | `description` | red warning icon (`text-red-600`, `mdi:warning-outline`) |
  | `confirmLabel` (e.g. "Delete", "Cancel task", "Remove") | danger-styled confirm button |
  | `cancelLabel` (e.g. "Cancel", "Keep task") | overall layout |
  | `onConfirm` (the action) |  |
  | `isLoading` |  |

But as it started to impact 10+ files, so I stopped. Do you think it would be a good direction?

For now, I'll just create a variant inside `ModalConfirm` "danger" so represent these common items.

polmichel: Ok based on your last feedback https://github.com/opsmill/infrahub/pull/9754#discussion_r3502236274, I've finally decicted to do a commit with the refactor for `ModalDanger` idea

frontend/app/src/entities/tasks/domain/task-action-result.ts
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/a39b2e381131e47eb284f1cc496b23a9e325136f but superseded by https://github.com/opsmill/infrahub/pull/9754/changes/f92c0c0260da50c66d1534a3446ae637f7e83bb5

frontend/app/src/entities/tasks/domain/task-action-result.ts
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/f92c0c0260da50c66d1534a3446ae637f7e83bb5

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/f92c0c0260da50c66d1534a3446ae637f7e83bb5

frontend/app/src/pages/objects/object-details/task-details.tsx
polmichel: resolved in https://github.com/opsmill/infrahub/pull/9754/changes/bf869b040aaac53d7a8274ff195e4fa71953d550

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/bf59535119a2bb57b3d627f798c2cecef62b8e3f

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/29c2c4f658d28696f9cbf3a777332d03f9b32e20

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/0a3533331a2188096dc260443bfa14b219777dac

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Resolved in https://github.com/opsmill/infrahub/pull/9754/changes/ea15440e7d10be4356b521e41e6993f61e14cf01

frontend/app/src/shared/components/modals/modal-confirm.tsx
polmichel: Resolved in https://github.com/opsmill/infrahub/commit/da4ce98dd7c144a7e372681fac472873045d8e4f

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const canRetry = isActionAvailable(task, "RETRY");
+  const canCancel = isActionAvailable(task, "CANCEL");
+
+  return (
+    <div className={classNames("flex items-center gap-2", className)}>
+      {canRetry && (
+        <Button
+          variant="outline"
+          onPress={() => setIsRetryConfirmOpen(true)}
+          className="flex items-center gap-2"
```
bilalabbad: you can drop className for both buttons, it does nothing

frontend/app/src/entities/tasks/ui/task-actions.tsx
```diff
@@ -0,0 +1,132 @@
...
+    onError: (error) => {
+      toast(<Alert type={ALERT_TYPES.ERROR} message={`Error: ${error.message}`} />);
+    },
+  });
+
+  const canRetry = isActionAvailable(task, "RETRY");
+  const canCancel = isActionAvailable(task, "CANCEL");
+
+  return (
+    <div className={classNames("flex items-center gap-2", className)}>
```
bilalabbad: when we want a div with `flex items-center gap-2`, we have a dedicated component called `<Row/>`

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
```diff
@@ -0,0 +1,16 @@
...
+export type RetryTaskParams = RetryTaskFromApiParams;
+
+export const retryTask = async (params: RetryTaskParams): Promise<string | undefined> => {
+  const { data, errors } = await retryTaskFromApi(params);
+
+  if (errors?.length) {
+    throw new Error(errors.map((error) => error.message).join("; "));
+  }
+
+  return data?.InfrahubTaskRetry?.task?.id ?? undefined;
```
bilalabbad: when we want to return an "empty" value, we prefer using `null` instead of `undefined`. `undefined` means "absent/not provided," null means "explicitly set to empty."

(same for cancelTask)

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Good point!

polmichel: Resolved in 92731b9805f0d5c2f983e3eedbbc597a68bf51b2

frontend/app/src/entities/tasks/ui/task-actions.tsx
polmichel: Resolved in 92731b9805f0d5c2f983e3eedbbc597a68bf51b2

frontend/app/src/entities/tasks/domain/retry-task/retry-task.ts
polmichel: Resolved in 92731b9805f0d5c2f983e3eedbbc597a68bf51b2

bilalabbad (approved):

backend/infrahub/graphql/mutations/task.py
```diff
@@ -0,0 +1,171 @@
...
+        webhook_id = delivery.parameters["webhook_id"]
+        workflow = await graphql_context.active_service.workflow.submit_workflow(
+            workflow=WEBHOOK_SEND,
+            context=graphql_context.get_context(),
+            parameters={
+                "webhook_id": webhook_id,
+                "webhook_kind": delivery.parameters["webhook_kind"],
+                "webhook_name": delivery.parameters["webhook_name"],
+                "payload": delivery.parameters["payload"],
+                "branch_name": delivery.parameters.get("branch_name"),
```
ogenstad: I'm not certain that it will make a practical difference but it looks like the branch here could be different from the branch used to initiate this specific request. 

If we look here we'd be using the branch from the graphql context when determining permissions.

```python
def build_delivery_action_authorizer(graphql_context: GraphqlContext) -> DeliveryActionAuthorizer:
    return DeliveryActionAuthorizer(
        action_generator=TaskActionGenerator(),
        permissions=graphql_context.active_permissions,
        branch=graphql_context.branch,
    )
```

It could possibly lead to a scenario where it's possible to resend a webhook depending on which branch is currently selected.

In reality it could be that we'd only see the tasks for the current branch in the frontend so that it would never show up in the frontend.

Perhaps not a huge issue, but did you consider any of this?

backend/infrahub/graphql/schema.py
```diff
@@ -111,6 +112,9 @@ class InfrahubBaseMutation(ObjectType):
...
+    InfrahubTaskCancel = InfrahubTaskCancel.Field()
```
ogenstad, line 116: I see that we've already diverged a bit from this with other mutations here, I think it would be nice if we could keep the mutations organized alphabetically within this case as it's easier to find something when just scanning. But I'm not sure we should reorganize the rest of them in this PR as it's unrelated.

ogenstad (approved):
Very clean and easy to follow! Left a comment around a potential corner case issue.

backend/infrahub/graphql/mutations/task.py
polmichel: Good point, I hadn't thought this through. I think you're right, and the best solution would be to authorize against the same branch the webhook send task is triggered from.

backend/infrahub/graphql/schema.py
polmichel: If I think about it I'll try to do some PRs against develop once the webhook feature is merged

PR #9755, opened 2026-06-30
feat(proposed-change): precise regeneration triggers for generators

ogenstad:
# Why

Generators re-run on every file change in any linked repository - the same blunt
gate INFP-409 already replaced for artifacts. A proposed change whose only
repository edit is unrelated (a README, a sibling generator's source, an
untouched query) still dispatches and re-runs every generator's instances,
wasting pipeline work and adding noise to the change.

Goal: re-run only the generators a change actually affects, reusing the
already-shipped INFP-409 closure/predicate machinery rather than new design.

Non-goals: no new predicate logic, no cross-branch fingerprint compare, no
AST-precise import analysis. Artifact selection and log wording are deliberately
left byte-for-byte unchanged. This PR delivers the two gate swaps and their test
coverage (IFC-2810); the user-declared `watch:` field, read-only-repo and
backward-compat verification, docs and changelog land in follow-up tickets.

Ref: IFC-2810, IFC-2738

### Behavioral changes
- A proposed change whose only repository change is outside every generator's
  dependency closure (e.g. a README edit, or a `.py` file no generator reads)
  now dispatches **zero** generators.
- Editing a generator's source file - or a sibling in its package directory -
  re-runs only that generator's instances.
- Editing a `.gql` query re-runs only the generators that use it; a query shared
  by two generators re-runs both; editing both a generator's query and its source
  dispatches it once.
- Every run/skip decision is explained in the task log (which file, query, or
  definition change triggered it), in generator-correct wording.

### Implementation notes
- `run_generators` (definition-level gate): the `DefinitionSelect.FILE_CHANGES`
  "any file changed in a synced repository" heuristic is replaced by
  `_query_changed OR _definition_changed OR _transform_changed`, evaluated per
  definition against its stored dependency closure via `_repo_diff_or_none(...)`.
  The `MODIFIED_KINDS` data-change clause is unchanged.
- `request_generator_definition_check` (per-member gate): `managed_branch` is
  derived from the same three predicates instead of `source_branch_sync_with_git`.
  `_run_generator` itself is untouched, preserving the never-under-run invariant -
  new members (`instance_id is None`) and data-changed instances
  (`impacted_instances`) always run; legacy (`dependencies=null`) or incomplete
  (`dependencies_complete=False`) closures fall back to
  regenerate-on-any-file-change.
- No new abstraction: this reuses the structural `RegenerationDefinition` Protocol,
  the three predicates, the closure builder and the diagnostic wording shipped by
  INFP-409. `ProposedChangeGeneratorDefinition` already satisfies the Protocol.

### What stayed the same
- No schema changes, no GraphQL/API changes in this PR (the schema attributes and
  SDK field landed in the foundation work).
- Artifact selection and log wording are byte-for-byte unchanged - guarded by the
  existing artifact predicate/selection/logging suites plus an explicit noun guard.
- The `MODIFIED_KINDS` data-change path and the `impacted_instances` per-member
  path are unchanged.

### Suggested review order
1. `backend/infrahub/proposed_change/tasks.py` - the two gate swaps (the per-member
   gate in `request_generator_definition_check` is the primary-risk area).
2. `backend/tests/component/proposed_change/test_generator_regen_selection.py` -
   end-to-end selection scenarios.
3. The unit tests (predicate variants, never-under-run, logging).

---
## Summary by cubic
Make generator re-runs precise. Only generators touched by a query/definition change or a file edit inside their dependency closure are selected; unrelated commits dispatch none. Per-member gating now targets only instances of selected generators while keeping never-under-run guarantees.

- **New Features**
  - Replace broad file-change heuristic in `run_generators` with per-definition predicates: `_query_changed` OR `_definition_changed` OR `_transform_changed` using the diff and stored closures; log reasons at INFO. Data-change selection via modified kinds stays the same, and source-branch-only new definitions are selected.
  - Swap the per-member gate in `request_generator_definition_check`: compute `managed_branch` from the same predicates and keep `_run_generator` logic unchanged so new members and impacted instances always run; closure/query/definition changes re-run all existing instances.
  - Add component and unit tests for IFC-2810 (US1–US3, US5): unrelated file edits dispatch none; package-floor closure selects only the owning generator; data changes still select; query edits target one or multiple consumers; combined query+source changes dispatch once; diagnostic nouns and artifact wording are correct.

- **Refactors**
  - Replace deprecated `@expand` usage in `generators/cartags.gql` test fixture with explicit fields to match current GraphQL behavior.

<sup>Written for commit 10d679983b07e15f5ed3629a02a7d2d869866fc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/fixtures/repos/car-dealership/initial__main/generators/cartags.gql
+++ b/backend/tests/fixtures/repos/car-dealership/initial__main/generators/cartags.gql
@@ -1,7 +1,12 @@
 query CarOwner($name: String!) {
   TestingPerson(name__value: $name) {
     edges {
-      node @expand {
+      node {
+        __typename
+        id
+        name {
+          value
+        }
         cars {
           edges {
             node {
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -341,6 +341,7 @@ async def run_proposed_change_data_integrity_check(model: RequestProposedChangeD
 async def run_generators(model: RequestProposedChangeRunGenerators, context: InfrahubContext) -> None:
     await add_tags(branches=[model.source_branch], nodes=[model.proposed_change], db_change=True)
 
+    log = get_run_logger()
     client = get_client()
 
     generators = await client.filters(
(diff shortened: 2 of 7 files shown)


backend/tests/fixtures/repos/car-dealership/initial__main/generators/cartags.gql
```diff
@@ -1,7 +1,12 @@
 query CarOwner($name: String!) {
   TestingPerson(name__value: $name) {
     edges {
-      node @expand {
+      node {
+        __typename
+        id
+        name {
+          value
+        }
```
ogenstad, line 9: This change was because the test previously triggered due to the branch being synced with git so we needed to add `name` as a field otherwise the generator wouldn't be executed.

gmazoyer (approved):

PR #9757, opened 2026-06-30
chore(tasks): build code-generation paths from REPO_BASE, not the escaped path

polmichel:
## Summary

The schema and protocol code-generation tasks built their filesystem paths from `ESCAPED_REPO_PATH`. That value is **shell-escaped** (`escape_path` backslash-escapes characters such as `-`), but the paths are consumed by Python file APIs — Jinja's `FileSystemLoader`, `sys.path`, and `open` — which expect the real path. When the repository path contains an escaped character (for example a worktree directory named `dazed-secure`), generation reads from and writes to a path that does not exist.

## Symptom & paths

Running `uv run invoke backend.generate` from a worktree whose path contains a character that `escape_path` rewrites (here the directory `dazed-secure`, and the `.superset` segment) fails while loading the generation templates:

```text
jinja2.exceptions.TemplateNotFound: 'generate_schema.j2' not found in search path:
'/Users/Pol/\.superset/worktrees/infrahub/dazed\-secure/backend/templates'
```

`escape_path` backslash-escapes shell-special characters (`.`, `-`, …). The result is correct as a shell argument but, when handed to Python's file APIs (`FileSystemLoader`, `sys.path`, `open`), it is a literal path that does not exist.

| | Path used | On disk |
|---|-----------|---------|
| Before — `ESCAPED_REPO_PATH` | `/Users/Pol/\.superset/worktrees/infrahub/dazed\-secure/backend/templates` | ❌ does not exist → `TemplateNotFound` |
| After — `REPO_BASE` | `/Users/Pol/.superset/worktrees/infrahub/dazed-secure/backend/templates` | ✅ resolves, template loads |

The same escaped-vs-real mismatch applies to the `sys.path` insertion and the generated-file output directories, so generation would also resolve those under the wrong, non-existent location.

## Change

Use the raw `REPO_BASE` instead of `ESCAPED_REPO_PATH` for the generation input/output paths in `tasks/backend.py` (schema templates, `sys.path` insertion, and the generated-file output directories).

## Testing

`uv run invoke backend.generate` resolves the template and output paths correctly from a worktree whose path contains a `-`.

---
## Summary by cubic
Use the real repo path (`REPO_BASE`) for schema and protocol code generation instead of the shell-escaped `ESCAPED_REPO_PATH`. This fixes broken Jinja `FileSystemLoader`, `sys.path`, and output paths when the repo path includes characters like `-`, so generation reads/writes to the correct directories.

<sup>Written for commit 856774aa78ee366b442c0355ca795a98cc127507. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9757?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tasks/backend.py
+++ b/tasks/backend.py
@@ -10,7 +10,7 @@ from .shared import (
     PYTHON_PRIMITIVE_MAP,
     execute_command,
 )
-from .utils import ESCAPED_REPO_PATH
+from .utils import ESCAPED_REPO_PATH, REPO_BASE
 
 MAIN_DIRECTORY = "backend"
 NAMESPACE = "BACKEND"
@@ -351,8 +351,8 @@ def _generate_schemas(context: Context) -> None:
         relationship_schema,
     )
 
-    env = Environment(loader=FileSystemLoader(f"{ESCAPED_REPO_PATH}/backend/templates"), undefined=StrictUndefined)
-    generated = f"{ESCAPED_REPO_PATH}/backend/infrahub/core/schema/generated"
+    env = Environment(loader=FileSystemLoader(f"{REPO_BASE}/backend/templates"), undefined=StrictUndefined)
+    generated = f"{REPO_BASE}/backend/infrahub/core/schema/generated"
     template = env.get_template("generate_schema.j2")
 
     attributes_rendered = template.render(schema="AttributeSchema", node=attribute_schema, parent="HashableModel")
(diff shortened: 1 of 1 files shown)


dgarros (approved):

ajtmccarty (approved):
verified that `inv backend.generate` ran correctly on this branch on my WSL

PR #9758, opened 2026-06-30
docs(webhooks): clarify branch scope for proposed change events

iddocohen:
## Why

A customer reported that webhooks for `proposed_change.approved` / `rejected` / `merged` never fire when scoped to **Other Branches**, while `proposed_change_thread.created` does. The webhook form offers every `event_type` × `branch_scope` combination from the schema, with no indication that some combinations can never match.

A proposed change is a branch-agnostic object, so its lifecycle events are emitted on the **default branch**:
- `merged` is explicitly stamped to the default branch.
- `approved` / `rejected` currently land on the default branch as well (they inherit the branch the review mutation runs on, which is the default-branch context).

Events generated by activity *inside* a proposed change (threads/comments, validators, data changes) are emitted on the source branch, so they work with **Other Branches**.

## What

Adds a note to `docs/docs/webhooks/overview.mdx` documenting which branch scope values are meaningful for each proposed-change event, so users don't configure webhooks that silently never fire.

## Scope

Documentation only — describes current (1.10) behaviour. The consistency follow-ups (enforcing approve/reject on the default branch in the backend, reconciling `BranchMergedEvent`, and webhook-form guardrails) are tracked in #9761, #9760 and #9759.


changelog/+webhook-branch-scope.fixed.md
```diff
@@ -0,0 +1 @@
+Documented which branch scope values are meaningful for proposed change webhook events (merge/approve/reject fire on the default branch).
```
polmichel: Do we need to add a changelog entry for documentation changes? Doing so would automatically include this content in the next published release notes. I'm not sure that's what we want.
Internal guidelines do not mention this is required for documentation update -> see `dev/guidelines/changelog.md`.
fyi @saltas888 since you were working on the changelog skill.

docs/docs/webhooks/overview.mdx
```diff
@@ -53,6 +53,23 @@ Webhooks can also be configured to trigger based on branch types:
...
 
+:::warning Branch scope for proposed change events
+
+The branch scope is matched against the branch an event is emitted on. A [proposed change](../reference/infrahub-events/proposed.mdx) is a branch-agnostic object, so the events for its lifecycle are emitted on the **default branch**, regardless of the proposed change's source branch:
+
+- `infrahub.proposed_change.merged`
+- `infrahub.proposed_change.approved`
+- `infrahub.proposed_change.rejected`
+
+Webhooks for these events must use the **Default Branch** or **All Branches** scope. Scoping them to **Other Branches** means the webhook will never fire.
```
PhillSimonds, line 64: LGTM. From a user perspective, It might make sense to open a feature request to disallow scoping these events to branches other than the default branch as that's the only place from which they actually fire.

PhillSimonds (approved):

PR #9763, opened 2026-06-30
feat(lock): add TTL to global initialization locks

### LINKED ISSUE #9277: feature: Add an out-of-band command to clear deadlock / initialization locks when Infrahub is down
### Component

API Server / GraphQL, infrahubctl CLI

### Describe the Feature Request

Provide a way to clear stuck Infrahub locks (e.g. the Redis-backed `lock.global.taskmgr.init` lock and other `lock.*` keys) that does **not** require an Infrahub API server or task worker to be running.

Today the existing in-process "scheduled deadlock cleaner" runs inside Infrahub itself, which means it cannot help when a held lock is preventing Infrahub from starting in the first place.

This could land as:

- a new `infrahubctl` subcommand (e.g. `infrahubctl admin locks list` / `infrahubctl admin locks clear [<key>|--all-stale]`) that talks directly to the configured Redis using the same connection settings Infrahub uses, or
- a standalone management script shipped with the server image that operators can `kubectl exec` to run (e.g. `python -m infrahub.management.locks clear`), or
- both, with the management script being the engine that the `infrahubctl` command wraps.

Whichever shape it takes, the key constraints are:

1. It must work when Infrahub server / task workers are **not** running.
2. It must use the same Redis connection configuration the server uses, so operators don't have to re-derive it.
3. It should be able to list current locks and selectively clear one by name (e.g. `lock.global.taskmgr.init`) in addition to a "clear all stale" mode.

### Describe the Use Case

Real-world incident:

Prefect workers were rebooted while the task-manager init lock (`lock.global.taskmgr.init`) was held in Redis. The lock was never released. On restart, Infrahub's task worker hangs during `setup_task_manager()` because the lock is still present; the API server's startup stalls indirectly as a result. Gunicorn binds to `:4200`, the lifespan handler waits on Prefect, Prefect waits on the lock, and the lock never clears because **the only mechanism Infrahub has to clear stale locks runs inside the very service that's blocked from starting.**

The recovery path today is:

1. Find someone with `kubectl` access to the customer's cluster.
2. `kubectl exec` into the Redis pod.
3. Run `redis-cli`.
4. `keys lock.*` to list locks and confirm `lock.global.taskmgr.init` is held.
5. `del lock.global.taskmgr.init`.
6. Restart Infrahub.

This is workable if you're on the team that owns the cluster. It is genuinely painful when the operator running into the failure is a customer — and unworkable in cases where the SE doesn't have cluster access at all.

Per CS, this pattern has come up multiple times across customer engagements — this is not a one-off.

A first-class CLI command would:

- Make recovery a one-line, documented step.
- Avoid teaching customers to `redis-cli del` keys by hand.
- Let CS hand customers a safe, copy-pasteable procedure instead of asking them to exec into the data store.
- Be usable from CI / runbooks as part of an "Infrahub won't start" recovery playbook.

### Additional Information

Relevant context:

- The specific lock in this incident: `lock.global.taskmgr.init` — the task-manager initialization lock that `infrahub.services.adapters.workflow.worker.setup_task_manager()` acquires.
- Symptoms on the customer side: Prefect server's gunicorn binds on `:4200`, then the worker stalls during lifespan; Infrahub task worker raises `httpx.ReadTimeout` against `http://prefect-server:4200/api/`; application startup fails; gunicorn exits with `Reason: Worker failed to boot`.
- The existing in-process deadlock cleaner only runs when Infrahub is up — by definition it cannot solve "Infrahub won't start because a lock is held."

Suggested UX (sketch):

```
$ infrahubctl admin locks list
KEY                                TTL       AGE      HOLDER
lock.global.taskmgr.init           none      00:42:11 task-worker@<pod-name>
lock.global.schema.load            5m 12s    00:00:18 api-worker@<pod-name>

$ infrahubctl admin locks clear lock.global.taskmgr.init
Cleared lock 'lock.global.taskmgr.init'.

$ infrahubctl admin locks clear --all-stale
Cleared 1 stale lock(s): lock.global.taskmgr.init
```

Docs ask: once the command lands, please also add a "Recovering from a stuck initialization lock" runbook to docs.infrahub.app.

fatih-acar:
## Summary

When a worker died while holding a **global initialization lock**, the lock was never released. On restart, Infrahub stalled during initialization — the task worker hung in `setup_task_manager()`, the Prefect bootstrap hung in `init_prefect()`, and the API server's startup stalled behind them. The only existing recovery was for an operator to `kubectl exec` into Redis and `del` the lock key by hand. The in-process deadlock cleaner can't help here — it runs inside the very service that's blocked from starting.

This gives the global initialization locks a **time-to-live** so Redis expires a stale lock on its own and the platform recovers without manual intervention.

## Key Changes

- Global init locks now auto-expire after a configurable TTL (**20 minutes by default**, `INFRAHUB_CACHE_INIT_LOCK_TTL_MINS`). A dead holder can no longer deadlock startup indefinitely.
- Covers all three init locks: `global.init` (registry/Root init), `global.taskmgr.init` (Prefect server bootstrap — the lock from the incident), and `global.worker.taskmgr.init` (worker task-manager setup). The TTL is applied centrally by lock name, so every acquisition site gets the safety net.
- Releasing a lock that already expired is treated as a no-op instead of raising.
- Centralized the three init-lock names in `lock.py` and removed a duplicate `GLOBAL_TASKMGR_INIT_LOCK` constant (a different value lived in `prefect_server/app.py`).

Enforced with the Redis cache driver (the default, and the driver in the reported incident). NATS locks are unchanged — its cache uses fixed TTL buckets rather than per-key TTL.

## Related Context

Closes #9277.

The issue also proposed a separate out-of-band `infrahubctl admin locks` CLI command to list/clear locks. That is a larger, independent piece and is **not** included here — this PR delivers the TTL-based auto-recovery, which removes the "permanent deadlock" failure mode at its root.

## Documentation Updates

- Regenerated `docs/docs/reference/configuration.mdx` with the new `INFRAHUB_CACHE_INIT_LOCK_TTL_MINS` setting.

## Test Plan

- `uv run pytest backend/tests/unit/test_lock.py backend/tests/unit/adapters/test_lock.py` — TTL wiring, default value, and the release-after-expiry swallow/re-raise behavior.
- `uv run pytest backend/tests/component/services/adapters/redis/test_lock_ttl.py` — against a real Redis (testcontainers): a TTL'd lock is dropped by Redis on its own and releasing it afterward does not raise; a lock without a TTL persists until released.
- CI: `ruff`, `mypy`, and generated-doc validation.

---
## Summary by cubic
Adds a TTL to global initialization locks so stale locks auto-expire and don’t deadlock startup after a worker crash. Default is 20 minutes via `INFRAHUB_CACHE_INIT_LOCK_TTL_MINS` (addresses NI-508 / #9277).

- **Bug Fixes**
  - TTL applied automatically to `global.init`, `global.taskmgr.init`, and `global.worker.taskmgr.init` via `InfrahubLockRegistry`; lock constants centralized in `lock.py` and used by Prefect and worker.
  - Releasing an already expired TTL’d lock is a no-op; without a TTL, losing the lock re-raises (with a warning if expired).
  - Redis-only behavior; NATS unchanged. Docs and `docker-compose` updated for `INFRAHUB_CACHE_INIT_LOCK_TTL_MINS`.

<sup>Written for commit 0b3dfe5a9d800668e93ddf6ac60c7a6a20b58d6c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ajtmccarty (approved):

fatih-acar:
Thanks @ajtmccarty!

PR #9769, opened 2026-06-30
feat(cache): NATS per-key TTL + init lock TTL on NATS

fatih-acar:
## Summary

Builds on the Redis init-lock TTL work (this PR targets that branch). It modernizes the NATS cache backend and brings the global initialization-lock TTL safety net to NATS as well, so a worker that dies mid-initialization can no longer deadlock startup regardless of cache driver.

## Key Changes

**Commit 1 — `feat(cache): switch NATS cache to per-key TTL`**
- Bumps `nats-py` 2.7.2 → 2.15.0.
- Replaces the per-bucket TTL workaround (one KV bucket per supported TTL, keys routed by prefix) with real per-message (per-key) TTL, available since NATS server 2.11.
- A single KV bucket now applies the caller-supplied expiry directly via `msg_ttl`, so NATS finally honors the same per-key expiry Redis already did. Existing buckets are upgraded in place to enable per-message TTL.
- The cache `set(expires=…)` value is now seconds (`KVTTL` members remain valid as they are integers). `RedisCache` aligned accordingly.
- Test NATS image pinned to `nats:2.11-alpine`.

**Commit 2 — `fix(lock): enforce init lock TTL on the NATS backend`**
- Fixes NATS-backed distributed locks: `NATSLock.acquire()` did not accept the `token` `InfrahubLock` passes, so acquiring any NATS lock raised a `TypeError`.
- Global initialization locks acquired through NATS now carry the configurable TTL via the per-key cache expiry, matching the Redis behavior. A TTL'd lock is released only if we still own it, so an expired-and-reacquired lock is not clobbered.

## Related Context

Follow-up to the Redis global-init-lock TTL (issue #9277). The TTL value/config (`INFRAHUB_CACHE_INIT_LOCK_TTL_MINS`, default 20 min) is defined on the base branch and reused here.

## Test Plan

Validated against a real NATS server (`nats:2.11-alpine`, server v2.11.17) with the NATS cache driver:

- `INFRAHUB_CACHE_DRIVER=nats uv run pytest backend/tests/component/services/adapters/nats` — cache per-key TTL (create + put paths), no-TTL persistence, list_keys; init lock auto-expiry, configured-TTL wiring, acquire/release cycle.
- `uv run pytest backend/tests/component/services/adapters/redis/test_redis.py` and `.../redis/test_lock_ttl.py` — Redis path unchanged.
- `uv run pytest backend/tests/unit/test_lock.py backend/tests/unit/adapters/test_lock.py` — local/Redis lock units.
- `ruff`, `ty` clean on changed files.

## Notes

- NATS message-bus adapter API usage was checked for compatibility with the `nats-py` bump (publish/subscribe/add_stream/add_consumer signatures unchanged; `publish` gains an optional `msg_ttl`).
- These NATS tests run in CI only when the suite is exercised with the NATS cache driver.

---
## Summary by cubic
Switches the NATS cache to real per-key TTL and adds a configurable TTL to all global initialization locks so a crashed worker can’t block startup. Follow-up to #9277, aligning NATS with Redis behavior and defaulting TTL to 20 minutes via `INFRAHUB_CACHE_INIT_LOCK_TTL_MINS`.

- New Features
  - NATS uses per-message (per-key) TTL in a single KV bucket; existing buckets auto-upgrade to enable per-message TTL.
  - Global init locks (`global.init`, `global.taskmgr.init`, `global.worker.taskmgr.init`) now auto-expire on both Redis and NATS; TTL is passed in seconds across caches.
  - `cache.set(expires=...)` now takes seconds; `KVTTL` values still work since they’re ints. `RedisCache` and NATS both accept int seconds.
  - Requires NATS server 2.11+; bumped `nats-py` to `2.15.0`; tests pin `nats:2.14.3-alpine`.

- Bug Fixes
  - Fixed NATS lock acquire TypeError: `NATSLock.acquire()` now accepts an optional token to match `InfrahubLock` and keeps the `timestamp::worker_id` value.
  - Releasing TTL’d locks is safe: no-op if the lock already expired or we no longer own it (Redis handles `LockNotOwnedError`; NATS checks the token).

<sup>Written for commit 36d09628d8e4ab092bdbe222839e1acc00bf6834. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/lock.py
+++ b/backend/infrahub/lock.py
@@ -90,10 +90,12 @@ class InfrahubMultiLock:
 class NATSLock:
     """Context manager to lock using NATS."""
 
-    def __init__(self, service: InfrahubServices, name: str) -> None:
+    def __init__(self, service: InfrahubServices, name: str, ttl: int | None = None) -> None:
         self.name = name
-        self.token = None
+        self.token: str | None = None
         self.service = service
+        # Maximum lifetime in seconds for the lock.
+        self.ttl = ttl
 
     async def __aenter__(self) -> None:
         await self.acquire()
@@ -106,8 +108,8 @@ class NATSLock:
     ) -> None:
         await self.release()
 
-    async def acquire(self) -> None:
-        token = f"{current_timestamp()}::{WORKER_IDENTITY}"
+    async def acquire(self, token: str | None = None) -> None:
+        token = token or f"{current_timestamp()}::{WORKER_IDENTITY}"
         while True:
             if await self.do_acquire(token):
                 self.token = token
@@ -115,10 +117,16 @@ class NATSLock:
             await sleep(0.1)  # default Redis GlobalLock value
 
     async def do_acquire(self, token: str) -> bool | None:
-        return await self.service.cache.set(key=self.name, value=token, not_exists=True)
+        return await self.service.cache.set(key=self.name, value=token, not_exists=True, expires=self.ttl)
 
     async def release(self) -> None:
+        if self.ttl is not None and await self.service.cache.get(key=self.name) != self.token:
+            # The TTL elapsed and the lock may have been re-acquired by another worker; nothing of ours
+            # to release.
+            self.token = None
+            return
         await self.service.cache.delete(key=self.name)
+        self.token = None
 
     async def locked(self) -> bool:
         return await self.service.cache.get(key=self.name) is not None
@@ -160,7 +168,7 @@ class InfrahubLock:
         elif config.SETTINGS.cache.driver == config.CacheDriver.Redis:
             self.remote = GlobalLock(redis=self.connection, name=f"{LOCK_PREFIX}.{self.name}", timeout=ttl)
         else:
-            self.remote = NATSLock(service=self.connection, name=f"{LOCK_PREFIX}.{self.name}")
+            self.remote = NATSLock(service=self.connection, name=f"{LOCK_PREFIX}.{self.name}", ttl=ttl)
 
     @property
     def acquire_time(self) -> int:
--- a/backend/infrahub/services/adapters/cache/nats.py
+++ b/backend/infrahub/services/adapters/cache/nats.py
@@ -1,45 +1,40 @@
 from __future__ import annotations
 
+import contextlib
 import ssl
+from typing import TYPE_CHECKING
 
 import nats
+import nats.js.api
+import nats.js.errors
 
 from infrahub import config
-from infrahub.message_bus.types import KVTTL
 from infrahub.services.adapters.cache import InfrahubCache
 
-# Inherit from BaseModel to avoid implementing a `__init__` otherwise `cls(...)` fails.
+if TYPE_CHECKING:
+    from infrahub.message_bus.types import KVTTL
 
 
 class NATSCache(InfrahubCache):
     connection: nats.NATS
     jetstream: nats.js.JetStreamContext
-    kv: dict[int, nats.js.kv.KeyValue]
-    kv_buckets: dict[str, KVTTL]
+    kv: nats.js.kv.KeyValue
+    bucket: str
 
     def __init__(
         self,
         connection: nats.NATS,
         jetstream: nats.js.JetStreamContext,
-        kv: dict[int, nats.js.kv.KeyValue],
-        kv_buckets: dict[str, KVTTL],
+        kv: nats.js.kv.KeyValue,
+        bucket: str,
     ) -> None:
         self.connection = connection
         self.jetstream = jetstream
         self.kv = kv
-        self.kv_buckets = kv_buckets
+        self.bucket = bucket
 
     @classmethod
     async def new(cls) -> NATSCache:
-        # FIXME: remove once NATS supports TTL for keys (2.11)
-        kv_buckets = {
-            NATSCache._tokenize_key_name("validator_execution_id:"): KVTTL.TWO_HOURS,
-            NATSCache._tokenize_key_name("workers:primary:"): KVTTL.FIFTEEN,
-            NATSCache._tokenize_key_name("workers:schema_hash:branch:"): KVTTL.TWO_HOURS,
-            NATSCache._tokenize_key_name("workers:active:"): KVTTL.FIFTEEN,
-            NATSCache._tokenize_key_name("workers:worker:"): KVTTL.TWO_HOURS,
-        }
-
         tls_context = None
         if config.SETTINGS.cache.tls_enabled:
             tls_context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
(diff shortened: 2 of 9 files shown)


fatih-acar:
Not an important one, we do not officially support NATS today anyway.
It was done in the background as bonus to stay consistent with the recent changes regard to init lock TTLs.

backend/infrahub/services/adapters/cache/nats.py
```diff
@@ -57,36 +48,40 @@ async def new(cls) -> NATSCache:
...
+            stream = await jetstream.stream_info(f"KV_{bucket}")
+            stream.config.allow_msg_ttl = True
+            await jetstream.update_stream(config=stream.config)
+            return await jetstream.key_value(bucket)
 
     @staticmethod
     def _tokenize_key_name(key: str) -> str:
         return key.replace(":", ".")
 
-    # FIXME: remove once NATS supports TTL for keys (2.11)
```
ajtmccarty, line 75: wow. a FIXME getting cleaned up. nice to see

backend/infrahub/services/adapters/cache/redis.py
```diff
@@ -55,8 +50,10 @@ async def list_keys(self, filter_pattern: str) -> list[str]:
 
         return [key.decode() for key in keys]
 
-    async def set(self, key: str, value: str, expires: KVTTL | None = None, not_exists: bool = False) -> bool | None:
-        return await self.connection.set(name=key, value=value, ex=expires.value if expires else None, nx=not_exists)
+    async def set(self, key: str, value: str, expires: int | None = None, not_exists: bool = False) -> bool | None:
+        # int() normalizes both plain ints and KVTTL (IntEnum) members to a value redis-py can encode.
```
ajtmccarty: the comment makes it look like the typing should be `expires: int | KVTTL | None = None`, but I don't think that's right

backend/infrahub/lock.py
```diff
@@ -90,10 +90,13 @@ async def release(self) -> None:
...
     """Context manager to lock using NATS."""
 
-    def __init__(self, service: InfrahubServices, name: str) -> None:
+    def __init__(self, service: InfrahubServices, name: str, ttl: int | None = None) -> None:
         self.name = name
-        self.token = None
+        self.token: str | None = None
         self.service = service
+        # Maximum lifetime in seconds for the lock. When set, the key carries a per-key TTL so a
+        # crashed holder cannot deadlock the system indefinitely.
```
ajtmccarty: only the first sentence is necessary

backend/infrahub/lock.py
```diff
@@ -106,19 +109,25 @@ async def __aexit__(
     ) -> None:
         await self.release()
 
-    async def acquire(self) -> None:
-        token = f"{current_timestamp()}::{WORKER_IDENTITY}"
+    async def acquire(self, token: str | None = None) -> None:
```
ajtmccarty, line 111: is supporting `token` in here really necessary? it appears unused

backend/tests/component/services/adapters/nats/test_lock_ttl.py
```diff
@@ -0,0 +1,72 @@
...
+    await asyncio.sleep(3)
+    # NATS expired the key without anyone releasing the lock, so a crashed holder cannot deadlock startup.
+    assert await service.cache.get(key) is None
+
+    # Releasing a lock that already expired must not raise back to the caller.
+    await lock_obj.release()
+    await service.cache.close_connection()
+
+
+async def test_init_locks_carry_configured_ttl(nats: dict[int, int] | None) -> None:
```
ajtmccarty: this one just seems like it is testing a setting and maybe not necessary

ajtmccarty (approved):

backend/infrahub/services/adapters/cache/nats.py
fatih-acar: personal affair to me, happy to clean that up 2 years later :D

backend/infrahub/services/adapters/cache/redis.py
fatih-acar: yeah dumb agent, we should keep `expires: KVTTL | None = None` actually since all callers use KVTTL
good catch ty

backend/infrahub/lock.py
fatih-acar: good point, it's not necessary
when NATS was introduced the value was not set at all even in the redis implementation but iirc the NATS one makes the value mandatory
but now that we set the value to a timestamp regardless of the implementation, I think we can keep that for NATS as well and remove that token argument

backend/infrahub/lock.py
fatih-acar: Trimmed to the first sentence.

backend/tests/component/services/adapters/nats/test_lock_ttl.py
fatih-acar: Removed this test.

backend/infrahub/lock.py
fatih-acar: Removed the token argument. NATSLock.acquire() now generates its own timestamp::worker_id value; InfrahubLock only forwards a token to the Redis lock (which stores it as its value).

backend/infrahub/services/adapters/cache/redis.py
fatih-acar: I kept the shared `set` signature as `expires: int | None` (seconds) rather than `KVTTL | None`. It turns out not all callers use `KVTTL`: the NATS init-lock TTL passes a configurable seconds value (`init_lock_ttl_mins * 60`, default 1200), which isn't a `KVTTL` member. Since `InfrahubCache.set` is a single shared signature, narrowing only `RedisCache` to `KVTTL` while the base stays broader would be an LSP violation.

Existing cache callers still pass `KVTTL` (fine, it's an `IntEnum`). I removed the misleading comment; redis just coerces the `IntEnum` to a plain int for the driver. Happy to switch to a dedicated `KVTTL`-typed cache API + a separate seconds path for the lock if you'd prefer keeping `expires` strictly `KVTTL`.

backend/infrahub/lock.py
fatih-acar: Correction to my previous reply: kept the token argument after all. InfrahubLock now generates the token once and passes the same value to both the Redis and NATS locks, so the NATS lock stores the identical timestamp::worker_id value (rather than generating its own).

backend/infrahub/services/adapters/cache/redis.py
fatih-acar: Updated: typed the cache `set` expiry as `KVTTL | int | None` across the abstract and both adapters. Cache callers keep passing `KVTTL`; the NATS init lock passes its configurable seconds value as an `int`. (A single shared type is required — ty rejects a narrower `KVTTL` override of an `int` abstract as an LSP violation.)

PR #9777, opened 2026-07-01
feat(webhook): suppress the traceback for classified delivery failures

polmichel:
## Why

Prefect's flow/task engine logs every raised exception with a full stacktrace. A classified webhook delivery failure (connection, TLS, timeout, HTTP status, configuration) is an expected operational outcome that is already reported as a clean, user-facing reason, so its traceback is noise in the run logs (the Tasks tab).

Ref: https://opsmill.atlassian.net/browse/IFC-2846

## What changed

- A classified delivery failure no longer emits a stacktrace to the Prefect run logs. The run still fails, and its clean classified reason is preserved, with a remediation hint and, while retries remain, the next-attempt note.
- A failed delivery now records a single classified line per attempt, with no redundant task-level failure record echoing the same reason.
- Unexpected errors are untouched: they keep their full traceback and surface as genuine crashes.

## Implementation

Two parts: a generic logging layer that drops registered tracebacks, and the send path shaped so the only exception Prefect logs for an expected failure is the registered one.

Logging layer (`backend/infrahub/log.py`):

- `TracebackSuppressionFilter` — a `logging.Filter` that drops a record whose exception is a registered type. Matching is by **exact type identity** against an injected registry.
- `@suppress_traceback_in_logs` — a class decorator that registers an exception type into that shared registry. `WebhookDeliveryError` opts in with it, so the contract lives on the class and cannot drift from it.
- `configure_logging` installs the filter on the `prefect.flow_runs` / `prefect.task_runs` loggers, after the `prefect.main` import so it survives Prefect's logging reset. The registry is read per record, so a type registered after install still takes effect.

Send path (`backend/infrahub/webhook/tasks/process.py`):

- An expected delivery failure is classified and raised as `WebhookDeliveryError` at the point of the POST, so the exception Prefect records for the failed attempt is the registered one (its traceback is dropped) rather than the raw transport stacktrace.
- The POST runs inline within the `webhook-send` flow instead of as its own task, so a failed attempt does not add a second, redundant task-state failure record to the run logs. Retries stay on the send flow, so delivery behavior is unchanged.

The filter and registry carry no Prefect knowledge; only the install site (the two logger names) is Prefect-specific.

### Rejected implementations

- `isinstance` matching (would suppress unrelated subclasses)
- a sniffed marker attribute (drifts from the class)

## How to test

```bash
cd backend && uv run pytest \
  tests/unit/test_log.py \
  tests/unit/webhook/test_classifier.py \
  tests/unit/webhook/test_process.py \
  tests/component/webhook/test_traceback_suppression.py \
  tests/functional/webhook/test_process.py
```

The component test drives real flows through Prefect's engine: a classified failure logs no traceback while retaining the clean reason; an unclassified failure still logs its full stacktrace. The functional test drives the real send path and asserts a failed delivery leaves no traceback in the run logs.

## Checklist

- [x] Tests added/updated

### Failing webhook send task
image

---
## Summary by cubic
Suppress stack traces for classified webhook delivery failures in Prefect run logs, keeping only the clean reason. Also run the delivery POST inline to avoid redundant failure records. Addresses IFC-2846.

- **New Features**
  - Added `@suppress_traceback_in_logs` and `TracebackSuppressionFilter`; installed on `prefect.flow_runs` and `prefect.task_runs` in `configure_logging`. Exact-type matching via a shared registry; `WebhookDeliveryError` is registered so its traceback is dropped.
  - Classified expected failures during POST and run `webhook_post` inline in `webhook_send`; expected transport/config/status failures raise `WebhookDeliveryError` (no traceback), while unexpected errors keep full tracebacks.
  - Updated tests/docs for flow/task log suppression, HTTP 400/500 bounds, transport-error precedence, fallback message, per-status remediation, filter behavior, and spelling exception for “traceback”.

<sup>Written for commit f952f98136b6542b9425e3c53585cd81381bc635. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -0,0 +1,57 @@
...
+    raise RuntimeError("boom")
+
+
+@pytest.fixture
+def configured_logging() -> None:
+    # Register the traceback filter on the Prefect run loggers, as production startup does.
+    configure_logging(production=False, log_level="DEBUG")
+
+
+@pytest.mark.usefixtures("configured_logging")
```
ogenstad: In a way I like this better for some parts when there's a fixture that isn't actively used and we don't need to have typing ignores for unused parameters within the test. However, I'm not sure we use markers like this anywhere else and from a consistency point of view it might be that we shouldn't include start to use them. One thing that I think we might loose is any kind of "go to definition" when we set it up like this?

backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -0,0 +1,57 @@
...
+        caplog.at_level(logging.INFO, logger="prefect.flow_runs"),
+        pytest.raises(WebhookDeliveryError, match=r"^The target responded with HTTP 404\.$"),
+    ):
+        await _raise_classified_failure()
+
+    # The traceback-bearing record is dropped, so nothing the logger emits carries an exception...
+    assert [record for record in caplog.records if record.exc_info] == []
+    # ...while the classified reason still surfaces (wrapped in Prefect's own state summary, so the
+    # owned reason is matched as a substring rather than pinning Prefect's phrasing).
+    assert CLASSIFIED_MESSAGE in caplog.text
```
ogenstad, line 65: Just a sanity check for this test, did we validate that the logger that outputs to Prefect also follows these rules and that this isn't just filtering what's spit out to the console?

ogenstad (approved):

backend/tests/component/webhook/test_traceback_suppression.py
polmichel: Currently testing it, this is the last step before putting the PR ready for review. I'll add a dedicated application testing section inside the PR description

backend/tests/component/webhook/test_traceback_suppression.py
polmichel: Good point, I also think it's better to use direct reference. I've missed that one.

PR #9783, opened 2026-07-02
Detect failed merges and hold write protection

ajtmccarty:
# Why

When a task worker is killed mid-merge (e.g. SIGKILL), the source branch is left stuck in `MERGING` with a partially-merged graph and the global merge lock still held by the dead worker. Nothing detected this state, so the default-branch write-protection either lingered indefinitely or — worse — writes could resume against a half-merged default branch, and users got no signal that anything was wrong.

**Goal:** deterministically detect a dead merge, record it durably as `MERGE_FAILED`, and keep the default branch write-blocked with a distinct, non-retryable error that tells the operator recovery is required — while never mis-flagging a healthy in-progress merge.

**Non-goals (follow-ups):**
- The recovery itself — `infrahub recover` / range rollback / branch + proposed-change reset (US3, separate PR).
- Structured detection/recovery logging and richer visibility (US4).

Part of the merge-failure-recovery epic **IFC-2559**; this is the **US2 (detection)** slice of umbrella bug **IFC-2437** and advances **IFC-2565** (surface a crashed-merge error to the operator). It does **not** close IFC-2437 — recovery lands in a follow-up.

Closes IFC-2565

## What changed

**Behavioral changes (what you'll observe differently):**

- New durable branch status **`MERGE_FAILED`**, auto-exposed in the GraphQL `BranchStatus` enum. It is set only by detection and cleared only by recovery (not yet in this PR).
- A recurring **`MERGE_WATCHER`** workflow (1-minute cron, single-flighted across workers) plus a best-effort check at **API-server startup** flip a branch that is stuck in `MERGING` whose merge-lock holder is no longer a live worker (after a grace period) to `MERGE_FAILED`, and hold the shared write-protection.
- Writes blocked by a **failed** merge now return the distinct **`MERGE_RECOVERY_REQUIRED`** GraphQL error code (HTTP 423) with a message naming `infrahub recover` and "contact an administrator" — separate from the transient **`MERGE_IN_PROGRESS`** used during a healthy merge. Payload: `{branch_name, merging_branch}`.
- A healthy in-progress merge (live lock holder, or younger than the grace period) is **never** flagged.
- New config setting **`INFRAHUB_MERGE_FAILURE_GRACE_PERIOD_SECONDS`** (default `180`).

**Implementation notes:**

- Liveness reuses the existing merge-lock token (`{timestamp}::{worker_id}`) compared against `component.list_workers()`' active set — the same pattern as `clean_up_deadlocks`. The decision is a side-effect-free predicate (`is_failed_merge`) that is unit-tested without a DB; the grace period is injected at construction.
- `MergeFailureIdentifier.scan()` runs detection then reconciliation in one pass. Reconciliation re-aligns the volatile `merge:protected` cache key with the durable `MERGING`/`MERGE_FAILED` branch status, so protection self-heals after a restart or cache flush.
- `MergeRecoveryRequiredError` is deliberately a **sibling** of `MergeInProgressError`, not a subclass: the two states demand opposite reactions (retry vs. run recovery), and subclassing would let any handler written for the retryable error silently absorb the durable one.
- Lock-token parsing now lives in `lock.py` beside the token format (a `get_worker_id_from_lock_token` helper), and a multi-status filter (`BranchListFilters.statuses` → `n.status IN …`) lets reconcile query protected branches in one round-trip.
- The `python_sdk` submodule is bumped so SDK clients can deserialize a `MERGE_FAILED` branch without a validation error.

**What stayed the same:** no database migration; the healthy-merge happy path is unchanged; the recovery flow is out of scope.

### Suggested review order

1. `945f7a18c` + `421f3828b` — the `MergeRecoveryRequiredError` exception and regenerated error-catalogue bindings/tests for `MERGE_RECOVERY_REQUIRED`.
2. `cbff3a97c` — add the `MERGE_FAILED` status (+ GraphQL schema regen).
3. `426833218` — detection primitives (grace config, lock-token parser, merge-lock read, multi-status filter).
4. `4ce36eba2` — the detector, watcher cron, startup scan, and the write gate (the meat + tests).
5. `2db7d9548` — docs + spec alignment.
6. The remaining commits are generated-file/CI syncs for the new config setting, the `python_sdk` submodule bump, hardening of the SIGKILL integration test, and post-review refactors (rename to `MergeFailureIdentifier`/`failure_identifier.py`, grace-period constructor injection, docstring corrections).

## How to review

- Focus: `backend/infrahub/core/merge/failure_identifier.py` (the predicate + `scan`), and `backend/infrahub/branch/status_checker.py` (the `MERGE_FAILED` gate raising `MergeRecoveryRequiredError`).
- Worth extra scrutiny: the failure predicate's conditions (why an **absent** lock is deliberately *not* auto-flagged), and the sibling-not-subclass choice for `MergeRecoveryRequiredError`.
- End-to-end coverage: `backend/tests/integration_docker/test_merge_kill_recovery.py` SIGKILLs a task worker mid-merge against a real docker stack (with a ~1s grace period in the test compose file) and asserts the idle scan flips the branch to `MERGE_FAILED`. The write-gate contract test lives in `backend/tests/integration/merge/test_failed_merge_protection.py`.
- Mechanical / generated (skim for drift only): `schema/schema.graphql`, `schema/error-catalogue.json`, `schema/openapi.json`, `frontend/app/src/shared/api/**/generated`, `frontend/app/src/shared/api/rest/types.generated.ts`, `docs/docs/reference/error-catalogue.mdx`, `docs/docs/reference/configuration.mdx`, regenerated docker-compose env vars.

## Impact & rollout

- **Backward compatibility:** additive only — `MERGE_FAILED` is a new `BranchStatus` enum value and `MERGE_RECOVERY_REQUIRED` a new catalogue code (marked `evolving`). Clients that exhaustively switch on either enum should handle the new value. No migration.
- **⚠️ Operational caveat:** recovery (`infrahub recover`) is **not** in this PR. Until the recovery PR lands, a branch that reaches `MERGE_FAILED` stays write-blocked with no in-product way to clear it (manual DB intervention only). Land the recovery follow-up close behind this.
- **Performance:** the write gate is unchanged (one cache `GET` per top-level mutation). Detection adds one parameterized branch query + one cache read per worker per minute.
- **Config/env changes:** new `INFRAHUB_MERGE_FAILURE_GRACE_PERIOD_SECONDS` (default 180).
- **Deployment notes:** safe to deploy; the `MERGE_WATCHER` cron deployment is created automatically once the workflow definition is present.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — intentionally omitted; to be folded into the larger merge-failure-recovery docs update once recovery lands
- [x] External docs updated (generated error-catalogue reference reflects `MERGE_RECOVERY_REQUIRED`; configuration reference reflects the new grace-period setting)
- [x] Internal .md docs updated (`dev/knowledge/backend/branch-status.md`, `dev/specs/ifc-2437-merge-failure-recovery/*`)

---
## Summary by cubic
Detects and guards against failed merges to keep the default branch consistent. Adds `MERGE_FAILED` with a distinct `MERGE_RECOVERY_REQUIRED` (HTTP 423) response, runs a 1‑min single‑flight `MERGE_WATCHER` plus an API‑server startup scan, and exposes a grace‑period setting (IFC-2437).

- **New Features**
  - Detection: flip branches stuck in `MERGING` to `MERGE_FAILED` when the merge‑lock holder is dead and past grace; never flag a healthy merge. Hold shared write protection and reconcile the `merge:protected` key with durable status so it self‑heals after restarts or cache flushes. Runs via `MERGE_WATCHER` (every minute, single‑flight) and on API‑server startup.
  - Write gates: raise `MergeInProgressError` for `MERGING` and `MergeRecoveryRequiredError` (code `MERGE_RECOVERY_REQUIRED`, HTTP 423) for `MERGE_FAILED` with payload `{branch_name, merging_branch}`; applies to the source and default branches. Cache‑down fallback enforces from DB and prioritizes `MERGE_FAILED`.  
  - Runtime/contracts: new `BranchStatus.MERGE_FAILED` (GraphQL/SDK); server‑side multi‑status filter (`statuses`) for safe DB fallback; lock‑token worker‑id parser and read‑only merge‑lock holder lookup. Error catalogue/formatter updated; frontend error bindings and GraphQL/OpenAPI/REST types regenerated; SDK submodule bumped. Tests include unit, component, functional, and a skipped-by-default SIGKILL integration.

- **Migration**
  - No breaking changes. Clients may see `MERGE_FAILED` and `MERGE_RECOVERY_REQUIRED`.
  - New setting `merge_failure_grace_period_seconds` (default 180), exposed as `INFRAHUB_MERGE_FAILURE_GRACE_PERIOD_SECONDS`.
  - Ensure workers run the internal `MERGE_WATCHER` workflow.

<sup>Written for commit f55ee890fda579ef13030e3b11b781f5fb278804. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/docs/reference/error-catalogue.mdx
```diff
@@ -284,6 +284,41 @@ The write was rejected because a branch merge is in progress. The block is trans
 }
 ```
 
+### MERGE_RECOVERY_REQUIRED
+
+The write was rejected because a previous branch merge failed and left the default branch protected. Recovery is required: an administrator must run `infrahub recover`. Unlike MERGE_IN_PROGRESS this is not retryable.
```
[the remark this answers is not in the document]
ajtmccarty: it will be added in this release

schema/error-catalogue.json
```diff
@@ -300,6 +300,30 @@
         "type": "object"
       }
     },
+    "MERGE_RECOVERY_REQUIRED": {
+      "description": "The write was rejected because a previous branch merge failed and left the default branch protected. Recovery is required: an administrator must run `infrahub recover`. Unlike MERGE_IN_PROGRESS this is not retryable.",
```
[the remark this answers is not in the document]
ajtmccarty: added in the SDK and the submodule commit has been bumped to include it

dev/specs/ifc-2437-merge-failure-recovery/contracts/internal-api.md
```diff
@@ -179,42 +179,48 @@ deleted/reopened; the `updated_at >= merge_at` filter excludes already-restored
...
     def __init__(
         self,
         db: InfrahubDatabase,
-        diff_merger: DiffMerger,       # owns the range rollback query (no get_affected_node_uuids needed)
-        cache: InfrahubCache,          # merge:protected key set/update/delete; read the lock-holder token
-        component: InfrahubComponent,  # list_workers() -> active-worker set for the liveness predicate
-        merge_locker: MergeLocker,     # read-side helper: current merge-lock holder worker_id (or None)
+        cache: InfrahubCache,                    # read the merge-lock holder token
+        component: InfrahubComponent,            # list_workers() -> active-worker set for the liveness check
+        merge_write_blocker: MergeWriteBlocker,  # owns the merge:protected key (set/get/delete)
```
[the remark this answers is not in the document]
ajtmccarty: I'll clean the spec up for the next PR in the series

backend/infrahub/exceptions.py
```diff
@@ -468,6 +468,26 @@ def __init__(self, identifier: str, message: str, merging_branch: str) -> None:
...
+    """Write rejected because a failed merge needs operator recovery.
+
+    `merging_branch` is the source branch that was being merged (the one whose merge died), matching
+    `MergeInProgressError`; `identifier` is the branch the rejected write targeted (the source branch
+    itself or the default branch).
+
+    Deliberately a sibling of MergeInProgressError, not a subclass: the error-catalogue resolver
+    walks the MRO to map an exception to a code, so subclassing would make this resolve to the
+    transient MERGE_IN_PROGRESS code and be indistinguishable from a retryable in-progress merge.
+    This is the durable, non-retryable signal that an administrator must run `infrahub recover`.
```
polmichel: I am not sure to understand this comment. I think we just remove the constructor from `MergeRecoveryRequiredError` and if we make it inherit from `MergeInProgressError`, the unit test `test_merge_recovery_required_resolves_to_distinct_code` passes.
If a code exists inside the catalogue for the subtype, it returns it. If not, it looks for parent types.

backend/infrahub/core/query/branch.py
```diff
@@ -208,6 +208,11 @@ def _build_raw_filter(self) -> str:
             self._branch_filter_params[param_name] = self.branch_filters.status.value
             conditions.append(f"n.status = ${param_name}")
 
+        if self.branch_filters.statuses:
+            param_name = "filter_statuses"
+            self._branch_filter_params[param_name] = [status.value for status in self.branch_filters.statuses]
+            conditions.append(f"n.status IN ${param_name}")
+
```
polmichel, line 215: Is it worth to keep both of these filters? Does it make a difference in Cypher to use the `IN` for only asserting against one status?

backend/infrahub/core/merge/failure_recovery.py
```diff
@@ -0,0 +1,191 @@
...
+    def is_failed_merge(
+        self,
+        *,
+        status: BranchStatus,
+        lock_holder_worker_id: str | None,
+        active_worker_ids: set[str],
+        merge_started_at: Timestamp | None,
+        now: Timestamp,
+        grace_period_seconds: int,
+    ) -> bool:
```
polmichel: As this is a component isolated from the rest, this is not really important, but having a method named `is_failed_merge` which does not consider the case when a branch comes in with a status of `MERGE_FAILED` looks confusing at first glance.

Based on the current usage inside the code, a branch can only come in with `MERGING` status and this is the first condition of the method, so I guess it's fine. 

I get that this is to determine if a `MERGING` branch should transition to `MERGE_FAILED` status but I think there should be a dedicated comment explaining why `MERGE_FAILED` comes out as false or maybe the method's name is not totally accurate.

Maybe I'm also missing something.

backend/infrahub/core/merge/failure_recovery.py
```diff
@@ -0,0 +1,191 @@
...
+        for branch in merging:
+            merge_started_at = Timestamp(branch.merge_started_at) if branch.merge_started_at else None
+            if not self.is_failed_merge(
+                status=branch.status,
+                lock_holder_worker_id=lock_holder_worker_id,
+                active_worker_ids=active_worker_ids,
+                merge_started_at=merge_started_at,
+                now=now,
+                grace_period_seconds=grace_period_seconds,
+            ):
```
polmichel: If the upper comment makes sense around `is_failed_merge` there would be an impact over here

backend/tests/component/core/merge/test_failure_detection.py
```diff
@@ -0,0 +1,119 @@
...
+    def _recovery(self, db: InfrahubDatabase, cache: MemoryCache, component: InfrahubComponent) -> MergeFailureRecovery:
+        return MergeFailureRecovery(
+            db=db,
+            cache=cache,
+            component=component,
+            merge_write_blocker=MergeWriteBlocker(cache=cache),
+        )
+
+    @pytest.fixture
+    async def merging_branch(self, db: InfrahubDatabase, default_branch: Branch) -> Branch:
```
polmichel: It looks like this fixture should be named differently since this is a merging branch that specifically started 10 minutes+ ago. In the `test_within_grace_is_not_flagged`, there is also a merging branch created.

backend/tests/integration/merge/test_failed_merge_protection.py
```diff
@@ -0,0 +1,105 @@
...
+        merge_write_blocker = MergeWriteBlocker(cache=service.cache)
+        await merge_write_blocker.set(branch=source_name, state=MergeProtectionState.MERGE_FAILED)
+
+        yield source_name
+
+        await merge_write_blocker.delete()
+        backend_branch.status = BranchStatus.OPEN
+        await backend_branch.save(db=db)
+
+    async def test_failed_merge_blocks_with_recovery_code(
```
polmichel, line 72: Nice test, it clearly illustrates the different steps of the workflow

backend/tests/integration_docker/test_merge_kill_recovery.py
polmichel, line 1: Just thinking out loud here, but shouldn't we keep such long tests from running in every pipeline? Maybe we should have a test suite that we run once per release, or something like that?

Based on my tests, this test takes approximately 90–100s, which impacts the `integration_docker` test suite. That's also the longest test suite in the current CI since the component tests were split out.

This gives a lof of confidence, but also costs a lot in my opinion.

backend/infrahub/exceptions.py
ajtmccarty: yes I agree. this docstring has been updated. looks like you were reviewing older code

backend/infrahub/core/query/branch.py
ajtmccarty: I kept it for backwards compatibility. I didn't think it was worth it to search for and update uses of `status` vs `statuses`

backend/tests/component/core/merge/test_failure_detection.py
ajtmccarty: updated to `merging_branch_for_10m`

backend/tests/integration_docker/test_merge_kill_recovery.py
ajtmccarty: yes this is a good point. I'll mark as skip for now and make a follow-up item to enable it for release testing only
https://opsmill.atlassian.net/browse/IFC-2878

backend/infrahub/core/merge/failure_recovery.py
ajtmccarty: renamed to `should_mark_as_failed_merge`

backend/infrahub/exceptions.py
polmichel: Ah sorry, I missed that

polmichel (approved):
LGTM !

PR #9787, opened 2026-07-02
feat(git): compute and store definition fingerprints during import

ogenstad:
# Why

**The problem.** To decide whether an imported definition (a query, transform, artifact definition, or generator) needs regenerating, we need to know whether the code/inputs that affect its *output* actually changed. Today the only signals are blunt:

- The **git commit id** changes on every commit - including edits to unrelated files, comments, or other definitions - so keying off it over-triggers.
- Determining the truth otherwise means re-reading and diffing the actual repository file contents, which the consumer shouldn't have to do.

**The goal.** Give each definition a deterministic **fingerprint**: a content hash of exactly its output-affecting inputs. Because it's a normal branch-aware attribute, a consumer can tell whether a definition changed by a plain data diff of the fingerprint - without inspecting the git commit and without re-reading file bytes. When an input genuinely changes, the fingerprint changes; when nothing output-affecting changes (an unrelated commit, a comment-only edit), it stays stable.

**The invariant.** Inherited from INFP-409: over-regeneration is acceptable, under-regeneration is not. Every fallback (unresolved dependency, absent `watch`, incomplete closure) folds the commit id back in and errs toward regenerating.

**Scope / non-goals.** This is the foundation only: it computes and stores the fingerprint but wires up no consumer, rewires no trigger, and changes no regeneration gate. The schema `fingerprint` attribute landed separately on the base branch; this PR adds the computation and storage.

## What changed

**New fingerprint package.** `backend/infrahub/git/fingerprint/` holds small, constructor-injected components:

- `FingerprintHasher` - deterministic SHA-256 over length-prefixed canonical terms, plus canonical JSON for structured inputs.
- `GitBlobResolver` (behind a `BlobResolver` protocol) - resolves `(path, blob_sha)` pairs from the git tree at the imported commit, reading git metadata only, never file contents.
- `FingerprintRegistry` - a per-import snapshot so a dependent definition composes the freshly-computed value of its inputs from the same import, never a stale stored one.
- `FingerprintComposer` - layered composers: query, transformation (Python + Jinja2), artifact definition, and generator definition.

**Import wiring.** `git/integrator.py` builds one composer per import and threads it through the phases in dependency order (queries, then transforms/generators, then artifact definitions), folding the fingerprint into the existing SDK create/update payloads. Writes are conditional on a value change, so a no-op re-import produces an empty diff and no extra saves.

**Composition rules (what is and isn't hashed).**

- The manifest (`.infrahub.yml`) is excluded from the hashed closure but kept in stored dependencies; its output-affecting fields are hashed as parsed scalars instead of raw bytes, so a comment-only manifest edit doesn't churn every fingerprint.
- The commit id is folded in only when inputs can't be pinned down - `watch` absent, closure incomplete, or an upstream fingerprint unresolved - always erring toward over-regeneration.
- Group **identity** (target group id) is hashed; group **membership** is not.
- `timeout`, `description`, and `execute_in_proposed_change` / `execute_after_merge` are excluded as non-output-affecting.

**Tests.** Unit tests for the hasher, blob resolver, registry, closure selector, and watch / incomplete-closure / unresolved-upstream folding, plus one composer test per layer. Integration tests over the car-dealership fixture cover import-and-store, re-import stability, change detection, the consistent-snapshot guarantee, and branch-aware behaviour.

---

**What stayed the same.** No trigger, regeneration gate, or consumer changed - confirmed by the existing import and artifact-regeneration regression suites staying green with identical selection behaviour. No new dependency (stdlib `hashlib`), and the importer keeps writing through the SDK with no direct-DB access.

---
## Summary by cubic
Compute and persist deterministic fingerprints for queries, transformations, generators, and artifact definitions during repository import to enable reliable change detection and branch-aware diffs. Implements IFC-2858.

- **New Features**
  - Added `FingerprintComposer` using git tree blob SHAs; no file content reads.
  - Queries: fingerprint from stored query text; persisted as a branch-aware attribute.
  - Transformations (Python/Jinja2): fingerprint from source + closure blob SHAs (manifest excluded) + watch config; fold commit id when `watch` is absent, closure is incomplete, or an upstream fingerprint is unresolved.
  - Generators and artifact definitions: fingerprint from their config plus dependent fingerprints.
  - Import integrator computes and stores fingerprints; values participate in branch diffs.
  - Helpers: `GitBlobResolver`, `FingerprintHasher` (canonical JSON + length‑prefixed SHA‑256), per-import `FingerprintRegistry`.
  - Tests: unit and integration coverage for blob resolution, closure selection, determinism, dependency chaining (query → transform → generator/artifact), branch inheritance, watch-state handling, snapshots, and re-imports.

- **Bug Fixes**
  - Fixed a test to pass under Python 3.14.

<sup>Written for commit fdeb10ca1b74912fa71b46139472af5e567288e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/fingerprint/blob_resolver.py
+++ b/backend/infrahub/git/fingerprint/blob_resolver.py
@@ -1,0 +1,44 @@
+"""Resolve repo-relative paths to their git blob SHAs from the tree at a commit."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from git import Repo
+
+
+class BlobResolver(Protocol):
+    """Resolve repo-relative paths to sorted `(path, blob_sha)` pairs."""
+
+    def resolve(self, paths: Sequence[str]) -> list[tuple[str, str]]: ...
+
+
+class GitBlobResolver:
+    """Resolve `(repo_relative_path, git_blob_sha)` pairs from the git tree at a commit.
+
+    Only git metadata (the tree entries) is read; file contents are never loaded.
+    """
+
+    def __init__(self, *, repo: Repo, commit: str) -> None:
+        self._repo = repo
+        self._commit = commit
+
+    def resolve(self, paths: Sequence[str]) -> list[tuple[str, str]]:
+        """Return the sorted `(path, blob_sha)` pairs for the given paths.
+
+        A path with no corresponding tree entry resolves to an empty SHA so the path
+        still contributes to the ordering and reappears distinctly once it is tracked.
+        """
+        tree = self._repo.commit(self._commit).tree
+        resolved: list[tuple[str, str]] = []
+        for path in paths:
+            try:
+                entry = tree / path
+            except KeyError:
+                resolved.append((path, ""))
+                continue
+            resolved.append((path, entry.hexsha))
+        return sorted(resolved)
--- a/backend/infrahub/git/fingerprint/composer.py
+++ b/backend/infrahub/git/fingerprint/composer.py
@@ -1,0 +1,215 @@
+"""Layered composers that turn a definition's output-affecting inputs into a fingerprint."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, assert_never
+
+from infrahub.git.closure_builder.post_processing import MANIFEST_PATH
+from infrahub.git.fingerprint.blob_resolver import GitBlobResolver
+from infrahub.git.fingerprint.hasher import FingerprintHasher, canonical_json
+from infrahub.git.fingerprint.registry import FingerprintKind, FingerprintRegistry
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from git import Repo
+    from infrahub_sdk.schema.repository import InfrahubWatchConfig
+
+    from infrahub.git.fingerprint.blob_resolver import BlobResolver
+
+
+def fold_commit_id(*, commit: str, watch: InfrahubWatchConfig | None, closure_complete: bool) -> str | None:
+    """Return the commit id to fold into a fingerprint, or None to omit it.
+
+    The commit id is folded (making the fingerprint change on every commit, the safe
+    over-regenerating default) whenever the definition's inputs cannot be pinned down:
+
+    - `watch` is absent (`None`): the definition has not declared its dependencies.
+    - the dependency closure is incomplete: some output-affecting dependency could not be
+      resolved, so hashing only the resolved paths would leave the fingerprint stable over
+      an input the system already knows is unknown - an under-regeneration risk.
+
+    Only a definition with a present `watch` and a complete closure omits the commit id and
+    gets a stable, precise fingerprint.
+    """
+    if watch is None or not closure_complete:
+        return commit
+    return None
+
+
+class ClosurePathSelector:
+    """Select the dependency paths that contribute to a fingerprint's hashed closure.
+
+    The manifest is excluded even though it is merged into every definition's stored
+    dependencies: its blob changes on any unrelated or comment-only edit, which would churn
+    every definition's fingerprint. Its output-affecting fields are folded in separately as
+    parsed scalars, so dropping its bytes keeps the fingerprint stable against manifest
+    noise while remaining sensitive to the fields that actually matter.
+    """
+
+    def __init__(self, *, excluded_paths: frozenset[str]) -> None:
+        self._excluded_paths = excluded_paths
+
+    def select(self, dependencies: Iterable[str]) -> list[str]:
+        return [path for path in dependencies if path not in self._excluded_paths]
+
+
+@dataclass(frozen=True, kw_only=True)
+class QueryFingerprintInput:
+    name: str
+    query_text: str
+
+
+@dataclass(frozen=True, kw_only=True)
+class PythonTransformationFingerprintInput:
+    name: str
+    query_name: str
+    dependencies: tuple[str, ...]
+    dependencies_complete: bool
+    watch: InfrahubWatchConfig | None
+    class_name: str
+    convert_query_response: bool
+
+
+@dataclass(frozen=True, kw_only=True)
+class Jinja2TransformationFingerprintInput:
+    name: str
+    query_name: str
+    dependencies: tuple[str, ...]
+    dependencies_complete: bool
+    watch: InfrahubWatchConfig | None
+    template_path: str
+
+
+@dataclass(frozen=True, kw_only=True)
+class ArtifactDefinitionFingerprintInput:
+    name: str
+    transformation_name: str
+    parameters: dict[str, Any]
+    content_type: str
+    artifact_name: str | None
+    target_group_id: str | None
+
+
+@dataclass(frozen=True, kw_only=True)
+class GeneratorDefinitionFingerprintInput:
+    name: str
+    query_name: str
+    dependencies: tuple[str, ...]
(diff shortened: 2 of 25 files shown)


backend/infrahub/git/fingerprint/blob_resolver.py
```diff
@@ -0,0 +1,44 @@
...
+        """Return the sorted `(path, blob_sha)` pairs for the given paths.
+
+        A path with no corresponding tree entry resolves to an empty SHA so the path
+        still contributes to the ordering and reappears distinctly once it is tracked.
+        """
+        tree = self._repo.commit(self._commit).tree
+        resolved: list[tuple[str, str]] = []
+        for path in paths:
+            try:
+                entry = tree / path
```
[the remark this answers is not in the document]
ogenstad: Not changing this one, deliberately.

`dependencies` only ever holds file paths: the Python/Jinja2 closure builders emit files, and `watch` directory entries are expanded to their tracked files via `git ls-files` before they reach the resolver. So a non-blob entry isn't a real input here.

Even in the hypothetical directory case, `tree / path` returning a Tree's `.hexsha` is still a valid content identifier - it changes exactly when the directory's contents change, i.e. the safe over-detecting direction. The governing invariant is "over-regeneration is acceptable, under-regeneration is not."

The suggested fallback (empty SHA for non-blob) would actually move it the wrong way: an empty SHA is a stable value, so a genuinely-unresolved path would silently stop influencing the fingerprint - an under-regeneration risk. That's the one outcome this feature must avoid, so I've left the current behavior.

backend/infrahub/git/integrator.py
```diff
@@ -376,7 +409,11 @@ async def import_jinja2_transforms(
         local_transforms = await self._build_jinja2_transform_definitions(
             branch_name=branch_name, commit=commit, config_file=config_file
         )
-        await self._apply_jinja2_transform_definitions(branch_name=branch_name, local_transforms=local_transforms)
+        await self._apply_jinja2_transform_definitions(
+            branch_name=branch_name,
+            local_transforms=local_transforms,
+            fingerprint_composer=self._build_fingerprint_composer(commit=commit),
```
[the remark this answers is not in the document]
ogenstad: Addressed, though not by sharing/seeding the registry.

Two things make this safe now:

1. In production this path can't happen: repository import always runs through `apply_import_plan`, which builds one `FingerprintComposer` and threads it through every phase in dependency order (queries → transforms/generators → artifact definitions), so the registry is fully populated. The per-phase `import_*` tasks flagged here have no production caller - they're only invoked directly from tests.

2. For those direct-call paths, the composer no longer goes silently stale on a missing upstream. A referenced fingerprint that's absent from the registry is now treated as an unresolved input, so the commit id is folded in and the fingerprint over-regenerates every commit rather than pinning to a stale value. That's the safe direction under the invariant (over-regeneration acceptable, under-regeneration not), and it holds regardless of which entry point is used.

So the silent-miss failure mode is gone. I opted for the conservative fold over seeding each task's registry because seeding would add real complexity to test-only helpers, and folding is correct for any caller. If we later wire a consumer that needs precise fingerprints from these standalone tasks, seeding from persisted fingerprints is the right follow-up.

backend/tests/unit/git/fingerprint/test_blob_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+
+from git import Repo
+
+from infrahub.git.fingerprint.blob_resolver import GitBlobResolver
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+def _commit_files(repo_dir: Path, files: dict[str, str]) -> str:
```
polmichel, line 13: Clean test suite!
An optional comment: adding a test that ensures that a file is correctly tied to its own sha could be a plus.
I've a proposition for this, but feel free to ignore it and this comment since this might be nitpicking

```
  # git blob SHAs for the file contents below — `git hash-object` of the raw bytes.
  _SHA_ALPHA = "7e74e68b2a782a3aead46d987a63ca1c91091c13"
  _SHA_BETA = "e1d65540f4fdf72431ec47e001282cc7e8ed7c0c"
  _SHA_GAMMA = "f6ce3c605d8e619e80eb03eb65fc984a940abde7"


  def test_resolve_links_each_path_to_its_own_blob_sha(tmp_path: Path) -> None:
      commit = _commit_files(tmp_path, {"a.py": "alpha", "b.py": "beta", "sub/c.py": "gamma"})
      resolver = GitBlobResolver(repo=Repo(tmp_path), commit=commit)

      pairs = dict(resolver.resolve(["sub/c.py", "a.py", "b.py"]))

      assert pairs == {
          "a.py": _SHA_ALPHA,
          "b.py": _SHA_BETA,
          "sub/c.py": _SHA_GAMMA,
      }
```

backend/infrahub/git/fingerprint/blob_resolver.py
```diff
@@ -0,0 +1,44 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from git import Repo
+
+
+class BlobResolver(Protocol):
+    """Resolve repo-relative paths to sorted `(path, blob_sha)` pairs."""
+
+    def resolve(self, paths: Sequence[str]) -> list[tuple[str, str]]: ...
```
polmichel, line 16: More a thought : having a `list[tuple[str, str]]` returned from a `resolve` method forces the reader to get inside the method and its documentation to know what is going on inside and what's returned. Maybe considering a simple object that gives a sense to this data collection could help.

That said, I think it's fine since this is a quite condense and internal package.

backend/infrahub/git/fingerprint/composer.py
```diff
@@ -0,0 +1,231 @@
...
+    query_text: str
+
+
+@dataclass(frozen=True, kw_only=True)
+class PythonTransformationFingerprintInput:
+    name: str
+    query_name: str
+    dependencies: tuple[str, ...]
+    dependencies_complete: bool
+    watch: InfrahubWatchConfig | None
```
polmichel, line 70: These 4 attributes seem to be linked: they are also used inside `Jinja2TransformationFingerprintInput` and `GeneratorDefinitionFingerprintInput` would a common object shell makes sense?

backend/tests/unit/git/fingerprint/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+from infrahub.git.closure_builder.post_processing import MANIFEST_PATH
+from infrahub.git.fingerprint.composer import ClosurePathSelector, FingerprintComposer
+from infrahub.git.fingerprint.hasher import FingerprintHasher
+from infrahub.git.fingerprint.registry import FingerprintRegistry
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+
+class StaticBlobResolver:
```
polmichel, line 14: nice test adapter

polmichel (approved):
Very clean ! The test suite seems to have a very high coverage and the responsibilities are split between meaningful components.
Comments and thoughts are just optional.

ogenstad:
> Very clean ! The test suite seems to have a very high coverage and the responsibilities are split between meaningful components. Comments and thoughts are just optional.

Thanks Pol! I've added a commit with some additional tests like you suggested, other than that I'll merge it as it is now.

PR #9791, opened 2026-07-02
fix(git): delete stale artifacts when their target leaves the definition's target group

### LINKED ISSUE #9790: bug: artifact of a removed target is left stale on main after a proposed change merge
### Component

### Infrahub version

1.10.1 (reproduced on latest `stable`, commit `144e7d1dd`)

### Current Behavior

When a device is removed from the target group of an artifact definition and that removal is merged into `main` via a proposed change, the device's artifact is left on `main` untouched. The post-merge artifact pipeline (`request_artifact_definitions_generate`) runs, regenerates artifacts for the group's *current* members, and silently skips artifacts whose owner is no longer a member — nothing ever deletes them. The artifact remains `Ready` on `main` with its last-rendered content, even though its target is no longer selected by the definition.

### Expected Behavior

After the removal merges, the artifact of the removed target should be cleaned up from `main` (or at minimum flagged as no longer managed) — it should not persist indefinitely as a stale copy that looks like a live, managed artifact.

### Steps to Reproduce

Using the demo dataset + `demo-edge` repository (`Startup Config for Edge devices` targets the `edge_router` group):

1. On a branch, add a device that is not an artifact target (e.g. `atl1-core1`, a core router) to the `edge_router` group; open a proposed change and merge it. The post-merge pipeline generates a `startup-config` artifact for the device on `main` (status `Ready`).
2. On a second branch, remove the device from `edge_router`; open a proposed change and merge it. The membership removal lands on `main`.
3. Query `CoreArtifact` on `main` for the device: the `startup-config` artifact is still there, `Ready`, indefinitely — task-worker logs confirm `artifact-definition-generate` ran after the merge and skipped it.

An automated reproduction exists as an SDK-driven e2e test (`tests/e2e/proposed-changes/test_proposed_changes_artifact_target_removal.py` on branch `fac/reproduce-artifact-removal-issue-xsscp`): it fails only at the final assertion, waiting 300s for the artifact to disappear from `main`.

### Additional Information

Where the behavior lives: `generate_request_artifact_definition` (`backend/infrahub/git/tasks.py`) builds `artifacts_by_member` only for artifacts whose `object` is still in the target group's members and iterates current members to dispatch generation; artifacts of removed members are never touched. The branch-side artifact checks in the proposed-change pipeline behave the same way, and nothing in `post_process_branch_merge` compensates.

Related (regeneration triggers, not cleanup): #9641

fatih-acar:
## Summary

Removing a device from an artifact definition's target group (e.g. via a merged proposed change) left its generated artifact on `main` indefinitely — a stale, `Ready`-looking artifact for a target the definition no longer selects. Artifact generation now cleans these up, so the artifact set on a branch always reflects the definition's current targets.

Fixes #9790

## Key Changes

- A full artifact-generation pass over a definition (post-merge trigger, definition create/update, or the generate endpoint without a node filter) now deletes artifacts whose target is no longer a member of the target group.
- Limited runs (an explicit list of artifacts to regenerate, e.g. the per-artifact re-generate button) are unchanged: they only regenerate what was requested and never delete.
- Regression coverage reuses the existing `objects/test_artifact_definition.py` e2e spec instead of adding a new one: before generating, `atl1-core1` is added to the `edge_router` target group (the pass must produce its artifact); it is then removed and a second Generate pass must delete the artifact. The original proposed-change-merge reproduction was validated separately during development (commit cea3e246e, later folded into this spec) — both paths funnel into the same generation flow.

## Documentation Updates

- `docs/docs/artifacts/overview.mdx`: the "When artifacts regenerate" section now documents that a target leaving the group deletes its artifact on the next generation pass.

## Test Plan

- `INFRAHUB_TESTING_IMAGE_VER=local INFRAHUB_TESTING_DOCKER_PULL=false uv run pytest -c tests/e2e/pytest.ini tests/e2e/objects/test_artifact_definition.py` — verified locally: passes on an image built from this branch (3m07s), fails at the deletion deadline on an image built without the fix.
- CI: `E2E-testing-pytest-playwright` shard `shard_branches_repo` runs the extended spec.


backend/infrahub/git/tasks.py
```diff
@@ -563,9 +563,24 @@ async def generate_request_artifact_definition(
...
+
+    # A full pass over the definition also cleans up artifacts whose target is no
+    # longer a member of the target group; a limited run only regenerates the
+    # requested artifacts and must not delete anything else.
+    if not model.limit:
+        log = get_run_logger()
+        for artifact in stale_artifacts:
+            log.info(
+                f"Deleting artifact {artifact.id} ({artifact.name.value}): "
+                f"its target {artifact.object.id} is no longer a member of the definition's targets"
```
gmazoyer: It could probably be beneficial for debug purpose to also have the branch in this log.

backend/infrahub/git/tasks.py
```diff
@@ -563,9 +563,24 @@ async def generate_request_artifact_definition(
...
+            )
+            await artifact.delete()
```
gmazoyer: If one `artifact.delete()` raises here, the whole flow fails before any regeneration is dispatched. I guess the cleanup is somehow secondary so a stale artifact that fails to delete should not block regenerating current members. Adding a try/except to log a warning and continue could help. Moving the loop after the regeneration batch would also work.

PhillSimonds (approved):
LGTM

backend/infrahub/git/tasks.py
fatih-acar: Done in 4aedd3d3c — the cleanup log lines now include the branch.

backend/infrahub/git/tasks.py
fatih-acar: Done in 4aedd3d3c — did both: the cleanup loop now runs after the regeneration batch is dispatched, and each delete is wrapped in try/except that logs a warning (with traceback) and continues, so a failing delete can neither block regeneration nor skip the remaining stale artifacts.

PR #9793, opened 2026-07-02
fix(core): cap node mutation event related resources at the Prefect maximum

fatih-acar:
## Summary

Creating or updating a node with a large cardinality-many relationship (roughly 250+ peers — e.g. a trunk interface tagging a 4k VLAN range from an object file) silently produced **no event at all**: no entry in the node's activity log, no automation triggers fired, no webhook delivery. The node mutation event listed every relationship peer twice in its Prefect related resources with no upper bound, so the event exceeded `PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES` (500 in our image) and the Prefect API rejected it outright.

## Key Changes

- Node mutation events are always recorded now, regardless of how many relationship peers the mutation touches.
- Related resources are built in priority order — node-scoped entries first (attribute updates, parent, the node's own related-node entry), then relationship updates (which automation triggers match on), then per-peer related-node entries — and truncated at the Prefect maximum with a warning log. The full peer list remains available in the event payload's changelog.
- The Prefect maximum reader moved from the computed-attribute tasks to `infrahub/events/constants.py` so both call sites share one definition.
- New e2e reproducer (`tests/e2e/events/test_node_event_related_resources.py`, foundation shard, self-contained): a control interface tagging 1 VLAN and a target tagging 300 must both get their `infrahub.node.created` event recorded. Fails on a pre-fix image (target event never appears, 180s timeout), passes with this fix (1m19s).
- Unit tests pin the cap and the priority order (`backend/tests/unit/event/test_node_action.py`).

## Related Context

Diagnosed from a customer object-file load where trunk interfaces carry ~4,000 `tagged_vlans` each: the interface events would have carried ~8,200 related resources. The dumped events that *did* land (a VLAN with 47 cvlan peers → 107 related resources) also showed every peer duplicated across the `relationship_update` and `related.node` roles.

Not addressed here (possible follow-ups):

- Every peer still appears under two roles below the cap — the duplication is the wire format automations and event queries rely on today.
- One secondary `NodeUpdatedEvent` is still emitted per touched peer (4k peers → 4k events); each is small and individually valid.
- The `payload` (full changelog) is not capped; it stays under Prefect's 1.5MB size limit even at 4k peers.

## Documentation Updates

- `dev/knowledge/backend/events.md`: documented the related-resources cap and the priority order.
- Changelog fragment: `changelog/+node-event-related-resources-cap.fixed.md`.

## Test Plan

- `uv run pytest backend/tests/unit/event/ backend/tests/unit/computed_attribute/` (50 passed; full unit suite: 1314 passed)
- `INFRAHUB_TESTING_IMAGE_VER=<local-image> INFRAHUB_TESTING_DOCKER_PULL=false uv run pytest -c tests/e2e/pytest.ini tests/e2e/events/test_node_event_related_resources.py` — verified failing on a pre-fix image and passing on an image with this fix.

---
## Summary by cubic
Prevent dropped node mutation events by capping and prioritizing related resources so each event stays within Prefect’s limit and gets recorded.

- **Bug Fixes**
  - Build `related` in priority order (node-scoped, then relationship updates, then per-peer) and truncate at `PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES` with a warning; full peer list stays in the changelog payload.
  - Added Docker integration reproducer and unit tests to verify the cap and ordering.

- **Refactors**
  - Moved `get_prefect_max_related_resources()` to `infrahub/events/limits.py`, reading the env var on each call with a safe default; updated imports/docs and added a rule clarifying `constants.py` holds constants only.

<sup>Written for commit 35423d2bca5c96d0a4ad943156bbe929c3b6bcef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/events/constants.py
```diff
@@ -1,9 +1,18 @@
+import os
 from enum import StrEnum
 
 EVENT_NAMESPACE = "infrahub"
 ACCOUNT_EVENT_PREFIX = f"{EVENT_NAMESPACE}.account."
 
 
+def get_prefect_max_related_resources() -> int:
```
ogenstad: This doesn't look like a constant.. We should have it in some other file. Not sure what the name would be, alternatively of we can just set a static variable that does what this function does, unless you want to be able to alter the environment variable during runtime?

Even if this function is trivial in scope I think we should avoid having functions in a file called constants.py to avoid muddying the waters.

fatih-acar: Moved to `backend/infrahub/events/limits.py` in 93d146ddf, and `constants.py` is back to constants only. Kept it a function rather than an import-time static: the value is read from the environment at call time, which the computed-attribute chunking tests rely on (they parametrize the env var at runtime), and it keeps the limit tunable per process. Also codified the guideline in `.agents/rules/python-module-layout.md` so agents stop putting functions in `constants.py`.

backend/infrahub/events/constants.py
```diff
@@ -1,9 +1,18 @@
...
+    """Return the maximum number of related resources the Prefect API accepts per event."""
+    max_related_resources = int(os.environ.get("PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES", "500"))
```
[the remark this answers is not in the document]
fatih-acar: Hardened in 93d146ddf: a malformed value now falls back to the default (500) instead of raising.

ogenstad (approved):

Issue #9499, opened 2026-06-08
bug: Git merge skipped if Proposed Change merged before repo sync completes


lancamat1:
### Component

Git Integration

### Infrahub version

1.9.6

### Current Behavior

On an active `CoreRepository` with a Git remote, if a Proposed Change is merged
*before* the repository has finished syncing the source branch's latest commit
into Infrahub's internal store, the git-side merge into the default branch is
**silently skipped** — no merge commit, no push to the remote default branch,
and no error surfaced.

Files added on the branch (e.g. a transform + artifact definition) therefore
never reach `main`, and views that depend on them (artifact detail) error on
`main`. Repo sync can take several minutes, so this is easy to hit. Waiting
after the merge does not recover — the merge for that PC is already complete.

Mechanism (read against `infrahub-v1.9.6`):

- `merge_git_repository` (`backend/infrahub/git/tasks.py`) for an active
  `CoreRepository` runs the `else` branch: `await repo.merge(source_branch, dest_branch)`.
- `InfrahubRepository.merge` (`backend/infrahub/git/repository.py`) computes
  `commit = get_commit_value(source_branch, remote=False)` — Infrahub's
  *internal* copy of the branch. If the sync hasn't landed the branch's new
  commit internally yet, this is still the old base, so the merge produces no
  new commit: `commit_after == commit_before` → `return False` → `push()` is
  never called, and no error is raised.

### Expected Behavior

Merging a Proposed Change should reliably merge the corresponding git branch
into the repository's default branch (and push to the remote) regardless of
sync timing — or, if the branch's commit isn't yet available internally, block
on the sync or fail loudly. The PC merge should not be able to "succeed" while
silently dropping the git-side merge.

### Steps to Reproduce

1. Connect a Git-remote-backed `CoreRepository` to Infrahub (active, not staging).
2. Create a branch with `--sync-with-git` and commit a new file on it (e.g. a
   Jinja transform + an artifact definition) so the branch carries a commit the
   default branch doesn't have.
3. Immediately open a Proposed Change for that branch and merge it *before* the
   repository finishes syncing (sync can take several minutes).
4. Inspect the git remote's default branch (activity feed / `git log` on `main`).

Observed: no merge commit on the default branch; the branch's commit never lands
on `main`. An artifact/transform that lives only on the branch errors when viewed
on `main`. Waiting and refreshing does not fix it.

### Additional Information

- Reproduced on Infrahub 1.9.6 (screen recording to be attached).
[...]

PR #9798, opened 2026-07-03
test(git): multi-environment single-repo validation (Approach A)

lancamat1:
## Context: multi-environment Infrahub from a single Git repository

The common deployment pattern today is one Git repository per Infrahub environment — clean isolation, but schema/object definitions are duplicated and promotion is a cross-repo copy. The pattern under validation here uses **one** Git repository with **one long-lived branch per environment** (`develop` → dev, `main` → prod, …), where each environment is a separate Infrahub instance that maps *its own* branch onto its internal `main` and never sees the other environments' branches. Promotion between environments is git-native (PR / tag on the Git host).

Two topologies exist for the consumer environments:

- **Approach A (this PR):** the dev instance is a read-write `CoreRepository` pinned via `default_branch`; staging/prod are **read-only** `CoreReadOnlyRepository` consumers pinned via `ref` (branch, tag, or commit SHA). Consumers receive promotions through an explicit reimport or by bumping the `ref` to a new release tag. Isolation is structural — a read-only consumer can never push back.
- **Approach B (out of scope, follow-up):** all environments read-write, each filtered to its own branch via `INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES`, receiving promotions automatically via periodic sync.

The goal of this PR is to find and pin the gaps in Approach A so it can be recommended and documented: every mechanism the pattern relies on is either verified by a green regression guard or captured as a reproducible, strict-`xfail` defect test that flips loudly the day the defect is fixed.

## What

Three additive test suites, no product-code changes:

- **`backend/tests/integration/git/test_multi_env_writeback.py`** (19 tests) — the read-write side: branch mapping (`default_branch=develop` → internal `main`), the write-back of in-Infrahub merges to the remote default branch, conflict/divergence behaviour, and sync-filter semantics. Real Gogs remote + in-process app; runs in the normal integration job.
- **`backend/tests/integration/git/test_multi_env_promotion.py`** (12 tests) — the read-only consumer side: promotion via reimport and ref bumps (new tags, annotated tags, rollback to an older tag, commit SHAs, force-moved tags, force-pushed branches), what content a promotion carries (object replacement vs removal, schema removal), and staging a ref bump inside an Infrahub branch before merging. Local bare remotes + in-process app.
- **`backend/tests/integration_docker/test_multi_env_approach_a.py`** (6 tests) — the faithful two-instance demonstration: two full Infrahub stacks bind-mounting one shared remote; consumer isolation, promotion-invisible-before-reimport, reimport, and tag-bump promotion verified across live instances. Runs in the Docker integration job (requires an image built from the branch under test — a stale image gives false failures in the periodic-sync path).

**Canonical promotion workflow** is verified end-to-end and treated as the primary contract: nothing acts on the consumer's `main` directly — the ref bump is staged in an Infrahub branch (import loads there; `main` provably keeps the previous release while under review) and lands via a **proposed change** whose merge carries the recorded ref/commit and content. Also guarded: merging an *unrelated* proposed change opened before a promotion does not move the promoted pin.

## Confirmed defects (strict xfail — flip to failures when fixed)

- **#9568 — the swallowed write-back family. Operational headline: on a multi-worker deployment, in-Infrahub merges are silently lost non-deterministically** — only the worker that performed the initial import holds a local copy of a non-`main` default branch, so whichever worker picks up the merge decides whether the write-back lands (the original report observed ~1 in 4 lost on a 2-worker pool). **Until fixed, write-back environments must run a single task worker and verify the remote tip after merges.** `push()` swallows every push failure and `merge()` records the commit before the push is confirmed. Three confirmed triggers (missing local default branch; non-fast-forward after an out-of-band advance; **push rejected by branch protection**) and confirmed permanence: the dropped write-back is never re-delivered, the repo is left stranded diverged, and no `GIT_CONFIG_GLOBAL`/`pull.rebase` lever rescues it. Blocks the read-write side of the pattern; read-only consumers are unaffected.
- **Per-flow repository factory never learns `default_branch`** — the same under-population class as #9600/#9601, at the `get_initialized_repo` layer used by artifacts, transforms, proposed-change checks, generators and computed attributes (likely the root of #8749). New issue to be filed.
- **Force-moved tag is silently stale on reimport** (fetch refuses to clobber tags; branches are force-updated — asymmetry pinned by tests). New issue to be filed.
- **Object-file removal orphans its objects** (no import run → no group-tracked GC; replacement within a listed file reconciles fine). New issue to be filed.
- **`default_branch` set after registration strands a duplicate standalone branch** — pins the "always set it at creation" rule.

## Verified working (green guards)

Branch mapping with no phantom branch (#9601), import-not-frozen, merge-conflict surfacing with clean worktree, divergent-pull recovery, filter semantics (incl. the documented unset-filter trap), the happy-path write-back on the importing clone, tag/SHA ref-bump promotion incl. rollback (pin + file-defined content only — schema and applied migrations never revert, and restored objects return as new nodes) and annotated tags, staged-in-branch ref bump promoting on merge, object replacement GC, and two-instance isolation full-stack.

## Related issues

**Blocking for the pattern (reproduced as strict-xfail here):**
- #9568 — silent write-back loss + permanent stranding (3 confirmed triggers incl. branch protection; scope expanded in issue comments)
- #9825 — per-flow repository factory never learns `default_branch` (artifacts, transforms, PC checks, generators; likely the root of #8749)

**Non-blocking, filed from findings here:**
- #9826 — force-moved tag silently stale on reimport (workaround: always cut a new tag)
- #9827 — commit recording re-triggers an echo re-import that can auto-sync new commits into read-only repos
- #9828 — feature request: reconcile schema removals imported from git

**Context:**
- #9601 / #9600 — the non-`main`-default import fix this suite guards / the bug it fixed
- #8749 — artifacts on non-`main` default branch (adjacent; likely same root as #9825)
- #9499 — merge-before-sync no-op write-back loss (related, distinct mechanism; kept separate)

## Status

Draft until the Docker integration job confirms the two-stack suite on a current image. Issue write-ups for the new findings are drafted and will be filed separately.


backend/tests/integration_docker/test_multi_env_approach_a.py
```diff
@@ -0,0 +1,500 @@
...
+        shared_remote: Path,
+        repo_name: str,
+    ) -> None:
+        """Bumping a read-only consumer's ref to a new tag promotes it — the production mechanism.
+
+        Instead of the documented reimport, a real release flow pins the consumer to a git tag and
+        bumps the ref to the next tag. Bumping the ref must advance the recorded commit to the tag's
+        commit. (Runs last: it mutates the shared consumer repo's ref.)
+        """
+        release_sha = _stage_release_with_object(shared_remote, tag_name="release-1", object_tag="release-1-tag")
```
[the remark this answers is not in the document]
lancamat1: Fixed: the dead helper is removed (the staging helper superseded it).

backend/tests/integration/git/test_multi_env_promotion.py
```diff
@@ -0,0 +1,563 @@
...
+        v1_sha = remote.branch_tip()
+        remote.tag("v1", v1_sha)
+
+        repo_name = "promo-stale-branch-repo"
+        node_id = await _register_readonly(client, repo_name, str(remote.bare_path), "v1")
+        await _import_ref(node_id, repo_name, "v1")
+        assert await _recorded_commit(db, node_id) == v1_sha
+
+        # The unrelated branch snapshots the repository at v1.
+        stale_branch = await client.branch.create(branch_name="unrelated-work")
```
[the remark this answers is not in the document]
lancamat1: Adopted, with a correction that turned out to matter: the suggested pre-merge assertion is now in place, but the branch reads its creation-time snapshot (v1), not the promoted value — and chasing that difference revealed the guard was previously passing vacuously (the dispatched repository merge lands on a queue nothing consumes in this harness). The test now also executes the dispatched flow explicitly after the merge, standing in for the task worker, and pins the full merge-then-dispatch sequence. Good catch — the extra assertion exposed a real hole in the guard.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+    async def test_nonff_writeback_lost_permanently_after_resync(
+        self,
+        db: InfrahubDatabase,
+        client: InfrahubClient,
+        gogs_server: GogsServer,
+    ) -> None:
+        """After a non-ff write-back drop, a subsequent sync does not re-deliver the write-back.
+
+        Probes whether the drop is transient (self-heals on the next sync) or permanent.
+        """
```
polmichel, line 562: A test docstring should describe the expected behavior of the system.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+            db=db, id=ds["node_id"], kind=InfrahubKind.REPOSITORY, raise_on_error=True
+        )
+        repo = await InfrahubRepository.init(
+            id=repository.id, name=ds["repo_name"], client=client, default_branch_name=DEV_BRANCH
+        )
+
+        _push_commit_to_remote(gogs_server.container, ds["repo_name"], "out_of_band.txt", branch=DEV_BRANCH)
+        writeback_commit = await _writeback_merge(repo, "feature-perm", "wb.txt")
+
+        await sync_remote_repositories()
```
polmichel, line 574: 1. I don't think the "sync a repository" operation should push anything to the remote, its purpose is to pull changes from the git remote.
  2. That said, the failed push should not be silently swallowed, it should surface as a clear failure.

From that perspective, this test is incorrect: even once we fix the silently-swallowed push, the test still won't do what it's meant to, because the fix will make an exception raise at the `_writeback_merge`
step, so before the assertion is ever reached. 
And the final assertion will never pass (and, in my opinion, shouldn't, since a sync isn't supposed to re-deliver the write-back). So under `xfail(strict=True)`, the test will never turn red: a raised exception is always treated as the expected failure, so it stays green whether the bug is present or fixed.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+        client: InfrahubClient,
+        gogs_server: GogsServer,
+    ) -> None:
+        """The write-back drop is a push-layer failure — it happens with explicit merge commits too."""
+        ds = await _register_and_import(client, gogs_server, "multi-env-mergecommit-repo")
+        repository: CoreRepository = await NodeManager.get_one(
+            db=db, id=ds["node_id"], kind=InfrahubKind.REPOSITORY, raise_on_error=True
+        )
+        repo = await InfrahubRepository.init(
+            id=repository.id, name=ds["repo_name"], client=client, default_branch_name=DEV_BRANCH
```
polmichel, line 671: Same comments than those related to `test_nonff_writeback_lost_permanently_after_resync`.
Additionally: 

1. The explicit merge commit parameter should not change anything from merge capability perspective since this is just acting on the commit history, so I am not sure that this test is useful.
2. This test is kind of linked to `test_nonff_writeback_lost_permanently_after_resync` if I get it right since the difference is about the merge commit parameter. There is no clear way of seeing they are related.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+
+        # The remote turns protected AFTER the import — the realistic sequence.
+        _install_reject_all_pushes_hook(bare_path)
+        develop_before = str(Repo(bare_path).commit(DEV_BRANCH))
+
+        await _writeback_merge(repo, "feature-protected", "wb.txt")
+
+        develop_after = str(Repo(bare_path).commit(DEV_BRANCH))
+        # Intended contract: the write-back reached the remote default branch (or the merge failed).
+        assert develop_after != develop_before
```
polmichel, line 814: Same comments than those related to `test_nonff_writeback_lost_permanently_after_resync`.
Additionally:
1. Since there is a specific hook here, it's impossible that a fix makes the assertion true.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+            "receives the write-back"
+        ),
+    )
+    async def test_writeback_rejected_by_remote_not_silently_dropped(
+        self,
+        db: InfrahubDatabase,
+        client: InfrahubClient,
+        tmp_path: Path,
+    ) -> None:
+        """A write-back rejected by the remote must land or fail loudly — not vanish.
```
polmichel, line 787: `Not land` is wrong since there is a hook here.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+        assert "staging" in await client.branch.all()
+
+        repo_node = await client.get(kind=InfrahubKind.REPOSITORY, id=node.id)
+        repo_node.default_branch.value = "staging"
+        await repo_node.save()
+        await sync_remote_repositories()
+        await sync_remote_repositories()
+
+        # Intended contract: the standalone duplicate of the (new) default branch is reconciled away.
+        assert "staging" not in await client.branch.all()
```
polmichel, line 916: Regarding this test, I am not sure that we would want the system to silently delete this branch. It sounds dangerous. If this is true, the assertion is not correct.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+        repo = await InfrahubRepository.init(
+            id=repository.id,
+            name=repo_name,
+            client=client,
+            default_branch_name=DEV_BRANCH,
+        )
+
+        # Non-importer state: a local primary branch worktree with a commit to write back, and NO
+        # local develop branch (only origin/develop from the fetch).
+        assert DEV_BRANCH not in repo.get_branches_from_local(include_worktree=False)
```
polmichel, line 227: This assertion is the one failing (and not the final assertion).
But the repository is created with a `default_branch_name` initialized to `DEV_BRANCH`. The repo initialization is currently creating the branch through `create_locally` method, so I am not sure if this assertion is really expected.

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+        (Path(str(repo_a.working_dir)) / "conflict.txt").write_text("content from a\n")
+        repo_a.index.add(["conflict.txt"])
+        repo_a.index.commit("conflict-a change")
+
+        repo_b = repo.get_git_repo_worktree(identifier="conflict-b")
+        (Path(str(repo_b.working_dir)) / "conflict.txt").write_text("content from b\n")
+        repo_b.index.add(["conflict.txt"])
+        repo_b.index.commit("conflict-b change")
+
+        with pytest.raises(RepositoryError):
```
polmichel, line 343: We should assert the exact error message raised inside the exception. At the same time, it describes precisely what we should expect from the system and it prevents this test to catch unwanted exceptions

backend/tests/integration/git/test_multi_env_writeback.py
```diff
@@ -0,0 +1,950 @@
...
+            raise_on_error=True,
+        )
+
+        # Fresh clone with default_branch=develop, mimicking a worker's local clone.
+        repo = await InfrahubRepository.init(
+            id=repository.id,
+            name=repo_name,
+            client=client,
+            default_branch_name=DEV_BRANCH,
+        )
```
polmichel, line 223: This repo instantiation is not the real code path that occurs when a merge occurs. I mean this may be the real code executed behind, but the test is then very fragile. One change to the real code path, and this test does not make any sense anymore.

From what I capture this test should exerce the real entry point of the application which should be `infrahub.git.tasks.merge_git_repository`. Doing this, we'll go through the real exercised repo
initialization, the real merge path and we will make sure to fix the real "user-facing" path.

This is a comment that may be true for all these tests. A double-check about what path is really executed when the application is managing the related workflow, and then apply this code path here would be great.

polmichel (commented):
I've only done the review of `test_multi_env_writeback`file for now.

1. strict xfail is only valid if the fix flips it loudly.
2. Additionally, I was wondering about the total execution time required for this test suite and how it'd impact the total CI execution time. Before we can answer this question, I think all of these tests should be assigned to the category they belong to. If a test does not need a real integration test setup, it should not be in this file. I asked Claude for a categorization of these tests, and I'll post the result at the end of this comment. I haven't reviewed it in detail against each test, but I think this should be considered and in each of those files before reviewing the entire PR right away.

## Component vs Integration — multi-env write-back suite (PR #9798)

  **Dividing line: what the assertion READS, not what the code touches (they all touch git).**
  - Reads **graph / branch-set after `sync_remote_repositories()`** → emergent from flow × db × git → **Integration**
  - Reads **remote or worktree git state after `repo.merge()/push()`** → pure git-repository layer → **Component** (dummy client + local bare remote, no testcontainers)
  - Reads a **pure decision** (predicate, field population) → **Unit**

  ### Integration (must stay) — assertion reads the graph/flow outcome

  | Test | Why |
  |---|---|
  | `test_nonmain_default_maps_to_primary_no_phantom` | branch set after sync |
  | `test_nonmain_default_import_not_frozen` | recorded commit after sync |
  | `test_divergent_default_branch_recovers_on_resync` | recorded commit after 2 syncs (flow recovery) |
  | `test_filter_excludes_branch` | branch set after sync |
  | `test_fetch_tolerates_problematic_excluded_ref` | recorded commit after sync |
  | `test_nonff_writeback_lost_permanently_after_resync` | needs the sync to probe re-delivery |
  | `test_writeback_drop_then_sync_converges_to_remote` | recorded commit after sync |
  | `test_pull_rebase_config_does_not_rescue_writeback_drop` | recorded commit after sync |
  | `test_unset_filter_imports_every_branch` | branch set after sync |
  | `test_default_branch_update_reconciles_phantom` | branch set after sync |
  | `test_conflicting_remote_branch_imports_without_blocking_others` | branch set + commit after sync |

  ### Component-downgradable — assertion reads git state after merge/push

  | Test | Why |
  |---|---|
  | `test_merge_conflict_surfaced_and_worktree_clean` | `RepositoryError` + `git status` only — no remote/graph read, zero barrier: start here |
  | `test_writeback_succeeds_on_importer_clone` | remote SHA after merge (happy path) |
  | `test_writeback_dropped_when_default_branch_absent_locally` | remote SHA after merge (missing-local-branch) |
  | `test_nonff_writeback_not_silently_dropped` | remote SHA after merge (out-of-band advance) |
  | `test_repeated_nonff_writebacks_each_dropped` | remote SHA after 2 merges, no sync |
  | `test_writeback_drop_independent_of_explicit_merge_commit` | remote SHA after merge (config flag) |
  | `test_writeback_rejected_by_remote_not_silently_dropped` | remote SHA after merge — already on a local bare remote, readiest port |

  ### Unit-ish (closest to a pure decision)

  | Test | Why |
  |---|---|
  | `test_initialized_repo_resolves_configured_default_branch` | factory field population (`default_branch`) — heaviest harness for the lightest assert |

Feasibility of the component ports: already low-friction — the existing dummy_async_request client swallows the one graph write `repo.merge()` makes (update_commit_value), and the repo has helpers for local file/bare remotes (_build_repository_with_conflict, _seed_local_bare_remote). No new infrastructure. Start with the conflict test (reads only git status, so the graph write isn't even on the path) as the sub-second proof-of-pattern; the other 6 follow the same shape, snapshotting the local remote's tip instead.

PR #9805, opened 2026-07-03
feat(backend): (INFP-589) Collect more metrics and introduce 24h windowing in telemetry collection

saltas888:
# Why

Phase 1 telemetry only reports a point-in-time snapshot (counts, versions) and a 7-day rolling event tally bounded by Prefect's retention. There's no adoption signal (accounts, active branches) and no way to see what happened over a fixed, comparable window.

Adds adoption metrics and a 24h activity window, computed the same way the product counts nodes so the numbers are correct across branches. Does not change the collection schedule, opt-out behavior, or storage.

## What changed

- `activity_24h`: logins (+unique), checks started/passed/failed, artifacts   created/updated, branches created/merged/deleted, and webhook success/failure, over the previous full UTC calendar day — a deterministic, half-open `[start, end)` window that tiles exactly with no gaps or overlap.
- New counts: `accounts.active`, `accounts.groups`, `branches.active`.
- `database.node_count.corenode` (all managed nodes) and `database.node_count.user` (user-defined-namespace nodes only); the existing raw `total` is unchanged.
- Per-metric graceful degradation: one failing source reports `null` for just that field; a source that succeeds with nothing to count reports `0`. The
  per-source seams in `gather_anonymous_telemetry_data` are plain function
  parameters defaulted to the real gatherers — no mocking, per the project's
  no-mock testing rule.
- `payload_format` bumped to `20260628`. All changes are additive.

Implementation notes, known gaps, and the temporal model (point-in-time vs.
rolling vs. windowed) are documented in
[`dev/knowledge/backend/telemetry.md`](../dev/knowledge/backend/telemetry.md),
including field-level semantics for `activity_24h` (e.g. `logins` counts
interactive sign-ins only — password/OIDC/OAuth2 — not per-request API-key
authentication). Notably, `checks_started` can currently exceed
`checks_passed + checks_failed` for validators with no real checks to run —
documented and covered by a unit test, not fixed here (pre-existing
event-emission gap, out of scope).

## Example payload

Here is a representative daily payload with this PR's additions. New fields are marked `// NEW`; everything else is unchanged. `null` means the source failed, `0` means measured-and-empty.

```jsonc
{
  "kind": "community",
  "payload_format": "20260628",              // BUMPED (was "20250318")
  "checksum": "9f2c…(sha256 of the data object)",
  "data": {
    "deployment_id": "6f1c9b2e-3a4d-4c8e-9f0a-1b2c3d4e5f60",
    "execution_time": 2.41,
    "infrahub_version": "1.4.0",
    "infrahub_type": "community",
    "python_version": "3.14.3",
    "platform": "aarch64",

    "workers": { "total": 2, "active": 2 },

    "branches": {
      "total": 12,
      "active": 4                            // NEW — open, non-system branches (int | null)
    },

    "accounts": {                            // NEW object
      "active": 7,                           // int | null
      "groups": 3                            // int | null
    },

    "activity_24h": {                        // NEW object — previous full UTC day [00:00, 00:00)
      "logins": 19,                          // int | null
      "unique_logins": 6,
      "checks_started": 88,                  // denominator; NOT additive with passed/failed
      "checks_passed": 80,
      "checks_failed": 8,
      "artifacts_created": 14,
      "artifacts_updated": 31,
      "branches_created": 9,
      "branches_merged": 5,
      "branches_deleted": 4,
      "webhooks_fired_success": 41,          // COMPLETED webhook-process runs
      "webhooks_fired_failure": 2            // FAILED / CRASHED webhook-process runs
    },

    "features": {
      "CoreArtifact": 12,
      "CoreResourcePool": 3,
      "CoreRepository": 5,
      "CoreGroup": 21,
      "CoreProfile": 4,
      "CoreProposedChange": 37,
      "CoreObjectTemplate": 2,
      "CoreTransformation": 6,
      "CoreWebhook": 3
    },

    "schema_info": {
      "node_count": 84,
      "generic_count": 19,
      "last_update": "2026-06-27T14:22:10.512000+00:00"
    },

    "database": {
      "database_type": "neo4j-community",
      "relationship_count": {
        "total": 481902
        // … existing per-graph-label counts unchanged …
      },
      "node_count": {
        "total": 154233,                     // unchanged — raw vertex total
        // … existing per-graph-label counts unchanged …
        "corenode": 4821,                    // NEW — all managed nodes (int | null)
        "user": 3902                         // NEW — user-defined-namespace nodes (int | null)
      },
      "servers": [],                         // populated on Neo4j Enterprise only
      "system_info": {
        "memory_total": 17179869184,
        "memory_available": 8231473152,
        "processor_available": 10
      }
    },

    "prefect": {
      "events": {                            // unchanged — cumulative over Prefect's ~7-day event retention
        "infrahub.account.logged_in": 142,
        "infrahub.account.logged_out": 60,
        "infrahub.artifact.created": 33,
        "infrahub.artifact.updated": 88,
        "infrahub.branch.created": 25,
        "infrahub.branch.deleted": 14,
        "infrahub.branch.merged": 12,
        "infrahub.branch.migrated": 0,
        "infrahub.branch.rebased": 7,
        "infrahub.group.auto_created": 5,
        "infrahub.group.member_added": 18,
        "infrahub.group.member_removed": 3,
        "infrahub.node.created": 1204,
        "infrahub.node.deleted": 231,
        "infrahub.node.updated": 5892,
        "infrahub.proposed_change.approved": 22,
        "infrahub.proposed_change.merged": 19,
        "infrahub.proposed_change.rejected": 2,
        "infrahub.proposed_change.review_requested": 24,
        "infrahub.repository.update_commit": 41,
        "infrahub.validator.started": 233,
        "infrahub.validator.passed": 205,
        "infrahub.validator.failed": 18
        // … remaining event names omitted for brevity …
      },
      "automations": {
        "action_trigger_rule": 0,
        "builtin": 6,
        "webhook": 3,
        "computed_attr_jinja2": 4,
        "computed_attr_python": 1,
        "computed_attr_python_query": 0,
        "display_label_jinja2": 12,
        "human_friendly_id": 8,
        "profile": 4
      },
      "work_pools": [
        { "name": "infrahub-worker", "type": "infrahubasync", "total_workers": 2, "active_workers": 2 }
      ]
    }
  }
}
```

> The three `checks_*` fields are **not additive** (`checks_started` is the denominator; derive pass/fail rate from it). `webhooks_*` is outcomes-only this phase. See [`contracts/telemetry-payload.md`](../dev/specs/telemetry-collection-infp-589/contracts/telemetry-payload.md) for the full field semantics.

## How to review

- Core logic: `backend/infrahub/telemetry/window.py` (windowing),
  `task_manager.py` (windowed Prefect event/flow-run counters),
  `database.py` (node counts), `tasks.py` (payload assembly, degradation).
- `dev/specs/telemetry-collection-infp-589/` has the full spec/plan trail;
  `contracts/telemetry-payload.md` is the payload contract.

## How to test

```bash
uv run invoke backend.test-unit backend.test-component --pattern telemetry
```

Manual end-to-end validation steps (trigger a collection, cross-check each
field against GraphQL, validate the window with the ad-hoc probe) are in
[`dev/specs/telemetry-collection-infp-589/manual-test-runbook.md`](https://github.com/opsmill/infrahub/tree/telemetry-collection-infp-589/dev/specs/telemetry-collection-infp-589/manual-test-runbook.md).

## Impact & rollout

- **Backward compatibility:** additive only; no fields removed or retyped.
  Consumers should tolerate unknown fields and `null`.
- **Config/env changes:** none. Daily schedule and local snapshot storage unchanged.
- **Deployment notes:** safe to deploy independently.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [x] External docs updated (not needed — `docs/deploy-manage/run-observe/telemetry.mdx`
      documents the storage/export mechanism, not individual payload fields)


dev/specs/telemetry-collection-infp-589/spec.md
```diff
@@ -0,0 +1,401 @@
...
+  snapshots tile exactly.
+- Activity counts are best-effort trend signals (event dispatch can drop), not
+  billing-grade exact figures; correctness is judged on windowing behavior, not
+  against an external ground truth.
+- Counts are gathered on the default branch through the standard branch-safe
+  count path, consistent with how the product computes them elsewhere.
+- The "graceful degradation" contract (per-metric isolation, `null` on failure,
+  `0` on genuine empty) applies uniformly to every in-scope metric.
+
+## Out of Scope
```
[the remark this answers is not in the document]
saltas888: Verified against the Jira card comments: DB-core reporting *was* pulled into Phase 1 per the 1 Jul scope-change call, but was deferred back out during implementation — the candidate setting (`server.threads.worker_count`) is REST-only and doesn't govern the Bolt path Infrahub uses, so it would only duplicate `processor_available` today. Still an open dependency on Fatih confirming the correct "licensed cores" setting. Added this reasoning to the Out of Scope section so it reads as a deliberate call, not an oversight (see `data-model.md` for the fuller investigation note).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

dev/specs/telemetry-collection-infp-589/tasks.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Tests for User Story 3 (write first, must fail) ⚠️
+
+- [x] T019 [P] [US3] Component test in `backend/tests/component/telemetry/test_datatabase.py`: seed N managed nodes via existing schema fixtures (`backend/tests/helpers/schema/`), independently compute N, assert `node_count["corenode"] == N` exactly and that `node_count["total"]` (raw) is unchanged and `>= N`.
+
+### Implementation for User Story 3
+
+- [x] T020 [US3] In `backend/infrahub/telemetry/database.py`, set `node_count["corenode"]` via `NodeManager.count(db, schema=InfrahubKind.NODE, branch=<default>)`, wrapped so a failure sets `corenode=None` without affecting `node_count["total"]` or the existing graph-label keys (do NOT use raw `count_nodes(label=...)`).
+- [x] T021 [US3] Add/extend a docstring or module note distinguishing the three node metrics at the namespace level: `total` (raw vertices), `corenode` (all managed nodes — `Core` + `Builtin` + user-defined namespaces), and `user` (customer-facing subset excluding the `Core` management namespace), noting they nest `user ⊆ corenode ⊆ total`. No tickets/IDs in source.
```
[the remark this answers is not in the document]
saltas888: Valid — the work was done but the checklist wasn't updated. `node_count["user"]` is implemented in `database.py` and covered by `test_gather_database_information_user_counts_only_user_namespaces` in `test_datatabase.py`. Checked off T021b/T021c.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

dev/specs/telemetry-collection-infp-589/data-model.md
```diff
@@ -0,0 +1,124 @@
...
+| `checks_passed`           | `int \| None` | `validator.passed` events in-window                     | `0`   | `null`  |
+| `checks_failed`           | `int \| None` | `validator.failed` events in-window                     | `0`   | `null`  |
+| `artifacts_created`       | `int \| None` | `artifact.created` events in-window                     | `0`   | `null`  |
+| `artifacts_updated`       | `int \| None` | `artifact.updated` events in-window                     | `0`   | `null`  |
+| `branches_created`        | `int \| None` | `branch.created` events in-window                       | `0`   | `null`  |
+| `branches_merged`         | `int \| None` | `branch.merged` events in-window                        | `0`   | `null`  |
+| `branches_deleted`        | `int \| None` | `branch.deleted` events in-window                       | `0`   | `null`  |
+| `webhooks_fired_success`  | `int \| None` | `webhook-process` flow runs in-window ending `COMPLETED`| `0`   | `null`  |
+| `webhooks_fired_failure`  | `int \| None` | `webhook-process` flow runs in-window ending `FAILED`/`CRASHED`/`TIMEDOUT` | `0` | `null` |
```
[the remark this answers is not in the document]
saltas888: Valid, and already flagged internally — `opsmill-implement-report.md` had noted `TIMEDOUT` isn't a real Prefect `StateType` and deferred the doc fix. `TIMEDOUT` removed from this table plus the payload contract, tasks.md, and research.md to match `WEBHOOK_FAILURE_STATES = [FAILED, CRASHED]` in `task_manager.py`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/task_manager.py
```diff
@@ -1,16 +1,35 @@
...
 from infrahub.trigger.constants import NAME_SEPARATOR
 from infrahub.trigger.models import TriggerType
 from infrahub.trigger.setup import gather_all_automations
 
-from .models import TelemetryPrefectData, TelemetryWorkPoolData
+from .models import TelemetryActivity24hData, TelemetryPrefectData, TelemetryWorkPoolData
+from .utils import safe_metric
+from .window import get_activity_window
+
+WEBHOOK_FLOW_NAME = "webhook-process"
```
dgarros: instead of hardcoding the value here, it would be best to pull the name of the workflow from the catalogue of workflows

backend/infrahub/telemetry/utils.py
```diff
@@ -1,11 +1,28 @@
...
+async def safe_metric[T](coro: Awaitable[T]) -> T | None:
+    """Run one metric coroutine in isolation, degrading a failure to ``None``.
+
+    A falsy result like ``0`` is preserved (measured, nothing to count); only an exception
+    yields ``None`` (logged), so one broken source nulls its own field, not the whole payload.
+    """
+    try:
+        return await coro
+    except Exception as exc:
+        log.warning("Telemetry metric collection failed; reporting null for this field: %s", exc)
```
dgarros, line 30: the default logger isn't gonna show in the logs in prefect, if you want this log to be displayed in prefect we need to use the prefect logger or `infrahub.tasks`

.specify/feature.json
```diff
@@ -0,0 +1,3 @@
+{
```
dgarros: FYI, we need to remove this file, I just merged a fix in stable to exclude it from git in the future 
https://github.com/opsmill/infrahub/pull/9806

backend/infrahub/telemetry/task_manager.py
saltas888: Done, `WEBHOOK_FLOW_NAME` now derives from `WEBHOOK_PROCESS.name` in the workflow catalogue instead of the hardcoded string.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/utils.py
saltas888: Fixed, `safe_metric` now logs via `infrahub.log.get_run_logger()` (the `infrahub.tasks` logger the worker registers with Prefect), so the degradation warning surfaces in task logs.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.specify/feature.json
saltas888: Removed the file in this push. Thanks for the heads-up on #9806 handling the gitignore going forward.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/database.py
```diff
@@ -47,8 +51,35 @@ async def get_system_info(db: InfrahubDatabase) -> TelemetryDatabaseSystemInfoDa
     )
 
 
+async def count_corenode(db: InfrahubDatabase) -> int:
```
ajtmccarty, line 55: so this is really a count of the active objects on the default (and maybe global) branch at this point in time. not sure if that is expected. and it should probably be documented somewhere if so (I have not looked)

backend/infrahub/telemetry/database.py
```diff
@@ -47,8 +51,35 @@ async def get_system_info(db: InfrahubDatabase) -> TelemetryDatabaseSystemInfoDa
...
+
+
+async def count_user_nodes(db: InfrahubDatabase) -> int:
+    """Count concrete nodes in user-editable namespaces.
+
+    Group-generic kinds are excluded because they do not carry the ``CoreNode`` label, which
+    would otherwise let ``user`` exceed ``corenode``.
+    """
+    default_branch = registry.get_branch_from_registry()
+    schema_branch = registry.schema.get_schema_branch(name=default_branch.name)
```
ajtmccarty: we should avoid using the `registry` wherever possible. we will hopefully delete it someday. I think you can use `db.schema` instead, which is a slight improvement

backend/infrahub/telemetry/task_manager.py
```diff
@@ -37,21 +64,124 @@ async def gather_prefect_events(client: PrefectClient) -> dict[str, Any]:
...
+
+
+@task(name="telemetry-gather-windowed-event", task_run_name="Gather Windowed Event Count", cache_policy=NONE)
+async def count_windowed_event(
+    client: PrefectClient, event_name: str, window_start: datetime, window_end: datetime
+) -> int:
+    """Count events of one name that occurred within ``[window_start, window_end)``."""
+    payload = _windowed_event_filter(event_name=event_name, window_start=window_start, window_end=window_end)
+    buckets = await _post_count_by(client=client, path="/events/count-by/event", payload=payload)
+    return sum(bucket["count"] for bucket in buckets)
```
ajtmccarty: is it possible for a bucket to be returned without `"count"`? typing doesn't enforce it, so you might want to use `.get("count", 0)`

backend/infrahub/telemetry/task_manager.py
```diff
@@ -37,21 +64,124 @@ async def gather_prefect_events(client: PrefectClient) -> dict[str, Any]:
...
+    after = DateTime.fromisoformat(window_start.isoformat())
+    before = DateTime.fromisoformat(inclusive_end(window_end).isoformat())
+
+    def runs_in_states(states: list[StateType]) -> FlowRunFilter:
+        return FlowRunFilter(
+            start_time=FlowRunFilterStartTime(after_=after, before_=before),
+            state=FlowRunFilterState(type=FlowRunFilterStateType(any_=states)),
+        )
+
+    # Count server-side: read_flow_runs caps at the API page size and would undercount a busy day.
```
ajtmccarty: I assume there's nothing to be done about this?

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,14 +70,72 @@ async def gather_feature_information() -> dict[str, int]:
...
+    )
+
+    return TelemetryAccountData(active=active, groups=groups)
+
+
+async def count_active_branches() -> int:
+    """Count open non-system branches.
+
+    Registry members minus the default and global branches (closed branches are evicted from
+    the registry). Async only so it composes with the degradation helper.
```
ajtmccarty: "Async only so it composes with the degradation helper." does not belong here. it's obviously async and the "degradation helper" is not the responsibility of this class

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,14 +70,72 @@ async def gather_feature_information() -> dict[str, int]:
...
+    """
+    return len([branch for branch in registry.branch.values() if not branch.is_default and not branch.is_global])
```
ajtmccarty: would be nice to skip the `registry`, but I think that would require going to the database for the Branches, which would be a fast query, but slower than memory

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,14 +70,72 @@ async def gather_feature_information() -> dict[str, int]:
...
+        return await gather_activity_24h(client=prefect_client)
+
+
 @task(name="telemetry-gather-data", task_run_name="Gather Anonynous Data", cache_policy=NONE)
-async def gather_anonymous_telemetry_data() -> TelemetryData:
+async def gather_anonymous_telemetry_data(
+    account_gatherer: Callable[[InfrahubDatabase], Awaitable[TelemetryAccountData]] = gather_account_information,
+    activity_gatherer: Callable[[], Awaitable[TelemetryActivity24hData]] = _default_activity_24h_gatherer,
+    active_branch_counter: Callable[[], Awaitable[int]] = count_active_branches,
+) -> TelemetryData:
```
ajtmccarty: perhaps too much for this PR, but this should be built using dependency injection instead of a chain of functions
the `gather_anonymous_telemetry_data` task should build an `AnonymousTelemetryGatherer` class that accepts `AccountTelementryGatherer`, `ActivityTelemetryGatherer` and `ActiveBranchCounter` as dependencies injected during construction. then `AnonymousTelemetryGatherer` can call the `gather` method (can be decorated with `@task`) on each of its dependencies and use their responses to build the `TelemetryData` response object

backend/infrahub/telemetry/utils.py
```diff
@@ -1,11 +1,60 @@
...
+def get_activity_window(now: datetime | None = None) -> tuple[datetime, datetime]:
+    """Return the half-open ``[window_start, window_end)`` for the previous full UTC day.
+
+    Anchored to the midnight-UTC boundary, not the raw ``now`` instant, so consecutive daily
+    runs tile exactly — no overlaps or gaps — whatever time the (jittered) cron fires.
+    """
+    reference = now if now is not None else datetime.now(tz=UTC)
+    window_end = floor_to_midnight_utc(reference)
+    window_start = window_end - WINDOW_LENGTH
+    return window_start, window_end
```
ajtmccarty, line 45: can this return the correct window instead of requiring the callers using it to call `inclusive_end` on the window end?

ajtmccarty (commented):
I haven't reviewed the tests yet

backend/infrahub/telemetry/database.py
saltas888: @saltas888 note to myself. Should this be ONLY THE DEFAULT BRANCH?

backend/infrahub/telemetry/task_manager.py
saltas888: Nothing to change functionally. We deliberately use `count_flow_runs` here for a server-side count, `read_flow_runs` isn't used because it pages at the API limit and would undercount a busy day. I reworded the comment so it says that plainly, it was leading with read_flow_runs which made it read like the method in use.

backend/infrahub/telemetry/utils.py
saltas888: Kept `get_activity_window` returning the half-open [start, end) on purpose. Prefect's filters are inclusive on both ends, and our start is already inclusive so it maps straight to since/after_, only the exclusive end needs the -1µs nudge to until/before_. Folding that into the window would make `window_end` mean "inclusive" in some paths and "exclusive" in others, and it would push the boundary-correctness onto. So the single `inclusive_end` at the two query sites is the one place our convention meets Prefect's inclusive API.

backend/infrahub/telemetry/database.py
saltas888: Yes, that's intended. It's the branch-safe count on the default branch at gather time, so the number matches what the product reports for that branch, distinct from the raw vertex total. Tightened the docstring to say that (default branch, at gather time), and it's also covered in the telemetry knowledge doc under Temporal models.

backend/infrahub/telemetry/database.py
saltas888: Done, `count_user_nodes` reads the schema branch from `db.schema` now instead of the registry singleton. Left the default-branch lookup on the registry since that's the branch source, happy to revisit if there's a db-side equivalent you'd prefer.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/task_manager.py
saltas888: Good call

backend/infrahub/telemetry/tasks.py
saltas888: Agreed, removed that sentence.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
saltas888: Went ahead and did this. The gather is now an `AnonymousTelemetryGatherer` built via DI, the account, activity, and active-branch collaborators are injected as constructor deps (each behind a `Protocol`), wired by `build_anonymous_telemetry_gatherer` at the flow entry point. Tests inject protocol doubles instead of callables. Not the aggregator/`supports()` pattern since the three sources return different types, just plain heterogeneous constructor DI.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/task_manager.py
```diff
@@ -37,21 +64,125 @@ async def gather_prefect_events(client: PrefectClient) -> dict[str, Any]:
...
+
+
+@task(name="telemetry-gather-windowed-event", task_run_name="Gather Windowed Event Count", cache_policy=NONE)
+async def count_windowed_event(
+    client: PrefectClient, event_name: str, window_start: datetime, window_end: datetime
+) -> int:
+    """Count events of one name that occurred within ``[window_start, window_end)``."""
+    payload = _windowed_event_filter(event_name=event_name, window_start=window_start, window_end=window_end)
+    buckets = await _post_count_by(client=client, path="/events/count-by/event", payload=payload)
+    return sum(bucket.get("count", 0) for bucket in buckets)
```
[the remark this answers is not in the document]
saltas888: We're safe here. Count-by returns Prefect's `EventCount` model where `count` is a required field (`count: int = Field(...)`), so a validated bucket can't come back without it — the `.get("count", 0)` default is just defensive and never actually fires. There's no real path where a malformed measurement is masked as `0`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,37 +70,145 @@ async def gather_feature_information() -> dict[str, int]:
...
+            database=await gather_database_information(db=self.database),
+            prefect=await gather_prefect_information(),
+        )
+
+        data.execution_time = time.time() - start_time
+
+        return data
+
+
+async def build_anonymous_telemetry_gatherer() -> AnonymousTelemetryGatherer:
```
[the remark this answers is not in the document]
saltas888: Good catch, that was doc-rot from the DI refactor. Updated the collection-flow diagram in `dev/knowledge/backend/telemetry.md` to name the current path (`build_anonymous_telemetry_gatherer()` → `AnonymousTelemetryGatherer.gather()` → `TelemetryData`) instead of the removed function.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/tests/unit/telemetry/test_workflow.py
```diff
@@ -36,20 +36,22 @@ def telemetry_mocks() -> Iterator[dict[str, Any]]:
     """Combined fixture providing all mocks needed for telemetry workflow tests."""
     repo = create_autospec(TelemetrySnapshotRepository, spec_set=True, instance=True)
     repo.save.return_value = None
+    gatherer = MagicMock()
```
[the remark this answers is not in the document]
saltas888: Done, switched the gatherer mock to `create_autospec(AnonymousTelemetryGatherer, spec_set=True, instance=True)`, matching the repository mock on the line above. Now if `gather` is renamed or removed the test fails instead of silently passing.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
saltas888: Update on this: we went ahead and switched it to the database after all. `count_active_branches` now uses `Branch.get_list(exclude_global=True, exclude_default=True)` instead of reading `registry.branch`. The daily gather is a cold path, so the round-trip is fine and it keeps this off the registry per the direction. I also reframed the guidance in `query-pattern.md` as hot-path (registry is fine) vs cold-path (use the DB) so the two stay consistent.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,37 +70,145 @@ async def gather_feature_information() -> dict[str, int]:
...
+            branch=default_branch,
+        )
+    )
+
+    return TelemetryAccountData(active=active, groups=groups)
+
+
+async def count_active_branches(db: InfrahubDatabase) -> int:
+    """Count open non-system branches, excluding the default and global branches."""
+    branches = await Branch.get_list(db=db, exclude_global=True, exclude_default=True)
```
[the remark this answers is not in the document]
saltas888: Fixed. `count_active_branches` now uses `Branch.get_list_count(exclude_default=True, exclude_terminal=True)` instead of `len(Branch.get_list(...))`. `get_list_count` runs `Query.count()`, which drops LIMIT/OFFSET, so the result is no longer capped at the default 1000.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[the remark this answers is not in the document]
saltas888: Fixed. `count_active_branches` now passes `exclude_terminal=True`, so merged/terminal branches are no longer counted and the metric keeps its open/non-system semantics.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

dev/knowledge/backend/query-pattern.md
```diff
@@ -4,6 +4,23 @@
...
+1. **Inject `SchemaManager`** into the constructor, built at the entry point (task/flow/API/CLI) — never import `registry` inside a component.
+2. **`db.schema`** when injection isn't practical — temporally correct against the operation's branch/time; `registry.schema` is not.
+3. **`registry.schema`** — avoid.
+
+```python
+class MyComponent:
+    def __init__(self, schema_manager: SchemaManager) -> None: ...
+```
+
+Components already accept `schema_manager`; entry points still pass `registry.schema`, concentrating the access at the boundary on the way to a `get_schema_manager()` accessor and retiring `registry` imports.
```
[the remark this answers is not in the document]
saltas888: Two parts here. Components **do** accept `schema_manager` today — e.g. `SchemaUpdateCoordinator`, the merge orchestrator/analyzer, `DiffCalculator`, and the schema number-pool synchronizer/upserter — wired at their entry points with `schema_manager=registry.schema` (`merge/builder.py`, `pools/tasks.py`, `api/schema.py`, `cli/db.py`). So that part isn't fabricated.

You're right that `get_schema_manager()` doesn't exist yet, though — it was referenced as the target, not as existing. I reworded the note to name the concrete components and point to the existing `get_database()` / `get_component()` accessors as the model, so it can't read as claiming a function that isn't there.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,37 +70,144 @@ async def gather_feature_information() -> dict[str, int]:
...
+            branch=default_branch,
+        )
+    )
+
+    return TelemetryAccountData(active=active, groups=groups)
+
+
+async def count_active_branches(db: InfrahubDatabase) -> int:
+    """Count open non-system branches (excludes the default, global, and terminal branches)."""
+    return await Branch.get_list_count(db=db, exclude_default=True, exclude_terminal=True)
```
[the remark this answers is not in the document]
saltas888: Behavior was correct — `get_list_count` hardcoded `exclude_global=True` internally, so global was always excluded — but you're right that the call site didn't show it while the docstring claimed it. Promoted `exclude_global` to a parameter (default `True`, so the two GraphQL callers stay unchanged), and `count_active_branches` now passes `exclude_global=True` explicitly. All three exclusions are visible at the call site and match the docstring.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/graphql/queries/branch.py
```diff
@@ -135,6 +135,7 @@ async def infrahub_branch_resolver(
             graphql_context=info.context,
             branch_filters=branch_filters,
             node_ordering=node_ordering,
+            exclude_global=True,
```
saltas888, line 138: this is so we align the signature of get_list_count and get_list

backend/infrahub/telemetry/task_manager.py
```diff
@@ -37,21 +64,118 @@ async def gather_prefect_events(client: PrefectClient) -> dict[str, Any]:
...
     events: dict[str, int] = {}
 
-    async def count_events(event_name: str) -> int:
-        payload = {"filter": {"event": {"name": [event_name]}}}
-        response = await client._client.post("/events/count-by/event", json=payload)
-        response.raise_for_status()
-        data = response.json()
-        if not isinstance(data, list) or len(data) == 0:
-            return 0
-        return data[0]["count"]
```
saltas888, line 47: align this as well so it uses _post_count_by

backend/infrahub/telemetry/models.py
```diff
@@ -59,6 +90,8 @@ class TelemetryData(BaseModel):
     platform: str
     workers: TelemetryWorkerData
     branches: TelemetryBranchData
+    accounts: TelemetryAccountData = Field(default_factory=TelemetryAccountData.default)
+    activity_24h: TelemetryActivity24hData = Field(default_factory=TelemetryActivity24hData.default)
```
ajtmccarty: I don't think the `.default` gains us much. it's just an extra step to get to `TelemetryAccountData()`

backend/infrahub/telemetry/database.py
```diff
@@ -47,8 +51,26 @@ async def get_system_info(db: InfrahubDatabase) -> TelemetryDatabaseSystemInfoDa
...
+
+async def count_user_nodes(db: InfrahubDatabase) -> int:
+    """Count concrete nodes in user-editable namespaces, excluding group-generic kinds."""
+    default_branch = registry.get_branch_from_registry()
+    schema_branch = db.schema.get_schema_branch(name=default_branch.name)
+    user_namespaces = [namespace.name for namespace in schema_branch.get_namespaces() if namespace.user_editable]
+    total = 0
+    for node_schema in schema_branch.get_schemas_for_namespaces(namespaces=user_namespaces):
+        if isinstance(node_schema, NodeSchema) and InfrahubKind.GENERICGROUP not in node_schema.inherit_from:
+            total += await NodeManager.count(db=db, schema=node_schema.kind, branch=default_branch)
```
ajtmccarty: this looks like it will run pretty quickly, but it could be a few hundred cypher queries for a complex schema. a single dedicated cypher query could be an improvement if necessary

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,37 +70,135 @@ async def gather_feature_information() -> dict[str, int]:
...
+class ActiveBranchCounter(Protocol):
+    async def count(self) -> int: ...
+
+
+class DefaultAccountGatherer:
+    def __init__(self, db: InfrahubDatabase) -> None:
+        self.db = db
+
+    async def gather(self) -> TelemetryAccountData:
+        return await gather_account_information(db=self.db)
```
ajtmccarty, line 111: I think that `gather_account_information` (and `gather_activity_24h` below) can be inside of each of these classes and still have the `@task` decorator, but I'm not entirely sure. if that's possible, then it would make sense to also split these components into their own files

backend/infrahub/telemetry/tasks.py
```diff
@@ -59,37 +70,135 @@ async def gather_feature_information() -> dict[str, int]:
...
+    return TelemetryAccountData(active=active, groups=groups)
+
+
+async def count_active_branches(db: InfrahubDatabase) -> int:
+    """Count open non-system branches (excludes the default, global, and terminal branches)."""
+    return await Branch.get_list_count(db=db, exclude_global=True, exclude_default=True, exclude_terminal=True)
+
+
+class AccountGatherer(Protocol):
+    async def gather(self) -> TelemetryAccountData: ...
```
ajtmccarty: all these `DefaultAbcGatherer/Counter` classes look like they are implementations of the same generic interface
```
class GathererInterface[T]:
    async def gather(self) -> T: ...
```

with this generic, the `AnonymousTelemetryGatherer` would look like this
```
class AnonymousTelemetryGatherer:
    def __init__(
        ...,
        account_gatherer: GathererInterface[TelemetryAccountData],
        activity_gatherer: GathererInterface[TelemetryActivity24hData],
        active_branch_counter: GathererInterface[int],
    ) -> None:
```

backend/tests/component/telemetry/test_task_manager.py
```diff
@@ -1,6 +1,314 @@
...
+        ActivityFieldCase(
+            name="artifacts_updated", field="artifacts_updated", event_name=ArtifactUpdatedEvent.event_name
+        ),
+        ActivityFieldCase(name="branches_created", field="branches_created", event_name=BranchCreatedEvent.event_name),
+        ActivityFieldCase(name="branches_merged", field="branches_merged", event_name=BranchMergedEvent.event_name),
+        ActivityFieldCase(name="branches_deleted", field="branches_deleted", event_name=BranchDeletedEvent.event_name),
+    ],
+    ids=lambda case: case.name,
+)
+async def test_gather_activity_24h_checks_artifacts_branches(
```
ajtmccarty, line 268: this looks like it calls the same function 8 times and then verifies a different part of the response based on the parameter. I think it all belongs in one test and it will run faster

ajtmccarty (approved):

backend/infrahub/telemetry/models.py
saltas888: Agreed, the classmethods were just `cls()` with extra steps. Removed them in c18d56a10, the fields now use `Field(default_factory=TelemetryAccountData)` directly.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
saltas888: Done in c18d56a10, the three protocols are now a single generic `GathererInterface[T]`, and the branch counter's `count()` became `gather()` so it fits the same interface.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/tests/component/telemetry/test_task_manager.py
saltas888: You're right, each case was re-running the full gather just to check one field. Collapsed into a single test in c18d56a10, one gather, eight assertions.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/tasks.py
saltas888: Prefect 3 does support `@task` on methods, so this should work. It's a broader restructuring than I want to fold into the open PRs though, captured it as [INBOX-140](https://opsmill.atlassian.net/browse/INBOX-140) to do as its own refactor.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/database.py
saltas888: Went ahead and optimized it here in 9b2fed57f. `count_user_nodes` now runs a single `CountNodesByKindsQuery`, one pass that resolves each node's winning `IS_PART_OF` edge and counts active nodes grouped by kind, instead of one `NodeManager.count` round trip per kind. Component tests pin the branch semantics (deleted and branch-only nodes excluded) and assert equivalence with the old per-kind counting as the oracle.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

backend/infrahub/telemetry/queries.py
```diff
@@ -28,6 +32,59 @@ def _normalize_end_date(value: str) -> str:
...
+        super().__init__(**kwargs)
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at)
+        self.params.update(branch_params)
+        self.params["kinds"] = self.kinds
+
+        query = """
+        MATCH (n:Node)
+        WHERE n.kind IN $kinds
```
ajtmccarty, line 70: one thing to note in the docstring perhaps: this query won't work for a generic kind b/c the `kind` property of all `:Node` vertices is always its concrete kind. this works for this use case b/c the query is passed all concrete and generic kinds, but if only generic kinds were passed, this would always return 0. if generic support is needed, then the query would need to use the labels of the vertex instead of the `kind` property

saltas888: Good point, went one step further than the docstring in 90bb0bde3. The constructor now takes `list[NodeSchema]` instead of `list[str]` and derives the kind strings itself, so passing a `GenericSchema` is a type error mypy/ty catch in CI rather than a silent zero at runtime. The docstring also documents the property-vs-labels constraint and the double-counting caveat a label-based sum would have.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #9807, opened 2026-07-03
fix: reduce integration-test memory usage and shard the CI job

fatih-acar:
## Summary

Two related changes that came out of profiling the backend integration suite:

1. **A product memory fix.** API worker processes retain a full GraphQL schema (~40MB of graphene types, ~1s to build) for **every schema version a branch has ever served** — the registry caches managers by schema hash but only evicts on branch deletion, so long-lived servers under schema churn grow until restart.
2. **CI sharding + test-suite memory reductions.** The same profiling showed memory (async-driver garbage + host-sized neo4j + the registry growth above) was the bottleneck capping pytest-xdist parallelism. With that removed, the `backend-tests-integration` job is split into parallel shards so the suite runs across runners instead of one.

## Key Changes

- **API servers release the previous GraphQL schema when a branch's schema changes.** `get_manager_for_branch` records each branch's active hash (using the previously dead `_branch_hash_activation_by_branch_name` field) and evicts a hash once no branch is on it, pruning registered types no remaining schema references — the cleanup `_reference_hash_schema_map`'s docstring always described. `purge_inactive` reuses the same path.
- **Integration test classes reuse the pristine core-schema GraphQL manager** instead of clearing the registry and rebuilding it per class (~1s CPU + ~40MB churn × ~74 classes).
- **Integration test workers stay small enough to scale out:** cyclic garbage from the async neo4j driver is collected and freed pages returned to the OS once per test class, and neo4j testcontainers get a bounded heap/page cache instead of sizing off host memory.
- **The `backend-tests-integration` CI job is sharded across parallel runners** (`shard: [a, b]`), the same way the component and docker-integration jobs are sharded. Shard membership lives in `INTEGRATION_TEST_SHARDS` (`tasks/backend.py`) and `backend.validate-integration-shards` guards that the shards partition the full suite exactly once, so a new test directory can never silently go unrun. Two shards at the default `-n 4` give the parallelism of a single `-n 8` job while spreading neo4j containers across runners.

## Measurements (memory ceiling lifted — 62GB/32-core host, full `backend/tests/integration`)

| Configuration | Result |
|---|---|
| `-n 4` before | green, 664s, worker RSS ends at 1.9–2.7GB |
| `-n 8` before | fails (host OOM → cascading fixture timeouts) |
| `-n 8` with the memory changes | **green, 461s, worker RSS peaks at 1.1–1.7GB** |

The `-n 8` run shows the memory work removed the parallelism ceiling; CI delivers that parallelism as two `-n 4` shards rather than one `-n 8` job, so no single runner carries eight neo4j containers. In `test_generic_migrations` the registry previously accumulated one full GraphQL schema per schema update within the class (~1GB growth per class); with eviction it holds at most the active ones.

## Test Plan

- New unit tests for the registry: `backend/tests/unit/graphql/test_registry.py` (hash reuse across calls and branches, activation-based eviction, protection while another branch still uses a hash, orphaned-type pruning, `purge_inactive`).
- `backend.validate-integration-shards` confirms the shard partition covers the suite exactly once (504 tests across 2 shards).
- Full `backend/tests/integration` suite green locally at `-n 4` and `-n 8` with the memory changes.

---
## Summary by cubic
Stops API worker memory growth from stale GraphQL schema caches and cuts integration-test memory so CI can shard the suite across runners. Upgrades mypy to 2.3.1 and moves `python-lint` to an arm64 runner, cutting backend type-checking from ~341s to ~15s.

**Bug Fixes**
- Record each branch's active schema hash and evict its previous hash when it moves on, keeping it if another branch still uses it.
- Generate the new schema before retiring the old hash so the two versions share registered types instead of rebuilding them.
- Centralize eviction in `evict_schema_hash`, which prunes types no remaining schema references; `purge_inactive` reuses it and clears per-branch activations.
- Fix the 25 type errors the mypy upgrade surfaced, plus the unannotated module-level `set()`/`{}` literals.

**Performance**
- Reuse the pristine core-schema GraphQL manager across test classes instead of rebuilding it per class.
- Bound neo4j testcontainer heap to 1GB and page cache to 512m.
- After each test class, collect cycles and call `malloc_trim` to return freed pages to the OS.
- CI runs the integration suite in two shards across parallel runners; `validate-integration-shards` guards that the shards cover the suite exactly once.
- mypy moved out of the test jobs into `python-lint`, and the heavy test jobs gate on unit tests passing.
- Results: full suite green at `-n 8` in 461s (worker RSS 1.1–1.7GB), where `-n 8` previously failed OOM and `-n 4` peaked at 1.9–2.7GB.

<sup>Written for commit 8d7d65d8c21a5e227aceacd880663176582a6592. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graphql/registry.py
```diff
@@ -138,8 +138,10 @@ def purge_inactive(self, active_branches: list[str]) -> set[str]:
         for schema_hash in list(self._branch_name_by_hash.keys()):
             if not self._branch_name_by_hash[schema_hash]:
                 # If no remaining branch is using the schema remove it completely
-                del self._branch_name_by_hash[schema_hash]
-                del self._branch_details_by_hash[schema_hash]
+                self.evict_schema_hash(schema_hash=schema_hash)
+
+        for branch_name in inactive_branches:
+            self._branch_hash_activation_by_branch_name.pop(branch_name, None)
```
ogenstad, line 144: One thing to consider here is if a user searches using `at` then we might end up with another schema hash. I think there are problems with that now if the schema differs, at the very least if the old schema has node kinds that have been removed in the newer schema. It could be that we shouldn't consider that aspect just now but I think at some point we should have a timestamp for when the schema was created and perhaps only delete schemas if they are unused and older than 15 minutes or so. Not requesting a change only mentioning this.

.github/workflows/ci.yml
```diff
@@ -609,16 +615,20 @@ jobs:
       - name: Install dependencies
         run: "uv sync --all-groups"
       - name: "Mypy Tests"
+        if: matrix.shard == 'a'
```
ogenstad: I'd suggest that we move mypy to the end of the python-lint job. While it takes a bit to run I think we can also shift it around so that the python-lint job runs as the same time as the unittests and if either of these fails we don't run any of the other tests.

backend/infrahub/graphql/registry.py
```diff
@@ -166,6 +203,7 @@ def get_manager_for_branch(self, branch: Branch, schema_branch: SchemaBranch) ->
             branch_details = self.cache_branch(branch=branch, schema_branch=schema_branch, schema_hash=schema_hash)
 
         self._add_branch_hash(branch_name=branch.name, schema_hash=schema_hash)
+        self._activate_branch_hash(branch_name=branch.name, schema_hash=schema_hash)
```
ogenstad, line 210: Do we have an issue here in the sense that the old hash would be removed before we're done registering the new schema and the hashes for that type? Might we end up deleting stuff that we later end up regenerating and as such create new types?

backend/tests/unit/graphql/test_registry.py
```diff
@@ -0,0 +1,155 @@
...
+def object_type() -> "type[InfrahubObject]":
+    return cast("type[InfrahubObject]", ObjectTypeStub)
+
+
+def input_type() -> "type[graphene.InputObjectType]":
+    return cast("type[graphene.InputObjectType]", ObjectTypeStub)
+
+
+def mutation_type() -> "type[InfrahubMutation]":
+    return cast("type[InfrahubMutation]", ObjectTypeStub)
```
ogenstad: Do these type functions with cast actually do anything? Or can we just add `InfrahubMutation` directly as input wherever this gets used? While it's not what is tested it just looks like a convoluted way to fool the type checker.

backend/infrahub/graphql/registry.py
fatih-acar: You're right. `GraphQLSchemaManager.__init__` only claims the `CoreNode` interface; every other type is claimed through `registry.get_*/set_*` inside `generate()`, which callers run after `get_manager_for_branch` has returned. So the previous hash was retired while 989/990 reference entries were still only claimed by it, and the caller then rebuilt every type from scratch (0 reused, ~0.93s on the core schema) instead of reusing the ~880 shared ones (~0.53s).

Fixed by generating the new manager's schema on the cache-miss path before `_activate_branch_hash` retires the old hash (free for callers, they all call `get_graphql_schema()` right after anyway), plus a unit test that fails without the reorder. No correctness impact either way: a single `generate()` is sync and internally consistent, and in-flight requests keep their own manager alive. It was pure churn.

backend/tests/unit/graphql/test_registry.py
fatih-acar: Agreed, replaced the helpers with the real classes (`InfrahubObject`, `graphene.InputObjectType`, `InfrahubMutation`); they import fine at runtime in the unit suite. The one cast left is `ManagerStub` for `_register_manager`, since a real `GraphQLSchemaManager` would build the very schema the test is counting.

.github/workflows/ci.yml
fatih-acar: Done: mypy is now the last step of `python-lint` (timeout 5→15 min, it takes ~3 min on the huge runners) and removed from the integration and functional jobs. `backend-tests-unit` no longer waits for `python-lint`, and the component/integration/functional/docker-integration/E2E/benchmark jobs need both, so a lint or unit failure stops them before they take a huge runner. The validation-only jobs (schema exports, docs) still only need `python-lint`.

backend/infrahub/graphql/registry.py
fatih-acar: Today `at` doesn't influence which schema is used: `prepare_graphql_params` always resolves `registry.schema.get_schema_branch(branch.name)`, so the manager is picked by the branch's current hash regardless of `at`, and nothing that was evicted is ever requested. If `at` later resolves the schema as of that timestamp, activation-based eviction would thrash between the historical and current hash on alternating requests. A created-at timestamp plus a grace period like you describe is the right shape for that; leaving it for a follow-up.

backend/infrahub/graphql/registry.py
```diff
@@ -164,8 +201,13 @@ def get_manager_for_branch(self, branch: Branch, schema_branch: SchemaBranch) ->
             branch_details = self._branch_details_by_hash[schema_hash]
         else:
             branch_details = self.cache_branch(branch=branch, schema_branch=schema_branch, schema_hash=schema_hash)
+            # A manager claims the registered types it is made of while it generates its schema.
+            # Generate it before retiring the branch's previous hash, otherwise every type the two
+            # schema versions share is pruned here and rebuilt from scratch by the caller.
+            branch_details.gql_manager.get_graphql_schema()
```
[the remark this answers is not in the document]
fatih-acar: Every production caller asks for the full schema: all eight `prepare_graphql_params` call sites use the defaults, and `endpoints.py`, `core/initialization.py` and `tasks/registry.py` request it explicitly. So this is the build they perform right after anyway, and the retained schema is the one the endpoint serves. Partial schemas are only requested by three calls in one component test.

A types-only path wouldn't be sufficient either: input and mutation types (~480 of the ~990 registry entries) are only claimed while the mutation mixin is generated, so they would still be pruned. Leaving as is.

AGENTS.md
```diff
@@ -11,7 +11,7 @@ Style: be direct and substantive. No filler, preamble, or pleasantries. Challeng
 - **Backend:** Python 3.14, FastAPI 0.131.0, Neo4j 2026.05 (driver 6.2), Pydantic 2.12
 - **Frontend:** TypeScript 5.9, React 19.2, Vite 8.0, Tailwind CSS 4.2
 - **Testing:** pytest 9.0, Vitest 4.1, Playwright 1.60
-- **Linting:** ruff 0.15, mypy 1.15, Biome 2.4
+- **Linting:** ruff 0.15, mypy 2.3, Biome 2.4
```
ogenstad, line 14: I'm wondering if we should get rid of this. I'm not sure I see the point of defining versions here and we should instead refer to the project or lock files. Fine for this change I guess, but thinking ahead.

changelog/+graphql-registry-eviction.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed unbounded memory growth in API server processes: the GraphQL type registry now evicts the previous schema of a branch when its schema changes, instead of retaining every historical GraphQL schema until restart.
```
ogenstad, line 1: I thing there's still some types that we aren't tracking within this registry. But we have the bulk of them.

ogenstad (approved):

PR #9814, opened 2026-07-04
feat: separate user-facing schema (write/read) from internal model [INFP-234]

dgarros:
# Why

The user-facing schema was a direct dump of Infrahub's internal model: it advertised internal-only fields (e.g. `inherited`) as settable and described constrained fields (e.g. attribute `kind`) as free-form strings. Users — and especially LLM agents — produced technically-invalid schemas as a result.

Goal: generate a distinct **write** contract (exactly what a user may submit) and **read** contract (adds visible-but-not-settable fields) from the single source of truth, validate submissions against the write contract at the load boundary, and let clients validate offline via the SDK.

Non-goals (this PR): retiring the SDK's hand-written schema models, publishing the write contract in the REST OpenAPI, and extracting schema code generation off the attribute definition model — all three documented as follow-ups in `dev/specs/archive/002-user-facing-schema/opsmill-implement-followups.md`.

Closes INFP-234 

## What changed

**Behavioral** — `POST /api/schema/load` now validates every submitted node, generic, and extension against the user-facing write contract:

- **Invalid *settable* values are rejected** with a field-level error naming the dotted path and the offending value: a constrained field outside its allowed set (attribute `kind`, relationship `kind`/`cardinality`, …), a value outside its allowed range or pattern, or a missing required field. Attribute `kind` and the other bounded fields are now published enums rather than free-form strings, so a kind that does not exist is caught at the boundary instead of failing later.
- **Read-only, internal, and unknown fields are tolerated and dropped**, at every nesting level — a schema read back from `GET /api/schema`, an export, or an older/hand-edited payload still loads without stripping anything. This is a change of direction taken during review: the first implementation rejected those fields, which made the read-back → edit → load round trip a breaking workflow for no benefit. Requirement FR-003 in `spec.md` still states the original reject rule and has not been rewritten.
- `version` is required on the write contract. The server already required it; the generated SDK model did not, so offline validation used to pass payloads the server rejected.
- Schemas can be validated offline with only the SDK installed, and the offline verdict reproduces the server's.

**Implementation**

- New `Visibility` axis on schema field definitions; every field classified write/read/internal from the single source `internal.py`.
- Generator emits write/read model families into the SDK (dedicated `generate_schema_sdk.j2`), driven by `SdkSchemaGenerator` in `tasks/backend.py`; internal models unchanged (byte-identical output).
- The load boundary is a `model_validator(mode="before")` on `SchemaLoadAPI` delegating to the SDK's `validate_schema` — one implementation, so server and SDK cannot drift. The generated write models are `extra="ignore"`, so pydantic itself drops non-write fields; there is no hand-rolled projection step.
- Reserved-namespace and reserved-suffix checks unified behind `SchemaRoot.validate_reserved_names()`.
- Incidental: generator no longer breaks on worktree paths containing shell/regex metacharacters.

**Unchanged:** `GET /api/schema` response shape; stored-schema read-back; no database or migration changes.

### Suggested review order

1. `backend/infrahub/core/schema/definitions/internal.py` — the `Visibility` axis + per-field classification.
2. `tasks/backend.py` + `backend/templates/generate_schema_sdk.j2` — write/read generation into the SDK.
3. `backend/infrahub/api/schema.py` — the `model_validator(mode="before")` boundary gate.
4. Tests, then generated SDK models (mechanical).

## How to review

Focus on the boundary gate and the visibility classification. Companion SDK PR (land first): https://github.com/opsmill/infrahub-sdk-python/pull/1135.

## How to test

```bash
uv run pytest backend/tests/unit/api/test_schema_load_contract.py
uv run pytest backend/tests/unit/core/schema/test_generated_visibility.py
uv run pytest backend/tests/functional/api/test_load_schema.py   # requires Neo4j testcontainers
uv run pytest backend/tests/component/api/test_40_schema.py
```

Full functional load suite: 17 passed locally (Neo4j testcontainers).

## Impact & rollout

- **Backward compatibility:** a payload is rejected only for an invalid *settable* value — an out-of-enum or out-of-range value, or a missing required field. Payloads carrying derived or internal fields (`inherited`, `used_by`, `hierarchy`, derived `kind`, …) keep loading, so the `GET` → edit → `POST` round trip needs no stripping. Changelog fragments and an upgrade note are included.
- **Performance:** negligible — schema load is infrequent and not a hot path.
- **Deployment notes:** requires a coordinated SDK release; the submodule pointer references the SDK branch, so land the companion SDK PR first.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [x] External docs updated (upgrade note)
- [x] Internal .md docs updated (dev/knowledge)

---
## Summary by cubic
Separates the user-facing schema into SDK-generated write/read models, wires `/api/schema` and `/api/schema/{load,check}` to those contracts, and serves the generated read models directly. `/api/schema/load` drops read‑only/internal/unknown fields, requires `version`, and validates constrained values with dotted, field‑level errors. Addresses INFP‑234.

- **New Features**
  - Generated `infrahub_sdk` models: `InfrahubSchemaWrite`/`InfrahubSchemaRead`; discriminated `Attribute` union; typed `computed_attribute` and dropdown `choices`; shared enums; read models compute `kind` and include `hash`; profile/template read models; write‑only `extensions`.
  - API/OpenAPI: `GET /api/schema` returns the generated read models (`NodeSchemaRead`, `GenericSchemaRead`, `ProfileSchemaRead`, `TemplateSchemaRead`); `POST /api/schema/{load,check}` declares `InfrahubSchemaWrite` and validates via `infrahub_sdk.schema.validate_schema()` (extras are dropped). OpenAPI and frontend REST types regenerated; CI guards generated SDK schema/protocol drift.
  - Internals: added `Visibility` on field definitions to drive write/read generation; removed the hand‑rolled write projection; unified reserved‑name checks under `validate_reserved_names()`; ADR 0010 recorded; corrected code‑generation docs.

- **Migration**
  - You can submit schemas that include read/internal/unknown fields (e.g., a read‑back/export); the server strips them, then enforces the write contract. `version` is now required.
  - Validate offline with `infrahub_sdk.schema.validate_schema()`; server and SDK give the same verdicts.
  - The deprecated attribute kind `String` is removed from SDK read models and OpenAPI; use `Text`.

<sup>Written for commit 34f98ee9706f2993e3462e535d81222d42d233cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/api/schema.py
```diff
@@ -75,35 +84,26 @@ def from_schema(cls, schema: MainSchemaTypes) -> Self:
         data["relationships"] = [
             relationship.model_dump() for relationship in schema.relationships if not relationship.internal_peer
         ]
+        # ``kind`` is a computed field on the generated read model (derived from namespace+name);
+        # only ``hash`` needs to be supplied here since it is computed by the server.
```
ajtmccarty: I don't think this comment is necessary

backend/infrahub/api/schema.py
ajtmccarty: not created by this PR, but `Mixin`s are a code smell and this one looks pretty easy to remove. it's only called in `get_schema()` and would be better as a private helper method

backend/infrahub/api/schema.py
```diff
@@ -115,8 +115,35 @@ class SchemaReadAPI(BaseModel):
...
     version: str
+    _internal_schema: SchemaRoot = PrivateAttr()
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_write_contract(cls, data: Any) -> Any:
+        # The user-facing "write" contract lives in the SDK; each submitted node/generic is
+        # validated against it so that fields the user may not set (read-level, internal, or
+        # unknown) and out-of-range constrained values are rejected at the boundary with a
+        # field-level message, before the payload is coerced into the richer internal models.
```
ajtmccarty: I think this comment is unnecessary. There's no information here that is not covered by either where `validate_write_schema` is imported from or the names of the functions.

backend/infrahub/api/schema.py
```diff
@@ -115,8 +115,35 @@ class SchemaReadAPI(BaseModel):
...
+            if not result.valid:
+                raise ValueError("; ".join(result.messages))
+        return data
+
+    @model_validator(mode="after")
+    def build_internal_schema(self) -> Self:
+        # The published request contract only exposes user-settable fields; the processing
+        # engine operates on the richer internal SchemaRoot. Building it here surfaces
+        # internal-only invariants (e.g. reserved keywords) as request-validation errors,
+        # and lets the endpoints feed the pipeline without re-deriving the conversion.
```
ajtmccarty: 4-line comment is overkill
`# Build the internal model here to surface validation errors at construction`

backend/infrahub/api/schema.py
```diff
@@ -357,10 +384,13 @@ async def load_schema(
...
     warnings: list[SchemaWarning] = []
+    candidate_schemas: list[SchemaRoot] = []
     for schema in schemas.schemas:
-        errors += schema.validate_namespaces()
-        errors += schema.validate_reserved_suffixes()
-        warnings += schema.gather_warnings()
+        internal_schema = schema.internal_schema
+        candidate_schemas.append(internal_schema)
+        errors += internal_schema.validate_namespaces()
+        errors += internal_schema.validate_reserved_suffixes()
```
ajtmccarty: looks like these two should be one function on `SchemaRoot`

backend/infrahub/core/schema/definitions/internal.py
```diff
@@ -575,15 +708,15 @@ def to_dict(self) -> dict[str, Any]:
             "Mainly relevant for internal object.",
             default_value=False,
             optional=True,
-            extra={"update": UpdateSupport.MIGRATION_REQUIRED},
+            extra={"update": UpdateSupport.MIGRATION_REQUIRED, "visibility": Visibility.WRITE},
```
ajtmccarty, line 718: is `WRITE` correct for this? I thought `read_only` was something that we set internally

backend/infrahub/core/schema/definitions/internal.py
```diff
@@ -709,7 +850,7 @@ def to_dict(self) -> dict[str, Any]:
             description="The ID of the relationship schema",
             kind="Text",
             optional=True,
-            extra={"update": UpdateSupport.NOT_APPLICABLE},
+            extra={"update": UpdateSupport.NOT_APPLICABLE, "visibility": Visibility.WRITE},
```
ajtmccarty, line 869: I think this only needs to be `WRITE` visible to allow renaming attributes and relationships. not an issue for this PR, but maybe there's a better way to do this where we don't have to make the ID-field writable. something like a`rename_from` field in the API, but not in internal schema. definitely something for a different time though

backend/infrahub/core/schema/definitions/internal.py
```diff
@@ -138,12 +155,113 @@ def default_definition(self) -> str:
 
         return f"default={formatted_default}"
 
+    @property
```
ajtmccarty, line 159: all of this logic for generating the code for a schema should be on a component, `SchemaCodeGenerator` or something, separate from the `SchemaAttribute`. the `SchemaAttribute` already has enough responsibilities.

backend/templates/generate_schema_sdk.j2
```diff
@@ -0,0 +1,57 @@
...
+
+from typing import Annotated, Any, Literal, Union
+
+from pydantic import BaseModel, ConfigDict, Field{% if with_computed_field %}, computed_field{% endif %}
+
+from .enums import (
+{%- for name in enum_names %}
+    {{ name }},
+{%- endfor %}
+)
```
ajtmccarty, line 13: could this be empty? maybe it's fixed by linting

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+        response = await test_client.post(
+            "/api/schema/load",
+            json=payload,
+            headers={"X-INFRAHUB-KEY": api_admin_token},
+        )
+
+        assert response.status_code == 422
+        body = response.text
+        assert "inherited" in body
+        assert "not_a_real_field" in body
```
ajtmccarty: would be good to assert a little more of the error message here

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+        response = await test_client.post(
+            "/api/schema/load",
+            json=payload,
+            headers={"X-INFRAHUB-KEY": api_admin_token},
+        )
+
+        assert response.status_code == 422
+        body = response.text
+        assert "kind" in body
+        assert "NotARealKind" in body
```
ajtmccarty: and here too

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+        initial_dataset: str,
+        test_client: InfrahubTestClient,
+        api_admin_token: str,
+    ) -> None:
+        """Reject non-write fields nested under `extensions.nodes`, naming each with its location.
+
+        An extension node's attribute carrying a read-level field (`inherited`) and an unknown
+        field must both be rejected, each located under `extensions.nodes[*]`. (An out-of-enum
+        `kind` is covered separately: it fails the attribute discriminator, which short-circuits
+        the per-variant field checks, so it cannot be reported alongside co-located fields.)
```
ajtmccarty: seems like too many implementation details for a description of a functional test

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+        assert response.status_code == 200
+        schema = response.json()
+
+        generics = {item["kind"]: item for item in schema["generics"]}
+        assert "TestDog" in generics["TestAnimal"]["used_by"]
+
+        nodes = {item["kind"]: item for item in schema["nodes"]}
+        dog = nodes["TestDog"]
+        inherited_attributes = [attr for attr in dog["attributes"] if attr["inherited"]]
+        assert inherited_attributes  # the `name` attribute inherited from TestAnimal reads back as inherited
```
ajtmccarty: you could ensure that this is checking the `name` attribute and only the `name` attribute here

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+            json=payload,
+            headers={"X-INFRAHUB-KEY": unprivileged_token},
+        )
+
+        assert response.status_code == 403
+        assert response.json()["errors"][0]["message"] == "You are not allowed to manage the schema"
+
+        # The targeted object was neither renamed nor deleted.
+        still_present = registry.schema.get(name="TestDevice", branch=default_branch.name)
+        assert still_present.id == existing_id
```
ajtmccarty, line 874: could be extra sure and validate `still_present.name == "Device"` here

backend/tests/functional/api/test_load_schema.py
```diff
@@ -596,3 +600,333 @@ async def test_remove_optional_text_field_value(
...
+        response_invalid = await test_client.post(
+            "/api/schema/load",
+            json={"schemas": [invalid_schema_root]},
+            headers={"X-INFRAHUB-KEY": api_admin_token},
+        )
+        assert response_invalid.status_code == 422
+
+        body = response_invalid.text
+        offending_fields = {error.field.rsplit(".", 1)[-1] for error in offline_invalid.errors}
+        assert {"inherited", "not_a_real_field"} <= offending_fields
```
ajtmccarty: seems like you should be able to assert `==` instead of `<=` here

docs/docs/release-notes/deprecation-guides/schema-load-write-contract.mdx
ajtmccarty, line 1: I'm not exactly sure how much of a pain this could be for users, but I wonder if it is worth adding an option to the api/load endpoint that is something like `ignore_extras` to allow loading a schema that include read-only fields and just accept that they will be ignored, regardless of value

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+    Deriving the grouping from the backend mapping keeps the generated discriminated union in
+    step with it; the caller supplies the set of parameters models its variants cover.
+
+    Raises:
+        ValueError: If the backend mapping yields a parameters model set that differs from
+            ``expected_parameters``, so the variant definitions are updated in lockstep.
+
+    """
+    from infrahub.core.schema.attribute_parameters import get_attribute_parameters_class_for_kind
+    from infrahub.types import ATTRIBUTE_KIND_LABELS
```
ajtmccarty, line 419: can this move to the top?

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+
+    Every node model exposes the derived ``kind`` (from namespace+name) on both write and read so
+    attribute access works when reading a locally-authored schema. It only *serializes* on read:
+    on write it is a plain property (not part of the payload), so ``kind`` never round-trips into
+    the write contract where ``extra="forbid"`` would reject it. The read variant additionally
+    carries the server-computed ``hash`` field. All propagate to node/generic/profile/template via
+    this base.
+    """
+    from infrahub.core.constants import UpdateSupport, Visibility
+    from infrahub.core.schema.definitions.internal import SchemaAttribute
```
ajtmccarty: samesies

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+
+
+def _sdk_profile_template_families(suffix: str) -> list[dict[str, Any]]:
+    """Read-only node families for profiles and templates (no write variant exists).
+
+    Both are a base node plus the list of generics they inherit from, mirroring their internal
+    counterparts; only the extra field is declared, the rest comes from the parent.
+    """
+    from infrahub.core.constants import UpdateSupport, Visibility
+    from infrahub.core.schema.definitions.internal import SchemaAttribute
```
ajtmccarty, line 505: and here

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+        BranchSupportType,
+        ComputedAttributeKind,
+        HashableModelState,
+        RelationshipCardinality,
+        RelationshipDeleteBehavior,
+        RelationshipDirection,
+        RelationshipKind,
+        SchemaAttributeDisplay,
+    )
+    from infrahub.types import ATTRIBUTE_KIND_LABELS
```
ajtmccarty, line 590: and some more

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+    "IPNetwork": "IPNETWORK",
+    "Boolean": "BOOLEAN",
+    "Checkbox": "CHECKBOX",
+    "List": "LIST",
+    "JSON": "JSON",
+    "Any": "ANY",
+}
+
+
+def _sdk_enum_specs() -> tuple[dict[str, list[tuple[str, str]]], dict[str, str], dict[str, str]]:
```
ajtmccarty: a dataclass response might make this more readable

tasks/backend.py
ajtmccarty: just check all the new functions for internal imports

tasks/backend.py
ajtmccarty, line 1: this new code looks like it should be its own class or module. it more than doubles the size of this file

tasks/backend.py
```diff
@@ -379,8 +397,583 @@ def _generate_schemas(context: Context) -> None:
...
+def _write_sdk_generated_init(generated: str) -> None:
+    init_content = (
+        '# Generated by "invoke backend.generate", do not edit directly\n'
+        "from . import enums, read, write\n\n"
+        '__all__ = ["enums", "read", "write"]\n'
+    )
+    Path(f"{generated}/__init__.py").write_text(init_content, encoding="utf-8")
+
+
+def _generate_schemas_sdk(context: Context) -> None:
```
ajtmccarty: this function has 8 functions defined inside of it. should probably be a class that defines methods

ajtmccarty (commented):
looks pretty good. I think some refactoring of where the code-generation logic lives is in order. that's the only significant issue though

backend/infrahub/core/schema/definitions/internal.py
dgarros: It's a supported field, it's actually mandatory to set a computed_attribute as read_only today

backend/templates/generate_schema_sdk.j2
dgarros: there are a lot of enums so very unlikely be be empty

tasks/backend.py
dgarros: I agree, I'm planning to do a refactoring of this code in another PR

tasks/backend.py
dgarros: These imports are inside the functions intentionally. This is a documented convention in the repository: PLC0415 (“import should be at the top level of a file”) is explicitly ignored globally and for tasks/**.py in pyproject.toml.

tasks/__init__.py eagerly imports every task submodule to assemble the Collection:

  from . import backend, bundle, demo, dev, docs, frontend, main, performance, release, schema, sdk

As a result, every invoke command—including invoke --list and unrelated frontend/docs tasks—imports tasks/backend.py from top to bottom. If we move the infrahub.* and jinja2 imports to the module level, they will be imported on every invoke invocation, even when they are not needed, which would add unnecessary startup overhead.

The lightweight, always-needed imports (invoke, the standard library, and sibling .shared/.utils modules) are already at the top of the file. Only the heavier or optional application imports are loaded lazily, which is the intended design.

I agree that the long-term solution is to move this logic out of the tasks module altogether, perhaps into a dedicated command or module. I’m planning to work on that. For this PR, though, I’d prefer to keep the scope small and merge it sooner so we can identify any side effects before making a larger structural change.

backend/infrahub/api/schema.py
ajtmccarty: still think this can be removed

backend/infrahub/api/schema.py
```diff
@@ -115,8 +116,37 @@ class SchemaReadAPI(BaseModel):
...
+    _internal_schema: SchemaRoot = PrivateAttr()
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_write_contract(cls, data: Any) -> Any:
+        # A submitted schema may carry fields that are not part of the write contract: read-only or
+        # internal fields (a schema read back from Infrahub, or a full internal dump), or unknown
+        # fields from an older or hand-edited payload. All of them are dropped silently by the
+        # projection below rather than rejected, so an otherwise-valid schema still loads. What
+        # remains is the user-facing write contract, validated for out-of-range constrained values.
```
ajtmccarty: looks like a docstring for `project_to_write_contract`

backend/infrahub/api/schema.py
ajtmccarty: still too long

backend/infrahub/api/schema_write_projection.py
```diff
@@ -0,0 +1,95 @@
...
+
+from infrahub_sdk.schema.generated.write import InfrahubSchemaWrite
+from pydantic import BaseModel
+from pydantic.fields import FieldInfo
+
+_UNION_ORIGINS = {typing.Union, types.UnionType}
+_SEQUENCE_ORIGINS = {list, tuple}
+
+
+def project_to_write_contract(schema: dict[str, Any]) -> dict[str, Any]:
```
ajtmccarty: this looks like it could use some tests. I'd also expect this all to be something that pydantic can do for us, so it might be good to explain why that's not the case. I'd think that something like `WriteModelClass.model_validate(schema, ignore_extra=True)` would work for this

ajtmccarty (commented):
looks like many of my comments from earlier review on this one are still outstanding, so I'm going to leave it for now

backend/templates/generate_schema_sdk.j2
```diff
@@ -0,0 +1,57 @@
...
+
+class {{ family.class_name }}({{ family.parent }}):
+    model_config = ConfigDict({{ model_config_args }})
+{%- for attribute in family.attributes %}
+    {{ attribute.name }}: {{ attribute.external_type_annotation }} = Field(
+        {{ attribute.external_default_definition }},
+        description="{{ attribute.description }}",
+        {{ attribute.external_pattern }}
+        {{ attribute.min }}
+        {{ attribute.max }}
```
saltas888, line 42: in case we have this. Should we add a validtor that checks min < max? Or we do this elsewhere?

saltas888, line 42: in case we have this. Should we add a validtor that checks min < max? Or we do this elsewhere?

bilalabbad (approved):

backend/templates/generate_schema_sdk.j2
dgarros: we do that elsewhere, this code is just to convert from our internal format to Pydantic

tasks/backend.py
```diff
@@ -379,8 +379,94 @@ def _generate_schemas(context: Context) -> None:
...
+            },
+            {
+                "class_name": f"GenericSchema{suffix}",
+                "parent": f"BaseNodeSchema{suffix}",
+                "attributes": _visible(generic_stripped),
+            },
+        ]
+
+    def _root(suffix: str, with_version: bool) -> dict[str, Any]:
+        version_field = ["version: str | None = None"] if with_version else []
```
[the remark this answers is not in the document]
dgarros: Valid — fixed in 34b7f95f3.

Confirmed the parity gap: `validate_schema()` validates against `InfrahubSchemaWrite`, whose generated `version` field was `str | None = None`, so a payload with no `version` was reported valid offline and then rejected by `POST /api/schema/load` (`SchemaLoadAPI` required it, on develop too).

Fixed at the source rather than in the generated file: `_sdk_root` in `tasks/backend.py` now emits `version: str` for the write variant only, and the models were regenerated via `invoke backend.generate`. The read root still carries no `version` field at all. With the parent requiring it, the `version: str` override on `SchemaLoadAPI` became redundant and was dropped — verified the field stays required, keeps its position, and `schema/openapi.json` is byte-identical (so no GraphQL or frontend-type regeneration either).

Coverage: a new unit test asserts the offline validator and the load contract reach the same verdict for every load-contract case, including one with no `version`, plus an SDK-side test that a missing `version` is rejected naming that field. Swept both repos for payloads built without a version — every schema fixture already carries one, so nothing else needed changing.

The generated-model change lives in the submodule (`3262c61`) and still has to land on the companion SDK pull request before the pointer bump here is valid for other checkouts.

python_sdk
```diff
@@ -1 +1 @@
-Subproject commit 25cf9394da583b43ffd124f2bcbcd748a4ec86b7
+Subproject commit d15ce11be0d0466d169f5fbfe3fccb49b283095e
```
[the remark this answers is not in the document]
dgarros: Valid — fixed in 34b7f95f3.

Confirmed the parity gap: `validate_schema()` validates against `InfrahubSchemaWrite`, whose generated `version` field was `str | None = None`, so a payload with no `version` was reported valid offline and then rejected by `POST /api/schema/load` (`SchemaLoadAPI` required it, on develop too).

Fixed at the source rather than in the generated file: `_sdk_root` in `tasks/backend.py` now emits `version: str` for the write variant only, and the models were regenerated via `invoke backend.generate`. The read root still carries no `version` field at all. With the parent requiring it, the `version: str` override on `SchemaLoadAPI` became redundant and was dropped — verified the field stays required, keeps its position, and `schema/openapi.json` is byte-identical (so no GraphQL or frontend-type regeneration either).

Coverage: a new unit test asserts the offline validator and the load contract reach the same verdict for every load-contract case, including one with no `version`, plus an SDK-side test that a missing `version` is rejected naming that field. Swept both repos for payloads built without a version — every schema fixture already carries one, so nothing else needed changing.

The generated-model change lives in the submodule (`3262c61`) and still has to land on the companion SDK pull request before the pointer bump here is valid for other checkouts.

python_sdk
```diff
@@ -1 +1 @@
-Subproject commit b25c093ed0680540dbbcfda47766afd9e49dd379
+Subproject commit dc46764a45e9cb43bd4ab7e0af987119b3777389
```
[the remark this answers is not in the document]
dgarros: Valid — fixed in 34b7f95f3.

Confirmed the parity gap: `validate_schema()` validates against `InfrahubSchemaWrite`, whose generated `version` field was `str | None = None`, so a payload with no `version` was reported valid offline and then rejected by `POST /api/schema/load` (`SchemaLoadAPI` required it, on develop too).

Fixed at the source rather than in the generated file: `_sdk_root` in `tasks/backend.py` now emits `version: str` for the write variant only, and the models were regenerated via `invoke backend.generate`. The read root still carries no `version` field at all. With the parent requiring it, the `version: str` override on `SchemaLoadAPI` became redundant and was dropped — verified the field stays required, keeps its position, and `schema/openapi.json` is byte-identical (so no GraphQL or frontend-type regeneration either).

Coverage: a new unit test asserts the offline validator and the load contract reach the same verdict for every load-contract case, including one with no `version`, plus an SDK-side test that a missing `version` is rejected naming that field. Swept both repos for payloads built without a version — every schema fixture already carries one, so nothing else needed changing.

The generated-model change lives in the submodule (`3262c61`) and still has to land on the companion SDK pull request before the pointer bump here is valid for other checkouts.

backend/infrahub/api/schema_write_projection.py
dgarros: You're right, pydantic does do this for us. Every generated write model sets `extra="ignore"` (with `use_enum_values=True`), at every nesting level, and the per-kind discriminated unions resolve before the extras are dropped — so `InfrahubSchemaWrite.model_validate(payload)` already does exactly what the hand-rolled projection did. Rather than add tests for the helper, it is deleted in 156a0d9fa.

Verified before removing: 16 payloads validated through `SchemaLoadAPI` with the projection and with it replaced by the identity function, comparing the accept/reject verdict, the exact error message, `model_dump()` and `internal_schema.model_dump()`. All 16 identical — full internal `SchemaRoot` dump, minimal write payload, unknown/read-only keys at node, attribute, relationship and extension level, a Jinja2 and a TransformPython `computed_attribute` (with a junk key inside), cross-kind `parameters`, `extensions: null`, and the rejections (out-of-enum attribute kind, out-of-enum computed-attribute kind, too-short attribute name, missing version).

The boundary behaviour itself is now covered in `backend/tests/unit/api/test_schema_load_contract.py`: the parametrized contract cases gained attribute-level, relationship-level and `computed_attribute` cases plus `extensions: null`, and there are three focused tests — non-write keys never reach the internal schema (a submitted `inherited: true` stays False, while the settable `state` does survive), a computed attribute keeps its own union variant, and an out-of-enum attribute kind is still rejected with the message naming the bad tag.

backend/infrahub/api/schema.py
dgarros: Agreed — the comment is gone in 156a0d9fa, along with the projection it was describing.

backend/infrahub/api/schema.py
dgarros: Removed in 156a0d9fa: the projection it described is deleted, and the validator now carries a single line about why it raises.

backend/infrahub/api/schema.py
```diff
@@ -115,8 +117,45 @@ class SchemaReadAPI(BaseModel):
...
+    @model_validator(mode="before")
+    @classmethod
+    def validate_write_contract(cls, data: Any) -> Any:
+        # A submitted schema may legitimately carry read-only/internal fields — a schema read back
+        # from Infrahub, or a full internal dump. The internal schema accepts those but still
+        # rejects fields that are unknown or invalid for their context (a typo, or a node-only
+        # field on an extension), drawing the tolerate/reject line a uniform extra="forbid" cannot.
+        if isinstance(data, dict):
+            try:
+                SchemaRoot.model_validate(data)
```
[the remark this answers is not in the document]
dgarros: No longer applies. The `SchemaRoot.model_validate(data)` call in the before-validator that this refers to does not exist any more; the before-validator only runs the offline write-contract validation. A payload with `extensions: null` is accepted — verified directly, and it is now a case in `backend/tests/unit/api/test_schema_load_contract.py`.

docs/docs/release-notes/deprecation-guides/schema-load-write-contract.mdx
dgarros: That behaviour is already the default, so there is no need for an opt-in flag. The load boundary tolerates and drops read-only, internal and unknown fields at every nesting level (nodes, generics, extensions, and their nested attributes and relationships) — the generated write models are `extra="ignore"`, so a schema read back from `GET /api/schema` and re-submitted unchanged loads fine, and Infrahub re-derives `inherited`, `used_by`, `hierarchy` and the derived `kind` itself. Only genuinely invalid *settable* values are rejected: out-of-set values, out-of-range or wrong-typed values, missing required fields, and a missing `version`.

You were right to flag the page, though — for a different reason than the one you were pointing at. The guide had not caught up with that change: it still claimed payloads carrying derived fields are rejected and walked the reader through stripping them by hand, and its example validator output (`nodes[0].inherited: Extra inputs are not permitted`) is no longer something the validator can emit. That was simply wrong. Corrected in c9a243d7d — the page now states plainly that extra fields are dropped, reframes who is affected around invalid values, and uses error strings captured from the shipped validator rather than invented ones.

backend/tests/functional/api/test_load_schema.py
dgarros: Done in ef44ea0c7: it now asserts exactly one validation message and that it contains the path `nodes[0].attributes[0]:`, the constraint (`Input tag 'NotARealKind' found using 'kind' does not match any of the expected tags`) and the received value, instead of bare `kind`/`NotARealKind` substring checks. The 22-member enum listing is deliberately left out as too volatile.

backend/tests/functional/api/test_load_schema.py
dgarros: The test this was on (`test_schema_load_rejects_non_write_and_unknown_fields`) no longer rejects — extra/read-level fields are now dropped, so it asserts a 200 and has no error message to check. Applied the point to the two rejection assertions that survive, in ef44ea0c7: the relationship-cardinality test now asserts one message containing `nodes[0].relationships[0].cardinality:`, `Input should be 'one' or 'many'` and `(received: 'both')`, and the SDK/endpoint parity test now asserts the single offline message names the path and the bad value and appears verbatim in the server's message.

backend/tests/functional/api/test_load_schema.py
dgarros: Added in ef44ea0c7: alongside the id it now asserts `still_present.name == "Device"` and `still_present.namespace == "Test"`, plus that `TestDeviceRenamed` is absent from the schema, so a rename could not slip through under a new kind.

backend/tests/functional/api/test_load_schema.py
dgarros: Tightened in ef44ea0c7: the inherited attributes are collected as a set of names and asserted to be exactly `{"name"}`, and the full attribute set is asserted to be `{"name", "breed"}` so a stray inherited or missing local attribute fails.

backend/tests/functional/api/test_load_schema.py
dgarros: Agreed — that docstring explained the discriminator short-circuit and where the extra fields come from rather than what the test checks. Its successor (the extensions tolerance test) is now a single line stating the accepted behaviour, and the same trim was applied to its non-extension sibling, in ef44ea0c7.

backend/tests/functional/api/test_load_schema.py
dgarros: No longer applies: that `<=` compared the offline offending-field set against `{"inherited", "not_a_real_field"}`, and it went away with the reject-extra-fields behaviour — those fields are now dropped, so there is no field set to compare. The nearest surviving subset-style check was switched to exact equality in ef44ea0c7 (`used_by == ["TestDog"]` instead of `"TestDog" in used_by`).

backend/infrahub/api/schema.py
dgarros: Agreed — the mixin is gone. `from_schema` is now a module-level private helper `_api_schema_from_schema(model=..., schema=...)` with a PEP 695 type parameter bound to `BaseNodeSchemaRead`, so the return type stays precise without `Any` or a cast. The four API classes now inherit only from their generated read model. Confirmed nothing outside this file referenced the mixin or `from_schema`; all five call sites (four in the schema-read endpoint plus the by-kind endpoint) were updated. Class names are unchanged, so `schema/openapi.json` regenerates byte-identical. Done in fd04d2d73.

backend/infrahub/api/schema.py
dgarros: Comment removed in fd04d2d73 — the `data["hash"] = ...` line stays.

backend/infrahub/api/schema.py
dgarros: Agreed - they were only ever called as a pair and both results went into the same errors list. Folded into a single public `SchemaRoot.validate_reserved_names()` that returns the combined error list, with `_validate_namespaces()` / `_validate_reserved_suffixes()` demoted to private helpers. Both call sites in `backend/infrahub/api/schema.py` (schema load and schema check) now call the one method, and the component tests exercise it while still keeping one test per error class. 738123eca2846015c7f3ef46ea6503cb096fafab

tasks/backend.py
dgarros: Agreed — all eight closures shared the same locals, so they were instance state waiting for a class. `SdkSchemaGenerator` now holds the schema definitions, the jinja environment, the enum lookups and the field building blocks as attributes, and the eight nested functions are methods. The per-variant 6-tuple also became a frozen `SdkVariant` dataclass. `_generate_schemas_sdk` remains as a thin entry point so the invoke task surface is unchanged, and the backend/jinja2 imports stay method-local so `tasks/backend.py` still imports without the backend installed.

`_sdk_base_node_family` and `_render_sdk_enums` moved onto the class too since they read that same state; the other module-level `_sdk_*` helpers are standalone and stayed. Regenerated output is byte-identical (diffed against a pre-refactor snapshot; `backend.validate-generated` clean). 29cc68e69

backend/infrahub/api/schema.py
dgarros: Trimmed to a single line in bc92789: `# Built here to surface validation errors at construction.`

dev/specs/002-user-facing-schema/PRD-user-facing-schema-separation.md
```diff
@@ -0,0 +1,126 @@
...
+- **FR-005**: The read model MUST include `read`-level fields and exclude `internal` fields. *Verify:* GET returns `inherited`; it never returns internal bookkeeping fields.
+- **FR-006**: The write and read models MUST be generated into the Python SDK and be importable with only the SDK installed (no server, no backend package). *Verify:* import and validate a payload in an SDK-only environment.
+- **FR-007**: The backend MUST validate write submissions and serialise read responses using the SDK-hosted models, not a backend-local copy. *Verify:* a rule change in the generated model changes both server and SDK behaviour.
+- **FR-008**: The SDK's existing hand-written schema models MUST be replaced by the generated write/read models, leaving no third parallel definition. *Verify:* old model classes are removed; SDK callers use the generated ones.
```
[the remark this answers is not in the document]
dgarros: Good catch — the PRD was misleading. In bc92789 FR-008 is kept but flagged as deferred and not delivered in this cycle, and Out of Scope now states explicitly that the generated models become the single source of truth for the write boundary while the SDK's hand-written read models and their consumers stay in place, cross-referencing the follow-ups file (which now carries an entry with the reason and the recommended approach). The Implementation Decisions line that claimed the hand-written models are removed was corrected too.

dev/specs/archive/002-user-facing-schema/plan.md
```diff
@@ -0,0 +1,99 @@
...
+
+**Project Type**: Web service backend + shipped client library (SDK). No frontend surface this cycle.
+
+**Performance Goals**: Not a hot path — schema load/read is infrequent and human/agent-driven. Generation is a dev-time step. No specific latency target.
+
+**Constraints**: Generated files must be byte-stable (idempotent regeneration, validated in CI). SDK write/read models must import with zero backend dependency **and be committed, shipped artifacts in the SDK package** (not build-time-only), so a consumer installing only the SDK obtains them. Backward compatibility: previously loadable write-shaped schemas must still load; stored schemas must still read back. Server and SDK ship one contract and must be released compatibly; the submission `version` field is the skew anchor (local validation advisory, server authoritative).
+
+**Scale/Scope**: Four schema families (node, generic, attribute, relationship) × ~20–30 fields each; one generator template; two API endpoints; one SDK schema module.
+
+## Constitution Check
```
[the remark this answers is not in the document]
dgarros: Reworded in bc92789. Principle VII no longer claims the hand-written models are removed here: it now says the generated models become the sole definition of the write boundary consumed by both server and SDK, and that retiring the remaining hand-written models is the tracked follow-up which realises the net reduction of parallel definitions, with a link to the follow-ups file. The gate result is unchanged.

backend/infrahub/core/schema/definitions/internal.py
dgarros: Agreed on the direction, but this is deferred to a follow-up rather than done here — recorded in dev/specs/002-user-facing-schema/opsmill-implement-followups.md in bc92789, with the reasoning and a recommended approach.

The reason: most of these properties (type_annotation, object_kind, default_definition, pattern, min, max) predate this change and are consumed directly by the jinja templates as attribute properties, so extracting them reshapes the template contract for both the internal and the SDK generator. Moving only the new external_* half would leave the same logic split across two homes, which is worse than where it is now. The follow-up covers both halves in one pass, and notes SdkSchemaGenerator in tasks/backend.py as a natural home for the SDK-side part.

frontend/app/src/entities/nodes/object/ui/object-table/object-table-schema-selector.tsx
```diff
@@ -60,7 +60,7 @@ export function ObjectTableSchemaSelector() {
       >
         <ComboboxList shouldFilter>
           <ComboboxItem
-            value={baseSchema.hash}
+            value={baseSchema.hash!}
```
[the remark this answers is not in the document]
dgarros: The API always populates `hash`. Every schema in a `GET /api/schema` response is built by `_api_schema_from_schema` in `backend/infrahub/api/schema.py`, which does `data["hash"] = schema.get_hash()` unconditionally for all four `API*Schema` models, and every `ModelSchema` value reaching this component comes from that response (schema provider -> schema atoms -> `getSchema`). The `hash: str | None` optionality is a property of the shared read model, which is also used to parse a locally-authored schema with no server-computed hash, not of the response, so there is no runtime case here to fall back from.

If the type-level `!` is what is objectionable, the fix would be at the model level (making `hash` required on the read contract), not a call-site fallback — but it is deliberately optional there.

backend/tests/integration/diff/test_diff_rebase.py
```diff
@@ -257,7 +257,7 @@ async def add_branch_3_changes(
         manufacturer_schema = branch_3_schema.get(name=TestKind.MANUFACTURER, duplicate=True)
         manufactuer_continent_rel = manufacturer_schema.get_relationship("continents")
         manufactuer_continent_rel.state = HashableModelState.ABSENT
-        schemas_to_load = {"version": "1.0", "nodes": [continent_schema.model_dump(), manufacturer_schema.model_dump()]}
+        schemas_to_load = to_write_schema_dict(SchemaRoot(version="1.0", nodes=[continent_schema, manufacturer_schema]))
```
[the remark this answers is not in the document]
dgarros: No longer applies. The `SchemaRoot(...)` construction this refers to was reverted — the line now reads `schemas_to_load = {"version": "1.0", "nodes": [continent_schema.model_dump(), manufacturer_schema.model_dump()]}` — and the file is unchanged by this PR.

backend/tests/helpers/schema/__init__.py
```diff
@@ -26,6 +26,29 @@
...
+    ``jinja2_template`` and ``transform``; the write contract's per-kind discriminated union rejects
+    the one that does not match ``kind``, so the ``None`` side is stripped here.
+    """
+    data = schema.model_dump(exclude_unset=True, exclude_defaults=True)
+    # Read-only fields that may hold a non-default value on a schema read back from the database
+    # (so exclude_defaults keeps them) but are still rejected by the write contract.
+    for collection in ("nodes", "generics"):
+        for node in data.get(collection, []) or []:
+            node.pop("used_by", None)
+            for attribute in node.get("attributes", []) or []:
```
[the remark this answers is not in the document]
dgarros: No longer applies. `to_write_schema_dict` was removed (no references left anywhere in the repo) and this file is unchanged by this PR. The schema-load boundary now tolerates and drops non-write fields itself, so nothing needs stripping test-side.

ajtmccarty (approved):

Issue #7498, opened 2025-10-29
bug: Too verbose errors when trying to create GraphQL queries from git that already exists


ogenstad:
### Component

Git Integration

### Infrahub version

1.4.12

### Current Behavior

When you import a repository and the import tries to create a GraphQL query that previously existed but wasn't tied to the current repository there's a verbose tracelog that can be seen in the worker logs:

```python
2025-10-27T09:11:02.383075Z [info     ] New Graphql Query 'backbone_service' found, creating [prefect.task_runs]
2025-10-27T09:11:02.405639Z [error    ] Task run failed with exception: GraphQLError('An error occurred while executing the GraphQL Query \nmutation ($value_18724e16673987e53297c5137afc20ec: String!) {\n    CoreGraphQLQueryCreate(\n        data: {\n            name: {\n                value: "backbone_service"\n                is_protected: true\n                source: "18724e15-38ed-4bc5-32fb-c51a00d32704"\n            }\n            query: {\n                value: $value_18724e16673987e53297c5137afc20ec\n                is_protected: true\n                source: "18724e15-38ed-4bc5-32fb-c51a00d32704"\n            }\n            repository: {\n                id: "18724e15-38ed-4bc5-32fb-c51a00d32704"\n                _relation__is_protected: true\n                _relation__source: "18724e15-38ed-4bc5-32fb-c51a00d32704"\n            }\n        }\n    ){\n        ok\n        object {\n            id\n        }\n    }\n}\n, [{\'message\': "Violates uniqueness constraint \'name\'", \'locations\': [{\'line\': 3, \'column\': 5}], \'path\': [\'CoreGraphQLQueryCreate\']}]') [prefect.task_runs]
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/prefect/task_engine.py", line 1464, in run_context
    yield self
  File "/usr/local/lib/python3.12/site-packages/prefect/task_engine.py", line 1543, in run_task_async
    await engine.call_task_fn(txn)
  File "/usr/local/lib/python3.12/site-packages/prefect/task_engine.py", line 1481, in call_task_fn
    result = await call_with_parameters(self.task.fn, parameters)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/source/backend/infrahub/git/integrator.py", line 573, in import_all_graphql_query
[...]

Issue #4589, opened 2024-10-09
task: Excessive logs for user errors


ogenstad:
### Component

API Server / GraphQL

### Task Description

If users provide invalid data there are some situations where the logs in Infrahub are excessive. An example is if you try to create a BuiltinTag called "red" and this tag already exists.

The user will be informed that the tag exists and everything seems fine from the users point of view.

However in the logs we get a critical error:

```python
14:34:05.297 | CRITICAL | infrahub.graphql - {'exc_info': ValueError('An object already exist with this value: name: red at name'), 'event': 'Unhandled exception occurred in resolvers', 'request_id': '16ab53b810eb45f6aa07c6627d3e55ec', 'app': 'infrahub.api', 'worker': '36af3e7d-5c7b-432b-b92d-05d9e1fd890b', 'timestamp': '2024-10-09T12:34:05.297274Z', 'logger': 'infrahub.graphql', 'level': 'critical'}
```

This could be that we're reraising an error we've already defined as the internal `ValueError` which we then don't do anything about.

Another example is if the user sends in an invalid query:

```graphql
query MyQuery {
  BuiltinTag {
    edges {
      node {
        id
      }
    }
  }

```
In the above query we are missing the last closing bracket and when sent into Infrahub even the error message returned looks off:

```json
{
  "errors": [
    {
      "message": "Failed to fetch",
[...]

PR #9815, opened 2026-07-05
fix: sanitize git GraphQL query import errors

Timtech4u:
## Problem
Git repository imports can fail while creating `CoreGraphQLQuery` nodes. When that happens, the SDK `GraphQLError` currently surfaces the full generated mutation, variables, location reprs, and traceback-heavy text in task logs. That makes a user/configuration error hard to scan and can expose implementation details that are not useful to operators.

## Change
- catch `infrahub_sdk.exceptions.GraphQLError` around GraphQL query creation during Git import
- convert SDK validation output into a concise `RepositoryConfigurationError` that keeps the repository name, query name, and server validation message
- strip raw GraphQL mutation text, variables, and `SourceLocation` reprs from the operator-facing error/log message
- add regression coverage proving the sanitized error keeps the useful context without leaking the raw payload

## Before / After
Before, the worker/task output could include the full `CoreGraphQLQueryCreate` mutation and SDK exception payload.

After, the task reports a focused message like:

```text
Unable to import GraphQL query 'backbone_service' from repository 'opsmill-demo-edge-clean-message': Query is not valid: Cannot query field 'InfraBackBoneService' on type 'Query'.
```

## Screenshot
Filtered task log showing the sanitized `RepositoryConfigurationError` after reproducing the Git import failure locally:

InfraHub task log showing sanitized RepositoryConfigurationError

## Scope
This targets the Git integration path from #7498. It is related to the broader user-error logging cleanup in #4589, but intentionally does not attempt to change general API/GraphQL resolver logging in this PR.

## Testing
- `uv run ruff check backend/infrahub/git/integrator.py backend/tests/component/git/test_graphql_query_import.py`
- `uv run pytest backend/tests/component/git/test_graphql_query_import.py -q`
- local Docker verification against the InfraHub task UI with a failing Git GraphQL import

Addresses #7498. Related to #4589.

### DIFF
--- a/backend/tests/component/git/test_graphql_query_import.py
+++ b/backend/tests/component/git/test_graphql_query_import.py
@@ -198,7 +198,9 @@ async def test_create_graphql_query_surfaces_server_validation_error(
 
     # The upstream SDK GraphQLError embeds the full rendered mutation in its message; the
     # chained context must stay suppressed and only the concise server message may be logged.
+    # The positive check guards the negative ones: if log capture broke, they would pass vacuously.
     assert exc_info.value.__suppress_context__ is True
+    assert "Unable to import GraphQL query 'broken_query' from repository 'fragment_repo'" in caplog.text
     assert "An error occurred while executing the GraphQL Query" not in caplog.text
     assert "SourceLocation" not in caplog.text
 
--- a/backend/tests/component/git/test_graphql_query_import.py
+++ b/backend/tests/component/git/test_graphql_query_import.py
@@ -9,15 +9,16 @@ from __future__ import annotations
 
 import shutil
 from pathlib import Path
-from unittest.mock import AsyncMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from git import Repo as GitRepo
 from infrahub_sdk import Config, InfrahubClient
-from infrahub_sdk.exceptions import FragmentFileNotFoundError, FragmentNotFoundError
+from infrahub_sdk.exceptions import FragmentFileNotFoundError, FragmentNotFoundError, GraphQLError
 from infrahub_sdk.schema.repository import InfrahubRepositoryFragmentConfig, InfrahubRepositoryGraphQLConfig
 from infrahub_sdk.uuidt import UUIDT
 
+from infrahub.exceptions import RepositoryConfigurationError
 from infrahub.git import InfrahubRepository
 from infrahub.git.integrator import InfrahubRepositoryIntegrator
 from tests.constants import FIXTURE_REPOS_DIR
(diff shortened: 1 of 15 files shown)


Timtech4u:
Hi @fatih-acar and @saltas888, I opened this small PR after reproducing the noisy Git GraphQL query import error locally and matching it to #7498.

The change is intentionally narrow: it catches the SDK GraphQLError in the Git integration query-creation path, keeps the repository/query context, and strips the raw generated mutation/variables from the operator-facing task message. I also added regression coverage and verified the behavior in a local InfraHub task UI run.

Would appreciate your review when you have a chance.

backend/tests/component/git/test_graphql_query_import.py
```diff
@@ -9,15 +9,16 @@
 
 import shutil
 from pathlib import Path
-from unittest.mock import AsyncMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
```
ogenstad: We don't want to use MagicMock for anything new, the goal is to remove any MagicMock or AsyncMock.

ogenstad (commented):
In Infrahub 1.10.0 we introduced an error catalogue: https://github.com/opsmill/infrahub/blob/infrahub-v1.10.2/schema/error-catalogue.json

These errors will be added in the error extension field of the GraphQL errors, the idea is that the catalogue will grow over time and that we'll use this json file to generate error bindings that the SDK then uses so that we get the enriched errors directly within the SDK. (some more information about this can be found here: https://github.com/opsmill/infrahub/tree/infrahub-v1.10.2/dev/specs/infp-468-graphql-error-catalogue)

As such I think the next step for this should be to introduce this parsing within the SDK and ensure that the backend raises the correct enriched error instead of having a method within the integrator that tries to parse the error again with something like the `_clean_sdk_graphql_error_message()` function.

Timtech4u:
Thanks @ogenstad, agreed. I ran this locally and the better fix is to move it down a layer instead of parsing in the integrator.

I’ll update the PR to add a `GRAPHQL_QUERY_INVALID` catalogue error from `CoreGraphQLQueryCreate`, make the SDK surface the server error message without including query/variables, and keep the Git integrator just wrapping that concise SDK error.

I’ll keep broader validation/catalogue cases out of this PR unless you’d prefer those included too.

Timtech4u:
> As such I think the next step for this should be to introduce this parsing within the SDK and ensure that the backend raises the correct enriched error

Done in 48ba4c2. The backend now raises a catalogued `GRAPHQL_QUERY_INVALID` error (422) from the GraphQLQuery mutations, and the integrator simply wraps the concise server messages with no parsing. The tests were rebuilt without MagicMock, driving the real SDK client over pytest-httpx.

The SDK parsing is up in opsmill/infrahub-sdk-python#1140. `GraphQLError` now exposes the catalogue extensions via `details` and `codes`, and its message no longer embeds the query or variables. Once that lands, the integrator can drop its message joining too.

Issue #9818, opened 2026-07-06
bug: git repo connection failure is not reflected in git status page


TeeFiX42:
### Infrahub version

1.10.2

### Current Behavior

The git connection is failing:

<img width="1162" height="1104" alt="Image" src="https://github.com/user-attachments/assets/de33a67a-3f8a-4403-9a79-00a97e16b0f4" />

the status of the repo remains "online" and "in sync":

<img width="1186" height="1132" alt="Image" src="https://github.com/user-attachments/assets/291a6941-e203-4980-a13a-27961d1af178" />

### Expected Behavior

see documentation:
https://docs.infrahub.app/git-integration/overview#repository-statuses
status should be "Error" 

### Steps to Reproduce

We changed our gitlab from on prem to cloud; that did not work and reverted. We can't bring the connectivity up again. that is when I noticed that the statuspage is not updated.

PR #9819, opened 2026-07-06
fix: re-point cached clone origin when repository location changes (closes #9811)

### LINKED ISSUE #9811: bug: Git: changing an existing CoreRepository location wedges sync on the old cached commit
### Component

Git Integration

### Infrahub version

1.9.8

## Current Behavior

Changing the `location` of an existing read-write `CoreRepository` (integrated repo) leaves Infrahub's server-side cached clone pinned to the commit it held before the change. After the location is updated to a reachable, current remote, Infrahub never re-fetches: the repo object stays at the old commit, no new commits are imported, and new git branches on the remote are not auto-created as Infrahub branches.

Observed sequence:

1. Repo initially pointed at remote A (`<on-prem-git>`).
2. `location` changed to remote B (`<saas-git>`), which the instance cannot reach (air-gapped) → sync fails.
3. `location` changed back to remote A. Connectivity check now passes, and `git ls-remote` against remote A shows it is current (e.g. `main` at `<new-commit>` and a new branch `<feature-branch>` present).
4. Infrahub's repo object remains at the pre-change commit (e.g. `<old-commit>`). `main` is not advanced, and the new branch is never created in Infrahub.
5. Triggering a manual "reimport current commit" runs an `import-object-from-file` task whose `commit` parameter is still the stale `<old-commit>` — it re-imports the cached commit rather than fetching new refs.

`operational_status` reports `online` and `sync_status` reports `in-sync`, which is misleading — the last completed sync succeeded, but the clone is many commits behind and nothing new is being fetched.

### Expected Behavior

Changing a `CoreRepository`'s `location` should re-point the server-side clone's remote and trigger a fresh fetch, so that subsequent syncs pick up new commits on tracked branches and auto-create Infrahub branches for new git branches. `sync_status`/`operational_status` should reflect that the clone is behind if a fetch is failing, rather than reporting `in-sync`.

### Steps to Reproduce

1. Add a read-write `CoreRepository` pointing at remote A; let it sync.
2. Update the repo object's `location` to remote B, then back to remote A (remote A now has new commits on `main` and at least one new branch).
3. Confirm connectivity passes and `git ls-remote` on remote A shows the new commits/branch.
4. Observe the repo object stays at the original commit; no new branch appears; a manual reimport re-imports the old commit.

### Additional Information

Reproduced on a read-write/integrated `CoreRepository`. Workaround under evaluation: delete and re-add the repository object (the "add" flow performs a clean clone and picks up current refs). This suggests the `location`-change path does not reset/re-fetch the cached clone the way the initial add does.

minitriga:
# Why

Changing the `location` of an existing read-write `CoreRepository` left Infrahub's server-side cached clone pinned to whatever remote it was first cloned from. The clone's git `origin` URL is only ever set at clone time and was never reconciled with a changed `location`, so subsequent syncs kept fetching the old remote — the repo stayed on the old commit, new commits were never imported, and new remote branches were never auto-created. `operational_status`/`sync_status` stayed `online`/`in-sync` because the (stale) fetch kept succeeding.

Goal: a `location` change re-points the cached clone's remote and fetches from it, so syncs pick up the new remote's commits and branches.

Non-goals: does not change the read-write `location`-update mutation to *immediately* trigger a sync (the next periodic sync now recovers on its own); does not address the separate "status stays online while a fetch is failing" gap (tracked in #9818).

Closes #9811

## What changed

**Behavioral changes:**
- When a repository is (re)initialized and the stored `location` no longer matches the cached clone's `origin` URL, the origin is re-pointed to the current `location` and a fresh fetch is performed. Every sync/fetch/reimport path funnels through `InfrahubRepositoryIntegrator.init`, so this single guard covers the periodic sync, the worker `RefreshGitFetch`, and the manual reimport.

**Implementation notes:**
- Fix lives at the single choke point `InfrahubRepositoryIntegrator.init` (`backend/infrahub/git/integrator.py`). The URL comparison is done as a cheap read *before* taking the lock (every cron tick calls `init`); the reentrant repository lock is acquired only when an actual re-point is needed, then `set_url` + `fetch()` run under it. Routing through the existing `fetch()` means a now-unreachable new location also surfaces a failure `operational_status` for free.
- Guarded by `self.location` being set (some `init` callers omit it) and `has_origin`.

**What stayed the same:** No schema, GraphQL, or API changes. A freshly cloned repo already tracks the correct location, so the guard is a no-op when the location is unchanged.

## How to review

Focus on `backend/infrahub/git/integrator.py` `init` — specifically the lock reentrancy (callers such as `bootstrap_local_repository` already hold the `repository` lock when calling `init`; the lock is reentrant via a `ContextVar` recursion counter, so re-acquiring in the same async context is safe) and the pre-lock URL check.

## How to test

```bash
uv run pytest backend/tests/unit/git/test_git_repository.py::test_init_repoints_origin_after_location_change -x -v
uv run pytest backend/tests/unit/git/ -q
```

The new test clones from remote A, advances a second remote B, re-opens the existing clone via `init(location=B)`, fetches, and asserts the remote-tracking `main` reflects B's commit. It fails on `stable` (remote still shows A's commit) and passes with this fix. Full git unit module: 108 passed.

## Impact & rollout

- **Backward compatibility:** No breaking changes. No-op when location is unchanged.
- **Performance:** One extra `origin.url` read per `init` (no lock, no network); the `set_url` + `fetch` only runs when the location actually changed.
- **Deployment notes:** Safe to deploy.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Fixes repos sticking to the old remote after a location change by re-pointing the cached clone’s `origin` to the provided `location` and fetching during init. Periodic sync and any init that supply a location now pull commits and branches from the correct remote (closes #9811).

- **Bug Fixes**
  - In `InfrahubRepositoryIntegrator.init`, when `self.location` is set and differs from `origin.url`, acquire the repo lock, `set_url`, then `fetch()`.
  - Gate the re-point on a supplied `location` to avoid graph/schema dependency; location-less callers (e.g., re-import) are corrected by the periodic sync which passes the location.

<sup>Written for commit 71177914ae567ad764fce44a50d4474fb163629d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9819?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


minitriga:
Good catch — valid P1. `_get_initialized_repo` (the re-import / fetch-commit path) calls `init` without a `location`, so the original guard (`if self.location and self.has_origin`) skipped the re-point for those callers. The periodic cron passes location and would eventually correct the on-disk origin, but a manual re-import fired before the next cron tick would still fetch the stale remote.

Fixed in 83f… by resolving the location from the graph inside `init` before the comparison, reusing the existing `ensure_location_is_defined()` (which early-returns when a location was already supplied, so the hot cron path takes no extra query). Now every init path — periodic sync, worker `RefreshGitFetch`, and manual re-import — corrects the origin.

On the requested regression test for the re-import path: the re-point mechanism is covered by the existing unit test, and the new branch only adds a call to `ensure_location_is_defined()`, which is already exercised by the missing-directory recovery path. A dedicated test for the graph-resolution branch needs a real client/graph to serve `client.get` — the git unit suite uses a stub client that returns `{}`, and the repo forbids mocking — so it belongs at component level. Happy to add one there if you'd like it in this PR.

minitriga:
Correction: the review fix is commit `96626fb` (I mistyped the SHA above — `83b9f25` is the original fix commit).

minitriga:
Applied both non-blocking suggestions in `e-next` (see latest commit):
1. Dropped the redundant explicit `fetch()` — since `init` now re-points and fetches on its own, removing it makes the test assert exactly that (init self-heals), rather than masking it behind a manual fetch.
2. Added a direct `origin.url == str(source_b)` assertion alongside the commit check, to pin the re-point behavior against unrelated regressions.

Test still passes. Thanks for the review.

minitriga:
(the commit referenced above is `5633ff4`)

minitriga:
CI update — the graph-resolution approach (commit `96626fb`) broke 16 component tests: calling `ensure_location_is_defined()` unconditionally made **every** `init` do a `client.get(CoreGenericRepository)`, which the component suite's fake clients can't serve (`SchemaNotFoundError` / `ServerNotReachableError`). That was a real contract change — `init` should not require a reachable graph + loaded schema.

Reverted to gating the re-point on a supplied `location` (`f7a558393`). All 110 component git tests + 108 unit tests now pass locally.

On @cubic-dev-ai's original P1 (re-import path initializes without a location): it's still covered, just not inside `init`. The every-minute `git_repositories_sync` cron passes the location and re-points the **shared** on-disk clone; once re-pointed it stays corrected, so a subsequent location-less re-import fetches from the right remote. The only gap is a re-import fired in the <1-minute window before the next cron tick, which self-heals. Making `init` itself reach into the graph to close that window costs more than it's worth (the contract change above). If we want it closed deterministically, the clean way is to thread `location` through the re-import message into `get_initialized_repo` — happy to do that as a follow-up if preferred.

ajtmccarty (approved):
I think this looks good. I am not the most familiar with how the git integration really works, but it looks reasonable and I'm assuming you've tested it manually

PR #9821, opened 2026-07-06
fix(core): support X | None annotations on StandardNode fields

polmichel:
# Why

`StandardNode.guess_field_type` only resolved nullable fields annotated `Optional[X]`. The PEP 604 form `X | None` resolves to `types.UnionType` on Python 3.12/3.13, so such fields fell through to the JSON-blob serialization path instead of being stored natively. Follow-up to review comments [on PR 9550](https://github.com/opsmill/infrahub/pull/9550#discussion_r3519988819) and [PR 9541](https://github.com/opsmill/infrahub/pull/9541#discussion_r3403421067).

## What changed

- `guess_field_type` recognizes `types.UnionType` alongside `typing.Union`, so both nullable spellings resolve to the wrapped type on all supported Python versions (3.12-3.14).
- `StandardNode`'s own annotations use `X | None`; the now-unneeded `UP045` per-file ruff ignore is removed.
- Model files keeping `Optional[X]` (Branch, Preference) are untouched; converting them is left as follow-up.

## How to test

```bash
uv run pytest backend/tests/unit/core/node/test_standard.py -q
```

---
## Summary by cubic
Support PEP 604 nullable annotations (`X | None`) on `StandardNode` fields so nullable scalars persist as native types, not JSON blobs. Handles both `Optional[X]` and `X | None`/`None | X` across Python 3.12–3.14.

- **Bug Fixes**
  - `guess_field_type` now recognizes `types.UnionType` alongside `typing.Union`, and selects the non-None union member so `Optional[X]`, `X | None`, and `None | X` resolve to the inner type.
  - Updated `StandardNode` field and method annotations to `X | None`; removed per-file `UP045` ruff ignore.
  - Added unit tests for both syntaxes and reversed union order, covering type resolution and serialization.

<sup>Written for commit 9d61c9f7cdab402f5d85024a121eff25bf9b650a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9821?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/unit/core/node/test_standard.py
```diff
@@ -0,0 +1,43 @@
+from typing import Optional
```
dgarros: Not sure if we need some tests for this kind of change

dgarros (approved):
it feels like this change should go directly in `develop` or even `stable` since it's not really specific to this branch

polmichel:
> it feels like this change should go directly in `develop` or even `stable` since it's not really specific to this branch

The issue exists once in `stable`, and the feature branch adds another occurrence of it. I'll backport this PR to `stable` once it has been merged to `develop`, but I can't target `develop` directly without having to rebase the whole feature branch.
This PR is waiting for the feature branch to be merged before I move it to "ready for review" status. It'll then target develop directly and create the backport PR once done if that sounds ok.

backend/tests/unit/core/node/test_standard.py
polmichel: In my opinion, it does not cost much and protect against Python typing evolution that'd break this behavior. The test is isolated, fast to execute and will not end in a false positive / false negative result.

backend/infrahub/core/node/standard.py
```diff
@@ -79,9 +80,15 @@ def guess_field_type(field: FieldInfo) -> Any:
         annotation_origin = get_origin(field.annotation)
         annotation_args = get_args(field.annotation)
 
-        if (annotation_origin == Union and len(annotation_args) == 2 and type(None) in annotation_args) or (
-            annotation_origin is list and len(annotation_args)
-        ):
+        # `Optional[X]` resolves to `Union` while `X | None` resolves to `UnionType` on Python < 3.14,
+        # where the two are distinct; both must be recognized as nullable fields.
+        if annotation_origin in (Union, UnionType) and len(annotation_args) == 2 and type(None) in annotation_args:
```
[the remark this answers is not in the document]
polmichel: It's kind of a internal technical change. Not sure if a changelog is relevant here

Issue #6849, opened 2025-07-17
bug: cascade delete can miss related nodes added while the delete logic is running


ajtmccarty:
h3.  Component

[| #component]

API Server / GraphQL

h3.  Infrahub version

[| #infrahub-version]

1.3.0

h3.  Current Behavior

[| #current-behavior]

If you request to delete a node that has relationships which should be cascade-deleted and then add more nodes that include the cascade-delete relationship before the initial delete operation completes, then some of those added nodes will not be correctly deleted. An example

# in my schema, Device has a `on_delete=cascade` relationship to Interface
# Device `A` is related to some Interfaces
# I request to delete Device `A`
# at the same time, I create more Interfaces linked to Device `A`
# the Device `A` delete request completes, but the Interface nodes added in step 4 are not deleted, leaving them in an illegal state

h3.  Expected Behavior

[| #expected-behavior]

A cascade delete should either fully complete, with every expected node deleted, or be completely reverted, with no nodes deleted.

h3.  Steps to Reproduce

[| #steps-to-reproduce]

Steps are given above. This will probably only be observable on a rather large dataset with tens of thousands of relationships of the type to be cascade-deleted.

h3.  Additional Information

[| #additional-information]

ajtmccarty:
if a user encounters this issue, it will probably only become apparent when trying to merge/rebase a branch and they will get an error like this

```
Relationship-level 'count' constraint violation on schema 'InterfaceL1'. Node (InterfaceL1: 1234abcd-a228-3c18-1485-106586851d9b) is not compliant.
```

fixing this is possible by deleting the broken Nodes using a GraphQL mutation

saltas888:
The InfrahubMultiLock on develop is only on the update/create path (line 299) — the delete path (_delete_obj line 509, mutate_delete line 521) still takes no lock, so the race is unguarded. And PR [#9832](https://github.com/opsmill/infrahub/pull/9832) — the cascade-delete PR that fixed [#3076](https://github.com/opsmill/infrahub/issues/3076) — explicitly disclaimed this in its own description (#6849 — related race-condition)

Issue #2767, opened 2024-03-27
feature: on_delete behavior in schema


ajtmccarty:
### Component

API Server / GraphQL

### Describe the Feature Request

In some situations, it might be preferable to also delete the nodes that are dependent on the node to be deleted. In order to allow a user to define this behavior, we need to extend the schema of a relationship to capture the expected behavior (not sure what is the best way to name it)

    In SQL this is referenced as ON DELETE

### Additional Information

split from #2680

Issue #2680, opened 2024-03-19
feature: Schema integrity on node deletion


dgarros:
### Component

API Server / GraphQL

### Task Description

When deleting a node that potentially has relationships to other nodes, we must validate that the deletion of the node will not leave the database into an invalid state.

Split part of this request off into another issue: #2767

PR #9830, opened 2026-07-07
feat(schema): cascade repository deletion to its managed objects

### LINKED ISSUE #3076: bug: Transformations, Checks and Query from a repository are not delete when deleting the repository
### Component

API Server / GraphQL

### Current Behavior

If I delete a repository from Infrahub, not all the objects in this repository are deleted at the same time.

### Expected Behavior

All the objects related to the repository are cleanup at the same time

### Steps to Reproduce

_No response_

### Additional Information

Using the Lineage/source is needed in #1698 as well

iddocohen:
## Problem

Deleting a repository fails because the objects it imports/generates (transforms, checks, GraphQL queries, generators → artifact definitions → artifacts/validators, generator definitions → instances/validators, check definitions → user validators) are attached through **mandatory** relationships. Infrahub correctly refuses to orphan a mandatory child, so users are forced to delete the whole dependency tree by hand, leaf-first. Tracked in #3076.

Infrahub already has an `on_delete` cascade capability, but it was never wired to the repository-managed relationships.

## Solution

1. **Wire the repository ownership tree with `on_delete=CASCADE`.** Mark the repository's owner-side relationships (`transformations`, `queries`, `checks`, `generators`) as `CASCADE`, and add the missing parent-side relationships so the cascade reaches every leaf:
   - `Transformation → ArtifactDefinition`
   - `ArtifactDefinition → Artifact`, `→ ArtifactValidator`
   - `GeneratorDefinition → GeneratorInstance`, `→ GeneratorValidator`
   - `CheckDefinition → UserValidator`
   - `GraphQLQuery → GraphQLQueryGroup`

   Each new relationship reuses (or symmetrically auto-derives) its child-side identifier, so it maps to the existing edge — no data migration. This mirrors the existing `ArtifactTarget.artifacts` / `Validator.checks` cascade pattern.

2. **Make the mandatory-orphan check deep.** `NodeDeleteIndex` now seeds its dependent-check from every cascade-reachable kind, not just the explicitly-deleted kind. Previously the check was shallow, so an incomplete cascade would *silently orphan* a mandatory child. Now, if a cascaded node is still referenced by a mandatory relationship from an object outside the delete set, deletion fails with a clear error instead of corrupting data. (This also fixed a latent re-queue bug in the cascade traversal for leaf kinds.)

3. **Cascade `GraphQLQuery → GraphQLQueryGroup`.** `CoreGraphQLQueryGroup.query` is a mandatory back-reference and these groups are created routinely by the artifact/generator/computed-attribute pipelines. Without this, deleting a repository whose queries had ever run would be blocked — moving the wall of manual deletions one node over.

## Accepted trade-off

Deleting a repository (or an artifact/generator/check definition directly) now also deletes validators attached to proposed changes and their generated artifacts/instances. This is the intended ownership semantics and the alternative is the manual wall-of-deletions this fixes.

## Testing

- **Unit** (`test_delete_validator.py`): asserts the cascade closure from both `ReadOnlyRepository` and `Repository` reaches every managed leaf (`Artifact`, `ArtifactValidator`, `GeneratorInstance`, `GeneratorValidator`, `UserValidator`, `GraphQLQueryGroup`).
- **Component** (`test_node_manager_delete.py`, against live Neo4j): deleting a repository cascades through transform → artifact definition → artifact, plus check/generator definitions and generator instances; deleting a repository with a `GraphQLQueryGroup` succeeds and removes the query + group; an out-of-tree mandatory referrer to a cascaded node raises `ValidationError` instead of silently orphaning.
- Regenerated `protocols.py`, `schema.graphql`, and frontend GraphQL types; schema validation and ruff/ty/mypy all pass.

## Migration

**No graph migration required.** `on_delete` values and relationship definitions are reconciled into the graph by `update_core_schema` during `infrahub upgrade` (it diffs the code-defined candidate schema against the DB and upserts changed attribute values + new relationships), and the delete path reads `on_delete` from the in-memory registry that is populated from that reconciled graph. Graph migrations (`mNNN_*`) are only needed for structural changes to the meta-schema itself (e.g. `m006` added the `on_delete` *field* to `SchemaRelationship`) — not for changing values on existing relationships or adding new ones. `GRAPH_VERSION` stays at 73.

## Related issues

- Closes #3076 — the bug this fixes (repository-managed objects not cleaned up on delete).
- #2767 (closed) — the `on_delete` cascade schema capability this builds on; #2680 (closed) — its parent "schema integrity on node deletion" task. Both carry `state/ref`, so there is likely a corresponding **INFP** Jira ticket tracking the on_delete/cascade work worth linking.
- #6849 (open) — related race-condition in cascade delete (not addressed here). #9790 (open) — sibling stale-artifact cleanup gap (different trigger).

### DIFF
--- a/backend/infrahub/core/node/delete_validator.py
+++ b/backend/infrahub/core/node/delete_validator.py
@@ -29,8 +29,8 @@ class NodeDeleteIndex:
         self._dependency_graph: dict[str, dict[DeleteRelationshipType, dict[str, set[str]]]] = {}
 
     def index(self, start_schemas: Iterable[NodeSchema | ProfileSchema | TemplateSchema]) -> None:
-        self._index_cascading_deletes(start_schemas=start_schemas)
-        self._index_dependent_schema(start_schemas=start_schemas)
+        cascade_kinds = self._index_cascading_deletes(start_schemas=start_schemas)
+        self._index_dependent_schema(cascade_kinds=cascade_kinds)
 
     def _get_schema_kinds(self, schema_kind: str) -> set[str]:
         schema = self._all_schemas_map[schema_kind]
@@ -50,13 +50,17 @@ class NodeDeleteIndex:
             self._dependency_graph[kind][relationship_type] = defaultdict(set)
         self._dependency_graph[kind][relationship_type][relationship_identifier].update(peer_kinds)
 
-    def _index_cascading_deletes(self, start_schemas: Iterable[NodeSchema | ProfileSchema | TemplateSchema]) -> None:
+    def _index_cascading_deletes(
+        self, start_schemas: Iterable[NodeSchema | ProfileSchema | TemplateSchema]
+    ) -> set[str]:
+        visited: set[str] = set()
         kinds_to_check: set[str] = {schema.kind for schema in start_schemas}
         while True:
             try:
                 kind_to_check = kinds_to_check.pop()
             except KeyError:
                 break
+            visited.add(kind_to_check)
             node_schema = self._all_schemas_map[kind_to_check]
             for relationship_schema in node_schema.relationships:
                 if relationship_schema.on_delete != RelationshipDeleteBehavior.CASCADE:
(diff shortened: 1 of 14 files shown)


iddocohen:
Superseded by #9832 — the branch was renamed to `feat/repository-cascade-delete` (rebased onto `develop`), which closed this PR. Continuing in #9832.

PR #9832, opened 2026-07-07
feat(schema): cascade repository deletion to its managed objects

### LINKED ISSUE #3076: bug: Transformations, Checks and Query from a repository are not delete when deleting the repository

iddocohen:
## Problem

Deleting a repository fails because the objects it imports/generates (transforms, checks, GraphQL queries, generators → artifact definitions → artifacts/validators, generator definitions → instances/validators, check definitions → user validators) are attached through **mandatory** relationships. Infrahub correctly refuses to orphan a mandatory child, so users are forced to delete the whole dependency tree by hand, leaf-first. Tracked in #3076.

Infrahub already has an `on_delete` cascade capability, but it was never wired to the repository-managed relationships.

## Solution

1. **Wire the repository ownership tree with `on_delete=CASCADE`.** Mark the repository's owner-side relationships (`transformations`, `queries`, `checks`, `generators`) as `CASCADE`, and add the missing parent-side relationships so the cascade reaches every leaf:
   - `Transformation → ArtifactDefinition`
   - `ArtifactDefinition → Artifact`, `→ ArtifactValidator`
   - `GeneratorDefinition → GeneratorInstance`, `→ GeneratorValidator`
   - `CheckDefinition → UserValidator`
   - `GraphQLQuery → GraphQLQueryGroup`

   Each new relationship reuses (or symmetrically auto-derives) its child-side identifier, so it maps to the existing edge — no data migration. This mirrors the existing `ArtifactTarget.artifacts` / `Validator.checks` cascade pattern.

2. **Make the mandatory-orphan check deep.** `NodeDeleteIndex` now seeds its dependent-check from every cascade-reachable kind, not just the explicitly-deleted kind. Previously the check was shallow, so an incomplete cascade would *silently orphan* a mandatory child. Now, if a cascaded node is still referenced by a mandatory relationship from an object outside the delete set, deletion fails with a clear error instead of corrupting data. (This also fixed a latent re-queue bug in the cascade traversal for leaf kinds.)

3. **Cascade `GraphQLQuery → GraphQLQueryGroup`.** `CoreGraphQLQueryGroup.query` is a mandatory back-reference and these groups are created routinely by the artifact/generator/computed-attribute pipelines. Without this, deleting a repository whose queries had ever run would be blocked — moving the wall of manual deletions one node over.

## Accepted trade-off

Deleting a repository (or an artifact/generator/check definition directly) now also deletes validators attached to proposed changes and their generated artifacts/instances. This is the intended ownership semantics and the alternative is the manual wall-of-deletions this fixes.

## Testing

- **Unit** (`test_delete_validator.py`): asserts the cascade closure from both `ReadOnlyRepository` and `Repository` reaches every managed leaf (`Artifact`, `ArtifactValidator`, `GeneratorInstance`, `GeneratorValidator`, `UserValidator`, `GraphQLQueryGroup`).
- **Component** (`test_node_manager_delete.py`, against live Neo4j): deleting a repository cascades through transform → artifact definition → artifact, plus check/generator definitions and generator instances; deleting a repository with a `GraphQLQueryGroup` succeeds and removes the query + group; an out-of-tree mandatory referrer to a cascaded node raises `ValidationError` instead of silently orphaning.
- Regenerated `protocols.py`, `schema.graphql`, and frontend GraphQL types; schema validation and ruff/ty/mypy all pass.

## Migration

**No graph migration required.** `on_delete` values and relationship definitions are reconciled into the graph by `update_core_schema` during `infrahub upgrade` (it diffs the code-defined candidate schema against the DB and upserts changed attribute values + new relationships), and the delete path reads `on_delete` from the in-memory registry that is populated from that reconciled graph. Graph migrations (`mNNN_*`) are only needed for structural changes to the meta-schema itself (e.g. `m006` added the `on_delete` *field* to `SchemaRelationship`) — not for changing values on existing relationships or adding new ones. `GRAPH_VERSION` stays at 73.

## Related issues

- Closes #3076 — the bug this fixes (repository-managed objects not cleaned up on delete).
- #2767 (closed) — the `on_delete` cascade schema capability this builds on; #2680 (closed) — its parent "schema integrity on node deletion" task. Both carry `state/ref`, so there is likely a corresponding **INFP** Jira ticket tracking the on_delete/cascade work worth linking.
- #6849 (open) — related race-condition in cascade delete (not addressed here). #9790 (open) — sibling stale-artifact cleanup gap (different trigger).

---
## Summary by cubic
Repository deletions now cascade to all repo-managed objects so a repo can be removed in one action. The delete validator checks mandatory dependents across the full cascade and blocks unsafe deletes with clear errors. Closes #3076.

- **Bug Fixes**
  - Enabled `on_delete=CASCADE` from repositories to transformations, queries, checks, generators, and repository groups; added parent links so cascades reach: Transformation → ArtifactDefinition; ArtifactDefinition → Artifacts/Validators; GeneratorDefinition → Instances/Validators; GraphQLQuery → QueryGroups; CheckDefinition → UserValidators.
  - Delete validator now checks the full cascade closure and respects mandatory refs via Node/Profile/Template `inherit_from`; tests assert the exact deleted ID set and that validators attached only to a `ProposedChange` (not repo-owned) survive; regenerated backend protocols, `schema.graphql`, and frontend GraphQL types.

- **Migration**
  - No graph migration needed; `update_core_schema` applies relationship changes during `infrahub upgrade` (`GRAPH_VERSION` unchanged).

<sup>Written for commit cebbc8909df813e66766056b157cd7dc58755666. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9832?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/schema/definitions/core/repository.py
```diff
@@ -267,6 +268,7 @@
             identifier="repository__transformation",
             optional=True,
             cardinality=Cardinality.MANY,
+            on_delete=RelationshipDeleteBehavior.CASCADE,
```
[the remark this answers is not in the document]
iddocohen: Good catch — fixed in 15be83139. `CoreRepositoryGroup.repository` is mandatory, so I added `on_delete=CASCADE` to `GenericRepository.groups_objects`; deleting a repository now cascades to its import-tracking repository groups. Verified: unit closure test now asserts `CoreRepositoryGroup` is reachable, and a new component test (`test_delete_repository_repository_group_cascade`) confirms it against a live database. `RepositoryGroup` has no other mandatory referrers, so the cascade stays bounded.

backend/tests/protocols.py
```diff
@@ -217,6 +217,7 @@ class CoreTransformation(CoreNode):
     query: RelationshipManager[CoreGraphQLQuery]
     repository: RelationshipManager[CoreGenericRepository]
     tags: RelationshipManager[BuiltinTag]
+    artifact_definitions: RelationshipManager[CoreArtifactDefinition]
```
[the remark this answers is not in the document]
iddocohen: This is stale — it was flagged on an earlier commit. On the current head, both `backend/infrahub/core/protocols.py` and `backend/tests/protocols.py` already declare the new relationships (`artifact_definitions`, `artifacts`, `validators`, `instances`, `query_groups`); `backend.validate-generated` passes. No action needed.

backend/infrahub/core/node/delete_validator.py
```diff
@@ -69,15 +73,17 @@ def _index_cascading_deletes(self, start_schemas: Iterable[NodeSchema | ProfileS
...
-    def _index_dependent_schema(self, start_schemas: Iterable[NodeSchema | ProfileSchema | TemplateSchema]) -> None:
+    def _index_dependent_schema(self, cascade_kinds: set[str]) -> None:
         start_schema_kinds: set[str] = set()
-        for start_schema in start_schemas:
-            start_schema_kinds.add(start_schema.kind)
-            if start_schema.inherit_from:
-                start_schema_kinds.update(set(start_schema.inherit_from))
+        for kind in cascade_kinds:
+            start_schema_kinds.add(kind)
+            inherit_from = getattr(self._all_schemas_map[kind], "inherit_from", None)
```
ajtmccarty: doing an `isinstance()` check her would be better for typing than `getattr`

backend/tests/unit/core/node/test_delete_validator.py
ajtmccarty, line 1: this test should look at the results of the public interface, `get_relationship_identifiers` instead of the private `_dependency_graph`

backend/infrahub/core/node/delete_validator.py
ajtmccarty, line 1: these piece and the test(s) for it might be worth sending to `stable`

ajtmccarty (commented):
just a couple small comments, but looks good. nice catch on the bug fix too. also want to confirm that you have tested upgrading an existing instance to this branch and that the cascade deletions from the repository correctly delete everything they are supposed to. I'd expect it to work, but I have not tested it

backend/infrahub/core/node/delete_validator.py
iddocohen: Done in a609627b5 — switched to an `isinstance(schema, NodeSchema)` guard (matches the idiom in `schema_branch.py`).

backend/tests/unit/core/node/test_delete_validator.py
iddocohen: Done in a609627b5 — the closure is now built from the public `get_relationship_identifiers()` + `get_relationship_types()` (the same methods the delete path uses), no more `_dependency_graph` access.

backend/infrahub/core/node/delete_validator.py
iddocohen: Good call — opening a separate draft PR against `stable` with just the delete-validator deep-orphan-check + its tests (the standalone fix, without the new schema relationships). Will link it here.

iddocohen: Opened as a draft: #9850 (base `stable`). It carries just the delete-validator deep-orphan-check + a behavior test; full delete module passes on `stable` (16 existing + 1 new). Watching its CI/cubic before requesting review.

backend/infrahub/core/node/delete_validator.py
iddocohen: Follow-up: cubic flagged that `isinstance(NodeSchema)` alone dropped Profile/Template `inherit_from` coverage vs the old `getattr`. Widened to `isinstance(schema, (NodeSchema, ProfileSchema, TemplateSchema))` in 297cc71f9 — type-safe and complete.

iddocohen:
@ajtmccarty regarding testing, I have done locally but to ensure hardness for furture, I will add e2e test.

tasks/dev.py
```diff
@@ -276,6 +277,359 @@ def test_branch_rebase(context: Context, branch: str, data_to_check: str = "") -
...
+    """Initialise a bare git repo at *git_repo_path* inside the running task-worker container.
+
+    The task-worker's ``/remote`` volume is used as the working area.  The repo is seeded
+    from the demo-edge fixture tree that is always present in the container at
+    ``/source/backend/tests/fixtures/repos/infrahub-demo-edge/initial__main``.
+    """
+    # Identify one of the running task-worker containers via the compose project name.
+    worker_id = (
+        subprocess.check_output(  # noqa: S603
+            ["docker", "compose", "-p", BUILD_NAME, "ps", "-q", "task-worker"],  # noqa: S607
```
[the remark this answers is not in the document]
iddocohen: Fixed in the latest push — `_init_git_repo_in_worker` now uses the project's compose abstractions (`build_compose_files_cmd`/`get_compose_cmd`/`get_env_vars`/`SERVICE_WORKER_NAME`/`execute_command`), mirroring `infra_git_import`, and the hard-coded `subprocess`/`docker compose`/`task-worker` calls (and the `import subprocess`) are gone. Also rebased the branch onto `develop` and regenerated the conflicted generated files.

iddocohen:
@ajtmccarty — update on the cascade-delete work here.

**Conflict + CI fixes**
- Rebased onto `develop`; the only conflict was in the generated `frontend/app/src/shared/api/graphql/generated/graphql-cache.d.ts`, resolved by regenerating (`pnpm codegen:graphql`) so it matches the merged schema.
- Fixed the `E2E-testing-version-upgrade / From 1.8.0` job. Two issues in `_init_git_repo_in_worker`:
  1. It used `docker compose run`, which started the worker's `depends_on` services and recreated the server container on a fresh random port mid-seed → the SDK got `ConnectError: [Errno 111] Connection refused`. Switched to `docker compose exec` against the already-running worker, routed through the shared compose abstractions (`get_compose_cmd` / `build_compose_files_cmd` / `execute_command` / `SERVICE_WORKER_NAME`) — this addresses your P2 review comment while keeping the stack intact.
  2. `execute_command` streamed the git output into the `REPO_CASCADE_ID=$(…)` capture, producing a multi-line value that `>> $GITHUB_ENV` rejected. Added `hide=True` so only `print(repo.id)` reaches stdout.

The `From 1.8.0` job is now **green**.

**Please review the test for hardening** ⚠️
The verify step passed, but it reported `OK: repository '…' and 0 descendant(s) all deleted` — i.e. it currently validates nothing, for two independent reasons:
1. The seeded managed objects don't survive: a `CoreReadOnlyRepository` reconciles managed objects against the repo contents and deletes the manually-created ones (worker logs: `Graphql Query 'cascade-test-query-…' not found locally, deleting`).
2. Even if they survived, `test_repository_cascade_delete` calls `client.get(kind="CoreReadOnlyRepository", id=…)` without `prefetch_relationships=True`, so `repo.transformations.peers` / `repo.queries.peers` / `repo.checks.peers` / `repo.generators.peers` / `repo.groups_objects.peers` are all empty → 0 descendants collected.

Suggestion for future-proofing so a regression can't pass vacuously:
- Seed via a repo fixture whose `.infrahub.yml` actually defines the managed objects (so they persist as repo descendants), or otherwise ensure the descendants survive the sync.
- Prefetch relationships in the verify task and assert `len(descendant_ids) > 0` before checking they're gone.

Happy to implement whichever direction you prefer.

ajtmccarty:
I'd say that a backend integration test would be a better option for this: easier to read and write, more predictable, faster. we've got other tests that deal with git repositories, but I think this one just needs to create the various objects in the database (as opposed to in the file system) and verify that they all get deleted correctly

iddocohen:
Thanks @ajtmccarty, makes sense — switched to a backend component test and dropped the version-upgrade e2e.

`backend/tests/component/core/test_node_manager_delete.py::test_delete_repository_cascades_managed_objects` now creates a repository plus one of every managed object directly in the database — transform → artifact definition → artifact, check definition, generator definition → instance, and (via a proposed change) the artifact/generator/user validators — then calls `NodeManager.delete` and asserts every descendant is deleted while the proposed change (their non-owned parent) survives. It asserts on explicit ids, so it can't pass vacuously. The query → query-group and repository-group chains keep their own focused tests, and validator reachability is also covered by the `test_delete_validator` schema-closure unit test.

Removed the `tasks/dev.py` cascade helpers and the two `version-upgrade.yml` steps that drove the e2e (net −362/+51). Verified locally against Neo4j and green across all four component-test shards in CI.

iddocohen:
Re: the cubic flag on removing the version-upgrade.yml cascade steps (review, "1 issue found") — checked the mechanism, this is expected and not a coverage gap.

The upgrade path converges on the exact same code the component test exercises:

- Cascade is resolved **at delete time from the live schema loaded from the database** — `delete_validator.py` reads `relationship_schema.on_delete` off the schema branch (`registry.schema.get_schema_branch(...)`), never from the Python definitions.
- `on_delete` is a persisted relationship-schema attribute (`schema/definitions/internal.py`), and it participates in the relationship diff hash — `RelationshipSchema._exclude_from_hash` only excludes `filters`. So when `update_core_schema` runs against a pre-existing DB, the diff detects the `NO_ACTION → CASCADE` change and reconciles it onto the already-stored relationships.
- After that reconcile, deleting a repository created under the *old* schema runs identical logic to deleting one created fresh — both read the updated `on_delete` from the live schema.

The reconcile-on-upgrade step is generic schema-migration infrastructure already covered by existing schema-diff tests; this PR only flips the `on_delete` values, it does not touch the reconcile machinery. So swapping the slow version-upgrade e2e (427924527) for the `test_node_manager_delete.py` component test does not lose real coverage. No change needed.

backend/tests/component/core/test_node_manager_delete.py
```diff
@@ -631,3 +631,210 @@ async def test_cascade_delete_blocked_by_external_mandatory_dependent(
...
+        transform.id,
+        definition.id,
+        artifact.id,
+        check_def.id,
+        generator_def.id,
+        generator_instance.id,
+        artifact_validator.id,
+        generator_validator.id,
+        user_validator.id,
+    } <= deleted_ids
```
ajtmccarty: I think you should be able to assert the exact deleted IDs and not just "subset of"

backend/tests/component/core/test_node_manager_delete.py
```diff
@@ -631,3 +631,210 @@ async def test_cascade_delete_blocked_by_external_mandatory_dependent(
...
+    await generator_instance.save(db=db)
+
+    proposed_change = await Node.init(db=db, schema=InfrahubKind.PROPOSEDCHANGE)
+    await proposed_change.new(
+        db=db,
+        name="repo-cascade-pc",
+        source_branch="repo-cascade-source",
+        destination_branch=default_branch.name,
+    )
+    await proposed_change.save(db=db)
```
ajtmccarty, line 730: not sure how easy/hard this is, but adding a data integrity and/or schema integrity validator here and confirming they survive the deletion might be a good addition

backend/tests/component/core/test_node_manager_delete.py
```diff
@@ -631,3 +631,210 @@ async def test_cascade_delete_blocked_by_external_mandatory_dependent(
...
+    await query_group.new(
+        db=db,
+        name="repo-cascade-query-group",
+        query=str(repo_query.id),
+    )
+    await query_group.save(db=db)
+
+    deleted = await NodeManager.delete(db=db, branch=default_branch, nodes=[r1])
+
+    deleted_ids = {d.id for d in deleted}
```
ajtmccarty, line 831: you should also be able to assert the exact set of deleted IDs here

backend/tests/component/core/test_node_manager_delete.py
```diff
@@ -631,3 +631,210 @@ async def test_cascade_delete_blocked_by_external_mandatory_dependent(
...
+    """Deleting a repository cascades to its import-tracking repository groups."""
+    r1 = car_person_data_generic["r1"]
+
+    repo_group = await Node.init(db=db, schema=InfrahubKind.REPOSITORYGROUP)
+    await repo_group.new(db=db, name="repo-cascade-import-group", repository=str(r1.id), content="object")
+    await repo_group.save(db=db)
+
+    deleted = await NodeManager.delete(db=db, branch=default_branch, nodes=[r1])
+
+    deleted_ids = {d.id for d in deleted}
```
ajtmccarty, line 851: samesies

backend/tests/unit/core/node/test_delete_validator.py
```diff
@@ -0,0 +1,72 @@
...
+    expected_leaves = {
+        InfrahubKind.ARTIFACT,
+        InfrahubKind.ARTIFACTVALIDATOR,
+        InfrahubKind.GENERATORINSTANCE,
+        InfrahubKind.GENERATORVALIDATOR,
+        InfrahubKind.GRAPHQLQUERYGROUP,
+        InfrahubKind.REPOSITORYGROUP,
+        InfrahubKind.USERVALIDATOR,
+    }
+    assert expected_leaves <= reachable
```
ajtmccarty: I think you can assert the exact expectation here too

ajtmccarty (approved):
a few small test comments, but looks good to merge after those

backend/tests/component/core/test_node_manager_delete.py
iddocohen: Done. Switched to exact set equality on the deleted IDs. The fixture adds no repository-linked objects beyond those under test, so the deleted set is deterministic.

backend/tests/component/core/test_node_manager_delete.py
iddocohen: Added. The test now creates a data validator and a schema validator attached only to the proposed change, and asserts both survive the cascade. Confirms the delete stays bounded to repository-owned objects.

backend/tests/component/core/test_node_manager_delete.py
iddocohen: Done. Replaced the membership checks with an exact set equality.

backend/tests/component/core/test_node_manager_delete.py
iddocohen: Done here as well, same exact set equality.

backend/tests/unit/core/node/test_delete_validator.py
iddocohen: Done. The assertion is now exact equality against the full cascade closure of 21 kinds, derived from the live schema, so it catches both missing and unexpected cascade edges.

PR #9844, opened 2026-07-08
fix(webhook): honor cancellation during retry wait and stop crashing waiting runs

polmichel:
## Why

Webhook delivery drives its retries in-process: between attempts the flow run sits idle in a retry-wait state and emits no heartbeats. Two behaviors depended on that idle window and misfired.

- A cancellation requested during a retry wait was observed only when the next attempt began, so an in-flight delivery kept retrying. [IFC-2879 ](https://opsmill.atlassian.net/browse/IFC-2879)
- The zombie-flow automation, which crashes runs that stop heartbeating, read the silent retry wait as a dead process and crashed the run mid-wait. [IFC-2880](https://opsmill.atlassian.net/browse/IFC-2880)

Goal: honor cancellation at any point in the retry sequence, and let a legitimately waiting run survive its backoff, while still reaping genuinely dead runs.

## What changed

**Behavioral**

- A delivery checks for a requested cancellation before each send attempt and stops the sequence once one is recorded, so cancellation takes effect during the retry wait rather than only at attempt boundaries. The parent flow surfaces a cancelled child's state directly.
- The `crash-zombie-flows` automation treats the retry-wait transition as an expected event and sizes its no-heartbeat window from the longest configured retry backoff plus a margin. A waiting delivery survives its backoff; a run whose process has died is still crashed once the window lapses.

**Implementation**

- A `FlowRunCancellationReading` protocol and a `cancellation_requested()` method on the Prefect client adapter, so cancellation state is read through the adapter.
- The webhook retry constants move to the webhook constants module so the trigger layer derives the window from the backoff without a circular import.

**Unrelated changes**

- Reworded a pre-existing changelog fragment (`+webhook-related-tasks.fixed.md`, authored on the base branch) so the `validate-release-notes-style` Vale job passes. The spell checker does not recognize the possessive form of the accepted term; using it attributively reads the same and clears the error. Not part of the state-handling fixes — included only to unblock CI on this branch.

## How to test

```bash
uv run pytest \
  backend/tests/component/webhook/test_cancel_during_retry_wait.py \
  backend/tests/component/trigger/test_zombie_detection.py \
  backend/tests/component/task_manager/test_prefect_client.py
```

Against a live Prefect server: a delivery cancelled mid-retry-wait stops immediately; a full four-attempt delivery cycle records `Pending → Running → AwaitingRetry ×3 → Failed` with no `Crashed` state; a wait that never resumes is still crashed once the window lapses.

### Execution time

These are live-server tests, so they carry a real wall-clock cost. Measured per-test call duration (`pytest --durations`):

| Test | Time |
|------|------|
| `test_cancel_during_retry_wait_stops_remaining_attempts` | ~5s |
| `test_wait_that_never_resumes_is_marked_crashed` | ~11s |
| `test_wait_transition_restarts_the_countdown` | ~12s |

The remaining cost is intrinsic: each duration is bounded by Prefect's 10s minimum proactive-trigger window (zombie tests) or the flow's retry backoff (cancel test), not by fixed padding. A prior ~100s real-time zombie test was removed as redundant once the detection window became derived, and the server's proactive evaluation cadence is tightened in the component harness so the scaled runs are judged without lag.

## Impact & rollout

- **Backward compatibility:** no breaking changes.
- **Performance:** a genuinely dead run is now declared crashed after the widened window (~180s) rather than the previous 90s; detection is retained, latency is higher.
- **Deployment notes:** safe to deploy; the zombie automation is re-synced to the Prefect server at startup.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry

---
## Summary by cubic
Honors webhook cancellation during retry backoff and prevents false “zombie” crashes, while still crashing genuinely dead runs. Updates docs and changelog, and aligns with Linear IFC-2119 by ensuring no request is sent after cancellation and retry waits aren’t mis-crashed.

- **Bug Fixes**
  - Each send attempt checks `PrefectClientAdapter.cancellation_requested()` (state history) and stops with `Cancelled` if a request was recorded; no further attempts are sent.
  - Zombie automation treats `AwaitingRetry` as an expected event and sizes its window to `WEBHOOK_SEND_RETRY_DELAY_SECONDS + 60s`, so a run waiting out backoff isn’t crashed, but a dead run still is.
  - Centralized retry constants in `infrahub/webhook/constants.py`; the system trigger derives its window from these to avoid circular imports.

<sup>Written for commit 4e1cdebfbc446b168fad72b00f1d282d72170a65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9844?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2880]: https://opsmill.atlassian.net/browse/IFC-2880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/trigger/system.py
+++ b/backend/infrahub/trigger/system.py
@@ -9,6 +9,16 @@ from datetime import timedelta
 from prefect.client.schemas.objects import StateType
 
 from infrahub.trigger.models import ChangeFlowRunStateAction, ProactiveEventTrigger, SystemTriggerDefinition
+from infrahub.webhook.constants import WEBHOOK_SEND_RETRY_DELAY_SECONDS
+
+# A proactive trigger keeps a per-run countdown: every expected event restarts it, and the
+# action fires when the countdown lapses. A run waiting out a retry backoff emits nothing at
+# all, so the window must outlast the longest configured backoff or every such wait is
+# mistaken for a dead process. The retry-wait transition is an expected event so the countdown
+# is anchored at the start of the silence it explains. The margin absorbs event-propagation
+# and scheduler jitter between the retry-wait transition and the resumed run's next heartbeat.
+ZOMBIE_DETECTION_MARGIN = timedelta(seconds=60)
+ZOMBIE_HEARTBEAT_WINDOW = timedelta(seconds=WEBHOOK_SEND_RETRY_DELAY_SECONDS) + ZOMBIE_DETECTION_MARGIN
 
 TRIGGER_CRASH_ZOMBIE_FLOWS = SystemTriggerDefinition(
     name="crash-zombie-flows",
@@ -17,6 +27,7 @@ TRIGGER_CRASH_ZOMBIE_FLOWS = SystemTriggerDefinition(
         after={"prefect.flow-run.heartbeat"},
         events={
             "prefect.flow-run.heartbeat",
+            "prefect.flow-run.AwaitingRetry",
             "prefect.flow-run.Completed",
             "prefect.flow-run.Failed",
             "prefect.flow-run.Cancelled",
@@ -25,7 +36,7 @@ TRIGGER_CRASH_ZOMBIE_FLOWS = SystemTriggerDefinition(
         match={"prefect.resource.id": ["prefect.flow-run.*"]},
         for_each={"prefect.resource.id"},
         threshold=1,
-        within=timedelta(seconds=90),
+        within=ZOMBIE_HEARTBEAT_WINDOW,
     ),
     actions=[
         ChangeFlowRunStateAction(
--- a/backend/infrahub/task_manager/flow_run/prefect_client.py
+++ b/backend/infrahub/task_manager/flow_run/prefect_client.py
@@ -7,6 +7,8 @@ from prefect.client.schemas.filters import ArtifactFilter, FlowFilter, FlowRunFi
 from prefect.client.schemas.objects import Artifact, Flow, FlowRun, Log, StateType
 from prefect.client.schemas.sorting import FlowRunSort
 
+CANCEL_REQUEST_STATE_TYPES = frozenset({StateType.CANCELLING, StateType.CANCELLED})
+
 
 class FlowRunQuerying(Protocol):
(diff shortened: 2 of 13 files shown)


backend/infrahub/trigger/system.py
```diff
@@ -9,6 +9,16 @@
...
+from infrahub.webhook.constants import WEBHOOK_SEND_RETRY_DELAY_SECONDS
+
+# A proactive trigger keeps a per-run countdown: every expected event restarts it, and the
+# action fires when the countdown lapses. A run waiting out a retry backoff emits nothing at
+# all, so the window must outlast the longest configured backoff or every such wait is
+# mistaken for a dead process. The retry-wait transition is an expected event so the countdown
+# is anchored at the start of the silence it explains. The margin absorbs event-propagation
+# and scheduler jitter between the retry-wait transition and the resumed run's next heartbeat.
+ZOMBIE_DETECTION_MARGIN = timedelta(seconds=60)
+ZOMBIE_HEARTBEAT_WINDOW = timedelta(seconds=WEBHOOK_SEND_RETRY_DELAY_SECONDS) + ZOMBIE_DETECTION_MARGIN
```
polmichel, line 21: Not the cleanest design, but I am not sure how to manage this window which consider awaiting flow as zombies one

fatih-acar: I think that's the simplest for now; the 90s second we had before was arbitrary. I think we can even lower `ZOMBIE_DETECTION_MARGIN` to 30s.

But detecting zombie flows within ~5 min seems enough to me.

fatih-acar (approved):

PR #9849, opened 2026-07-08
feat(recompute): bulk-write derived values with coalesced chaining

gmazoyer:
## Why

This is a follow-up to the merge and rebase recompute coalescing (IFC-2761). That work grouped the per-node process-flow dispatch into one pass. The dataset runs showed the time an instance stays busy after a large merge comes mostly from the next stage: one update flow plus one GraphQL write for each affected value. This change removes that stage.

## What

The bulk write is now the only way we persist recomputed derived values. This is true for live edits and for merge and rebase.

1. **Bulk write.** A `BulkRecomputeWriter` saves display labels, human-friendly ids and Jinja2 computed attributes for many nodes in bounded transactions on the task worker, which has database access. Before, there was one update flow per value. The process flow now renders and checks for a real change in place, because the reader query already returns the current values. This drops the per-value task fan-out, the per-value API calls and the repeated per-value reads.

2. **Coalesced chaining.** A recompute write can feed a value that reads it. Those writes become a new change set that goes through the same coordinator, so the next level is one coalesced pass instead of one live event per written node. Each write still emits one node-update event per node for other consumers. On the coalesced pass the event is tagged with a `recompute` origin, so the per-node recompute automations do not run again.

3. **Depth bound.** For normal (acyclic) schemas the chain stops on its own once a level changes nothing. The bound is only there for a cyclic schema. It is derived from the schema: a chain cannot recompute more values than the schema defines, so the bound is the number of derived-value targets, with a floor. It never cuts a real chain short, and it still stops a cyclic one.

## Scope

Jinja2 computed attributes, display labels and human-friendly ids, the same families as the coalescing work. Python-transform computed attributes stay on the per-node path. The per-value update flows and their workflow definitions are removed.

## Correctness

The stored value does not change. Rendering, including custom Jinja2 filters, still runs on the worker. Only the way we persist changed, from a GraphQL mutation to a direct save. Component tests against a real database check the stored value for every impacted node across the three families. They also check no-op suppression (per node and per field), one event per node across transaction chunks, the origin tag, a single write chaining to exactly one coalesced next level, the depth bound, and a cyclic schema that stops at the bound. The value-based integration tests check the recomputed data is correct after live edits, merges and rebases.

## Notes

- The number of events per written node is the same as before. We keep it for the non-recompute consumers. The win is removing the per-value task and write storm.
- Measured on a dedicated machine, the per-value update flow counts drop to zero (100 to 0 on a heavy merge, 300 to 0 on a rebase), and the derived values stay correct.

---
## Summary by cubic
Bulk-write recomputed display labels, HFIDs, and Jinja2 computed attributes, and coalesce recompute chaining to cut fan-out and speed up recompute after merges and live edits. Removes per-value update flows and updates events/docs; follow-up to IFC-2761.

- **New Features**
  - `BulkRecomputeWriter` groups writes per node, loads the full node, saves in bounded transactions with a dedicated DB session, includes cascaded fields in the change/event, and emits one `NodeUpdated` per node.
  - Coalesced chaining via `RecomputeChainSubmitter` with a schema-derived depth bound (floor 10); chained writes use `NodeMutationOrigin.RECOMPUTE` so per-node automations skip; dispatch only when values changed.
  - The bulk path is now the only recompute write for live edits, merges, and rebases; per-value workflows and their workflow definitions are removed; extracted `BulkRecomputeDispatcher`/`build_bulk_recompute_dispatcher` and injected collaborators; dropped per-node flow-run tags.

- **Bug Fixes**
  - Preserve author attribution; tag runs as database-changing; skip no-op writes/events; tolerate branches deleted between read and persist; log remaining kinds/ids at the chain bound; note per-chunk commits are non-atomic.
  - Load the whole node so same-node derived values (display label, HFID, computed siblings) refresh; include cascaded fields in the changelog and event so cross-node readers chain; docs and event metadata updated to add the `recompute` origin.

<sup>Written for commit ccd22eaa3bf1ddc62209b9ad0472949cda022e05. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/computed_attribute/tasks.py
```diff
@@ -359,6 +366,13 @@ async def process_jinja2(
         if not found:
             log.debug("No nodes found that requires updates")
 
+        if bulk_write_enabled:
+            for node in found:
+                value = await jinja_template.render(variables=node.variables)
+                if value != node.computed_attribute_value:
+                    writes.append(AttributeValueWrite(node_id=node.node_id, field=attribute.name, value=value))
+            continue
```
saltas888: this is mostly about taste, but I'd prefer an if/else here instead of goto-like expressions(continue)

dev/knowledge/backend/merge-recompute.md
```diff
@@ -0,0 +1,91 @@
...
+The per-node path emits one node event per changed node, and Prefect matches each event against every recompute automation. A large merge therefore starts one flow per changed node per family, and nothing dedups across nodes. The coalesced path replaces that fan-out with a single deduplicated pass: it works out once which derived values the change set affects, then submits one flow per affected value. The stored values are the same. The work scales with the number of affected derived values, not with the changed-node count.
+
+## The flow
+
+```text
+merge / rebase
+  -> stamp node events with origin = merge | rebase   (suppresses the 3 families' per-node triggers)
+  -> CoalescedRecomputeBuilder.build(changes, branch)  (diff changelog -> deduplicated target set)
+  -> CoalescedRecomputeSubmitter.submit(...)           (one process flow per target and source kind)
+       -> computed_attribute_process_jinja2 / display-label-process-jinja2 / hfid-process
```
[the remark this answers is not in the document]
gmazoyer: The actual flow name is `display-label-process-jinja2`

backend/infrahub/core/recompute/bulk_write.py
```diff
@@ -0,0 +1,155 @@
...
+    field: str
+    value: str | list[str] | None
+
+
+@dataclass(frozen=True)
+class WrittenNode:
+    """A node whose derived values were persisted, and the fields that were written.
+
+    Returned so the caller can chain: these writes become the change set of the next coalesced
+    recompute level, reaching values that read them.
```
ajtmccarty: not the responsibility of this class

backend/infrahub/core/recompute/bulk_write.py
```diff
@@ -0,0 +1,155 @@
...
+HFID_FIELD = "human_friendly_id"
+
+
+@dataclass(frozen=True)
+class AttributeValueWrite:
+    """One recomputed derived value to persist on one node.
+
+    ``field`` is the attribute to set: the display-label attribute, the human-friendly-id
+    attribute, or a computed attribute's own name. ``value`` is a list of parts for the
+    human-friendly id and a string for the others.
```
ajtmccarty: seems obvious. maybe not the "human-friendly ID uses a list" part, but also not really sure if that matters

backend/infrahub/core/recompute/dispatch.py
```diff
@@ -0,0 +1,60 @@
...
+) -> None:
+    """Persist recomputed values in one bulk write; on a coalesced pass, dispatch the next level.
+
+    An empty write set is a no-op. A coalesced pass stamps the write with the recompute origin and
+    drives the next level here; a live pass stamps it live and lets the per-node path carry it.
+    """
+    if not writes:
+        return
+
+    await add_tags(nodes=sorted({item.node_id for item in writes}), db_change=True)
```
ajtmccarty: I think this should be at the flow-level. also I think there might be a max on the number of linked items you can add to a flow

backend/infrahub/core/recompute/dispatch.py
```diff
@@ -0,0 +1,60 @@
...
+from infrahub.workers.dependencies import get_database, get_event_service, get_workflow
+from infrahub.workflows.utils import add_tags
+
+if TYPE_CHECKING:
+    from infrahub.core.recompute.bulk_write import AttributeValueWrite
+    from infrahub.core.schema.schema_branch import SchemaBranch
+    from infrahub.events.models import EventContext
+
+
+async def persist_and_chain(
```
ajtmccarty: from what I can tell, the dependencies for this function look something like this
```
persist_and_chain()
  BulkRecomputeWriter().write() 
  submit_recompute_chain()
    CoalescedRecomputeBuilder().build()
    CoalescedRecomputeSubmitter().submit()
```

all of these should be built as components with their dependencies injected at the flow-level. `persist_and_chain` needs new name. `BulkRecomputeOrchestrator` or something, not entirely sure. `submit_recompute_chain` should also become a class with dependencies. the three flows that use `persist_and_chain` can call a shared `build_<new_class_name>` function

I could be missing something that prevents this approach, but the mix of classes and functions feels like spaghetti to me

ajtmccarty (commented):
I am complaining about dependency injection. as usual.

backend/infrahub/core/recompute/dispatch.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+async def build_bulk_recompute_dispatcher(schema_branch: SchemaBranch) -> BulkRecomputeDispatcher:
+    """Wire a bulk recompute dispatcher from the flow-level dependencies."""
+    return BulkRecomputeDispatcher(
+        db=await get_database(),
+        event_service=await get_event_service(),
+        workflow=get_workflow(),
+        schema_branch=schema_branch,
+    )
```
gmazoyer, line 79: @ajtmccarty do you think a factory such as one like that is a good idea? Or should we build the object in place when needed?

ajtmccarty: I think that, ideally, `get_database()`, `get_workflow()`, and the other singleton getters should only be called at application entry points. so if you want to call those at the top of the `@flow` function and then pass `db`, `workflow`, etc. in here, then I think that is better. although it does not gain you much at that point. building the component inside of the `@flow` function is fine too. if the component needs to be constructed in multiple places, then it would be better to have a function to help with it though

gmazoyer: Thanks. I kept the factory because the three flows all build the dispatcher, so it gives us one place to wire it. I left the getters inside it for now, moving them up to each flow does not buy much like you said. Easy to change later if we want it stricter.

backend/infrahub/core/recompute/dispatch.py
gmazoyer: I removed the nodes IDs from the tags as 250 node tags per run bloats the flow-run JSON and the tag namespace and per-node tags probably aren't meaningful on a bulk run. I haven't found a limit in prefect doc regarding this though.

backend/infrahub/core/recompute/dispatch.py
```diff
@@ -0,0 +1,78 @@
...
+        writes: list[AttributeValueWrite],
+        branch_name: str,
+        context: EventContext,
+        coalesced: bool,
+        recompute_depth: int,
+    ) -> None:
+        if not writes:
+            return
+
+        await add_tags(db_change=True)
```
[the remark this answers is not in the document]
gmazoyer: almost 250 IDs for tags is probably a bit too many so it actually useful, especially for a bulk operation. so we want to keep it that way for now.

ajtmccarty (approved):
great stuff. very solid

PR #9861, opened 2026-07-10
feat(backend): close relationship data on schema relationship removal

### LINKED ISSUE #2474: task: Schema migration for node.relationship.remove
### Component

API Server / GraphQL

### Task Description

Implement a migration to delete an existing relationship from the schema.

> We need to ensure that all references to this relationship identifier have been deleted before updating the data in the database

ajtmccarty:
# Close relationship data on schema relationship removal (IFC-402)

Closes #2474 

## Summary

Removing a relationship from a schema was a **no-op** on the underlying data:
`"node.relationship.remove"` mapped to `PlaceholderDummyMigration`, so the schema definition
changed but the `IS_RELATED` edges and `Relationship` vertices stayed `status="active"` in the
graph. This "works" b/c removing the relationship from a schema makes it unreachable for anything
other than direct database access.

This PR replaces that placeholder with a real, branch-aware migration that closes the
relationship data, and fixes two graph-correctness bugs uncovered while validating it.

Consequences of the old behaviour (all fixed here):

- **Resurfacing hazard** — re-adding a relationship with the same identifier reconnected the stale
  peers.
- **Incorrect diff** — relationship data for deleted relationship schemas would not show up in the
  diff as removed.

### New migration: `node.relationship.remove`

`backend/infrahub/core/migrations/schema/node_relationship_remove.py` closes the graph data for a
removed relationship.

- **Close only when the schema no longer references it.** Relationship data is shared by both sides
  of a relationship (an outbound relationship and its inbound inverse share one identifier and the
  same `Relationship` vertices). The migration closes the data only once *no* active schema still
  declares that identifier for the affected peers. It computes the set of kinds that still declare
  the identifier (a surviving inverse side, or an inheriting kind that overrides a generic and keeps
  its own relationship) and skips their vertices; if the node's own kind or its peer still declares
  the identifier, the migration short-circuits and runs nothing. Because data is only closed once
  every schema reference is gone, edge direction never has to be considered.
- **Profile and object-template copies.** Profiles and templates replicate the relationship under
  `profile_`- and `template_`-prefixed identifiers, and the schema diff never emits a separate
  migration for those kinds, so the migration closes all three variants together.

The migration owns the schema analysis and short-circuiting; the query is a pure Cypher builder that
takes a precomputed set of parameters (`node_kinds`, `kinds_to_skip`, `rel_identifiers`).

`backend/infrahub/core/migrations/__init__.py` wires `NodeRelationshipRemoveMigration` into
`MIGRATION_MAP` and `placeholder_dummy.py` is deleted (it had no remaining users).

### 1. Rebase used the wrong schema baseline for remove migrations

Bug from merge refactor only present on `develop`.

`backend/infrahub/core/branch/tasks.py` — `rebase_branch()` computed migrations with
`pre_rebase_schema = schema_analyzer.destination_schema` (the default branch's schema). That schema
is missing anything removed on either side of the rebase, so a remove migration had no origin
definition to resolve *what* to close and the rebase crashed. The baseline is now the
branch-creation (common-ancestor) schema via `schema_analyzer.get_initial_source_branch()`, which
still contains any element removed on either branch.

This is a general rebase bug, not specific to relationships — removing an attribute or a node on a
branch and then rebasing hit the same missing-baseline crash. The fix covers all schema-element
remove migrations during a rebase.

### 2. `RelationshipDeleteAllQuery` created duplicate `deleted` edges

`backend/infrahub/core/query/relationship.py` — when more than one edge existed between a
`Relationship` vertex and a peer (e.g. an active edge on a parent branch plus a `deleted` shadow on
the child, as happens when the same relationship is removed on two branches independently and then
rebased), the delete-all query matched *every* active edge and created a `deleted` shadow for each,
producing duplicate parallel edges. It now resolves the single latest edge per `(relationship, peer)`
pair and closes only that one if it is still active. No user would have been aware of this issue and
it would not have caused any problems, but it was technically incorrect.

This was surfaced by the added "remove on both branch and main, then rebase" test. The duplicated
edges landed on the internal schema-representation graph (`schema__node__*`), so it affected schema
rebases regardless of this feature.

## Checklist

- [x] Tests added/updated

---
## Summary by cubic
Replaces the no-op `node.relationship.remove` with a branch-aware migration that closes relationship data and fixes diff/rebase edge cases to prevent orphaned edges and stale peer resurfacing (IFC-402).

- **Bug Fixes**
  - Closes `Relationship` vertices and sub-edges (`IS_RELATED`, `IS_PROTECTED`, `HAS_SOURCE`, `HAS_OWNER`) across node, `Profile*`, and `Template*` copies; runs only when no schema still references the identifier (inverse side or overriding inheriting kinds keep their data; handles `profile_`/`template_` identifiers).
  - Branch-aware and idempotent: closes in place on the operating branch; shadows parent edges with `deleted` edges on child branches; resolves the latest edge per peer to prevent duplicate deleted edges.
  - Diff correctness: pre-existing data is REMOVED; same-branch create+remove is UNCHANGED.
  - Rebase correctness: uses the branch-creation common-ancestor schema as the migration baseline; short-circuits when the relationship is absent from both previous and new schemas to avoid crashes.

<sup>Written for commit b4ecf52d751663d41409a89975e3697f5a76dfa8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/migrations/schema/node_relationship_remove.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        query = query_template % {
+            "branch_filter": branch_filter,
+            "node_kinds": "|".join(params.node_kinds),
+        }
+        self.add_to_query(query)
+
+
+class NodeRelationshipRemoveMigration(RelationshipSchemaMigration):
+    name: str = "node.relationship.remove"
```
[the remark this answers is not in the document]
ajtmccarty: updating a relationship's identifier is currently blocked, so this will be a problem for that migration when it is added

backend/infrahub/core/branch/tasks.py
```diff
@@ -191,7 +191,9 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
             if error_messages:
                 raise ValidationError(",\n".join(error_messages))
 
-        pre_rebase_schema = schema_analyzer.destination_schema.duplicate()
```
ajtmccarty, line 201: bug found during testing. it's caused by an earlier merge/rebase refactor that is only in `develop`, so does not need a backport. the issue is that using the `destination_schema` at this point would be the current state of the default branch schema, which would be missing anything deleted on that branch and prevent "remove" migrations from accessing the entity to delete

backend/infrahub/core/migrations/schema/placeholder_dummy.py
ajtmccarty, line 1: no longer used now that `node.relationship.remove` has a real migration and `node.relationship.identifier` is prevented

backend/infrahub/core/query/relationship.py
```diff
@@ -1381,10 +1381,15 @@ async def query_init(self, db: InfrahubDatabase, **kwargs) -> None:
...
+                # Resolve the latest edge to each peer and close it
                 sub_query = """
                     CALL (rl) {
                         MATCH (rl)%(arrow_left)s[active_edge:%(edge_type)s]%(arrow_right)s(n)
-                        WHERE %(active_rel_filter)s AND active_edge.status ="active"
+                        WHERE %(active_rel_filter)s
+                        WITH rl, active_edge, n
+                        ORDER BY %(id_func)s(rl), %(id_func)s(n), active_edge.from DESC
+                        WITH rl, n, head(collect(active_edge)) AS active_edge
+                        WHERE active_edge.status = "active"
```
ajtmccarty, line 1392: small bug found during testing. the filter was not quite right here, it only checked for an "active" edge, but this doesn't handle the case of a delete on a branch correctly. when deleting on a branch we add a new `[:EDGE_TYPE {branch: "my_branch", status: "deleted"}]` edge, so just filtering for an "active" edge would still surface the active edge on the default branch. fix is to order them chronologically, get the latest edge, and only do the delete if the latest edge is active

backend/tests/component/core/test_node_manager_delete.py
ajtmccarty, line 1: add a bunch of `verify_graph()` calls to catch any other sneaky delete bugs

backend/infrahub/core/branch/tasks.py
gmazoyer: wow nasty one

gmazoyer (approved):

PR #9867, opened 2026-07-10
Add Sort button to objects manager with searchable field picker

bilalabbad:
## Summary

  Adds sorting to the objects manager: a **Sort** button in the toolbar opens a picker for building a multi-key sort across a schema's attributes, cardinality-one relationship's attributes, and node metadata (`Created at` / `Updated at`). The sort is persisted in the URL and applied to the objects query.

  ## UX

  The add-sort picker has two modes driven by the search input:

  - **Browse (empty search):** the node's own attributes flat, one submenu per sortable relationship listing the peer's attributes (ordered by `order_weight`), then metadata fields. Relationships whose peer has no sortable attribute are hidden.
  - **Search (typing):** all fields flat with `Peer › Attribute` labels, so typing a relationship name surfaces every field reachable through it. Clearing the search restores the grouped layout.

Every field opens an Ascending/Descending submenu. Selected sorts append to the current sort keys; re-selecting an already-sorted field replaces it and moves it to the end. Sort keys can be reordered by drag & drop in the editor.

  ## Behavior & state

  - Sort state lives in the `sort` query param as `{field}__{direction}` tokens (nuqs), so sorts are shareable and survive reloads.
  - URL input is validated against the schema (`getValidSorts` builds the allowlist of sortable fields once per list): stale or hand-edited sort params are silently dropped instead of erroring the object table — e.g. after a schema change or branch switch.
  - With no explicit sort, the default is derived from the schema's `order_by`.
  - Non-orderable attribute kinds (JSON, List, Any, Password, HashedPassword) are excluded everywhere; only cardinality-one relationships are sortable-through.

  ## Architecture

  - **Domain** (`entities/nodes/sort/domain/`): predicates (`isSortableAttribute`, `isSortableRelationship`), sort-field token builders (`buildAttributeSortField`,`buildRelationshipSortField`), URL validation (`getValidSorts`), default sort, and token parse/serialize — pure rules, unit-tested.
  - **UI** (`entities/nodes/sort/ui/`): components own their arrangement — `AddSortPicker` composes the menu from domain predicates, resolves peers reactively via `useSchema`, and switches flat/grouped rendering at a single point. Labels and menu structure never leak into domain.
  - Field tokens are built exclusively through the domain builders, so the picker and the URL allowlist cannot drift.

  ## Testing

  - Unit tests for every domain rule (token building, validation allowlist incl. hostile/stale tokens, sortability predicates, default sort).
  - Browser-mode component tests for the picker: grouped layout, the 3-level submenu cascade, flat search with breadcrumb labels, label fallback for unlabeled attributes, metadata
  selection (both tokens pinned), empty-peer hiding, and grouped-layout restore on clearing search.


frontend/app/src/entities/nodes/sort/domain/rules/get-valid-sorts.ts
```diff
@@ -0,0 +1,48 @@
...
+import { isSortableAttribute } from "@/entities/nodes/sort/domain/rules/is-sortable-attribute";
+import { isSortableRelationship } from "@/entities/nodes/sort/domain/rules/is-sortable-relationship";
+import {
+  buildAttributeSortField,
+  buildRelationshipSortField,
+} from "@/entities/nodes/sort/domain/rules/sort-field";
+import type { ModelSchema } from "@/entities/schema/domain/model/schema";
+import { getSchema } from "@/entities/schema/domain/use-cases/get-schema";
+
+const NODE_METADATA_SORT_FIELDS: SortField[] = [
```
pa-lem: **Single source of truth for metadata sort fields.** These field IDs are declared twice: `NODE_METADATA_SORT_FIELDS` here and `METADATA_SORTABLE_FIELDS` in `add-sort-picker.tsx:26-29` (field + label). They agree today, but they're now separate sources of truth — the picker controls what's *offered* and this allowlist controls what *survives* URL validation, so adding a metadata field to one but not the other silently breaks (offered-but-stripped, or valid-but-hidden).

Worth noting the attribute/relationship parts *can't* drift — both paths go through the same `isSortableAttribute` / `isSortableRelationship` / `build*SortField` rules. Metadata is the only duplicated data.

Suggest one labeled constant in the domain layer both consume, e.g. `sort/domain/rules/metadata-sort-fields.ts`:

```ts
import type { SortField } from "@/entities/nodes/sort/domain/model/sort";

export const NODE_METADATA_SORT_FIELDS: { field: SortField; label: string }[] = [
  { field: "node_metadata__created_at", label: "Created at" },
  { field: "node_metadata__updated_at", label: "Updated at" },
];
```

Then here spread the ids (`...NODE_METADATA_SORT_FIELDS.map(({ field }) => field)`), and in the picker import this constant instead of the local list (which lets the local `SortableField` interface at `add-sort-picker.tsx:21` go away too).

frontend/app/src/entities/nodes/sort/ui/add-sort-picker.tsx
```diff
@@ -0,0 +1,237 @@
...
+  RelationshipSchema,
+} from "@/entities/schema/domain/model/schema";
+import { useSchema } from "@/entities/schema/ui/hooks/useSchema";
+
+interface SortableField {
+  field: SortField;
+  label: string;
+}
+
+const METADATA_SORTABLE_FIELDS: SortableField[] = [
```
pa-lem: Part of the same finding — see the comment on `get-valid-sorts.ts:11`. This `METADATA_SORTABLE_FIELDS` list duplicates the metadata field IDs in `NODE_METADATA_SORT_FIELDS`; import a single shared labeled constant from the domain layer instead so the picker and the URL allowlist can't drift.

frontend/app/src/entities/nodes/sort/ui/sort-editor.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+export function SortEditor({ schema }: SortEditorProps) {
+  const { sort, setSort } = useSort(schema);
+
+  const currentSort: Sort[] = sort ?? getSchemaDefaultSort(schema) ?? [];
+
+  const addSort = (newSort: Sort) => {
+    const withoutField = currentSort.filter((existing) => existing.field !== newSort.field);
+    setSort([...withoutField, newSort]);
+  };
```
pa-lem, line 18: **Extract & test the "re-select replaces and moves to end" logic.** This is the headline sort behavior and it has no test — `sort-editor.test.tsx` only asserts menu visibility. It's also currently **unreachable through the UI**: active fields are filtered out of the picker and the active-sort list is still the placeholder (`:24` `// TODO`), so there's no way to re-select an already-sorted field yet. The logic is correct, but it's dead until that UI lands.

Since it's pure list logic, extracting it to a domain rule makes it testable now and puts it with the other rules:

```ts
// sort/domain/rules/append-sort.ts
import type { Sort } from "@/entities/nodes/sort/domain/model/sort";

/** Adds a sort key, keeping fields unique: re-adding a field replaces it and
 *  moves it to the end (least-significant key) with its new direction. */
export function appendSort(currentSort: Sort[], newSort: Sort): Sort[] {
  const withoutField = currentSort.filter((existing) => existing.field !== newSort.field);
  return [...withoutField, newSort];
}
```

`addSort` then becomes `setSort(appendSort(currentSort, newSort))`, and `append-sort.test.ts` covers: append new field; re-add existing → replaced, moved to end, new direction; field stays unique; input not mutated.

pa-lem (commented):
Nice work — the domain-rules decomposition is clean and the test coverage is genuinely strong (the interaction tests in `add-sort-picker.test.tsx` are thorough, and the refactor kept the hostile-token validation case). A few things before merge, none blocking, left as inline comments.

I checked that the schema default sort not being sent from the client is fine — the backend applies `order_by` by default, so `useSort` returning `null` on an empty URL is correct.

Minor/optional: the PR description says `getValidSorts` "builds the allowlist of sortable fields once per list" — it's actually rebuilt on every render of every `useSort` consumer (not memoized). Perf is fine at this scale, so no change needed; just flagging the wording (the code comment itself is accurate).

pa-lem (approved):

PR #9876, opened 2026-07-10
fix(migrations): repair HFID values stored with non-string elements

gmazoyer:
# Why

After an upgrade, some human-friendly ID (HFID) values were left stored with non-string elements (numbers) instead of all strings. The same object type could end up with mixed HFID types, and re-saving an object was needed to fix it. A customer hit this on a large database.

Goal: HFID values are consistently stored as all-strings after an upgrade, both for fresh upgrades and for instances that already ran the affected upgrade.

Non-goals: no change to how HFID values are computed or written at runtime.

## What changed

Behavioral:

- Fresh upgrades: migration 072 now normalizes every HFID value, not only those that fit in a single batch. The old code skipped values because it paginated over the same value nodes it was rewriting.
- Already-upgraded instances: a new migration 074 repairs values that 072 left non-string and indexed. It normalizes them, keeps the indexed label when the value still fits the index size limit, and removes the old value node. It runs automatically on the next upgrade, no manual step.

Implementation notes:

- Both migrations paginate over the HFID attribute nodes, a set the writes never create or delete, and flush each batch. Memory stays flat regardless of database size, and no value is skipped.
- Normalization uses `str()` on every element, so floats and negative numbers are handled, not only integers.
- 074 matches the runtime byte-based index-size check and never relabels a plain value node that a non-HFID attribute shares.

What stayed the same:

- No schema changes. The HFID compute and write path is unchanged. Migration 072's indexing step is unchanged.

## How to test

```bash
GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null INFRAHUB_USE_TEST_CONTAINERS=1 \
  uv run pytest \
    backend/tests/component/core/migrations/graph/test_072_index_hfid_values.py \
    backend/tests/component/core/migrations/graph/test_074_normalize_indexed_hfid_values.py \
    backend/tests/unit/core/graph/test_graph_version.py
```

All pass: 10 migration tests plus 3 graph-version guardrail tests. The multi-page tests use a batch size of 2 to prove no value is skipped across pagination batches.

## Impact & rollout

- Backward compatibility: no breaking changes.
- Performance: normalization runs in batches with bounded memory, so it scales to very large databases.
- Deployment notes: safe to deploy. Instances already on the affected release are repaired on their next upgrade.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (not needed, HFID behavior is unchanged for users)

---
## Summary by cubic
Normalizes all HFID values to strings during upgrades and adds a repair migration for databases that stored non-string elements. Ensures consistent HFID storage and keeps indexing when the value fits.

- **Bug Fixes**
  - Migration 072 now paginates over HFID attributes and normalizes every value; converts all element types via str() (ints, floats, negatives).
  - Added migration 074 to fix existing non-string HFID values (indexed and plain); keeps the index only if within the byte limit, dedupes across HFID attrs, deletes orphaned nodes, avoids relabeling shared plain nodes, and is idempotent.

- **Migration**
  - Runs automatically on upgrade; processes in batches with bounded memory for large databases. Bumps graph version to 74. No schema changes.

<sup>Written for commit e59abf03ab129307e9f95c6811fcc682b40ba37f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/migrations/graph/m074_normalize_indexed_hfid_values.py
```diff
@@ -0,0 +1,159 @@
...
+    MERGE (new_av:AttributeValue:AttributeValueIndexed {value: update.new_value, is_default: old_av.is_default})
+    WITH attr, old_r, old_av, new_av
+    LIMIT 1
+    CREATE (attr)-[new_r:HAS_VALUE]->(new_av)
+    SET new_r = properties(old_r)
+    DELETE old_r
+    WITH old_av
+    WHERE NOT (old_av)--()
+    DELETE old_av
+}
```
ajtmccarty: I think you should do these `IN TRANSACTIONS OF XYZ ROWS`
might be unnecessary given they are already batched, but better to be safe. same for the PLAIN_QUERY

backend/tests/component/core/migrations/graph/test_074_normalize_indexed_hfid_values.py
```diff
@@ -0,0 +1,212 @@
...
+    for _ in range(2):  # second run must be a no-op
+        assert not (await migration.execute(MigrationInput(db=db))).errors
+        results = await db.execute_query(query=check_query)
+        assert len(results) == 1
+        assert results[0].get("value") == '["650001"]'
+        assert results[0].get("is_indexed") is True
+
+    # The old integer-typed node is removed once it no longer carries any edge.
+    orphan = await db.execute_query(query="MATCH (av:AttributeValue {value: '[650001]'}) RETURN count(av) AS c")
+    assert orphan[0].get("c") == 0
```
ajtmccarty, line 39: should check this is 1 before the migration runs

ajtmccarty (approved):
this is a sneaky one. great work. just a couple small comments

backend/infrahub/core/migrations/graph/m074_normalize_indexed_hfid_values.py
gmazoyer: I wasn't sure about that but I thought about it. Thanks for confirming my hunch, I will do that.

PR #9879, opened 2026-07-12
feat(api): priority-aware server-side API backpressure

dgarros:
# Why

When an Infrahub instance runs heavy background work (generators, artifacts, diffs, repo syncs, computed attributes) while a human uses the frontend, both compete for the same finite uvicorn worker pool and Neo4j connection pool. Today the API is origin-blind — frontend and background requests are indistinguishable — so under heavy background load the API can no longer serve the frontend and the app appears to hang.

**Goal:** keep the frontend responsive under background overload by shedding load *by priority*, with no per-customer tuning.

**Non-goals (out of scope, tracked separately):** the client/SDK backoff that consumes `Retry-After`, the frontend setting `X-Priority: high`, and enforcement that a `high` claim is legitimate. This PR is **server-side only**.

Implements Jira **IFC-2886** (which implements JPD idea INFP-636).

## What changed

**Behavioral changes (what operators/callers observe):**
- Requests may declare priority via a new `X-Priority` header (`high` / `normal` / `low`); missing or invalid → `normal`, so the layer is safe to deploy before any caller is updated.
- Under overload the server sheds low-priority requests first and protects high-priority (interactive) requests, returning a fast `429 Too Many Requests` + `Retry-After` with **no handler work performed**. Shedding is adaptive (CoDel on measured slot-wait / sojourn time), not a fixed threshold, and self-terminates within a bounded window after overload ends.
- Eight new per-priority metric families on the **existing** `/metrics` endpoint (`infrahub_admission_*`): offered, admitted, rejected-by-reason (`codel`/`backstop`), in-flight, waiters, sojourn histogram, derived `max_concurrency`, and a no/invalid-priority counter for adoption tracking.
- New config knobs (`INFRAHUB_API_BACKPRESSURE_*`) including a kill-switch (`backpressure_enabled`, default on) and a new `INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE` (default `100`, now passed explicitly to the Neo4j driver).

**Implementation notes:**
- New package `backend/infrahub/api/admission/`: a pure-ASGI `AdmissionMiddleware` (registered outermost), a cancellation-safe `PrioritySlotPool` (per-class FIFO, priority hand-off, modelled on `asyncio.Semaphore`), a pure `CoDelController` with an injected clock, an `AdmissionController` composing them, capacity derivation, and the metrics module.
- Per-worker and coordination-free by design (no shared/global limiter, no Redis). `max_concurrency` is derived from the process's own Neo4j pool size — no hard-coded constant.

**What stayed the same:**
- No database schema or migration change; no GraphQL change; no persistence.
- No new hard dependency (`prometheus-client` was already present transitively and imported in-tree; it is now declared explicitly).
- Default behaviour is unchanged under normal load: with no caller sending `X-Priority`, everything is `normal` and nothing is shed until load exceeds the derived cap.

### Suggested review order
1. `contracts/x-priority-header.md` and `contracts/metrics.md` in the spec dir — the external contract.
2. `slot_pool.py` and `codel.py` — the concurrency core (cancellation safety, CoDel state machine).
3. `controller.py` → `middleware.py` → `server.py` — how they wire together.
4. Tests under `backend/tests/unit/api/admission/` and `backend/tests/component/api/test_admission_middleware.py`.

## Related Context

Full spec, plan, research, data model, contracts, critique, and implementation report live in `dev/specs/ifc-2886-priority-api-backpressure/`. Key success criteria: `high` shed rate ≈ 0% under sustained overload while `low` sheds first (SC-002); sub-interval bursts shed zero (SC-003); every shed is a fast `429 + Retry-After` with no handler work (SC-004); shedding self-terminates after overload (SC-005).

The headline latency bound (SC-001) is intentionally discovery-measured against a live overload scenario (see Test Plan) rather than pre-committed.

## Documentation Updates

- Added changelog fragment `changelog/+ifc-2886.added.md`.
- Regenerated `docs/docs/reference/configuration.mdx` for the new settings.

## How to review / test

```bash
# Deterministic unit tests (fake-clock CoDel, slot-pool cancellation, capacity, parser)
uv run pytest backend/tests/unit/api/admission/ -v

# Component tests (middleware end-to-end: classification, 429+Retry-After, gradient, metrics, kill-switch, excluded paths)
uv run pytest backend/tests/component/api/test_admission_middleware.py -v
```

Both suites pass locally with no external infrastructure (42 tests). CI to watch: backend unit/component, ruff/mypy lint, and `validate-generated-documentation` (the regenerated config reference is committed).

**Live overload / SC-001 discovery (manual, needs a running server):**
```bash
export INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE=4
export INFRAHUB_API_BACKPRESSURE_CODEL_TARGET_SECONDS=0.005
export INFRAHUB_API_BACKPRESSURE_CODEL_INTERVAL_SECONDS=0.1
# drive a saturating `X-Priority: low` stream + a steady `X-Priority: high` stream, then:
curl -s localhost:8000/metrics | grep infrahub_admission_
# expect rejected_total{priority="low"} climbing while {priority="high"} stays ≈ 0
```

## Impact & rollout

- **Backward compatibility:** fully compatible. No `X-Priority` header → `normal`, no behaviour change under normal load. No request/response body change for non-shed requests.
- **Performance:** protects contended API/Neo4j capacity under load; hot-path cost is a header read + slot attempt + CoDel eval (O(1), per-worker).
- **Config/env changes:** new `INFRAHUB_API_BACKPRESSURE_*` settings and `INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE` (default 100 preserves current driver behaviour).
- **Deployment notes:** safe to deploy. Ships inert under normal load; kill-switch `INFRAHUB_API_BACKPRESSURE_ENABLED=false` is the instant rollback. Interactive protection is fully realized once the frontend sends `X-Priority: high` (separate ticket); until then interactive traffic is `normal`.
- **Governance heads-up:** new `X-Priority` request header + `429` behaviour (public API surface); the middleware sits in the admission path and reads a client-controlled header (borderline auth, no auth logic changed); `X-Priority` is untrusted and cooperative in v1 (enforcement deferred).

## Checklist

- [x] Tests added/updated (42 unit + component tests)
- [x] Changelog entry added (`changelog/+ifc-2886.added.md`)
- [x] External docs updated (regenerated `configuration.mdx`)
- [x] Internal .md docs updated (spec/plan/contracts/report under `dev/specs/ifc-2886-priority-api-backpressure/`)

---
## Summary by cubic
Adds priority-aware, server-side API backpressure with CoDel- and database‑stress–driven shedding and adaptive Retry-After to keep interactive requests responsive under load. The per-worker cap derives from the DB pool with a default concurrency factor of 0.5 (pool × 0.5), enabled by default behind `INFRAHUB_API_BACKPRESSURE_ENABLED`.

- New Features
  - Priority admission (`high`/`medium`/`low`), exact excluded-path matching, CORS preflight bypass, and CORS allow‑list for `x-priority`; per‑priority `infrahub_admission_*` metrics; `prometheus-client` declared.
  - Database stress-aware shedding from a rolling median of the reference permission query with tiered, graduated fractions (`reason="stress"`), alongside CoDel; new `INFRAHUB_API_BACKPRESSURE_*` knobs; DB pool default 100.
  - Adaptive Retry-After scaled by shed intensity and sustained-load duration (new sustained-load metric), capped under the SDK per‑retry ceiling.
  - Client integration: the frontend sends `X-Priority` (high by default; low opt‑in; preserved on retries/uploads; only for the Infrahub API origin). Background flows set SDK `RequestContext.priority`; artifact generation no longer forces HIGH. The `python_sdk` tracks `infrahub-develop` with a 429 retry budget of 10.

- Bug Fixes & Refactors
  - Slot‑pool observer failures can’t leak slots or strand waiters; non‑finite concurrency factor is rejected at config load.
  - Admission controller is built at app startup and read from `app.state`; excluded paths and CORS preflights bypass admission before reading startup state; exclusions match exact path or descendant and include `/api/config`; middleware decision branching is exhaustive.
  - CI: disable backpressure only for the legacy Playwright E2E job via `INFRAHUB_API_BACKPRESSURE_ENABLED=false`; default remains enabled elsewhere.
  - Internal cleanup: `RetryAfterPolicy` now pushes sustained‑load duration to a registered observer (metrics‑free, sink pattern), wired in the controller; renamed `codel` map to `codel_priority_map` for clarity.
  - Docs: extracted IFC‑2886/IFC‑2890 into ADRs and backend/frontend knowledge; archived specs as “extracted” and fixed index links and an archived‑spec knowledge link depth; no behavior change.

<sup>Written for commit 6189969f521ef59c86127da3005fb68c6d9c9e44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9879?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Update since original description

- **Tier rename `normal` → `medium`.** Header values are now `high` / `medium` / `low`; a missing or invalid value resolves to `medium`. Read every `normal` above as `medium`.
- **Database-stress signal added.** The admission decision now also consumes a per-worker signal that times the global permission (reference) query: past per-class stress-ratio triggers it sheds a graduated fraction of a class, reported as `reason="stress"` alongside `codel`/`backstop`. This augments the CoDel sojourn signal described above.
- **Config defaults retuned.** DB connection pool `100 → 50`, admission concurrency factor `1.0 → 0.5`, plus new `INFRAHUB_API_BACKPRESSURE_*` stress knobs.
- **New backend knowledge doc** `dev/knowledge/backend/api-backpressure.md` documents the admission layer and the stress signal (the IFC-2886 spec predates the stress signal and the rename).


dev/specs/archive/ifc-2886-priority-api-backpressure/contracts/metrics.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Histogram buckets** (`infrahub_admission_sojourn_seconds`): `[0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 5]` (seconds; fine at the low end where the CoDel `target` sits).
+
+## Invariants (assertion-testable)
+
+| ID | Invariant |
+|----|-----------|
+| M-1 | For each class, for every request the server adjudicates: `offered_total == admitted_total + rejected_total{codel} + rejected_total{backstop}`. A request whose client disconnects while it is still queued is counted in `offered_total` only — it is neither admitted nor shed by the server — so under in-flight cancellation `offered_total` may transiently exceed `admitted + rejected` by the number of abandoned waiters. The `waiters`/`in_flight` gauges stay accurate across that case (they are driven by the pool's own enqueue/dequeue transitions, including cancellation). |
+| M-2 | `max_concurrency` gauge equals `derive_max_concurrency(pool_size, factor)` and is > 0 (no magic number; FR-009/FR-OBS-6). |
+| M-3 | `in_flight{priority}` never exceeds `max_concurrency`; `sum(in_flight)` never exceeds `max_concurrency`. |
```
polmichel, line 33: The first condition is included in the second one, right? I'm not suggesting a change here, I just want to make sure I've understood this correctly.

backend/infrahub/api/admission/slot_pool.py
```diff
@@ -0,0 +1,148 @@
...
+
+        self._pool = pool
+        self._released = False
+
+    def release(self) -> None:
+        """Return the slot to the pool. Idempotent."""
+        if self._released:
+            return
+        self._released = True
+        self._pool._release_acquisition(priority=self.priority)
```
polmichel: Shouldn't it be up to the pool to own its own release method? Wherever the acquisition is held in a method, we should have access to the pool anyway, in my opinion.
Having it this way could lead to maintainability issues, I think, since it means we can operate on the pool wherever we manipulate these `Acquisition` objects.

And I think it's fine if we have a `release(Acquisition)` method inside the pool.

One example pointing in this direction: if you try to replace the `Acquisition.release()` calls with a `Pool.release(Acquisition)` call right now, you can do it, since the controller already has access to the pool via self.pool.

polmichel: And in that case you could get rid of the dependency related to the pool inside this object.
Obviously, I can miss something here

backend/infrahub/api/admission/slot_pool.py
```diff
@@ -0,0 +1,149 @@
...
+
+        This lets an external observer (e.g. metric gauges) track queue depth the moment
+        a request enqueues or leaves the queue, rather than only when some other request is
+        admitted or released. The primitive itself stays free of any metrics dependency.
+        """
+        self._on_change = on_change
+
+    def _notify(self, priority: Priority) -> None:
+        if self._on_change is not None:
+            self._on_change(priority)
```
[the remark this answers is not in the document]
polmichel: I'm not sure this is the best solution exactly, but it's true that, given that we can pass any callable to `_on_change`, having some extra safety there seems important.

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
...
+            self._first_above_time = None
+            self._dropping = False
+            return False
+
+        ok_to_drop = False
+        if self._first_above_time is None:
+            # Start of an excursion above target; tolerate it for one full interval.
+            self._first_above_time = now + self._interval
+        elif now >= self._first_above_time:
+            ok_to_drop = True
```
polmichel, line 63: `ok_to_drop` then means `!(self._first_above_time is None || now < self._first_above_time)`

So here is an optional suggestion, would it not be more readable to do something like
```
if self._first_above_time is None:
     # Start of an excursion above target; tolerate it for one full interval.
     self._first_above_time` = now + self._interval
     return False
elif now < self._first_above_time:
     return False
```

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if self._dropping:
+            if now >= self._drop_next:
+                self._count += 1
+                self._drop_next = self._control_law(reference=self._drop_next, count=self._count)
+                return True
+            return False
```
polmichel, line 70: Still an optional comment:
To enter this block, dropping must be `True`, which means that `self._first_above_time is not None`. So I think this block could be moved before the previous one. This block represents more of a general state, while the previous one is more like internal machinery.

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if ok_to_drop:
```
polmichel, line 72: If you apply the optional comment about `ok_to_drop` the if block is not required anymore, neither the last return.

backend/infrahub/api/admission/priority.py
```diff
@@ -0,0 +1,52 @@
...
+    This never raises.
+    """
+    if header_value is None:
+        return PriorityHeaderParseResult(priority=Priority.MEDIUM, was_explicit=False)
+
+    normalized = header_value.strip().lower()
+    try:
+        priority = Priority[normalized.upper()]
+    except KeyError:
+        return PriorityHeaderParseResult(priority=Priority.MEDIUM, was_explicit=False)
```
polmichel, line 50: No error logged in that case?

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,151 @@
...
+            try:
+                await self.app(scope, receive, send)
+            finally:
+                self._controller.release(acquisition=decision.acquisition)
+            return
+        if isinstance(decision, Rejected):
+            # Short-circuit: the shed response is written directly, so the downstream app and
+            # every handler dependency (auth, routing, DB query) never runs.
+            response = _build_shed_response(path=path, retry_after=decision.retry_after)
+            await response(scope, receive, send)
```
polmichel: Instead of using the `isinstance()` operator, I would tend to use a method defined on an ABC on top of `Admitted` and `Rejected` (`Decision`?), which both of these classes would implement.
From a maintainability perspective, it's better in the long run since:

- we can easily reuse this behavior, which is encapsulated in a meaningful method
- a new decision will have to explicitly define the behavior
- this avoids any future code duplication

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,151 @@
...
+        "data": None,
+        "errors": [{"message": _SHED_MESSAGE, "extensions": {"code": HTTP_429_TOO_MANY_REQUESTS}}],
+    }
+
+
+def _graphql_shed_envelope() -> dict[str, Any]:
+    return {
+        "data": None,
+        "errors": [{"message": _SHED_MESSAGE, "extensions": {"code": HTTP_429_TOO_MANY_REQUESTS}}],
+    }
```
polmichel, line 156: Is this meant to return different outputs? If not, maybe it's not worth to differentiate those?
I am maybe missing something here

backend/tests/component/api/test_admission_middleware.py
```diff
@@ -0,0 +1,664 @@
...
+        magic number.
+    (b) A burst of concurrent requests that completes within a single CoDel interval cannot be
+        shed: the first above-target sojourn only arms the interval timer, and dropping starts
+        only after a full interval of continuous overload. A fast handler and a long interval keep
+        the whole burst inside one window, so the shed count is unchanged by construction.
+    """
+    pool_size = 40
+    factor = 0.25
+    max_concurrency = derive_max_concurrency(pool_size=pool_size, factor=factor)
+    assert max_concurrency == 10
```
polmichel, line 341: Is this not an assertion that would deserve its own unit test?

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,151 @@
...
+    from .controller import AdmissionController
+
+EXCLUDED_PATHS: tuple[str, ...] = (
+    "/health",
+    "/metrics",
+    "/assets",
+    "/favicons",
+    "/docs",
+    "/api/schema",
+)
```
polmichel, line 26: would `/api/config` also be a good candidate?

backend/tests/component/api/test_cors_priority.py
```diff
@@ -0,0 +1,90 @@
...
+    assert response.status_code == 200
+    assert "x-priority" in response.headers["access-control-allow-headers"].lower()
+
+
+async def test_non_preflight_request_is_still_shed() -> None:
+    """The preflight exemption is narrow: a real cross-origin request is still admission-gated."""
+    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=_build_app()), base_url="http://test") as client:
+        response = await client.post("/work", headers={"Origin": _ORIGIN, "X-Priority": "low"})
+
+    assert response.status_code == 429
```
polmichel, line 97: Should the shed request also carry headers? Is this the case since the CORS middleware is not the outermost layer?

backend/tests/unit/api/admission/test_controller.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    controller, _ = _build(ratio=case.ratio, samples=case.samples, rng_value=case.rng_value, max_concurrency=10)
+
+    result = await controller.admit(priority=case.priority)
+    if isinstance(result, Admitted):
+        result.acquisition.release()
+        assert case.expect_shed is False
+    else:
+        assert case.expect_shed is True
+        assert result.reason == "stress"
```
polmichel, line 167: If the `isinstance()` is removed in the production code, it should be removed over there aswell

backend/tests/unit/api/admission/test_priority.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_priority_ordering() -> None:
+    assert Priority.HIGH < Priority.MEDIUM < Priority.LOW
+
+
+def test_priority_labels() -> None:
+    assert Priority.HIGH.label == "high"
+    assert Priority.MEDIUM.label == "medium"
+    assert Priority.LOW.label == "low"
```
polmichel, line 50: If I am not wrong, these tests are just testing the `IntEnum` Python behavior

backend/infrahub/api/admission/codel.py
ajtmccarty, line 1: using the actual name makes it easier to understand. so `controlled_delay.py`. `codel` looks like a typo to me

backend/tests/unit/api/admission/test_slot_pool.py
```diff
@@ -0,0 +1,211 @@
...
+    """Manually advanced monotonic clock so sojourn is deterministic in tests."""
+
+    def __init__(self) -> None:
+        self.now = 0.0
+
+    def __call__(self) -> float:
+        return self.now
+
+    def advance(self, seconds: float) -> None:
+        self.now += seconds
```
polmichel: I think this is the third time I've seen this class. Have you considered moving it into a test helper file?

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
+
+import math
+import time
+from typing import Callable
+
+
+class CoDelController:
```
ajtmccarty, line 8: `ControlledDelayController` sounds like a joke name. `DelayController` or `ControlledDelayer` is better

polmichel (commented):
The backend part of the code is now reviewed.

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, *, target: float, interval: float, clock: Callable[[], float] = time.monotonic) -> None:
+        self._target = target
+        self._interval = interval
+        self._clock = clock
+
+        self._dropping = False
+        self._first_above_time: float | None = None
+        self._drop_next = 0.0
+        self._count = 0
+        self._last_count = 0
```
ajtmccarty, line 32: a lot of these names don't really describe the variable and it makes it hard to understand what each of these objects is really for
- `target` is more like a maximum of acceptable request time or a minimum of unacceptable request time. maybe `request_time_threshold_seconds` or something
- `dropping` I think `over_threshold` explains a little more about what is happening. could also be a `state` or `status` enum `NORMAL` / `OVER_THRESHOLD`
- `first_above_time` is more like `start_dropping_after_timestamp`
- `drop_next` is more like `drop_next_at_timestamp`
- `count` - > `total_dropped`
- `last_count` -> `previous_total_dropped`

backend/infrahub/api/admission/codel.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _control_law(self, reference: float, count: int) -> float:
```
ajtmccarty, line 34: this is `get_next_drop_timestamp`. `reference` should be `from_timestamp` or something. `count` is an instance variable, so does not need to be included

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,229 @@
...
+        self._stress_min_samples = stress_min_samples
+        self._retry_after = retry_after
+        # Source of the [0, 1) draw used to shed a fraction of a stressed class's requests.
+        self._rng = rng
+        # HIGH gets a larger effective target so it sheds last; MEDIUM and LOW share the base target.
+        self._codel: dict[Priority, CoDelController] = {
+            Priority.HIGH: CoDelController(target=target * high_target_multiplier, interval=interval, clock=clock),
+            Priority.MEDIUM: CoDelController(target=target, interval=interval, clock=clock),
+            Priority.LOW: CoDelController(target=target, interval=interval, clock=clock),
+        }
```
ajtmccarty: should be injected at construction, not built within the constructor.
you could 
- pass 3 arguments in: 1 for each priority
- a `dict` like this
- set priority as an instance variable on the `CoDelController` and add a `supports` method that determines which controller to apply to a request. so this class would accept `Sequence[CoDelController]` and then do something like `for controller in self.controllers: if controller.supports(req.priority): ...` we use this pattern elsewhere

backend/infrahub/api/admission/slot_pool.py
```diff
@@ -0,0 +1,148 @@
...
+    ``release`` returns the slot to the pool and is safe to call more than once; use it
+    in a ``finally`` so the slot is always returned even when the handler raises.
+    """
+
+    def __init__(self, *, priority: Priority, sojourn: float, pool: PrioritySlotPool) -> None:
+        self.priority = priority
+        """Priority class the slot was acquired for."""
+
+        self.sojourn = sojourn
+        """Seconds spent waiting to acquire the slot (``0.0`` on the fast path)."""
```
ajtmccarty, line 41: these should be comments or in an actual docstring

backend/infrahub/api/admission/slot_pool.py
```diff
@@ -0,0 +1,148 @@
...
+from collections import deque
+from typing import TYPE_CHECKING, Callable
+
+from .priority import Priority
+
+if TYPE_CHECKING:
+    from asyncio import Future
+
+
+class Acquisition:
```
ajtmccarty, line 28: seems more like a `PrioritySojourn` to me. it should be just those two data points.

backend/infrahub/api/admission/priority.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass(frozen=True)
+class PriorityHeaderParseResult:
+    """Outcome of parsing the priority request header.
+
+    Kept distinct from ``Priority`` so adoption can be recorded without a second parse:
+    ``was_explicit`` distinguishes a caller-supplied class from the ``MEDIUM`` default.
+    """
+
+    priority: Priority
+    was_explicit: bool
```
ajtmccarty, line 33: good name

backend/infrahub/api/admission/priority.py
```diff
@@ -0,0 +1,52 @@
...
+    """Classify a priority header value into a priority class.
+
+    The match is case-insensitive and ignores surrounding whitespace. A missing, empty,
+    whitespace-only, or unrecognized value resolves to ``MEDIUM`` with ``was_explicit=False``.
+    This never raises.
+    """
+    if header_value is None:
+        return PriorityHeaderParseResult(priority=Priority.MEDIUM, was_explicit=False)
+
+    normalized = header_value.strip().lower()
```
ajtmccarty: this could be `upper()`, then you don't need to `upper()` a `lower()` on the next line

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,229 @@
...
+    def release(self, *, acquisition: Acquisition) -> None:
+        """Return a served request's slot; the pool observer refreshes the live gauges.
+
+        The release flows through the pool, whose observer drives ``in_flight``/``waiters``
+        back down, so a finished request is reflected immediately rather than lingering
+        until the next admit.
+        """
+        acquisition.release()
+
+    def _sync_gauges(self, priority: Priority) -> None:
```
ajtmccarty: this looks like a sneak to skip doing dependency injection. the code currently passes this callback, `_sync_gauges`, into the dependent `PrioritySlotPool` so that the callback can access properties on the dependent `PrioritySlotPool`. it's unnecessarily confusing and sets reverses the dependency chain.

I think that the right approach is probably to pass the metrics into the `PrioritySlotPool` and have that class update the metrics.

backend/infrahub/api/admission/slot_pool.py
```diff
@@ -0,0 +1,148 @@
...
+
+        sojourn = self._clock() - enqueue_time
+        return self._make_acquisition(priority=priority, sojourn=sojourn)
+
+    def release(self) -> None:
+        """Return a slot and hand it to the highest-priority waiter, if any."""
+        self._available += 1
+        self._wake_up_next()
+
+    def _wake_up_next(self) -> None:
```
ajtmccarty, line 160: a little docstring here might be nice for those of us who aren't that familiar with Futures

backend/infrahub/api/admission/slot_pool.py
ajtmccarty: agree. this is a cycle in the dependency chain `Acquisition` <-> `PrioritySlotPool`. the chain should only go one way

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,229 @@
...
+    of database stress. Either firing sheds the request; when both fire the shed is attributed to
+    stress. The backstop is a third, unconditional per-class memory-safety cap that bounds the
+    waiter queue regardless of either signal.
+    """
+
+    def __init__(
+        self,
+        *,
+        slot_pool: PrioritySlotPool,
+        target: float,
```
ajtmccarty: more descriptive name would be helpful for readability

backend/infrahub/database/load_signal.py
```diff
@@ -0,0 +1,164 @@
...
+        self._samples: deque[tuple[float, float]] = deque()
+        # The same in-window values kept sorted: front is the minimum, middle is the median.
+        self._sorted: list[float] = []
+        self._observer: Callable[[ReferenceQueryLoadTracker], None] | None = None
+
+    def set_observer(self, observer: Callable[[ReferenceQueryLoadTracker], None] | None) -> None:
+        """Register a callback invoked after each recorded observation.
+
+        Lets an external sink (the metric gauges) refresh from the latest state without the
+        tracker depending on that sink.
```
ajtmccarty, line 82: the tracker **does** depend on the metrics. it is just pretending not to. I do not think we should use this pattern. it creates a loop in the dependency chain

backend/infrahub/database/load_signal.py
```diff
@@ -0,0 +1,164 @@
...
+    window_min = tracker.window_min()
+    if window_min is not None:
+        REFERENCE_QUERY_WINDOW_MIN_SECONDS.set(window_min)
+    REFERENCE_QUERY_STRESS_RATIO_MEDIAN.set(tracker.stress_ratio_median())
+
+
+# Process-global singleton fed by the database layer and read by the admission layer. Built
+# with the default window; the startup wiring overrides it from settings.
+reference_query_load_tracker = ReferenceQueryLoadTracker()
+reference_query_load_tracker.set_observer(_publish_metrics)
```
ajtmccarty: this is bad. it looks like what we do with `registry` and we have been trying to undo that for some time. this is always instantiated when the module is imported and it is an immediate dependency loop: `ReferenceQueryLoadTracker` <-> `metrics.REFERENCE...`

for a singleton, a pattern like `get_database()` is better than setting this as a module-level variable

backend/infrahub/database/load_signal.py
```diff
@@ -0,0 +1,164 @@
...
+    minimum ("floor") plus a rolling window of the most recent ``window_seconds`` of samples,
+    and derives how much slower the database currently is than at its best. A stress ratio of
+    ``5`` means the recent query time is five times the floor: sustained load.
+
+    The window uses the median rather than the mean for its central tendency: with sparse idle
+    traffic a single slow outlier (a GC pause, event-loop scheduling delay) would dominate a
+    mean, but barely moves a median.
+
+    State is per worker process and lives on a single asyncio event loop, so — like the
+    admission slot pool — it takes no lock: ``record`` never awaits, so no two updates
```
ajtmccarty: `admission slot pool` does not belong here

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:
+        self.app = app
+        if controller is None:
+            config.SETTINGS.initialize_and_exit()
+            self._controller = build_admission_controller()
+            self._enabled = config.SETTINGS.api.backpressure_enabled if enabled is None else enabled
+        else:
+            self._controller = controller
+            self._enabled = True if enabled is None else enabled
+        self._excluded_paths = excluded_paths
```
ajtmccarty: judging from the docstring, the init has these conditionals in it for testing purposes, but I don't really understand why. we should be able to correctly build this class when the app is starting up. then it can be correctly typed as `controller: AdmissionController, enabled: bool` instead of `| None`. it is not the responsibility of this middleware class to build `AdmissionController`

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,229 @@
...
+        until the next admit.
+        """
+        acquisition.release()
+
+    def _sync_gauges(self, priority: Priority) -> None:
+        metrics.IN_FLIGHT.labels(priority=priority.label).set(self._slot_pool.in_flight(priority=priority))
+        metrics.WAITERS.labels(priority=priority.label).set(self._slot_pool.waiters(priority=priority))
+
+
+def build_admission_controller() -> AdmissionController:
```
ajtmccarty: the settings required to build this class should be passed in to this method. or the class should be constructed at the application entry point where it is needed. the `AdmissionController` class should not be dependent on `settings`

ajtmccarty (commented):
a lot of naming and SOLID comments.
regardless of how many comments I left, this does look like a very cool bit of rate-limiting / metrics functionality

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,229 @@
...
+        # Once a slot is held, any exception before returning Admitted would strand it (the
+        # caller only releases what it receives), so guard the whole window and release on error.
+        try:
+            metrics.SOJOURN_SECONDS.labels(priority=priority.label).observe(acquisition.sojourn)
+
+            # Two independent shed signals, evaluated every request so CoDel keeps observing
+            # sojourn continuously. Database stress sheds a fraction of the class (a random draw
+            # against the escalating schedule); CoDel sheds on sojourn overrun. Stress is
+            # attributed first when both fire, so the stress dimension stays visible.
+            shed_fraction = self._stress_shed_fraction(priority=priority)
```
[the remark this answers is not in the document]
dgarros: Valid, fixed in `01bb58c`. The stress fraction and random draw are now evaluated *before* `acquire()`, so a stress-shed request returns its fast 429 without queueing behind a saturated pool or consuming waiter/backstop capacity. CoDel stays after acquisition since it keys off the measured sojourn. Added `test_stress_sheds_before_queueing_for_a_slot`.

backend/infrahub/database/load_signal.py
```diff
@@ -0,0 +1,170 @@
...
+        while self._samples and self._samples[0][0] <= horizon:
+            self._running_sum -= self._samples.popleft()[1]
+        while self._window_min and self._window_min[0][0] <= horizon:
+            self._window_min.popleft()
+
+    def floor(self) -> float | None:
+        """The all-time minimum observation, or ``None`` before any observation."""
+        return self._floor
+
+    def sample_count(self) -> int:
```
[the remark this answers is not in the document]
dgarros: Valid, fixed in `01bb58c`. The window now expires on read — `sample_count`, `window_min`, `window_median` and `stress_ratio_median` prune by wall-clock — so a stale stressed window can no longer shed the first request arriving after an idle gap, before it reaches the reference query. Added `test_reads_expire_stale_window_after_idle_without_recording`.

backend/infrahub/api/admission/slot_pool.py
dgarros: Agreed, done in `01bb58c`. `Acquisition` is now a plain data holder with no back-reference to the pool, and the pool owns release via `PrioritySlotPool.release(*, acquisition)`. The dependency runs one way now.

backend/infrahub/api/admission/controller.py
dgarros: Addressed in `01bb58c`. The pool now *pushes* its live in-flight/waiter counts to the observer instead of the callback reading them back off the pool, and gauge publishing is a module-level pure function (`_publish_slot_metrics`) rather than a bound method on the controller. Dependency is pool → sink only; the pool stays metrics-free.

backend/infrahub/database/load_signal.py
dgarros: Fixed in `01bb58c`. Replaced the import-time instance + observer wiring with a lazy `get_reference_query_load_tracker()` accessor, matching the `get_database()` pattern. Importing the module now has no side effect; the sink is attached on first use.

backend/infrahub/database/load_signal.py
dgarros: The import-time coupling — the real problem — is gone in `01bb58c`: the observer is now wired inside the lazy accessor, not at module import. I kept the observer indirection itself so the tracker doesn't import the metrics module directly, mirroring how the slot pool stays metrics-free. Happy to inline the dependency instead if you'd rather it be explicit.

backend/infrahub/api/admission/middleware.py
dgarros: Changed in `01bb58c` — replaced the `isinstance` chain with an exhaustive `match`/`assert_never`. I went with `match` over a method on a `Decision` base because the two branches do ASGI-specific work (run the app + release vs. write the 429), which I didn't want to push onto the decision dataclasses; the backend design rule also endorses `match` for closed unions. Let me know if you'd still prefer the ABC.

backend/infrahub/api/admission/slot_pool.py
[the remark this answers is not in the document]
dgarros: Good catch — fixed in `35bc33d`. `_notify` now wraps the observer in a try/except: it's a best-effort metrics sink invoked mid-transition, so a failing callback is logged and swallowed and can no longer skip the slot return (leaking a slot / stranding waiters) or surface to the caller. Added `test_failing_observer_does_not_corrupt_admission_state`, which asserts acquire/release still complete and the slot recovers when the observer raises on every transition.

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,136 @@
...
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http" or not self._enabled:
+            await self.app(scope, receive, send)
+            return
+
+        path = scope.get("path", "")
+        if any(path.startswith(excluded) for excluded in self._excluded_paths):
```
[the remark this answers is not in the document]
dgarros: Fixed in `def462a`. Matching is now exact-path-or-slash-delimited-descendant (`path == excluded or path.startswith(f"{excluded}/")`), so mounted trees like `/assets/…` and their roots stay excluded but `/healthcheck` no longer slips past `/health`. Verified every real excluded route is either exact (`/metrics`) or a descendant (`/api/schema/…`, `/assets/…`), so nothing legitimate regresses. Added `test_prefix_of_excluded_path_is_not_bypassed` (asserts `/healthcheck` and `/metrics-internal` reach the layer and are shed).

backend/infrahub/api/admission/capacity.py
```diff
@@ -0,0 +1,19 @@
...
+    Args:
+        pool_size: Number of connections in the process's database pool.
+        factor: Multiplier applied to ``pool_size``.
+
+    Returns:
+        The slot cap, floored at ``1`` so a pool/factor product below one still
+        admits a single request at a time.
+
+    """
+    return max(1, int(pool_size * factor))
```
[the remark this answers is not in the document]
dgarros: Fixed in `03f2b75`. Went with `allow_inf_nan=False` on `backpressure_max_concurrency_factor`, so `inf`/`nan` are rejected at config validation with a clear error rather than reaching `int(pool_size * factor)` and raising `OverflowError` at startup. No guard test — per the repo testing rule we don't test Pydantic field constraints — but confirmed manually that `inf` and `nan` now raise `ValidationError` while finite values still load.

backend/infrahub/database/load_signal.py
```diff
@@ -0,0 +1,164 @@
...
+
+        """
+        now = self._clock()
+        observation = max(execution_seconds, MIN_OBSERVATION_SECONDS)
+
+        if self._floor is None or observation < self._floor:
+            self._floor = observation
+
+        self._samples.append((now, observation))
+        bisect.insort(self._sorted, observation)
```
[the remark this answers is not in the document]
dgarros: Thanks — I looked at this closely but I'm going to leave it as is. The concern assumes an unbounded, high-rate hot path, but this list only grows with the *reference query*, which runs periodically (not per API request), and it's bounded by the rolling window (~20s), so in practice it holds on the order of tens–hundreds of small floats. `bisect.insort` / eviction on a list that size is a cheap C-level memmove — not an event-loop hazard — and the tradeoff is documented on the class. Swapping in an O(log n) structure (skip list / order-statistic tree) would add real complexity for no measurable gain at this scale. Happy to revisit if profiling ever shows this list growing large, but I don't think the change is justified today.

backend/infrahub/api/admission/codel.py
dgarros: I'm not a fan of this name too but CoDel is a real algorithm name defined in RFC 8289 so I think that make sense to keep it as it
https://datatracker.ietf.org/doc/html/rfc8289

docker-compose.yml
```diff
@@ -52,6 +66,7 @@ x-infrahub-config: &infrahub_config
   INFRAHUB_DB_DATABASE:
   INFRAHUB_DB_MAX_CONCURRENT_QUERIES: ${INFRAHUB_DB_MAX_CONCURRENT_QUERIES:-0}
   INFRAHUB_DB_MAX_CONCURRENT_QUERIES_DELAY: ${INFRAHUB_DB_MAX_CONCURRENT_QUERIES_DELAY:-0.01}
+  INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE: ${INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE:-50}
```
saltas888: is this intentional? Speckit says 100. IF its intentional would be nice to explain how this number was decided

backend/tests/component/api/test_admission_middleware.py
```diff
@@ -0,0 +1,693 @@
...
+    async def healthcheck() -> dict[str, bool]:
+        return {"ok": True}
+
+    @app.get("/metrics-internal")
+    async def metrics_internal() -> dict[str, bool]:
+        return {"ok": True}
+
+    app.add_middleware(AdmissionMiddleware, controller=_shed_everything_controller(), enabled=True)
+
+    offered_before = _offered_total()
```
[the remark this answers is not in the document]
dgarros: Good catch, that assertion was missing. Added `missing_before = metrics.MISSING_PRIORITY_TOTAL._value.get()` before the request and asserted the per-invocation delta is 1, matching the sibling tests, so the header-less request is now validated as processed end-to-end by the admission layer. Fixed in 611468f.

dev/specs/ifc-2886-priority-api-backpressure/research.md
```diff
@@ -0,0 +1,123 @@
...
+- Auth in Infrahub is a **FastAPI dependency** (`Depends(get_current_user)` in `backend/infrahub/api/dependencies.py`), resolved per-route *after* all middleware. So the admission middleware runs pre-auth and can only classify origin-blind on the header — which is exactly what FR-006 mandates ("classify solely on `X-Priority`"). No auth interaction is required.
+
+**Alternatives considered**:
+- `@app.middleware("http")` decorator (used by 3 existing middlewares): rejected — it wraps `BaseHTTPMiddleware`, which fully buffers the response and is a poor fit for a hot admission path.
+- A FastAPI dependency instead of middleware: rejected — dependencies run after routing/auth and per-route wiring; a middleware gives one uniform admission point across all endpoints.
+
+**Path exclusions**: The middleware MUST bypass admission for `/health` and `/metrics` (liveness and scraping must never be shed) and SHOULD bypass static/docs paths (`/assets`, `/favicons`, `/docs`, `/api/schema`) to match the existing `ConditionalGZipMiddleware` skip set. Non-`http` scopes (WebSocket, lifespan) pass through untouched.
```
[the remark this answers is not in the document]
dgarros: Thanks — both edge cases hold up, and here's where each lands:

1. **CORS OPTIONS preflight** is already handled in code: `AdmissionMiddleware.__call__` calls `_is_cors_preflight(scope)` (OPTIONS carrying `Access-Control-Request-Method`) and passes those straight through to the downstream CORS middleware before the gate, so a preflight is never shed. I've now reflected that bypass in the R1 path-exclusions note in `research.md` too, so the spec matches the code.

2. **M-1 disconnect gap** is real. The `OFFERED_TOTAL` increment is placed at the entry of `admit()`, before the slot acquire, deliberately — it must count true offered load regardless of what happens next. Rather than move it or add a counter (both change behaviour, out of scope for this spec), I've documented the resulting discrepancy as an explicit exception to M-1 in `contracts/metrics.md`: a client disconnecting while queued is counted in `offered_total` only, so under in-flight cancellation `offered_total` may transiently exceed `admitted + rejected` by the number of abandoned waiters (the `waiters`/`in_flight` gauges stay accurate).

Both doc edits are in commit 7c67b54.

docker-compose.yml
dgarros: this number was changed during the testing, with large query the DB would be overloaded very quickly. 
@fatih-acar what do you think ? should we keep it at 100?

backend/infrahub/api/admission/middleware.py
saltas888: I agree with this. Claude help  me discover that this is a must because we use it for the k8s and coker compose probes github.com/opsmill/infrahub-helm/blob/stable/charts/infrahub/values.yaml

backend/infrahub/api/admission/middleware.py
dgarros: Agreed — the middleware shouldn't own construction. Reworked in e0c27296f: the controller and kill-switch are now built once during startup and published on `app.state`; the middleware reads them per request and no longer builds anything, resolves settings, or carries the `| None` seam. It stays at startup (the lifespan) rather than eager construction at `add_middleware` because settings are deliberately initialized in the lifespan, not at import — building at import would fire config init on the ~10 test modules that import `app` at collection. Tests now set the two `app.state` attributes instead of injecting via the constructor.

docker-compose.yml
fatih-acar: image

here's a fresh trace over the connection pool metric when running the test scenario I'm working on
this is with the changes in this PR but with `INFRAHUB_DB_MAX_CONNECTION_POOL_SIZE=100`

changing the default from 50 to 100 didn't change the test behavior nor errors, and we seem to be below the 50 threshold

without even those test results, I'd be in favor of keeping this value of 50 because the connection pool on the [server side is 400](https://neo4j.com/docs/operations-manual/current/performance/bolt-thread-pool-configuration/#_configuration_options) by default and 50 fits better the default (50 * 4 gunicorn workers) + 4 task workers (task workers are not expected to use that much connections)

backend/infrahub/api/admission/controller.py
dgarros: Done in 27c4b3aeb — `AdmissionController` now takes the per-class CoDel controllers as an injected `codel: dict[Priority, CoDelController]` instead of building them in the constructor from `target`/`interval`/`high_target_multiplier`/`clock`. The factory builds the dict (HIGH given the larger effective target so it sheds last). Went with the dict option over `Sequence[CoDelController]` + `supports()` since the priority→controller relationship is a fixed, closed mapping — a direct dict lookup reads clearer here than iterating predicates.

backend/infrahub/api/admission/controller.py
dgarros: Done in 27c4b3aeb — `build_admission_controller` now takes `settings: config.Settings` as a parameter instead of reading the global `config.SETTINGS`; the lifespan passes `config.SETTINGS.active_settings`. Together with building the controller at the app entry point (e0c27296f), the factory no longer depends on global settings and `AdmissionController` itself stays settings-free.

bilalabbad (approved):

pa-lem (approved):

backend/infrahub/api/admission/middleware.py
dgarros: Good call — added `/api/config` to the admission exclusion set in a9a0ac9. It's a public no-auth endpoint used as a k8s/compose probe, so shedding it under load would fail probes exactly when we don't want pods restarting.

backend/infrahub/api/admission/middleware.py
```diff
@@ -0,0 +1,154 @@
...
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        # The controller and kill-switch are published on app.state during startup, before any
+        # request reaches this gate; a non-http (lifespan) scope short-circuits above, so state is
+        # only read once it is guaranteed present.
+        state = scope["app"].state
```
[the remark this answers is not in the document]
dgarros: Valid — reordered `__call__` in a9a0ac9 so the excluded-path and CORS-preflight checks run *before* reading `state.admission_enabled`. Probes and preflights now bypass the gate without depending on startup state, so they can't 500 if the lifespan wiring hasn't published `app.state` (added a test that an excluded path is served with no admission state set).

backend/infrahub/database/load_signal.py
dgarros: Agreed — dropped the `admission slot pool` cross-reference in 15fb634; the docstring now justifies the lock-free design on its own terms (per-worker process, single event loop, `record` never awaits). The database layer shouldn't name an admission-layer concrete.

docker-compose.yml
```diff
@@ -19,6 +19,20 @@ x-infrahub-config: &infrahub_config
   INFRAHUB_ANALYTICS_API_KEY:
   INFRAHUB_ANALYTICS_ENABLE: ${INFRAHUB_ANALYTICS_ENABLE:-true}
   INFRAHUB_ANONYMOUS_ACCESS_ROLE: ${INFRAHUB_ANONYMOUS_ACCESS_ROLE:-Anonymous User}
+  INFRAHUB_API_BACKPRESSURE_BACKSTOP_HIGH_MULTIPLIER: ${INFRAHUB_API_BACKPRESSURE_BACKSTOP_HIGH_MULTIPLIER:-4.0}
```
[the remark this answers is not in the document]
dgarros: Checked — none of these are dead config. All fifteen `INFRAHUB_API_BACKPRESSURE_*` vars map to fields on the API settings via pydantic-settings' prefix + field-name convention: the `INFRAHUB_API_` prefix selects `Settings.api`, and each suffix maps to a `backpressure_*` field there (`config.py:547-604`) — e.g. `BACKSTOP_HIGH_MULTIPLIER` -> `backpressure_backstop_high_multiplier`, `SHED_MEDIUM_STRESS_RATIO` -> `backpressure_shed_medium_stress_ratio`. The factory (`build_admission_controller`) reads them all.

dev/specs/archive/ifc-2886-priority-api-backpressure/contracts/metrics.md
dgarros: Yes, you've read it right. `sum(in_flight) <= max_concurrency` implies each per-class `in_flight{priority} <= max_concurrency` (a single term can't exceed the sum, which is bounded by the cap). The per-class bound is spelled out separately only so it's directly assertable per `priority` label rather than something a reader has to derive.

backend/infrahub/api/admission/priority.py
dgarros: Deliberately not logged. The header is untrusted client input, so logging on every unrecognized value would be a trivial log-flood vector. Invalid/missing values are instead counted in the `missing_priority_total` metric (incremented in the middleware, `middleware.py:91-92`) for adoption tracking — that's the intended signal, and it aggregates instead of emitting a line per request.

backend/infrahub/api/admission/middleware.py
dgarros: They return identical bodies today, kept as two functions on purpose: the REST and GraphQL error envelopes diverge elsewhere in the API, and the call site already routes by path, so a future GraphQL-shaped shed body would be a one-function change rather than a refactor of a shared builder. That said, it's scaffolding for a divergence that doesn't exist yet — happy to collapse them into one if you'd rather not carry it. Your call.

backend/tests/component/api/test_admission_middleware.py
```diff
@@ -0,0 +1,605 @@
...
+
+        async def fire(priority: str) -> int:
+            response = await client.get("/work", headers={"X-Priority": priority}, timeout=10)
+            return response.status_code
+
+        low_tasks = [asyncio.create_task(fire("low")) for _ in range(LOW_REQUESTS)]
+
+        # Let the pool saturate and sojourn climb past the CoDel interval before the
+        # interactive stream starts, so shedding is already active.
+        await asyncio.sleep(0.05)
```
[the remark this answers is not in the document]
dgarros: Leaving the sleep here. This is a component test driving the gate over HTTP (httpx/ASGI transport), so it has no handle on the controller's internal `PrioritySlotPool` — there's no `pool.waiters(...)` to poll the way the unit tests do; `_wait_until_waiting` lives at the unit layer where the pool is constructed directly. Under the LOW flood, saturation is reached well inside the 50ms buffer, and the key assertion (every HIGH request returns 200) isn't timing-sensitive.

backend/tests/component/api/test_admission_middleware.py
dgarros: The derivation itself is already unit-tested exhaustively in `backend/tests/unit/api/admission/test_capacity.py::test_derive_max_concurrency` (parametrized over the floor, fractional-truncation, identity and above-one cases). The assertion here checks the *wiring* — that the gauge ends up set from that derivation end-to-end — which the unit test can't reach. So it's intentionally at the component layer, not a gap.

backend/tests/unit/api/admission/test_controller.py
dgarros: No coupling to worry about — the production `admit()` path decides with `match`/`case` on `Admitted()`/`Rejected()`, not `isinstance`, and that's not changing. The `isinstance(result, Admitted)` here is just the test narrowing the returned union to assert on it; it stands on its own.

backend/tests/unit/api/admission/test_priority.py
dgarros: The ordering test guards a real contract, not stock `IntEnum` behaviour: the numeric values (`HIGH=0 < MEDIUM=1 < LOW=2`) *are* the priority order the slot pool depends on when it iterates `for priority in Priority` to serve the highest-priority waiter first. A reshuffle of those values would silently break shedding, and this test catches it. `test_priority_labels` covers the custom `.label` property (`name.lower()`), which is our code too. Keeping both.

backend/infrahub/api/admission/slot_pool.py
dgarros: These are attribute docstrings — a string literal placed right after the attribute assignment. It's a recognized pattern that IDEs and doc tooling surface on hover for that specific attribute, which an inline `#` comment doesn't. Keeping them as-is.

backend/infrahub/api/admission/slot_pool.py
dgarros: `Acquisition` is more than the two data points — it's the handle you hand back to `release()`, and it carries the `_released` idempotency guard that makes releasing in a `finally` safe even after an earlier release. The name reflects that it models a *held slot* (a resource you acquire and release), not just a sojourn measurement; `PrioritySojourn` would hide that lifecycle role. Keeping it.

backend/infrahub/api/admission/controller.py
dgarros: This was flagged on the old `target: float` parameter of `AdmissionController.__init__`, which no longer exists — the constructor now takes injected `codel: dict[Priority, CoDelController]` controllers rather than raw target/interval floats (refactored later on this branch). The name it pointed at is gone, so there's nothing to rename here.

backend/infrahub/api/admission/codel.py
dgarros: Keeping `codel.py`. CoDel (Controlled Delay) is the actual published name of the algorithm — RFC 8289 and the Nichols/Jacobson ACM Queue paper — and the class docstring spells it out (`CoDel (Controlled Delay)`). It reads as a typo only until you've met the algorithm; to anyone who has, `controlled_delay.py` would be the surprising name. Matching the reference term keeps the file searchable against the literature it implements.

backend/infrahub/api/admission/codel.py
dgarros: Keeping these names. They mirror the canonical CoDel reference (RFC 8289 appendix and the ns-2 / Linux `codel` sources): `target`, `interval`, `dropping`, `first_above_time`, `drop_next`, `count`, `last_count` are the algorithm's own terms of art. Renaming them to local descriptions would make the state machine harder to line up against the reference it implements — which for an algorithm like this is the bigger readability win. The plain-language explanation lives in the docstring and comments.

backend/infrahub/api/admission/codel.py
dgarros: Same reasoning as the state variables — `_control_law` and its `reference` argument track the CoDel reference's control-law formulation (`drop_next = reference + interval/sqrt(count)`). `count` is passed in explicitly (rather than read from `self`) to keep `_control_law` a pure function of its inputs, which is what makes it easy to unit-test in isolation. Keeping it aligned with the algorithm's vocabulary.

backend/infrahub/api/admission/codel.py
dgarros: Thanks — I'm keeping the current `ok_to_drop` structure. It maps 1:1 onto the CoDel reference's `should_drop()`: the arm-then-check split on `first_above_time` is exactly how the algorithm expresses "tolerate one full interval above target before dropping". Restructuring for local readability would drift the core control law from that reference, which I'd rather not do here. Appreciate the suggestion — noting it was flagged optional.

backend/infrahub/api/admission/codel.py
dgarros: Same call as the related comment — leaving the block order as-is to stay aligned with the CoDel reference's control flow, where the dropping-state machinery and the arm/check step are ordered the way the algorithm defines them. Noting it was optional.

backend/infrahub/api/admission/codel.py
dgarros: Follows from the previous two: since I'm keeping the `ok_to_drop` structure to stay aligned with the CoDel reference, this block and the final return stay as well. Keeping as-is.

backend/tests/component/api/test_cors_priority.py
dgarros: Good question — and no, a shed `429` to a real cross-origin request does not carry CORS headers. `AdmissionMiddleware` is the outermost layer (it sheds before anything downstream runs), so a shed response never reaches the CORS middleware, and a browser surfaces it as a CORS error rather than a readable 429. That's an accepted limitation for v1: the preflight bypass keeps normal cross-origin traffic working, and the shed path is a transport-level load-shed where the browser honoring `Retry-After` matters more than JS reading the exact status. Worth revisiting if we later want first-party browser clients to react to the 429 specifically.

dev/specs/ifc-2886-priority-api-backpressure/contracts/metrics.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Histogram buckets** (`infrahub_admission_sojourn_seconds`): `[0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 5]` (seconds; fine at the low end where the CoDel `target` sits).
+
+## Invariants (assertion-testable)
+
+| ID | Invariant |
+|----|-----------|
+| M-1 | For each class, for every request the server adjudicates: `offered_total == admitted_total + rejected_total{codel} + rejected_total{backstop}`. `offered_total` is incremented at the entry of `admit()`, before the slot acquire, by design — so it counts true offered load regardless of what happens next. The consequence is a documented exception to the equality: a request whose client disconnects while it is still queued (the acquire await is cancelled) is counted in `offered_total` only — it is neither admitted nor shed by the server — so under in-flight cancellation `offered_total` may transiently exceed `admitted + rejected` by the number of abandoned waiters. The `waiters`/`in_flight` gauges stay accurate across that case (they are driven by the pool's own enqueue/dequeue transitions, including cancellation). |
```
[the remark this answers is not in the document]
dgarros: Fixed in 48af6e6 — M-1 now reads `offered_total == admitted_total + rejected_total{codel} + rejected_total{backstop} + rejected_total{stress}`, matching the three `RejectionReason` values the controller actually emits. Also added `stress` to the `rejected_total` reason enumeration elsewhere in metrics.md and to the shed-reasons line in x-priority-header.md so the contracts agree.

dev/specs/ifc-2886-priority-api-backpressure/contracts/x-priority-header.md
```diff
@@ -0,0 +1,81 @@
+# Contract: `X-Priority` request header & `429` admission response
+
+**Feature**: IFC-2886 | Applies to: every HTTP endpoint served by the Infrahub API (except the excluded liveness/scrape/static paths below).
+
+This is a **transport-layer contract**. It adds one request header and one possible response outcome. It does **not** change any REST/GraphQL request body, GraphQL schema, or existing 2xx/4xx/5xx semantics of the handlers themselves.
+
+> **Naming note**: the middle tier was renamed `normal` → `medium` after this contract was written. The shipped header values are `high` / `medium` / `low`; read `normal` as `medium` below.
```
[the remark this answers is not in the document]
dgarros: Done in 48af6e6 — replaced `normal` with `medium` throughout the contract (allowed values, defaults, classification rules, C-1/C-4) and deleted the naming-note band-aid, so the doc uses the shipped tier name directly.

dev/specs/ifc-2886-priority-api-backpressure/research.md
```diff
@@ -0,0 +1,125 @@
...
+
+**CORS preflight bypass**: A CORS preflight (`OPTIONS` carrying `Access-Control-Request-Method`) MUST also bypass the gate. It carries no `X-Priority` and would classify as `normal`; shedding it under load would strip the CORS response and break every cross-origin request precisely when the backend is busy. Preflights therefore pass straight through to the downstream CORS middleware and are never counted as offered load.
```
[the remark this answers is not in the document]
dgarros: Fixed in 48af6e6 — the CORS-preflight note now says a missing `X-Priority` classifies as `medium`, matching `parse_priority` returning `Priority.MEDIUM`. Cleaned up the other `normal` tier stragglers in research.md too.

backend/tests/unit/api/admission/test_slot_pool.py
dgarros: Done in b561c75 — pulled `FakeClock` into a shared `backend/tests/unit/api/admission/helpers.py`, imported by both `test_slot_pool.py` and `test_codel.py`. (Left `_StepClock` and the database test's own `FakeClock` alone — different classes/packages.)

backend/infrahub/api/admission/priority.py
dgarros: Done in a04df60 — `strip().upper()` once, then `Priority[normalized]` directly.

backend/infrahub/api/admission/slot_pool.py
dgarros: Done in a04df60 — added a docstring to `_wake_up_next` describing the priority-order scan and how the freed slot is handed off by resolving the first not-yet-done `Future`.

backend/infrahub/api/admission/codel.py
ajtmccarty: I disagree with naming things poorly based on an RFC or paper b/c it requires reading an RFC or a paper, but I'm willing to accept that the names are following a standard of some sort

backend/infrahub/api/admission/controller.py
```diff
@@ -0,0 +1,234 @@
...
+    keys off the measured sojourn, so it only runs once a slot is held — a request shed by stress
+    never reaches it. The backstop is a third, unconditional per-class memory-safety cap that
+    bounds the waiter queue regardless of either signal.
+    """
+
+    def __init__(
+        self,
+        *,
+        slot_pool: PrioritySlotPool,
+        codel: dict[Priority, CoDelController],
```
ajtmccarty: `codel_priority_map` is a better name

backend/infrahub/api/admission/middleware.py
polmichel: This option looks fine to me!

backend/infrahub/api/admission/middleware.py
polmichel: Would tend to follow the YAGNI principle and collapse these methods. But this is a nitpick comment in my opinion, it's fine kept as-is.

backend/tests/component/api/test_admission_middleware.py
polmichel: Yes you're right I've seen them later, but forgot to update this comment.
Not sure the assert is required then. But this is also a nitpick comment, you don't have to change it if you don't want to.

polmichel (approved):

backend/infrahub/database/load_signal.py
ajtmccarty: the `set_observer` pattern is still breaking the dependency chain. it requires two entities to have knowledge of one another. I think that the correct pattern (both here and in `PrioritySlotPool`) is to have an `ObserverInterface` that can be passed into `ReferenceQueryLoadTracker` and `PrioritySlotPool` and then the concrete implementation of the `ObserverInterface` (that actually uses the metrics gauges) is only imported where these classes are constructed.

this branch shows what I have in mind https://github.com/opsmill/infrahub/compare/dga/feat-rate-limiting-api-zb38x...ajtm-rate-limiting-metric-injection

backend/infrahub/api/admission/retry_policy.py
```diff
@@ -0,0 +1,86 @@
...
+        A ratio at or above the significant-load line starts the episode (if not already running);
+        anything below resets it. Ratios in the warm-up zone therefore never accrue sustained time.
+        """
+        now = self._clock()
+        if ratio >= self._significant_load_ratio:
+            if self._overload_since is None:
+                self._overload_since = now
+        else:
+            self._overload_since = None
+        metrics.SUSTAINED_LOAD_SECONDS.set(self._sustained_seconds(now))
```
ajtmccarty: I think that just importing and using the metric like this is easier to read and understand, but it doesn't fit the pattern of the other classes that update metrics. also I think that `notify` or `update_metrics` or something makes more sense than`observe` for this one if it continues to update the metric directly

ajtmccarty (commented):
still reviewing tests

backend/tests/component/api/test_admission_middleware.py
```diff
@@ -0,0 +1,752 @@
...
+        statuses = await asyncio.gather(*[fire() for _ in range(burst)])
+        elapsed = time.monotonic() - start
+
+    rejected_after = _rejected_total()
+
+    # The burst really did finish inside one interval, so no drop was even possible.
+    assert elapsed < interval
+    # Every request was served; none was shed.
+    assert statuses == [200] * burst
+    assert rejected_after - rejected_before == 0
```
ajtmccarty, line 391: interesting way to assert `rejected_after == rejected_before`

ajtmccarty (approved):
one primary outstanding comment about how the metrics should be injected into the components using them

backend/infrahub/api/admission/retry_policy.py
dgarros: Good call — moved it to the same metrics-free-with-sink pattern as `PrioritySlotPool` and the load tracker: the policy pushes the sustained-load seconds to a registered observer (`set_observer`), and the controller wires `_publish_sustained_load_metric`. Kept `observe` since it now only observes the ratio and no longer writes the metric itself. `cb1b2578`

backend/infrahub/api/admission/controller.py
dgarros: Renamed to `codel_priority_map` (constructor param + attribute + all call sites). `cb1b2578`

PR #9882, opened 2026-07-13
fix: restore account_id on outbound webhook payloads (closes #9881)

### LINKED ISSUE #9881: bug: outbound webhook payloads have account_id=None
## What happens

Every outbound webhook request emitted since upgrading from 1.9.6 to 1.10.x carries `account_id: null`, regardless of which account triggered the event. Sample payload from a `infrahub.branch.created` event:

    {
        "data": {"branch_id": "18c0f7a2-...", "branch_name": "test4", "sync_with_git": false},
        "id": "673a2d93-a9e0-464a-a635-aba2b42ff2b0",
        "branch": null,
        "account_id": null,
        "occured_at": "2026-07-10 15:35:29.674921+00:00",
        "event": "infrahub.branch.created"
    }

## What should happen

The webhook payload's `account_id` should be the ID of the account that triggered the event.

## Steps to reproduce

1. Configure a webhook subscribed to any event (e.g. branch creation).
2. As an authenticated account, trigger the event (create a branch).
3. Inspect the delivered webhook payload → `account_id` is `null`.

## Environment

- Infrahub 1.10.x (regression from 1.9.6)
- Backend / webhooks

## Notes

The event context was restructured so the triggering account's ID now serializes at the top level of the event payload's `context` block, while the webhook payload builder still reads it from a nested `context["account"]["account_id"]` path that no longer exists. Branch extraction is unaffected, which is why only `account_id` regressed.

polmichel:
# Why

Since 1.9.6 → 1.10.x, every outbound webhook payload carried `account_id: null` regardless of which account triggered the event, dropping attribution for all webhook consumers.

Closes #9881

## What changed

- Outbound webhook payloads again include the `account_id` of the triggering account.

The event-context refactor moved `account_id` to the top level of the event payload's `context` block and removed the nested `account` object. The webhook payload builder still read the old nested path (`context["account"]["account_id"]`), which no longer exists, so it always resolved to `None`. It now reads the top-level `account_id`.

Branch extraction was unaffected and is unchanged. No schema or event-model changes.

## How to review

Single change in `WebhookEventContext.from_event` (`backend/infrahub/webhook/models.py`). The added unit tests build the payload from the real `EventContext.to_event()` serialization, so they stay coupled to the actual contract.

## How to test

```bash
uv run pytest backend/tests/unit/webhook/test_models.py -q
```

Verified the new `from_event` account_id test fails without the fix and passes with it.

The functional payload-transform test exercises the full delivery flow against the testcontainers stack:

```bash
uv run pytest --neo4j "backend/tests/functional/webhook/test_process.py::TestWebhookProcess::test_webhook_check_payload_transform"
```

### Live end-to-end verification

Validated on a local stack built from this branch (`dev.deps` → single-platform image build → `dev.start`), through the real Prefect `webhook-process` → `webhook-send` delivery path:

1. Logged in as `admin` and created a `CoreCustomWebhook` pointing at a local HTTP receiver.
2. Created a branch as `admin`, firing `infrahub.branch.created`.
3. The receiver captured the delivered payload, whose `account_id` equalled the `admin` account id — the same event shape as the reported regression, now carrying the triggering account instead of `null`.

Confirmed the running task-worker executed this branch's image before triggering, so the observed payload reflects the fixed code.

---
## Summary by cubic
Restores `account_id` in outbound webhook payloads by reading it from the top-level `context.account_id` after the event-context refactor. Fixes the 1.10.x regression that sent `account_id: null`.

- **Bug Fixes**
  - Read from `context.account_id` instead of `context.account.account_id`.
  - Tests: unit checks using `SourceEventContext.to_event()`; functional fixture now derives `context` via `EventContext.to_event()`; verifies global branch is `None`.
  - No schema changes; branch extraction unchanged.

<sup>Written for commit bc74427feea113006d47b928057b0fa2ac154eb5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/webhook/models.py
+++ b/backend/infrahub/webhook/models.py
@@ -191,14 +191,13 @@ class EventContext(BaseModel):
     def from_event(cls, event_id: str, event_type: str, event_occured_at: str, event_payload: dict[str, Any]) -> Self:
         """Extract the context from the raw event we are getting from Prefect."""
         infrahub_context: dict[str, Any] = event_payload.get("context", {})
-        account_info: dict[str, Any] = infrahub_context.get("account", {})
         branch_info: dict[str, Any] = infrahub_context.get("branch", {})
 
         return cls(
             id=event_id,
             # We use `GLOBAL_BRANCH_NAME` constant instead of `registry.get_global_branch().name` to the flow from depending on the registry
             branch=branch_info.get("name") if branch_info and branch_info.get("name") != GLOBAL_BRANCH_NAME else None,
-            account_id=account_info.get("account_id"),
+            account_id=infrahub_context.get("account_id"),
             occured_at=event_occured_at,
             event=event_type,
         )
--- a/backend/tests/functional/webhook/conftest.py
+++ b/backend/tests/functional/webhook/conftest.py
@@ -24,19 +24,9 @@ if TYPE_CHECKING:
 
 BRANCH_CREATED_PAYLOAD: dict[str, Any] = {
     "context": {
-        "event": {
-            "id": "24790022-2bc8-42ab-a447-bf3e84675901",
-            "name": "infrahub.branch.created",
-            "ancestors": [],
-            "parent_id": None,
-        },
         "branch": {"id": "182853ef-58a3-b3cc-3e80-c5161f4171c1", "name": "-global-"},
-        "account": {
-            "auth_type": "api",
-            "account_id": "182853f2-3a43-c7f9-3e84-c5152eff4b17",
-            "session_id": None,
-            "authenticated": None,
-        },
+        "account_id": "182853f2-3a43-c7f9-3e84-c5152eff4b17",
+        "parent_event": None,
     }
 }
 
--- a/backend/tests/functional/webhook/conftest.py
+++ b/backend/tests/functional/webhook/conftest.py
@@ -23,8 +23,6 @@ if TYPE_CHECKING:
     from infrahub.database import InfrahubDatabase
 
 
-# Built from the source event context so the fixture tracks the real serialization
-# instead of restating it as a literal that can silently drift.
 BRANCH_CREATED_PAYLOAD: dict[str, Any] = {
     "context": EventContext(
         branch=EventBranchContext(name=GLOBAL_BRANCH_NAME, id="182853ef-58a3-b3cc-3e80-c5161f4171c1"),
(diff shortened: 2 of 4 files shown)


backend/infrahub/webhook/models.py
```diff
@@ -191,14 +191,13 @@ class EventContext(BaseModel):
...
         infrahub_context: dict[str, Any] = event_payload.get("context", {})
-        account_info: dict[str, Any] = infrahub_context.get("account", {})
         branch_info: dict[str, Any] = infrahub_context.get("branch", {})
 
         return cls(
             id=event_id,
             # We use `GLOBAL_BRANCH_NAME` constant instead of `registry.get_global_branch().name` to the flow from depending on the registry
             branch=branch_info.get("name") if branch_info and branch_info.get("name") != GLOBAL_BRANCH_NAME else None,
-            account_id=account_info.get("account_id"),
+            account_id=infrahub_context.get("account_id"),
```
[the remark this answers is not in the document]
polmichel: This has been fixed by using the dynamic construction in the integration test and not a plain text input

backend/tests/functional/webhook/conftest.py
```diff
@@ -22,22 +23,13 @@
...
-            "parent_id": None,
-        },
-        "branch": {"id": "182853ef-58a3-b3cc-3e80-c5161f4171c1", "name": "-global-"},
-        "account": {
-            "auth_type": "api",
-            "account_id": "182853f2-3a43-c7f9-3e84-c5152eff4b17",
-            "session_id": None,
-            "authenticated": None,
-        },
-    }
```
polmichel, line 40: This payload was at stale. I've plugged directly the `EventContext` serialization so it does not happen anymore.

backend/tests/functional/webhook/conftest.py
```diff
@@ -22,22 +23,13 @@
     from infrahub.database import InfrahubDatabase
 
 
+# Built from the source event context so the fixture tracks the real serialization
+# instead of restating it as a literal that can silently drift.
```
gmazoyer: Probably useless comment

gmazoyer (approved):

backend/tests/functional/webhook/conftest.py
polmichel: Good catch, thanks!

PR #9766, opened 2026-06-30
feat(webhook): log the delivery attempt, request headers, and payload

polmichel:
## Summary

Enrich the webhook delivery run logs so an operator can see, for every attempt:

- the attempt number (`attempt N/4`) on the request, success, and failure log lines
- the outgoing request: target URL, the headers with secret values masked, and the payload (truncated inline, logged in full at debug level)

Environment-sourced header values, the request signature, and well-known credential headers (`Authorization`, `Cookie`, `X-API-Key`, …) are masked regardless of how the header was configured, matched case-insensitively; other standard and static headers, and the payload, are kept verbatim.

The attempt number is read once from the flow run and passed into the send task, rather than read from the Prefect runtime inside the task.

Resolves [IFC-2832](https://opsmill.atlassian.net/browse/IFC-2832) and [IFC-2833](https://opsmill.atlassian.net/browse/IFC-2833).

## Example log output

Take a signed `StandardWebhook` named `notify-slack`, pointing at `https://hooks.example.com/services/T000`, configured with a static header `X-Custom-Source: infrahub` and an environment-sourced header `X-Api-Token` (reading `SLACK_TOKEN`). The signature and the env-sourced token are masked; the static header and the standard headers are kept.

**Attempt succeeds on the first try:**

```
INFO  Webhook 'notify-slack' attempt 1/4: POST https://hooks.example.com/services/T000 with headers {'Accept': 'application/json', 'Content-Type': 'application/json', 'X-Custom-Source': 'infrahub', 'X-Api-Token': '***', 'webhook-id': 'msg_1f0a…', 'webhook-timestamp': '1751459200', 'webhook-signature': '***'} and payload {"data": {"name": "device-01"}, "event_id": "…", "event_type": "infrahub.node.updated", …}
INFO  Webhook delivered to https://hooks.example.com/services/T000 attempt 1/4, HTTP 200 in 143 ms
```

**Target returns a 5xx, so the attempt fails and a retry is scheduled** (`status_class` drives the remediation hint; the retry note is appended because attempts remain):

```
INFO   Webhook 'notify-slack' attempt 2/4: POST https://hooks.example.com/services/T000 with headers {…} and payload {…}
ERROR  Webhook delivery failed [HTTP_SERVER_ERROR] attempt 2/4 after 88 ms: The target responded with HTTP 503. The target returned a server error; retry, or check the target. Retrying in 120s (attempt 3/4).
```

[...]

PR #9887, opened 2026-07-13
feat(webhook): format the delivery logs for readability

polmichel:
## Summary

Follow-up to the webhook delivery logging: make the delivery logs easy to read. Stacked on `pmi-20260618-ifc-2119-resend-webhook` (PR #9810).

Resolves [IFC-2905](https://opsmill.atlassian.net/browse/IFC-2905).

The outgoing-request log is laid out over labeled lines instead of one dense string, so an operator can pick out each element at a glance:

```
Webhook 'notify-slack' attempt 1/4
POST https://hooks.example.com/services/T000
Headers:
  Accept: application/json
  X-Api-Token: ***
  webhook-signature: ***
Payload:
  {
    "data": { "name": "device-01" },
    "event_type": "infrahub.node.updated"
  }
```

- Headers are printed one per line; secret values stay masked (`***`).
- The payload is pretty-printed (indented JSON) and truncated inline; the full untruncated payload is emitted at debug level.
- Success and classified-failure lines remain single readable sentences.

## What changed

- **`WebhookLogFormatter`** (`backend/infrahub/webhook/log_formatter.py`) — a dedicated component whose sole responsibility is the wording and layout of every delivery log line. It returns strings only (emitting stays in the flow) and does not redact (the caller passes already-masked headers). The send flow builds its messages through it.
- **`test_log_formatter.py`** — unit tests that specify the expected output: attempt phrasing, the labeled layout, headers one-per-line, `(none)` when empty, pretty-printed payload, inline truncation + overflow marker, untruncated debug payload, and the failure line's class/reason/remediation + retry note.
- **Frontend** (`logs.tsx`) — render task log messages with their whitespace preserved so the multi-line layout shows (same font as before), and migrate the rows to the Table's `{ value, display }` shape, which restores the severity and timestamp columns (previously rendered `-`).

## Verification

- Backend and frontend exercised end-to-end against a live stack: triggered a real `webhook-send` delivery and confirmed the stored task logs render un-stacked in the UI, including the debug full-payload line with debug logging enabled.
- `WebhookLogFormatter` and webhook flow unit tests pass; ruff, `ty`, and Biome clean; Betterer improved (removed one tracked `logs.tsx` type error).

## Out of scope

The `debug` severity has no badge in the logs table (`getSeverityBadge`), so debug rows show an empty Severity cell — tracked as a separate follow-up.

---
## Summary by cubic
Improves webhook delivery logs for quick scanning by laying out outgoing requests over labeled lines with masked headers and pretty JSON. The logs UI preserves multi-line formatting and restores the Severity and Timestamp columns; the formatter is now built lazily via a cached factory.

- New Features
  - Added `WebhookLogFormatter` and switched the send flow to use it; messages include the attempt, POST URL, per-line masked headers, pretty-printed JSON, a debug full-payload line, and clear success/failure lines with status, time, and retry notes. The formatter is constructed lazily with a cached factory to avoid import-time work.
  - Updated `logs.tsx` to use the Table `{ value, display }` shape and render messages with preserved whitespace; Severity and Timestamp columns are restored.

- Bug Fixes
  - Fall back to the payload’s plain representation with a warning when JSON serialization fails, so logging never blocks deliveries.
  - Apply the inline character limit after indentation so the shown payload stays within the configured bound.

<sup>Written for commit c43bd0b4cf6ba7fa21aa74528643d4206fcb72f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2905]: https://opsmill.atlassian.net/browse/IFC-2905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/webhook/log_formatter.py
+++ b/backend/infrahub/webhook/log_formatter.py
@@ -1,0 +1,86 @@
+from __future__ import annotations
+
+from typing import Any
+
+import ujson
+
+PAYLOAD_LOG_LIMIT: int = 2048  # characters shown inline; the full payload is logged at debug level
+
+
+class WebhookLogFormatter:
+    """Builds the log messages emitted during a webhook delivery.
+
+    Its sole responsibility is the wording and layout of those messages, laid out over several
+    labeled lines so an operator can find the request target, headers, payload, outcome, and retry
+    position at a glance. It only returns strings; emitting them is left to the caller. Header
+    redaction is not done here: the caller passes headers that are already masked.
+    """
+
+    def __init__(self, *, attempts: int, retry_delay_seconds: float, payload_log_limit: int) -> None:
+        self._attempts = attempts
+        self._retry_delay_seconds = retry_delay_seconds
+        self._payload_log_limit = payload_log_limit
+
+    def attempt_phrase(self, attempt: int | None) -> str:
+        """Position a send within its retry sequence, or note that no flow run is driving retries."""
+        if attempt is None:
+            return "outside a flow run"
+        return f"attempt {attempt}/{self._attempts}"
+
+    def outgoing_request(
+        self, *, webhook_name: str, url: str, headers: dict[str, Any], payload: Any, attempt: int | None
+    ) -> str:
+        """Summarize the outgoing request: target, headers one per line, and the truncated payload."""
+        return (
+            f"Webhook '{webhook_name}' {self.attempt_phrase(attempt)}\n"
+            f"POST {url}\n"
+            f"Headers:\n{self._headers_block(headers)}\n"
+            f"Payload:\n{self._indent(self._truncate(self._to_json(payload)))}"
+        )
+
+    def full_payload(self, *, webhook_name: str, payload: Any, attempt: int | None) -> str:
+        """Render the full, untruncated payload for the debug-level line."""
+        return (
+            f"Webhook '{webhook_name}' {self.attempt_phrase(attempt)} full payload:\n"
+            f"{self._indent(self._to_json(payload))}"
+        )
+
+    def delivery_succeeded(self, *, url: str, status_code: int, attempt: int | None, elapsed_ms: float) -> str:
+        return f"Webhook delivered to {url} {self.attempt_phrase(attempt)}, HTTP {status_code} in {elapsed_ms:.0f} ms"
+
+    def delivery_failed(
+        self, *, status_class: str, message: str, remediation: str, attempt: int | None, elapsed_ms: float
+    ) -> str:
+        return (
+            f"Webhook delivery failed [{status_class}] {self.attempt_phrase(attempt)} "
+            f"after {elapsed_ms:.0f} ms: {message.rstrip('.')}. {remediation}{self._retry_note(attempt)}"
+        )
+
+    def _retry_note(self, attempt: int | None) -> str:
+        """State when the next attempt runs, or that none remain. Empty when no flow run drives retries."""
+        if attempt is None:
+            return ""
+        if attempt < self._attempts:
+            return f" Retrying in {self._retry_delay_seconds:.0f}s (attempt {attempt + 1}/{self._attempts})."
+        return " No retries remaining."
+
+    def _headers_block(self, headers: dict[str, Any]) -> str:
+        if not headers:
+            return "  (none)"
+        return "\n".join(f"  {key}: {value}" for key, value in headers.items())
+
+    def _truncate(self, text: str) -> str:
+        if len(text) <= self._payload_log_limit:
+            return text
+        remaining = len(text) - self._payload_log_limit
+        return (
+            f"{text[: self._payload_log_limit]}… (+{remaining} characters; enable debug logging for the full payload)"
+        )
+
+    @staticmethod
+    def _to_json(payload: Any) -> str:
+        return ujson.dumps(payload, indent=2)
+
+    @staticmethod
+    def _indent(text: str) -> str:
+        return "\n".join(f"  {line}" for line in text.splitlines())
--- a/backend/infrahub/webhook/log_formatter.py
+++ b/backend/infrahub/webhook/log_formatter.py
@@ -1,8 +1,11 @@
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 import ujson
+
+logger = logging.getLogger(__name__)
 
 PAYLOAD_LOG_LIMIT: int = 2048  # characters shown inline; the full payload is logged at debug level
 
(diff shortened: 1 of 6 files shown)


backend/infrahub/webhook/tasks/process.py
```diff
@@ -41,21 +42,11 @@
...
-def _attempt_phrase(attempt: int | None) -> str:
-    """Position this send within its retry sequence, or note that no flow run is driving retries."""
-    if attempt is None:
-        return "outside a flow run"
-    return f"attempt {attempt}/{WEBHOOK_SEND_ATTEMPTS}"
+_LOG_FORMATTER = WebhookLogFormatter(
+    attempts=WEBHOOK_SEND_ATTEMPTS,
+    retry_delay_seconds=WEBHOOK_SEND_RETRY_DELAY_SECONDS,
+    payload_log_limit=PAYLOAD_LOG_LIMIT,
+)
```
polmichel: This component cannot be injected in the current state since there is no object to inject it into. As webhooks may be refactored into classic events, refactoring this whole file is out of the current scope.

frontend/app/src/entities/tasks/ui/logs.tsx
```diff
@@ -39,9 +39,18 @@ export const Logs = ({ logs }: LogsProps) => {
...
-      ...log,
-      severity: getSeverityBadge[log.severity],
-      timestamp: <DateDisplay date={log.timestamp} />,
+      message: {
+        value: log.message,
+        display: <span className="whitespace-pre-wrap break-words">{log.message}</span>,
+      },
+      severity: {
+        value: log.severity,
+        display: getSeverityBadge[log.severity],
```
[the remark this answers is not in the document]
polmichel: Preexisting and out of scope

backend/infrahub/webhook/log_formatter.py
```diff
@@ -0,0 +1,86 @@
...
+        if attempt is None:
+            return ""
+        if attempt < self._attempts:
+            return f" Retrying in {self._retry_delay_seconds:.0f}s (attempt {attempt + 1}/{self._attempts})."
+        return " No retries remaining."
+
+    def _headers_block(self, headers: dict[str, Any]) -> str:
+        if not headers:
+            return "  (none)"
+        return "\n".join(f"  {key}: {value}" for key, value in headers.items())
```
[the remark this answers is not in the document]
polmichel: Key and value are configured inside the application. So the risk is low I think

backend/infrahub/webhook/log_formatter.py
```diff
@@ -0,0 +1,98 @@
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import ujson
+
+logger = logging.getLogger(__name__)
+
+PAYLOAD_LOG_LIMIT: int = 2048  # characters shown inline; the full payload is logged at debug level
```
ajtmccarty, line 10: should this be a config value? truly don't know, just something to consider

backend/infrahub/webhook/tasks/process.py
ajtmccarty: I think this should be a singleton behind a `get_webhook_log_formatter` (or something) function. setting it as a constant means it is initialized whenever this file is imported. it doesn't take much to initialize right now, but it could grow in the future and it's not best practice

backend/infrahub/webhook/log_formatter.py
```diff
@@ -0,0 +1,98 @@
...
+    position at a glance. It only returns strings; emitting them is left to the caller. Header
+    redaction is not done here: the caller passes headers that are already masked.
+    """
+
+    def __init__(self, *, attempts: int, retry_delay_seconds: float, payload_log_limit: int) -> None:
+        self._attempts = attempts
+        self._retry_delay_seconds = retry_delay_seconds
+        self._payload_log_limit = payload_log_limit
+
+    def attempt_phrase(self, attempt: int | None) -> str:
```
ajtmccarty: looks like it should be private

backend/tests/unit/webhook/test_log_formatter.py
```diff
@@ -0,0 +1,221 @@
...
+
+    payload_section = message.split("Payload:\n", 1)[1]
+    unindented = "\n".join(line[2:] for line in payload_section.splitlines())
+    assert unindented.startswith('{\n  "blob": "' + "x" * 100)
+    assert unindented.endswith("characters; enable debug logging for the full payload)")
+    assert "… (+" in unindented
+
+
+def test_outgoing_request_does_not_truncate_a_payload_at_the_limit(formatter: WebhookLogFormatter) -> None:
+    small = WebhookLogFormatter(attempts=ATTEMPTS, retry_delay_seconds=RETRY_DELAY_SECONDS, payload_log_limit=100)
```
ajtmccarty: is this really "at the limit" as the name of the test suggests? looks like the limit is 100 and the payload is significantly less than that

backend/tests/unit/webhook/test_log_formatter.py
```diff
@@ -0,0 +1,221 @@
...
+class _Unserializable:
+    def __repr__(self) -> str:
+        return "<unserializable>"
+
+
+def test_outgoing_request_falls_back_to_plain_repr_when_payload_is_not_json_serializable(
+    formatter: WebhookLogFormatter, caplog: pytest.LogCaptureFixture
+) -> None:
+    payload = _Unserializable()
+    with pytest.raises((TypeError, ValueError)) as exc_info:
```
ajtmccarty: seems like this should be predictable, either `Type` or `ValueError`

ajtmccarty (approved):

backend/infrahub/webhook/log_formatter.py
polmichel: The goal behind is to limit the length of the payload displayed in the task logs (you can see an example here https://github.com/opsmill/infrahub/pull/9766).
The full payload is still accessible inside debug logs, so I think this is not required to add this parameter to global configuration options.

backend/tests/unit/webhook/test_log_formatter.py
polmichel: True, I change that

backend/infrahub/webhook/log_formatter.py
polmichel: Good point, there are some unit tests attached to it though, I'll change them so they use the public ways of interacting with the object

backend/tests/unit/webhook/test_log_formatter.py
polmichel: Good point this is misleading, I changed that

PR #9901, opened 2026-07-14
fix(schema): handle node removal during branch merge migration (closes #9899)

### LINKED ISSUE #9899: bug: merge fails with SchemaNotFoundError when removing any schema node
### Component

API Server / GraphQL

### Infrahub version

1.10.4 (SDK 1.20.1)

### Current Behavior

Merging a branch (via a proposed change or `infrahubctl branch merge`) that removes any schema node fails. Both the Schema Integrity and Data Integrity validators pass and the proposed change is reported "eligible to be merged", but the merge task fails during migration calculation and rolls back:

```
infrahub.exceptions.SchemaNotFoundError:
    Unable to find the schema '<NodeKind>' in the registry
    main | <NodeKind>
```

Traceback path:
`core/branch/tasks.py:_do_merge_branch` (~L440) -> `core/merge/branch_merger.py:calculate_migrations` (L151) -> `core/models.py:SchemaUpdateValidationResult.init -> process_diff` (L194) -> `core/schema/schema_branch.py:get` (L405), which raises.

`calculate_migrations` builds the post-merge (target) schema, in which the node is already removed, then `process_diff` calls `schema.get(name='<NodeKind>', duplicate=False)` against that target schema and fails because the node is gone.

A secondary task also fails while serializing the exception:

```
TypeError: SchemaNotFoundError.__init__() missing 1 required positional argument: 'identifier'
```

The exception is not picklable (its `__init__` requires `identifier`), so Prefect cannot round-trip it, which can mask the real error in some views.

### Expected Behavior

The merge removes the node and completes. `process_diff` should resolve a removed schema element against the pre-merge (source/root) schema, or otherwise handle removals without raising `SchemaNotFoundError`. (Separately, `SchemaNotFoundError` should be picklable.)

### Steps to Reproduce

Reproduced on a clean instance with fresh schema and zero objects; no inheritance or relationships required.

1. Add a schema node to `main`:
   ```yaml
   version: "1.0"
   nodes:
     - name: ReproSolo
       namespace: Testing
       attributes:
         - { name: name, kind: Text, unique: true }
   ```
2. Create a branch and set that node `state: absent` (or delete it).
3. Merge the branch to `main` (proposed-change merge or `infrahubctl branch merge`).
4. The merge task fails: `SchemaNotFoundError: Unable to find the schema 'TestingReproSolo' in the registry`, then rolls back.

### Additional Information

- Reproduced on 1.10.4, fresh schema, zero objects. Happens for a standalone node and for a node that inherits a generic.
- Removing the node via `infrahubctl schema load` directly (not via a merge) works. Only the merge path (`calculate_migrations`) is affected.
- Removing relationships via a merge works; only node removal fails.
- Both integrity validators pass; the failure is only at merge time.
- Earlier failure on the same effort (a different bug, diff tracking after a branch rebase): #9898.
- A redacted diagnostic bundle is available on request.

iddocohen:
# Why

Merging a branch that removes any schema node failed. Both integrity validators passed and the proposed change was reported "eligible to be merged", but the merge task then failed during migration calculation and rolled back with:

```
SchemaNotFoundError: Unable to find the schema '<NodeKind>' in the registry
```

Closes #9899

## What changed

**Behavioral change:** merging a branch that removes a schema node now completes instead of failing and rolling back.

**Root cause.** Migration calculation builds a 3-way schema diff (`initial ↔ source` + `initial ↔ destination`). A removed node lands in the `removed` set from the source side, but also appears in the `changed` set from the destination side, because `load_schema_from_db(at=created_at)` does not byte-for-byte match the live in-memory schema. `process_diff` then looked the node up in the post-merge target schema — where it has already been deleted — and raised `SchemaNotFoundError`.

- `process_diff` now skips any node already present in the diff's `removed` set when iterating `changed`. The removal is fully handled by the `removed` loop; field-level migrations for a node that is disappearing are moot.
- `SchemaNotFoundError` is now picklable (`__reduce__`). Its `__init__` requires `branch_name`/`identifier`, so the default pickle round-trip (args = message only) raised `TypeError: __init__() missing 1 required positional argument: 'identifier'`, masking the real error when transported between processes (e.g. Prefect).

**What stayed the same:** no schema or API changes. Nodes that are only *changed* (not removed) still produce their migrations exactly as before — verified by the existing `test_merge_update_schema` component test.

## How to review

1. `backend/infrahub/core/models.py` — the guard in `process_diff`.
2. `backend/infrahub/exceptions.py` — `SchemaNotFoundError.__reduce__`.
3. Tests: `backend/tests/unit/core/test_models.py::test_process_diff_node_removed_and_changed` reproduces the original crash (node in both `removed` and `changed`, target schema missing the node); `backend/tests/unit/test_exceptions.py` guards the pickle round-trip.

## Test Plan

```bash
uv run pytest backend/tests/unit/core/test_models.py backend/tests/unit/test_exceptions.py -q
uv run pytest backend/tests/component/core/test_branch_merge.py::test_merge_update_schema -q
```

Both green. The unit test fails on `stable` with the exact reported traceback (`models.py` → `schema_branch.py` `SchemaNotFoundError`) and passes with the fix.

---
## Summary by cubic
Fixes a merge crash when a branch removes a schema node by correctly handling nodes that show up as both removed and changed. Also makes `SchemaNotFoundError` picklable so real errors aren’t masked when crossing process boundaries.

- **Bug Fixes**
  - Skip nodes found in `diff.removed` when iterating `diff.changed` in `process_diff` to avoid looking up a node that no longer exists.
  - Add `SchemaNotFoundError.__reduce__` to support pickle round-trips (e.g., Prefect) and preserve error context.
  - Added targeted unit tests; no schema or API changes, and migrations for non-removed nodes are unchanged.

<sup>Written for commit 52556ca4526d5c69fc70edb353ca48be1d233f5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9901?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/models.py
+++ b/backend/infrahub/core/models.py
@@ -191,6 +191,12 @@ class SchemaUpdateValidationResult(BaseModel):
             )
 
         for schema_name, schema_diff in self.diff.changed.items():
+            # A node can appear in both `removed` and `changed` of a merged 3-way diff
+            # when it is removed on one side and modified on the other. Its removal is
+            # already handled above, and the post-merge target schema no longer holds it,
+            # so processing it as a changed element would raise SchemaNotFoundError.
+            if schema_name in self.diff.removed:
+                continue
             schema_node = schema.get(name=schema_name, duplicate=False)
             if "inherit_from" in schema_diff.changed:
                 self.migrations.append(
--- a/backend/infrahub/exceptions.py
+++ b/backend/infrahub/exceptions.py
@@ -295,6 +295,12 @@ class SchemaNotFoundError(Error):
         self.message = message or f"Unable to find the schema {identifier} in the database."
         super().__init__(self.message)
 
+    def __reduce__(self) -> tuple[type[SchemaNotFoundError], tuple[str, str, str]]:
+        # The default pickling only preserves `args` (the message), but __init__ requires
+        # branch_name and identifier, so a round-trip through pickle (e.g. Prefect) would
+        # otherwise fail with a missing-argument TypeError.
+        return (self.__class__, (self.branch_name, self.identifier, self.message))
+
     def __str__(self) -> str:
         return f"""
         {self.message}
--- a/backend/tests/unit/core/test_models.py
+++ b/backend/tests/unit/core/test_models.py
@@ -1,7 +1,8 @@
 from typing import Any
 
 from infrahub.core.constants import UpdateValidationErrorType
-from infrahub.core.schema import SchemaRoot
+from infrahub.core.models import SchemaUpdateValidationResult
+from infrahub.core.schema import AttributeSchema, SchemaRoot
 from infrahub.core.schema.schema_branch import SchemaBranch
 
 WIDGET_GADGET_SCHEMA: dict[str, Any] = {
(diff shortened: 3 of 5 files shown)


backend/infrahub/core/models.py
```diff
@@ -191,6 +191,12 @@ def process_diff(self, schema: SchemaBranch) -> None:
             )
 
         for schema_name, schema_diff in self.diff.changed.items():
+            # A node can appear in both `removed` and `changed` of a merged 3-way diff
+            # when it is removed on one side and modified on the other. Its removal is
+            # already handled above, and the post-merge target schema no longer holds it,
+            # so processing it as a changed element would raise SchemaNotFoundError.
+            if schema_name in self.diff.removed:
+                continue
```
saltas888, line 199: The fix works, but I'm wondering if it belongs in `SchemaDiff.__add__` rather than in `process_diff`, so we handle this where the schema diff lives and not here in the **consumer**.

If we reconcile there instead, dropping anything in `removed` from `changed`/`added`, the combined diff stays internally consistent and every consumer sees a coherent object, not just this one loop. The `process_diff` guard can then go away.

Something like:

```python
def __add__(self, other: SchemaDiff) -> SchemaDiff:
    merged_dict = deep_merge_dict(self.model_dump(), other.model_dump())
    merged = self.__class__(**merged_dict)
    # Removal is terminal: a node deleted on one side must not survive as
    # changed/added from the other. Keep the buckets mutually exclusive.
    for name in merged.removed:
        merged.changed.pop(name, None)
        merged.added.pop(name, None)
    return merged
```

Only real trade-off, this bakes "removed wins over changed" into the combined diff, but that's correct for the merge model here (removal is terminal, the post-merge target reflects it), so I think it's fine.

@ajtmccarty can u take a look as well?

ajtmccarty:
replaced by #9905

PR #9904, opened 2026-07-14
fix: recompute relationship on_delete when kind changes away from Component

lancamat1:
# Why

A relationship's cascade-on-delete behavior is derived from its `kind` (a `Component` relationship implies `on_delete=CASCADE`). Once that value was derived and persisted, it was never recomputed when the kind later changed. So a relationship switched from `Component` to `Attribute` kept cascading deletes to its related nodes as if it were still a Component — observed at a customer.

Goal: changing a relationship's kind away from `Component` stops the cascade, while still honoring an explicit `on_delete` the user set on purpose.

Non-goal: this does not retroactively repair schemas that are already in the stale state (see Impact & rollout).

## What changed

Behavioral changes:
- Changing a relationship's `kind` away from `Component` — without restating `on_delete` — now recomputes the delete behavior from the new kind, so deletes stop cascading.
- An explicit `on_delete` that diverges from the kind's default (e.g. a `Generic` relationship deliberately set to cascade, as the built-in `artifacts` relationship does) is preserved across a kind change.

Implementation notes:
- `on_delete` is reconciled at schema-merge time in `RelationshipSchema.update`, the only layer that sees both the old kind/value and the incoming kind. A stored value that matches the *old* kind's default is treated as auto-derived and cleared so it re-derives from the new kind; a divergent value is kept as a deliberate override.
- Derivation is centralized in `RelationshipSchema.default_on_delete_for_kind` and reused by `SchemaBranch.process_relationships` (single source of truth).

What stayed the same: no database-schema or GraphQL-contract changes; cascade deletes still key off `on_delete`.

## How to review

- `backend/infrahub/core/schema/relationship_schema.py` — the `update` override and the derivation helper; this is the core of the fix and the part worth extra scrutiny (the "auto-derived vs. explicit override" heuristic).
- `backend/infrahub/core/schema/schema_branch.py` — `process_relationships` now calls the shared helper.
- `backend/tests/unit/core/schema/test_schema_branch.py` — the reproduction test plus a guard that an explicit override survives a kind change.

## How to test

```bash
uv run pytest backend/tests/unit/core/schema/test_schema_branch.py -v
uv run pytest backend/tests/component/core/schema_manager/test_manager_schema.py -k on_delete -v
```

Regression suites run locally and green: schema unit (60) + migrations (7), `schema_manager` component (200), node delete/cascade component (17).

## Impact & rollout

- **Backward compatibility:** fixes the bug going forward. Schemas *already* corrupted (kind already switched away from `Component` while the stale `CASCADE` is persisted) are **not** auto-healed — the stale value is now indistinguishable from an intentional override. Remediation is to set `on_delete: no-action` explicitly on the affected relationship, or a follow-up data migration.
- **Performance:** no impact.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (N/A — no user-facing doc change)
- [ ] Internal .md docs updated (N/A)

---
## Summary by cubic
Fixes stale cascade deletes when a relationship’s kind changes away from Component. We now recompute on_delete from the new kind unless an explicit override exists.

- **Bug Fixes**
  - Recompute on_delete on kind change to stop unwanted cascades.
  - Keep explicit on_delete overrides intact.
  - Centralize default derivation in RelationshipSchema.default_on_delete_for_kind and use it in SchemaBranch.
  - Add unit tests for both cases.

- **Migration**
  - Existing schemas already persisting CASCADE won’t auto-heal. Set on_delete: no-action on affected relationships or run a data migration.

<sup>Written for commit a87bef53851c3d095c734e1992999b1a4817faa7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9904?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/schema/relationship_schema.py
+++ b/backend/infrahub/core/schema/relationship_schema.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any
 from pydantic import BaseModel
 
 from infrahub import config
-from infrahub.core.constants import RelationshipDirection, RelationshipKind
+from infrahub.core.constants import RelationshipDeleteBehavior, RelationshipDirection, RelationshipKind
 from infrahub.core.query import QueryNode, QueryRel, QueryRelDirection
 from infrahub.core.relationship import Relationship
 from infrahub.exceptions import InitializationError
@@ -14,7 +14,10 @@ from infrahub.exceptions import InitializationError
 from .generated.relationship_schema import GeneratedRelationshipSchema
 
 if TYPE_CHECKING:
+    from typing_extensions import Self
+
     from infrahub.core.branch import Branch
+    from infrahub.core.models import HashableModel
     from infrahub.core.query import QueryElement
     from infrahub.core.schema import MainSchemaTypes
     from infrahub.database import InfrahubDatabase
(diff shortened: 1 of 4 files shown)


lancamat1:
Closing: this fix addresses an unnecessary problem and adds avoidable complexity. Scrapping the branch.

PR #9905, opened 2026-07-14
schema diff cleanup

### LINKED ISSUE #9899: bug: merge fails with SchemaNotFoundError when removing any schema node

ajtmccarty:
# Why

The schema diff (`SchemaBranch.diff`) decides what changes when loading a schema, merging a branch,
rebasing, and upgrading. Several independent defects made that diff report changes that aren't real
— or compute the wrong baseline — which surfaced as hard failures:

- Merging a branch that **removes a node schema** fails with `SchemaNotFoundError` (the removed node
  is spuriously reported as *changed*, not just *removed*).
- Loading a schema on a **rebased branch** is rejected because core kinds (`CoreProfile`) show as
  changed on `update: not_supported` fields.
- The `infrahub upgrade` core-schema diff is **never empty** on a fresh, up-to-date install
  (`LineageSource`/`LineageOwner` always reported as changed).
- The real error is often **masked**: `SchemaNotFoundError` isn't picklable, so it fails to
  serialize across the workflow engine and surfaces as an unrelated `TypeError`.

**Goal:** make the schema diff faithful so schema load, merge, rebase, and upgrade behave correctly,
and make errors legible when they do occur.

**Non-goals:** no changes to the schema data model or the GraphQL/REST API contracts.

Closes #9899

## What changed

Grouped by intent:

- **Compare id-less schema elements by value.** `_diff_element` previously flagged any element with
  no id on either side as *added* (guarded by a `PYTEST_RUNNING` env check, so test and production
  behaved differently). Virtual relationships (e.g. `profiles`) are never persisted and have no id,
  so this manufactured phantom changes — the root cause of the node-remove-merge
  `SchemaNotFoundError`. These elements are now diffed by value, and the `PYTEST_RUNNING` shim is
  removed so tests exercise the production path.

- **Persist deprecated schema models on first-time initialization.** Fresh installs now store the
  deprecated-model markers, so the `infrahub upgrade` core-schema diff starts empty instead of
  always reporting `LineageSource`/`LineageOwner` as changed.

- **Correct the merge baseline on rebased branches.** `BranchMerger.get_initial_source_branch` now
  uses the destination schema at `branched_from` as the 3-way diff baseline. A rebase folds
  destination changes into the source branch, so basing the diff on the original branch point
  double-counted those as branch changes; `branched_from` (which a rebase advances) is the correct
  divergence point.

- **Fully resolve late-introduced schemas in `SchemaBranch.process`.** The profile/template
  management steps can pull in base generics (e.g. `CoreProfile`) *after* the early resolution
  passes, leaving their attributes unresolved (branch-support at `None`, legacy `min_length`/
  `max_length` unsynced with the nested `parameters`). Attribute parameter reconciliation and
  branch-support resolution now re-run at the end of `process`, so a subsequent schema load on a
  rebased branch isn't rejected on `not_supported` fields.

- **Make `SchemaNotFoundError` picklable.** A base-class `Error.__reduce__` reconstructs via
  `__new__` + instance state, so every `Error` subclass round-trips through pickle regardless of its
  constructor signature — the real error now propagates instead of a masking `TypeError`.

What stayed the same: no schema-model or API contract changes; `process()` remains idempotent for
already-complete schemas (the re-run passes are no-ops there).

## How to review

Suggested order:

1. `core/schema/basenode_schema.py` — the id-less diff change (the central fix).
2. `core/schema/schema_branch.py` — the end-of-`process` re-resolution.
3. `core/merge/branch_merger.py` — the `branched_from` baseline.
4. `exceptions.py` — the picklable base-class `__reduce__`.
5. `core/initialization.py` — deprecated-model persistence.

Extra scrutiny welcome on the `SchemaBranch.process` change: it re-runs two resolution steps at the
end of the pipeline. It is guarded by idempotency (verified by `test_process_idempotency`), but it
sits in a hot path exercised on every schema mutation.

## Impact & rollout

- **Backward compatibility:** no breaking changes; no data migration required.
- **Performance:** `SchemaBranch.process` runs two additional resolution passes per call; both are
  cheap and no-op when nothing needs resolving.
- **Config/env changes:** none. (`PYTEST_RUNNING`, a test-only env shim, is removed.)

## Checklist

- [x] Tests added/updated
- [x] Changelog entries added

---
## Summary by cubic
Fixes false-positive schema diffs (including on rebased branches), aligns fresh installs with the deprecated overlay, and makes custom errors picklable across workflows.

- **Bug Fixes**
  - Schema diffs: compare id-less attributes/relationships by value (remove `PYTEST_RUNNING`); don’t clear optional text fields when already empty; identical branches and data-only changes now produce empty diffs.
  - Rebases: use the destination schema at `branched_from` as the common ancestor in 3‑way diffs; re-run attribute-parameter reconciliation and branch-support at the end of pre-validation so rebased branches don’t resurface folded changes or flag `CoreProfile` fields.
  - Fresh install: load `deprecated_models` during `first_time_initialization` so the core-schema diff starts empty.
  - Exceptions: make `Error` subclasses pickle-safe via a custom `__reduce__`, so real errors propagate across the workflow engine.

<sup>Written for commit 278bb5f4bb506765a039f96f6e70b4172d1cfcc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/branch_merger.py
```diff
@@ -94,7 +94,12 @@ def workflow(self) -> InfrahubWorkflow:
         return self._workflow
 
     async def get_initial_source_branch(self) -> SchemaBranch:
```
dgarros: Should we update the name of the method , if feels strange to have source_branch in the name and to pull from destination_branch within the function

backend/infrahub/core/initialization.py
```diff
@@ -561,6 +562,7 @@ async def first_time_initialization(db: InfrahubDatabase) -> None:
     schema = SchemaRoot(**internal_schema)
     schema_branch = registry.schema.register_schema(schema=schema, branch=default_branch.name)
     schema_branch.load_schema(schema=SchemaRoot(**core_models))
+    schema_branch.load_schema(schema=SchemaRoot(**deprecated_models))
```
dgarros, line 565: is it a long term fix or a workaround ? I'm surprised we have to load the deprecated models into the schema for everyone

ajtmccarty: this is actually just to be symmetrical with `check_core_schema_diff` and `update_core_schema` so that `LineageSource` and `LineageOwner` won't show a diff for `display_labels` (`[] <> NULL/None`) where none really exists

backend/infrahub/core/merge/branch_merger.py
ajtmccarty: yes good point. updated to `common_ancestor_schema` throughout

PR_DESCRIPTION.md
```diff
@@ -0,0 +1,45 @@
...
+- When recalculating a branch's tracked diff, the coordinator now deletes every *other* root carrying the same tracking id — querying by tracking id alone instead of within the aggregation's time range — so a tracking id always maps to a single diff.
+- Because the merge flow re-runs the diff update before the failing lookup, this also unblocks databases already in the broken state on the next merge attempt.
+
+<!-- What stayed the same -->
+- No schema changes; no API contract changes. Only the diff-coordinator cleanup path is affected.
+
+## How to review
+
+- Core change: `backend/infrahub/core/diff/coordinator.py` — the stale-root cleanup in `_update_diffs`.
+- Regression test: `test_tracking_diff_not_duplicated_after_rebase` reproduces the rebase → duplicate-root scenario (fails with the exact issue error before the fix).
```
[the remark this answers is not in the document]
ajtmccarty: this was included by mistake. deleted now

dgarros (approved):

ajtmccarty:
also verified with manual testing against a local infrahub instance
- load a schema
- create a branch
- add an attribute to the schema on main
- add an attribute to the schema on the branch
- rebase the branch successfully
- remove the schema on the branch
- merge the branch
- verify success

PR #9924, opened 2026-07-16
feat(merge): selective post-merge generator & artifact regeneration

polmichel:
## Why

Merging a branch currently re-runs **every** generator and regenerates **every** artifact,
regardless of what the merge actually changed. On a large dataset — the originating report
(IFC-2306) cites ~4,000 artifacts across ~40 branches — a single merge spawns thousands of
artifact-generation tasks and leaves the API and UI effectively unusable for 10–15 minutes.

**Goal:** make post-merge execution selective — run only the generators and regenerate only the
artifacts whose inputs the merge diff affected, scoped to the affected group members.

**Guiding invariant:** over-execution is acceptable, under-execution is not. Whenever the affected
set cannot be determined with confidence, the merge falls back to regenerating everything rather
than risk leaving an artifact stale.

**Non-goals:** no change to any regeneration caller outside the merge follow-up (proposed-change
pipeline, manual regeneration are untouched); the shared-primitive refactor is deferred (see below).

Closes IFC-2893, IFC-2894, IFC-2895, IFC-2898

## Scope within IFC-2704

Where this PR sits in the parent-feature breakdown:

| Child | Summary | This PR? |
|---|---|---|
| IFC-2893 | Setup | ✅ closes (already listed) |
| IFC-2894 | Foundational (blocking prereqs) | ✅ closes (already listed) |
| IFC-2895 | MVP — selective regen on merge | ✅ closes (already listed) |
| IFC-2898 | Reversible rollout via configuration | ✅ this is the config flag — closes |
| IFC-2908 | internal: refactoring into regeneration package | 🔜 referenced as follow-up (Phase 1) |
| IFC-2911 | internal: remove SDK usage from selective regen | 🔜 follow-up (Phase 2 of same doc) |
| IFC-2896 | Safe fallback to full regeneration | ⚠️ mechanism is in this PR, story is a later increment |
| IFC-2897 | Repository code changes still regenerate | ❌ later increment (US2) |
| IFC-2899 | Direct-merge generator cascade (D7) | ⚠️ only validated here (test #8), not closed |
| IFC-2900 / 2907 | Polish / additional thoughts | ❌ later |

## What changed

**Behavioral changes**

- On merge, only definitions whose inputs are in the merge diff are
  dispatched, narrowed to the affected members.
- Applies identically to direct branch merges and merges through a proposed change.
- New config flag `selective_execution_after_merge` gates the behavior.
- Comprehensive fallback to full regeneration on any uncertainty: feature disabled, no diff summary
  captured, an unloadable summary, or any failure during selection/dispatch.

**Implementation notes**

- The enriched branch diff is captured in the merge orchestrator, serialized into the `NodeDiff` summary the selection predicates consume, and cached keyed by the freeze-stable diff-root uuid. The follow-up loads the summary by key.
- A new `selective_regen/` package builds the regeneration plan from the diff summary.
- Selection reuses the proposed-change regeneration-trigger primitives rather than reimplementing
  them.

**What stayed the same:** no schema changes; no change to any regeneration caller outside the two
post-merge trigger submissions; disabled flag reproduces prior behavior byte-for-byte.

## How to review

The commits are clean and ordered to be read top to bottom; each is self-contained.

1. **`feat(config): add selective_execution_after_merge flag`** — the on/off surface and default,
   plus the regenerated artifacts (docker-compose, configuration reference, OpenAPI + frontend REST
   types). Start here.
2. **`feat(merge): add merge diff serializer/cache components and share PC primitives`** — how the
   enriched diff becomes the `NodeDiff` summary and is cached, and the proposed-change primitives
   that become shared. Foundational; no merge behavior wired yet.
3. **`feat(merge): add selective post-merge regeneration selection engine`** — the `selective_regen/`
   package (orchestrator → gate → selectors → impact resolver). The core logic; the dependency
   graph above maps this commit.
4. **`feat(merge): dispatch and thread the merge diff summary into the follow-up`** — the wiring:
   diff capture in the merge orchestrator, the dispatcher's enable/fallback decision tree, and the
   composition root in `branch/tasks.py`. The correctness surface for the no-under-execution invariant.
5. **`test(merge): selective post-merge regeneration`** — component-tier selection coverage of the
   selective dispatch (recorder-based, against a real graph) plus the engine unit suites. Real
   regenerated-output is deferred to an API-tier test — the integration testcontainer harness
   cannot execute the render flow's worker→server callback.
6. **`docs(specs): record IFC-2704 US1 progress, D7 spike, and interim default`** — spec/plan
   context; skim.
7. **`chore(config): regenerate docker-compose.yml`** — generated file; mechanical.
8. **`docs(specs): reconcile IFC-2704 test tiers and perf-validation status`** — aligns the
   spec/plan/tasks with the shipped test tiers and the perf-validation results; skim.

### Selective regeneration package

```text
COMPONENT / DEPENDENCY GRAPH
============================

core/branch/tasks.py :: _build_post_merge_regeneration_dispatcher(log)   [composition root]
        |   wires everything per merge follow-up
        v
+------------------------------------------------------------------+
| PostMergeRegenerationDispatcher      (regeneration_dispatcher.py) |
|   inject: workflow . selector . summary_cache . log              |
|   dispatch():                                                    |
|     flag off / no cache key / cache miss / error -> blanket      |
|                                    (TRIGGER_* every definition)  |
|     else -> selector.build_plan() -> submit only what it returns |
+------------------------------------------------------------------+
        |  selector : RegenerationSelector (Protocol)
        v
+------------------------------------------------------------------+
| MergeSelectiveRegeneration                     (orchestrator.py) |
|   built by build_merge_selective_regeneration(client, log)       |
|   modified_kinds = get_modified_kinds(diff)                      |
|   generator_selector.select() + artifact_selector.select()      |
|   -> SelectiveRegenerationPlan { generator_runs, artifact_gens } |
+------------------------------------------------------------------+
        |
        v
+------------------------------------------------------------------+
| DefinitionSelectorBase[Def, Req]   (definition_selector/base.py) |
|   ABC / template method                                          |
|   inject: client . gate . impacted_resolver . log                |
|   per definition:                                                |
|     1. gate.evaluate()            -> selected? / all-members?    |
|     2. impacted_resolver.resolve()-> which members changed       |
|     3. _narrow_members_filter()   -> subset, or empty = "all"    |
|     4. _build_request()           -> skip if no members render   |
+------------------------------------------------------------------+
     |  abstract hooks: _load_definitions . _fetch_group . _should_render . _build_request
     +-- GeneratorSelector   generators flagged execute_after_merge
     +-- ArtifactSelector    artifact definitions

  shared, injected into BOTH selectors:
    - DefinitionGate (gate.py)                 -> GateResult { managed_branch, selected }
        selects on: query changed | definition/fingerprint changed |
                    modified kind read by query | target group touched (membership-only)
    - ImpactedSubscriberResolver (impacted.py)
        field-level impact -> ALL (widen to all members) | NONE (skip) | specific ids

  Guiding invariant: every uncertain signal WIDENS the selection; it never narrows below
  the set of members whose inputs changed, so nothing affected is left stale.
```

### Follow-up refactor (uniformization, deferred)

The selective-regeneration path currently reuses the proposed-change selection primitives by
importing underscored internals from `proposed_change/tasks.py`, which forces a lazy import in
`branch/tasks.py` to avoid an import cycle. A separate refactor will **uniformize the two paths** by
extracting the shared cluster into a dependency-free leaf package (`core/regeneration/`), promoting
the shared names to public API and breaking the cycle (Phase 1, IFC-2908), then moving the merge
path off the SDK client to direct database access (Phase 2, IFC-2911). Both are behavior-neutral and
intentionally **not** part of this increment.

## How to test

Automated coverage: unit suites for the selection engine (gate, selectors, orchestrator,
diff serializer/cache, dispatcher) plus a component-tier selection test driving the dispatcher
against a real graph (`backend/tests/component/proposed_change/test_merge_selective_regen.py`).

Manual A/B validation on a real dataset:

### Environment / build

- Dataset: infra demo + infrahub-demo-edge (read-only). 30 devices; 2 artifact defs
  (20 artifacts); 4 generator defs (all `execute_after_merge`). Baseline blanket = 80 leaf
  regen flows (40 query-group + 10 openconfig + 10 startup-config + 20 generator-member).
- Flag: `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE`. Same-build A/B: off, then on.
- Metric: total regeneration-family Prefect flow runs in the drained follow-up window
  (leaf flows + orchestration wrappers), per the plan's step-5 counting recipe.

### Results

Headline = total regen-family flows; flag-off is scenario-independent = 88.

| # | Scenario | Change | Flag off | Flag on | Under-exec |
|---|----------|--------|:--------:|:-------:|:----------:|
| 1 | Single-kind change | `InfraInterfaceL3.description`, 1 arista device | 88 | 24 | 0 |
| 2 | No-op read | `BuiltinTag.description` (no def reads it) | 88 | 0 | 0 |
| 4 | Membership-only | add existing core device to arista group | 88 | 3 | 0 |
| 8 | Direct-merge gen. cascade | `InfraCircuit.vendor_id` (generators only) | 88 | 3 | 0 |

## Impact & rollout

- **Backward compatibility:** selective execution is on by default; set
  `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE=false` to restore the exact prior full-regeneration
  behavior. The no-under-execution guarantee holds either way via the full-regeneration fallback.
- **Performance:** the point of the change — dispatched-task count for a small merge drops from
  proportional-to-instance-size to proportional-to-affected-set (73–100% fewer regen flows across
  the validated scenarios).
- **Config/env changes:** new `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE` flag, default `true`
  (docker-compose and configuration reference regenerated).
- **Deployment notes:** safe to deploy; selective path is active by default and revertible via the flag.


backend/infrahub/core/branch/tasks.py
```diff
@@ -462,11 +467,33 @@ async def _get_diff_root(
...
 
 
+async def _build_post_merge_regeneration_dispatcher(
+    log: Logger | LoggerAdapter[Logger],
+) -> PostMergeRegenerationDispatcher:
+    # Imported lazily: the selection module depends on the proposed-change tasks, which import this
+    # module, so a top-level import would be circular.
+    from infrahub.core.merge.diff_summary_cache import MergeDiffSummaryCache
+    from infrahub.core.merge.regeneration_dispatcher import PostMergeRegenerationDispatcher
+    from infrahub.core.merge.selective_regen.orchestrator import build_merge_selective_regeneration
```
polmichel, line 479: This will be tackled in the next PR against this feature branch. I'll extract the common logic around selective regeneration from `proposed_change` and `merge/selective_regen` package inside a new package at the `core/` level. This is described in the PR description and will be done before moving on regarding other subitems of the feature.

backend/infrahub/core/merge/selective_regen/impacted.py
```diff
@@ -0,0 +1,40 @@
...
+
+from typing import TYPE_CHECKING
+
+from infrahub.proposed_change.tasks import ImpactScope, get_field_level_impacted_subscribers
+
+if TYPE_CHECKING:
+    from infrahub_sdk.client import InfrahubClient
+    from infrahub_sdk.diff import NodeDiff
+
+    from infrahub.proposed_change.tasks import ImpactedSubscribers
```
polmichel, line 11: Hopefully after the refactor, we should not see these imports targeting `proposed_change.tasks` file

backend/infrahub/message_bus/types.py
```diff
@@ -133,6 +133,15 @@ def source_noun(self) -> str:
...
         return "artifacts"
 
+    def reads_kind(self, kind: str) -> bool:
+        """Whether a data change to ``kind`` is relevant because the query reads that kind.
+
+        An artifact query may read a kind through its Profile, so the profile-stripped kind matches too.
+        """
+        return kind in self.query_models or (
+            kind.startswith("Profile") and kind.replace("Profile", "", 1) in self.query_models
+        )
```
polmichel, line 143: This behavior is fragile. An internal task list this as an additional item to change https://opsmill.atlassian.net/browse/IFC-2907

polmichel: And there is also an existing asymmetry between artifacts and generator, which we can now notice in the difference between `reads_kind` methods.
I am not certain that this is expected from the system. The same internal task contains a follow-up item to verify this point.

backend/tests/unit/core/merge/selective_regen/definition_selector/test_base.py
```diff
@@ -0,0 +1,217 @@
...
+        query_models=[],
+        query_payload="query {}",
+        repository_id="repo-1",
+    )
+
+
+class _StubGate(DefinitionGate):
+    """A gate whose verdict is fixed per definition name, bypassing diff evaluation."""
+
+    def __init__(self, results: dict[str, GateResult]) -> None:
```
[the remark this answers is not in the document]
polmichel: Let's not over-anticipate these needs. this would break one of the repo rule which is not using mock objects

backend/infrahub/config.py
```diff
@@ -216,6 +216,11 @@ class MainSettings(BaseSettings):
         default=False,
         description="When enabled, the Infrahub branch is automatically deleted after a successful merge.",
     )
+    selective_execution_after_merge: bool = Field(
+        default=True,
```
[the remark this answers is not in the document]
polmichel: Set to `default=False` to match tasks.md T001 / research.md D9 for this increment (the default-True rationale assumes the US3/cascade fallbacks, which are not yet built); regenerated the compose files and configuration reference to match. Resolved in "feat(config): add selective_execution_after_merge flag".

backend/infrahub/core/merge/selective_regen/definition_selector/artifact_selector.py
```diff
@@ -0,0 +1,135 @@
...
+    subscriber_kind = InfrahubKind.ARTIFACT
+
+    async def _load_definitions(
+        self, *, target_branch: str
+    ) -> list[LoadedDefinition[ProposedChangeArtifactDefinition]]:
+        definition_information = await self.client.execute_graphql(
+            query=GATHER_ARTIFACT_DEFINITIONS, branch_name=target_branch
+        )
+        edges = definition_information[InfrahubKind.ARTIFACTDEFINITION]["edges"]
+        group_id_by_definition = {edge["node"]["id"]: edge["node"]["targets"]["node"]["id"] for edge in edges}
```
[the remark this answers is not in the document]
polmichel: `_load_definitions` now skips definitions whose `targets` is null (with a warning) instead of raising, so a single invalid definition no longer forces the whole merge back to blanket regeneration. Resolved in "feat(merge): add selective post-merge regeneration selection engine".

backend/infrahub/core/merge/selective_regen/definition_selector/base.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Each request is narrowed to the members whose inputs the diff actually changed; a definition
+        left with no members to render produces no request. Every uncertain signal widens the
+        selection rather than narrows it, so the result never omits a member the merge may have
+        affected.
+        """
+        requests: list[RequestT] = []
+        for loaded in await self._load_definitions(target_branch=target_branch):
+            definition = loaded.definition
+            gate_result = self.gate.evaluate(
```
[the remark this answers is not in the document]
polmichel: The repository code/template path (US2) is intentionally deferred in this increment, so to avoid shipping the under-execution window enabled the flag now defaults to `False` — selective execution is off until the fallbacks land. Resolved in "feat(config): add selective_execution_after_merge flag".

backend/tests/unit/core/merge/selective_regen/definition_selector/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+def client() -> InfrahubClient:
+    return InfrahubClient(config=Config(address="http://mock"))
+
+
+@pytest.fixture
+def artifact_selector(client: InfrahubClient) -> ArtifactSelector:
+    log = logging.getLogger("test")
+    return ArtifactSelector(
+        client=client,
+        gate=DefinitionGate(log=log),
```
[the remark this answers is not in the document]
polmichel: Extracted shared `log`, `gate` and `impacted_resolver` fixtures so the dependency wiring is defined once. Resolved in "feat(merge): add selective post-merge regeneration selection engine".

backend/tests/unit/core/merge/selective_regen/test_orchestrator.py
```diff
@@ -0,0 +1,86 @@
...
+    async def select(
+        self, *, diff_summary: list[NodeDiff], target_branch: str, modified_kinds: list[str]
+    ) -> list[RequestT]:
+        self.calls.append((diff_summary, target_branch, modified_kinds))
+        return self.result
+
+    async def _load_definitions(self, *, target_branch: str) -> list[LoadedDefinition[DefinitionT]]:
+        raise NotImplementedError
+
+    async def _fetch_group(self, *, definition: DefinitionT, target_branch: str) -> CoreGroup:
```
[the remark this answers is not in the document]
polmichel: Removed the dead `_fetch_group` override (the base abstract method is `_fetch_member_ids` after the selector refactor). Resolved in "feat(merge): add selective post-merge regeneration selection engine".

backend/infrahub/git/models.py
```diff
@@ -26,6 +26,10 @@ class RequestArtifactDefinitionGenerate(BaseModel):
         default_factory=list,
         description="List of targets to limit the scope of the generation, if populated only the included artifacts will be regenerated",
     )
+    members: list[str] = Field(
```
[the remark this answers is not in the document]
polmichel: Added a focused component test asserting `generate_request_artifact_definition` generates a selected member that has no existing artifact (so the `limit`/artifact-id path cannot silently drop it) and that an empty `members` list means all members. Resolved in "test(merge): selective post-merge regeneration".

backend/tests/integration/proposed_change/test_merge_selective_regen.py
```diff
@@ -0,0 +1,383 @@
...
+        )
+
+    async def _run_follow_up(
+        self,
+        *,
+        dataset: dict[str, Any],
+        default_branch: Branch,
+        admin_account: CoreAccount,
+        memory_cache: MemoryCache,
+        diff_summary: list[dict[str, Any]] | None,
```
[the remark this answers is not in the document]
polmichel: This test was moved to a component-tier selection test that seeds the cache with real `NodeDiff` objects (`make_node_diff`), so the `list[dict]` vs `list[NodeDiff]` contract mismatch no longer exists. Resolved in "test(merge): selective post-merge regeneration".

backend/tests/unit/core/merge/selective_regen/definition_selector/test_base.py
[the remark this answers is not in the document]
polmichel: Not applied: the suggested `unittest.mock.create_autospec` conflicts with this repo's testing rule (`.agents/rules/testing-python.md` — no `unittest.mock`; adapter/protocol stubs only). The stubs override every method that reads `self.client`/`self.log`, so the uninitialized parent state is unreachable; leaving as-is.

backend/infrahub/core/branch/tasks.py
polmichel: The PR is already too big, doing the refactor here would increase the difficulty to review this MVP

backend/infrahub/core/merge/regeneration_dispatcher.py
```diff
@@ -0,0 +1,117 @@
...
+            for generator_run in plan.generator_runs:
+                await self.workflow.submit_workflow(
+                    workflow=REQUEST_GENERATOR_DEFINITION_RUN, context=context, parameters={"model": generator_run}
+                )
+            for artifact_generate in plan.artifact_generates:
+                await self.workflow.submit_workflow(
+                    workflow=REQUEST_ARTIFACT_DEFINITION_GENERATE,
+                    context=context,
+                    parameters={"model": artifact_generate},
+                )
```
polmichel, line 108: For now, there is one run by element. I've created an internal task to make sure we cannot gain some time by changing that

backend/infrahub/core/merge/selective_regen/definition_selector/artifact_selector.py
```diff
@@ -0,0 +1,73 @@
...
+    """Selects the artifact definitions a merge changed, narrowed to the members it affects."""
+
+    subscriber_kind = InfrahubKind.ARTIFACT
+
+    async def _load_definitions(
+        self, *, target_branch: str
+    ) -> list[LoadedDefinition[ProposedChangeArtifactDefinition]]:
+        definition_information = await self.client.execute_graphql(
+            query=GATHER_ARTIFACT_DEFINITIONS, branch_name=target_branch
+        )
```
polmichel, line 29: I do not think that using the SDK is the correct way to manage this definition loading, but this is the path used inside PC selection related code.
An internal task will have as follow-up items to use directly db queries instead https://opsmill.atlassian.net/browse/IFC-2911

backend/infrahub/core/merge/diff_serializer.py
```diff
@@ -0,0 +1,84 @@
...
+class MergeDiffSerializer:
+    """Serialize an enriched merge diff into the NodeDiff summary the selection predicates consume.
+
+    Nodes are tagged with the merge target (destination) branch rather than the source branch the
+    diff was computed on: post-merge the changed data lives on the target branch, which is also
+    where selection runs its live lookups, so the tag stays valid even after the source branch is
+    deleted. Actions are emitted as the uppercase enum name to match the proposed-change summary.
+    Unchanged nodes and unchanged elements are dropped; a node whose only change was a conflict
+    resolved to the base branch is still emitted, since it remains a real change for regeneration.
+    """
```
ajtmccarty: most of this is not the responsibility of this class. it just turns an EnrichedDiffRoot into a JSON blob, leaving out UNCHANGED items

backend/infrahub/core/merge/diff_summary_cache.py
```diff
@@ -0,0 +1,52 @@
...
+
+        Raises:
+            ResourceNotFoundError: the summary is missing or cannot be loaded.
+
+        """
+        try:
+            summary_payload = await self._cache.get(key=self._key(diff_id))
+            if not summary_payload:
+                raise ResourceNotFoundError(message=f"Merge diff summary for diff {diff_id} was not found in the cache")
+            return _NODE_DIFFS_ADAPTER.validate_json(summary_payload)
```
ajtmccarty: I think this logic should move to the `MergeDiffSerializer`. serialization and deserialization are two halves of the same operation. then this component might need `MergeDiffSerializer` as a dependency, but I think that makes sense

is there a corresponding `_NODE_DIFFS_ADAPTER.model_dump` or something that can be used for serializing?

backend/infrahub/core/merge/diff_summary_cache.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class MergeDiffSummaryCache:
+    """Store and load a per-merge diff summary, keyed by the (freeze-stable) diff-root uuid."""
+
+    def __init__(self, cache: InfrahubCache) -> None:
+        self._cache = cache
+
+    async def set(self, diff_id: str, diff_summary: list[NodeDiff]) -> None:
+        await self._cache.set(key=self._key(diff_id), value=json.dumps(diff_summary), expires=KVTTL.TWO_HOURS)
```
ajtmccarty: the diff summary could be quite large for a big diff, not sure if there is a limit on the size of the JSON blob

backend/infrahub/core/diff/summary_serializer.py
ajtmccarty, line 1: I think this can live in `backend/infrahub/core/diff/`. it's not really related to a merge specifically. so it's more of a `DiffSummarySerializer`

backend/infrahub/core/merge/diff_summary_cache.py
ajtmccarty, line 1: same for this one the only thing that seems specifically merge-related is the cache key name, but that could be anything. this is really more of a `DiffSummaryCacheRepository` than `MergeDiffSummaryCache`, but `DiffSummaryCache` is fine

backend/tests/unit/core/diff/test_summary_serializer.py
```diff
@@ -0,0 +1,210 @@
...
+    NodeIdentifier,
+)
+from infrahub.core.merge.diff_serializer import MergeDiffSerializer
+from infrahub.core.timestamp import Timestamp
+
+SOURCE_BRANCH = "feature"
+TARGET_BRANCH = "main"
+
+
+def _node(
```
ajtmccarty, line 28: you don't have to use them, but this file has some factory classes for creating in-memory EnrichedDiff objects `backend/tests/component/core/diff/factories.py`

backend/infrahub/core/merge/selective_regen/orchestrator.py
```diff
@@ -0,0 +1,69 @@
...
+    """Select the generator and artifact definitions a merge changed, narrowed to affected members.
+
+    Orchestrates the generator and artifact selectors over a single computation of the diff's
+    modified kinds, returning the combined plan a merge follow-up should dispatch.
+    """
+
+    def __init__(
+        self,
+        generator_selector: DefinitionSelectorBase[ProposedChangeGeneratorDefinition, RequestGeneratorDefinitionRun],
+        artifact_selector: DefinitionSelectorBase[ProposedChangeArtifactDefinition, RequestArtifactDefinitionGenerate],
```
ajtmccarty, line 42: defining the specific type group of `DefinitionSelectorBase`s you need like this kind of defeats the purpose. Ideally this would accept `list[DefinitionSelectorBase]` and then run them all through `.select()` in `build_plan`. although this also requires some changes to the `SelectiveRegenerationPlan` data structure. it would probably have to link a workflow to each `run` somehow

backend/infrahub/core/merge/selective_regen/gate.py
```diff
@@ -0,0 +1,56 @@
...
+        diff_summary: list[NodeDiff],
+        modified_kinds: list[str],
+        group_id: str,
+    ) -> GateResult:
+        query_outcome = _query_changed(definition=definition, diff_summary=diff_summary)
+        definition_outcome = _definition_changed(definition=definition, diff_summary=diff_summary)
+        for outcome in (query_outcome, definition_outcome):
+            if outcome.reason is not None:
+                self.log.info(outcome.reason)
+        managed_branch = query_outcome.matched or definition_outcome.matched
```
ajtmccarty: I'm not clear on what "managed_branch" means

ajtmccarty (commented):
very nice component design.

backend/infrahub/core/merge/diff_serializer.py
polmichel: You're right, I've missed it. I'll change that

backend/infrahub/core/merge/diff_summary_cache.py
polmichel: True, it now belongs to `DiffSummarySerializer` object (renamed from `MergeDiffSerializer`), using `dump` (`_NODE_DIFFS_ADAPTER.dump_json`) and `load` (`_NODE_DIFFS_ADAPTER.validate_json`) method

backend/infrahub/core/diff/summary_serializer.py
polmichel: True, this is done and renamed

backend/infrahub/core/merge/diff_serializer.py
polmichel: Resolved in `"refactor(merge): move diff summary serializer to core/diff with paired codec"`

backend/infrahub/core/merge/diff_summary_cache.py
polmichel: Resolved in `"refactor(merge): move diff summary serializer to core/diff with paired codec"`

backend/infrahub/core/diff/summary_serializer.py
polmichel: Resolved in `"refactor(merge): move diff summary serializer to core/diff with paired codec"`

backend/infrahub/core/merge/diff_summary_cache.py
polmichel: That's a good point, I'll post this comment inside an internal task to take some time to analyze this point https://opsmill.atlassian.net/browse/IFC-2907

backend/infrahub/core/merge/selective_regen/gate.py
polmichel: It has slipped through the code coming from the parallel block inside PC package. I'll rename it into `regenerate_all` or something like that.
I think this is linked to some external git branch concept

backend/infrahub/core/merge/diff_summary_cache.py
polmichel: Renamed into `DiffSummaryCache`

backend/infrahub/core/merge/selective_regen/orchestrator.py
polmichel: Yes I agree and I thought the same. But I think this is still doable in a later PR, and the groundwork is already done. I would leave this for a future improvement if that's okay in your opinion

backend/infrahub/core/merge/diff_summary_cache.py
polmichel: Resolved in `refactor(diff): generalize MergeDiffSummaryCache to DiffSummaryCache in core/diff`

backend/infrahub/core/merge/selective_regen/orchestrator.py
polmichel: Hopefully I'll do it onto the develop branch tied to 1.12 https://opsmill.atlassian.net/browse/IFC-2916

backend/tests/unit/core/diff/test_summary_serializer.py
polmichel: As these are unit tests, I've moved the factories from component test suite to shared helpers and I used them. Thanks.

polmichel: Resolved in `test(diff): move EnrichedDiff factories to tests/helpers and adopt in serializer test`

backend/infrahub/core/merge/selective_regen/gate.py
polmichel: Resolved in `refactor(merge): rename gate flag managed_branch to regenerate_all_members`

backend/infrahub/config.py
```diff
@@ -216,6 +216,11 @@ class MainSettings(BaseSettings):
         default=False,
         description="When enabled, the Infrahub branch is automatically deleted after a successful merge.",
     )
+    selective_execution_after_merge: bool = Field(
```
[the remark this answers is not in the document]
polmichel: Will be added in the final PR of this feature.

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -100,6 +109,8 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
             changelog_collector = DiffChangelogCollector(diff=branch_diff, branch=self.source_branch, db=self.db)
             node_events = changelog_collector.collect_changelogs()
 
+            serialized_diff_summary = self._serialize_diff_summary(branch_diff=branch_diff)
```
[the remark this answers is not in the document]
polmichel: Will be enabled once the feature is totally finished and testing process is completed

backend/infrahub/proposed_change/tasks.py
```diff
@@ -1004,27 +1005,26 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
...
 
+    Such orphan rows can appear when a target node has been removed via a path that does not
+    cascade-delete the subscriber.
     """
-    artifacts_by_member: dict[str, str] = {}
-    for artifact in existing_artifacts:
-        object_id = artifact.object.id
+    subscriber_by_member: dict[str, str] = {}
+    for subscriber in existing_subscribers:
+        object_id = subscriber.object.id
```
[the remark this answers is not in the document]
polmichel: Resolved in `b2ad13b34`. The shared mapper now resolves each member via `subscriber.object.peer.id` (store-backed) instead of `subscriber.object.id`, which is `None` for generator instances whose object peer is only reachable through the store — so generator narrowing works instead of rerunning whole groups.

backend/infrahub/core/branch/tasks.py
```diff
@@ -477,16 +507,11 @@ async def post_process_branch_merge(source_branch: str, target_branch: str, cont
...
 
-        await get_workflow().submit_workflow(
-            workflow=TRIGGER_ARTIFACT_DEFINITION_GENERATE,
-            context=context,
-            parameters={"branch": target_branch},
-        )
-
-        await get_workflow().submit_workflow(
-            workflow=TRIGGER_GENERATOR_DEFINITION_RUN,
+        dispatcher = await _build_post_merge_regeneration_dispatcher(log=log)
```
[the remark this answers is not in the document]
polmichel: Resolved in `2d1e9c7f9`. With the feature off, the follow-up now submits blanket regeneration directly and never builds the selective dispatcher, so the fallback path no longer depends on SDK client / cache initialization.

backend/infrahub/core/merge/orchestrator.py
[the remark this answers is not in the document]
polmichel: Resolved in `2d1e9c7f9`. Diff-summary serialization is now gated on `selective_execution_after_merge`, so a flag-off merge skips the serialization and cache write entirely.

backend/infrahub/core/merge/selective_regen/definition_selector/artifact_selector.py
```diff
@@ -0,0 +1,73 @@
...
+            if target_node is None:
+                self.log.warning(
+                    f"Artifact definition {edge['node']['id']} has no target group; "
+                    "excluding it from selective regeneration"
+                )
+                continue
+            group_id_by_definition[edge["node"]["id"]] = target_node["id"]
+            selectable_edges.append(edge)
+        return [
+            LoadedDefinition(definition=definition, group_id=group_id_by_definition[definition.definition_id])
```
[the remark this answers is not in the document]
polmichel: Resolved in `0cb5f6ebe`. Each definition is now paired with its group id inside the gather loop, removing the implicit `definition_id` lookup coupling.

dev/specs/ifc-2704-incremental-merge-regen/plan.md
```diff
@@ -213,3 +213,14 @@ The Spec Kit block in `CLAUDE.md` will be updated (via the `after_plan`
 Re-evaluated after design: no new violations. The design introduces no new schema, no new
 dependency, and no new query on the hot path; it adds two typed module files and one message
 field, and extracts one shared helper serving two callers. **PASS.**
+
+### Revision: Implementation Sync 2026-07-16
+
+- Reason: reconciled the testing strategy to the tiers that shipped. Selective-dispatch
+  selection is covered at the unit tier (dispatcher branch matrix, gate, selector) and by a
+  real-graph selection test at the component tier driving the dispatcher through a recording
+  workflow backend. The planned `integration_docker/` full live-stack matrix is not used for this
```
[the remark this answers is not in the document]
polmichel: Resolved in `0cb5f6ebe`. The stale `integration_docker/` tier was dropped from the plan test-structure tree to match the shipped unit / component / api tiers described in the revision note.

backend/infrahub/core/merge/regeneration_dispatcher.py
```diff
@@ -0,0 +1,117 @@
...
+        self.log = log
+
+    async def dispatch(
+        self,
+        *,
+        context: InfrahubContext,
+        target_branch: str,
+        merge_diff_cache_key: str | None,
+    ) -> None:
+        if not config.SETTINGS.main.selective_execution_after_merge:
```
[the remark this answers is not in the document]
polmichel: Intentional: this feature ships opt-in. A new merge-behavior change defaults off for a safe rollout, so enabling it is a deliberate operator choice and the `False` default matches the intended contract. Not changing the default.

backend/infrahub/core/merge/selective_regen/definition_selector/artifact_selector.py
```diff
@@ -57,9 +55,9 @@ async def _fetch_member_ids(self, *, definition: ProposedChangeArtifactDefinitio
         )
         return [relationship.peer.id for relationship in group.members.peers]
 
-    def _should_render(self, *, subscriber_id: str | None, managed_branch: bool, impacted: list[str]) -> bool:
+    def _should_render(self, *, subscriber_id: str | None, regenerate_all_members: bool, impacted: list[str]) -> bool:
         return _should_render_artifact(
-            artifact_id=subscriber_id, managed_branch=managed_branch, impacted_artifacts=impacted
+            artifact_id=subscriber_id, managed_branch=regenerate_all_members, impacted_artifacts=impacted
```
ajtmccarty: looks like a leftover `managed_branch=` here

backend/infrahub/core/merge/selective_regen/definition_selector/generator_selector.py
```diff
@@ -56,8 +56,10 @@ async def _fetch_member_ids(
         )
         return [relationship.peer.id for relationship in group.members.peers]
 
-    def _should_render(self, *, subscriber_id: str | None, managed_branch: bool, impacted: list[str]) -> bool:
-        return _run_generator(instance_id=subscriber_id, managed_branch=managed_branch, impacted_instances=impacted)
+    def _should_render(self, *, subscriber_id: str | None, regenerate_all_members: bool, impacted: list[str]) -> bool:
+        return _run_generator(
+            instance_id=subscriber_id, managed_branch=regenerate_all_members, impacted_instances=impacted
```
ajtmccarty: and here

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -175,3 +196,20 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
...
         )
+
+    def _serialize_diff_summary(self, branch_diff: EnrichedDiffRoot) -> list[NodeDiff] | None:
+        try:
+            return self.diff_serializer.serialize(root=branch_diff, target_branch_name=self.destination_branch.name)
+        except Exception:
+            self.log.exception("Failed to serialize merge diff summary; falling back to full regeneration")
+            return None
+
+    async def _write_diff_summary_cache(self, diff_id: str, diff_summary: list[NodeDiff] | None) -> str | None:
```
ajtmccarty: could this be combined with `_serialize_diff_summary` so that this accepts `branch_diff` instead of `diff_summary`? I don't think `branch_diff` needs to be serialized separate from being saved to the cache but maybe I am missing something

ajtmccarty (approved):

backend/infrahub/core/merge/selective_regen/definition_selector/artifact_selector.py
polmichel: Thank you

backend/infrahub/core/merge/orchestrator.py
polmichel: Good point, I'll do it

PR #9932, opened 2026-07-16
feat(computed-attribute): selective recompute of transform-based computed attributes on git updates

gmazoyer:
# Why

On any commit to any linked git repository, every Python transform-based computed attribute is recomputed for every node of its kind, even when the commit touched nothing that feeds that attribute. The commit event carries no diff, so the recompute falls back to a full sweep. On large datasets this is slow and wasteful, and it buries the work that actually matters.

Goal: recompute a Python transform-based computed attribute only when the transform that feeds it actually changed.

Non-goals: artifact-definition and generator consumers of the fingerprint (separate tickets), dropping `.infrahub.yml` from the dependency closure (IFC-2775), and Jinja2 computed attributes (git-independent, already skip via `template_hash`).

Closes IFC-2804

## What changed

Behavioral:

- A commit now recomputes only the computed attributes whose Python transform actually changed, not every transform-based attribute on the branch.
- The change signal is the branch-aware `fingerprint` attribute from IFC-2844. An unchanged fingerprint means no update event and no recompute. An edit and its exact revert produce the same fingerprint, so they trigger no work.
- A transform that declares no `watch` keeps per-commit recompute, but scoped to its own attributes only (the safe over-regenerate default).

Implementation:

- Recompute is driven by the transform node's own lifecycle instead of the commit event. Three triggers on `CoreTransformPython` (created, updated-on-fingerprint, deleted) replace the old commit trigger.
- The fired flow resolves the transform to the attributes it feeds (by name or id) and recomputes only those, reusing the existing recompute fan-out.
- The lifecycle flow also reconciles the node-input recompute automations on every event. That reconciliation used to happen as a side effect of the removed commit trigger, so it runs here in a `finally` and never gets skipped when the recompute leg fails. This is what keeps a transform-only import from silently leaving those automations unbuilt.
- Loop safety comes from the kind and field match (recompute writes land on the attribute's own node kind, never on the transform) and from the live origin match (merge and rebase replays are excluded).

What stayed the same:

- No schema change (the `fingerprint` attribute already exists from IFC-2844), no migration, no generated files.
- The schema-change recompute path and the merge/rebase recompute path are untouched.

```
before:  commit event (no diff)          -> recompute EVERY transform attr x EVERY node
after:   CoreTransformPython fingerprint  -> recompute only that transform's attrs
```

### Suggested review order

1. `dev/specs/ifc-2804-selective-recompute/` (spec, research, plan, contracts) for the intent and the decisions.
2. `backend/infrahub/computed_attribute/recompute_resolution.py` (the pure resolver).
3. `backend/infrahub/computed_attribute/tasks.py` (`process_transform_lifecycle` and the shared reconcile helper).
4. `backend/infrahub/computed_attribute/triggers.py` plus the two catalogue swaps.
5. Tests: unit, then component, then integration.

## How to review

The load-bearing correctness piece is the reconcile-on-every-event in the lifecycle flow. It guards against under-regeneration on transform-only imports (a code review of the MVP caught the hole where removing the commit trigger would otherwise strand the node-input automations). Extra scrutiny there is welcome.

`dev/specs/ifc-2804-selective-recompute/review-followups.md` lists the deferred low-severity items (a reconcile lock, an idempotent double-recompute overlap on first import, and a rename-by-name edge case).

## How to test

```bash
# unit
uv run pytest backend/tests/unit/computed_attribute/ -q
# component (testcontainers)
uv run pytest backend/tests/component/computed_attribute/test_transform_lifecycle_recompute.py --neo4j -q
# integration (testcontainers)
uv run pytest backend/tests/integration/git/test_selective_recompute.py --neo4j -q
```

All of the above pass locally, along with `uv run invoke lint` (ruff, ty, mypy) and `uv run invoke docs.validate`.

## Impact & rollout

- **Backward compatibility:** fingerprints are null until each repository's first import after upgrade. That first import stamps a fingerprint and recomputes each transform's attributes once. This is a one-time cost per transform, not a per-commit sweep.
- **Performance:** recompute work now scales with the number of changed transforms per commit, not the total number of computed attributes on the branch.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (if user-facing or ops-facing change) - not needed, internal recompute behavior, covered by the changelog
- [x] Internal .md docs updated (dev/specs and review-followups)

---
## Summary by cubic
Recomputes Python transform-based computed attributes only when their transform actually changes, replacing the commit-wide sweep with transform lifecycle triggers. First import after upgrade stamps each transform’s branch-aware `fingerprint` and recomputes its attributes once.

- **New Features**
  - Three `CoreTransformPython` lifecycle triggers (create/update-on-`fingerprint`/delete) run `computed-attribute-process-transform-lifecycle`; the old commit trigger is removed.
  - Branch-aware `fingerprint`: unchanged or edit+revert does nothing; transforms without `watch` keep per-commit recompute scoped to themselves.
  - Live-only: filters on `origin=live` to ignore merge/rebase replays; waits for schema convergence; tolerates a missing transform.

- **Bug Fixes**
  - Resolves transforms by name and UUID and dedupes by (kind, attribute) to avoid double or missed recomputes.
  - Always reconciles node-input automations in `finally`, gathering once under a single registry lock to prevent race-induced drops.

<sup>Written for commit b966b43f2fd30cf5a73a346e5091508a79d652ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9932?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/computed_attribute/recompute_resolution.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import TYPE_CHECKING, Protocol
+
+if TYPE_CHECKING:
+    from infrahub.core.schema.schema_branch_computed import PythonDefinition
+
+
+class TransformAttributeMap(Protocol):
+    """The transform -> computed-attributes lookup the resolver reads, keyed by name or id."""
+
+    def get(self, key: str, /) -> list[PythonDefinition] | None: ...
```
polmichel: Why using a specific protocol and not the map typing? `Mapping[str, list[PythonDefinition]]` or `dict[str, list[PythonDefinition]]`

ajtmccarty: looks like `dict[str, list[PythonDefinition] | None]`

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -46,13 +48,53 @@
...
+
+
+async def _reconcile_python_computed_attribute_automations(db: InfrahubDatabase) -> None:
+    """Reconcile the node-input (data-path) automations against the current schema.
+
+    These automations recompute an attribute when a node feeding its transform's query
+    changes. Each trigger type is gathered and applied while holding its trigger-registry
+    lock, so a concurrent reconcile cannot delete an automation another run just created by
+    applying a stale desired set. Reconciling drops the automations of transforms that no
+    longer exist and builds those of transforms that just appeared.
```
ajtmccarty: not sure if all of this really belongs to this function

backend/tests/component/computed_attribute/test_transform_lifecycle_recompute.py
```diff
@@ -0,0 +1,319 @@
...
+    ) -> None:
+        await process_transform_lifecycle(
+            branch_name=default_branch.name,
+            transform_id="00000000-0000-0000-0000-000000000000",
+            action=MutationAction.UPDATED.value,
+            context=self._context(admin_account, default_branch),
+        )
+
+        assert self._submitted_attribute_names(workflow_recorder) == set()
+        assert await self._python_automation_names(), "a missing transform must still reconcile the automations"
```
polmichel, line 185: Optional comment: can we directly assert the exact automation names based on the existing dataset?

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -46,13 +48,53 @@
...
+    triggers_python, _ = await gather_trigger_computed_attribute_python(db=db)
+    return triggers_python
+
+
+async def _gather_computed_attr_python_query_triggers(
+    db: InfrahubDatabase | None,
+) -> list[ComputedAttrPythonQueryTriggerDefinition]:
+    if db is None:
+        raise ValueError("A database session is required to gather Python computed-attribute query triggers")
+    _, triggers_python_query = await gather_trigger_computed_attribute_python(db=db)
```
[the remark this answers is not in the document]
ajtmccarty: seems like reasonable feedback. is this what the docstring below is describing?

backend/tests/component/computed_attribute/test_transform_lifecycle_recompute.py
```diff
@@ -0,0 +1,319 @@
...
+        await process_transform_lifecycle(
+            branch_name=default_branch.name,
+            transform_id=transform_dataset["transform_by_id"],
+            action=MutationAction.UPDATED.value,
+            context=self._context(admin_account, default_branch),
+        )
+
+        assert self._submitted_attribute_names(workflow_recorder) == {"computed_by_id"}
+
+    async def test_created_recomputes_only_changed_transform_attribute(
```
polmichel: should the test name rather be `test_created_recomputes_only_created_transform_attribute`?

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -594,22 +636,73 @@ async def computed_attribute_setup_python(
...
+    name="computed-attribute-process-transform-lifecycle",
+    flow_run_name="Process computed attributes for transform {transform_id} lifecycle ({action})",
+)
+async def process_transform_lifecycle(
+    branch_name: str,
+    transform_id: str,
+    action: str,
+    context: EventContext,
+    event_name: str | None = None,
+) -> None:
```
ajtmccarty, line 636: this flow looks like it should be a thin wrapper around a class (with its dependencies injected) that actually does the work

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -594,22 +636,73 @@ async def computed_attribute_setup_python(
...
+
+                client = get_client()
+                transform = await client.get(
+                    kind=InfrahubKind.TRANSFORMPYTHON, id=transform_id, branch=branch_name, raise_when_missing=False
+                )
+                if transform is None:
+                    log.warning(
+                        f"Transform {transform_id} not found on {branch_name} for {event_name or action}; "
+                        "skipping recompute, still reconciling automations"
+                    )
```
ajtmccarty: I think you can return early here

backend/tests/integration/git/test_selective_recompute.py
ajtmccarty, line 1: these looks like good tests. they don't really seem related to the code changes, but maybe I am missing how they connect.

ajtmccarty (approved):

backend/tests/integration/git/test_selective_recompute.py
```diff
@@ -0,0 +1,229 @@
...
+
+class TestSelectiveRecompute(FingerprintImportTestBase):
+    @pytest.fixture(scope="class")
+    async def prefect_client(self, prefect: str) -> AsyncGenerator[PrefectClient, None]:
+        # `prefect` sets PREFECT_API_URL to the same server the import flow publishes events to,
+        # so the client reads the events a real import actually emitted.
+        async with get_client(sync_client=False) as client:
+            yield client
+
+    async def test_source_edit_then_revert_restores_original_fingerprint(
```
polmichel: This test and the 3 following are exercising the fingerprint capabilities which have been implemented in a previous PR by @ogenstad right? Is this not a duplication ?

For example this one `test_reimport_same_commit_leaves_every_transform_fingerprint_unchanged` seems already tested here `TestFingerprintTransformation.test_reimport_leaves_fingerprint_unchanged`

polmichel (approved):
LGTM, only optional comments

PR #9896, opened 2026-07-14
run fewer constraint validators

ajtmccarty:
# Why

Merging, rebasing, or running proposed-change checks on a branch that only
changed *data* used to select every constraint validator against every kind in
the schema, even kinds the change never touched. That work grows with the size
of the schema, not the size of the change, and dominates merge/rebase time on
large schemas.

**Goal:** scope constraint validation to the kinds a change can actually
affect, so a data-only branch validates only the kinds whose data changed plus
the few kinds that transitively depend on them.

**Non-goals:** no change to what any validator checks, nor to the set of
validators that exist. This only changes *which* validators are selected to run
for a given diff.

Closes #9246 

## What changed

**Behavioral changes:**

- A branch with only data changes no longer triggers constraint validators for
  unrelated kinds. Node-level property constraints are scoped to the kinds
  present in the data diff, the generics they inherit from, and any peer whose
  uniqueness constraint reads an attribute of a changed kind (e.g. a
  `owner__name` uniqueness on `Car` when `Person` data changes).
- Schema-transition-only checkers (`node.generate_profile.update`,
  `node.inherit_from.update`) are no longer run from the data-diff path. A data
  change cannot violate them — they only fire when the schema property itself
  changes — so they stay covered by the schema-diff path and are dropped from
  the data path.
- Internal schema kinds (`SchemaNode`/`SchemaGeneric`/`SchemaAttribute`/
  `SchemaRelationship`) are validated when they appear in the diff (i.e. when a
  schema is edited on the branch) rather than being appended unconditionally.

**Implementation notes:**

- New `triggered_by_data_change` flag on `ConstraintCheckerInterface`
  (defaults `True`), set `False` on `NodeGenerateProfileChecker` and
  `NodeInheritFromChecker`. The determiner consults `CONSTRAINT_VALIDATOR_MAP`
  and skips any constraint whose checker is not data-triggered — the
  classification lives with the checker that embodies it, and a new checker is
  forced to answer the question rather than silently defaulting into the data
  path.
- `ConstraintValidatorDeterminer` computes an "impacted kinds" set (diff kinds +
  inherited generics + peers whose uniqueness reads them) and scopes node-level
  property constraints to it instead of iterating every kind.
- Peer-uniqueness detection uses `parse_schema_path` for a structured view of
  each constraint element instead of string partitioning.
- A kind that is absent from the schema (e.g. a kind whose schema the branch
  also deletes) is skipped instead of raising.

[...]

PR #9943, opened 2026-07-17
fix(backend): scope node-level constraint validators to the changed field/path

ajtmccarty:
# Why

IFC-2604 (#9896) scoped constraint validation to the *kinds* that have a data diff, but within a
changed kind the determiner still emitted every node-level constraint (uniqueness, hierarchy)
regardless of *which* field changed. A data change to a non-unique attribute still triggered
`node.uniqueness_constraints.update` for the whole kind.

**Goal:** emit a node-level constraint only when the specific attribute/relationship/path it guards
is actually part of the diff.

**Non-goals:** the node-scoped *execution* of the uniqueness and other constraint validators
(follow-ups IFC-2796 / IFC-2797, which assume this trigger gate exists); no change to the
schema-diff constraint source, which is already field-scoped.

Part of epic IFC-2706. Closes IFC-2795.

## What changed

**Behavioral changes** — a data change to field X of kind K now emits a node-level constraint only
when X participates in that constraint's guarded path:

- Uniqueness (`node.uniqueness_constraints.update`) fires only when a `unique=True` attribute or a
  `uniqueness_constraints` path element changed. This includes the peer case: a constraint like
  `owner__name` fires only when the related node's `name` changed, not when an unrelated peer
  attribute (e.g. `height`) changed.
- Hierarchy (`node.parent.update` / `node.children.update`) fires only when the parent/children
  relationship changed on that kind — re-parenting a child no longer triggers the children check on
  the parent, and vice-versa.
- Generic-level uniqueness still fires when an implementation's inherited field changes (a
  generic's constraint spans every implementing node).

**Implementation notes:**

- All "only-when-required" trigger logic lives in the determiner (`ConstraintValidatorDeterminer`),
  per the epic's design. A new `_node_property_triggered_by_diff` gate dispatches to
  `_diff_triggers_uniqueness` (unique attributes + constraint-path fields, local and peer) and a
  parent/children relationship-diff check.
- `_field_in_diff` handles generic -> implementation propagation: an inherited field keeps its name
  on the child, so a change on an implementation still triggers the generic's uniqueness gate.
- The coarser `_has_uniqueness_constraint_on_peer` (which fired on any change to a peer *kind*) is
  replaced by the field-precise `_diff_triggers_uniqueness`, reusing `get_attribute_path_from_string`.
- Unknown node-level properties default to emitting, so a newly-added node-level constraint is
  never silently missed.

**What stayed the same:**

- Attribute and relationship constraints were already field-scoped — unchanged.
- The schema-diff source (`MergeSchemaAnalyzer.calculate_validations` ->
  `SchemaUpdateValidationResult.process_diff`) was already scoped per changed property — no code
  change; a guard test locks that in.
- No DB schema or migration changes. Validator execution, constraint payloads, and the API contract
  are unchanged — only *which* validators are triggered.

## How to review

Suggested order:

1. `backend/infrahub/core/validators/determiner.py` — the gate plus the two helpers
   (`_diff_triggers_uniqueness`, `_field_in_diff`) and the peer-selection change in
   `_get_property_constraints_for_impacted_kinds`.
2. `backend/tests/component/core/constraint_validators/test_determiner.py` — narrowed and new
   exact-set assertions.
3. `backend/tests/unit/core/schema/test_schema_update_validation.py` — the schema-diff guard (also
   the new home for the relocated schema-update-validation unit tests, moved out of the
   grab-bag `test_models.py`).

The risk to scrutinize is *under*-selection — narrowing which validators fire must not drop one a
change could actually violate. Extra scrutiny welcome on the peer/generic uniqueness cases in
`_diff_triggers_uniqueness` / `_field_in_diff`; the uniqueness-merge integration suite guards the
end-to-end behavior.

## Impact & rollout

- **Backward compatibility:** no API or schema changes; validation is narrowed only, never widened.
- **Performance:** the motivating win — a data change now runs only the node-level validators whose
  guarded field actually changed, reducing constraint validation work on proposed-change checks,
  merges, and rebases.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (n/a — internal validation behavior)
- [ ] Internal .md docs updated (n/a — internal validation behavior)

---
## Summary by cubic
Scopes node-level constraint validation (uniqueness, hierarchy) to only the specific changed field/path, reducing unnecessary checks and speeding up diff validation. Aligns with IFC-2795; adds precise path parsing, inheritance-aware peer checks, and logs unparseable paths.

- **Bug Fixes**
  - Uniqueness fires only when a unique attribute or a `uniqueness_constraints` path segment changed (local relationship or peer like owner__name); paths parsed via `schema.parse_schema_path`; unparseable elements are skipped and warned.
  - Inheritance-aware: changes to inherited fields trigger generic-level uniqueness; a path that targets a generic peer fires when an implementation’s attribute changes.
  - Hierarchy fires only when `parent` or `children` changed on that kind; each endpoint emits only its changed side.
  - The determiner gates node-level properties per field/path; unknown properties default to emit so new constraints aren’t missed.
  - Schema-diff validation stays field-scoped; guard test added.

<sup>Written for commit 76eeb6ee37279365dad5c9e8646db49484c25548. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9943?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/validators/determiner.py
```diff
@@ -105,27 +105,74 @@ def _get_impacted_kinds(self) -> set[str]:
...
+        """
+        for attribute_schema in schema.unique_attributes:
+            if self._field_in_diff(schema=schema, field_name=attribute_schema.name, is_relationship=False):
+                return True
+        for constraint_group in schema.uniqueness_constraints or []:
+            for constraint_path in constraint_group:
                 try:
                     schema_path = schema.parse_schema_path(path=constraint_path, schema=self.schema_branch)
                 except AttributePathParsingError:
                     continue
```
polmichel, line 140: Not directly tied to this PR but would it not be better to log the message coming from the Error in case this happen?

backend/tests/component/core/constraint_validators/test_determiner.py
```diff
@@ -191,7 +191,7 @@ async def test_many_relationship_update(
         constraints = await determiner.get_constraints(node_diffs=[node_diff])
 
         assert len(constraints) >= len(constraint_info_set)
-        assert constraint_info_set < set(constraints)
+        assert constraint_info_set <= set(constraints)
```
polmichel: Then, this test should assert using a direct equality? The determiner used to retrieve additional unrelated constraints, but this is not the case anymore thanks to your fix. In that case `assert len(constraints) == len(constraint_info_set)` is implicitly already tested through the equality operator.

If another test is already testing this path, this is maybe not necessary as the regression would still be caught elsewhere

polmichel (approved):
LGTM! Two optional comments have been posted

backend/infrahub/core/validators/determiner.py
ajtmccarty: good point. added a log and test for it

backend/tests/component/core/constraint_validators/test_determiner.py
ajtmccarty: updated to `==`

PR #9947, opened 2026-07-17
fix(schema): backport repository cascade delete to stable (closes #9934)

### LINKED ISSUE #9934: git-sync: import fails deleting a CheckDefinition pinned by a validator
**Infrahub version:** 1.10.5 (first observed on 1.10.4; persists on 1.10.5)

**Component:** Backend / Git integration (repository import / git-sync)

### Current Behavior

When a git-sourced check definition is renamed or removed from `.infrahub.yml`, the repository import (git-sync) tries to delete the old `CoreCheckDefinition`. If that check was ever evaluated by a proposed change, a `CoreUserValidator` still references it through the **mandatory `check_definition` relationship**, so the delete is rejected and the **entire repository import aborts** (`sync_status: error-import`).

The import had already loaded all schema and imported the renamed check + generators successfully; only the final cleanup-delete fails, but it fails the whole flow. Re-imports keep failing on the same delete, so the repository is stuck on the old commit and never advances to newer commits.

Error (UUIDs/names redacted):

```
CheckDefinition '<old_name>' not found locally, deleting
Encountered exception during execution: GraphQLError(
  mutation { CoreCheckDefinitionDelete(data: { id: "<check-def-id>" }) { ok } },
  [{'message': "Cannot delete CoreCheckDefinition '<check-def-id>'. It is linked to
    mandatory relationship check_definition on node CoreUserValidator '<validator-id>' ..."}])
Finished in state Failed('Flow run encountered an exception: GraphQLError ...')
```

### Expected Behavior

Removing/renaming a git-sourced check definition should let the import complete. The importer should do one of:

- detach / cascade the historical `CoreUserValidator` references so the check can be deleted, or
- soft-deprecate the `CoreCheckDefinition` when validators still reference it, or
- at minimum, not abort the whole import for a cleanup-delete failure: log a warning, leave the old check in place, but still advance the commit and apply everything else.

A check rename should not require manually deleting validator records from (often already-merged) proposed changes.

### Steps to Reproduce

1. Add a check definition (e.g. `my_check`) to a git repository's `.infrahub.yml`; sync it.
2. Run a proposed change so the check is evaluated (this creates a `CoreUserValidator` linked to the `CoreCheckDefinition` via `check_definition`); merge or close the proposed change.
3. Rename or remove `my_check` in `.infrahub.yml`.
4. Trigger a repository import (or let git-sync run).
5. The import aborts on `CoreCheckDefinitionDelete` with "Cannot delete ... linked to mandatory relationship `check_definition` on node `CoreUserValidator` ..."; the repository `sync_status` becomes `error-import` and stays stuck on the previous commit.

### Additional Information

- Workaround: delete the referencing `CoreUserValidator` node(s), then re-import; the delete then succeeds. This means editing historical validator records from (possibly other people's) merged proposed changes, which is undesirable.
- The same failure mode should apply to any node the importer must delete while a mandatory relationship from a historical record still holds it.
- The schema and the renamed check applied before the abort; only the final delete + commit-advance are blocked.

iddocohen:
# Why

Backport of #9832 to `stable`.

On `stable`, renaming or removing a git-sourced check definition (or transform, generator, or GraphQL query) freezes repository import: the reconcile tries to delete the stale object, but a historical `CoreUserValidator` still pins it through the **mandatory `check_definition` relationship**, so the delete is rejected and the entire import aborts (`sync_status: error-import`). The repository stays stuck on the old commit and every subsequent sync fails identically. This is #9934.

The root cause is that the repository-managed ownership relationships were never wired with `on_delete=CASCADE`. #9832 fixed that on `develop`; this PR backports it so the reconcile delete cascades to the pinning validator and succeeds, letting the import complete.

Non-goals: no integrator-level workaround (the superseded approach in #9937 caught the error and left the stale object behind, which reviewers correctly flagged as too general and prone to silent inconsistency).

Closes #9934
Refs #3076

## What changed

- The repository ownership tree is wired with `on_delete=CASCADE` (`transformations`, `queries`, `checks`, `generators`), with the missing parent-side relationships added so the cascade reaches every managed leaf, including `CheckDefinition → UserValidator`.
- Deleting a repository-managed object now cascades to the historical validators/artifacts/instances that pin it, instead of the delete being rejected. As a result, removing a check definition during git-sync no longer aborts the import.
- Added an integration test proving the git-sync import path: a check definition pinned by a `CoreUserValidator` is removed from config, the import deletes it, the cascade removes the validator, and the import completes.

Implementation notes: this is a faithful backport of #9832. The schema definitions are cherry-picked as-is; `protocols.py`, `schema.graphql`, and the frontend GraphQL types were regenerated on `stable` (not diff-merged from `develop`). `stable` already carries the deep delete-validator engine, so no `on_delete` logic change was needed — only the wiring.

What stayed the same: no graph migration required (`update_core_schema` reconciles the `on_delete` values and new relationships during upgrade); `GRAPH_VERSION` unchanged.

## Accepted trade-off

As in #9832: deleting a repository (or a check/generator/artifact definition directly) now also deletes validators attached to proposed changes and their generated artifacts/instances. This is the intended ownership semantics and lands on `stable` with this backport.

## How to review

- `backend/infrahub/core/schema/definitions/core/*.py` — the `on_delete=CASCADE` wiring and new parent-side relationships (the actual fix).
- Generated files (`protocols.py`, `schema.graphql`, frontend GraphQL types) are mechanical regenerations.
- Tests: `test_delete_validator.py` and `test_node_manager_delete.py` are from #9832; `test_git_repository.py::TestInfrahubClient::test_import_deletes_check_definition_pinned_by_validator` is the added git-sync regression for #9934.

## How to test

```bash
cd backend
uv run pytest tests/unit/core/node/test_delete_validator.py
uv run pytest tests/integration/git/test_git_repository.py::TestInfrahubClient::test_import_deletes_check_definition_pinned_by_validator -x
```

## Impact & rollout

- **Backward compatibility:** delete semantics change per the accepted trade-off above; no API contract change.
- **Deployment notes:** no graph migration; safe to deploy.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated — N/A
- [ ] Internal .md docs updated — N/A

---
## Summary by cubic
Backports cascade delete for repository-owned objects to stable so git imports don’t freeze when a removed definition is still referenced. Deletes now cascade to validators, instances, and artifacts, allowing imports to advance to the new commit. Closes #9934.

- **Bug Fixes**
  - Wired cascade from repositories to their managed objects and dependents, adding missing parent-side relationships (e.g., CheckDefinition → UserValidator, ArtifactDefinition → Artifact/Validator, GeneratorDefinition → Instance/Validator, Transformation → ArtifactDefinition, GraphQLQuery → Group).
  - Git import no longer aborts when a removed check definition is pinned by a validator; reconcile delete succeeds and the repository sync completes.
  - Regenerated `protocols.py`, `schema.graphql`, and frontend GraphQL types; added unit coverage for cascade resolution and a git-sync regression test verifying the fix.

<sup>Written for commit 94314239bac2af33413c6191b6e870a226363f52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9947?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


iddocohen:
@ajtmccarty As you suggested, this backports the schema-level cascade from #9832 to `stable` instead of handling it in the importer, so removing a git-sourced check definition now cascades to its pinning `CoreUserValidator` and the repository import completes rather than aborting. The fix is verified against the original scenario with a new git-sync import regression test, alongside the cascade unit and live-database component suites from #9832.

ajtmccarty (approved):

PR #9948, opened 2026-07-17
feat(frontend): column-header sort & filter menu [IFC-2794]

bilalabbad:
Closes [IFC-2794](https://opsmill.atlassian.net/browse/IFC-2794)

Spec: `specs/002-header-sort-menu/` (spec.md, plan.md, contracts/header-menu-ui.md)

## Summary

Column headers in object lists and IPAM IP address/prefix lists now open a react-aria menu instead of directly opening the per-column filter popover. The menu offers:

- **Sort ascending / Sort descending** on sortable attribute columns — selecting one replaces the entire active sort with a single-field sort on that column; selecting the already-active direction again toggle-clears back to the schema default order.
- **Sort by ▸** submenu on cardinality-one relationship columns, listing the peer's sortable attributes with a direction choice (e.g. sort devices by their site's name).
- **Filter…** (always last, after a separator) — opens the existing per-column filter form in a controlled popover, pre-filled when a filter is active on the column.

The header shows an ↑/↓ indicator when it drives the active user-applied sort (never for the schema default). Sort state is shared with the toolbar Sort control through the existing `?sort=` URL param via `useSort`, so both surfaces always agree and sorted views survive reload and link-sharing. Cardinality-many or unresolvable-peer relationship columns offer Filter… only; columns that are neither sortable nor filterable keep their plain non-interactive header.

The IPAM IP-address and IP-prefix lists previously ignored sorting entirely; this PR wires `sort` end-to-end (API call → use-case → query hook → table) so the shared header's sort actions work there too (release-blocking per spec Assumptions).

## User-facing interaction change

Clicking a column header previously opened the filter form directly. It now opens the menu, and filtering sits under **Filter…** — one click further. The filter experience itself is unchanged and remains fully consistent with the toolbar path: same form, same active-filter tags, same `?filters=` URL, editable/removable from either place (spec Story 3).

A Towncrier fragment covers this: `changelog/+header-sort-menu.changed.md`.

## Implementation notes

- New pure domain rule `getColumnActiveSort(customSort, columnSchema)` (`src/entities/nodes/sort/domain/rules/get-column-active-sort.ts`) decides when a column drives the active sort — token-aware matching (split on `__`), so relationship `site` never matches attribute `site_code`.
- `TableColumnHeader` reworked from a Radix popover to `@infrahub/ui` `MenuTrigger` + `Menu`, with a controlled popover for the filter form (pattern donor: the toolbar sort picker).
- IPAM API calls spread `addOrderByToRequest(sort)` into the GraphQL `__args`, mirroring the object-list wiring.

## Accepted migration debt

- **`TableColumnHeader` location (critique E2)**: the component stays under `entities/nodes/object/ui/object-table/` while now also imported by `entities/ipam/` tables. Moving it to a shared location is deliberate follow-up work outside this PR's scope; the cross-entity import is the accepted interim state.
- **Softened FR-001b of `specs/ifc-2428-filters`** (spec Assumptions): that Draft spec's FR-001b ("column headers are no longer clickable filter triggers") is softened, not contradicted — headers now reuse the unified filter flow as a second entry point. The Draft spec should be amended when next worked on; `specs/002-header-sort-menu/spec.md` is the current source of truth for header behavior.

## IPAM Prefix-column caveat

On the IPAM prefix list, the prefix value renders in the identifier column, which has no header menu (`prefix` is excluded from the list columns by design). Sorting the prefix list therefore goes through the other columns (e.g. Description), which is what the E2E spec exercises.

## Test coverage

- **Unit/domain**: `get-column-active-sort.test.ts` — attribute/relationship match, near-miss names, multi-field sort → null.
- **Component** (vitest browser mode): `table-column-header.test.tsx` — menu structure, full-replace write, toggle-clear, active direction marked, non-sortable kinds, relationship submenu contents, cardinality-many/unresolvable peer → Filter… only, filter form pre-fill and state parity.
- **E2E** (Playwright): new `objects/object-header-sort.spec.ts` (sort, persistence, toggle-clear, toolbar replace semantics, relationship submenu incl. keyboard-only path) and `ipam/ip-prefix-list-sort.spec.ts`; `objects/object-filters.spec.ts` extended with header-menu paths and explicit header/toolbar parity assertions; `objects/object-sort.spec.ts` (toolbar sort) passes unmodified.
- Full frontend unit suite green (134 files / 931 tests); the four affected E2E specs green (14 tests); a scripted browser pass over the quickstart flows confirmed no console errors from overlay/focus management in the menu → filter-popover sequence.

[IFC-2794]: https://opsmill.atlassian.net/browse/IFC-2794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Adds a column‑header menu to sort and filter in object lists and IPAM IP address/prefix lists. Sorting stays in sync with the toolbar via `?sort=`, persists on reload, and IPAM lists now support sorting end‑to‑end; implements IFC‑2794.

- **New Features**
  - Column headers open a menu: Sort ↑, Sort ↓, and Filter; to‑one relationships add “Sort by ▸” with peer attributes; clicking the active direction clears to default; headers show ↑/↓ and menu items show active checks and icons.
  - Sort state is shared with the toolbar (`useSort`, `?sort=`); header sort replaces the whole sort, multi‑field remains toolbar‑only.
  - IPAM IP address and prefix lists wired for sorting (API → use‑case → query); available IPs are hidden when a custom sort is active or filters conflict (and the availability tag hides); filtering now sits under “Filter”.

- **Refactors**
  - `TableColumnHeader` rebuilt with `@infrahub/ui` `MenuTrigger`/`Menu`; prop `disabled` → `isDisabled`; shared `SortableFieldMenuItem`, `SortableAttributeMenuItem`, and `SortDirectionMenuItem` with icons across header and the sort picker.
  - `useSort.setCustomSort` accepts `null` to clear; added `findSortForField`; `getValidSorts` keeps schema‑default sub‑property fields (e.g. `prefix__version`).
  - Consolidated IP availability filter tag into a shared component; removed per‑entity tags; cleanup: stop tracking gitignored `.specify/feature.json`.

<sup>Written for commit 85120333e300f0a5a946911b2fe5d7214eb43e5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9948?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pa-lem (commented):
Reviewed with parallel agents + manual testing. The core claims all check out in code: token-aware sort matching, replace/toggle-clear semantics, multi-field staying toolbar-only, IPAM wiring with query re-keying, cardinality gating. `?sort=` is parsed and schema-allowlisted before reaching GraphQL. Tests are thorough and the TS/Python e2e mirrors are consistent. 👍

Also verified manually: sorting while deep in the list is **fine** — the infinite query refetches the loaded window so the result stays coherent (just a brief skeleton flash). No change needed there.

1 blocker + 1 decision + 3 small items inline. Nits, take or leave:

- `willResetsToDefault` → `willResetToDefault` (sort-editor.tsx)
- relationship "Sort by" submenu omits `variant="picker"` unlike the direction submenu
- `(peerSchema.attributes ?? []).filter(isSortableAttribute)` + weight-sort repeated in `use-sortable-fields.ts` / `add-sort-picker.tsx` / `table-column-header.tsx` — could be one `getSortablePeerAttributes` rule
- `seedSortInUrl`/`seedFiltersInUrl` duplicated across two test files
- PR description says `getColumnActiveSort`; the actual export is `findSortForField`

frontend/app/src/entities/nodes/object/ui/object-table/cells/table-column-header.tsx
```diff
@@ -1,57 +1,273 @@
...
+            <MenuItem textValue="Filter" onAction={() => setShowFilterForm(true)}>
+              <ListFilterIcon />
+              <span>Filter</span>
+            </MenuItem>
+          </Menu>
+        </Popover>
+      </MenuTrigger>
+
+      <Popover
+        triggerRef={triggerRef}
```
pa-lem, line 260: **Blocker (a11y/keyboard):** the filter form renders in a bare `Popover` — no `PopoverDialog` wrapper. So: no dialog role, focus depends on a racy `autoFocus`, and Escape only dismisses once focus is inside. The `.focus()` workaround in `object-filters.spec.ts` is this bug surfacing in CI. Wrapping in `PopoverDialog` (exported by `@infrahub/ui`) with a "Filter by {label}" heading fixes dismissal + a11y and restores the visible title the old radix popover had.

```text
🤖 Prompt: Verify before fixing. In table-column-header.tsx, confirm the controlled
filter Popover renders AttributeFilterForm/RelationshipFilterForm without the
PopoverDialog wrapper exported by @infrahub/ui (frontend/packages/ui/src/components/popover/popover.tsx).
If confirmed: wrap the form in PopoverDialog with accessible name "Filter by {label}",
add a component test that Escape closes the popover WITHOUT focusing it first,
then remove the focus workaround in tests/e2e/objects/object-filters.spec.ts and its
Python mirror. Run the component tests and the object-filters e2e.
```

frontend/app/src/entities/ipam/ip-prefixes/api/get-ip-prefix-list-from-api.ts
```diff
@@ -132,6 +141,7 @@ export function buildGetIpPrefixListWithAvailabilityQuery({
           [AVAILABLE_IP_FILTER_NAME]: true,
           ...(objectKind !== IP_PREFIX_GENERIC ? { kinds: [objectKind] } : {}),
           ...(filters ? addFiltersToRequest(filters) : {}),
+          ...(sort?.length ? addOrderByToRequest(sort) : {}),
```
pa-lem, line 144: **Decision needed:** `order_by` works on top-level lists, but inside a parent prefix the backend merges the available pseudo-rows and re-sorts everything numerically, silently discarding the user's sort — while the header still shows ↑/↓. Since gap rows only make sense in numeric order, honoring the sort server-side isn't realistic. Options: hide sort items in availability views, or drop availability rows while a custom sort is active.

```text
🤖 Prompt: Verify before changing anything. Read backend/infrahub/graphql/resolvers/ipam.py:
confirm _annotate_result only runs when include_available is true AND a parent prefix
resolves (parent__ids filter), and that _resolve_available_prefix_nodes /
_resolve_available_address_nodes end with sorted(existing + available, key=n.prefix.obj),
discarding order_by. If confirmed, this is a product decision — do NOT pick silently:
summarize the two UI options (no sort menu in availability views vs. dropping availability
rows when a custom sort is active) and ask the human which to implement.
```

frontend/app/src/entities/nodes/sort/ui/sort-editor.tsx
```diff
@@ -113,6 +113,20 @@ interface SortFieldSelectProps {
   onChange: (field: SortField) => void;
 }
 
+/**
+ * Readable label for a sort field the picker doesn't offer — a schema default order can target an
+ * attribute sub-property (e.g. IP prefixes sort on `prefix__version`) while the picker only exposes
+ * `__value` fields. Renders as "Prefix › version", matching the peer-attribute label style.
+ */
+function describeUnlistedSortField(field: SortField, schema: ModelSchema): string {
```
pa-lem: This assumes every unlisted field is an attribute path: a relationship default sort like `site__name__value` renders as "site › name · value", leaking the internal `__value` suffix users never see elsewhere. Also `[attributeName = field]` is dead code — `split` can't return an empty first token.

```text
🤖 Prompt: Verify before fixing. Call describeUnlistedSortField with field "site__name__value"
and a schema where site is a relationship (not an attribute): confirm it returns a label
containing the raw "value" segment. If confirmed: check schema.relationships for the first
token and label via the peer relationship, or fall back to the raw field when the shape
isn't a known attribute sub-property. Remove the dead `= field` destructuring default.
Add a unit test for the relationship case.
```

frontend/app/src/entities/nodes/object/ui/object-table/cells/table-column-header.tsx
```diff
@@ -1,57 +1,273 @@
...
+                  <ArrowDownIcon className="size-4 text-indigo-700" />
+                  <span className="sr-only">sorted descending</span>
+                </>
+              ) : (
+                <>
+                  <ArrowUpIcon className="size-4 text-indigo-700" />
+                  <span className="sr-only">sorted ascending</span>
+                </>
+              ))}
+            {currentColumnFilters && <ListFilterIcon className="size-4 text-indigo-700" />}
```
pa-lem, line 239: The sort arrows above have `sr-only` text but this active-filter icon doesn't — filtered state is invisible to screen readers. One-liner: `<span className="sr-only">filtered</span>` next to the icon.

```text
🤖 Prompt: Verify before fixing. Confirm the ListFilterIcon rendered when
currentColumnFilters is truthy has no accompanying sr-only text, unlike the sort
indicator right above it. If confirmed, add <span className="sr-only">filtered</span>
beside it and extend the existing indicator component test to assert it.
```

frontend/app/src/entities/nodes/object/ui/object-table/utils/get-object-table-columns.tsx
```diff
@@ -89,7 +91,7 @@ export function getObjectGenericColumns(schema: ModelSchema): Array<ColumnDef<No
 
 export function getObjectFieldsColumns(
   schema: ModelSchema,
-  headerProps?: PopoverTriggerProps
+  headerProps?: Partial<TableColumnHeaderProps>
```
pa-lem, line 94: `Partial<TableColumnHeaderProps>` spread after `schema`/`columnSchema` lets a caller silently override the column schema. Only `isDisabled` is passed today — `Pick<TableColumnHeaderProps, "isDisabled">` states intent and closes that door (same shape in the IPAM column files).

```text
🤖 Prompt: Verify before fixing. Grep all callers passing headerProps: confirm only
isDisabled is ever passed (relationship-table.tsx) and that the spread comes after the
explicit schema/columnSchema props. If confirmed, narrow the type to
Pick<TableColumnHeaderProps, "isDisabled"> here and in the IPAM get-*-table-columns
files, and check typecheck still passes.
```

frontend/app/src/entities/nodes/object/ui/object-table/cells/table-column-header.test.tsx
```diff
@@ -0,0 +1,386 @@
...
+    await component.getByRole("button", { name: "Owner" }).click();
+
+    // THEN
+    await expect.element(component.getByRole("menuitem", { name: "Filter" })).toBeVisible();
+    await expect
+      .element(component.getByRole("menuitem", { name: "Sort by" }))
+      .not.toBeInTheDocument();
+  });
+
+  test("keeps the pagination offset unchanged when sorting from the header", async () => {
```
pa-lem, line 337: Correction to my earlier "sort-while-paginated is fine" note: with a bigger dataset (10k tags) we hit a page-stitching bug. Repro: sort asc → jump deep into the list (~row 4000) → switch to desc from the header → scroll back up. The table showed the *tail* of the desc list (…3, 2, 1) immediately followed by rows from a different offset (9780, 9779, …) — adjacent pages from inconsistent offsets. Resetting to page 1 on sort change is the safer semantics (and the common convention); this test would then assert the reset instead of pinning the current behavior.

```text
🤖 Prompt: Reproduce before fixing. Load ~10k objects (e.g. BuiltinTag), sort ascending,
navigate deep into the list (~row 4000), switch to descending via the column header,
then scroll/page back up. Confirm the table stitches pages from inconsistent offsets
(end-of-list rows followed by mid-list rows). If confirmed: reset the pagination offset
to 0 whenever the sort changes (both the header menu and toolbar setCustomSort paths),
update this test to assert the reset instead of the preserved offset, and re-run the
object-header-sort e2e plus the toolbar sort e2e.
```

frontend/app/src/entities/ipam/ip-prefixes/api/get-ip-prefix-list-from-api.ts
bilalabbad: done

frontend/app/src/entities/nodes/object/ui/object-table/utils/get-object-table-columns.tsx
bilalabbad: `Partial<TableColumnHeaderProps>` already communicates the intent: "any of the header's props." Narrowing to `Pick<..., "isDisabled"> `optimizes for a constraint we don't actually have and would have to re-widen the moment requirements change.

frontend/app/src/entities/nodes/object/ui/object-table/cells/table-column-header.tsx
bilalabbad: No it's expected to not have a PopoverDialog. Adding it  breaks the enum/dropdown/relationship filters.
those render a defaultOpen combobox in a portal outside the dialog, and react-aria's useDialog steals focus on mount, closing the combobox

frontend/app/src/entities/nodes/object/ui/object-table/cells/table-column-header.test.tsx
bilalabbad: I'll do it in another PR

pa-lem (approved):
is it possible to completely remove the `.specify/feature.json` file?

other than that seems good

PR #9952, opened 2026-07-17
fix(graphql): normalize non-str node_type in NodeNotFoundError (closes #9926)

### LINKED ISSUE #9926: bug: non-str node_type in NodeNotFoundError crashes GraphQL error formatter
## Component

API Server / GraphQL

## Infrahub version

1.10.5

## Current Behavior

A GraphQL HTTP request that produces a `NodeNotFoundError` whose `node_type` attribute is a **class** (a `type[SchemaProtocol]`) rather than a string causes the GraphQL error formatter itself to raise a second, unhandled exception. The request returns **HTTP 500** with an ASGI `ExceptionGroup` ("unhandled errors in a TaskGroup" / uvicorn "Exception in ASGI application") instead of a normal GraphQL error response.

The failure is in the error-formatting path, so it converts what should be a clean, catalogued `NODE_NOT_FOUND` GraphQL error into an uncaught 500:

- `backend/infrahub/graphql/error_formatter.py`, `_build_payload()`:
  ```python
  case "NODE_NOT_FOUND" if isinstance(exc, NodeNotFoundError):
      payload = NodeNotFoundData(node_kind=exc.node_type, identifier=exc.identifier)
  ```
- `NodeNotFoundData.node_kind` is typed `str` (`backend/infrahub/errors/payloads.py`), so when `exc.node_type` is a class, pydantic raises:
  ```
  pydantic_core.ValidationError: 1 validation error for NodeNotFoundData
  node_kind
    Input should be a valid string [type=string_type, input_value=<class 'infrahub.core.protocols.CoreMenuItem'>, input_type=type]
  ```
- Because this happens inside `format_graphql_errors()` (called from `graphql/app.py` `_handle_http_request`), it is not caught and propagates as an `ExceptionGroup` → 500.

How `node_type` ends up as a class: `NodeNotFoundError.__init__` annotates `node_type: str` but does not coerce it, and several `NodeManager` methods raise the error with the caller's unresolved `kind` argument, which is typed `str | type[SchemaProtocol]`. For example `NodeManager.get_one()`:

```python
if not result:
    if raise_on_error:
        raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
```

(`get_one_by_id_or_default_filter()` has the same pattern.) When a caller passes a protocol class as `kind`, `node_type` becomes that class. Sibling methods in the same module resolve the kind to a string first (`kind_str = node_schema.kind`) before raising, so the behavior is inconsistent.

Abbreviated traceback:

```
File "backend/infrahub/graphql/app.py", line 296, in _handle_http_request
    response["errors"] = format_graphql_errors(list(result.errors))
File "backend/infrahub/graphql/error_formatter.py", line 160, in format_graphql_errors
    out.append(catalogue_error_formatter(error))
File "backend/infrahub/graphql/error_formatter.py", line 134, in catalogue_error_formatter
    extensions = build_catalogue_extensions(error.original_error)
File "backend/infrahub/graphql/error_formatter.py", line 127, in build_catalogue_extensions
    "data": _build_payload(exc, code),
File "backend/infrahub/graphql/error_formatter.py", line 52, in _build_payload
    payload = NodeNotFoundData(node_kind=exc.node_type, identifier=exc.identifier)
pydantic_core._pydantic_core.ValidationError: 1 validation error for NodeNotFoundData
node_kind
  Input should be a valid string [type=string_type, input_value=<class 'infrahub.core.protocols.CoreMenuItem'>, input_type=type]
```

## Expected Behavior

A `NodeNotFoundError` surfaced through GraphQL should always produce a normal, catalogued GraphQL error response (code `NODE_NOT_FOUND`, HTTP 404), regardless of whether `node_type` was passed as a string or a schema/protocol class internally. The error formatter should not be able to raise while formatting an error.

## Steps to Reproduce

The trigger is any GraphQL request whose resolver raises `NodeNotFoundError` with a class-valued `node_type`. Minimal mechanism:

1. Cause a `NodeManager` lookup that raises with an unresolved class kind, e.g. call a resolver path that reaches `NodeManager.get_one(kind=<SchemaProtocol subclass>, raise_on_error=True, id=<missing id>)` (or `get_one_by_id_or_default_filter` with a class `kind`). `node_type` is then the class object rather than its kind string.
2. Surface that error through a GraphQL HTTP request.
3. Observe an HTTP 500 / ASGI `ExceptionGroup` with the pydantic `ValidationError` above, instead of a `NODE_NOT_FOUND` GraphQL error.

A direct unit-level repro of the formatter alone:

```python
from infrahub.exceptions import NodeNotFoundError
from infrahub.graphql.error_formatter import build_catalogue_extensions
from infrahub.core.protocols import CoreMenuItem  # any class

build_catalogue_extensions(NodeNotFoundError(node_type=CoreMenuItem, identifier="x"))
# -> pydantic ValidationError instead of a formatted extensions dict
```

## Additional Information

Observed in a self-hosted 1.10.5 deployment. The crash fired on every affected GraphQL request for ~5 minutes (09:25:37–09:30:36), 6,688 uncaught 500s total (~22/sec), and **zero** GraphQL errors formatted successfully in that window — i.e. this error path never recovers on its own; it always 500s.

Scope of impact, for what it's worth: only the affected GraphQL requests were hit. REST endpoints (including `/api/menu`) continued returning normally throughout, and this did not take the server down — the main symptom is uncaught 500s plus heavy error-log spam.

Context on how we hit it (circumstantial, not a proven cause): the crashes began ~2 seconds after a Git repository object-import created a batch of `CoreMenuItem` menu entries on a working branch, while a browser session was viewing that branch. The crashes stopped when that session navigated back to `main`; deleting the branch a few minutes later was unrelated to the crashes stopping. We could not reproduce the exact resolver afterward because the branch was deleted, but the formatter bug itself is independent of the trigger.

Suggested fix directions (any one closes the crash; ideally the first for defence in depth): (a) make `_build_payload` robust, e.g. `node_kind=str(exc.node_type)`; (b) normalize `node_type` to a string in `NodeNotFoundError.__init__`; (c) resolve `kind` to its kind string at the `NodeManager.get_one` / `get_one_by_id_or_default_filter` raise sites, matching the sibling methods that already use `kind_str`.

iddocohen:
## Analyst's findings (summary)

> **Root cause:** `NodeNotFoundError.node_type` can hold a schema/protocol **class** (`type[SchemaProtocol]`) rather than a string. The GraphQL error formatter feeds it straight into `NodeNotFoundData(node_kind=...)`, whose `node_kind` field is typed `str`, so pydantic raises a `ValidationError` **inside** the error-formatting path — turning a would-be `NODE_NOT_FOUND` (404) response into an uncaught ASGI 500.
>
> **Affected files:**
> - `backend/infrahub/graphql/error_formatter.py` — `_build_payload()` builds `NodeNotFoundData(node_kind=exc.node_type, ...)` with no coercion (the crash site).
> - `backend/infrahub/errors/payloads.py` — `NodeNotFoundData.node_kind: str` (`extra="forbid", frozen=True`).
> - `backend/infrahub/exceptions.py` — `NodeNotFoundError.__init__` annotates `node_type: str` but stores it verbatim with no coercion.
> - `backend/infrahub/core/manager.py` — `get_one` and `get_one_by_id_or_default_filter` raise with an unresolved class `kind`, unlike sibling sites that resolve to `kind_str` first.

## Replication test

**Test file:** `backend/tests/unit/graphql/test_error_formatter.py`
**Test name:** `test_build_catalogue_extensions_handles_class_node_type`

**What it tests:** Formatting a `NodeNotFoundError` whose `node_type` is a class produces a normal catalogued `NODE_NOT_FOUND` / 404 payload with a string `node_kind`, instead of raising.

**Verification:** Test confirmed FAILING on the pre-fix code.
**Failure reason:** `build_catalogue_extensions` raised a pydantic `ValidationError` on `node_kind` inside `_build_payload`, proving the formatter can itself crash while formatting an error.

<details><summary>Failure output (last 20 lines)</summary>

```
backend/tests/unit/graphql/test_error_formatter.py:140: in test_build_catalogue_extensions_handles_class_node_type
    extensions = build_catalogue_extensions(exc)
backend/infrahub/graphql/error_formatter.py:127: in build_catalogue_extensions
    "data": _build_payload(exc, code),
backend/infrahub/graphql/error_formatter.py:52: in _build_payload
    payload = NodeNotFoundData(node_kind=exc.node_type, identifier=exc.identifier)
E   pydantic_core._pydantic_core.ValidationError: 1 validation error for NodeNotFoundData
E   node_kind
E     Input should be a valid string [type=string_type, input_value=<class 'infrahub.core.protocols.CoreMenuItem'>, input_type=type]
E       For further information visit https://errors.pydantic.dev/2.12/v/string_type
```

</details>

## Test expectations

The test asserts the observable behavior of the error formatter for a class-valued `node_type`: it must not raise, and it must return `code == "NODE_NOT_FOUND"`, `http_status == 404`, `data["identifier"] == "x"`, and a string `data["node_kind"]` that conveys the kind name. The `node_kind` assertion is deliberately tolerant of the exact coercion strategy (kind name vs. repr) so it validates the contract without over-constraining the fix.

---
## Summary by cubic
Fixes a crash in the GraphQL error formatter when `NodeNotFoundError` carried a non-string `node_type`, which returned a 500 instead of a 404. The formatter now always returns a `NODE_NOT_FOUND` payload with a string `node_kind`.

- **Bug Fixes**
  - Normalize kinds at `NodeManager` raise sites (`get_one`, `get_one_by_id_or_default_filter`) using a new `get_kind_str` helper to always pass a string into `NodeNotFoundError`.
  - Coerce `node_kind` with `str()` in `_build_payload` so formatting never raises; added unit tests for class-valued `node_type` and for `get_kind_str`.

<sup>Written for commit f2fe776885929295507997e03e2c33f33cfc56a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9952?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -876,7 +876,8 @@ class NodeManager:
             branch_agnostic=branch_agnostic,
         )
         if not node:
-            raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+            kind_str = get_schema(db=db, branch=branch, node_schema=kind).kind
+            raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=id)
         return node
 
     @overload
@@ -1059,7 +1060,8 @@ class NodeManager:
 
         if not result:
             if raise_on_error:
-                raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+                kind_str = get_schema(db=db, branch=branch, node_schema=kind).kind if kind else "Node"
+                raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=id)
             return None
 
         node = result[id]
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -84,6 +84,16 @@ def get_schema[SchemaProtocol](
     return node_schema
 
 
+def get_kind_str(kind: str | type | MainSchemaTypes | None) -> str:
+    if isinstance(kind, str):
+        return kind
+    if isinstance(kind, type) and issubclass(kind, CoreNode):
+        return kind.__name__
+    if isinstance(kind, MainSchemaTypes):
+        return kind.kind
+    return "Unknown kind"
+
+
 class NodeManager:
     @overload
     @classmethod
@@ -876,8 +886,7 @@ class NodeManager:
             branch_agnostic=branch_agnostic,
         )
         if not node:
-            kind_str = get_schema(db=db, branch=branch, node_schema=kind).kind
-            raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=id)
+            raise NodeNotFoundError(branch_name=branch.name, node_type=get_kind_str(kind), identifier=id)
         return node
 
     @overload
@@ -1060,8 +1069,7 @@ class NodeManager:
 
         if not result:
             if raise_on_error:
-                kind_str = get_schema(db=db, branch=branch, node_schema=kind).kind if kind else "Node"
-                raise NodeNotFoundError(branch_name=branch.name, node_type=kind_str, identifier=id)
+                raise NodeNotFoundError(branch_name=branch.name, node_type=get_kind_str(kind), identifier=id)
             return None
 
         node = result[id]
--- a/backend/infrahub/exceptions.py
+++ b/backend/infrahub/exceptions.py
@@ -244,12 +244,15 @@ class NodeNotFoundError(Error):
     HTTP_CODE: int = 404
 
     def __init__(
-        self, node_type: str, identifier: str, branch_name: str | None = None, message: str | None = None
+        self, node_type: str | type, identifier: str, branch_name: str | None = None, message: str | None = None
     ) -> None:
-        self.node_type = node_type
+        # Callers may pass a schema/protocol class as the kind rather than its kind string;
+        # normalize to a string so downstream consumers (e.g. error payload serialization)
+        # always see a plain kind name.
+        self.node_type = node_type if isinstance(node_type, str) else getattr(node_type, "__name__", str(node_type))
         self.identifier = identifier
         self.branch_name = branch_name
-        self.message = message or f"Unable to find the node {identifier} / {node_type} in the database."
+        self.message = message or f"Unable to find the node {identifier} / {self.node_type} in the database."
         super().__init__(self.message)
 
     def __str__(self) -> str:
--- a/backend/infrahub/exceptions.py
+++ b/backend/infrahub/exceptions.py
@@ -244,15 +244,12 @@ class NodeNotFoundError(Error):
     HTTP_CODE: int = 404
 
     def __init__(
-        self, node_type: str | type, identifier: str, branch_name: str | None = None, message: str | None = None
+        self, node_type: str, identifier: str, branch_name: str | None = None, message: str | None = None
     ) -> None:
-        # Callers may pass a schema/protocol class as the kind rather than its kind string;
-        # normalize to a string so downstream consumers (e.g. error payload serialization)
-        # always see a plain kind name.
-        self.node_type = node_type if isinstance(node_type, str) else getattr(node_type, "__name__", str(node_type))
+        self.node_type = node_type
         self.identifier = identifier
         self.branch_name = branch_name
-        self.message = message or f"Unable to find the node {identifier} / {self.node_type} in the database."
+        self.message = message or f"Unable to find the node {identifier} / {node_type} in the database."
         super().__init__(self.message)
 
     def __str__(self) -> str:
--- a/backend/infrahub/graphql/error_formatter.py
+++ b/backend/infrahub/graphql/error_formatter.py
@@ -49,7 +49,10 @@ def _build_payload(exc: BaseException | None, code: str) -> dict[str, Any]:
     payload: BaseModel = UndefinedErrorData()
     match code:
         case "NODE_NOT_FOUND" if isinstance(exc, NodeNotFoundError):
-            payload = NodeNotFoundData(node_kind=exc.node_type, identifier=exc.identifier)
+            # ``node_type`` is normally a kind string, but guard against a non-str value
+            # so formatting an error can never itself raise.
+            node_kind = exc.node_type if isinstance(exc.node_type, str) else getattr(exc.node_type, "__name__", None)
+            payload = NodeNotFoundData(node_kind=node_kind or str(exc.node_type), identifier=exc.identifier)
         case "AUTHENTICATION_REQUIRED":
             payload = AuthenticationRequiredData()
         case "TOKEN_EXPIRED":
--- a/backend/infrahub/graphql/error_formatter.py
+++ b/backend/infrahub/graphql/error_formatter.py
@@ -49,10 +49,7 @@ def _build_payload(exc: BaseException | None, code: str) -> dict[str, Any]:
     payload: BaseModel = UndefinedErrorData()
     match code:
         case "NODE_NOT_FOUND" if isinstance(exc, NodeNotFoundError):
-            # ``node_type`` is normally a kind string, but guard against a non-str value
-            # so formatting an error can never itself raise.
-            node_kind = exc.node_type if isinstance(exc.node_type, str) else getattr(exc.node_type, "__name__", None)
-            payload = NodeNotFoundData(node_kind=node_kind or str(exc.node_type), identifier=exc.identifier)
+            payload = NodeNotFoundData(node_kind=str(exc.node_type), identifier=exc.identifier)
         case "AUTHENTICATION_REQUIRED":
             payload = AuthenticationRequiredData()
         case "TOKEN_EXPIRED":
--- a/backend/tests/unit/core/test_manager.py
+++ b/backend/tests/unit/core/test_manager.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from infrahub.core.manager import get_kind_str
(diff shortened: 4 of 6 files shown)


backend/infrahub/graphql/error_formatter.py
```diff
@@ -49,7 +49,10 @@ def _build_payload(exc: BaseException | None, code: str) -> dict[str, Any]:
     payload: BaseModel = UndefinedErrorData()
     match code:
         case "NODE_NOT_FOUND" if isinstance(exc, NodeNotFoundError):
-            payload = NodeNotFoundData(node_kind=exc.node_type, identifier=exc.identifier)
+            # ``node_type`` is normally a kind string, but guard against a non-str value
+            # so formatting an error can never itself raise.
```
polmichel: I do not think that this comment is necessary, the code looks self-explanatory

backend/infrahub/exceptions.py
```diff
@@ -244,12 +244,15 @@ class NodeNotFoundError(Error):
     HTTP_CODE: int = 404
 
     def __init__(
-        self, node_type: str, identifier: str, branch_name: str | None = None, message: str | None = None
+        self, node_type: str | type, identifier: str, branch_name: str | None = None, message: str | None = None
     ) -> None:
-        self.node_type = node_type
+        # Callers may pass a schema/protocol class as the kind rather than its kind string;
+        # normalize to a string so downstream consumers (e.g. error payload serialization)
+        # always see a plain kind name.
```
polmichel: I do not think that this comment is necessary, the code looks self-explanatory

backend/infrahub/exceptions.py
```diff
@@ -244,12 +244,15 @@ class NodeNotFoundError(Error):
     HTTP_CODE: int = 404
 
     def __init__(
-        self, node_type: str, identifier: str, branch_name: str | None = None, message: str | None = None
+        self, node_type: str | type, identifier: str, branch_name: str | None = None, message: str | None = None
```
polmichel: The API is clear and expects a `str`. In my opinion, we should keep this API simple and keep only a `str`-typed parameter, the caller should adapt its code and provide a `str`.

Our IDEs show a warning in `infrahub.core.manager.NodeManager.get_one` (manager.py:1062): "_Expected type 'str', got 'str | type[SchemaProtocol] | None' instead_". I think the `type` → `str` transformation should happen in the caller.
Another caller seems to pass something other than a `str` and would need the same fix: `infrahub.core.manager.NodeManager.get_one_by_id_or_default_filter` (manager.py:879).

polmichel (commented):
Thanks for the fix.
A comment about API parameters typing, the unit test will also need to be changed since the API to be tested should be those containing the type variable typed as `type` and not `str`.

backend/infrahub/exceptions.py
polmichel: Otherwise, we could fairly say that this exception has two different responsibilities:
- Being an `NodeNotFoundError` and related `__str__` capabilities
- Serializing the type object into a string

ajtmccarty: I agree. I think the real fix is making sure that the `NodeNotFoundError` gets the `str` it expects and not a `type`

backend/infrahub/graphql/error_formatter.py
iddocohen: Removed the comment so the str() coercion stands on its own, please review again.

backend/infrahub/exceptions.py
iddocohen: Removed this comment as suggested, please review again.

backend/infrahub/exceptions.py
iddocohen: Reverted the constructor to a str-only node_type and moved the type to str resolution into both NodeManager.get_one and get_one_by_id_or_default_filter to match the sibling raise sites, please review again.

iddocohen: The exception no longer serializes the type and only holds the resolved kind string, with the conversion handled by the callers, please review again.

backend/infrahub/core/manager.py
```diff
@@ -876,7 +876,8 @@ async def get_one_by_id_or_default_filter(
             branch_agnostic=branch_agnostic,
         )
         if not node:
-            raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+            kind_str = get_schema(db=db, branch=branch, node_schema=kind).kind
```
ajtmccarty: this works, but is seems like overkill. this function is meant to get the whole schema and we really just need the name of it. I think a small helper function like this would be better 

```python
def get_kind_str(kind: type[SchemaProtocol] | MainSchemaTypes | str) -> str:
    if isinstance(str):
        return str
    if isinstance(kind, type) and issubclass(kind, CoreNode):
        return kind.__name__
    if isinstance(kind, MainSchemaTypes):
        return kind.get_kind()
    else:
        return "Unknown kind"
```

this is not tested, but something like this should do it

iddocohen: Added a get_kind_str helper that resolves the kind name from a str, a schema or protocol class, or a schema object without a database lookup, used it at both raise sites, and added unit tests for it, please review again.

saltas888:
Resolved by http://github.com/opsmill/infrahub/pull/9977

PR #9956, opened 2026-07-17
fix(schema): keep the cached derived-value filter for own-id recomputes

gmazoyer:
## Why

A node's display label and human-friendly id can include a value taken from a related node (read across a relationship). When Infrahub had to rebuild that value for the node itself, it sometimes did not update and kept showing the old value.

## How it was found

This turned up while working on #9845 (refresh a reader when its read peer is deleted). After deleting a peer and merging, the reader's computed attribute refreshed, but its display label kept naming the deleted peer. The real cause was a separate, older bug in how display labels and HFIDs are rebuilt, so it gets its own fix here.

## The bug

Each kind keeps one cached template for its display label and one for its HFID. Two paths share that cached template: one rebuilds the node by its own id, the other rebuilds it because a related node changed. The related-node path changed a lookup setting on the shared template and never reset it. So once a related-node rebuild had run, the own-id path used the wrong lookup, matched no node, and left the stored value stale.

## The fix

The related-node path now works on a copy instead of editing the shared cached template, so the own-id path keeps its correct lookup.

## Why this is separate from #9845 and does not replace it

They fix different bugs in different places:

- #9845 fixes the merge itself. When a peer is deleted, the old code looked for readers starting from the deleted peer and found none, so it never asked to rebuild them. #9845 makes the merge ask for the rebuild by the reader's own id. That alone fixes the main case (the computed attribute), so #9845 is worth merging on its own.
- This PR fixes an older, unrelated bug in how display labels and HFIDs are rebuilt, in a different part of the code. It is wrong with or without #9845.

So this PR does not supersede #9845. #9845 stays the main fix; this one clears the remaining display label gap. The delete-peer display label refresh only works once both are in.

## Tests

- New unit test: rebuilding from a related node must not change the cached own-id lookup. It fails on the current code and passes with the fix.
- Checked end to end with #9845 applied on top: after a related-node change followed by a peer delete, the reader's display label refreshes (`dnode via beta` becomes `dnode via None`) instead of staying stale.

### DIFF
--- a/backend/infrahub/core/schema/schema_branch_display.py
+++ b/backend/infrahub/core/schema/schema_branch_display.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import hashlib
 from copy import deepcopy
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


backend/infrahub/core/schema/schema_branch_display.py
```diff
@@ -2,7 +2,7 @@
 
 import hashlib
 from copy import deepcopy
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
```
gmazoyer, line 5: TIL about `replace()`

ajtmccarty (approved):
looks like a tough one to find. maybe worth sending to `stable` first?

PR #9970, opened 2026-07-19
feat: add IPAddress attribute kind for bare IP addresses [INFP-551]

PhillSimonds:
## Summary

Adds a new `IPAddress` attribute kind for storing **bare** IP addresses (IPv4 or IPv6) with no netmask/prefix, alongside the existing `IPHost` and `IPNetwork` kinds. Resolves the recurring pain of `IPHost` silently appending `/32`·`/128` for data that has no subnet context (DNS records, NTP/syslog/SNMP server addresses, monitoring targets, etc.), and sidesteps the mask-inconsistency class of bug (#8896) for bare-IP use cases.

Implements INFP-551 (targets the 1.11 release). Built via the Spec Kit flow — spec/plan/tasks/research live in `dev/specs/infp-551-ipaddress-attribute-kind/`.

| Kind | Python mapping | Stores | Example |
|---|---|---|---|
| `IPAddress` (new) | `ipaddress.ip_address` | bare IP, no prefix | `192.168.1.1` |
| `IPHost` | `ipaddress.ip_interface` | IP + prefix | `192.168.1.1/24` |
| `IPNetwork` | `ipaddress.ip_network` | network + prefix | `192.168.1.0/24` |

## Design decision — dedicated DB node type

`IPAddress` gets its **own** `AttributeIPAddress` graph value-node label (not a reuse of `AttributeIPHost`), with its own `binary_address` RANGE index and write-path Cypher, storing `value` + `version` + `binary_address` but **no** `prefixlen`. This keeps bare addresses self-describing in the graph and preserves range/containment querying. See `dev/knowledge/backend/database-schema.md`.

## Changes

**Backend**
- `types.py`: register the `IPAddress` data type; `core/attribute.py`: `IPAddress`/`IPAddressOptional` classes (validate as IP, reject any `/`).
- Dedicated storage: `AttributeDBNodeType.IPADDRESS` flag, `GraphAttributeIPAddress*` graph schema node, `attr_ipaddress_bin` index, write/update-path Cypher.
- GraphQL `IPAddress` type (`value`, `version`).
- `GRAPH_VERSION` 74 → 75 + migration `m075` creating the index (no data backfill — new label applies to new writes only).

**Frontend**
- Additive kind constant + list-view membership, icon, and input/display/table-cell/filter switch cases (plain text input, no prefix selector). Regenerated GraphQL types.

**Docs**
- Attribute-kinds reference (regenerated) + `nodes-and-attributes` topic + a `dev/knowledge/backend` note on the new DB node type.

## Testing

- Backend: 1844 unit tests pass; component tests for attribute validation/`to_db`, node DB round-trip + filtering, GraphQL query + HFID round-trip, create-mutation prefix rejection, graph constraints, and the `m075` migration.
- Frontend: list-view unit test; biome + betterer clean.
- **End-to-end** on a locally-built image + the Python SDK: an `IPAddress` attribute reads back as a bare `IPv4Address`/`IPv6Address`, while `IPHost`/`IPNetwork` still return prefixed objects (no regression).

## Cross-repo / merge order

The Python SDK companion change is **opsmill/infrahub-sdk-python#1190** (base `infrahub-develop`). Per the submodule rules, this PR does **not** yet bump the `python_sdk` pointer — that lands as a follow-up commit here once #1190 is merged.

## Out of scope

Bulk data-migration tooling to convert existing `IPHost` `/32`·`/128` attributes into `IPAddress` (the generic attribute-kind-change path still works for a manual change).

---
## Summary by cubic
Adds a new `IPAddress` attribute kind for storing bare IPv4/IPv6 addresses (no prefix), fixing the `/32`/`/128` suffix issue with `IPHost` and meeting INFP-551’s requirement to handle IPs with no subnet context.

- New Features
  - Backend: `IPAddress` attribute with strict validation (rejects any `/`), dedicated `AttributeIPAddress` value node and `binary_address` RANGE index; query/utils wiring for binary range lookups.
  - GraphQL: new `IPAddress` type exposing `value` and `version`.
  - Frontend: added `IPAddress` kind, icon, and input/display/filter handling; regenerated GraphQL types, including `graphql-env.d.ts` and `graphql-cache.d.ts` (gql.tada).
  - Docs/specs and tests updated; added DB constraint assertions for `AttributeIPAddress`; GraphQL types regenerated.

- Migration
  - Graph version bumped to 75 with migration `m075` creating the `attr_ipaddress_bin` index.
  - No data backfill; new label applies to new writes only.
  - Run standard migrations; no breaking changes for existing `IPHost`/`IPNetwork`.

<sup>Written for commit ddf8729b23a69f5a803a916eeb587a27047e039b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9970?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ajtmccarty:
adding a new `AttributeIPAddress` type vertex to the database feels excessive for this case. we've already got two IP vertices that get special handling throughout our data serialization, de-serialization, and some cypher queries. adding a third one to handle just to handle the "no `/32` or `/128` for this host case" feels like too much. I have not investigated this approach, but perhaps adding a `bare` or `no_prefix` Boolean property to the `AttributeIPHost` vertex on the database would be simpler.

I'm also assuming that there are cases where users want to have an IP host with a `/32` or `/128`, but I'm not really sure about that. if this is a case that users would never want, then we should be able to handle in Python without any database changes

ajtmccarty:
replaced by the PR chain starting with #10081

PR #9977, opened 2026-07-20
fix(backend): resolve NodeNotFoundError node_type to a string

### LINKED ISSUE #9926: bug: non-str node_type in NodeNotFoundError crashes GraphQL error formatter

saltas888:
# Why

A GraphQL request that produced a `NodeNotFoundError` whose `node_type` was a schema/protocol **class** (not a kind string) crashed the error formatter itself — pydantic rejected the non-`str` `node_kind`, the exception escaped the formatter, and the request returned an uncaught **HTTP 500** instead of a normal `NODE_NOT_FOUND` (404). Observed in production as thousands of uncaught 500s.

Closes #9926

Supersedes #9952, which took a different approach: widening `NodeNotFoundError.node_type` to accept a class plus a defensive `str()` guard in the formatter. This PR instead fixes the two raise sites at the source and keeps the exception's `str` contract.

## What changed

**Behavioral:** a node-not-found lookup now always returns a catalogued `NODE_NOT_FOUND` (404) GraphQL error, whether the internal `kind` was a string or a class.

**Implementation:**

- `NodeManager.get_one` and `get_one_by_id_or_default_filter` resolve `kind` to its kind string at the raise site, so `NodeNotFoundError.node_type` is always a plain string (the sibling lookup methods already did this).
- Re-enabled mypy `arg-type` on `infrahub.core.manager` so a non-`str` `node_type` is caught at type-check time; fixed the pre-existing violations it surfaced (or scoped the two genuine type-system limitations with a documented `# type: ignore`).
- Documented the practice ("a wrong-type bug means a checker was suppressed — re-enable it and fix the typing chain") in `dev/guidelines/backend/python.md`, and tightened comment guidance in `.agents/rules/code-doc-style.md`.

**Unchanged:** `NodeNotFoundError.node_type` keeps its `str` contract (not widened to accept a class); the error formatter is untouched — the fix is at the source, not a defensive coercion.

## How to review

- Start with `backend/infrahub/core/manager.py` — the two raise sites.
- `pyproject.toml` — one line: `arg-type` re-enabled for `infrahub.core.manager`.
- Tests: `test_manager_node.py` (3 raise-site cases) and `test_20_graphql.py` (end-to-end 404).

## How to test

```bash
uv run pytest backend/tests/component/core/test_manager_node.py -k reports_str_node_type
uv run pytest backend/tests/component/api/test_20_graphql.py -k menu_delete_missing
uv run invoke backend.lint   # arg-type now enforced on manager
```

## Impact & rollout

Backend-only. No schema, API-contract, or migration changes.

A remaining hardening follow-up — two other modules (`infrahub.core.node`, `infrahub.graphql.mutations.main`) raise `NodeNotFoundError` with `arg-type` still disabled, both passing `str` today — is tracked in INBOX-1.

---
## Summary by cubic
Fixes GraphQL requests that returned HTTP 500 when a missing node used a class kind. Missing nodes now return `NODE_NOT_FOUND` (404), and stricter typing prevents regressions; also fixes a `fetch_peers` KeyError by normalizing peer IDs.

- **Bug Fixes**
  - Resolve `kind` to a string at raise sites in `NodeManager.get_one` and `get_one_by_id_or_default_filter`; ensure lookups pass kind strings (e.g., `get_one_by_hfid`).
  - Normalize peer IDs to `str` on both sides of `fetch_peers` mapping to prevent `KeyError`.
  - Re-enabled `mypy` `arg-type` for `infrahub.core.manager`; added targeted `# type: ignore[...]`; `NodeNotFoundError.node_type` stays `str`.
  - Tests: GraphQL `CoreMenuItemDelete` with a missing ID returns `NODE_NOT_FOUND` (404); manager errors assert string `node_type`. Updated backend guidelines and `.agents/rules/code-doc-style.md`; added changelog entry.

- **Refactors**
  - Trimmed the `get_one` kind-normalization comment to keep only the non-obvious why.

<sup>Written for commit f115a2bfcbee3fbd17964682b12f0c83e27efbff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9977?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/rules/code-doc-style.md
+++ b/.agents/rules/code-doc-style.md
@@ -43,6 +43,6 @@ Where IDs *do* belong:
 
 ## What good documentation looks like
 
-- Explains *why* the code exists when the why is non-obvious (a constraint, an invariant, a workaround for a specific upstream bug).
+- Comment the *why*, never the *what*: a constraint, an invariant, a workaround, a deliberate deviation from the obvious approach. Never paraphrase the line below it or restate the type signature.
 - Documents the contract of a public function (inputs, outputs, errors raised) when it crosses a module boundary.
-- Stays silent by default. A comment that restates the code is worse than no comment — it adds noise and rots the moment the code changes.
+- Stays silent by default. If code needs a comment to explain *what* it does, rename or extract until it doesn't. A comment that restates the code is worse than none — noise that rots the moment the code changes.
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -344,7 +344,7 @@ class NodeManager:
                 data=peer,
             )
             if fetch_peers:
-                result.set_peer(value=peer_nodes[peer.peer_id])
+                result.set_peer(value=peer_nodes[str(peer.peer_id)])
             results.append(result)
 
         return results
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -163,7 +163,7 @@ class NodeManager:
             node = await cls.get_one_by_hfid(
                 db=db,
                 hfid=filters["hfid"],
-                kind=schema,
+                kind=node_schema.kind,
                 fields=fields,
                 at=at,
                 branch=branch,
@@ -318,7 +318,7 @@ class NodeManager:
             return []
 
         if fetch_peers:
-            peer_ids = [peer.peer_id for peer in peers_info]
+            peer_ids = [str(peer.peer_id) for peer in peers_info]
             peer_nodes = await cls.get_many(
                 db=db,
                 ids=peer_ids,
@@ -336,7 +336,7 @@ class NodeManager:
                 branch=branch,
                 source_kind=peer.source_kind,
                 at=at,
-                node_id=peer.source_id,
+                node_id=str(peer.source_id),
             ).load(
                 db=db,
                 id=peer.rel_node_id,
@@ -864,10 +864,11 @@ class NodeManager:
         if node:
             return node
 
-        node = await cls.get_one_by_default_filter(
+        return await cls.get_one_by_default_filter(
             db=db,
             id=id,
             kind=kind,
+            raise_on_error=True,
             fields=fields,
             at=at,
             branch=branch,
@@ -875,9 +876,6 @@ class NodeManager:
             prefetch_relationships=prefetch_relationships,
             branch_agnostic=branch_agnostic,
         )
-        if not node:
-            raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
-        return node
 
     @overload
     @classmethod
@@ -1059,7 +1057,11 @@ class NodeManager:
 
         if not result:
             if raise_on_error:
-                raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+                # `kind` is a protocol class, a kind string, or None because callers look nodes up
+                # in all three ways; derive the name from it directly, since resolving it through the
+                # schema could raise on the not-found path we are already handling.
+                node_kind = kind.__name__ if isinstance(kind, type) else (kind or "Node")
+                raise NodeNotFoundError(branch_name=branch.name, node_type=node_kind, identifier=id)
             return None
 
         node = result[id]
@@ -1175,7 +1177,7 @@ class NodeManager:
             node_class = identify_node_class(node=node)
             node_branch = await registry.get_branch(db=db, branch=node.branch)
             item = await node_class.init(schema=node.schema, branch=node_branch, at=at, db=db)
-            await item.load(**new_node_data, db=db)
+            await item.load(**new_node_data, db=db)  # type: ignore[arg-type]  # mypy cannot match a heterogeneous **dict splat against load()'s typed params
             item._set_created_at(node.created_at)
             item._set_created_by(node.created_by)
             item._set_updated_at(node.updated_at)
@@ -1303,9 +1305,9 @@ class NodeManager:
             if insert_peer_node:
                 rel_peers: list[Node | str] = []
                 for peer_id in peer_ids:
-                    peer = nodes_by_id.get(peer_id)
-                    if peer:
-                        rel_peers.append(peer)
+                    peer_node = nodes_by_id.get(peer_id)
+                    if peer_node:
+                        rel_peers.append(peer_node)
             # if only getting some relationships, make sure we want THIS relationship for THIS node schema
             elif cardinality_one_identifiers_by_kind:
                 required_direction = cardinality_one_identifiers_by_kind.get(node_schema.kind, {}).get(
@@ -1342,7 +1344,8 @@ class NodeManager:
                 rel_peers_with_metadata.append(peer_with_metadata)
 
             rel_manager.has_fetched_relationships = True
-            await rel_manager.update(db=db, data=rel_peers_with_metadata)
+            # invariant list parameter; update() only reads data, so the narrower element type is safe
+            await rel_manager.update(db=db, data=rel_peers_with_metadata)  # type: ignore[arg-type]
 
     @classmethod
     async def delete(
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -1057,9 +1057,7 @@ class NodeManager:
 
         if not result:
             if raise_on_error:
-                # `kind` is a protocol class, a kind string, or None because callers look nodes up
-                # in all three ways; derive the name from it directly, since resolving it through the
-                # schema could raise on the not-found path we are already handling.
+                # Avoid resolving kind via the schema here: it can raise on the not-found path.
                 node_kind = kind.__name__ if isinstance(kind, type) else (kind or "Node")
                 raise NodeNotFoundError(branch_name=branch.name, node_type=node_kind, identifier=id)
             return None
--- a/backend/tests/component/core/test_manager_node.py
+++ b/backend/tests/component/core/test_manager_node.py
@@ -9,6 +9,7 @@ from infrahub.core.constants import MetadataOptions
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager, identify_node_class
 from infrahub.core.node import Node
+from infrahub.core.protocols import CoreMenuItem
 from infrahub.core.protocols_base import CoreNode
 from infrahub.core.query.node import NodeToProcess
 from infrahub.core.registry import registry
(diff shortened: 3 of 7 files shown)


backend/infrahub/core/manager.py
```diff
@@ -318,7 +318,7 @@ async def query_peers(
             return []
 
         if fetch_peers:
-            peer_ids = [peer.peer_id for peer in peers_info]
+            peer_ids = [str(peer.peer_id) for peer in peers_info]
```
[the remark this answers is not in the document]
saltas888: Good catch — fixed in 0bdb4f074. My `str(peer.peer_id)` change made `peer_nodes` string-keyed but left the lookup using the raw `peer.peer_id`, so a non-str peer id would `KeyError`. Normalized the lookup to `peer_nodes[str(peer.peer_id)]` so both sides match (this also clears the previously-suppressed `index` type error on that line).

backend/infrahub/core/manager.py
```diff
@@ -1059,7 +1057,11 @@ async def get_one(
 
         if not result:
             if raise_on_error:
-                raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+                # `kind` is a protocol class, a kind string, or None because callers look nodes up
+                # in all three ways; derive the name from it directly, since resolving it through the
+                # schema could raise on the not-found path we are already handling.
+                node_kind = kind.__name__ if isinstance(kind, type) else (kind or "Node")
+                raise NodeNotFoundError(branch_name=branch.name, node_type=node_kind, identifier=id)
```
saltas888, line 1062: @polmichel I know we discussed this, but I think this is very specific to `get_one` method and the fact that it expects 3 different types.

.agents/rules/code-doc-style.md
polmichel, line 1: Great change

backend/infrahub/core/manager.py
```diff
@@ -1059,7 +1057,11 @@ async def get_one(
 
         if not result:
             if raise_on_error:
-                raise NodeNotFoundError(branch_name=branch.name, node_type=kind, identifier=id)
+                # `kind` is a protocol class, a kind string, or None because callers look nodes up
+                # in all three ways; derive the name from it directly, since resolving it through the
+                # schema could raise on the not-found path we are already handling.
```
polmichel: The comment looks too long for what it explains. 
Maybe removing it completely would be better, another option could be to remove the first part:
```
 `kind` is a protocol class, a kind string, or None because callers look nodes up in all three ways; 
```

backend/infrahub/core/manager.py
```diff
@@ -1342,7 +1344,8 @@ async def _enrich_one_node_with_relationships(
                 rel_peers_with_metadata.append(peer_with_metadata)
 
             rel_manager.has_fetched_relationships = True
-            await rel_manager.update(db=db, data=rel_peers_with_metadata)
+            # invariant list parameter; update() only reads data, so the narrower element type is safe
+            await rel_manager.update(db=db, data=rel_peers_with_metadata)  # type: ignore[arg-type]
```
polmichel, line 1346: Optional as out of the scope of the PR:

While taking a look at this ignore comment, Claude is telling me that a solution could be to type the parameter inside the `update` method as a `Sequence`.

Since we are removing the warning, it's always better to properly fix the warning. https://docs.python.org/3/library/stdtypes.html#common-sequence-operations Sequence is not mutable by default (see ABC class https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence)

Quote from the Python documentation for `MutableSequence`
_Inherited [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) methods and append, clear, reverse, extend, pop, remove, and __iadd___

backend/tests/component/core/test_manager_node.py
polmichel, line 1: Nice, this is covering the different code paths

polmichel (approved):
LGTM! 
Minor comments about code comments which are too long in my opinion, and a typing warning comment

backend/infrahub/core/manager.py
polmichel: This is fine!

backend/infrahub/core/manager.py
saltas888: Trimmed in f115a2bfc — dropped the type-list first clause and kept only the non-obvious why:
```python
# Avoid resolving kind via the schema here: it can raise on the not-found path.
```
Thanks!

backend/infrahub/core/manager.py
```diff
@@ -1303,9 +1305,9 @@ async def _enrich_one_node_with_relationships(
             if insert_peer_node:
                 rel_peers: list[Node | str] = []
                 for peer_id in peer_ids:
-                    peer = nodes_by_id.get(peer_id)
-                    if peer:
-                        rel_peers.append(peer)
+                    peer_node = nodes_by_id.get(peer_id)
+                    if peer_node:
+                        rel_peers.append(peer_node)
```
ajtmccarty, line 1310: much better name

ajtmccarty (approved):

backend/infrahub/core/manager.py
saltas888: I will create a new ticket to tackle this separately! Good point!

PR #9994, opened 2026-07-22
queries for node-scoped uniqueness constraints

ajtmccarty:
# Why

When uniqueness validation fires from a data change (proposed-change checks, merges, rebases), the
`UniquenessChecker` scans the **entire population** of each implicated kind via
`NodeUniqueAttributeConstraintQuery` — cost tracks how many objects exist, not how many changed.
The scan also returns every population-wide duplicate (>100k rows on that dataset) for Python-side
post-processing, including violations that pre-date the change.

**Goal:** provide the query and component foundation to validate only the *changed* nodes: a
batched, index-anchored query whose cost tracks the size of the change, plus the pieces that
identify which nodes a data diff implicates.

**Non-goals:** this PR does not yet *wire* any of it. `UniquenessChecker` still runs the full scan;
the determiner does not yet emit changed-node uuids. The checker integration, the diff-side scoping,
and generic/implementation dedup are stacked follow-ups. Nothing in this PR changes runtime
behavior.

Part of IFC-2796. Later PR(s) will actually wire the changes into the tasks and test them.

## What changed

**Behavioral changes:** none — every addition is unused by production code paths in this PR.

**New pieces (grouped by role in the upcoming targeted path):**

- **`TargetedUniquenessValidationQuery`** (`validators/uniqueness/query/targeted_validation.py`) —
  one Cypher query per (kind, constraint group) validating ALL changed nodes against the full
  population in a single round trip. It resolves the first constraint element's value for the
  changed nodes only, probes the population for other nodes sharing it through the
  `AttributeValueIndexed`/`Node(uuid)` indexes, drops changed nodes with no match, and resolves each
  subsequent element only for the survivors — a changed node whose first element collides with
  nothing never has its remaining values read.
- **`AffectedUniquenessDependentsQuery` + `UniquenessDependentResolver`**
  (`query/affected_dependents.py`, `dependent_resolver.py`) — cross-kind resolution: when a
  constraint reads a peer's attribute (e.g. `owner__name`), a change to the *peer* kind implicates
  nodes of the constrained kind; this resolves the changed peers into those nodes. Each hop resolves
  to a single winner across the input/default/global branches (latest edge on the deepest branch),
  so a relationship deleted on the input branch excludes the node — matching what the merge will
  produce — while the default branch's post-fork state decides wherever the input branch is silent.
- **`NodeDiffIndex`** (`validators/node_diff_index.py`) — per-kind index over a data diff's changed
  attributes, relationships, and node uuids; `NodeDiffFieldSummary` gains the `node_uuids` field it
  reads.
- **Query package restructure:** `validators/uniqueness/query.py` moves to
  `validators/uniqueness/query/validation.py` inside a new package whose `__init__` re-exports
  everything — existing import paths are unchanged.

**Design decisions worth knowing:**

- Peer-attribute constraint elements (`owner__name`) are **rejected** by the targeted query
  (`ValueError`): schema updates disallow attributes-of-relationships in `uniqueness_constraints`
  today, so the targeted path never legitimately receives one.

**What stayed the same:** `NodeUniqueAttributeConstraintQuery` and `UniquenessValidationQuery` are
untouched (the latter still serves the node-save constraint in
`core/node/constraints/grouped_uniqueness.py`); `UniquenessChecker` behavior is unchanged; no DB
schema or migration changes.

# Performance

Regarding the new targeted uniqueness query, on a large local dataset (≈500k `DcimInterface` nodes),
a single uniqueness-constraints scan ran for 8+ minutes and then OOM-killed Neo4j. A sibling kind's
scan ran for 30 minutes before being killed.

Benchmarked against a large generated dataset (2%/5%-of-population change branches): the targeted
query validates 9,400 changed nodes of a 503k-node kind in ~12s and 23,500 in ~29s, where the
current full scan of the same kind ran 8.4 minutes before OOM-killing Neo4j (and >30 minutes on a
464k-node sibling kind before being killed). All 31 implicated kinds complete in ~32s / ~60s total.

## Impact & rollout

- **Backward compatibility:** no behavior change — additions only, plus an import-transparent file
  move. API contract unchanged.
- **Performance:** no actual changes in this PR.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry (deferred to the follow-up PR that wires the behavior change)
- [ ] External docs updated (n/a — no user-facing change yet)
- [ ] Internal .md docs updated (n/a)

---
## Summary by cubic
Adds index-backed, node-scoped uniqueness validation that checks only changed nodes and their dependents instead of scanning all nodes. Supports IFC-2796 and reduces validation time on large kinds; honors branch visibility, post-fork default updates, and migrated-kind duplicates.

- **New Features**
  - `TargetedUniquenessValidationQuery` validates only given node UUIDs for one constraint, comparing attributes and 1:1 relationships while enforcing node liveness and branch visibility; supports large attribute types (no index), generic kinds across implementations, and ignores stale same-UUID duplicates from kind migrations.
  - Added `UniquenessDependentResolver` (uses `AffectedUniquenessDependentsQuery`) to resolve constrained nodes referencing changed peers; added `NodeDiffIndex` and extended `NodeDiffFieldSummary` with `node_uuids` to drive scoping.

- **Refactors**
  - Moved uniqueness queries under `core/validators/uniqueness/query` and re-exported to keep imports stable.
  - Fixed a value check so falsy attribute values compare correctly.

<sup>Written for commit 38e6b1c194ca92d1762903f4967690133f53a789. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/9994?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/validators/node_diff_index.py
+++ b/backend/infrahub/core/validators/node_diff_index.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from infrahub.core.diff.model.path import NodeDiffFieldSummary
+
+
+class NodeDiffIndex:
+    """Which fields and nodes changed, per kind, in a data diff.
+
+    `initialize` must be called for a given diff before any query method is used.
+    """
+
+    def __init__(self) -> None:
+        self._initialized: bool = False
+        self._kinds: set[str] = set()
+        self._attribute_names_by_kind: dict[str, set[str]] = {}
+        self._relationship_names_by_kind: dict[str, set[str]] = {}
+        self._node_uuids_by_kind: dict[str, set[str]] = {}
+
+    def initialize(self, node_diffs: list[NodeDiffFieldSummary]) -> None:
+        self._kinds = set()
+        self._attribute_names_by_kind = {}
+        self._relationship_names_by_kind = {}
+        self._node_uuids_by_kind = {}
+        for node_diff in node_diffs:
+            self._kinds.add(node_diff.kind)
+            self._attribute_names_by_kind.setdefault(node_diff.kind, set()).update(node_diff.attribute_names)
+            self._relationship_names_by_kind.setdefault(node_diff.kind, set()).update(node_diff.relationship_names)
+            self._node_uuids_by_kind.setdefault(node_diff.kind, set()).update(node_diff.node_uuids)
+        self._initialized = True
+
+    def _ensure_initialized(self) -> None:
+        if not self._initialized:
+            raise RuntimeError("NodeDiffIndex must be initialized with initialize() before its query methods are used")
+
+    @property
+    def kinds(self) -> set[str]:
+        self._ensure_initialized()
+        return self._kinds
+
+    def has_attribute_diff(self, kind: str, name: str) -> bool:
+        self._ensure_initialized()
+        return name in self._attribute_names_by_kind.get(kind, set())
+
+    def has_relationship_diff(self, kind: str, name: str) -> bool:
+        self._ensure_initialized()
+        return name in self._relationship_names_by_kind.get(kind, set())
+
+    def uuids_for_kinds(self, kinds: set[str]) -> set[str]:
+        self._ensure_initialized()
+        uuids: set[str] = set()
+        for kind in kinds:
+            uuids |= self._node_uuids_by_kind.get(kind, set())
+        return uuids
--- a/backend/infrahub/core/validators/uniqueness/query/__init__.py
+++ b/backend/infrahub/core/validators/uniqueness/query/__init__.py
@@ -1,0 +1,14 @@
+from ..model import QueryAttributePathValued, QueryRelationshipPathValued
+from .affected_dependents import AffectedUniquenessDependentsQuery
+from .targeted_validation import TargetedUniquenessValidationQuery, TargetedUniquenessViolation
+from .validation import NodeUniqueAttributeConstraintQuery, UniquenessValidationQuery
+
+__all__ = [
+    "AffectedUniquenessDependentsQuery",
+    "NodeUniqueAttributeConstraintQuery",
+    "QueryAttributePathValued",
+    "QueryRelationshipPathValued",
+    "TargetedUniquenessValidationQuery",
+    "TargetedUniquenessViolation",
+    "UniquenessValidationQuery",
+]
--- a/backend/infrahub/core/validators/uniqueness/query/targeted_validation.py
+++ b/backend/infrahub/core/validators/uniqueness/query/targeted_validation.py
@@ -1,0 +1,379 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from infrahub.core.graph.schema import GraphAttributeValueIndexedNode, GraphAttributeValueNode
+from infrahub.core.query import Query, QueryType
+from infrahub.types import is_large_attribute_type
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from infrahub.core.schema.basenode_schema import SchemaAttributePath
+    from infrahub.database import InfrahubDatabase
+
+
+@dataclass(frozen=True)
+class TargetedUniquenessViolation:
+    """One changed node sharing its full constraint-value tuple with at least one other node."""
+
+    changed_uuid: str
+    element_values: tuple[Any, ...]
+    partner_uuids: tuple[str, ...]
+
+
+class TargetedUniquenessValidationQuery(Query):
+    """Find uniqueness violations for a set of changed nodes.
+
+    For one uniqueness constraint group, resolve the changed nodes' current value for a first
+    element only, probe the whole population for other nodes sharing that value, and drop every
+    changed node with no match. Each subsequent element is resolved only for the changed nodes
+    that still have live matches, and each candidate set shrinks by comparing the candidates'
+    current value for that element. A changed node is reported only if at least one other node
+    still shares its full value tuple after the last element.
+
+    Values are compared as stored in the graph: enum values in their raw form and null attribute
+    values as the null sentinel, so two nulls collide. A node without a live value for an element
+    (e.g. a relationship with no peer) contributes no tuple and cannot collide on that group.
+
+    Changed nodes and candidate partners must themselves be live: their latest IS_PART_OF edge on
+    a visible branch must be active. This excludes nodes deleted on the branch and stale same-uuid
+    duplicates left behind by kind or inheritance migrations.
+
+    Supported constraint elements are node attributes (compared by value) and cardinality-one
+    relationships (compared by peer uuid). Attributes of related peers are not supported.
+    """
+
+    name = "uniqueness_constraint_validation_targeted"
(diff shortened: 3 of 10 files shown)


backend/tests/unit/core/validators/test_node_diff_index.py
```diff
@@ -0,0 +1,75 @@
...
+        index.has_attribute_diff(kind="TestCar", name="name")
+    with pytest.raises(RuntimeError, match=match):
+        index.has_relationship_diff(kind="TestCar", name="owner")
+    with pytest.raises(RuntimeError, match=match):
+        index.uuids_for_kinds({"TestCar"})
+
+
+def test_indexes_fields_and_uuids_per_kind() -> None:
+    index = NodeDiffIndex()
+    index.initialize(
```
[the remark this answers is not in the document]
ajtmccarty: updated in latest commit

backend/infrahub/core/validators/node_diff_index.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from infrahub.core.diff.model.path import NodeDiffFieldSummary
+
+
+class NodeDiffIndex:
```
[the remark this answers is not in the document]
ajtmccarty: will be linked up in following PR(s)

backend/infrahub/core/validators/uniqueness/query/targeted_validation.py
```diff
@@ -0,0 +1,379 @@
...
+        self,
+        kind: str,
+        constraint_elements: list[SchemaAttributePath],
+        node_uuids: list[str],
+        **kwargs: Any,
+    ) -> None:
+        if not constraint_elements:
+            raise ValueError("At least one constraint element is required")
+        for element in constraint_elements:
+            if element.relationship_schema is not None and element.attribute_schema is not None:
```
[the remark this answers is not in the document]
ajtmccarty: we don't actually allow relationship-attributes in uniqueness constraints when updating a schema. some of the backend code supports them, but it is not actually used for now

backend/infrahub/core/validators/uniqueness/query/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+
+__all__ = [
+    "AffectedUniquenessDependentsQuery",
+    "NodeUniqueAttributeConstraintQuery",
+    "QueryAttributePathValued",
+    "QueryRelationshipPathValued",
+    "TargetedUniquenessValidationQuery",
+    "TargetedUniquenessViolation",
+    "UniquenessValidationQuery",
+]
```
polmichel, line 14: Is this import redirection intentional? 
I guess that's fine since this is a query package there is no risk of creating an import cycle, and it allows callers to access all the queries directly, but I'd rather ask than assume.

backend/infrahub/core/validators/uniqueness/query/targeted_validation.py
```diff
@@ -0,0 +1,379 @@
...
+        """Pick the element whose population probe anchors the query.
+
+        The anchor is the only population-wide MATCH, so prefer an element whose value can be
+        found through an index: any relationship element (peer uuid) or any attribute whose kind
+        has an indexed value label. Fall back to the first element when none qualifies.
+        """
+        for index, element in enumerate(self.constraint_elements):
+            if not self._is_large_type(element):
+                return index
+        return 0
```
polmichel, line 119: That's interesting! Have you considered these options from a performance standpoint:

- Having multiple indexed anchors to reduce the number of matched nodes after the first iteration?
- Choosing an element that is indexed AND likely to return a limited number of nodes? For instance, a boolean would have a high probability of retrieving a lot of nodes with the searched value, whereas I would assume an integer would not.

backend/tests/component/core/constraint_validators/test_targeted_uniqueness_query.py
polmichel, line 1: Nice test suite

polmichel: Would it be interesting to add a test for `render_liveness_check` role? 

```
The latest IS_PART_OF edge on a visible branch decides: a node deleted on the branch, or a
stale same-uuid duplicate left by a kind or inheritance migration, resolves to a non-active
winner and the row is eliminated.
```

backend/infrahub/core/validators/node_diff_index.py
```diff
@@ -0,0 +1,56 @@
...
+        self._kinds = set()
+        self._attribute_names_by_kind = {}
+        self._relationship_names_by_kind = {}
+        self._node_uuids_by_kind = {}
+        for node_diff in node_diffs:
+            self._kinds.add(node_diff.kind)
+            self._attribute_names_by_kind.setdefault(node_diff.kind, set()).update(node_diff.attribute_names)
+            self._relationship_names_by_kind.setdefault(node_diff.kind, set()).update(node_diff.relationship_names)
+            self._node_uuids_by_kind.setdefault(node_diff.kind, set()).update(node_diff.node_uuids)
+        self._initialized = True
```
polmichel, line 32: Can we not just make this object more immutable while doing the initialization in the constructor part? I feel I am missing something here

polmichel (approved):
LGTM, optional comments posted

backend/infrahub/core/validators/uniqueness/query/__init__.py
ajtmccarty: it's for backwards compatibility for the existing queries that were in `query.py` previously

backend/infrahub/core/validators/uniqueness/query/targeted_validation.py
ajtmccarty: I think that using multiple anchors would probably reduce performance b/c there would be more path matching to do at the beginning instead of reducing the number of possible violating objects as quickly as possible. the ordering of uniqueness constraints is important and we suggest schema builders put the constraints in order of least-likely-to-overlap to most-likely-to-overlap so that this staged reduction will perform better. I think that trying to analyze uniqueness constraints to determine the best ordering is probably something that would fit better as part of a schema optimization tool of some sort. maybe it could fit in this flow somewhere, but I think it would be a separate issue.

backend/infrahub/core/validators/node_diff_index.py
ajtmccarty: unfortunately, we can't do the `intialize()` logic in the constructor b/c we don't have access to the `NodeDiffFieldSummary`s until runtime, so we have to wait until the flow actually starts to run the indexing

backend/tests/component/core/constraint_validators/test_targeted_uniqueness_query.py
ajtmccarty: yes good idea. adding more Node kind/inheritance update testing is always a good idea. it has caused quite a few bugs in the past

PR #10000, opened 2026-07-22
fix(backend): delete an account's Preference row on account deletion (IFC-2867)

pa-lem:
## Summary

Follow-up from the user/global preferences work (IFC-2718): a `Preference` is a `StandardNode` keyed by a plain-string `owner_id`, so the schema cannot cascade it with `on_delete: cascade` when its account is deleted — the row was left behind as unreachable dead data.

This hooks the account-delete mutation to drop the row explicitly (ticket option 1, the cleanup hook). Accounts are branch-agnostic, so the deletion is effective everywhere at once and the hook needs no branch condition.

## Changes

- `PreferenceRepository.delete_for_owner(owner_id)` — deletes every Preference row for an owner; no-op when there is none.
- `InfrahubAccountMutation.mutate_delete` — after the account delete succeeds, drops the account's Preference row. Best-effort: a cleanup failure logs a warning instead of failing the mutation (the account is already gone, and the orphan stays benign since account ids are UUIDs and never reused).
- Updated the `Preference` docstring, which pointed at IFC-2867 as the tracking ticket.
- Changelog fragment.

## Tests

- Component (repository): `delete_for_owner` removes only the target owner's row (other owners + global row untouched); no-op when no row exists.
- Component (GraphQL): `CoreAccountDelete` removes the deleted account's Preference row and leaves the global row untouched.
- Ran locally: ruff, mypy, preference unit tests, and both component test files (14 passed).

Jira: [IFC-2867](https://opsmill.atlassian.net/browse/IFC-2867)

[IFC-2867]: https://opsmill.atlassian.net/browse/IFC-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Deletes an account’s `Preference` row when the account is deleted to prevent orphaned data. Global preferences stay untouched. (IFC-2867)

- **Bug Fixes**
  - Added `PreferenceRepository.delete_for_owner(owner_id)` using a single-query `PreferenceDeleteByOwnerQuery` (`DETACH DELETE` by `owner_id`).
  - Updated `CoreAccountDelete` to run cleanup under the per-owner preference lock; on failure catch `DatabaseError`, `QueryTimeoutError`, `DriverError`, `Neo4jError`, and `RedisError`, log, and do not fail. Trimmed comments, removed ticket IDs, and dropped the unreleased changelog fragment.

<sup>Written for commit 63cbee95976b742011753aaf23d2d9dc1952d59f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/preferences/models.py
+++ b/backend/infrahub/core/preferences/models.py
@@ -17,10 +17,10 @@ class Preference(StandardNode):
     client's built-in default).
 
     `owner_id` is a plain string, not a graph relationship: a StandardNode cannot declare a schema
-    relationship with `on_delete: cascade` (that is a schema-Node feature), so the schema cannot
-    cascade a Preference row when its account is deleted. Instead the account-delete mutation drops
-    the row explicitly (IFC-2867). A row orphaned by a deletion that bypasses the mutation stays
-    benign: account ids are UUIDs and never reused, so it is permanently unreachable.
+    relationship with `on_delete: cascade` (that is a schema-Node feature), so account deletion
+    cannot cascade to the Preference row through the schema and deletes it explicitly instead. A row
+    orphaned by a deletion path that skips that cleanup stays benign: account ids are UUIDs and
+    never reused, so it is permanently unreachable.
     """
 
     owner_id: str
--- a/backend/infrahub/core/preferences/models.py
+++ b/backend/infrahub/core/preferences/models.py
@@ -17,10 +17,10 @@ class Preference(StandardNode):
     client's built-in default).
 
     `owner_id` is a plain string, not a graph relationship: a StandardNode cannot declare a schema
-    relationship with `on_delete: cascade` (that is a schema-Node feature), so deleting an account
-    leaves its Preference row behind as unreachable dead data. Account ids are UUIDs and never reused,
-    so such a row is permanently unreachable and benign. Cleanup is out of scope for V1 and tracked in
-    Jira (IFC-2867).
+    relationship with `on_delete: cascade` (that is a schema-Node feature), so the schema cannot
+    cascade a Preference row when its account is deleted. Instead the account-delete mutation drops
+    the row explicitly (IFC-2867). A row orphaned by a deletion that bypasses the mutation stays
+    benign: account ids are UUIDs and never reused, so it is permanently unreachable.
     """
 
     owner_id: str
--- a/backend/infrahub/core/preferences/repository.py
+++ b/backend/infrahub/core/preferences/repository.py
@@ -52,3 +52,11 @@ class PreferenceRepository:
     async def save(self, preference: Preference, actor_id: str = SYSTEM_USER_ID) -> None:
         """Persist the preference, creating the row on first write or updating it in place."""
         await preference.save(db=self.db, user_id=actor_id)
+
+    async def delete_for_owner(self, owner_id: str) -> None:
+        """Delete every Preference row owned by `owner_id`; a no-op when there is none."""
+        query = await PreferenceGetByOwnerQuery.init(db=self.db, owner_ids={owner_id})
+        await query.execute(db=self.db)
+
+        for preference in query.get_preferences():
+            await preference.delete(db=self.db)
--- a/backend/infrahub/core/preferences/repository.py
+++ b/backend/infrahub/core/preferences/repository.py
@@ -3,7 +3,11 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from infrahub.core.constants import SYSTEM_USER_ID
-from infrahub.core.query.preference import PreferenceGetAllQuery, PreferenceGetByOwnerQuery
+from infrahub.core.query.preference import (
+    PreferenceDeleteByOwnerQuery,
+    PreferenceGetAllQuery,
+    PreferenceGetByOwnerQuery,
+)
 
 if TYPE_CHECKING:
     from infrahub.core.preferences.models import Preference
@@ -55,8 +59,5 @@ class PreferenceRepository:
 
     async def delete_for_owner(self, owner_id: str) -> None:
         """Delete every Preference row owned by `owner_id`; a no-op when there is none."""
-        query = await PreferenceGetByOwnerQuery.init(db=self.db, owner_ids={owner_id})
+        query = await PreferenceDeleteByOwnerQuery.init(db=self.db, owner_id=owner_id)
         await query.execute(db=self.db)
-
-        for preference in query.get_preferences():
-            await preference.delete(db=self.db)
--- a/backend/infrahub/core/query/preference.py
+++ b/backend/infrahub/core/query/preference.py
@@ -83,7 +83,6 @@ class PreferenceGetByOwnerQuery(PreferenceReadQuery):
         # Cypher parameters cannot bind a set, so pass the ids as a list.
         self.params["owner_ids"] = list(self.owner_ids)
 
-        # The label is the Preference StandardNode type name (Cypher labels can't be parameterised).
         query = """
         MATCH (n:Preference)
         WHERE n.owner_id IN $owner_ids
@@ -106,7 +105,6 @@ class PreferenceDeleteByOwnerQuery(StandardNodeQuery):
     async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
         self.params["owner_id"] = self.owner_id
 
-        # The label is the Preference StandardNode type name (Cypher labels can't be parameterised).
         query = """
         MATCH (n:Preference { owner_id: $owner_id })
         DETACH DELETE n
@@ -120,7 +118,6 @@ class PreferenceGetAllQuery(PreferenceReadQuery):
     name = "preference_get_all"
 
     async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
-        # The label is the Preference StandardNode type name (Cypher labels can't be parameterised).
         query = """
         MATCH (n:Preference)
         """
--- a/backend/infrahub/core/query/preference.py
+++ b/backend/infrahub/core/query/preference.py
@@ -92,6 +92,28 @@ class PreferenceGetByOwnerQuery(PreferenceReadQuery):
         self._set_return_shape()
 
 
+class PreferenceDeleteByOwnerQuery(StandardNodeQuery):
+    """Delete every Preference row owned by one owner, in ONE query (no fetch-then-delete-each)."""
+
+    name = "preference_delete_by_owner"
+    type = QueryType.WRITE
+    insert_return = False
+
+    def __init__(self, owner_id: str, **kwargs: Any) -> None:
+        self.owner_id = owner_id
+        super().__init__(**kwargs)
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        self.params["owner_id"] = self.owner_id
+
+        # The label is the Preference StandardNode type name (Cypher labels can't be parameterised).
+        query = """
+        MATCH (n:Preference { owner_id: $owner_id })
+        DETACH DELETE n
+        """
+        self.add_to_query(query=query)
+
+
 class PreferenceGetAllQuery(PreferenceReadQuery):
     """Fetch every Preference row, across all owners."""
 
--- a/backend/infrahub/graphql/mutations/account.py
+++ b/backend/infrahub/graphql/mutations/account.py
@@ -11,6 +11,7 @@ from infrahub.core.constants import InfrahubKind
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
 from infrahub.core.order import OrderModel
+from infrahub.core.preferences.repository import PreferenceRepository
 from infrahub.core.protocols import CoreAccount, CoreNode, InternalAccountToken
 from infrahub.core.schema import NodeSchema
 from infrahub.core.timestamp import Timestamp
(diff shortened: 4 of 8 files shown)


backend/infrahub/core/preferences/repository.py
```diff
@@ -52,3 +52,11 @@ async def get_all(self) -> list[Preference]:
...
         """Persist the preference, creating the row on first write or updating it in place."""
         await preference.save(db=self.db, user_id=actor_id)
+
+    async def delete_for_owner(self, owner_id: str) -> None:
+        """Delete every Preference row owned by `owner_id`; a no-op when there is none."""
+        query = await PreferenceGetByOwnerQuery.init(db=self.db, owner_ids={owner_id})
+        await query.execute(db=self.db)
+
+        for preference in query.get_preferences():
+            await preference.delete(db=self.db)
```
saltas888: this is going to execute N+1 query for every single preference.

Can't we write a delete query with the condition (owner_id)? See the doc here https://neo4j.com/docs/cypher-manual/current/clauses/delete/

backend/infrahub/graphql/mutations/account.py
```diff
@@ -220,4 +224,16 @@ async def mutate_delete(
...
+        result = await super().mutate_delete(info=info, data=data, branch=branch)
+
+        # Accounts are branch-agnostic, so the deletion is effective everywhere at once; drop the
+        # account's Preference row (a StandardNode keyed by owner_id, which cannot cascade via the
+        # schema) instead of leaving it behind as unreachable dead data (IFC-2867). Best-effort: the
+        # account is already deleted, so a cleanup failure must not fail the mutation — the orphaned
+        # row is benign (account ids are UUIDs and never reused, so it stays unreachable).
+        try:
+            await PreferenceRepository(db=graphql_context.db).delete_for_owner(owner_id=obj.id)
+        except Exception as exc:
```
saltas888: Can we handle whats only likely to be thrown? handling any exception is a documented anti-pattern.

backend/infrahub/graphql/mutations/account.py
```diff
@@ -220,4 +226,19 @@ async def mutate_delete(
...
+
+        # Accounts are branch-agnostic, so the deletion is effective everywhere at once; drop the
+        # account's Preference row (a StandardNode keyed by owner_id, which cannot cascade via the
+        # schema) instead of leaving it behind as unreachable dead data (IFC-2867). The per-owner
+        # preference lock serialises this with InfrahubSetPreferences' read-modify-write, so an
+        # in-flight upsert cannot resurrect the row mid-delete. Best-effort: the account is already
+        # deleted, so a cleanup failure must not fail the mutation — the orphaned row is benign
+        # (account ids are UUIDs and never reused, so it stays unreachable).
+        try:
+            async with lock.registry.get(name=obj.id, namespace=PREFERENCE_LOCK_NAMESPACE, local=False):
```
saltas888, line 242: Why is acquiring the lock necessary here?

backend/infrahub/core/preferences/repository.py
pa-lem: Good catch — replaced the fetch-then-delete-each loop with a single `PreferenceDeleteByOwnerQuery` (`MATCH (n:Preference { owner_id: $owner_id }) DETACH DELETE n`) in 802981b684.

backend/infrahub/graphql/mutations/account.py
pa-lem: Agreed — narrowed to what the block can actually raise: `DatabaseError` (infrahub's wrapper), `Neo4jError`/`DriverError` (query execution), and `RedisError` (the distributed lock). Fixed in 802981b684.

backend/infrahub/graphql/mutations/account.py
pa-lem: `InfrahubSetPreferences` upserts under this per-owner lock as a read-then-save (preferences.py `_set`). Without taking the same lock here, the delete can interleave between that read and save — the save then re-creates the row for the just-deleted account, i.e. a permanent orphan. The lock serialises the two, so an in-flight upsert either completes before the delete (row removed) or starts after it. I've expanded the code comment to spell this out (802981b684).

backend/infrahub/graphql/mutations/account.py
```diff
@@ -220,4 +228,20 @@ async def mutate_delete(
...
+        # schema) instead of leaving it behind as unreachable dead data (IFC-2867). The per-owner
+        # preference lock is required because InfrahubSetPreferences upserts under it as a
+        # read-then-save: without the lock this delete can run between that read and save, and the
+        # save re-creates the row for the just-deleted account — a permanent orphan. Best-effort:
+        # the account is already deleted, so a cleanup failure must not fail the mutation — the
+        # orphaned row is benign (account ids are UUIDs and never reused, so it stays unreachable).
+        try:
+            async with lock.registry.get(name=obj.id, namespace=PREFERENCE_LOCK_NAMESPACE, local=False):
+                await PreferenceRepository(db=graphql_context.db).delete_for_owner(owner_id=obj.id)
+        except (DatabaseError, DriverError, Neo4jError, RedisError) as exc:
```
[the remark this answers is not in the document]
pa-lem: Partially valid — assessed both: `QueryTimeoutError` is real (raised by `execute_query_with_metadata` on `TransactionTimedOut`, and it subclasses `Error`, not `DatabaseError`), added to the tuple. `RuntimeError` is deliberately not caught: `__aexit__` only releases after a successful acquire, so an escaping `RuntimeError` means a lock-implementation bug — a programming error this best-effort handler should not swallow (per dev/guidelines/backend/python.md). Keeping the named tuple rather than reverting to bare `Exception`.

backend/infrahub/graphql/mutations/account.py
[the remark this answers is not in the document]
pa-lem: Assessed: the in-flight interleaving is closed by taking the lock (802981b). The residual — a writer that acquires the lock entirely after cleanup — is accepted by design: it requires an authenticated request from an already-deleted account, and it produces exactly the benign-orphan class the Preference docstring documents (unreachable forever, ids never reused). Rejecting writes for deleted owners would add an account-existence query to every InfrahubSetPreferences call, which isn't worth it for that residual.

backend/infrahub/graphql/mutations/account.py
```diff
@@ -220,4 +234,20 @@ async def mutate_delete(
...
+        result = await super().mutate_delete(info=info, data=data, branch=branch)
+
+        # Accounts are branch-agnostic, so the deletion is effective everywhere at once; drop the
+        # account's Preference row (a StandardNode keyed by owner_id, which cannot cascade via the
+        # schema) instead of leaving it behind as unreachable dead data (IFC-2867). The per-owner
+        # preference lock is required because InfrahubSetPreferences upserts under it as a
+        # read-then-save: without the lock this delete can run between that read and save, and the
+        # save re-creates the row for the just-deleted account — a permanent orphan. Best-effort:
+        # the account is already deleted, so a cleanup failure must not fail the mutation — the
+        # orphaned row is benign (account ids are UUIDs and never reused, so it stays unreachable).
```
saltas888: this comment is quite big to read. I understand its something specific, but can we trim it down to 2-3 lines so its more reader friendly?

dev/knowledge/backend/preferences.md
```diff
@@ -108,7 +108,9 @@ source.
 ## Caveats
 
 - **Orphan rows.** A `StandardNode` cannot declare an `on_delete: cascade` schema relationship, so
-  deleting an account leaves its `Preference` row behind. Account ids are UUIDs and are never reused,
-  so the row is permanently unreachable and benign. Cleanup is out of scope for V1 (tracked in Jira).
+  the schema cannot cascade a `Preference` row when its account is deleted. Instead the account-delete
+  mutation drops the row explicitly, best-effort and under the per-owner preference lock (IFC-2867).
+  A deletion path that bypasses the mutation still leaves the row behind; that orphan is benign:
+  account ids are UUIDs and are never reused, so it is permanently unreachable.
```
saltas888, line 119: great to see the knowledge to be maintained`!

saltas888 (approved):
approved with a minor comment about comments!

backend/infrahub/core/preferences/models.py
```diff
@@ -17,10 +17,10 @@ class Preference(StandardNode):
     client's built-in default).
 
     `owner_id` is a plain string, not a graph relationship: a StandardNode cannot declare a schema
-    relationship with `on_delete: cascade` (that is a schema-Node feature), so deleting an account
-    leaves its Preference row behind as unreachable dead data. Account ids are UUIDs and never reused,
-    so such a row is permanently unreachable and benign. Cleanup is out of scope for V1 and tracked in
-    Jira (IFC-2867).
+    relationship with `on_delete: cascade` (that is a schema-Node feature), so the schema cannot
+    cascade a Preference row when its account is deleted. Instead the account-delete mutation drops
+    the row explicitly (IFC-2867). A row orphaned by a deletion that bypasses the mutation stays
```
ajtmccarty: jira ID in here

backend/infrahub/core/query/preference.py
```diff
@@ -92,6 +92,28 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
+    insert_return = False
+
+    def __init__(self, owner_id: str, **kwargs: Any) -> None:
+        self.owner_id = owner_id
+        super().__init__(**kwargs)
+
+    async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa: ARG002
+        self.params["owner_id"] = self.owner_id
+
+        # The label is the Preference StandardNode type name (Cypher labels can't be parameterised).
```
ajtmccarty: unnecessary comment. also not true. cypher labels can be parameterized: `MATCH (n:$($node_label))`, I believe

backend/infrahub/graphql/mutations/account.py
```diff
@@ -220,4 +234,16 @@ async def mutate_delete(
         if graphql_context.account_session and obj.id == graphql_context.account_session.authenticating_account_id:
             raise ValidationError(input_value="Cannot delete your own account")
 
-        return await super().mutate_delete(info=info, data=data, branch=branch)
+        result = await super().mutate_delete(info=info, data=data, branch=branch)
+
+        # Drop the account's Preference row (no schema cascade for StandardNodes — see the
+        # Preference docstring, IFC-2867). The lock serialises with InfrahubSetPreferences'
+        # read-then-save so an in-flight upsert cannot re-create the row. Best-effort: the account
+        # is already deleted, so a cleanup failure only logs (the orphaned row is benign).
```
ajtmccarty: I think the whole comment is unnecessary, but definitely the Jira ID.

changelog/+ifc-2867.fixed.md
ajtmccarty, line 1: I don't think this is necessary for something that has not been released

dev/knowledge/backend/preferences.md
```diff
@@ -108,7 +113,9 @@ source.
 ## Caveats
 
 - **Orphan rows.** A `StandardNode` cannot declare an `on_delete: cascade` schema relationship, so
-  deleting an account leaves its `Preference` row behind. Account ids are UUIDs and are never reused,
-  so the row is permanently unreachable and benign. Cleanup is out of scope for V1 (tracked in Jira).
+  the schema cannot cascade a `Preference` row when its account is deleted. Instead the account-delete
+  mutation drops the row explicitly, best-effort and under the per-owner preference lock (IFC-2867).
```
ajtmccarty: another Jira ID

ajtmccarty (approved):
everything looks good functionally and the tests cover it. mostly comments about comments

backend/infrahub/core/preferences/models.py
pa-lem: Removed in 63cbee9597 (docstring reworded without the ID).

backend/infrahub/core/query/preference.py
pa-lem: Right on both counts — comment removed (from all three preference queries, same false claim). Also captured the dynamic-label syntax in the cypher guide via #10005 so it doesn't get repeated.

backend/infrahub/graphql/mutations/account.py
pa-lem: Trimmed to two lines in 63cbee9597 — kept only the lock rationale and the best-effort contract, no ID.

changelog/+ifc-2867.fixed.md
pa-lem: Agreed — fragment removed in 63cbee9597 (preferences are unreleased, so this fix is not user-observable). Also added that carve-out to dev/guidelines/changelog.md via #10005.

dev/knowledge/backend/preferences.md
pa-lem: Removed in 63cbee9597. The code-doc-style rule only scoped to backend/**/*.py, which is why this slipped through — #10005 extends it to the dev/ docs.

backend/infrahub/graphql/mutations/account.py
pa-lem: Trimmed to two lines in 63cbee9597.

PR #10005, opened 2026-07-22
docs(agents): branching strategy, component-test Docker note, set-based Cypher writes

pa-lem:
## Summary

Three small context improvements from the IFC-2867 session retrospective (PR #10000):

- **Branching** (new `AGENTS.md` section): feature/bugfix branches start from `develop`; verify the code a ticket references exists on the chosen base before cutting a branch — the session first cut its branch off `stable`, where the referenced module doesn't exist yet.
- **Component tests & Docker** (`AGENTS.md` testing section): `backend/tests/component/` needs a running Docker daemon, and testcontainers only looks at `/var/run/docker.sock` — documented the `DOCKER_HOST` override for daemons exposing a user socket only.
- **Set-based Cypher writes** (`neo4j-cypher-guide` skill): prefer one conditioned write query over fetch-then-mutate-each (N+1) — a reviewer had to point this out on #10000.

---
## Summary by cubic
Clarifies branching rules, documents Docker needs for component tests, and expands the Neo4j Cypher guide with set-based writes and dynamic label/type guidance. Also links the backend feature checklist, tightens exception handling guidance, and updates changelog/docs rules (IFC-2867 follow-up).

- **Guideline updates**
  - Branching in `dev/guidelines/git-workflow.md`: features from `develop`; bug fixes to the oldest affected branch; verify the base contains the code.
  - Component tests: default path needs a Docker daemon; set `INFRAHUB_USE_TEST_CONTAINERS=false` to reuse an existing DB.
  - Backend docs: link `dev/guidelines/backend/checklist.md`; add an exception-hierarchy note; fix the coding-standards list glitch.
  - Cypher guide: add set-based write guidance and dynamic labels/types (`$($param)`, Neo4j 5.26+) with planner/plan-cache tradeoffs.
  - Changelog/docs: skip fragments for fixes to unreleased features; avoid ticket/spec IDs in documentation.

<sup>Written for commit ded472a4bc6d5142537e2e21b1585f687f7c032f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/rules/code-doc-style.md
+++ b/.agents/rules/code-doc-style.md
@@ -2,16 +2,11 @@
 paths:
   - "backend/**/*.py"
   - "python_testcontainers/**/*.py"
-  - "dev/knowledge/**/*.md"
-  - "dev/guidelines/**/*.md"
-  - "dev/guides/**/*.md"
 ---
 
 # Code documentation style
 
-Applies to docstrings, comments, and any inline documentation in Python source files. The
-work-item-ID rule below also applies to the internal docs under `dev/knowledge/`,
-`dev/guidelines/`, and `dev/guides/`.
+Applies to docstrings, comments, and any inline documentation in Python source files.
 
 ## No references to other code
 
@@ -31,9 +26,7 @@ Acceptable exceptions:
 
 ## No work-item or spec IDs
 
-Do not reference Jira tickets, GitHub issues, or spec-kit identifiers in docstrings, comments, test
-names, or `dev/knowledge`/`dev/guidelines`/`dev/guides` pages — describe the constraint or behavior
-itself, not the ticket that tracked it. Examples of what to avoid:
+Do not reference Jira tickets, GitHub issues, or spec-kit identifiers in docstrings, comments, or test names. Examples of what to avoid:
 
 - `# Fixes INFP-556`
 - `"""Implements FR-003: ..."""`
--- a/.agents/rules/code-doc-style.md
+++ b/.agents/rules/code-doc-style.md
@@ -2,11 +2,16 @@
 paths:
   - "backend/**/*.py"
   - "python_testcontainers/**/*.py"
+  - "dev/knowledge/**/*.md"
+  - "dev/guidelines/**/*.md"
+  - "dev/guides/**/*.md"
 ---
 
 # Code documentation style
 
-Applies to docstrings, comments, and any inline documentation in Python source files.
+Applies to docstrings, comments, and any inline documentation in Python source files. The
+work-item-ID rule below also applies to the internal docs under `dev/knowledge/`,
+`dev/guidelines/`, and `dev/guides/`.
 
 ## No references to other code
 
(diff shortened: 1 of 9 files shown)


backend/AGENTS.md
```diff
@@ -31,13 +31,19 @@ uv run invoke backend.generate         # Regenerate schemas/protocols
...
 - Pydantic models
 - Docstring conventions
 - Naming conventions
 - Query patterns
+- Exception handling (catch the narrowest types the call path actually raises)
+
+When planning or implementing a backend feature, also walk
+`dev/guidelines/backend/checklist.md` — migrations, query efficiency (set-based, no N+1),
+permissions, error handling.
 - Type hints
```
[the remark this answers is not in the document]
pa-lem: Valid — the paragraph was inserted mid-list, stranding the Type hints bullet. Fixed: Type hints is back with the bullets, the checklist paragraph follows the complete list.

.agents/rules/code-doc-style.md
```diff
@@ -2,11 +2,16 @@
 paths:
   - "backend/**/*.py"
   - "python_testcontainers/**/*.py"
+  - "dev/knowledge/**/*.md"
+  - "dev/guidelines/**/*.md"
+  - "dev/guides/**/*.md"
```
ajtmccarty: this doesn't seem quite right. those aren't directories for actual code
also, a possibly overlapping PR: #9991

.agents/skills/neo4j-cypher-guide/SKILL.md
```diff
@@ -211,6 +220,17 @@ WHERE n:Label1&Label2  // AND
 WHERE n:!Archived      // NOT
 ```
 
+### Dynamic Labels and Types
+
+Labels and relationship types accept parameters (Neo4j 5.26+) — do not interpolate them into the
+query string with `%`/f-strings, and do not claim labels "can't be parameterized":
```
ajtmccarty: we don't always want to use dynamic labels b/c it has performance implications. the tradeoff is that a query using interpolated labels will be cached differently from a query using dynamic labels, but the query planner should do a better job with the interpolated labels than the dynamic labels. so I think it actual requires some testing to be sure which approach is correct for a given query. generally, I think that if the label is the primary filter then it should use a static/interpolated string. if the `MATCH` can be filtered using other indexes, then a dynamic label is more likely to be acceptable

.agents/rules/code-doc-style.md
pa-lem: Agreed — reverted; code-doc-style is back to develop's version (paths and text). Scoping it to dev/ pages was wrong twice over: they aren't code, and the rule's no-code-references section would forbid knowledge docs from naming classes, which is their whole job. The no-work-item-IDs point for internal docs now lives as one line in the Don't list of dev/guidelines/documentation.md instead. Thanks for the #9991 pointer — no remaining overlap after the rebase.

.agents/skills/neo4j-cypher-guide/SKILL.md
pa-lem: Thanks — reworded with exactly that framing: the section now presents dynamic vs interpolated as a performance tradeoff (per-variant plan cache vs planner quality), recommends a static/interpolated label when the label is the primary filter, allows dynamic when other indexed predicates carry the filtering, and says to PROFILE when in doubt. Dropped the 'never interpolate' imperative.

ajtmccarty (approved):

PR #10014, opened 2026-07-22
fix: show changed nodes in diff tree when their parent is unchanged

### LINKED ISSUE #10010: bug: diff view tree omits changed nodes whose parent node is unchanged
### Component

Frontend UI

### Infrahub version

1.10.0 (regression present since #5676, Feb 2025)

### Current Behavior

On the branch/proposed-change diff Data view, the left-hand tree omits changed nodes when their hierarchical parent has no changes of its own. The right-hand diff list shows them correctly, so the two panes disagree.

The `DiffTree` GraphQL response (queried with `include_parents: true`) intentionally contains the parent nodes with `status: UNCHANGED` as hierarchy context (added by `parent_node_adder.py`). The frontend filters out all `UNCHANGED` nodes in `entities/diff/ui/node-diff/index.tsx` **before** building the tree, and `buildDiffTreeItems` (`entities/diff/utils/build-diff-tree-items.ts`) then silently drops any node whose `parent.uuid` cannot be found among the tree items.

The drop is transitive: a *changed* parent that was itself dropped takes its changed children with it (e.g. an `ADDED` prefix under an `UPDATED` prefix under an unchanged root prefix disappears too).

### Expected Behavior

Every changed node visible in the right-hand diff list also appears in the left tree, nested under its (unchanged) parent — as it did before the regression, where unchanged parents were rendered as context items in the tree.

### Steps to Reproduce

Reproducible on https://sandbox.infrahub.app (1.10.0), demo login, no data changes needed:

1. Open **Branches › `jfk1-update-edge-ips` › Data**.
2. The right pane shows i.a. two `Updated Infra › Interface L3 Ethernet1` (parents `jfk1-edge1`/`jfk1-edge2` are unchanged), `Updated Ipam › IP Prefix 10.1.0.0/16`, and `Added Ipam › IP Prefix 10.1.0.32/31`.
3. The left tree contains only `IP Address → 10.1.0.33/31, 10.1.0.32/31` and `IPAM Namespace → default`. Both interfaces and both prefixes are missing (verified via the tree DOM — not a collapsed-node issue).

Originally reported by a customer (Slack, #ext-rolex, 2026-07-22): an updated Loadbalancer Pool Member shown in the right pane had no entry in the left tree because its pool was unchanged.

### Additional Information

- Regression origin: commit `1c4a0f9ce` (#5676, "Incrementally load node diff in UI view") moved the `UNCHANGED` filter from the right-pane render into the shared `nodes` computation feeding both panes. Before that (#4199), the tree received all nodes and only the right pane filtered `UNCHANGED`.
- Drop site: `frontend/app/src/entities/diff/utils/build-diff-tree-items.ts` — `if (!parentTreeItem) return acc;`. This is also order-sensitive for multi-level parent chains.
- Screenshot of the sandbox reproduction available.

lancamat1:
# Why

On the branch/proposed-change diff Data view, the left-hand tree silently omitted changed nodes whenever their hierarchical parent had no changes of its own — while the right-hand diff list showed them correctly, so the two panes disagreed. The drop was transitive: a changed parent that was itself dropped took its changed children with it. Regression since #5676 (1c4a0f9ce), originally reported by a customer whose updated Loadbalancer Pool Member had no entry in the tree because its pool was unchanged.

This PR makes every changed node visible in the right-hand diff list also appear in the left tree, nested under its (unchanged) parent — as it did before the regression. Per review, it also fixes the same pane disagreement under the Conflicts filter (`?status=CONFLICT`), where unchanged ancestors were dropped because `contains_conflict` never propagates to them.

Non-goals: no backend or GraphQL changes (the backend already returns unchanged parents on purpose via `include_parents: true`), no change to `contains_conflict` semantics, no pagination changes.

Closes #10010

## What changed

Behavioral changes:

- The diff tree again renders `UNCHANGED` parent nodes as hierarchy context, so changed nodes nest under them instead of disappearing.
- With the Conflicts filter active, the tree keeps conflicted nodes plus their ancestor chains, so a conflicted node under an unchanged parent stays anchored; the right pane shows only changed nodes with conflicts, as before.
- The right-hand diff list and its empty state are otherwise unaffected: they still show only changed nodes.
- Multi-level parent chains no longer depend on the order nodes arrive in: `buildDiffTreeItems` pre-creates one tree item per node in a `Map` keyed by `uuid`, so ancestors delivered after descendants (possible across infinite-query pages) no longer cause drops.

Implementation notes:

- Root cause: #5676 moved the `UNCHANGED` filter from the right-pane render into the shared `nodes` computation feeding **both** panes, starving `buildDiffTreeItems` of the parents it needs to anchor children (`if (!parentTreeItem) return acc;` then silently dropped them).
- Fix restores the pre-regression split of concerns in `node-diff/index.tsx`: the tree receives all nodes, a `changedNodes` list feeds the right pane. The client-side CONFLICT QSP filter is likewise split: `contains_conflict` filters only the right pane, while the tree keeps conflicted nodes plus their ancestor chains (`keepConflictNodesWithAncestors`: a `Map` of nodes by `uuid`, recursive marking of each conflicted node's parents, then a filter to the marked set).
- `buildDiffTreeItems` pairs each node with its tree item upfront (`{ node, treeItem }`), indexes the items in a `Map` keyed by `uuid`, then attaches children in a single `for...of` pass — the parent item always exists in the `Map` regardless of input order, so no retry passes are needed.
- Nodes whose parent genuinely never appears in the payload are still skipped, as before.

What stayed the same: no schema or API changes; no changes to `DiffTree` rendering — the `UNCHANGED` badge variant already existed from the pre-regression behavior.

## How to review

- `frontend/app/src/entities/diff/ui/node-diff/index.tsx` — the root-cause fix (filter split, for both the `UNCHANGED` and CONFLICT filters).
- `frontend/app/src/entities/diff/utils/build-diff-tree-items.ts` — order-independent single-pass attach via a `Map` of pre-created tree items.
- `frontend/app/src/entities/diff/ui/node-diff/index.test.tsx` — replication tests written first and verified failing before the fix (component-level, mocks only the two GraphQL API functions): changed node under unchanged parent, and conflicted node under unchanged parent with `?status=CONFLICT`.
- `frontend/app/.betterer.results` — mechanical fingerprint/line shifts for the edited files, no new type issues.

## How to test

```bash
cd frontend/app && pnpm run test src/entities/diff
```

The new tests fail without the fix and pass on this branch. Manual verification: on a demo dataset, open **Branches › `jfk1-update-edge-ips` › Data** — the updated `Interface L3 Ethernet1` nodes (under unchanged devices) and the updated/added IP prefixes (under an unchanged root prefix) now appear in the left tree.

Full gates run locally: frontend suite (118 files / 807 tests), backend unit tests (1400), `biome ci`, `knip`, `betterer ci`.

## Impact & rollout

- **Backward compatibility:** none — frontend-only rendering fix restoring intended behavior.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (not needed — restores documented/expected behavior)
- [ ] Internal .md docs updated (not needed — no architectural change)

---
## Summary by cubic
Fixes the diff tree dropping changed nodes under UNCHANGED parents and keeps conflicted nodes anchored when the Conflicts filter is active. Children stay visible even if their parent arrives later. Addresses #10010.

- **Bug Fixes**
  - Keep UNCHANGED parents in the tree; under `?status=CONFLICT`, keep conflicted nodes with their ancestors. The right pane still shows only changed nodes (and only conflicted ones under the filter).
  - Attachment is order-independent across pages; sibling order follows input.

- **Refactors**
  - Rewrote `buildDiffTreeItems` to a single pass with a `Map` keyed by `uuid`, removing the recursive lookup and a tracked type error; tests added/updated.
  - Formatted `node-diff/index.tsx` to match CI; updated Betterer results.

<sup>Written for commit 042b0d5b13f47d4dd98cd82e5f29a243fc9c73fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10014?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/app/src/entities/diff/ui/node-diff/index.tsx
+++ b/frontend/app/src/entities/diff/ui/node-diff/index.tsx
@@ -93,15 +93,19 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
     );
   }
 
+  // UNCHANGED nodes are kept: the backend returns them (include_parents) so the tree
+  // can nest changed nodes under their unchanged parents as hierarchy context
   const nodes =
     data.pages
       .flatMap((page) => page?.nodes)
       .flatMap((node) => {
-        if (!node || node.status === "UNCHANGED") return [];
+        if (!node) return [];
         // Manually filter conflicts items since it's not available yet in the backend filters
         if (qspStatus === DIFF_STATUS.CONFLICT && !node.contains_conflict) return [];
         return node as unknown as DiffNodeType;
       }) ?? [];
+
+  const changedNodes = nodes.filter((node) => node.status !== "UNCHANGED");
 
   return (
     <div className="flex h-[calc(100vh-14rem)] flex-col overflow-hidden">
@@ -128,8 +132,8 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
         </nav>
 
         <main className="col-start-2 col-end-5 space-y-4 overflow-auto bg-stone-100 p-4">
-          {nodes.length ? (
-            nodes.map((node) => (
+          {changedNodes.length ? (
+            changedNodes.map((node) => (
               <DiffNode
                 key={node.uuid}
                 node={node}
--- a/frontend/app/src/entities/diff/ui/node-diff/index.tsx
+++ b/frontend/app/src/entities/diff/ui/node-diff/index.tsx
@@ -95,17 +95,22 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
 
   // UNCHANGED nodes are kept: the backend returns them (include_parents) so the tree
   // can nest changed nodes under their unchanged parents as hierarchy context
-  const nodes =
+  const allNodes =
     data.pages
       .flatMap((page) => page?.nodes)
-      .flatMap((node) => {
-        if (!node) return [];
-        // Manually filter conflicts items since it's not available yet in the backend filters
-        if (qspStatus === DIFF_STATUS.CONFLICT && !node.contains_conflict) return [];
-        return node as unknown as DiffNodeType;
-      }) ?? [];
+      .flatMap((node) => (node ? [node as unknown as DiffNodeType] : [])) ?? [];
 
-  const changedNodes = nodes.filter((node) => node.status !== "UNCHANGED");
+  // The CONFLICT filter is client-side only (buildFilters excludes it from the backend
+  // status filter) and contains_conflict never propagates to ancestor nodes, so the
+  // tree keeps conflicted nodes plus their ancestor chains as hierarchy context
+  const nodes =
+    qspStatus === DIFF_STATUS.CONFLICT ? keepConflictNodesWithAncestors(allNodes) : allNodes;
+
+  const changedNodes = nodes.filter(
+    (node) =>
+      node.status !== "UNCHANGED" &&
+      (qspStatus !== DIFF_STATUS.CONFLICT || node.contains_conflict)
+  );
 
   return (
     <div className="flex h-[calc(100vh-14rem)] flex-col overflow-hidden">
@@ -149,3 +154,20 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
     </div>
   );
 };
+
+function keepConflictNodesWithAncestors(nodes: Array<DiffNodeType>): Array<DiffNodeType> {
+  const nodesByUuid = new Map(nodes.map((node) => [node.uuid, node]));
+  const keptUuids = new Set<string>();
+
+  const markNodeAndAncestors = (node?: DiffNodeType) => {
+    if (!node || keptUuids.has(node.uuid)) return;
+    keptUuids.add(node.uuid);
+    if (node.parent) markNodeAndAncestors(nodesByUuid.get(node.parent.uuid));
+  };
+
+  for (const node of nodes) {
+    if (node.contains_conflict) markNodeAndAncestors(node);
+  }
+
+  return nodes.filter((node) => keptUuids.has(node.uuid));
+}
--- a/frontend/app/src/entities/diff/ui/node-diff/index.tsx
+++ b/frontend/app/src/entities/diff/ui/node-diff/index.tsx
@@ -108,8 +108,7 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
 
   const changedNodes = nodes.filter(
     (node) =>
-      node.status !== "UNCHANGED" &&
-      (qspStatus !== DIFF_STATUS.CONFLICT || node.contains_conflict)
+      node.status !== "UNCHANGED" && (qspStatus !== DIFF_STATUS.CONFLICT || node.contains_conflict)
   );
 
   return (
--- a/frontend/app/src/entities/diff/utils/build-diff-tree-items.ts
+++ b/frontend/app/src/entities/diff/utils/build-diff-tree-items.ts
@@ -29,11 +29,17 @@ export function buildDiffTreeItems(nodes: Array<DiffNode>): Array<DiffTreeItem> 
 
   const [nodesWithoutParent, nodesWithParent] = partition(nodes, (node) => !node.parent);
 
+  // Pre-create one tree item per node so a parent can be found in O(1) even when
+  // it appears after its children in the input
+  const nodeTreeItemsByUuid = new Map(
+    nodes.map((node) => [node.uuid, createNodeTreeItemFromDiffNode(node)])
+  );
+
   const baseTreeItems = nodesWithoutParent.reduce<DiffTreeItem[]>((acc, node) => {
     const { kind } = node;
     const topLevelKindTreeItem = acc.find((treeItem) => treeItem.kind === kind);
 
-    const newDiffTreeItem = createNodeTreeItemFromDiffNode(node);
+    const newDiffTreeItem = nodeTreeItemsByUuid.get(node.uuid)!;
 
     if (!topLevelKindTreeItem) {
       const newKindTreeItem = createGroupTreeItemFromDiffNode(node);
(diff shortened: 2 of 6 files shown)


lancamat1:
Addressed the CI failure and the review findings in 6f1166b89:

- **frontend-lint (biome formatting):** the offending multi-pass block was removed entirely by the refactor below.
- **Performance (P2):** `buildDiffTreeItems` now pre-creates one tree item per node in a `Map` keyed by uuid, so parent lookup is O(1) and the whole build is a single pass over the input — no more repeated recursive tree scans. Order independence is preserved by construction (a parent's item exists before any child looks it up, regardless of input order).
- **Sibling ordering (P3):** children attach in a single pass in input order, so sibling order always follows the source `nodes` order — including when a child precedes its parent.
- **Test hygiene (P3):** the `NodeDiff` test now restores the shared Jotai store's `nodeSchemasAtom` after the suite and clears mocks between tests.

Side effect: removing `findTreeItemById` also eliminated one tracked betterer type error (211 remaining, was 212). All local gates green: biome ci (2.5.1, same as CI), knip, betterer, 807/807 frontend tests.

frontend/app/src/entities/diff/ui/node-diff/index.tsx
```diff
@@ -93,16 +93,20 @@ export const NodeDiff = ({ branch, filters }: NodeDiffProps) => {
...
+  // UNCHANGED nodes are kept: the backend returns them (include_parents) so the tree
+  // can nest changed nodes under their unchanged parents as hierarchy context
   const nodes =
     data.pages
       .flatMap((page) => page?.nodes)
       .flatMap((node) => {
-        if (!node || node.status === "UNCHANGED") return [];
+        if (!node) return [];
         // Manually filter conflicts items since it's not available yet in the backend filters
         if (qspStatus === DIFF_STATUS.CONFLICT && !node.contains_conflict) return [];
```
pa-lem: This same bug still exists under the Conflicts filter, and I think we should fix it in this PR rather than a follow-up.

The CONFLICT check is client-side only (`buildFilters` deliberately excludes CONFLICT from the backend status filter), and it runs inside the shared `nodes` list feeding the tree. Synthetic `UNCHANGED` parents always have `contains_conflict: false` — the backend computes a node's `contains_conflict` only from its own attributes/relationship elements (`summary_counts.py`), and child conflicts propagate only into relationship groups (`_add_child_nodes_to_summaries`), never into an ancestor node's `contains_conflict`, which is the field this filter reads. So with `?status=CONFLICT`, a conflicted node under an unchanged parent shows in the right pane but is dropped from the tree — the exact pane disagreement this PR fixes.

Caveat for whoever fixes it: a node with only a node-level `conflict` (e.g. delete-vs-update) but no attribute/element conflicts may also have `contains_conflict: false` — pre-existing and separate, but know the field's semantics before touching the filter.

Suggested prompt:
```
In frontend/app/src/entities/diff/ui/node-diff/index.tsx, the CONFLICT QSP filter is
applied inside the shared `nodes` computation, so UNCHANGED parent nodes (which always
have contains_conflict=false) are dropped and conflicted children lose their tree
anchor — the same bug as #10010, in the filtered view. Fix: apply the contains_conflict
filter only to `changedNodes` (right pane); for the tree, when the CONFLICT filter is
active, keep conflicted nodes PLUS their ancestor chains (build a Map of nodes by uuid,
recursively mark each conflicted node's parents, then filter nodes to the marked set).
No `let`, no `while` — use a recursive marker or reduce. Add a component test:
conflicted child under an unchanged parent with ?status=CONFLICT appears in both panes.
```

frontend/app/src/entities/diff/utils/build-diff-tree-items.ts
```diff
@@ -29,11 +29,17 @@ export function buildDiffTreeItems(nodes: Array<DiffNode>): Array<DiffTreeItem>
...
+  const nodeTreeItemsByUuid = new Map(
+    nodes.map((node) => [node.uuid, createNodeTreeItemFromDiffNode(node)])
+  );
+
   const baseTreeItems = nodesWithoutParent.reduce<DiffTreeItem[]>((acc, node) => {
     const { kind } = node;
     const topLevelKindTreeItem = acc.find((treeItem) => treeItem.kind === kind);
 
-    const newDiffTreeItem = createNodeTreeItemFromDiffNode(node);
+    const newDiffTreeItem = nodeTreeItemsByUuid.get(node.uuid)!;
```
pa-lem: Nit: the two `.get(node.uuid)!` assertions are safe by construction but avoidable — partition `nodes.map((node) => ({ node, item: createNodeTreeItemFromDiffNode(node) }))` instead, so each entry carries its own item and only the *parent* lookup needs the Map.

Suggested prompt:
```
In buildDiffTreeItems, remove the two non-null assertions by partitioning pairs of
{ node, item } created upfront, keeping the Map only for parent lookup by uuid.
```

pa-lem (commented):
Nice fix — the Map-based single pass in `buildDiffTreeItems` is the right shape, and the tests are solid. Two general points, plus two inline comments:

**1. PR description is stale.** The Implementation notes describe a "fixed-point iteration with a no-progress break" / "as many passes as needed", but the final code is a single `for...of` pass over a pre-built `Map` (no `while`, no `let`). Please update the body so it describes the actual approach.

Suggested prompt:
```
The PR body for #10014 describes an earlier implementation ("fixed-point iteration
with a no-progress break"). The actual code pre-creates all tree items in a Map keyed
by uuid and attaches children in one for...of pass. Rewrite the Implementation notes
section of the PR body to describe that approach.
```

**2. (Optional nit, file not in this diff)** In `diff-tree.tsx`, every node row gets an `href` with `hash: #<uuid>`, but the right pane only renders changed nodes (cards anchor via `id={node.uuid}` in `node-diff/node.tsx`). Clicking an `UNCHANGED` tree row is therefore a silent no-op. Fine to leave, but rendering unchanged rows without `href` would avoid the dead link.

frontend/app/src/entities/diff/ui/node-diff/index.tsx
lancamat1: Fixed in 7d0a26460. The CONFLICT QSP filter is now split like the UNCHANGED one: `contains_conflict` filters only `changedNodes` (right pane), and the tree input goes through `keepConflictNodesWithAncestors` — a Map of nodes by uuid, recursive marking of each conflicted node's ancestor chain, then a filter to the marked set (no `let`/`while`). Added the component test: a conflicted interface under an unchanged device with `?status=CONFLICT` appears in both panes, and an updated-but-not-conflicted node stays filtered out of the right pane.

frontend/app/src/entities/diff/utils/build-diff-tree-items.ts
lancamat1: Fixed in 7d0a26460 — `buildDiffTreeItems` now pairs each node with its tree item upfront (`{ node, treeItem }`) and partitions those pairs; the Map is only used for the parent lookup, so both `.get(node.uuid)!` assertions are gone.

lancamat1:
Review addressed in 7d0a26460:

1. **PR description** — Implementation notes rewritten to describe the actual approach (pre-created tree items in a `Map` keyed by uuid, single `for...of` attach pass), and extended to cover the Conflicts-filter fix now in scope.
2. **Unchanged tree rows' dead `href`** — left as-is per the "fine to leave" note.
3. **Conflicts filter** — fixed in this PR, see inline thread.
4. **Non-null assertions** — removed, see inline thread.

pa-lem (approved):

PR #10019, opened 2026-07-23
scope uniqueness constraints to affected objects

ajtmccarty:
# Why

Uniqueness validation scanned the **entire population** of a kind on every data change. Merging, rebasing, or proposing a change that touched a handful of nodes of a large kind still ran a population-wide scan — slow on big kinds (hundreds of thousands of nodes) and it shipped large duplicate result sets back to Python.

**Goal:** scope a data-triggered uniqueness check to only the nodes the diff actually affects, so the work is proportional to the size of the change rather than the size of the kind.

**Non-goals:** this does not change node-save-time uniqueness (`grouped_uniqueness`), does not touch other constraint validators, and does not change the uniqueness error surfaced to users.

**Note:** uniqueness constraints do not support attribute of relationships (e.g. `"device__name"`). some parts of the code support this kind of uniqueness constraint, but not all of it and it is still blocked when creating/updating a schema.

**Planned follow-ups**
- refactor `MergeConstraintValidator` construction. `schema_branch` should probably be an argument for this class and its dependents, not a dependency
- check on integration/functional-level testing for uniqueness constraint validation and add more if necessary
- issue for supporting attributes on relationships in uniqueness constraints

Part of IFC-2796 (epic IFC-2706).

## What changed

**Behavioral**

- Data-triggered uniqueness validation (merge, rebase, proposed-change integrity) now validates **only the changed/affected nodes** via a batched, index-anchored targeted query instead of a full-population scan.
- A uniqueness check is scoped to exactly the nodes whose changed field participates in the constraint — changing one node's unique field no longer drags in every other changed node of the kind.
- A **newly added or broadened** constraint (schema-diff origin, with no affected-node set) still runs the full-population scan, as before.
- Cross-kind constraints that read a peer's attribute (e.g. `owner__name`) fall back to full-population validation, which supports them; the targeted query does not. **Again, this type of uniqueness constraint cannot be added to schemas currently.**
- Uniqueness **error messages are unchanged** — attribute values render identically to the full-population path regardless of their type.

**Implementation notes**

- A `node_uuids` carrier is threaded end to end: `None` means full scan, a list means validate exactly those nodes. It rides from the diff summary through the determiner, the constraint info, and the validator request into the checker.
- `NodeDiffFieldSummary`/`NodeDiffIndex` now keep the correlation between each changed field and the nodes that changed it (per-field UUID maps), so scoping targets a specific field's nodes rather than every changed node of the kind.
- New here (constructor-injected, wired by `build_constraint_validator_determiner`): `UniquenessConstraintScoper` decides, per kind, whether uniqueness must run and which nodes it affects (`None` → full scan); the `ConstraintValidatorDeterminer` update threads the affected set through; constraint dedup/merge helpers collapse a node's uniqueness check into its generic's where the generic already covers it.
- `UniquenessChecker` now runs the targeted query for a scoped set, and falls back to the full-population scan when a constraint reads a peer attribute.

**What stayed the same**

- The full-population path and the node-save-time uniqueness checker (`grouped_uniqueness`) are untouched.
- No GraphQL/API contract changes, no database schema or migration changes.

### Suggested review order

1. `core/diff/model/path.py`, `core/diff/query/field_summary.py`, `core/validators/node_diff_index.py` — how per-field changed-node UUIDs are captured.
2. `core/validators/uniqueness/scope.py` + `core/validators/determiner.py` — how the affected-node set (or `None`) is decided.
3. `core/validators/uniqueness/checker.py` — running the targeted query for a scoped set vs the full-population fallback.
4. `proposed_change/tasks.py`, `merge/constraints.py`, `branch/tasks.py` — the three call sites that thread the affected set through.
5. Tests.

## How to review

Focus on the scope decision (`UniquenessConstraintScoper._compute_scope`) and the checker's targeted-vs-full-population branch (`UniquenessChecker.check` / `_supports_targeted`) — those govern correctness. Mechanical: the `attribute_node_uuids` conversions across the test files.

## Impact & rollout

- **Backward compatibility:** no breaking changes; uniqueness error messages are identical to before.
- **Performance:** the headline win. On a large local dataset (≈500k-node kind) the full-population scan ran 8+ minutes before OOM-killing Neo4j; the targeted query validates ~9,400 changed nodes of that kind in ~12s and ~23,500 in ~29s, and all implicated kinds complete in tens of seconds. Work is now bounded by the change, not the population.
- **Deployment notes:** safe to deploy; no coordinated release or migration required.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added
- [ ] External docs updated (no user-facing surface change)

---
## Summary by cubic
Scope node-level uniqueness validation to only the nodes that changed each unique field (and their dependents), using diff-derived IDs and batched, index-backed queries. Full scans now run only when a constraint is added/broadened or when peer-attribute elements are present, cutting worst-case merge and schema-update time (IFC-2796).

- **Bug Fixes**
  - Fall back to full-population validation for peer-attribute paths (e.g., owner__name) unsupported by the targeted query.
  - Reset scoper cache on determiner re-init to prevent stale scopes between diffs.
  - Stringify targeted values in error paths to match full-scan messages.

- **Refactors**
  - Batch checks per constraint group via `TargetedUniquenessValidationQuery`, paging over changed nodes with one read-only session per change.
  - `UniquenessChecker` reads branch/schema from the validator request; dependency builder and m018 migration updated.
  - Carry `node_uuids` through `SchemaUpdateConstraintInfo` and validator requests across proposed-change, merge, rebase, and migrations; merge/dedup constraints and standardize on `ConstraintIdentifier.NODE_UNIQUENESS_CONSTRAINTS_UPDATE`.
  - Consolidate chunking into `infrahub.utilities.chunks.chunked` and replace local helpers.
  - Refactor `UniquenessConstraintScoper` for clarity; trim docstrings.

<sup>Written for commit b7c6994889b3327c2325f8afbfc93a1a178f9f1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/constraints.py
```diff
@@ -95,27 +99,35 @@ class MergeConstraintValidator:
...
         self.db = db
         self.branch = branch
         self.diff_repository = diff_repository
+        self.migration_validator = migration_validator
 
     async def validate(
         self, candidate_schema: SchemaBranch, schema_diff_constraints: list[SchemaUpdateConstraintInfo]
     ) -> MergeConstraintValidationResult:
-        determiner = ConstraintValidatorDeterminer(schema_branch=candidate_schema)
+        determiner = build_constraint_validator_determiner(
```
ajtmccarty, line 117: this should be built as part of construction, but the dependencies require `schema_branch` which is not ready until runtime. I think the right solution is to move `schema_branch` from dependency to runtime argument for all of these, but I'm going to do that separately

backend/infrahub/core/validators/determiner.py
```diff
@@ -319,3 +265,21 @@ async def _get_constraints_for_one_field(
 
             constraints.append(SchemaUpdateConstraintInfo(constraint_name=constraint_name, path=schema_path))
         return constraints
+
+
+def build_constraint_validator_determiner(
```
saltas888, line 270: only matter of taste but I like this pattern when we need this factory functions for collocation and less imports. Would it be possible here?

```
class ConstraintValidatorDeterminer
   @classmethod
    def from_(cls, ...) -> "ConstraintValidatorDeterminer":
        ....
        return cls(....)
```

saltas888: @polmichel explained that this is intentional and we prefer this pattern to separate concerns

polmichel: My opinion is that we'd rather limit these class methods to prevent increasing class responsibilities. A small component or method can take care of the object instantiation.

backend/infrahub/core/validators/uniqueness/checker.py
```diff
@@ -46,32 +56,175 @@ def get_attribute_path_from_string(
 
 class UniquenessChecker(ConstraintCheckerInterface):
     def __init__(
-        self, db: InfrahubDatabase, branch: Branch | str | None = None, max_concurrent_execution: int = 5
+        self,
+        db: InfrahubDatabase,
+        max_concurrent_execution: int = 5,
+        query_batch_size: int = 500,
```
saltas888, line 57: I think a comment on how we decided to be 500, even if its a  heurestic would be useful for the future

backend/infrahub/core/validators/constraint_merge.py
```diff
@@ -0,0 +1,46 @@
...
+    then node-level uniqueness checks already covered by an implicated generic are dropped.
+    """
+
+    def __init__(self, deduplicator: UniquenessConstraintDeduplicator) -> None:
+        self.deduplicator = deduplicator
+
+    def merge(self, *constraint_lists: list[SchemaUpdateConstraintInfo]) -> list[SchemaUpdateConstraintInfo]:
+        # Collapse the same constraint from multiple producers onto one entry. A constraint both
+        # broadened by a schema change (full population) and hit by a data change (node-scoped) must
+        # re-check every node, so a None node_uuids always wins; two node-scoped entries union.
```
saltas888: should this be a docstring?

backend/infrahub/core/validators/uniqueness/checker.py
```diff
@@ -19,13 +18,24 @@
...
+    from infrahub.core.schema.schema_branch import SchemaBranch
     from infrahub.database import InfrahubDatabase
 
     from ..model import SchemaConstraintValidatorRequest
+    from .query import TargetedUniquenessViolation
+
+
+def _chunked(items: list[str], size: int) -> Iterator[list[str]]:
+    for start in range(0, len(items), size):
+        yield items[start : start + size]
```
saltas888: this is very similar with https://github.com/opsmill/infrahub/blob/develop/backend/infrahub/core/merge/recompute_coalescing.py#L176

backend/infrahub/core/validators/determiner.py
ajtmccarty: yes I agree with Pol here. constructing the class is not the responsibility of the class

backend/infrahub/core/validators/constraint_merge.py
ajtmccarty: yes it should. updated

backend/infrahub/core/validators/uniqueness/checker.py
ajtmccarty: small comment added

backend/infrahub/core/validators/uniqueness/checker.py
ajtmccarty: added consolidated `chunks` utility in https://github.com/opsmill/infrahub/pull/10019/commits/25fbe7f73b64247d050d0de20873570f4916b946

backend/infrahub/core/diff/model/path.py
```diff
@@ -102,10 +102,41 @@ def __str__(self) -> str:
 
 @dataclass
 class NodeDiffFieldSummary:
+    """Which fields changed for a kind, and which nodes changed each one.
+
+    The per-field maps keep the correlation between a changed field and the nodes that changed it,
+    so a consumer can scope work to the nodes touching a specific field rather than every changed
+    node of the kind. ``attribute_names``/``relationship_names``/``node_uuids`` are derived views.
+    """
+
```
polmichel, line 111: This looks pretty useful, I'll keep this data structure in mind.

backend/infrahub/core/validators/uniqueness/scope.py
```diff
@@ -0,0 +1,190 @@
...
+        node_uuids = set(scope.object_uuids)
+        for peer_change in scope.cross_kind_peer_changes:
+            if not peer_change.changed_peer_uuids:
+                # the changed peers are not identified, so the constrained nodes cannot be either
+                return None
+            node_uuids |= await self.dependent_resolver.resolve(
+                node_kind=schema.kind,
+                relationship_identifier=peer_change.relationship_identifier,
+                peer_uuids=sorted(peer_change.changed_peer_uuids),
+            )
```
polmichel, line 102: `DependentResolver` object usage is clearer now. Nice injection

backend/infrahub/core/validators/uniqueness/scope.py
```diff
@@ -0,0 +1,190 @@
...
+            return None
+        return sorted(node_uuids)
+
+    def _diffed_kinds_with_field(self, schema: MainSchemaTypes, field_name: str, is_relationship: bool) -> set[str]:
+        """Return the diffed kinds where `field_name` changed, on `schema` or a kind that inherits it.
+
+        An inherited field keeps its name on the implementing kind, so a generic-level constraint
+        is implicated when an implementation's copy of the field is what changed in the diff. A
+        generic knows its implementing kinds through `used_by`; the diff check then keeps only the
+        kinds actually present in the diff.
```
polmichel: Not sure whether we should keep the last sentence since the `used_by` is an implementation detail, which should be documented on its own.

backend/infrahub/core/validators/uniqueness/scope.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _compute_scope(self, schema: MainSchemaTypes) -> UniquenessScopeForKind:
+        """Compute why and how a data change implicates `schema`'s uniqueness.
+
+        Uniqueness spans single unique attributes and multi-field constraint groups. A group
+        element such as "owner__name" reads an attribute of a related peer, so a data change on
+        the peer kind can create a violation without any change to the constrained kind itself:
+        that case yields a cross-kind peer change. Directly changed nodes of the constrained kind
+        are collected as object uuids.
+        """
```
polmichel, line 148: This algorithm is really clean.

The method is a bit long though. From a readability perspective, would it be better to extract the two parts (unique_attributes / uniqueness_constraints) into private methods, each returning `"UniquenessScopeFragment"` objects? This method would then only assemble them and pass them to the `UniquenessScopeForKind` constructor. This would bring a better readability without having to know about the implementation details, and would give `UniquenessScopeForKind` a certain level immutability

Just an idea, feel free to ignore it or to do it differently.

polmichel (approved):
Great PR and feature. Nice use of dependency injection, which makes the code readable and maintainable. The tests aren't all integration tests, and in my opinion that's a sign the design is good.

A few optional comments.

backend/infrahub/core/validators/uniqueness/scope.py
ajtmccarty: good point. trimmed the last sentence

backend/infrahub/core/validators/uniqueness/scope.py
ajtmccarty: good suggestion. refactored and it is easier to read now

PR #10020, opened 2026-07-23
test(backend): merge/rebase recompute of cross-rel HFID + label cascade

saltas888:
## What

Adds `integration_docker` coverage for two merge/rebase recompute cases the existing suite missed — its synthetic schema used a self-only HFID and self-only chain labels, so neither a cross-relationship HFID nor a display-label cascade was exercised:

1. **Cross-relationship HFID** — an interface whose `human_friendly_id` reads its device across the relationship (`["name__value", "device__name__value"]`). Rename the device, merge/rebase, expect the stored HFID to refresh.
2. **Two-level display-label cascade** — metro → site → rack; rename the metro, expect both levels' display labels to refresh two hops down (through a computed short-name).

Plus a component-level check that the coalesced builder *schedules* an HFID recompute for a cross-relationship HFID (vs none for a self-only HFID).

## Results on `develop`

| Case | merge | rebase |
|---|:---:|:---:|
| Cross-relationship HFID | ❌ stale | ❌ stale |
| Two-level label cascade | ✅ | ✅ |

- **Label cascade: fixed** by the coalescing — both pass.
- **Cross-relationship HFID: broken** — after the merge/rebase the stored HFID keeps the *old* peer value, while the display label on the same node (reading the same renamed peer) refreshes correctly. The coalesced builder *does* schedule the HFID recompute; the gap is downstream in the render/write step. Filed as **IFC-2937**.

The two HFID tests are committed as `@pytest.mark.xfail(strict=True)`, so they document the intended behavior and turn into a hard failure the moment the bug is fixed — remove the marker then.

## Notes

- **Draft** — opened to carry the repro alongside IFC-2937; not intended to merge as-is.
- Test-only change; no changelog fragment.
- `integration_docker` tier (full stack). Run:
  `uv run pytest backend/tests/integration_docker/test_merge_recompute.py -k "cross_relationship_hfid or two_level_display_label_cascade"`

---
## Summary by cubic
Adds `integration_docker` and component tests to verify recompute behavior for a cross-relationship HFID and a two-level display-label cascade on merge/rebase. The cascade passes; the cross-relationship HFID stays stale and is tracked in IFC-2937 (tests are marked xfail, strict).

- **New Features**
  - Integration tests:
    - Cross-relationship HFID (interface reads device): rename device, then merge/rebase; expect HFID refresh. Currently fails on `develop` and is xfail (IFC-2937).
    - Two-level label cascade (metro → site → rack): top-level rename refreshes both levels’ labels. Passes.
  - Component test: verifies the coalesced builder schedules an HFID recompute for cross-relationship HFID on merge and rebase.
  - Test helpers: add schema variants for device/interface HFID and location cascade, plus a flag to build a cross-relationship HFID in the profile schema.

<sup>Written for commit 2541c40203db652f71ee673d26c7ee184f9c5b70. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10020?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/component/merge_recompute_coalescing/test_merge_submits_cross_relationship_hfid.py
+++ b/backend/tests/component/merge_recompute_coalescing/test_merge_submits_cross_relationship_hfid.py
@@ -1,0 +1,141 @@
+"""A human-friendly id reading a peer across the relationship is recomputed on merge and rebase.
+
+When the id reads only the node's own name a peer rename never touches it. When the id reads the
+peer across the relationship the peer rename has to schedule an HFID recompute like the other two
+families. These assert the latter, the case a self-only id never exercises.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from uuid import uuid4
+
+from infrahub import lock
+from infrahub.auth.session import AccountSession
+from infrahub.auth.types import AuthType
+from infrahub.context import InfrahubContext
+from infrahub.core.branch.tasks import merge_branch, rebase_branch
+from infrahub.core.diff.coordinator import DiffCoordinator
+from infrahub.dependencies.registry import get_component_registry
+from infrahub.workers.dependencies import build_cache, build_database, build_event_service, build_workflow
+from infrahub.workflows.catalogue import (
+    COMPUTED_ATTRIBUTE_PROCESS_JINJA2,
+    DISPLAY_LABELS_PROCESS_JINJA2,
+    HFID_PROCESS,
+)
+from tests.adapters.cache import MemoryCache
+from tests.adapters.event import MemoryInfrahubEvent
+from tests.adapters.workflow import WorkflowRecorder
+from tests.helpers.merge_recompute.dataset import load_profile_schema, seed_branch
+
+if TYPE_CHECKING:
+    from fast_depends import Provider
+
+    from infrahub.core.branch import Branch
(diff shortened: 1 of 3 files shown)


gmazoyer:
Thanks for the reproduction Dimitris. I checked this on a real stack and there is no server bug. The coalesced merge and rebase recompute already refreshes the stored cross-relationship HFID.

The stale value comes from the SDK. `InfrahubNode.hfid` is computed on the client side from the relationship peer via the client store, so it reads stale (or `None`) unless the peer is prefetched. `display_label` comes fresh from the server, which is why the same object showed a stale hfid next to a fresh label. On the server, `get_one_by_hfid` finds the node by its new id and no longer by the old one, for both merge and rebase.

I carried your tests over in #10069 (merged to `release-1.11`). The two cross-relationship HFID tests now assert the server value through `get_one_by_hfid` and pass, so the `xfail` markers are gone. The two component tests are replaced by one unit test, and the cascade tests are kept. More detail is in the IFC-2937 comment.

Closing this in favor of #10069.

PR #10018, opened 2026-07-22
fix(git): correct merge()/rebase() return annotations to str | Literal[False]

saltas888:
# Why

`InfrahubRepository.merge()` and `rebase()` were annotated `-> bool`, but they actually return the new **commit-hash `str`** on a successful merge and `False` on a no-op. The `bool` annotation is a lie that defeats static type-checking for callers — a legitimate `isinstance(result, str)` check looks dead, and code that assumes a plain boolean goes unchecked.

**Goal:** make the declared return type match the real contract so callers and the type checker are no longer misled.

**Non-goals:** no behavior change; not removing the module's mypy `return-value` suppression (a separate follow-up); not touching `pull()`'s annotation or the broader git-module SOLID refactor.

Tracked in Jira: **[INBOX-20](https://opsmill.atlassian.net/browse/INBOX-20)** (from the "Infrahub Git refactoring 26Q2" SOLID analysis §4.2 / §9, High priority).

## What changed

Behavioral changes: **none — annotation-only.**

- `merge()` and `rebase()` return annotations: `-> bool` → `-> str | Literal[False]` (commit-hash string on success, `False` on no-op).
- Added `Literal` to the `typing` import in `backend/infrahub/git/repository.py`.
- Added a `housekeeping` changelog fragment.

What stayed the same: no DB schema/migration changes, API contract unchanged, no auth changes, no new dependencies, no CI changes, no generated files. Runtime behavior and control flow are unchanged — the existing `return False` / `return str(commit_after)` statements are untouched.

> The diff also includes a spec-kit design record under `dev/specs/inbox-20-merge-return-type/` (spec / plan / critique / tasks / report) produced by the workflow that generated this change. It is documentation only.

## How to review

- The entire behavioral change is **2 return annotations + 1 import** in `backend/infrahub/git/repository.py`. Everything else is docs/metadata.
- Callers: the sole production `merge()` caller (`backend/infrahub/git/tasks.py`) discards the result, and `rebase()` has no production callers — so no caller changes were required (mypy confirms).

## How to test

```bash
uv run ruff check . --exclude python_sdk
uv run mypy backend/infrahub/git/repository.py backend/infrahub/git/tasks.py
uv run pytest backend/tests/component/git/test_git_repository.py
```

Expected: ruff clean, mypy reports no issues, git component tests pass. (A pre-existing, unrelated `test_pull_branch_conflict` failure can appear locally on some git versions — it lives in `pull()`, not `merge`/`rebase`, and is unaffected by this change.)

[...]

PR #9953, opened 2026-07-17
docs(git-integration): multi-environment pattern and promotion guide

lancamat1:
## What

Adds two spokes under **Branches & Change Control → Git Integration** documenting how to run separate development / staging / production Infrahub instances from a single Git repository, and how to promote changes between them.

- `docs/docs/git-integration/multi-environment.mdx` — **concept**: the all-read-only, branch-per-environment model; what moves through Git (`.infrahub.yml` content); Git-vs-Infrahub terminology; the staged promotion model (import into an Infrahub branch, review via a proposed change); and reaching production safely.
- `docs/docs/git-integration/promote-between-environments.mdx` — **how-to**: register a read-only repository per instance, author a change, the staged promotion flow, and an optional/experimental CI automation note.
- `docs/sidebars.ts` — two entries under Git Integration.

## Key points of the documented model

- Every environment is a **read-only** consumer pinned to one long-lived Git branch via `ref`; no instance uses a read-write `Repository`, and nothing is pushed back to Git.
- Read-only repositories **never auto-import** — imports are always triggered manually (UI) or via API.
- Promotion is **staged**: import into an Infrahub branch and merge via a proposed change; importing straight to `main` is a caveated side-note, not recommended.
- The environment count is an **example** (dev/staging/prod) — 2 or 5 work identically.
- No clean rollback is claimed; safety comes from rehearsing on a pre-production instance restored from a current production backup.

## For reviewers

- **Placement** under Git Integration is provisional — flag if this should be a top-level section instead.
- The **"Automate the promotion"** section is intentionally marked optional/experimental. A fully automated PC-merge gate is blocked by open behavior: a proposed-change merge currently carries a read-only repository's branch-aware `ref`/`commit` onto the destination branch, which would repoint an instance's `main`. That behavior is under review (internal ticket; relates to IFC-1361) — the docs deliberately keep production promotion manual until it's resolved.
- **Product framing** (positioning / "who this is for") is expected to be iterated post-merge by product.
- `<repository-id>` and host placeholders are intentional; the reader substitutes their own.

## Verification

- Facts verified against backend source (mutation/type names, `ref` vs `default_branch`, `CoreValidator` `state`/`conclusion`, `CoreProposedChange.validations`, `.infrahub.yml` importable content).
- `npm run build` passes; all cross-links and anchors resolve; Vale/banned-qualifier scan clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
[...]

PR #10030, opened 2026-07-23
docs(guidelines): internal-doc writing style, size discipline, six harvest runs

saltas888:
## Why

Reviewers kept raising the same things, and each lesson lived in a review thread and died there. This PR consolidates six `/harvesting-review` passes over PRs reviewed 2026-07-16 → 2026-08-07 into the internal-doc layer, so a future author (human or agent) follows the convention the first time. It also introduces the discipline the harvests themselves turned out to need: a harvest that only adds makes the layer worse, so every addition here is paid for with compression or pruning.

Every lesson was reconstructed against the before/after code, verified against the current codebase, and checked against existing coverage before landing, per `.agents/skills/harvesting-review/`.

**Closed in favor of this PR** (their commits are cherry-picked here): [#10098](https://github.com/opsmill/infrahub/pull/10098), [#10145](https://github.com/opsmill/infrahub/pull/10145), [#10157](https://github.com/opsmill/infrahub/pull/10157).

Docs only — no code, no generated files, no schema.

## The conventions now in place

One row per lesson: the source PR whose review surfaced it, and where it now lives.

### Backend

| Convention | Lives in | Source |
|---|---|---|
| Ask whether the `try`/`except` is needed at all; `# noqa: BLE001` needs a named reason; inside a transaction a caught-and-returned failure **commits** (verified against `InfrahubDatabase.__aexit__`) | `dev/guidelines/backend/exceptions.md` (new, split from `python.md`) | #10002 |
| Don't write `T \| Sequence[T]` unions — take `Sequence[T]` and let callers wrap (a bare `str` satisfies `Sequence[str]` and gets shredded) | `dev/guidelines/backend/python.md` | #10079 |
| `tasks/*.py` backend imports stay function-local — and the exception covers a thin wrapper, not logic hosted in `tasks/` | `dev/guidelines/backend/python.md` | #9814, #10095 |
| A path-exclusion check must not use a bare `startswith` | `dev/guidelines/backend/asgi-middleware.md` (new, split from `python.md`) | #9879 |
| `retry_db_transaction` replays the whole wrapped scope (on `TransientError` or `EntityNotFound`): wrap only rollback-able scopes, skip under `db.is_transaction`, mind nested retried callers | `dev/knowledge/backend/mutations.md` | #10121 |
| Pass `limit`/`offset` to the base `Query`; `insert_limit = False` without an own LIMIT never finishes paging; auto-pagination needs a total-order `ORDER BY`; read `collect()` columns with `get_as_list_of_type` | `dev/knowledge/backend/query-pattern.md` | #10106, #10132 |
| `n.kind` is concrete-only — generics exist as labels; type concrete-only query inputs as `list[NodeSchema]` | `dev/knowledge/backend/database-schema.md` | #9805 |
| Changelog models mask secrets only at construction — post-construction assignment leaks them | `dev/knowledge/backend/events.md` | #10105 |
| `InfrahubBatch` is concurrent, not ordered; the best-effort contract also covers a single inline side effect; `add_tags()` mid-run drops earlier tags (cross-linked where the guide teaches the pattern) | `dev/knowledge/backend/async-tasks.md`, `dev/guides/backend/creating-async-tasks.md` | #10113, #10034 |
| Fix data bugs in the migration layer, not lazily in `save()`; per-item try/collect into `MigrationResult.errors`; batch by the memory-bounding unit | `dev/guides/backend/creating-migrations.md` (now routed from `backend/AGENTS.md` — it was loaded by nothing) | #10105, #10132 |
| A new `Settings` field must reach the hand-maintained `development/` compose anchor, not just the generated one | `dev/guidelines/backend/checklist.md` **Configuration** | #10122 |
| New REST endpoints join **Ask First** — extend the existing GraphQL query instead | `backend/AGENTS.md` | #8594 |

### Backend testing

| Convention | Lives in | Source |
|---|---|---|
| Verify wiring by parsing source (`ast`/`inspect`), never by instrumenting production code | `dev/guidelines/backend/testing.md`, `.agents/rules/testing-python.md` | #10121 |
| Don't add a return value just so a test can assert on it; assert an intermediate signal when a result converges via two code paths | `dev/guidelines/backend/testing.md` | #10031, #10038 |
| Poll, don't sleep, for async outcomes; make removal assertions branch-attributable; extract a pure function before claiming the cheap test tier is unreachable | `dev/guidelines/backend/testing.md` | #10133, #10132, #10137 |
| Parametrize with dataclasses — the dict-keys-plus-in-test-lookup hybrid is now named and forbidden | `dev/guidelines/backend/testing.md` | #10095 |
| A benchmark must exercise the exact input shape a conditional optimization gates on | `dev/knowledge/backend/testing.md` | #10062 |

### Frontend

| Convention | Lives in | Source |
|---|---|---|
| Reset shared Jotai atoms, mocks (`vi.resetAllMocks`), and URL state between tests in one file | `dev/guides/frontend/writing-component-tests.md` | #10014 |
| A test must fail when the behavior breaks: assert the value, not the static text beside it; check the component actually reads the flag the setup varies | `dev/guides/frontend/writing-component-tests.md` | #10111, #10128, #10129 |
| A lazily-memoized module-level singleton needs a test reset hook | `dev/guides/frontend/writing-unit-tests.md` | #10059 |
| A `Map` lookup seeded from the array you index it with is still a type lie under `!` | `dev/guidelines/frontend/typescript.md` | #10014 |
| Don't mirror owned state (URL, forms, server) into a Jotai atom — generalized from the old forms-only rule | `dev/guidelines/frontend/page-architecture.md` | #10111 |

### Component design and code style (`.agents/rules/` — edits to existing rules, no new files)

| Convention | Lives in | Source |
|---|---|---|
| Build the component graph at the entry point in one pass; a wiring mismatch raises at construction, not as a runtime fallback; `initialize()`/`reset()` clears every derived cache | `.agents/rules/backend-component-design.md` | #10055, #10053/#10019/#10082 |
| Spec vocabulary is as unreadable as a spec ID in a test name or docstring | `.agents/rules/code-doc-style.md` | #10096 |

### Process and documentation

| Convention | Lives in | Source |
|---|---|---|
| PR ceremony scales with the change: no changelog fragment or `dev/specs/` scaffolding on a trivial fix; a lint-suppression PR ships the suppression only and files the bug separately | `dev/guidelines/git-workflow.md`, `dev/guidelines/repository-organization.md` | #10018, #10079 |
| Internal-doc writing style: rule first, plain words, no provenance, no padding, a size budget, pay-for-it | `dev/guidelines/documentation.md` *For Internal Docs* | the harvests themselves |
| No `file.py:line` citations in internal docs; measurements go in as relative comparisons | `dev/guidelines/documentation.md` | #10038, #10108 |
| Verify identifier/metric/config-default claims against the diff before writing them into fragments, PR text, or ADRs | root `AGENTS.md` | #9845 et al. |
| Changelog fragments: GitHub issue number or `+`, never a tracker ID; the `+` orphan-prefix failure mode; `housekeeping` is not a catch-all — no user-facing effect means no fragment | `.agents/skills/creating-changelog-entries/SKILL.md`, `docs/docs/development/changelog.mdx` | #9845, #10146/#10147, #10018 |
| `invoke lint` runs `ruff check --diff`, blind to unfixable violations — `/pre-ci` carries the CI-parity `ruff check` step | `.agents/commands/pre-ci.md`, root `AGENTS.md` | #10146/#10147 |
| Don't escalate from a stale "commits behind" count | `.agents/skills/monitoring-pull-requests/SKILL.md` | #10146 |
| Grafana dashboards: defined datasource variables only, sweep every link surface, regenerate the standalone compose from `development/` | `development/grafana/AGENTS.md` (new) | #10153 |
| Excalidraw exports: opaque white background | `docs/AGENTS.md` | #9953 |

### Pruned in the same PR

Stale `PR #NNNN`/`Migrated in #NNNN` citations across 5 frontend docs (17 hits) and `templates.md` (4 hits, plus the renamed method the rot was hiding: `_apply_template()` → `handle_object_template()`); `typescript.md`'s `useMemo` dependency-array teaching, which contradicted `react.md`'s React-Compiler guidance — a reviewer had to ask which was right; `schema-definitions.md`'s pointer to an issue closed 2026-07-20; `query-pattern.md`'s five near-identical walkthroughs collapsed into one example plus an accessor table (520 → 459).

## The harvesting-review skill, final state (0.8.0)

**Refine, don't accrete** leads the skill and outranks its other rules: measure the target file against its size range first, compress or split a file at its limit (repointing every inbound reference, including `.agents/skills`/`.agents/commands` routes), cut what the new rule supersedes, and report added/removed line counts — zero deletions is a finding about the harvest. Every run sweeps the destination layer (including `.agents/rules`) for citation rot, fixed defect-snapshots, and superseded narrow rules, and **fixes every hit in the same PR** — a punch list is not pruning. A lesson whose root cause is a fragile pattern becomes a rule steering away from the pattern, not a survival guide for it. A lesson owned by an existing skill (`creating-changelog-entries`, `pre-ci`, the org plugin's `pruning-residues`) is routed into that skill instead of parallel `dev/` prose.

## Also decided along the way

- **Base branch — `develop`, deliberately** (cubic suggested `stable`, discussed with @polmichel): `git-workflow.md` has no docs-target-stable rule, the lessons were verified against `develop`'s code, and several describe modules that don't exist on `stable`. If internal-docs-to-stable should be policy, that's a `git-workflow.md` rule to add first.
- **Dropped, not restored**: four develop-only lessons from a brief `stable` retarget (diff-tree node sets #10014, merge-recompute invariant #9845, `get_query_arrows()` #10053-cluster, `ComputedAttribute` drift guard #9814) — re-harvestable later.
- **Read with nothing durable**: #10051, #10090, #10104 (bot findings retracted), #10123, #10125, #10136, #10114 (already covered by the verify-claims rule), #10141, #10154, #10149, #10144, #10099, #10124, #10080, #10089, #10081, #7569, and a set of bare approvals.
- **Confirmed-real code issues flagged as follow-ups, not codified**: `migrations/query/attribute_add.py` missing the documented `r.status ASC` tiebreaker; develop's dashboards carrying 42 undefined-`$datasource` references (#10153 fixed them only on `release-1.11`); `changelog/models.py`'s `set_value_previous` bypassing the masking validator; #10133's broker-durability lessons parked until the Redis driver merges; the `stable`/`develop` ruff `BLE` divergence.
- **All review feedback on this PR addressed in-branch** (@polmichel's six comments, cubic's two rounds — every finding verified against code before fixing): the `str`/`Sequence` doc section became the rule against the pattern, the extract skill's error-handling route repointed to `exceptions.md`, `backend/AGENTS.md`'s list made contiguous, the retry-trigger wording completed, the test examples made reproducible and actually-isolating, `vi.resetAllMocks` over `clearAllMocks`, and the Grafana regeneration command corrected.

## Context budget

**+702/−338 over `develop` (40 files, after merging `develop` back in)**, and the shape matters more than the total: no new file in `.agents/rules/` (the only layer injected into every agent turn — three existing rules edited), the two `python.md` splits load only on their trigger, and every harvest paid with deletions.

| File | Lines | Range |
|---|---|---|
| `dev/knowledge/backend/query-pattern.md` | 490 | 200–400 — the `## Internals` split is the flagged next change |
| `dev/guidelines/backend/python.md` | 414 | 100–400 — the ASGI split had it at 393; `develop`-side additions re-grew it; the next split is its own change |
| `dev/guides/frontend/writing-component-tests.md` | 398 | 200–400 |
| `dev/guidelines/backend/testing.md` | 393 | 100–400 |

## How to test

Docs only. `markdown-lint` (docs/docs), `validate-dev-guideline-links` (lychee over `dev/**`, `backend/**`), and `validate-documentation-style` (Vale over `docs/`) cover these paths in CI; relative links, section anchors, and code-fence balance checked locally across all 37 changed markdown files.


dev/guidelines/changelog.md
```diff
@@ -50,6 +50,14 @@ uv run towncrier create -c "Added breadcrumb navigation for hierarchical schemas
 uv run towncrier create -c "Updated dependencies to latest versions" +deps-update.housekeeping.md
 ```
 
+## When to Skip
+
+Skip the fragment entirely when the change has **no user-facing effect** — an internal
```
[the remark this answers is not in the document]
polmichel: I do agree

backend/AGENTS.md
```diff
@@ -38,6 +38,9 @@ backend Python, including:
 - Async-first patterns
 - Pydantic models
 - Docstring conventions
+
+Exception handling lives in `dev/guidelines/backend/exceptions.md` — load it when writing a
```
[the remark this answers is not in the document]
polmichel: I do agree with Cubic here @saltas888

dev/guidelines/backend/python.md
```diff
@@ -267,6 +267,22 @@ async def set(self, key: str, value: str, expires: KVTTL | int | None = None) ->
```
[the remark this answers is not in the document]
polmichel: Do these skills need to reference this exact file? It creates this kind of issue

dev/guidelines/backend/python.md
```diff
@@ -296,6 +296,22 @@ async def set(self, key: str, value: str, expires: KVTTL | int | None = None) ->
...
+# ❌ Bad - a bare id like "abc-123" satisfies Sequence and gets shredded into one item per character
+if not isinstance(data, Sequence):
+    data = [data]
+
+# ✅ Good - str is excluded first, so a single id stays a single item
+if isinstance(data, str) or not isinstance(data, Sequence):
+    data = [data]
+```
+
+Widening a parameter from `list[T]` to `Sequence[T]` means widening this runtime check in step: an annotation that newly accepts `str` while the `isinstance` check still assumes a `list` misroutes every bare string. Test both the bare `str` and the newly-accepted non-`list` sequence (a `tuple`).
```
polmichel: This is a personal take but I am not a huge fan of this pattern. I think we should get rid of them instead of patching potential typing bugs with a dedicated section inside our documentations.
In my opinion, let's use a `Sequence` or a `list` directly.

dev/guidelines/documentation.md
```diff
@@ -40,6 +40,26 @@ Infrahub documentation follows the [Diataxis framework](https://diataxis.fr/), s
...
+  path". Name a symbol only when the reader has to go find it
+- **Show instead of qualifying**: a ✅/❌ pair replaces a paragraph of caveats
+- **No provenance**: where a rule came from — a review thread, a ticket, a PR, a spec — belongs in the
+  commit message, not the doc (see `.agents/rules/code-doc-style.md`)
+- **No padding**: cut "note that", "it's important to", "keep in mind", bolded whole sentences, and any
+  sentence that restates the one before it
+- **Say it once**: if the rule is already written elsewhere, link to it instead of restating it
+- **Budget**: a new rule is a few lines plus one example. If it needs more than that, it's an
+  explanation for `dev/knowledge/`, not a guideline
+- **Pay for it**: cut the prose the new rule supersedes, and check the file against its size range
```
polmichel, line 63: Should we use the `pruning-residues` skill and iterate over it if there is any enhancement to do on it?

AGENTS.md
```diff
@@ -99,6 +99,11 @@ cd frontend/app && pnpm biome:fix     # Format/lint frontend
 uv run invoke docs.lint               # Lint documentation
 ```
 
+`invoke lint`/`invoke backend.lint` run `ruff check --diff`, which reports only auto-fixable issues
+and exits 0 on an unfixable violation (e.g. an unsuppressed `BLE001`) that CI's plain `ruff check`
+still fails on. Run `uv run ruff check <path>` directly to see the full picture before trusting a
+clean local run.
+
```
polmichel, line 104: Just a comment in case this sounds useful: a `/pre-ci` skill exists with a `--fast` option that trigger a bunch of CI checks locally. These checks include these lint steps

dev/guidelines/backend/python.md
saltas888: Agreed, replaced the section with a rule against the pattern itself, take Sequence[T] and let callers wrap. Kept one line on the str carve-out for existing code.

dev/guidelines/documentation.md
saltas888: The harvesting skill now routes lessons into the skill that owns the workflow and runs the diff through pruning-residues when available. It's not in our skills-lock yet, will add it as a follow-up.

saltas888:
@polmichel 
> Do these skills need to reference this exact file? It creates this kind of issue

Repointed them to exceptions.md, and the harvesting skill now requires repointing all inbound references on a split, including .agents, so this shouldn't recur.

AGENTS.md
saltas888: right, done

dev/guidelines/changelog.md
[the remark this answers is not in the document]
saltas888: I know but there's a big deviation on the patterns that are described in develop and stable, therefore in stable we will have to describe code that actually doesn't exist. I'd need to check if we can split some parts

.agents/skills/creating-changelog-entries/SKILL.md
```diff
@@ -88,6 +88,7 @@ uv run towncrier create -c "Migrated the frontend build to pnpm workspaces" +pnp
   reference and either skips the fragment or renders a bogus issue link instead of an orphan entry.
 - **Hand-writing the fragment file.** Use `towncrier create` so the name and location are correct.
 - **Placing it in a sub-package directory** (e.g. `backend/changelog/`) instead of the configured fragments directory.
+- **Naming the fragment after an internal tracker ID** (Jira, Linear, etc.) instead of the GitHub issue number or `+`. Towncrier's `issue_format` turns the fragment name straight into a GitHub issue URL — an internal ticket ID renders a broken link in the shipped changelog.
```
polmichel: Just to be sure: do we agree that the following naming is correct since the `+` is the first character? `+ifc-2546-lorem-ipsum.md`.

.agents/skills/harvesting-review/SKILL.md
```diff
@@ -228,23 +241,76 @@ must fire while coding.** Confirm the target file exists (`ls`/grep it, match si
...
 at all: changelog conventions, for one, moved out to the `creating-changelog-entries` skill.
 
-### 5. Report
+### 5. Sweep for rot (prune before you add)
+
+Every run of this skill only adds. Nothing else revisits what a previous run wrote, so the layer
+grows monotonically — a doc entry that was true and useful the week it landed can quietly become
+stale, redundant, or wrong, and stays in place forever unless a run like this one checks it. Do this
+sweep every time, not as an occasional separate cleanup — it is cheap (a handful of greps, not a
+re-read of every doc) and it is what keeps "harvested" from becoming a synonym for "bloated."
```
polmichel, line 261: Wdyt about explicitly referring to an existing skill? Like "/pruning-residues"?
It would be interesting to see whether this skill uses it without explicit statement, or not.

polmichel: I saw your following commit about that

dev/knowledge/backend/testing.md
```diff
@@ -113,6 +113,14 @@ async def test_query_performance(aio_benchmark, db):
...
 ```
 
+**The benchmark's input must actually exercise what it claims to measure.** When a performance
+change is conditional on a specific input shape (a particular field selection, a branch of an
+`if`, a fast path), add or update a benchmark whose input takes that path. A benchmark that
+takes a different path reports "no change" and gives a false regression guard — CodSpeed stays
+green while the optimization silently regresses later. Before trusting a benchmark result on a
+perf PR, check that the query/input it drives actually hits the changed code, not just a
+similarly-named neighbor.
+
```
polmichel, line 121: I'm not sure about this part of the text:

1. It ties pytest-benchmark to CodSpeed, which is the tool we use to create and interpret the results of pytest-benchmark.
2. This text says "Do not trust CodSpeed results". In that case, I think we should either write better tests or drop CodSpeed.

.agents/rules/backend-component-design.md
```diff
@@ -10,18 +10,18 @@ Applies when creating a new backend component or making significant changes to a
...
 
 ## Build components near the application entry point
 
 Construct components as close to the application entry point as possible. Use a builder class or factory function when wiring is non-trivial, and inject each sub-component rather than constructing it inside a parent component's `__init__`.
 
+**The whole graph is built there, in one pass, before the work starts.** A component that builds a collaborator part-way through a run — when the flow reaches the step that needs it — has the same problem as one that builds it in `__init__`, and is harder to follow because the wiring is spread across the execution. Pass the collaborator in, and pass per-run values to the entry method instead of holding them as constructor state.
+
+**Invalid wiring fails while you build the graph.** Validate the combination at construction and raise; never let a missing or mismatched collaborator surface later as a silently skipped step or a degraded fallback. Where a generic ties two sides together, parameterize both so the mismatch is a type error rather than a runtime discovery — a `Handler[Any]` paired with `Output[Any]` type-checks against anything and defers the failure to the run.
```
polmichel, line 25: Nice

.agents/skills/harvesting-review/SKILL.md
```diff
@@ -192,7 +196,11 @@ A reviewer flagged this, so the verdicts are not a pass/fail of the docs — the
   someone fixes it, and nothing revisits it. Phrase it as a forward-looking convention that stays true
   regardless of whether this exact instance ever gets fixed ("when adding a value generated this way,
   derive the fields rather than hand-listing them"), or drop it and suggest filing a GitHub issue
-  instead of writing it into `dev/knowledge`/`dev/guidelines`.
+  instead of writing it into `dev/knowledge`/`dev/guidelines`. The same test applies when the root
+  cause is a fragile *pattern* rather than a defect: if the honest fix is to stop writing the pattern
+  (a `str | Sequence[str]` union and its `isinstance` dispatch, for one), the lesson is the one-line
```
polmichel: Not sure about this specific example in the text, it feels over-customized for this use case.

.agents/skills/creating-changelog-entries/SKILL.md
saltas888: Almost — the + does suppress the issue link so it's safe, but it's missing the type segment (needs `.changed.md` etc.) or towncrier won't pick it up. The ticket ID never renders in the output either way, so the slug is better spent on a description. Fixed in b7d06e06d. Also found two fragments on develop missing the + (IFC-2747.fixed.md, infp-677.changed.md) that will ship broken links — queued a fix for those.

.agents/skills/harvesting-review/SKILL.md
saltas888: Already explicit since ae7986fbd (pushed before this comment) — named in both the routing rule and the apply step. Sounds like your follow-up found the same commit.

dev/knowledge/backend/testing.md
saltas888: Fair, reworded in b7d06e06d. The point isn't that CodSpeed is untrustworthy — it's that a benchmark hitting a different path measures the neighbor, so "no change" is an honest report of the wrong thing. Dropped the tool from that sentence.

.agents/skills/harvesting-review/SKILL.md
saltas888: Agreed, dropped in b7d06e06d — that was the str|Sequence example naming this PR's own case inside a general rule, exactly the over-customization the skill tells authors to avoid elsewhere.

polmichel (approved):

saltas888:
Superseded by #10421, which carries this content on a branch cut from `stable`.

Closing rather than retargeting: this branch has `develop` merged into it three times, so flipping the base turns a 35-file docs change into 157 commits / 870 files — a develop-to-stable merge. #10270 hit the same wall in August and solved it the same way, by carving the content out onto a fresh `stable` branch.

On the base question this PR left open — `stable` won the argument in the end. Internal docs on `stable` reach `develop` through the routine sync (#10270's content is byte-identical on both branches today), while docs landed on `develop` do not flow back. The objection recorded here, that some lessons describe modules absent from `stable`, no longer holds: the four genuinely develop-only lessons were dropped from this branch in August and never restored, and every symbol the remaining rules name resolves on `stable` (`retry_db_transaction`, `get_as_list_of_type`, `insert_limit`, `handle_object_template`, `handle_template_relationships`, `set_value_previous`, `collect_graph_violations`).

The port needed exactly one conflict resolved, in `creating-migrations.md`, and it was a context mismatch rather than a content clash — the hunk's context included a develop-only paragraph. All links and anchors resolve, and there is no duplication against what #10270 and #10259 already landed.

Review feedback from this PR carried over unchanged; nothing was dropped in the port.

PR #10031, opened 2026-07-24
merge failure recovery improvements

ajtmccarty:
# Why

Live failure-mode testing of the merge-failure-recovery feature surfaced two ways a crashed branch merge leaves the system in a degraded state that `infrahub recover merge` did not clean up:

- **stale destination schema metadata.** When a merge crashes after the schema graph is written, recovery rolls the graph back but left the destination branch's persisted `schema_hash` / `schema_changed_at` at their failed-merge values. Those are plain `Branch`-vertex properties written in place mid-merge, so the versioned range rollback cannot revert them. The persisted metadata then disagrees with the rolled-back graph, so every worker keeps detecting a "new" schema hash and churning through reconciles.
- **leaked distributed locks.** A hard-killed merge leaks no-TTL locks (`lock.global.graph`, `lock.global.schema`, and the per-branch `lock.diff-update.*` locks). Recovery cleared only the merge lock, so the leaked locks silently blocked the next merge until the 15-minute deadlock reaper cleared them — with no log line before the blocking acquire to diagnose the hang.

**Goal:** make `infrahub recover merge` leave the destination branch's schema metadata and the distributed locks in the same state a healthy merge would, and make a leaked-lock hang diagnosable from logs.

**Non-goals:** no change to the merge happy path, the rollback range logic, or the failure-detection/`MERGE_FAILED` flagging.

## What changed

**Behavioral changes**

- Recovery now realigns the destination branch's `schema_hash` (recomputed from the rolled-back graph) and `schema_changed_at` (restored from the value captured at merge start), so a recovered merge no longer triggers spurious "New hash detected" reconcile churn.
- Recovery now clears the diff-update / global-graph / global-schema locks a dead merge leaked, deleting only those whose holder is no longer an active worker — a lock a live flow still holds is left in place — so the next merge is not silently blocked until the deadlock reaper runs.
- The merge path now logs before the blocking merge-lock and global-graph-lock acquisitions, so a leaked-lock hang is diagnosable from the logs.

**Implementation notes**

- New `Branch.pre_merge_destination_schema_changed_at` (optional, default `None`) is stamped onto the source branch at merge start (extracted into `BranchMergeOrchestrator._record_merge_start`) and is the source recovery restores from.
- New `StaleLockCleaner` component (liveness-guarded deletes) and the shared active-worker lookup consolidated onto `InfrahubComponent.list_active_worker_ids`.
- The recovery schema load is isolated behind a single override point (`MergeFailureRecoverer._load_destination_schema`) so tests can substitute an in-memory schema and skip the slow database round-trip.

## Impact & rollout

- **Backward compatibility:** new optional `Branch` field, no migration; existing in-flight/older `MERGE_FAILED` records without the field are handled (schema hash still recomputed, `schema_changed_at` left untouched).
- **Performance:** no runtime impact; the schema-load override is a test-only seam.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — n/a, the merge-failure-recovery feature is still unreleased

---
## Summary by cubic
Improves merge failure recovery by restoring destination schema metadata and clearing stale distributed locks so subsequent merges and workers don’t get stuck or recompute endlessly. Adds merge-start bookkeeping and better lock diagnostics for easier debugging.

- **New Features**
  - Capture the destination branch’s `schema_changed_at` at merge start on the source branch (new `Branch` field: `pre_merge_destination_schema_changed_at`).
  - Clear stale diff-update and global graph/schema locks during recovery using a new `StaleLockCleaner` (deletes only locks held by dead workers).
  - Introduce `WorkerLiveness` protocol and expose global `InfrahubComponent.list_active_worker_ids`; log before/after acquiring global merge/graph locks; CLI recovery wires in `SchemaManager`, `StaleLockCleaner`, and `DiffLocker`.

- **Bug Fixes**
  - Recovery realigns the destination schema metadata with the rolled-back graph: recomputes schema hash and restores `schema_changed_at` from the captured pre-merge value, including when it was `None`.
  - Recovery sequence is idempotent and failure-safe: rollback → restore schema metadata → reset proposed change → release merge lock → clear stale locks → reset branch → lift write protection.

<sup>Written for commit 56b44cad9bf24e27ddae933252150ca4d1313e91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/locks/cleaner.py
```diff
@@ -0,0 +1,48 @@
...
+        A key with no token, an unparseable token, or a token naming a live worker is left in place.
+        """
+        active_worker_ids = await self._active_worker_ids()
+        deleted: list[str] = []
+        for key in keys:
+            token = await self.cache.get(key=key)
+            worker_id = get_worker_id_from_lock_token(token)
+            if worker_id is None or worker_id in active_worker_ids:
+                continue
+            await self.cache.delete(key=key)
```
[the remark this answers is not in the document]
ajtmccarty: technically true, but incredibly unlikely. the gap in here is 5 lines of Python

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -184,3 +182,16 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
...
+    async def _record_merge_start(self, *, merge_at: Timestamp, user_id: str) -> None:
+        """Persist the merge-start markers a recovery depends on.
+
+        ``merge_started_at`` gives an out-of-process recovery the point to roll the graph back from, and
+        the destination's pre-merge ``schema_changed_at`` is captured alongside so recovery can restore
+        it after the rollback.
+        """
+        self.source_branch.status = BranchStatus.MERGING
+        self.source_branch.merge_started_at = merge_at.to_string()
+        self.source_branch.pre_merge_destination_schema_changed_at = self.destination_branch.schema_changed_at
```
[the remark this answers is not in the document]
ajtmccarty: this is resolved

backend/tests/component/merge_recompute_coalescing/test_merge_submits_coalesced.py
```diff
@@ -63,6 +63,10 @@ async def test_merge_submits_one_coalesced_recompute_per_target(
         account=AccountSession(account_id=str(uuid4()), auth_type=AuthType.NONE),
     )
 
+    # Captured before the merge so it can be compared against what the merge stamps on the source branch.
+    pre_merge_destination_changed_at = default_branch.schema_changed_at
+    assert pre_merge_destination_changed_at is not None
```
polmichel, line 68: Is this assertion useful?

backend/tests/component/core/merge/conftest.py
```diff
@@ -36,7 +40,18 @@ def find_logged_event(caplog: pytest.LogCaptureFixture, *, event: str, branch: s
...
-class FailAtBranchResetRecoverer(MergeFailureRecoverer):
+class InMemorySchemaRecoverer(MergeFailureRecoverer):
+    """Recoverer that recomputes the schema hash from the in-memory registry instead of the database.
+
+    Loading the schema from the database is a slow round-trip; the recovery logic under test does not
+    depend on that round-trip, so the schema is read from the registry to keep these tests fast.
+    """
+
+    async def _load_destination_schema(self, branch: Branch) -> SchemaBranch:
+        return registry.schema.get_schema_branch(name=branch.name)
```
polmichel, line 51: Nice one

backend/infrahub/locks/cleaner.py
```diff
@@ -0,0 +1,47 @@
...
+    reaper this applies no age threshold: recovery already proves the merge dead. The per-key
+    liveness check is what protects a lock legitimately held by a live flow — a running branch-diff
+    computation or a schema load on another branch — so those are left untouched.
+    """
+
+    def __init__(self, cache: InfrahubCache, component: InfrahubComponent) -> None:
+        self.cache = cache
+        self.component = component
+
+    async def clear_if_holder_dead(self, keys: list[str]) -> list[str]:
```
polmichel: This is an optional comment, which may be subjective:
I assume the return value is there for testing purposes, but in my opinion a method should either act or return something and not both. A method that does both is usually harder to maintain, and bending that rule just for testing is probably not a good enough reason.

backend/tests/unit/locks/test_cleaner.py
```diff
@@ -0,0 +1,73 @@
...
+MISSING_KEY = f"{LOCK_PREFIX}.diff-update.main__feature__incremental"
+
+
+def _lock_token(worker_id: str) -> str:
+    return f"{Timestamp().to_string()}::{worker_id}"
+
+
+async def _build_cleaner(cache: MemoryCache) -> StaleLockCleaner:
+    # list_active_worker_ids reads only the cache heartbeats, so a cache-backed component needs no database.
+    db = cast("InfrahubDatabase", None)
```
polmichel: I'm not sure my idea is better, but since the only reason for injecting `component` into `StaleLockCleaner` is to call `list_active_worker_ids`, would a dedicated `Protocol` be a better solution, making the object easier to fake in tests?

backend/tests/unit/locks/test_cleaner.py
```diff
@@ -0,0 +1,73 @@
...
+    # MISSING_KEY is intentionally never set.
+
+    cleaner = await _build_cleaner(cache)
+
+    deleted = await cleaner.clear_if_holder_dead(keys=[DEAD_KEY, LIVE_KEY, TOKENLESS_KEY, MISSING_KEY])
+
+    # Only the dead-worker lock is dropped; the live-worker lock and the unparseable/missing keys stay.
+    assert deleted == [DEAD_KEY]
+    assert await cache.get(DEAD_KEY) is None
+    assert await cache.get(LIVE_KEY) == live_token
```
polmichel: If my comment above about a method not doing both is valid, then the second and third assertions should be enough here

polmichel (approved):
LGTM! Optional comments posted

backend/tests/component/merge_recompute_coalescing/test_merge_submits_coalesced.py
ajtmccarty: it's mostly to ensure that the test is running correctly. so that it is easier to troubleshoot if this is not set as expected. if just the new assertion on line 81 failed, that would look like a functional problem, but if `default_branch.schema_changed_at` is never set, then it is more a test configuration issue

backend/infrahub/locks/cleaner.py
ajtmccarty: I think you are right. I've removed the return and updated the test to actually check the cache for existing locks

backend/tests/unit/locks/test_cleaner.py
ajtmccarty: good idea. I've added a protocol to cover this and this test is much clearer now

PR #10033, opened 2026-07-24
feat(regeneration): target artifacts from an after-merge generator's output

polmichel:
## Summary

Refines the post-merge generator → artifact cascade: instead of regenerating **every** artifact once
a generator runs, regenerate only the artifacts its output affects.

## Change

A generator dispatched by the follow-up runs after the merge diff was captured, so its writes are
absent from that diff. Once every generator has run, this diffs the target branch over the window
they ran in and selects only the artifacts that read the changed data (reusing the artifact half of
the merge selector). Full regeneration stays the fallback if that output can't be captured or
selected, so no artifact is left stale. Direct and proposed-change merges take the same path.

## How tested

| Level | Coverage |
|-------|----------|
| Unit | targeted dispatch, generator-before-capture sequencing, blanket fallback |
| Component | real selector + graph, stubbed generator output → only the affected artifact |
| Integration | real follow-up flow, empty capture → merge-diff selection stands |
| Backend unit suite | passes |

### Live validation (full battery, demo stack, `SELECTIVE_EXECUTION_AFTER_MERGE=true`)

Driven over the API against a live demo stack, asserting rendered artifact content (not workflow calls):

| # | Scenario | Verdict | Observation |
|---|----------|---------|-------------|
| 1 | Single-member narrowing | ✅ | diff=1 → rendered 1/3, others untouched |
| 2 | Multi-member | ✅ | diff=2 → rendered 2/3, third untouched |
| 4 | Generator → artifact cascade | ✅ | after-merge generator writes `description`, artifact re-renders |
| 5 | Wide signal (query changed) | ✅ | `regenerate_all_members=True`, rendered 4/4 |
| 6 | Arista `mtu` (demo-edge) | ✅ | startup-config updated to new mtu; oc-interfaces correctly excluded (does not read `mtu`) |
| 9 | Diverged main (main gains a member) | ✅ | branch member refreshed, no fallback |
| — | Empty diff | ✅ | 0 generator / 0 artifact — no wasted work |
| 7 | Composition cascade (artifact inlines another) | 🔴 gap | composite permanently inlines the **stale** section |

### A/B performance (demo stack, 250-member group, people-only)

Dataset: 250 `TestingPerson` in a single `people` group; each member feeds 1 generator
(`person_description_generator`, writes `description`) and 3 artifacts — `Section config`,
`Composite config` (inlines the section), and the generator-fed `Person cascade` — so a full
regeneration touches ~250 generators + ~750 artifacts.

Rename one member and merge; wall-clock from merge to regeneration drain, median of 3 reps:

| Mode | Median | Regen flows dispatched |
|------|--------|------------------------|
| OFF (full regeneration) | ~302 s | ~1001 (250 generators + 750 artifacts) |
| ON (selective) | ~27 s | ~6 (the single impacted member) |

~11× faster wall-clock — a conservative floor, since the ON time is dominated by the drain-detection
window; the dispatched regeneration work itself drops ~167× (1001 → 6 flows).

### Real-git validation (served bare remote in the demo stack)

Pushed a change to a served bare repository, let the sync import it, then merged — asserting on rendered
artifact content. Each push advanced the repository commit and the definition fingerprint, and the merge
ran the chain against the new code; no `CommitNotFoundError`, no fallback.

| Change in the git repo | Result after push → sync → merge |
|------------------------|----------------------------------|
| Generator Python code (`v1-`→`v2-`) | generator runs the new code; artifact shows `v2-…` |
| Jinja2 artifact template (`Code:`→`CODE2:`) | artifact renders with the new template |
| New artifact definition in `.infrahub.yml` | definition imported and selected on the next merge |

Exercises the repository commit-resolution path end-to-end (the `CommitNotFoundError` in the earlier A/B
report was a test-setup artifact — a non-bare remote committed in place — not a product issue).

### Closure-completeness gating (served bare remote in the demo stack)

Pushed successive commits to a served bare repo, let the sync import each, then read the definition's
`dependencies_complete` / `fingerprint` (and, for the escalation cases, what a merge actually
regenerated). Each row is also covered by the closure-builder unit tests in
`backend/tests/unit/git/closure_builder/`.

| Change in the git repo | Result after sync/merge |
|------------------------|-------------------------|
| Jinja2 dynamic `{% include var %}` | transform `dependencies_complete=false`, fingerprint moved |
| `watch: {files: [...]}` declared | closure trusted again (`dependencies_complete=true`), watched directory pulled into the dependency list |
| Broken transform (missing template) beside a healthy one | broken → `false`, sibling stays `true`, repository import not aborted |
| Static `{% include "templates/partial.j2" %}` | included file resolved into the dependency closure, `true` |
| Read-only repository | closure and fingerprint computed identically to a standard repository |
| Definition with no computed fingerprint (pre-feature state) | its whole repository is force-regenerated on the next merge — even one that changes none of its data — and self-heals once the repository is re-imported |
| One repository's code changed | only that definition's fingerprint moves; other repositories' fingerprints stay put |

## How to review

1. `backend/infrahub/core/merge/selective_regen/generator_diff_capturer.py`
2. `select_artifacts` in `selective_regen/orchestrator.py`
3. `regeneration_dispatcher.py` — `_dispatch_plan` + `_artifacts_from_generator_output`
4. `_build_post_merge_regeneration_dispatcher` in `core/branch/tasks.py`
5. Tests — unit → component → integration
6. (Optional) run live: add a group member, merge, watch for `Targeted N artifact definition(s)`

https://claude.ai/code/session_012crFmdtPR3Qm2ZCDK8e2Wk

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10033?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/selective_regen/generator_diff_capturer.py
```diff
@@ -0,0 +1,99 @@
...
+        node_ids = await self._output_node_ids(generator_definition_names=generator_definition_names)
+
+        diff = await self.diff_coordinator.create_or_update_arbitrary_timeframe_diff(
+            base_branch=self.branch,
+            diff_branch=self.branch,
+            from_time=since,
+            to_time=Timestamp(),
+            name=str(uuid4()),
+        )
+        filters = EnrichedDiffQueryFilters(ids=sorted(node_ids)) if node_ids else None
```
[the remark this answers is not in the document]
polmichel: Fixed in 8c37bdf1a. `GeneratorTrackingGroupDiffCapturer.capture` now widens to an unscoped window read whenever *any* requested definition has no resolved tracking group, instead of narrowing on the aggregate ids (which would drop the unresolved generator's output). Covered by `test_capture_widens_when_any_definition_lacks_a_tracking_group`.

backend/infrahub/core/merge/regeneration_dispatcher.py
```diff
@@ -116,36 +121,82 @@ async def _dispatch_plan(
...
+            target_branch=target_branch,
+            since=cascade_started_at,
+            generator_definition_names=[run.generator_definition.definition_name for run in plan.generator_runs],
+        )
+        if targeted is None:
+            # Every artifact was already regenerated wholesale, which covers the merge-diff selection too.
+            return
+        # Dispatched only after the capture, so the capture window never sees these artifact generations'
+        # own writes.
+        await self._submit_artifacts(context=context, artifact_generates=[*plan.artifact_generates, *targeted])
```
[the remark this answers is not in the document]
polmichel: Fixed in 7be7ab02f. `_consolidate_artifact_generates` merges requests by `artifact_definition_id` before dispatch, unioning member/limit filters and letting an empty (all-members) filter subsume specific ones, so an artifact selected by both the merge diff and the generator output is dispatched once.

backend/infrahub/core/branch/tasks.py
```diff
@@ -473,14 +474,26 @@ async def _get_diff_root(
...
         summary_cache=DiffSummaryCache(
             cache=await get_cache(), serializer=DiffSummarySerializer(), key_namespace="branch_merge"
         ),
+        generator_diff_capturer=GeneratorTrackingGroupDiffCapturer(
+            diff_coordinator=diff_coordinator,
+            diff_repository=diff_repository,
+            serializer=DiffSummarySerializer(),
+            client=get_client(),
+            branch=branch,
+        ),
```
polmichel, line 496: This is another SDK dependency I'd like to remove later -> https://opsmill.atlassian.net/browse/IFC-2911

backend/infrahub/core/merge/selective_regen/definition_selector/base.py
```diff
@@ -89,7 +89,7 @@ async def select(
             )
             forced_triggers = [trigger for trigger in candidate_triggers if trigger is not None]
             for trigger in forced_triggers:
-                self.log.info(trigger.detail)
+                self.log.debug(trigger.detail)
```
polmichel, line 92: @ajtmccarty fyi I've removed the `log.info()`

backend/infrahub/core/merge/selective_regen/generator_diff_capturer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    async def capture(self, *, since: Timestamp, generator_definition_names: list[str]) -> list[NodeDiff]:
+        node_ids = await self._output_node_ids(generator_definition_names=generator_definition_names)
+
+        diff = await self.diff_coordinator.create_or_update_arbitrary_timeframe_diff(
+            base_branch=self.branch,
+            diff_branch=self.branch,
+            from_time=since,
+            to_time=Timestamp(),
+            name=str(uuid4()),
```
ajtmccarty, line 69: as these diffs are saved to the database, it might be good to give them some sort of special name or mark them as deletable in some way. and then clean them up ... somehow. not for this PR probably, but maybe a follow-up

ajtmccarty (approved):

backend/infrahub/core/merge/selective_regen/generator_diff_capturer.py
polmichel: This is a good idea, I'll encapsulate it in a dedicated task. https://opsmill.atlassian.net/browse/IFC-2941

PR #10034, opened 2026-07-24
perf(computed-attribute): batch Python transform recompute

saltas888:
# Why

When a change touches data that Python-transform computed attributes read, every affected node was recomputed through its own full path: a git repository init per node, one GraphQL update mutation per node, and every mutation's events re-triggered target queries and further recompute dispatches. One device-type rename could keep an instance churning for minutes, and at scale it takes the instance down. This is the "merge leaves the instance unusable" pain reported by large deployments.

Goal: absorb a wide fan-out as one bounded pass per chunk, with the same final values and the same events for real changes.

Non-goals: scoping the fan-out itself (every node of the kind vs only the affected ones) and a recovery surface for failed nodes, both are follow-ups.

# The improvements

The scenario for all of them is a Python computed attribute whose source data changes with many reader nodes, most visibly a branch merge or rebase, but the same path fires on a direct data change or a transform update that fans out to a whole kind.

1. **One git repository init per batch instead of one per node.** The recompute flow receives a chunk of node ids (up to ~250) and initializes the transform's repository once, before this every single node paid its own repository init.
2. **One bulk write instead of one GraphQL mutation per node.** Values persist through the shared bulk recompute writer in bounded transactions (100 nodes per transaction), no per-node `InfrahubUpdateComputedAttribute` mutation, no per-node client round-trip.
3. **Unchanged values emit no events, which removes the echo storm.** Before, every per-node mutation emitted `NodeUpdated` events even when the recomputed value was identical to the stored one, and those events re-triggered the `query-computed-attribute-transform-targets` and `graphql-query-group-update` automations, which dispatched further recomputes, a self-sustaining storm. Now a node whose save produces no effective change emits nothing and dispatches nothing, so a wave of recomputes that mostly lands on already-correct values simply stops.
4. **One bad node no longer hurts its siblings.** A transform that raises or returns a non-string for one node is skipped with its previous value intact and a logged reason, the rest of the batch persists, and the flow ends with a `submitted/written/skipped` summary line.

Plus one visibility fix: the branch tag is set on each submission at creation so the recompute runs stay visible in branch-filtered task views, tags added mid-run can be dropped by later in-flow tag updates.

# How we know it is an improvement

Measured, not inferred. The `TestMergeRecomputePython` suite in infrahub-private-tests restores the same dataset backup, renames a device type on a branch, merges, waits until every affected device is recomputed, and measures how long the instance keeps churning afterwards and how many flow runs the merge caused. We ran it twice on `develop_x-small`, once with develop ([baseline run](https://github.com/opsmill/infrahub-private-tests/actions/runs/30017807672)) and once with this branch ([run with this PR](https://github.com/opsmill/infrahub-private-tests/actions/runs/30017809470)), same dataset, same scenario, the only difference between the two runs is the commits in this PR.

| Metric (merge) | develop | this branch |
|---|---|---|
| Trailing recompute churn after merge | 275.2 s | ~1 s |
| `query-computed-attribute-transform-targets` runs | 1,321 | 1 |
| Process flow runs | 169 | 1 |
| Client-visible update mutations | 1 per node | 0 |

At scale develop does not degrade gracefully, it collapses: on `develop_large`, [the same scenario on develop](https://github.com/opsmill/infrahub-private-tests/actions/runs/30022700392) spawned 73,149 flow runs in 110 minutes (99% dispatch echo, 53k group updates plus 19k target queries, near-zero actual transform work) and died with Neo4j bolt-thread and file-descriptor exhaustion. [The same scenario with this branch](https://github.com/opsmill/infrahub-private-tests/actions/runs/30072597101), on a deliberately smaller database because the shared CI hosts kept OOM-killing full-size setups, survived without taking the host down. The churn that remains there is the reverse-index bookkeeping from dispatching to every node of the kind, which this PR deliberately does not touch, so the at-scale data confirms fan-out scoping as the necessary follow-up rather than a nice-to-have.

# How we know nothing broke

- The perf suite asserts correctness, not just speed: after the merge, every affected device's computed value matches the expected new value, on both runs above.
- The existing functional computed-attribute suite (real git repository, real transforms, end to end) runs unchanged and passes.
- The write path is not new code, it is the same bulk writer the Jinja2 computed attributes already persist through.
- The behavior contract is preserved: same final values, same per-node `NodeUpdated` events for real changes, the public `InfrahubUpdateComputedAttribute` mutation stays (just no longer used internally), and the flow parameters are unchanged so the Prefect deployment contract is stable.
- New tests pin the new behavior: unit tests for the partition/skip logic, a functional test asserting an oversized fan-out splits into chunks with every id exactly once and the branch tag on every submission, and a component test characterizing today's whole-kind fan-out so the future scoping work has a baseline to assert against.

# Built on the Jinja2 work, and what Python transforms still cannot do

This deliberately reuses the bulk recompute machinery Guillaume built for the Jinja2 computed attributes (`BulkRecomputeDispatcher`/`BulkRecomputeWriter`), no second write path, and it is verified with his `TestMergeRecomputePython` suite from infrahub-private-tests #29.

The part we cannot inherit: Jinja2 values are template renders the backend computes directly, so on merge they are recomputed in one coalesced in-process pass. Python transforms execute user code from a git repository on the task workers, they cannot join that pass, so they remain event-dispatched flows and a merge still dispatches a recompute for every node of the kind rather than only the changed ones. This PR makes that dispatch cheap and non-amplifying, scoping it to the affected nodes is the next big lever (see next steps).

### Suggested review order

1. `backend/infrahub/computed_attribute/tasks.py`, the `process_transform` rewrite plus the tagged submission sites.
2. Tests: unit (partition/skip, submitter tags), functional (chunked submission and tags), component (fan-out characterization).
3. The `add_tags` commit pair (`de6664ded` reverted by `58680ba2e`) documents a rejected fix, kept for history, the final approach tags at submission instead.

The design artifacts (spec, plan, critique, tasks) live in `dev/specs/004-batch-python-recompute/`.

# How to test

```bash
uv run pytest backend/tests/unit/computed_attribute/ -q
uv run pytest backend/tests/functional/computed_attributes/ -q
```

At-scale validation runs through infrahub-private-tests `test-dataset.yml` with `test_filter=TestMergeRecomputePython`, links above.

# Impact & rollout

- **Backward compatibility:** no schema changes, no API changes, event contract unchanged for real value changes.
- **Performance:** see above, the trailing recompute after a wide-impact change drops from minutes to about a second on the reference dataset.
- **Deployment notes:** safe to deploy, rollback is a clean revert, no migration.

# Next steps

- Handle failed/skipped nodes more gracefully, likely raise after the dispatch so the flow run shows failed while healthy siblings stay persisted:

  ```python
  await dispatcher.dispatch(writes=writes, ...)   # healthy siblings already persisted
  if skipped:
      raise RecomputeError(f"{len(skipped)} node(s) failed: {skipped}")
  ```

- A queryable recompute-failure surface with a one-click re-run (the `RecomputeComputedAttribute` mutation is the existing lever).
- Scope the merge fan-out to affected nodes instead of every node of the kind, that is the next big lever for giant-branch merges and the one thing the Jinja2 path also still needs.
- Phase-level spans/counters (repo-init, read, execute, write) to restore the per-node observability the batch path removed.

# Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (not user-facing, behavior-preserving)
- [x] Internal .md docs updated (`dev/knowledge/backend/computed-attributes.md`)

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10034?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dev/knowledge/backend/computed-attributes.md
```diff
@@ -121,6 +121,17 @@ All three run `process_transform_lifecycle`. On create or update it waits for th
...
 
+### Batch Execution
+
+`process_transform` processes its node ids as one batch per attribute, not one task per node:
+
+- The transform's git repository is initialized once for the whole batch and shared across the per-node executions. Transform execution must not mutate the shared checkout.
+- Each node's read still runs individually with `update_group=True`, keeping the node subscribed to the transform's query group (the reverse index that routes future source changes to affected readers).
+- The recomputed values persist through the shared bulk recompute writer (bounded transactions), not via per-node GraphQL mutations. The writer's skip-unchanged gating means a value identical to the stored one emits no event and dispatches no follow-on recompute, which is what keeps a wide fan-out from echoing into further waves.
```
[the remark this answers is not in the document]
saltas888: Good catch, the gate is per node rather than per value, a save that changes another field on the same node still emits an event. Reworded in c8734c805.

dev/knowledge/backend/computed-attributes.md
```diff
@@ -121,6 +121,17 @@ All three run `process_transform_lifecycle`. On create or update it waits for th
...
+- A node whose transform raises or returns a non-string is skipped with its previous value intact and a logged reason; the rest of the batch persists. The flow ends with a `submitted/written/skipped` summary line.
+- Each submission carries the branch tag at creation so the flow run stays visible in branch-filtered task queries; tags added mid-run do not survive later in-flow tag updates.
```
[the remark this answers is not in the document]
saltas888: The trigger flow was not actually invisible, it tags itself once mid-run and never goes through the dispatcher, so nothing rebuilds its tags. That said, creation tags are more robust than relying on call ordering, so I added the branch tag to the submitter's submissions too in c8734c805.

backend/tests/component/computed_attribute/test_merge_fanout_python.py
```diff
@@ -0,0 +1,109 @@
...
+
+CAR_COUNT = 5
+
+
+class TestMergeFanoutPython(ScopedRecomputeTestBase):
+    # The fan-out is dispatched as one process-transform per chunk of node ids.
+    WORKFLOW = COMPUTED_ATTRIBUTE_PROCESS_TRANSFORM
+
+    @pytest.fixture(scope="class")
+    async def transform_dataset(
```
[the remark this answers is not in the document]
saltas888: Extracted the shared query/repo/transform setup into create_transform01 in _base.py, both tests use it now, c8734c805.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -220,26 +219,50 @@ async def process_transform(
...
                 class_name=transform.class_name,
                 convert_query_response=transform.convert_query_response,
-                context=context,
+                repo=repo,
             )
-        _ = [r async for _, r in batch.execute()]
+        results: list[tuple[str, AttributeValueWrite | Exception]] = [
+            (oid, result) async for oid, result in batch.execute()
+        ]
+        writes, skipped = _partition_transform_results(results)
```
ajtmccarty, line 252: looks like `_partition` could handle logging the failures and only return the `writes` instead of splitting them and returning both.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -273,6 +296,8 @@ async def trigger_update_python_computed_attributes(
                 "computed_attribute_kind": computed_attribute_kind,
                 "context": context,
             },
+            # Must be a creation tag: in-flow tag updates drop tags added mid-run.
+            tags=[WorkflowTag.BRANCH.render(identifier=branch_name)],
```
ajtmccarty, line 304: do we need a task to cover the rest of the places where we add tag(s) mid-run?

ajtmccarty (approved):
looks good to me. @gmazoyer might want to look at it too

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -222,26 +223,51 @@ async def process_transform(
...
+            writes=writes,
+            branch_name=branch_name,
+            context=context,
+            coalesced=False,
+            recompute_depth=0,
+        )
+        log.info(
+            f"Recompute of '{attribute_name}' complete: submitted={len(results)} "
+            f"written={len(writes)} skipped={len(skipped)}"
+        )
```
gmazoyer, line 267: `written=len(writes)` counts what we send to the writer, not what really got saved. The writer skips unchanged nodes, but `dispatch()` throws away its return value. So the number of skipped no-ops never shows up. The plan asked for written/unchanged/skipped, but that is not what we log. Maybe return the written list from `dispatch()` and log `written` and `unchanged`. Or just rename this to `dispatched=`.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -222,26 +223,51 @@ async def process_transform(
...
             )
-        _ = [r async for _, r in batch.execute()]
+        results: list[tuple[str, AttributeValueWrite | Exception]] = [
+            (oid, result) async for oid, result in batch.execute()
+        ]
+        writes, skipped = _partition_transform_results(results)
+        for skipped_id, reason in skipped:
+            log.warning(f"Skipping recompute of '{attribute_name}' for node {skipped_id}: {reason}")
+
+        dispatcher = await build_bulk_recompute_dispatcher(schema_branch=schema_branch)
```
gmazoyer, line 256: `schema_branch` does not change in the loop. So you can build the dispatcher once, above the loop, like `process_jinja2` does.

backend/infrahub/computed_attribute/tasks.py
gmazoyer: I would keep it returning both. The function is probably easier to unit-test as it stands. The flow also still needs `len(skipped)` for the summary line. So moving the logging inside would probably not help much.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -191,8 +179,21 @@ async def process_transform(
...
+
+    Raises:
+        ValueError: if a computed attribute has no transform configured or the transform cannot be fetched.
+
+    """
+    log = get_run_logger()
     all_ids = list({*([object_id] if object_id else []), *(object_ids or [])})
     await add_tags(branches=[branch_name], nodes=all_ids)
+    if not all_ids:
+        return
```
gmazoyer, line 193: The `if not all_ids: return` is after `add_tags`. So an empty batch still tags. Moving the guard up matches what T008 wanted.

python_testcontainers/infrahub_testcontainers/docker-compose.test.yml
```diff
@@ -173,6 +183,12 @@ services:
       replicas: ${INFRAHUB_TESTING_API_SERVER_COUNT}
     image: "${INFRAHUB_TESTING_DOCKER_IMAGE}:${INFRAHUB_TESTING_IMAGE_VERSION}"
     command: ${INFRAHUB_TESTING_DOCKER_ENTRYPOINT}
+    ulimits:
+      # Raise the file-descriptor ceiling so a wide recompute fan-out doesn't hit
+      # "[Errno 24] Too many open files" on the API server's outbound connections.
+      nofile:
+        soft: 1048576
+        hard: 1048576
```
gmazoyer, line 191: @fatih-acar's input about this could be nice to have. I wonder if this needs to be documented somewhere for production environments too.

PR #10035, opened 2026-07-24
docs(git-integration): clarify one-way model, repo types, and sync_with_git

iddocohen:
## Summary

Correct and expand the Git Integration overview so it matches how the integration actually behaves. The page previously described the read-write Repository type as "full bidirectional integration" with data flowing "from Infrahub back to Git" — this is inaccurate (and contradicted the page's own ff-only merge statement), and it is the single most common customer misconception hit in the field.

Use case: smaller update / feature improvement (no structural change).

## Content changes (all in `git-integration/overview.mdx`)

- **Two version-control planes** (new section) — establishes the graph (data + schema, versioned in Infrahub's DB) vs Git (code intent) split, and that the link runs one way, Git → Infrahub.
- **Repository types** — reframed the standard type from "full bidirectional integration" to read-write integration: automatic Git → Infrahub import plus a single narrow write-back (the linked branch's default ref) on Proposed Change merge. Fixed the "creates a merge commit / pushes it back" paragraph to match the ff-only default and to state that only code intent moves, never data.
- **Read-only repositories** — clarified that they are not polled; you update via `ref` change or **Import latest commit**.
- **Controlling whether a branch reaches Git** (new subsection) — documents the per-branch `sync_with_git` flag and the common misreading of its name.
- **Common misconceptions** (new section) — three verified pitfalls.

## What needs reviewer attention

New prose to read carefully (the rest is small edits to existing lines):

- **"Two version-control planes".** Mental model; claims verified against `core/branch/models.py`, `git/tasks.py:696` (`merge_git_repository`), `git/repository.py:271` (`merge` + push).
- **`sync_with_git` subsection.** Matches the field description in `core/branch/models.py:40-42` and the PC-merge gate in `core/merge/branch_merger.py:286`.

## What didn't change

- No claims invented; every new statement traces to backend code.
- The `~60s` poll interval from field notes was intentionally **not** added — it isn't sourced from code, and the page already hedges ("multiple times per minute").

## Out of scope (deferred, not touched)

- Pre-existing voice drift ("bridges these two worlds"; "simpler, unidirectional") left as-is to keep this PR scoped. Small follow-up.
- Multi-environment promotion content (three pages in flight elsewhere).

## Verification

- `markdownlint-cli2` (repo `.markdownlint-cli2.yaml`): 0 issues.
- Vale: 0 errors, 0 warnings, 0 suggestions.
- Word check (`simple`/`easy`/`just`): none in new content.
- `cd docs && npm run build` **not** run locally — docs deps not installed in this checkout; relying on CI for the render check.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10035?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/docs/git-integration/overview.mdx
+++ b/docs/docs/git-integration/overview.mdx
@@ -4,7 +4,7 @@ title: Git Integration
 
 import ReferenceLink from "../../src/components/Card";
 
-Connect an external Git repository to keep the code that drives Infrahub — schemas, transformations, generators, checks, GraphQL queries, and artifact definitions — under version control, next to your infrastructure data. Infrahub imports that code and keeps it in sync as the repository changes.
+Connect an external Git repository to keep the code that drives Infrahub — schemas, Transformations, Generators, checks, GraphQL queries, and artifact definitions — under version control, next to your infrastructure data. Infrahub imports that code and keeps it in sync as the repository changes.
 
 To set one up, see [connect a repository](./connect-repository.mdx), [configure `.infrahub.yml`](./infrahub-yml.mdx), and [filter which branches sync](./branch-synchronization.mdx).
 
@@ -13,7 +13,7 @@ To set one up, see [connect a repository](./connect-repository.mdx), [configure 
 Infrahub has two separate version-control systems, and understanding the split is the key to the rest of this section.
 
 - **The graph** — your infrastructure data (objects) and schema. Infrahub versions the graph itself: you branch it, change it, review the diff in a [Proposed Change](../proposed-changes/overview.mdx), and merge. This is Git-like version control, but the store is Infrahub's database, not a Git repository.
-- **Git** — the code that describes intent: schemas, transformations, generators, checks, GraphQL queries, and artifact definitions, declared in [`.infrahub.yml`](./infrahub-yml.mdx). Infrahub reads this code from the repository and imports it.
+- **Git** — the code that describes intent: schemas, Transformations, Generators, checks, GraphQL queries, and artifact definitions, declared in [`.infrahub.yml`](./infrahub-yml.mdx). Infrahub reads this code from the repository and imports it.
 
 The link between the two runs one way: from Git into Infrahub. Git is where you author code intent; Infrahub imports it and applies it to the graph. Data moves the other way only inside Infrahub — when you merge a branch, the change is applied to the graph, not pushed back to Git. A read-write repository does write one thing back to Git, and only one; the [repository types](#read-only-vs-core) section covers it.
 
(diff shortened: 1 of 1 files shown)


lancamat1 (commented):
Huge thanks to @iddocohen for this — these changes were badly needed. The old "full bidirectional integration" framing was actively misleading customers into expecting data pushed back to Git, and it contradicted the page's own ff-only merge statement. This fixes a real, recurring source of confusion. I verified every new claim against the backend — the one-way import, `sync_with_git` semantics (including that it does *not* gate import into Infrahub), read-only-vs-read-write behavior, and the read-only-only "Import latest commit" restriction all hold up against the code. The "two planes" mental model and the `sync_with_git` misconception callout are exactly the right framing.

**Verdict: Approve with three minor fixes.**

**1. Line 35 — write-back is stated unconditionally, but it's gated by `sync_with_git`. (P2, valid)**
The summary says the Git write-back happens "on a Proposed Change merge" with no condition, which contradicts line 53 and the "Controlling whether a branch reaches Git" section — and the backend gates it (`branch_merger.py:286`). Suggest:

> …and on a Proposed Change merge it updates the linked Git branch — for branches that sync with Git (see [Controlling whether a branch reaches Git](#controlling-whether-a-branch-reaches-git)).

**2. Lines 7 & 16 — capitalize feature names. (P3, valid)**
`docs/AGENTS.md → Infrahub Terminology` requires **Transformations** and **Generators** capitalized as feature names. Both lists use lowercase. (`schemas`, `checks`, `GraphQL queries`, `artifact` stay lowercase — not in the terminology table, so only those two words change, in both lines.)

> Connect an external Git repository to keep the code that drives Infrahub — schemas, **Transformations**, **Generators**, checks, GraphQL queries, and artifact definitions — …

**3. Line 53 — "fast-forward" is *when possible*, not unconditional.**
The code runs plain `git merge` (default `--ff`), which fast-forwards only for linear history and otherwise still creates a merge commit. Suggest "By default this is a fast-forward **when possible**, so no merge commit is created." *(Related, pre-existing: line 111's "fast-forward only (ff-only)" is technically inaccurate — the code isn't `--ff-only`, which would fail rather than merge-commit. Worth a reword if you're in the file, but out of this PR's scope.)*

iddocohen:
@lancamat1 Thanks for the careful review — all three fixes are in, plus the line 111 `ff-only` reword you flagged, since leaving it would have contradicted the corrected line 53. Mind taking another look?

lancamat1 (approved):

PR #10055, opened 2026-07-28
refactor(regeneration): inject selectors so each owns its kind-specific logic

polmichel:
## Summary

Refactors the post-merge selective-regeneration selectors so the merge follow-up no longer reaches
into kind-specific details. Addresses the review feedback that typing the orchestrator against a
specific pair of selectors "defeats the purpose" of the shared `DefinitionSelectorBase`, and the
follow-up that selection and output capture were two concerns bundled onto a single selector.

Jira: [IFC-2916](https://opsmill.atlassian.net/browse/IFC-2916) (internal, under IFC-2704)
Origin: https://github.com/opsmill/infrahub/pull/9924#discussion_r3597524020

## Change

Each selector now owns only its kind's **selection** — loading definitions, narrowing to affected
members, building the run request, and consolidating its own requests. It no longer carries its
cascade role or knows how its output is captured.

A `CascadeParticipant[RequestT]` composes a selector with its place in the cascade and is the single
surface the follow-up drives — it delegates to its selector rather than exposing it:

- **`CascadeSource`** carries a `CascadeSourceOutputFactory` that builds a `CascadeSourceOutput` from
  the requests the selector chose, so the source's writes are captured without the selector handling
  that.
- **`CascadeTerminal`** carries none — nothing downstream re-reads what it produces.

The `RegenerationPlanner` builds the plan and the dispatcher runs it by cascade role: sources run to completion. The dispatcher holds no kind-specific logic and its failure fallbacks are generic too.

Behavior is unchanged: same selective dispatch, same generator→artifact cascade, same per-generator
failure isolation and safe fallbacks.

## How tested

- Unit: the participant entry mapping (a source forwards its requests to its output factory and
  carries the built capture; a terminal carries none); `build_plan` tags each entry and shares the
  modified-kind computation; `reselect_from_cascade_output` excludes the sources and escalates
  repository-wide; `terminal_full_regenerations` covers every terminal and excludes sources;
  `GeneratorOutputFactory` derives the capture scope from a generator's runs; `GeneratorSelector`
  tags its blanket regeneration as merge-sourced; `ArtifactSelector` consolidates per definition; the
  planner routes consolidation to each owning participant; the dispatcher submits what the planner
  returns and falls back to blanket terminal regeneration on failure.
- Component: the real-graph selection flow drives the rebuilt dispatcher end-to-end.

Housekeeping changelog fragment added (backend-only refactor; no schema, config, event, CLI, docs, or
frontend changes).

## How to review

1. `core/merge/selective_regen/participant.py` — `CascadeParticipant` / `CascadeSource` /
   `CascadeTerminal`: the composition, `to_entry`, and the façade (`load` / `plan` /
   `consolidated_entry` / `full_regeneration`) that keeps the selector private.
2. `core/merge/selective_regen/definition_selector/base.py` — the per-selector contract: selection
   (`workflow`, `load_definitions`, `select`, `consolidate`) plus the blanket
   `full_regeneration_workflow` / `full_regeneration_parameters`.
3. `core/merge/selective_regen/generator_output.py` — `GeneratorOutputFactory` (the cascade-output
   builder) and the tracking-group capturer; production impls inherit their `Protocol` explicitly.
4. `core/merge/selective_regen/orchestrator.py` — the `RegenerationPlanner` protocol, the injected
   participants, and `build_plan` / `reselect_from_cascade_output` / `consolidate_submissions` /
   `terminal_full_regenerations`.
5. `core/merge/regeneration_dispatcher.py` — role-based dispatch and the generic terminal fallback,
   free of kind-specific logic.
6. `core/branch/tasks.py` — the wiring (the capturer is composed into the generator source).

## Flow

```
merge committed
   ▼
dispatch()  ──(feature off · no diff summary · summary load fails · selection fails)──▶  FULL REGEN (every kind)
   ▼
build_plan()   per participant: load → select → plan entry
               CascadeSource (generators) carry an output capture · CascadeTerminal (artifacts) don't
   ▼
run the sources (generators)        [awaited, so their writes have landed]
   ├─ a source run fails ─────────────────────────────────────▶  regenerate every TERMINAL (never re-run sources)
   ▼
capture each source's own writes  →  reselect the terminals that read them
   ├─ capture or reselect fails ──────────────────────────────▶  regenerate every TERMINAL
   ▼
submit terminals + reselected       [fire-and-forget, deduped per workflow, only after capture]
```

[IFC-2916]: https://opsmill.atlassian.net/browse/IFC-2916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10055?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/selective_regen/models.py
```diff
@@ -21,12 +53,26 @@ class LoadedDefinition[DefinitionT: DefinitionModel]:
...
 
 
+@dataclass(frozen=True)
+class PlannedRegeneration:
+    """One selector's selected requests, tagged with the workflow and dispatch mode that run them."""
+
+    workflow: WorkflowDefinition
+    cascade_role: CascadeRole
+    requests: Sequence[RegenerationRequest]
+    output: CascadeSourceOutput | None = None
```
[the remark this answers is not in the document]
polmichel: Fixed in 2eca76f0e. `build_plan` now routes every entry through a `_plan_entry` helper that enforces the invariant at plan construction: a cascade source must carry an output capture, a terminal must not — a mismatch raises a `ValueError` the follow-up turns into safe full regeneration, rather than a source that runs while its terminals are never reselected. Both the initial plan and the cascade re-selection go through the helper, so the validation holds for both. Covered by two unit tests asserting the exact error message for a source-without-output and a terminal-with-output.

backend/infrahub/core/merge/selective_regen/orchestrator.py
```diff
@@ -9,105 +9,154 @@
...
-            diff_summary=diff_summary,
-            target_branch=target_branch,
-            modified_kinds=modified_kinds,
-        )
+        entries: list[PlannedRegeneration] = []
+        for selector in self.selectors:
+            if selector.cascade_role is CascadeRole.SOURCE:
+                continue
+            loaded_definitions = await selector.load_definitions(target_branch=target_branch)
+            forced_repositories = repositories_forcing_full_regeneration(
```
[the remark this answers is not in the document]
polmichel: Fixed in 717e43403. `reselect_from_cascade_output` now aggregates the loaded definitions across every non-source selector and computes the forced-repository set once, mirroring `build_plan`. A repository escalated by any missing fingerprint therefore regenerates all of its definitions, including a sibling terminal kind that shares the repository, rather than escalating each kind in isolation. Covered by a unit test where a null-fingerprint terminal escalates a populated-fingerprint sibling in the same repository.

backend/infrahub/core/merge/selective_regen/definition_selector/generator_selector.py
```diff
@@ -10,15 +12,46 @@
...
+    workflow = REQUEST_GENERATOR_DEFINITION_RUN
+    cascade_role = CascadeRole.SOURCE
+
+    def __init__(
+        self,
+        client: InfrahubClient,
+        gate: DefinitionGate,
+        impacted_resolver: ImpactedSubscriberResolver,
+        log: logging.Logger | logging.LoggerAdapter[logging.Logger],
+        output_capturer: GeneratorMutationDiffCapturer,
```
ajtmccarty: if possible, I think this should be restructured in a few ways
1. `output_capturer` should just be a dependency of `GeneratorTrackingOutput`. `GeneratorSelector` does not actually use it, so `GeneratorMutationDiffCapturer` should not be a dependency of `GeneratorSelector`
2. `definition_names` looks more like an argument to the `GeneratorTrackingOutput` constructor, but maybe that is not possible. if that's the case, then `GeneratorTrackingOutput` can have a `set_definition_names` method and guard on `capture()` that prevents it from being called with no definition names. or a no-op in that case would be fine too
3. constructing an instance of `CascadeSourceOutput` does not look like it is really the responsibility of `GeneratorSelector`. it looks like you need some way to link this `GeneratorSelector` to its implementation of `CascadeSourceOutput`, but I don' think this class should be building the other class

maybe some sort of shared `CachedDefinitionLoader` class could make some of this work where multiple components share a single definition loader instance

backend/infrahub/core/merge/selective_regen/orchestrator.py
```diff
@@ -9,105 +9,178 @@
...
+        A source feeds the cascade, so it must capture output for its terminals to reselect from; a
+        terminal ends the chain, so it must not. A mismatch is a wiring error caught here rather than a
+        source that runs but whose terminals are silently never reselected.
+
+        Raises:
+            ValueError: When a source carries no output capture, or a terminal carries one.
+
+        """
+        output = selector.output_capture(requests)
+        if selector.cascade_role is CascadeRole.SOURCE and output is None:
```
ajtmccarty: kind of goes with my above structuring comment, but, ideally, this kind of thing is raised during construction instead of runtime

backend/infrahub/core/merge/selective_regen/definition_selector/generator_selector.py
polmichel: Thanks for the review. Good point, and you're right that the responsibilities are mixed and the timing feels off.

The output selection logic should live in a separate component: the `XXXSelector` handles the input side, and `CascadeSourceOutput` handles the output side.

I've created an object that gathers the inputs and outputs and can generate an entry in the selective regeneration plan for the merge. With this object (`CascadeParticipant`), all the logic is encapsulated in one place and won't leak into the orchestrator, which also covers the point you raise in your next comment.

backend/infrahub/core/merge/selective_regen/generator_output.py
```diff
@@ -34,7 +39,39 @@ class GeneratorMutationDiffCapturer(Protocol):
...
+    Binds the generators' definition names to the capturer so the follow-up can capture their output
+    without handling the names itself.
+    """
+
+    def __init__(self, *, capturer: GeneratorMutationDiffCapturer, definition_names: list[str]) -> None:
+        self._capturer = capturer
+        self._definition_names = definition_names
+
+    async def capture(self, *, since: Timestamp) -> list[NodeDiff]:
+        return await self._capturer.capture(since=since, generator_definition_names=self._definition_names)
```
ajtmccarty: I might be missing something here, but I think that you can simplify the construction of these objects so that all of the classes are built at the beginning of the flow instead of building this class as the flow runs. it's kind of hard to keep it all straight, but I think that if this `capture()` method accepted `generator_definitions` as an argument instead of a dependency, then `GeneratorTrackingOutput` could be built along with everything else in `_build_post_merge_regeneration_dispatcher`

polmichel: True, this is now done.
`capture()` now takes the source's selected requests instead of binding the definition names as a dependency, so the output object holds only its capturer. It's built once in `_build_post_merge_regeneration_dispatcher` and injected through the planner, so nothing on this path is constructed mid-flow anymore.

polmichel: `GeneratorTrackingOutput` is now `GeneratorCascadeOutput` and the `GeneratorOutputFactory`/`GeneratorSourceOutputFactory` are now gone.

changelog/+selective-regen-cascade-participant.housekeeping.md
ajtmccarty, line 1: not sure this is necessary

ajtmccarty (approved):

changelog/+selective-regen-cascade-participant.housekeeping.md
polmichel: As this change lands on `develop`, I wanted to note somewhere that the code has been refactored. But you're right, I do not think this brings much value to the user to know about it

PR #10058, opened 2026-07-28
fix(regeneration): profile-aware kind selection for post-merge and proposed-change regeneration

polmichel:
## Summary

Regeneration selection now widens a diff's changed kinds to include the node kind behind each changed profile, resolved from the branch's live schema. A change touching only a profile selects every generator and artifact whose query reads the profiled node kind — on both the post-merge selective regeneration and the proposed-change refresh.

## What changed

- **New `SchemaProfileExpander`** (`core/regeneration/profiles.py`): given a branch and a set of changed kinds, adds the profiled node kind for every changed profile kind, using the schema's profile→node relationship. A kind that names no profile passes through. The `SchemaManager` is a constructor dependency (resolved to `registry.schema` at the wiring sites), so the component no longer reaches into the global registry.
- **Wired into both selection paths** through one component: injected as the merge planner's `kinds_expander`, which runs it in the planner's shared plan step so both the initial plan and the cascade reselection widen alike; constructed at the proposed-change flow level for the generator and artifact loops. The proposed-change generator loop, which selected on raw `query_models` membership, now runs through the same widened kinds.
- **`reads_kind` moved to `predicates.py`**: with profile handling now living in the expander, it reduces to `kind in query_models`, so the name-stripping heuristic (`kind.replace("Profile", "", 1)`) is removed and the predicate lives beside `query_changed` / `definition_changed` as a free function over `RegenerationDefinition` — off the definition models, which stay plain data holders.

The widening only ever adds kinds, so the no-under-execution guarantee is preserved.

## Tests

- unit: `SchemaProfileExpander` against an injected in-memory schema branch that generates profiles
- unit: gate widening selects both a generator and an artifact for a profiled kind
- unit: the planner passes the expanded kinds to every selector
- unit: `reads_kind` matches only a kind the definition's query reads
- component: `run_generators` dispatches every generator reading a profiled kind on a profile-only change

## Notes

- Resolves https://opsmill.atlassian.net/browse/IFC-2907 in full — both items the ticket tracks: the fragile profile heuristic, and profile handling applied to artifacts but not generators.
- Rebased onto `pmi-20260727-refactor-selectors`, which is now participant/planner based; the profile-kind widening integrates into the planner's shared plan step, and the tests moved to the participant wiring. Review feedback addressed: `SchemaProfileExpander` takes an injected `SchemaManager` (no global registry), and `reads_kind` moved off the definition models into `predicates.py`.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/regeneration/profiles.py
+++ b/backend/infrahub/core/regeneration/profiles.py
@@ -1,0 +1,28 @@
+from __future__ import annotations
+
+from typing import Protocol
+
+from infrahub.core import registry
+
+
+class ModifiedKindsExpander(Protocol):
+    """Widens the changed kinds a regeneration selection considers, beyond those literally in the diff."""
+
+    def expand(self, *, modified_kinds: list[str], branch: str) -> list[str]: ...
+
+
+class SchemaProfileExpander:
+    """Widen changed kinds with the node kind behind each changed profile, from a branch's live schema.
+
+    A profile holds attribute values that apply to the nodes it targets, so changing a profile node
+    effectively changes those nodes. A profile schema's name is the kind of the node it targets, so a
+    changed profile widens the set to that node kind; a kind that names no profile passes through.
+    """
+
+    def expand(self, *, modified_kinds: list[str], branch: str) -> list[str]:
+        schema_branch = registry.schema.get_schema_branch(name=branch)
+        expanded = set(modified_kinds)
+        for kind in modified_kinds:
+            if kind in schema_branch.profiles:
+                expanded.add(schema_branch.get_profile(name=kind, duplicate=False).name)
+        return list(expanded)
--- a/backend/infrahub/core/regeneration/profiles.py
+++ b/backend/infrahub/core/regeneration/profiles.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
-from typing import Protocol
+from typing import TYPE_CHECKING, Protocol
 
-from infrahub.core import registry
+if TYPE_CHECKING:
+    from infrahub.core.schema.manager import SchemaManager
 
 
 class ModifiedKindsExpander(Protocol):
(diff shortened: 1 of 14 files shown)


backend/infrahub/core/regeneration/profiles.py
```diff
@@ -0,0 +1,28 @@
...
+class SchemaProfileExpander:
+    """Widen changed kinds with the node kind behind each changed profile, from a branch's live schema.
+
+    A profile holds attribute values that apply to the nodes it targets, so changing a profile node
+    effectively changes those nodes. A profile schema's name is the kind of the node it targets, so a
+    changed profile widens the set to that node kind; a kind that names no profile passes through.
+    """
+
+    def expand(self, *, modified_kinds: list[str], branch: str) -> list[str]:
+        schema_branch = registry.schema.get_schema_branch(name=branch)
```
ajtmccarty: ideally, `SchemaManager` is a dependency of this class and `registry` is removed

backend/infrahub/generators/models.py
```diff
@@ -104,7 +104,11 @@ def instance_noun(self) -> str:
         return "instances"
 
     def reads_kind(self, kind: str) -> bool:
-        """Whether a data change to ``kind`` is relevant because the query reads that kind."""
+        """Whether a data change to ``kind`` is relevant because the query reads that kind.
+
+        Profile changes are matched by widening the changed kinds to the profiled node kind before
+        this check, so a profile read need not be reconstructed here.
```
ajtmccarty: not certain this belongs here

ajtmccarty (approved):

backend/infrahub/generators/models.py
polmichel: Good catch! I also moved this method to `predicates.py` since this is used as a predicate.

backend/infrahub/core/regeneration/profiles.py
polmichel: Good point, dependency injected.

PR #10059, opened 2026-07-28
refactor(frontend): replace Apollo Client with urql for the GraphQL transport [INFP-563]

bilalabbad:
## What & why

The frontend used `@apollo/client` as a **transport-only** GraphQL layer — no Apollo hooks, no Apollo cache (`fetchPolicy: "no-cache"`, `queryDeduplication: false`), no subscriptions. TanStack Query is and remains the sole caching and request-deduplication authority. Apollo was paying for a cache and a hooks layer the app never used.

This replaces it with `@urql/core` + `@urql/exchange-auth`, keeping the imperative `graphqlClient.query()` / `graphqlClient.mutate()` interface so the ~95 call sites are untouched by design.

Dependencies: `-@apollo/client`, `-apollo-upload-client`, `-@types/apollo-upload-client`, `+@urql/core`, `+@urql/exchange-auth`.

**Bundle:** 4,574,160 → 4,539,212 bytes gzipped JS (**−34,948 bytes, ≈ −34 KB**), measured by summing gzipped `dist/assets/*.js` across a production build on the same machine.

No user-facing behaviour change. Auth, priority header, error routing, token refresh-and-replay, and file uploads are all preserved.

## Notable implementation details

Four places where urql's defaults differ from Apollo's and parity had to be restored deliberately:

**`preferGetMethod: false`** — urql's `within-url-limit` default sends small queries as GET, which the backend's `/graphql` answers with the SPA fallback `index.html`. Forced to POST, with a regression test.

**`__typename` injection** — Apollo's `InMemoryCache` added it automatically and several dynamic object queries relied on it. Restored with a `mapExchange` running `formatDocument`.

**One urql client per endpoint** — this is the subtle one. urql's `Client` merges concurrent operations by `hash(query, variables)` and **ignores the URL**, but this app carries branch and point-in-time *in the URL*. A single shared client therefore served one branch's data to both callers of two concurrent identical queries on different branches — a branch-safety hazard, not a caching nicety. `getGraphqlClient(branch, date)` memoises a client per endpoint, so each branch/date keeps its own in-flight operations. Verified against `@urql/core` directly: with one shared client the two-branch case produces **1** fetch and identical data for both callers; with per-endpoint clients it produces 2, each correct. Same-endpoint identical queries still share one request, as before.

**Error routing** — `handleGraphQLErrors` keeps the catalogue-code routing (`TOKEN_EXPIRED` / `AUTHENTICATION_REQUIRED` → `/login`, `PERMISSION_DENIED` silent, generic → the caller's `processErrorMessage`). Both network and GraphQL errors reject, so a resolved result carries `data` only — every use-case already threw on `errors`, and the ones that did not dereferenced `null`.

The transport also dropped things that turned out to be dead: the `X-Priority` header is a client-level constant (no call site ever overrode it), and `GraphQLRequestContext.date` is now `Date | null` (nothing ever passed a string).

## Testing

- `pnpm test` — 1096 passing (159 files), including new regression tests for POST-not-GET, `__typename` injection, endpoint targeting, the priority header, the refresh-and-replay contract (one refresh, one replay, then bail), cross-branch / cross-date concurrent operations, the multipart shape of a `File` upload, and `X-Priority` surviving the replayed operation after a refresh. The last two cover guarantees that rest entirely on `@urql/*` internals and had no fast-suite coverage.
- `pnpm exec biome ci .` — clean.
- `pnpm knip` — clean.
- `pnpm build` — succeeds.
- `pnpm exec betterer ci` — clean (`190 issues`, unchanged).

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/app/src/shared/api/graphql/client.ts
```diff
@@ -0,0 +1,134 @@
...
+    refreshAuth: async () => {
+      await queryClient.fetchQuery(refreshAccessTokenQueryOptions()).catch((error) => {
+        redirectToLogin();
+        throw error;
+      });
+    },
+  };
+});
+
+const clientsByEndpoint = new Map<string, Client>();
```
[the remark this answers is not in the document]
bilalabbad: it requires hundreds/thousands of branch before really being problematic

frontend/app/src/shared/api/graphql/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface GraphQLRequestContext {
+  branch?: string | null;
+  date?: Date | null;
+  processErrorMessage?: (message: string) => void;
+}
+
+export interface GraphQLResult<TData> {
+  data: TData;
+  error?: CombinedError;
+  errors?: Array<{ message: string }>;
```
pa-lem, line 12: **`error` and `errors` can never be populated now.** `toGraphQLResult` throws on both `networkError` and `graphQLErrors`, so these two fields are always `undefined` — which keeps ~67 unreachable `if (errors) throw` guards under `entities/` compiling silently. Dropping them lets `tsc` drive the cleanup.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: confirm that `toGraphQLResult` in
frontend/app/src/shared/api/graphql/client.ts throws on both networkError and graphQLErrors, so
`error` and `errors` on `GraphQLResult` can never be populated. If confirmed, remove both fields
and the now-unused `CombinedError` import from types.ts. Then run `pnpm exec tsc --noEmit` and
delete every unreachable `if (errors...) throw` guard it flags under frontend/app/src/entities/.
One flagged site is NOT dead code: shared/libs/graphiql/use-graphiql-fetcher.ts uses `errors` to
fall back to `baseFetcher` — fix that one with `.catch(() => null)` rather than deleting it. Report
any other site where removing the field changes behaviour instead of deleting dead code.
```

</details>

frontend/app/src/shared/api/graphql/client.ts
```diff
@@ -0,0 +1,134 @@
...
+    addAuthToOperation: (operation) => {
+      const accessToken = getAccessToken();
+      if (!accessToken) return operation;
+
+      return authUtilities.appendHeaders(operation, { Authorization: `Bearer ${accessToken}` });
+    },
+    didAuthError: (error) => {
+      return hasCatalogueCode(error, ERROR_CODES.TOKEN_EXPIRED);
+    },
+    refreshAuth: async () => {
```
pa-lem, line 48: **Lost guard: refresh resolving without an `access_token`.** The old `retryWithRefreshedToken` called `redirectToLogin()` and errored in that case. Now the operation replays with the stale token, gets `TOKEN_EXPIRED` again, and only then redirects — one extra round-trip, and the caller resolves rather than rejects.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: trace what happens in
frontend/app/src/shared/api/graphql/client.ts when the token refresh inside `refreshAuth` resolves
successfully but without an `access_token`. Compare it with the deleted Apollo client's
`retryWithRefreshedToken`, which called redirectToLogin and errored in that case. If the guard is
genuinely missing, restore it — redirectToLogin() and throw — and add a test covering it.
```

</details>

frontend/app/src/shared/api/graphql/client.ts
pa-lem, line 58: **Reopening the unbounded-growth thread on a different cause.** The rebuttal was about branch count, but the key is `CONFIG.GRAPHQL_URL(branch, date)`, which embeds `date.toISOString()` at millisecond precision — so it's point-in-time selection that grows this map, not branches, and nothing ever evicts. A small LRU is safe here because per-endpoint clients only matter for in-flight dedup.

Separately: this map is module-global with no reset hook, so a client and its `authExchange` state leak between tests in `client.test.tsx` / `client.dedup.test.tsx`.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: read CONFIG.GRAPHQL_URL in
frontend/app/src/shared/config/config.ts and work out what varies in the `clientsByEndpoint` key.
Note whether the date is embedded at millisecond precision, and whether any code path removes
entries. If the map can grow without bound, bound it with a small LRU (cap ~20, evict
least-recently-used); eviction is safe because per-endpoint clients only matter for in-flight
dedup. Keep the existing cross-branch and cross-date tests passing.

Then check whether the map retains clients and their authExchange state across tests. If state
leaks, export a `__resetGraphqlClients()` helper that clears it and call it in `afterEach` in
client.test.tsx and client.dedup.test.tsx.
```

</details>

frontend/app/src/shared/api/graphql/client.ts
```diff
@@ -0,0 +1,134 @@
...
+// but this app carries the branch/point-in-time in the URL, not in variables.
+// Without this, two concurrent identical query+variables on DIFFERENT branches share one network request
+// and both receive one branch's data.
+function getGraphqlClient(branch?: string | null, date?: Date | null): Client {
+  const url = CONFIG.GRAPHQL_URL(branch, date);
+  let client = clientsByEndpoint.get(url);
+  if (!client) {
+    client = new Client({
+      url,
+      requestPolicy: "network-only",
```
pa-lem: **`requestPolicy` has no effect here.** There's no `cacheExchange` in the chain, so nothing reads it. Harmless, but it reads as if a cache policy is in play.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: does `requestPolicy: "network-only"` at
frontend/app/src/shared/api/graphql/client.ts line 69 have any effect given the exchanges in the
chain? If there is no cacheExchange, drop it or add a comment explaining why it is kept.
```

</details>

frontend/app/src/shared/api/graphql/client.ts
```diff
@@ -0,0 +1,134 @@
...
+        },
+      },
+      exchanges: [addTypenameExchange, authenticationExchange, fetchExchange],
+    });
+    clientsByEndpoint.set(url, client);
+  }
+  return client;
+}
+
+// Map urql result to the preserved `{ data, errors }` shape and run error routing.
```
pa-lem, line 83: **This comment and the one at line 115 both describe a shape the client no longer returns.** Since `5ef09c9` it returns `{ data }` and throws on network or GraphQL errors — there is no `errors` in the result and no Apollo result-shape parity.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: read the comment above `toGraphQLResult`
(line 83) and above `graphqlClient` (line 115) in frontend/app/src/shared/api/graphql/client.ts and
compare each against what the code actually returns. If they describe a `{ data, errors }` result
the client no longer produces, rewrite them: the client returns `{ data }` and throws on network or
GraphQL errors. Drop any claim of Apollo result-shape parity. Leave accurate comments alone.
```

</details>

frontend/app/src/shared/api/graphql/error-handling.ts
```diff
@@ -0,0 +1,76 @@
...
+      `[GraphQL error]: Code: ${parsed.code}, Message: ${graphQLError.message}, ` +
+        `Path: ${graphQLError.path}`
+    );
+
+    switch (parsed.code) {
+      case ERROR_CODES.TOKEN_EXPIRED:
+        redirectToLogin();
+        return;
+
+      case ERROR_CODES.AUTHENTICATION_REQUIRED:
```
pa-lem: Three small things in this switch:

- `TOKEN_EXPIRED` and `AUTHENTICATION_REQUIRED` are identical arms — could fall through.
- `default` and `UNDEFINED_ERROR` `return`, so sibling errors in the same response stop routing. `PERMISSION_DENIED` uses `continue` for exactly that reason, and the Apollo handler didn't return here either.
- The Apollo handler also logged `graphQLError.locations`; that's dropped.

<details>
<summary>Prompt</summary>

```
Check each of these three independently in frontend/app/src/shared/api/graphql/error-handling.ts
and fix only the ones that are real:
1. Are the TOKEN_EXPIRED and AUTHENTICATION_REQUIRED cases identical? If so, merge them into one
   fall-through arm.
2. Do the `default` and UNDEFINED_ERROR arms `return` and thereby stop sibling errors in the same
   response from routing? If so, use `continue`, matching the PERMISSION_DENIED intent.
3. Did the deleted Apollo handler log `graphQLError.locations`? If so, include it in the
   console.error here.
```

</details>

frontend/app/src/shared/api/graphql/client.test.ts
```diff
@@ -0,0 +1,352 @@
...
+      .mockResolvedValueOnce(jsonResponse(tokenExpiredBody))
+      .mockResolvedValueOnce(jsonResponse({ data: { __typename: "Query" } }));
+    fetchQuerySpy.mockResolvedValue({ access_token: "new-token", refresh_token: "new-refresh" });
+
+    // WHEN
+    const result = await graphqlClient.query({ query: PING });
+
+    // THEN
+    expect(result.data).toEqual({ __typename: "Query" });
+    expect(result.errors).toBeUndefined();
```
pa-lem, line 309: **This assertion can never fail** — `result.errors` is never set by the client, so it's vacuous.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: determine whether `result.errors` can
ever be defined given what graphqlClient.query returns. If it cannot, the assertion at
frontend/app/src/shared/api/graphql/client.test.tsx line 284 is vacuous — remove it and assert on
`result.data` only.
```

</details>

dev/knowledge/frontend/entities-structure.md
```diff
@@ -362,10 +359,10 @@ See `dev/guidelines/frontend/naming-conventions.md` for the full naming conventi
 
 ## GraphQL transport vs server-state hooks
 
-Apollo Client is kept as the GraphQL transport (auth links, error handling, retry) only. All server-state hooks are TanStack Query. Do not use `useQuery` / `useMutation` / `useLazyQuery` from `@apollo/client` — they were removed in 2026-05.
+`@urql/core` is the GraphQL transport (auth, error routing, token refresh, uploads) only — there is no GraphQL-layer cache and there are no GraphQL hooks. All server-state hooks are TanStack Query.
```
pa-lem, line 362: **urql doesn't handle uploads.** `fetchExchange` implements no multipart-request spec — that would need `@urql/exchange-multipart-fetch`. Dropping `apollo-upload-client` was right (uploads go over REST at `/api/storage/upload/file`), but as written this will lead someone to push a `File` through `graphqlClient.mutate` and get a silent null. Same claim in `dev/knowledge/frontend/request-priority.md:33` ("Uploads ride the same terminating `fetchExchange`").

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: determine whether urql's fetchExchange
has any multipart-request support in this codebase, and find where file uploads actually go. If
uploads are REST-only, remove "uploads" from the list of things the urql transport handles in
dev/knowledge/frontend/entities-structure.md (~line 362) and in
dev/knowledge/frontend/request-priority.md (~line 33), and state explicitly that GraphQL has no
multipart path. If GraphQL uploads do work, leave the docs and explain how.
```

</details>

pa-lem (commented):
Reviewed the transport swap end to end. The migration itself is solid — the imperative interface is preserved across ~95 call sites, the bundle win is real, the regression tests for POST-not-GET / `__typename` / per-endpoint clients are the right ones, and `a337d42` fixes a duplicate React resolution in the lockfile.

Two things I'd fix before merge: the GraphiQL fallback below, and the dead `error`/`errors` fields on `GraphQLResult` (inline). Everything else is docs, comments and cleanup.

Four items that can't be anchored to a line in this diff:

---

### 1. `shared/libs/graphiql/use-graphiql-fetcher.ts:72` — the one functional regression

The file isn't in this diff, so I can't comment inline. It does:

```ts
const { data, errors } = await getObjectsCountFromApi({ objectKind, branchName, atDate });
if (errors) {
  return baseFetcher(graphQLParams, opts);   // graceful degrade
}
```

This is the only `errors` consumer in the app that recovers rather than throws. Now that the client throws, the branch is unreachable and the exception escapes the fetcher (no try/catch in the file — the `.catch` at line 37 is on `data.json()`). GraphiQL surfaces an error instead of falling back to the non-parallel path, which is most likely to bite on a kind the user can't read.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: confirm that getObjectsCountFromApi now
throws on GraphQL errors rather than resolving with an `errors` array, and that nothing catches it
between there and the caller of the fetcher. If confirmed, replace the
`const { data, errors } = await getObjectsCountFromApi(...)` destructuring and its `if (errors)`
guard with a `.catch(() => null)` so a failed count query still falls back to
`baseFetcher(graphQLParams, opts)`. Keep the existing totalCount handling unchanged. If the guard
turns out to still be reachable, say so and change nothing.
```

</details>

---

### 2. PR description — three claims no longer hold

<details>
<summary>Prompt</summary>

```
Verify each of these three claims against the branch before editing the PR description, and fix
only the ones that are actually wrong:
1. The stated bundle delta. Recompute it from the figures in the description (4,574,160 ->
   4,539,212 bytes) and check the units.
2. "File uploads are preserved." Check whether urql's fetchExchange implements the GraphQL
   multipart-request spec and where this app actually uploads files.
3. "The `{ data, errors }` result shape is preserved 1:1." Check what graphqlClient.query and
   .mutate return today.
```

</details>

---

### 3. Transport-level dedup is now on, and it used to be deliberately off

The deleted Apollo client set `queryDeduplication: false` with a comment naming TanStack Query the single dedup authority. urql dedups in-flight operations by `hash(query, variables)`, and `client.dedup.test.tsx` now asserts that. I think it's benign — per-caller error routing still runs in `toGraphQLResult` — but it's worth recording as a decision rather than leaving it as a silent reversal.

<details>
<summary>Prompt</summary>

```
Check whether this issue is real before changing anything: confirm from git history that the
deleted Apollo client set `queryDeduplication: false` with a comment naming TanStack Query the
single dedup authority, and confirm that urql's Client dedups in-flight operations by
hash(query, variables). If the invariant really did change, either add a comment in client.ts
explaining why sharing one in-flight request per endpoint is now safe, or restore the previous
behaviour. Do not leave the reversal undocumented.
```

</details>

---

### 4. `a23401a` deleted some load-bearing comments

That pass removed the explanations of the non-obvious parity decisions (urql's GET default, dedup ignoring the URL, the Response-stream-read-once gotcha, TOKEN_EXPIRED terminality) while leaving the two comments that are actually wrong (flagged inline). The test comments in particular were the reason a reader can tell why each regression guard exists.

<details>
<summary>Prompt</summary>

```
Review commit a23401a, which deleted comments across
frontend/app/src/shared/api/graphql/{error-handling.ts,types.ts,client.test.tsx,client.dedup.test.tsx}.
For each deleted comment, decide whether it explained something non-obvious that is not documented
elsewhere in the repo — in particular: urql's within-url-limit GET default versus the backend's SPA
fallback; the Client dedup ignoring the URL while branch and date live in the URL; a Response body
stream only being readable once; and TOKEN_EXPIRED being terminal because the auth exchange already
refreshed and replayed. Restore only those, keeping the bare GIVEN/WHEN/THEN markers alongside
them. Leave deleted comments that merely restated the code.
```

</details>

dev/knowledge/frontend/entities-structure.md
bilalabbad: wrong urql does handle it. https://nearform.com/open-source/urql/docs/advanced/persistence-and-uploads/#file-uploads

frontend/app/src/shared/api/graphql/client.ts
bilalabbad: `/api/auth/refresh` returns AccessTokenResponse, where `access_token` is a required field, and FastAPI validates the response body so a 200 without a token can't come back.
 
 The previous assertion was not needed so I removed it.

frontend/app/src/shared/api/graphql/client.ts
bilalabbad: tbh I do not see the issue here? date is the point-in-time selected by the user. It's not like we are going to create thousand on client (there are all cleared on page refresh). I measured it: one client holds ~18 KB. So 100 clients is 1.8 MB, in an app that already uses 50–150 MB. You'd need some thousands before it matters.

New client only created when you pick a new time, switch to a new branch etc... The map also dies on page refresh. A heavy session ends up in the low hundreds at max.

frontend/app/src/shared/api/graphql/client.ts
bilalabbad: why not, I'll update it

frontend/app/src/shared/api/graphql/types.ts
bilalabbad: I know, it's the same behavior as current Apollo (surprisingly), I did not change it to avoid the need to touch error management in all usecases

pa-lem:
### PR description — two claims left to fix

**Line 9** — the stated delta doesn't match its own figures: 4,574,160 → 4,539,212 bytes is **34,948 bytes (~34 KB)**, not `−76.68.9 KB` (which also looks like a typo).

**Line 23** — "the `{ data, errors }` result shape is preserved 1:1 including `errors` never being `[]` on success" stopped being true at `5ef09c9`; the client now throws on GraphQL errors and never returns `errors`.

Line 11 on file uploads is correct — that was my mistake earlier, `@urql/core` does implement the multipart spec.

<details>
<summary>Prompt</summary>

```
Verify each of these two claims against the branch before editing the PR description for
opsmill/infrahub#10059, and fix only the ones that are actually wrong:
1. Line 9's bundle delta. Recompute it from the figures in the description (4,574,160 ->
   4,539,212 bytes) and check both the arithmetic and the units.
2. Line 23's claim that the `{ data, errors }` result shape is preserved 1:1 and that `errors` is
   never `[]` on success. Check what graphqlClient.query and .mutate actually return today, and
   what toGraphQLResult does when the response carries graphQLErrors.

Leave the file-uploads claim on line 11 alone — it is accurate.
```

</details>

pa-lem (approved):

PR #10061, opened 2026-07-28
refactor(merge): extract a post-commit follow-up guard

polmichel:
## Summary

Consolidates the post-commit follow-up error handling on the merge and rebase paths. The post-merge and post-rebase follow-ups run after the operation is irreversibly committed, so each wrapped its call in `try / except Exception / log.exception(...)` to keep a single follow-up failure from surfacing an already-committed operation as failed.

Those five near-identical blocks are replaced with a `merge_follow_up_guard` context manager, so the "why broad" rationale lives in one place. Where a guard genuinely needs to stay broad but must also react to the failure (set a flag, fall back to full regeneration), it keeps its explicit handling, and its docstring now states why the broad catch is deliberate.

Behaviour is unchanged: the catch breadth stays `except Exception` (so `asyncio.CancelledError` still propagates), and every follow-up is still logged and skipped on error.

Jira: [IFC-2944](https://opsmill.atlassian.net/browse/IFC-2944)

## Alternatives considered

**Narrowing each guard to specific exception types.** Rejected. The wrapped calls are Prefect workflow submissions, event-service sends, cache reads and writes, and diff serialization, none of which expose a stable, bounded set of failure types. Naming precise types would either miss a failure that still has to be absorbed (an unexpected type would escape and report an already-committed merge or rebase as failed) or short-circuit the full-regeneration fallback, reintroducing the exact failure mode the broad catch exists to prevent. The invariant on this path is that over-execution (re-running a follow-up, regenerating more than strictly needed) is acceptable while under-execution (a committed operation failing, or a stale artifact left behind) is not, so any uncertainty deliberately resolves to the broad catch rather than a precise type.

**Folding the reacting guards into the context manager.** Rejected. The guards that set a failure flag or return `None` to trigger full regeneration run recovery code on the failure path, not only logging. Routing them through the log-and-continue guard would hide that behaviour, so they keep an explicit `try/except` whose comment or docstring states why the broad catch is intended.

The change is therefore an audit and a factoring of the shared log-and-continue pattern, not a move to precise exception types.

## Test plan

- New unit tests for the guard: a failure inside the block is logged and absorbed while execution continues, a clean block stays silent, and one failing iteration in a loop does not abort the siblings.

[IFC-2944]: https://opsmill.atlassian.net/browse/IFC-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10061?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/follow_up.py
```diff
@@ -0,0 +1,24 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from infrahub.log import InfrahubLogger
+
+
+@contextmanager
+def merge_follow_up_guard(log: InfrahubLogger, message: str) -> Iterator[None]:
```
ajtmccarty: nothing about this appears to be specific to the merge logic. this looks like it is more of a `log_exception_guard` or something and could probably move to `backend/infrahub/utilities/`

ajtmccarty (approved):

PR #10065, opened 2026-07-28
some updates for dev documentation

ajtmccarty:
trying to distill the changes from #10052 into something more repeatable

truly not sure if some of this might be better as  `dev/guides/backend/creating-new-components.md`

---
## Summary by cubic
Improves backend dev docs on constructor-first wiring, interfaces/protocols, configuration invariants, and testing doubles. Extends the API backpressure example to show named-event observers, a single `_notify`, and how protocols keep third‑party deps out of core logic.

- **Documentation**
  - Wiring: no late registration; pass deps to `__init__` (0+ as required `list[...]`); build near entry points; factories resolve settings and accept adapters; configure at construction; process‑global registries build lazily in their own module.
  - Interfaces/Protocols: choose implicit `Protocol` or explicit interface module (prefer explicit); never put an ABC in the consumer; name methods in the consumer’s vocabulary; adapters live in their own module; only the wiring imports both; verify no third‑party imports under logic.
  - Configuration: tightly bound numeric fields with `allow_inf_nan=False`; enforce cross‑field rules with `@model_validator` and clear, greppable messages; test the validator and shipped defaults; use enums for values that leave the process and pass enum members.
  - Testing: two doubles for collaborators — a recording double for exact call sequences/values, and a failing double to prove failure containment; keep helpers shared and prefer adapters over patching.
  - API backpressure example: events are named methods; one `_notify` fans out with per‑observer failure containment; collaborators arrive via the constructor; wiring in `server.py` chooses sinks so `api/admission` stays decoupled behind `Protocol`s.

<sup>Written for commit f7fbfbf355fac9362ecdf27de44d14dfbeaf990a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10065?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


.agents/rules/backend-component-design.md
```diff
@@ -18,12 +18,21 @@ Constructor dependencies for new code are required parameters - not `collaborato
...
+
 ## Build components near the application entry point
 
 Construct components as close to the application entry point as possible. Use a builder class or factory function when wiring is non-trivial, and inject each sub-component rather than constructing it inside a parent component's `__init__`.
 
 Prefect `@flow` functions are application entry points: resolve singleton getters (`get_database()`, `get_workflow()`, …) at the top of the flow only — never inside helpers or component internals — then build the component and delegate to it. The flow body stays a thin composition root; the business logic lives in the component.
 
+Anything that comes from outside the component's own domain — loaded settings, an external service client, a telemetry sink — is resolved at that entry point, never inside the component:
+
+- **Settings resolve in the factory, not the component.** A component takes plain values (`window_seconds: float`, `max_retries: int`), never a `Settings` object and never a module-global read. The factory is then the only place that knows a value came from configuration, which is also what makes the component directly testable with hand-picked values.
```
ogenstad, line 31: Do we have any examples of this? In situations where there's a high number of settings i thing a `Settings` object can be cleaner. An example would be the Config object for the SDK.

dev/guidelines/backend/python.md
```diff
@@ -370,6 +416,19 @@ except Exception as exc:
...
 A broad `except Exception` is justified only at a top-level boundary (a task worker loop, a request handler) whose job is to prevent one failure from taking down the process — and even there, log the exception and re-raise or record it, never discard it.
 
+The one other justified case is a **best-effort side-effect boundary**: notifying an observer or other sink whose failure must not affect the operation that triggered it. There the exception is logged and deliberately not re-raised, and the `try` must wrap the *individual* call so one bad sink cannot skip the ones behind it:
+
+```python
+# ✅ Good - a failing sink must not fail the operation that produced the event
+for observer in self._observers:
+    try:
+        observer.on_depth_changed(queued=queued, running=running)
+    except Exception:
```
ogenstad: I don't think we should include `except Exception` as a good example. (https://docs.astral.sh/ruff/rules/blind-except/)

Instead we should have a specific exception even if it's made up so that the LLM doesn't pick up on this part and start to introduce additional broad exceptions. We also have #10002.

ogenstad (commented):
Looks good to me overall, but added two comments.

.agents/rules/backend-component-design.md
ajtmccarty: I think the `Config` is a special case b/c it is a settings object. its only responsibility should be tracking many settings. any components that actually do things should have settings injected when they are constructed.

dev/guidelines/backend/python.md
ajtmccarty: I think in the case of an "observer" object (or any other "best-effort side-effect" components that should not block the primary logic during failure) we want to catch the bare `Exception`. we could move the bare `Exception` catch into `on_depth_changed` in this example and re-raise a different error class, but the end result would be the same

.agents/rules/backend-component-design.md
```diff
@@ -57,6 +66,22 @@ A single implementation does not need an interface yet; introduce one when the s
...
+
+Split it in three:
+
+- **Use a `Protocol`, not an ABC, and declare it beside the component that depends on it.** Structural typing is what makes this work: the adapter satisfies the protocol by having matching signatures, so it never imports the protocol's module and the dependency does not run backwards. An ABC does not have that property — the adapter must subclass it, and therefore import whatever module it lives in. If you genuinely need an ABC, it goes in a module of its own that both sides import, never in the consumer's module.
+- **Name the methods in the depending component's vocabulary**, not the adapter's — `on_depth_changed(*, queued: int, running: int)`, not `set_gauges(...)` — and pass the values as arguments rather than handing over `self`, so the adapter can never read back into the component. The component then depends on a shape it defined, has no idea what is on the other side, and stays free to change its internals.
+- **Put the concrete adapter in a separate, purpose-named module** that is the only place importing the library, and **let only the wiring layer import both** (see "Build components near the application entry point").
```
polmichel, line 83: Is there anything worth saying about explicitly declaring the protocol in the class definition?

I mean this:
```
class MyClass(MyProtocol):
    # Protocol implemented
```

And not:
```
class MyClass():
    # Protocol implemented implicitly
```

The upside of the first option is the direct link between the implementation and the protocol it satisfies. The downside is that we would need to import the protocol. Another option would be to make sure these protocols live in a package that is safe to import from anywhere.

ajtmccarty: I'm always ready to say more about dependency injection. added some more detail to describe either using a Protocol without importing it or using a Protocol/ABC interface defined in a separate module

polmichel (approved):
LGTM besides the merge conflicts

PR #10083, opened 2026-07-30
feat(attribute): store bare IP values on write when allow_prefix is disabled [INFP-551]

ajtmccarty:
**Stack position: 2 of 5.** Base is **#10081** (PR 1), which declares `allow_prefix`. See
`dev/specs/infp-551-bare-ip-attribute/pr-split-plan.md` for the full split. Merge strictly 1 → 5.

**~85% of this PR is test code.** The production change is ~40 lines in `backend/infrahub/core/attribute.py`.

# Why

PR 1 gave an `IPHost` attribute the vocabulary to say it holds a bare address. This PR makes the write
path honour it: a declared attribute stores exactly one spelling of an address, whichever spelling the
caller used. Reads are untouched — making the masked spelling resolve as *lookup input* is PR 3.

## What changed

`backend/infrahub/core/attribute.py`:

- `IPHost._normalize_value` returns the bare address when the attribute refuses a prefix, and the
  existing `with_prefixlen` form otherwise. The undeclared path is byte-for-byte what it was.
- `validate_format` rejects a real subnet prefix on a declared attribute. The host-length comparison is
  `interface.ip.max_prefixlen`, never a hardcoded 32/128 — a `/32` is a redundant host mask on IPv4 but a
  subnet prefix on IPv6, and the version-agnostic comparison is what gets that right.
- A new `BaseAttribute._normalize_assigned_value` hook, a no-op on the base class, called on the update
  path (`from_db`-sourced assignment and the pre-save validation). `IPHost` overrides it so an **edit**
  converges on the value a **creation** would have stored. Every other kind is provably unaffected
  because the base implementation returns its argument.
- `iphost_allows_prefix()`, a module-level helper, answers permissively for a schema whose `parameters`
  is not the per-kind class, so a base-classed schema cannot accidentally acquire the restriction.

`to_db()` and the derived properties (`ip`, `version`, `prefixlen`, `binary_address`) are deliberately
untouched — they read `self.value`, so they follow the stored spelling without needing to know about the
flag.

## The uniqueness bypass this closes

The commit `normalise bare IP values on the update path` is worth reading on its own. The pre-save
uniqueness check ran against the **un-normalised** value, so on a declared attribute `10.0.0.1/32` could
be written alongside an existing `10.0.0.1` — two rows for one address on an attribute marked
`unique=True`. Normalising before the check closes it, and
`TestUniquenessAcrossInputForms::test_bare_and_host_masked_input_collide` pins it, with the undeclared
attribute asserted in the same test to still collide only on its own stored spelling.

## Test shape

Every behavioural test pairs a **declared** attribute with an **undeclared control** asserting the old
behaviour, on one kind (`TestingDnsRecord`) that carries both flavours side by side. The classes:

| Class | Covers |
|---|---|
| `TestValueValidationAndNormalisation` | input forms accepted, rejected, and how each is stored |
| `TestStorageAndDerivedProperties` | the `:AttributeValue` vertex, and the derived properties following it |
| `TestUniquenessAcrossInputForms` | the bypass above |
| `TestTheUpdatePath` | an edit converging on the creation value |
| `TestGeneratedKindsInheritTheDeclaration` | profile and object-template kinds |
| `TestBranchMerge` | the value surviving a branch merge bare |
| `TestAttributeKindChange` | a kind change off/onto `IPHost` |

The two read-only classes load their schema once per class rather than once per test, and share a single
saved record — per-test schema loading was the dominant cost in the suite.

The `default_value` cases live in their own file, `test_iphost_default_value.py`, rather than being added
to `test_manager_schema.py`, which was already too large. That file is untouched by this branch.

`TestAllowPrefixIsImmutable` (integration) proves the flag cannot be toggled **through the real
`/api/schema/load` endpoint**, not merely through the model. That test is why the SDK pointer bump in PR 1
is load-bearing rather than cosmetic: without `IPHostAttributeWrite` in the SDK's generated write models,
`allow_prefix` is silently dropped from every schema-load payload and 4 of these cases fail.

This PR touches no generated file and does not move the submodule pointer (still `525b28f`, inherited
from PR 1).

## Changelog

No fragment here. Per constraint C6 in the split plan, all documentation and changelog content is
consolidated into PR 5 so the user-facing wording is reviewed in one place.

### DIFF
--- a/backend/infrahub/core/attribute.py
+++ b/backend/infrahub/core/attribute.py
@@ -1047,17 +1047,6 @@ class IPHost(BaseAttribute):
     value: str
 
     @staticmethod
-    def _allows_prefix(schema: AttributeSchema) -> bool:
-        """Return whether the attribute accepts an address carrying a prefix.
-
-        Anything other than the per-kind parameters class cannot express the restriction, so the
-        permissive answer is the safe one for a base-classed schema.
-        """
-        if isinstance(schema.parameters, IPHostAttributeParameters):
-            return schema.parameters.allow_prefix
-        return True
-
-    @staticmethod
     def get_allowed_property_in_path() -> list[str]:
         return [
             "binary_address",
@@ -1150,6 +1139,12 @@ class IPHost(BaseAttribute):
         return convert_ip_to_binary_str(obj=self.obj)
 
     @classmethod
+    def _allows_prefix(cls, schema: AttributeSchema) -> bool:
+        if isinstance(schema.parameters, IPHostAttributeParameters):
+            return schema.parameters.allow_prefix
+        return True
+
+    @classmethod
     def validate_format(cls, value: Any, name: str, schema: AttributeSchema) -> None:
         """Validate the format of the attribute.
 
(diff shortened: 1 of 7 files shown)


ajtmccarty:
replaced by #10090

PR #10085, opened 2026-07-30
fix: render Jinja2 computed attributes sourced from an unallocated number pool (closes #7836)

### LINKED ISSUE #7836: bug: Jinja2 Computed Attribute Error

polmichel:
## Why

Creating a node whose Jinja2 computed attribute formats a value sourced from a number pool failed with `%d format: a real number is required, not NoneType`. This regressed in 1.5.0 when node creation began building a throwaway preview object with `process_pools=False` (to compute lock names): the pool value is intentionally left unallocated on that preview, yet macro rendering still ran against it, so filters like `'%09d' | format(...)` blew up on `None`.

Goal: a computed attribute that depends on a pool value renders correctly once the pool has been allocated, instead of failing during creation.

Closes #7836

Ref: [IFC-2095](https://opsmill.atlassian.net/browse/IFC-2095)

## What changed

Behavioral changes:

- A Jinja2 computed attribute whose template reads a pool-sourced local attribute is left unrendered while that attribute has no allocated value, and renders normally once the pool allocation has happened. Macros that do not depend on an unallocated pool are unaffected.

Implementation notes:

- A dedicated predicate decides whether a macro can be rendered yet. `_resolve_jinja2_variables` stays a pure variable-to-value mapping; the create path and the recompute path both consult the predicate before resolving and rendering.
- Skips cascade in dependency order: a computed attribute that references a macro skipped for an unallocated pool is deferred as well, instead of rendering against the unresolved value.
- The root fields a template depends on are read through `InfrahubJinja2Template.get_referenced_root_fields()`, shared by both macro paths.

## How to test

```bash
uv run pytest \
  backend/tests/component/core/node/test_create_node_computed_pool.py \
  backend/tests/component/graphql/test_mutation_create_jinja2_attributes.py
```

## Impact & rollout

- **Performance:** negligible; one extra pass over a template's variables per computed attribute during creation.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added
- [ ] External docs updated (not user-facing beyond the bug fix)
- [ ] Internal .md docs updated (no architectural change)

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-2095]: https://opsmill.atlassian.net/browse/IFC-2095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/infrahub/core/node/__init__.py
```diff
@@ -821,6 +839,9 @@ async def _recompute_local_jinja2(
                 failed_attributes.add(target.attribute.name)
                 continue
 
+            if self._has_pending_pool_dependency(schema_branch=schema_branch, jinja_template=jinja_template):
+                continue
+
             variables = await self._resolve_jinja2_variables(
                 db=db, schema_branch=schema_branch, jinja_template=jinja_template
             )
```
polmichel, line 853: Quite fragile since it needs to be manually called before a Jinja2 variable resolution. I think the proper way to manage this need is to have a dedicated component with clear inputs which will manage the Jinja2 variables resolution.

backend/infrahub/core/node/__init__.py
```diff
@@ -754,10 +774,16 @@ async def _process_macros(self, db: InfrahubDatabase) -> None:
                 continue
 
             jinja_template = InfrahubJinja2Template(template=attr_schema.computed_attribute.jinja2_template)
+            referenced_attributes = {variable.split("__")[0] for variable in jinja_template.get_variables()}
```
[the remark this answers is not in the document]
polmichel: This one is a false positive in practice, but it prompted a good cleanup.

`skipped` only ever contains **computed-attribute** names, and a node cannot have a relationship and an attribute sharing a name — both resolve through the same instance namespace (`get_attribute`/`get_relationship` key off the same `getattr(self, name)`, and the schema treats attributes + relationships as one combined field set). So a relationship root extracted by `split("__")[0]` can never be an element of `skipped`; the intersection only matches a genuine local-attribute dependency, which is the intended behavior. The identical root-extraction is already used in the recompute path (`depends_on_failed`), so this is consistent with existing code.

That said, the `split("__")[0]` extraction was duplicated across the create and recompute paths, so I've pulled it into a single `_referenced_root_fields` helper in bc7fd5dee.

backend/infrahub/core/node/__init__.py
ogenstad: Could you create an issue in Jira to capture these changes and label it as tech debt?

ogenstad (approved):
LGTM

backend/infrahub/core/node/__init__.py
polmichel: https://opsmill.atlassian.net/browse/IFC-2955 Here is the internal issue.

PR #10087, opened 2026-07-30
Fix Jinja2 comparison during repository imports

### LINKED ISSUE #3094: bug: Jinja2 templates are always updated when a new commit is added to a repo
### Component

Git Integration

### Current Behavior

The controls to determine if a Jinja2 template needs to be updated in the database after a new commit is added to the repo isn't working as intended.

Even if the Jinja2 template/transform hasn't been modified a message like this will be displayed within the git-agent log:

2024-04-26T11:23:00.500158Z [info     ] New version of the Jinja2 Transform 'blue' found, updating [infrahub.git] branch=main repository=blue routing_key=request.git.sync

```python
    @classmethod
    async def compare_jinja2_transform(
        cls, existing_transform: InfrahubNode, local_transform: InfrahubRepositoryJinja2
    ) -> bool:
        # pylint: disable=no-member
        if (
            existing_transform.description.value != local_transform.description
            or existing_transform.template_path.value != local_transform.template_path
            or existing_transform.query.id != local_transform.query
        ):
            return False

        return True
```

The problem seems to be that "existing_transform.template_path.value" is a string and local_transform.template_path is a `Path`

### Expected Behavior

The template in the database should only be updated if it is out of date

### Steps to Reproduce

_No response_

### Additional Information

_No response_

ogenstad:
# Why

When importing a repository we'd always see a log entry stating that the Jinja2 transforms were updated even though there were no changes. The reason is that we were comparing a filepath as a string vs one defined as a `Path` object.

Targeting `release-1.11` as I just want this in place for now while testing something else, it's a two year old bug with no real urgency since it's just producing some extra logging and an attempt to update something that doesn't need updating.

Closes #3094

## What changed

* Minor fix
* Added unittests to vaoid regressions

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/integrator.py
+++ b/backend/infrahub/git/integrator.py
@@ -563,7 +563,7 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
     ) -> bool:
         if (
             existing_transform.description.value != local_transform.description
-            or existing_transform.template_path.value != local_transform.template_path
+            or existing_transform.template_path.value != str(local_transform.template_path)
             or existing_transform.query.id != local_transform.query
             or existing_transform.dependencies.value != local_transform.dependencies
             or existing_transform.dependencies_complete.value != local_transform.dependencies_complete
--- a/backend/tests/unit/git/test_transform_jinja2_information.py
+++ b/backend/tests/unit/git/test_transform_jinja2_information.py
@@ -1,0 +1,142 @@
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+import pytest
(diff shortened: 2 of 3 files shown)


gmazoyer (approved):
Good find!

PR #10095, opened 2026-07-31
feat(schema): report the schema fields the load endpoint does not apply

dgarros:
## Why

#9814 made `POST /api/schema/load` validate against a generated user-facing *write* contract, and — after review reversed that half of the design — set `extra="ignore"` so a field the user may not set is dropped rather than rejected. That kept the read-back, edit, re-load round trip working, but it was too permissive in one direction: **any** extra key was accepted and the author was told nothing, so a misspelled key produced a schema quietly different from the one they wrote. ADR 0010 flagged this as the follow-up to revisit.

## What changed

Extra keys are now classified rather than uniformly dropped:

- a name the contract knows at that location but the user may not set → **warning**, value still dropped;
- any other name → **error**, since the only ways to produce one are a typo and a field that no longer exists.

```console
$ infrahubctl validate schema models/base/dcim.yml     # with `inherited: true` added
nodes[0].attributes[0].inherited: Read-only field, the submitted value is ignored (received: True)
Schema is valid !!

$ infrahubctl validate schema typo.yml                 # `optionl: true`
Schema not valid, found 1 error(s)
  nodes[0].attributes[0].optionl: Unknown field, it is not part of the schema (received: True)
```

`POST /api/schema/load` and `/check` return the warnings on the response, one per distinct field, naming every kind and element that carried it:

```text
'inherited' is a read-only field, the submitted value is ignored [InfraDevice.name, InfraDevice.interfaces]
```

**The published request schema is unchanged.** The endpoint still validates through `InfrahubSchemaWrite` alone — no read-only field is advertised as settable, and `schema/openapi.json` is byte-identical after regeneration. That was the design constraint: keep the load contract clean and put the feedback in a separate layer.

## Implementation notes

**A generated table, not a third model family.** `SdkSchemaGenerator._read_only_fields` emits `python_sdk/infrahub_sdk/schema/generated/contract.py` (17 entries) from two diffs:

1. the read variant of a class against its write variant — `inherited`, `used_by`, `hierarchy`, `hash`, the derived `kind`;
2. the internal pydantic counterpart of a value model against its generated write model.

The second diff is the one that isn't obvious. The parameters, choice, computed-attribute and extension models are declared in the generator rather than derived from `internal.py`, so they omit the `id`/`state` bookkeeping every internal schema model carries, and each computed-attribute variant omits its siblings' fields. All of those names appear in a full `SchemaRoot.model_dump()` — which integration tests submit — so classifying them as unknown would have broken the round trip. Deriving from the internal models rather than hardcoding also means a new field on an internal parameters model is tolerated automatically instead of becoming a spurious error.

Root-level keys that `GET /api/schema` adds over the write root (`main`, `profiles`, `templates`, `namespaces`) are listed explicitly in the generator and pinned against `SchemaReadAPI` by a test, so a raw response body resubmitted warns rather than fails.

**Applying the table needs to know which model governs each place in the payload.** `_collect_extra_fields` (SDK) walks the raw payload alongside the *validated* write document: the document resolves the model at every location, including which member of a discriminated union an attribute matched, so raw keys are compared against the fields that location actually accepts. Alternatives considered and rejected — a generated `extra="allow"` model family, and a `mode="before"` hook writing into the pydantic validation context — are recorded in ADR 0010.

One consequence: extra fields surface only once the payload is otherwise valid. A payload rejected for another reason names them on the next run.

**Warning transport.** `SchemaLoadAPI`'s validator becomes `mode="wrap"` so the warnings, which are only visible on the raw payload, can be set on the instance the handler returns and merged into the response next to `gather_warnings()`. They ride the existing `deprecation` warning type rather than a new enum member, so an SDK released before this change can still parse a load response; a dedicated type has to wait until a tolerant SDK is the supported floor.

## Behaviour changes to be aware of

Two shapes that previously loaded now fail. Both were silently dropped before, so the loaded schema differed from the authored one:

1. a misspelled field name;
2. attribute `parameters` belonging to a different attribute `kind` — e.g. `start_range` on a `Number` attribute. (`Dropdown` + `regex` was already rejected pre-#9814, so that one is unchanged.)

This means a repository whose committed schema carries either one stops importing until corrected. Covered by the `changed` changelog fragment and the deprecation guide.

Also fixed: `load_schema` returned early when the submitted schema matched the one already loaded and dropped `warnings` entirely, so a deprecation warning went unreported on every load after the first.

## Testing

- Backend unit suite: **2177 passed**. `test_schema_load_contract.py` reworked — the five `*-tolerated` cases split into accepted-with-warning and rejected; new tests for warning aggregation, the root-key pin, and unknown-field rejection. `test_generated_visibility.py` gains an invariant test that the generated table is exactly the read/write delta per family.
- Swept every schema file in the repo (`models/**`, `backend/tests/**`, `development/**` — 28 files) through the new validator: **no unknown-field errors anywhere**; only `models/base/dcim.yml` with a deliberately-added `inherited` produces a warning.
- `invoke lint` (ruff/ty/mypy) and `invoke docs.lint` clean. `schema/openapi.json` byte-identical after `invoke schema.generate-jsonschema`.
- Functional tests in `test_load_schema.py` are updated but **not run locally** — they need the testcontainers stack, so CI is the first execution.

## Submodule

`python_sdk` is bumped to the head of the companion PR, opsmill/infrahub-sdk-python#1223, whose parent is the commit `release-1.11` previously pinned — so the pointer moves strictly forward. **This needs re-pointing at the merged commit once #1223 lands**, as was done for #9814.

Note #1223 targets `infrahub-develop` while this PR targets `release-1.11`, per the requested targets.

## Docs

- `dev/adr/0010-generated-user-facing-schema-contract.md` — decision updated, the anticipated negative consequence resolved, and the two rejected alternatives recorded.
- `dev/knowledge/backend/schema-definitions.md` — new section on how the warning/error split is derived and applied.
- `docs/docs/release-notes/deprecation-guides/schema-load-write-contract.mdx` — the "extra fields are dropped" section was now wrong; rewritten with both categories and the remediation steps.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10095?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/unit/core/schema/test_generated_visibility.py
```diff
@@ -110,6 +115,41 @@ def test_write_model_publishes_allowed_values(family: str) -> None:
...
+
+def _declared_read_only_fields(model: type[BaseModel]) -> set[str]:
+    """Read-only names the table declares for a model, including those it inherits."""
+    names: set[str] = set()
+    for klass in model.__mro__:
+        names |= READ_ONLY_FIELDS.get(klass.__name__, frozenset())
+    return names
+
+
+@pytest.mark.parametrize("family", list(WRITE_TO_READ_MODEL))
```
ajtmccarty: this is kind of an unusual way to parametrize a test. usually we do either a list of some type of dataclass or set the actual values in the `parametrize` decorator, but this one is doing kind of a mix where it uses the dict keys in `parametrize` and then access the dict within the test

changelog/+schema-load-extra-field-feedback.changed.md
ajtmccarty, line 1: should this be used to update an existing changelog for the release instead of adding a new one for this unreleased change?

docs/docs/release-notes/deprecation-guides/schema-load-write-contract.mdx
```diff
@@ -32,17 +35,48 @@ Schema fields are now classified by who may see or set them:
...
+version — is an error naming the field path and the value received:
+
+```text
+nodes[0].attributes[0].optionl: Unknown field, it is not part of the schema (received: True)
+```
+
+Earlier 1.x versions dropped these silently, which meant a misspelled key produced a schema
+that quietly differed from the one you wrote. Attribute `parameters` that belong to a
+different attribute `kind` fall in this category too, since they configure nothing on the kind
+they were set on:
```
ajtmccarty: this was never released right? or is this referring to `1.2` or some much older version?

tasks/backend.py
```diff
@@ -833,6 +867,7 @@ def generate(self) -> None:
             rendered = rendered.replace("__VARIANT__", variant.suffix)
             Path(f"{self.generated}/{variant.name}.py").write_text(rendered, encoding="utf-8")
 
+        _write_sdk_generated_contract(self.generated, self._read_only_fields())
```
ajtmccarty, line 872: probably out of scope, but all of this logic should be contained in a class or classes outside of this tasks file

tasks/backend.py
```diff
@@ -1043,6 +1078,91 @@ def _families(self, minimum: "Visibility", suffix: str) -> list[dict[str, Any]]:
             },
         ]
 
+    def _families_for_contract(self, minimum: "Visibility") -> list[dict[str, Any]]:
+        """Every generated write family, at the given field visibility."""
+        suffix = "Write"
```
ajtmccarty, line 1085: looks like it should be an enum or `Literal`

tasks/backend.py
ajtmccarty: this would also allow cleaning up all this in-function imports b/c they could all move to the top of the module containing the actual logic for this task and then the `SdkSchemaGenerator` could be imported in one task instead of dozens of imports in the wrong places

ajtmccarty (approved):
approved b/c it looks like everything is working, but `tasks/backend.py` could really use some cleaning up

backend/tests/unit/core/schema/test_generated_visibility.py
dgarros: Good catch — switched to the dataclass pattern in `1353cd5`^ (`3db160a2a`), with the pytest ids unchanged.

I copied the shape from `DEFINITION_TO_WRITE_MODEL` directly above, which uses the same dict-keys-plus-lookup style. Left that one alone rather than widening this PR, but it has the same issue if you want it cleaned up separately.

PR #10096, opened 2026-07-31
fix(fingerprint): trust dependencies_complete for Jinja2 transforms

ogenstad:
# Why

Make the commit-id fold gate builder-aware instead of uniform. A Jinja2 transform with a complete dependency closure now produces a stable fingerprint without an explicit watch declaration, so unrelated commits no longer churn its fingerprint.

Jinja2 completeness is sound: it comes from a real transitive parse of the template include/import/extends graph, so a complete closure alone is enough. A dynamic or unresolvable reference still drops completeness to False and keeps the commit id folded. Python transforms and generators are unchanged - their completeness flag is only the package-directory floor, so a present watch remains required to omit the commit id.

Amends spec FR-016/FR-019 with a builder-aware note.

Ref: IFC-2952

## What changed

Made the fingerprint commit-id fold gate **builder-aware** so a Jinja2 transform with a
fully-resolved dependency closure no longer needs an explicit `watch:` declaration to get a
stable fingerprint.

Previously the gate was uniform: whenever `watch` was absent (`None`), the commit id was
folded into the fingerprint regardless of closure completeness, so a Jinja2 transform's
fingerprint churned on *every* commit - even when the change touched none of its real
dependencies. The only workaround was to declare a no-op `watch: {}`, which was pure
ceremony.

The gate now branches on the builder:

- **Jinja2** trusts `dependencies_complete` on its own. It comes from a real transitive parse
  of the template `include`/`import`/`extends` graph and is only `True` when every reference
  resolved to a tracked template, so a complete closure yields a stable, precise fingerprint
  with no `watch` needed. A dynamic or unresolvable reference drops completeness to `False`,
  which keeps the commit id folded.
- **Python transforms and generators** are unchanged. Their `dependencies_complete = True` is
  only the unsound package-directory floor (an out-of-directory import is silently missed), so
  a present `watch` remains the only signal trusted to omit the commit id.

### Files

- `backend/infrahub/git/fingerprint/composer.py` - `fold_commit_id` gains a `watch_required`
  flag; `compose_transformation` sets it from the input type (`False` for Jinja2, `True` for
  Python), and generators pass `True`.
- `backend/tests/unit/git/fingerprint/test_composer_transformation.py`,
  `test_watch_state.py` - parametrized unit coverage of the builder-aware truth table.
- `backend/tests/integration/git/test_fingerprint_transformation.py` - end-to-end test: an
  unrelated commit leaves a complete, watch-free Jinja2 transform's fingerprint stable while
  still changing a watch-free Python transform's fingerprint.
- `dev/specs/ifc-2844-definition-fingerprint/spec.md` - amendment notes on FR-016/FR-019.

### Safety

No under-regeneration is introduced. Python and generators are untouched, incomplete Jinja2
closures still fold the commit id, and an unresolved connected query still folds it too. No
consumer reads the fingerprint yet, so the only observable effect is that some Jinja2
fingerprints *stop* churning - a safe over-regeneration → stable transition.

## Checklist

- [x] Tests added/updated

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/integration/git/test_fingerprint_transformation.py
```diff
@@ -54,6 +54,33 @@ async def test_connected_query_edit_changes_transformation_fingerprint(
         after = (await client.get(kind=CoreTransformJinja2, name__value="person_with_cars")).fingerprint.value
         assert before != after
 
+    async def test_unrelated_commit_keeps_complete_jinja2_stable_but_folds_python(
+        self, repository_id: str, client: InfrahubClient, file_repo: FileRepo
+    ) -> None:
+        # Neither transform declares a watch. person_with_cars is a Jinja2 transform whose
+        # template has no includes, so its closure is complete; CarSpecMarkdown is a Python
+        # transform whose completeness is only the unsound package-directory floor.
```
gmazoyer: "unsound package-directory floor" I know it comes from the specs, but if I read this without knowing it, this does not mean anything to me.

gmazoyer (approved):

backend/tests/integration/git/test_fingerprint_transformation.py
ogenstad: Good point!

PR #10105, opened 2026-07-31
fix(attribute): persist inherited default-backed attribute updates

### LINKED ISSUE #9284: bug: Inherited default-backed attribute update reports success but does not persist
### Component

API Server / GraphQL

### Infrahub version

Enterprise 1.9.5

### Current Behavior

When a concrete node gains an inherited Dropdown attribute from a Generic through schema evolution, and the node is using the Generic schema default, the attribute is returned with `id: null` and `is_default: true`.

Updating that attribute through GraphQL returns `ok: true` and echoes the requested value in the mutation response. A subsequent read returns the original schema default again. No persisted attribute value is created.

Filtering also does not match the default-backed values. For example, a node whose `status.value` reads as `active` is not included in `status__value: "active"` results.

### Expected Behavior

When an update mutation sets an inherited default-backed attribute to a non-default value:

- The mutation persists an explicit attribute value on the node.
- A subsequent read returns the new value.
- The attribute has a real `id`.
- `is_default` becomes `false`.
- Attribute filters match the persisted value.

If updating inherited default-backed attributes on concrete nodes is unsupported, the mutation should fail with a validation error instead of returning `ok: true`.

### Steps to Reproduce

1. Start with a clean namespace, for example `Bugb`.

2. Load an initial schema that defines only a concrete node:

```yaml
---
version: "1.0"
nodes:
  - name: Server
    namespace: Bugb
    attributes:
      - name: name
        kind: Text
        unique: true
```

3. Create one `BugbServer` object before the inherited attribute exists:
```json
mutation {
  BugbServerCreate(data: { name: { value: "server-1" } }) {
    ok
    object { id name { value } }
  }
}
```

4. Load a second schema version that adds a Generic with a default-backed status attribute, then makes BugbServer inherit from it:
```yaml
---
version: "1.0"

generics:
  - name: TrackedThing
    namespace: Bugb
    attributes:
      - name: status
        kind: Dropdown
        default_value: active
        choices:
          - name: active
            label: Active
          - name: planned
            label: Planned

nodes:
  - name: Server
    namespace: Bugb
    inherit_from:
      - BugbTrackedThing
    attributes:
      - name: name
        kind: Text
        unique: true
```

5. Read the existing object. It now returns `status.value: active`, `status.id: null`, and `is_default: true`.

Update the existing object to set status to planned:

```json
mutation {
  BugbServerUpdate(
    data: {
      id: "SERVER_ID_FROM_PREVIOUS_READ"
      status: { value: "planned" }
    }
  ) {
    ok
    object {
      id
      status {
        id
        value
        is_default
      }
    }
  }
}
```

The mutation returns `ok: true` and echoes `status.value: planned`, but `status.id` is still `null` and `is_default` is still `true`.

Read the same object again. The value has not persisted: it returns `status.value: active`, `status.id: null`, and `is_default: true`.

Expected result: the update should create a persisted attribute row, return `status.value: planned`, assign a non-null `status.id`, and set `is_default: false`.

### Additional Information

Tested update paths that did not persist the new value:

Single and batched GraphQL update mutations.
infrahubctl object load.
Python SDK node update.
Profile attachment.

AI assisted a human in writing this bug report.

polmichel:
## What this does

An attribute that a node inherits from a generic and that carries a schema default now persists a written value even when the node predates the inheritance. Writing a non-default value to such an attribute materialises its database row on save, so the value survives a reload, gains a real id, and is no longer flagged as the default.

Closes #9284.

## Root cause (history)

The bug is a chain, and its origin is a previous fix:

- An attribute inherited from a generic lives on the generic, not on each concrete node. Nodes created **before** their schema started inheriting the generic therefore have **no database row** for the attribute.
- #7408 (`03f7…`, prep 1.10.0 era) made the add-attribute migration a **no-op for inherited attributes** to stop duplicate rows being created on nodes that already inherited the generic. That fix was correct for those nodes, but it removed the only path — the node-kind–matched migration — that used to materialise a row for **pre-existing** instances. Those nodes were left with no row.
- The attribute save path bailed out on any attribute without an id (`if not self.id: return None`), so a write to a rowless inherited attribute was silently dropped.

So #7408 fixed a duplicate-row bug for already-inheriting nodes and, as a side effect, opened this persistence gap for nodes that predate the inheritance. This branch closes the write side of that gap.

## The change

- `save()` now materialises the missing row on the first non-default write, then reconciles the value and `is_default` through the regular update path. Two guards keep it safe: it stays virtual when the value equals the default, and it does not create a second row when one already exists (e.g. an attribute rebuilt without its id).
- The materialisation query is **shared with the schema migration** rather than duplicated. `AttributeAddQuery` was relocated to `core/query` and generalised to optionally target specific node uuids; the ~100-line near-duplicate that the first cut of the fix introduced was removed.
- Two correctness fixes surfaced by code review: the changelog now reports the **schema default** as the previous value (it was being read back from the transient placeholder row), and materialisation matches the node by its **own kind** label, so it also reaches Profile / Template / Schema-namespace nodes rather than only `Node`-labelled ones.

## Known limitation — residual read-side gap

The fix restores **writes**, not **queryability at the default**. A node that predates the inheritance and still holds the default has no value row, so an attribute-value filter on that default does not match it until a value is written and the row is materialised. This is captured by an `xfail(strict=True)` test that asserts the desired behaviour, so it flips to a failure the day the gap is closed.

## Alternative considered — Solution B (root-cause fix)

Instead of materialising lazily on write, the structurally correct fix would act at the **migration / schema-evolution layer**: when a node starts inheriting a generic, materialise (or relabel) the rows for existing instances so the single generic-migration path covers them. That would:

- remove the need for any materialisation logic in `save()`, and
- close the residual queryability gap above (every node would carry a real default row).

It is heavier and touches the migration layer, so it was not taken here. This branch keeps the smaller, save-side fix and shares one query instead of duplicating it. Solution B is the recommended follow-up if the team wants default-value queryability restored.


backend/infrahub/core/query/attribute.py
```diff
@@ -508,3 +508,135 @@ async def default_attribute_query_filter(
...
+            MATCH (av:%(attr_value_label)s { value: $attr_value, is_default: true })
+            WHERE NOT av:AttributeValueIndexed
+            LIMIT 1
+            """ % {"attr_value_label": attr_value_label}
+
+        node_kinds_str = "|".join(self.node_kinds)
+        node_match = "MATCH (n:%(node_kinds_str)s)" % {"node_kinds_str": node_kinds_str}
+        if self.uuids:
+            self.params["node_uuids"] = self.uuids
+            node_match += "\n        WHERE n.uuid IN $node_uuids"
```
polmichel, line 598: This part has been modified in this query

backend/infrahub/core/query/attribute.py
```diff
@@ -508,3 +508,135 @@ async def default_attribute_query_filter(
...
+        )
+        """ % {
+            "match_query": match_query,
+            "branch_filter": branch_filter,
+            "node_match": node_match,
+            "uuid_generation": db.render_uuid_generation(node_label="a", node_attr="uuid"),
+        }
+
+        self.add_to_query(query)
+        self.return_labels = ["n.uuid", "a.uuid", "a"]
```
polmichel, line 641: and this one

backend/tests/component/core/test_attribute_default_value.py
```diff
@@ -0,0 +1,403 @@
...
+        db=db, schema="BugfilterServer", filters={"status__value": "active"}, branch=default_branch
+    )
+    assert non_matches == []
+
+
+@pytest.mark.xfail(
+    reason="A predating node still at its default has no value row, so a value filter does not match it until the "
+    "attribute is materialised by an explicit write. The save-side fix does not close this read-side gap.",
+    strict=True,
+)
```
polmichel, line 355: Still failing and would be fixed by the long-term approach I think

backend/infrahub/core/query/attribute.py
```diff
@@ -508,3 +508,135 @@ async def default_attribute_query_filter(
         query_params[f"{param_prefix}_{clean_filter_name}"] = filter_value
 
     return query_filter, query_params, query_where
+
+
+class AttributeAddQuery(Query):
```
polmichel, line 513: This query has been moved to this file. Modification are underlined in lower comments, or in the specific related commit changes

backend/infrahub/core/query/attribute.py
```diff
@@ -508,3 +508,135 @@ async def default_attribute_query_filter(
...
+        %(node_match)s
+        CALL (n) {
+            MATCH (:Root)<-[r:IS_PART_OF]-(n)
+            WHERE %(branch_filter)s
+            WITH n, r AS is_part_of_e
+            OPTIONAL MATCH (n)-[r:HAS_ATTRIBUTE]-(:Attribute { name: $attr_name })
+            WHERE %(branch_filter)s
+            WITH is_part_of_e, r AS has_attr_e
+            RETURN is_part_of_e, has_attr_e
+            ORDER BY has_attr_e.branch_level DESC, has_attr_e.from ASC, is_part_of_e.branch_level DESC, is_part_of_e.from ASC
```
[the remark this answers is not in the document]
polmichel: This ordering is pre-existing code: it was relocated verbatim from the add-attribute schema migration query and is not introduced by this PR.

backend/infrahub/core/query/attribute.py
```diff
@@ -508,3 +508,135 @@ async def default_attribute_query_filter(
...
+            LIMIT 1
+        }
+        WITH n, is_part_of_e, has_attr_e, av, is_protected_value
+        WHERE is_part_of_e.status = "active" AND (has_attr_e IS NULL OR has_attr_e.status = "deleted")
+        CREATE (a:Attribute { name: $attr_name, branch_support: $branch_support })
```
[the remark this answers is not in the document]
polmichel: This is pre-existing behaviour of the shared add-attribute query (the same non-atomic check-then-create the schema migration already relied on); it is not introduced by this PR.

backend/infrahub/core/attribute.py
```diff
@@ -377,17 +378,72 @@ def _normalize_value(self, value: Any) -> Any:
...
+    def _value_diverges_from_default(self) -> bool:
+        """Return True when the in-memory value differs from the schema default.
+
+        A missing id is ambiguous: it also covers attributes that were not fetched in a partial read or
+        rebuilt without their id, both of which stay at their default. Only a value that diverges from the
+        default marks content a caller actually intends to persist.
+        """
+        current = self.value.value if isinstance(self.value, Enum) else self.value
+        if self.schema.default_value is not None:
+            return self.schema.default_value != current
```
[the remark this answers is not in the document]
polmichel: Pre-existing code: the previous is_default logic already compared against the raw, un-normalized schema default. This change only relocated that comparison into a helper.

ajtmccarty (changes requested):
I don't think we should do this change. the root cause is that the schema migrations are not working correctly when updating inheritance, so this should be fixed in the schema migrations. the PR description mentions this approach, but rejects it as being heavier. I think that is wrong. We should prioritize doing this one the right way next week. Pol, do not even think about trying to work on this on vacation

ajtmccarty:
replaced by #10113 and it's upcoming follow-up

PR #10106, opened 2026-07-31
perf(diff): batch node field summary retrieval over paged queries

ajtmccarty:
Encountered an issue with the query to retrieve diff summaries of nodes when running merge performance testing locally on a very large branch. Could be that it was just too much data for my machine to handle, but better safe than sorry and this batching will cost us very little.

# Why

Retrieving the node field summaries for a diff currently runs one query that aggregates the changed field names and node UUIDs of every changed node per kind in a single transaction. On large diffs (tens of thousands of changed nodes) that aggregation does not scale and can exhaust the Neo4j heap (`Java heap space` errors during merge/rebase validation).

This PR pages the retrieval over the changed nodes so memory use per query is bounded by `database.query_size_limit`, regardless of diff size.

Non-goals: no change to what the summaries contain or how consumers (schema-constraint validation) use them.

## What changed

Behavioral changes:

- `DiffRepository.get_node_field_summaries()` now issues one query per page of changed nodes instead of a single aggregate query. Results are identical; only the retrieval strategy changed.

Implementation notes:

- `EnrichedDiffNodeFieldSummaryQuery` now selects one page of changed nodes (`SKIP`/`LIMIT` over a strict `ORDER BY n.uuid, elementId(n)`) and collects each node's altered attribute/relationship names per row, rather than aggregating across the whole diff per kind.
- `get_node_field_rows()` yields one single-node `NodeDiffFieldSummary` per row; the repository folds them into per-kind summaries via the new `NodeDiffFieldSummary.merge()`.
- The `DIFF_HAS_NODE` match stays `OPTIONAL` so a diff root with no changed nodes still produces a (node-less) row. Those rows are skipped when yielding summaries; this cannot terminate pagination early because nulls sort after every real node in the `ORDER BY`.
- Pagination stops when a page yields fewer nodes than the limit.

What stayed the same: the shape of `NodeDiffFieldSummary` (per-field node-UUID maps and derived views) and the repository's public API are unchanged.

## How to review

The part deserving the most scrutiny is the pagination-termination logic: the loop breaks when a page yields fewer rows than `query_size_limit`, and the node-less rows from the `OPTIONAL MATCH` occupy page slots without being counted. This is safe only because nulls sort last in the query's `ORDER BY` — see the `get_node_field_rows()` docstring.

## Impact & rollout

- **Backward compatibility:** none — no API, schema, or data changes.
- **Performance:** bounds per-query memory on large diffs; adds one extra round trip per `query_size_limit` changed nodes.

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../.agents/skills/creating-changelog-entries/SKILL.md) added (`uv run towncrier create ...`)

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/diff/query/field_summary.py
+++ b/backend/infrahub/core/diff/query/field_summary.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from collections.abc import Generator
 from typing import Any
 
 from infrahub.core.constants import DiffAction
@@ -8,16 +8,13 @@ from infrahub.database import InfrahubDatabase
 from ..model.path import NodeDiffFieldSummary, TrackingId
 
 
-@dataclass
-class FieldNodeUuidsRow:
-    """One changed field of a kind and the uuids of the nodes that changed it, as projected by the query."""
-
-    name: str
-    node_uuids: list[str]
-
-
 class EnrichedDiffNodeFieldSummaryQuery(Query):
-    """Get node kind and names of all altered attributes and relationships for each kind."""
+    """Get the names of all altered attributes and relationships for one page of changed nodes.
+
+    Pagination is over the changed nodes, strictly ordered, with each row carrying every altered
+    field name of one node; aggregating the fields of every node in one transaction does not scale
+    with large diffs.
+    """
 
     name = "enriched_diff_node_field_summary"
     type = QueryType.READ
@@ -25,12 +22,16 @@ class EnrichedDiffNodeFieldSummaryQuery(Query):
     def __init__(
         self,
         diff_branch_name: str,
+        node_offset: int,
+        node_limit: int,
         tracking_id: TrackingId | None = None,
         diff_id: str | None = None,
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
         self.diff_branch_name = diff_branch_name
+        self.node_offset = node_offset
+        self.node_limit = node_limit
         self.tracking_id = tracking_id
         self.diff_id = diff_id
 
@@ -42,6 +43,8 @@ class EnrichedDiffNodeFieldSummaryQuery(Query):
             "diff_branch": self.diff_branch_name,
             "tracking_id": self.tracking_id.serialize() if self.tracking_id else None,
             "diff_id": self.diff_id,
+            "node_offset": self.node_offset,
+            "node_limit": self.node_limit,
         }
         query = """
         MATCH (diff_root:DiffRoot)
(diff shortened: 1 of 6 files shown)


backend/infrahub/core/diff/query/field_summary.py
```diff
@@ -51,48 +54,40 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
+        produces one node-less row instead; it is skipped here, which is safe for the caller's
+        consumed-node count because null nodes sort after every real node.
+        """
+        for result in self.get_results():
+            kind = result.get_as_str("kind")
+            node_uuid = result.get_as_str("node_uuid")
+            if not kind or not node_uuid:
+                continue
+            summary = NodeDiffFieldSummary(kind=kind)
+            for attr_name in result.get_as_type(label="attr_names", return_type=list[str]):
```
gmazoyer: Is it better than using `get_as_list_of_type(return_type=str)`?

backend/infrahub/core/diff/query/field_summary.py
```diff
@@ -8,29 +8,30 @@
...
+    """
 
     name = "enriched_diff_node_field_summary"
     type = QueryType.READ
 
     def __init__(
         self,
         diff_branch_name: str,
+        node_offset: int,
+        node_limit: int,
```
gmazoyer: It there a reason not to use the base class `self.limit`/`self.offset`?

gmazoyer (approved):

backend/infrahub/core/diff/query/field_summary.py
ajtmccarty: no there is not. and this mistake would actually cause a weird double-execution when the query tries to execute with `query_with_size_limit()` b/c `self.limit` and `self.offset` are both `None`. so thank you for pointing it out

backend/infrahub/core/diff/query/field_summary.py
ajtmccarty: I forgot this option existed. updating to use it

backend/infrahub/core/diff/query/field_summary.py
```diff
@@ -51,48 +54,40 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: Any) -> None:  # noqa
...
-        WITH DISTINCT diff_root, n.kind AS kind
-        CALL (diff_root, kind) {
-            OPTIONAL MATCH (diff_root)-[:DIFF_HAS_NODE]->(n:DiffNode {kind: kind})-[:DIFF_HAS_ATTRIBUTE]->(a:DiffAttribute)
-            WHERE n.action <> $unchanged_str
-            AND a.action <> $unchanged_str
-            WITH a.name AS attr_name, collect(DISTINCT n.uuid) AS attr_node_uuids
-            WHERE attr_name IS NOT NULL
-            RETURN collect({name: attr_name, node_uuids: attr_node_uuids}) AS attr_name_uuids
+        WITH n
+        ORDER BY n.uuid, elementId(n)
```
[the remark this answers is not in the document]
ajtmccarty: the memory cost that caused this query to throw the out-of-memory error will not be triggered by looking at the UUID and elementId of every `:DIFF_NODE` in a given `:DIFF_ROOT`, even if there are 100,00. so this is not a concern. this is the pattern that we follow elsewhere. maybe someday we will switch to using cursors for queries, but this is not the place to make that change

backend/infrahub/core/diff/repository/repository.py
```diff
@@ -588,11 +588,32 @@ async def get_affected_node_uuids(
...
+                num_nodes += 1
+                if not node_summary.attribute_node_uuids and not node_summary.relationship_node_uuids:
+                    continue
+                kind_summary = summaries_by_kind.setdefault(
+                    node_summary.kind, NodeDiffFieldSummary(kind=node_summary.kind)
+                )
+                kind_summary.merge(node_summary)
+            if num_nodes < node_limit:
+                break
+            node_offset += node_limit
```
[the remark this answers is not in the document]
ajtmccarty: same as reported below

PR #10121, opened 2026-08-04
fix(graphql): retry transient database errors on object create

### LINKED ISSUE #10119: bug: object create is the only CRUD mutation without transient-error retry, so a raw Neo4j "no available threads" error reaches users
### Component

API Server / GraphQL

### Infrahub version

1.10.6

### Current Behavior

`mutate_create` is the only CRUD mutation on `InfrahubMutationMixin` that is **not** wrapped in `@retry_db_transaction`. Under database contention, a transient Neo4j error during an object create is not retried and is surfaced verbatim to the caller. In the UI this shows up as a raw driver message on an otherwise ordinary "create object" action:

```
there are no available threads to serve this request at the moment, you can retry at a
later time or consider increasing max thread pool size for bolt connectors
```

The asymmetry is visible in `backend/infrahub/graphql/mutations/main.py`:

| Mutation | Line | Decorator |
| --- | --- | --- |
| `mutate_create` | 248-249 | **none** |
| `mutate_update` | 321-323 | `@retry_db_transaction(name="object_update")` |
| `mutate_upsert` | 389-391 | `@retry_db_transaction(name="object_upsert")` |
| `mutate_delete` | 519-521 | `@retry_db_transaction(name="object_delete")` |

The IPAM mutations already retry their creates, which makes the generic path the odd one out:

- `backend/infrahub/graphql/mutations/ipam.py:143` — `@retry_db_transaction(name="ipaddress_create")`
- `backend/infrahub/graphql/mutations/ipam.py:306` — `@retry_db_transaction(name="ipprefix_create")`

So under the same load, an update, an upsert or a delete recovers transparently, and only a create fails in the user's face.

Note the create body is *already* executed under retry on one path: `mutate_upsert` is decorated and calls `cls.mutate_create` directly (`main.py:459`). The behaviour therefore differs purely by entry point — the same code retries when reached through `Upsert` and does not when reached through `Create`.

### Expected Behavior

A transient database error during an object create is retried with the same exponential backoff as every other CRUD mutation, and only surfaces to the user if it is still failing after `retry_limit` attempts (default 3). Creating an object should be no more fragile under load than updating one.

### Steps to Reproduce

Reproducing this requires Neo4j to run out of bolt threads while a create is in flight.

1. Start Infrahub against a Neo4j with a deliberately small bolt thread pool. The existing component test does exactly this — see `backend/tests/component/graphql/test_graphql_read_retry.py`, which runs Neo4j with `NEO4J_server_bolt_thread__pool__max__size: "10"` and confirms the resulting error arrives as a `neo4j.exceptions.TransientError`.
2. Drive enough concurrent GraphQL traffic to saturate that pool.
3. Issue a `<Kind>Create` mutation while the pool is saturated → the raw "no available threads" message is returned to the client.
4. Issue a `<Kind>Update`, `<Kind>Upsert` or `<Kind>Delete` under the same conditions → the request succeeds after a retry.

The same asymmetry is reachable without a thread-pool ceiling by injecting a `TransientError` into the create path.

### Additional Information

**Root cause.** `retry_db_transaction` (`backend/infrahub/database/__init__.py:571-609`) catches `neo4j.exceptions.TransientError` and retries with exponential backoff, incrementing the `TRANSACTION_RETRIES` counter under the label it is given. `mutate_create` never opts in, so the error escapes to the GraphQL layer unchanged.

**Real-world impact.** This was hit during a support engagement on a production-scale deployment. The driver connection pools across the deployment's server and task-worker processes can collectively demand more bolt threads than Neo4j is configured to serve (a default 100-connection pool per process against Neo4j's default 400 bolt threads adds up quickly once you have several server pods with multiple workers each, plus task workers). Under that pressure, users creating objects in the UI saw the raw Neo4j message, while updates and deletes under the identical load recovered silently. The result reads as "Infrahub is broken" rather than "the database is briefly busy", and it is not actionable for the end user.

Increasing Neo4j's bolt thread pool or lowering the per-process connection pool is a sensible deployment-side mitigation, but the missing retry is a product defect in its own right: transient contention should not be user-visible on one CRUD verb and invisible on the other three.

**Proposed fix.** Add `@retry_db_transaction(name="object_create")` to `mutate_create`, applied exactly as on `mutate_update` (below `@classmethod`). This is retry-safe:

- `create_node` (`backend/infrahub/core/node/create.py:358`) wraps the node writes in `db.start_transaction()`, so a `TransientError` rolls the attempt back and the retry starts from a clean slate.
- File uploads are not re-run: `_process_file` executes in `mutate()` (`main.py:176`) *before* `mutate_create` is called, so it sits outside the retried scope.
- Mutation events are emitted after `mutate()` returns (`main.py:238`), also outside the retried scope.
- The upsert and IPAM-create paths already retry this same work, so no new class of side effect is being placed under retry.

The distinct metric label (`object_create`) also makes create-side contention visible in `TRANSACTION_RETRIES` for the first time, alongside the existing `object_update` / `object_upsert` / `object_delete` series.

BeArchiTek:
Fixes #10119

## Problem

`mutate_create` was the only CRUD mutation on `InfrahubMutationMixin` not wrapped in `@retry_db_transaction`:

| Mutation | Decorator before this PR |
| --- | --- |
| `mutate_create` | **none** |
| `mutate_update` | `@retry_db_transaction(name="object_update")` |
| `mutate_upsert` | `@retry_db_transaction(name="object_upsert")` |
| `mutate_delete` | `@retry_db_transaction(name="object_delete")` |

The IPAM mutations already retry their creates (`ipaddress_create`, `ipprefix_create` in `graphql/mutations/ipam.py`), which left the generic create path as the sole outlier.

The practical effect is that a transient Neo4j error during an object create is handed to the user verbatim:

```
there are no available threads to serve this request at the moment, you can retry at a
later time or consider increasing max thread pool size for bolt connectors
```

Under identical load, an update, upsert or delete recovers transparently. This came out of a support engagement on a production-scale deployment where the aggregate driver connection pools across server and task-worker processes outran Neo4j's configured bolt thread count. Increasing the bolt thread pool is a reasonable deployment-side mitigation, but the inconsistency is a defect on its own: transient contention should not be user-visible on exactly one CRUD verb.

## Change

One line in `backend/infrahub/graphql/mutations/main.py`, applied exactly as on `mutate_update` (below `@classmethod`):

```python
@classmethod
@retry_db_transaction(name="object_create")
async def mutate_create(
```

## Why this is retry-safe

- `create_node` (`core/node/create.py`) wraps the node writes in `db.start_transaction()`, so a `TransientError` rolls the attempt back and the retry starts from a clean slate.
- File uploads are not repeated: `_process_file` runs in `mutate()` *before* `mutate_create` is called, so it sits outside the retried scope.
- Mutation events are emitted after `mutate()` returns, also outside the retried scope.
- The create body already runs under retry today on another path — `mutate_upsert` is decorated and calls `cls.mutate_create` directly. This PR makes the two entry points behave the same instead of differing by how the create was reached.

Two notes for reviewers, both pre-existing rather than introduced here:

- **Nested retry on the upsert path.** `mutate_upsert` (retried) calling `mutate_create` (now retried) means the create leg can attempt up to `retry_limit²` times. This already happens for IPAM kinds, whose generated `Upsert` inherits the retried `mutate_upsert` and dispatches to an already-retried `mutate_create`. Worth flagging if you'd rather bound it, but it is not new behaviour.
- **Post-commit profile application.** `create_node` applies profiles and re-saves after the transaction commits. A transient error in that narrow window would be retried and could duplicate the node. That window is identical on the already-retried upsert and IPAM-create paths, so this PR neither introduces nor widens it.

## Tests

New `backend/tests/unit/graphql/test_mutation_retry.py`:

- `test_mutate_create_retries_transient_error` — drives `mutate_create` with `create_node` raising `TransientError` once, asserts the call succeeds on the second attempt and that `TRANSACTION_RETRIES{name="object_create"}` incremented. This also pins the metric label.
- `test_crud_mutations_are_retry_wrapped` — parametrized over all four CRUD mutations, so none can silently lose its retry again.

Verified red/green: both new tests fail with the decorator removed and pass with it. Full run of `backend/tests/unit/database/test_retry_db_transaction.py` plus the new file is 14 passed. `ruff check` and `ruff format --check` are clean on both touched files.

## Release line

Targeted at `stable` so the fix ships on the released line and merges forward into `release-1.11` with the usual sync.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10121?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/unit/graphql/test_mutation_retry.py
fatih-acar, line 1: you can remove this I think
we dont want magicmock etc

fatih-acar (approved):

backend/tests/unit/graphql/test_mutation_retry.py
```diff
@@ -0,0 +1,48 @@
+"""Unit tests for transient-error retry on the CRUD mutations of `InfrahubMutationMixin`."""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, patch
```
ogenstad: Can you rewrite this test so that it doesn't require any Async or MagicMock as per https://github.com/opsmill/infrahub/blob/infrahub-v1.10.6/.agents/rules/testing-python.md?plain=1#L20-L35

ogenstad (changes requested):

backend/tests/unit/graphql/test_mutation_retry.py
BeArchiTek: Rewritten in c1399836a1 — the file now imports nothing from `unittest.mock` and has no async test.

The behavioral retry test was duplicate coverage: `backend/tests/unit/database/test_retry_db_transaction.py` already covers retry-on-`TransientError`, backoff, exhaustion and the metric increment against the decorator itself. The only thing the mutation-layer test added was proof that `mutate_create` is wired to the decorator, and that needed no mocks — it only needed the wiring to be observable.

So `retry_db_transaction` now records its metric name on the wrapper it returns, and the test asserts the exact mapping:

```python
EXPECTED_RETRY_METRIC_NAMES = {
    "mutate_create": "object_create",
    "mutate_update": "object_update",
    "mutate_upsert": "object_upsert",
    "mutate_delete": "object_delete",
}
```

Mock-free behavioral coverage at the mutation layer was not reachable without patching: `create_node` is a module-level import in `mutations/main.py`, not an injected collaborator, so there is no seam to hand a test double to. Refactoring `InfrahubMutationMixin` for injection felt out of proportion to a one-line decorator fix — happy to file that separately if you think it is worth doing.

backend/tests/unit/graphql/test_mutation_retry.py
BeArchiTek: Removed in c1399836a1 — the mock-based test is gone and the file no longer imports `unittest.mock` at all. What remains is a mock-free assertion that all four CRUD mutations are retry-wrapped, under the right metric label.

backend/tests/unit/graphql/test_mutation_retry.py
```diff
@@ -0,0 +1,48 @@
...
+
+    assert obj is created_node
+    assert create_node_mock.await_count == 2
+    assert TRANSACTION_RETRIES.labels("object_create")._value.get() == initial_retries + 1
+
+
+@pytest.mark.parametrize("mutation_name", CRUD_MUTATIONS)
+def test_crud_mutations_are_retry_wrapped(mutation_name: str) -> None:
+    """Every CRUD mutation stays wrapped in `retry_db_transaction`, so none regresses to no retry."""
+    assert hasattr(getattr(InfrahubMutationMixin, mutation_name), "__wrapped__")
```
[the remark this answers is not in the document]
BeArchiTek: Valid — `hasattr(..., "__wrapped__")` passed for any `functools.wraps` decorator, so it proved neither that `retry_db_transaction` was the wrapper nor that the metric label was right.

Fixed in c1399836a1 with the explicit marker you suggested: `retry_db_transaction` records its metric name on the wrapper, and the test asserts the exact `{method: metric_name}` mapping for all four CRUD mutations with full dict equality.

Verified the guard discriminates — undecorated methods on the mixin (`mutate`, `mutate_update_object`, `_call_mutate_update`) report `None`, so removing a decorator fails the test.

Behavioral retry coverage stays where it belongs, on the decorator itself, in `backend/tests/unit/database/test_retry_db_transaction.py`. Adding per-mutation behavioral retry tests would have meant patching module-level imports, which this repository's testing rules disallow.

backend/infrahub/database/__init__.py
```diff
@@ -599,6 +599,9 @@ async def wrapper(*args: Any, **kwargs: Any) -> R:
                         break
             raise error
 
+        # Exposes the metric name so callers can assert a function is retry-wrapped and under which label.
+        wrapper.retry_db_transaction_name = name  # type: ignore[attr-defined]
```
ogenstad: We're trying to improve the typing throughout the project, this line adds a type ignore that's not even used within the app. It feels like a workaround for the test and should be removed.

A test like below should work without touching the production code:

```python
from __future__ import annotations

import ast
import inspect

from infrahub.graphql.mutations import main

EXPECTED_RETRY_LABELS = {
    "mutate_create": "object_create",
    "mutate_update": "object_update",
    "mutate_upsert": "object_upsert",
    "mutate_delete": "object_delete",
}


def test_crud_mutations_are_retry_wrapped() -> None:
    """Every CRUD mutation retries transient database errors, under its own metric label."""
    class_def = next(
        node
        for node in ast.parse(inspect.getsource(main)).body
        if isinstance(node, ast.ClassDef) and node.name == "InfrahubMutationMixin"
    )

    retry_labels = {
        method.name: keyword.value.value
        for method in class_def.body
        if isinstance(method, (ast.FunctionDef, ast.AsyncFunctionDef))
        for decorator in method.decorator_list
        if isinstance(decorator, ast.Call)
        and isinstance(decorator.func, ast.Name)
        and decorator.func.id == "retry_db_transaction"
        for keyword in decorator.keywords
        if keyword.arg == "name" and isinstance(keyword.value, ast.Constant)
    }

    assert retry_labels == EXPECTED_RETRY_LABELS
```

ogenstad (changes requested):

backend/infrahub/database/__init__.py
BeArchiTek: Removed in 233f2348be — the marker and its `type: ignore` are gone, `database/__init__.py` is no longer in the diff at all, and the production change is back to the single decorator line.

Took your test, with one change: factored the decorator/keyword walk out into a `_retry_label()` helper. The 6-clause comprehension (3 `for`, 3 `if`) states the yielded value before the clauses that explain it, and the repo already has a house style for this kind of structural test — `backend/tests/unit/workflows/test_flow_session_convention.py` uses module-level `_`-prefixed helpers with explicit loops for the same `ast` + `inspect.getsource` technique. Kept your scoping decision (`class_def.body`, not `ast.walk`, so it does not recurse into nested defs) and the exact-dict assertion.

```python
def _retry_label(method: ast.FunctionDef | ast.AsyncFunctionDef) -> str | None:
    for decorator in method.decorator_list:
        if not (
            isinstance(decorator, ast.Call)
            and isinstance(decorator.func, ast.Name)
            and decorator.func.id == "retry_db_transaction"
        ):
            continue
        for keyword in decorator.keywords:
            if keyword.arg == "name" and isinstance(keyword.value, ast.Constant):
                return keyword.value.value
    return None


def test_crud_mutations_are_retry_wrapped() -> None:
    """Every CRUD mutation retries transient database errors, under its own metric label."""
    class_def = next(
        node
        for node in ast.parse(inspect.getsource(main)).body
        if isinstance(node, ast.ClassDef) and node.name == "InfrahubMutationMixin"
    )
    methods = [node for node in class_def.body if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))]

    retry_labels = {method.name: label for method in methods if (label := _retry_label(method)) is not None}

    assert retry_labels == EXPECTED_RETRY_LABELS
```

Checked that it still fails on a real regression: parsing a copy of the source with the decorator stripped drops `mutate_create` from the dict, and a label typo (`object_delete` → `object_remove`) mismatches — both fail the assertion. `ruff` and `mypy` clean, no `type: ignore` needed anywhere.

BeArchiTek:
Separate from the review feedback — one consequence of this change worth a decision before merge.

`mutate_upsert` is itself retry-wrapped (`object_upsert`) and calls `cls.mutate_create` directly at `main.py:460`. Now that `mutate_create` carries the decorator too, the upsert-as-create path has nested retry: the inner wrapper exhausts `retry_limit` attempts, re-raises the `TransientError`, and the outer wrapper retries the whole upsert — inner create included.

With the defaults (`retry_limit=3`, `retry_base_delay=0.1`, `retry_max_delay=2.0`):

| | before | after |
|---|---|---|
| worst-case create attempts | 3 | 9 |
| backoff before final failure | ~0.7-1.0s | ~3-4s |
| `TRANSACTION_RETRIES` | `object_upsert` only | `object_create` + `object_upsert` for one failure sequence |

The plain create path is unaffected — `mutate` (`main.py:177`) carries no decorator, so a direct `<Kind>Create` gets exactly one layer, same as update and delete.

What makes me raise it rather than shrug: 9 attempts against a Neo4j already reporting "there are no available threads" adds load in exactly the saturation scenario this PR is meant to handle gracefully.

The update path already avoids this by construction — `_call_mutate_update` calls the undecorated `mutate_update_object` (`main.py:302`), not the wrapped `mutate_update`, so the decorator stays on the public entry point only. The symmetric fix for create is to extract the `mutate_create` body into an undecorated classmethod, leave `mutate_create` as the decorated public entry, and have `mutate_upsert` call the inner one — roughly 10 lines.

Happy to do that here, or leave it and file it separately. Deliberately not changing it unilaterally since it is production behaviour beyond the one-line gap this PR set out to close.

BeArchiTek:
Reworked in d76cb0c51f. The one-line version I pushed earlier had a real defect, so the fix is now larger than the single decorator line you both reviewed — flagging the reasoning rather than sliding it in.

## The one-liner could commit a duplicate node

`create_node` commits its own transaction (`core/node/create.py:358`), and `mutate_create` then builds the GraphQL response from reads that run *after* that commit — `build_graphql_response` → `obj.to_graphql(db=db)`. With `@retry_db_transaction` on `mutate_create`, a `TransientError` from those reads was caught by the wrapper, which replayed the whole method including `create_node`:

- no uniqueness constraint → a second node is committed silently, and the first never reaches `emit_node_mutation_events`, so no webhook, no computed-attribute recompute and no trigger fires for it
- HFID or uniqueness constraint → attempt 2 raises "already exists" for an object the same request just created
- pool-backed field → attempt 1's allocation is committed and attempt 2 allocates a second one

That is the property separating create from the three methods the decorator was written for: `_call_mutate_update` keeps the response build *inside* the transaction (`main.py:309-320`), so replaying it is safe.

## Fix: the retry wraps only the creation transaction

`_create_object_with_retry` carries the decorator and wraps nothing but `create_node`; the response build sits outside the retried scope. This is the placement the decorator historically had — before `35b7f08e10` ("Support converting repository type", Sep 2025) it was on `mutate_create_object`, whose body was the create alone. That refactor collapsed the three create methods into one and dropped it; my one-liner re-added it at a wider scope than it ever had.

## Second case: a caller-supplied transaction

`proposed_change.py:118` and `action.py:109` both open `db.start_transaction()` and pass `database=dbt` into `mutate_create`. Retrying there is actively harmful: the transient error has already failed that transaction, so replaying on it raises `TransactionError`, whose MRO is `TransactionError → DriverError → GqlError` — not caught by `except (TransientError, ClientError)`. The client would get "Transaction failed" and the caller's own retry would never see a retryable error. So `mutate_create` runs the creation undecorated when `db.is_transaction`, leaving the retry to whoever can actually roll back and reopen.

## Known residual, not fixed here

`create_node` does post-commit work of its own at `create.py:376-379` (`get_profile_ids` / `apply_profiles` / `obj.save(db=db)`), outside its transaction. A transient error there still replays the creation, so the duplicate-node window is narrowed, not closed. Pre-existing on the upsert path, which has wrapped `create_node` plus that tail all along. Left alone because closing it changes `create_node` for every caller — happy to file it.

The nested-retry point from my earlier comment still stands, and is worse than I first measured: `CoreNumberPoolUpsert` routes upsert → `resource_manager_create` → `object_create`, so 3³ = 27 create attempts, each re-taking the per-kind lock. A re-entrancy guard in `retry_db_transaction` so only the outermost wrapper retries is the real fix, and that helper has ~40 call sites — its own change.

Verified: 61 component tests in `test_mutation_create.py` + `test_mutation_upsert.py` pass against a throwaway Neo4j, unit test passes, `ruff` and `mypy` clean.

ogenstad (changes requested):
Reviewed the rework. Both defects you found in the one-liner check out: `create_node` does commit at `create.py:520-555` before `build_graphql_response` runs, and `TransactionError`'s MRO is `DriverError → GqlError`, so it would have escaped `except (TransientError, ClientError)` and surfaced as "Transaction failed". The `is_transaction` guard is earning its place for `CoreProposedChangeCreate`, which stays at 3 attempts instead of 9.

Three things I'd like to see in this PR, then an answer on your nested-retry question.

**Close the post-commit window on the Create path.** You listed this as a known residual, pre-existing on upsert, which is accurate. I'd upgrade it to a fix here, because for `<Kind>Create` it isn't pre-existing: that path had no retry at all before, so `create.py:557-562` (`_has_profiles_set` → `apply_profiles` → `obj.save`) could never be replayed. Now it can, and the result is a silently duplicated profiled node whose first copy never reaches `emit_node_mutation_events`. Same class of failure you correctly rejected the one-liner for, narrower window, on the one path this PR set out to improve. The fix looks like moving those lines inside the `async with db.start_transaction() as dbt` branch. The `is_transaction` branch already runs that tail on the caller's transaction, so it makes the two consistent rather than changing anything for transaction-passing callers. The comment above it explains why no read-back is needed, not why it sits outside the transaction, so the placement reads incidental. Worth confirming with whoever wrote it.

One approach I'd steer away from: having `mutate_create` open the transaction itself and pass it down to mirror `_call_mutate_update`. It inverts the lock/transaction ordering, since `create_node` takes `InfrahubMultiLock` internally, and you'd end up committing after releasing the lock.

**The docstring on `_create_object_with_retry` states an invariant the code doesn't hold.** "The retried scope must stay inside the creation transaction" isn't true while the scope includes `create_node`'s post-commit tail. Fixing the above makes it true.

**The test docstring claims more than the test checks.** `EXPECTED_RETRY_LABELS` asserts a decorator sits on a private helper. Nothing asserts `mutate_create` still calls it, so deleting the `else:` at `main.py:302-305` leaves the test green while creates lose their retry. The `ast` approach is fine and the mock-free constraint is real, I'd just have the docstring claim decorator-presence-and-labels rather than implying create-is-retried coverage.

### On the nested retry

Worth acting on, and I think there's a sharper argument than attempt count. The inner wrapper restarts its backoff from `retry_base_delay` on each outer attempt, so nesting doesn't buy a longer retreat, it crams more load into roughly the same window:

- nested 3×3: 9 attempts across ~2.8-4.0s of sleep
- flat `retry_limit=9`: 9 attempts across ~11s (0.1, 0.2, 0.4, 0.8, 1.6, then capped at 2.0)

If 9 attempts were wanted, raising `retry_limit` would be strictly better than the nesting. That makes "more chances to recover" hard to defend, which is usually the shape of the counter-argument. Each attempt also re-takes the per-kind lock, global on the default branch, so it lengthens the saturation it's meant to ride out. Your `TRANSACTION_RETRIES` point compounds it: one failure sequence increments `object_create`, `object_upsert`, and for number pools `resource_manager_create`, so those labels stop being additive for anyone alerting on them.

Traced the paths, your counts hold including the 27:

| Path | Before | After |
|---|---|---|
| `<Kind>Create` | 1 | 3 (the fix) |
| `<Kind>Upsert` create leg | 3 | 9 |
| `CoreNumberPoolCreate` | 3 | 9 |
| `CoreNumberPoolUpsert` create leg | 9 | 27 |
| `CoreProposedChangeCreate` | 3 | 3 |
| IPAM | unchanged | unchanged |

Your earlier "symmetric fix" idea won't work here: `mutate_upsert:495` calls `cls.mutate_create` specifically so subclass overrides apply, which is how number-pool range validation gets in, so bypassing it loses that dispatch. The re-entrancy guard is the right answer. I'd push back gently on the "~40 call sites" caution though: that's blast radius, not edit size, and the radius is one-directional. A `ContextVar` plus an early `return await func(...)` when a retry is already active can only reduce attempts, never increase them. It also fixes the pre-existing 9 on IPAM upsert and number-pool create.

The one thing to test carefully is contextvar propagation into spawned tasks. A `ContextVar` is copied at task creation, so concurrent work started inside a retried scope inherits `True` and skips its own retry loop. Defensible under "outermost owns the retry", but I'd confirm it's what you want for the fan-out inside `bulk_merge_*` and `enriched_diff_*` before this reaches `stable`.

### Separate tickets

Both pre-existing and out of scope, worth filing so the `is_transaction` guard here doesn't read as a one-off: the update path has the same hazard unguarded (`action.py:132` calls the decorated `mutate_update` with `database=dbt`), and the wrapper sleeps before checking exhaustion (`database/__init__.py:619-622`), burning a pointless final delay that nesting multiplies.

PR #10128, opened 2026-08-04
refactor: derive the current branch from the URL instead of syncing an atom

bilalabbad:
## What changed and why

`BranchesProvider` held the selected branch in two places — the `branch` query string parameter and `currentBranchAtom` — and kept them in step with an effect. Two representations of one fact meant the provider had to block rendering while they disagreed, and the branch selector had to write both on every switch.

The query string parameter is now the single source of truth. `currentBranch` is derived during render:

```ts
const currentBranch = branches ? findSelectedBranch(branches, branchInQueryString) : null;
```

`currentBranchAtom` is deleted. `setCurrentBranch` stays on the context but writes the query string parameter, so the "default branch means no parameter" convention lives in one place instead of being re-derived by each caller.

## Fixes a latent bug

The removed render guard compared the resolved branch against a hardcoded `"main"`:

```ts
if (currentBranch?.name !== (branchInQueryString ?? DEFAULT_BRANCH_NAME)) return <InfrahubLoading />;
```

The branch itself is resolved by `is_default`, but `INFRAHUB_INITIAL_DEFAULT_BRANCH` lets a deployment name its default branch anything. On such a deployment the comparison never succeeded and the app never rendered past "Loading branches...". Deriving the branch removes the comparison entirely.

Five other hardcoded `"main"` checks remain out of scope here (`shared/config/config.ts`, `branches-to-select-options.ts`, `branches-table.tsx`, `diff/ui/checks/data-conflict.tsx`, and `DEFAULT_BRANCH_NAME` itself) — `config.ts` still falls back to `/graphql/main`, so those deployments need a follow-up.

## Notable implementation details

- The effect survives, reduced to the side effects it always owned: the not-found toast and `navigate("/")`. It no longer writes state, so it cannot re-trigger itself.
- `isFieldDisabled` was reaching into the global Jotai store for one boolean. It now takes `isDefaultBranch`, threaded from `getFormFieldsFromSchema` through `getFormFieldFromAttribute` to the 8 form call sites, and is a pure function again. The relationship builder passes `permissions: undefined`, so the branch bit is unreachable there and is not threaded.
- Four tests dropped `store.set(currentBranchAtom, ...)` in favour of a plain argument, and the component-test wrapper no longer needs `useState` to fake the branch.

## Testing performed

```
pnpm exec biome ci .    ✅
pnpm exec betterer ci   ✅ 190 issues (baseline unchanged)
pnpm knip               ✅
pnpm test               ✅ 166 files, 1139 tests
```

Behaviour verified by hand against the existing e2e coverage of the unknown-branch redirect (`tests/e2e/branches/branch-selector.spec.ts` and `tests/e2e/branches/test_branch_selector.py`), which asserts the toast, the `Other` button and the cleared URL — all preserved. Branch creation was checked for a regression risk: `useCreateBranchMutation` awaits `refetchQueries(["branches"])` in its own `onSuccess`, which React Query runs before the per-call callback, so the new branch is in the cache before the query string parameter flips and no spurious not-found toast fires.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/app/src/entities/resource-manager/ui/ip-address-pool-form.tsx
```diff
@@ -84,7 +86,7 @@ export const IpAddressPoolForm = ({
       }
       return field;
     });
-  }, [props.schema.kind, genericAddressSchema?.kind, currentObject, isUpdate]);
+  })();
```
saltas888, line 89: no need for a memo anymore?

frontend/app/src/shared/components/form/utils/isFieldDisabled.ts
```diff
@@ -16,19 +15,18 @@ export type IsFieldDisabledParams = {
 export const isFieldDisabled = ({
   owner,
   auth,
+  isDefaultBranch,
   isProtected,
   isReadOnly,
   permissions,
 }: IsFieldDisabledParams) => {
-  const currentBranch = store.get(currentBranchAtom);
```
saltas888, line 23: I think this is great hygiene honestly. We made this utility not requiring external deps. 🦾

frontend/app/src/entities/resource-manager/ui/ip-address-pool-form.tsx
bilalabbad: We use react compiler, so yes we do not need it anymore

saltas888 (approved):

PR #10135, opened 2026-08-05
docs: rework IPAM documentation

BaptisteGi:
## Summary by cubic
Reworks IPAM docs into task-focused guides and updates hierarchy, utilization, namespaces, branching, and allocation guidance so users can set up schemas, query data, and automate safely. Replaces legacy pages and fixes navigation and cross-links to reduce confusion and broken references.

- Adds five guides and removes three legacy pages; updates `ipam/overview`, `docs/sidebars.ts`, and fixes the link in `resource-manager/overview` to `ipam/build-your-ipam-schema`.
- Documents Resource Pool allocation: idempotency via `identifier`, lock-based concurrency (`InfrahubMultiLock`) for pools and prefixes on any branch, address locks only on the default branch, and merge/rebase-time hierarchy reconciliation; clarifies next-available GraphQL queries as read-only without cross-branch guarantees.
- Clarifies `IPHost`/`IPNetwork`/`IPAddress` attribute kinds vs IPAM nodes, and the Python SDK mapping to `ipaddress` interfaces; adds a Vale spelling exception for “netmask”.
- Required action: Update internal links that referenced removed pages to the new guides.

<sup>Written for commit 08fdbf6b5676e112894b91eb1888ee5614b48bb3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wvandeun:
You will need to merge this branch into release-1.11 or only merge after the 1.11 release. The attribute kind `IPAddress` will only ship in 1.11

docs/docs/ipam/build-your-ipam-schema.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+```shell
+infrahubctl marketplace get infrahub/ipam --stdout
+```
+
+See [Infrahub Marketplace](../schema/marketplace/index.mdx) for the full fetch-and-load workflow.
+
+## VLAN, VRF, RIR, and other extensions
+
+VLAN, VRF, RIR, and aggregate are Marketplace extensions, not part of the base IPAM schema — fetch them individually:
```
minitriga, line 62: **Blocking — RIR and aggregate aren't in the marketplace catalog.**

Checked the live API: `infrahub/rir` and `infrahub/aggregate` both return `{"detail":"Schema not found"}`. `infrahub/ipam`, `infrahub/vrf`, and `infrahub/vlan` do resolve to published items. As written, the heading and this sentence point readers at `infrahubctl marketplace get` commands for two items that don't exist.

```suggestion
## VLAN, VRF, and other extensions

VLAN and VRF are Marketplace extensions, not part of the base IPAM schema — fetch them individually:
```

docs/docs/ipam/build-your-ipam-schema.mdx
```diff
@@ -0,0 +1,131 @@
...
+infrahubctl schema load schemas/
+```
+
+To inspect the IPAM schema alone without loading it, fetch it with `--stdout`:
+
+```shell
+infrahubctl marketplace get infrahub/ipam --stdout
+```
+
+See [Infrahub Marketplace](../schema/marketplace/index.mdx) for the full fetch-and-load workflow.
```
minitriga, line 26: **Content dropped with no replacement.** `overview.mdx:19` tells the reader the three generics are there "to inherit from when building your IPAM schema", but this page — the one titled *Build your IPAM schema* — now only says to fetch a schema from the Marketplace.

The deleted `building-your-schema.mdx:8-27` had the YAML for exactly that. Nothing in the new set replaces it, so a reader defining their own IPAM nodes has no example to work from. Suggested as a plain block rather than an applyable suggestion because it contains a fenced code block:

````markdown
## Define your own IPAM nodes

To model IPAM without the Marketplace schema, inherit from the generics directly:

```yaml title="Schema definition implementing IPAM generics"
# yaml-language-server: $schema=https://schema.infrahub.app/develop/schema.schema.json
---
version: "1.0"
nodes:
  - name: "IPPrefix"
    namespace: "Ipam"
    inherit_from:
      - "BuiltinIPPrefix"
    description: "IPv4 or IPv6 network"
    icon: "mdi:ip-network"
    label: "IP Prefix"
    menu_placement: "IpamNamespace"
  - name: "IPAddress"
    namespace: "Ipam"
    inherit_from:
      - "BuiltinIPAddress"
    description: "IPv4 or IPv6 address"
    icon: "mdi:ip"
    label: "IP Address"
    menu_placement: "IpamNamespace"
```
````

File: docs/docs/ipam/build-your-ipam-schema.mdx [hunk 3]
```diff
@@ -0,0 +1,131 @@
...
+extensions:
+  nodes:
+    - kind: IpamIPPrefix
+      relationships:
+        - name: service
+          peer: ServiceGeneric
+          cardinality: one
+          direction: outbound
+          identifier: service_prefix
+```
```
minitriga, line 163: Missing the closing Next/Related section — the hub-and-spokes pattern has every spoke end with one pointing at adjacent spokes. `query-ipam-data.mdx` uses `## Next` and `plan-changes-on-a-branch.mdx` uses `## Related`; this page, `use-namespaces.mdx`, and `allocate-ips-and-prefixes.mdx` have none. Worth settling on one of the two heading names across all five.

````suggestion
```

## Next

- [Use namespaces](./use-namespaces.mdx) — isolating the prefixes and addresses this schema holds
- [Allocate IPs and prefixes](./allocate-ips-and-prefixes.mdx) — assigning addresses and subnets from a pool
- [Query IPAM data](./query-ipam-data.mdx) — reading computed values and the hierarchy
````

File: docs/docs/ipam/overview.mdx [hunk 4]
```diff
@@ -4,44 +4,57 @@ title: IP Address Management
...
+
+## Utilization
+
+Every IP prefix exposes a read-only `utilization` attribute, computed on read from the graph, using one of two formulas based on `member_type`:
+
+- **`prefix` mode** — utilization is the address space of the child prefixes, divided by the address space of the prefix itself. For example, `192.0.2.0/24` with one `192.0.2.0/26` child reports 25%.
+- **`address` mode** — utilization is the number of IP addresses assigned to the prefix, divided by the number of usable addresses in the prefix. For example, `192.0.2.0/26` holds up to 62 usable addresses (excluding network and broadcast); with 20 addresses assigned, it reports 32%.
 
-IPAM generics and schema nodes that inherit from them have relationships and a hierarchy. An IP prefix can be related to other prefixes (as a parent or a child), an IP address can be related to an IP prefix, and both are related to an IP namespace.
+Network and broadcast addresses count toward the usable total when `is_pool` is `true`, for IPv4 prefixes with a `/31` length (per [RFC 3021](https://datatracker.ietf.org/doc/html/rfc3021)), and for `/32` prefixes.
```
minitriga: Two gaps against `backend/infrahub/core/ipam/size.py:10-16`. The `-2` subtraction is gated on `version == 4`, so **IPv6 prefixes always report full address space** — unstated here, and the case a reader is most likely to get wrong. And `/31` and `/32` are one condition in the code (`prefixlen < 31` is the guard), so listing them separately reads as more special-casing than there is.

```suggestion
Network and broadcast addresses count toward the usable total when `is_pool` is `true`, when an IPv4 prefix is `/31` or longer (per [RFC 3021](https://datatracker.ietf.org/doc/html/rfc3021)), and for every IPv6 prefix.
```

docs/docs/ipam/overview.mdx
```diff
@@ -4,44 +4,57 @@ title: IP Address Management
...
 
-- [IP Namespaces](./ip-namespaces.mdx) — Understand namespace isolation, when to use multiple namespaces, and multi-tenant patterns
-- [Building your IPAM schema](./building-your-schema.mdx) — Define your IPAM nodes, load the schema, and query IP resources via GraphQL
-- [Automate with Resource Manager](./automate-with-resource-manager.mdx) — Use IPAM prefixes as pool sources for automated IP allocation
+Infrahub also exposes lower-level GraphQL queries — `InfrahubIPAddressGetNextAvailable` and `InfrahubIPPrefixGetNextAvailable` — that return the next free address or subnet without creating a record and without the pool's cross-branch allocation guarantees. Use Resource Pool for production workflows; reach for these only to look up a value without persisting it.
```
minitriga: "reach for these" is figurative. The voice rule asks for words in their literal sense, because much of the audience reads English as a second or third language.

```suggestion
Infrahub also exposes lower-level GraphQL queries — `InfrahubIPAddressGetNextAvailable` and `InfrahubIPPrefixGetNextAvailable` — that return the next free address or subnet without creating a record and without the pool's cross-branch allocation guarantees. Use Resource Pool for production workflows; use these only to look up a value without persisting it.
```

docs/docs/ipam/allocate-ips-and-prefixes.mdx
```diff
@@ -0,0 +1,59 @@
+---
+title: Allocate IPs and prefixes
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
```
minitriga: `Tabs` and `TabItem` are imported but never used on this page.

```suggestion

```

docs/docs/ipam/allocate-ips-and-prefixes.mdx
```diff
@@ -0,0 +1,59 @@
...
+## Use a resource pool
```
minitriga, line 9: **Content dropped — the when-to-use guidance.** The deleted `automate-with-resource-manager.mdx:22-31` carried the decision content: the four triggers (automated provisioning, generators, CI/CD, idempotent allocation) and the counterpart, "If you are only storing and querying IP data, IPAM alone is sufficient."

This page opens straight into mechanics, and `overview.mdx:58` keeps only a fragment. The hub-and-spokes rule is specific that a feature has to answer *when to use this*, not only *how*.

```suggestion
## When to allocate from a pool

Allocate from a Resource Pool when something other than a person assigns the address — a provisioning pipeline creating devices, a Generator rendering configuration, a CI/CD job reserving space for a new service. If you are only recording and querying IP data that already exists, the IPAM nodes alone are enough.

## Use a resource pool
```

docs/docs/ipam/allocate-ips-and-prefixes.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+interconnect_subnet = await client.allocate_next_ip_prefix(
+    resource_pool=interconnection_pool,
+    kind=IpamIPPrefix,
+    identifier=network.identifier,
+    branch=branch,
+)
+```
+
+`allocate_next_ip_address` and `allocate_next_ip_prefix` take the same `resource_pool` and `identifier` arguments as the underlying `InfrahubIPAddressPoolGetResource` and `InfrahubIPPrefixPoolGetResource` mutations — passing a stable `identifier` (for example, the device name) keeps a Generator idempotent across reruns. See the infrastructure-edge topology model in `models/infrastructure_edge.py` for a worked example that allocates loopback and management addresses, and interconnection subnets, this way.
```
minitriga: `models/infrastructure_edge.py` is a repo-internal path — readers on docs.infrahub.app don't have a checkout. The sample itself checks out against `models/infrastructure_edge.py:1457,1535-1537`, so it's a good example to have picked; it just needs a reachable link.

```suggestion
`allocate_next_ip_address` and `allocate_next_ip_prefix` take the same `resource_pool` and `identifier` arguments as the underlying `InfrahubIPAddressPoolGetResource` and `InfrahubIPPrefixPoolGetResource` mutations — passing a stable `identifier` (for example, the device name) keeps a Generator idempotent across reruns. The [infrastructure-edge topology model](https://github.com/opsmill/infrahub/blob/stable/models/infrastructure_edge.py) is a worked example that allocates loopback and management addresses, and interconnection subnets, this way.
```

docs/docs/ipam/allocate-ips-and-prefixes.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+## Advanced: next-available GraphQL primitives
+
+`InfrahubIPAddressGetNextAvailable` and `InfrahubIPPrefixGetNextAvailable` are read-only GraphQL queries that return the next free address or subnet from a given prefix. Unlike a Resource Pool allocation, they don't create a record, and they don't carry the pool's cross-branch allocation guarantee — two callers on different branches can compute the same "next available" value independently and collide.
```
minitriga: "they don't **carry** the pool's cross-branch allocation guarantee" — `carry` meaning "have" is called out explicitly in the figurative-language rule. Same pattern at `build-your-ipam-schema.mdx:45` ("A value **carrying** prefix notation"), `:88` ("**carry** its own relationships"), `:41` ("it **lives on** the node"), and `query-ipam-data.mdx:8` ("**walk** the hierarchy tree").

```suggestion
`InfrahubIPAddressGetNextAvailable` and `InfrahubIPPrefixGetNextAvailable` are read-only GraphQL queries that return the next free address or subnet from a given prefix. Unlike a Resource Pool allocation, they don't create a record, and they don't provide the pool's cross-branch allocation guarantee — two callers on different branches can compute the same "next available" value independently and collide.
```

docs/docs/ipam/allocate-ips-and-prefixes.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+Use Resource Pool for production workflows. Reach for these queries only to look up a value without persisting it.
```
minitriga:
```suggestion
Use Resource Pool for production workflows. Use these queries only to look up a value without persisting it.
```

docs/docs/ipam/use-namespaces.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: Use namespaces
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
```
minitriga: `Tabs` and `TabItem` are unused here (`VideoPlayer` on the next line is used).

```suggestion

```

docs/docs/ipam/use-namespaces.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+Consider an Internet Service Provider (ISP) that provides managed network services to three customers. Each customer uses the 10.0.0.0/8 private address range internally.
+
+Without namespaces, the ISP could not track these overlapping ranges in a single IPAM system. With namespaces, the ISP creates three namespaces (one per customer) and manages each customer's IP space independently:
+
+- **customer-a** namespace: Contains Customer A's 10.0.0.0/8 and its subnets
+- **customer-b** namespace: Contains Customer B's 10.0.0.0/8 and its subnets
+- **customer-c** namespace: Contains Customer C's 10.0.0.0/8 and its subnets
+
+Each namespace maintains its own hierarchy, utilization calculations, and address assignments. The ISP can query, report on, and manage each customer's IP space without any overlap or confusion.
```
minitriga: This section came across from `ip-namespaces.mdx` unrewritten, and the extraction rule asks for the rewrite on the way over. Three things: third person where it means a concrete actor ("the ISP could not track", "The ISP can query, report on, and manage"); "without any overlap or confusion" as filler; and bare `10.0.0.0/8` throughout while the rest of the file backticks addresses (lines 33, 42).

```suggestion
An Internet Service Provider runs managed network services for three customers, and each customer uses `10.0.0.0/8` internally. A single IPAM without namespaces cannot hold three copies of the same range. With one namespace per customer, each range is tracked independently:

- **customer-a** — Customer A's `10.0.0.0/8` and its subnets
- **customer-b** — Customer B's `10.0.0.0/8` and its subnets
- **customer-c** — Customer C's `10.0.0.0/8` and its subnets

Each namespace has its own hierarchy, utilization figures, and address assignments, so a query scoped to one customer never returns another customer's addresses.
```

docs/docs/ipam/query-ipam-data.mdx
```diff
@@ -0,0 +1,144 @@
+---
+title: Query IPAM data
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Read the computed values on IP prefixes and addresses, or walk the hierarchy tree. Use these patterns for reporting, automation, or verification.
```
minitriga: **Content dropped — mutations.** The deleted `building-your-schema.mdx:34-49` listed the generated mutations: `IpamIPPrefixCreate`/`Update`/`Upsert`/`Delete` and the `IpamIPAddress` equivalents.

Across the new set this page covers reads and `allocate-ips-and-prefixes.mdx` covers pool allocation — nothing documents creating, updating, or deleting a prefix or address directly. That's the most basic write path in the feature, and it now has no home.

The suggestion below only fixes "walk" (figurative — see the voice rule); the mutations gap needs a new section, either here or back on `build-your-ipam-schema.mdx`.

```suggestion
Read the computed values on IP prefixes and addresses, or follow the hierarchy relationships. Use these patterns for reporting, automation, or verification.
```

docs/docs/ipam/query-ipam-data.mdx
```diff
@@ -0,0 +1,144 @@
...
+```
+
+The Python SDK maps an `IPHost` attribute's value to the standard library's `ipaddress.IPv4Interface` or `IPv6Interface` — `netmask` and `hostmask` come from that object, not from a separate attribute.
+
+</TabItem>
+</Tabs>
+
+## Navigate the hierarchy
+
+IP prefixes expose `parent`, `children`, `ancestors`, and `descendants` relationships. They are read-only, Infrahub manages them automatically.
```
minitriga: Comma splice. The claim itself is correct and matches what the deleted page said about hierarchy relationships not being usable in mutations.

```suggestion
IP prefixes expose `parent`, `children`, `ancestors`, and `descendants` relationships. They are read-only — Infrahub manages them automatically.
```

minitriga (dismissed):
Structurally this is a good rework — the five-spoke split is the right shape, and the branching/concurrency page is genuinely new value. Vale and markdownlint are clean on all six pages, and I checked a long list of factual claims against the source (verified list at the bottom). Most inline comments carry an applyable suggestion.

## Blocking

**`infrahub/rir` and `infrahub/aggregate` are not marketplace items** (`build-your-ipam-schema.mdx:28-30`). Verified against the live catalog: `GET /api/v1/schemas/infrahub/{rir,aggregate}` returns `{"detail":"Schema not found"}`, while `infrahub/ipam`, `infrahub/vrf`, and `infrahub/vlan` all return published items. As written, the heading and that sentence point readers at `infrahubctl marketplace get` commands for two items that don't exist.

## Not anchored to a line: redirects

Three published URLs are deleted with no redirect — `/ipam/ip-namespaces`, `/ipam/building-your-schema`, `/ipam/automate-with-resource-manager`. `docs/redirects-pending/ipam.yml` is untouched and still lists all three under `new_pages`. It needs redirect entries pointing at `/ipam/use-namespaces`, `/ipam/build-your-ipam-schema`, and `/ipam/allocate-ips-and-prefixes`, and the `new_pages` block refreshed for the new slugs.

## Verified correct — no need to re-check

The `IPHost`/`IPNetwork`/`IPAddress` attribute-kind section, against the `release-1.11` implementation: `IPAddress` is registered (`types.py:273-279`, `:350`) and listed in the generated reference (`reference/schema/attribute.mdx:187`); `validate_format` runs `ipaddress.ip_address(value)`, so prefix notation really is rejected; `type = str` with `_normalize_value` returning `str(ip_address(value))`, so `region.ntp_server.value` really is the bare string; and `IPHost._normalize_value` returns `ip_interface(value).with_prefixlen`, so `192.0.2.1` really does become `192.0.2.1/32`.

Also verified: prefix node-level `utilization`/`netmask`/`hostmask`/`network_address`/`broadcast_address` (`schema/schema.graphql:519-550`); the IPHost sub-fields `ip`/`netmask`/`hostmask`/`prefixlen`/`version`; the SDK `IPv4Interface`/`IPv6Interface` mapping (`core/attribute.py:1053`); the concurrency claim — address locks default-branch-only, prefix locks on any branch (`graphql/mutations/ipam.py:117-121`, `:281-285`); `default` namespace undeletable (`ipam.py:92-93`); both SDK allocation samples, which match `models/infrastructure_edge.py:1457,1535-1537` almost verbatim; `InfraInterfaceL3.ip_addresses` -> `IpamIPAddress` with `kind: Component` (`models/base/dcim.yml:317-321`); the marketplace `--collection`/`--stdout` syntax and `infrahub/base-schemas`; both utilization percentages (25%, 32%); all 8 cross-links and 3 anchors resolve; the sidebar hub `link` is present; dropping "In this section" from the hub is correct per the hub-and-spokes rule; and the `netmask` Vale exception is justified by the bare use in the attribute-kinds table (`hostmask` is always backticked, so it needs none).

docs/docs/ipam/build-your-ipam-schema.mdx
BaptisteGi: They will be together with V2 of schema library

docs/docs/ipam/build-your-ipam-schema.mdx
BaptisteGi: fixed!

saltas888 (approved):

PR #10137, opened 2026-08-05
refactor(git): enforce invalid-argument-type for the integrator

ogenstad:
# Why

`backend/infrahub/git/integrator.py` was listed under four ty overrides. This clears one of them outright: **`invalid-argument-type` is now enforced for the file** and its entry is gone from `pyproject.toml`. That is the first file in `git/` to shed a rule since the module-wide override was split per file in #10125, and integrator.py is the largest file in the module at 91 KB.

Goal: take a rule from "suppressed for this file" to "enforced for this file", so no new violation of it can land here again.

Non-goals: the other three rules integrator.py is still listed under (`unresolved-attribute` 9, `no-matching-overload` 9, `invalid-assignment` 7). Those need the surrounding function signatures tightened, which is judgement work, not substitution.

## What changed

Nine `invalid-argument-type` diagnostics, four mechanical causes plus one behaviour fix.

**Mechanical (no runtime effect):**

- `source=self.id` at four sites passed a `UUID` where the SDK declares `str | None`. `generate_payload_create` does `item_metadata["source"] = str(source)` internally, so wrapping it produces the identical string. Two other sites in the same file already did this; now all six agree.
- `load_class(import_root=...)` received a `Path` where `str | None` is declared. The SDK does `Path(import_root, relative_path, module_path.name)`, which accepts either, so `str()` is a no-op.
- Two artifact early-returns fed `CoreArtifact.storage_id.value` (a `StringOptional`) into `ArtifactGenerateResult.storage_id: str`. The invariant is real, the storage id is written in the same `save()` as the checksum, but it is not visible at the point of use. The returns now guard on both fields, so the impossible state falls through and regenerates rather than reporting a result with no storage id.

**Behaviour change, one place:** `import_schema_files` detected unloadable schema files, logged them, and then left them in the list. The failure was therefore produced accidentally by Pydantic choking on `None` further down, so a user with an empty schema file was told:

```
Schema not valid, found '1' error(s) in foo.yml : Input should be a valid dictionary or instance of InfrahubSchemaWrite
```

rather than the reason `YamlFile.load_content()` had already recorded. Now:

```
Unable to load 1 schema file(s): foo.yml (Empty YAML/JSON file)
```

Every invalid file is still logged individually first, exactly as before, and all of them are named in the error instead of whichever one Pydantic happened to reach first. Payloads are collected into a narrowed list because re-reading `.content` in the later loops would lose the non-`None` narrowing.

**Also included:** the eight remaining kind-string node fetches in the file now pass the generated SDK protocol classes, finishing what #10125 started for `tasks.py`. That required correcting two return annotations from `InfrahubNode` to the precise protocol type, since the generated protocols are **not** `InfrahubNode` subclasses (`CoreArtifactDefinition.__mro__` is `CoreArtifactDefinition -> CoreTaskTarget -> CoreNode -> CoreNodeBase -> object`). Both callers discard the return value, so there is no cascade. Only two of the module's diagnostics now come from the SDK `__getattr__` union, down from 57 before this work began.

`integrator.py` 37 -> 27 suppressed ty diagnostics, the module 108 -> 98.

## How to review

Start with the one-line `pyproject.toml` diff, that is the point of the PR. Then the `import_schema_files` restructure, which is the only behaviour change. The rest is `str()` wrappers and one-token kind substitutions.

**Known inconsistency worth a decision:** six functions now read `schema.get(kind=InfrahubKind.X)` a few lines above `sdk.create(kind=CoreX)`. I stopped at the node calls deliberately, because `schema.get` returns a schema rather than a node so there is no typing gain, but the two spellings sitting three lines apart do look odd. `schema.get` accepts `type[SchemaType | SchemaTypeSync] | str`, so converting them is safe if you would rather have the consistency. Say the word and I will.

## How to test

```bash
uv run ty check .                                  # All checks passed!
uv run invoke backend.mypy                         # Success: no issues found in 1601 source files
uv run invoke backend.ruff
uv run pytest backend/tests/unit/git/ --no-cov -q  # 178 passed
```

The ratchet is armed, not just asserted. With a deliberate violation injected into integrator.py, `ty check .` now fails on it, where before the override would have swallowed it:

```
backend/infrahub/git/integrator.py:835:34: error[invalid-argument-type] Argument to bound method
`InfrahubSchemaBase.validate` is incorrect: Expected `dict[str, Any]`, found `None`
```

## Tests

The behaviour change is covered by `backend/tests/unit/git/test_integrator.py`: 6 unit tests, 0.09s, no database and no containers. `find_unloadable_schema_files` and `format_unloadable_schema_files` are extracted as pure functions over `list[SchemaFile]`, which is directly constructible, so no fixtures or patching are needed. The cases include a document containing only `{}`, which loads to an empty dict rather than `None`, so the loader's `valid` flag rather than the content is the signal. The tests were mutation-checked: dropping that condition makes them fail.

The `storage_id` guard is deliberately not tested. It only changes behaviour when the checksum matches while `storage_id` is empty, and in that state the previous code did not work either, it raised `ValidationError: storage_id` from Pydantic because the field is declared `str`. The guard replaces that crash with a re-upload that repairs the record, so it cannot regress that state. The state should also be unreachable, since `checksum` and `storage_id` are only ever written together in the same `save()`.

Where this logic belongs is a separate question from whether it is tested. `infrahubctl schema load` performs the identical flow using SDK helpers that Infrahub cannot call because they are coupled to a rich `Console` and `typer.Exit`, which is why this hand-rolled copy diverged and lost the invalid-file handling in the first place. Tracked in **opsmill/infrahub-sdk-python#1225**, which proposes a library-level `load_schema_files()`. The two functions extracted here are shaped to be replaced by it.

## Impact & rollout

- **Backward compatibility:** one user-visible error message improves; no API or schema change.
- **Performance:** none.

## Follow-ups

- opsmill/infrahub-sdk-python#1225, after which this file loses ~20 lines and gains real test coverage.
- Deduplicating `generate_artifact` and the near-identical method below it, roughly 35 lines each, which would collapse the two storage-id guards into one.
- The remaining three rules for this file, plus `repository.py`, where `assignment` and `call-overload` are one mypy error each and would remove two more override entries.

## Checklist

- [ ] Tests added/updated 
- [ ] Changelog entry added 

### DIFF
--- a/backend/infrahub/git/integrator.py
+++ b/backend/infrahub/git/integrator.py
@@ -207,6 +207,17 @@ class ObjectImportPlan:
     artifact_definitions: dict[str, InfrahubRepositoryArtifactDefinitionConfig]
 
 
+def find_unloadable_schema_files(schemas_data: list[SchemaFile]) -> list[SchemaFile]:
+    """Return the schema files whose content never loaded, and which therefore carry no payload."""
+    return [schema_file for schema_file in schemas_data if not schema_file.valid or schema_file.content is None]
+
+
+def format_unloadable_schema_files(unloadable: list[SchemaFile]) -> str:
+    """Report the reason the loader recorded for each file, rather than an opaque validation error."""
+    details = ", ".join(f"{item.identifier} ({item.error_message})" for item in unloadable)
+    return f"Unable to load {len(unloadable)} schema file(s): {details}"
+
+
 class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
     """This class provides interfaces to read and process information from .infrahub.yml files and can perform.
 
@@ -859,29 +870,21 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
             # and send an empty list to the API
             return
 
-        # A file that failed to load carries no content, so it cannot be validated or sent to the
-        # API. Report the reason the loader recorded rather than letting the empty content surface
-        # later as an opaque schema validation failure. Each payload travels next to its file
-        # because re-reading the content attribute in the loops below would lose the narrowing.
-        unloadable: list[SchemaFile] = []
-        payloads: list[tuple[SchemaFile, dict[str, Any]]] = []
-        for schema_file in schemas_data:
-            if not schema_file.valid or schema_file.content is None:
-                unloadable.append(schema_file)
-                log.error(f"Unable to load the file {schema_file.identifier}, {schema_file.error_message}")
-                continue
-            payloads.append((schema_file, schema_file.content))
-
+        # A file that failed to load carries no content, so it can be neither validated nor sent to
+        # the API. Report the reason the loader recorded rather than letting the missing content
+        # surface later as an opaque schema validation failure. Rejecting them here is also what
+        # makes `payload` below safe: it substitutes an empty dict for absent content, which would
+        # otherwise turn an unloadable file into a misleading schema error.
+        unloadable = find_unloadable_schema_files(schemas_data)
+        for schema_file in unloadable:
+            log.error(f"Unable to load the file {schema_file.identifier}, {schema_file.error_message}")
         if unloadable:
-            details = ", ".join(f"{item.identifier} ({item.error_message})" for item in unloadable)
-            raise ValidationError(
-                identifier=str(self.id), message=f"Unable to load {len(unloadable)} schema file(s): {details}"
-            )
+            raise ValidationError(identifier=str(self.id), message=format_unloadable_schema_files(unloadable))
 
         # Valid data format of content
-        for schema_file, payload in payloads:
+        for schema_file in schemas_data:
             try:
-                self.sdk.schema.validate(payload)
+                self.sdk.schema.validate(schema_file.payload)
             except PydanticValidationError as exc:
                 log.error(f"Schema not valid, found '{len(exc.errors())}' error(s) in {schema_file.identifier} : {exc}")
                 raise ValidationError(
@@ -890,7 +893,7 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
                 ) from exc
 
         response = await self.sdk.schema.load(
-            schemas=[payload for _, payload in payloads], branch=branch_name, wait_until_converged=True
+            schemas=[item.payload for item in schemas_data], branch=branch_name, wait_until_converged=True
         )
 
         if response.errors:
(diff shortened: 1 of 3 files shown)


ogenstad:
Thanks, both flagged the same thing: behaviour changes without coverage. Taking them separately.

**Issue 1, `import_schema_files` (integrator.py:875) — agreed, fixed in e511c1cb8.**

You were right and my "needs the full integration fixture" reasoning was wrong. That is true of the *method*, but the part I changed is a pure function over `list[SchemaFile]`, and `SchemaFile` is directly constructible with no fixtures and no patching:

```python
SchemaFile(identifier="empty.yml", location=Path("empty.yml"), valid=False, error_message="Empty YAML/JSON file")
```

So I extracted `find_unloadable_schema_files` and `format_unloadable_schema_files` and covered them with the project's dataclass test case pattern. 6 tests, 0.09s, no DB and no containers. They include the case that is easy to get wrong: a document containing only `{}` loads to an empty dict rather than `None`, so the loader's `valid` flag is the signal, not the content. I verified the tests fail when that condition is dropped.

The extraction also removed an awkward `tuple[list[tuple[SchemaFile, dict[str, Any]]], list[SchemaFile]]` I had been using to carry narrowed content forward. Rejecting unloadable files up front makes `SchemaFile.payload` safe, which is how the SDK's own CLI does it, so the validate loop and the load call return to their original shape.

**Issue 2, the artifact `storage_id` guard (integrator.py:2046) — I do not think a test is warranted here, and I would rather not add one for an unreachable state.**

The guard only changes behaviour when the checksum matches while `storage_id` is empty. In that state the old code did not work either, it crashed:

```
>>> ArtifactGenerateResult(changed=False, checksum="abc", storage_id=None, artifact_id="id1")
ValidationError: storage_id
  Input should be a valid string
```

`storage_id` is declared `str` on that Pydantic model, so the previous code raised a Pydantic `ValidationError` from inside artifact generation. The guard replaces that crash with a re-upload that repairs the record. It cannot introduce a regression in that state, it can only turn a hard failure into a recovery.

The state should also be unreachable: `checksum` and `storage_id` are only ever written together, in the same `save()`, a few lines below the guard. Constructing a test for it would mean building an artifact whose persisted fields violate that invariant, using the full artifact fixture, to assert a repair path that exists only as belt and braces.

Happy to add one if you disagree, but I would rather spend the fixture cost on `opsmill/infrahub-sdk-python#1225`, which moves this loading logic to where it can be tested cheaply for both consumers.

ogenstad:
I decided to close this as we're planning to move some of this code to the SDK regardless so I don't think this is that helpful.

PR #10141, opened 2026-08-05
Clone git repositories on their own default branch

### LINKED ISSUE #8749: bug: Artifact generation fails for read-write repositories when default branch is not "main"

### LINKED ISSUE #9825: bug: per-flow repository factory never learns configured default_branch (artifacts, transforms, PC checks, generators)
## Component
Git Integration

## Infrahub version
1.10.1 (stable)

## Current Behavior
The shared factory that downstream flows use to build their repository object (`get_initialized_repo` → `InfrahubRepository.init(id=..., name=..., commit=..., client=...)`, `backend/infrahub/git/repository.py:428`) never passes `default_branch_name`, and `init()` never queries the repository node. The `default_branch` property then silently falls back to `registry.default_branch` (`"main"`, `base.py:191`).

For a repository whose git `default_branch` is not `main` (e.g. `develop`), every flow built on this factory operates on a repo object that believes its default branch is `main`:

- artifact generation (`artifacts/tasks.py:21`)
- transformations (`transformations/tasks.py:21,45`)
- proposed-change checks incl. merge-conflict validation (`proposed_change/tasks.py:719,1039`, `branch_diff.py:72`, `git/tasks.py:1061→get_conflicts`)
- generators (`generators/tasks.py:46`), computed attributes (`computed_attribute/tasks.py:94`), git file ops
- `git_branch_create` (`git/tasks.py:428`) has the same omission

The proposed-change pipeline contains no reference to `default_branch` at all. Consequences concentrate wherever branch-*name* resolution happens: local-branch lookups fail (the clone has local `develop` + `master`, **no** local `main`), and the mapping helpers (`_get_mapped_target_branch`, `_get_mapped_remote_branch`, `_resolve_worktree_identifier`) misroute. Commit-scoped operations work by accident (worktrees are keyed by SHA, and the worktree *named* `main` is checked out on `develop`), which is why the breakage is partial and hard to attribute — #9600 (periodic sync) and most likely #8749 (artifacts) are the same underlying gap surfacing at different call sites; #9601 fixed only the periodic-sync fetch.

## Expected Behavior
A repository object built from its id/name resolves the repository's configured `default_branch` (query it in `init()`, or pass it at every factory call site), so branch resolution in downstream flows targets the configured default branch.

## Steps to Reproduce
1. Register a read-write `CoreRepository` with `default_branch=develop` (remote has `main` + `develop`) and let it import.
2. Build the repo object the way the flows do: `get_initialized_repo(repository_id=..., name=..., repository_kind="CoreRepository")`.
3. Observe `repo.default_branch == "main"` (expected `"develop"`), and `repo.get_commit_value(branch_name=repo.default_branch, remote=False)` raises `Branch main not found` (no local `main` exists in the clone).

Deterministic reproduction: `test_initialized_repo_resolves_configured_default_branch` in `backend/tests/integration/git/test_multi_env_writeback.py` (PR #9798), `xfail(strict)` — flips to a failure signal when fixed.

## Additional Information
Suggested systemic fix: make `init()` load `default_branch` from the repository node (one query) instead of fixing call sites one by one — #9601 already demonstrated the per-call-site approach doesn't close the class of bugs. Blocks the multi-environment single-repo pattern (one non-`main` branch per environment) for any flow beyond plain import/sync.

gmazoyer:
# Why

A worker without a local copy of a repository clones it on demand. That path only carries the repository id and name, so the repository object never learns which git ref to track. It fell back to the platform default branch. Git then failed on any repository whose default branch is not `main`, the repository went to error, and artifact generation failed with it.

It takes more than one worker to see this. Each worker keeps its own git directory, so only the worker that added the repository has the clone.

Closes #8749. It does not close #9825: `repo.default_branch` is still wrong when the clone already exists.

## What changed

`InfrahubRepositoryBase` gets one abstract method, `resolve_checkout_ref`. The re-clone path calls it and passes the result to `create_locally`. `CoreRepository` reads `default_branch`, `CoreReadOnlyRepository` reads `ref`. No schema change, no migration.

```
init()
  validate_local_directories()      clone missing
    ensure_location_is_defined()    -> location
    resolve_checkout_ref()          -> default_branch or ref
      create_locally()
```

## What to look at

`backend/infrahub/git/integrator.py` has the behaviour change. The graph lookup sits inside the missing directory branch, so an existing clone makes no extra query. `init()` is on the hot path of artifacts, transforms and generators.

## How to test

```bash
uv run pytest backend/tests/functional/git/test_repository_default_branch.py
uv run pytest backend/tests/integration_docker/test_artifact_composition.py
```

The end to end case needs two workers, so it lives in `integration_docker`. The artifact composition test now uses a repository on a non default branch, instead of a new test class with its own compose stack.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -191,6 +191,11 @@ class InfrahubRepositoryBase(BaseModel, ABC):
     def default_branch(self) -> str:
         return self.default_branch_name or registry.default_branch
 
+    @abstractmethod
+    async def resolve_checkout_ref(self) -> str:
+        """Return the git ref the primary clone has to be checked out on, reading it from the graph if needed."""
+        raise NotImplementedError()
+
     @property
     def legacy_directory_root(self) -> Path:
         """Return the legacy path to the root directory for this repository."""
--- a/backend/infrahub/git/integrator.py
+++ b/backend/infrahub/git/integrator.py
@@ -197,7 +197,11 @@ class InfrahubRepositoryIntegrator(InfrahubRepositoryBase):
             self.validate_local_directories()
         except RepositoryInvalidFileSystemError:
             await self.ensure_location_is_defined()
-            await self.create_locally(infrahub_branch_name=self.infrahub_branch_name, update_commit_value=False)
+            await self.create_locally(
+                checkout_ref=await self.resolve_checkout_ref(),
+                infrahub_branch_name=self.infrahub_branch_name,
+                update_commit_value=False,
+            )
             self.reinitialized = True
             log.info(f"Initialized the local directory for {self.name} because it was missing.")
 
--- a/backend/infrahub/git/repository.py
+++ b/backend/infrahub/git/repository.py
@@ -9,6 +9,7 @@ from cachetools.keys import hashkey
 from cachetools_async import cached
 from git.exc import BadName, GitCommandError
 from infrahub_sdk.exceptions import GraphQLError
+from infrahub_sdk.protocols import CoreReadOnlyRepository, CoreRepository
 from prefect import task
 from prefect.cache_policies import NONE
 from pydantic import Field
@@ -82,6 +83,15 @@ class InfrahubRepository(InfrahubRepositoryIntegrator):
         )
         log.info("Created new repository locally.", repository=self.name)
         return self
+
+    async def resolve_checkout_ref(self) -> str:
+        if not self.default_branch_name:
+            repository = await self.sdk.get(
+                kind=CoreRepository, name__value=self.name, exclude=["tags", "credential"], raise_when_missing=True
+            )
+            self.default_branch_name = repository.default_branch.value
+
+        return self.default_branch
 
     def get_commit_value(self, branch_name: str, remote: bool = False) -> str:
         branches = {}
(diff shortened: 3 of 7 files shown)


polmichel (approved):
LGTM!

PR #10154, opened 2026-08-06
fix: skip computed attribute query triggers for kinds no field is read from

gmazoyer:
# Why

A Python computed attribute creates one trigger per node kind its query reads. When the query goes through a relationship whose peer is a generic, the analyzer reports every member kind of that generic, even the ones we read no field from. We created a trigger for each of them.

Those triggers get no field filter, so they fire on any update to that kind. And it cascades: the recompute writes values, those writes are updates, and they match the same triggers again. One perf run ended up with about 26000 `query-computed-attribute-transform-targets` flows.

## What changed

In `gather.py`, skip a kind when its read set is empty. Kinds with real fields keep their filtered trigger. Kinds read through `display_label` keep theirs too, their read set is not empty.

The `gqlquery01` test fixture had an invalid query (`name` selected directly under the root node). That is why the old test could only check that some trigger existed. Fixed it and made the assertion exact.

## One thing to look at

The analyzer knows about `display_label` but not `hfid`. So a query that only reads `hfid` from a peer loses its trigger now. I did not fix that here.

I checked whether we could filter on those fields instead of matching any field, and we can: `display_label` and `human_friendly_id` do arrive as `infrahub.field.name` in the node mutation event. So the right fix is to record `hfid` in the analyzer, map `hfid` to `human_friendly_id`, and drop the match any field fallback. That is IFC-2983.

## Test

```bash
uv run pytest --neo4j backend/tests/component/computed_attribute/ backend/tests/component/graphql/test_query_analyzer.py
```

I did not run the trigger audit in `infrahub-private-tests`.

Same bug on `release-1.10` and `develop`, the file is identical there, so I will backport.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added


ogenstad (approved):
LGTM, though I think it would be good to add the hfid part as well to avoid under generation. This should only happen if there's a query that only looks for the hfid of a generic or node and nothing else, the chances of such a query might be low but I think we have to include it as an option. We we should also consider if these updates (to hfid & display_labels) are presented as fields in node mutation events. I think that might be missing now.

PR #10180, opened 2026-08-09
fix: advance number pool past values already present on the target kind

### LINKED ISSUE #10179: bug: CoreNumberPool parks on a value already present on the target kind instead of advancing
## Component

API Server / GraphQL

## Infrahub version

1.10.6 (present on current `stable` and `develop`)

## Current Behavior

A `CoreNumberPool` chooses its next value from its own reservations only. It has no view of values already present on the target kind (created directly, or brought in by a brownfield import). When a value inside the pool range already exists on the target, the pool still offers it, the attribute's uniqueness constraint rejects the save, and because the failed allocation reserves nothing, the pool re-offers the same value on every subsequent attempt. It parks permanently on that value until the conflicting object is removed.

A single pre-existing value inside the range therefore stops the pool at that value rather than degrading gracefully.

## Expected Behavior

The pool skips a value that already exists on the target kind and advances to the next free one, so a pool can be declared over a range that already contains data. This should track the same visibility the uniqueness constraint uses (branch scoping and deletions), and only apply where a duplicate would actually be rejected (a globally unique attribute).

## Steps to Reproduce

1. Define a node with a unique Number attribute (e.g. `Ticket.ticket_id`, `unique=True`).
2. Create a `CoreNumberPool` over that attribute with range `1..10`.
3. Create one object of that kind directly with `ticket_id = 1` (inside the range, not via the pool).
4. Allocate `ticket_id` from the pool for a new object.
5. The pool offers `1`, the save fails on the uniqueness constraint, and repeated allocations keep offering `1` instead of advancing to `2`. Deleting the object holding `1` and reallocating returns `1`, confirming the cursor was parked.

## Additional Information

- Allocation is derived only from the pool's `IS_RESERVED` edges (`backend/infrahub/core/node/resource_manager/number_pool.py`, `backend/infrahub/core/query/resource_manager.py`), so any value not handed out by the pool is invisible to it.
- Impact: any brownfield deployment where a pool range overlaps already-deployed values is stalled at the first such value.
- Related: #7276 (static allocations in number pools), #7064 (multiple number ranges in a single pool).

iddocohen:
# Why

A `CoreNumberPool` picked its next value from its own reservations only, so a value already present on the target kind (created directly, or brought in by a brownfield import) was invisible to it. The pool offered that value, the uniqueness constraint rejected the save, and because the failed allocation reserved nothing, the pool re-offered the same value on every attempt, parking permanently at it.

Goal: a pool can be declared over a range that already contains data, skipping occupied values and advancing to the next free one.

Non-goals: no support for per-relationship / composite `[relationship, value]` uniqueness (those stay on their current path); no performance change to the occupancy lookup.

Closes #10179

## What changed

Behavioral:
- For a globally unique attribute, allocation now skips values already held on the target kind and advances to the next free value, instead of stalling on the first pre-existing one.

Implementation:
- New `NumberPoolGetTaken` query reads values currently held on the target kind for the pool's attribute, within range. Its branch/temporal handling mirrors the uniqueness-constraint validator: `is_isolated=False` (so a value created on the origin branch after this branch's point still counts), and deletions drop out via the active-edge check.
- `get_taken` method plus a guarded hook in `get_next`: taken values are unioned into the existing exclusion set, reusing the current skip loop.
- Gated on `attribute.unique`, so only allocations where a duplicate would actually be rejected are constrained.

What stayed the same:
- Reservation-based `get_free` is unchanged; non-unique and per-relationship-unique attributes remain fully allocatable. No schema changes, no GraphQL/API changes.

## How to review

1. `backend/infrahub/core/query/resource_manager.py`: `NumberPoolGetTaken` (the new Cypher).
2. `backend/infrahub/core/node/resource_manager/number_pool.py`: `get_taken` and the guarded hook in `get_next`.
3. `backend/tests/component/core/resource_manager/test_number_pool.py`: 4 new tests.

Extra scrutiny on the branch/temporal scoping of the new query (the `is_isolated=False` choice, which mirrors the uniqueness validator).

## How to test

```bash
uv run pytest backend/tests/component/core/resource_manager/test_number_pool.py
```

Covers: skipping a pre-existing value, the `attribute.unique` guard (non-unique stays allocatable), recovery after the conflicting object is deleted, and origin-branch visibility from a child branch.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        # branch after this branch's point still collides here, so it must count as taken.
+        branch_filter, branch_params = self.branch.get_query_filter_path(
+            at=self.at.to_string(), branch_agnostic=self.branch_agnostic, is_isolated=False
+        )
+
+        self.params.update(branch_params)
+        self.params["attribute_name"] = self.pool.node_attribute.value
+
+        query = """
+        MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]->(attr:Attribute { name: $attribute_name })-[hv:HAS_VALUE]->(av:AttributeValueIndexed)
```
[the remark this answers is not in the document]
iddocohen: This case is already handled and covered by a test, so no `IS_PART_OF` resolution is needed here.

Deleting a node cascades to `attr.delete()`, which closes the `HAS_VALUE` edge. The `CALL` subquery orders by `from DESC, status ASC` and takes the freshest edge, so a deleted node's value returns `is_active = False` and drops out. This matches the uniqueness-constraint validator, which also anchors on the `HAS_ATTRIBUTE`/`HAS_VALUE` path rather than `IS_PART_OF`.

`test_allocate_reuses_value_after_conflicting_target_deleted` verifies the behaviour: it creates a conflicting object in range, deletes it, then allocates and asserts the pool hands back the freed value.

backend/infrahub/core/node/resource_manager/number_pool.py
```diff
@@ -147,6 +162,14 @@ async def get_next(self, db: InfrahubDatabase, branch: Branch, attribute: Attrib
         if effective_start > effective_end:
             raise PoolExhaustedError("There are no more values available in this pool.")
 
+        # A value already present on the target kind only blocks allocation when a duplicate would be
+        # rejected, i.e. when the attribute is unique on its own. For a per-relationship uniqueness
+        # constraint or a non-unique attribute the value is reusable, so skip it only when unique.
```
saltas888: I think we can be more concise here with the commenting. One day our Claude.md will understand how to comment... Im tired of giving AI this feedback

backend/infrahub/core/node/resource_manager/number_pool.py
```diff
@@ -71,6 +72,20 @@ async def get_free(
 
         return query.get_result_value()
 
+    async def get_taken(
```
saltas888, line 75: should this be a private (_) member? not sure about our convention

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        """ % {
+            "branch_filter": branch_filter,
+            "node": self.pool.node.value,
+        }
+
+        self.add_to_query(query)
+        self.return_labels = ["value"]
+        self.order_by = ["value"]
+
+    def get_taken_values(self) -> set[int]:
```
saltas888, line 624: I'd prefer to have a common contract/interface and have this generic like `get_data` when we need to do extra transformation

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        branch_filter, branch_params = self.branch.get_query_filter_path(
+            at=self.at.to_string(), branch_agnostic=self.branch_agnostic, is_isolated=False
+        )
+
+        self.params.update(branch_params)
+        self.params["attribute_name"] = self.pool.node_attribute.value
+
+        query = """
+        MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]->(attr:Attribute { name: $attribute_name })-[hv:HAS_VALUE]->(av:AttributeValueIndexed)
+        WHERE toInteger(av.value) >= $start_range and toInteger(av.value) <= $end_range
```
saltas888: toInteger makes the range predicate non-indexable, so the only plan left is a full scan of the target kind, per allocation, under the pool lock. Number values are stored as native integers and NumberPoolGetAllocated already compares av.value directly, I'd drop the cast so the planner can at least consider a range seek on the value index.

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        CALL (n, attr, av) {
```
saltas888: The outer MATCH binds ha/hv but never uses them, so the subquery runs once per edge pair instead of once per (n, attr, av). A `WITH DISTINCT n, attr, av` before the CALL avoids that, same as the `WITH DISTINCT node` in the uniqueness query.

backend/infrahub/core/node/resource_manager/number_pool.py
iddocohen: Done, condensed to a single line.

backend/infrahub/core/node/resource_manager/number_pool.py
iddocohen: Kept public to match its siblings `get_used`/`get_free` on this class, which are public and also called from tests and the utilization resolver.

backend/infrahub/core/query/resource_manager.py
iddocohen: Left as-is for this PR: each number-pool query exposes its own typed accessor today (`get_result_value`, `iter_results`, `get_free_data`), so `get_taken_values` matches the current pattern. A shared `get_data` contract is worth doing, but as its own refactor rather than folded into this bug fix.

backend/infrahub/core/query/resource_manager.py
iddocohen: Done, dropped the `toInteger` cast so the range predicate can use the value index. `av.value` is a native integer, which is what `NumberPoolGetAllocated` already relies on.

backend/infrahub/core/query/resource_manager.py
iddocohen: Done, added `WITH DISTINCT n, attr, av` before the CALL so it runs once per `(n, attr, av)` instead of per edge pair, mirroring the uniqueness query.

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+    Sees values created outside the pool, which allocation must skip or it stalls on a value the
+    uniqueness constraint rejects.
+    """
+
+    name = "number_pool_get_taken"
+    type = QueryType.READ
+
+    def __init__(
+        self,
+        pool: CoreNumberPool,
```
ajtmccarty, line 575: the query shouldn't need to know about the structure of a `CoreNumberPool` instance. this creates a cross-dependency where the query needs to know about `CoreNumberPool` and `CoreNumberPool` needs to know about the query. we want to make sure that the dependency chain always points in a single direction. it looks like `min_value` and `max_value` should be calculated with `pool.start/end_range` before being passed into the query and `attribute_name` and `schema_kind` should be their own parameters

backend/infrahub/core/node/resource_manager/number_pool.py
```diff
@@ -147,6 +157,12 @@ async def get_next(self, db: InfrahubDatabase, branch: Branch, attribute: Attrib
         if effective_start > effective_end:
             raise PoolExhaustedError("There are no more values available in this pool.")
 
+        # Only a globally unique attribute rejects a duplicate, so skip existing values only then.
+        if attribute.unique:
+            excluded_values |= await self.get_taken(
+                db=db, branch=branch, min_value=effective_start, max_value=effective_end
+            )
+
```
ajtmccarty, line 165: not a problem caused by this PR, but we really should be able to get the next value (or even the next series of values) with a single cypher query

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        # this branch point still collides here.
+        branch_filter, branch_params = self.branch.get_query_filter_path(
+            at=self.at.to_string(), branch_agnostic=self.branch_agnostic, is_isolated=False
+        )
+
+        self.params.update(branch_params)
+        self.params["attribute_name"] = self.pool.node_attribute.value
+
+        query = """
+        MATCH (n:%(node)s)-[ha:HAS_ATTRIBUTE]->(attr:Attribute { name: $attribute_name })-[hv:HAS_VALUE]->(av:AttributeValueIndexed)
```
ajtmccarty: looks like the `ha` and `hv` variables can be dropped in this match b/c they are never used

backend/infrahub/core/query/resource_manager.py
```diff
@@ -560,6 +560,71 @@ def get_result_value(self) -> int | None:
...
+        WHERE av.value >= $start_range and av.value <= $end_range
+        WITH DISTINCT n, attr, av
```
ajtmccarty: I think that you actually want to restrict this to just `n` and `attr`. an `:Attribute` will always have, at most, 1 active `:AttributeValue` on a given branch at a given time. if you include the `:AttributeValue` in this grouping, then you could unnecessarily check past `AttributeValue`s that have been deleted. you'll also need to include another filter like `WHERE av.value >= $start_range and av.value <= $end_range` following the subquery to ensure that only in-range values are identified

backend/tests/component/core/resource_manager/test_number_pool.py
```diff
@@ -52,6 +54,98 @@ async def test_allocate_from_number_pool(
...
+    await np1.save(db=db)
+
+    branch = await create_branch(db=db, branch_name="feat")
+
+    # Created on the origin branch after the branch point.
+    manual_ticket = await Node.init(db=db, schema=TICKET.kind)
+    await manual_ticket.new(db=db, title="manual", ticket_id=5)
+    await manual_ticket.save(db=db)
+
+    assert await np1.get_taken(db=db, branch=branch, min_value=1, max_value=10) == {5}
```
ajtmccarty: this is the only test that calls `get_taken()` should  they all be calling `get_taken()` and/or should this one also test creating a new ticket instance on the default branch?

backend/tests/component/core/resource_manager/test_number_pool.py
```diff
@@ -52,6 +54,98 @@ async def test_allocate_from_number_pool(
     assert await np1.get_free(db=db, branch=default_branch) == 3
 
 
+async def test_allocate_skips_value_already_present_on_target(
```
ajtmccarty: I think these tests could be combined into 1 class based test that loads a schema with a unique and non-unique number pool-based attribute and then confirms that valid values are assigned when attributes are created/updated/deleted on the default and user branches. that would save some time that is currently spent on duplicated test setup

changelog/10179.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed `CoreNumberPool` allocation stalling on a value that already exists on the target kind but was created outside the pool. Previously the pool kept offering such a value, the attribute's uniqueness constraint rejected it, and because the failed allocation reserved nothing the pool re-offered the same value on every subsequent attempt, parking permanently at that value. When the target attribute is unique, the pool now skips values already present on the target and advances to the next free one, so pools can be declared over ranges that already contain data. Attributes that are not unique on their own (including per-relationship uniqueness constraints) are unaffected and remain fully allocatable.
```
ajtmccarty, line 1: I'm a little surprised that we allow a non-unique pool-backed attribute. wouldn't this just always return the first value in the pool?

backend/infrahub/core/query/resource_manager.py
iddocohen: Done. The initial match now groups by `n, attr` only, and the subquery returns the latest active value per attribute, so it no longer fans out over historical in-range values. The range predicate is re-applied after the subquery, so an attribute whose active value sits outside the range is dropped. I kept the in-range filter on the initial match as well, so the value index still narrows candidates before the per-attribute lookup.

backend/infrahub/core/query/resource_manager.py
iddocohen: Done, folded into the regrouping above: the initial match no longer binds `ha`/`hv`.

backend/infrahub/core/query/resource_manager.py
iddocohen: All four number-pool queries (`GetAllocated`, `GetUsed`, `GetFree`, `GetTaken`) take `pool: CoreNumberPool` and read `start_range`/`end_range`/`node_attribute`/`node` off it. Decoupling only `GetTaken` makes it inconsistent with its siblings for no functional gain, and decoupling all four is wider than this bugfix should carry. I'd rather track that as its own change. Want me to open an issue?

backend/tests/component/core/resource_manager/test_number_pool.py
iddocohen: The other tests cover `get_taken` indirectly through allocation. This one calls it directly because the branch-visibility case (an origin value added after the branch point) can't be observed through an allocation assertion. I can add a default-branch creation case here too if you'd like the direct call to cover both.

changelog/10179.fixed.md
iddocohen: Allocation advances by reservation, not by scanning the target, so a non-unique pool still counts up 1, 2, 3 rather than repeating the first value. Uniqueness only controls the extra skip of values already present on the target from outside the pool; without it, a pre-existing external value is simply reused. Whether a non-unique pool-backed attribute is desirable is a separate product question, but nothing here makes it stall on the first value.

backend/tests/component/core/resource_manager/test_number_pool.py
iddocohen: Extracted the shared setup: a `ticket_number_pool` fixture now holds the schema load, registry init and pool creation, and the three new same-setup tests use it. The non-unique test keeps its own setup since it needs a modified schema. I held off on collapsing everything into one parametrized matrix so each scenario stays readable as its own named test.

ajtmccarty: I'm not suggesting a parametrized matrix. I'm suggesting one class that loads the schema and data one time at the beginning using a class-based fixture, then runs the three tests as methods on the test class in a known order against the loaded schema/data. that would prevent needing to load the same schema/data 3 times.

ajtmccarty (approved):

backend/tests/component/core/resource_manager/test_number_pool.py
iddocohen: Ah I see - my misunderstanding. Will do the change. Thank you!

PR #10187, opened 2026-08-10
fix: reject artifacts whose transform returned no payload (closes #5303)

### LINKED ISSUE #5303: bug: Unhandled error if artifact is rendered as `None`
### Component

Git Integration

### Infrahub version

v1.1.0b2

### Current Behavior

When creating artifacts in a pipeline and you have an artifact definition that uses a transform that would return a `None` value instead of some payload there is an unhandled error in the worker. Which results in a failure message in the CI that reads:

```
Failed to render artifact: encoding without a string argument
```

### Expected Behavior

The error should instead say that the transform didn't return any payload.

### Steps to Reproduce

1. Create an artifact definition with a transform that doesn't return any data (this would typically happen if you have a try/except statement or conditional logic that returns early)
2. Try to generate some artifacts

In this location we need to check if "artifact_content" is `None` and if so raise an error:

https://github.com/opsmill/infrahub/blob/4d32ce6204e295dbd8f4ca2cf541db84c2c6fc41/backend/infrahub/git/integrator.py#L1338-L1356

This is the error we are seeing:

```python
>>> bytes(None, encoding="utf-8")
  File "<stdin>", line 1, in <module>
TypeError: encoding without a string argument
```

### Additional Information

_No response_

saltas888:
# Why

When an artifact definition uses a Python transform that returns nothing — the normal outcome of a
`try/except` or an early `return` — artifact rendering had nothing to serialize and never noticed.
The reporter saw the pipeline fail with `Failed to render artifact: encoding without a string argument`,
which says nothing about the real problem.

The symptom has since changed, and got worse. Commit `3bcf0e136` replaced the `text/plain` branch with a
`str()` fallback, so current `stable` raises nothing at all: `str(None)` produces the four-character
string `"None"`, which is checksummed, uploaded to the object store and saved with status `Ready`. The
defect degraded from a wrong error message into silent data corruption — for an `application/json`
artifact, into a body that is not even valid JSON.

Goal: fail the artifact with an error that names the transform that produced no payload.

Non-goals: the neighbouring `str()` fallback for non-`dict` JSON/YAML payloads is left alone (a transform
returning a `list` for `application/json` still yields a Python repr). Real, but a separate defect with a
different blast radius.

Closes #5303

## What changed

Behavioral changes:

- An artifact whose transform returns no payload now fails with
  `The transform at <location> did not return a payload` instead of storing the text `None` and reporting
  success. Surfaced by the artifact check as
  `Failed to render artifact: The transform at transforms/foo.py::Foo did not return a payload`.
- The artifact is left with status `Error` and nothing is uploaded to the object store.

Implementation notes:

- The serialization block was duplicated verbatim in `render_artifact` and `artifact_generate`, so a guard
  clause would have had to be written twice and could drift. It now lives in one module-level
  `serialize_artifact_content()` in `backend/infrahub/git/integrator.py`, which rejects a missing payload
  before serializing. Both call sites are validated by construction.
- The guard is `content is None`, not a truthiness test: `""` and `{}` are legitimate artifact payloads and
  must keep working. There is a test case pinning that.
- `TransformError` was chosen because it is what `execute_python_transform` already raises for its own
  failures, and because both callers of `render_artifact` already catch `Exception`, set the artifact status
  to `Error` and surface `str(exc)`. No change was needed in `artifacts/tasks.py` or `git/tasks.py`.
- In `artifact_generate` the Jinja2 branch now assigns `transformation_location` before use, so the helper
  can be given a location for both transform types.

What stayed the same: no schema changes, no GraphQL or REST API changes, no migrations. The check-message
plumbing in `backend/infrahub/artifacts/tasks.py` and `backend/infrahub/git/tasks.py` is untouched. The
`isinstance(..., dict)` dispatch semantics for JSON and YAML are unchanged, so existing artifact content
and checksums are unaffected.

## How to review

Start with `backend/infrahub/git/integrator.py` — the new `serialize_artifact_content()` and the two call
sites that replace the duplicated block. Everything else is test scaffolding: the component test in
`backend/tests/component/git/test_git_repository.py` drives the production path end to end, and
`backend/tests/unit/git/test_integrator.py` pins the serialization matrix around it, including the
empty-string case.

The part worth extra scrutiny is the deliberate absence of an object-store mock in the new test. Adding one
would make the test fail `pytest_httpx`'s "all registered responses were requested" check now that the fix
short-circuits before the upload — so its absence is what proves nothing is stored.

Alternative considered and rejected: pushing the check down into `execute_python_transform`. That task is
shared with webhooks, computed attributes and the REST transform endpoint, where a `None` result means
something different. "An artifact must have content" is a contract of artifact rendering, not of running a
transform.

## How to test

```bash
uv run pytest backend/tests/component/git/test_git_repository.py::test_render_artifact_python_without_payload -x -v
```

Passes on this branch. On `stable` it fails inside `integrator.py:1818` — `object_store.upload()` is called
with the string `"None"`, one line past the missing validation, which is the bug.

Wider run, no regressions:

```bash
uv run pytest backend/tests/unit/git backend/tests/component/git
```

`224 passed, 1 xfailed` for the git suites and `1407 passed` for the backend unit suite. `invoke format`
clean, `mypy` clean on `integrator.py`, `invoke backend.lint` reports the same 103 pre-existing `ty`
diagnostics as before this change (none in the files touched here — the two hits inside them are the
`from tests.conftest import TestHelper` warning that also appears in five sibling files). The generation
tasks (`backend.generate`, `schema.generate-*`, `docs.generate`) produce no diff, as expected for a change
that touches no schema, event, CLI or config surface.

## Impact & rollout

- **Backward compatibility:** a behavior change for anyone whose transform silently returned nothing. Those
  artifacts currently contain the text `None` and report success; they will now fail loudly with a named
  cause. That is the point of the fix, but it can surface pipelines that looked green.
- **Performance:** none.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (if user-facing or ops-facing change) — N/A, no documented behavior changes
- [ ] Internal .md docs updated — N/A, no architectural change to record

### DIFF
--- a/backend/tests/component/git/test_git_repository.py
+++ b/backend/tests/component/git/test_git_repository.py
@@ -14,6 +14,8 @@ from infrahub_sdk.node import InfrahubNode
 from infrahub_sdk.uuidt import UUIDT
 from pytest_httpx._httpx_mock import HTTPXMock
 
+from infrahub.auth.session import AnonymousSession
+from infrahub.context import BranchContext, InfrahubContext
 from infrahub.core.branch import Branch
 from infrahub.core.constants import InfrahubKind
 from infrahub.core.registry import registry
@@ -35,6 +37,7 @@ from infrahub.git.integrator import (
     ArtifactGenerateResult,
     CheckDefinitionInformation,
 )
+from infrahub.git.models import RequestArtifactGenerate
 from infrahub.git.sync import RepositoryFileImporter, RepositorySyncer
 from infrahub.git.worktree import Worktree
 from infrahub.lock import InfrahubLockRegistry
@@ -915,6 +918,50 @@ async def test_artifact_generate_jinja2_new(
         artifact_id=artifact_node_01.id,
     )
     assert result == expected_data
+
+
+async def test_render_artifact_python_without_payload(
+    client: InfrahubClient,
+    prefect_test_fixture: None,
+    git_repo_transforms_w_client: InfrahubRepository,
+    artifact_node_01: InfrahubNode,
+    mock_gql_query_03: HTTPXMock,
+) -> None:
+    repo = git_repo_transforms_w_client
+    commit_main = repo.get_commit_value(branch_name="main", remote=False)
+    branch = Branch(name="main", uuid=uuid4())
+    registry.branch[branch.name] = branch
+
+    message = RequestArtifactGenerate(
+        artifact_name="myartifact",
+        artifact_definition="c4908d78-7b24-45e2-9252-96d0fb3e2c78",
+        artifact_definition_name="artifactdef01",
+        commit=commit_main,
+        content_type="text/plain",
+        transform_type=InfrahubKind.TRANSFORMPYTHON,
+        transform_location="transform03.py::Transform03",
+        repository_id=str(repo.id),
+        repository_name=repo.name,
+        repository_kind=InfrahubKind.REPOSITORY,
+        branch_name=branch.name,
+        target_id="b663d7a4-5f95-48dd-b04d-e03169e7fcf3",
+        target_kind="TestElectricCar",
+        target_name="bolt",
+        query="my_query",
+        query_id="47800bff-adf1-450d-8388-b04ef2ffb129",
+        timeout=10,
+        variables={"name": "bolt"},
+        context=InfrahubContext(branch=BranchContext(name=branch.name), account=AnonymousSession()),
+    )
+
+    with pytest.raises(
+        TransformError, match=r"^The transform at transform03\.py::Transform03 did not return a payload$"
+    ):
+        await repo.render_artifact(artifact=artifact_node_01, artifact_created=True, message=message)
+
+    assert artifact_node_01.status.value == "Pending"
+    assert artifact_node_01.checksum.value is None
+    assert artifact_node_01.storage_id.value is None
 
 
 async def test_execute_python_transform_file_missing(
--- a/backend/tests/unit/git/test_integrator.py
+++ b/backend/tests/unit/git/test_integrator.py
@@ -2,8 +2,6 @@ from dataclasses import dataclass
 from typing import Any
 
 import pytest
-import ujson
-import yaml
 
 from infrahub.core.constants import ContentType
 from infrahub.exceptions import TransformError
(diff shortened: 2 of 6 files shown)


backend/tests/component/git/test_git_repository.py
```diff
@@ -917,6 +920,50 @@ async def test_artifact_generate_jinja2_new(
     assert result == expected_data
 
 
+async def test_render_artifact_python_without_payload(
```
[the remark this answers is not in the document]
saltas888: This review ran against the test-only commit, that red state is the intended first half of the bug pipeline's red-green sequence. The fix landed on the same PR right after in 0807ba467, render_artifact now goes through serialize_artifact_content() which raises TransformError when the transform returns no payload, so the test passes at the branch head, all backend-tests-component shards are green in CI. Adding xfail now would break the suite since the test passes.

backend/tests/unit/git/test_integrator.py
```diff
@@ -0,0 +1,73 @@
...
+    content_type: str
+    expected: str
+
+
+SERIALIZATION_CASES = [
+    SerializationCase(
+        name="dict_as_json",
+        content={"key1": "value1"},
+        content_type=ContentType.APPLICATION_JSON.value,
+        expected=ujson.dumps({"key1": "value1"}, indent=2),
```
polmichel: I'm not sure about this, but I wonder whether we should assert against the concrete raw value instead of using a third-party library, which makes it unclear what we expect from the system. Same for the `SerializationCase` below, which uses `yaml.dump`.

polmichel (approved):
Nice fix, code is cleaner now. 
Just an optional comment regarding test good practices, but I am not even sure if that's better. My take is that it's slightly better since we are not sure to rely on these `ujson` or `yaml` middleware forever

backend/tests/unit/git/test_integrator.py
saltas888: Agreed, deriving the expectation with the same library call the implementation makes wasn't really asserting anything. Both cases now pin the raw strings in 20f86e1dd, the ujson and yaml imports are gone from the test.

PR #10189, opened 2026-08-10
fix: keep kinds with no read field as a kind-level dependency

gmazoyer:
# Why

Follow up of #10154 and #10162.

Any Python computed attribute whose query follows a relationship to a generic is recomputed on every schema change.

`TransformReadSet.from_read_fields` marks the whole set imprecise as soon as one kind reads no field. That case is not rare. When a query traverses a relationship whose peer is a generic, the analyzer reports every member kind, and the members the query reads nothing from come back empty. `RecomputeScoper` then selects the attribute unconditionally.

Same "we cannot tell, so do everything" behaviour on the schema change side that #10154 removed on the data change side.

## What changed

If the query reads no field from a kind, skipping that kind is the right answer. Kinds with an empty field set are now dropped from `read_fields` but kept in `read_kinds`. Adding or removing such a kind still triggers a recompute, a field change on it does not.

The empty set branch used to have a reason. The analyzer could not record a `hfid` read, so an hfid only read arrived as an empty set and had to be caught here. #10162 records `hfid` under its schema name, so `IMPRECISE_READ_FIELDS` covers it now and that reason is gone.

`Jinja2DependencyDeriver` in `computed_attribute/scoping.py` already had this shape. The two derivers agree again.

## One thing to look at

`test_kind_with_no_mapped_fields_marks_imprecise` pinned the old behaviour and flips. It now asserts the read set the analyzer really produces for a query through a generic, the one from the component test added in #10154.

## How to test

```bash
uv run pytest backend/tests/unit/computed_attribute/
uv run pytest --neo4j backend/tests/component/computed_attribute/
```

On my machine: 64 unit tests pass, 28 component tests pass, ruff and mypy clean.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


changelog/+ifc2986.fixed.md
ajtmccarty, line 1: not completely sure this is necessary for unreleased changes. if so, the file name should remove the IFC reference

ajtmccarty (approved):

PR #10198, opened 2026-08-10
fix: validate preference timezone at write and correct the UI source hint (closes #10174)

### LINKED ISSUE #10174: bug: preferences timezone accepts any string, and the UI then claims an unusable zone is in effect
### Component

API Server / GraphQL, Frontend UI

### Infrahub version

1.11.0b1

### Current Behavior

`InfrahubSetPreferences` accepts and persists any string as `timezone`. All of the following return `ok: true`, are stored verbatim, and are returned by `InfrahubEffectivePreferences` with `source: USER`: `Not/AZone`, `UTC+25`, `'; DROP TABLE--`, and a 300-character string.

The UI then misreports the result: with `Not/AZone` stored, the Preferences card displays `Not/AZone` and its info hint says **"Your preference."** — but timestamps render in the *browser's* timezone, because the frontend validates via `Intl` and silently discards the unusable zone.

### Expected Behavior

A non-IANA timezone is rejected at write. And where a stored zone cannot be applied, the UI does not claim it is in effect.

### Steps to Reproduce

1. `mutation { InfrahubSetPreferences(scope: USER, timezone: "Not/AZone") { ok timezone } }` → `ok: true`.
2. Open Account settings → Preferences: the Timezone field shows `Not/AZone`, and its info hint says "Your preference."
3. View a timestamp whose rendered *date* differs by zone (e.g. a DateTime attribute at `2026-01-15T22:00:00Z`): it renders `2026-01-15` (browser zone), whereas with `Asia/Tokyo` stored it renders `2026-01-16`.

### Additional Information

`date_format` is constrained by a closed enum at both the GraphQL and model layers; `timezone` is a bare `String` (`backend/infrahub/graphql/mutations/preferences.py:53`, vs the enum-typed `date_format` on line 52) assigned straight to `Optional[str]` (`core/preferences/models.py:32`). `zoneinfo` / `available_timezones` / `pytz` appear nowhere under `backend/infrahub`. Both the GraphQL description and `docs/…/managing-preferences.mdx:41` document an IANA contract that is unenforced. Constitution VI (`dev/constitution.md:150`) requires GraphQL input to be validated.

Mitigations, for triage: no injection risk (parameterised Cypher); nothing crashes (the frontend hardening in `shared/utils/date.ts:14-28` is deliberate and works); not reachable via the UI, since the picker only offers `Intl.supportedValuesOf('timeZone')` — realistic paths are SDK/API, a script, or a migration.

Three traps for whoever fixes this:

- **A model-level Pydantic validator would be actively harmful.** `Preference(...)` is constructed on *read* (`core/query/preference.py:55`, `StandardNode.from_db`), so existing bad rows would raise on read — and because user+global are fetched in one query, a single bad GLOBAL row would break effective-preference reads for **every** user. Reject at write; stay lenient on read.
- **No `tzdata` package is installed** (win32 marker only). Validation works solely because the image ships `/usr/share/zoneinfo` — verified 486 zones in the running container. On an image without it, `available_timezones()` is empty and a validator would reject everything. Adding `tzdata` is an "Ask First" dependency per AGENTS.md.
- **Server and browser zone sets are neither equal nor nested.** Verified: the container lacks `Europe/Kiev`, `Asia/Calcutta`, `US/Pacific` (backward aliases), while browser ICU enumerates `Europe/Kiev` and omits `UTC` from enumeration while still accepting it. A naive `available_timezones()` allowlist can reject a value the picker itself offered.

Because of that last point the frontend half is needed regardless: it cannot distinguish "invalid" from "valid but unknown to this browser", so its message should state what it *did* ("this browser cannot render X; times shown in Y") and never call the value invalid.

Minor related inconsistency: an empty-string timezone reads back as unset (`QueryResult.get_as_str` treats falsy as `None`, so it correctly falls through to GLOBAL) — but the mutation's own response in that same request still echoes `timezone: ""`. The write reply disagrees with every later read.

A regression test needs a positive case alongside the rejection, so a validator broken by an empty zone set cannot pass by rejecting everything.

Found while verifying user preferences (INFP-512) against `1.11.0b1`.

polmichel:
## Why

`InfrahubSetPreferences` accepted and persisted any string as `timezone` (`Not/AZone`, injection strings, ...), and the preferences screen then reported an unusable zone as "Your preference." while timestamps silently rendered in the browser's own zone.

**Goal:** reject a non-IANA `timezone` at write, and stop the UI claiming a stored zone is in effect when it cannot be applied.

Closes #10174

## What changed

**Behavioral changes**

- Setting a `USER` or `GLOBAL` preference `timezone` now rejects a value that is not a resolvable IANA zone — including implementation-defined keys like `localtime`, `posix/*`, `right/*` — with a `ValidationError` (HTTP 422); nothing is persisted on rejection.
- An empty-string `timezone` normalizes to unset, so the mutation's echoed value agrees with every later read (previously the write reply echoed `""` while reads treated it as unset).
- The preferences screen no longer labels a stored zone "Your preference." when the viewer's browser cannot render it; the info hint now reports the browser fallback ("This browser can't display X; times are shown in Y.").

**Implementation notes**

- Validation lives at the **write path only** (`normalize_timezone`).
- Validation resolves the value with `zoneinfo.ZoneInfo(...)` rather than checking membership in `available_timezones()`: the enumerated set is environment- and version-dependent. `ZoneInfoNotFoundError`, `ValueError`, and `OSError` (e.g. an over-long key) all map to a rejection. Implementation-defined keys (`localtime`, `posix/*`, `right/*`) are rejected before construction, so the outcome does not depend on which zone tree the runtime ships (verified: the slim runtime resolves `localtime`).
- The frontend reuses the existing `Intl`-based `supportedTimezone` probe (now exported) to decide renderability.

## How to review

- Backend: `backend/infrahub/core/preferences/validation.py` (the helper) and its single call site in `backend/infrahub/graphql/mutations/preferences.py`.
- Frontend: `frontend/app/src/entities/preferences/ui/preference-fields.tsx` (`timezoneSourceMessage`) and the `supportedTimezone` export in `frontend/app/src/shared/utils/date.ts`.
- Knowledge doc: `dev/knowledge/backend/preferences.md` updated to reflect write-only validation.

## How to test

```bash
# Backend unit (validation matrix: reject + accept + empty-normalization)
uv run pytest backend/tests/unit/core/test_preference_validation.py -v

# Backend component (mutation rejects a non-IANA zone and persists nothing)
uv run pytest backend/tests/component/graphql/mutations/test_preferences.py -v

# Frontend (tooltip reports browser fallback for an unrenderable stored zone)
cd frontend/app && pnpm test -- --run src/entities/preferences/ui/user-preferences-card.test.tsx
```
## Checklist

- [x] Tests added/updated

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10198?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/preferences/validation.py
+++ b/backend/infrahub/core/preferences/validation.py
@@ -4,9 +4,9 @@ from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from infrahub.exceptions import ValidationError
 
-# Implementation-defined entries browsers reject; rejected explicitly so the guarantee holds
-# regardless of which zone tree the runtime ships.
-_REJECTED_KEYS = frozenset({"localtime"})
+# Implementation-defined entries a full zone tree resolves but browsers reject; listed explicitly
+# so they are refused regardless of which zone tree the runtime ships.
+_REJECTED_KEYS = frozenset({"localtime", "posixrules"})
 _REJECTED_PREFIXES = ("posix/", "right/")
 
 
--- a/backend/infrahub/core/preferences/validation.py
+++ b/backend/infrahub/core/preferences/validation.py
@@ -10,14 +10,14 @@ _REJECTED_KEYS = frozenset({"localtime", "posixrules"})
 _REJECTED_PREFIXES = ("posix/", "right/")
 
 
-def normalize_timezone(value: str | None) -> str | None:
-    """Return a stored-ready timezone, rejecting anything a client should not store.
+def validate_timezone(value: str | None) -> str | None:
+    """Validate a timezone for storage; empty becomes None, a valid value is returned unchanged.
 
-    An empty value is normalized to None (nothing stored) so the write reply agrees with every
-    later read, which treats a falsy timezone as unset. A non-empty value must resolve against the
-    runtime's zone database and not be an implementation-defined entry; validating by construction
-    (rather than an enumerated allowlist) keeps the check independent of an enumeration that varies
-    by runtime and avoids a filesystem walk.
+    A non-empty value must resolve against the runtime's zone database and not be an
+    implementation-defined key. It is checked by construction rather than an allowlist, so
+    backward-compatibility aliases are accepted and stored as given. The accepted set follows the
+    interpreter's own tzdata, so confirm new accept/reject behavior against a running instance,
+    not by unit test alone.
 
     Raises:
         ValidationError: the value is non-empty but is not a storable IANA timezone.
--- a/backend/infrahub/core/preferences/validation.py
+++ b/backend/infrahub/core/preferences/validation.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+from infrahub.exceptions import ValidationError
+
+
+def normalize_timezone(value: str | None) -> str | None:
+    """Return a stored-ready timezone, rejecting anything the runtime cannot resolve.
+
+    An empty value is normalized to None (nothing stored) so the write reply agrees with every
+    later read, which treats a falsy timezone as unset. A non-empty value must resolve against the
+    runtime's zone database; validating by construction accepts every zone that has a data file
+    (backward-compatibility aliases included), rather than an enumerated allowlist that would reject
+    aliases a client legitimately offers.
+
+    Raises:
+        ValidationError: the value is non-empty but names no timezone the runtime can resolve.
+
+    """
+    if not value:
+        return None
+    try:
+        # OSError covers keys the resolver rejects at the filesystem layer (e.g. an over-long name);
+        # ValueError covers malformed keys; ZoneInfoNotFoundError covers well-formed-but-absent ones.
+        ZoneInfo(value)
+    except (ZoneInfoNotFoundError, ValueError, OSError) as exc:
+        raise ValidationError(input_value=f"'{value}' is not a valid IANA timezone") from exc
+    return value
--- a/backend/infrahub/core/preferences/validation.py
+++ b/backend/infrahub/core/preferences/validation.py
@@ -4,22 +4,29 @@ from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from infrahub.exceptions import ValidationError
 
+# Implementation-defined entries browsers reject; rejected explicitly so the guarantee holds
+# regardless of which zone tree the runtime ships.
+_REJECTED_KEYS = frozenset({"localtime"})
+_REJECTED_PREFIXES = ("posix/", "right/")
+
 
 def normalize_timezone(value: str | None) -> str | None:
-    """Return a stored-ready timezone, rejecting anything the runtime cannot resolve.
+    """Return a stored-ready timezone, rejecting anything a client should not store.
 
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/preferences/validation.py
```diff
@@ -0,0 +1,36 @@
...
+
+    """
+    if not value:
+        return None
+    if value in _REJECTED_KEYS or value.startswith(_REJECTED_PREFIXES):
+        raise ValidationError(input_value=f"'{value}' is not a valid IANA timezone")
+    try:
+        # OSError covers keys the resolver rejects at the filesystem layer (e.g. an over-long name);
+        # ValueError covers malformed keys; ZoneInfoNotFoundError covers well-formed-but-absent ones.
+        ZoneInfo(value)
```
saltas888, line 33: are we missing a return here? And are we normalizing or validating the timezone here?

frontend/app/src/entities/preferences/ui/preference-fields.tsx
saltas888, line 103: should we do the same here?

backend/tests/component/graphql/mutations/test_preferences.py
```diff
@@ -189,6 +189,29 @@ async def test_user_rejects_unknown_date_format(
...
+    rejected at the write path and no Preference row is created.
+    """
+    result = await run_mutation(
+        db=db,
+        branch=default_branch,
+        account_session=session_first_account,
+        variables={"scope": "USER", "timezone": "Not/AZone"},
+    )
+    assert result.errors is not None
+    assert len(result.errors) == 1
```
ajtmccarty, line 211: maybe assert the text of the error is as expected

ajtmccarty (approved):
backend looks good

backend/tests/component/graphql/mutations/test_preferences.py
polmichel: Good point, this is fixed

backend/infrahub/core/preferences/validation.py
polmichel: This is a validation step

frontend/app/src/entities/preferences/ui/preference-fields.tsx
polmichel: No we do not since:

1. It's already handled in `formatWithPreferences` which makes sure the value is correct.
2. We modified in `formatWithPreferences` because the software was lying in that case: it was pretending using the user defined format, but silently used the browser fallback. this is not the same use case here

saltas888 (approved):
looks good!

backend/infrahub/core/preferences/validation.py
polmichel: Thanks for the feedback, I'll change the name of the method which is not intuitive. I'll go for `validate_timezone` format and I'll trim a bit this PR description

PR #10203, opened 2026-08-10
refactor(frontend): unify border colours onto semantic tokens

bilalabbad:
# Why

Tailwind v4's preflight resets `border-color` to `currentColor` via `border: 0 solid`, so every border in the app needed an explicit colour class. Four neutral families had drifted across ~150 call sites for the same visual role — `border-gray-200` ×62, `border-gray-300` ×22, `border-stone-300` ×14, `border-neutral-300` ×13, plus `divide-gray-200` ×15 and friends. Picking a border meant picking from four greys, and dark mode would have meant editing all 150.

Goal: one default colour, applied automatically, so `border-b` alone renders the right line and changing it app-wide is a one-line edit.

Non-goals: dark mode itself, `packages/schema-visualizer` (submodule — needs its own upstream PR), `ring-*`/`outline-*` colours, and state/brand borders.

## What changed

**Behavioural changes**

- Borders and dividers across the app now share one colour instead of four near-identical greys.
- Divider and separator **components** — `Separator`, `MenuSeparator`, `Toolbar.Divider`, `ToolbarDivider`, the IPAM header `Divider` — now track the tokens too. These paint their rule with `bg-*` rather than a border, so the border sweep was structurally blind to them and they had kept hardcoded greys.
- Six panels that had accidentally near-black `neutral-700` outlines on white (`data-viewer` ×4, `csv-table` ×2) plus one `divide-gray-600` (`attribute-display`) now match the rest of the UI.

Only the last bullet is a visible change; everything else is value-matched and should look identical.

**Implementation notes**

- `theme.css` gains `--border` (`neutral-200`) and `--border-strong` (`neutral-300`), exposed through `@theme inline`, plus a base-layer rule re-defaulting `border-color` so width utilities alone are enough. This also covers `divide-y`, which sets width only.
- Call sites swept value-matched — `*-100/200` → default tier, `*-300/400` → strong tier — which keeps the sweep visually neutral rather than turning a refactor into a redesign. The same rule decides the divider components' `bg-border` vs `bg-border-strong`.
- Inline one-off rules are deliberately left alone — the three `<div className="w-px bg-gray-300">` separators in the event filters are not divider components.
- `--card-border` deleted; `Card`/`CardHeader` use the shared token. An alias with one consumer and no divergence was dead weight.
- The base rule alone changes nothing visible — explicit utilities outrank a base-layer rule — which is why the sweep and the rule land together rather than in separate PRs.

**Deliberate exceptions** (3 raw palette borders remain, all intentional)

| Kept | Why |
|------|-----|
| `tooltip.tsx:25` `border-neutral-800` | Dark surface; belongs to the dark palette, revisited with dark mode |
| `global-filter-tag.tsx:21` `data-hovered:border-gray-600` | Hover emphasis — a signal, not chrome |
| `filter-suggestion-tag.tsx:18` `hover:border-stone-600` | Same |

The sortable-list drop indicator and the timeline rule keep their brand colours.

**What stayed the same**

No schema, API, or dependency changes. `border-transparent` (×32) untouched — it reserves layout space to prevent shift on hover/selection. State and brand borders (`red`, `custom-blue`, `cyan`, `amber`, `green`, `rose`) untouched.

## How to review

Start with `frontend/packages/ui/src/styles/theme.css` — the two tokens and the base rule are the whole design. Then `packages/ui/src/components/card/card.tsx` for the `--card-border` removal. The second commit (divider components) is 7 files and readable on its own. The remaining ~105 files are mechanical.

Worth a closer look: `shared/components/ui/badge.tsx`, `shared/components/table/style.tsx`, `packages/ui/src/components/button/button.tsx` and `shared/components/ui/popover.tsx` — colour-only variant strings where the class carried no width, so the class was removed entirely and the colour now comes from the base rule.

## Test plan

- `pnpm exec biome ci .` — clean
- `pnpm knip` — exit 0
- `pnpm exec betterer ci` — exit 0; 188 issues unchanged (`.betterer.results` refreshed because the edits shifted stored per-file hashes)
- `pnpm test` — 176 files / 1190 tests pass
- `pnpm build` — verified in the emitted CSS: the base rule (`*,:after,:before,::backdrop`), `.border-border`, `.border-border-strong`, `.bg-border`, `.bg-border-strong`, and `:where(.divide-border-strong>:not(:last-child))`
- **Chromatic** covers the `border-strong` tier (select, autocomplete, checkbox, checkbox-card, modal, popover, card) and `MenuSeparator` — expect no diffs
- **Manual pass needed** where nothing is tested: object list, object details, branches table, IPAM, proposed changes, tasks, schema browser. Also confirm none of the six `data-viewer`/`csv-table` dark outlines was deliberate.


frontend/app/src/shared/libs/graphiql/parallel-mode-plugin.tsx
```diff
@@ -101,7 +101,7 @@ function ParallelModePlugin() {
                       updateConfig({ maxConcurrent });
                     }
                   }}
-                  className="mt-1 block w-full rounded border border-gray-300 px-3 py-2 text-sm focus:border-blue-500 focus:outline-none focus:ring-1 focus:ring-blue-500"
+                  className="mt-1 block w-full rounded border border-border-strong px-3 py-2 text-sm focus:border-blue-500 focus:outline-none focus:ring-1 focus:ring-blue-500"
```
saltas888, line 104: should we create a border-border-focus var?

frontend/packages/graph/src/components/export-menu/export-menu.tsx
```diff
@@ -65,7 +65,7 @@ export function ExportMenu({ onExport, label = "Export diagram" }: ExportMenuPro
       {open && (
         <div
           id={menuId}
-          className="absolute bottom-full left-1/2 mb-2 min-w-[120px] -translate-x-1/2 rounded-lg border border-gray-200 bg-white py-1 shadow-lg"
```
saltas888, line 68: why are we just removing this?

frontend/packages/ui/src/components/button/button.tsx
```diff
@@ -61,7 +61,7 @@ const buttonVariants = tv({
         "data-pressed:bg-neutral-200",
       ],
       input: [
-        "border-neutral-200 bg-white text-stone-800",
+        "bg-white text-stone-800",
```
saltas888, line 64: why are we just removing this?

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
```diff
@@ -86,7 +86,7 @@ function DataViewerContent({
     case "image/bmp":
     case "image/x-icon": {
       return (
-        <div className="flex justify-center rounded-lg border border-neutral-700 bg-white p-4">
```
saltas888, line 89: also here?

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
```diff
@@ -74,7 +74,7 @@ function DataViewerContent({
         <iframe
           src={`data:application/pdf;base64,${content}`}
           title="PDF Preview"
-          className="h-150 w-full rounded-lg border border-neutral-700"
+          className="h-150 w-full rounded-lg border"
```
saltas888: and here

frontend/app/src/entities/nodes/object/ui/filters/filter-tag.tsx
```diff
@@ -29,13 +29,13 @@ export function FilterTag({ label, value, condition, ref, ...props }: FilterTagP
...
         <>
-          <Separator orientation="vertical" className="h-6 bg-stone-300" />
+          <Separator orientation="vertical" className="h-6" />
           <span>{condition}</span>
         </>
       )}
       {(value || value === 0 || value === false) && (
         <>
-          <Separator orientation="vertical" className="h-6 bg-stone-300" />
+          <Separator orientation="vertical" className="h-6" />
```
saltas888, line 38: is the bg-color removal intentional here?

saltas888 (approved):
Approving with comments

frontend/app/src/entities/nodes/object/ui/filters/filter-tag.tsx
bilalabbad: yes

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
bilalabbad: good catch, this one no

frontend/app/src/shared/components/data-viewer/data-viewer.tsx
bilalabbad: yes

frontend/packages/ui/src/components/button/button.tsx
bilalabbad: because it's the new default border color

frontend/app/src/shared/libs/graphiql/parallel-mode-plugin.tsx
bilalabbad: maybe in the future! I need to investigate first on tokenizing highlight colors

PR #10209, opened 2026-08-11
fix: create branches imported from git with sync_with_git enabled [IFC-3001]

### LINKED ISSUE #10208: bug: branches imported from a Git repository are created with sync_with_git disabled, so merging them skips the Git side
## Component

Git Integration

## Infrahub version

1.10.6 (all 1.10.x releases are affected, as are `stable`, `release-1.10`, `release-1.11` and `develop`; 1.9.x is not affected)

## Current Behavior

When the Git agent discovers a new branch on a repository remote and creates the corresponding
branch in Infrahub, that branch is created with `sync_with_git = False`.

`InfrahubRepositoryBase.create_branch_in_graph` (`backend/infrahub/git/base.py:655`) does not pass
`sync_with_git` and relies on the SDK default:

    branch = await self.sdk.branch.create(branch_name=branch_name)

The SDK default flipped from `True` to `False` in
opsmill/infrahub-sdk-python#1025 (commit `9b75f1c`, 2026-05-25). The SDK always sends the value
explicitly in the mutation input (`infrahub_sdk/branch.py:121`), so the server-side default
(`BranchCreateModel.sync_with_git = True`, `backend/infrahub/graphql/mutations/models.py:10`) never
applies -- the mutation receives an explicit `False`.

The regression reached Infrahub with the submodule bump `f2c63f085`
("chore(sdk): sync python_sdk to infrahub-develop", 2026-06-02).

A large amount of downstream logic is gated on `sync_with_git`, and all of it now takes the wrong
branch for a Git-originated branch on an ACTIVE managed repository:

- `backend/infrahub/core/merge/repository_merge_dispatcher.py:81` -- **the most severe effect**.
  `GIT_REPOSITORIES_MERGE` is only submitted when `source_branch.sync_with_git` is true or the
  repository is in `STAGING`. With `False` on an ACTIVE repository the workflow never runs, so
  `repo.merge(source_branch, dest_branch)` (`backend/infrahub/git/tasks.py:774`) is skipped
  entirely: the Git branch is never merged into the repository default branch, nothing is pushed to
  the remote, and the commit recorded on `main` never advances.
- `backend/infrahub/proposed_change/tasks.py:1190` -- `_validate_repository_merge_conflicts` is
  skipped, so a proposed change can report as mergeable while the underlying Git branches conflict.
- `backend/infrahub/proposed_change/tasks.py:628` and `:734` -- internal repository checks and
  repository test execution are skipped for the proposed change.
- `backend/infrahub/git/tasks.py:593` -- artifact generation resolves the repository from `main`
  instead of the branch, so artifacts can be produced from the wrong commit.
- `backend/infrahub/core/branch/tasks.py:417` -- `delete_git_branch_after_merge` becomes a no-op for
  these branches.
- `backend/infrahub/core/branch/creator.py:85` -- `GIT_REPOSITORIES_CREATE_BRANCH` is not submitted,
  so the branch is not created in the *other* repositories (the originating repository still gets it
  via `backend/infrahub/git/repository.py:181`).
- `backend/infrahub/git/base.py:821-825` -- when `import_sync_branch_names` is configured, an
  existing branch without `sync_with_git` is only kept on subsequent syncs if it still matches a
  configured pattern.
- `BranchCreatedEvent` / `BranchDeletedEvent` carry `sync_with_git=False` for these branches, so any
  consumer gated on the flag also misbehaves.

## Expected Behavior

A branch that originates from a Git repository is created in Infrahub with `sync_with_git = True`.
Merging such a branch merges the corresponding Git branch into the repository's default branch,
pushes it to the remote, and advances the commit recorded on the destination branch -- as it did
before 1.10.0.

## Steps to Reproduce

1. Add a `CoreRepository` pointing at a Git remote and let it sync.
2. Push a new branch to that remote (matching `import_sync_branch_names` if that setting is
   configured).
3. Wait for the repository sync to import the branch into Infrahub.
4. Inspect the branch, e.g. `infrahubctl branch list`, or:

       query { Branch(name: "<branch>") { name sync_with_git } }

   Observed: `sync_with_git` is `false`. Expected: `true`.

5. Commit a change on that branch in Git, create a proposed change from it and merge it. The
   Infrahub-side data merges, but the Git branch is not merged into the repository default branch
   and the repository commit on the destination branch is unchanged.

### Automated reproduction

The following integration test fails on the affected versions and passes once the branch is created
with the flag enabled. Note that a `INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES` value exported in the
environment leaks into pytest and filters the remote branch out of the sync, so the test must be run
with that variable unset.

    env -u INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES \
      uv run pytest backend/tests/integration/git/test_sync_branch_flag.py

```python
from __future__ import annotations

from typing import TYPE_CHECKING

from git.repo import Repo

from infrahub.core.constants import InfrahubKind
from infrahub.core.node import Node
from infrahub.git import InfrahubRepository
from tests.helpers.test_app import TestInfrahubApp

if TYPE_CHECKING:
    from pathlib import Path

    from tests.helpers.file_repo import FileRepo

NEW_BRANCH = "branch-from-git"


class TestSyncBranchFlag(TestInfrahubApp):
    async def test_branch_imported_from_git_syncs_with_git(
        self,
        db: InfrahubDatabase,
        client: InfrahubClient,
        git_repo_car_dealership: FileRepo,
        git_repos_dir: Path,
    ) -> None:
        obj = await Node.init(schema=InfrahubKind.REPOSITORY, db=db)
        await obj.new(
            db=db,
            name=git_repo_car_dealership.name,
            description="test repository",
            location=git_repo_car_dealership.path,
        )
        await obj.save(db=db)

        repo = await InfrahubRepository.new(
            id=obj.id,
            name=git_repo_car_dealership.name,
            location=git_repo_car_dealership.path,
            client=client,
        )

        Repo(git_repo_car_dealership.path).git.branch(NEW_BRANCH, "main")

        collected = await repo.collect_pending_imports()
        assert [pending.infrahub_branch_name for pending in collected.imports] == [NEW_BRANCH]

        branch = await client.branch.get(branch_name=NEW_BRANCH)
        assert branch.sync_with_git is True
```

### Affected versions

Determined by checking the `python_sdk` submodule pointer at each tag for the presence of
SDK commit `9b75f1c`:

| Version | Affected |
| --- | --- |
| 1.9.0 - 1.9.10 | no |
| 1.10.0 - 1.10.6 | yes |
| 1.11.0 pre-releases | yes |
| `stable`, `release-1.10`, `release-1.11`, `develop` | yes |

The fix should target `stable`.

### Notes for whoever picks this up

- The SDK default change itself is reasonable -- a user-created branch should not sync to Git by
  default. The defect is that the Git agent depended on that default rather than stating its intent
  at the call site.
- Branches already imported while the defect was present carry `sync_with_git = False` in the graph
  and will not be corrected by a code fix alone; existing deployments may need those branches
  repaired.

ogenstad:
# Why

Branches discovered on a Git repository remote are created in Infrahub with `sync_with_git` disabled, so merging one silently skips the Git side: the Git branch is never merged into the repository's default branch, nothing is pushed to the remote, and the recorded commit never advances.

`create_branch_in_graph` did not pass `sync_with_git` and relied on the SDK client default, which flipped from `True` to `False` in opsmill/infrahub-sdk-python#1025 (`9b75f1c`). The SDK always sends the value explicitly in the mutation input, so the server-side default (`BranchCreateModel.sync_with_git = True`) never applies. The regression reached Infrahub with the submodule bump `f2c63f085` and affects every 1.10.x release; 1.9.x is unaffected.

Goal: a branch that originates from a Git repository is created with `sync_with_git` enabled, so merges, repository checks and artifact generation behave as they did before 1.10.0.

Non-goals: repairing branches already imported with the flag disabled (see Impact & rollout), and changing the SDK default itself, which is correct for user-created branches.

Closes #10208

## What changed

Behavioral changes:

- A branch created in the graph by the Git agent for a branch found on a repository remote now has `sync_with_git` enabled. Merging it merges the corresponding Git branch into the repository's default branch and advances the recorded commit again.

Implementation notes: the flag is passed explicitly at the call site rather than relying on the client default, and the docstring records why, so it does not get pruned as redundant later. The eight downstream code paths gated on `sync_with_git` are untouched - they were behaving correctly given the input they received.

What stayed the same: no schema changes, no migration, no API contract change. The SDK default is unchanged, so branches created by users or by `infrahubctl` still default to not syncing with Git.

## How to review

The whole change is one keyword argument in `backend/infrahub/git/base.py`. The interesting part is the test, which exercises the real import path end to end rather than asserting on the call.

The test pins `import_sync_branch_names` to an empty list through the new `import_every_remote_branch` fixture. Without it, an `INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES` value exported in your shell leaks into pytest, filters the new remote branch out of the sync and fails the first assertion with an empty import list - a failure that looks like the fix, but is not.

## How to test

```bash
uv run pytest backend/tests/integration/git/test_sync_branch_flag.py -q

# neighbouring sync/merge coverage (not guarded against the env var, so clear it)
env -u INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES \
  uv run pytest backend/tests/integration/git/test_sync_merged_branch.py \
                backend/tests/integration/git/test_repository_branch.py -q
```

The new test fails on `stable` without the fix (`assert False is True`) and passes with it, including with `INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES='["nomatch-.*"]'` exported. The neighbouring tests pass both before and after (3 passed). `ruff format --check` and `ruff check` are clean on the changed files.

## Impact & rollout

- **Backward compatibility:** No breaking changes. Branches imported **while the defect was present** keep `sync_with_git = False` and are not repaired by this PR. There is no reliable way to identify them - `Branch` carries no provenance, so an imported branch and a deliberately non-syncing user branch are indistinguishable on the node - and `BranchUpdateInput` accepts only `description`, so the flag cannot be changed after creation via the API. Repairing one means re-importing the branch or writing the property directly in the database. Wrongly flipping a user branch would be worse than leaving it, so no migration is proposed.
- **Performance:** None.
- **Deployment notes:** Safe to deploy. Targets `stable`.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (if user-facing or ops-facing change) - not needed, restores documented behaviour
- [ ] Internal .md docs updated (internal knowledge and AI code tools knowledge) - not needed, no new concepts

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10209?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -653,11 +653,15 @@ class InfrahubRepositoryBase(BaseModel, ABC):
     async def create_branch_in_graph(self, branch_name: str) -> BranchData:
         """Create a new branch in the graph.
 
+        The branch originates from a git repository, so it must sync with git: merging it, running
+        its repository checks and generating its artifacts are all gated on that flag. It is passed
+        explicitly rather than left to the client default, which does not sync with git.
+
         NOTE We need to validate that we are not gonna end up with a race condition
         since a call to the GraphQL API will trigger a new RPC call to add a branch in this repo.
         """
         # TODO need to handle the exception properly
-        branch = await self.sdk.branch.create(branch_name=branch_name)
+        branch = await self.sdk.branch.create(branch_name=branch_name, sync_with_git=True)
 
         log.debug(f"Branch {branch_name} created in the Graph", repository=self.name, branch=branch_name)
         return branch
--- a/backend/tests/integration/git/conftest.py
+++ b/backend/tests/integration/git/conftest.py
@@ -214,3 +214,16 @@ def delete_git_branch_after_merge_reset_config() -> Generator[None, None, None]:
     original = config.SETTINGS.git.delete_git_branch_after_merge
     yield
     config.SETTINGS.git.delete_git_branch_after_merge = original
+
+
+@pytest.fixture
+def import_every_remote_branch() -> Generator[None, None, None]:
+    """Import every remote branch, whatever INFRAHUB_GIT_IMPORT_SYNC_BRANCH_NAMES holds in the ambient environment.
+
+    An empty list disables branch-name filtering. Without this, a value exported in the developer's
+    shell leaks into the test process and silently drops the branches the test relies on.
+    """
+    original = config.SETTINGS.git.import_sync_branch_names
+    config.SETTINGS.git.import_sync_branch_names = []
+    yield
+    config.SETTINGS.git.import_sync_branch_names = original
--- a/backend/tests/integration/git/test_sync_branch_flag.py
+++ b/backend/tests/integration/git/test_sync_branch_flag.py
@@ -1,0 +1,59 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from git.repo import Repo
+
+from infrahub.core.constants import InfrahubKind
+from infrahub.core.node import Node
+from infrahub.git import InfrahubRepository
+from tests.helpers.test_app import TestInfrahubApp
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from infrahub_sdk import InfrahubClient
+
+    from infrahub.database import InfrahubDatabase
+    from tests.helpers.file_repo import FileRepo
+
+NEW_BRANCH = "branch-from-git"
+
+
+class TestSyncBranchFlag(TestInfrahubApp):
+    async def test_branch_imported_from_git_syncs_with_git(
+        self,
+        db: InfrahubDatabase,
+        client: InfrahubClient,
+        git_repo_car_dealership: FileRepo,
+        git_repos_dir: Path,
+    ) -> None:
+        """A branch discovered on the remote must be created with sync_with_git enabled.
+
+        Merging the branch, running its repository checks and generating its artifacts are all gated
+        on that flag, so a branch imported without it silently skips the git side of a merge.
+        """
+        obj = await Node.init(schema=InfrahubKind.REPOSITORY, db=db)
+        await obj.new(
+            db=db,
+            name=git_repo_car_dealership.name,
+            description="test repository",
+            location=git_repo_car_dealership.path,
+        )
+        await obj.save(db=db)
+
+        repo = await InfrahubRepository.new(
+            id=obj.id,
+            name=git_repo_car_dealership.name,
+            location=git_repo_car_dealership.path,
+            client=client,
+        )
+
+        # The branch is created after the clone so the sync treats it as a new remote branch.
+        Repo(git_repo_car_dealership.path).git.branch(NEW_BRANCH, "main")
+
+        collected = await repo.collect_pending_imports()
+        assert [pending.infrahub_branch_name for pending in collected.imports] == [NEW_BRANCH]
+
+        branch = await client.branch.get(branch_name=NEW_BRANCH)
+        assert branch.sync_with_git is True
--- a/backend/tests/integration/git/test_sync_branch_flag.py
+++ b/backend/tests/integration/git/test_sync_branch_flag.py
@@ -27,6 +27,7 @@ class TestSyncBranchFlag(TestInfrahubApp):
         client: InfrahubClient,
         git_repo_car_dealership: FileRepo,
         git_repos_dir: Path,
+        import_every_remote_branch: None,
     ) -> None:
         """A branch discovered on the remote must be created with sync_with_git enabled.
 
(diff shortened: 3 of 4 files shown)


polmichel (approved):
LGTM, the root cause, the fix and the test are clear.

PR #10218, opened 2026-08-11
chore: remove ruff ignore rules that no longer suppress anything

ogenstad:
## What changed

Removed 73 lines of dead `ruff` configuration from the root `pyproject.toml`.

Every ignore entry was tested mechanically: drop it from the full config, re-run `ruff check .`, keep the removal only if ruff stays clean. Nothing here relies on judgement about whether a rule "should" apply.

### Whole blocks removed

Their last remaining rule was dead, or the block was already empty:

- `backend/infrahub/core/node/__init__.py` — empty list, ignored nothing
- `backend/infrahub/core/query/**.py` (UP007), `core/query/relationship.py` (PLR0915), `core/schema/manager.py` (PLR0915)
- `backend/tests/integration_docker/test_schema_migration.py` (ASYNC230), `backend/tests/component/graphql/test_schema_sort.py` (W293)
- `python_testcontainers/**.py` and `python_testcontainers/infrahub_testcontainers/helpers.py`

The last two were never consulted at all. `python_testcontainers/` carries its own `[tool.ruff]` config, so `ruff check --show-settings` resolves those files against `python_testcontainers/pyproject.toml` and the root patterns never appear in the resolved settings.

### Individual dead rules

`tests/e2e` (ARG003, S106) · `backend/tests/functional` (PT011) · `backend/tests/integration` (PT006, plus a duplicated PT007 line) · `backend/tests/component` (FURB118, RUF010) · `models/**` (SIM102, UP031) · `tasks/**` (UP031) · `utilities/**` (N806, PLR0912, UP031)

## What was deliberately left alone

**The global `ignore` list is unchanged.** D203/D213 (and ISC in the `python_testcontainers` config) test as removable only because ruff auto-disables them as incompatible with D211/D212 and the formatter. Dropping them would trade a clean run for permanent `warning: ... are incompatible` output.

**A pre-existing config bug, not addressed here.** In ruff's per-file-ignores globs `*` crosses `/`, so `backend/tests/*.py` is equivalent to `backend/tests/**.py` and matches the whole test tree recursively. Removing that block surfaces 601 violations in `backend/tests/unit/` alone. It sits directly below an explicit `backend/tests/**.py` block, so the two are duplicates with different rule lists, and it silently shadows roughly 45 entries in the narrower sub-blocks (`component`, `integration`, `functional`, `helpers`, `benchmark`, `adapters`, `scale`, `test_data`, `integration_docker`).

Those shadowed entries are individually deletable today, but only because the broad block covers them. Tightening `backend/tests/*.py` to be non-recursive would re-expose real violations across the test tree, so it needs its own decision and its own PR.

## Testing

- `ruff check .` clean from the repo root
- `ruff check .` clean inside `python_testcontainers/` under its own config

No changelog fragment: lint configuration only, no user-facing change.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -602,6 +602,10 @@ allow-dunder-method-names = [
     "UP035",    # `typing.List` is deprecated, use `list` instead
 ]
 
+"**/graphql_queries/*.py" = [
+    "E501",  # ruff check --fix produces a long single-line union before ruff format wraps it
+]
+
 "backend/infrahub/cli/db.py" = [
     ##################################################################################################
     # Refactor code and remove the ignore rule
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -602,10 +602,6 @@ allow-dunder-method-names = [
     "UP035",    # `typing.List` is deprecated, use `list` instead
 ]
 
-"**/graphql_queries/*.py" = [
-    "E501",  # ruff check --fix produces a long single-line union before ruff format wraps it
-]
-
 "backend/infrahub/cli/db.py" = [
     ##################################################################################################
     # Refactor code and remove the ignore rule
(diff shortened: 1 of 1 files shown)


saltas888 (approved):
awesome!

PR #10228, opened 2026-08-12
feat(backend): map uniqueness violations to a UNIQUENESS_VIOLATION code

ogenstad:
# Why

## The short version

The goal of this is to improve error reporting both within Infrahub but also to expose additional errors that the SDK can use. It's part of the effort to improve the sync logs with the git integration, but can of course be useful in other places as well. The SDK changes will come as a separate PR once this is in place but the idea is to expand the error catalogue with a UNIQUENESS_VIOLATION error that can return data to indicate the node_kind as well as the violating fields.

A duplicate create previously returned an uncatalogued error with an empty payload:

```json
{
  "message": "Violates uniqueness constraint 'name'",
  "path": ["TestingTagCreate"],
  "extensions": {
    "code": "UNDEFINED_ERROR",
    "http_status": 422,
    "data": {}
  }
}
```

The same mutation now returns:

```json
{
  "message": "Violates uniqueness constraint 'name'",
  "path": ["TestingTagCreate"],
  "extensions": {
    "code": "UNIQUENESS_VIOLATION",
    "http_status": 422,
    "data": { "node_kind": "TestingTag", "fields": ["name"] }
  }
}
```

Changes include:
* updates to error definitions
* management to add the error to the catalogue
* Updates where we raise these errors in the first place
* Tests to validate the expected behaviour
* Regeneration of frontend bindings and small changes to the code

## Longer AI summary

A uniqueness constraint violation on a GraphQL mutation returns `UNDEFINED_ERROR` with an empty `data` payload instead of a catalogued code, so consumers have to string-match `"Violates uniqueness constraint"` to detect a collision.

`UniquenessViolationError` had no catalogue entry, and no typed attributes to build a payload from. The raise sites already computed the node kind and the constraint members, formatted them into a message, and discarded the structured values. `resolve_catalogue_code` then walked the MRO to `ValidationError` (deliberately unmapped, since its routing is delegated to the typed `ATTRIBUTE_*` subclasses) and fell through.

Goal: uniqueness collisions surface as a distinct, catalogued code with a payload that names the colliding constraint members.

Non-goals: no SDK changes, and no consumer wiring. The `infrahub.git.integrator` change that turns this code into a concise repository-import failure is a follow-up PR.

Related to #7498

## What changed

Behavioral changes:

- Uniqueness violations on GraphQL mutations now return `extensions.code = "UNIQUENESS_VIOLATION"` with `data: {node_kind, fields}` instead of `UNDEFINED_ERROR` with `data: {}`.
- HFID collisions resolve to the same code, inherited through the formatter's MRO walk, since `HFIDViolatedError` subclasses `UniquenessViolationError`.
- `NodeAttributeUniquenessConstraint`'s message loses the trailing `" at <field>"` suffix, because its `input_value` moved from a dict to a plain string. That constraint is not wired into `NodeConstraintRunner`, so this is not observable on the production path.

Implementation notes:

- `UniquenessViolationError` gains required keyword-only `node_kind` and `fields`, and its input contract narrows from `str | dict | list` to `str`. Required rather than optional on purpose: an optional attribute that a raise site forgets would fail Pydantic validation *inside the error formatter*, turning a 422 into a 500 while formatting an error. There are only three raise sites and all three had the values in hand.
- The typed attributes live on the existing class in `backend/infrahub/exceptions.py` rather than a new subclass under `backend/infrahub/errors/`. `mutate_upsert` catches `HFIDViolatedError` for create-vs-update control flow, so forking that hierarchy would break the upsert path.
- The payload deliberately carries `fields: list[str]` rather than a single `field_name`: uniqueness constraints are multi-field and their members can be relationships. A consumer can branch on `fields.length` to decide between field-level and form-level presentation.
- `matching_nodes_ids` is deliberately **not** in the payload. It is the single most useful field for a "fetch the existing node instead" flow, but exposing it confirms the existence of a node the caller may have no read permission on (FR-013). Revisit alongside a permission-aware filter.
- `fields: list[str]` is the first array in the catalogue's schema vocabulary, so both generators gained array support: `frontend/app/scripts/generate-error-bindings.mjs` (previously threw on arrays by design) and the two helpers in `tasks/docs.py` (which would otherwise have rendered `"fields": null` in the worked example).

What stayed the same:

- The wire envelope is unchanged; this is an additive code.
- `http_status` stays 422 rather than a semantically correct 409, because `ValidationError.HTTP_CODE` is 422 and REST shares the exception class. Emitting 409 here would diverge from the REST body for the same failure. Recorded under "Known gaps" in the reference page.
- The `exc.__class__ is ValidationError` guard in `graphql/mutations/main.py` is untouched. It correctly lets already-typed subclasses through, and routing by class identity means uniqueness never needed the classification bridge.

## How to review

Start with `backend/infrahub/exceptions.py` and the three raise sites, then the catalogue entry and formatter arm. `schema/error-catalogue.json`, `frontend/app/src/shared/api/errors/catalogue.generated.ts`, and `docs/docs/reference/error-catalogue.mdx` are generated; regenerate with `uv run invoke frontend.regenerate-error-bindings`.

Worth extra scrutiny:

- `node_kind` at `attribute_uniqueness.py` uses `node_schema.kind` (the kind the caller submitted), not `comparison_schema.kind` (the schema the constraint is enforced against, which can be a generic parent). The submitted kind is what a consumer can act on, and it matches what the grouped path reports. `test_attribute_uniqueness_reports_the_submitted_kind_not_the_constraint_scope` pins this using `TestingContinent`, which inherits a unique `name` from the `TestingLocation` generic; it is the only fixture where the two kinds diverge.
- Whether `stability="evolving"` is right for a first release of this code.

Alternatives considered: folding this into `ATTRIBUTE_CONSTRAINT_VIOLATION`. Rejected because that payload carries a single `field_name`, and because the consumer action differs in kind. A regex failure means change the value; a collision means upsert, fetch the existing node, or pick a different value.

## How to test

```bash
uv run pytest backend/tests/unit/graphql/test_error_formatter.py backend/tests/unit/errors/ -q
uv run pytest backend/tests/component/core/constraint_validators/ -q
uv run pytest backend/tests/functional/graphql/test_error_catalogue.py -q
cd frontend/app && pnpm check:error-bindings
```

A duplicate create previously returned an uncatalogued error with an empty payload:

```json
{
  "message": "Violates uniqueness constraint 'name'",
  "path": ["TestingTagCreate"],
  "extensions": {
    "code": "UNDEFINED_ERROR",
    "http_status": 422,
    "data": {}
  }
}
```

The same mutation now returns:

```json
{
  "message": "Violates uniqueness constraint 'name'",
  "path": ["TestingTagCreate"],
  "extensions": {
    "code": "UNIQUENESS_VIOLATION",
    "http_status": 422,
    "data": { "node_kind": "TestingTag", "fields": ["name"] }
  }
}
```

The message and `path` are unchanged, so anything currently matching on the message text keeps working. The `422` in the "before" is not the catalogue's value: it is the `UNDEFINED_ERROR` fallback branch reading `HTTP_CODE` off `ValidationError`, which is itself evidence the exception reached the formatter as an unmapped subclass. It now comes from the catalogue entry instead.

`path` stops at the mutation field rather than extending to the offending input: the constraint is evaluated deep inside `save()`, and a composite constraint has no single offending field. This is a deliberate partial exception to FR-017, documented on the reference page; `data.fields` carries the equivalent information.

## Impact & rollout

- **Backward compatibility:** additive. Consumers already required to treat unrecognised codes as `UNDEFINED_ERROR` are unaffected, and anything previously matching on the message text still works since the wording is unchanged on the live path.
- **Deployment notes:** safe to deploy independently. No SDK release required. The follow-up integrator change reads `extensions.code` off the raw errors array that `GraphQLError` already bundles.

## Checklist

- [x] Tests added/updated


backend/infrahub/errors/catalogue.py
```diff
@@ -120,6 +122,19 @@ class CatalogueEntry(BaseModel):
                 exception_class=AttributeConstraintViolationError,
             ),
         ),
+        (
+            "UNIQUENESS_VIOLATION",
+            CatalogueEntry(
+                description=(
+                    "The submitted values collide with an existing node on a uniqueness constraint. "
```
ajtmccarty: I think we try to say "object" instead of "node" for user-facing stuff

backend/tests/component/core/constraint_validators/test_node_uniqueness.py
```diff
@@ -37,10 +40,36 @@ async def test_node_validate_constraint_attribute_uniqueness_failure(
...
+
+async def test_attribute_uniqueness_reports_the_submitted_kind_not_the_constraint_scope(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_internal_models_schema: SchemaBranch,
+    register_core_models_schema: SchemaBranch,
+) -> None:
+    # TestingContinent inherits `name` (unique) from the TestingLocation generic, so uniqueness is
+    # enforced against the generic. node_kind must still name the kind the caller submitted,
+    # otherwise a consumer gets back a kind it never mentioned.
```
ajtmccarty, line 58: maybe this is pre-existing, but I would think that `TestingLocation` would be more correct in this case. `TestingContinent` might not have any uniqueness constraints defined on it, so the error really is coming from the `TestingLocation` definition. maybe an issue for a different PR

backend/tests/functional/graphql/test_error_catalogue.py
```diff
@@ -124,6 +124,33 @@ async def test_multi_field_validation_returns_one_entry_per_failing_field(
...
 
+    async def test_uniqueness_violation_on_duplicate_create(
+        self,
+        initial_dataset: str,
+        test_client: InfrahubTestClient,
+        api_admin_token: str,
+    ) -> None:
+        # TestingTag.name is unique, so the second create collides. The constraint is evaluated
+        # deep inside save(), so `path` stops at the mutation field and `data.fields` carries the
+        # colliding members instead.
```
ajtmccarty: I think you can drop this whole comment. the test name and purpose seem pretty obvious

ajtmccarty (approved):
backend changes look good

backend/infrahub/errors/catalogue.py
ogenstad: Ah right, good point. Will push an update.

backend/tests/functional/graphql/test_error_catalogue.py
ogenstad: Yes, it's gone.

backend/tests/component/core/constraint_validators/test_node_uniqueness.py
ogenstad: Not sure, I think it might actually be easier for users if the error is reported with the node kind that the user tried to modify. But we can perhaps revisit later.

saltas888 (approved):

PR #10233, opened 2026-08-12
docs(objects): add Create objects overview page

BaptisteGi:
# Why

New users had no single place to see every way to create objects in Infrahub —
each method (web UI, CLI, YAML files, SDK, GraphQL, Generators, Sync) was only
discoverable by already knowing which page to look at.

## What changed

- Added `docs/docs/objects/create-objects.mdx`: a comparison table of all object-creation
  methods with guidance on when to use each, linked from the sidebar and from
  `manage-from-cli.mdx`.
- Removed the redundant "What you can do with objects" list from `overview.mdx` (superseded
  by the new page and existing sidebar).
- Shortened two page titles (`Manage objects from the command line` → `Manage objects with
  infrahubctl`, `Load data in bulk using YAML file` → `Load data using YAML file`) for
  consistency with the new page's table.

No schema, API, or code changes — docs only.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10233?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.vale/styles/spelling-exceptions.txt
+++ b/.vale/styles/spelling-exceptions.txt
@@ -32,6 +32,7 @@ Ceph
 changelog
 conftest
 Conftest
+cron
 check_color_tags_name
 check_definitions
 class_name
@@ -173,6 +174,7 @@ Observium
 OIDC
 Okta
 Onboarding
+onboarding
 OpenAPI
 order_by
 order_weight
--- a/docs/docs/objects/create-objects.mdx
+++ b/docs/docs/objects/create-objects.mdx
@@ -5,31 +5,31 @@ title: Create objects
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-How you create an object depends on what drives it. Create objects directly through one of Infrahub's interfaces, declare them in a file that lives in Git alongside your schema, or have a workflow create them from data that already exists.
+How you create an object depends on what starts the work. Create objects directly through an Infrahub interface, declare them in a YAML file stored in Git alongside your schema, or configure a workflow that creates them from data you already have.
 
 ## Create objects directly
 
-Infrahub has four interfaces, and all of them produce the same object and enforce the same schema constraints. Pick the one that fits how you are working.
+Four interfaces create objects. All four produce the same object and enforce the same schema constraints, so pick the one that matches how you are already working.
 
 The example below creates a device named `atl1-edge1` at site `atl1`. On `InfraDevice`, `name` and `type` are required and `site` is a relationship.
 
 <Tabs groupId="method" queryString>
   <TabItem value="web" label="Web interface" default>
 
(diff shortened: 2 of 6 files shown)


yjouffrault:
Pushed some changes to `create-objects.mdx` and the Objects hub. Summary of what changed from the original page and why.

### The page now groups the methods into three categories

The original listed seven creation methods in one table. The page now groups them by what starts the work — **directly**, **from a file**, **from a workflow** — with a short summary at the top saying what each is for, so a reader can go straight to the one they need.

The table came out. The useful content per method turned out to be a descriptive paragraph rather than a cell value, and three categories are easier to hold in your head than seven rows. Your "Use when" wording is preserved — it moved into the section and tab openers and got expanded there.

### Each method now shows how it works

The four direct interfaces are in a `<Tabs groupId="method">` block, matching `groups/create.mdx` and the rest of the site, so a reader's choice persists across pages. One running example — `InfraDevice` `atl1-edge1` — is shown four ways. Every method opens with when to use it, then the example or the mechanism, then where to read more.

### Fixes

- The web UI button is **Add Device**, not **Create \<Kind\>** — verified in `object-create-form-trigger.tsx:43`. The **Create Device** panel title is the string that comes after it.
- Generators and Infrahub Sync now explain *when* objects are actually created: per-target runs and their four triggers for Generators, and the `sync` run for Sync.
- The Infrahub Sync description says it's bidirectional. The original line ("Import from an external system") covers only one direction — it also publishes Infrahub data outward into monitoring and CMDB systems.

### Two additions

**Templates, Profiles, and resource pools** — folded into "After an object is created", since they determine where an attribute's value came from, which is what metadata and lineage records. This wasn't documented anywhere in the Objects section.

**`## In this section` on `objects/overview.mdx`** — the hub had lost its only links to its own child pages, so navigation depended entirely on the sidebar. Restored in the format `ipam/overview.mdx` and `schema/overview.mdx` use.

### Deliberately left out

CSV import. There's no CSV import feature or page today, so a row for it would link nowhere. It belongs with the *Import data from CSV* page when that lands.

### Two things for you

1. **"Open the Device list from the left menu"** (web interface tab) is the one claim I couldn't verify — the menu is schema-configurable, so that path may not match what a reader sees. Worth a check against a running instance.
2. **Tab order** is web → infrahubctl → SDK → GraphQL. `groups/create.mdx` uses web → graphql → sdk. Happy to match whichever you prefer.

Separately, `manage-from-cli.mdx:87,90` uses `--set location=DC1` on `InfraDevice`, which has a `site` relationship and no `location` — so that example fails as written. Left alone here since it predates this PR; happy to fix it in a follow-up.

Verified: `npm run build` passes (`onBrokenLinks: "throw"`, so all links resolve), markdownlint clean with the repo config. `docs.lint` and Vale weren't run locally — no `uv` or `vale` in my environment — so CI is the check on those.

yjouffrault (approved):

PR #10238, opened 2026-08-12
refactor: dispatch kind-update migrations by name [IFC-2619]

ajtmccarty:
# Why

The cubic review on #10224 raised four follow-ups. All of them land in files introduced by #10113, which was already merged, so none could be fixed at source and all were deferred. Two are real gaps rather than polish: the serialization that prevents two concurrent vertex duplications from each creating a replacement vertex had **no test at all**, and a node rename was repeating attribute-creation work already done by the inheritance migration.

Goal: close all four, and put a test behind the phase ordering so a regression there — a data-corruption class bug — cannot ship silently.

Non-goals: no change to what any migration does to the graph, and no change to any `SchemaPath`. Both were considered and deliberately rejected (see *How to review*).

## What changed

**Behavioral**

- A `node.name.update` / `node.namespace.update` that coincides with newly-gained inherited attributes no longer re-runs the `force_inherited` attribute-adds already performed by `node.inherit_from.update`. Fewer redundant writes; the end state is unchanged, since the creates were idempotent.
- Nothing else about migration behavior changes.

**Implementation notes**

- `schema_apply_migrations` orchestration moved out of the flow body into `SchemaMigrationsApplier`, which owns *policy* — phase split, concurrency cap per phase, error accumulation, the phase-one early return. The injected `SchemaMigrationExecutor` owns *mechanism*: `TaskSchemaMigrationExecutor` builds the `InfrahubBatch` for a group of requests with the cap it was handed and runs each through the Prefect task. The flow is now a thin composition root and resolves `get_database()` once instead of per migration — it is a singleton, so no behavior change.
- The three registered kind-update names now resolve to their own classes (`NodeInheritFromUpdateMigration`, `NodeNameUpdateMigration`, `NodeNamespaceUpdateMigration`). Only the first carries the attribute-add sub-migrations, so a rename structurally cannot run them — there is no conditional to get wrong. `NodeKindUpdateMigration` keeps the duplication and stays usable on its own, which ~26 call sites across 13 component test files rely on.
- Migration names are declared once as `MigrationIdentifier`, following the existing `ConstraintIdentifier` precedent in `core/validators/enum.py`. The enum is a general registry other names can join, so the phase split groups on `KIND_UPDATE_MIGRATION_NAMES`, an explicit subset, rather than on membership of the enum as a whole.

### Suggested review order

1. `backend/infrahub/core/migrations/schema/tasks.py` — the policy/mechanism split, and `KIND_UPDATE_MIGRATION_NAMES`.
2. `backend/infrahub/core/migrations/{enum.py,__init__.py}` plus `schema/node_kind_update.py` — the name→class dispatch.
3. `backend/tests/unit/core/migrations/schema/test_tasks.py` — the recording executor and the four orchestration tests.
4. The two test files under `component/` and `integration/` are adjustments to existing suites.

## Impact & rollout

- **Backward compatibility:** no breaking changes. No `SchemaPath`, migration-name, or persisted-data changes, so validator records and in-flight branches are unaffected.
- **Performance:** strictly less work — a rename no longer issues redundant attribute-create writes. Phase-one serialization is unchanged from #10113.
- **Deployment notes:** safe to deploy; no coordinated release needed.

## Checklist

- [x] Tests added/updated
- [x] [Changelog entry](../.agents/skills/creating-changelog-entries/SKILL.md) added — the two existing `.fixed` fragments were reworded so neither restates the other. No new fragment: the behavior being changed is unreleased.
- [ ] External docs updated — n/a, no user-facing or ops-facing change.
- [ ] Internal .md docs updated — checked and none apply: `dev/guides/backend/creating-migrations.md` covers *graph* migrations (`m0XX`, `GRAPH_VERSION`), and nothing under `dev/knowledge/` documents schema-migration dispatch.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/migrations/enum.py
gmazoyer, line 1: Should the value of this enum also be used in `backend/infrahub/core/models.py` around line 200. There is a `migration_name` parameter of an object that uses `"node.inherit_from.update"`.

backend/infrahub/core/migrations/schema/tasks.py
gmazoyer: Not from this PR but noticed that `Timestamp` is imported twice.

backend/infrahub/core/migrations/schema/tasks.py
```diff
@@ -40,38 +51,109 @@ def split_migrations_by_phase(
...
+    def __init__(self, database: InfrahubDatabase) -> None:
+        self.database = database
+
+    async def execute(
+        self,
+        requests: Sequence[SchemaMigrationRequest],
+        max_concurrent_execution: int | None = None,
+    ) -> list[SchemaMigrationPathResponseData]:
+        if max_concurrent_execution:
+            batch = InfrahubBatch(max_concurrent_execution=max_concurrent_execution)
```
ogenstad, line 93: I know the use of InfrahubBatch didn't get introduced in this PR, and I think it's fine with regards to how it's used here. But I'm wondering if we should take a step back and replace it with something else. While the typing in the SDK for InfrahubBatch.execute is a bit vague it's really returning optional `InfrahubNode` objects from the SDK. While we just ignore the "node" and return the results this approach seems like a code smell to me.

ogenstad (approved):
Approved with some thoughts for future changes or to start a discussion.

backend/infrahub/core/migrations/enum.py
ajtmccarty: updated to use the new enum everywhere

backend/infrahub/core/migrations/schema/tasks.py
ajtmccarty: removed the duplicate import

backend/infrahub/core/migrations/schema/tasks.py
ajtmccarty: added a note about removing `InfrahubBatch` to the tech debt item for cleaning up schema migrations https://opsmill.atlassian.net/browse/IFC-2399

PR #10242, opened 2026-08-13
fix(core): reserve headroom below the Prefect related-resources maximum [IFC-3008]

### LINKED ISSUE #10241: bug: mutation event related resources are truncated to exactly the Prefect maximum, so the EventsWorker's run-context append overflows it and kills the events websocket
### Component

API Server / GraphQL

### Infrahub version

1.11 (`release-1.11`, `0.0.0.dev0` build). The node path is affected since 1.10.2, the group path since 1.10.7.

### Summary

A single large update - roughly 250+ changed relationship peers on a node, or 500+ members on a group, in one mutation - produces an event the Prefect server refuses during validation. The websocket connection Infrahub uses to deliver events is closed as a result, and it does not come back on its own: no further events can be sent until the Infrahub server is restarted. The task manager does not need restarting; the problem is on Infrahub's side of the connection.

The mutation itself succeeds and the data is correct, so nothing surfaces to the caller. What stops is everything Infrahub drives off events: the Activity log, event rules and their actions, webhooks, and generators. Log forwarding runs on a separate path and is unaffected - a change that reaches your logs but never appears in the Activity log is a good signal you have hit this.

### Current Behavior

`NodeMutatedEvent.get_related()` (`backend/infrahub/events/node_action.py:119-129`) and
`GroupMutatedEvent.get_related()` (`backend/infrahub/events/group_action.py:67-77`) truncate to
**exactly** `get_prefect_max_related_resources()` (500 in the image, `development/Dockerfile:129`).

Prefect's `EventsWorker` then appends run-context related resources after the event has left us:

```python
# prefect/events/worker.py:116
event.related += await related_resources_from_run_context(...)
```

`Event.model_config` has no `validate_assignment`, so `+=` does not re-run
`_validate_related_resources` (`prefect/events/schemas/events.py:103-111`). Verified: appending 600
entries to an already-validated `Event` raises nothing. The event goes on the wire at 502+.

The server rejects it, and the rejection is not contained:

```python
# prefect/server/api/events.py:73-75
async for event_json in websocket.iter_text():
    event = Event.model_validate_json(event_json)   # no try/except
```

Only `NORMAL_DISCONNECT_EXCEPTIONS` is caught around that loop, so
`enforce_maximum_related_resources` (`prefect/server/events/schemas/events.py:170-181`) raising a
`ValidationError` propagates out of the ASGI handler and tears down the `/events/in` websocket.

Everything above is confirmed from a running stack (see the logs below).

**Why emission stops entirely rather than losing a single event.** `_emit` appends to
`_unconfirmed_events` before sending (`prefect/events/clients.py:507`) and never removes it on
failure, while `_reconnect` resends the whole buffer (`clients.py:439-440`). The rejected event is
therefore replayed on every reconnect and kills the socket again; after 11 attempts `_emit` raises
and the item is dropped from the queue with `Service 'EventsWorker' failed to process item
Event(...)`, but it remains in `_unconfirmed_events`, so subsequent events keep failing the same way.

Two observations support this. A node event failed roughly 57 minutes after it was emitted, so the
worker was an hour behind. And delivery was restored by restarting the **Infrahub server** while the
task manager kept running - the broken state therefore lived in the emitting process's memory, not
in the Prefect server, on disk, or in the database, which is what a poisoned in-memory buffer
predicts.

The chain has not been reproduced deliberately, and one detail remains open: the client also runs a
30-second time-based checkpoint (`_force_checkpoint`, `clients.py:467-482`) that clears
`_unconfirmed_events` unconditionally after a successful ping. Whether that can ever win the race
against the resend loop, and so recover without a restart, is unknown - it did not do so before the
restart was performed.

**The caps changed the failure mode, not just the threshold.** Before #9793 / #10163, an oversized
event was rejected client-side inside `emit_event`, swallowed by the bare `except Exception` at
`prefect/events/utilities.py:110-112`, and logged as `Error emitting event: ...` - exactly what #9794
and #10127 describe. One event lost, websocket unharmed. After the caps, the event passes client-side
validation at exactly 500 and is rejected server-side instead, which is where the connection-level
damage comes from.

Secondary defect visible in the same data: the run-context resources should not be attached to an
API-emitted event at all. The observed events carry `prefect.flow-run ... Setup Task Manager` and
`prefect.flow ... task-manager-setup`, i.e. a leaked `FlowRunContext` from
`backend/infrahub/workflows/initialization.py:125`. With no flow-run context,
`related_resources_from_run_context` returns `[]` and the overflow cannot happen.

### Expected Behavior

Related resources are trimmed to a budget that stays under the Prefect maximum after the
`EventsWorker` run-context append, so the event is accepted and the websocket stays alive.

Separately: a single rejected event should not be able to stop delivery of subsequent events.

### Steps to Reproduce

1. Stack from `release-1.11` with the image default `PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES=500`.
2. Perform a mutation producing more than 500 related resources on a single node or group event
   (a group membership change touching 500+ members, or a node with 250+ relationship peers - each
   peer contributes up to two related resources).
3. Infrahub logs `Truncating the related resources of a ... mutation event to the Prefect maximum`.
4. The Prefect server logs a `ValidationError` for `related` at `stream_events_in`, followed by
   `Exception in ASGI application`.
5. Infrahub logs `Service 'EventsWorker' failed to process item Event(...)` for that event, with a
   growing lag between the event's `occurred` timestamp and the log line.
6. Events emitted after it are no longer delivered. Restarting the Infrahub server restores
   delivery; the task manager does not need restarting.

Step 3 is not yet evidenced. Steps 4 and 5 are directly evidenced from a running stack (the
server-side `ValidationError`, with the appended `task-manager-setup` resource visible at the tail of
`input_value`; and a node event failing ~57 minutes after it was emitted). Step 6 is evidenced as an
outcome - delivery had stopped and came back after the restart - but the sequence has not been
driven deliberately from step 2.

### Additional Information

<details><summary>Prefect server log</summary>

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Event
related
  Value error, The maximum number of related resources is 500 [type=value_error,
  input_value=[{'prefect.resource.id': ...: 'task-manager-setup'}], input_type=list]
08:31:00.706 | ERROR | uvicorn.error - Exception in ASGI application
  ...
  File "prefect/server/api/events.py", line 74, in stream_events_in
    event = Event.model_validate_json(event_json)
```

</details>

<details><summary>Infrahub-side log</summary>

```
9:39:29.845 | ERROR | GlobalEventLoopThread | prefect._internal.concurrency -
Service 'EventsWorker' failed to process item Event(occurred=datetime.datetime(2026, 8, 13, 8, 42, 49, ...),
event='infrahub.node.updated', ... related=[... 'prefect.resource.name': 'Setup Task Manager' ...])
```

Note the ~57 minute gap between `occurred` and the log line: the worker was already an hour behind.
Also note that this line carries no traceback - `QueueService._main_loop` passes
`exc_info=logger.isEnabledFor(logging.DEBUG)` (`prefect/_internal/concurrency/services.py:210-217`),
so the cause is invisible unless Prefect debug logging is on. Worth enabling while investigating.

</details>

**Sizing the headroom.** `related_resources_from_run_context` appends up to 6 fixed entries (flow-run,
task-run, flow, deployment, work-queue, work-pool) plus **one per tag** in
`flow_run.tags | deployment.tags`. Our own tags come from `WorkflowTag`
(`backend/infrahub/workflows/constants.py:34`), so roughly 4-6 today, but the tag term is
structurally unbounded and a fixed `-6` would be wrong.

**Config-level mitigation, independent of any code fix.**
`PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES` and `PREFECT_EVENTS_MAXIMUM_RELATED_RESOURCES` are
two aliases of the same field (`server.events.maximum_related_resources`). Raising it on the
**task-manager** service only, while leaving the app and worker containers at the image's 500, makes
the server tolerant of the overflow. Any residual overflow is then rejected client-side, which is the
benign failure mode.

**Test gap.** The unit assertions (`test_node_action.py:97`, `test_group_action.py:304,320`) pin the
truncated count *to* the maximum, which is the behaviour in question.

The Docker reproducers cannot catch it either, but not because they stay small.
`backend/tests/integration_docker/test_node_event_related_resources.py` does truncate - 300 peers
produce roughly 600 entries - yet nothing in that environment leaks a flow-run context into the
emitting process, so Prefect appends no run-context resources and the event arrives at exactly the
maximum, which is accepted. `test_group_event_related_resources.py` stays under the cap entirely
after the consolidation in #10163. Reproducing this end to end needs a flow-run context on the
emitting side, which is the same condition as the `Setup Task Manager` leak above.

**Related.**

- #9794 / #9793 introduced the node cap (first in `infrahub-v1.10.2`).
- #10127 / #10163 introduced the group cap (first in `infrahub-v1.10.7`). #10163 explicitly deferred
  the `limits.py` 500-vs-100 default divergence, which is the same failure mode from the other
  direction: where the env var is unset, the server enforces 100 while we trim to 500.
- #9068 is adjacent but distinct - flow-run terminal events with thousands of related nodes leaving
  tasks stuck in `RUNNING`.

ogenstad:
# Why

Node and group mutation events truncated their related resources to **exactly** the configured
Prefect maximum (500 in the Infrahub image). Prefect's events worker then extends that list in place
with run-context resources - flow run, task run, flow, deployment, work queue, work pool, and one per
flow-run tag:

```python
# prefect/events/worker.py:116
event.related += await related_resources_from_run_context(...)
```

`Event.model_config` has no `validate_assignment`, so `+=` does not re-run the client-side
`_validate_related_resources`. The enlarged event goes on the wire above the maximum, and the server
rejects it inside a websocket handler that does not guard the parse
(`prefect/server/api/events.py:73-75`, only `NORMAL_DISCONNECT_EXCEPTIONS` is caught around the
loop). Uvicorn logs `Exception in ASGI application` and closes the transport, so a single oversized
event takes down the `/events/in` stream rather than being dropped on its own.

Observed on a running stack as a `ValidationError` for `related` on the Prefect server, with the
appended `task-manager-setup` resource visible at the tail of `input_value`.

Closes #10241 [IFC-3008]

Goal: an event that leaves Infrahub is still acceptable to the Prefect API after Prefect has enlarged
it.

Non-goals, both tracked separately:

- The leaked `FlowRunContext` from `setup_task_manager()` that attaches run-context resources to
  API-emitted events in the first place. It is the deeper cause; fixing it here would mask whether
  this change works.
- Prefect's own behaviour of never dropping a failed event from `_unconfirmed_events` while
  `_reconnect` resends the whole buffer, which is the suspected reason delivery stops entirely rather
  than losing one event. That belongs upstream.

## What changed

Behavioral change: both mutation event types truncate to a budget that sits below the Prefect
maximum instead of on it. At the image's 500 the budget is 450, so an already-truncated event carries
up to 50 fewer related resources. Those events are lossy by construction and the full peer or member
list stays in the payload changelog.

Implementation notes:

- `events/limits.py`: new `get_related_resource_budget()`, reserving a tenth of the maximum with a
  floor of 20. `get_prefect_max_related_resources()` is unchanged and still returns the raw maximum -
  it remains the honest reader of the Prefect setting and the base for `get_submission_chunk_size()`,
  which is a different constraint (flow-run parameter size) and keeps its 250.
- `events/node_action.py`, `events/group_action.py`: truncate to the budget; the warning log now
  reports `budget` rather than `maximum`.

### Why the headroom is sized this way

The fixed part of the append is at most 6 objects. The unbounded part is one related resource per tag
in `flow_run.tags | deployment.tags`. Two things bound it in practice:

- Deployment tags are namespace + workflow type + declared tags (`models.py:88-94`), so ~3, and no
  `submit_workflow` call site passes more than one extra tag at creation.
- `add_tags(nodes=all_ids)` in `computed_attribute/tasks.py:191` can add ~250 node tags, but it only
  PATCHes the server. `setup_run_context` (`prefect/flow_engine.py:1757`) refreshes the flow run once
  at run start, before `FlowRunContext` is entered, so tags added during the flow body never reach
  the object Prefect reads for the append.

That puts the realistic append at ~10. The reservation is deliberately much larger than that because
over-reserving costs a handful of related resources on events that are already truncated, while
under-reserving costs the event stream.

## How to review

- `events/limits.py` is the whole fix; the two call sites are one identifier each.
- In the tests, only three assertions changed - the ones that genuinely break.
  `test_related_resources_stay_within_prefect_maximum_for_large_relationships` is untouched on
  purpose: 450 ≤ 500 still holds and it still states the external contract.
- Extra scrutiny on the headroom sizing above. If you can find a path that creates a flow run with
  many tags up front, the reservation needs to grow.

## How to test

```bash
uv run pytest backend/tests/unit/event/
```

The new `test_event_on_the_budget_survives_the_prefect_run_context_append` builds a real Prefect
`Event` at the budget, applies the worker's in-place append, and validates it. It fails without this
change with `assert 516 <= 500`.

There is no end-to-end reproducer here. Triggering it for real needs a flow-run context on the
emitting side, which is the `Setup Task Manager` leak listed as a non-goal - the existing Docker
reproducers truncate but never see an append, which is why they pass today.

## Impact & rollout

- **Backward compatibility:** none affected. No schema, API, wire-format or config change.
- **Performance:** unchanged.
- **Deployment notes:** safe to deploy on its own. Independently of this change, raising
  `PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES` on the **task-manager** service only (both
  `PREFECT_SERVER_EVENTS_…` and `PREFECT_EVENTS_…` are aliases of the same field) makes the server
  tolerant of any residual overflow, which then fails client-side as a benign single-event drop.
- **Backport:** targeted at `release-1.11` for validation ahead of the release; the same commit
  applies to `stable`, where both truncation sites are identical.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (`changelog/10241.fixed.md`)
- [ ] External docs updated - n/a, no user-facing API change
- [x] Internal .md docs updated (`dev/knowledge/backend/events.md`)

[IFC-3008]: https://opsmill.atlassian.net/browse/IFC-3008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/events/limits.py
```diff
@@ -19,6 +20,20 @@ def get_prefect_max_related_resources() -> int:
...
+
+    Prefect's events worker appends run-context resources to an event after it has been handed
+    over - flow run, task run, flow, deployment, work queue, work pool, and one per flow-run tag -
+    by extending the list in place, which does not re-run the client-side validation. An event
+    that leaves on the maximum therefore arrives above it, and the Prefect API answers by closing
+    the event stream rather than by dropping the single event. The budget stays below the maximum
+    so the enlarged event is still accepted.
+    """
+    maximum = get_prefect_max_related_resources()
+    return max(1, maximum - max(_MIN_RUN_CONTEXT_HEADROOM, maximum // 10))
```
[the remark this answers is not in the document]
ogenstad: Valid. `MAX_RUN_CONTEXT_RESOURCES` now declares the append size once in `limits.py`, the floor derives from it, and the survival test imports it and runs at 500/200/100 so the floor is covered too.

changelog/10241.fixed.md
```diff
@@ -0,0 +1 @@
+Node and group mutation events no longer overflow the Prefect related-resources maximum after Prefect enlarges them. Events were truncated to exactly the configured maximum, and Prefect's events worker then appended its own run-context resources (flow run, flow, deployment, work queue, work pool and one per flow-run tag) to the list in place, which skips the client-side validation. The enlarged event was refused by the Prefect API, which closes the event stream rather than dropping the single event. Events are now truncated to a budget that leaves room for that append.
```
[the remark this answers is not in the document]
ogenstad: Applied.

backend/infrahub/events/limits.py
polmichel, line 50: I'm not sure this makes a real difference, but maybe we should change this one too, for consistency?

backend/tests/unit/event/test_limits.py
polmichel, line 1: Do you see any value here to add this test which demonstrate that an error is raised when the maximum has been reached?

```

@pytest.mark.parametrize("case", [pytest.param(case, id=case.name) for case in SURVIVAL_CASES])
def test_event_on_the_maximum_is_rejected_after_the_prefect_run_context_append(
    case: SurvivalCase, monkeypatch: pytest.MonkeyPatch
) -> None:
    """An event emitted on the maximum is refused once Prefect has enlarged it.

    This is the failure the budget exists to prevent, and the control that proves the reservation
    does real work: an event that leaves on the maximum has no room for the in-place run-context
    append, so the enlarged event lands above the limit and no longer validates. Building the same
    event on the budget instead is what the companion test shows still validates.
    """
    monkeypatch.setenv(ENV_VAR, str(case.configured_max))
    with temporary_settings({PREFECT_SERVER_EVENTS_MAXIMUM_RELATED_RESOURCES: case.configured_max}):
        event = Event(
            event="infrahub.node.updated",
            resource=Resource({"prefect.resource.id": "infrahub.node.abc"}),
            related=[
                RelatedResource(
                    {"prefect.resource.id": f"infrahub.node.{index}", "prefect.resource.role": "infrahub.related.node"}
                )
                for index in range(get_prefect_max_related_resources())
            ],
        )
        event.related += [
            RelatedResource({"prefect.resource.id": f"prefect.tag.{index}", "prefect.resource.role": "tag"})
            for index in range(MAX_RUN_CONTEXT_RESOURCES)
        ]

        assert len(event.related) > get_prefect_max_related_resources()
        with pytest.raises(ValidationError, match=r"The maximum number of related resources is \d+"):
            Event.model_validate(event.model_dump())


```

polmichel (approved):
LGTM, two optional comments

backend/infrahub/events/limits.py
```diff
@@ -2,6 +2,12 @@
 
 _DEFAULT_MAX_RELATED_RESOURCES = 500
 
+# Six fixed entries - flow run, task run, flow, deployment, work queue, work pool - plus one per
+# flow-run tag. Only tags present when the run was created reach an event: a run refreshes its tags
+# once before its context is entered, so anything a flow tags itself with later stays out. Infrahub
+# renders four tag kinds today; the rest of the allowance absorbs tags added later.
+MAX_RUN_CONTEXT_RESOURCES = 6 + 14
```
saltas888, line 9: I dont really get where the 14 is coming from. Is it 4 tag kinds + the rest (buffer)?

ogenstad: Yes pretty much.

backend/infrahub/events/limits.py
ogenstad: No, this is used for something else.

PR #10245, opened 2026-08-13
feat(regeneration): narrow relationship-reached changes to owning members (IFC-2946)

polmichel:
## What

Post-merge (and proposed-change) selective regeneration now narrows to the affected members when the changed data is read **through a relationship** — an owner's name, an interface's field, multi-hop chains, a kind reached by several relationships, and a change behind a generic peer — instead of regenerating every member of the definition.

## How

- **Analyzer** (`ReachedPathResolver`): reconstructs, per related kind, every relationship chain the query traverses down to it; a change reached through a generic peer is resolved to the generic's concrete implementations.
- **Classifier**: a new `RelationshipReachedChanges` assessment carrying those chains; it widens the whole assessment when a change can't be pinned (a generic owner along the chain, a fragment refinement, or a kind also read at a root).
- **Impact** (`ReachedMemberResolver`): walks each chain back to the owning members via the existing dependent-node resolver and unions them with the directly-changed members.

## Invariant

Over-execution is acceptable, under-execution is not: every hop returns a superset and any ambiguity widens to the whole group, so no member that genuinely needs regenerating is skipped.

## Known limitation

A change reached through a relationship whose **owner** is a generic still widens.

Example — a query reading a field on a kind reached through a relationship **defined on a generic**:

```graphql
query { Rack { edges { node {
  devices {              # peer is the generic InterfaceHolder -> resolved to concrete impls (narrows)
    endpoints {          # a relationship on the generic -> its peer kind is owned by a generic
      status { value }   # a change here widens the whole group
    }
  }
} } } }
```

## Tests

- Unit: the two resolvers in isolation (chains and widen cases including generic peer/owner; hop-chaining and multi-chain union, recording double, no mocks) and the classifier variants.
- Integration: real queries through the analyzer assert the reconstructed chains.
- Component: a relationship-reached change narrows to the owning member end-to-end on a real database.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/regeneration/impact.py
+++ b/backend/infrahub/core/regeneration/impact.py
@@ -118,7 +118,7 @@ class ReachedMemberResolver:
 
     Each hop resolves the current node ids to the owners referencing them, feeding the next hop, so
     a chain ends at the root members. Every hop returns a superset of the truly-related nodes, so the
-    resolved member set is a superset too -- it never omits a member that genuinely needs to run.
+    resolved member set is a superset too.
     """
 
     resolver: UniquenessDependentResolverInterface
--- a/backend/infrahub/core/regeneration/impact.py
+++ b/backend/infrahub/core/regeneration/impact.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, assert_never
 
 from infrahub.core import registry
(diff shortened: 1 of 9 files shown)


backend/infrahub/core/regeneration/impact.py
```diff
@@ -20,6 +21,8 @@
     from infrahub_sdk.client import InfrahubClient
     from infrahub_sdk.diff import NodeDiff
 
+    from infrahub.core.validators.uniqueness.dependent_resolver import UniquenessDependentResolverInterface
+
```
polmichel, line 26: I'll modify this weird dependency in the following PR https://github.com/opsmill/infrahub/pull/10246

PR #10246, opened 2026-08-13
refactor(core): generalize the dependent-node resolver out of uniqueness

polmichel:
Stacked on #10262.

## What

The reverse "which nodes reference these changed peers through a relationship" traversal is generic; it lived in `validators/uniqueness/` only because that was its first caller. This moves it to a neutral home so the selective-regeneration narrowing depends on a general component rather than the uniqueness validators.

- `DependentNodeResolver` (+ interface) → `core/relationship/` (its domain).
- `DependentNodesQuery` → `core/query/` (it is a query, alongside the existing relationship queries).
- Uniqueness callers repointed; behavior unchanged.

## The import cycle

Moving the resolver into `core/relationship/` surfaced a latent import cycle in that package. Its root: `relationship_schema` imported the `Relationship` **class** only to read `Relationship.rel_type` through a `get_class()` that nothing else calls. Dropping the dead `get_class()` and reading the edge type from `DatabaseEdgeType.IS_RELATED` removes `relationship_schema`'s dependency on the `relationship` package and breaks the cycle at its root.

## Tests

Uniqueness scope/determiner and the relocated resolver component tests stay green; the resolver imports in isolation without triggering the cycle.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/query/dependent_nodes.py
+++ b/backend/infrahub/core/query/dependent_nodes.py
@@ -10,12 +10,12 @@ if TYPE_CHECKING:
     from infrahub.database import InfrahubDatabase
 
 
-class AffectedUniquenessDependentsQuery(Query):
+class DependentNodesQuery(Query):
     """Return the nodes of a kind related, through a named relationship, to any of a set of peer nodes.
 
     The path is traversed in the relationship's own direction, so a kind whose relationship points at
-    its own kind resolves only the nodes on the constrained side of the peers, not those on the
-    opposite side.
+    its own kind resolves only the nodes on the owning side of the peers, not those on the opposite
+    side.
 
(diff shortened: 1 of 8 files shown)


ajtmccarty (commented):
looks good to me. only not approving b/c it is still a draft

ajtmccarty (approved):

PR #10247, opened 2026-08-13
refactor(frontend): tokenize primary and muted text colors

bilalabbad:
# Why

Neutral text colors were spread across **three ramps** — `gray` (297 uses), `neutral` (64), `stone` (52). Measured in oklch these are lightness-identical at every shade (max ΔL 0.011, imperceptible); what differs is temperature: `gray` is cool (hue ≈260), `neutral` is achromatic (C=0), `stone` is warm (hue ≈34–74). So the same visual weight rendered in three temperatures, frequently side by side.

Meanwhile `--foreground` already existed in `theme.css` as `stone-800` — the design system had already chosen warm — and had **zero users**.

Goal: one token pair owns primary and muted text, on the warm direction `--foreground` already picked.

Non-goals: no dark-mode values, no status/brand colors, no third or fourth text level.

## What changed

**Behavioral / visual**

- 166 text-color sites now read from tokens (60 `text-foreground`, 106 `text-foreground-muted`).
- For the **800 and 600** tones the change is **temperature, not weight**: `gray-800`→`stone-800` and `gray-600`→`stone-600` are within ΔL 0.01, so nothing gets lighter or darker — previously-cool sites are simply now warm.
- For the **900** tone it *is* a weight change: headings, labels and table headers move from 17.5–17.9:1 to **15.2:1** on white (ΔL ≈ 0.055). Perceptibly lighter, still far above AA. This is the one visual regression risk in the PR and the thing worth eyeballing.

**Implementation**

- `packages/ui/src/styles/theme.css` gains `--foreground-muted: var(--color-stone-600)`, exposed via `@theme inline` as `--color-foreground-muted` → `text-foreground-muted`.
- `text-{gray,neutral,stone}-800` → `text-foreground` — **32 sites, all of them**.
- `text-{gray,neutral,stone}-900` → `text-foreground` — **28 sites, all of them**. Same role as 800 and split across the same three ramps: the identical "page heading" role was cool in `task-item-details`, achromatic in `login`, warm in `sheet.stories`.
- `text-{gray,neutral,stone}-600` → `text-foreground-muted` — **100 sites**, selected by the rule below.
- `text-{gray,neutral,stone}-700` → `text-foreground-muted` — **6 sites only**, in the three files where 700 pairs with a 900 (`pagination.tsx`, `badge.tsx`, `link-toggle-button.tsx`), so `{900,700}` lands on `{foreground, muted}`.
- Variant prefixes preserved throughout: `hover:`, `data-hovered:`, `data-selected:`, `data-focused:`, and `[&_[cmdk-group-heading]]:`.

## The rule for the 600 tone

600 becomes `--foreground-muted` only where that text is **subordinate to more prominent content where it renders** — which is not the same as "the file contains a darker shade". A badge has no primary text of its own; it inherits it from whatever page renders it. `branch-default-badge.tsx` is a `text-xs` "default" pill annotating a branch name that lives in a different file entirely.

Migrated categories:

| category | examples |
|---|---|
| badges & pills | `"default"`, `"schema differs from default"` |
| descriptions | branch, schema, IP-namespace, object-template, `content.tsx` |
| labels & captions | `property-list` `<td>` labels, `"After"`/`"Before"`, cmdk group headings, breadcrumb items |
| helper text | form hints, token/password tab copy, `error-fallback` |
| event phrasings | `"logged out"`, `"merged the proposed change"` |
| icon-only controls | `LockIcon`, `ChevronsUpDownIcon`, 6 zoom/layout buttons in `graph-controls` |
| ghost button label | `"Add sort"` |
| empty states | `"No ip prefix"` |

## The 700 tone is mostly left raw (28 sites)

700 only becomes `muted` where a 900 in the same file makes it unambiguously the second level. It stays raw elsewhere for two reasons:

- **In seven files it coexists with 600 as a genuine distinct level** — `parallel-mode-plugin`, `search-nodes`, `path-results-list`, `generator-definition-run-button`, `event-details-popover`, `modal.stories`, `export-menu`. Mapping both onto `--foreground-muted` would flatten them.
- **`event-details-popover.tsx:12` would have broken outright**: it is `text-foreground-muted … hover:text-gray-700`, so mapping 700→muted turns the hover into a no-op.

Also left raw: standalone 700 that is a component's *own* primary, e.g. the `new.tsx` "Create a proposed change" h1. 700 sits at 10.3:1 — nearer `--foreground-muted` (7.6:1) than `--foreground` (15.2:1) — so it is really a third level, and this PR only defines two.

## Deliberately left on raw shades (12 sites at the 600 tone)

Nine components where 600 is that component's **own primary text**, with 400 as its secondary — an internal hierarchy an absolute two-level scale can't express:

`file-dropzone` ×2 · `object-hierarchy-tree` ×2 · `global-filter-tag` ×2 · `event-attributes` · `event-relationships` · `sidebar` · `ipam-layout` · `sortable-list.stories`

Plus `ipam-details-header.tsx` ×1, where `Value` (`font-medium text-gray-600`) holds the actual data next to a `text-custom-blue-800` `Title`. Calling the data *muted* would be backwards; if anything it wants `--foreground`, but that's a visual change rather than a rename.

This is the real finding behind the scoping: **only 45 of 201 files (22%) anchor their darkest text at 800/900. 139 files (69%) never go darker than 600.** The most common per-file primary is 600, not 800, and the step to the next level is 100–200 in ~79% of cases. Components anchor their scale to how prominent they are on the page. A fuller fix is a *relative* foreground per surface — the `--card-header-foreground` precedent generalised — which is a design-system change, not a rename.

## How to review

Two commits: the first does the 800/600 tones, the second the 900 tone plus the three paired-700 files. Reviewing them separately is easier, since only the second changes text weight.

Start with `packages/ui/src/styles/theme.css` — 2 lines, the only non-mechanical change. The rest are one-line class swaps.

Note **15 files contain reformatting noise**: `text-foreground-muted` is longer than `text-gray-600`, so Biome re-wrapped some JSX and re-sorted classes. That is why the diff is 204/176 rather than a clean 166/166. `git diff --word-diff` cuts through it.

## How to test

```bash
cd frontend/app
pnpm exec biome ci .      # 1532 files, clean
pnpm knip                 # clean
pnpm exec betterer ci     # 188 issues, unchanged
pnpm test                 # 1190 tests / 176 files
pnpm build
```

Manually: compare a branch list, an object detail page, and the events feed against `develop`. Text at the 800/600 tones should sit at the same weight, slightly warmer. Headings and labels previously at 900 will be slightly lighter — that is the one thing to judge.

## Impact & rollout

- **Backward compatibility:** CSS-only, nothing affected.

## Known gaps, out of scope here

- **`packages/schema-visualizer/src`** has 7 files at these shades, but it is a git submodule (`opsmill/infrahub-schema-visualizer`) and needs its own upstream PR. `index.css` does `@source` it, so the tokens would resolve there — it will keep rendering cool-grey text until migrated separately.
- **`markdown.css`** hardcodes six hex text colors, including `#1f2328`/`#656d76` which are functionally a foreground/muted pair. It deliberately mimics GitHub's rendering, so folding it in is its own decision.
- **`text-gray-80`** is a typo at `repository-objects-tab.tsx:34` and `object-tabs.tsx:39`/`:62`. Not a valid Tailwind shade, so it emits no CSS and those badge counts inherit their parent color. Left alone since it is not the 800 tone.
- Shades **300/400/500** (~240 sites), most of **700** (28 sites, see above), status colors (111) and brand text (72) are untouched.

## Checklist

- [x] Tests added/updated — no new tests; existing 1190 pass and no test asserts on the migrated classes
- [x] Changelog entry — intentionally skipped (`ci/skip-changelog`), no user-facing behaviour change


saltas888 (approved):

frontend/app/src/entities/branches/ui/branch-list-item/branch-default-badge.tsx
```diff
@@ -5,7 +5,7 @@ export function BranchDefaultBadge({ className, ...props }: BadgeProps) {
   return (
     <span
       className={classNames(
-        "rounded-full border border-border-strong bg-transparent px-1.5 py-0.5 text-stone-600 text-xs",
+        "rounded-full border border-border-strong bg-transparent px-1.5 py-0.5 text-foreground-muted text-xs",
```
saltas888, line 8: I think its important to add in the linter the no-arbitrary-value from tailwindcss so we prevent any future deviations from the defined tokens

PR #10256, opened 2026-08-14
chore(backend): enforce import-at-top (PLC0415) for backend/**

polmichel:
## Why

`PLC0415` ("import should be at the top of the file") was suppressed via ruff per-file-ignores across the backend, so accidental function-local imports went unflagged. This removes the blanket suppression for `backend/**` and enforces the rule, keeping only genuinely-justified lazy imports as explicit exceptions.

Goal: import-at-top enforced across `backend/**` and `python_testcontainers/**`; every remaining function-local import is a deliberate, documented exception.

**This is the first step toward removing all of these exceptions.** Dropping the blanket suppression is the enabler: each function-local import that must remain is now surfaced as an individual `# noqa: PLC0415` with a stated reason, instead of being hidden behind a file-wide ignore. That makes the exceptions visible and reviewable, so they can be revisited and eliminated one by one in follow-up PRs (see Follow-ups). This PR does not attempt to remove every exception at once — it removes the suppression and pays down only the accidental cases.

Non-goals: the `tasks/**` and `tests/e2e/**` suppressions are intentionally left in place (out of scope for this change).

## What changed

- Removed the `PLC0415` per-file-ignore for `backend/infrahub/**` and the `component` / `functional` / `integration` test trees.
- Hoisted the accidental function-local imports to the module top, including the whole `infrahub`/SDK block of the CLI `shell` command (those modules already load at the CLI top level, so nothing was actually deferred).
- Kept the genuinely-lazy imports behind a targeted `# noqa: PLC0415`, each with a reason.
- Moved the "imports at the top" instruction from `.agents/rules/testing-python.md` into the general `.agents/rules/python-module-layout.md` so it applies to all backend Python, not just tests.

What stayed the same: no schema, API, or runtime behavior changes. Every moved import was verified in a fresh interpreter to confirm it introduces no circular import, and the two cycles kept behind `noqa` were confirmed by reproducing the `ImportError`.

## Follow-ups

The remaining `# noqa: PLC0415` markers are candidates for removal in later PRs, in decreasing order of confidence:

- **Removable but needs owner sign-off** — the Prefect-testsuite isolation in `prefect_server/app.py` and the deferred app load in the two API test fixtures: no cycle, but each carries a deliberate runtime/test-design rationale to confirm before changing.
- **Keep** — the optional `IPython` dependency and the two verified circular imports (`workers/utils.py`, `core/migrations/shared.py`).

## How to test

```bash
uv run invoke main.lint
uv run invoke backend.lint          # ruff + ty + mypy
```

All green locally (1427 unit tests passed); `ruff check backend --select PLC0415` reports zero violations.

### DIFF
--- a/backend/infrahub/cli/__init__.py
+++ b/backend/infrahub/cli/__init__.py
@@ -45,20 +45,22 @@ async def _init_shell(config_file: str) -> None:
 @app.command()
 def shell() -> None:
     """Start a python shell within Infrahub context (requires IPython)."""
-    from infrahub_sdk import InfrahubClient
-    from IPython import start_ipython
-    from traitlets.config import Config
+    # Imported lazily: IPython/traitlets are optional dependencies, and this keeps the
+    # heavy Infrahub imports off every other CLI invocation.
+    from infrahub_sdk import InfrahubClient  # noqa: PLC0415
+    from IPython import start_ipython  # noqa: PLC0415
+    from traitlets.config import Config  # noqa: PLC0415
 
-    from infrahub import config
-    from infrahub.components import ComponentType
-    from infrahub.core.branch import Branch
-    from infrahub.core.initialization import initialization
-    from infrahub.core.manager import NodeManager
-    from infrahub.core.registry import registry
-    from infrahub.dependencies.registry import build_component_registry
-    from infrahub.lock import initialize_lock
-    from infrahub.services import InfrahubServices
-    from infrahub.workers.dependencies import (
+    from infrahub import config  # noqa: PLC0415
+    from infrahub.components import ComponentType  # noqa: PLC0415
+    from infrahub.core.branch import Branch  # noqa: PLC0415
+    from infrahub.core.initialization import initialization  # noqa: PLC0415
+    from infrahub.core.manager import NodeManager  # noqa: PLC0415
+    from infrahub.core.registry import registry  # noqa: PLC0415
+    from infrahub.dependencies.registry import build_component_registry  # noqa: PLC0415
+    from infrahub.lock import initialize_lock  # noqa: PLC0415
+    from infrahub.services import InfrahubServices  # noqa: PLC0415
+    from infrahub.workers.dependencies import (  # noqa: PLC0415
         get_cache,
         get_component,
         get_database,
--- a/backend/infrahub/cli/__init__.py
+++ b/backend/infrahub/cli/__init__.py
@@ -1,10 +1,25 @@
 import typer
+from infrahub_sdk import InfrahubClient
 from infrahub_sdk.async_typer import AsyncTyper
 
 from infrahub import config
+from infrahub.components import ComponentType
+from infrahub.core.branch import Branch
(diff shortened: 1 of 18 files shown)


backend/infrahub/cli/__init__.py
```diff
@@ -1,10 +1,25 @@
...
+from infrahub.services import InfrahubServices
+
+from ..workers.dependencies import (
+    get_cache,
+    get_component,
+    get_database,
+    get_http,
+    get_workflow,
+    set_component_type,
+)
```
ogenstad, line 22: As these are only needed for the dev shell command, I wonder if these imports should remain in that location. Mainly since the imports could potentially slow down other cli commands that don't need them.

backend/infrahub/core/query/__init__.py
```diff
@@ -13,6 +13,9 @@
 from neo4j.graph import Path as Neo4jPath
 from neo4j.graph import Relationship as Neo4jRelationship
 from opentelemetry import trace
+from rich import print as rprint
+from rich.console import Console
+from rich.table import Table
```
ogenstad, line 18: A sanity check here could be if we import rich regardless or if this results in more imports than we had before in the production code for something that's only used for debugging.

ogenstad (approved):

backend/infrahub/cli/__init__.py
polmichel: These dependencies were already indirectly imported through these lines:
- `from infrahub.core.initialization import initialization`
    - `infrahub.core.branch` (Branch)
    - `infrahub.core.manager` (NodeManager)
    - `infrahub.core.registry` (registry)
    - `infrahub.dependencies.registry` (build_component_registry) — via `infrahub.core.node.create`

- `from ..workers.dependencies import get_database`
    - infrahub.components (ComponentType)
    - infrahub.services (InfrahubServices) — via infrahub.services.adapters.cache

So, this will not impact the performance of those commands right? Or maybe is that we would want to break this dependency?

backend/infrahub/core/query/__init__.py
polmichel: Good point! I checked that and the only place which is used in "production" code is when miscellaneous settings are configured in a certain way

see `infrahub.core.query.Query.execute`

```
 @trace.get_tracer(__name__).start_as_current_span("Query.execute")
    async def execute(self, db: InfrahubDatabase, timeout_seconds: float | None = None) -> Self:
        # Ensure all mandatory params have been provided
        # Ensure at least 1 return obj has been defined

        if config.SETTINGS.miscellaneous.print_query_details:
            self.print(include_var=True)
```

backend/infrahub/cli/__init__.py
ogenstad: If we import them regardless I think that's fine. There might come a point where we revert this to have the various `infrahub` cli commands load faster and only load what they need on demand, but looks good to me for now.

ogenstad (approved):

PR #10262, opened 2026-08-14
feat(regeneration): narrow relationship-reached changes to owning members (IFC-2946)

polmichel:
## What

Post-merge (and proposed-change) selective regeneration now narrows to the affected members when the changed data is read **through a relationship** — an owner's name, an interface's field, multi-hop chains, a kind reached by several relationships, a change behind a generic peer, and a change reached through a relationship defined on a generic — instead of regenerating every member of the definition.

## How

- **Analyzer** (`ReachedPathResolver`): reconstructs, per related kind, every relationship chain the query traverses down to it; a change reached through a generic peer is resolved to the generic's concrete implementations, and a hop whose owner is a generic is keyed on the generic kind.
- **Classifier**: a new `RelationshipReachedChanges` assessment carrying those chains; it widens the whole assessment when a change can't be pinned (a fragment refinement along the chain, or a kind also read at a root).
- **Impact** (`ReachedMemberResolver`): walks each chain back to the owning members via the existing dependent-node resolver and unions them with the directly-changed members.

## Invariant

Over-execution is acceptable, under-execution is not: every hop returns a superset and any ambiguity widens to the whole group, so no member that genuinely needs regenerating is skipped.

## How to review

Suggested order, with the invariant above as the lens:

1. **`graphql/analyzer.py` — `ReachedPathResolver`** (the core): `_hop_chain_to_root` rebuilds a chain of owners from a related node up to the root, keying a generic owner on the generic kind; `_reached_kinds` keys a generic peer by its concrete implementations. Focus on the **widen conditions** (fragment refinement, kind also read at a root) — this is where under-execution would hide.
2. **`core/regeneration/impact_classifier.py` — `assess`**: emits `RelationshipReachedChanges` when every reached kind resolves, and widens the whole assessment as soon as one cannot be pinned.
3. **`core/regeneration/impact.py`**: `ReachedMemberResolver` walks each chain back to members (via the existing dependent-node resolver) and unions them; `get_field_level_impacted_subscribers` wires it and keeps the widen fallback.
4. **Tests**: `tests/unit/graphql/test_reached_path_resolver.py` pins every narrow/widen case (including generics × multiple relationships, and the generic-owner narrowing); `tests/component/graphql/test_relationship_reached_paths.py` re-checks them through the real analyzer; `tests/component/core/regeneration/test_impact.py` proves end-to-end narrowing on a database, including a change reached through a generic-owned relationship.

The question to keep asking: **can any path resolve a related change to a *subset* of its owners?** Every uncertainty must widen.

## Tests

- Unit: the two resolvers in isolation (chains and widen cases including generic peer and generic owner; hop-chaining and multi-chain union, recording double, no mocks) and the classifier variants.
- Integration: real queries through the analyzer assert the reconstructed chains.
- Component: a relationship-reached change narrows to the owning member end-to-end on a real database, including a change reached through a relationship defined on a generic.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/regeneration/impact.py
```diff
@@ -96,13 +99,47 @@ async def get_field_level_impacted_subscribers(
...
+
+    async def resolve(self, changes: RelationshipReachedChanges) -> set[str]:
+        members: set[str] = set(changes.direct_member_node_ids)
+        for change in changes.reached:
+            for path in change.paths:
+                peer_uuids = set(change.node_ids)
+                for hop in path.hops:
+                    if not peer_uuids:
+                        break
+                    peer_uuids = await self.resolver.resolve(
```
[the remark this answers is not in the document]
polmichel: Investigated — this is a false positive, no under-execution.

When a node read through a relationship is deleted, the reader's own edge to it is removed too, and the diff records that removal **on the reader node itself** (e.g. a `TestingCar` surfaces `previous_owner: REMOVED`). The reader is therefore selected through its **direct member change** (the root path in `assess`), independently of the reverse traversal returning nothing for the now-deleted peer. The member is never skipped.

This generalises to multi-hop: a deletion always surfaces a relationship change on some still-existing node along the chain, which resolves back to the owning member.

changelog/+ifc-2946-relationship-reached-narrowing.changed.md
```diff
@@ -0,0 +1 @@
+Post-merge regeneration now narrows to the affected members when the changed data is read through a relationship (for example an owner's name rendered by an artifact, or an interface's field on a device), instead of regenerating every member of the definition. Changes that cannot be attributed to specific members unambiguously still regenerate the whole group.
```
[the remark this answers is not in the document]
polmichel: Done — renamed to `+ifc-2946-relationship-reached-narrowing.changed.md`, matching the `<issue-id>-<slug>.<type>.md` convention (fe6c5038a).

backend/infrahub/core/regeneration/impact_classifier.py
gmazoyer: This docstring is unchanged and now says the opposite of `assess()`. It still claims a relationship-reached change "cannot be mapped back to a subscriber" and widens. It is the only place the widen rule is written down, so the next person will code against it.

backend/infrahub/graphql/analyzer.py
```diff
@@ -140,6 +142,26 @@ def fields(self) -> set[str]:
...
+    relationship_direction: RelationshipDirection
+
+
+@dataclass(frozen=True, slots=True)
+class ReachedPath:
+    """The relationship chain a query follows from a root object down to a related kind."""
+
+    hops: tuple[RelationshipHop, ...]
+
+
```
gmazoyer: `RelationshipHop` and `ReachedPath` have nothing GraphQL about them and `core` imports them from there. Maybe these should be placed somewhere else.

backend/infrahub/core/regeneration/impact_classifier.py
```diff
@@ -54,24 +80,38 @@ class QueryImpactClassifier:
...
-            return EveryTarget()
-
-        return ChangedNodes(node_ids=changed_node_ids)
-
-    def _must_widen(self, *, diff_summary: list[NodeDiff], changed_node_ids: list[str]) -> bool:
         if not self.only_has_unique_targets:
             # Any number of objects can answer the query, so a changed node cannot be traced back to
             # the targets reading it.
-            return bool(changed_node_ids)
+            any_relevant_change = self._changed_node_ids(diff_summary=diff_summary, kinds=self.readable_fields_by_kind)
```
gmazoyer: `any_relevant_change` holds a list, not a bool. I'd rename it to `relevant_change_ids` or drop the local and test the call directly.

backend/infrahub/graphql/analyzer.py
```diff
@@ -272,6 +294,100 @@ def get_models(self) -> list[GraphQLQueryModel]:
...
+        return [model.kind]
+
+    def _iter_object_nodes(self, node: GraphQLQueryNode) -> Iterator[GraphQLQueryNode]:
+        """Yield every node in the subtree that stands for an Infrahub object."""
+        if node.infrahub_model is not None:
+            yield node
+        for child in node.children:
+            yield from self._iter_object_nodes(child)
+
+    def _clean_hop_chain(self, node: GraphQLQueryNode) -> ReachedPath | None:
```
gmazoyer: `_clean_hop_chain` builds a chain, it does not clean one. `_hop_chain_to_root` or something like that would say it better.

ajtmccarty:
the Known Limitation in the PR description had me worried for a second, but it is actually incorrect

> adding a generic to a kind's inheritance does not relabel existing instances (inherit_from is validate-only, no migration), so nodes created before the change lack the label and would be missed

this is correct that `inherit_from` is set to `VALIDATE_CONSTRAINT` when it should be `MIGRATION_REQUIRED`, but this attribute has special handling in `SchemaUpdateValidationResult.process_diff()` that ensures it is run as a migration. And that migration relabels all active objects on the branch where it runs with the new labels appropriate for the updated `inherit_from` value.

not sure how this might affect this PR

backend/infrahub/core/regeneration/impact.py
ajtmccarty, line 101: a dataclass should not be calling a component. dataclasses just hold data. they should only be inputs and outputs of functions/methods.

backend/infrahub/core/regeneration/impact.py
```diff
@@ -96,13 +99,47 @@ async def get_field_level_impacted_subscribers(
         case EveryTarget():
             return TargetSelection(ids=every_target, widened=True)
         case ChangedNodes(node_ids=node_ids):
-            subscribers = await _get_subscribers_for_nodes(node_ids=node_ids, branch=query_branch, client=client)
-            ids = [subscriber.subscriber_id for subscriber in subscribers if subscriber.kind == subscriber_kind]
-            log.debug(f"SELECTIVE_REGEN field-impact resolved subscribers: {len(ids)}")
-            return TargetSelection(ids=ids, widened=False)
+            member_ids = node_ids
+        case RelationshipReachedChanges():
+            dependent_resolver = UniquenessDependentResolver(db=db, branch=query_branch_obj)
```
ajtmccarty, line 72: looks like this method should be a class that has the `UniquenessDependentResolver` injected as a dependency at construction time

backend/infrahub/core/regeneration/impact_classifier.py
polmichel: Agreed. And changed!

backend/infrahub/graphql/analyzer.py
polmichel: Since `core` import them I do agree. Good point.

backend/infrahub/core/regeneration/impact.py
polmichel: Good point I changed that

backend/infrahub/core/regeneration/impact_classifier.py
polmichel: Resolved in c20b4473f — docs(regeneration): correct assess() docstring to describe path-based narrowing

backend/infrahub/graphql/analyzer.py
polmichel: Resolved in 23b988ef4 — refactor(regeneration): move RelationshipHop/ReachedPath into core models

backend/infrahub/core/regeneration/impact_classifier.py
polmichel: Resolved in 67eb34db3 — refactor(regeneration): name the empty-diff guard for the bool it tests

backend/infrahub/graphql/analyzer.py
polmichel: Resolved in a49ccb276 — refactor(analyzer): rename _clean_hop_chain to _hop_chain_to_root

backend/infrahub/core/regeneration/impact.py
polmichel: Resolved in 325117676 — refactor(regeneration): make ReachedMemberResolver a class, not a dataclass

backend/infrahub/core/regeneration/impact.py
polmichel: Resolved in 325117676 — refactor(regeneration): make ReachedMemberResolver a class, not a dataclass

polmichel:
> the Known Limitation in the PR description had me worried for a second, but it is actually incorrect
> 
> > adding a generic to a kind's inheritance does not relabel existing instances (inherit_from is validate-only, no migration), so nodes created before the change lack the label and would be missed
> 
> this is correct that `inherit_from` is set to `VALIDATE_CONSTRAINT` when it should be `MIGRATION_REQUIRED`, but this attribute has special handling in `SchemaUpdateValidationResult.process_diff()` that ensures it is run as a migration. And that migration relabels all active objects on the branch where it runs with the new labels appropriate for the updated `inherit_from` value.
> 
> not sure how this might affect this PR

Thanks for the feedback, you're right I missed this part of the code. 

`_hop_chain_to_root` now builds a hop for a generic owner, keyed on the generic kind, instead of returning None, and the "Known limitation" section is removed from the description.

Tested by:
- Unit: `test_a_kind_reached_through_a_relationship_on_a_generic_owner_narrows` (test_reached_path_resolver.py) asserts the reconstructed chain instead of the former widen.
- Component: `TestGenericOwnerFieldLevelImpact` (test_impact.py) creates a card reached through a relationship defined on a generic slot and asserts the change narrows to the owning rack alone, proving the label-keyed reverse traversal reaches the concrete instance.

backend/infrahub/core/regeneration/impact.py
```diff
@@ -76,13 +78,47 @@ async def get_field_level_impacted_subscribers(
...
+            member_ids = sorted(await ReachedMemberResolver(resolver=dependent_resolver).resolve(assessment))
         case _ as unreachable:
             assert_never(unreachable)
 
+    subscribers = await _get_subscribers_for_nodes(node_ids=member_ids, branch=query_branch, client=client)
+    ids = [subscriber.subscriber_id for subscriber in subscribers if subscriber.kind == subscriber_kind]
+    log.debug(f"SELECTIVE_REGEN field-impact resolved subscribers: {len(ids)}")
```
ajtmccarty: leftover, I think

gmazoyer (approved):

backend/infrahub/core/regeneration/impact.py
```diff
@@ -29,24 +32,22 @@ async def get_field_level_impacted_subscribers(
...
+
+    A change matters only when a modified field is one the query reads. The query analysis,
+    the diff-summary tag and the subscriber lookup all run on `query_branch`, so the caller
+    passes the branch the changed data lives on (a proposed change's source branch, a merge's
+    target branch).
+
+    `every_target` is the fallback when a change cannot be traced to specific subscribers;
+    taking it as an argument keeps "process everything" out of the return type, so the caller
+    always gets one authoritative list.
     """
```
ajtmccarty, line 42: probably an issue for a different day, but I think the `get_field_level_impacted_subscribers` function can be turned into its own component and use dependency injection within its call sites in `ImpactedSubscriberResolver`, `validate_artifacts_generation`, and `request_generator_definition_check`

backend/infrahub/core/regeneration/impact.py
ajtmccarty: also leftover perhaps

backend/infrahub/graphql/analyzer.py
```diff
@@ -272,6 +275,97 @@ def get_models(self) -> list[GraphQLQueryModel]:
         return models
 
 
+def _reached_path_sort_key(path: ReachedPath) -> tuple[tuple[str, str, str], ...]:
+    return tuple((hop.node_kind, hop.relationship_identifier, hop.relationship_direction.value) for hop in path.hops)
+
+
+@dataclass(frozen=True)
+class ReachedPathResolver:
```
ajtmccarty, line 283: I don't think there's a good way to do this within the current structure, but this class looks like it should be a dependency of whatever builds a `GraphQLQueryReport` instead of being used within a property of `GraphQLQueryReport`. I don't think this is really possible right now (w/o a lot of refactoring) b/c `InfrahubGraphQLQueryAnalyzer` builds `GraphQLQueryReport` and `InfrahubGraphQLQueryAnalyzer` does not really use dependency injection.

backend/infrahub/graphql/analyzer.py
```diff
@@ -321,6 +415,11 @@ def traversed_kinds(self) -> set[str]:
 
         return kinds
 
+    @cached_property
+    def relationship_reached_paths(self) -> dict[str, tuple[ReachedPath, ...]]:
```
ajtmccarty: maybe `relationship_reached_paths_by_kind` to indicate this is a map/dict?

backend/infrahub/core/regeneration/impact_classifier.py
```diff
@@ -28,50 +30,88 @@ class ChangedNodes:
...
+    A kind read both at a root and through a relationship widens: the two read paths are
+    indistinguishable once a change is in hand, so narrowing would drop the members reached only
+    by the relationship.
     """
 
     query_branch: str
     only_has_unique_targets: bool
     traversed_kinds: set[str]
     readable_fields_by_kind: dict[str, set[str]]
+    reached_paths: dict[str, tuple[ReachedPath, ...]] = field(default_factory=dict)
```
ajtmccarty: `reach_paths_by_kind` goes with the name of the above variable

backend/infrahub/core/regeneration/impact.py
```diff
@@ -76,13 +78,47 @@ async def get_field_level_impacted_subscribers(
...
+
+    async def resolve(self, changes: RelationshipReachedChanges) -> set[str]:
+        members: set[str] = set(changes.direct_member_node_ids)
+        for change in changes.reached:
+            for path in change.paths:
+                peer_uuids = set(change.node_ids)
+                for hop in path.hops:
+                    if not peer_uuids:
+                        break
+                    peer_uuids = await self.resolver.resolve(
```
ajtmccarty, line 101: maybe worth a followup. this is a database query at the bottom of three loops, so this could be a performance issue. maybe the query can be updated to support multiple relationships and these can be grouped before being sent to the database?

ajtmccarty (approved):
excellent work on the unit tests. I think the code is a bit hard to understand, but I'm not sure if there is anything to be done about that b/c it is handling some pretty complex logic. the tests definitely make it easier to understand. great work

backend/infrahub/core/regeneration/impact.py
polmichel: This was actually intentional, since I thought it would be a good idea to have those logs during the first weeks after this feature ships, in case we want to analyze the behavior. But you're right, I can reapply them later if troubleshooting is needed. I'll remove them.

backend/infrahub/core/regeneration/impact.py
polmichel: Yes it could and it would be definitely better. Noted and saved inside another task of the internal Epic https://opsmill.atlassian.net/browse/IFC-2950

backend/infrahub/graphql/analyzer.py
polmichel: Yes you are right, I'll also capture this thought inside an item in case I'll be able to spend some time on it

backend/infrahub/graphql/analyzer.py
polmichel: Good idea I'll do it

backend/infrahub/core/regeneration/impact.py
polmichel: I'll look into it during this sprint, captured inside the internal item https://opsmill.atlassian.net/browse/IFC-3069. Thanks

PR #10284, opened 2026-08-17
Adapt remaining app to dark theme

bilalabbad:
# Dark theme

Infrahub gets a dark theme. Earlier PRs in this series gave the design system its dark palette; this one carries it through the rest of the application and adds the machinery a person needs to actually choose it.

It ships behind an experimental flag, off by default, while the theme is alpha. A later PR will come which allows users to save this in their user preferences

## Screenshots

Captured on a live stack at 1440×900 — same pages, same data, only the theme differing. These cover this series as a whole, including the follow-up work merged in from #10295.

**Home**

| light | dark |
|---|---|
| home light | home dark |

**Object list**

| light | dark |
|---|---|
| devices light | devices dark |

**Proposed change with a Mermaid diagram** — the diagram itself follows the theme:

| light | dark |
|---|---|
| proposed change with mermaid, light | proposed change with mermaid, dark |

**GraphQL sandbox** — follows the app theme instead of being pinned light:

| light | dark |
|---|---|
| graphql sandbox light | graphql sandbox dark |

**Login** — themed before a session exists:

| light | dark |
|---|---|
| login light | login dark |

## Checklist

- [x] Tests added/updated — theme provider, gating rule, resolved-theme hook, account-menu switch, and visual tests asserting the colours Mermaid actually bakes in
- [x] Changelog entry — `+dark-theme.added.md`
- [x] External docs updated — FAQ entry; the configuration reference carries the setting through generation
- [x] Internal .md docs updated — `dev/knowledge/frontend/theming.md`

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/config.py
```diff
@@ -990,6 +990,13 @@ class AnalyticsSettings(BaseSettings):
 class ExperimentalFeaturesSettings(BaseSettings):
     model_config = SettingsConfigDict(env_prefix="INFRAHUB_EXPERIMENTAL_")
     graphql_enums: bool = False
+    # Four generated artifacts record this default: the root compose file, the OpenAPI schema, the
+    # configuration reference, and the frontend REST types. The frontend one is regenerated from the
+    # OpenAPI schema rather than from this file, so it is the easy one to miss when this changes.
```
ajtmccarty: this looks unnecessary

backend/infrahub/config.py
```diff
@@ -990,6 +990,13 @@ class AnalyticsSettings(BaseSettings):
...
+    dark_theme: bool = Field(
+        default=False,
+        description="Offer the dark theme in the web interface. Alpha: some surfaces still render incorrectly.",
+    )
```
gmazoyer: Should this actually be a user preference instead of a global switch?

ogenstad: I think this was just something that @saltas888 requested to make it easy to test this during development since the development of dark mode is different from the configuration of the feature. But it should be something that we can control with the user defined settings introduced in 1.11.

saltas888: I should have mentioned that in the PR description. This is so we can release it internally, in an alpha version. I would prefer to send a different PR with the BE changes to save it

saltas888:
UI tests have been checked on https://github.com/opsmill/infrahub/pull/10295 earlier. Can be ignored for now

ogenstad (approved):

PR #10306, opened 2026-08-18
fix(graphql): reject user-supplied branched_from on branch creation

ajtmccarty:
`BranchCreate` persisted a client-supplied branched_from verbatim, letting a new branch appear to have been created in the past. The mutation now rejects the field with an error, and both branched_from and origin_branch (already restricted to the default branch) are marked deprecated on `BranchCreateInput`.

We might somewhat support creating a historical branch, but I am not completely certain and I bet that trying to merge or rebase one would cause problems, so I think it is safest to make sure this is completely blocked until we have a reason to support it.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10306?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/component/graphql/mutations/test_branch.py
```diff
@@ -24,6 +26,158 @@
...
+        name="branched_from_empty_string_rejected",
+        branch_name="empty-branched-from",
+        variables={"branchedFrom": ""},
+        expected_message=BRANCHED_FROM_ERROR,
+    ),
+    RejectedInputTestCase(
+        name="origin_branch_other_than_default_rejected",
+        branch_name="other-origin-branch",
+        variables={"originBranch": "not-the-default-branch"},
+        expected_message="origin_branch must be 'main'",
```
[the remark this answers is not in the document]
ajtmccarty: "main" is the default branch used in almost all tests

backend/infrahub/graphql/mutations/branch.py
```diff
@@ -76,6 +80,10 @@ async def mutate(
     ) -> Self:
         if data.origin_branch and data.origin_branch != registry.default_branch:
             raise ValueError(f"origin_branch must be '{registry.default_branch}'")
+        if data.branched_from:
```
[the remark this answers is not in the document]
ajtmccarty: using `data.get("branched_from") is not None` instead so that someone sending in a `null` branched_from can still proceed

gmazoyer (approved):

backend/infrahub/graphql/mutations/branch.py
```diff
@@ -46,8 +46,12 @@ class BranchCreateInput(InputObjectType):
...
     name = String(required=True)
     description = String(required=False)
-    origin_branch = String(required=False)
-    branched_from = String(required=False)
+    origin_branch = InputField(
+        String(required=False), deprecation_reason="Branches can only be created from the default branch"
+    )
+    branched_from = InputField(
+        String(required=False), deprecation_reason="branched_from is set by the server and cannot be provided"
+    )
```
gmazoyer, line 56: I guess that technically this would need to be in a minor release and not a patch one. But I'm pretty sure nobody use this anyway.

ogenstad: I think it would be good to include an Infrahub version for when this will be removed. Though I think we might have this situation elsewhere as well and we could have a follow up item to search for these strings without a set version for removal and then add those along with follow up items in the correct housekeeping epics for future versions.

ajtmccarty: updated description to include "Will be removed after versions 1.12."
added this one to find and mark all deprecated fields as part of 1.12 https://opsmill.atlassian.net/browse/IFC-3029
added this one to remove these fields in 1.13

Issue #10310, opened 2026-08-18
tech-debt: branch rebase memory use scales with the branch diff size and exceeds Neo4j's default transaction limit


lancamat1:
### Component

API Server / GraphQL

### Describe the Feature Request

Rebasing a branch requires an amount of Neo4j transaction memory proportional to that branch's diff, so whether a rebase succeeds depends on how much was changed on the branch rather than on how the instance is sized. On a default Neo4j configuration this fails outright:

```
neo4j.exceptions.TransientError: {neo4j_code: Neo.TransientError.General.MemoryPoolOutOfMemoryError}
{message: The allocation of an extra 6.3 GiB would use more than the limit 4.2 GiB.
Currently using 97.8 MiB. dbms.memory.transaction.total.max threshold reached}
```

The save is issued from `backend/infrahub/core/diff/repository/repository.py` (`_save_node_batch`, reached from `save`). In the observed case the branch's diff was roughly 100,000 added and 13,000 updated nodes, and the 4.2 GiB ceiling was Neo4j's stock `dbms.memory.transaction.total.max` (70% of default heap) on a 24 GB host, with no memory settings overridden.

The need: rebasing a branch completes on a default Neo4j configuration, without the memory requirement growing with the amount of data on the branch.

### Describe the Use Case

Version affected: 1.10.6 (Enterprise); the code path is unchanged in 1.10.8 and 1.11.0b2. Encountered during `infrahub upgrade --rebase-branches`, whose Step 6/6 rebases every open branch, on an instance with 8.25M nodes, 10.4M relationships and 12 branches. Branches with large diffs failed while smaller ones rebased fine. Disabling `dbms.memory.transaction.total.max` and raising the heap let every rebase complete, so the requirement is a property of the operation rather than of the host.

The practical consequence is that upgrading an instance needs bespoke Neo4j tuning determined by its largest branch, and the threshold moves as customer data grows.

### Additional Information

Recurring pattern rather than a one-off: #4511 (closed) was the same `TransientError` for a 5.2k-node diff during diff generation, and #9889 (closed) was the same limit during `BranchDelete`. `docs/docs/deploy-manage/maintain-upgrade/upgrade/overview.mdx` states "We are working on a more robust solution to handle large migrations without hitting this limit in future releases"; this issue is the evidence for the rebase path.

Issue #10309, opened 2026-08-18
bug: an error during one branch rebase aborts the remaining branches with a raw traceback instead of being reported per branch


lancamat1:
### Component

API Server / GraphQL

### Infrahub version

1.10.6 (Enterprise). Present in 1.10.8 at `backend/infrahub/cli/db.py` in `trigger_rebase_branches`; observed while upgrading a database restored from 1.10.6.

### Current Behavior

The per-branch rebase loop only treats two exception types as a branch-level failure:

```python
    for branch in branches:
        ...
        try:
            await registry.schema.load_schema(db=db, branch=branch)
            await rebase_branch(...)
        except (ValidationError, MigrationFailureError):
            console.log(f"Rebased branch '{branch.name}' (ID: {branch.uuid}) {FAILED_BADGE}")
        else:
            console.log(f"Rebased branch '{branch.name}' (ID: {branch.uuid}) {SUCCESS_BADGE}")
```

Any other exception — for example a database error such as `neo4j.exceptions.TransientError` — propagates out of the loop and terminates the whole command, so the branches after it are never attempted.

Observed in one run of `infrahub upgrade --rebase-branches` over 11 branches: two branches failed with `ValidationError` and were reported as `FAILED`, the third raised `TransientError` from the diff save, and the command exited with an unhandled traceback. The remaining 8 branches were never attempted and no closing line was printed.

### Expected Behavior

A failure while rebasing one branch is reported for that branch and the run continues with the others, so one problem branch does not cost the remaining branches a maintenance window.

### Steps to Reproduce

1. Have several branches needing rebase, at least one of which triggers a database-level error during rebase (for example a branch large enough to exceed Neo4j's transaction memory limit).
2. Run `infrahub upgrade --rebase-branches`.
3. Branches failing with `ValidationError` are reported as `FAILED`; the first branch raising anything else aborts the command.

### Additional Information

Traceback tail from the observed run: `core/diff/repository/repository.py _save_node_batch` -> `core/query/__init__.py execute` -> `neo4j.exceptions.TransientError ... MemoryPoolOutOfMemoryError`.

Issue #10308, opened 2026-08-18
bug: infrahub upgrade prints "Upgrade complete SUCCESS" and exits 0 when branch rebases failed


lancamat1:
### Component

API Server / GraphQL

### Infrahub version

1.10.6 (Enterprise). The unconditional success log is present in 1.10.8 (`backend/infrahub/cli/upgrade.py:220`); observed while upgrading a database restored from 1.10.6.

### Current Behavior

`infrahub upgrade --rebase-branches` on a database with 11 branches needing rebase reported 4 branch failures and still finished with:

```
Upgrade complete SUCCESS
```

and exit code 0. Step 6/6 had printed one line per failed branch, with no reason:

```
Rebased branch 'feature/<name>' (ID: ...) FAILED
```

The final line is logged unconditionally after the rebase step, regardless of what the per-branch results were:

```python
    console.log(f"[bold]Upgrade complete[/bold] {SUCCESS_BADGE}")
```

Nothing in the output or the exit status distinguishes this run from one where every branch rebased. The reasons were only retrievable from the underlying task state (constraint violations on the branch, and branches carrying conflicts with the default branch).

### Expected Behavior

An upgrade that did not complete everything it attempted is distinguishable from one that did:

- a non-zero exit code, so scripts, CI jobs and Helm hooks can detect it
- a final status that reflects the partial failure rather than `SUCCESS`
- the per-branch reason surfaced in the command's own output, or an explicit pointer to where it can be read

### Steps to Reproduce

1. Have open branches carrying conflicts or constraint violations against `main`.
2. Run `infrahub upgrade --rebase-branches`.
3. Observe `FAILED` lines without reasons, a final `Upgrade complete SUCCESS`, and `echo $?` returning 0.

### Additional Information

`infrahub upgrade --check` afterwards does report the truth (`4 branches need rebase`), so the state is discoverable — but only by an operator who knows not to trust the command's own success report.

Issue #10307, opened 2026-08-18
bug: validate_database does not retry TransientError, so infrahub upgrade aborts with DatabaseUnavailable while Neo4j is still recovering


lancamat1:
### Component

API Server / GraphQL

### Infrahub version

1.10.6 (Enterprise). Code path unchanged in 1.10.8 and 1.11.0b2; encountered while running `infrahub upgrade` against a 4.5 GB database restored from 1.10.6.

### Current Behavior

Running `infrahub upgrade` shortly after starting the stack aborts within seconds with exit code 1:

```
neo4j.exceptions.DatabaseUnavailable: {neo4j_code: Neo.TransientError.General.DatabaseUnavailable}
{message: The database is not currently available to serve your request ...}
{gql_status: 08N09} The database `neo4j` is currently unavailable.
```

Call path: `workers/dependencies.py build_database` -> `database/__init__.py get_db` (called with `retry=5`) -> `database/__init__.py validate_database` -> `session.run("SHOW TRANSACTIONS")`.

`validate_database` only catches `ClientError`:

```python
    except ClientError as exc:
        ...
        if retry == 0:
            raise
        await asyncio.sleep(retry_interval)
        await validate_database(..., retry=retry - 1, create_db=False)
```

`DatabaseUnavailable` is a `TransientError`, not a `ClientError` (`DatabaseUnavailable.__mro__` -> `TransientError` -> `Neo4jError`; `issubclass(DatabaseUnavailable, ClientError)` is `False`), so the `retry=5` budget never applies and the first attempt propagates. `is_retriable_db_error` in the same module already classifies `TransientError` as retriable for transactions, so the two paths disagree about the same error class.

The window exists because a multi-gigabyte Neo4j store keeps recovering after the container reports healthy: the compose healthcheck is `wget -O /dev/null http://localhost:7474` (`development/docker-compose-database-neo4j.yml` and both testcontainers compose files), and Neo4j answers on 7474 before individual databases are online.

### Expected Behavior

A database that is still starting is waited out rather than aborting the command, since `get_db` is already called with a retry budget for exactly this situation.

### Steps to Reproduce

1. Stack with a multi-gigabyte Neo4j store.
2. Restart the database, wait only for the container to report healthy.
3. Run `infrahub upgrade` -> exits 1 with `DatabaseUnavailable`.
4. Wait until `cypher-shell -d system "SHOW DATABASE neo4j YIELD currentStatus"` returns `online`, re-run -> succeeds.

[...]

PR #10311, opened 2026-08-18
docs: cover branch rebase in the upgrade docs

lancamat1:
## Why

Three gaps in the upgrade docs, each of which an operator hits in practice. Kept deliberately minimal — one addition per gap, no restructuring.

## What changed, and why each line is there

**`overview.mdx` — "Before you upgrade" gains an "Open branches" line**

The section covers upgrade path, backups and release notes, but says nothing about branches, even though Step 6/6 rebases every open branch. A branch that cannot be rebased is therefore discovered mid-upgrade. The line points at `infrahub upgrade --check`, which already reports the branches the upgrade will rebase, so they can be dealt with beforehand.

**`overview.mdx` — the transaction memory known issue gains two sentences**

1. The section is scoped to "large database/schema migrations", but the branch rebase in Step 6/6 can reach the same limit, and its requirement grows with the number of changes on the branch rather than with the size of the database. An operator reading the current text does not expect a rebase failure.
2. Setting `dbms.memory.transaction.total.max` to `0` removes the ceiling, but the memory is still allocated from the JVM heap, so `server.memory.heap.max_size` has to be large enough as well. Without this, the documented workaround can be applied and still fail.

**`community.mdx` — "Verify the upgrade" also uses `--check`**

Verification stopped at `infrahub db showmigrations`, which passes even when the core schema or a branch rebase is still outstanding, so an upgrade can look verified while it is not. `infrahub upgrade --check` reports all three areas.

## Related issues

Reported separately, against 1.10.6:

- #10307 — `validate_database` does not retry `TransientError`, so a command can abort with `DatabaseUnavailable` while Neo4j is still recovering
- #10308 — `infrahub upgrade` prints `Upgrade complete SUCCESS` and exits 0 when branch rebases failed
- #10309 — the per-branch rebase loop only catches `ValidationError` and `MigrationFailureError`, so any other error aborts the remaining branches
- #10310 — branch rebase memory use scales with the branch diff size

The two paragraphs about the memory limit are interim guidance; if #10310 lands they should be revisited.

## Verification

- `uv run invoke docs.lint` — 0 markdownlint errors; vale reports 0 warnings on the changed files (the 25 warnings in the run are pre-existing in other pages)
- `cd docs && npm run build` — succeeds, no broken links or anchors

### DIFF
--- a/docs/docs/deploy-manage/maintain-upgrade/upgrade/community.mdx
+++ b/docs/docs/deploy-manage/maintain-upgrade/upgrade/community.mdx
@@ -92,7 +92,7 @@ invoke demo.start
 
 ## Verify the upgrade
 
-After the `infrahub upgrade` command exits with `Upgrade complete SUCCESS`, run `infrahub db showmigrations` against the same database to confirm every migration shows `[X]` and that the database version matches the target version reported in the header. If any migrations are still pending, re-run `infrahub upgrade`.
+After the `infrahub upgrade` command exits with `Upgrade complete SUCCESS`, run `infrahub db showmigrations` against the same database to confirm every migration shows `[X]` and that the database version matches the target version reported in the header. Then run `infrahub upgrade --check`, which reports pending migrations, core schema differences and branches that still need rebase, so all three parts of the upgrade are covered rather than the migrations alone. Re-run `infrahub upgrade` until it reports nothing outstanding.
 
 If you skipped `--rebase-branches` and the upgrade reported open branches that need rebase, they are left in the `needs-rebase` state. You can rebase them manually later, or re-run with `infrahub upgrade --rebase-branches` (optionally combined with `--interactive` to confirm each branch).
 
--- a/docs/docs/deploy-manage/maintain-upgrade/upgrade/overview.mdx
+++ b/docs/docs/deploy-manage/maintain-upgrade/upgrade/overview.mdx
@@ -12,6 +12,8 @@ Upgrading Infrahub involves pulling the latest container images, running databas
 
 **Backup first**: Even though a smooth migration is anticipated, we strongly recommend creating a backup before upgrading. See [Backup and restore](../database-backup/backup-and-restore.mdx).
 
+**Open branches**: Run `infrahub upgrade --check` before you start. It reports the branches the upgrade will rebase, so a branch that needs attention first can be resolved or deleted instead of failing during the upgrade.
+
 **Release notes**: Review the [release notes](../../../release-notes/infrahub/) for any version-specific upgrade instructions. In Infrahub 1.2 and later, the upgrade process was streamlined with a unified upgrade command. For earlier versions, refer to the release notes for specific instructions.
 
 ## What the upgrade does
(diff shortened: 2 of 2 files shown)


docs/docs/deploy-manage/maintain-upgrade/upgrade/overview.mdx
```diff
@@ -176,11 +178,11 @@ Select the guide for your edition and deployment:
 
 ### Migration failing because of transaction memory limit reached in Neo4j
 
-For large database/schema migrations, you may encounter a `Transaction memory limit reached` error in Neo4j.
+For large database/schema migrations, you may encounter a `Transaction memory limit reached` error in Neo4j. The branch rebase in Step 6/6 can reach the same limit, because the memory it needs grows with the number of changes on the branch being rebased.
```
gmazoyer, line 181: I'd avoid having step number in the doc. If we ever introduce a new step at some point that number would be stale.

docs/docs/deploy-manage/maintain-upgrade/upgrade/community.mdx
```diff
@@ -92,7 +92,7 @@ invoke demo.start
 
 ## Verify the upgrade
 
-After the `infrahub upgrade` command exits with `Upgrade complete SUCCESS`, run `infrahub db showmigrations` against the same database to confirm every migration shows `[X]` and that the database version matches the target version reported in the header. If any migrations are still pending, re-run `infrahub upgrade`.
+After the `infrahub upgrade` command exits with `Upgrade complete SUCCESS`, run `infrahub db showmigrations` against the same database to confirm every migration shows `[X]` and that the database version matches the target version reported in the header. Then run `infrahub upgrade --check`, which reports pending migrations, core schema differences and branches that still need rebase, so all three parts of the upgrade are covered rather than the migrations alone. Re-run `infrahub upgrade` until it reports nothing outstanding.
```
ajtmccarty, line 95: if you keep running `infrahub upgrade` without the `--rebase-branches` option, I think it will keep reporting branches that need to be rebased, so it wouldn't ever reach the "until it reports nothing outstanding" state. I think the migration operator would understand how to proceed, but it doesn't seem quite right

docs/docs/deploy-manage/maintain-upgrade/upgrade/overview.mdx
```diff
@@ -12,6 +12,8 @@ Upgrading Infrahub involves pulling the latest container images, running databas
 
 **Backup first**: Even though a smooth migration is anticipated, we strongly recommend creating a backup before upgrading. See [Backup and restore](../database-backup/backup-and-restore.mdx).
 
+**Open branches**: Run `infrahub upgrade --check` before you start. It reports the branches the upgrade will rebase, so a branch that needs attention first can be resolved or deleted instead of failing during the upgrade.
```
BaptisteGi, line 15: Talking about open branches, should we consider adding a recommendation to delete stale branches so it avoids potential issue and speedup the migration.

PR #10312, opened 2026-08-18
fix(backend): retire branch-agnostic field edges on object deletion (…

ajtmccarty:
# Why

Removing a branch-agnostic attribute or relationship from a branch-aware schema always left global-
branch edges open. The removals wrote a branch-scoped `deleted` edge, which hid the field on the
branch being updated, but the edges carrying the value live on the global branch and were never
closed. For a unique attribute backed by a pool, the value stayed reserved with nothing able to
read it and that value became permanently unallocatable.

Goal: at the schema-removal enforcement point, a branch-agnostic field's global edges are closed
if and only if they are no longer reachable by ANY branch.

Non-goals — each is its own slice, so this PR is deliberately narrow:

- The remaining enforcement points: branch merge, rebase, branch deletion.
- The repair migration for damage already in customer databases.
- Rollback and metadata updates both have issues with branch-agnostic fields on branch-aware
  objects that will be addressed in forthcoming PRs
- The changelog entry and user-facing documentation, deferred until the enforcement points are
  complete so the behaviour is described once rather than five times.

Closes 

> **Stacked PR.** This targets `retire-agnostic-edges-ifc-2843-slice1`, not the base branch. Review
> that one first; the retention predicate this composes lives there.

## What changed

Behavioral changes:

- Removing a branch-agnostic attribute or relationship from a schema now closes its global edges,
  releasing the value, when no branch can still read the field.
- If any branch can still read it — typically one that forked before the removal — nothing is
  released, and that branch goes on reading the value.

Implementation notes:

- The closure is folded into `AttributeRemoveQuery` and `NodeRelationshipRemoveMigrationQuery` as a
  shared Cypher unit subquery, `CLOSE_UNRETAINED_AGNOSTIC_FIELDS`, rather than running as a separate
  pass afterwards. The removal has already resolved which vertices belong to the field — the kind plus
  its profile and template variants, minus the kinds that still declare it — and it is the removal's
  own writes that decide retention. A later pass would have to re-derive both.
- The subquery returns no rows. That is what keeps both host queries' `RETURN` and their
  `nbr_migrations_executed` counts untouched while the predicate filters candidates inside it, and it
  puts the closure in the same statement, and therefore the same transaction, as the removal.
- Candidates are handed in as a collected list rather than re-selected from the graph, which also
  makes the branch lookup inside the predicate run once per statement instead of once per candidate.
- Retention is not reimplemented here: the fragment composes the existing
  `UNRETAINED_AGNOSTIC_FIELD_PREDICATE` verbatim.

What stayed the same:

- No graph migration, no `GRAPH_VERSION` bump, no schema changes.
- No change to the rollback, to merge, rebase, or recovery.
- Both removal queries keep their existing `RETURN` and their migration-count semantics.
- Branch-aware fields are untouched: a cheap `EXISTS` gate drops them before any per-branch work runs.

### Suggested review order

1. `backend/infrahub/core/query/agnostic_field_closure.py` — the fragment, and why candidates are
   passed in as a list.
2. The two splice sites in `attribute_remove.py` and `node_relationship_remove.py` — note the
   `collect` / `UNWIND` round-trip, which is what preserves the migration count.
3. `backend/tests/component/core/migrations/schema/test_agnostic_field_removal.py` — the closure cases
   first, then the deferral cases, which are the ones a naive implementation breaks.

## Testing

Ten component tests, asserting graph shape — edge presence, `status`, `to` — rather than API
responses, because this is a graph-shape bug the API hides.

Each was mutation-checked rather than only observed to pass:

| Mutation | Result |
|---|---|
| Closure fragment emptied | the closure cases fail, the deferral cases pass |
| Retention predicate dropped, so closure is unconditional | the deferral cases fail, the closure cases pass |
| Field-edge axis dropped, making retention a disjunction across axes | both cross-axis tests fail |
| Peer count leaked across candidates instead of per-vertex | only the divergent-retention test fails |

The existence axis is exercised by slice 1's suite rather than this one: `Node.delete` tombstones the
branch-scoped field edge alongside the existence edge, so no scenario reachable from a schema removal
holds a live field edge over a dead owner.

## Impact & rollout

- **Backward compatibility:** no breaking changes. Orphaned edges already in a database are not
  repaired — that needs the migration in a later slice.
- **Performance:** the retention predicate's cost grows with open-branch count, not graph size, and the
  `EXISTS` gate keeps branch-aware fields out of it entirely. Not benchmarked here; the timing gate is
  attached to the branch-deletion slice, which is the one that adds a query the others do not.
- **Deployment notes:** safe to deploy. Behaviour changes only when a branch-agnostic field is removed
  from a schema.

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../.agents/skills/creating-changelog-entries/SKILL.md) added (`uv run towncrier create ...`) — deferred, see non-goals
- [ ] External docs updated (if user-facing or ops-facing change) — deferred, see non-goals

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -25,6 +25,7 @@ from infrahub.core.metadata.model import MetadataInfo
 from infrahub.core.protocols import CoreNumberPool, CoreObjectTemplate
 from infrahub.core.protocols_base import CoreNode
 from infrahub.core.query.node import NodeCheckIDQuery, NodeCreateAllQuery, NodeDeleteQuery, NodeUpdateMetadataQuery
+from infrahub.core.query.node_agnostic_retirement import RetireNodeAgnosticFieldsQuery
 from infrahub.core.schema import (
     AttributeSchema,
     NodeSchema,
@@ -1249,6 +1250,18 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
 
         query = await NodeDeleteQuery.init(db=db, node=self, at=delete_at, user_id=user_id)
         await query.execute(db=db)
+
+        retirement_query = await RetireNodeAgnosticFieldsQuery.init(db=db, node_uuid=self.get_id(), at=delete_at)
+        await retirement_query.execute(db=db)
+        retired = retirement_query.get_data()
+        if retired.edges_closed:
+            log.debug(
+                "Retired branch-agnostic fields of a deleted node",
+                node_id=self.get_id(),
+                node_kind=self.get_kind(),
+                edges_closed=retired.edges_closed,
+                at=delete_at.to_string(),
+            )
 
         self._node_changelog = node_changelog
 
--- a/backend/infrahub/core/query/agnostic_retention.py
+++ b/backend/infrahub/core/query/agnostic_retention.py
@@ -1,0 +1,112 @@
+"""The one predicate that decides whether a branch-agnostic field is still retained by any branch.
+
+A branch-agnostic field keeps its value on edges carrying the global branch name, which every branch
+reads. Those edges may only be closed once **NO** branch can still reach a live owner over a live field
+edge. Cypher for this is consolidated here to ensure consistency.
+
+Retention is decided per branch **and** per linked vertex: under that branch's view, the vertex's
+existence edge and its edge to the field must both resolve to `active`. Surviving edges are summed
+per branch, and only then is the maximum taken across branches. Retention is a disjunction of what
+each branch holds live on its own, never a pool the branches contribute to jointly.
+
+A `:Relationship` needs two qualifying field edges rather than one, because a relationship missing a
+peer is not a relationship. A `:Relationship` must also have two distinct active peers to be
+considered active.
+
+Assumption: every branch forks from the default branch. A branch-of-branch feature would not extend
+this logic.
+"""
+
+# Expects `field` in scope, one row per candidate vertex, plus the `$global_branch_name` and `$at`
+# parameters. Emits the candidates no branch retains, with `field` as the only variable in scope.
+UNRETAINED_AGNOSTIC_FIELD_PREDICATE = """
+WITH collect(field) AS candidates
+
+// ----------------------
+// The branches are read once for the whole run and carried as a list.
+// ----------------------
+MATCH (branch:Branch)
+WHERE branch.name <> $global_branch_name
+WITH
+    candidates,
+    collect({
+        name: branch.name,
+        origin_name: CASE WHEN branch.is_default THEN NULL ELSE branch.origin_branch END,
+        origin_at: CASE
+            WHEN branch.is_default THEN NULL
+            WHEN branch.branched_from < $at THEN branch.branched_from
+            ELSE $at
+        END
+    }) AS branch_windows
+
+UNWIND candidates AS field
+WITH field, branch_windows, CASE WHEN field:Relationship THEN 2 ELSE 1 END AS required_live_peers
+
+CALL (field, branch_windows) {
+    UNWIND branch_windows AS branch_window
+    WITH
+        field,
+        branch_window.name AS branch_name,
+        branch_window.origin_name AS origin_name,
+        branch_window.origin_at AS origin_at
+
+    // ----------------------
+    // Count this `field`'s  active links to :Node vertices on this branch
+    // ----------------------
+    MATCH (node:Node)-[field_edge:HAS_ATTRIBUTE|IS_RELATED]-(field)
+    WHERE (field_edge.branch IN [$global_branch_name, branch_name]
+           AND field_edge.from <= $at
+           AND (field_edge.to IS NULL OR field_edge.to > $at))
+       OR (field_edge.branch = origin_name
+           AND field_edge.from <= origin_at
+           AND (field_edge.to IS NULL OR field_edge.to > origin_at))
+    WITH
+        branch_name,
+        origin_name,
+        origin_at,
+        node,
+        field_edge.branch_level AS field_edge_level,
+        field_edge.from AS field_edge_from,
+        field_edge.status AS field_edge_status
+    ORDER BY field_edge_level DESC, field_edge_from DESC, field_edge_status ASC
+    WITH
+        branch_name,
+        origin_name,
+        origin_at,
+        node,
+        collect(field_edge_status)[0] AS winning_field_edge_status
+    WHERE winning_field_edge_status = "active"
+
+    // ----------------------
+    // Check that each linked :Node vertex is active on this branch.
+    // The global branch stays in the existence match so that an owner which is itself branch-agnostic
+    // reads as live on every branch and is therefore retained.
+    // ----------------------
+    MATCH (node)-[existence:IS_PART_OF]->(:Root)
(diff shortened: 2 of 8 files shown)


backend/infrahub/core/node/__init__.py
```diff
@@ -1250,6 +1251,18 @@ async def delete(self, db: InfrahubDatabase, user_id: str = SYSTEM_USER_ID, at:
         query = await NodeDeleteQuery.init(db=db, node=self, at=delete_at, user_id=user_id)
         await query.execute(db=db)
 
+        retirement_query = await RetireNodeAgnosticFieldsQuery.init(db=db, node_uuid=self.get_id(), at=delete_at)
+        await retirement_query.execute(db=db)
```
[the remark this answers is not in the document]
ajtmccarty: updated `run_check_merge_conflicts()` in `git/tasks.py` to use a transaction instead of a session. this is a pre-existing bug b/c `NodeManager.delete()` will run multiple `Node.delete()` calls in sequence, so it is possible for one to fail partway through, leaving the group semi-deleted. the move to use a transaction should fix this

backend/infrahub/core/query/agnostic_retention.py
```diff
@@ -0,0 +1,112 @@
...
+MATCH (branch:Branch)
+WHERE branch.name <> $global_branch_name
+WITH
+    candidates,
+    collect({
+        name: branch.name,
+        origin_name: CASE WHEN branch.is_default THEN NULL ELSE branch.origin_branch END,
+        origin_at: CASE
+            WHEN branch.is_default THEN NULL
+            WHEN branch.branched_from < $at THEN branch.branched_from
```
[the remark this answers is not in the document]
ajtmccarty: non-isolated branches do not exist anymore. the references to them are dead code

backend/infrahub/core/query/node_agnostic_retirement.py
```diff
@@ -0,0 +1,73 @@
...
+// -----------------
+MATCH (anchor_node:Node {uuid: $node_uuid})-[anchor:HAS_ATTRIBUTE|IS_RELATED]-(field:Attribute|Relationship)
+WHERE anchor.branch = $global_branch_name
+  AND anchor.status = "active"
+  AND anchor.to IS NULL
+WITH DISTINCT field
+%(unretained_predicate)s
+
+MATCH (field)-[e]-()
+WHERE e.branch = $global_branch_name
```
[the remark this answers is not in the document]
ajtmccarty: deletion is always going to run at the current_time +/- a few microseconds. updating the query to make sure $at is > the from time just to be safe

backend/infrahub/core/query/agnostic_retention.py
```diff
@@ -0,0 +1,112 @@
...
+        existence.status AS existence_status
+    ORDER BY existence_level DESC, existence_from DESC, existence_status ASC
+    WITH branch_name, node, collect(existence_status)[0] AS winning_existence
+    WHERE winning_existence = "active"
+
+    // ----------------------
+    // Peers are counted by uuid. Kind/inheritance migration leaves multiple Node vertices with
+    // the same uuid for a single entity
+    // ----------------------
+    WITH branch_name, count(DISTINCT node.uuid) AS live_peer_count
```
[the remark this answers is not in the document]
ajtmccarty: we don't support an object referencing itself in a relationship

backend/infrahub/git/tasks.py
```diff
@@ -1170,8 +1170,8 @@ async def run_check_merge_conflicts(model: CheckRepositoryMergeConflicts) -> Val
         await check.save()
 
     database = await get_database()
-    async with database.start_session() as db:
-        await NodeManager.delete(db=db, nodes=list(existing_checks.values()))
+    async with database.start_transaction() as dbt:
+        await NodeManager.delete(db=dbt, nodes=list(existing_checks.values()))
```
gmazoyer: Is this related to this PR or just a fix along the way?

backend/tests/helpers/agnostic_edges.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def edge_summary(edges: list[EdgeState]) -> list[tuple[str, str, str]]:
+    """Order-independent view of a vertex's edges, for before/after comparison."""
+    return sorted((edge.edge_type, edge.status, edge.to_time or "") for edge in edges)
+
+
+def branch_summary(edges: list[EdgeState]) -> list[tuple[str, str, str]]:
+    """Order-independent view keyed on the branch, for the assertions where that is the point."""
+    return sorted((edge.branch, edge.status, edge.to_time or "") for edge in edges)
```
gmazoyer: This does not seem to be used anywhere.

backend/tests/helpers/agnostic_edges.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def expected_closed_at(edges: list[EdgeState], at: Timestamp) -> list[tuple[str, str, str]]:
+    """What `edge_summary` should return once every type present has been closed at `at`."""
+    return sorted((edge_type, "active", at.to_string()) for edge_type in {edge.edge_type for edge in edges})
+
+
+def to_times(edges: list[EdgeState]) -> set[str | None]:
+    return {edge.to_time for edge in edges}
```
gmazoyer: This does not seem to be used anywhere.

backend/tests/helpers/agnostic_edges.py
ajtmccarty: removed

backend/tests/helpers/agnostic_edges.py
ajtmccarty: removed

backend/infrahub/git/tasks.py
ajtmccarty: cubic complained about this for some reason. I've updated it again to give each call to `NodeManager.delete` its own transaction, which I think makes more sense. the validator updates in this method should be idempotent. they will get recomputed and updated from scratch each time, so it is not the end of the world if one of these deletes fails, but the _whole_ delete should fail and not just part of it

backend/infrahub/git/tasks.py
```diff
@@ -1170,8 +1170,9 @@ async def run_check_merge_conflicts(model: CheckRepositoryMergeConflicts) -> Val
         await check.save()
 
     database = await get_database()
-    async with database.start_session() as db:
-        await NodeManager.delete(db=db, nodes=list(existing_checks.values()))
+    for check in existing_checks.values():
+        async with database.start_transaction() as dbt:
+            await NodeManager.delete(db=dbt, nodes=[check])
```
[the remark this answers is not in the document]
ajtmccarty: I don't want to do one transaction in case it runs out of memory. the function containing this call, `run_check_merge_conflicts()`, is idempotent. it recomputes the expected state and sets it every time. so re-running following a partial delete will fix it

gmazoyer (approved):

Issue #10063, opened 2026-07-28
bug: trigger rule objects in a repository's objects/ cannot resolve generator definitions on first import


dgarros:
## Component

Git Integration

## Infrahub version

1.10.5

## Current Behavior

A repository cannot declare its action/trigger rules (`CoreGeneratorAction` + `CoreNodeTriggerRule`) as objects under `.infrahub.yml`'s `objects:` section. On the **first** import of the repository the import fails, because `CoreGeneratorAction.generator` is a mandatory relationship to a `CoreGeneratorDefinition` that does not exist yet at the point objects are imported.

`apply_import_plan` (`backend/infrahub/git/integrator.py`) applies a single fixed-order pass:

```
schema → queries → python transforms → jinja2 transforms
       → OBJECTS (+menus) → check definitions → GENERATOR DEFINITIONS → artifact definitions
```

Objects are imported several steps before generator definitions, with this rationale in the source:

> Checks, generators and artifact definitions are imported after objects because their targets reference groups that are defined as objects in the repository.

That ordering is required by `CoreGeneratorDefinition.targets` (mandatory, → `CoreGroup`), which typically resolves to a `CoreStandardGroup` declared in the repo's `objects/`. But trigger rules need the opposite order, producing a cycle that a single pass cannot satisfy:

```
CoreGeneratorAction  ──needs──>  CoreGeneratorDefinition  ──needs──>  CoreGroup  ──from──>  objects/
   (objects step)                  (generator defs step)              (objects step)
```

Two aggravating factors:

- **It only fails on the first import.** Every subsequent import finds the `CoreGeneratorDefinition` nodes left behind by the previous run, so the objects step resolves and the import succeeds. This makes the failure look intermittent or environment-specific when it is actually deterministic on an empty graph.
- **There is no ordering workaround.** `import_objects` walks `config_file.objects` then `config_file.menus` with no kind-aware deferral and no second pass, so renaming the file (`99_triggers.yml`) keeps it in the same step. `InfrahubRepositoryConfig` also exposes no `triggers:` section that would be applied after generator definitions — the available keys are `check_definitions, schemas, jinja2_transforms, artifact_definitions, python_transforms, generator_definitions, queries, graphql_fragments, objects, menus`.

[...]

Issue #9248, opened 2026-05-13
CoreAccountRole cannot be added to CoreRepositoryGroup.members — repository import fails (also affects CoreGraphQLQuery)


fatih-acar:
### Summary

Importing objects of kind `CoreAccountRole` from a Git repository fails because the `members` relationship of `CoreRepositoryGroup` (inherited from `CoreGroup`) refuses to accept `CoreAccountRole` as a peer. The same root cause affects `CoreGraphQLQuery`.

### Where it surfaced

While importing the bootstrap account roles from a repository (`data/bootstrap/02_account_roles.yml`, 4 `CoreAccountRole` entries), the import fails with:

```
CoreAccountRole - 18af1968-... cannot be added to relationship,
must be of type: [...long list...] at members
```

That long list includes `CoreAccount`, `CoreAccountGroup`, `CoreGlobalPermission`, `CoreObjectPermission`, but **not** `CoreAccountRole`.

The integrator (`backend/infrahub/git/integrator.py`, `_import_file_paths`) wraps each repository import in:

```python
async with self.sdk.start_tracking(
    identifier=f"group-repo-{object_type.value}-{self.id}",
    delete_unused_nodes=True,
    branch=branch_name,
    group_type="CoreRepositoryGroup",
    group_params={"content": object_type.value, "repository": str(self.id)},
):
    ...
```

so every freshly-imported object is registered as a member of the tracking group `group-repo-object-<repo-id>` (a `CoreRepositoryGroup`). When the upsert for that group tries to attach the 4 newly-created `CoreAccountRole` IDs to `members`, the peer-kind constraint rejects them and the whole upsert — and therefore the whole import — fails.

### Root cause

`CoreGroup.members` is defined with `peer=InfrahubKind.NODE` (the `CoreNode` generic). The peer-kind constraint at `backend/infrahub/core/relationship/constraints/peer_kind.py:34` uses `CoreNode.used_by` as the allowlist of acceptable peer kinds.

`CoreNode.used_by` is populated in `SchemaBranch.process_inheritance` (`backend/infrahub/core/schema/schema_branch.py:1819`):

```python
if node.inherit_from or node.namespace == "Builtin" or node.namespace not in RESTRICTED_NAMESPACES:
    generics_used_by[InfrahubKind.NODE].append(node.kind)
```

A node in a restricted namespace (e.g. `Core`) is only added when it has `inherit_from`. The only two `Core`-namespace nodes without `inherit_from` are:

[...]

Issue #6886, opened 2025-07-24
bug: Object created using object file not linked to my repository


BaptisteGi:
### Component

Git Integration

### Infrahub version

1.3.3

### Current Behavior

I'm creating a `CoreStandardGroup` using an object file and git integration. The object is properly created but the `objects` tab of my repository is empty.

### Expected Behavior

I expect the group to appear in the object tab

### Steps to Reproduce

- spin infrahub
- connect a repository
- create a branch in the repository
- add the following file
    ```yml objects/groups.yml
    ---
    apiVersion: infrahub.app/v1
    kind: Object
    spec:
      kind: CoreStandardGroup
      data:
        - name: "widgets"
    ```
- import the object in the `.infrahub.yml`
    ```yml .infrahub.yml
    ---
    objects:
      - "objects/groups.yml"
    ```
- commit and push
- going back to infrahub > select the correct branch
=> Group object is created however looking at the repository the object list is empty

<img width="2610" height="1527" alt="Image" src="https://github.com/user-attachments/assets/fc56f87f-f25d-47f5-ae12-c113bb6dcdb7" />

ogenstad:
My guess is that it's because the tracking is done by members of the `CoreRepositoryGroup` and a group can't add another group as a member because the `members` relationship on `CoreGroup` points to `CoreNode`. And group nodes aren't inheriting from `CoreNode` so they can't be members.

Either we'd need to change that behaviour or track these "group member nodes" in a different way.

PR #10316, opened 2026-08-19
fix: emit group.member_added only for the group named in the mutation (closes #10314)

### LINKED ISSUE #10314: bug: adding a group as a member of another group emits group.member_added on the wrong node, so every repository import fires a failing generator run
### Component

### Infrahub version

Observed on 1.9.7. The code paths are unchanged on `stable` (1.10.8), and identical in 1.9.8 and 1.11.0b2.

### Current Behavior

When a node that is itself a group is added as a member of another group, the resulting `infrahub.group.member_added` event names the **enrolled group** as the group that gained a member, and the **enclosing group** as the added member. The relationship is inverted.

Every repository import hits this: `_import_file_paths` wraps each object pass in a `CoreRepositoryGroup` tracking group, so every group defined in an `objects/*.yml` file is enrolled into it, and each enrolment emits a `member_added` event against the imported group.

If a `CoreGroupTriggerRule` with `member_update: added` watches one of those groups, the event matches it and its `CoreGeneratorAction` runs with the tracking group as the target. The run then fails:

```
ValueError: Target <repository-group-id> is not part of the group <generator-group-id>
```

Observed on a 1.9.7 instance — a FAILED flow `Running generator`, `created_by` AUTOMATION `action_trigger_rule::l3vpn-service-intent-enrolled`, with:

```json
"parameters": {"members": [{"id": "18cd03d3-…", "kind": "CoreRepositoryGroup"}], …}
```

The watched group gained no member; no service intent was enrolled; nothing in the graph changed.

The failure raises before any generator run is submitted, so no data is written. The cost is a failed flow on every import, indistinguishable in the task list from a real generator failure — which makes "did the pipeline succeed?" unanswerable without inspecting each failure by hand.

### Expected Behavior

- Adding group B as a member of group A emits `member_added` for A with B as the member, and emits nothing for B.
- Importing a repository whose `objects/` define groups produces no failed generator runs.

### Steps to Reproduce

1. Start an empty Infrahub and register a repository containing:
   - `objects/groups.yml` defining a `CoreGeneratorGroup` named `widgets`
   - a `generator_definitions` entry in `.infrahub.yml` whose `targets` is `widgets`
   - `objects/triggers.yml` defining a `CoreGeneratorAction` for that definition, and a `CoreGroupTriggerRule` with `group: widgets`, `member_update: added`, `branch_scope: default_branch`
2. Import it. On current `stable` the trigger objects must land after the generator definitions exists, so use two commits — first without `objects/triggers.yml`, then with it (see #10063 / #10196).
3. Push any further commit that re-imports the objects.
4. A FAILED `Running generator` flow appears with the `ValueError` above, and `parameters.members[0].kind == "CoreRepositoryGroup"`. `widgets` has no members.

In the observed run, the event fired once per object-import pass (three passes, 30s apart); only the passes after the trigger rule existed produced a failed run, which makes the defect look like a race when it is deterministic.

### Additional Information

Trace, against `stable`:

1. `git/integrator.py:1286` — object import runs inside `sdk.start_tracking(group_type="CoreRepositoryGroup")`; the SDK closes it by upserting that group with every touched node in `members` (`infrahub_sdk/query_groups.py`, `update_group`). Groups created from object files are included.
2. `core/changelog/models.py:594` — `RelationshipChangelogGetter._process_added_peers` builds the peer-side changelog by resolving the reverse relationship **by identifier alone**: `peer_schema.get_relationship_by_identifier(id=str(rel_schema.identifier))`.
3. `core/schema/basenode_schema.py:367` — that helper returns the first relationship whose identifier matches, without considering direction.
4. `core/schema/schema_branch.py:2274` — `add_groups()` skips any schema inheriting `CoreGroup` when injecting `member_of_groups`. A group kind therefore has exactly one relationship with identifier `group_member`: its own `members` (`core/schema/definitions/core/group.py`).
5. Consequently the peer-side changelog for a group peer is written onto `members` rather than `member_of_groups`, stating that the enrolled group gained the enclosing group as a member.
6. `groups/parsers.py` — `GroupNodeMutationParser._get_applicable_events` sees a group-kind node with an added `members` peer and emits `GroupMemberAddedEvent` for it.
7. `actions/models.py:208` — `CoreGroupTriggerRule._from_group_trigger` matches on `infrahub.node.kind` + `infrahub.node.id` only, and passes `payload.data.members` through to the workflow unfiltered.
8. `actions/tasks.py:274` — `_run_generators` intersects those ids with the target group's real members, gets an empty set, and raises.

Note for anyone adding a guard on the event: `infrahub.node.action` is `"created"` for *every* `GroupMemberAddedEvent` (`events/group_action.py`, `action: MutationAction = MutationAction.CREATED`), so it does not distinguish this case.

Related:

- #6886 — groups created from an object file are not linked to their repository. In the run above the tracking group's members did not include the imported group either, which would explain why the enrolment is recorded as newly *added* on every import pass rather than only the first.
- #9248 — also concerns which peers `CoreRepositoryGroup.members` accepts during import.
- #10063 / #10196 — the import ordering for repository-defined trigger rules. Separate defect in the same import path; it changes when this one is reachable, not whether it happens.

iddocohen:
# Why

Adding a group as a member of another group emitted a second `infrahub.group.member_added` event naming the *enrolled* group as the one that gained a member, with the enclosing group as the member. Every repository import hits this, because object import wraps each pass in a `CoreRepositoryGroup` tracking group and enrols every group defined in `objects/*.yml` into it. A `CoreGroupTriggerRule` watching one of those groups matched the spurious event and ran its `CoreGeneratorAction` against the tracking group, which then failed with `ValueError: Target <id> is not part of the group <id>`.

Goal: one membership change produces one `member_added` event, for the group named in the mutation.

Non-goals: this PR does not change the import path, the tracking group, or which peers it collects (#6886, #9248), the import ordering for repository-defined trigger rules (#10063 / #10196), or the fact that a group-to-group membership is readable from both groups (see *Risky or uncertain parts*).

Closes #10314

## What changed

Behavioral changes:

- Enrolling group B into group A now emits a single `member_added` event, for A with B as the member. Previously a second event fired for B, claiming it had gained A.
- The peer-side `NodeUpdatedEvent` that accompanied that spurious event is gone too, so a `CoreNodeTriggerRule` watching `members` on B no longer fires either.
- Secondary changelogs now pick the correct side of a hierarchical relationship. `parent` and `children` share one identifier on hierarchical kinds, and the previous lookup returned whichever appeared first in the schema's relationship list.

Implementation notes:

- Root cause was in `RelationshipChangelogGetter`: it resolved the peer's "reverse" relationship with `get_relationship_by_identifier`, which matches on identifier alone. Group kinds never receive `member_of_groups` (`add_groups()` skips anything inheriting `CoreGroup`), so a group peer's only `group_member` relationship is its own inherited `members` — the same side that was just changed, not the reverse.
- Added `BaseNodeSchema.get_reverse_relationship()`, which requires the candidate to point the opposite way (`RelationshipDirection.neighbor_direction`) and never accepts a bidirectional relationship as its own reverse. It reuses the existing `get_relationships_by_identifier` rather than introducing new machinery.
- Both `_process_added_peers` and `_process_removed_peers` now call it. The existing `if peer_relation:` guard already handles "no reverse relationship", so no control flow changed.

What stayed the same:

- No schema changes, no migration, no GraphQL contract change. `add_groups()` was deliberately **not** changed to inject `member_of_groups` onto group kinds: that would add a field to every group in the public API and still leave two bidirectional `group_member` relationships on one kind to disambiguate.
- `get_relationship_by_identifier` is untouched and keeps its eight other callers.
- No data repair is needed. The fix changes only which events are derived, never what is written.

## How to review

Focus on `backend/infrahub/core/schema/basenode_schema.py` — specifically whether the two rejection rules in `get_reverse_relationship` are the right ones. Everything else follows from it; the changelog change is a two-line call-site swap.

The second rule compares relationship *names*, which deliberately also rejects two different kinds that share a name: the reported failure is exactly that shape, `CoreRepositoryGroup.members` and `CoreGeneratorGroup.members`, both inherited from `CoreGroup`. Restricting the rule to a self-referential relationship instead does not fix the report, and there is no identity token to compare on — inherited relationships get `id = None` and carry no `source_*_id`. The replication test therefore enrols a group into a group of a *different* kind; it fails again if the rule is narrowed to an identity comparison.

Risky or uncertain parts:

- **Worth a maintainer's opinion:** while adding the persistence assertions requested in review, the test showed that a group-to-group membership is readable from *both* groups — after adding B to A's `members`, reloading B also lists A. Both kinds read the one stored link through the same `members` relationship, so the graph genuinely cannot say which group gained a member; only the mutation can. This makes the issue's statement that "nothing in the graph changed" inaccurate, and it means suppressing the second event is a decision about event semantics (report the intent of the mutation) rather than a correction of a claim the graph contradicts. The test pins the current storage behavior so a future change to the group data model surfaces here. If you would rather group-to-group membership were asymmetric, that is a separate modelling change, adjacent to #9248.
- The direction filter would stop emitting a secondary changelog for a pair whose two sides declare mismatched directions (e.g. one `OUTBOUND`, one `BIDIR`). Core schemas contain no such pair: the only explicit directions are the properly inverse hierarchy pair and the resource-pool relationship, whose identifier exists on one side only and already resolved to nothing. A user-defined mismatched pair cannot be traversed from the peer's side anyway.

Alternatives considered: guarding in `GroupNodeMutationParser` or in `CoreGroupTriggerRule`, and tolerating an empty intersection in `_run_generators`. All three leave the fabricated changelog in place and only hide its effects; the last one would also mask genuine misconfiguration, which is what that error exists to report. Gating on `infrahub.node.action` does not work: it is `CREATED` for every `GroupMemberAddedEvent`.

## How to test

```bash
# Replication test (fails on stable, passes here) plus the regression canary in the same file
uv run pytest backend/tests/component/graphql/mutations/test_group_event_collection.py -v --neo4j

# Blast radius: both call sites of RelationshipChangelogGetter, and the action/trigger path
uv run pytest -n 4 --neo4j \
  backend/tests/component/core/changelog/ \
  backend/tests/component/graphql/mutations/ \
  backend/tests/component/graphql/test_mutation_relationship.py \
  backend/tests/component/actions/

```

Results locally: 2 passed in the test file (it failed as `assert [enclosing] == [enclosing, enrolled]` before the fix), 158 passed across the blast-radius suites, 1435 unit tests passed. `uv run invoke format`, `main.lint` and `backend.lint` (ruff, ty, mypy) clean. `backend.generate`, `schema.generate-graphqlschema`, `schema.generate-jsonschema` and `docs.generate` produce no diff, so frontend codegen cannot be affected either. `docs.lint` / `docs.format` could not run locally (`markdownlint-cli2` not installed); no markdown other than the changelog fragment changed.

## Impact & rollout

- **Backward compatibility:** no breaking change to any API or schema. Automation that relied on the spurious second event will stop receiving it — that is the point of the fix.
- **Performance:** unchanged. The new lookup walks the same relationship list as before.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (if user-facing or ops-facing change) — N/A, no documented behavior changes
- [ ] Internal .md docs updated (internal knowledge and AI code tools knowledge) — N/A for the fix; the group-to-group symmetry noted above may deserve a line in `dev/knowledge/backend/events.md` once maintainers decide whether it is intended

### DIFF
--- a/backend/infrahub/core/schema/basenode_schema.py
+++ b/backend/infrahub/core/schema/basenode_schema.py
@@ -395,9 +395,14 @@ class BaseNodeSchema(GeneratedBaseNodeSchema):
         expected_direction = relationship.direction.neighbor_direction
 
         for candidate in self.get_relationships_by_identifier(id=relationship.get_identifier()):
-            if candidate.direction != expected_direction:
-                continue
-            if candidate.direction == RelationshipDirection.BIDIR and candidate.name == relationship.name:
+            # A peer has to point the opposite way, which rules out every candidate on the same side.
+            # Both sides of a bidirectional pair share BIDIR, so there the name is the only thing that
+            # separates a real peer from the relationship we started on: an identical name means two
+            # kinds inherited the same relationship from a common generic (CoreGroup.members on both
+            # CoreStandardGroup and CoreGeneratorGroup), not two sides of one relationship.
+            if candidate.direction != expected_direction or (
+                candidate.direction == RelationshipDirection.BIDIR and candidate.name == relationship.name
+            ):
                 continue
             return candidate
 
--- a/backend/infrahub/core/schema/basenode_schema.py
+++ b/backend/infrahub/core/schema/basenode_schema.py
@@ -11,7 +11,12 @@ from infrahub_sdk.utils import compare_lists, intersection
 from pydantic import ConfigDict, ValidationError, field_validator
 
 from infrahub.computed_attribute.jinja2 import InfrahubJinja2Template
-from infrahub.core.constants import HashableModelState, RelationshipCardinality, RelationshipKind
+from infrahub.core.constants import (
+    HashableModelState,
+    RelationshipCardinality,
+    RelationshipDirection,
+    RelationshipKind,
+)
 from infrahub.core.models import HashableModel, HashableModelDiff
 
 from .attribute_schema import AttributeSchema, get_attribute_schema_class_for_kind
@@ -377,6 +382,26 @@ class BaseNodeSchema(GeneratedBaseNodeSchema):
                 rels.append(item)
 
         return rels
+
+    def get_reverse_relationship(self, relationship: RelationshipSchema) -> RelationshipSchema | None:
+        """Return the relationship on this schema that traverses back along the provided relationship.
+
+        Sharing an identifier is not enough: two kinds can both declare the same identifier on the same
+        side, and a single kind can declare both sides of it. The peer must point the opposite way, and a
+        bidirectional relationship is never its own reverse because both sides are stored asymmetrically.
+
+        Returns None when this schema declares no relationship pointing back.
+        """
+        expected_direction = relationship.direction.neighbor_direction
+
+        for candidate in self.get_relationships_by_identifier(id=relationship.get_identifier()):
+            if candidate.direction != expected_direction:
+                continue
+            if candidate.direction == RelationshipDirection.BIDIR and candidate.name == relationship.name:
+                continue
+            return candidate
+
+        return None
 
     def get_relationships_of_kind(self, relationship_kinds: Iterable[RelationshipKind]) -> list[RelationshipSchema]:
         return [r for r in self.relationships if r.kind in relationship_kinds]
--- a/backend/infrahub/core/changelog/models.py
+++ b/backend/infrahub/core/changelog/models.py
@@ -591,7 +591,7 @@ class RelationshipChangelogGetter:
         primary_changelog: NodeChangelog,
     ) -> list[NodeChangelog]:
         secondaries: list[NodeChangelog] = []
-        peer_relation = peer_schema.get_relationship_by_identifier(id=str(rel_schema.identifier), raise_on_error=False)
+        peer_relation = peer_schema.get_reverse_relationship(relationship=rel_schema)
         if peer_relation:
             node_changelog = NodeChangelog(node_id=peer_id, node_kind=peer_kind, display_label="n/a")
             if peer_relation.cardinality == RelationshipCardinality.ONE:
@@ -625,7 +625,7 @@ class RelationshipChangelogGetter:
         primary_changelog: NodeChangelog,
     ) -> list[NodeChangelog]:
         secondaries: list[NodeChangelog] = []
-        peer_relation = peer_schema.get_relationship_by_identifier(id=str(rel_schema.identifier), raise_on_error=False)
+        peer_relation = peer_schema.get_reverse_relationship(relationship=rel_schema)
         if peer_relation:
             node_changelog = NodeChangelog(node_id=peer_id, node_kind=peer_kind, display_label="n/a")
             if peer_relation.cardinality == RelationshipCardinality.ONE:
--- a/backend/tests/component/graphql/mutations/test_group_event_collection.py
+++ b/backend/tests/component/graphql/mutations/test_group_event_collection.py
@@ -184,3 +184,62 @@ async def test_node_mutation_to_group_event(
     assert len(orphan_group_event.members) == 1
     assert EventNode(id=person_id, kind="TestPerson") in orphan_group_event.members
     assert len(orphan_group_event.ancestors) == 0
+
+
+async def test_group_enrolled_into_group_only_emits_event_for_enclosing_group(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    node_group_schema: None,
+    standard_group_schema: None,
+    session_first_account: AccountSession,
+) -> None:
+    """Enrolling a group into another group reports the enclosing group as the one that gained a member.
+
+    The enrolled group gains nothing, so it must not be reported as having gained a member itself.
+    """
+    enclosing_group = await Node.init(db=db, schema="CoreStandardGroup", branch=default_branch)
+    await enclosing_group.new(db=db, name="enclosing_group")
+    await enclosing_group.save(db=db)
+    enrolled_group = await Node.init(db=db, schema="CoreStandardGroup", branch=default_branch)
+    await enrolled_group.new(db=db, name="enrolled_group")
+    await enrolled_group.save(db=db)
+
+    memory_event = MemoryInfrahubEvent()
+    service = await InfrahubServices.new(event=memory_event)
+    default_branch.update_schema_hash()
+    gql_params = await prepare_graphql_params(
+        db=db, branch=default_branch, service=service, account_session=session_first_account
+    )
+
+    update_query = """
+    mutation($group: String!, $member: String!) {
+        CoreStandardGroupUpdate(data:
+            {
+                id: $group,
+                members: [{id: $member}]
+            }
+        ) {
+            ok
+        }
+    }
+    """
+    result = await graphql(
+        schema=gql_params.schema,
+        source=update_query,
+        context_value=gql_params.context,
+        root_value=None,
+        variable_values={"group": enclosing_group.get_id(), "member": enrolled_group.get_id()},
+    )
+
+    assert not result.errors
+    assert gql_params.context.background
+    await gql_params.context.background()
+
+    member_added_events = [event for event in memory_event.events if isinstance(event, GroupMemberAddedEvent)]
+    assert [event.node_id for event in member_added_events] == [enclosing_group.get_id()]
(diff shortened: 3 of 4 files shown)


backend/infrahub/core/schema/basenode_schema.py
```diff
@@ -378,6 +383,26 @@ def get_relationships_by_identifier(self, id: str) -> list[RelationshipSchema]:
...
+        bidirectional relationship is never its own reverse because both sides are stored asymmetrically.
+
+        Returns None when this schema declares no relationship pointing back.
+        """
+        expected_direction = relationship.direction.neighbor_direction
+
+        for candidate in self.get_relationships_by_identifier(id=relationship.get_identifier()):
+            if candidate.direction != expected_direction:
+                continue
+            if candidate.direction == RelationshipDirection.BIDIR and candidate.name == relationship.name:
```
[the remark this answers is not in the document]
iddocohen: The mechanism is real but the conclusion is inverted for this codebase, so keeping the name comparison.

The reported failure *is* two different kinds using the same name: `CoreRepositoryGroup.members` and `CoreGeneratorGroup.members`, both inherited from `CoreGroup` with identifier `group_member`. Skipping that is the fix, not a regression.

Comparing relationship identity cannot separate the two cases. Inherited relationships get `id = None` (`node_inheritance_handler.py:81`) and, unlike attributes, carry no `source_*_id`, so there is no token shared between the two `members`. Comparing owning kinds instead only skips a same-kind pair, which leaves the reported cross-kind case emitting the spurious event. Verified: swapping the guard for an identity comparison makes the replication test fail again.

Your comment did expose a genuine gap, and thanks for it: the test used one kind on both sides, so it could not tell the two implementations apart. `41b14eb8` enrols a `CoreStandardGroup` into a `TestTrackingGroup` instead, matching the reported scenario, and it now fails under the identity-based guard.

On the residual risk: a same-name bidirectional pair across two kinds that is a genuine two-sided relationship would lose its peer-side changelog. No such pair exists in the core schema — every two-sided pair names its sides differently (`checks`/`validator`, `comments`/`thread`, `roles`/`permissions`, `ip_addresses`/`ip_prefix`), and the one identifier shared by two same-named relationships (`ippool__resource`) is on sibling pool kinds that are never each other's peer.

saltas888: Can we unify this if in one and also leave a code comment explaining the logic?

iddocohen: Done in bc26b3d1d: one condition plus a comment on why the name is what separates an inherited same-name pair (`CoreGroup.members` on two kinds) from a genuine two-sided relationship, and I checked the new guard against the old two-if version across every schema/relationship pair in the core schema (33,028 comparisons, 2,830 resolved peers, 0 mismatches).

ogenstad (commented):
>Adding a group as a member of another group

I think the premise of this bug report and fix needs some work. In Infrahub it's not currently possible to add a group as a member of another group. So it feels like something is wrong today. Possibly the root cause is that the mutation doesn't immediately reject such actions. Is this the case?

While we can argue about what should be possible the CoreGroup member relationship (https://github.com/opsmill/infrahub/blob/infrahub-v1.10.8/backend/infrahub/core/schema/definitions/core/group.py#L46) points to `CoreNode` and a group is not a `CoreNode` type object.

iddocohen:
@ogenstad groups, as far as I understand the code, can be members of other groups today. 

The check allows any kind that inherits something, and every group kind inherits CoreGroup. Infrahub relies on this. Every repository import puts the repository's groups into a CoreRepositoryGroup so they appear in the repository's Objects tab, which is exactly the mutation in question. So rejecting the mutation is, as far as I think, not the right choice.

The use case: a customer's repository declares a group in objects/groups.yml and a CoreGroupTriggerRule that runs a generator when something is added to that group — the documented enrol-then-materialise pattern for auto-provisioning services. That is a normal, supported way to wire a repository.

The harm: every time they push a commit, the import fabricates a "member added" event for that group, so the generator runs against Infrahub's internal repository group and fails. They get a red generator run per trigger rule on every import, forever — failures that look exactly like real generator failures in the task list.

Why it matters: it makes "did my pipeline run clean?" unanswerable without opening each failure by hand, and it rules out any automated gate on failed tasks, since the baseline is never zero. For a customer running this in CI or during a DR rebuild, that is the difference between a usable signal and permanent noise they learn to ignore.

PR #10322, opened 2026-08-19
test: stop the webhook traceback fixture from reconfiguring logging process-wide

fatih-acar:
# Why

`backend/tests/component/webhook/test_traceback_suppression.py` installed its Prefect traceback filter by calling application startup code:

```python
@pytest.fixture
def configured_logging() -> None:
    configure_logging(production=False, log_level="DEBUG")
```

`configure_logging` owns the process when it runs — it sets the root log level, swaps the root handler and reconfigures structlog — and, being startup code, undoes none of it. Called from a fixture it left the root logger at `DEBUG` for the rest of the xdist worker, overriding the `WARNING` level `pytest_configure` pins. Every test that ran after it in that worker then logged one line per Bolt message from the Neo4j driver.

The enterprise CI job that surfaced it ([infrahub-private run 32120976740](https://github.com/opsmill/infrahub-private/actions/runs/32120976740/job/95673712574)) produced **185,123 of 193,263 log lines** as `neo4j.io` / `neo4j.pool` DEBUG output, starting at the first test of this file and continuing on that worker for the next 37 minutes. The three tests that failed — all `Failed: Timeout >300.0s`, all on `gw1`, the poisoned worker — ran the suite past its 1800s session timeout.

Goal: the fixture leaves logging exactly as it found it, and the rule is written down so the next fixture doesn't repeat it.

Non-goals: no change to production logging behaviour, and no attempt to pin `neo4j*` / `httpcore` levels in `configure_logging` (see below).

Targets `stable`. The three Python files this touches are byte-identical on `stable` and `develop`, `backend/tests/helpers/log.py` is new on both, and the two docs additions land outside the one region where those files differ between the branches — so the change applies unchanged wherever it flows next.

## What changed

- **`backend/tests/helpers/log.py`** (new) — `traceback_suppression()`, a context manager that installs the filter on the Prefect run loggers and removes it again. One implementation shared by the webhook fixture and the regression guard below, so a future fixture that installs the filter inherits the guard.
- **`backend/tests/component/webhook/test_traceback_suppression.py`** — the fixture now wraps that context manager instead of calling `configure_logging`. Renamed `configured_logging` → `traceback_suppression_installed`, since that is what it does now.
- **`backend/infrahub/log.py`** — the filter installation is extracted from `configure_logging` into `install_traceback_suppression_filter()`, so production startup and the test share one implementation and the test still exercises the real registry (`suppress_traceback_in_logs` opt-in) rather than a hand-built set. Pure extraction — same operations, same order, same call site.
- **`backend/tests/unit/test_log.py`** — two guards: `test_traceback_suppression_leaves_logging_state_unchanged` (the install/remove cycle restores the run loggers' filters and the root level) and `test_startup_installs_the_filter_on_the_prefect_run_loggers` (startup wiring is still in place).
- **`dev/guidelines/backend/testing.md`** — new section "Leave process-global state as you found it": what counts as global state, and never call an application startup routine from a test.
- **`.agents/rules/testing-python.md`** — the same rule in short form for coding agents.

What stayed the same: `configure_logging`'s behaviour, the three suppression tests' assertions, and the traceback suppression contract.

## How to review

Start with the test file — the fixture is the fix. Then `log.py` to confirm the extraction is behaviour-preserving.

The judgement call worth scrutiny: the fixture no longer calls `configure_logging`, so that wiring is covered by the startup assertion in `test_log.py` instead. The alternative — keep calling `configure_logging` and unwind it in teardown — is worse: restoring `root.handlers` wholesale re-attaches pytest's own capture handler (`LogCaptureHandler` subclasses `StreamHandler`, so `configure_logging` removes it too), which would leak records into a stale handler for the rest of the session.

## How to test

```bash
uv run pytest backend/tests/unit/test_log.py backend/tests/component/webhook/ -o addopts="" -q   # 18 passed
```

Two committed guards, both hermetic — no dependence on collection order or on which xdist worker takes the file, and both live in the unit suite because neither needs a database or a Prefect flow run:

- `test_traceback_suppression_leaves_logging_state_unchanged` drives the `traceback_suppression()` context manager the fixture wraps: while it is open the Prefect run loggers carry exactly the filter it installed; once it closes their filter lists and the root log level are back to their previous values. Fails if that cycle reverts to `configure_logging` (verified: the assertion reports the extra filter that routine leaves behind).
- `test_startup_installs_the_filter_on_the_prefect_run_loggers` asserts the wiring `configure_logging` performs at import, so dropping its `install_traceback_suppression_filter()` call fails a test (verified: `[] == ['prefect.flow_runs', 'prefect.task_runs']`). Nothing is reconfigured to check it.

Before/after on the original leak, measured with a throwaway probe module collected after the suppression tests in the same worker:

| | root level | traceback filters | `neo4j.io` DEBUG record |
|---|---|---|---|
| before | `DEBUG` | `[4, 4]` | emitted |
| after | `WARNING` | `[1, 1]` | none |

Also run: `backend/tests/unit/log_forwarding` (4 passed), `ruff format --check backend/`, `ruff check backend/tests/`, `mypy` on the three changed test files, `invoke docs.lint` (0 errors).

## Impact & rollout

- **Backward compatibility:** none affected; `configure_logging`'s public behaviour is unchanged.

## Follow-up, not in this PR

`configure_logging` mutes `httpx` explicitly but leaves `neo4j`, `neo4j.io`, `neo4j.pool` and `httpcore` at the root level, which is why a single leaked `DEBUG` was this loud. Worth deciding separately whether those should be pinned. Two smaller things noticed nearby: the handler-removal loop in `configure_logging` mutates `root_logger.handlers` while iterating it (so it can skip a handler), and the enterprise `pyproject.toml`'s `[tool.pytest_env]` omits `INFRAHUB_PRODUCTION = false`, which community sets — that is why the enterprise job renders JSON logs where community renders console.

## Checklist

- [x] Tests added/updated (two regression guards)
- [ ] Changelog entry added (test-only change; no user-visible behaviour)
- [ ] External docs updated (not user-facing)

### DIFF
--- a/backend/tests/component/webhook/test_traceback_suppression.py
+++ b/backend/tests/component/webhook/test_traceback_suppression.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 import logging
+from contextlib import contextmanager
 from typing import TYPE_CHECKING
 
 import httpx
 import pytest
 from prefect import flow, task
 
-from infrahub.log import PREFECT_RUN_LOGGERS, install_traceback_suppression_filter
+from infrahub.log import PREFECT_RUN_LOGGERS, TracebackSuppressionFilter, install_traceback_suppression_filter
 from infrahub.webhook.classifier import (
     EXPECTED_DELIVERY_ERRORS,
     ClassifiedFailure,
@@ -17,7 +18,7 @@ from infrahub.webhook.classifier import (
 )
 
 if TYPE_CHECKING:
-    from collections.abc import Generator
+    from collections.abc import Generator, Iterator, Sequence
 
 CLASSIFIED_MESSAGE = "The target responded with HTTP 404."
 
@@ -49,21 +50,29 @@ async def _send_classifying_in_task() -> None:
     await _classify_transport_failure()
 
 
-@pytest.fixture
-def traceback_suppression_installed() -> Generator[None, None, None]:
+@contextmanager
+def _traceback_suppression() -> Iterator[TracebackSuppressionFilter]:
     """Register the traceback filter on the Prefect run loggers, as production startup does, then remove it.
 
     Only the filter is installed, not the whole of configure_logging: that startup routine also raises
     the root log level, replaces the root handler and reconfigures structlog, none of which these
-    assertions need and none of which it undoes. Called from a fixture it would leak that state into
-    every test that follows in the same worker — overriding the WARNING root level the suite pins in
+    assertions need and none of which it undoes. Called per test it would leak that state into every
+    test that follows in the same worker — overriding the WARNING root level the suite pins in
     pytest_configure, so unrelated tests drown in DEBUG records from the database driver and the HTTP
     client.
     """
     traceback_filter = install_traceback_suppression_filter()
-    yield
-    for prefect_logger_name in PREFECT_RUN_LOGGERS:
-        logging.getLogger(prefect_logger_name).removeFilter(traceback_filter)
+    try:
+        yield traceback_filter
+    finally:
+        for prefect_logger_name in PREFECT_RUN_LOGGERS:
+            logging.getLogger(prefect_logger_name).removeFilter(traceback_filter)
+
+
+@pytest.fixture
+def traceback_suppression_installed() -> Generator[None, None, None]:
+    with _traceback_suppression():
+        yield
 
 
 async def test_classified_failure_logs_no_traceback(
@@ -112,3 +121,27 @@ async def test_unclassified_failure_logs_a_traceback(
     logged_exceptions = [record.exc_info[1] for record in caplog.records if record.exc_info]
     assert [type(exc) for exc in logged_exceptions] == [RuntimeError]
     assert str(logged_exceptions[0]) == "boom"
+
+
+def _run_logger_filters() -> dict[str, Sequence[object]]:
+    # Logger.filters is a union of filter forms; the identity of what is attached is all that matters here.
+    return {name: list(logging.getLogger(name).filters) for name in PREFECT_RUN_LOGGERS}
+
+
+def test_traceback_suppression_leaves_logging_state_unchanged() -> None:
+    """The suppression context must hand logging back exactly as it found it.
+
+    The assertions above cannot tell this apart from calling configure_logging, which installs the same
+    filter — only the process-wide state that routine also changes, and never restores, separates them.
+    """
+    root_logger = logging.getLogger()
+    level_before, filters_before = root_logger.level, _run_logger_filters()
+
+    with _traceback_suppression() as traceback_filter:
+        assert _run_logger_filters() == {
+            name: [*filters_before[name], traceback_filter] for name in PREFECT_RUN_LOGGERS
+        }
+        assert root_logger.level == level_before
+
+    assert _run_logger_filters() == filters_before
+    assert root_logger.level == level_before
--- a/backend/tests/component/webhook/test_traceback_suppression.py
+++ b/backend/tests/component/webhook/test_traceback_suppression.py
@@ -1,14 +1,12 @@
 from __future__ import annotations
 
 import logging
-from contextlib import contextmanager
 from typing import TYPE_CHECKING
 
 import httpx
 import pytest
 from prefect import flow, task
 
-from infrahub.log import PREFECT_RUN_LOGGERS, TracebackSuppressionFilter, install_traceback_suppression_filter
 from infrahub.webhook.classifier import (
     EXPECTED_DELIVERY_ERRORS,
     ClassifiedFailure,
@@ -16,9 +14,10 @@ from infrahub.webhook.classifier import (
     WebhookDeliveryError,
     WebhookFailureClassifier,
 )
+from tests.helpers.log import traceback_suppression
 
 if TYPE_CHECKING:
-    from collections.abc import Generator, Iterator, Sequence
+    from collections.abc import Generator
 
 CLASSIFIED_MESSAGE = "The target responded with HTTP 404."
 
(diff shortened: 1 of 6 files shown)


backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -46,12 +50,25 @@ async def _send_classifying_in_task() -> None:
 
 
 @pytest.fixture
-def configured_logging() -> None:
-    # Register the traceback filter on the Prefect run loggers, as production startup does.
-    configure_logging(production=False, log_level="DEBUG")
-
-
-async def test_classified_failure_logs_no_traceback(configured_logging: None, caplog: pytest.LogCaptureFixture) -> None:
+def traceback_suppression_installed() -> Generator[None, None, None]:
```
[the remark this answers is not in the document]
fatih-acar: Valid point, and it lands on the choice rather than being answered by it: with the old fixture the three assertions pass identically, so nothing committed separated the two.

Guard added in 18801432b. The fixture body is now a `_traceback_suppression()` context manager, and `test_traceback_suppression_leaves_logging_state_unchanged` drives it directly: while it is open the Prefect run loggers carry exactly the filter it installed, and once it closes both their filter lists and the root log level are back to their previous values. Verified it fails when the fixture is reverted to call `configure_logging` (the extra filter that routine leaves behind, and the root level it raises).

Driving the context manager rather than probing from a later test module is deliberate — a separate probe module can be assigned to a different xdist worker under `--dist loadscope` and would then pass without ever observing the state it checks.

I left out the "no neo4j DEBUG records" assertion: in a hermetic test that only restates the suite's ambient `WARNING` root level (pinned in `pytest_configure`), so it would still pass if this fixture leaked again. `root_logger.level == level_before` is the invariant that actually belongs to this fixture.

backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -46,12 +50,25 @@ async def _send_classifying_in_task() -> None:
...
+    """Register the traceback filter on the Prefect run loggers, as production startup does, then remove it.
+
+    Only the filter is installed, not the whole of configure_logging: that startup routine also raises
+    the root log level, replaces the root handler and reconfigures structlog, none of which these
+    assertions need and none of which it undoes. Called from a fixture it would leak that state into
+    every test that follows in the same worker — overriding the WARNING root level the suite pins in
+    pytest_configure, so unrelated tests drown in DEBUG records from the database driver and the HTTP
+    client.
+    """
+    traceback_filter = install_traceback_suppression_filter()
```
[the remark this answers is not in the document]
fatih-acar: Valid — the call site was left uncovered, and that is a real gap.

Covered in d579fb950, but not by calling `configure_logging` from a test: that is the process-wide reconfiguration this PR exists to remove, so re-introducing it here would trade the coverage for the bug.

`test_startup_installs_the_filter_on_the_prefect_run_loggers` in `backend/tests/unit/test_log.py` asserts on what importing `infrahub.log` already did — the module-level `configure_logging` call at log.py:148 *is* the production wiring — and checks both `PREFECT_RUN_LOGGERS` carry a `TracebackSuppressionFilter`. Verified it fails when the `install_traceback_suppression_filter()` call is dropped from `configure_logging` (assertion goes to `[] == ['prefect.flow_runs', 'prefect.task_runs']`), and it mutates nothing.

backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -95,3 +121,27 @@ async def test_unclassified_failure_logs_a_traceback(
...
     assert [type(exc) for exc in logged_exceptions] == [RuntimeError]
     assert str(logged_exceptions[0]) == "boom"
+
+
+def _run_logger_filters() -> dict[str, Sequence[object]]:
+    # Logger.filters is a union of filter forms; the identity of what is attached is all that matters here.
+    return {name: list(logging.getLogger(name).filters) for name in PREFECT_RUN_LOGGERS}
+
+
+def test_traceback_suppression_leaves_logging_state_unchanged() -> None:
```
ajtmccarty: this looks like a unit test to me

backend/tests/component/webhook/test_traceback_suppression.py
```diff
@@ -45,13 +50,34 @@ async def _send_classifying_in_task() -> None:
     await _classify_transport_failure()
 
 
+@contextmanager
+def _traceback_suppression() -> Iterator[TracebackSuppressionFilter]:
+    """Register the traceback filter on the Prefect run loggers, as production startup does, then remove it.
+
+    Only the filter is installed, not the whole of configure_logging: that startup routine also raises
```
ajtmccarty: I don't think docstrings should describe what they are NOT doing. I think you can remove all or almost all of this docstring under the first line

ajtmccarty (approved):

backend/tests/component/webhook/test_traceback_suppression.py
fatih-acar: Agreed — it needs no database, no Prefect flow run and none of the component fixtures, so it was paying component-suite cost for a pure logging-state assertion.

Moved to `backend/tests/unit/test_log.py` in 32323eb2e, next to the other `infrahub.log` tests. The install/remove cycle it drives moved with it, to `backend/tests/helpers/log.py` as `traceback_suppression()`, so the webhook fixture and the guard still share one implementation across the two suites — and any future fixture that installs the filter inherits the guard rather than each suite growing its own copy.

Still bites where it matters, re-verified after the move: revert the cycle to `configure_logging` and the guard fails on the extra filter that routine leaves behind; drop `install_traceback_suppression_filter()` from `configure_logging` and `test_startup_installs_the_filter_on_the_prefect_run_loggers` fails with `[] == ['prefect.flow_runs', 'prefect.task_runs']`.

One thing the move does not carry over, worth stating plainly: the guard now asserts the contract of the shared helper, not of the fixture literally. If someone rewrote the fixture to stop using the helper and call `configure_logging` directly, the guard would keep passing — the written-down rule in `dev/guidelines/backend/testing.md` is what covers that case, not a test.

backend/tests/component/webhook/test_traceback_suppression.py
fatih-acar: Fair — and it applied to two docstrings, not one.

That whole paragraph is gone: the context manager moved to `backend/tests/helpers/log.py` in 32323eb2e keeping only its first line ("Register the traceback filter on the Prefect run loggers, as production startup does, then remove it"), and the fixture that wraps it now needs no docstring at all. The rationale it was carrying — why install the filter alone instead of calling `configure_logging` — belongs in `dev/guidelines/backend/testing.md`, which is where this PR puts it; the guideline now points at the helper as the example.

Applied the same cut in 8ca9a21bb to `test_startup_installs_the_filter_on_the_prefect_run_loggers`, which had the same shape: its second paragraph explained why it doesn't call `configure_logging`. First line kept, since that one says what the test asserts.

backend/tests/component/webhook/test_traceback_suppression.py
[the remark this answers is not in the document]
fatih-acar: Still covered after the rebase onto `stable` and the follow-up review round — the commit SHA moved from d579fb950 to f862fb102, and `test_startup_installs_the_filter_on_the_prefect_run_loggers` now sits alongside the logging-state guard in `backend/tests/unit/test_log.py`. Re-verified that dropping `install_traceback_suppression_filter()` from `configure_logging` fails it (`[] == ['prefect.flow_runs', 'prefect.task_runs']`).

ogenstad (approved):

Issue #2268, opened 2024-02-16
bug: Time travel continuum error: you can't edit the past!


petercrocker:
### Component

Frontend UI

### Current Behavior

If you roll back in time, you shouldn’t be allowed to click the edit button to modify data.

### Expected Behavior

Any time you've moved back in the past in the UI, the edit button should be disabled, and even better if we can give an indicator that the viewed state is a time in the past.

PR #10334, opened 2026-08-19
docs(immutable-history): restructure the hub and add a page for querying historical data

yjouffrault:
## What changed

Immutable History had a concept page but no task page. The only mention of `at` in the published
docs was a bullet reading "GraphQL API: Timestamp parameters for historical queries". The syntax,
the accepted time formats, and the limits were not documented anywhere.

- New page: `immutable-history/query-historical-data.mdx`. Covers how a branch and a time select the
  state, how a historical query is resolved, reading a past state from each interface, comparing two
  timestamps, time formats, and branch history limits.
- Rewritten: `immutable-history/overview.mdx`. Reordered to lead with the capability instead of the
  storage model.
- Changed: `docs/sidebars.ts`. Immutable History is now a category with the overview as its hub, so
  the category label is clickable.

## Why the hub was reordered

The old order was: architecture, then a benefits list, then use cases, then commits, then temporal
queries, then attribute storage, then an Implementation details section that repeated the temporal
queries section.

The new order is: what the capability does and when you need it, what you can do with it, how
historical data differs from the Activity log, how history is preserved, how branches use it, and
where to query it.

Two things were removed. "Risk-free rollbacks" overstated the guarantee, because immutable storage
by itself does not establish an operational rollback path. The Implementation details section
repeated content already covered under temporal queries.

## Source material

The starting point was the [temporal graph blog
post](https://opsmill.com/blog/temporal-graph-infrastructure-data/). Three ideas from it are on
these pages because they help explain how the product behaves:

- Historical state answers a different question than an operations log.
- Relationships carry history, not only attributes.
- A branch builds on the history that already existed at its branch point.

The blog's claim about hundreds of branches with no performance lag is not included, because there
is nothing in the repository to support it.

## Facts verified against the implementation

Every behavioural claim on both pages was read from the source rather than from the old page or the
blog. These were not documented before:

- A date with no time resolves to 12:00 UTC, not midnight
  (`python_sdk/infrahub_sdk/timestamp.py:74`).
- Relative offsets support seconds, minutes and hours only. `7d` raises `TimestampFormatError`
  (`timestamp.py:22-26`).
- `at` applies to reads only. If a query document contains a mutation, Infrahub replaces the
  timestamp with the current time (`backend/infrahub/graphql/app.py:223`).
- Schema is resolved for the requested time. When `at` is earlier than the branch's
  `schema_changed_at`, Infrahub calls `load_schema_from_db(at=at_ts)` and analyzes the query against
  that schema (`app.py:228-231`). A query for an earlier time therefore sees the attributes the
  schema defined then.
- REST has no generic historical object read. `at` is accepted on `/api/query/{query_id}`,
  `/api/artifact/{artifact_id}` and the transformation endpoints
  (`backend/infrahub/api/dependencies.py:77`). The REST tab documents running a stored
  `CoreGraphQLQuery`.
- SDK argument types: `at` takes a `Timestamp` on `all()`, `get()` and `filters()`, and also accepts
  a string on `execute_graphql()`. There is no `DateTime` form. A `datetime` passed to
  `Timestamp.__init__` falls through to the default branch and resolves to the current time, so a
  reader following the wrong type would get current data and believe it was historical.

## Review feedback from @ajtmccarty

All five points were correct. Each was checked against the code before it was written up.

- Deleting a branch removes the data and history recorded on it, and that history cannot be
  recovered. `DeleteBranchEdgesQuery` deletes every edge where `r.branch = $branch_name` and any
  vertex left with no edges (`backend/infrahub/core/query/branch.py:89`). The hub no longer implies
  that history is always kept.
- A branch's origin is always the default branch. `Branch.origin_branch` defaults to `main`
  (`backend/infrahub/core/branch/models.py:33`). The boundary rule and the workaround now both name
  the default branch, so nothing suggests branching from a branch.
- `DiffTree` returns a diff that has already been calculated. `DiffUpdate` calculates one for a
  period (`backend/infrahub/graphql/mutations/diff.py:32`). This is documented next to the `null`
  behaviour, because it explains why a request for an arbitrary period can return `null` the first
  time.
- A branch's Branch view shows its diff and allows a refresh whether or not a Proposed Change
  exists.
- Rebasing moves the timestamps of changes made on a branch up to the rebase time. `rebase_graph`
  updates relationships to start from the rebase point and deletes those that ended before it
  (`backend/infrahub/core/branch/models.py:433`). This is noted in the history limits section, where
  it affects what can be queried.

Aaron also raised schema temporal support twice. It was open in an earlier revision of this
description and is now resolved and documented, with the file reference above.

## Open questions for the reviewer

1. **Immutability and rebase read as a contradiction.** "How historical queries work" says that once
   a version is recorded, a later change does not alter it. "Understand branch history limits" then
   says a rebase moves the timestamps of changes made on a branch. Both statements are correct, but
   they are four sections apart and the reader has to connect them. Should the first sentence name
   rebase as the exception and link down to it, or is the current separation acceptable?

2. **Wording of the schema paragraph.** @ajtmccarty, the mechanism is verified, but please confirm
   the description reads correctly: "If the schema changed after the timestamp you request, Infrahub
   loads the schema as it was then, so the query sees the attributes and relationships that existed
   at that point rather than the current ones."

3. **`DiffTree` filters.** The page documents `kind`, `namespace` and `status`. `DiffTreeQueryFilters`
   also accepts `ids` (`backend/infrahub/graphql/queries/diff/tree.py:746`). Worth adding, or is a
   summary of the common three enough for this page?

4. **`updated_by` is the last account to modify an object, not a per-change record.** The hub says a
   change can be traced to the account that made it, and links to `objects/metadata.mdx`. Reading
   `updated_by` at a past timestamp gives the account behind the change that was current then, which
   is why the claim stands. Confirm that is the intended way to answer "who changed this".

## Out of scope

`tutorials/getting-started/**` is excluded from the Docusaurus build
(`docs/docusaurus.config.ts:83`). That series, including
`tutorials/getting-started/historical-data.mdx`, is not published. The build failed when the new
page linked to it, which is how this surfaced. There is currently no published tutorial for
historical data. Deciding what happens to that series belongs in its own change.

A full voice pass over the rest of `overview.mdx` is also out of scope for this PR.

## Verification

- `cd docs && npm run build` passes with no broken links.
- `npx markdownlint-cli2 "docs/docs/immutable-history/**/*.mdx"` reports 0 issues.
- Vale is not installed locally, so CI will be its first run. Both pages were checked by hand against
  `swap.yml`, `simple-easy.yml`, `oxford-comma.yml` and the sentence-case heading rule.

### DIFF
--- a/dev/specs/docs/immutable-history-temporal-queries.md
+++ b/dev/specs/docs/immutable-history-temporal-queries.md
@@ -1,0 +1,360 @@
+# Immutable History: the plan for documenting historical queries
+
+## What this document is
+
+This document records how the Immutable History documentation was planned and written, which facts
+were checked against the source code, and which questions are still unanswered.
+
+It exists so that a person who did not write these pages can understand why they are shaped the way
+they are, and so that a future change to them does not repeat work that has already been done.
+
+The pages it describes are:
+
+- `docs/docs/immutable-history/overview.mdx` — explains the capability. This page already existed
+  and was rewritten.
+- `docs/docs/immutable-history/query-historical-data.mdx` — explains how to query historical data.
+  This page is new.
+- `docs/sidebars.ts` — changed so that Immutable History became a section in the navigation menu,
+  with the overview page as its front page, instead of a single page.
+
+The work was delivered as pull request 10334: <https://github.com/opsmill/infrahub/pull/10334>
+
+The starting point for the work was an assessment of the company blog post about the temporal graph,
+which is recorded in `temporal-graph-blog-assessment.md` in this same folder. The blog post was used
+only to help decide how to frame the explanation. No sentence from the blog post was copied into the
+documentation.
+
+## Decisions that were taken before writing
+
+**A separate page about the "temporal graph" concept was not created.** The vocabulary used in the
+documentation stays as "immutable history" and "temporal queries". Existing pages were not renamed.
+The term "temporal graph" appears once on the overview page, as a definition, because it is a useful
+name for the storage design. It is not used as a heading or a page title.
+
+**The Python SDK does expose the `at` parameter.** This was confirmed in the code, so the overview
+page's claim that four interfaces support historical queries is correct.
+
+**The blog post's performance claim was excluded.** The blog states that Infrahub supports hundreds
+of branches with no performance penalty. Nothing in the repository supports that claim, so it does
+not appear in the documentation.
+
+**Comparing two arbitrary moments in time is a real capability, and it is in scope.** It works
+through the GraphQL interface. It is documented on the new page.
+
+## Where the page sits in the documentation, and why
+
+The Infrahub documentation groups pages into sections. The section called "Branches and Change
+Control" owns anything about branching, proposing changes, and validating changes. Immutable History
+was already the first entry in that section, as a single page.
+
+The Diátaxis model, which this documentation follows, separates pages into four kinds: explanation,
+how-to guides, tutorials, and reference. Before this work, Immutable History had an explanation page
+and nothing else. The missing kind was the how-to guide, which is what the new page provides.
+
+The pattern used is a front page with pages beneath it. The existing overview page became the front
+page, and the new page is the first page beneath it. This pattern is normally reserved for sections
+that will hold several pages. It was justified here because a second page is plausible; see "What was
+deliberately left out" at the end of this document.
+
+No changes were needed in any other repository. The `at` parameter belongs to the platform, so the
+main Infrahub repository is the correct home for the page. The Python SDK's own documentation may
+eventually want a link to it, but that would be a separate change.
+
+## Facts that were verified in the source code
+
+Everything in this section was read from the code in this repository rather than taken from the
+previous version of the page or from the blog post. Each entry gives the file and line so that it can
+be checked again when the code changes.
+
+### How each interface accepts a timestamp
+
+| Interface | How the timestamp is given | Where this is in the code |
+|---|---|---|
+| Web interface | A time selector next to the branch selector. It shows only an icon until a time is chosen; afterwards a bar displays "Current view time" with the chosen value, and a cross clears it. | `frontend/app/src/entities/navigation/ui/time-selector.tsx`, lines 44, 73, 83 |
+| GraphQL | A parameter named `at` in the address of the endpoint, written as `/graphql/<branch>?at=<time>` | `backend/infrahub/graphql/app.py`, line 206 |
+| REST | A parameter named `at`, described in the code itself as accepting "absolute or relative format" | `backend/infrahub/api/dependencies.py`, line 81 |
+| Python SDK | An argument named `at` on the methods `all()`, `get()`, `filters()` and `execute_graphql()` | `python_sdk/infrahub_sdk/client.py`, lines 277, 883, 1184, 2134 |
+
+### Which time formats are accepted
+
+These are read from `python_sdk/infrahub_sdk/timestamp.py`, lines 52 to 101. Infrahub tries them in
+this order:
+
+- A full date and time in ISO 8601 format, with a time zone or an offset.
+- A date and time with no time zone, which is assumed to be UTC.
+- **A date on its own**, such as `2026-03-09`. This resolves to **twelve o'clock midday UTC** on
+  that day, not midnight. This is stated on the page because it surprises people.
+- **A relative offset**, subtracted from the current time. Only three units are supported: seconds,
+  minutes, and hours. They can be combined, for example `2h30m`. **Days and weeks are not
+  supported**: writing `7d` produces an error. The unit definitions are at lines 22 to 26.
+- Anything else produces an error called `TimestampFormatError`, at line 101.
+
+### How far back a query can reach
+
+This is in `backend/infrahub/branch/query_time_validator.py`, lines 24 to 32.
+
+Only the earliest time is checked. The requested time must not be earlier than the effective
+creation time of the branch. For any branch that is not the default branch, the effective creation
+time is the creation time of the **default** branch, not of the branch itself. This is because
+`origin_branch` always refers to the default branch.
+
+The error message the reader will see is:
+
+```text
+Requested time '<time>' is before branch '<name>' was created at '<created_at>'.
+```
+
+**Nothing checks whether the requested time is in the future.** A future time is accepted, and the
+query returns current data with no indication that this happened. The reason is that the filter used
+to select versions is `from <= $at AND (to IS NULL OR to > $at)`, in
+`backend/infrahub/core/timestamp.py`, lines 23 to 24, and every currently valid version satisfies
+that condition for a future time. The web interface is not affected, because its picker only offers
+times in the past (`time-selector.tsx`, lines 56 to 57). This affects GraphQL, REST, and the SDK.
+See open item 2.
+
+### How comparing two moments in time works
+
+Comparing two moments requires two steps, and this was not understood correctly when the page was
+first written.
+
+First, a request called `DiffUpdate` asks Infrahub to calculate the difference and store it. Its
+input accepts `branch`, which is required, plus optional `name`, `from_time` and `to_time`. There is
+also a separate `wait_until_completion` argument, which makes the request return only once the
+calculation has finished instead of starting a background task. This is in
+`backend/infrahub/graphql/mutations/diff.py`, lines 24 to 36.
+
+**If a custom period of time is requested, a name must also be given.** Without a name, Infrahub
+refuses the request with the message "diff with specified time range requires a name". This is at
+line 59 of the same file. This was not documented anywhere before, and it is the real reason why
+asking for an arbitrary period appeared to return nothing on the first attempt.
+
+Second, the stored result is read using either `DiffTree`, which returns the changed objects, or
+`DiffTreeSummary`, which returns only counts. Their arguments are defined in
+`backend/infrahub/graphql/queries/diff/tree.py`, lines 752 to 777:
+
+- `name`, `branch`, `from_time`, `to_time`
+- `filters`, which accepts `ids`, `status`, `kind` and `namespace`. The `status`, `kind` and
+  `namespace` filters each accept lists named `includes` and `excludes`.
+- `limit` and `offset`, for reading results in pages. These are on `DiffTree` only.
+- `proposed_change_id`, which reads the difference belonging to a Proposed Change instead.
+
+When `from_time` is not given, the comparison starts at the point the branch diverged. When
+`to_time` is not given, it runs to the present. These defaults are at lines 463 to 476.
+
+The base of the comparison is always the default branch, at line 551. See open item 1, because a
+reviewer is not certain this is working as intended.
+
+### What happens to history when a branch is merged, rebased, or deleted
+
+**Merging** records the merged changes on the default branch using the timestamp of the merge, not
+the timestamps the changes originally had on the branch. Every merge operation is given one single
+timestamp, in `backend/infrahub/core/diff/merger/merger.py`, lines 94 to 210, which receives it from
+`backend/infrahub/core/merge/branch_merger.py`, line 197.
+
+The practical consequence is worth stating plainly, and it is now on the overview page. If somebody
+creates an object on a branch on Monday and the branch is merged on Friday, the default branch shows
+that object as created on Friday. The history of the default branch therefore contains the final
+result of the branch, but not the sequence of changes that produced it. Those intermediate versions
+stay readable on the branch itself, for as long as the branch exists.
+
+**Rebasing** moves the timestamps of changes made on a branch forward to the time of the rebase, in
+`backend/infrahub/core/branch/models.py`, line 433. A change recorded on the branch before a rebase
+can no longer be read at its original timestamp.
+
+**Deleting a branch** permanently removes the data and history recorded on that branch. The query
+deletes every edge belonging to the branch, and any vertex left with no edges at all, in
+`backend/infrahub/core/query/branch.py`, line 89. This cannot be recovered.
+
+### How the timestamp behaves in the web interface
+
+**The time picker does not use UTC.** The date picker component receives no time zone setting, and
+the applied time is displayed using a local-time format. This is in
+`frontend/app/src/entities/navigation/ui/time-selector.tsx`, lines 42 to 58 and line 75. The
+previous version of the page told readers to choose a time "in UTC", which was wrong and could put a
+reader in Central European Time one or two hours away from the moment they intended.
+
+**The selected time applies to more than object data, and less than the whole interface.** Thirty-
+seven files read the shared value that holds the selected time, which is called `datetimeAtom`.
+Among them are object data, relationships, IPAM, groups, profiles, permissions, search, the schema
+(`entities/schema/ui/queries/load-schema.query.ts`) and the navigation menu
+(`entities/navigation/ui/queries/get-menu.query.ts`). No file belonging to the diff, Proposed
+Change, task or event screens reads it, which means those screens continue to show current data
+while the header still displays "Current view time".
+
+**Editing is not prevented while a past time is applied.** There is no protection anywhere in
+`frontend/app/src/`. A deletion forwards the historical timestamp
+(`entities/nodes/object/ui/queries/delete-object.mutation.ts`, lines 18 and 26) and the backend then
+silently replaces it with the current time (`backend/infrahub/graphql/app.py`, line 223). No error
+is shown. See open item 3.
+
+**The selected time is part of the page address**, as a parameter named `at`
+(`shared/config/qsp.ts`, line 5), and it is preserved as the reader moves around the interface,
+alongside the branch (`shared/api/rest/fetch.ts`, line 83). This means a historical view can be
+saved as a bookmark or shared as a link.
+
+### The schema is also resolved for the requested time
+
+When the requested time is earlier than the moment the branch's schema last changed, Infrahub loads
+the schema as it was at that time and analyses the query against that older schema. This is in
+`backend/infrahub/graphql/app.py`, lines 228 to 231. A query for an earlier time therefore sees the
+attributes that the schema defined then, rather than the ones it defines now.
+
+### A timestamp applies to reading only
+
+If a query document contains a change as well as a read, Infrahub replaces the requested timestamp
+with the current time, in `backend/infrahub/graphql/app.py`, line 223.
+
+### There is a third way to ask about the past
+
+Besides reading one moment and comparing two moments, there is a third operation: asking which
+events happened between two moments. The event query accepts filters named `since` and `until`,
+along with `branches`, `account__ids`, `event_type` and node identifiers. `since` defaults to one
+hundred and eighty days in the past and `until` defaults to the current time. This is in
+`backend/infrahub/graphql/queries/event.py`, lines 48 and 140 to 143.
+
+This is why the new page opens by naming three ways to ask about the past rather than two. See open
+item 4, because these filters are not documented on any page.
+
+## How a comparable product documents this, and what that changed
+
+The equivalent page from Dolt, a database with similar version-control behaviour, was used as a
+benchmark: <https://www.dolthub.com/docs/sql-reference/version-control/querying-history>
+
+Dolt's page leads with the constraint before it shows any syntax. It states that the unit of a
+snapshot is the commit, then works through each interface, then closes with an explicit list of
+limitations.
+
+This changed one thing in the plan. The section describing constraints was moved **above** the
+per-interface syntax rather than being left as a footnote at the end. The reasoning is that a reader
+whose first historical query fails will have failed for one of two reasons — the branch boundary, or
+an unsupported time unit — and both of those belong before the examples rather than after them.
+
+## What changed during the writing, because the code contradicted the plan
+
+**REST has no general way to read an object at a past time.** The `at` parameter is accepted only on
+`/api/query/{query_id}`, `/api/artifact/{artifact_id}`, and the two transformation endpoints
+`/api/transform/python/{transform_id}` and `/api/transform/jinja2/{transform_id}`. This is in
+`backend/infrahub/api/dependencies.py`, lines 77 to 90, and
+`backend/infrahub/api/transformation.py`, lines 35 and 100. The section describing REST therefore
+explains saving a read as a stored query and running that, instead of describing an ad-hoc object
+read that does not exist.
+
+**`DiffTree` reads a stored result rather than calculating one.** This is described above. The page
+was originally written as though `DiffTree` performed the comparison itself.
+
+**The getting-started tutorials are not published.** The folder
+`docs/docs/tutorials/getting-started/` is excluded from the documentation build, in
+`docs/docusaurus.config.ts`, line 83. This was discovered when the build failed after a link was
+added to the existing tutorial about historical data. That whole tutorial series is unpublished. See
+"What was deliberately left out".
+
+**Two structural changes were made.** The four interfaces became four tabs within one section rather
+than four separate sections, following the pattern already used in `docs/docs/groups/create.mdx`.
+And a use case about testing automation against historical data was left out of the introduction
+entirely, rather than written and marked as unconfirmed, because nothing in the repository supported
+it.
+
+**Both pages are written with one paragraph on one line.** They were originally broken across
+several lines at roughly one hundred characters. A reviewer asked for this to change. It is the
+correct choice, because the line-length rule is switched off in `.markdownlint-cli2.yaml`, and
+neighbouring pages such as `docs/docs/branches/overview.mdx` are written as single long lines.
+
+## Open items
+
+These are grouped by who can decide them. None of these are writing problems. Each needs a decision
+or a confirmation from a person.
+
+### Items that need an engineer
+
+**Open item 1. Is the base of the comparison in `DiffTree` working correctly?**
+
+The page states that the base of a comparison is always the default branch, which is what the code
+does at `backend/infrahub/graphql/queries/diff/tree.py`, line 551. However, a reviewer commented
+that he was not certain this is working as intended. Documentation should not describe behaviour that
+nobody is willing to confirm. Please have an engineer confirm it. If the behaviour is wrong, remove
+that sentence from the page and raise a bug report instead.
+
+**Open item 2. Is it intended that a future timestamp is accepted?**
+
+As described above, only the earliest time is checked. A time in the future is accepted and returns
+current data silently.
+
+The page currently documents this as normal behaviour. That may be the wrong choice. If engineering
+would prefer future timestamps to be rejected in the same way that too-early timestamps are
+rejected, then the paragraph describing it should be deleted from the page, and this should become a
+bug report instead.
+
+### An item that needs a product decision
+
+**Open item 3. Should the page warn about editing in a past view, or should the behaviour be fixed?**
+
+As described above, nothing prevents somebody from editing or deleting data while they are viewing a
+past moment, and the change is applied to current data.
+
+There is history here. GitHub issue 2268 asked for exactly this protection in February 2024 and was
+closed in March of that year, but no protection exists in the current code.
+
+The page currently carries a warning box. That was a deliberate choice: it seemed better to warn
+people than to let them discover the problem by losing data. But warning about a problem is not the
+same as fixing it, and this is a product decision rather than a documentation one. If the behaviour
+is fixed, the warning box should be removed.
+
+### Items that need a decision about scope
+
+**Open item 4. The `since` and `until` filters are not documented anywhere.**
+
+The new page opens by naming three ways to ask about the past. The third is the event query with its
+`since` and `until` filters. The table links to the Activity log page, because that is the closest
+existing page, but that page only describes the filters available in the web interface. It does not
+describe the query parameters, and it does not mention that `since` defaults to one hundred and
+eighty days in the past.
+
+There are two reasonable options: extend the Activity log page as a separate change, or add a short
+section to the new page. This was not decided, because either choice changes the scope of this work.
+
+**Open item 5. A reviewer suggested a diagram.**
+
+On the subject of how merging and deleting affect history, a reviewer wrote that a graphic would
+explain the concepts much better. The written explanation now exists. A diagram would probably
+communicate it more clearly, and would need somebody who produces diagrams for the documentation.
+
+### Statements a reviewer asserted that were not confirmed in the code
+
+Both statements are now on the pages. Both came from a maintainer, so they are very likely correct.
+Neither was confirmed by reading the code, so both should be checked before merging.
+
+**Open item 6.** The page says that after a branch is deleted, the Activity log still records the
+operations that were performed on that branch. Events are stored separately from the graph data, so
+this is plausible, but it was not confirmed that the query which deletes a branch's data leaves the
+events untouched.
+
+**Open item 7.** The page describes the creation time of the default branch as the point at which
+Infrahub was first initialised. That phrasing came from the reviewer. The code only states that the
+limit is the creation timestamp of the default branch. The explanation helps the reader, but it
+should be confirmed.
+
+### An item that was in the original plan and is still open
+
+**Open item 8. Is `updated_by` the right way to answer "who changed this"?**
+
+The overview page says that a change can be traced to the account that made it, and links to
+`docs/docs/objects/metadata.mdx`. Reading `updated_by` at a past timestamp gives the account behind
+the change that was current at that time, which is why the claim stands. This should be confirmed as
+the intended way to answer that question.
+
+## What was deliberately left out
+
+These are not oversights. They were considered and excluded on purpose.
+
+**A second page about comparing two points in time.** This would cover `DiffTree` in depth alongside
+the Proposed Change difference. The current page gives it one worked example. A full treatment needs
+somebody to settle the relationship between comparisons based on timestamps and comparisons based on
+branches, which is a larger question than this work.
+
+**Any link to the getting-started tutorials**, because they are not published. Deciding what happens
+to that tutorial series is separate work.
+
+**The blog post's claim about performance**, because nothing in the repository supports it.
+
+**A full review of the writing style across the rest of the overview page.** Only the sections
+touched by this work were revised.
--- a/dev/specs/docs/immutable-history-temporal-queries.md
+++ b/dev/specs/docs/immutable-history-temporal-queries.md
@@ -264,6 +264,13 @@ neighbouring pages such as `docs/docs/branches/overview.mdx` are written as sing
 These are grouped by who can decide them. None of these are writing problems. Each needs a decision
 or a confirmation from a person.
 
+Each item below states the question and who can answer it. A handover comment on pull request 10334
+covers the same items in more depth. For each one it sets out the available options with their
+trade-offs, how the item could be checked, and what to do with each possible answer. That comment
+stays readable on the pull request after it is merged. Read it alongside this section rather than
+instead of it: this document records why the pages are shaped as they are, and the comment records
+what to do next.
+
 ### Items that need an engineer
 
 **Open item 1. Is the base of the comparison in `DiffTree` working correctly?**
--- a/dev/specs/docs/immutable-history-temporal-queries.md
+++ b/dev/specs/docs/immutable-history-temporal-queries.md
@@ -106,10 +106,11 @@ Requested time '<time>' is before branch '<name>' was created at '<created_at>'.
 
 **Nothing checks whether the requested time is in the future.** A future time is accepted, and the
 query returns current data with no indication that this happened. The reason is that the filter used
-to select versions is `from <= $at AND (to IS NULL OR to > $at)`, in
-`backend/infrahub/core/timestamp.py`, lines 23 to 24, and every currently valid version satisfies
-that condition for a future time. The web interface is not affected, because its picker only offers
-times in the past (`time-selector.tsx`, lines 56 to 57). This affects GraphQL, REST, and the SDK.
+to select versions, in `backend/infrahub/core/timestamp.py` lines 22 to 25, is two clauses joined by
+OR — `(r.from <= $at AND r.to IS NULL)` or `(r.from <= $at AND r.to >= $at)` — and every currently
+valid version satisfies the first of those for a future time. The web interface is not affected,
+because its picker only offers times in the past (`time-selector.tsx`, lines 56 to 57). This affects
+GraphQL, REST, and the SDK.
 See open item 2.
 
 ### How comparing two moments in time works
--- a/docs/docs/immutable-history/overview.mdx
+++ b/docs/docs/immutable-history/overview.mdx
@@ -2,79 +2,95 @@
 title: Immutable history
 ---
 
-# Understanding immutable history in Infrahub
+Infrahub keeps every state your infrastructure data has held. Query the graph as it was at any
+point in the past — yesterday's routes, last week's VLAN definitions, the ACL a rollout replaced —
+using the same queries you run against current data.
 
-At its foundation, Infrahub implements data immutability—a principle where information in the database cannot be deleted or modified in place. Instead, every change creates a new version while preserving all previous states. This approach mirrors version control systems like Git, providing a robust history of all infrastructure changes.
+## Why it matters
 
-This architectural decision provides several critical benefits for infrastructure management:
+Debugging a change means comparing what is true now with what was true before it. A data model
+that stores only current values answers the first half and loses the second: once a rollout has
+overwritten the configuration that worked, the comparison you need is gone.
 
-- **Complete audit trail**: Every modification is permanently recorded with who, what, and when, providing full traceability
-- **Time travel queries**: Access the exact state of your infrastructure at any point in history to understand past configurations
-- **Risk-free rollbacks**: Return to any previous state without data loss when issues are detected
-- **Compliance and forensics**: Meet regulatory requirements with immutable change history for audits
-- **Parallel workflows**: Enable multiple teams to work on infrastructure changes simultaneously using branches
-- **Change verification**: Review proposed changes before committing them to production environments
+That gap turns up in ordinary work:
 
-## Use cases
+- A deployment breaks connectivity, and you need the exact configuration it replaced rather than
+  an approximation of it.
+- Something has drifted from design intent, and the useful question is when it diverged.
+- An audit asks which interfaces changed between two dates.
+- A schema change is coming, and you want to know what it does to data that already exists.
+- An automation behaved correctly in testing and incorrectly in production, and you want to run it
+  against the data as it stood at the time.
 
-Immutable history in Infrahub supports several crucial infrastructure management scenarios:
+Because Infrahub records the state of the graph at every point in time, each of these is a query
+rather than an investigation.
 
-- **Historical analysis**: View how your infrastructure looked at a specific point in time to troubleshoot issues or understand past decisions
-- **Compliance auditing**: Extract all changes performed within a specific time frame for regulatory compliance
-- **Change impact assessment**: Compare infrastructure states before and after significant changes
-- **Security investigation**: Trace unauthorized or unexpected changes to their source
-- **Knowledge preservation**: Understand why specific configuration decisions were made, even as team members change
+## How Infrahub stores history
 
-## Core concepts
+Nothing in the database is modified in place. Updating an attribute adds a new value with its own
+timestamp; the previous value stays where it is, still carrying the timestamps that say when it
+was current. The same applies to relationships, so both the objects and the connections between
+them are timestamped.
 
-### Timestamps and commits
+A graph recorded this way is a **temporal graph**: it holds not only how objects relate, but when
+each of those relationships was valid. Naming a time in a query selects the values that were
+current then, which is what makes "the topology before Tuesday's pipeline run" a question the
+database can answer directly.
 
-Every change in Infrahub is organized into commits, each with an immutable timestamp. These commits capture:
+Two properties follow from storing history this way:
 
-- The specific data that changed
-- Who made the change
-- When the change occurred
+- **Change tracking is per attribute, not per object.** An update stores the value that changed
+  rather than a fresh copy of the whole object, and a diff can say which field moved rather than
+  only that the object was touched.
+- **History cannot be rewritten.** Because values are added rather than replaced, a past state
+  cannot be edited after the fact, and the record of who changed what remains attached to it.
 
-This approach ensures a complete and coherent historical record that can never be rewritten or deleted.
+## Temporal history and change logs
 
-### Temporal queries
+A change log records events: what changed, when, and by whom. Reconstructing a past state from one
+means finding an earlier snapshot and replaying the events on top of it, and the result is only as
+complete as the log.
 
-Infrahub's temporal query system allows you to retrieve data from any point in time. When you query the database, you're not just accessing current data—you're accessing a specific moment in the database's history. By default, queries return the latest state, but you can specify any timestamp to see exactly how your infrastructure looked at that moment.
+Infrahub stores the resulting state at every point in time instead of the operations that produced
+it. Reading the past is therefore a query against data that is already there, and it returns
+objects in the same shape as a query against current data — the same fields, the same filters, the
+same client code.
 
-This capability extends across all interfaces:
+Both are useful for different questions. Use the [activity
+log](../deploy-manage/run-observe/activity-log) to see the sequence of operations someone
+performed; query a past time to see the state those operations produced.
 
-- **Web UI**: Time navigation controls in the interface
-- **GraphQL API**: Timestamp parameters for historical queries
-- **REST API**: Temporal query support for retrieving historical data
-- **Python SDK**: Time-aware methods to access past states
+## Branches are parallel timelines
 
-For the syntax each one uses, the time formats they accept, and how far back a branch can be read, see [Query historical data](./query-historical-data.mdx).
+A branch is a second timeline over the same history. Creating one stores a pointer to the moment
+it diverged rather than a copy of the data, so the branch starts with the full history of its
+origin available to it and records only the values you change on it.
 
-### Temporal queries and change logs
+That is why a query names both a branch and a time: together they identify one state out of every
+state the graph has recorded. A branch created last week can be read as it was on Monday, and the
+default branch can be read as it was before that branch merged into it.
 
-A change log records the operations that were applied. Reading a past state from one means replaying those operations in order, and the result is only as complete as the log. Infrahub stores the resulting state at every point in time, so reading the past is a query against data that is already stored rather than a reconstruction from a record of what happened.
+## Where you can use it
 
-### Multiple timelines at once
+| Surface | Access |
+|---|---|
+| Web interface | The time selector beside the branch selector; every object you open reflects the selected time |
+| GraphQL API | `at` on the endpoint — see [Query historical data](./query-historical-data.mdx) |
+| REST API | `at` on stored queries, artifacts, and transformations |
+| Python SDK | `at` on `all()`, `get()`, `filters()`, and `execute_graphql()` |
 
-Branches share the same immutable baseline and record only the values that diverge from it. Creating a branch stores a pointer to a base timestamp rather than a copy of the data, so a branch starts with the full history of its origin available to it and adds only what you change. This is what allows a query to name both a branch and a time: the two together identify one state out of every state the graph has recorded.
-
-### Attribute-level change tracking
-
-Unlike systems that capture entire object snapshots, Infrahub's immutable history operates at the attribute level. This approach offers several advantages:
-
-- **Storage efficiency**: Only changed values are stored, not entire object copies
-- **Change clarity**: Easier identification of exactly what changed in each commit
-- **Performance optimization**: Faster queries and better scalability with large datasets
-
-## Implementation details
-
-There is a timestamp associated with every change in the database. This timestamp is immutable and cannot be changed or deleted. Every query to the database can be associated with a timestamp, allowing you to see the state of the database at that specific point in time.
+To compare two moments rather than read one, query `DiffTree` with a start and end time. A
+[Proposed Change](../proposed-changes/overview.mdx) does the same comparison scoped to a branch,
+which is the right tool when the two states you care about are a branch and its base.
 
 ## Related topics
 
-- [Branches](../branches/overview.mdx) — how branches isolate changes, and what merging does to the timeline
-- [Proposed Changes](../proposed-changes/overview.mdx) — review and validate a set of changes before they merge
+- [Branches](../branches/overview.mdx) — how branches isolate changes, and what merging does to
+  the timeline
+- [Proposed Changes](../proposed-changes/overview.mdx) — review and validate a set of changes
+  before they merge
 
 ## In this section
 
-- [Query historical data](./query-historical-data.mdx) — Read the graph as it stood at any past time, from the web interface, GraphQL, REST, or the Python SDK
+- [Query historical data](./query-historical-data.mdx) — Read the graph as it stood at a past
+  time from any interface, and compare two moments to see what changed
--- a/docs/docs/immutable-history/overview.mdx
+++ b/docs/docs/immutable-history/overview.mdx
@@ -2,24 +2,17 @@
 title: Immutable history
 ---
 
-Infrahub preserves previous values and relationships as your infrastructure data changes. Earlier
-versions remain available for queries, so you can inspect what existed at a specific time, compare
-changes across a period, and trace how an object or relationship changed.
+Infrahub preserves previous values and relationships as your infrastructure data changes. Earlier versions remain available for queries, so you can inspect what existed at a specific time, compare changes across a period, and trace how an object or relationship changed.
 
-This is useful for troubleshooting incidents, reviewing the effect of a change, answering audit
-questions, and understanding how topology or dependencies evolved. When you specify a timestamp,
-Infrahub returns the values and relationships that were valid at that time.
+This is useful for troubleshooting incidents, reviewing the effect of a change, answering audit questions, and understanding how topology or dependencies evolved. When you specify a timestamp, Infrahub returns the objects, attribute values, relationships, and schema that were valid at that time.
 
 ## What you can do with immutable history
 
-- Query objects and relationships as they existed at a specific point in time.
+- Query objects, attributes, relationships, and the schema as they existed at a specific point in time.
 - Compare two timestamps to identify which objects, attributes, or relationships changed.
-- Trace a change to the account that made it and the time it happened, through [object
-  metadata](../objects/metadata.mdx).
-- Preserve the history available before a branch was created while recording changes made on that
-  branch.
-- Use previous infrastructure data for troubleshooting, audits, security investigations, and
-  post-incident analysis.
+- Trace a change to the account that made it and the time it happened, through [object metadata](../objects/metadata.mdx).
+- Preserve the history available before a branch was created while recording changes made on that branch.
+- Use previous infrastructure data for troubleshooting, audits, security investigations, and post-incident analysis.
 
 For example, you can answer questions such as:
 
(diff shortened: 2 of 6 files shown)


docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
 title: Immutable history
 ---
 
-# Understanding immutable history in Infrahub
+Infrahub keeps every state your infrastructure data has held. Query the graph as it was at any
+point in the past — yesterday's routes, last week's VLAN definitions, the ACL a rollout replaced —
+using the same queries you run against current data.
```
ajtmccarty: this is not exactly true. if there are changes on a branch and that branch is deleted, we hard-delete everything on that branch and it is not recoverable

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-- **Change impact assessment**: Compare infrastructure states before and after significant changes
-- **Security investigation**: Trace unauthorized or unexpected changes to their source
-- **Knowledge preservation**: Understand why specific configuration decisions were made, even as team members change
+## How Infrahub stores history
 
-## Core concepts
+Nothing in the database is modified in place. Updating an attribute adds a new value with its own
+timestamp; the previous value stays where it is, still carrying the timestamps that say when it
+was current. The same applies to relationships, so both the objects and the connections between
+them are timestamped.
```
ajtmccarty: would be good to mention that this also applies to schema. so an historical query to before an attribute was added/removed would include the correct schema for that point in time

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
+Two properties follow from storing history this way:
 
-- The specific data that changed
-- Who made the change
-- When the change occurred
+- **Change tracking is per attribute, not per object.** An update stores the value that changed
+  rather than a fresh copy of the whole object, and a diff can say which field moved rather than
+  only that the object was touched.
+- **History cannot be rewritten.** Because values are added rather than replaced, a past state
+  cannot be edited after the fact, and the record of who changed what remains attached to it.
```
ajtmccarty: there's a slight exception here (maybe too technical for most readers, but not sure). when a branch is rebased all of its timestamps before the rebase are moved up to the rebase time. so if Object A is created on a branch 10 minutes ago and then I rebase that branch right now, Object A's new `create` time will be right now instead of 10 minutes ago

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
 
-This approach ensures a complete and coherent historical record that can never be rewritten or deleted.
+## Temporal history and change logs
 
-### Temporal queries
+A change log records events: what changed, when, and by whom. Reconstructing a past state from one
+means finding an earlier snapshot and replaying the events on top of it, and the result is only as
```
ajtmccarty: I don't think we offer this "replaying the events" an an Infrahub feature and it would probably be difficult to do manually

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
+complete as the log.
 
-Infrahub's temporal query system allows you to retrieve data from any point in time. When you query the database, you're not just accessing current data—you're accessing a specific moment in the database's history. By default, queries return the latest state, but you can specify any timestamp to see exactly how your infrastructure looked at that moment.
+Infrahub stores the resulting state at every point in time instead of the operations that produced
+it. Reading the past is therefore a query against data that is already there, and it returns
+objects in the same shape as a query against current data — the same fields, the same filters, the
```
ajtmccarty: maybe _this_ is the right place to mention that the schema also has temporal support

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-- **Web UI**: Time navigation controls in the interface
-- **GraphQL API**: Timestamp parameters for historical queries
-- **REST API**: Temporal query support for retrieving historical data
-- **Python SDK**: Time-aware methods to access past states
+## Branches are parallel timelines
 
-### Attribute-level change tracking
+A branch is a second timeline over the same history. Creating one stores a pointer to the moment
+it diverged rather than a copy of the data, so the branch starts with the full history of its
+origin available to it and records only the values you change on it.
```
ajtmccarty: the "origin" is always the default branch. we don't want to imply that you can branch off of a branch. maybe someday, but not yet

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
+| Surface | Access |
+|---|---|
+| Web interface | The time selector beside the branch selector; every object you open reflects the selected time |
+| GraphQL API | `at` on the endpoint — see [Query historical data](./query-historical-data.mdx) |
+| REST API | `at` on stored queries, artifacts, and transformations |
+| Python SDK | `at` on `all()`, `get()`, `filters()`, and `execute_graphql()` |
 
-There is a timestamp associated with every change in the database. This timestamp is immutable and cannot be changed or deleted. Every query to the database can be associated with a timestamp, allowing you to see the state of the database at that specific point in time.
+To compare two moments rather than read one, query `DiffTree` with a start and end time. A
+[Proposed Change](../proposed-changes/overview.mdx) does the same comparison scoped to a branch,
```
ajtmccarty: you can also view (and refresh) the diff for a branch in its Branch view page even if a proposed change does not exist for the branch

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-## Implementation details
+| Surface | Access |
+|---|---|
+| Web interface | The time selector beside the branch selector; every object you open reflects the selected time |
+| GraphQL API | `at` on the endpoint — see [Query historical data](./query-historical-data.mdx) |
+| REST API | `at` on stored queries, artifacts, and transformations |
+| Python SDK | `at` on `all()`, `get()`, `filters()`, and `execute_graphql()` |
 
-There is a timestamp associated with every change in the database. This timestamp is immutable and cannot be changed or deleted. Every query to the database can be associated with a timestamp, allowing you to see the state of the database at that specific point in time.
+To compare two moments rather than read one, query `DiffTree` with a start and end time. A
```
ajtmccarty: probably deserves its own topic if it does not have one yet, but you need to send a `DiffUpdate` request to calculate the diff for a given time before you can retrieve it via the API with `DiffTree`. the UI only allows accessing the latest diffs for existing branches.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,211 @@
...
+```
+
+`Timestamp` accepts every form in the table above, so `Timestamp("6h")` reads six hours back.
+Pointing a script at a past state is also how you test an automation against the data it ran on,
+rather than against data that has moved on since.
+
+</TabItem>
+</Tabs>
+
+## Compare two points in time
```
ajtmccarty: _this_ is probably where the `DiffUpdate` before `DiffTree` explanation belongs

ajtmccarty (approved):

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
 title: Immutable history
 ---
 
-# Understanding immutable history in Infrahub
+Infrahub preserves previous values and relationships as your infrastructure data changes. Earlier
+versions remain available for queries, so you can inspect what existed at a specific time, compare
+changes across a period, and trace how an object or relationship changed.
```
BaptisteGi: We should make sure Claude isn’t adding line breaks where they’re not needed - here  for instance it should be a single line

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
 
-At its foundation, Infrahub implements data immutability—a principle where information in the database cannot be deleted or modified in place. Instead, every change creates a new version while preserving all previous states. This approach mirrors version control systems like Git, providing a robust history of all infrastructure changes.
+This is useful for troubleshooting incidents, reviewing the effect of a change, answering audit
+questions, and understanding how topology or dependencies evolved. When you specify a timestamp,
+Infrahub returns the values and relationships that were valid at that time.
```
wvandeun: Infrahub returns the schema, object, attributes and relationships that were valid at that time.

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-This architectural decision provides several critical benefits for infrastructure management:
+## What you can do with immutable history
 
-- **Complete audit trail**: Every modification is permanently recorded with who, what, and when, providing full traceability
-- **Time travel queries**: Access the exact state of your infrastructure at any point in history to understand past configurations
-- **Risk-free rollbacks**: Return to any previous state without data loss when issues are detected
-- **Compliance and forensics**: Meet regulatory requirements with immutable change history for audits
-- **Parallel workflows**: Enable multiple teams to work on infrastructure changes simultaneously using branches
-- **Change verification**: Review proposed changes before committing them to production environments
+- Query objects and relationships as they existed at a specific point in time.
```
wvandeun: objects, attributes, relationships, schemas

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
 
-- **Historical analysis**: View how your infrastructure looked at a specific point in time to troubleshoot issues or understand past decisions
-- **Compliance auditing**: Extract all changes performed within a specific time frame for regulatory compliance
-- **Change impact assessment**: Compare infrastructure states before and after significant changes
-- **Security investigation**: Trace unauthorized or unexpected changes to their source
-- **Knowledge preservation**: Understand why specific configuration decisions were made, even as team members change
+## Historical data and the Activity log answer different questions
 
-## Core concepts
+The Activity log records operations: which objects were affected, when a change occurred, who made
```
wvandeun: I think this would be better:
which object was changed, when that change occurred and who made it; and the sequence of actions.

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-- When the change occurred
+## How Infrahub preserves history
 
-This approach ensures a complete and coherent historical record that can never be rewritten or deleted.
+Each change in Infrahub creates a new version instead of modifying the previous value in place.
+Every version is associated with a timestamp, and earlier values remain available for queries that
+specify an earlier time.
 
-### Temporal queries
+History is tracked at the attribute level. When one field changes, Infrahub records the new value
```
wvandeun: To me this reads a bit weird.

History is tracked at the object, attribute, relationship and schema level. We can then explain how we track value, related object changes.

It's also a bit weird how we explain the exception about branches.
Maybe to not mention that here as we have a section on it below.

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
+Deleting a branch is the exception. Infrahub removes the data and history recorded on that branch,
+and that history is not recoverable. Versions on the default branch are unaffected.
 
-Unlike systems that capture entire object snapshots, Infrahub's immutable history operates at the attribute level. This approach offers several advantages:
+## How branches use immutable history
 
-- **Storage efficiency**: Only changed values are stored, not entire object copies
-- **Change clarity**: Easier identification of exactly what changed in each commit
-- **Performance optimization**: Faster queries and better scalability with large datasets
+When you create a branch, it starts from the data and history available on the default branch at its
```
wvandeun: I think here we need to do a better explanation.

The history within a branch is available as long as that branch exists. If we delete the branch then that history will be removed permanently.

When a branch is merged that creates a new object:
- the creation date of that object in the default branch will be the time of the merge

When a branch is merged that modifies an objectL
- the modification date in the default branch will become the time of the merge

This means that the history in the default branch will not contain the changes that happened in the merged branch. The history within that branch for a given object within that branch would still be available for as long as that branch exists. Once you delete the branch that historical data will be gone from the branch. However the activity log of that object will still show these changes.

wvandeun: Maybe a graphic to explain these concepts would come a long way to explain this better

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,95 @@
...
-## Implementation details
+You can therefore develop and review multiple changes independently while retaining the history
+needed to compare and merge them. See [Branches](../branches/overview.mdx) for the branch creation,
+diff, and merge workflow.
 
-There is a timestamp associated with every change in the database. This timestamp is immutable and cannot be changed or deleted. Every query to the database can be associated with a timestamp, allowing you to see the state of the database at that specific point in time.
+## Query data at a specific time
 
-## Related topics
+By default, queries return the latest data on the selected branch. Set a time when you need the
```
wvandeun: Maybe change this to: You can a set a time when you need to the state of the data in Infrahub at an earlier time.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
+---
+title: Query historical data
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+You can query infrastructure data at previous points in time and compare how it changed between
```
wvandeun: and compare how the data changed between them.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+them. Use queries with a timestamp to investigate incidents, verify changes during a period, answer
+audit questions, and inspect how objects and relationships differed at an earlier time.
+
+To query a specific point, specify a branch and timestamp. Infrahub returns the values and
```
wvandeun: To query data at a specific point in time, ...

wvandeun: Infrahub returns the object, attribute values and related object that were valid at that time. (I would drop the part about the interfaces you use for current data)

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+relationships that were valid at that time through the same interfaces you use for current data. To
```
wvandeun: `To compare changes, specify two timestamps`. This is not true, AFAIK. To see the changes between 2 timestamps we actually have to use a dedicated queries.

Also there is a way to get the activity log between 2 given timestamps, to retrieve the audit log.

So I think we should explain here that we have 3 ways to query historical data:
1. Query the state of the of the database at a given timestamp in a specified branch (using the at and branch query param)
2. The changes that happened between 2 timestamps in a given branch
3. The auditrail of what happened between 2 given timestamps in a given branch

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+compare changes, specify two timestamps.
+
+## Specify a branch and a time
```
wvandeun, line 20: This is only relevant in case 1 above

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+is that you specify the time to evaluate the data.
+
+## How historical queries work
+
+When you add a timestamp to a query, Infrahub evaluates each requested object using the attribute
+values and relationships that were valid at that time. You use the same query structure as you do
+for current data; the timestamp changes which versions Infrahub returns.
+
+Infrahub can return those versions because changes create new attribute values and relationship
+versions instead of replacing the previous ones. History is tracked per attribute and relationship,
```
wvandeun: Here as well we should talk about history being tracked per object / attribute and relationship

wvandeun: Actually it would be good to point at the explanation you do in the overview page here

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+so a comparison can identify the specific fields or connections that changed, including the previous
```
wvandeun: this part is confusing.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+values.
+
+The schema is evaluated for the same point in time. If the schema changed after the timestamp you
+request, Infrahub loads the schema as it was then, so the query sees the attributes and relationships
```
wvandeun: so the query sees the object, attributes and relationship that existed...

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+that existed at that point rather than the current ones.
+
+A timestamp applies to reads only. If a query document contains a mutation, Infrahub ignores the
```
wvandeun: applies to read operations (query in GraphQL terms) only

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+timestamp and applies the mutation at the current time.
+
+Once Infrahub records a version, a change made after it does not alter that version. See [Immutable
+history](./overview.mdx) for more detail on how Infrahub preserves those versions and how immutable
+history relates to branches and the Activity log.
+
+## Query data at a specific time
+
+Use a timestamp when you need the values and relationships that existed at a known point — for
```
wvandeun: when you need the objects, attributes and relationships that existed at ..

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+the values valid at that timestamp. Select the × beside the displayed time to return to the current
+time.
+
+![The time selector, with a past time applied](../media/tutorial_2_historical.png)
+
+</TabItem>
+
+  <TabItem value="graphql" label="GraphQL">
+
+Apply the historical time to the same GraphQL query you use for current data. Use this when you can
```
wvandeun: this is confusing

I would just say: Use the URL at parameter to specify a desired timestamp in a GraphQL query.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+already state the data you need and want a repeatable query.
+
+`at` is a query-string parameter on the endpoint, so the query document itself is unchanged:
```
wvandeun: at is in fact a URL parameter

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+```graphql # Read a device as it existed at an earlier time
+# Endpoint : http://localhost:8000/graphql/main?at=2026-03-09T14:00:00Z
```
wvandeun: The endpoint here is commented out  and makes it hard to read in the docs. And in this case the endpoint containing the at URL parameter is exactly the IMPORTANT part.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+    }
+  }
+}
+```
+
+</TabItem>
+
+  <TabItem value="rest" label="REST API">
+
+Use the REST API when an external tool, scheduled process, or integration needs data for a specific
```
wvandeun: We don't need to explain when to use the REST API. We just only offer certain functionality over the REST API, that is not avialable over the GraphQL API. I think it would be better if we just specify which REST API endpoints support this and list them here.

wvandeun: IE:
stored graphql queries, artifacts, transfoms...

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+    at=Timestamp("2026-03-09T14:00:00Z"),
+)
+```
+
+</TabItem>
+</Tabs>
+
+## Compare changes between two timestamps
+
+Use a diff when you need to identify which objects or attributes changed between two timestamps.
```
wvandeun: This section will need to be redone. It is not accurate and clear.

We first have to send a mutation to calculate the differences between 2 timestamps (DiffUpdate).
Once the calculation is done, the diff can be retrieved using the `DiffTree` or `DiffTreeSummary` queries.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+  over time.
+- If no diff covering the requested period is available, the query returns `null` rather than an
+  empty result. Treat `null` as "not calculated," not "nothing changed."
+
+`DiffTree` retrieves a diff that has already been calculated. To calculate one for a period, send
+the `DiffUpdate` mutation first with the `branch` and, when you need a period other than the
+default, `from_time` and `to_time`. This is why a request for an arbitrary period can return `null`
+on the first attempt.
+
+In the web interface, a branch's Branch view shows its diff and lets you refresh it, whether or not
```
wvandeun: We can also consult  a diff for a given proposed change, here we seem to imply we can only do it from the branch page.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+On the default branch and the global branch, the earliest available time is that branch's creation.
+On any other branch, the boundary is the creation time of the default branch.
+
+```text
+Requested time '2026-01-05T00:00:00Z' is before branch 'main' was created at '2026-02-01T09:14:22.481000Z'.
+```
+
+If you need data from an earlier timestamp than the branch allows, query the default branch instead.
+
+Rebasing a branch moves the timestamps of changes made on it up to the rebase time, so a change
```
wvandeun: this should be part of the overview, not in how we query historical data

wvandeun: it should probably be together with how we explain how merges affect the history

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+Use a relative offset when you are investigating from the current time. Use an absolute timestamp
+when the query needs to resolve to the same time each time it runs — for example, for an audit
+answer, post-incident report, or reproducible analysis.
+
+| Form | Example | Resolves to |
+|---|---|---|
+| ISO 8601 with a zone or offset | `2026-03-09T14:00:00Z` or `2026-03-09T15:00:00+01:00` | The specified instant |
+| ISO 8601 without a zone | `2026-03-09T14:00:00` | The same wall-clock time, interpreted as UTC |
+| Date only | `2026-03-09` | 12:00 UTC on that day |
+| Offset from now | `30s`, `45m`, `6h`, `2h30m` | That interval before the current time |
```
wvandeun, line 198: Not sure if this is correctly working atm, but that is a different thing

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+## Understand branch history limits
+
+You can query a branch back to the point where its history begins. If you request a time earlier
+than that, Infrahub rejects the query rather than returning partial data.
+
+When you create a branch, Infrahub records where it diverged rather than copying the entire dataset.
+You can query the history available through the default branch, plus the changes recorded on the
+branch after it diverged.
+
+On the default branch and the global branch, the earliest available time is that branch's creation.
```
wvandeun: The default and global branch are the same thing. 

The earliest time you can query in the default branch is the time that you started the application for the first time. I think it will be clearer that way

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+A date without a time resolves to midday rather than midnight, so include an explicit time when a
+day boundary matters. Relative offsets support seconds, minutes, and hours, including combined
+values such as `2h30m`. They do not support day or week units; use an absolute timestamp for those
+intervals.
+
+## Understand branch history limits
+
+You can query a branch back to the point where its history begins. If you request a time earlier
+than that, Infrahub rejects the query rather than returning partial data.
```
wvandeun: This contradicts what we state below. The only exception here is the default branch

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+branch after it diverged.
+
+On the default branch and the global branch, the earliest available time is that branch's creation.
+On any other branch, the boundary is the creation time of the default branch.
+
+```text
+Requested time '2026-01-05T00:00:00Z' is before branch 'main' was created at '2026-02-01T09:14:22.481000Z'.
+```
+
+If you need data from an earlier timestamp than the branch allows, query the default branch instead.
```
wvandeun: This is not needed, because the history from before the branch was created is accessible.

wvandeun (changes requested):

docs/docs/immutable-history/query-historical-data.mdx
wvandeun, line 1: We should probably mention how to query the activity log for audit trail here as well

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+example, during an incident or before a change.
+
+You can specify a time when viewing or querying Infrahub data through the web interface, GraphQL,
+REST API, or Python SDK. Use the web interface for interactive investigation and the APIs or SDK
+when you need a repeatable or programmatic query.
+
+<Tabs groupId="method" queryString>
+  <TabItem value="web" label="Web interface" default>
+
+Select the time selector — the calendar and clock icon beside the branch selector — and choose a
```
lancamat1: the picker is not UTC. `DateTimePicker` in `frontend/app/src/entities/navigation/ui/time-selector.tsx:53-57` gets no timezone prop, so the calendar and the time list are both in the browser's local zone. `toISOString()` converts to UTC on the wire, and the applied time shown next to the icon goes through `useFormatDate`, which honours a per-user display preference that can carry its own timezone (`frontend/app/src/shared/context/date-preferences-context.tsx`).

So nothing the reader actually sees is guaranteed to be UTC. Someone in CET following this literally picks a time one or two hours off and gets the wrong state back, with nothing to signal it.

I would drop "in UTC" and say the picker uses your browser's time zone. The same wording is in `docs/docs/tutorials/getting-started/historical-data.mdx`, though that file is excluded from the build (`docs/docusaurus.config.ts:84`) so it is not published.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+date and time in UTC. Until you set a time, the selector displays only the icon; once you set one,
+the bar beside it displays **Current view time** with your selection.
+
+The selected time remains applied as you navigate, so objects and relationships are displayed using
```
lancamat1: this is right about the time staying applied, but it does not say where it stops applying. diff, proposed-changes, tasks and events have no time-aware query at all, so those screens stay current while the header still shows **Current view time**. someone who opens a diff in a past view and trusts it is being misled by this paragraph

worth adding in the other direction as well: the schema and the sidebar menu are also loaded at the selected time (`entities/schema/ui/queries/load-schema.query.ts:33`, `entities/navigation/ui/queries/get-menu.query.ts:19`), so the navigation tree and an object's fields can differ from what the current schema defines

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,239 @@
...
+the values valid at that timestamp. Select the × beside the displayed time to return to the current
+time.
```
lancamat1: one thing missing that I would add: editing is not blocked while a past time is applied. there is no read-only guard anywhere in `frontend/app/src/`, and a mutation carries the historical `at`, which the backend then resets to now (`backend/infrahub/graphql/app.py:223-224`) without an error. `delete-object.mutation.ts:18,26` forwards it, update and create do not

so a delete performed while you think you are in the past deletes live data with no warning. #2268 asked for exactly this guard in Feb 2024 and was closed in March, but nothing survives in the current tree

I do not love documenting this instead of fixing it, but it will not be fixed before these docs ship, so one sentence seems better than letting people find out. something like: editing is not blocked while a past time is applied, and any change you save applies to the current time rather than the time you are viewing

lancamat1: also worth one sentence: the selected time is in the page URL as `at` (`shared/config/qsp.ts:5`), hydrated from the URL on mount and preserved across in-app navigation alongside `branch` (`shared/api/rest/fetch.ts:83`). so a historical view is a bookmarkable and shareable link, which seems like the most useful property of the UI implementation and it is not mentioned anywhere

yjouffrault:
## Handover notes: what this pull request does, and the seven decisions it still needs

This comment is written for someone who has not worked on this pull request before. It explains what
the pull request does, what has already been done, and what still needs a decision.

### What this pull request does

Infrahub can answer questions about the past. You can ask what a device looked like yesterday, what
changed between two moments in time, and who performed which operation.

Before this pull request, the documentation explained this idea but never showed anyone how to do it.
The `at` parameter appeared only once in the published documentation, in a single bullet point on an
unrelated tutorial page. The syntax, the accepted time formats, and the limits were not written down
anywhere.

This pull request therefore does three things:

1. It adds a new page, `immutable-history/query-historical-data.mdx`, which explains how to query
   historical data.
2. It rewrites the existing page, `immutable-history/overview.mdx`, so that it explains the
   capability before it explains the storage design.
3. It changes the sidebar so that Immutable History is a section with the overview page at its front,
   instead of a single page.

### What has already been done

Four reviewers read the pull request and left forty-two comments. All of those comments have now been
addressed. The pages changed a great deal as a result: the introduction, the section about
comparing two moments, the section about how far back you can query, and the section about branch
behaviour were all rewritten rather than only corrected.

What remains is not a writing task. Seven questions need a person who knows the product to decide
them. Those seven questions are the reason this work is being handed over, and they are listed at the
end of this comment.

### What the two pages now cover

The new page explains that there are three different ways to ask about the past. These are genuinely
different operations, and the earlier version of the page did not separate them clearly:

1. Reading the state of the data at a single moment in time, using `at` and `branch`.
2. Finding which objects and attributes changed between two moments, using `DiffUpdate` and then
   `DiffTree`.
3. Finding which operations happened between two moments and who performed them, using the `since`
   and `until` filters on an event query.

The page then covers each interface separately — the web interface, GraphQL, REST, and the Python
SDK — followed by the accepted time formats and how far into the past a query can go.

The overview page explains what immutable history is, how it differs from the Activity log, how
history is recorded at the level of objects, attributes, relationships and the schema, and what
happens to that history when a branch is merged, rebased, or deleted.

### Three things the documentation used to say that were wrong

These are worth knowing beyond this pull request, because they are things we may have been telling
customers.

**The time selector in the web interface does not use UTC.** The page told readers to choose a date
and time "in UTC". In fact it uses the time zone of the reader's own browser. A reader in Central
European Time who followed that instruction would have seen data from a moment one or two hours
different from the moment they intended, and nothing on the screen would have told them.

**You cannot compare two moments by putting two timestamps on a single read request.** The page
suggested that you could. It actually requires two steps. First you send a request called
`DiffUpdate`, which asks Infrahub to calculate the difference and store it. Only then can you read
the result, using `DiffTree` or `DiffTreeSummary`.

There is also a requirement that was documented nowhere. If you ask for a custom period of time, you
must also give the calculation a name. If you do not, Infrahub refuses the request. This is the real
reason why asking for an arbitrary period seemed to return nothing the first time, which people had
reported as confusing.

**The rule about how far back you can query contradicted itself.** One paragraph said that each
branch can be read back to the point where its own history begins. A later paragraph said that the
limit is the creation time of the default branch. The second one is correct. Every branch reads its
history through the default branch, so the earliest moment you can query is always the creation time
of the default branch.

### The most important thing that was missing

Neither page said anything about what merging does to history, and the behaviour surprises people.

When a branch is merged, the merged changes are recorded on the default branch with the timestamp of
the **merge**, not with the timestamps they had on the branch. If somebody creates an object on a
branch on Monday, and that branch is merged on Friday, then the default branch shows that object as
created on Friday.

This is useful to understand for customer conversations. The history of the default branch contains
the final result of a branch, but not the sequence of changes that produced that result. Those
intermediate versions stay readable on the branch itself, but only while the branch still exists. If
the branch is deleted, they are removed permanently and cannot be recovered.

The overview page now explains merging, rebasing and deleting together, because all three change
what a historical query is able to return.

### The seven decisions that are needed

None of these are writing problems, and none of them stop you reading the pages. Each one is a
question about the product. Each entry below explains the issue, gives you the options, says who to
ask, and says what to do depending on the answer.

---

#### 1. Is the base of a `DiffTree` comparison behaving correctly?

**The issue.** The page tells the reader that the base of a comparison is always the default branch.
That is what the code does today (`backend/infrahub/graphql/queries/diff/tree.py`, line 551). One
reviewer commented that he was not certain this is working as intended.

**Why it matters.** The page tells a reader that if they name a feature branch and a time range, they
will get that branch compared against the default branch across that period. If the base is not
actually selected that way, we are describing a defect as though it were a feature, and readers will
trust comparison results that are wrong.

**Who to ask.** Aaron McCarty or Wim Van Deun. Wim raised the doubt, and Aaron has worked in
the diff code.

**How you could check it yourself.** Create a branch, change one object on it, then run `DiffUpdate`
followed by `DiffTree`, naming the branch and a time range that begins before the branch existed.
Look at which branch and which moment the returned difference actually used as its base.

**What to do with each answer:**

- If it is confirmed correct, reply on the review conversation and mark it resolved. No page change.
- If the base is the default branch but only at the branch point, rather than at the `from_time` you
  asked for, then the sentence needs qualifying. Say which one it is and I or someone else can adjust
  the wording.
- If it is not working correctly, remove that sentence from the page, raise a bug report, and add a
  short note saying the comparison base is not yet reliable for arbitrary time ranges.

---

#### 2. Should a timestamp in the future be rejected?

**The issue.** Infrahub only checks that the requested time is not too *early*. It never checks
whether the time is in the future. Ask for a moment next week and Infrahub accepts the request and
returns today's data, with nothing to indicate that this happened
(`backend/infrahub/branch/query_time_validator.py`, line 32).

**Why it matters.** A person typing a date by hand will rarely do this. A script will. If a script
builds the timestamp from a calculation, and gets the time zone offset backwards or runs on a machine
with a wrong clock, it receives current data that looks exactly like historical data. A silent wrong
answer is worse than an error message, because nobody investigates it.

**Who to ask.** Engineering, specifically whoever owns `BranchQueryTimeValidator`.

**What to do with each answer:**

- If this is intended behaviour, keep the paragraph as it is and mark the item closed.
- If future timestamps should be rejected in the same way as too-early ones, raise a bug report
  referencing that file and line. Keep the paragraph in the page for now, because it describes what
  Infrahub does today, and remove it in the pull request that adds the validation.
- If nobody wants to decide yet, keep the paragraph. Describing real behaviour accurately is correct
  even when that behaviour may change later.

**My suggestion.** Keep the paragraph either way. It should only come out once the validation
actually changes.

---

#### 3. Nothing prevents editing while you are viewing a past moment

**The issue.** There is no protection anywhere in the web interface. If you delete an object while
you believe you are looking at last Tuesday, you delete the current, live data. The historical
timestamp is sent with the request, the backend silently replaces it with the current time, and no
error or warning appears.

GitHub issue #2268 asked for exactly this protection in February 2024. It was closed in March of that
year, and no protection exists in the code today.

**Why it matters.** This is the most serious item on this list. Everything else on this page is about
a reader being confused. This one is about a reader losing production data while believing they are
only looking at it.

**Who to ask.** The product owner for the web application, together with the frontend lead. Point
them at issue #2268 so they can see the earlier discussion.

**What to do with each answer:**

- If the protection is going to be built soon, keep the warning box in the page, and remove it in the
  pull request that ships the protection.
- If it is not going to be built, keep the warning box and consider changing it from a warning to a
  stronger `:::danger` admonition, because the consequence is data loss rather than confusion.
- If anyone argues that this should not be documented at all, please escalate rather than quietly
  removing it. Saying nothing means readers discover the behaviour by losing data.

**My suggestion.** Keep the warning and raise the missing protection as a separate bug. Documenting
a hazard and fixing it are not alternatives; we should do both.

---

#### 4. The `since` and `until` filters are documented nowhere

**The issue.** The new page opens with a table naming three ways to ask about the past. The first two
rows link to sections further down the same page. The third row is the event query, and it links to
the Activity log page. That page only documents the filters available in the web interface. It does
not document the `since` and `until` query parameters, and it does not mention that `since` defaults
to one hundred and eighty days in the past. So a reader who wants the third option clicks the link
and does not find what the table promised.

**Why it matters.** The three-way framing came from Wim's review and it is a genuine
improvement, because these really are three different operations, and the earlier version of the page did not separate them clearly. But it now makes a promise the documentation does not keep.

**The complication.** These two pages sit in different top-level sections of the documentation.
`immutable-history` is under **Branches & Change Control**. The Activity log is under **Deploy &
Manage, Run & observe**. So this is a question about a boundary between two sections, not a question
about one page.

**Your options:**

- **Option A. Extend the Activity log page** with the query parameters, in its own pull request. This
  is the natural home, because event filtering belongs with events, and anyone looking for event
  filtering will go to that page rather than to a page about immutable history. The drawback is that
  the gap stays open until that second pull request lands.
- **Option B. Add a short section to the new page.** This closes the gap immediately and keeps
  everything in one pull request. The drawback is that it documents the event query inside a page
  about immutable history, so the same parameters end up described in two places over time, and the
  two descriptions will eventually disagree.
- **Option C. Remove the third row from the table** and do not describe the third option in this pull
  request. No gap and no extra work, but the page then describes less than the product can do, and it
  partly undoes the improvement Wim asked for.

**Who to ask.** Wim, because the three-way framing was his point, and whoever owns the Deploy &
Manage section of the documentation.

**My suggestion.** Option A, and leave the third row where it is. The link is not wrong, it is
incomplete, and a reader still lands on the right page for the topic. Raise an issue for the Activity
log page so the gap is tracked rather than forgotten. Choose Option B instead if you need this pull
request to be self-contained, for example if the Activity log page has no owner right now.

---

#### 5. A reviewer asked for a diagram

**The issue.** Wim wrote that a graphic would explain the merge and delete behaviour much better
than prose. The written explanation now exists and is correct, but his request is reasonable.

**Why prose struggles here.** The behaviour has two dimensions at once, time and branch. The same
change has one timestamp when you query the branch and a different timestamp when you query the
default branch. Prose can only describe that one step at a time, whereas a picture can show both at
once.

**What a diagram would need to show.** A time axis, a branch created from the default branch, a change
made on the branch on Monday, the merge on Friday, and the same change appearing at two different
moments depending on which branch you query. Then what disappears if the branch is deleted.

**What already exists that a diagram could reuse.** The `docs/media/` folder already contains Excalidraw
diagrams for branch behaviour: `branch_merge.excalidraw.svg`, `branch_rebase_no_conflict.excalidraw.svg`
and `branch_conflict.excalidraw.svg`. So there is an established visual style and an editable source
format for exactly this subject area. Separately, Mermaid is installed and enabled in the
documentation site, but it is currently used on only one page (`overview/architecture.mdx`).

**Your options:**

- **Option A. Extend the existing `branch_merge.excalidraw.svg`** with a time axis. This matches the
  style of the diagrams already used for branches, and the source stays editable. You need the
  Excalidraw source file and somebody comfortable editing it.
- **Option B. Draw a new Mermaid diagram inline in the page.** No image file to maintain, anyone can
  edit it in a pull request, and it adapts to light and dark themes automatically. The drawback is that Mermaid is used on only one page in this documentation today, and its diagram types may not express
  "the same change at two different timestamps" clearly.
- **Option C. Ship this pull request without a diagram** and raise it as a follow-up.

**Does this block the merge?** No. The prose is correct and complete. A diagram is an improvement, not
a correction.

**Who to ask.** Whoever produces diagrams for the documentation. It is also worth asking Wim
whether the written explanation is now sufficient for him, because he may be satisfied.

**My suggestion.** Option C, then Option A as a follow-up. Do not hold this pull request for a
diagram, but do raise it, because the existing branch diagrams make this the natural place for one and
it would be a genuine improvement.

---

#### 6. Does the Activity log survive a branch being deleted?

**The issue.** The overview page states that after a branch is deleted, the Activity log still records
the operations that were performed on that branch. This came from Wim. It was never checked
against the code.

**Why it matters.** This sentence is the reassurance we offer a reader who has just been told that
deleting a branch destroys its history permanently. If the sentence is wrong, we are offering false
comfort on a path that loses data.

**How to check it.** The query that removes a branch deletes every edge belonging to that branch, and
any vertex left with no edges at all (`backend/infrahub/core/query/branch.py`, line 89). Events are
stored separately from the graph data, through the task manager, so they are probably untouched. What
needs confirming is whether anything about event records depends on the branch data that was removed.

**Who to ask.** Aaron McCarty, or whoever owns the event store.

**What to do with each answer:**

- If the events do survive, reply on the conversation and mark it resolved. No page change.
- If the events are also removed, this is significant. Remove that sentence and strengthen the
  surrounding text, because it would mean deleting a branch destroys both the history and the record
  that the history ever existed.
- If the events survive but refer to objects that no longer exist, describe what a reader would
  actually see when they look, rather than leaving the current sentence as it stands.

---

#### 7. Is the default branch created when Infrahub is first initialised?

**The issue.** The page describes the earliest moment you can query as the creation time of the
default branch, and then explains that this is the point at which Infrahub was first initialised. The
explanation came from Wim. The code only states that the limit is the creation timestamp of the
default branch.

**Why it matters.** This is the least serious item on the list, but it shapes a reader's mental model.
Somebody troubleshooting why they cannot query before a particular date will reason from this
sentence. The two descriptions would differ in any situation where the default branch is not as old as
the deployment, for example after a restore from backup, a migration between deployments, or a
re-initialisation.

**Who to ask.** Anyone familiar with how Infrahub initialises. It is specifically worth asking
somebody who has restored a deployment from a backup, because that is the situation where the two descriptions would no longer match.

**What to do with each answer:**

- If the two are always the same, keep the sentence as it is.
- If they can differ, delete the explanatory part and keep only the precise statement, which is the
  creation time of the default branch. That version is always correct, and the reader loses nothing by it.

**My suggestion.** If nobody can answer quickly, drop the explanation and keep the precise statement.
It is not worth holding the pull request for.

---

#### And one question carried over from the original description

**Is `updated_by` the intended way to answer "who changed this object"?**

**The issue.** The overview page says that a change can be traced to the account that made it, and
links to `objects/metadata.mdx`. However, `updated_by` records the last account to modify an object.
It is not a list of every change and its author.

**Why it matters.** Reading `updated_by` at a past timestamp gives you the account behind the change
that was current at that moment, so the statement on the page is not wrong. But a reader may expect a
per-change author history, which is not what this field provides. If they expect that and do not get
it, they will conclude the feature is broken.

**Who to ask.** Wim or Aaron.

**What to do with each answer:**

- If reading `updated_by` at a timestamp is the intended way to answer the question, keep the claim
  and add one sentence making clear that it gives the account behind the change current at that
  moment, not a full author history.
- If the Activity log is the intended answer for authorship questions, change the overview page to
  point there instead, and keep the link to `objects/metadata.mdx` only for the metadata itself.

### Where the reasoning is recorded

Two documents are committed on this branch and explain the decisions behind these pages.

`dev/specs/docs/immutable-history-temporal-queries.md` records why the pages are shaped the way they
are. It explains why a separate page about the "temporal graph" was rejected, in favour of keeping
the vocabulary the documentation already uses. It explains why the section about limits sits above
the section about syntax. It lists every behavioural fact that was checked in the source code, with
its file and line number, so that each one can be checked again when the code changes. It also
describes the seven questions above in more detail.

`dev/specs/docs/temporal-graph-blog-assessment.md` records the assessment of the temporal graph blog
post that started this work. It includes the reason the blog post's claim about supporting hundreds
of branches with no performance penalty was left out: there is nothing in the repository to support
that claim.

### One thing to expect from CI

Vale, which checks writing style, is not installed on the machine where this work was done, so it has
never run against these pages. CI will be the first place it runs, and it may report problems. As a
partial check, the new text was tested by script against the four style rules this project uses most
often, and no problems were found. That is not the same as running Vale itself. If Vale does report
something, it should need only a small follow-up change, not a structural one.

yjouffrault:
Tagging people against the specific items in the handover comment above, so nobody has to read the
whole thing to find their part.

**@ajtmccarty** — three items need a backend confirmation:

- **Item 1.** Is the base of a `DiffTree` comparison behaving as intended? The page states it is
  always the default branch (`queries/diff/tree.py`, line 551).
- **Item 2.** Should a timestamp in the future be rejected? Today only the earliest time is validated
  (`branch/query_time_validator.py`, line 32), so a future time is accepted and returns current data
  with no indication. Please redirect this if `BranchQueryTimeValidator` is not yours.
- **Item 6.** Does the Activity log still hold the operations from a branch after that branch is
  deleted? The page says it does. Events look like they are stored outside the branch edges that
  `core/query/branch.py` line 89 removes, but nobody has confirmed it.

Also, all five of your original points were applied, and the details are in the handover comment. You
were never replied to on those conversations, which was an oversight — apologies.

**@wvandeun** — five items follow from your review:

- **Item 1.** You said you were not certain the comparison base is working correctly. Item 1 above is
  that question.
- **Item 4.** Your three-way framing is now on the page, and it is a real improvement. But the third
  row links to the Activity log page, which does not document the `since` and `until` parameters. The
  handover comment sets out three options for closing that gap.
- **Item 5.** Your diagram request. The prose is written now, so the first question is whether that is
  enough for you. If not, `docs/media/` already has Excalidraw diagrams for branch merge and rebase,
  which look like the natural thing to extend.
- **Item 7.** You described the default branch's creation time as the moment Infrahub was first
  initialised. Is that always true, including after a restore from backup?
- **The `updated_by` question.** Is reading `updated_by` at a past timestamp the intended way to answer
  "who changed this object"?

**@lancamat1** — one item, and it is the most serious one here:

- **Item 3.** The editing hazard you found is now documented as a warning on the page. Please check
  that the description is accurate: that nothing blocks editing in a past view, and that a saved
  change including a delete applies at the current time. The product decision about whether to build
  the guard instead of documenting it is being routed separately.

Your other three points are all applied — the time zone correction, which screens honour the applied
time, and the shareable URL.

**@BaptisteGi** — nothing open for you. The line breaks are gone; both pages are now one paragraph per
line, which also matches the neighbouring pages.

dev/specs/docs/immutable-history-temporal-queries.md
```diff
@@ -0,0 +1,367 @@
...
+
+The error message the reader will see is:
+
+```text
+Requested time '<time>' is before branch '<name>' was created at '<created_at>'.
+```
+
+**Nothing checks whether the requested time is in the future.** A future time is accepted, and the
+query returns current data with no indication that this happened. The reason is that the filter used
+to select versions is `from <= $at AND (to IS NULL OR to > $at)`, in
```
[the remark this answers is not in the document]
yjouffrault: Correct, and fixed in 58a37b7dd.

The code writes the filter as two clauses joined by OR, and uses `>=`:

```
(r.from <= $at AND r.to IS NULL)
OR (r.from <= $at AND r.to >= $at)
```

The quote now matches that. You are also right that the `>=`/`>` difference does not change the conclusion about future timestamps — only the quote was wrong.

Worth recording how it got there: the wording was inherited from the working notes this document was rewritten from, and carried across without being checked against the source.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,219 @@
...
+| ISO 8601 without a zone | `2026-03-09T14:00:00` | The same wall-clock time, interpreted as UTC |
+| Date only | `2026-03-09` | 12:00 UTC on that day |
+| Offset from now | `30s`, `45m`, `6h`, `2h30m` | That interval before the current time |
+
+A date without a time resolves to midday rather than midnight, so include an explicit time when a day boundary matters. Relative offsets support seconds, minutes, and hours, including combined values such as `2h30m`. They do not support day or week units; use an absolute timestamp for those intervals.
+
+## Understand branch history limits
+
+Every branch reads history through the default branch, so the earliest time you can query is the default branch's creation time — the point at which Infrahub was first initialized. That floor is the same on a feature branch as on the default branch, because creating a branch records where it diverged rather than copying the dataset. A branch created this morning can still be read back to the default branch's creation, with the changes recorded on the branch since it diverged added to that history.
```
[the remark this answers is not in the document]
yjouffrault: Not taking this one, because it describes what the code permits rather than what Infrahub does.

Three things against it:

- `create_branch`'s own docstring says "currently all the branches are based on Main" (`backend/infrahub/core/initialization.py`).
- `origin_branch` defaults to `"main"` (`backend/infrahub/core/branch/models.py:33`).
- @ajtmccarty asked for exactly this in review: "the 'origin' is always the default branch. we don't want to imply that you can branch off of a branch. maybe someday, but not yet."

Writing "your origin branch's creation time" would describe a capability that does not exist yet.

On the global branch, @wvandeun asked for the opposite of this suggestion in review — "The default and global branch are the same thing" — and for the floor to be described as when the application first started. Restoring the distinction would undo that.

So both halves of this would reintroduce wording two maintainers had removed. Leaving the page as it is. If branching from a non-default branch becomes supported, this paragraph needs revisiting.

docs/docs/immutable-history/overview.mdx
yjouffrault: Correct, and the hub no longer implies otherwise. `overview.mdx` now says deleting a branch removes the data and history recorded on it permanently, that versions on the default branch are unaffected, and that the versions themselves cannot be recovered. Verified against `DeleteBranchEdgesQuery` in `backend/infrahub/core/query/branch.py:89`. Done in 80885fd12.

docs/docs/immutable-history/overview.mdx
yjouffrault: Done in 80885fd12. Both pages are now one paragraph per line, with no hard wrapping. That also matches the neighbouring pages -- `branches/overview.mdx` and `proposed-changes/overview.mdx` are written the same way, and `MD013` is switched off in `.markdownlint-cli2.yaml`, so the wrapping was never required.

docs/docs/immutable-history/overview.mdx
yjouffrault: Done in 80885fd12. It now reads: "Infrahub returns the objects, attribute values, relationships, and schema that were valid at that time."

docs/docs/immutable-history/overview.mdx
yjouffrault: Done in 80885fd12. The bullet now reads "Query objects, attributes, relationships, and the schema as they existed at a specific point in time." The same four levels are used consistently across both pages.

docs/docs/immutable-history/overview.mdx
yjouffrault: Used your wording, in 80885fd12: "The Activity log records operations: which object was changed, when that change occurred and who made it, and the sequence of actions."

docs/docs/immutable-history/overview.mdx
yjouffrault: Added in 80885fd12, in both places. `overview.mdx` has "The schema is versioned as well", and the how-to's "How historical queries work" explains that the schema is evaluated for the same point in time. Verified against `backend/infrahub/graphql/app.py:228-231`, where Infrahub calls `load_schema_from_db(at=at_ts)` when `at` predates the branch's `schema_changed_at`.

docs/docs/immutable-history/overview.mdx
yjouffrault: Added in 80885fd12, and it was worth including. `overview.mdx` now has a section covering the three branch operations that change what history remains available, and rebasing is one of them: it moves the timestamps of changes made on a branch up to the rebase time, so a change recorded before a rebase is no longer readable at its original timestamp. Verified against `backend/infrahub/core/branch/models.py:433`.

docs/docs/immutable-history/overview.mdx
yjouffrault: Both points done in 80885fd12. The section now opens with "History is recorded at four levels: objects, attributes, relationships, and the schema", then explains each. And the branch exception has moved out of there into "How branches use immutable history", which is where merge, rebase and delete are now explained together.

docs/docs/immutable-history/overview.mdx
yjouffrault: Removed in 80885fd12. There is no claim about replaying events on either page now. The only remaining mention of reconstruction is a contrast -- that you query past state directly rather than rebuilding it from a backup plus later changes -- which describes what Infrahub does, not a replay feature.

docs/docs/immutable-history/overview.mdx
yjouffrault: Agreed, and that is where it went. `overview.mdx` covers the schema alongside objects, attributes and relationships in "How Infrahub preserves history", in 80885fd12. See also the reply on your earlier comment about the same thing.

docs/docs/immutable-history/overview.mdx
```diff
@@ -2,65 +2,94 @@
...
+for that field alone. This makes it possible to identify the specific attributes that changed
+between two timestamps.
 
-Infrahub's temporal query system allows you to retrieve data from any point in time. When you query the database, you're not just accessing current data—you're accessing a specific moment in the database's history. By default, queries return the latest state, but you can specify any timestamp to see exactly how your infrastructure looked at that moment.
+Relationships are versioned as well. A query for a specific timestamp therefore returns both the
+attribute values and the object relationships that were valid at that time. This is important for
+topology and dependency questions where the connections between objects are part of the answer.
 
-This capability extends across all interfaces:
+The schema is versioned the same way. If the schema changed after the timestamp you request,
```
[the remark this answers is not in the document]
yjouffrault: Answered rather than changed: the behaviour is established, so the paragraph stands. When `at` is earlier than the branch's `schema_changed_at`, Infrahub calls `load_schema_from_db(at=at_ts)` and analyses the query against that schema -- `backend/infrahub/graphql/app.py:228-231`. The PR description listed this as unresolved at the time you commented; it has since been verified and the description updated. No change needed here.

docs/docs/immutable-history/overview.mdx
yjouffrault: Done as you asked, in 80885fd12. Neither page mentions origin branches; the rule is stated as the default branch in every case, and the workaround that used to hint at branching from a branch is gone.

Worth flagging: cubic later argued the opposite on this same paragraph, asking for the origin-branch nuance to be restored. I declined it and cited you, because `create_branch`'s own docstring says "currently all the branches are based on Main" and `origin_branch` defaults to `"main"`. If that ever changes, this paragraph needs revisiting.

docs/docs/immutable-history/overview.mdx
yjouffrault: Written up in 80885fd12, using your explanation. `overview.mdx` now has a section on merging, rebasing and deleting a branch. It states that merged changes are recorded on the default branch with the timestamp of the merge rather than the ones they had on the branch, gives the Monday-created / Friday-merged example, and says the intermediate versions stay readable on the branch only while it exists. Verified against `backend/infrahub/core/diff/merger/merger.py:94-210`.

On the graphic: not done, and I think you are right that it would explain this better than prose can, since it is two dimensions at once. It is tracked as an open item on the PR. `docs/media/` already has `branch_merge.excalidraw.svg` and `branch_rebase_no_conflict.excalidraw.svg`, which look like the natural things to extend.

docs/docs/immutable-history/overview.mdx
yjouffrault: Used your wording, in 80885fd12: "Set a time when you need the state of the data at an earlier time."

docs/docs/immutable-history/overview.mdx
yjouffrault: Done in 80885fd12, and it changed the shape of the section. "Compare changes between two timestamps" now leads with the calculation step: send `DiffUpdate`, then read the result with `DiffTree` or `DiffTreeSummary`. It also documents something that was not written down anywhere -- a custom time range requires a `name`, or Infrahub rejects the request (`backend/infrahub/graphql/mutations/diff.py:59`). That is the actual reason an arbitrary period returned nothing on a first attempt.

docs/docs/immutable-history/overview.mdx
yjouffrault: Added in 80885fd12: "a branch's Branch view shows its diff and lets you refresh it, whether or not a Proposed Change exists for that branch". The page also now mentions that a Proposed Change shows the diff between its source and target branches, and that `proposed_change_id` retrieves it over the API.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Done in 80885fd12. The opening line is now "You can read infrastructure data as it existed at previous points in time, and compare how it changed between them."

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Used your wording, in 80885fd12. The section now opens "To read the state of the data at a specific point in time, specify a branch and a timestamp."

yjouffrault: Both done in 80885fd12. It now reads "Infrahub returns the objects, attribute values, and related objects that were valid at that time", and the clause about the interfaces you use for current data is gone.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: You were right that it was not true, and this reshaped the page. The opening now names three ways to ask about the past, in a table: state at one point in time (`at` and `branch`), what changed between two points (`DiffUpdate` then `DiffTree`), and which operations happened between two points (`since` and `until` on an event query). The third one I had not known about -- verified at `backend/infrahub/graphql/queries/event.py:140-141`, where `since` defaults to 180 days back. Done in 80885fd12.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Correct, and it is now scoped that way. That section is explicitly about the first of the three cases -- it opens "To read the state of the data at a specific point in time" -- and the other two have their own sections, linked from the table at the top. Done in 80885fd12.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Both done in 80885fd12. The paragraph now says history is recorded per object, per attribute and per relationship, and it links to "How Infrahub preserves history" on the overview page for the storage model, rather than repeating it here.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Rewritten in 80885fd12. It now reads: "Infrahub can return those versions because a change records a new value alongside the previous one rather than replacing it. History is recorded per object, per attribute, and per relationship, which is what lets a comparison identify the specific fields or connections that changed and what they held before." If it is still unclear, say so and I will take another pass.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Done in 80885fd12: "the query sees the objects, attributes, and relationships that existed at that point rather than the current ones".

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Used your wording, in 80885fd12: "A timestamp applies to read operations only -- queries, in GraphQL terms."

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Done in 80885fd12: "Use a timestamp when you need the objects, attributes, and relationships that existed at a known point".

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Used your wording, in 80885fd12. The tab now opens "Use the `at` URL parameter to specify the timestamp for a GraphQL query", and then says `at` goes on the endpoint rather than in the query document.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Corrected in 80885fd12 -- it is described as a URL parameter throughout now, on this page and on the overview.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Agreed, and fixed in 80885fd12. The endpoint is out of the code comment and in its own block above the query, so it renders as text rather than as a greyed-out comment:

```text
http://localhost:8000/graphql/main?at=2026-03-09T14:00:00Z
```

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Done as you suggested, in 80885fd12. The REST tab no longer explains when to reach for REST. It lists the four endpoints that accept `at` in a table: `/api/query/{query_id}`, `/api/artifact/{artifact_id}`, `/api/transform/python/{transform_id}` and `/api/transform/jinja2/{transform_id}`. Verified against `backend/infrahub/api/dependencies.py:77-90` and `backend/infrahub/api/transformation.py:35,100`.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Redone in 80885fd12, in the order you described. The section now says comparing two timestamps takes two steps, shows the `DiffUpdate` mutation first, then retrieval with `DiffTree` or `DiffTreeSummary`. It also documents a requirement that was not written down anywhere: a custom time range requires a `name`, or Infrahub rejects the request (`backend/infrahub/graphql/mutations/diff.py:59`).

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Agreed, and that is where it went -- in the how-to's "Compare changes between two timestamps", as the first step rather than a note at the end. Done in 80885fd12.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Corrected in 80885fd12. The page now says a Proposed Change shows the diff between its source and target branches, alongside the branch view, and documents `proposed_change_id` for retrieving a Proposed Change's diff over the API (`backend/infrahub/graphql/queries/diff/tree.py:764`).

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: You were right, and the contradiction is gone. There is one rule now: every branch reads its history through the default branch, so the earliest time you can query is always the default branch's creation. Done in 80885fd12.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Done in 80885fd12, using your framing. The section states one floor -- the default branch's creation time -- and describes it as the point at which Infrahub was first initialised. The separate default-and-global sentence is gone.

One thing worth confirming when you have a moment: the "first initialised" wording is yours rather than something I could verify in code, which only says the default branch's `created_at`. If those can differ, for instance after a restore from backup, tell me and I will drop the explanation and keep the precise statement.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Removed in 80885fd12. You were right that it was unnecessary -- the history from before the branch was created is already reachable, so there is nothing to work around.

docs/docs/immutable-history/query-historical-data.mdx
yjouffrault: Moved in 80885fd12, to exactly where you suggested. The rebase note is now in `overview.mdx`, in the same section as merging and deleting a branch, since all three change what a historical query can return.

docs/docs/immutable-history/query-historical-data.mdx
lancamat1: Checked this on a live instance, and it works. `DiffTree` returns `base_branch: "main"` in every form that returns data, and `DiffUpdate` hardcodes the base at `backend/infrahub/graphql/mutations/diff.py:62`. So the sentence on the page is correct as written.

Checking it did turn up something beside it. Retrieving a diff for a custom period by the name you stored it under returns `null` rather than the diff, because the resolver recomputes a default range and the lookup needs the stored range to match. The page now documents retrieval by `from_time` and `to_time`, and the behaviour is filed as #10592.

docs/docs/immutable-history/query-historical-data.mdx
```diff
@@ -0,0 +1,223 @@
...
+    wait_until_completion: true
+  ) {
+    ok
+  }
+}
+```
+
+Then retrieve it with the same branch and time range you calculated it for, using `DiffTree` for the changed nodes or `DiffTreeSummary` when you only need counts. The timestamps do not need to align with a branch point or a Proposed Change, so you can compare any useful period, such as the hours around an incident.
+
+Retrieve a custom period by that period rather than by its name. `DiffTree` looks for a diff covering the range it computes by default, from the branch point to the present, so it does not find one stored for a different range and returns `null`. `DiffTreeSummary` does not accept `name` at all and returns an error. A name is still worth setting, because `DiffUpdate` requires one for a custom period.
```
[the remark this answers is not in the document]
lancamat1: You are right about `DiffTree`, and I have corrected the page. A diff stored with no explicit period is found by name. Only one stored for a custom period is missed, because the range the resolver recomputes by default, branch point to now, does not match the stored range. The example here uses a custom period, so the guidance to retrieve by `from_time` and `to_time` is unchanged.

You are not right about `DiffTreeSummary`. It does not ignore `name`, it raises:

```
DiffTreeResolver.summary() got an unexpected keyword argument 'name'
```

`DiffTreeSummaryQuery` declares `name=String()` but the `summary()` signature has no `name` parameter, so graphene passes a kwarg the resolver cannot accept. Observed three times on a live instance.

Both halves are recorded on #10592.

lancamat1:
@wvandeun taking this over from @yjouffrault. I have resolved the 38 conversations that were already answered and applied, so there are two left, both yours.

**`:202`, the comparison base.** This one never got a reply. It works: `DiffTree` returns `base_branch: "main"` in every form that returns data, and `DiffUpdate` hardcodes it at `backend/infrahub/graphql/mutations/diff.py:62`. Answered on the thread, yours to close.

**`:220`, the default branch's creation time.** Still open for you. I confirmed the floor is the default branch's `created_at` and that the page's example error text matches the real message, but not that it is always the moment Infrahub was first initialised. If a restore from backup can make those differ, say so and the explanation comes out.

Since your review I verified every behavioural claim on both pages against a live instance rather than against the code, and validated every GraphQL example against `schema/schema.graphql`. Nine claims held. Three did not, and are fixed in three commits:

- retrieving a diff for a custom period by its name returns `null`, so the example now uses the time range (#10592)
- object metadata is not resolved at the requested time, so authorship on the hub now points at the Activity log instead (#10591)
- `attributes { action }` does not exist, `DiffAttribute` exposes `status`

I also moved the code fences to the convention used on `stable`. The ```` ```graphql # label ```` form puts the label in the metastring where Docusaurus never renders it, so the page had zero code block titles and now has three. The `# Endpoint:` comment lines are gone, which finishes your point about the commented-out endpoint being hard to read; the explicit endpoint block you asked for is untouched.

Last commit documents `since` and `until` on the Activity log page, closing the gap this pull request opened by linking there for event filtering.

Five issues came out of the verification: #10589, #10590, #10591, #10592, and #10593 for the diagram you asked for. Each behaviour is documented as it works today, and each issue names the paragraph to delete when it is fixed.

wvandeun (approved):
LGTM

docs/docs/immutable-history/query-historical-data.mdx
lancamat1: Done, in two parts.

The Activity log page did not document `since` and `until` at all, so the row in the table at the top of this page named them and then linked somewhere they were not explained. That page now has a "Querying the activity log over the API" section covering all fifteen parameters on the `InfrahubEvent` query, with an example, and it records that `since` defaults to 180 days back, so a query without it returns only the last 180 days and nothing in the response says so.

The row here now links straight to that section rather than to the top of the page.

I kept the detail on the Activity log page rather than duplicating it here, since that is where someone looking for event filtering will go, and two copies would drift.

PR #10339, opened 2026-08-20
fix(backend): write branch-agnostic attribute edges on the global branch

ajtmccarty:
# Why

The `node.attribute.add` schema migration stamped every edge it created with the branch the migration
ran on, ignoring the attribute's declared branch support. Adding an attribute declared
`branch: agnostic` to an existing kind therefore produced an `Attribute` vertex marked `agnostic`
whose `HAS_ATTRIBUTE`, `HAS_VALUE` and `IS_PROTECTED` edges sat on that branch instead of `-global-`,
where every other write path puts them.

**Goal:** make the migration resolve the write branch the same way the normal create path does, so a
branch-agnostic attribute added by a schema update is stored once and readable from every branch.

**Non-goals:** repairing databases that already contain the wrong shape (no graph migration here,
`GRAPH_VERSION` stays at `75`), and the branch-agnostic edge *lifecycle* work, which is separate.

### Behavioral changes

- A branch-agnostic attribute added by a schema migration now has its edges written to `-global-` at
  `branch_level: 1`, so it is readable from every branch rather than only the migration branch.
- A branch-local attribute of a branch-agnostic node (pretty unusual, but possible) follows the
  same rule, matching the second arm of `BaseAttribute.get_create_data()`.
- A branch-agnostic `NumberPool` attribute now reads back the value the migration allocated. Before,
  the migration wrote the default value on its own branch while the allocation went through the normal
  write path to `-global-`; the default edge was never closed and won the read, so every existing
  object reported no value while the pool treated the numbers as taken.
- `verify_graph` reports a new violation when an attribute's live edges are not on the branch its
  branch support implies.

### Implementation notes

The write destination is resolved **per matched node**, inside the query, from the vertex's own
`branch_support` — not once for the whole migration:

Resolving it per query instead would be wrong, and this is the reason the parameterised version was
abandoned. `_get_node_kinds` expands a generic to every `used_by` kind plus their profiles and
templates, and a generated schema takes its branch support from the node it is generated from
(`ProfileSchema`/`TemplateSchema` are constructed with `branch=node.branch`). One generic can therefore
be inherited by kinds with *different* branch support, and a single write branch is wrong for one of
them: a branch-local attribute of an aware `used_by` kind would be written to `-global-` because the
generic is agnostic. Reading the destination off the row makes that whole question disappear, and it is
the faithful translation of `get_create_data()`, which reads the concrete instance's own schema.

The new graph check judges only **open, active** edges. That is what keeps it correct rather than
noisy: `get_branch_for_delete()` deliberately writes *branch-scoped* `deleted` edges for an agnostic
attribute on an aware node, and same-uuid node copies leave closed owning edges behind. It anchors on
the `Node`/`Attribute` labels rather than enumerating kinds, so profile and template copies are covered
without being listed. It treats a *missing* `branch_level` as invalid alongside a wrong one, since
`branch_level <> 1` evaluates to `NULL` for an absent property and would otherwise drop the row;
`assert_no_global_edges_with_wrong_branch_level` had the same hole and is fixed the same way.

### What stayed the same

- **The match side of the query is untouched.** A branch's query filter already spans `-global-` at
  query time, so the existing re-run guard still finds an owning edge a previous run put there — there
  is a test for exactly this.
- **No schema, API or GraphQL changes. No new dependencies. No migration, and `GRAPH_VERSION` is
  unchanged.**
- Branch-aware attributes still land on the migration branch; there is a regression test asserting the
  specific branch and level, not merely "not global".

### Suggested review order

1. `backend/infrahub/core/migrations/query/attribute_add.py` — the rule, and the only behavioral change
   to the migration.
2. The three schema call sites and `m013`, which only *lose* an argument
   (`node_attribute_add.py`, `attribute_supports_generated_schema.py` ×2,
   `m013_convert_git_password_credential.py`).
3. `backend/infrahub/database/validation.py` — `_check_attribute_edge_branch`, independent of the fix.
4. Tests. `test_local_attribute_of_generic_follows_each_node_kind` is the one that pins the per-node
   behavior: one agnostic generic inherited by both an aware and an agnostic kind, one branch-local
   attribute, each instance's edges asserted against its own kind's branch support.

## Impact & rollout

- **Backward compatibility:** no API, schema or GraphQL change. Existing databases that already have
  agnostic attributes with branch-scoped edges are **not** repaired by this PR — the fix only corrects
  new writes.
- **Performance:** no new queries on any hot path. The migration query gains two `WITH` projections
  evaluated per matched node. `_check_attribute_edge_branch` runs only where
  `verify_graph`/`collect_graph_violations` is already called, which today is tests only.
- **Deployment notes:** safe to deploy. No coordinated release, no `GRAPH_VERSION` bump, no upgrade step.
- **Known follow-up:** `AttributeAddQuery` gates vertex metadata on the executing branch while
  `AttributeUpdateValueQuery` gates it on the write destination (`WHERE $branch_level = 1`), so an
  `Attribute` vertex's metadata depends on whether it was last touched by a create or an update. This
  predates the change and is not fixed here — the single `set_metadata` flag also gates the *Node*
  vertex's `updated_at`/`previous_updated_at`, which feeds the rollback snapshot, so a correct fix has
  to split the flag and pick one convention.

## Checklist

- [x] Tests added/updated
- [ ] External docs updated (if user-facing or ops-facing change) — n/a, this restores intended
      behavior rather than adding a user-facing concept

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10339?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/migrations/query/attribute_add.py
+++ b/backend/infrahub/core/migrations/query/attribute_add.py
@@ -29,7 +29,6 @@ class AttributeAddQuery(Query):
         attribute_name: str,
         attribute_kind: str,
         branch_support: str,
-        node_branch_support: str,
         default_value: Any | None = None,
         uuids: list[str] | None = None,
         **kwargs: Any,
@@ -38,18 +37,9 @@ class AttributeAddQuery(Query):
         self.attribute_name = attribute_name
         self.attribute_kind = attribute_kind
         self.branch_support = branch_support
-        self.node_branch_support = node_branch_support
         self.default_value = default_value
         self.uuids = uuids
         super().__init__(**kwargs)
-
-    def _writes_on_global_branch(self) -> bool:
-        if self.branch_support == BranchSupportType.AGNOSTIC.value:
-            return True
-        return (
-            self.branch_support == BranchSupportType.LOCAL.value
-            and self.node_branch_support == BranchSupportType.AGNOSTIC.value
-        )
 
     async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> None:  # noqa: ARG002
         branch_filter, branch_params = self.branch.get_query_filter_path(at=self.at.to_string())
@@ -70,20 +60,13 @@ class AttributeAddQuery(Query):
 
         self.params["user_id"] = self.user_id
 
-        if self._writes_on_global_branch():
-            edge_branch_name = GLOBAL_BRANCH_NAME
-            edge_branch_level = 1
-        else:
-            edge_branch_name = self.branch.name
-            edge_branch_level = self.branch.hierarchy_level
-
-        self.params["rel_props"] = {
-            "branch": edge_branch_name,
-            "branch_level": edge_branch_level,
-            "status": RelationshipStatus.ACTIVE.value,
-            "from": write_time,
-            "from_user_id": self.user_id,
-        }
+        self.params["is_branch_agnostic"] = self.branch_support == BranchSupportType.AGNOSTIC.value
+        self.params["is_branch_local"] = self.branch_support == BranchSupportType.LOCAL.value
+        self.params["agnostic_support"] = BranchSupportType.AGNOSTIC.value
+        self.params["global_branch_name"] = GLOBAL_BRANCH_NAME
+        self.params["branch_name"] = self.branch.name
+        self.params["branch_level"] = self.branch.hierarchy_level
+        self.params["edge_status"] = RelationshipStatus.ACTIVE.value
 
         self.params["is_protected_default"] = False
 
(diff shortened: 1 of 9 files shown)


backend/infrahub/core/migrations/query/attribute_add.py
```diff
@@ -49,9 +59,10 @@ async def query_init(self, db: InfrahubDatabase, **kwargs: dict[str, Any]) -> No
 
         self.params["user_id"] = self.user_id
 
+        writes_on_global_branch = self._writes_on_global_branch()
         self.params["rel_props"] = {
-            "branch": self.branch.name,
-            "branch_level": self.branch.hierarchy_level,
+            "branch": GLOBAL_BRANCH_NAME if writes_on_global_branch else self.branch.name,
+            "branch_level": 1 if writes_on_global_branch else self.branch.hierarchy_level,
```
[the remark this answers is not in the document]
ajtmccarty: captured in a follow-up issue https://opsmill.atlassian.net/browse/IFC-3032

backend/tests/component/database/test_validation.py
```diff
@@ -269,6 +296,24 @@ class GraphDamageCase:
...
+    ),
+    GraphDamageCase(
+        name="aware_attribute_edges_on_global_branch",
+        check=GraphCheck.ATTRIBUTE_EDGE_BRANCH,
+        damage=_move_aware_attribute_edges_to_global,
+        expected_violations=1,
+        scoped_violations=1,
+        included_kinds=["TestCar"],
+        excluded_kinds=["TestPerson"],
+    ),
```
polmichel, line 357: Optional comment: 
These test cases look relevant. Should we add other test cases regarding the following cases code inside the database check?
- branch_level is NULL
- branch_level is != 1

I do not see any test covering those but maybe I've missed something here.
```
    if result.get("reason") == "branch_level":
            problem = f"on {branch=} with {branch_level=}, expected branch_level 1"
        elif result.get("expects_global"):
            problem = f"on {branch=}, expected '{GLOBAL_BRANCH_NAME}'"
        else:
            problem = f"on {branch=}, expected a branch of its own"
```

backend/tests/component/core/migrations/schema/test_node_attribute_add.py
```diff
@@ -612,3 +616,378 @@ async def test_migration_numberpool_attribute(
...
+    assert (
+        await run_test_car_attribute_add(
+            db=db, branch=migration_branch, previous_schema=previous_schema, candidate=candidate, attribute_name="tag"
+        )
+        == 0
+    )
+    assert await count_nodes(db=db, label="Attribute") == attribute_count
+
+
+async def test_migration_agnostic_numberpool_attribute(
```
polmichel, line 843: Great test

polmichel (approved):
LGTM, nice fix!

Only one optional comment regarding additional test cases

backend/tests/component/database/test_validation.py
ajtmccarty: good point. tests added

PR #10347, opened 2026-08-20
test(frontend): make the pre-bundled dep list resolvable so vitest cannot reload mid-run

fatih-acar:
## Summary

Kills the `vi.mocked(...).mockX is not a function` CI flake that has hit **15+ PRs since week 31**, failing a different frontend unit-test file each time. The mock corruption was collateral damage of Vite re-optimizing dependencies mid-run and reloading the browser page while a test file was executing.

## Root cause

Five `optimizeDeps.include` entries in `frontend/app/vitest.config.ts` — `@dagrejs/dagre`, `html-to-image`, `@radix-ui/react-scroll-area`, `react-resizable-panels`, `tailwind-variants` — are dependencies of the source-consumed workspace packages (`infrahub-schema-visualizer`, `@infrahub/ui`), not of the app. Under pnpm's strict layout they are not resolvable from the app root, so Vite silently skipped pre-bundling them, warning at startup:

```
Failed to resolve dependency: @dagrejs/dagre, present in client 'optimizeDeps.include'
```

The first test importing them (`path-flow-graph.tsx` → `infrahub-schema-visualizer`) triggered a mid-run re-optimization → `optimized dependencies changed. reloading` → vitest's own warning `Vite unexpectedly reloaded a test…` → the reloaded test file re-ran against a fresh, un-mocked module graph, so `vi.mocked(realFn)` had no mock methods. Which file died was a pure timing race — hence the rotating victims (preferences, relationships, tasks, permission, hierarchy, refresh-button).

The causal model is already proven in-repo: 1dda41c5c made `@radix-ui/react-dropdown-menu` resolvable+included on 2026-08-11 and its mid-run re-optimizations stopped (12 CI logs before, none after).

## Key Changes

- The five packages are declared as app `devDependencies` at the exact versions the lockfile already pins via the workspace members — nothing new enters the dependency tree, the include list simply becomes effective.
- The `optimizeDeps` comment now records the constraint: every include entry must be resolvable from the app root, or Vite silently skips it and the mid-run reload comes back.
- `pnpm-lock.yaml` change is surgical (+15 lines: the five importer references).

## Test Plan

- Before (stable): `pnpm exec vitest run src/entities/path-traversal src/entities/nodes/relationships/domain/use-cases` logs all five `Failed to resolve dependency` warnings at startup.
- After (this branch): same run logs none, no `dependencies optimized:` events after startup, no `unexpectedly reloaded` warning; 43/43 tests pass.
- Watch `frontend-tests` in CI: the flake signature (`vitest-mock-corruption`) should disappear from the weekly flakiness ledger.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10347?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/app/package.json
```diff
@@ -109,9 +109,11 @@
     "@betterer/cli": "6.0.0-alpha.1",
     "@betterer/typescript": "6.0.0-alpha.1",
     "@biomejs/biome": "^2.5.1",
+    "@dagrejs/dagre": "^1.1.8",
```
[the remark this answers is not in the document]
fatih-acar: Not changing this — the premise is wrong: knip's Vite plugin **does** parse `optimizeDeps.include` strings as dependency usage, so `pnpm knip` (the exact frontend-lint command) exits 0 on this branch with no unused-dependency findings.

Verified both ways locally: as-is → exit 0; with the five entries temporarily stripped from `optimizeDeps.include` → knip flags exactly these five packages as unused devDependencies and fails. That counterfactual also shows why adding them to `ignoreDependencies` would be mildly harmful: knip currently enforces the include↔devDependency coupling for free — delete an include entry later and the orphaned devDep is flagged immediately; a blanket ignore would silence that.

frontend/app/package.json
```diff
@@ -109,9 +109,11 @@
...
     "@graphql-codegen/cli": "^7.2.0",
     "@graphql-codegen/typescript": "^6.1.0",
     "@playwright/test": "1.62.1",
+    "@radix-ui/react-scroll-area": "^1.2.18",
```
saltas888, line 116: this is weird to be dev dependency. its not tooling, its actual lib

frontend/app/package.json
```diff
@@ -125,8 +127,11 @@
     "@vitest/browser-playwright": "^4.1.10",
     "@vitest/coverage-v8": "^4.1.10",
     "babel-plugin-react-compiler": "catalog:",
+    "html-to-image": "^1.11.13",
     "knip": "~6.27.0",
     "openapi-typescript": "^7.13.0",
+    "react-resizable-panels": "^4.12.2",
```
saltas888, line 133: same for those 2

saltas888: I guess this is because they are side dependencies? This is a bit weird, but if it solves the issue Im fine.

In this case I find this a must have documented thing.

fatih-acar: added comments to keep track of this
will merge as-is for now

saltas888 (approved):

PR #10361, opened 2026-08-21
fix(backend): keep non-mutation GraphQL execution on the sync path

fatih-acar:
## Summary

GraphQL read queries were paying a hidden per-field tax: the branch-status middleware is an `async` function, so every resolved field of every query returned a coroutine and pushed the whole execution onto graphql-core's async completion path. For large reads this is dramatic — the GraphQL sandbox's schema introspection took **12.3s** of event-loop-pinning CPU per visit (0.85s on the sync path), during which the affected API worker served nothing else.

This is also the root cause of CI's top e2e flake, `test_should_display_breadcrumb_on_ipam_page` (12 hits across 3 weeks): the sandbox e2e test closes its page ~1.4s in, the abandoned introspection keeps pinning a worker, and HAProxy parks exactly one of the next page load's asset requests on that worker's idle keep-alive connection. `/ipam` is the next test in file order, and on shared runners the park exceeds the 30s expect timeout — a blank page with one asset request forever in flight (all six downloadable failure traces show exactly one request with status `-1` among ~165 sub-20ms ones).

## Key Changes

- Every non-mutation GraphQL request now executes on graphql-core's synchronous resolution path; the middleware dispatches to an async gate only for top-level mutation fields, so mutation blocking on merged/needs-rebase/merging branches behaves exactly as before.
- The GraphQL response-size metric reuses the already-rendered response body instead of serializing the multi-MB payload a second time.
- New unit test pins the invariant that query/introspection execution completes without touching the event loop (it fails against the previous middleware).

## Test Plan

- `backend/tests/unit/graphql/` — 125 tests pass, including the new sync-path regression tests.
- `backend/tests/functional/branch/test_branch_needs_rebase.py` — the mutation gate still blocks writes on a NEED_REBASE branch and still allows `BranchRebase`.
- Validated on the e2e testcontainers stack (fix hot-patched into the running image): introspection over the LB 12.3s → ~0.9s; a 25-iteration sandbox→/ipam browser loop reproducing the flake went from p50 10.1s / p90 12.4s page loads to p50 0.40s / max 1.75s with zero failures.
- Watch the `E2E-testing-pytest-playwright (branches_repo)` job — the flaky test lives there.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10361?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/graphql/middleware.py
+++ b/backend/infrahub/graphql/middleware.py
@@ -1,4 +1,7 @@
 from inspect import isawaitable
+from typing import Any
+
+from graphql import GraphQLResolveInfo
 
 from infrahub.branch.status_checker import BranchStatusChecker
 from infrahub.core.merge.write_blocker import MergeWriteBlocker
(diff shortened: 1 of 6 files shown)


backend/infrahub/graphql/middleware.py
```diff
@@ -9,20 +12,29 @@
 ALLOWED_MUTATIONS_DURING_MERGE = ["BranchCreate", "BranchDelete"]
 
 
-async def raise_on_mutation_for_branch_status(next, root, info, **kwargs):  # type: ignore  # noqa
+def raise_on_mutation_for_branch_status(next: Any, root: Any, info: GraphQLResolveInfo, **kwargs: Any) -> Any:  # noqa: A002
+    # Stay synchronous outside the gated case: an async middleware would return a coroutine for
+    # every resolved field, forcing the whole execution onto graphql-core's async completion path.
```
ajtmccarty, line 17: might be worth adding something to the dev/knowledge/backend files somewhere about keeping async out of middleware wherever possible

ajtmccarty (approved):
great work Fatih. cleaning up my mistake

backend/infrahub/graphql/middleware.py
fatih-acar: Good call — added in 5c6474ecf: [`dev/knowledge/backend/graphql-execution.md`](https://github.com/opsmill/infrahub/blob/fac/flaky-check-ej7fo/dev/knowledge/backend/graphql-execution.md) covers how graphql-core picks sync vs async completion per field, why an async middleware multiplies by the field count and pins the worker event loop (with the measured 0.85s → 12.3s introspection numbers), and the sync-dispatch-with-async-gate pattern to use instead. Also linked it from the dev/README knowledge index.

ajtmccarty (approved):

PR #10362, opened 2026-08-21
test(component): share read-only fixture datasets at class scope

saltas888:
## Measured improvements

Full `core-schema` shard, `-n 2`, same machine, before → after (identical results: 1187 passed / 10 skipped):

| file | in-shard setup time |
|---|---|
| `test_relationship_peer_common_parent_update.py` | 151.7s → **6.7s** |
| `test_attribute_numberpool_constraints.py` | 75.8s → **16.2s** |

Roughly ~110 worker-seconds saved per shard run after correcting for run-to-run environment variance. Read-only tests now share one class-scoped dataset (built via the `default_branch_scope_class` chain) instead of wiping and rebuilding per test — safe because pytest runs with `--dist loadscope`, so a class's tests always execute contiguously on one xdist worker. Tests that mutate shared state (`test_query_deleted_lag_members`, `test_validator_range`) keep their own function-scoped copies.

Profiling notes: the per-test full-graph `DETACH DELETE` wipe measured 0.04s avg over 577 wipes — it is *not* a meaningful cost; the expense is per-node writes in data fixtures (~0.1–0.9s per `Node.save` round-trip).

## 📓  How to review
!**Use side-by-side view!**
Screenshot 2026-08-21 at 11 21 52

## Follow-ups (not in this PR)

- Mutation-heavy files (`test_number_pool_query`, `test_relationship_count`, `test_attribute_choices_update`, `test_node_applier`) need the ipam pattern instead (module-scoped dataset + per-test `DeleteAfterTimeQuery` rollback, unique branch names).
- `create_default_account_groups` runs 42 sequential queries (~7–12s) in every `TestInfrahubApp` class boot.
- Fail-fast timeout around the Prefect task-manager setup (source of the recent 300s CI timeouts).

## Test Plan

- `uv run pytest -n 2 backend/tests/component/core/...` full shard: 1187 passed / 10 skipped, matching baseline.
- Both restructured files also verified standalone (8/8 and 6/6 pass).

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/component/core/constraint_validators/test_attribute_numberpool_constraints.py
+++ b/backend/tests/component/core/constraint_validators/test_attribute_numberpool_constraints.py
@@ -21,111 +21,133 @@ from infrahub.pools.schema_number_pool_upserter import SchemaNumberPoolUpserter
 from tests.helpers.schema.snow import SNOW_INCIDENT, SNOW_REQUEST, SNOW_TASK
 
 
-@pytest.fixture
-async def snow_incident_01(db: InfrahubDatabase, default_branch: Branch, register_core_models_schema: None) -> Node:
+async def build_snow_incident(db: InfrahubDatabase, branch: Branch) -> Node:
     schema = SchemaRoot(generics=[SNOW_TASK], nodes=[SNOW_INCIDENT, SNOW_REQUEST])
-    registry.schema.register_schema(schema=schema, branch=default_branch.name)
+    registry.schema.register_schema(schema=schema, branch=branch.name)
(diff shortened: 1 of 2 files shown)


fatih-acar (approved):
Nice one! Can we target stable instead?

PR #10368, opened 2026-08-21
chore: clear most mypy and ty suppressions for infrahub.database

ogenstad:
# Why

The `infrahub.database` override in `pyproject.toml` disabled six mypy error codes, and the matching ty override ignored five rules. Both were placeholders for "fix the code later". Two of the mypy entries (`index`, `misc`) suppressed nothing at all any more.

Goal: fix the underlying type errors rather than trim the list, and document precisely what would be required to remove whatever is left.

Non-goals: no behaviour changes, and no rename of the index managers' `list()` method (see below).

## What changed

**Behavioral changes:** none intended. One API signature is narrower:

- `DatabaseSchemaManager.set()` now returns `None`. Its `int` annotation was wrong on one of its two paths - the branch-local one returns `SchemaBranch.set`'s `str` hash - and no caller in this repo uses the value. Worth a grep in infrahub-enterprise.

**Fixes behind the removed suppressions:**

- `get_node_schema` / `get_generic_schema` narrow with `isinstance`, matching the shape the equivalent accessors on `SchemaManager` already use (`backend/infrahub/core/schema/manager.py:110-124`).
- `SchemaManager.set` widened to `MainSchemaTypes` - what it already forwards to `SchemaBranch.set` and what `DatabaseSchemaManager` already passes it. Annotation-only; note this means mypy no longer blocks passing a profile/template schema there.
- `InfrahubDatabase(schemas=...)` gets a `list` instead of `dict_values`.
- `__aexit__` reads the session/transaction through guarded accessors instead of dereferencing them unchecked.
- `run_query` split so the explicit-transaction and auto-commit paths each pass the query form their execution method accepts. The transaction acquisition deliberately stays outside the `try`, so a `ServiceUnavailable` raised by `begin_transaction()` still propagates raw instead of becoming a `DatabaseError` - that's why the `except` clause is duplicated.
- `trusted_certificates` annotated with `TrustStore`, the base the driver's own signature accepts. neo4j exports only the three concrete subclasses, so it is imported under `TYPE_CHECKING`; a future neo4j reshuffle would break type-checking, never runtime.

**Result:** mypy 16 violations -> 1, six disabled codes -> one. ty 26 diagnostics -> 14, five ignored rules -> three.

**What stayed the same:** no schema, migration, GraphQL or API changes. No new dependencies. No changelog fragment (internal typing debt).

## How to review

Focus on two things:

1. **`run_query`** - the only non-trivial rewrite. Equivalence was checked by characterising old and new against recording/failing doubles across six scenarios (both modes x success / `ServiceUnavailable` / other `Neo4jError`), capturing the exact object handed to `run()`, the return value and the exception chain. Output was byte-identical. That characterisation is now committed as `backend/tests/unit/database/test_run_query.py`; inverting the mode branch in the source makes all 6 of its cases fail, so it does detect regressions.
2. **The new runtime import** - `from infrahub.core.schema import GenericSchema, NodeSchema` moved from `TYPE_CHECKING` to module level, which adds 193 modules (79 of them `infrahub.*`) to `infrahub.database`'s import-time closure. No cycle exists: after importing `infrahub.core.schema` alone, `infrahub.database` is absent from `sys.modules`, so nothing in that closure imports back. All 964 modules under `backend/infrahub` import cleanly, matching the pre-change baseline exactly. Happy to move it into the two methods if you would rather keep the import surface flat.

The `pyproject.toml` diff is the summary of what's left and why.

## How to test

```bash
uv run invoke lint                                    # ruff + ty + mypy, with the reduced ignore lists
uv run pytest backend/tests/unit/database -q          # includes the new run_query tests
INFRAHUB_USE_TEST_CONTAINERS=false uv run pytest \
  backend/tests/component/core/migrations/graph/test_014.py -q   # drives the rewritten __aexit__
```

Two pre-existing failures are unrelated to this branch and reproduce identically on `develop`: `backend/tests/unit/services/adapters/workflow/test_local_stamping.py` (4 tests, needs a Prefect API on :4200) and `npx betterer` (34 new TS issues; this branch contains no frontend files).

## Impact & rollout

- **Backward compatibility:** `db.schema.set()` no longer returns a value; nothing in this repo consumed it.
- **Performance:** unchanged; `infrahub.database` now imports more modules at startup.

## Checklist

- [x] Tests added/updated
- [ ] [Changelog entry](../.agents/skills/creating-changelog-entries/SKILL.md) added (`uv run towncrier create ...`) - not applicable, internal typing debt
- [ ] External docs updated (if user-facing or ops-facing change) - not applicable
- [ ] Internal .md docs updated (internal knowledge and AI code tools knowledge) - not applicable

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/database/__init__.py
+++ b/backend/infrahub/database/__init__.py
@@ -35,6 +35,7 @@ from infrahub.core.constants import (
     GLOBAL_BRANCH_NAME,
 )
 from infrahub.core.query import QueryType
+from infrahub.core.schema import GenericSchema, NodeSchema
 from infrahub.exceptions import DatabaseError, QueryTimeoutError
 from infrahub.log import get_logger
 from infrahub.utils import InfrahubStringEnum
@@ -50,8 +51,11 @@ from .metrics import (
 if TYPE_CHECKING:
     from types import TracebackType
 
+    # neo4j only exports the concrete trust stores, not the base class the driver accepts.
+    from neo4j._conf import TrustStore
+
     from infrahub.core.branch import Branch
-    from infrahub.core.schema import GenericSchema, MainSchemaTypes, NodeSchema
+    from infrahub.core.schema import MainSchemaTypes
     from infrahub.core.schema.schema_branch import SchemaBranch
 
 validated_database = {}
@@ -98,7 +102,7 @@ class DatabaseSchemaManager:
 
     def get_node_schema(self, name: str, branch: Branch | str | None = None, duplicate: bool = True) -> NodeSchema:
         schema = self.get(name=name, branch=branch, duplicate=duplicate)
-        if schema.is_node_schema:
+        if isinstance(schema, NodeSchema):
             return schema
 
         raise ValueError("The selected node is not of type NodeSchema")
@@ -107,16 +111,17 @@ class DatabaseSchemaManager:
         self, name: str, branch: Branch | str | None = None, duplicate: bool = True
     ) -> GenericSchema:
         schema = self.get(name=name, branch=branch, duplicate=duplicate)
-        if schema.is_generic_schema:
+        if isinstance(schema, GenericSchema):
             return schema
 
         raise ValueError("The selected node is not of type GenericSchema")
 
-    def set(self, name: str, schema: MainSchemaTypes, branch: str | None = None) -> int:
+    def set(self, name: str, schema: MainSchemaTypes, branch: str | None = None) -> None:
         branch_name = get_branch_name(branch=branch)
         if branch_name not in self._db._schemas:
-            return registry.schema.set(name=name, schema=schema, branch=branch)
-        return self._db._schemas[branch_name].set(name=name, schema=schema)
+            registry.schema.set(name=name, schema=schema, branch=branch)
+            return
+        self._db._schemas[branch_name].set(name=name, schema=schema)
 
     def has(self, name: str, branch: Branch | str | None = None) -> bool:
         branch_name = get_branch_name(branch=branch)
@@ -227,7 +232,7 @@ class InfrahubDatabase:
             mode=InfrahubDatabaseMode.SESSION,
             db_type=self.db_type,
             default_neo4j_runtime=self.default_neo4j_runtime,
-            schemas=schemas or self._schemas.values(),
+            schemas=schemas or list(self._schemas.values()),
             driver=self._driver,
             session_mode=session_mode,
             queries_names_to_config=self.queries_names_to_config,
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1624,10 +1623,12 @@ include = ["backend/infrahub/database/**"]
 # The ignored rules below should be removed once the code has been updated, they are included    #
 # like this so that we can reactivate them one by one.                                           #
 ##################################################################################################
+# Queries are assembled at runtime, so they are `str` where the neo4j stubs require `LiteralString`.
+# Also covers the `resolver=None` violation described in the mypy override for infrahub.database.
 invalid-argument-type = "ignore"
-invalid-return-type = "ignore"
+# IndexManagerBase and its subclasses define a `list` method, which shadows the builtin for every
+# `list[...]` annotation in those class bodies. To remove this the method has to be renamed.
```
ogenstad, line 1630: This looked like a simple fix that I left out of scope but it seems like we'd have to rename the method from list to list_index or something similar.

gmazoyer (approved):

backend/infrahub/database/__init__.py
```diff
@@ -292,29 +297,40 @@ async def __aenter__(self) -> Self:
...
         return self
 
+    def _get_open_session(self) -> AsyncSession:
+        if self._session is None:
+            raise RuntimeError("No database session is currently open")
+        return self._session
+
+    def _get_open_transaction(self) -> AsyncTransaction:
+        if self._transaction is None:
+            raise RuntimeError("No database transaction is currently open")
```
ajtmccarty, line 307: I'm wondering if `RuntimeError` is the right way to handle this. it looks like this would be raised while trying to close a database session/txn. on the one hand, it seems weird to raise an error during the close if there is nothing to close, but it _does_ seem like unexpected state to end up in `__aexit__` with no active transaction or session. and the existing code looks like it would have run into this anyway trying to call `close`/`rollback`/`commit` on `None`.

ajtmccarty (approved):

backend/infrahub/database/__init__.py
ogenstad: @ajtmccarty Yeah, I'm not completely sure here either. As you mention today we have `await self._transaction.rollback()` so if `self._transaction` is `None` it would raise an `AttributeError`.

So this is mainly just a typing issue as I wanted to get rid of some of the typing overrides we have in the core db layer.

Would you prefer some other error aside from `RuntimeError` or is it just the fact that it looks weird with an error when we are trying to close it?

ajtmccarty: I think using `RuntimeError` is fine. I was more wondering if it makes sense to raise an error while attempting to close. but it would have broken anyway if it encountered this failure state, so it seems fine for now

PR #10374, opened 2026-08-21
Fix branch scoping that made per-node automations fire twice

gmazoyer:
# Why

A branch whose schema or repository commit differs from the default branch owns its own trigger
automation, and the default-branch automation has to skip that branch. It did so with a list of
negated names on one label: `match["infrahub.branch.name"] = ["!b1", "!b2"]`. Prefect combines the
patterns of one label with OR, and every value satisfies one of two different negations, so the
filter excluded nothing. With two or more divergent branches each event matched both automations
and the work ran twice. A single divergent branch was not affected.

## What changed

Prefect combines the entries of `match_related` with AND, so the exclusion is now one entry per
branch. `EventTrigger.exclude_branches()` builds them against the `infrahub.branch` related
resource that every event already carries, so the events themselves stay unchanged. It sorts the
names, because an automation is reconciled by comparing model dumps and the registry returns
branches in insertion order.

Six builders shared the broken shape: the three computed attribute ones, display labels,
human-friendly ids and profile refresh. The positive match of a branch-scoped automation stays on
the primary resource. A single negation there is still correct, so the branch scoping of action
rules and webhooks is unchanged.

## How to review

Start with `exclude_branches()` in `backend/infrahub/trigger/models.py`, then the six call sites.
`backend/tests/helpers/trigger.py` is worth a look too: the tests assert through Prefect's own
server-side matcher, because the shape of a generated filter says nothing about the events it
selects, which is how the broken filter passed its tests.

## How to test

```bash
uv run pytest backend/tests/unit/trigger backend/tests/unit/event
uv run pytest backend/tests/component/computed_attribute/test_gather.py --neo4j
```

## Impact & rollout

- **Backward compatibility:** no schema, API or event change. Automations reconcile on the next
  setup run.
- **Performance:** measured with the scenario harness. The rebase scenario drops from 148 to 74
  computed attribute flows and the unique target ids stay at 89, so nothing is under-recomputed.
  Numbers are on the ticket.
- **Target branch:** this targets `develop` on purpose, to stay with the rest of the recompute
  parity work. The same pattern exists on `stable`, so we can consider a backport if this proves
  useful.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/unit/trigger/test_branch_scope.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+@dataclass
+class BuilderCase:
+    name: str
+    build: Callable[[str, list[str]], TriggerBranchDefinition]
+
+
+BUILDER_CASES = [
+    BuilderCase(name="computed_attribute_jinja2", build=_jinja2_computed_attribute),
```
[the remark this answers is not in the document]
gmazoyer: Those two builders are covered, just not in this file. `test_gather_trigger_computed_attribute_python_fires_once_per_branch` in `backend/tests/component/computed_attribute/test_gather.py` runs both the owner trigger and the query trigger through the same one-owner-per-branch check, with two branches sitting on different repository commits. They are not unit cases because both need a working query analyzer, so a real schema and a built GraphQL schema, and we do not allow mocks in tests. Keeping them at the component tier also tests the real out-of-scope computation instead of a list I pass by hand.

backend/infrahub/computed_attribute/models.py
```diff
@@ -229,20 +228,21 @@ def from_object(
...
             kind=computed_attribute.computed_attribute.kind
         )
-        event_trigger.match_related = {
+        match_related: dict[str, Any] = {
             "prefect.resource.role": ["infrahub.node.attribute_update", "infrahub.node.relationship_update"],
         }
 
         if update_fields:
-            event_trigger.match_related["infrahub.field.name"] = update_fields
+            match_related["infrahub.field.name"] = update_fields
```
ajtmccarty: are these changes necessary? it looks like there is no functional difference, so it seem fine. but I don't think it should be required if `exclude_branches` accounts for the current state of `match_related` before making any changes

backend/tests/component/computed_attribute/test_gather.py
```diff
@@ -182,8 +184,10 @@ async def test_gather_trigger_computed_attribute_jinja2_different_branch(
     assert set(triggers_by_name.keys()) == {name_main, name_branch_first, name_branch_second}
 
     trigger_main = triggers_by_name[name_main]
-    assert "infrahub.branch.name" in trigger_main.trigger.match
-    assert trigger_main.trigger.match["infrahub.branch.name"] == ["!branch2"]
+    assert "infrahub.branch.name" not in trigger_main.trigger.match
+    assert trigger_main.trigger.match_related[1:] == [
```
ajtmccarty: should this just be `in trigger_main.trigger.match_related` instead of asserting exactly where in the list?

backend/tests/component/profiles/test_models.py
```diff
@@ -74,8 +74,11 @@ async def test_profile_refresh_trigger_definition_with_branches_out_of_scope(
         branches_out_of_scope=["branch1", "branch2"],
     )
 
-    assert "infrahub.branch.name" in trigger.trigger.match
-    assert trigger.trigger.match["infrahub.branch.name"] == ["!branch1", "!branch2"]
+    assert "infrahub.branch.name" not in trigger.trigger.match
+    assert trigger.trigger.match_related[1:] == [
+        {"prefect.resource.role": "infrahub.branch", "infrahub.resource.label": "!branch1"},
+        {"prefect.resource.role": "infrahub.branch", "infrahub.resource.label": "!branch2"},
+    ]
```
ajtmccarty: same here. does ordering really matter? or do we just want to be sure each of these items is in `match_related`?

backend/tests/helpers/trigger.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def automation_covers_event(trigger_definition: TriggerDefinition, event: InfrahubEvent) -> bool:
+    """Report whether the automation built from this definition selects the event.
+
+    The verdict comes from Prefect's own server-side matcher, the one that routes events to
+    automations, because the shape of a generated filter says nothing about what it selects.
+    """
+    server_trigger = PrefectServerEventTrigger.model_validate(trigger_definition.trigger.get_prefect().model_dump())
+    return server_trigger.covers_resources(
```
ajtmccarty, line 44: It looks like `covers_resources` is not called anywhere in our code. is there a method that we actually use that could work here? I'm no Prefect expert, so I could definitely be missing something and maybe this is just the best method available for the test, but it looks like we are testing Prefect here instead of our code

changelog/+ifc3041.fixed.md
```diff
@@ -0,0 +1 @@
+Computed attributes, display labels, human-friendly ids and Profile refresh no longer run twice for a change on a branch that carries its own schema or its own repository commit. The default-branch automation excluded the branches with their own automation through a list of negated branch names, but Prefect combines the patterns of one label with OR, so with two or more such branches the filter excluded nothing and every event matched both automations. The exclusion is now one negation per related-resource filter, which Prefect combines with AND.
```
ajtmccarty: I'm not sure if implementation details belong in the changelog. I think people are mostly just interested in the functional change/fix

ajtmccarty (approved):

changelog/+ifc3041.fixed.md
gmazoyer: Reworded with a user centric approach

backend/infrahub/computed_attribute/models.py
gmazoyer: left over from changes I made then partially reverted.

backend/tests/helpers/trigger.py
gmazoyer: You are right that we never call it, but the Prefect server does, it is what decides if an event matches an automation, so the test checks the filter we build and Prefect is only the judge. There is no client side equivalent for this, the method only exists on the server. Asserting the shape of the filter instead is what hid the bug in the first place, the old test pinned the negated list and passed while the filter excluded nothing.

I made the helper private since only one function uses it, and wrote in the docstring why we go through this method and that it ties us to a server side signature that an upgrade could move (which could be weak, but I don't have another solution yet).

ajtmccarty (approved):

PR #10375, opened 2026-08-21
docs: rework the computed attributes page

yjouffrault:
## Summary

Rework the Computed attributes page: what a computed attribute is, when to define one, which
of Infrahub's derived-value mechanisms to use, and when Infrahub recalculates the value. One
behavior claim on the page was wrong and is corrected.

Pattern: rework of an existing single page. No new pages, no sidebar change, no URL change.

## Content changes

| Section | Change |
|---|---|
| Intro | Rewritten. Defines a computed attribute as a read-only `Text` or `URL` value calculated from modeled data. |
| `## When to use a computed attribute` | New. Four kinds of value that can be calculated from data already modeled. |
| `## Choose the right derived value` | New. Computed attribute versus `display_label` versus `human_friendly_id`. |
| `## Choose between Jinja2 and Python` | Table rebuilt on relationship cardinality, mandatory support, and execution model, replacing "Lightweight" versus "More flexible and powerful". Carries an explicit anchor id — see below. |
| `## When Infrahub recalculates a value` | New, and it corrects the page. |
| `## Jinja2 computed attributes` | Opening and restrictions rewritten; example, YAML, and local-testing content unchanged. |
| `## Python computed attributes` | Opening and restrictions rewritten; the implementation steps are now a four-part list. Examples unchanged. |
| Implementation prose | Tutorial-voice transitions replaced ("In the following example we will create...", "First we prepare a GraphQL query...", "Tie everything together into `.infrahub.yml` file"). |
| `## Related` | New. |
| Limitations callout | Moved below the comparison table it constrains. Wording unchanged. |

## The correction

The page claimed automatic recalculation for Jinja2 only — "Any change to any field used to
compute the value will automatically update it" — and mentioned it for Python only in a
`:::success` callout after the worked example. Both kinds recalculate, and they differ in how
and when:

- Jinja2 inputs on the object itself are recalculated inside the update
  (`Node._update` calls `_recompute_local_jinja2`, `backend/infrahub/core/node/__init__.py:1145`).
  Inputs on a related object are handled by asynchronous automations — stated in that method's
  docstring at `:817`.
- Python computed attributes get one trigger per node kind the Transformation's GraphQL query
  reads (`backend/infrahub/computed_attribute/gather.py:189`), and the value updates when the
  worker task completes.

## Facts verified against the implementation

- A computed attribute must be `Text` or `URL` — `backend/infrahub/computed_attribute/constants.py:11`
- A computed attribute must be marked `read_only`; the schema validator rejects it otherwise —
  `backend/infrahub/core/schema/schema_branch.py:1448`
- Jinja2 templates may reference attributes and cardinality-one relationships only. Every
  template variable is resolved against the schema at load time, and an unknown path or a
  self-reference is rejected then — `backend/infrahub/core/schema/schema_branch.py:1468-1508`
- A Python computed attribute cannot be mandatory — `backend/infrahub/core/schema/schema_branch.py:1510`
- Changing a Python Transformation recomputes the attributes that use it — triggers fire on
  TransformPython create, update, and delete (`backend/infrahub/computed_attribute/triggers.py:37`);
  `backend/infrahub/computed_attribute/tasks.py:633` records that a create or fingerprint change
  recomputes the attributes it feeds.

## Open issues for the reviewer

- [x] **Confirm the Jinja2 synchronous / asynchronous split.** The page says a change to an
  input on the object itself is recalculated as part of the same update, and a change on a
  related object is processed asynchronously. That is what `_recompute_local_jinja2` and its
  docstring describe, read from the code rather than from a spec. The method also skips a
  target that has a pending pool dependency, so "as part of the same update" may need
  qualifying.
- [x] **Confirm the Python granularity.** Triggers are created per node kind the Transformation's
  query reads, so the page says a change to any object of those kinds queues a recomputation.
  If that is filtered further downstream, the sentence is too broad.
- [x] **Check the three rows of `## Choose the right derived value`.** `display_label` also takes
  a Jinja2 template, so the boundary between it and a computed attribute is the row most likely
  to be wrong. The distinction as written: `display_label` is how an object reads in the
  interface, `human_friendly_id` identifies it across systems, and a computed attribute is an
  attribute automation and integrations can query.
- [x] **`restrict_untrusted_jinja2_filters` is undocumented.** It gates which Jinja2 filters a
  computed attribute may use (`backend/infrahub/core/schema/schema_branch.py:1475`). Confirm
  whether it belongs on this page or in the configuration reference.
- [x] **`human_friendly_id` has no page of its own.** Both new links point at
  `schema/nodes-and-attributes.mdx#human-friendly-identifier-hfid`. Confirm that is the right
  destination.

## Out of scope (deferred)

- **A naming-conventions and data-modeling guidance page.** The source material argues that
  infrastructure detail belongs in structured data rather than in name strings, and that
  components should be classified by role, type, and status. That is modeling guidance with no
  home in the docs today; it is recorded on "Docs Revamp — Open Questions + Recommendations"
  under Schema & Data.
- **The `:::success` callout after the Python example** still says a change to a related
  attribute triggers "a reevaluation". Left as-is; the new recalculation section covers the
  behavior, and rewriting the callout would touch the worked example.

## Verification

- `cd docs && npm run build` succeeds with no broken links or anchors.
- The build caught one regression during this work: renaming the heading to "Choose between
  Jinja2 and Python" broke the anchor `#choosing-between-jinja2-and-python`, which
  `docs/docs/release-notes/infrahub/release-1_1_0.mdx:56` links to. The heading keeps an
  explicit anchor id, so the release note resolves without editing a published release note.
- Five other pages link to this one — `artifacts/content-composition.mdx`,
  `development-resources/developer-guide.mdx`,
  `development-resources/graphql/single-target-queries.mdx` (twice), and
  `schema/generics-and-inheritance.mdx`. All point at the page URL, which has not changed;
  `{#jinja-computed-attribute}` and `{#python-computed-attribute}` are preserved.
- Terminology pass: "computed attribute" lowercase, "Python Transformation" capitalised, per
  the terminology rules.
- **Docs lint not run locally.** `vale`, `markdownlint`, and `uv` are unavailable in this
  environment, so `uv run invoke docs.lint` did not run and CI is the first check. New prose was
  checked by hand against the repo's Vale rules: no `config`/`repo` substitutions, no
  `easy`/`simple`/`simply`, sentence-case headings, Oxford commas.


docs/docs/computed-attributes/overview.mdx
```diff
@@ -1,13 +1,45 @@
...
+A computed attribute is a read-only `Text` or `URL` value calculated from data already modeled in Infrahub. Use a computed attribute where that value should stay synchronized with the attributes and relationships it is built from, rather than being entered and maintained separately.
 
-Currently, Infrahub supports two main types of computed attributes:
+Define the calculation in the schema as a Jinja2 template, or reference a Python Transformation where the calculation needs cardinality-many relationships, nested relationships, or logic a template cannot express.
 
-* **Jinja2** – Lightweight, but limited in relationship handling.
-* **Python** – More flexible and powerful, but requires async execution.
+## When to use a computed attribute
+
+Computed attributes fit values that can be calculated from data you already model:
```
wvandeun: Computed attributes fit values that can be calculated from other data in Infrahub.

docs/docs/computed-attributes/overview.mdx
```diff
@@ -1,13 +1,45 @@
 ---
 title: Computed attributes
+description: What a computed attribute is, when to use it, and how Infrahub recalculates it as your data changes.
 ---
 
-The computed attributes feature enables the dynamic calculation of attribute values based on user-defined logic.
+A computed attribute is a read-only `Text` or `URL` value calculated from data already modeled in Infrahub. Use a computed attribute where that value should stay synchronized with the attributes and relationships it is built from, rather than being entered and maintained separately.
 
-Currently, Infrahub supports two main types of computed attributes:
+Define the calculation in the schema as a Jinja2 template, or reference a Python Transformation where the calculation needs cardinality-many relationships, nested relationships, or logic a template cannot express.
```
wvandeun: Not really sure this sentence add anything here. I actually liked what was here before.

BaptisteGi (approved):

PR #10376, opened 2026-08-21
fix(backend): pull the mapped remote branch when syncing a repository

fatih-acar:
## Summary

Repositories whose git default branch differs from the Infrahub default branch could be flipped to the `error` operational status by a routine background sync, even though nothing was actually wrong with the repository. This surfaced as the `backend-docker-integration` flake in `test_artifact_composition.py::test_section_artifacts` (`assert 'error' == 'online'`, [run 32490367256](https://github.com/opsmill/infrahub/actions/runs/32490367256/job/96796754359)), but it affects any multi-worker deployment using such a repository.

## Root cause

When a task worker processes a `refresh-git-fetch` broadcast without a pinned commit (the documented fallback when the publishing worker cannot resolve one), it falls back to `repo.pull()`. `pull()` passed the Infrahub branch name straight to `git pull origin <name>` — but when the repository's default branch is not the Infrahub one (the test repo uses `production`), the remote has no such ref, so the pull failed with `fatal: couldn't find remote ref main` and the error handler marked the repository `error`.

The flakiness came from the trigger, not the bug: the broken pull only runs when the pinned-commit resolution loses a race on the publishing worker; most runs take the `reset_to_commit` path and never hit it.

## Key Changes

- `pull()` now resolves the remote branch through `_get_mapped_remote_branch()` — the same mapping `push()` already uses for this exact case. The mapping is identity for every other caller: git-native branch names pass through unchanged, and a remote branch literally named after the Infrahub default is already rejected by `validate_remote_branch` when the defaults differ.
- New unit test reproducing the failure: it fails before the fix with the exact CI error (`fatal: couldn't find remote ref main`) and passes after.

## Documentation Updates

- `dev/knowledge/backend/git-sync.md` now documents the Infrahub-to-remote branch mapping alongside the existing remote-to-Infrahub one.

## Test Plan

- `uv run pytest backend/tests/unit/git/` — 199 tests, all passing (including the new reproduction test).
- `ruff` and `mypy` clean on the touched files; `docs.lint` clean.
- The flaked docker-integration test (`test_artifact_composition.py`) exercises this path end-to-end in CI; the failure was timing-dependent, so the unit test pinning the exact failing git command is the regression guard.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10376?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/git/base.py
+++ b/backend/infrahub/git/base.py
@@ -963,7 +963,7 @@ class InfrahubRepositoryBase(BaseModel, ABC):
         if repo is not None:
             try:
                 commit_before = str(repo.head.commit)
-                repo.remotes.origin.pull(branch_name)
+                repo.remotes.origin.pull(self._get_mapped_remote_branch(branch_name=branch_name))
             except GitCommandError as exc:
                 await self._raise_enriched_error(error=exc, branch_name=branch_name)
 
--- a/backend/tests/unit/git/test_git_repository.py
+++ b/backend/tests/unit/git/test_git_repository.py
@@ -206,6 +206,46 @@ async def test_init_repoints_origin_after_location_change(
     assert relocated.get_branches_from_remote()["main"].commit == commit_b
 
 
+async def test_pull_infrahub_default_branch_pulls_repository_default_branch(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Pulling the Infrahub default branch must pull the repository's own default branch.
+
+    When the two differ, the remote has no branch named after the Infrahub default branch,
+    so pulling the unmapped name fails and flips the repository into an error state.
+    """
+    repos_dir = tmp_path / "repositories"
+    repos_dir.mkdir()
+    monkeypatch.setattr(registry, "_default_branch", "main")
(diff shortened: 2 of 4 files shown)


backend/infrahub/git/base.py
```diff
@@ -963,7 +963,10 @@ async def pull(
         if repo is not None:
             try:
                 commit_before = str(repo.head.commit)
-                repo.remotes.origin.pull(branch_name)
+                # Pull the mapped remote branch, not the bare branch name: when the repository's
+                # default branch is not the Infrahub default, the remote has no branch named after
+                # the Infrahub one.
```
ajtmccarty: I don't think the comment is necessary

ajtmccarty (approved):

backend/infrahub/git/base.py
fatih-acar: Removed — the method name says it, and the reasoning now lives in `dev/knowledge/backend/git-sync.md`. Also squashed the branch down to a single commit.

PR #10377, opened 2026-08-21
test(backend): keep the Prefect log flush out of httpx-mocked component tests

fatih-acar:
## Summary

Fixes the `backend-tests-component (other)` flake seen in [run 32490367256](https://github.com/opsmill/infrahub/actions/runs/32490367256/job/96796754341): `test_askpass_password` failed with `--- Error logging to API ---` in its captured stderr plus a teardown error about an unexpected `POST http://127.0.0.1:17339/api/logs/`.

## Root cause

Each xdist worker runs a session-scoped ephemeral Prefect server, and Prefect's `APILogWorker` background thread flushes logs batched by **earlier** tests on its own interval. `pytest_httpx` patches the httpx transports process-wide, so a flush landing inside a mocked test window fails to match any registered response — Prefect prints the error banner into the CliRunner-captured stderr (failing `assert not result.stderr`) and the fixture's unmatched-request teardown assertion trips. Any component test using `httpx_mock` without a `should_mock` guard can lose this race.

## Key Changes

- Component tests that use `httpx_mock` now default to intercepting only the fake `http://mock` SDK host, so Prefect server traffic passes through to the real transport. The default lives in one place — a `pytest_collection_modifyitems` hook in `backend/tests/component/conftest.py` — so there is nothing to remember when a new module starts using `httpx_mock`.
- Tests that declare their own `httpx_mock` marker keep it untouched (`test_refresh_git_fetch.py`, `test_sync_new_branch`); the hook only adds the default when no marker exists, since a hook-added function-level marker would otherwise outrank an explicit module-level one. It is path-scoped to the component tree because a sub-directory conftest sees every collected item.
- The per-module markers are removed, including the pre-existing one in `test_git_repository.py`.

## Test Plan

- Deterministic repro validated locally: emitting a log through `APILogHandler` and forcing the worker flush inside the test window reproduces both CI failure modes exactly with no guard, and passes with the conftest default alone (no module marker).
- Collection probe confirms every marker-less component `httpx_mock` user resolves to the conftest default, and tests with explicit markers resolve to exactly their own.
- `git_credential`, `message_bus/operations/requests` and `git/test_git_repository.py` pass locally (73 passed, 1 xfailed).

### DIFF
--- a/backend/tests/component/git_credential/test_git_askpass.py
+++ b/backend/tests/component/git_credential/test_git_askpass.py
@@ -1,13 +1,7 @@
-import pytest
 from pytest_httpx import HTTPXMock
 from typer.testing import CliRunner
 
 from infrahub.git_credential.askpass import app
-
-# Only intercept the fake SDK endpoint: the per-worker Prefect server flushes
-# batched API logs on its own schedule, and a flush landing mid-test must reach
-# the real server instead of failing inside the mock and polluting stderr.
-pytestmark = pytest.mark.httpx_mock(should_mock=lambda request: request.url.host == "mock")
 
 runner = CliRunner()
 
--- a/backend/tests/component/git_credential/test_git_askpass.py
+++ b/backend/tests/component/git_credential/test_git_askpass.py
@@ -1,7 +1,13 @@
+import pytest
 from pytest_httpx import HTTPXMock
 from typer.testing import CliRunner
 
 from infrahub.git_credential.askpass import app
+
+# Only intercept the fake SDK endpoint: the per-worker Prefect server flushes
+# batched API logs on its own schedule, and a flush landing mid-test must reach
+# the real server instead of failing inside the mock and polluting stderr.
+pytestmark = pytest.mark.httpx_mock(should_mock=lambda request: request.url.host == "mock")
 
 runner = CliRunner()
 
--- a/backend/tests/component/conftest.py
+++ b/backend/tests/component/conftest.py
@@ -85,6 +85,27 @@ from tests.helpers.test_client import dummy_async_request
 from tests.helpers.utils import find_available_prefect_port
 from tests.test_data import dataset01 as ds01
 
+COMPONENT_TESTS_DIR = Path(__file__).parent
+
+
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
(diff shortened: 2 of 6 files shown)


backend/tests/component/git_credential/test_git_askpass.py
```diff
@@ -1,8 +1,14 @@
+import pytest
 from pytest_httpx import HTTPXMock
 from typer.testing import CliRunner
 
 from infrahub.git_credential.askpass import app
 
+# Only intercept the fake SDK endpoint: the per-worker Prefect server flushes
+# batched API logs on its own schedule, and a flush landing mid-test must reach
+# the real server instead of failing inside the mock and polluting stderr.
+pytestmark = pytest.mark.httpx_mock(should_mock=lambda request: request.url.host == "mock")
```
ajtmccarty: I'm not sure exactly how this works, but could these be consolidated somewhere? seems like it would be easy to forget if adding another test module that uses `httpx_mock`, although that should be pretty rare going forward

fatih-acar: Good call — folded in (15c776b7c).

`pytest_httpx` only reads its options from `httpx_mock` markers (closest marker wins; there's no config-level default), so the consolidation is a `pytest_collection_modifyitems` hook in `backend/tests/component/conftest.py`: every component test that requests `httpx_mock` and has no `httpx_mock` marker of its own gets `should_mock=lambda r: r.url.host == "mock"`. Tests with an explicit marker (`test_refresh_git_fetch.py`, `test_sync_new_branch`) keep theirs — that guard matters because a hook-added function-level marker would otherwise outrank an explicit module-level one. It's path-scoped to the component tree since a sub-directory conftest sees every collected item. The per-module markers are gone, including the pre-existing one in `test_git_repository.py`.

For the "how this works" part: the component harness runs a per-worker ephemeral Prefect server, and Prefect's `APILogWorker` thread flushes batched logs on its own schedule. `pytest_httpx` patches the httpx transports process-wide, so a flush landing inside a mocked test used to fail in the mock (stderr banner + unmatched-request teardown error). `should_mock` is evaluated per request, so restricting it to the fake SDK host lets that traffic through to the real server.

ajtmccarty (approved):

Issue #9762, opened 2026-06-30
bug: branch-agnostic attribute values are not tombstoned on node deletion, breaking uniqueness validation


lancamat1:
**Component:** API Server / GraphQL

**Infrahub version:** 1.10.0 (Neo4j 2025.10.x, SDK 1.22.0)

## Current Behavior

A `branch: agnostic` attribute does not have its value tombstoned when the owning node is deleted, which leaves orphaned "active" values that break uniqueness-constraint validation.

Observed with a `kind: NumberPool`, `read_only: true`, `branch: agnostic` attribute defined on a generic and inherited by several concrete kinds:

- A branch-agnostic attribute stores its value edge on the `-global-` branch. When the owning node is deleted on a branch (e.g. `main`), the node's existence is tombstoned at branch level — `(:Node)-[:IS_PART_OF]->(:Root)` gets a `main` edge with `status: deleted` — **but the branch-agnostic `HAS_VALUE` edge on `-global-` stays `status: active`** (never closed). So the deleted node keeps a "live" attribute value.
- Over repeated create/delete cycles (e.g. data re-imports) these orphaned values accumulate.
- During schema-integrity / uniqueness-constraint validation (e.g. when running a **data-only** proposed change such as a description edit on an unrelated node), these orphaned values are still scanned. The PC fails its uniqueness check, reporting conflicts on attribute values whose node UUIDs no longer resolve — `{ <Kind>(ids: ["..."]) { count } }` returns `0` and the UI shows "node not found".
- The same orphaned values also **block schema migrations that introduce a uniqueness constraint** on that attribute: the new `uniqueness_constraints` cannot load because the dead-node values appear as duplicate violations.

Quantified on one real dataset (a NumberPool attr on a generic with 3 child kinds): ~512 effectively-live nodes, but **~6,400 effectively-deleted nodes still holding an `active` `-global-` value** for that attribute. 13,378 active `-global-` value edges across 6,949 node objects for only 1,139 distinct values — e.g. a single value attached to 32 distinct (mostly deleted) node objects.

## Expected Behavior

- When a node is deleted, its branch-agnostic attribute values should be tombstoned/closed consistently, so they are no longer considered live.
- Uniqueness-constraint validation (and schema-migration constraint checks) should not scan attribute values belonging to nodes that are effectively deleted in the relevant branch view.

## Steps to Reproduce

[...]

PR #10381, opened 2026-08-22
fix(backend): rollback branch-agnostic fields on branch-aware objects correctly

ajtmccarty:
should follow #10358 

# Why

The rollback cypher queries that we use to clean up a failed merge or rebase operation does not
correctly handle cleaning up branch-agnostic fields on branch-aware objects.

This updates the rollback to correctly check the global branch for changes with the exact timestamp
being rolled back AND also undo metadata `updated_at/by` properties on changed Node, Attribute, and
Relationship vertices.

Non-goals: the upstream gap where a user-branch schema removal closes global edges without bumping
metadata at all (tracked in IFC-3032)

Part of #9762 (slice 2a of the agnostic-edge retirement plan, `dev/specs/ifc-2843-retire-agnostic-edges/`).

## What changed

Behavioral:

- The metadata restore matches the operation's **exact timestamp on every pass**, independent of
  branch and scope. The justification is an invariant the rollback already leans on for the global
  branch: every write of one operation lands at exactly the operation's `at`, so a vertex bumped at
  any other time — even inside a range-scoped window — belongs to another writer. A possible example
  is :
    - Merging Branch A begins
    - Update on Branch B alters a branch-agnostic field on an object included in the Branch A merge
    - Now the object has a updated_at/by AFTER the start of the merge operation
    - Merging Branch A fails and rollback begins
    - the branch-agnostic field on the object is left with its correct update time from the Branch
      B change
- The restore is now **unconditional**: the `restore_metadata` flag, its default/global-branch
  target guard, and the flag's three call sites are deleted. The visible flip is the schema-update
  coordinator, which rolled back with `restore_metadata=False` — a failed standalone schema update
  now restores the stamps its migrations wrote, the same way the same removal rolled back through a
  merge always did. On targets where the operation bumped nothing, the exact match makes the
  restore a correct no-op, which is what made the guard unnecessary.

Implementation notes:

- The pipeline no longer depends on the per-branch rollback row, and dedups its vertex set
  (`WITH DISTINCT restore_vertex`) so several edges in one batch restore a shared vertex once.
- The edge passes are untouched: target branch follows the scope, global branch always exact, and
  the restore-before-delete crash-safety ordering is preserved.
- The same-microsecond collision (an unrelated write landing at exactly `$at`, extremely unlikely)
  is accepted and documented at `_rollback_branches` — the same operation-identity assumption
  `AT_TIMESTAMP` scope already makes.

What stayed the same: no schema change, no API/GraphQL surface, no migration. Only the internal
`GraphRollbacker.rollback` signature changed, and all callers are updated in this PR.

## How to review

- `backend/infrahub/core/query/rollback.py` is the substance: `_render_restore_metadata_pipeline`
  and the two query bodies. The docstrings carry the reasoning.

## Impact & rollout

- **Backward compatibility:** internal-only signature change (`GraphRollbacker.rollback` loses
  `restore_metadata`); no persisted-data or API change. Failed standalone schema updates now
  restore vertex metadata where they previously left it bumped.
- **Performance:** the restore now runs on every rollback rather than optionally, but rollbacks are
  failure-path operations and the restore is bounded by the matched edge set with an exact-match
  filter.
- **Deployment notes:** safe to deploy; no migration, no coordination.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — carried at the feature level (task T051 in the spec's task list), not per slice
- [ ] External docs updated — no user-facing surface in this slice
- [x] Internal .md docs updated (`dev/specs/ifc-2843-retire-agnostic-edges/` plan and tasks)

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10381?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/component/core/test_rollback.py
+++ b/backend/tests/component/core/test_rollback.py
@@ -5,15 +5,25 @@ from typing import TYPE_CHECKING
 
 import pytest
 
-from infrahub.core.branch import Branch
+from infrahub.core import registry
+from infrahub.core.constants import GLOBAL_BRANCH_NAME
 from infrahub.core.initialization import create_branch
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
 from infrahub.core.query.rollback import RollbackScope
 from infrahub.core.rollback import GraphRollbacker
 from infrahub.core.timestamp import Timestamp
+from tests.helpers.agnostic_edges import (
+    VertexMetadata,
+    attribute_global_edges,
+    attribute_metadata,
+    edge_summary,
+    open_edges,
+)
+from tests.helpers.schema.agnostic_retirement import AGNOSTIC_RETIREMENT_SCHEMA, WIDGET_KIND
 
 if TYPE_CHECKING:
+    from infrahub.core.branch import Branch
     from infrahub.core.schema.schema_branch import SchemaBranch
     from infrahub.database import InfrahubDatabase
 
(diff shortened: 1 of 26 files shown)


backend/tests/component/core/test_rollback.py
polmichel, line 1: Great test suite. Optional comment:
maybe would it be worth considering splitting this file into subfiles? For instance generic rollback mechanism versus agnostic fields specific cases?

polmichel (approved):
LGTM.
One optional comment about splitting a test file and a independent comment about the following commit 758a99ab2f985c2c5e3d44a560d8b913e28089e0 which looks duplicated on the previous [PR](https://github.com/opsmill/infrahub/pull/10358) with a73d1df9f9da4cf44bf109b74b2da390c3a082b2 SHA1.

backend/tests/component/core/test_rollback.py
ajtmccarty: I suppose it is a little too big. I'll divide it in the next PR

PR #10382, opened 2026-08-23
retire branch-agnostic fields during merge/rebase

ajtmccarty:
# Why

Overall goal of this stack of PRs is to correctly close branch-agnostic fields on branch-aware objects when the object cannot be reached from any branch. This slice handles retiring/closing/deleting these fields when they become unreachable as part of a merge or a rebase.

A merge example:
- create object with branch-agnostic field on main
- create a branch
- delete the object on the branch (it is still accessible from main)
- merge the branch. now the object is completely unreachable, so its branch-agnostic field must be closed

A rebase example:
- create object with branch-agnostic field on main
- create a branch
- delete the object on main (it is still accessible on the branch)
- rebase the branch. now the object is completely unreachable, so its branch-agnostic field must be closed

Non-goals: branch-deletion enforcement, the repair migration for pre-existing leaks, and removal of the superseded retirement stack — all later slices of the spec (`dev/specs/ifc-2843-retire-agnostic-edges`).

Part of IFC-2843.

## What changed

**Behavioral**

- **Merge** — after the bulk merges and metadata updates, `DiffMerger.merge_graph` reads the merge diff's removed-node uuids and re-evaluates retention for them at the merge timestamp in batches of 500. A failure propagates into the existing `MergeFailedError` path, and the merge rollback reopens any closes (they land exactly at the merge `at`, which the rollbacker's global-exact pass restores).
- **Rebase** — `rebase_branch` reads the **base-branch** deletions via a second `DiffRepository` read under the branch's existing tracking id, and runs the closures **inside the rebase transaction**, so a retirement failure rolls back the entire rebase. A no-schema-diff rebase never reaches the schema-update coordinator, so transaction atomicity is the only rollback cover there.
- Both points log candidates and `edges_closed` per batch; candidates with zero closes is the visible signature of a deferral (another branch still retains).

**Implementation notes**

- `RetireNodeAgnosticFieldsQuery` now takes `node_uuids: list[str]` instead of a single uuid; the node-delete call site passes a one-element list.
- Rather than adding a parallel removed-nodes-only repository method, `DiffRepository.get_affected_node_uuids` gained optional `include_actions` / `exclude_actions` (`DiffAction`) filters — the default preserves the old exclude-`unchanged` behavior; the retirement call sites pass `include_actions=[DiffAction.REMOVED]`. Its query also dropped `at` / `target_branch` parameters that were bound but never used in the Cypher.
- The rebase closures run **after** `user_branch.rebase(...)` within the transaction. The retention predicate reads each branch's window from its `(:Branch)` vertex, so only once `branched_from` moves past the deletions does the rebased branch stop retaining them. If run before, then the enforcement point would be a permanent no-op.

**Tests** (all component, driving the real merge/rebase flows)

- Merge: closure of the last holder's deletion; deferral while another branch retains; a two-node mixed batch proving the predicate is evaluated per candidate, not in aggregate.
- Rebase: closure once the fork point passes the deletion; deferral while a third branch retains; a branch-local create+delete lifecycle asserting nothing is left open or orphaned (post-rebase reads are anchored on the field vertices' own uuids, since node-anchored reads go blind on exactly the orphan shape at issue); a retirement failure rolling back the whole rebase (fails if the call is moved outside the transaction).
- Diff layer: `test_aware_node_agnostic_fields.py` (placed with the diff-calculator tests) pins that retirement's `to` stamps on global edges never surface as changes in the diff of a branch that forked before the object existed — the diff traversal does include global edges, so this is not true by construction.

**What stayed the same**: no schema changes, no API changes, the retention predicate and the node-delete/schema-removal enforcement points are untouched, and `get_affected_node_uuids`'s default filtering behavior is unchanged for its existing metadata-update caller.

## How to review

- `backend/infrahub/core/branch/tasks.py` — the transaction placement and after-rebase ordering are the load-bearing decisions.
- `backend/infrahub/core/diff/merger/merger.py` — merge wiring and failure propagation.
- `backend/infrahub/core/diff/query/affected_diff_nodes.py` + `repository.py` — the filter refactor; check the default keeps the metadata caller's behavior.
- `dev/specs/ifc-2843-retire-agnostic-edges/tasks.md` is spec bookkeeping only.

## Impact & rollout

- **Backward compatibility:** none broken; no migrations.
- **Performance:** each merge/rebase adds one indexed diff read plus one retirement query per 500 recorded deletions; the spec's timing gate (T038) is measured with the branch-deletion slice.
- **Deployment notes:** safe to deploy; behavior only activates on merges/rebases that carry deletions of nodes with branch-agnostic fields.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added — deferred to the final IFC-2843 slice, which lands the user-visible fix
- [ ] External docs updated — not user-facing on its own
- [x] Internal .md docs updated (`dev/specs/ifc-2843-retire-agnostic-edges/tasks.md`)

---
## Summary by cubic
Retires branch-agnostic fields when deletions land via merge or rebase, preventing orphaned global values that break uniqueness (IFC-2843). Previously retirement ran only at delete; now merges and rebases re-evaluate candidates and close global edges only when no branch still retains the node.

- Merge: after bulk merges and metadata updates, `DiffMerger.merge_graph` collects `REMOVED` node UUIDs from the source diff and batches `RetireNodeAgnosticFieldsQuery` at the merge timestamp (`metadata_batch_size`); failures fail the merge.
- Rebase: `rebase_branch` reads base-branch deletions under the branch’s tracking id and runs retirement after `user_branch.rebase(...)` inside the same transaction (`RETIREMENT_BATCH_SIZE=500`); failures roll back the rebase.
- Repository/query: `RetireNodeAgnosticFieldsQuery` now accepts `node_uuids: list[str]`; `Node.delete` passes a list. `DiffRepository.get_affected_node_uuids` now takes `diff_branch_name` plus optional `include_actions`/`exclude_actions` (defaults exclude `UNCHANGED`); `AffectedDiffNodeUUIDsQuery` implements these filters.
- Tests: add merge/rebase enforcement coverage and a diff check that global-edge closes never surface as changes on branches that never saw the node; consolidate shared assertions in `tests.helpers.agnostic_edges`.

<sup>Written for commit a08b5dc99982664cc6d57fc9ed297f5b28fe05de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/diff/merger/merger.py
```diff
@@ -125,8 +125,42 @@ async def merge_graph(self, at: Timestamp) -> None:
...
+        log.info(
+            "Re-evaluating branch-agnostic retirement for merged deletions",
+            candidates=len(deleted_node_uuids),
+        )
+        if deleted_node_uuids:
+            await self._retire_agnostic_fields_of_deleted_nodes(node_uuids=deleted_node_uuids, at=at)
+
         log.info("Graph merge complete")
 
+    # TODO: not sure the retry_db_transaction decorator does anything b/c this is not running in a transaction
```
ajtmccarty, line 142: my bad. this is removed in the next slice. not worth rerunning all the tests for something that is going to end up squashed before going into `stable`

backend/tests/component/core/test_agnostic_retirement.py
```diff
@@ -427,3 +466,198 @@ async def test_a_value_freed_by_retirement_is_allocatable_again_from_its_pool(
...
+
+    @pytest.fixture(scope="class")
+    async def agnostic_schema(
+        self,
+        db: InfrahubDatabase,
+        default_branch: Branch,
+        register_core_models_schema_scope_class: SchemaBranch,
+    ) -> None:
+        """The rebase flow resolves core kinds while it diffs and validates, so the core schema rides along."""
+        registry.schema.register_schema(schema=AGNOSTIC_RETIREMENT_SCHEMA, branch=default_branch.name)
```
polmichel, line 642: Not that important, but this looks duplicated in other test classes

backend/tests/component/core/test_agnostic_retirement.py
```diff
@@ -427,3 +466,198 @@ async def test_a_value_freed_by_retirement_is_allocatable_again_from_its_pool(
...
+        assert {edge.status for edge in attribute_after} == {"active"}, (
+            "retirement is a time-close, never a status tombstone"
+        )
+        relationship_after = await relationship_global_edges(
+            db=db, node_id=widget.id, identifier=RELATIONSHIP_IDENTIFIER
+        )
+        assert open_edges(relationship_after) == [], (
+            "no branch reads both peers as live once the rebase lands, so the relationship goes with it"
+        )
+        assert to_times(relationship_after) == {rebase_at.to_string()}
```
polmichel: This also looks like duplication, not mandatory at all but wdyt about additional test helpers?
```
  def assert_attribute_retired_at(after: list[EdgeState], before: list[EdgeState], at: Timestamp) -> None:
      """The attribute's global edges were time-closed at `at`, never tombstoned."""
      assert edge_summary(after) == expected_closed_at(before, at)
      assert {edge.status for edge in after} == {"active"}, "retirement is a time-close, never a status tombstone"

  def assert_relationship_retired_at(after: list[EdgeState], at: Timestamp) -> None:
      """No branch reads both peers live once the operation lands, so the relationship closes at `at`."""
      assert open_edges(after) == []
      assert to_times(after) == {at.to_string()}
```

I think it would also increase the readability

polmichel (approved):
Also LGTM, a few optional comments regarding the test suite.

The commits split made the review very easy to make, thanks.

backend/tests/component/core/test_agnostic_retirement.py
ajtmccarty: good idea. added throughout the tests

PR #10387, opened 2026-08-24
feat(schema): add commit attribute to the generic repository [IFC-2954]

ogenstad:
## The short version

The CoreGenericRepository doesn't have a `commit` attribute so we always have to either have type ignores when we access `.commit` as both of the actual repository types carry this attribute or we need to use fragments when we issue a query. This is a small change to simplify the code and clean up some typing around repositories.

The SDK repo in https://github.com/opsmill/infrahub-sdk-python/pull/1272 will probably be merged first and then we'll update the SDK commit in this branch to the latest `infrahub-develop` commit.

## What

`CoreGenericRepository` now declares a `commit` attribute (`Text`, optional, `branch=local`, `order_weight=7500`), so a repository's current commit can be read without first narrowing to `CoreRepository` or `CoreReadOnlyRepository`.

Both concrete kinds keep their own local `commit` definition and branch scope - `local` for `CoreRepository`, `aware` for `CoreReadOnlyRepository`. A locally-defined attribute takes precedence over the inherited one, so neither kind changes behaviour.

## Why `branch=local` on the generic

Nothing reads the generic's value at runtime, so this is a documentation choice - but it is not a free one. Omitting `branch` does not leave it unset: `process_branch_support` fills a `None` branch with the node's own branch, and the generic is `agnostic`, which would assert that `commit` is identical across branches. That is false for both implementers, and `agnostic` is the one value with real semantics elsewhere (`core/attribute.py` special-cases it). `local` matches `CoreRepository`, the managed kind, and is the conservative value.

## Verifying the concrete kinds are untouched

The processed core schema was built four ways - without the generic attribute, and with `branch` omitted / `local` / `aware` - and all 100 kinds compared. In every variant the only kind that differs from baseline is `CoreGenericRepository` itself; `CoreRepository` and `CoreReadOnlyRepository` are byte-identical, keeping `branch: local` / `branch: aware` and `order_weight: 7000`.

Why it holds:

- `inherit_from_interface` skips any generic attribute already in the node's `valid_local_names`, so the generic's copy is never applied to either kind
- `process_branch_support` only rewrites attributes flagged `inherited`
- every `Node` is instantiated via `find_node_schema`, which returns only `NodeSchema`/`ProfileSchema`/`TemplateSchema` and skips generics - so the branch-support reads in `core/attribute.py` always see the concrete kind's attribute
- the GraphQL resolvers consult branch support only for relationships, never attributes
- the one consumer of the generic's value is the `node.attribute.add` migration the core-schema update emits; its Cypher only creates an attribute where none exists (`has_attr_e IS NULL OR has_attr_e.status = "deleted"`), and every repository node already has `commit`, so it is a no-op

## Workarounds removed

- five `# type: ignore[attr-defined]` on `.commit` access (`api/transformation.py` x4, `api/file.py` x1) plus the comment in `api/file.py` describing the gap
- the inline fragments in `computed_attribute/graphql_queries/transform_fetch.gql` and in `DESTINATION_ALLREPOSITORIES` (`proposed_change/tasks.py`), which existed only to reach `commit` through the concrete kinds. Collapsing the former drops a three-member discriminated union from the generated types (-35 lines) and turns `getattr(repo, "commit", None)` into direct attribute access

The three ignores in `core/diff/branch_differ.py` are **not** removable and are left alone - see below.

## Regenerated

`backend/infrahub/core/protocols.py`, `backend/tests/protocols.py`, `python_sdk/infrahub_sdk/protocols.py`, `schema/schema.graphql`, the `.gql` pydantic types, and the frontend gql.tada files + `types.ts`. `schema/openapi.json` and `docs/` are unaffected.

The `schema.graphql` diff is 55 insertions and 0 deletions: `commit: TextAttribute` on the interface, `commit: TextAttributeUpdate` on `CoreGenericRepositoryUpdateInput`, and `commit__*` filters on the generic plus nested `repository__commit__*` on the ten kinds that relate to it.

## Note on `CoreGenericRepositoryUpdate`

Generics with attributes get an update mutation (`graphql/manager.py`), built on plain `InfrahubMutation` rather than `InfrahubRepositoryMutation`. `commit` therefore becomes settable through `CoreGenericRepositoryUpdate`, which does not fire the read-only-repository import that `CoreReadOnlyRepositoryUpdate` triggers. Accepted deliberately: that mutation already writes `location`, `name`, `internal_status`, and `sync_status` with no repository-specific handling. Separately, the trigger in `mutations/repository.py:122` is already skipped for `commit: {value: null}` on the concrete kind - worth its own ticket.

## Two findings for follow-up

1. **The ticket overcounts the type-ignores.** Five of the eight are removable, not eight. The three in `core/diff/branch_differ.py:206,213,214` are on objects typed `Node`, not `CoreGenericRepository`, because the query passes the kind as a string (`schema=InfrahubKind.GENERICREPOSITORY`) and so selects the untyped overload. Different root cause, untouched by this change.
2. **Fixing those would surface a latent bug.** Passing the protocol class instead trades the three `attr-defined` errors for `Argument "commit_from" ... has incompatible type "str | None"; expected "str"`. It is reachable: the guard at line 206 only skips when both commits are *equal*, so a repository with no commit at `diff_from` and one at `diff_to` passes `None` into `GitDiffNamesOnly(first_commit=...)` and fails pydantic validation at runtime. Fixing it properly means widening `FileDiffElement.repository` (an API-serialised model) and deciding what a null commit should mean.

## Testing

- new unit test in `tests/unit/core/schema/test_schema_branch.py`: a node's locally-defined attribute keeps its own branch support when the generic declares the same attribute with a different one. The existing `test_schema_branch_process_branch_support` only covers the *inherited* path, so this override path was untested - and it is the invariant this change depends on
- `test_transform_query.py` payload updated: `commit` is now a required key in the regenerated model, so the unsupported-kind case has to include it
- `backend.mypy` clean (1630 files), `backend.ty` clean, `backend.ruff` clean
- backend unit suite 2339 passed; frontend 1222 passed
- all four repository GraphQL queries validate against the new schema; the collapsed `DESTINATION_ALLREPOSITORIES` is *invalid* against the pre-change schema, confirming the check is not vacuous
- frontend `tsc` error count identical before and after (197, same set - all pre-existing)

## Merge order

Depends on opsmill/infrahub-sdk-python#1272. `backend.validate-generated` runs `git -C python_sdk diff --exit-code infrahub_sdk/protocols.py`, so it stays red until the pointer includes that commit. The second commit here points `python_sdk` at the head of that PR's branch to get CI green during review; it must be re-pointed at the merge commit before this merges.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10387?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gmazoyer (approved):

ajtmccarty (approved):
much cleaner

PR #10396, opened 2026-08-24
fix(backend): branch registry updates

ajtmccarty:
two or three small fixes
1. update branch registry during a BranchUpdate mutation once it succeeds
2. update branch registry in the `migrate_branch` task when branch state is saved to the database
3. if an error is raised during the `refresh_branches` task, log the exception and continue, so that other branches can still be refreshed

the first two show the need for a `BranchSaver` component that would handle making the change to the database and then updating the registry so that they cannot get out of sync

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -101,6 +101,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
             log.info(f"No migrations detected for branch '{obj.name}'")
             obj.graph_version = GRAPH_VERSION
             await obj.save(db=db)
+            registry.branch[obj.name] = obj
             return
 
         # Branch status will remain as so if the migration process fails
@@ -108,6 +109,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
         if obj.status != BranchStatus.NEED_UPGRADE_REBASE:
             obj.status = BranchStatus.NEED_UPGRADE_REBASE
             await obj.save(db=db)
+            registry.branch[obj.name] = obj
 
         try:
             log.info(f"Running migrations for branch '{obj.name}'")
@@ -120,6 +122,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
             obj.status = BranchStatus.OPEN
         obj.graph_version = GRAPH_VERSION
         await obj.save(db=db)
+        registry.branch[obj.name] = obj
 
     if send_events:
         event_context = context.to_event_context()
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -101,7 +101,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
             log.info(f"No migrations detected for branch '{obj.name}'")
             obj.graph_version = GRAPH_VERSION
             await obj.save(db=db)
-            registry.branch[obj.name] = obj
+            registry.refresh_cached_branch(obj)
             return
 
         # Branch status will remain as so if the migration process fails
@@ -109,7 +109,7 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
         if obj.status != BranchStatus.NEED_UPGRADE_REBASE:
             obj.status = BranchStatus.NEED_UPGRADE_REBASE
             await obj.save(db=db)
-            registry.branch[obj.name] = obj
+            registry.refresh_cached_branch(obj)
 
         try:
             log.info(f"Running migrations for branch '{obj.name}'")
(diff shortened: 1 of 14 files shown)


backend/tests/component/tasks/test_registry.py
```diff
@@ -0,0 +1,51 @@
...
+    with caplog.at_level("ERROR", logger="infrahub"):
+        await refresh_branches(db=db)
+
+    # The branch it gave up on has to be reported, with its traceback: absorbed is not silent
+    failures = [
+        record.msg
+        for record in caplog.records
+        if isinstance(record.msg, dict)
+        and record.msg.get("event") == f"Failed to refresh branch '{broken_branch.name}' in the registry"
+    ]
```
gmazoyer: Could this reuse `find_logged_event` from `component/core/merge/conftest.py`? It does the same `record.msg` dict scan. Maybe worth moving it somewhere shared.

backend/infrahub/tasks/registry.py
```diff
@@ -95,10 +96,12 @@ async def refresh_branches(db: InfrahubDatabase) -> None:
...
-            if active_branch.name in registry.branch:
-                await update_branch_registry(db=db, branch=active_branch)
-            else:
-                await create_branch_registry(db=db, branch=active_branch)
+            # Absorb a failure on one branch rather than abandoning the sweep
+            with log_exception_guard(log, f"Failed to refresh branch {active_branch.name!r} in the registry"):
+                if active_branch.name in registry.branch:
+                    await update_branch_registry(db=db, branch=active_branch)
+                else:
+                    await create_branch_registry(db=db, branch=active_branch)
```
gmazoyer, line 107: wow the fact that these two functions have the same number of characters and that only the first words differ confused me. I had to read it 5 times to finally see they are 2 different functions 🫨
no action to take though, it's just come from my broken eyes

backend/infrahub/graphql/mutations/branch.py
```diff
@@ -253,6 +253,9 @@ async def mutate(
         async with graphql_context.db.start_transaction() as db:
             await obj.save(db=db, user_id=graphql_context.active_account_session.account_id)
 
+        # update registry after txn commit, so it cannot diverge from db on failure
+        registry.branch[obj.name] = obj
```
gmazoyer: Is this safe the branch is not already in this worker's registry? `Branch.get_by_name` never loads the schema, and `registry.get_branch()` and `create_branch_registry()` both do when they insert.

If the entry lands schema-less, does the next `refresh_branches` ever fix it?

Is this where the `BranchSaver` you mention in the description would come in handy?

gmazoyer (approved):

backend/tests/component/tasks/test_registry.py
ajtmccarty: yes good idea. this is done

backend/infrahub/graphql/mutations/branch.py
ajtmccarty: very good point. updated all the registry update sites to only update an existing branch in the registry and not add a new one. I think if everything is working correctly, then this would not be an issue that we could encounter b/c the branches and their schemas will always be up-to-date, but it is better to be safe

backend/infrahub/core/branch/tasks.py
```diff
@@ -101,13 +101,15 @@ async def migrate_branch(branch: str, context: InfrahubContext, send_events: boo
             log.info(f"No migrations detected for branch '{obj.name}'")
             obj.graph_version = GRAPH_VERSION
             await obj.save(db=db)
+            registry.refresh_cached_branch(obj)
```
ogenstad, line 104: I'm not completely sure what problem it is we are solving here. I'm not saying anything is wrong, however the way I read this is that we'd be solving a problem for the current worker but we don't have anything to indicate to other workers that they need to take action. I.e. to trigger the other workers to refresh their view. It might be that this happens elsewhere. Just asking as a sanity check if we need anything else to happen here as well (and in the other places where we refresh this. Another consideration would be if we need to update the hashes within the components to indicate that a branch gets out of sync until things converge. I'm not sure that the graph_version or status are currently included in that hash. I think it might only be the actual schema that gets included.

So this is mainly a sanity check.

ajtmccarty: it was possible for changes made to a branch to be saved to the database without updating the branch in the registry, so the persisted branch would drift from the cached branch. the only problem this PR aims to solve is updating the branch in the registry when a branch change is persisted to the database.

you are correct that non-schema changes to a branch would not be picked up by other workers, so I updated `update_branch_registry` to check for any non-schema changes when updating `registry.branch`

gmazoyer (approved):

Issue #10402, opened 2026-08-25
tech-debt: mutation responses render relationships from in-memory state, so a response depends on the save having read the peers


fatih-acar:
## Task Description

`Node.to_graphql` renders a node's relationships only from what the in-memory relationship managers happen to hold: it calls `rel_manager.get_one()` / `list(rel_manager)` inside a `try` block and swallows the `LookupError` those raise when a manager has not fetched its peers, omitting the field from the response entirely (`backend/infrahub/core/node/__init__.py`, the relationship loop at the end of `to_graphql`).

The GraphQL relationship resolvers cannot cover for that during a mutation. They read the database at `graphql_context.at`, which is the request's timestamp — and `InfrahubMutationMixin.mutate` only advances it *after* the mutation returns. So while a mutation response is being built, a resolver read cannot see the edges the mutation just wrote.

The result is that mutation responses depend on `RelationshipManager.save()` having read the peers as a side effect. That read is what populates the managers today; nothing in the response path asks for it.

Two consequences:

- A relationship that the mutation wrote but that is missing from the in-memory state comes back `null` in the response, silently, rather than erroring. Removing the peer read from `RelationshipManager.save()` for untouched relationships reproduces this immediately: three `backend/tests/component/graphql/mutations/test_ipam.py` tests fail with `ip_namespace: null` in a *create* response even though the edge is in the database.
- It blocks removing that read. `Node.save()` with no `fields` filter reads the peers of every relationship on the node — 5 queries for a node with 5 relationships, whether or not anything changed — and the read cannot be dropped while the response depends on it.

## Additional Information

Found while reducing `relationship_get_peer` queries on the update/upsert path (`4R+1` to `2R+1` for R relationships in the payload). The remaining per-update reads are documented there; this one is the blocker for the rest.

PR #10404, opened 2026-08-25
perf(backend): stop re-reading relationship peers during creates and updates

fatih-acar:
## Summary

Creating or updating a node issued the same `relationship_get_peer` query several times over — once per validation step — because each step recomputed its comparison against the database rather than against the peers already read. A create with two relationships in its payload made 3 of those queries and an upsert of the same node made 9. The create now makes none and the upsert makes 3. Bulk loads and idempotent upserts, where the same data is re-sent unchanged, carry the largest share of that cost.

## Key Changes

Each is a separate commit carrying a regression test that fails without it:

- A node that has not been written yet is no longer asked for its relationship peers — it cannot have any (`a746d6e90`)
- A read with `prefetch_relationships` now also covers the relationships that came back without a peer, instead of leaving each one to be looked up again on first access (`ec4a8caa7`)
- Applying the payload, checking the constraints and saving now compare against a single read of the peers. The record is discarded whenever the manager writes, and peers read before the node exists are never recorded — creating the node writes its edges without going through the manager (`83e91143e`)
- The preview node built to work out which locks to take no longer reads peers it never consults; the relationship lock names come from the payload, and a test pins that the names are unchanged (`7180d045c`)
- Profiles are no longer looked up for a node that was just created: the ones its payload named or its template copied onto it are already in memory (`32a744f45`, corrected in `bf8069e6e` after review — the first version gated on the payload alone and skipped the applier for template-carried profiles)
- Review follow-ups: the comparison computed before a node exists is no longer kept for reuse (`58ed1bc74`), and `save()`/`delete()` discard the recorded peers before their writes rather than after (`1ca6567d9`)
- An update that changes nothing no longer reads the node's parent for a changelog that is never turned into an event (`6e7632aa8`)

## Measured

`relationship_get_peer` queries per operation, with R relationships in the payload:

| Operation | Before | After |
|---|---|---|
| Create, 1 relationship | 2 | 0 |
| Create, 2 relationships | 3 | 0 |
| Upsert existing, attributes only | 2 | 2 (1 when nothing changed) |
| Upsert existing, 1 relationship | 5 | 2 |
| Upsert existing, 2 relationships | 9 | 3 |

For an existing node the shape went from `4R+1` to `R+1`; what remains is the one read that applying the payload genuinely needs.

## Related Context

Same class of redundancy as `3e57b5b7b` ("skip peer fetch for empty relationships on node create", 1.10.6), which removed one such read at the create query itself.

`Node.save()` with no field filter still reads the peers of every relationship on the node. That read cannot be removed yet: mutation responses render relationships from in-memory state and depend on it. Filed as #10402.

## Documentation Updates

- `dev/knowledge/backend/mutations.md` — when a mutation reads the relationship peers, and why the lock preview does not

## Test Plan

- `uv run pytest backend/tests/component backend/tests/unit` — green on this branch
- The measurements above come from a probe that wraps `InfrahubDatabase.execute_query_with_metadata` and records a call stack per query; it is not part of the branch

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10404?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -1252,6 +1252,9 @@ class NodeManager:
         peer_ids = grouped_peer_nodes.get_all_peers()
         # there are no peers to enrich the nodes
         if not peer_ids:
+            if prefetch_relationships:
+                for node in nodes_by_id.values():
+                    _mark_all_relationships_as_fetched(node=node)
             return
 
         missing_peers: dict[str, Node] = {}
@@ -1287,6 +1290,8 @@ class NodeManager:
         insert_peer_node: bool,
     ) -> None:
         if not grouped_peer_nodes.has_node(node_id=node.get_id()):
+            if insert_peer_node:
+                _mark_all_relationships_as_fetched(node=node)
             return
 
         node_schema = node.get_schema()
@@ -1296,10 +1301,14 @@ class NodeManager:
                 rel_name=rel_schema.get_identifier(),
                 direction=rel_schema.direction,
             )
+            rel_manager: RelationshipManager = getattr(node, rel_schema.name)
             if not peer_ids:
+                if insert_peer_node:
+                    # The batched query covered every relationship of this node, so an empty result
+                    # is authoritative and a later read does not need to ask the database again.
+                    rel_manager.has_fetched_relationships = True
                 continue
 
-            rel_manager: RelationshipManager = getattr(node, rel_schema.name)
             if insert_peer_node:
                 rel_peers: list[Node | str] = []
                 for peer_id in peer_ids:
@@ -1373,6 +1382,13 @@ class NodeManager:
             await node.delete(db=db, at=at, user_id=user_id)
 
         return nodes_to_delete
+
+
+def _mark_all_relationships_as_fetched(node: Node) -> None:
+    """Record that a batched relationship read has covered every relationship of this node."""
+    for rel_name in node._relationships:
+        rel_manager: RelationshipManager = getattr(node, rel_name)
+        rel_manager.has_fetched_relationships = True
 
 
 def _get_cardinality_one_identifiers_by_kind(
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -29,7 +29,7 @@ from infrahub.core.query.node import (
 )
 from infrahub.core.query.relationship import RelationshipGetPeerQuery
 from infrahub.core.registry import registry
-from infrahub.core.relationship import Relationship, RelationshipManager
+from infrahub.core.relationship import Relationship
 from infrahub.core.relationship.model import PeerWithRelationshipMetadata
 from infrahub.core.schema import (
     GenericSchema,
@@ -1254,7 +1254,7 @@ class NodeManager:
         if not peer_ids:
             if prefetch_relationships:
                 for node in nodes_by_id.values():
-                    _mark_all_relationships_as_fetched(node=node)
+                    node.mark_relationships_as_fetched()
             return
 
         missing_peers: dict[str, Node] = {}
@@ -1291,7 +1291,7 @@ class NodeManager:
     ) -> None:
         if not grouped_peer_nodes.has_node(node_id=node.get_id()):
             if insert_peer_node:
-                _mark_all_relationships_as_fetched(node=node)
+                node.mark_relationships_as_fetched()
             return
 
         node_schema = node.get_schema()
@@ -1301,7 +1301,7 @@ class NodeManager:
                 rel_name=rel_schema.get_identifier(),
                 direction=rel_schema.direction,
             )
-            rel_manager: RelationshipManager = getattr(node, rel_schema.name)
+            rel_manager = node.get_relationship(rel_schema.name)
             if not peer_ids:
                 if insert_peer_node:
                     # The batched query covered every relationship of this node, so an empty result
@@ -1382,13 +1382,6 @@ class NodeManager:
             await node.delete(db=db, at=at, user_id=user_id)
 
         return nodes_to_delete
-
-
-def _mark_all_relationships_as_fetched(node: Node) -> None:
-    """Record that a batched relationship read has covered every relationship of this node."""
-    for rel_name in node._relationships:
-        rel_manager: RelationshipManager = getattr(node, rel_name)
-        rel_manager.has_fetched_relationships = True
 
 
 def _get_cardinality_one_identifiers_by_kind(
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -244,6 +244,17 @@ async def get_profile_ids(db: InfrahubDatabase, obj: Node | CoreObjectTemplate) 
     return {pr.peer_id for pr in profile_rels}
 
 
+def _has_profiles_set(node: Node) -> bool:
+    """Whether the node was built with profiles, named by its payload or copied from its template.
+
+    A relationship manager initialised with data is marked as fetched, so this reads nothing.
+    """
+    if not hasattr(node, PROFILES_RELATIONSHIP_NAME):
+        return False
+    profiles: RelationshipManager = getattr(node, PROFILES_RELATIONSHIP_NAME)
+    return profiles.has_fetched_relationships and len(profiles) > 0
+
+
 async def _do_create_node(
     node_class: type[Node],
     node_constraint_runner: NodeConstraintRunner,
@@ -374,8 +385,9 @@ async def create_node(
                     user_id=user_id,
                 )
 
-    # A node created here can only be linked to the profiles its payload named.
-    if PROFILES_RELATIONSHIP_NAME in data and await get_profile_ids(db=db, obj=obj):
+    # The node was written from its in-memory state, so the profiles it is linked to are the ones
+    # its payload or its template set on it; there is nothing to read back.
+    if _has_profiles_set(node=obj):
         node_profiles_applier = NodeProfilesApplier(db=db, branch=branch)
         await node_profiles_applier.apply_profiles(node=obj)
         await obj.save(db=db, user_id=user_id)
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -251,8 +251,12 @@ def _has_profiles_set(node: Node) -> bool:
     """
     if not hasattr(node, PROFILES_RELATIONSHIP_NAME):
         return False
-    profiles: RelationshipManager = getattr(node, PROFILES_RELATIONSHIP_NAME)
-    return profiles.has_fetched_relationships and len(profiles) > 0
+    profiles = node.get_relationship(PROFILES_RELATIONSHIP_NAME)
+    if not profiles.has_fetched_relationships:
+        # Nothing populated this manager, so the node was built without profiles. Counting the peers
+        # is not an option: `len()` raises on a manager whose peers have never been read.
+        return False
+    return len(profiles) > 0
 
 
 async def _do_create_node(
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Mapping, overload
 from infrahub import lock
 from infrahub.core import registry
 from infrahub.core.constants import (
+    PROFILES_RELATIONSHIP_NAME,
     SYSTEM_USER_ID,
     MetadataOptions,
     RelationshipCardinality,
@@ -373,7 +374,8 @@ async def create_node(
                     user_id=user_id,
                 )
 
-    if await get_profile_ids(db=db, obj=obj):
+    # A node created here can only be linked to the profiles its payload named.
+    if PROFILES_RELATIONSHIP_NAME in data and await get_profile_ids(db=db, obj=obj):
         node_profiles_applier = NodeProfilesApplier(db=db, branch=branch)
         await node_profiles_applier.apply_profiles(node=obj)
         await obj.save(db=db, user_id=user_id)
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -93,6 +93,10 @@ class DbPeersRead:
     at: str
     branch_agnostic: bool
 
+    def is_valid_for(self, at: str, branch_agnostic: bool) -> bool:
+        """Whether a read under these conditions would return the peers recorded here."""
+        return self.at == at and self.branch_agnostic == branch_agnostic
+
 
 @dataclass
 class PeerWithRelationshipMetadata:
@@ -1143,8 +1147,11 @@ class RelationshipManager[RelationshipManagerPeerType]:
             self._relationship_id_details = None
             return details
 
-        peers = await self.get_db_peers(db=db, at=at, branch_agnostic=branch_agnostic)
-        self._last_db_peers = DbPeersRead(peers=peers, at=str(at or self.at), branch_agnostic=branch_agnostic)
+        # Resolve the timestamp here rather than leaving it to `get_db_peers`, so that the peers are
+        # recorded under the very timestamp the query read them at.
+        read_at = at or self.at
+        peers = await self.get_db_peers(db=db, at=read_at, branch_agnostic=branch_agnostic)
+        self._last_db_peers = DbPeersRead(peers=peers, at=str(read_at), branch_agnostic=branch_agnostic)
 
         return self._compare_with_db_peers(peers=peers)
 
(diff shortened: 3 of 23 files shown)


backend/infrahub/core/node/create.py
```diff
@@ -373,7 +374,8 @@ class that is not a node schema protocol is passed.
                     user_id=user_id,
                 )
 
-    if await get_profile_ids(db=db, obj=obj):
+    # A node created here can only be linked to the profiles its payload named.
+    if PROFILES_RELATIONSHIP_NAME in data and await get_profile_ids(db=db, obj=obj):
```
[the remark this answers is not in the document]
fatih-acar: Valid point — the guard read the request payload, but a template carries a `profiles` relationship of its own and `NodeTemplateApplier._apply_relationships` copies its peers into the fields the node is built from (never into the caller's `data`), so a node created from such a template skipped the applier and kept the template's copy of the profile values, which goes stale as soon as the profile changes. It now decides from the node's in-memory `profiles` manager, which covers both sources and still reads nothing for a node without profiles, in bf8069e6e — with a regression test where the profile changes after the template took its values.

backend/infrahub/core/relationship/model.py
```diff
@@ -1124,9 +1134,39 @@ async def fetch_relationship_ids(
         if not force_refresh and self._relationship_id_details is not None:
             return self._relationship_id_details
 
-        current_peer_ids = [rel.get_peer_id() for rel in self._relationships]
+        if not self.node._existing:
+            # A node that has never been written cannot have any relationship edge in the database.
+            # Creating it writes the edges without going through this manager, so this answer must
+            # not be recorded for reuse: it stops being true as soon as the node is saved.
+            self._last_db_peers = None
+            return self._compare_with_db_peers(peers=[])
```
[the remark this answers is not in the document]
fatih-acar: Valid point — the read was not recorded, but the comparison derived from it was, and `_create()` writes the edges without resetting it. The unsaved-node branch now discards `_relationship_id_details` as well, so the first `force_refresh=False` call after the node exists reads the database, in 58ed1bc74 (test included). One caveat the test had to work around: the manager's `at` is fixed at init, so a post-create read through the same manager only sees the new edges when given a later `at`.

backend/infrahub/core/relationship/model.py
```diff
@@ -1124,9 +1134,39 @@ async def fetch_relationship_ids(
         if not force_refresh and self._relationship_id_details is not None:
             return self._relationship_id_details
 
-        current_peer_ids = [rel.get_peer_id() for rel in self._relationships]
+        if not self.node._existing:
+            # A node that has never been written cannot have any relationship edge in the database.
+            # Creating it writes the edges without going through this manager, so this answer must
+            # not be recorded for reuse: it stops being true as soon as the node is saved.
+            self._last_db_peers = None
```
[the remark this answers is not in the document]
fatih-acar: Done in 1ca6567d9 — `save()` and `delete()` now discard the recorded peers right after the comparison is made, before any write, so a write failing part-way cannot leave the pre-write snapshot behind.

backend/tests/component/core/test_node.py
```diff
@@ -1811,3 +1813,35 @@ async def test_node_serialize_address(db: InfrahubDatabase, default_branch: Bran
...
+
+    counting_db.reset_counts()
+    loaded.nbr_seats.value = 4
+    await loaded.save(db=counting_db, fields=["nbr_seats"])
+
+    assert loaded.node_changelog.has_changes is True
+    assert loaded.node_changelog.parent is not None
+    assert loaded.node_changelog.parent.node_id == person.id
+    assert loaded.node_changelog.parent.node_kind == "TestPerson"
+    assert counting_db.count_for(RelationshipGetPeerQuery.name) == 1
```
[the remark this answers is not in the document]
fatih-acar: Done in 1ca6567d9 — the car is reloaded on the plain `db` after the second save and `nbr_seats == 4` asserted.

backend/tests/component/core/profiles/test_create_profile_reads.py
```diff
@@ -0,0 +1,51 @@
...
+    profile_schema = registry.schema.get(name=f"Profile{TestKind.CHILD}", branch=default_branch)
+    profile = await Node.init(db=db, schema=profile_schema, branch=default_branch)
+    await profile.new(db=db, profile_name="from-profile", profile_priority=1000, name="profile-name")
+    await profile.save(db=db)
+
+    child_schema = registry.schema.get_node_schema(name=TestKind.CHILD, branch=default_branch)
+
+    child = await create_node(data={"profiles": [profile.id]}, db=db, branch=default_branch, schema=child_schema)
+
+    assert child.name.value == "profile-name"
```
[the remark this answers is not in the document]
fatih-acar: Done in 1ca6567d9 — the child is reloaded with source metadata and the profile value, its source and the `profiles` relationship are asserted from that read.

backend/infrahub/core/manager.py
```diff
@@ -1296,10 +1301,14 @@ async def _enrich_one_node_with_relationships(
                 rel_name=rel_schema.get_identifier(),
                 direction=rel_schema.direction,
             )
+            rel_manager: RelationshipManager = getattr(node, rel_schema.name)
```
ajtmccarty: you can use `node.get_relationship()` instead of `getattr`

backend/tests/component/core/test_manager_node.py
```diff
@@ -827,3 +829,40 @@ async def test_get_one_global(db: InfrahubDatabase, default_branch: Branch, base
...
+        for rel_name in fetched.get_schema().relationship_names
+    }
+
+    assert peers_per_relationship == {
+        "cars": [car.id],
+        "cars_driven": [],
+        "member_of_groups": [],
+        "subscriber_of_groups": [],
+        "profiles": [],
+    }
```
ajtmccarty, line 870: I think this test should assert that `has_fetched_relationships` is set `True` on all the relationships for `fetched`.

backend/tests/component/core/test_manager_node.py
```diff
@@ -827,3 +829,40 @@ async def test_get_one_global(db: InfrahubDatabase, default_branch: Branch, base
...
+    fetched = await NodeManager.get_one(
+        db=counting_db, id=person.id, branch=default_branch, prefetch_relationships=True, raise_on_error=True
+    )
+
+    assert counting_db.count_for(RelationshipGetPeerQuery.name) == 0
+
+    peers_per_relationship = {
+        rel_name: [
+            str(rel.peer_id)
+            for rel in await getattr(fetched, rel_name).get_relationships(db=counting_db)
```
ajtmccarty: can use `get_relationship` instead of `getattr`

backend/infrahub/core/relationship/model.py
```diff
@@ -1124,10 +1134,39 @@ async def fetch_relationship_ids(
...
+    ) -> RelationshipUpdateDetails:
+        """Compare the local relationships with the peers last read from the database.
+
+        The peers are read again only when this manager has not read them yet under the same
+        conditions, or when it has written to the relationship since. Use this instead of
+        `fetch_relationship_ids` when the local relationships have changed but the database has
+        not, so that the comparison is recomputed without paying for another read.
+        """
+        last_read = self._last_db_peers
+        if last_read is None or last_read.at != str(at or self.at) or last_read.branch_agnostic != branch_agnostic:
```
ajtmccarty: looks like an equality method of some sort of `DbPeersRead`. something like `if last_read and last_read.is_valid_for(at, branch_agnostic): return self._compare_.. else return self.fetch_...`

backend/infrahub/core/relationship/model.py
```diff
@@ -1124,10 +1134,39 @@ async def fetch_relationship_ids(
...
+            # A node that has never been written cannot have any relationship edge in the database.
+            # Creating it writes the edges without going through this manager, so this answer must
+            # not be recorded for reuse: it stops being true as soon as the node is saved.
+            self._last_db_peers = None
+            return self._compare_with_db_peers(peers=[])
 
-        # A node that has never been written cannot have any relationship edge in the database.
-        peers = await self.get_db_peers(db=db, at=at, branch_agnostic=branch_agnostic) if self.node._existing else []
+        peers = await self.get_db_peers(db=db, at=at, branch_agnostic=branch_agnostic)
+        self._last_db_peers = DbPeersRead(peers=peers, at=str(at or self.at), branch_agnostic=branch_agnostic)
```
ajtmccarty: there's a mismatch here between the use of `at` in `get_db_peers` and the use of `(at or self.at)` in `DbPeersRead` b/c if `at` in `get_db_peers` is `None`, then it will end up defaulting to `Timestamp()` (current time) in any database queries eventually. and `self.at` is going to be an earlier timestamp from when the `RelationshipManager` was created.

Normally, the difference will be small and won't really matter, but simultaneous updates to peers could cause overlapping/conflicting updates. We might already completely prevent this through a combination of fine-grained locking and idempotent queries, but we'll need to be sure

Or maybe this is fine b/c `_last_db_peers` is nulled following a write

backend/infrahub/graphql/mutations/ipam.py
```diff
@@ -208,7 +212,7 @@ async def mutate_update(
             id=address.get_id(),
             branch=branch,
         )
-        await preview_obj.from_graphql(db=db, data=data, process_pools=False)
+        await apply_payload_for_lock_names(db=db, node=preview_obj, data=data)
```
ajtmccarty, line 215: I think this is a bit of a kludge if I am understanding it correctly. `apply_payload_for_lock_names` just sets `has_fetched_relationships` True for every rel on the preview object so that `from_graphql` won't retrieve the peers from the db. the `from_graphql` methods are already a bit of a mess, so this isn't that much worse, but an option to skip retrieving peers in `from_graphql` might be cleaner, although perhaps too big of a change for this PR b/c those methods are a mess

backend/infrahub/core/node/create.py
```diff
@@ -244,6 +244,17 @@ async def get_profile_ids(db: InfrahubDatabase, obj: Node | CoreObjectTemplate)
...
 
+def _has_profiles_set(node: Node) -> bool:
+    """Whether the node was built with profiles, named by its payload or copied from its template.
+
+    A relationship manager initialised with data is marked as fetched, so this reads nothing.
+    """
+    if not hasattr(node, PROFILES_RELATIONSHIP_NAME):
+        return False
+    profiles: RelationshipManager = getattr(node, PROFILES_RELATIONSHIP_NAME)
+    return profiles.has_fetched_relationships and len(profiles) > 0
```
ajtmccarty: is `has_fetched_relationships` necessary here? shouldn't `len(profiles) > 0` be enough?

ajtmccarty (commented):
my longest comment is the only real possible functional concern that I have. other than that, it looks like a solid improvement for performance. maybe a bit of a loss for readability, but these methods are already quite difficult to read, so definitely worth it

backend/infrahub/core/manager.py
fatih-acar: Done in ba8738c8d. `_enrich_one_node_with_relationships` now calls `node.get_relationship(rel_schema.name)`, which also drops the local annotation and the `RelationshipManager` import, now unused in this module.

The loop below that flipped `has_fetched_relationships` was doing the same `getattr` dance, and `lock_utils.py` had its own copy of it, so it has moved onto `Node` as `mark_relationships_as_fetched()` and both callers share it. Same treatment for the `getattr` in `get_lock_names_on_object_mutation` while I was in that file.

backend/tests/component/core/test_manager_node.py
fatih-acar: Added in ba8738c8d, before any peer is read:

```python
relationship_names = fetched.get_schema().relationship_names
assert all(fetched.get_relationship(rel_name).has_fetched_relationships for rel_name in relationship_names)
```

That is the property the change is actually about; the `count_for(...) == 0` after the reads is then the consequence rather than the whole claim.

backend/tests/component/core/test_manager_node.py
fatih-acar: Done in ba8738c8d, in the same test.

backend/infrahub/core/relationship/model.py
fatih-acar: Done in ba8738c8d. `DbPeersRead.is_valid_for(at, branch_agnostic)` owns the condition now, and the call site reads the way you wrote it:

```python
last_read = self._last_db_peers
if last_read and last_read.is_valid_for(at=str(at or self.at), branch_agnostic=branch_agnostic):
    return self._compare_with_db_peers(peers=last_read.peers)

return await self.fetch_relationship_ids(db=db, at=at, branch_agnostic=branch_agnostic, force_refresh=True)
```

Kept as a named predicate rather than `__eq__`: the two sides are not the same kind of thing — a read that happened, against the conditions of one we want — and the name says which way round it is.

backend/infrahub/core/node/create.py
fatih-acar: It is load-bearing, yes. `RelationshipManager.__len__` raises `LookupError` when the peers have never been read, and for a create whose payload has no `profiles` key — and no template that sets one — nothing populates that manager: `Node.new` leaves it at `has_fetched_relationships = False`. So `len(profiles) > 0` on its own would raise on the commonest create rather than return `False`.

Split in ba8738c8d so the reason is visible instead of riding along as a conjunct:

```python
profiles = node.get_relationship(PROFILES_RELATIONSHIP_NAME)
if not profiles.has_fetched_relationships:
    # Nothing populated this manager, so the node was built without profiles. Counting the peers
    # is not an option: `len()` raises on a manager whose peers have never been read.
    return False
return len(profiles) > 0
```

backend/infrahub/core/relationship/model.py
fatih-acar: The mismatch is not there: `get_db_peers` resolves the fallback itself — it builds `RelationshipGetPeerQuery` with `at=at or self.at` — so `at=None` on that call never reaches the database as "now". And `self.at` cannot be `None`: it is a required `Timestamp` on `RelationshipManager`, and `Node.init` always builds one (`attrs["at"] = Timestamp(at)`). Both lines were already resolving to the same value. Made that syntactic rather than something you have to go and check, in ba8738c8d:

```python
read_at = at or self.at
peers = await self.get_db_peers(db=db, at=read_at, branch_agnostic=branch_agnostic)
self._last_db_peers = DbPeersRead(peers=peers, at=str(read_at), branch_agnostic=branch_agnostic)
```

On the concurrency question, your own "or maybe this is fine" is most of the answer, but it is worth being precise about why:

- The peer query is strictly time-scoped. `get_query_filter_path` emits `r.from < $at AND (r.to IS NULL OR r.to > $at)`, so a fresh read at the same `at` returns the same rows as the recorded one. A concurrent update stamped later than our `at` is invisible to both the snapshot and a re-read — time-travel semantics do the work here, not luck.
- The one case where re-reading would differ is an edge whose `from` precedes our `at` and which is committed between our two reads: a request that started before us and wrote after us. That window is real, and it existed before this PR too, between the old read and the write that followed it.
- `_last_db_peers` is dropped before every write on the manager (`save()`, `delete()`, and the `None`-item path in `update()`), so the record is only ever reused between reads, never across a write.
- In practice the reuse is one hop: the constraint runner's `refresh_update_details` and then the one in `save()`, both with `at=None`, both inside the `relationship_count.<identifier>.<node_id>` / `.<peer_id>` locks that `get_lock_names_on_object_mutation` takes for exactly the constrained relationships where a missed peer would matter. The writes are idempotent `MERGE`s per peer, so a peer we did not see is left in place rather than clobbered.

If you would rather not lean on that, the tighter variant is to reuse the snapshot only inside `save()` and put `NodeConstraintRunner` back on a forced read. That costs one `relationship_get_peer` per relationship on an upsert — the `R+1` shape becomes `2R+1` — and I will take it if you prefer the guarantee to the query. Your call.

backend/infrahub/graphql/mutations/ipam.py
fatih-acar: You are reading it right, and I agree it is a kludge. What I have done in ba8738c8d:

- Given it a name. The loop is now `node.mark_relationships_as_fetched()`, a `Node` method whose docstring states when marking is legitimate: the peers it holds are known to match the database, or the node is a preview that is never saved. `manager.py` was running the same loop for the batched read, so both callers share one implementation.
- Said in the `apply_payload_for_lock_names` docstring that the marking is what keeps `from_graphql` from reading, so the next reader does not have to infer it from a bare flag assignment.

What I have not done is touch `from_graphql`. A `fetch_peers=False` threaded into `RelationshipManager.update` is the cleaner shape, but `update()` uses the fetched peers as `previous_relationships`, which decides `changed` and which `Relationship` objects get reused rather than rebuilt — so the flag changes real behaviour on that path and wants its own change and tests, not a corner of a perf PR. Happy to file it. Worth noting the preview node is not a general case of that knob either: it wants *no* peers and is thrown away, whereas a caller of `from_graphql` that skipped the fetch would still be saved.

The risk worth checking here is whether the lock names move, and they do not: `from_graphql` only visits the keys present in the payload, and `update()` clears `_relationships` and rebuilds it from the payload's items, so the peers `get_lock_names_on_object_mutation` iterates are the payload's either way — the read only ever fed `previous_relationships`. `test_lock_names_do_not_depend_on_the_stored_peers` pins that against a hydrated node that has a different `owner` stored.

fatih-acar:
Thanks — the `at` one was the right thing to push on, even though it turned out to be a false alarm: `get_db_peers` was already resolving `at or self.at` before it built the query, so the key and the query never diverged. I have made that explicit rather than implied, and written down in the thread what actually bounds the reuse window, including the one case where a re-read could legitimately differ.

Everything else is addressed in ba8738c8d — `get_relationship()` over `getattr`, `DbPeersRead.is_valid_for`, the marking loop given a name and shared between the node manager and the lock preview, and the profiles check split so it says why the flag is load-bearing. Replies on each thread.

One thing beyond the review while I was here: 3d7040e90 folds this branch's six changelog fragments into a single entry. Six of them described one improvement from six angles, each in terms of the queries it removed, which is not what a reader of the release notes is after.

ajtmccarty (approved):

PR #10405, opened 2026-08-25
chore(agents): make pre-ci detect changed areas and close the frontend gap

pa-lem:
## Problem

`/pre-ci` describes itself as running "all locally-executable CI checks", but it ran only part of the frontend gate. It could report all-green while `frontend-lint` and `frontend-tests` failed on the PR.

Checked against `.github/workflows/ci.yml`:

**Missing entirely**
- `pnpm knip` — step 2 of the `frontend-lint` job
- `pnpm test:coverage` — the `frontend-tests` job (Phase 4 ran only `backend.test-unit`)
- `frontend/packages/graph` tests — a separate step of the same job, outside `frontend/app`

**Wrong variants**
- `npx biome check --write .` instead of `pnpm exec biome ci .` — the auto-fix variant, not the check-only one CI asserts
- `npx betterer` instead of `pnpm exec betterer ci` — bare `betterer` rewrites the snapshot rather than failing on an increase
- `npx` / `npm` used throughout a pnpm workspace

## Change

**Phase 0 — detect what changed.** Diffs against the merge-base with `origin/develop` (falling back to `origin/stable`) *and* includes the working tree, since uncommitted work is about to be pushed too. Paths are classified with the same globs as `.github/file-filters.yml`, so each area's phases run only when relevant. `--all` forces every phase.

**Phase 3 — complete frontend gate,** at CI parity, with each command annotated with the job it mirrors.

Two things worth flagging for reviewers:

- `frontend-lint` has **no path filter** in `ci.yml` — it runs on every PR. A backend-only change still fails CI if frontend lint is broken. Phase 3A therefore always runs the lint trio regardless of detection.
- `frontend-tests` runs a **second suite** in `frontend/packages/graph`. It is also absent from the four-command gate documented in `frontend/app/AGENTS.md`.

The summary table now lists skipped rows explicitly (`skipped (no backend changes)`) rather than omitting them, so partial coverage is visible rather than reading as a full pass.

## Notes

- No changelog fragment: internal agent tooling, not user-facing.
- Detection is path-based only — no branch-name logic, so it behaves the same for everyone.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10405?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/commands/pre-ci.md
+++ b/.agents/commands/pre-ci.md
@@ -4,10 +4,12 @@ argument-hint: "[--fast] [--all]"
 allowed-tools:
   - Bash(git merge-base:*)
   - Bash(git diff:*)
-  - Bash(git status:*)
+  - Bash(git ls-files:*)
   - Bash(git rev-parse:*)
+  - Bash(sort:*)
   - Bash(uv run invoke:*)
   - Bash(uv run ruff check:*)
+  - Bash(uv run ty check:*)
   - Bash(uv run yamllint:*)
   - Bash(uv run --directory python_testcontainers pytest:*)
   - Bash(uv lock --check:*)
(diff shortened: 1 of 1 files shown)


saltas888:
Would it make sense to point stable?

.agents/commands/pre-ci.md
```diff
@@ -1,116 +1,313 @@
...
+  point. Preferring `develop` unconditionally diffs a `stable`-based branch tens of commits back.
+- **Every** ref is existence-checked, including the first. An unguarded `git merge-base` aborts
+  the block on a fork that carries only one branch, leaving `MERGE_BASE` empty and detecting
+  *nothing*. An empty `BASE_REF` must reach the all-areas fallback instead.
+- `git fetch` first: a stale ref can flip the choice between the two candidates. If it fails,
+  the run is best-effort — prefer `--all`.
+- `--no-renames` keeps both sides of a rename, so the area that *lost* a file is still flagged.
+  Do not swap in `git status --porcelain`, whose `R  old -> new` collapses into one bogus path.
+- The list covers **committed and uncommitted** work: CI sees the former, you are about to push
+  the latter.
```
polmichel: This section could be/become complex to read and maintain, and I think we could just tell the agent what to do. Have you run into issues with this lookup of what has changed in the code?
We might need to add `release-X.YZ` branches to this script, which would make this script even more complex.

Have you tried using a plain text? An example I've generated with Claude then modified:
_Identify which areas (table below) were touched on this branch — committed, staged, and untracked — relative to its base (stable, develop, whichever long-lived branch it forked from). Run only the phases for changed areas, plus those marked always-run. When unsure about the base or whether a file counts, include it. --all skips this._

polmichel (commented):
This is a great idea 👍 
Have you tried to use `/pruning-residues` skill on this skill? When I've run it locally it offers me to trim/condense a number of items.

polmichel (approved):
Cool improvement 👍

PR #10408, opened 2026-08-25
docs(dev): add a skill for testing against real data (AI/DC solution)

saltas888:
# Why

A skill so an agent can stand up a live Infrahub with real data — the AI/DC solution's generated fabric, tens of thousands of nodes — on the code under test, without cycling through discovery first. This started as a `dev/guides/` page; discussion landed on a skill, since the problem is the trigger rather than the content.

## What changed

- `.agents/skills/testing-against-real-data/`: `SKILL.md` with the workflow, plus references for the `/source`-mount dev stack and for measuring. `.claude/skills` symlinks to `.agents/skills`, so it loads for anyone working in this repo.
- Internal agent tooling only. No code, config, or user-facing change.

## How this was tested

- The skill loads. Dropped into `.agents/skills/`, it is picked up and offered by name and description in a session opened on this repo, which is the mechanism the whole change exists for. Frontmatter parses, pre-commit hooks pass.
- The procedure came out of a real run, not from reading. The load order, the scaling levers, the generator cascade behaviour and most of the failure-mode table come from standing this stack up from a worktree and driving it to a ~21.6k-node diff (19,185 added, 2,477 updated) on a proposed change.
- Everything that run did not cover was checked against source rather than executed: the solution's `tasks.py`, `Dockerfile`, `docker-compose.override.yml` and `tests/integration/stack_config.py`, plus this repo's `tasks/shared.py` and `development/docker-compose*.yml`.

Not done: a cold end-to-end run of path A (`INFRAHUB_BASE_VERSION=local`, `inv build`, `inv start`) on a clean machine with a fresh clone. That path is derived from the solution's own invoke tasks and compose override, so it is worth someone walking through once before we lean on it.

Naming: it is named for the job, so it triggers on "test against a live app with real data" as well as on AI/DC by name. Say if you would rather it were named for AI/DC directly.


dev/guides/testing-against-the-ai-dc-solution.md
```diff
@@ -0,0 +1,341 @@
...
+
+services:
+  server:
+    environment:
+      <<: *ai_dc_env
+  task-worker:
+    environment:
+      <<: *ai_dc_env
+    volumes:
+      - <AI_DC_CLONE>:/upstream
```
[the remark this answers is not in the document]
saltas888: Fixed in 951d2916. The paragraph before the block now says to replace `<AI_DC_CLONE>` with the absolute path to the clone, with an example, plus why a relative path is a trap here: it resolves against the compose project directory rather than the repository root, so it silently mounts the wrong thing instead of failing.

dev/guides/testing-against-the-ai-dc-solution.md
```diff
@@ -0,0 +1,341 @@
...
+## Pick a stack
+
+Two stacks can run the solution. They differ in where the Infrahub code comes from.
+
+| | Solution stack (path A) | This repo's dev stack (path B) |
+|---|---|---|
+| Compose files | Released stack downloaded by the solution, plus its own override | `development/docker-compose*.yml` from this repo |
+| Infrahub code | Baked into the image; rebuild to change it | Mounted from the worktree at `/source`; restart to change it |
+| Solution wiring | Already in the solution's override | You write an overlay |
+| Also published | MCP sidecar, Prefect, Neo4j browser and Bolt | Observability profile (Prometheus, Grafana, Tempo) |
```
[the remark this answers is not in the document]
saltas888: Correct, and verified against `development/docker-compose-observability.yml`: every service there is gated behind `debug` (Prometheus, Grafana, Loki, Alloy, node-exporter, cadvisor, the Prefect exporter) or `trace` (Tempo), so the documented command starts none of them.

Fixed in 951d2916, slightly differently from the suggestion. The table row now says the stack sits behind its own profiles, and the `--profile dev` bullet below the command names them, `--profile debug` for Prometheus, Grafana and the exporters, `--profile trace` for Tempo. Splitting it that way keeps the table cell short and puts the actionable part next to the command it applies to.

.agents/skills/testing-against-real-data/references/measuring.md
```diff
@@ -0,0 +1,65 @@
...
+tag, then run against it:
+
+```bash
+docker tag registry.opsmill.io/opsmill/infrahub:local registry.opsmill.io/opsmill/infrahub:dev-mybranch
+export INFRAHUB_BASE_VERSION=dev-mybranch
+uv run inv build
+uv run inv test-integration        # --tier full for the extended tier
+```
+
+The tag cannot be the literal `local` here: testcontainers treats that as a sentinel and
```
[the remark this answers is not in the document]
saltas888: You're right that `local` works here, `tests/e2e/README.md` runs with it and `tests/e2e/conftest.py` returns it. That paragraph was about the ai-dc solution's own integration suite though, run from that clone, where `tests/integration/stack_config.py` raises on a resolved tag of `local`, because infrahub-testcontainers re-resolves the sentinel from `INFRAHUB_TESTING_IMAGE_VER` and the stack can then run an image other than the one just built.

Still a real problem, so I rewrote it in 5b56e553e. Unscoped, in a skill that lives in this repo where `local` is the normal value, it reads as a claim about our harness. It now leads with that, names which suite refuses the tag and why, and drops the blame on testcontainers, which accepts it.

saltas888:
On the second item in the cubic summary (dev/README.md and the "don't delete, leave a pointer" rule): nothing to reconcile, `git diff origin/stable...HEAD -- dev/` is empty. The guide was added and removed inside this PR's own history, so it never existed on stable and there is no reader to redirect. The net diff here is the three skill files only.

.agents/skills/testing-against-real-data/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+  Stands up a live Infrahub built from this worktree, loaded with the AI/DC solution's dataset,
+  so a change can be exercised against realistic data at scale — tens of thousands of nodes,
+  deep hierarchies, generators, artifacts, computed attributes, large diffs — instead of test
+  fixtures. TRIGGER when: the user wants a change tried against a live app, real or realistic
+  data, a big dataset, a large diff, or "at scale"; wants to reproduce something that only shows
+  up on real data; wants a live Infrahub running to click through; or names the AI/DC solution.
+  DO NOT TRIGGER when: the unit, component, or E2E suites already cover the behaviour → run
+  those per AGENTS.md; the user only wants the dependencies up for a test run →
+  `uv run invoke dev.deps`; the work is inside the ai-dc solution repository itself rather than
+  in Infrahub.
```
polmichel, line 13: From what I've captured on the different reviews on `opsmill-skills` project, this is way too long for a skill description. It should only describes when the agents should use the skill.

.agents/skills/testing-against-real-data/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+  `uv run inv start`, which recreates the containers whose image changed. The database and its
+  data survive; only `inv destroy` drops them.
+
+When the user is iterating and a stack is already up, say what you found rather than rebuilding
+on autopilot.
+
+## Step 1 — Prerequisites
+
+- The `python_sdk` submodule must be initialized (`git submodule update --init python_sdk`). A
+  fresh worktree starts with an empty one.
```
polmichel, line 67: All submodules should be initialized right? I'd say the correct command to use is `git submodule update --init --recursive`.

.agents/skills/testing-against-real-data/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Failure modes that cost the most time
+
+| Symptom | Cause | Fix |
+|---|---|---|
+| Change under test appears to have no effect | Stack is running a released Infrahub | `INFRAHUB_BASE_VERSION` was unset; verify with `docker compose images` (Step 3) |
+| Repository stuck "syncing" | The `CoreRepository` object points at `/upstream`, which the compose override mounts from the clone | Check the mount before anything else |
+| Generator cascade stops after the first layer | `triggers.yml` loaded before the repository finished syncing | Re-load it (Step 4) |
+| Objects that should have been generated are missing | A generator failed on import; this surfaces as a failed task run, not a UI error | Check Prefect at `:4200` first |
+| Every SDK import fails | The worktree is mounted over the image's `/source`, and an uninitialized `python_sdk` shadows the packaged SDK | `git submodule update --init python_sdk`, rebuild |
```
polmichel, line 198: This is already listed as one of the mandatory steps

.agents/skills/testing-against-real-data/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+A proposed change's diff summary gives added and updated node counts; `docker stats` gives the
+memory picture during a cascade. Quote the numbers you measured. A claim like "tested at scale"
+without a node count is not a result.
+
+## Step 7 — Iterate
+
+- **Backend change** → rebuild both layers (Step 2) and `uv run inv start`. Or switch to the dev
+  stack in `references/dev-stack-overlay.md`, where a restart is enough.
+- **Frontend change** → no rebuild. Run `cd frontend/app && pnpm dev`; the dev server on `:8080`
+  calls the API at `http://localhost:8000` unconditionally. See `references/measuring.md`.
```
polmichel, line 184: A part of these informations is already contained in the referenced file. This looks like a duplication.

.agents/skills/testing-against-real-data/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+Loading objects only creates design intent. The node count comes from running the generators.
+
+In the UI: **Actions > Generator definitions > generate-fabric > Run**, then pick a fabric. One
+trigger is enough — the fabric generator writes a checksum to each pod, which triggers the pod
+generators in parallel, which do the same to their racks. Watch Prefect rather than guessing;
+a full fabric takes several minutes and the computed-attribute cascade continues after the last
+generator finishes.
+
+Each of the four fabrics (one per vendor) is a similar unit of scale: Fabric-A is 6 super spines,
+3 pods, 8 racks. To grow it, in rough order of effect:
```
polmichel, line 152: This could get out of sync if the AI-DC solution evolves, right? Maybe a pointer to a file reference would be better to avoid that.

polmichel (commented):
Cool idea.

1. Could you add a section that describes how you've tested the skill? This would also clarify how we should use this new skill.
2. Moreover I've run the `/opsmill-dev:pruning-residues` skill once per file. It trims at least 17 lines of these files. Do you think it would be a good idea for you to use this skill on this PR? I can also push what this skill renders if you want.

Issue #3890, opened 2024-07-22
feature: Long Text Attribute aren't displayed properly in the create/edit form


dgarros:
### Infrahub version

0.15.0

### Current Behavior

Currently a long Text attribute, like a description, won't be properly displayed in the create/edit form.
On my laptop, only 51 characters are visible

![Screenshot 2024-07-22 at 13 33 20](https://github.com/user-attachments/assets/9e857fe8-4ffc-4d1e-bd15-bf872b587435)

### Expected Behavior

To some limit, it should be easier to display a long Text Attribute in the side panel.
Either the side of the panel should be extended or the input field should use multiple lines if there isn't enough space to display the text properly

### Steps to Reproduce

Edit any objects with a description and set the value of the description to 
```
Very long description that is still smaller than the current limit of 128 characters. Would be good to fix that
```

_This sentence is 111 characters long_

pa-lem:
I dont think this is an actual issue

simple text fields are meant to be one line without line breaks

textarea are another kind of fields and also a dedicated kind in the schema attributes so if we know that long values will be applied + we want to include new lines, those ones should be used

we can currently scroll to the right + use the arrows to go through a long text in the filed value for a text field, so the default and expected behaviour is already fine now

other tools also have dedicated components for inputs and textarea, there is a real distinction here between the two

for example, the same is done in git, to check that you can:
- go in github
- click your avatar
- set your status to a very long name
- the name will allow a h-scroll to see the full value even if it doesn't fit the panel
- (the same applies if you change your full name in the settings too)

WDYT, is it ok to close this?

PR #10411, opened 2026-08-25
[PoC — reference, do not merge] CI-verified e2e reproduction for #3890 (bug pipeline proof runs)

saltas888:
## PoC: E2E proof runs for the bug pipeline

Experiment: make the bug pipeline able to use E2E reproductions by moving the RED/GREEN proof into CI, with before/after screenshots embedded in this description. Guinea pig: the long-Text-attribute clipping bug (relates to #3890 — do not auto-close).

**How it works**: the `bug-e2e-proof` job (added on this branch) runs exactly the new e2e test. While this body has no fix marker it expects the test to FAIL on its assertion (RED) and embeds the failure screenshot below; once the fix lands and the marker is added it expects PASS (GREEN) and embeds the after screenshot.

This PR intentionally starts test-only, mirroring the pipeline's `/bug-tdd` phase. The normal e2e CI jobs are expected to be red on this phase — the new test fails by design.

### Before (bug reproduced)

✅ `red_confirmed` -- test failed on its assertion: AssertionError: Description field clips its value: 340px of the 111-char text is hidden beyond the field's right edge ([run](https://github.com/opsmill/infrahub/actions/runs/32873414287))

before

### After (fix verified)

✅ `green_confirmed` -- reproduction test passes with the fix applied ([run](https://github.com/opsmill/infrahub/actions/runs/32874468897))

after

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10411?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


saltas888:
Status: this PR stays open as the **reference PoC** for E2E proof runs in the bug pipeline — it will not be merged.

- The actual fix for #3890 was extracted to #10412.
- The production integration of the proof workflow is being specced separately (spec-kit); that PR will link back here as the working example.
- Write-up: https://opsmill.atlassian.net/wiki/spaces/Product/pages/844922883

PR #10412, opened 2026-08-25
fix(frontend): let Text attribute fields wrap instead of clipping long values [closes #3890]

### LINKED ISSUE #3890: feature: Long Text Attribute aren't displayed properly in the create/edit form

saltas888:
Fixes #3890.

A long value in a single-line `Text` attribute (e.g. a description) was clipped at the field's right edge in the create/edit form — only ~50 characters visible with no way to read the rest. The `Text` kind now renders a `MultilineInput`: an input look-alike `textarea` (`rows=1`, `field-sizing: content`, capped height) that wraps and grows with its content. Enter still submits the form and pasted newlines are stripped, so single-line semantics are preserved. Structured kinds (email, URL, IP, MAC, …) keep the plain input.

## Review follow-ups

The cubic findings are addressed:

- **IME composition** — Enter that confirms an IME composition (`isComposing`) commits the text instead of submitting the form.
- **Double submit** — Enter is ignored while the form is already submitting or validating. The pending submit button is a react-aria button that is never natively `disabled`, so an unguarded `requestSubmit()` fired a second mutation (the single-line input's native implicit submission has the same flaw; the multiline path now guards it).
- **`field-sizing` fallback** — browsers without `field-sizing: content` (Firefox < 152, Safari < 26.2, both inside our browserslist targets) get a JS fallback that syncs the textarea height to its `scrollHeight`, so the value stays readable there too.
- **Edit-form coverage** — the new tests cover both the create form (typed value) and the edit form (pre-populated value).

## Testing: e2e demoted to vitest

The Playwright e2e reproduction is replaced by a colocated vitest browser-mode test, `frontend/app/src/shared/components/form/fields/input.field.test.tsx`. It renders `DynamicField` with the `Text` kind (so the kind→multiline wiring is covered) inside a drawer-width container and reuses the same geometry probe: a field that can scroll is a field that clips (`toBeVisible()` cannot catch this class of bug — the full value is in the DOM, only the rendering clips). Six tests: create-form clipping, edit-form clipping, newline stripping, Enter submits, Enter ignored during IME composition, Enter ignored while a save is pending.

Every test was verified to fail without its fix: reverting `Text` to the single-line input fails both clipping tests, removing the keydown guards fails the IME and pending tests, and disabling both `field-sizing` and its support flag confirms the JS fallback alone keeps the tests green.

## How this was built

This fix came out of the E2E-proof-run experiment for the bug pipeline ([PoC PR #10411](https://github.com/opsmill/infrahub/pull/10411), kept open as reference):

1. **Failing test first** — an e2e test reproduced the bug with the geometry probe described above.
2. **RED verified in CI** — the test-only phase ran against the published `1.11.0` image and failed on its assertion: *340px of the 111-char text hidden* ([run](https://github.com/opsmill/infrahub/actions/runs/32873414287)).
   before
3. **Fix, then GREEN verified in CI** — the same test passed on an image built with the fix ([run](https://github.com/opsmill/infrahub/actions/runs/32874468897)).
   after

Having served its RED/GREEN purpose, the e2e test is demoted to the vitest suite above, which exercises the same component chain at a fraction of the cost.

Local gates: vitest 1238/1238, biome, knip, betterer (188 → 188, results file refreshed for line shifts only), no new tsc errors.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pa-lem:
as stated in the issue, default behaviour is supposed to be inline field with h-scroll, we can navigate through the text by scrolling and by using arrows

it's more a feature request than a bug, so we'll revisit that later if the need comes up again

PR #5966, opened 2025-03-07
Improve node upsert performances

LucasG0:
Improve node upsert performances by going from a check-if-exist-then-create-or-update to a try-create-else-update behavior. This avoids the extra cost of checking whether the node already exists in the database. Some modifications around `NodeGroupedUniquenessConstraint` now differentiate hfid violation than the other ones, as violating hfid uniqueness constraint would mean the node already exists in the database.

Note that the upsert mutation does not benefit from this performance boost if:
- schema defines default_filter
- id is in payload -> we currently do not support (at least officiallly) creating a node with a given id, so it does not matter
- hfid is in payload -> currently, hfid is sent while we upsert sdk side. I believe we don't need it: if we attempt to upsert, it means we may want to create, and thus all mandatory fields required for the node creation should be in the payload, and thus hfid is redondant information. I have a sdk PR here: https://github.com/opsmill/infrahub-sdk-python/pull/312 to test that.

Regarding performances, the current benchmarks do not work properly yet, so they should be ignored (they are also failing here: https://github.com/opsmill/infrahub/pull/6109). I tested locally to create 5000 nodes on non-main branch with this fix versus release-1.2 and performance boost is 12% (76s vs 86s). Based on previous real scenarios observations (cf a slack discussion) I believe this may vary up to 15-20% depending on database state / schemas.

PR #10413, opened 2026-08-25
perf(backend): stop re-reading object templates and the peers a create holds

fatih-acar:
## Summary

Creating an object from an object template read the template, its subtemplates and the peers they name over and over: applying a template walked its relationships one at a time, the whole application ran three times per create, and every component it materialized read back nodes the mutation was already holding. Creating a device from a template carrying ten interfaces issued 118 queries, 37 of them `relationship_get_peer` returning no row. It now issues 32, none of them a peer read. Bulk loads from templates — object files, SDK sync, re-running the same load — pay that cost per object.

## Key Changes

Each is a separate commit carrying a regression test that fails without it:

- The template, and the peers of its component relationships, are read together with their relationships, so the relationships a template leaves empty cost no query (`7dd40c919`)
- The template is read once per create instead of three times: the preview that works out the lock names hands it to the create, and a node built without a template is no longer asked which template it came from — a query every create paid, template or not (`fdb1f0b2b`)
- **That read is now bounded.** Reading a node together with its relationships asked the database for every edge touching it, so reading a template pulled in `related_nodes`, which names every object already created from it, and loaded them all. A prefetch now covers the relationships the caller names (`fdb1f0b2b`)
- The peers the mutation already holds are no longer read back: the grouped uniqueness constraint takes the peer id from the relationship instead of materializing the peer, a component's parent and the node's own template are linked as the nodes they are, and the peer-kind check trusts the kind a peer states (`2623fe97c`, `b34628744`)
- A template's component peers are named from the ids and kinds their relationships already carry (`e01ddc49c`)

## Measured

Create through an upsert mutation. Total queries, with the `relationship_get_peer` among them in brackets:

| Operation | Before | After |
|---|---|---|
| Create, no template | 3 (1) | 2 (0) |
| From a template with no component | 16 (7) | 6 (0) |
| From a template with 3 components | 48 (16) | 18 (0) |
| From a template with 10 components | 118 (37) | 32 (0) |
| Upsert of an existing node from a template | 20 (8) | 14 (2) |

A component now costs 2 queries: the check of its uniqueness constraint and its write. Nothing a create does grows with the number of objects created from that template before — a guard test compares the rows read, not the query count, because the peers come back in one batched read whose size grows silently.

## Related Context

Stacked on #10404, which removed the repeated reads of the payload's own peers; this is the same class of redundancy one layer up, where whole nodes rather than peer ids are re-read. Measurements are taken on top of that branch.

Left in place deliberately, each measured:

- An upsert that updates an existing node still runs a whole create attempt first and throws the result away (about 5 of its 14 queries). #5966 dropped the id and hfid getters from the upsert path to make creates cheaper; reversing that trade costs roughly one query per create to save five per update, which is a call on the create/update mix rather than a bug.
- `_call_mutate_update()` re-reads the node the caller just read, to hold a pristine copy while the payload is applied to the preview (2 queries per update).
- `RelationshipManager.get_peers()` re-reads peers the manager already holds; reusing them safely first needs a node to record which metadata options it was read with.

## Documentation Updates

- `dev/knowledge/backend/templates.md` — what applying a template reads, and what it deliberately leaves alone
- `dev/knowledge/backend/mutations.md` — the scope of a prefetch, and the peers a mutation already holds

## Test Plan

- `uv run pytest backend/tests/unit backend/tests/component` — 6 225 passed at `e01ddc49c`. The one failure, `test_attribute_python_types_cover_every_kind`, is unrelated: `graphql/manager.py` writes a generated class into the module-level `ATTRIBUTE_TYPES` for every enum attribute it builds a schema for, so the test fails in any worker process that has built such a schema and passes in isolation.
- `uv run pytest backend/tests/component backend/tests/unit/graphql backend/tests/unit/core` — 4 551 passed at `f55c6839d`, the tip of this branch.
- The measurements come from a probe that wraps `InfrahubDatabase.execute_query_with_metadata` and records the call stack and the row count of every query; it is not part of the branch.

### DIFF
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -431,6 +431,7 @@ class NodeManager:
         branch: Branch | str | None = ...,
         id: str | None = ...,
         hfid: list[str] | None = ...,
+        prefetch_relationships: bool = ...,
     ) -> SchemaProtocol: ...
 
     @overload
@@ -443,6 +444,7 @@ class NodeManager:
         branch: Branch | str | None = ...,
         id: str | None = ...,
         hfid: list[str] | None = ...,
+        prefetch_relationships: bool = ...,
     ) -> Node: ...
 
     @classmethod
@@ -454,6 +456,7 @@ class NodeManager:
         branch: Branch | str | None = None,
         id: str | None = None,
         hfid: list[str] | None = None,
+        prefetch_relationships: bool = False,
     ) -> Node | SchemaProtocol:
         if id and is_valid_uuid(id):
             return await cls.get_one(
@@ -463,6 +466,7 @@ class NodeManager:
                 branch=branch,
                 at=at,
                 include_metadata=MetadataOptions.LINKED_NODES,
+                prefetch_relationships=prefetch_relationships,
                 raise_on_error=True,
             )
 
@@ -474,6 +478,7 @@ class NodeManager:
                 branch=branch,
                 at=at,
                 include_metadata=MetadataOptions.LINKED_NODES,
+                prefetch_relationships=prefetch_relationships,
                 raise_on_error=True,
             )
 
--- a/backend/infrahub/core/manager.py
+++ b/backend/infrahub/core/manager.py
@@ -431,7 +431,6 @@ class NodeManager:
         branch: Branch | str | None = ...,
         id: str | None = ...,
         hfid: list[str] | None = ...,
-        prefetch_relationships: bool = ...,
     ) -> SchemaProtocol: ...
 
     @overload
@@ -444,7 +443,6 @@ class NodeManager:
         branch: Branch | str | None = ...,
         id: str | None = ...,
         hfid: list[str] | None = ...,
-        prefetch_relationships: bool = ...,
     ) -> Node: ...
 
     @classmethod
@@ -456,7 +454,6 @@ class NodeManager:
         branch: Branch | str | None = None,
         id: str | None = None,
         hfid: list[str] | None = None,
-        prefetch_relationships: bool = False,
     ) -> Node | SchemaProtocol:
         if id and is_valid_uuid(id):
             return await cls.get_one(
@@ -466,7 +463,6 @@ class NodeManager:
                 branch=branch,
                 at=at,
                 include_metadata=MetadataOptions.LINKED_NODES,
-                prefetch_relationships=prefetch_relationships,
                 raise_on_error=True,
             )
 
@@ -478,7 +474,6 @@ class NodeManager:
                 branch=branch,
                 at=at,
                 include_metadata=MetadataOptions.LINKED_NODES,
-                prefetch_relationships=prefetch_relationships,
                 raise_on_error=True,
             )
 
@@ -1208,6 +1203,37 @@ class NodeManager:
         return nodes
 
     @classmethod
+    async def prefetch_relationships(
+        cls,
+        db: InfrahubDatabase,
+        nodes: Iterable[Node],
+        names: set[str],
+        branch: Branch | str | None = None,
+        at: Timestamp | str | None = None,
+        include_metadata: MetadataOptions = MetadataOptions.NONE,
+        branch_agnostic: bool = False,
+    ) -> None:
+        """Read the named relationships of these nodes, and their peers, in a single query.
+
+        The relationships that are not named are left untouched, to be read on demand.
+        """
+        nodes_by_id = {node.get_id(): node for node in nodes}
+        if not nodes_by_id or not names:
+            return
+
+        await cls._enrich_node_dicts_with_relationships(
+            db=db,
+            branch=await registry.get_branch(branch=branch, db=db),
+            at=Timestamp(at),
+            nodes_by_id=nodes_by_id,
+            branch_agnostic=branch_agnostic,
+            include_metadata=include_metadata,
+            prefetch_relationships=True,
+            fields=None,
+            prefetch_relationship_names=names,
+        )
+
+    @classmethod
     async def _enrich_node_dicts_with_relationships(
         cls,
         db: InfrahubDatabase,
@@ -1218,35 +1244,34 @@ class NodeManager:
         include_metadata: MetadataOptions,
         prefetch_relationships: bool,
         fields: dict[str, Any] | None,
+        prefetch_relationship_names: set[str] | None = None,
     ) -> None:
         if not prefetch_relationships and not fields:
             return
         cardinality_one_identifiers_by_kind: dict[str, dict[str, RelationshipDirection]] | None = None
-        outbound_identifiers: set[str] | None = None
-        inbound_identifiers: set[str] | None = None
-        bidirectional_identifiers: set[str] | None = None
-        if not prefetch_relationships:
+        if prefetch_relationships:
+            # Only the relationships these schemas declare are kept below, so reading the peers of
+            # the other edges pointing at these nodes would be reading them to discard them.
+            identifiers_by_direction = _get_relationship_identifiers_by_direction(
+                nodes=nodes_by_id.values(), names=prefetch_relationship_names
+            )
+        else:
             cardinality_one_identifiers_by_kind = _get_cardinality_one_identifiers_by_kind(
                 nodes=nodes_by_id.values(), fields=fields or {}
             )
-            outbound_identifiers = set()
-            inbound_identifiers = set()
-            bidirectional_identifiers = set()
-            for identifier_direction_map in cardinality_one_identifiers_by_kind.values():
-                for identifier, direction in identifier_direction_map.items():
-                    if direction is RelationshipDirection.OUTBOUND:
-                        outbound_identifiers.add(identifier)
-                    elif direction is RelationshipDirection.INBOUND:
-                        inbound_identifiers.add(identifier)
-                    elif direction is RelationshipDirection.BIDIR:
-                        bidirectional_identifiers.add(identifier)
-
+            identifiers_by_direction = _group_identifiers_by_direction(
+                identifier_directions=(
+                    (identifier, direction)
+                    for identifier_direction_map in cardinality_one_identifiers_by_kind.values()
+                    for identifier, direction in identifier_direction_map.items()
+                )
+            )
         query = await NodeListGetRelationshipsQuery.init(
             db=db,
             ids=list(nodes_by_id.keys()),
-            outbound_identifiers=None if outbound_identifiers is None else list(outbound_identifiers),
-            inbound_identifiers=None if inbound_identifiers is None else list(inbound_identifiers),
-            bidirectional_identifiers=None if bidirectional_identifiers is None else list(bidirectional_identifiers),
+            outbound_identifiers=list(identifiers_by_direction[RelationshipDirection.OUTBOUND]),
+            inbound_identifiers=list(identifiers_by_direction[RelationshipDirection.INBOUND]),
+            bidirectional_identifiers=list(identifiers_by_direction[RelationshipDirection.BIDIR]),
             branch=branch,
             at=at,
             branch_agnostic=branch_agnostic,
@@ -1259,7 +1284,7 @@ class NodeManager:
         if not peer_ids:
             if prefetch_relationships:
                 for node in nodes_by_id.values():
-                    node.mark_relationships_as_fetched()
+                    node.mark_relationships_as_fetched(names=prefetch_relationship_names)
             return
 
         missing_peers: dict[str, Node] = {}
@@ -1282,6 +1307,7 @@ class NodeManager:
                 nodes_by_id=nodes_by_id | missing_peers,
                 cardinality_one_identifiers_by_kind=cardinality_one_identifiers_by_kind,
                 insert_peer_node=prefetch_relationships,
+                prefetch_relationship_names=prefetch_relationship_names,
             )
 
     @classmethod
@@ -1293,14 +1319,19 @@ class NodeManager:
         nodes_by_id: dict[str, Node],
         cardinality_one_identifiers_by_kind: dict[str, dict[str, RelationshipDirection]] | None,
         insert_peer_node: bool,
+        prefetch_relationship_names: set[str] | None = None,
     ) -> None:
         if not grouped_peer_nodes.has_node(node_id=node.get_id()):
             if insert_peer_node:
-                node.mark_relationships_as_fetched()
+                node.mark_relationships_as_fetched(names=prefetch_relationship_names)
             return
 
         node_schema = node.get_schema()
         for rel_schema in node_schema.relationships:
+            if insert_peer_node and prefetch_relationship_names is not None:
+                if rel_schema.name not in prefetch_relationship_names:
+                    # The query did not cover this relationship, leave it to be read on demand.
+                    continue
             peer_ids = grouped_peer_nodes.get_peer_ids(
                 node_id=node.get_id(),
                 rel_name=rel_schema.get_identifier(),
@@ -1387,6 +1418,41 @@ class NodeManager:
             await node.delete(db=db, at=at, user_id=user_id)
 
         return nodes_to_delete
+
+
+def _group_identifiers_by_direction(
+    identifier_directions: Iterable[tuple[str, RelationshipDirection]],
+) -> dict[RelationshipDirection, set[str]]:
+    grouped: dict[RelationshipDirection, set[str]] = {
+        RelationshipDirection.OUTBOUND: set(),
+        RelationshipDirection.INBOUND: set(),
+        RelationshipDirection.BIDIR: set(),
+    }
+    for identifier, direction in identifier_directions:
+        grouped[direction].add(identifier)
+    return grouped
+
+
+def _get_relationship_identifiers_by_direction(
+    nodes: Iterable[Node], names: set[str] | None = None
+) -> dict[RelationshipDirection, set[str]]:
+    """Return the identifiers of the relationships the schemas of these nodes declare, per direction.
+
+    `names` restricts the result to the relationships bearing one of these names.
+    """
+    seen_kinds: set[str] = set()
+    identifier_directions: list[tuple[str, RelationshipDirection]] = []
+    for node in nodes:
+        node_schema = node.get_schema()
+        if not node_schema or node_schema.kind in seen_kinds:
+            continue
+        seen_kinds.add(node_schema.kind)
+        identifier_directions.extend(
+            (rel_schema.get_identifier(), rel_schema.direction)
+            for rel_schema in node_schema.relationships
+            if names is None or rel_schema.name in names
+        )
+    return _group_identifiers_by_direction(identifier_directions=identifier_directions)
 
 
 def _get_cardinality_one_identifiers_by_kind(
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -552,9 +552,9 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
             return set()
 
         self._object_template = template
-        # The node records which template it was created from; the template is already read, so it is
-        # named by the node it is rather than by an id to be read again. The rest of the payload is
-        # kept, as it carries the metadata of the relationship.
+        # Hand the template over as the node it is: it has just been read, and the relationship layer
+        # links a node it is given directly, where an id would send it back to the database. The rest
+        # of the payload is kept, as it carries the metadata of the relationship.
         fields[OBJECT_TEMPLATE_RELATIONSHIP_NAME] = {**object_template_field, "id": template}
         pool_allocator = DefaultPoolAllocator(db=db, branch=self._branch) if process_pools else NoOpPoolAllocator()
         applier = NodeTemplateApplier(db=db, branch=self._branch, pool_allocator=pool_allocator)
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -513,6 +513,7 @@ class Node(BaseNode, MetadataInterface, metaclass=BaseNodeMeta):
                 id=object_template_field.get("id"),
                 hfid=object_template_field.get("hfid"),
                 branch=self.get_branch_based_on_support_type(),
+                prefetch_relationships=True,
             )
         except NodeNotFoundError:
             errors.append(
--- a/backend/infrahub/core/node/__init__.py
+++ b/backend/infrahub/core/node/__init__.py
@@ -16,6 +16,7 @@ from infrahub.core.constants import (
     BranchSupportType,
     ComputedAttributeKind,
     InfrahubKind,
+    MetadataOptions,
     RelationshipCardinality,
     RelationshipKind,
 )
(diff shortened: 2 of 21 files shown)


backend/infrahub/templates/node_applier.py
```diff
@@ -23,6 +23,24 @@
     "subscriber_of_groups_for_instances": "subscriber_of_groups",
 }
 
+# The relationship kinds an object template is read for: the values it carries and the components it
+# holds. Its own group memberships describe the template, not its instances, and its TEMPLATE-kind
+# relationship lists every object already created from it, which creating one more never needs.
```
ajtmccarty: I don't know what this means. looks like these are the kinds of relationships that might use a template?

backend/infrahub/core/manager.py
```diff
@@ -1207,6 +1202,37 @@ async def get_many(
...
+
+        The relationships that are not named are left untouched, to be read on demand.
+        """
+        nodes_by_id = {node.get_id(): node for node in nodes}
+        if not nodes_by_id or not names:
+            return
+
+        await cls._enrich_node_dicts_with_relationships(
+            db=db,
+            branch=await registry.get_branch(branch=branch, db=db),
```
ajtmccarty, line 1226: this might need to be `get_branch_based_on_support_type` to handle agnostic objects, but not totally sure

backend/infrahub/core/relationship/model.py
```diff
@@ -1051,6 +1051,30 @@ async def get_peer(
...
+        """Return the id of the peer of this relationship without reading the peer itself.
+
+        A peer named by a human-friendly id or by a default filter value is still read, as reading it
+        is the only way to know which node it is.
+
+        Raises:
+            TypeError: When the relationship has a cardinality of many.
+
+        """
+        if self.schema.cardinality == "many":
```
ajtmccarty: should use the enum

backend/infrahub/core/node/__init__.py
```diff
@@ -517,6 +536,11 @@ async def handle_object_template(
             )
             return set()
 
+        self._object_template = template
+        # The node records which template it was created from; the template is already read, so it is
+        # named by the node it is rather than by an id to be read again. The rest of the payload is
+        # kept, as it carries the metadata of the relationship.
```
ajtmccarty: is this trying to say, "add the template ID so we don't have to look it up later"?

backend/infrahub/core/node/create.py
```diff
@@ -94,29 +113,32 @@ async def extract_peer_data(
...
+        peer_ids = [item.peer_id for item in relationships]
         if rel_manager.schema.kind in [
             RelationshipKind.COMPONENT,
             RelationshipKind.PARENT,
             RelationshipKind.PROFILE,
-        ] and list(peers_map.keys()) == [current_template.id]:
-            obj_peer_data[rel] = {"id": parent_obj.id}
+        ] and peer_ids == [current_template.id]:
+            # The peer is the node this one is created under, which the caller holds: naming it by
+            # its id would have every step that needs the peer read it back.
```
ajtmccarty: should describe what it is doing instead of what it is not doing

backend/infrahub/core/node/create.py
```diff
@@ -94,29 +113,32 @@ async def extract_peer_data(
...
+            obj_peer_data[rel] = parent_obj
             continue
 
         rel_peer_ids = []
-        for peer_id, peer_object in peers_map.items():
+        for item in relationships:
```
ajtmccarty: if the name is changing, might as well make it better: `rel` or `relationship`

backend/infrahub/core/relationship/constraints/peer_kind.py
```diff
@@ -38,26 +44,38 @@ async def check(self, relm: RelationshipManager, node_schema: MainSchemaTypes, n
...
-        async for peer_node in peers_query.get_nodes(db=self.db, duplicate=False):
-            if not peer_node.schema:
-                raise ValueError(f"Cannot identify schema for node {peer_node.node_uuid}")
-            if peer_node.schema.kind not in allowed_kinds:
+        # A peer that is already in hand states its own kind, only the ones named by an id are read.
+        peer_ids_to_read: list[str] = []
+        for relationship in relationships:
+            peer_id = relationship.peer_id
+            if peer_id is None:
+                continue
```
ajtmccarty, line 56: seems like an error. maybe `get_peer_id()` instead here

backend/tests/component/core/constraint_validators/test_relationship_peer_kind.py
```diff
@@ -0,0 +1,45 @@
...
+) -> None:
+    """A peer the caller hands over states its own kind."""
+    car = await Node.init(db=db, schema="TestCar", branch=default_branch)
+    await car.new(db=db, name="accord", nbr_seats=5, is_electric=False, owner=person_john_main)
+    counting_db = CountingInfrahubDatabase.from_db(db=db)
+    constraint = RelationshipPeerKindConstraint(db=counting_db, branch=default_branch)
+
+    await constraint.check(relm=car.owner, node_schema=car.get_schema(), node=car)
+
+    assert counting_db.count_for(NodeListGetInfoQuery.name) == 0
```
ajtmccarty, line 23: I guess this is the best option we have right now for testing this kind of thing based on how tangled and recursive our code is for updating/creating relationships, but this is not great testing practice. this kind of test should be concerned with functional outcomes, not precisely which queries are called how many times.

Unfortunately, I think the real most-correct solution is a big refactoring of how all of these parts fit together and that's not going to be a quick fix.

backend/tests/component/templates/test_template_reads.py
```diff
@@ -0,0 +1,162 @@
...
+            data={"name": f"device-{idx}", "object_template": {"id": template.id}},
+            db=counting_db,
+            branch=default_branch,
+            schema=device_schema_obj,
+        )
+        counts.append(sum(counting_db.query_counts.values()))
+        rows_read.append(sum(counting_db.row_counts.values()))
+
+    assert len(set(counts)) == 1, f"the cost of a create grew with the objects already created: {counts}"
+    assert len(set(rows_read)) == 1, f"a create read more of the database for each object created: {rows_read}"
```
ajtmccarty, line 154: I think this a good one though.

backend/infrahub/core/node/__init__.py
```diff
@@ -485,24 +487,41 @@ async def handle_pool(
...
+        """Read the template this node is created from, together with the relationships it is read for."""
+        branch = self.get_branch_based_on_support_type()
+        template: CoreObjectTemplate = await registry.manager.find_object(
+            db=db,
+            kind=self._schema.get_relationship(name=OBJECT_TEMPLATE_RELATIONSHIP_NAME).peer,
+            id=object_template_field.get("id"),
+            hfid=object_template_field.get("hfid"),
+            branch=branch,
+        )
+        await registry.manager.prefetch_relationships(
```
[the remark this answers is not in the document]
fatih-acar: Correct about the mechanism, and it is already listed under "Left in place deliberately" in the description: `RelationshipManager.get_peers()` calls `registry.manager.get_many()` on the peer ids regardless of what the relationships already hold.

It does not make the prefetch pointless, though:

- `get_peer()` (cardinality one) *does* reuse it — `Relationship.get_peer()` returns `self._peer` when it is set. The relationships `NodeTemplateApplier._apply_relationships` reads off a template (`Attribute`/`Generic`/`Profile` kinds) are mostly cardinality one, and those are served straight from the prefetch.
- For every relationship, hydrated or not, the prefetch replaces one `RelationshipGetPeerQuery` *per relationship* with one `NodeListGetRelationshipsQuery` for all of them. That is the 37 -> 0 `relationship_get_peer` drop in the table. What `get_peers()` repeats is one batched node read, not one read per peer.

Making `get_peers()` reuse the cached peers needs a node to record which metadata options it was read with — otherwise a cached peer is handed back missing the source/owner properties the caller asked for, silently. That is a separate change and it is called out in the description.

backend/infrahub/core/node/create.py
```diff
@@ -347,6 +370,9 @@ class that is not a node schema protocol is passed.
     await preview_obj.new(db=db, process_pools=False, **data)
     schema_branch = db.schema.get_schema_branch(name=branch.name)
     lock_names = get_lock_names_on_object_mutation(node=preview_obj, schema_branch=schema_branch)
+    # The preview read the object template to work out the lock names; the node created under the
+    # lock is built from that same template rather than reading it again.
+    object_template = preview_obj._object_template
```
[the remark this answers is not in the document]
fatih-acar: The lock never gave that guarantee.

`get_lock_names_on_object_mutation()` (`backend/infrahub/core/node/lock_utils.py`) builds names from resource pool ids, cardinality-one relationship identifiers and uniqueness constraint values **of the node being created**. None of them names the template. Updating a template is a mutation on `CoreObjectTemplate` and takes locks derived from *its* schema, so the two never contend: the template could always change between the read and the write, inside the lock or outside it.

The read this commit removes was also never the first one. `preview_obj.new()` read the template before the lock on `develop` as well, and the lock names themselves are computed from that pre-lock preview. What changed is that the template is read once instead of three times, not which side of the lock the first read falls on.

backend/tests/component/core/constraint_validators/test_node_grouped_uniqueness.py
```diff
@@ -148,6 +150,20 @@ async def test_uniqueness_constraint_no_conflict_one_relationship(
 
         await self.__call_system_under_test(db=db, branch=default_branch, node=car_node)
 
+    async def test_uniqueness_constraint_relationship_does_not_read_the_peer(
+        self, db: InfrahubDatabase, default_branch: Branch, car_person_generics_data_simple: dict[str, Node]
+    ) -> None:
+        """The constraint compares the peer this node points at, which its id identifies."""
+        car_node: Node = car_person_generics_data_simple["c1"]
+        car_node.get_schema().uniqueness_constraints = [["previous_owner"]]
```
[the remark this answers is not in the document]
fatih-acar: Valid — thanks, this one was worth catching. `c1` never gets a `previous_owner` in the fixture, so `get_peer_id()` returns `None` off an empty relationship list and the old `get_peer()` would have returned `None` just as cheaply. The test asserted nothing.

The suggested patch does not fix it, though: `update(data=person_1)` sets `Relationship._peer` to the `Node` object itself, and `Relationship.get_peer()` returns a cached `_peer` without a query. I applied that exact suggestion against the pre-commit implementation and it still passes — so it would not guard either.

The relationship has to carry the peer *id* alone, which is what it looks like when the node comes back from the database. Fixed by assigning the peer, saving, and reading the car back before counting:

```python
person: Node = car_person_generics_data_simple["p2"]
car: Node = car_person_generics_data_simple["c1"]
await car.previous_owner.update(db=db, data=person)
await car.save(db=db)
# Read the car back, so its relationship carries the id of its peer rather than the peer itself.
car_node = await registry.manager.get_one(db=db, id=car.get_id(), branch=default_branch, raise_on_error=True)
car_node.get_schema().uniqueness_constraints = [["previous_owner"]]
counting_db = CountingInfrahubDatabase.from_db(db=db)

await self.__call_system_under_test(db=counting_db, branch=default_branch, node=car_node)

assert await car_node.previous_owner.get_peer_id(db=db) == person.get_id()
assert counting_db.count_for(UniquenessValidationQuery.name) > 0
assert counting_db.count_for(NodeListGetInfoQuery.name) == 0
assert counting_db.count_for(NodeListGetAttributeQuery.name) == 0
```

Verified both ways: it passes on this branch, and against the reverted `get_peer()` implementation it fails with `assert 1 == 0` on `node_list_get_info`.

backend/infrahub/templates/node_applier.py
fatih-acar: The other way round, and that is my fault for writing it as a list of exclusions. These are the relationship kinds read *off the template* when one is applied — a template is read for the values it carries and the components it holds, and those live under `Attribute`, `Generic`, `Profile`, `Component` and `Parent`. Nothing here is about which relationships may use a template.

Reworded, and it now says why the excluded kinds matter rather than just that they exist:

```python
# Applying a template reads the values it carries and the components it holds; these are the
# relationship kinds those live under. Every other kind is left unread, which matters most for
# TEMPLATE: a template's TEMPLATE-kind relationship lists every object already created from it, so
# reading it would make each create more expensive than the last. GROUP is left out too — a
# template's own group memberships describe the template, not its instances, whose memberships come
# from the `*_for_instances` relationships mapped above.
```

backend/infrahub/core/manager.py
fatih-acar: `get_branch_based_on_support_type()` is a method on `Node`/`Attribute`/`Relationship` — it answers "is *this* schema branch-agnostic", so the manager cannot call it for a list of nodes that need not share a schema. Agnostic reads reach this method the same two ways they reach the rest of `NodeManager`:

- the caller resolves the branch — `Node._read_object_template()`, the only caller today, passes `self.get_branch_based_on_support_type()`;
- `branch_agnostic=True`, which `prefetch_relationships()` accepts and threads into `NodeListGetRelationshipsQuery` and the peer `get_many()`.

`registry.get_branch(branch=branch, db=db)` here is exactly what `get_many()` does. `prefetch_relationships()` is `get_many(prefetch_relationships=True)` with the relationship names narrowed, so I kept the branch resolution identical rather than making this one method behave differently from its neighbours. Leaving it as is — say the word if you would rather I spell that out in the docstring.

backend/infrahub/core/relationship/model.py
fatih-acar: Agreed. Done — and since `RelationshipCardinality` had to be imported into the module anyway, I swept the five other string comparisons in the same file (`get_peer`, `__iter__`, `get`, `to_graphql`) so there is no mix of both styles left.

backend/infrahub/core/node/__init__.py
fatih-acar: The opposite, actually — which means the comment failed. The payload already carried the template id; that is how the template was found a few lines up. What this line does is *replace* that id with the template node object we have just read, so the relationship layer links the node it is handed instead of reading it back out of the database.

Reworded:

```python
# Hand the template over as the node it is: it has just been read, and the relationship layer
# links a node it is given directly, where an id would send it back to the database. The rest
# of the payload is kept, as it carries the metadata of the relationship.
```

backend/infrahub/core/node/create.py
fatih-acar: Fair. Rewritten to say what the branch does:

```python
# This relationship points back at the template being applied, so the peer on the new
# object is the node created from that template — which the caller holds. Hand the node
# over, so every step that needs the peer already has it.
```

backend/infrahub/core/node/create.py
fatih-acar: Done — `relationship` throughout, in the comprehension above and in this loop.

`rel` was not available: the outer loop binds it to a relationship *name* (`for rel in template_peer.get_schema().relationship_names`), which is its own bit of confusion. I renamed that one to `rel_name` in the same pass so `relationship` is unambiguous here.

backend/infrahub/core/relationship/constraints/peer_kind.py
fatih-acar: `Relationship.get_peer_id()` raises `ValueError` when `peer_id` is unset, and it is legitimately unset here: a peer given by hfid keeps `peer_id = None` and holds only `peer_hfid` until `resolve()` runs — and `resolve()` runs *after* this constraint. `RelationshipPeerKindConstraint` is built before `RelationshipCountConstraint` in `dependencies/builder/constraint/grouped/node_runner.py`, and the count constraint is the one that calls `relm.resolve()`. Raising here would reject every hfid-named peer.

The `continue` also preserves the behaviour exactly: before this commit, `peer_ids = [r.peer_id for r in relationships]` put that `None` straight into `NodeListGetInfoQuery`, where it matched nothing. Such a peer is not left unchecked either — resolving an hfid goes through `get_one_by_hfid(kind=self.schema.peer)`, so a node of another kind is simply not found.

You are right that it reads as a swallowed case, so I added a comment:

```python
peer_id = relationship.peer_id
if peer_id is None:
    # The peer is named by an hfid the relationship has not resolved yet: this constraint
    # runs before RelationshipCountConstraint, which is what calls resolve(). Resolving an
    # hfid reads it as the schema's peer kind, so a peer of another kind is not found.
    continue
```

backend/tests/component/core/constraint_validators/test_relationship_peer_kind.py
fatih-acar: Agreed, and I would rather not have written it either.

The two tests below it are the functional ones — a wrong-kind peer is rejected whether it is handed over as a node or named by an id — and both pass on either implementation. Line 23 exists only so that the read this commit removes does not come back unnoticed. It is the guard, not the coverage.

On the refactor: agreed that is the real fix, and well out of scope for a perf branch. Want me to open an issue for it so it is tracked somewhere other than this thread?

backend/tests/component/templates/test_template_reads.py
fatih-acar: That one compares rows read rather than queries run, deliberately: the peers of a template come back in one batched read whose *size* grows with the number of objects already created from it, so a query count would sit flat at 1 while the cost kept climbing.

backend/infrahub/templates/node_applier.py
```diff
@@ -23,6 +23,27 @@
...
+# from the `*_for_instances` relationships mapped above.
+TEMPLATE_APPLICATION_RELATIONSHIP_KINDS = frozenset(
+    {
+        RelationshipKind.ATTRIBUTE,
+        RelationshipKind.COMPONENT,
+        RelationshipKind.GENERIC,
+        RelationshipKind.PARENT,
+        RelationshipKind.PROFILE,
+    }
+)
```
gmazoyer: This is now a hard list, and anyone adding a new relationship kind to a template has to remember to come here (not sure if it'll happen though). Can we get a comment saying that, and a test that fails if a template kind is left unread?

Also this comment seems over-complicate to me, the code is pretty explicit and such comments is already be part of the LLM knowledge doc about the object template feature (added in this PR).

backend/tests/component/templates/test_template_reads.py
gmazoyer, line 1: The new tests in only go one level deep, template with interfaces. The nested path, where a component is itself a template holding its own components, is covered only by `test_create_with_object_template`, which predates this PR. That test is what guards the customer scenario (graphql mutation), so if this backports to 1.10 it is the one to keep an eye on. A query count test at depth two would be a good addition here.

gmazoyer (commented):
There are a lot of changelogs in this PR and they seem to read most about code internals. It could be a good thing to rewrite them to make them about what a regular user would care about/understand.

fatih-acar:
@gmazoyer good call — they did read as a list of reads the code no longer makes, which is the shape of the commits rather than anything a user can see.

The six fragments this PR added are now a single entry (`98255e0`):

> Creating objects from an object template is now significantly faster, and no longer slows down as more objects are created from the same template. Bulk loads from templates, such as loading object files or synchronizing with the SDK, benefit the most.

I deliberately left the query counts out of it — "32 queries instead of 118" is exactly the internal unit you flagged, and the measured table in the description is the right home for it, as review material rather than release notes.

The fragments still in `changelog/` belong to #10404, the branch this one is stacked on; I have not touched those.

backend/infrahub/templates/node_applier.py
fatih-acar: Both done in `c709cc3`.

The comment is gone — you are right that it duplicates `dev/knowledge/backend/templates.md`, which this PR adds and which already lists the kinds and why the others are left out. What is left is only the bit that is not in the doc, that the list has to be maintained:

```python
# A hard list: a relationship kind that is not named here is never read when a template is applied,
# so a kind newly added to a template has to be classified here.
# `test_relationship_kinds_on_templates_are_classified` fails while one is left unclassified.
```

And the test, in `test_template_applier.py`: it collects every relationship kind reachable on a generated template schema and fails unless each one is either in `TEMPLATE_APPLICATION_RELATIONSHIP_KINDS` or in an explicit `TEMPLATE_RELATIONSHIP_KINDS_NOT_READ` list next to the test, so adding a kind to a template forces a decision rather than silently skipping the read. Six of the eight kinds are reachable under the device schema (Component, Generic, Group, Parent, Profile, Template). I checked it actually bites: dropping `PARENT` from the hard list fails it with

> `[RelationshipKind.PARENT] appears on a template but is neither read when one is applied nor listed as deliberately unread.`

backend/tests/component/templates/test_template_reads.py
fatih-acar: Added in `c709cc3` — `test_creating_from_a_nested_template_reads_no_peer_at_either_depth`. The test schema already had the shape: device -> interface -> SFP, so `_build_template` now optionally hangs an SFP subtemplate off each interface subtemplate.

It holds at depth two: no `relationship_get_peer` at either level, the SFPs are materialized with the right serials, and the cost stays constant per interface across 1, 3 and 5 interfaces.

It did turn up something worth knowing, though, which is why I measured it per query rather than just asserting a total. **The second level is not as cheap as the first.** An interface costs the same 2 queries a first-level component costs, but the SFP under it costs 6:

| | per extra interface, flat | per extra interface, with an SFP |
|---|---|---|
| `node_create_all` | 1 | 2 |
| `uniqueness_constraint_validation` | 1 | 2 |
| `node_list_get_info` / `node_list_get_attribute` | 0 | 1 / 1 |
| `node_list_get_relationship` | 0 | 1 |
| `relationship_count_per_node` | 0 | 1 |
| **total** | **2** | **8** |

The three read queries are the recursion prefetching each subtemplate's own relationships **one parent at a time** instead of once for the whole level — with three interfaces you get three prefetches where one would do. It is constant per interface, so a wide tree does not degrade, but a deep one pays per branch.

That is not a regression from this PR — the recursion has that shape on `develop` too, and this PR is what removed the per-relationship peer reads underneath it. I left it alone here rather than widen the branch, recorded it in the knowledge doc as the next saving to take, and the new test locks in the current figure so it cannot drift while someone gets to it. Happy to open an issue for prefetching a level at a time if you would rather it be tracked.

dev/knowledge/backend/templates.md
```diff
@@ -64,6 +64,35 @@ When a user creates an object with `object_template={id: ...}`, `Node.handle_obj
...
+
+The cost of a create is then the template read (3 queries), the node's own uniqueness check and
+write, and 2 queries per component (its uniqueness check and its write). It does not grow with the
+number of objects created from the template before.
+
+That 2-query figure is for a component at the first level. A component that is itself a template
+holding components of its own costs more, because the recursion prefetches each subtemplate's
+relationships one parent at a time rather than once for the whole level: with the test schema's
+device -> interface -> SFP, an interface costs 2 and the SFP under it costs 6 (a batched
+relationship read, a batched peer read, its own count constraint, its uniqueness check and its
```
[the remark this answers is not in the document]
fatih-acar: Valid, and the total was the right number — the enumeration was wrong. Reading a peer back is two queries, not one: `node_list_get_info` for the node and `node_list_get_attribute` for its attributes. I collapsed them into "a batched peer read", so five phrases stood where six queries were measured.

Fixed in `8d27698`; the peer read now says it is two queries, so the list adds up to 6:

> an interface costs 2 and the SFP under it costs 6: a batched relationship read, a batched peer read (two queries, one for the node and one for its attributes), its own count constraint, its uniqueness check, and its write.

The measured per-extra-interface breakdown behind it, for the record:

| query | flat | with an SFP |
|---|---|---|
| `node_create_all` | 1 | 2 |
| `uniqueness_constraint_validation` | 1 | 2 |
| `node_list_get_info` | 0 | 1 |
| `node_list_get_attribute` | 0 | 1 |
| `node_list_get_relationship` | 0 | 1 |
| `relationship_count_per_node` | 0 | 1 |
| **total** | **2** | **8** |

The remaining saving is now tracked in #10419.

backend/tests/component/templates/test_template_reads.py
fatih-acar: Opened as #10419, with the per-query breakdown and a pointer to the test that measures it.

ajtmccarty (approved):

PR #10414, opened 2026-08-26
test(backend): stop one wedged Prefect test server from costing a whole component job

fatih-acar:
# Why

One Prefect test server that stops answering currently costs a whole `backend-tests-component` job. On develop run [32874262393](https://github.com/opsmill/infrahub/actions/runs/32874262393/job/97888958473), gw0's ephemeral server went unresponsive at 16:57 and the job died 35 minutes later at the 1800s xdist session timeout — 3 failures, 15 errors. The same family hit develop again the same morning (run 32818780299), that time against the Prefect container.

Two things made it that expensive, and this PR fixes both. It does **not** fix the wedge itself — that root cause is still unknown, which is precisely what the second commit is for.

## What changed

**A hung task-manager setup is now remembered the first time it hangs.**

`setup_task_manager_once()` already memoizes a failure so that one wedged server doesn't cost a 300s pytest timeout per test. It never armed: pytest-timeout runs in signal mode, so SIGALRM is raised in whichever frame the main thread is in — for an async fixture that is the event loop driver, several frames above the coroutine. `run_once`'s `except BaseException` never saw it, nothing was recorded, and the next test started the setup over. Three tests burned the full 300s each before the memo finally armed (when an abandoned coroutine raised `ReadTimeout` on its own), and those abandoned coroutines stayed pending on the session-scoped loop, going on to hang `profiles/test_gather.py` — which never touches Prefect — for another 900s.

The setup now carries its own 180s ceiling, so it fails *inside* the coroutine. The margin is deliberate: a false timeout is memoized too, and would condemn a healthy server for the rest of the session, so the budget sits ~6x above a healthy setup and well under the 300s pytest timeout.

**A wedged server can now be asked what it is doing.**

Previously nothing described the server itself — blocked, it logs nothing, and the tests only ever see their own timeouts. The ephemeral server now boots through an entry point that answers SIGUSR1 with the stack of every one of its threads (`chain=False`; chaining would let the signal's default action kill the server), its output goes to a file per port, and both the stacks and the log tail are printed when a Prefect call gives up:

```
--- Prefect test server on port 28354 (pid 3060674) ---
Thread ... [Thread-3 (_conn] (most recent call first):
  File ".../aiosqlite/core.py", line 59 in _connection_worker_thread
...
Current thread ... [python] (most recent call first):
  File ".../uvicorn/server.py", line 65 in run
```

That is enough to confirm or kill the standing SQLite-contention hypothesis (see the comment on `PREFECT_SERVER_NONESSENTIAL_SERVICE_ENV_VARS`) the next time a server wedges.

**Note for the merge-up:** `stable` globally ignores the `BLE` ruff rules and `develop` does not, so the blanket `except Exception` guarding the report (a broken diagnostic must never replace the failure that triggered it) needs a `# noqa: BLE001` on `develop` and must not carry one here — RUF100 flags it as unused on `stable`. That is the one line that will need adding when this reaches `develop`.

**What stayed the same:** no product code touched — this is entirely test infrastructure. The server's own configuration (env, log level, port range) is unchanged.

## How to review

`backend/tests/helpers/task_manager.py` and `backend/tests/helpers/prefect_diagnostics.py` are the substance; the conftest change is just the launcher wiring.

The value worth scrutiny is `SETUP_TIMEOUT_SECONDS = 180`. Too low and a slow-but-healthy CI runner gets its server condemned for the whole session; too high and the failure escapes to the pytest timeout, which is the bug being fixed here.

## How to test

```bash
uv run pytest --no-cov backend/tests/unit/helpers
uv run pytest --no-cov --neo4j backend/tests/component/test_prefect_test_server_diagnostics.py
```

Validated locally beyond CI:

- The real ephemeral server answers a stack-dump request **and survives it** (asserted by the new component test) — a server killed by the signal asking what it was doing would take every later test with it.
- The whole failure path, by SIGSTOPping the live server: `TimeoutError` at the budget → memoized → next call fails fast with `RuntimeError: already failed` → report printed. A stopped server reports "neither logged nor answered the stack dump", which usefully distinguishes unschedulable from blocked-in-Python.
- `backend/tests/component/api/test_40_schema.py` + `backend/tests/component/profiles` (the modules that failed in the run above): 37 passed.
- `uv run invoke format`, `lint`, and `backend.validate-component-shards` clean.

## Impact & rollout

- **Backward compatibility:** test-only; no runtime behaviour changes.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry added (test-infrastructure only; no user-visible change)

### DIFF
--- a/backend/tests/component/conftest.py
+++ b/backend/tests/component/conftest.py
@@ -81,6 +81,7 @@ from tests.helpers.constants import (
     PREFECT_TEST_SERVER_PORT_RANGE,
 )
 from tests.helpers.file_repo import FileRepo
+from tests.helpers.prefect_diagnostics import register_prefect_test_server
 from tests.helpers.test_client import dummy_async_request
 from tests.helpers.utils import find_available_prefect_port
 from tests.test_data import dataset01 as ds01
@@ -128,9 +129,17 @@ def neo4j_factory() -> _GraphHydrator:
 
 
 @pytest.fixture(scope="session", autouse=True)
-def prefect_test_fixture() -> Generator[None, None, None]:
+def prefect_test_fixture(tmp_path_factory: pytest.TempPathFactory) -> Generator[None, None, None]:
+    log_dir = tmp_path_factory.getbasetemp()
+
     def _run_uvicorn_command(self: Any) -> subprocess.Popen[Any]:
-        """Patched version of prefect method to call the test server, pointing at the Infrahub entrypoint instead."""
+        """Patched version of prefect method to call the test server, pointing at the Infrahub entrypoint instead.
+
+        The server is launched through a wrapper that answers SIGUSR1 with the stacks of all its
+        threads, and its output is redirected to a file rather than left to interleave with the
+        worker's own. A server that stops answering is otherwise invisible in CI: blocked, it
+        logs nothing at all, and the tests only ever see their own timeouts.
+        """
         # used to turn off serving the UI
         server_env = {
             "PREFECT_UI_ENABLED": "0",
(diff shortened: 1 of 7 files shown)


fatih-acar:
reopening to pick up the recomputed merge ref for the new base

ogenstad (approved):

PR #10416, opened 2026-08-26
Add the Python transform target deriver

gmazoyer:
# Why

A merge or a rebase recomputes Jinja2 computed attributes, display labels and human-friendly ids
through one coalesced pass. Python transform computed attributes are excluded from it, so their
automations start one task per changed node: 68 tasks against 3 coalesced ones on a site merge,
148 on a rebase.

This is the first of three PRs ([IFC-3017](https://opsmill.atlassian.net/browse/IFC-3017)). It adds
the missing piece and nothing else. IFC-3018 then runs the coalesced Python pass alongside the
per-node automations, and IFC-3019 gates those automations on `origin == live`, which is what
removes the fan-out.

## What changed

Nothing calls the resolver yet, so behaviour does not change.

* A resolver that maps a change signature `(kind, action, changed fields)` to the affected Python
  `(kind, attribute)` pairs and to the nodes to recompute for each. Read sets come from the
  transform GraphQL queries, readers from the query-group subscriber index.
* A `whole_kind` flag on the affected target. A widened target carries no node ids, and chunking an
  empty id set produces no submission, so without the flag the fallback becomes a silent skip.
* The setting `INFRAHUB_COALESCE_PYTHON_RECOMPUTE_AFTER_MERGE`, on by default so the scenario
  harness in the private test repository takes the coalesced path with no change to its CI setup.
  Expect the switch to go away once the three PRs are in and the measured counts prove the pass
  works.

Chain plumbing on `process_transform` stays out of scope, it belongs to IFC-3018, so the Python
family carries no `recompute_depth` yet.

## How to review

Is the narrowing correct, and does it fail safe when the reader lookup fails? Over recompute is
acceptable, under recompute is not. Start with `python_target_resolution.py` and its unit tests.
Two calls worth a look: reader lookups are memoised per set of changed ids rather than shared
across every attribute, and a deleted node has no query-group membership left on the destination
branch, so its readers have to arrive as updated nodes with a changed relationship field.

The changes to the generated files all come from the new setting.

## How to test

```bash
uv run pytest backend/tests/unit/core/merge -q
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/merge_recompute_coalescing -q
```

Green locally. The two component suites are there to show that behaviour is unchanged.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry: one fragment covers the whole stack, on IFC-3019

[IFC-3017]: https://opsmill.atlassian.net/browse/IFC-3017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -0,0 +1,247 @@
...
+    set, so that it widens instead of dropping out of the change set unnoticed.
+    """
+
+    async def read_sets(self, *, branch: str) -> list[PythonAttributeReadSet]: ...
+
+
+class PythonSubscriberSource(Protocol):
+    """The nodes subscribed to a query group that holds any of ``node_ids`` as a member."""
+
+    async def subscribers(self, *, node_ids: list[str], branch: str) -> list[SubscriberRef]: ...
```
polmichel, line 64: Regarding the work I've done on generator part, I'll replace my ProposedChangeSubscriber by your object as this looks pure duplication
https://opsmill.atlassian.net/browse/IFC-3070

backend/infrahub/computed_attribute/read_sets.py
```diff
@@ -0,0 +1,40 @@
...
+        for kind, fields in read_fields_by_kind.items()
+        if derived_reads_are_scopable(schema_branch=schema_branch, kind=kind, read_fields=frozenset(fields))
+    }
+
+    return TransformReadSet.from_read_fields(read_fields_by_kind, scopable_derived_kinds=scopable_derived_kinds)
+
+
+def derived_reads_are_scopable(*, schema_branch: SchemaBranch, kind: str, read_fields: frozenset[str]) -> bool:
+    """Whether every derived field read on one kind can be held against that kind alone."""
+    derived_reads = read_fields & IMPRECISE_READ_FIELDS
```
polmichel, line 35: I've not captured this imprecise field concept inside the code I've done regarding generators and artifacts, I'll see if I need this concept in this release https://opsmill.atlassian.net/browse/IFC-3071

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -0,0 +1,247 @@
...
+    async def resolve(self, *, changes: Iterable[MergeChange]) -> list[AffectedTarget]:
+        """Derive the affected Python computed attributes and the nodes to recompute for each.
+
+        Changes are grouped by their (kind, action, changed fields) signature so the narrowing runs
+        once per distinct shape. Targets are deduplicated per (kind, attribute) across the whole
+        change set and returned in a deterministic order.
+        """
+        ids_by_signature: dict[ChangeSignature, set[str]] = {}
+        for change in changes:
+            signature = ChangeSignature(kind=change.kind, action=change.action, changed_fields=change.changed_fields)
```
polmichel: It looks like a derivative of a `MergeChange` object, do you think that would be a good idea to create a property inside `MergeChange` returning this object?

polmichel: Moreover this code looks duplicated over here `infrahub.core.merge.recompute_coalescing.CoalescedRecomputeBuilder.build` wdyt about factorising it somewhere?

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -0,0 +1,247 @@
...
+    async def subscribers(self, *, node_ids: list[str], branch: str) -> list[SubscriberRef]: ...
+
+
+@dataclass(frozen=True)
+class _Selection:
+    """Why one change signature selects one attribute, and how exactly."""
+
+    self_target: bool
+    widen: bool
+    precise: bool
```
polmichel, line 78: Optional comment, maybe it'd be worth to define what these attributes - and the future `reader_lookup`/`self_ids` - mean inside the docstring

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -0,0 +1,247 @@
...
+
+    """
+    if signature.action == CREATED:
+        # A created node subscribes to no query group yet, so it can only be its own target.
+        return _Selection(self_target=True, widen=False, precise=True) if attribute.kind == signature.kind else None
+
+    if signature.action not in {UPDATED, DELETED}:
+        raise ValueError(f"Unknown change action: {signature.action!r}")
+
+    return _select_reader(signature=signature, read_set=attribute.read_set)
```
polmichel: Could `signature.action` be a `StrEnum` instead of a raw string? It would allow a `match case`/`assert_never` structure here. Wdyt from maintainbility perspective?

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -0,0 +1,275 @@
...
+    """Why one change signature selects one attribute, and how exactly.
+
+    ``self_ids`` and ``reader_lookup`` are independent: a changed node can be both a target of
+    its own and a source whose readers have to be resolved.
+    """
+
+    widen: bool
+    self_ids: bool
+    reader_lookup: bool
+    precise: bool
```
polmichel, line 78: Optional comment as well: regarding `_Selection`: I find it a little hard to use. It carries four booleans, and most of their combinations aren't valid, only `self_ids` and `reader_lookup` are meant to be combined, while `widen` excludes the rest. What do you think about making the legal cases explicit instead?

Two possible options come to my mind to do so:
- A small class hierarchy where each variant implements the behavior and the base `_Selection` is abstract:
```
@dataclass(frozen=True)
  class _WidenSelection(_Selection):
      def accumulate_into(self, acc, *, node_ids, deleted):
          acc.mark_whole_kind()


  @dataclass(frozen=True)
  class _SelfTargetSelection(_Selection):
      def accumulate_into(self, acc, *, node_ids, deleted):
          acc.add_self(node_ids)


  @dataclass(frozen=True)
  class _ReaderSelection(_Selection):
      self_ids: bool = False
      precise: bool = True

      def accumulate_into(self, acc, *, node_ids, deleted):
          if self.self_ids:
              acc.add_self(node_ids)
          acc.add_source(node_ids, deleted=deleted)
          if not self.precise:
              acc.mark_imprecise()
```
- - A union of small dataclasses, one per case: 
```
 @dataclass(frozen=True)
  class _DeletedReader:
      """A deletion of a read kind: resolve its readers, kept apart from live sources."""


  @dataclass(frozen=True)
  class _LiveReader:
      """An update of a read kind: resolve its readers, and itself when it is the own kind.

      ``precise`` is False when the selection is a deliberate over-approximation, never when wrong.
      """

      self_ids: bool
      precise: bool
  _Selection = _Widen | _SelfTarget | _DeletedReader | _LiveReader
```

It would allow us to lock down the configuration available and which makes sense regarding these boolean while giving a clear meaning to each combination but I am sure other solutions exist.
Maybe I am also missing something here.

backend/tests/adapters/python_target_sources.py
polmichel, line 1: Nice fake x3! This is clean

backend/tests/unit/core/merge/test_python_target_resolution.py
```diff
@@ -0,0 +1,325 @@
...
+
+    targets = await resolver.resolve(
+        changes=[MergeChange(node_id="s1", kind=SITE, action="updated", changed_fields=frozenset({"name"}))]
+    )
+
+    assert subscribers.calls == [("s1",)]
+    assert _identities(targets) == [(DEVICE, "summary")]
+    assert targets[0].whole_kind is True
+    assert targets[0].precise is False
+    assert targets[0].reader_lookups == frozenset()
```
polmichel, line 211: Nice test, this is key 👍

backend/tests/unit/core/merge/test_python_target_resolution.py
```diff
@@ -0,0 +1,325 @@
...
+    targets = await resolver.resolve(
+        changes=[
+            MergeChange(node_id="d1", kind=DEVICE, action="created"),
+            MergeChange(node_id="d2", kind=DEVICE, action="created"),
+        ]
+    )
+
+    assert _identities(targets) == [(DEVICE, "label"), (DEVICE, "summary")]
+    for target in targets:
+        assert target.family == PYTHON_COMPUTED_ATTRIBUTE
```
polmichel, line 101: `RecomputeFamily` is a `Literal[str]` but since there are constant re-referencing its values, I think it looks like a `StrEnum`. Wdyt?

backend/tests/unit/core/merge/test_python_target_resolution.py
```diff
@@ -0,0 +1,325 @@
...
+
+    assert subscribers.calls == [("s1", "s2")]
+    by_identity = dict(zip(_identities(targets), targets, strict=True))
+    assert set(by_identity) == {(DEVICE, "summary"), (ROUTER, "tag")}
+    # The subscriber index reports a node once per matching group; the duplicate collapses.
+    assert _ids(by_identity[DEVICE, "summary"]) == frozenset({"d1", "d2"})
+    assert by_identity[DEVICE, "summary"].precise is True
+    # The site is read through a derived field, so the site change is a deliberate over-selection.
+    assert _ids(by_identity[ROUTER, "tag"]) == frozenset({"r1"})
+    assert by_identity[ROUTER, "tag"].precise is False
```
polmichel: Nitpick comment which can be ignored, encapsulate `by_identity[DEVICE, "summary"]` and the second one which looks alike in a local variable could help from a readability standpoint

backend/tests/unit/core/merge/test_python_target_resolution.py
```diff
@@ -0,0 +1,325 @@
...
+    assert _ids(by_identity[DEVICE, "summary"]) == frozenset({"d1"})
+    assert _ids(by_identity[DEVICE, "label"]) == frozenset({"d9"})
+
+
+async def test_an_updated_node_of_the_target_kind_is_its_own_target() -> None:
+    """The reverse lookup reaches a node only through the group its last compute subscribed it to.
+
+    A node that never computed is in no group, so relying on the lookup alone leaves it stale.
+    """
+    subscribers = RecordingSubscriberSource(subscribers={})
```
polmichel, line 269: The key is just there right? Maybe it would worth a comment to highlight this.

backend/tests/unit/core/merge/test_python_target_resolution.py
```diff
@@ -0,0 +1,325 @@
...
+async def test_deleted_nodes_select_their_readers() -> None:
+    """Every field the query read is gone with the node, so dropping the field filter is exact."""
+    subscribers = RecordingSubscriberSource(subscribers={"s1": [("d1", DEVICE)]})
+    resolver = _resolver(read_sets=[SUMMARY, LABEL], subscriber_source=subscribers)
+
+    targets = await resolver.resolve(changes=[MergeChange(node_id="s1", kind=SITE, action="deleted")])
+
+    assert _identities(targets) == [(DEVICE, "summary")]
+    assert _ids(targets[0]) == frozenset({"d1"})
+    assert targets[0].precise is True
```
polmichel, line 241: Maybe it'd be worth to also assert other booleans since they can all be modified and set?

polmichel (commented):
The test suite looks exhaustive and has interesting helpers and fakes. I just posted comments regarding code maintainability, nothing from correctness perspective. 
I am obviously open to discuss those as some of them may be subjective.
This PR certainly helped me regarding selective artifacts/generators regeneration, I've open two items under my Epic. 

Thanks for the Confluence document which was pretty useful to give an overview of the feature. This has helped the review

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: Both done. `MergeChange` has a signature property now, and one `group_ids_by_signature` helper does the grouping for the builder and the resolver. Good catch, it was copy paste.

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: I would rather fix this with your other suggestion. If `_Selection` becomes a union then each case has a name and the fields explain themselves. If we keep the booleans I will document them instead.

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: You are right, `widen=True` makes the other 3 meaningless. I prefer your union of small dataclasses over the class hierarchy, it keeps the accumulate logic in one place. The next PR in the stack changes `_Selection` again, so doing it here is a sure conflict. I will address this in the follow up.

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: Agree. `MutationAction` already exists, so it is mostly about typing `MergeChange.action` with it. It is not only my code though, `post_merge` and the rebase flow build those changes too, and the Jinja2 builder reads the same field. I prefer a follow up ticket so this PR stays about the deriver.

backend/tests/unit/core/merge/test_python_target_resolution.py
gmazoyer: Same idea as the one on the action field, and I agree. It would drop the four constants. It touches the 3 other families and their tests, so I would put it in the same follow up as `MergeChange.action`.

backend/tests/unit/core/merge/test_python_target_resolution.py
gmazoyer: Done. I did the same in the unscoped update test just below, it had the same repetition.

backend/tests/unit/core/merge/test_python_target_resolution.py
gmazoyer: Added `whole_kind is False`. The id assert already fails if a deletion widens, so this one is mostly saying it out loud but better to be explicit.

backend/tests/unit/core/merge/test_python_target_resolution.py
gmazoyer: Yes, the empty map is the whole point. I added one line saying the node is in no query group.

polmichel (approved):
LGTM! Nice one.
I'll may take another look at the open comments which will be tackled later, and at the resolver once it is plugged into the real engine in the following PRs since it'll be clearer to see what's the impact of the caching system and the lifecycle of the object. Even if the Confluence document already gives an overview.

Issue #9760, opened 2026-06-30
feature: consistent event branch scoping for branch-agnostic nodes


iddocohen:
## Summary

Define a consistent rule for the branch on which an event is emitted when a mutation targets a branch-agnostic node.

## Problem

Branch-agnostic objects, for example proposed changes and threads, currently emit events on whichever branch the mutation runs on. This produces inconsistent webhook branch scoping. A blanket rule of "a branch-agnostic node always emits on the default branch" is insufficient, because a branch-agnostic node can still carry branch-aware or local attributes and relationships, where the working branch remains meaningful.

## Proposed direction

Establish a rule along these lines:

- Emit on the default branch when all mutated members are branch-agnostic.
- Preserve the working branch when branch-aware or local attributes or relationships form part of the mutation.

Apply the rule centrally in the node-mutation event path rather than per mutation.

## Notes

Supersedes the narrow proposed change fix in #9761 once delivered. Requires design agreement before implementation.

Related: #9761 (proposed change consistency fix), #9759 (webhook form validation).

Issue #9759, opened 2026-06-30
feature: prevent webhook configurations that can never trigger


iddocohen:
## Summary

Prevent webhook configurations that cannot trigger any event.

## Problem

The webhook form exposes every combination of `event_type` and `branch_scope` permitted by the schema. Some combinations cannot match any event. For example, an Other Branches scope on proposed change lifecycle events, which are always emitted on the default branch. Neither the backend nor the frontend warns the user, so a webhook can be created that silently never fires.

## Proposed direction

- Surface guidance, or validation, that flags or blocks combinations that cannot trigger.
- This requires conditional validation that the schema-driven form does not support today. Treat as a request for change rather than a defect.

## Priority

Low. Documentation now covers the meaningful combinations; this is a usability improvement.

PR #10422, opened 2026-08-26
fix: scope branch-merge and proposed-change-review events to the default branch (closes #9761)

### LINKED ISSUE #9761: bug: branch merge and proposed change review events use an inconsistent branch for webhook scoping
## Summary

Emit proposed change lifecycle events on the default branch consistently, so webhook branch scoping behaves predictably for every client.

## Current behaviour

- `infrahub.proposed_change.merged` is explicitly stamped to the default branch. Intentional.
- `infrahub.proposed_change.approved` and `rejected` inherit the branch of the review mutation. They reach the default branch only because the review UI runs in the default-branch context, not by enforcement. A non-UI client (SDK, REST, automation) operating on another branch would emit these events on that branch.
- `infrahub.branch.merged` is emitted on the global branch (`-global-`), so a Default Branch scoped webhook never matches it and an Other Branches scoped webhook does. This is the inverse of the proposed change merge event.

## Required change

- Add an explicit default-branch override to the proposed change review mutation, covering approve, reject, and the cancel-approve and cancel-reject variants. Enforce server-side so behaviour is independent of the client.
- Reconcile the branch merge event scoping to the default branch.

## Constraint

For the branch merge event, only the scoping branch changes. The payload (`branch_name`, `branch_id`) must continue to identify the branch that was merged, so the event never implies the default branch was the subject of the merge.

## Impact

- Approve and reject: no observable change today, as both already reach the default branch. Removes reliance on UI context.
- `infrahub.branch.merged`: observable change to webhook matching. Requires a changelog entry and a release note.

## Out of scope

- General rule for branch-agnostic nodes: #9760.
- Webhook form validation: #9759.

polmichel:
## Why

Webhook branch scoping matches an event against the branch carried on its metadata. Two proposed-change / branch lifecycle events carried the wrong branch there, so branch-scoped webhooks matched them inconsistently:

- `infrahub.branch.merged` was emitted on the global branch, so a **Default Branch** scoped webhook never matched it while an **Other Branches** one did, exactly the inverse of `infrahub.proposed_change.merged`, which is already default-scoped.
- `infrahub.proposed_change.approved` / `rejected` / `approval_revoked` / `rejection_revoked` inherited the branch of the review mutation. They reached the default branch only because the review UI happens to run in the default-branch context — a non-UI client (SDK/REST/automation) reviewing from another branch would emit them on that branch.

**Goal:** emit both consistently on the default branch so webhook branch scoping is deterministic and client-independent.

**Non-goals:** the general rule for branch-agnostic node events (#9760) and webhook form validation (#9759). `branch.created` / `branch.deleted` carry the same pattern but are intentionally left for #9760.

### Future thoughts

Changes related to the global branch are matched by the "All other branches" filter and not by the "Default branch" filter. Events tied to the global branch can be misleading from a user's perspective, since this behavior relies on the fact that some objects are branch-agnostic and others are not. A dedicated "Global branch" filter could be a way to solve this.

Closes #9761

## What changed

**Behavioral changes:**
- `infrahub.branch.merged` is now scoped to the default branch.
- The proposed-change review events (approve / reject / cancel-approve / cancel-reject) are now scoped to the default branch server-side, independent of the branch the review mutation ran on.

**Implementation notes:**
- `PostMergeDispatcher` now stamps the merge event with `self.default_branch` instead of `self.global_branch`. The event payload (`branch_name`, `branch_id`) still identifies the branch that was merged.
- The review mutation resolves the default branch and passes it to `EventMeta.from_context(...)`, so only the event's scoping branch is overridden.

**What stayed the same:** no schema changes, no GraphQL/REST contract changes, event payloads unchanged.

## How to review

- `backend/infrahub/core/merge/post_merge.py` — merge event scoping (one-line change) + dropped `global_branch` param.
- `backend/infrahub/graphql/mutations/proposed_change.py` — review event scoping override.
- `backend/infrahub/core/merge/builder.py` — wiring updated for the dropped param.
- Tests are the largest part of the diff.

## How to test

```bash
uv run pytest \
  backend/tests/component/core/merge/test_post_merge.py::TestPostMergeBranchMergedEvent \
  "backend/tests/functional/proposed_change/test_review.py::TestProposedChangeReview::test_review_event_scoped_to_default_branch" \
  -v
```

## Impact & rollout

- **Backward compatibility:** `infrahub.branch.merged` webhook matching changes — a Default-Branch webhook starts matching it and an Other-Branches webhook stops. Documented in the changelog / release notes. Approve/reject scoping is unchanged for UI clients.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


polmichel (commented):
I'm not super confident about shipping this behavior change directly in the next minor release, since it changes the branch that some events are tied to. Maybe for this bug we should target `develop` branch and add a section in the 1.12 release note about this breaking change.
I'll target `develop` branch for now.

backend/tests/component/core/merge/test_post_merge.py
```diff
@@ -21,6 +21,30 @@
     from infrahub.database import InfrahubDatabase
 
 
+def _build_dispatcher(
```
polmichel, line 24: extracted as this is now reused in other test classes

dev/knowledge/backend/events.md
```diff
@@ -158,6 +158,15 @@ The `EventMeta` class provides rich context:
...
 Use `EventMeta.from_parent()` to create child events that maintain hierarchy.
 
+## Scoping branch for webhook matching
+
+Webhook branch scoping matches an event against `meta.context.branch` (see [Webhooks](webhooks.md)). Not every branch-agnostic event overrides the caller's context, so the scoping branch is set per event:
+
+- Proposed change merge and review events (merged, approved, rejected, and the approval/rejection revoke variants) are stamped to the default branch, so scoping is independent of the branch the mutation ran on.
+- `branch.merged` is stamped to the default branch as well, since the merge lands there. Its payload still carries the merged branch in `branch_name` / `branch_id`; only the scoping branch is the default one.
+- `branch.created` and `branch.deleted` are stamped to the global branch, pending a general rule for branch-agnostic node events.
+- `branch.rebased` and `branch.migrated` inherit the caller's context branch; they are not overridden.
```
polmichel, line 168: I wonder whether it would be relevant to assign the target branch for those operations

backend/tests/component/core/merge/test_post_merge.py
```diff
@@ -99,15 +99,58 @@ async def test_no_schema_updated_event_when_no_schema_change(
...
+    ) -> None:
+        source_branch = await create_branch(branch_name="feature", db=db)
+        memory_event = MemoryInfrahubEvent()
+        dispatcher = _build_dispatcher(db, source_branch, default_branch, memory_event)
+
+        await dispatcher.dispatch_events(
+            branch=source_branch,
+            proposed_change_id=None,
+            node_events=[],
+            context=_context(default_branch),
```
gmazoyer: Should this be `context=_context(source_branch)` to validate that the fix works properly?

gmazoyer (approved):

backend/tests/component/core/merge/test_post_merge.py
polmichel: Yes you're right, thanks @gmazoyer

ogenstad (approved):

PR #10423, opened 2026-08-26
perf(backend): read an object template a level at a time

### LINKED ISSUE #10419: tech-debt: creating an object from a nested template reads each subtemplate one parent at a time, so every level below the first costs 3 reads per branch
## Task Description

Creating an object from an object template materializes the template's components recursively: `handle_template_relationships()` walks the new node's `COMPONENT` relationships, and for each peer it reads the subtemplate together with the relationships materializing it consults, then recurses into that subtemplate's own components (`backend/infrahub/core/node/create.py`).

The read is issued **per parent**, not per level. Every subtemplate at depth two gets its own `prefetch_relationships()` call, so a template with N interfaces issues N prefetches for the interface level rather than one covering all N. The peers of a whole level are known before any of them is walked, so they could be read together the way the first level already is.

The cost, measured on the test schema's `device -> interface -> SFP`, per extra interface:

| query | flat (interface only) | nested (interface + SFP) |
|---|---|---|
| `node_create_all` | 1 | 2 |
| `uniqueness_constraint_validation` | 1 | 2 |
| `node_list_get_info` | 0 | 1 |
| `node_list_get_attribute` | 0 | 1 |
| `node_list_get_relationship` | 0 | 1 |
| `relationship_count_per_node` | 0 | 1 |
| **total** | **2** | **8** |

A first-level component costs 2 queries (its uniqueness check and its write). The SFP under it costs 6: the same 2, plus its mandatory-parent count constraint, plus 3 read queries — one batched relationship read and one batched peer read (`node_list_get_info` + `node_list_get_attribute`) — that exist only because that subtemplate was prefetched on its own.

The figure is constant per interface, so a *wide* tree does not degrade. A *deep* one pays those 3 reads per branch at every level below the first, and templates that nest several levels (a device holding interfaces holding transceivers) multiply it.

## Additional Information

Found while adding a depth-two read test in #10413, which removed the per-relationship peer reads underneath this recursion (a create from a template with ten interfaces went from 118 queries to 32, none of them a peer read). The recursion's per-parent shape predates that PR and was left alone there to keep the branch scoped.

- The measurement above is reproducible from `test_creating_from_a_nested_template_reads_no_peer_at_either_depth` in `backend/tests/component/templates/test_template_reads.py`, which asserts the cost stays constant per interface and bounds it at the current figure, so a change here will show up as that assertion moving.
- Recorded in `dev/knowledge/backend/templates.md`, under "What applying a template reads".

fatih-acar:
## Summary

Materializing the components of an object template walked it one parent at a time. Every subtemplate was read together with the relationships materializing it consults, so a template with ten interfaces issued ten reads for the interface level rather than one, and each level below the first paid those 3 reads again on every branch of the tree. A wide tree did not degrade, but a deep one multiplied.

The peers of a whole level are known before any of them is read, so the template is now read level by level: `read_template_components()` names the subtemplates every parent of a level holds, reads them in one call, and returns the tree it read. That read brings back the peers those subtemplates name, which is what leaves the level below already in memory — so the first level is no longer read back either, having arrived with the template.

Creating is a separate, depth-first walk over that tree (`create_components()`), which reads nothing. The two walks are deliberately different: a resource pool serving more than one level hands its resources out in the order the components are created, so creating a level at a time would move the resource each component is given. That regression was caught in review by @gmazoyer, who wrote the test pinning it.

Closes #10419.

## Measured

On the test schema's `device -> interface -> SFP`, per extra interface at depth two:

| query | before | after |
|---|---|---|
| `node_create_all` | 2 | 2 |
| `uniqueness_constraint_validation` | 2 | 2 |
| `relationship_count_per_node` | 1 | 1 |
| `node_list_get_info` | 1 | 0 |
| `node_list_get_attribute` | 1 | 0 |
| `node_list_get_relationship` | 1 | 0 |
| **total** | **8** | **5** |

Every read is now constant per level. What is left per interface is the two objects' uniqueness checks and writes, plus the count constraint on the SFP's mandatory parent.

Total queries for a create through `create_node()`:

| Template | Before | After |
|---|---|---|
| 1 interface, each holding an SFP | 20 | 18 |
| 3 interfaces, each holding an SFP | 36 | 28 |
| 5 interfaces, each holding an SFP | 52 | 38 |
| 1 interface, flat | 14 | 12 |
| 5 interfaces, flat | 22 | 20 |
| 10 interfaces, flat | 32 | 30 |

The flat rows improve by the same 2 queries at any width: the subtemplates of the first level came back with the template read and are no longer read a second time. That was the `RelationshipManager.get_peers()` re-read #10413 left in place — this resolves it for the template path, where both reads ask for the same metadata and the walk owns both ends. The general case in `get_peers()` is untouched.

## Test Plan

- `test_creating_from_a_nested_template_reads_no_peer_at_either_depth` — its bound moves from 8 queries per interface to 5, which is the assertion #10419 predicted would move.
- `test_a_level_of_a_template_is_read_once_however_many_parents_it_hangs_from` — new. Asserts 3 relationship reads and 4 node reads whatever the width of the second level; before this branch they were 3/5/7 and 5/7/9 for 1/3/5 interfaces.
- `test_one_pool_shared_by_two_component_levels_allocates_depth_first` — added by @gmazoyer in review as a strict xfail; kept as written, with the marker removed. A component and the component it holds take adjacent addresses from a pool serving both levels.
- `uv run pytest backend/tests/component backend/tests/unit/graphql backend/tests/unit/core` — 4 549 passed, 29 skipped. `backend/tests/component/core/test_branch_rebase.py` is deselected: its two failures are Redis connection refusals from the local environment and reproduce identically with this branch's code reverted.
- `uv run invoke lint` (ruff, ty, mypy over 1 631 files) and `uv run invoke docs.validate` are clean.

## Documentation Updates

- `dev/knowledge/backend/templates.md` — the two walks and why they differ, and the cost of a create restated in terms of level reads: 3 queries per level, one for the template and one for each level of subtemplates under it. A level's read also brings back the parents it points back at, which the walk already holds — recorded rather than fixed.
- `dev/knowledge/backend/mutations.md` — the mechanism the walk leans on: a prefetched relationship carries its peers as nodes, so a caller that prefetched does not read them back.
No changelog fragment: `+faster-creates-from-object-templates.changed.md`, added by the branch below, already covers this for a reader of the release notes.

## Related Context

Stacked on #10413, which removed the per-relationship peer reads underneath this recursion and left the recursion's per-parent shape alone to keep that branch scoped. Measurements are taken on top of that branch.

### DIFF
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -217,9 +217,7 @@ async def extract_peer_data(
             # deeper templates are handled in the next level of recursion
             if await _peer_is_a_template(db=db, relationship=relationship):
                 continue
-            # The subtemplate was read with the peers its relationships name, so hand over the node
-            # where it is in hand: an id sends the checks on the new object, and the write itself,
-            # back to the database for a node already in memory.
+            # The peer is already read, so an id would send the checks and the write back to the database.
             rel_peers.append({"id": relationship.get_peer_in_hand() or relationship.peer_id})
 
         # Only set the relationship data if there are actual peers to set
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -8,7 +8,6 @@ from infrahub.core.constants import (
     PROFILES_RELATIONSHIP_NAME,
     SYSTEM_USER_ID,
     MetadataOptions,
-    RelationshipCardinality,
     RelationshipKind,
 )
 from infrahub.core.constants.schema import RESOURCE_POOL_REL_SUFFIX
@@ -29,32 +28,44 @@ if TYPE_CHECKING:
     from infrahub.core.node import SchemaProtocol
     from infrahub.core.protocols import CoreObjectTemplate
     from infrahub.core.relationship.model import Relationship, RelationshipManager
-    from infrahub.core.schema import MainSchemaTypes, NonGenericSchemaTypes, RelationshipSchema
+    from infrahub.core.schema import MainSchemaTypes, NonGenericSchemaTypes
     from infrahub.core.timestamp import Timestamp
     from infrahub.database import InfrahubDatabase
 
 
-async def get_template_relationship_peers(
-    db: InfrahubDatabase, template: CoreObjectTemplate, relationship: RelationshipSchema
-) -> Mapping[str, CoreObjectTemplate]:
-    """For a given relationship on the template, fetch the related peers."""
-    template_relationship_manager: RelationshipManager = getattr(template, relationship.name)
-    peers: Mapping[str, CoreObjectTemplate]
-    if relationship.cardinality == RelationshipCardinality.MANY:
-        peers = await template_relationship_manager.get_peers(db=db, include_metadata=MetadataOptions.SOURCE)
-    else:
-        template_relationship_peer = await template_relationship_manager.get_peer(db=db)
-        peers = {template_relationship_peer.id: template_relationship_peer} if template_relationship_peer else {}
+async def get_component_subtemplates(
+    db: InfrahubDatabase, obj: Node, template: CoreObjectTemplate, fields: list
+) -> list[CoreObjectTemplate]:
+    """The subtemplates the component relationships of this template name.
 
-    # Every peer is a subtemplate of its own, read here with the relationships materializing it reads.
+    The template was read with those relationships and with the peers they name, so naming its
+    subtemplates costs no query.
+    """
+    subtemplates: list[CoreObjectTemplate] = []
+    for relationship in obj.get_relationships(kind=RelationshipKind.COMPONENT, exclude=fields):
+        template_relationship_manager: RelationshipManager = getattr(template, relationship.name)
+        for template_relationship in await template_relationship_manager.get_relationships(db=db):
+            subtemplates.append(cast("CoreObjectTemplate", await template_relationship.get_peer(db=db)))
+    return subtemplates
+
+
+async def read_subtemplates(db: InfrahubDatabase, branch: Branch, subtemplates: list[CoreObjectTemplate]) -> None:
+    """Read a whole level of subtemplates, with the relationships materializing them consults.
+
+    The peers those relationships name are read with them, which is what leaves the level below
+    already in memory by the time it is walked.
+    """
     await registry.manager.prefetch_relationships(
         db=db,
-        nodes=cast("Iterable[Node]", peers.values()),
-        names={name for peer in peers.values() for name in get_relationship_names_to_read(schema=peer.get_schema())},
-        branch=template_relationship_manager.branch,
+        nodes=cast("Iterable[Node]", subtemplates),
+        names={
+            name
+            for subtemplate in subtemplates
+            for name in get_relationship_names_to_read(schema=subtemplate.get_schema())
+        },
+        branch=branch,
         include_metadata=MetadataOptions.SOURCE,
     )
-    return peers
 
 
 async def _peer_is_a_template(db: InfrahubDatabase, relationship: Relationship) -> bool:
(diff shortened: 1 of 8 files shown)


backend/infrahub/core/node/create.py
```diff
@@ -198,68 +209,97 @@ async def allocate_from_resource_pools(
...
+                    branch=branch,
+                    constraint_runner=constraint_runner,
+                    parent_obj=parent_obj,
+                    parent_template=parent_template,
+                    subtemplate=subtemplate,
+                    at=at,
+                    user_id=user_id,
+                )
+                next_level.append((obj_peer, subtemplate))
+        level = next_level
```
gmazoyer: This one is introducing a behaviour change regarding how pool allocation will work.

I created a quick tesst with a schema that gave both `PhysicalInterface` and `Sfp` a `primary_ip` relationship (stupid to assign an IP to a SFP but it just to make a point), pointed both subtemplates at one shared IP pool over 10.10.9.0/27, and built a template with 2 interfaces each holding an SFP. Then I ran the same test twice, swapping only `create.py`:

Stable behaviour is: eth0 iface → .1, eth0 sfp → .2, eth1 iface → .3, eth1 sfp → .4
Behaviour with this PR is: eth0 iface → .1, eth0 sfp → .3, eth1 iface → .2, eth1 sfp → .4

That said, I won't say it is a bug, I'm not even sure if people share pools across level like that. But I suppose we should mention it somewhere just in case.

backend/tests/component/templates/test_template_reads.py
```diff
@@ -219,9 +219,47 @@ async def test_creating_from_a_nested_template_reads_no_peer_at_either_depth(
...
+        relationship_reads[nbr_interfaces] = counting_db.count_for(NodeListGetRelationshipsQuery.name)
+        node_reads[nbr_interfaces] = counting_db.count_for(NodeListGetInfoQuery.name)
+
+    # One relationship read per level: the template, the interface subtemplates, the SFP subtemplates.
+    assert set(relationship_reads.values()) == {3}, (
+        f"a level was read once per parent rather than once: {relationship_reads}"
+    )
+    # Each of those reads brings back the peers it names, which is how a level is in memory before
+    # it is walked. Nothing reads a subtemplate again afterwards, so the nodes read are those three
+    # sets of peers plus the template the create was given.
```
gmazoyer: I think this comment might not be entirely true, for such schema I'd say the read would look like:

```
read#1: [DEVICE_TPL]
read#2: [IFACE_eth0, IFACE_eth1, IFACE_eth2]
read#3: [DEVICE_TPL, SFP_0, SFP_1, SFP_2] <- device template again
read#4: [IFACE_eth0, IFACE_eth1, IFACE_eth2] <- whole interface level again
```

But could be I don't get the comment quite right.

dev/knowledge/backend/mutations.md
gmazoyer: Stale line here now

backend/infrahub/core/node/create.py
```diff
@@ -198,68 +209,97 @@ async def allocate_from_resource_pools(
...
+    The subtemplates of a level are all named before any of them is walked, so the level is read
+    once for every parent it hangs from rather than once per parent. `template` itself was read that
+    way by the caller, which is why the walk starts with a level already in memory.
+    """
+    component_registry = get_component_registry()
+    constraint_runner = await component_registry.get_component(NodeConstraintRunner, db=db, branch=branch)
+
+    level: list[tuple[Node, CoreObjectTemplate]] = [(obj, template)]
+    while level:
+        parents_with_subtemplates: list[tuple[Node, CoreObjectTemplate, list[CoreObjectTemplate]]] = []
```
gmazoyer: `list[tuple[Node, CoreObjectTemplate, list[CoreObjectTemplate]]]` is built, flattened, then re-walked. A small @`dataclass ParentLevel` makes the loop readable and is safe even in a fix.

backend/infrahub/core/node/create.py
fatih-acar: Good catch, and it is a bug rather than a wrinkle — thank you for the test, I took it as is.

Fixed in d7f0edf30 by separating the two walks. Reading and creating do not have to happen in the same order: `read_template_components()` walks the template level by level and returns the tree it read, then `create_components()` walks that tree depth first and creates without reading anything. So the reads stay one per level and the creation order is the one on the branch below.

Your allocation ordering is back: eth0 iface → .1, eth0 sfp → .2, eth1 iface → .3, eth1 sfp → .4. The measured query counts did not move (18/28/38 for 1/3/5 nested interfaces, 30 for a flat ten). I removed the strict xfail marker and your test passes as written, and I recorded in `dev/knowledge/backend/templates.md` that the creation order is load-bearing so the next person does not merge the two walks back together.

backend/tests/component/templates/test_template_reads.py
fatih-acar: You are right, and the comment was wrong. I logged the reads to check:

```
1. node_list_get_info:         [DEVICE_TPL]
2. node_list_get_relationship: [DEVICE_TPL]
3. node_list_get_info:         [IFACE_eth0, IFACE_eth1, IFACE_eth2]
4. node_list_get_relationship: [IFACE_eth0, IFACE_eth1, IFACE_eth2]
5. node_list_get_info:         [DEVICE_TPL, SFP_0, SFP_1, SFP_2]
6. node_list_get_relationship: [SFP_0, SFP_1, SFP_2]
7. node_list_get_info:         [IFACE_eth0, IFACE_eth1, IFACE_eth2]
```

Exactly your sketch: a level names the level below it *and* the parents it points back at, so 5 re-reads the device template and 7 re-reads the whole interface level. The batched read has no way to be told a node is already in hand.

Corrected in 73829b134, along with the assertion message — the figure the test is really guarding is that neither read count grows with the width of a level. The same clarification went into the knowledge doc.

dev/knowledge/backend/mutations.md
fatih-acar: Fixed in 73829b134 — it now reads `# template instantiation, read a level at a time`.

backend/infrahub/core/node/create.py
fatih-acar: Done in d7f0edf30. Two small dataclasses now: `TemplateComponent` (a subtemplate, the schema of the object it describes, and the components it holds) and `_SubtemplateParent` for the frontier of the level walk. The flatten-then-re-walk is gone with it — the read walk builds the tree and the create walk consumes it, so nothing is flattened and re-paired.

gmazoyer (approved):

PR #10424, opened 2026-08-26
fix(preferences): expose the inherited layer to fix the date preview

### LINKED ISSUE #10200: bug: The date-format Example preview cannot show the inherited timezone while a personal override is being cleared
### Component

Frontend UI

### Current Behavior

In Account settings → Preferences, the date-format `Example:` preview renders in the timezone the form currently holds, falling back to the **inherited** zone while the timezone field is empty (`inheritedTimezone`, `entities/preferences/domain/rules/resolve-date-preferences.ts`).

When the caller has a personal timezone override and clears it, the inherited zone is unknowable client-side, so the preview falls back to the browser zone. If a **global** timezone is set, that is not what will apply after saving — the preview is wrong for the seconds between clearing the override and saving.

Two API facts cause this:

- `resolve_effective_preferences` (`backend/infrahub/graphql/queries/preferences.py`) returns only the winning value per field; once a `USER` override wins, the global value it resolved against is not in the response.
- `resolve_global_preferences` is gated on `MANAGE_GLOBAL_PREFERENCES_PERMISSION`, so a non-admin cannot read it directly.

### Expected Behavior

While editing, the preview reflects the zone that will actually apply after saving, including the case where a personal override is being cleared and a global default exists.

### Suggested approach

Expose the inherited resolution (global → default) to any authenticated caller, e.g. an additional field on `InfrahubEffectivePreferences` alongside the resolved value. This leaks nothing new: a caller with no override already receives the global value through the same query. The frontend would then use it as the fallback instead of `null`.

This is a GraphQL schema change, so it wants a deliberate decision rather than being folded into a bugfix.

### Additional Information

Follow-up from the cubic review on https://github.com/opsmill/infrahub/pull/10193 (the `Example:` timezone fix for #10175). That PR narrowed the defect — the preview is now correct whenever no global timezone is set, and no longer shows a zone the user just removed — but cannot close it fully without the API change above.

Impact is limited: transient, edit-time only, and it self-corrects on save.

pa-lem:
Closes #10200. Follow-up to #10193, which narrowed the defect but could not close it without this API change.

## The bug

On Account → Preferences, the date-format `Example:` preview renders against the timezone the form currently holds, falling back to the inherited zone while the timezone field is empty. That fallback could only be computed when the effective source was `GLOBAL`: once a caller had a personal (`USER`) timezone override, the API had already resolved the inherited layer away, so `inheritedTimezone` returned `null` and the preview silently fell back to the browser zone — wrong whenever a global timezone is set.

`EffectivePreferences._resolve_field` returned only the winning value and its source; the value it shadowed was discarded before reaching GraphQL, and nothing else in the response let a client reconstruct it. The frontend's unit test pinned `USER → null` with the comment "unknowable here", so the bug was encoded as intended behaviour in the suite.

## The API change

The schema shape was an explicit decision, per the `state/need-decision` label:

```graphql
type EffectiveDateFormat { inherited: DateFormat  source: PreferenceSource!  value: DateFormat }
type EffectiveTimezone   { inherited: String      source: PreferenceSource!  value: String }
```

`inherited` is the value the caller would fall back to with no override of their own, null when there is nothing to inherit and the client applies its own default.

Additive only — no new root query or mutation, and `graphql/schema.py` / `QUERIES_REQUIRING_AUTHENTICATION` are untouched.

A nullable scalar rather than an object carrying its own `source`: `_resolve_field(user_value=None, …)` returns `GLOBAL` iff the global value is set, and `DEFAULT` with a null value otherwise, so a source field and nullability would be bijective — one scalar says everything two types would. The resolution still runs through `_resolve_field` in the model layer, so a third layer stays addable without a breaking change; only the GraphQL edge projects down to the value. Applied to both `date_format` and `timezone`.

No extra DB round-trip: `resolve_effective_preferences` already read the global row in the same request.

## The permission boundary

`inherited` is returned for every field, and `InfrahubGlobalPreferences` returns only `date_format` and `timezone`. On the read path, the ungated `InfrahubEffectivePreferences` is now a complete substitute for the gated query.

That is deliberate. `resolve_effective_preferences` has never had a permission gate and already returned the global value verbatim, labelled `source: GLOBAL`, for every field a caller had not overridden. The delta is not new data — it removes a caller's ability to suppress that disclosure by setting an override, and "user with an override" is not a lower-privileged principal than "user without one". `GLOBAL_OWNER_ID` is a single deployment-wide sentinel, so there is no tenant dimension. The values are an IANA zone name and one of five enum keys.

`MANAGE_GLOBAL_PREFERENCES_PERMISSION` is unchanged. What it guards is the management surface — reading the org row as an editable scope and writing it via `InfrahubSetPreferences(scope: GLOBAL)` — not the confidentiality of two display settings.

The resolver docstring previously claimed the global row was "read internally, never exposed as raw org values". That was already false before this PR; it now states the boundary accurately. `test_effective_inherited_readable_without_manage_global_permission` asserts both halves: the gated raw read is denied for a caller with no role/group/permission — with the exact denial message and a null `data` — and the effective read still returns the inherited value.

## Tooltip: provenance-only

The (i) source tooltip was half-live, half-saved. Its source label came from the saved preference while its rendered date sample was built from the live watched timezone field. Changing the date-format dropdown left the tooltip stale next to an updated `Example:` line; changing the timezone moved the sample while the label stayed pinned — with `source: GLOBAL` it rendered the org's format in the user's unsaved zone, a value corresponding to neither the org default nor what would be saved.

The tooltip now describes saved state only and names the format by label, with no rendered sample:

| Case | Date format | Timezone |
|---|---|---|
| `USER`, inherited `GLOBAL` | `Your preference, overriding the organisation default: dd/MM/yyyy HH:mm.` | `Your preference, overriding the organisation default: Asia/Tokyo.` |
| `USER`, nothing inherited | `Your preference.` | `Your preference.` |
| `GLOBAL` | `From the organisation default: dd/MM/yyyy HH:mm.` | `From the organisation default: Asia/Tokyo.` |
| `DEFAULT` | `From your browser.` | `From your browser: <browserZone>.` |

Division of labour: **tooltip** = where the saved value comes from; **`Example:` line** = what saving would produce (follows the live form, including the unsaved zone); **`Automatic (inherited)` placeholder** = this field has no override. `sourceMessage` takes a required `formatValue` and an optional `browserValue` whose clause is dropped when absent — the only date-vs-timezone divergence. The `supportedTimezone` pre-check still runs ahead of the shared message, now against the pending zone rather than the stored one.

## Tests

Backend:

- `inherited` is now selected in `EFFECTIVE_QUERY`, so every pre-existing whole-dict assertion pins the invariant `source != USER ⇒ inherited == value`.
- A parametrized invariant test: `inherited` agrees with the effective value wherever the user has no override.
- New cases for USER-shadowing-GLOBAL, nothing-to-inherit-when-no-global, per-attribute independence, and the permission test above (asserting the exact denial message and that `data` is null).

Frontend — reintroducing the bug's shape (deriving the fallback from `source` instead of reading `inherited`) fails the preview test; breaking the tooltip's inherited read fails five more:

- `user-preferences-card.test.tsx` — at card level through the real wiring: clearing a saved date-format override returns the `Example:` line to the inherited organisation format.
- Five tooltip tests across `preference-fields.test.tsx` and `user-preferences-card.test.tsx` name the organisation default that a user override is shadowing, for both the date format and the timezone.

## Verification

| Check | Result |
|---|---|
| `invoke schema.validate-graphqlschema` | clean |
| frontend codegen drift (both pipelines) | none |
| backend unit — preferences | 14 passed |
| backend component — preferences (Docker) | 12 passed |
| frontend — full suite | 1237 passed (180 files) |
| biome / knip / betterer | clean / clean / unchanged (188) |

## Deliberately not done

- **No card-level test for the timezone field's cleared-override path.** Opening a virtualized combobox under vitest browser mode needs a `process.env.VIRT_ON` define in `vitest.config.ts`; carrying a suite-wide config change was not worth it here. The date-format field covers the same code path at card level, and a tooltip test covers the timezone field's inherited read. `vitest.config.ts` is untouched.
- **No e2e test.** `tests/e2e/preferences/` is untouched; the three-layer (user + global + default) interaction remains unexercised end-to-end.
- **No change to `InfrahubGlobalPreferences` or its permission gate.**
- **No Python SDK addition** — additive field on an existing type, not new query surface.
- **No validity check on the inherited zone in the tooltip.** The clause says "the organisation default is X", not "times are shown in X", so naming an unrenderable zone remains a true statement about the org default.
- **No third preference layer**, no persistence or migration change, no change to save behaviour.


frontend/app/src/entities/preferences/ui/user-preferences-card.test.tsx
```diff
@@ -160,11 +170,57 @@ describe("UserPreferencesCard", () => {
...
+
+    // The example must now follow the INHERITED organisation zone (UTC+9 → the next calendar day),
+    // which the caller's own override was hiding.
+    await expect.element(component.getByText("Example: 2026-06-12 08:30")).toBeVisible();
+
+    // …and not the browser's zone, which is what the bug fell back to. On a machine already in
+    // Asia/Tokyo this guard contradicts the assertion above and fails loudly rather than passing
+    // for the wrong reason.
+    expect(
+      component.getByText(`Example: ${format(FIXED_INSTANT, "yyyy-MM-dd HH:mm")}`).elements()
```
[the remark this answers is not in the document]
pa-lem: Valid, and fixed in 03a9094 — the guard is removed rather than made conditional.

You are right that `format(FIXED_INSTANT, ...)` renders in the host zone, which the fake timers do not freeze, so on a UTC+9 host it produces exactly `2026-06-12 08:30` and contradicts the positive assertion two lines above. The comment defending it ("fails loudly rather than passing for the wrong reason") was wrong reasoning: failing when the implementation is correct is a false positive, not a safety property.

Removed instead of conditioned, because the guard had no discriminating power either way. Where the host and inherited renderings differ, the positive `Example: 2026-06-12 08:30` assertion already catches the regression; where they coincide, no negative assertion can tell the two apart. A conditional guard would only ever fire in the cases already covered.

Verified: with `inheritedValue` reverted to its pre-fix behaviour, 5 tests in this entity fail, including this one on the positive assertion. The residual weakness is that on a UTC+9 host this test passes without discriminating; pinning `timezoneId` in the browser context would fix that suite-wide and is worth its own change.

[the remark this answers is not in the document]
pa-lem: Valid, and fixed in 03a9094 — the guard is removed rather than made conditional.

You are right that `format(FIXED_INSTANT, ...)` renders in the host zone, which the fake timers do not freeze, so on a UTC+9 host it produces exactly `2026-06-12 08:30` and contradicts the positive assertion two lines above. The comment defending it ("fails loudly rather than passing for the wrong reason") was wrong reasoning: failing when the implementation is correct is a false positive, not a safety property.

Removed instead of conditioned, because the guard had no discriminating power either way. Where the host and inherited renderings differ, the positive `Example: 2026-06-12 08:30` assertion already catches the regression; where they coincide, no negative assertion can tell the two apart. A conditional guard would only ever fire in the cases already covered.

Verified: with `inheritedValue` reverted to its pre-fix behaviour, 5 tests in this entity fail, including this one on the positive assertion. The residual weakness is that on a UTC+9 host this test passes without discriminating; pinning `timezoneId` in the browser context would fix that suite-wide and is worth its own change.

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -280,3 +360,52 @@ async def test_global_denied_for_normal_account(
...
+
+async def test_effective_inherited_readable_without_manage_global_permission(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    default_permission_backend: None,
+    register_core_models_schema: None,
+    first_account: Node,
+    session_first_account: AccountSession,
+) -> None:
+    """The asymmetry between the two queries is deliberate, not an oversight.
```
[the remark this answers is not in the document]
pa-lem: Valid — fixed in 03a9094. Condensed to the observable contract, with the cross-code names, the rejected-approach narration, and the multi-paragraph rationale removed; the design discussion stays in the PR description where the rule puts it.

The one reference kept is react-stately in `vitest.config.ts`, under the rule's explicit exception for a workaround depending on an upstream library symbol — now stated with a version constraint.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -19,11 +19,29 @@
...
+
 async def resolve_effective_preferences(root: dict, info: GraphQLResolveInfo) -> dict:  # noqa: ARG001
     """Resolve the caller's effective preferences (user override → global default → DEFAULT).
 
-    Open to any authenticated caller; the global row is read internally, never exposed as raw org
-    values.
+    Open to any authenticated caller. The global layer's values ARE reported here — labelled GLOBAL
+    when nothing shadows them, and as the `inherited` layer when a user override does. Since
+    `inherited` is returned for every field, this query now discloses the same two org values that
+    the gated `InfrahubGlobalPreferences` returns. That is deliberate:
```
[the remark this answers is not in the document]
pa-lem: Valid — fixed in 03a9094. Condensed to the observable contract, with the cross-code names, the rejected-approach narration, and the multi-paragraph rationale removed; the design discussion stays in the PR description where the rule puts it.

The one reference kept is react-stately in `vitest.config.ts`, under the rule's explicit exception for a workaround depending on an upstream library symbol — now stated with a version constraint.

backend/infrahub/graphql/queries/preferences.py
```diff
@@ -19,11 +19,29 @@
     from infrahub.graphql.initialization import GraphqlContext
 
 
+def _effective_field(*, resolved: ResolvedPreference[Any], inherited: ResolvedPreference[Any]) -> dict[str, Any]:
+    """Plain dicts, not ResolvedPreference: the frozen dataclass has no `inherited` attribute for graphene to find.
```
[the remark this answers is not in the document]
pa-lem: Valid — fixed in 03a9094. Condensed to the observable contract, with the cross-code names, the rejected-approach narration, and the multi-paragraph rationale removed; the design discussion stays in the PR description where the rule puts it.

The one reference kept is react-stately in `vitest.config.ts`, under the rule's explicit exception for a workaround depending on an upstream library symbol — now stated with a version constraint.

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -116,9 +136,19 @@ async def test_effective_user_override_source_user(
     assert result.errors is None
     assert result.data is not None
     prefs = result.data["InfrahubEffectivePreferences"]
-    # User override wins for both attributes: source USER, value is the user's.
-    assert prefs["date_format"] == {"value": "EU_DATETIME", "source": "USER"}
-    assert prefs["timezone"] == {"value": "Europe/Paris", "source": "USER"}
+    # User override wins for both attributes: source USER, value is the user's. The shadowed global
+    # layer stays readable as `inherited` — this is the case #10200 needed, so a client can preview
```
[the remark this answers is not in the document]
pa-lem: Valid — fixed in 03a9094. Sweeping the rest of the diff for the same rule turned up a third issue reference the review did not flag, in `resolve-date-preferences.test.ts`; all three are gone and the behavioural point is kept in each case.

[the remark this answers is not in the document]
pa-lem: Valid — fixed in 03a9094. Sweeping the rest of the diff for the same rule turned up a third issue reference the review did not flag, in `resolve-date-preferences.test.ts`; all three are gone and the behavioural point is kept in each case.

frontend/app/tests/components/utils.ts
```diff
@@ -3,8 +3,18 @@
...
-}) {
+export async function initPointerTracking(locator: PointerLocator) {
   await locator.click({ position: { x: 0, y: 0 } });
 }
+
+/**
+ * Moves the pointer off whatever it is hovering, so an open React Aria tooltip closes before the
+ * next interaction or render.
+ */
+export async function closeTooltip(locator: PointerLocator) {
```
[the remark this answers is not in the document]
pa-lem: Valid — fixed in c54c6aa. `closeTooltip` is now an alias of `initPointerTracking` rather than a second body, so the two cannot drift.

Worth noting how it got there: the helper existed only to kill eight copies of a comment that restated the click, flagged under the code-doc-style rule. That fix traded comment duplication for implementation duplication, which is the worse of the two. The alias keeps the call-site intent (a reader after an assertion wants "close the tooltip", not "warm up pointer tracking") with one implementation behind it, and the shared docstring now states both intents.

Verified: 1246 tests pass, biome and knip clean, betterer unchanged.

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -16,11 +16,12 @@
     from infrahub.core.node import Node
     from infrahub.database import InfrahubDatabase
 
+# Selecting `inherited` on every effective read pins the invariant on all of the paths below.
```
polmichel: This comment looks kind of misplaced here, I think we can get rid of it (or place it somewhere else)

backend/infrahub/graphql/types/preferences.py
```diff
@@ -46,6 +46,20 @@ class PreferenceWriteScope(StrEnum):
...
+
+    value = Field(DateFormat, required=False)
+    source = Field(PreferenceSource, required=True)
+
+
+class InheritedTimezone(ObjectType):
+    """The timezone the caller would inherit with no override of their own; source is GLOBAL or DEFAULT."""
+
+    value = Field(String, required=False)
+    source = Field(PreferenceSource, required=True)
```
polmichel: Is source useful in `InheritedTimezone` and `InheritedDateFormat`? If I'm not mistaken, DEFAULT <=> value == null and GLOBAL <=> value != null.

Also, having a docstring that tries to constrain the set of allowed values, without any data validation, could be hard to maintain.

The visual symmetry is not really a plus here, since the underlying concepts are not the same.

polmichel: This is fine to reuse the `ResolvedPreferences` resolution logic in the lower layer of the backend, but there should be a distinction here compared to other classic preferences

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -138,10 +164,55 @@ async def test_effective_mixed_per_attribute_sources(
...
+    }
+
+
+async def test_effective_mixed_inherited_layers_per_attribute(
+    db: InfrahubDatabase,
+    default_branch: Branch,
+    register_core_models_schema: None,
+    first_account: Node,
+    session_first_account: AccountSession,
+) -> None:
```
polmichel, line 186: Nice test 👍

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -280,3 +357,37 @@ async def test_global_denied_for_normal_account(
...
+    not the readability of the values themselves.
+    """
+    await PreferenceRepository(db=db).save(Preference(owner_id=GLOBAL_OWNER_ID, timezone="UTC"))
+    await PreferenceRepository(db=db).save(Preference(owner_id=first_account.id, timezone="Europe/Paris"))
+
+    global_result = await run_query(
+        db=db, branch=default_branch, query=GLOBAL_QUERY, account_session=session_first_account
+    )
+    assert global_result.errors is not None
+    assert global_result.data is None or global_result.data.get("InfrahubGlobalPreferences") is None
```
polmichel: We should be able to tell what a test's execution flow is at a glance. Having an `or` nested in an assertion forces us to run the test to know what the concrete system behavior is.

backend/tests/component/graphql/queries/test_preferences.py
```diff
@@ -280,3 +357,37 @@ async def test_global_denied_for_normal_account(
...
+    The asymmetry is deliberate: the permission gates the ability to manage organisation defaults,
+    not the readability of the values themselves.
+    """
+    await PreferenceRepository(db=db).save(Preference(owner_id=GLOBAL_OWNER_ID, timezone="UTC"))
+    await PreferenceRepository(db=db).save(Preference(owner_id=first_account.id, timezone="Europe/Paris"))
+
+    global_result = await run_query(
+        db=db, branch=default_branch, query=GLOBAL_QUERY, account_session=session_first_account
+    )
+    assert global_result.errors is not None
```
polmichel, line 380: We should always assert exact error messages, as this gives us a clear description of how the system behaves.

polmichel: Claude missed it in `.claude/rules/testing-python.md`. 
Maybe we should add other examples that include this data structure, `global_result.errors`, so that it can be detected as an "error message".

```
## Assert exact expectations

Exact-match is not only for error messages. Assert the exact collection (full set/dict equality, not `in`/`issubset`), never mere non-emptiness (`!= frozenset()`, `len() > 0`), and a positive count where the number matters (so a run that silently measures zero fails). A denial test must also reload the target and assert nothing changed. Pin literal expected values — never compute the expectation with the same serializer/library the implementation calls. Full guidance in `dev/guidelines/backend/testing.md` §"Assert exact expectations".
```

backend/tests/unit/core/test_preference_resolution.py
```diff
@@ -15,6 +15,9 @@ class ResolutionCase:
     global_: Preference | None
     expected_date_format: ResolvedPreference
     expected_timezone: ResolvedPreference
+    # No defaults: a default would let a row's inherited expectation go silently unasserted.
```
polmichel: I think we can remove this comment as well

backend/infrahub/graphql/types/preferences.py
polmichel: A solution is proposed in the comment on the related tests

backend/tests/unit/core/test_preference_resolution.py
```diff
@@ -62,3 +96,31 @@ def test_effective_preference_resolution(case: ResolutionCase) -> None:
...
+    assert effective.inherited_timezone() == case.expected_inherited_timezone
+
+
+@pytest.mark.parametrize("case", RESOLUTION_CASES, ids=lambda case: case.name)
+def test_inherited_never_reports_the_user_layer(case: ResolutionCase) -> None:
+    """The inherited projection suppresses the user layer, so it can only ever be GLOBAL or DEFAULT."""
+    effective = EffectivePreferences(user=case.user, global_=case.global_)
+
+    assert effective.inherited_date_format().source != PreferenceSource.USER
+    assert effective.inherited_timezone().source != PreferenceSource.USER
```
polmichel: In my opinion, this test is the signal the code is not built as it should. This comment is linked to the first one (https://github.com/opsmill/infrahub/pull/10424/changes#r3902536886). 
I think the correct approach is then to change the structure 
```
class EffectiveDateFormat(ObjectType):
      value = Field(DateFormat, required=False)
      source = Field(PreferenceSource, required=True)
      inherited = Field(DateFormat, required=False)   # and not an InheritedDateFormat
```

And then the front-end should determine based on inherited nullability if the source is default or global.

changelog/10200.fixed.md
```diff
@@ -0,0 +1 @@
+In Account settings → Preferences, the date format and timezone fields now preview what saving would actually produce. Clearing a personal override shows the organisation default that takes over — previously the date-format `Example:` fell back to your browser's timezone, and disappeared altogether when the date format itself was cleared. The information icon now describes the value you are about to save rather than the one already stored, so it no longer names a personal preference you have just cleared. To make this possible, `InfrahubEffectivePreferences` now returns an `inherited` layer alongside each field's `value` and `source`, reporting what the field would resolve to if the caller set nothing of their own.
```
polmichel: Since this will be posted in the Fixed section of the release notes, maybe only what has been fixed is necessary, and not the implementation details or what the broken behavior was? Wdyt?

polmichel (commented):
Backend code is now reviewed. Some comments regarding the test suite and the code maintainibility.

polmichel (approved):
Backend code looks good to me!

PR #10426, opened 2026-08-26
perf(backend): hand a template component the peer it carries

fatih-acar:
Stacked on #10423 (`fac/template-level-reads-10419`), which is the branch this work was measured against.

## What changed

Materializing a component of an object template named the peers it copies from the subtemplate by their id (`{"id": relationship.peer_id}`), although the level read had just brought those peers back as nodes. The component is now handed the node itself — as it already was for the parent it hangs from.

- `Relationship.get_peer_in_hand()` returns the peer as the node it is, or `None` when the relationship holds only its id, so the caller can tell the two apart without triggering a read.
- `extract_peer_data()` hands that node over. A peer the template read did not bring back still falls through to its id, unchanged.

## Why

A trace of an upsert creating a device with 66 components from a customer template showed 135 `node_list_get_info` queries. 128 of them were two reads of the *same* node, repeated once per component:

1. `RelationshipPeerKindConstraint` read the peer's kind, because the copied id carried none.
2. `NodeCreateAllQuery.query_init` → `Relationship.get_create_data()` → `get_peer()` read the whole node, for nothing but `peer.id` and `peer.get_branch()`.

In that trace the peer was one shared SFP node (`sfp_model` on `NetworkPhysicalInterface`), read 64 times over. The info queries were cheap (0.33 s), but the attribute read that follows each one averaged 106 ms — 6.8 s of a 15.1 s mutation.

## Measured

A device template with 65 interfaces, each naming the same 5-attribute peer:

| | before | after |
|---|---|---|
| `node_list_get_info` | 131 | **3** |
| `node_list_get_attribute` | 67 | **3** |
| `uniqueness_constraint_validation` | 66 | 66 |
| `node_create_all` | 66 | 66 |
| `node_list_get_relationship` | 2 | 2 |
| **total** | **332** | **140** |

A component carrying a peer now costs the 2 queries a component without one costs: its uniqueness check and its write. Node reads stay flat at 3, measured identical at 5, 20 and 65 interfaces.

Projected onto the traced mutation with its own per-query timings: 345 → 153 queries, ~7.2 s saved of 15.1 s. The per-component HFID uniqueness check (3.6 s, growing with the sibling count) becomes the next term.

## Second commit: the same for peers of cardinality many

`NodeCreateAllQuery.query_init` batched every peer of a many-cardinality relationship through `RelationshipManager.get_peers()`, which reads whatever it is given — so a component naming its peers through such a relationship still read them back, once per component, even when they were handed over as nodes. The create query now asks for the peers held only by id (`RelationshipManager.read_peers_not_in_hand()`), in the same single call (1161a6898).

Same shape as above, the interface naming the tag through a relationship of cardinality many:

| | before | after |
|---|---|---|
| `node_list_get_info` | 68 | **3** |
| `node_list_get_attribute` | 68 | **3** |
| `uniqueness_constraint_validation` | 66 | 66 |
| `node_create_all` | 66 | 66 |
| `node_list_get_relationship` | 2 | 2 |
| **total** | **270** | **140** |

Node reads flat at 3, measured identical at 5, 20 and 65 interfaces. `test_a_peer_a_component_carries_is_not_read_back_for_every_component` now runs once per cardinality; the `many` case failed on the previous commit with node reads of 3 + N. Green: the suites below plus `component/graphql/resource_manager`, `test_mutation_relationship`, `mutations/test_profile_resource_pool_validation` and `queries/test_resource_pool` — 798 tests.

## Testing

- New component test `test_a_peer_a_component_carries_is_not_read_back_for_every_component` pins 3 node reads / 3 attribute reads and 2 queries per component, and reloads the created objects from the database to assert the peer was actually written.
- Green: `component/templates`, `component/core/constraint_validators`, `component/core/profiles`, `component/core/node`, `component/graphql/test_mutation_{create,upsert}`, `component/core/test_{node,relationship,relationship_manager,relationship_metadata}` — 728 tests.
- `mypy` and `ruff` clean on the touched files.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -217,9 +217,7 @@ async def extract_peer_data(
             # deeper templates are handled in the next level of recursion
             if await _peer_is_a_template(db=db, relationship=relationship):
                 continue
-            # The subtemplate was read with the peers its relationships name, so hand over the node
-            # where it is in hand: an id sends the checks on the new object, and the write itself,
-            # back to the database for a node already in memory.
+            # The peer is already read, so an id would send the checks and the write back to the database.
             rel_peers.append({"id": relationship.get_peer_in_hand() or relationship.peer_id})
 
         # Only set the relationship data if there are actual peers to set
--- a/backend/infrahub/core/node/create.py
+++ b/backend/infrahub/core/node/create.py
@@ -212,16 +212,19 @@ async def extract_peer_data(
             obj_peer_data[rel_name] = parent_obj
             continue
 
-        rel_peer_ids = []
+        rel_peers = []
         for relationship in relationships:
             # deeper templates are handled in the next level of recursion
             if await _peer_is_a_template(db=db, relationship=relationship):
                 continue
-            rel_peer_ids.append({"id": relationship.peer_id})
+            # The subtemplate was read with the peers its relationships name, so hand over the node
+            # where it is in hand: an id sends the checks on the new object, and the write itself,
+            # back to the database for a node already in memory.
+            rel_peers.append({"id": relationship.get_peer_in_hand() or relationship.peer_id})
 
         # Only set the relationship data if there are actual peers to set
-        if rel_peer_ids:
-            obj_peer_data[rel_name] = rel_peer_ids
+        if rel_peers:
+            obj_peer_data[rel_name] = rel_peers
 
         if rel_manager.schema.kind == RelationshipKind.PROFILE:
             obj_peer_data[rel_name] = peer_ids
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -1129,9 +1129,14 @@ class RelationshipManager[RelationshipManagerPeerType]:
 
         A relationship already holding its peer as a node is left as it is: the caller handed the node
         over, and reading it back is what this avoids. A relationship without a peer id yet, such as one
-        waiting on a resource pool, is left to `Relationship.resolve()`.
+        waiting on a resource pool, or naming its peer by a default filter value, is left to
+        `Relationship.resolve()`, which reads it the only way it can be read.
         """
-        peer_ids = [rel.peer_id for rel in self._relationships if rel.peer_id and rel.get_peer_in_hand() is None]
+        peer_ids = [
+            rel.peer_id
+            for rel in self._relationships
+            if rel.peer_id and is_valid_uuid(rel.peer_id) and rel.get_peer_in_hand() is None
+        ]
         if not peer_ids:
             return
 
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -204,6 +204,13 @@ class Relationship(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
 
         return self._resolved_peer_kind
 
+    def get_peer_in_hand(self) -> Node | None:
+        """Return the peer as the node it is, or None when this relationship holds only its id."""
+        if self._peer is None or isinstance(self._peer, str):
+            return None
+
+        return self._peer
+
     @property
     def node_id(self) -> str:
         if self._node_id:
--- a/backend/infrahub/core/relationship/model.py
+++ b/backend/infrahub/core/relationship/model.py
@@ -192,17 +192,13 @@ class Relationship(FlagPropertyMixin, NodePropertyMixin, MetadataInterface):
         return self.peer_id
 
     def get_peer_kind(self) -> str:
-        if not self._peer or isinstance(self._peer, str):
-            return self.schema.peer
-
-        return self._peer.get_kind()
+        peer = self.get_peer_in_hand()
+        return peer.get_kind() if peer is not None else self.schema.peer
 
     def get_concrete_peer_kind(self) -> str | None:
         """Return the peer's concrete kind, or None when only the schema's (possibly generic) peer kind is known."""
-        if self._peer and not isinstance(self._peer, str):
-            return self._peer.get_kind()
-
-        return self._resolved_peer_kind
+        peer = self.get_peer_in_hand()
+        return peer.get_kind() if peer is not None else self._resolved_peer_kind
 
     def get_peer_in_hand(self) -> Node | None:
         """Return the peer as the node it is, or None when this relationship holds only its id."""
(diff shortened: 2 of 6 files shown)


changelog/+template-component-peer-handover.changed.md
```diff
@@ -0,0 +1 @@
+Creating objects from an object template no longer reads back the objects its components point at, such as the transceiver model every interface of a device template names, once for each component created.
```
gmazoyer: This entry describes how it works rather than what a user gets. It is also not accurate as written. For cardinality-many relationships the peers are still read, because `core/query/node.py` around lines 216-222 calls `get_peers()`, which always goes to the database. I'd rewrite to something like: "Creating objects from an object template is now faster when the objects it creates share a related object, such as the transceiver model used by every interface of a device template." but it might be redundant with other fixes so far.

backend/infrahub/core/node/create.py
```diff
@@ -212,16 +212,19 @@ async def extract_peer_data(
...
-        rel_peer_ids = []
+        rel_peers = []
         for relationship in relationships:
             # deeper templates are handled in the next level of recursion
             if await _peer_is_a_template(db=db, relationship=relationship):
                 continue
-            rel_peer_ids.append({"id": relationship.peer_id})
+            # The subtemplate was read with the peers its relationships name, so hand over the node
+            # where it is in hand: an id sends the checks on the new object, and the write itself,
+            # back to the database for a node already in memory.
```
gmazoyer: I think "The peer is already read, so an id would send the checks and the write back to the database" would be enough as a comment here

backend/tests/component/templates/test_template_reads.py
```diff
@@ -265,3 +295,58 @@ async def test_a_level_of_a_template_is_read_once_however_many_parents_it_hangs_
...
+async def test_a_peer_a_component_carries_is_not_read_back_for_every_component(
+    db: InfrahubDatabase, default_branch: Branch, device_schema_with_tagged_interfaces: None
+) -> None:
+    """A peer a subtemplate names is handed to the object created from it, not read again per object.
+
+    The subtemplates arrive with the peers their relationships name, so a component that carries
+    one costs what a component without one costs: its uniqueness check and its write. Naming the
+    peer by its id instead sent both the kind check on the new object and the write back to the
+    database for a node already in memory, once per component.
+    """
```
gmazoyer: The last two sentences describe the old behavior. Keeping the first line is enough.

backend/infrahub/core/relationship/model.py
```diff
@@ -204,6 +204,13 @@ def get_concrete_peer_kind(self) -> str | None:
 
         return self._resolved_peer_kind
```
gmazoyer: `get_concrete_peer_kind` repeats the check that now has a name, and the two are slightly different: this one tests `if self._peer and ...` while `get_peer_in_hand` tests `is None`. Same result only because `Node` is always truthy.

backend/tests/component/templates/test_template_reads.py
```diff
@@ -265,3 +295,58 @@ async def test_a_level_of_a_template_is_read_once_however_many_parents_it_hangs_
     # read has no way to be told a node is already in hand. What matters here is that neither figure
     # grows with the width of a level.
     assert set(node_reads.values()) == {4}, f"reading the subtemplates grew with the width of a level: {node_reads}"
+
+
+async def test_a_peer_a_component_carries_is_not_read_back_for_every_component(
```
gmazoyer, line 304: That is a good test 👍 . I put it on the base commit and ran it against the unfixed code. It seems to prove what it tries to.

changelog/+template-component-peer-handover.changed.md
fatih-acar: Reworded to what a user gets, in 8c7a3db56. You're right on cardinality-many: `NodeCreateAllQuery.query_init` still batches those peers through `get_peers()` → `get_many()` once per component, so only the peer-kind read is skipped there; the measured case and the test are cardinality-one. Kept it as its own fragment rather than folding into `+faster-creates-from-object-templates`, which lives on #10423's branch.

backend/infrahub/core/node/create.py
fatih-acar: Done in 8c7a3db56.

backend/tests/component/templates/test_template_reads.py
fatih-acar: Done in 8c7a3db56.

backend/infrahub/core/relationship/model.py
fatih-acar: Done in 8c7a3db56 — `get_concrete_peer_kind` and `get_peer_kind` (the same check five lines up) both go through `get_peer_in_hand()` now.

changelog/+template-component-peer-handover.changed.md
fatih-acar: Follow-up: 1161a6898 closes the cardinality-many gap too — the create query now reads only the peers held by id (`RelationshipManager.read_peers_not_in_hand()`), 270 → 140 queries on the same 65-interface shape. Measurement is in the PR description.

backend/infrahub/core/relationship/model.py
```diff
@@ -1121,6 +1124,24 @@ async def get_peers(
...
         )
 
+    async def read_peers_not_in_hand(self, db: InfrahubDatabase, branch_agnostic: bool = False) -> None:
+        """Read, in one query, the peers the relationships hold only the id of, and hand each its node.
+
+        A relationship already holding its peer as a node is left as it is: the caller handed the node
+        over, and reading it back is what this avoids. A relationship without a peer id yet, such as one
+        waiting on a resource pool, is left to `Relationship.resolve()`.
+        """
+        peer_ids = [rel.peer_id for rel in self._relationships if rel.peer_id and rel.get_peer_in_hand() is None]
```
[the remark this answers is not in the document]
fatih-acar: Applied in 1c6e7ed9b — the batch is gated on `is_valid_uuid`, the rule `get_peer_id()` already uses. Two corrections to the premise: a peer named by hfid never had a `peer_id` (it goes to `peer_hfid`), so it was already skipped; and the default-filter read was not new — `get_peers()` passed the same ids before this commit — and cannot happen on the `create_node` path, where the constraint runner resolves relationships before the save. The gate matters for a direct `Node.save()` whose many-cardinality peers are all named by a default filter value.

gmazoyer (approved):

PR #10432, opened 2026-08-27
perf(ci): upgrade mypy to 2.3.1 and run python-lint on arm64

fatih-acar:
Follow-up to #9807. Two commits, independently reviewable.

## The problem

#9807 moved mypy into `python-lint`, which makes that job the critical path for every backend PR — component, integration, functional, docker-integration, E2E and benchmark all gate on it. Step timings from the last run on #9807 ([job 98499925076](https://github.com/opsmill/infrahub/actions/runs/33067127650)):

| Step | Duration |
| --- | --- |
| set up job + checkout + Python + uv | 27s |
| `uv sync --all-groups` | 5s |
| ruff check + ruff format | 1s |
| ty check | 5s |
| **mypy** | **341s** |
| **total** | **379s** |

## 1. `ci: run python-lint on an arm64 runner`

mypy has no internal parallelism, so the runner's four vCPUs buy nothing and a larger runner would not help either — only a faster core does. `ubuntu-24.04-arm` has a higher per-core throughput than the x64 runners, is free for public repositories, and is already used by `ci-docker-image.yml` and `push-bench-image.yml`.

Verified safe on aarch64 before pushing:

- Every package in `uv.lock` that resolves to a Linux binary wheel also ships an `aarch64` wheel — re-checked after the mypy bump, including its new `librt` and `ast-serialize` dependencies. The only sdist-only package is `pytest-clarity`, which is pure Python.
- The `actions/python-versions` manifest carries a `linux/arm64` build of 3.14.3, so `actions/setup-python@v7` resolves without a source build.
- `actionlint` accepts the label.

Nothing in the job is architecture-dependent: no Docker builds, no compiled fixtures, no database.

## 2. `chore(deps): upgrade mypy to 2.3.1`

This is the real fix. **mypy was running as pure Python, not as the mypyc-compiled build.**

`mypy>=1.15,<1.16` resolved to a release that predates Python 3.14 and publishes compiled wheels only up to `cp313`, so on 3.14.3 both CI and local environments installed `mypy-1.15.0-py3-none-any.whl` — the interpreted build. Confirmed in a local venv: `mypy/build.py`, no `.so`. The first release with a `cp314` wheel is 1.17.1. The `<1.16` ceiling dates back to the uv migration and has no recorded reason.

Cold-cache runs over the same 1632 source files on one machine:

| mypy | build | cold run | vs before |
| --- | --- | --- | --- |
| 1.15.0 (before) | pure Python | 106s | — |
| 1.17.1 | mypyc | 32s | 3.4x |
| **2.3.1 (this PR)** | **mypyc** | **15s** | **6.9x** |

Warm-cache re-runs land at 3.8s.

2.3.1 reports 25 errors the interpreted 1.15.0 missed. All are fixed at the source, none suppressed:

- **4 × `var-annotated`** — `validated_database` (`database/__init__.py`), `background_tasks` (`services/scheduler.py`), `subscription_tasks` (`graphql/app.py`) and `publish_tasks` (`message_bus/nats.py`) were bare `{}` / `set()` literals. Annotated as `dict[str, bool]` and `set[asyncio.Task[None]]` from their actual use sites.
- **3 × `name-defined`** — the nats adapter annotated `on_callback`, `on_message` and one local with `nats.aio.msg.Msg` while only importing `nats`, so the name never resolved for any checker. `Msg` is now imported under `TYPE_CHECKING`.
- **18 × `union-attr`** — `test_migration_hierarchy_change` read `.parent` / `.children` off `db.schema.get()`, whose `MainSchemaTypes` union only carries them on `NodeSchema`. Switched to `db.schema.get_node_schema()`, which the sibling migration tests in the same directory already use (387 call sites across `backend/tests/`).

`AGENTS.md` records the linter versions, so its mypy entry moves with the pin.

## Combined effect — measured

This PR's own `python-lint` run ([job 98512726116](https://github.com/opsmill/infrahub/actions/runs/33070901530)):

| Step | Before (#9807) | This PR |
| --- | --- | --- |
| set up job + checkout + Python + uv | 27s | 20s |
| `uv sync --all-groups` | 5s | 5s |
| ruff check + ruff format | 1s | 1s |
| ty check | 5s | 4s |
| **mypy** | **341s** | **39s** |
| **total job** | **379s** | **74s** |

**8.7x off the mypy step, and 305s off the gate every backend test job waits on.** The ARM runner picked up the job with no meaningful queue wait.

The two commits cannot be separated within a single run, so the 39s is their combined effect. Indirectly: the same local machine runs interpreted 1.15.0 in 106s and compiled 2.3.1 in 15.3s, so the x64 runner was ~3.2x slower than local and the arm64 runner is ~2.5x slower — putting arm64 roughly 25% ahead per core, in line with expectations. The bump is doing the bulk of the work; the runner change is the smaller half, kept because it is free.

## Validation

- `uv run mypy --show-error-codes backend` → clean, 1632 files, 15.3s cold / 3.8s warm.
- `ruff check`, `ruff format --check`, `ty check` → all clean.
- `actionlint .github/workflows/ci.yml` → clean.
- Every touched module imports at runtime; the four annotations and the `TYPE_CHECKING` import have no runtime effect.
- The remaining proof is this PR's own `python-lint` run against the 341s / 379s baseline.

## Still on the table (not in this PR)

- **Split mypy into its own job.** ruff + ty finish 33s in; mypy then holds the gate for the rest of the job while every heavy test job waits. A separate `python-mypy` job would take it off the critical path entirely. Much less urgent now that the step is seconds rather than minutes.
- **Cache `.mypy_cache`.** 227MB / 10.7k files, and it would need write-on-default-branch-only to avoid eating the repo's 10GB Actions cache budget. Not worth the complexity against a 15s cold run.
- **Eventually drop mypy for `ty`.** Not actionable at `ty` 0.0.32 with this repo's per-module override set, but it is the direction that removes the problem rather than shrinking it.


changelog/+mypy-2.housekeeping.md
```diff
@@ -0,0 +1 @@
+Upgraded mypy to `2.3.1`. The previous `<1.16` pin resolved to a release that predates Python 3.14 and therefore ships no `cp314` wheel, so mypy installed as its interpreted build and type-checking the backend took over five minutes in CI. The compiled `2.3.1` wheel checks the same 1632 files in a fraction of that.
```
ogenstad: This doesn't hurt anything but I don't think it's that useful for end users.

ogenstad (approved):
Great find! We should have done this earlier..

changelog/+mypy-2.housekeeping.md
fatih-acar: Yeah slop I've just removed too late while you were already reviewing... sorry for that :sweat_smile:

ogenstad: No worries.

PR #10435, opened 2026-08-27
fix: let an SSO user sign in after their account is deleted (closes #10304)

### LINKED ISSUE #10304: bug: deleting an SSO user's account leaves its external identity behind, permanently locking the user out of SSO login
### Component

API Server / GraphQL

### Infrahub version

1.10.x (reproduced on `stable` @ `982a6c018`)

### Current Behavior

An SSO user (reported with Keycloak / OIDC) whose Infrahub account is deleted can never sign in again. Every subsequent login attempt fails with `(0) Failed to complete sign-in. Please try again.` in the UI and an unhandled 500 on the server.

Deleting the `CoreAccount` does not remove the `InternalExternalIdentity` node that was created alongside it, so the identity survives with a dangling `account` relationship. The next login still matches that orphan on `(sub, provider_name, protocol)`, takes the "returning user" path, and dereferences the missing peer:

```
File "/source/backend/infrahub/auth/auth.py", line 238, in _account_from_existing_identity
    account = await identity_node.get_relationship(name="account").get_peer(
        db=db, peer_type=CoreAccount, raise_on_error=True
    )
File "/source/backend/infrahub/core/relationship/model.py", line 1029, in get_peer
    raise LookupError("Unable to find the peer")
LookupError: Unable to find the peer
```

`LookupError` is not an `InfrahubError`, so nothing converts it into an API error — it escapes the endpoint as a 500.

The user is locked out permanently. The only workaround found by the reporter is to delete the user in the identity provider and recreate them, which issues a fresh `sub` that no longer matches the orphaned identity.

### Expected Behavior

Deleting an SSO user's account in Infrahub should return them to a clean slate: the next SSO login creates a new account, exactly as a first-time login does. Whatever the state of the data, a login must never surface as an unhandled 500.

### Steps to Reproduce

1. Configure SSO (OIDC, e.g. Keycloak) and sign in as a user, letting Infrahub create the account.
2. Delete that user's account in Infrahub.
3. Sign in again through SSO as the same provider user.
4. The UI shows `(0) Failed to complete sign-in. Please try again.` and the server logs the `LookupError: Unable to find the peer` traceback above.

### Additional Information

Reproduced against a live Neo4j:

```
DELETED: [('CoreAccount', '18cc9327-…')]     # succeeds, identity not removed
REMAINING IDENTITIES: 1
SECOND LOGIN RAISED: LookupError Unable to find the peer
```

Two separate defects meet here:

- **The account delete should never have left the identity behind.** `InternalExternalIdentity.account` is `optional=False`, so `NodeDeleteValidator` classifies it as a mandatory dependent and ought to refuse the delete. It does not, because the query it relies on unconditionally excludes the `Internal` namespace (`backend/infrahub/core/query/relationship.py:1156`, `NOT src.namespace IN $excluded_namespaces`), so `_analyze_delete_dependencies` (`backend/infrahub/core/node/delete_validator.py:145`) sees no dependents. Disabling that exclusion and repeating the delete raises the expected error:

  ```
  ValidationError: Cannot delete CoreAccount '...'. It is linked to mandatory relationship account
  on node InternalRefreshToken '...', Cannot delete CoreAccount '...'. It is linked to mandatory
  relationship account on node InternalExternalIdentity '...'
  ```

  The same gap orphans `InternalAccountToken` and `InternalRefreshToken` rows on any account delete. Note it cannot be closed by lifting the exclusion alone: the guard would then make any account that has ever logged in undeletable, since its refresh token is also a mandatory dependent.

- **Independently, `signin_sso_account` assumes an identity node always has a live account peer** (`backend/infrahub/auth/auth.py:192-242`). Deployments that already contain orphaned identities stay broken even once the delete side is corrected, so this path has to cope on its own.

### Related issues

The delete side of this bug has surfaced twice before under different symptoms, both closed as not planned:

- #7202 — deleting an account does not invalidate active sessions/tokens (closed 2025-09-11, not planned).
- #8955 — deleting the admin account breaks worker communication (closed 2026-05-20, not planned); also asks for a delete guard on the built-in service accounts and for self-deletion prevention.

The same underlying fact drives all three: deleting a `CoreAccount` cleans up none of its Internal-namespace children, because the delete-dependency query cannot see them.

Related but a distinct defect in the same component — a race rather than a visibility gap, so not a duplicate:

- #6849 — cascade delete can miss related nodes added while the delete logic is running.

gmazoyer:
# Why

An SSO user whose account is deleted can never sign in again. The delete leaves the `InternalExternalIdentity` behind, the next login matches it on `(sub, provider_name, protocol)`, finds no account and fails with an unhandled 500.

Closes #10304

Depends on https://github.com/opsmill/infrahub-sdk-python/pull/1307

## What changed

- **The delete stops orphaning.** `on_delete=CASCADE` on the account's `external_identities`, `tokens` and `refresh_tokens`. The account side of `tokens` also had no identifier, so it never paired with the token's own `account__token` and matched no edge in the graph.
- **The cascade can reach them.** `RelationshipGetByIdentifierQuery` forced `Internal` into its exclusion list, which hid all three children from `NodeDeleteValidator`. The exclusion moved to the public GraphQL relationship query, the only caller that wants it.
- **`Migration077`** clears what earlier releases already orphaned, for all three kinds. `GRAPH_VERSION` goes to 77.

All three children have to cascade, not only the two that caused visible trouble. A mandatory relationship that is not cascaded becomes a blocking dependent, and every account that has ever logged in owns a refresh token, so leaving that one out would make those accounts impossible to delete.

No API surface change: `refresh_tokens` is an internal peer, so it stays out of GraphQL and the schema API.

## How to review

Dropping the `Internal` exclusion is only safe because of a fact worth checking rather than trusting: across the core schema there are exactly three mandatory relationships declared on any `Internal` kind, and all three point at `CoreGenericAccount`. Every other one is optional, so it cannot block a delete, and the two IPAM internal kinds are virtual and never stored.

Two commits cancel out. The login path first gained a guard for an identity with no account, then lost it again: with the cascade and the migration in place that state cannot exist on an upgraded database, so `auth.py` ends up untouched.

`Migration077` is plain Cypher rather than `NodeManager`, matching every migration since 069. A migration has to keep working against an old graph, so it cannot depend on the current node layer.

## How to test

```bash
uv run pytest backend/tests/unit/core/node/test_delete_validator.py
uv run pytest backend/tests/component/core/test_signin_sso_account.py --neo4j
uv run pytest backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py --neo4j
```

`test_login_after_account_deletion_creates_a_new_account` is the regression test. With the cascade reverted it fails with `LookupError: Unable to find the peer`, the error that locked the user out. Every other test here was checked the same way, by reverting its fix and confirming it fails.

## Impact & rollout

Deleting an account now also deletes its external identities, API tokens and refresh tokens. Needs `infrahub upgrade` for the new migration.

The submodule pins the SDK branch commit that carries the generated `refresh_tokens` protocol. That SDK PR has to merge first, then the pointer needs re-pinning to the squashed commit on SDK `stable`.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added


backend/infrahub/auth/auth.py
```diff
@@ -196,13 +196,17 @@ async def signin_sso_account(
...
     existing_identity = await _find_existing_identity(db=db, external_identity=external_identity)
-    is_first_login = existing_identity is None
+    account: CoreAccount | None = None
     if existing_identity is not None:
         account = await _account_from_existing_identity(
             db=db, identity_node=existing_identity, external_identity=external_identity
         )
-    else:
+
+    if account is None:
```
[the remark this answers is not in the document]
gmazoyer: If this is indeed valid, this is pre-existing, so it would worth a follow-up issue and PR to fix it.

backend/infrahub/auth/auth.py
```diff
@@ -233,15 +237,34 @@ async def _find_existing_identity(*, db: InfrahubDatabase, external_identity: Ex
...
+    """
+    account = await _peer_account_or_none(db=db, identity_node=identity_node)
+    if account is None:
+        log.warning(
+            "sso identity has no account, dropping it to sign in as a new user",
+            provider_name=external_identity.provider_name,
+            protocol=external_identity.protocol,
+        )
+        await identity_node.delete(db=db)
+        return None
```
polmichel: If the relationship deletion cascade and the migration work, the `account is None` block should no longer be reachable, right?

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
```diff
@@ -0,0 +1,133 @@
...
+
+    migration = Migration077()
+
+    # A clean validation after the run only means something if it was dirty before.
+    before = await migration.validate_migration(db=db)
+    assert before.errors
+    assert orphaned_identity.get_id() in " ".join(before.errors)
+    assert orphaned_token.get_id() in " ".join(before.errors)
+    assert live_identity.get_id() not in " ".join(before.errors)
+
```
polmichel, line 67: Can we do a strict assertion here regarding `before.errors` content?

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
```diff
@@ -0,0 +1,133 @@
...
+RETURN count(rel) AS removed
+        """,
+        params={"uuid": identity.get_id()},
+        type=QueryType.WRITE,
+    )
+    assert removed[0]["removed"] > 0
+
+    migration = Migration077()
+    before = await migration.validate_migration(db=db)
+    assert identity.get_id() in " ".join(before.errors)
```
polmichel: Can we do another strict assertion here? It would make the system's behavior more readable.

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -475,6 +475,90 @@ async def test_unknown_group_is_silently_ignored(
...
 
+async def test_login_recovers_from_an_orphaned_identity(
+    db: InfrahubDatabase, default_branch: Branch, register_core_models_schema: SchemaBranch
+) -> None:
+    """A deployment upgraded from a release that orphaned identities must still let the user in.
+
+    Node.delete leaves the identity behind, which is the state those releases produced. The login
+    has to drop that identity and start over, since the uniqueness constraint on
+    (sub, provider_name, protocol) would otherwise reject the replacement.
+    """
```
polmichel: After the migration has been executed, there should not be any orphaned identities anymore right?
If that's true and I am not missing something else, this test would not be required anymore since `test_migration_077_identity_without_relationship_node` is already covering this test case.

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -475,6 +475,90 @@ async def test_unknown_group_is_silently_ignored(
...
+        schema=InfrahubKind.EXTERNALIDENTITY,
+        filters={"sub__value": "sub-orphaned-001", "provider_name__value": "provider1", "protocol__value": "oidc"},
+    )
+    assert len(identity_nodes) == 1
+    assert identity_nodes[0].get_id() != orphan_id
+    linked_account = await identity_nodes[0].account.get_peer(db=db)
+    assert linked_account.id == second_login.account_id
+
+
+async def test_login_after_account_deletion_creates_a_new_account(
```
polmichel, line 478: In my opinion, this one is describing the true scenario that should be protected against regressions

backend/tests/unit/core/node/test_delete_validator.py
```diff
@@ -84,3 +84,46 @@ def test_repository_cascade_reaches_exactly_expected_kinds(case: RepositoryCase)
...
+        (full_id.source_kind, full_id.identifier, full_id.destination_kind)
+        for full_id in account_index.get_internal_cascade_relationship_identifiers()
+    } == {
+        (InfrahubKind.ACCOUNT, "account__external_identity", InfrahubKind.EXTERNALIDENTITY),
+        (InfrahubKind.ACCOUNT, "account__token", InfrahubKind.ACCOUNTTOKEN),
+    }
+
+    # A cascade to a peer outside the Internal namespace stays on the general lookup.
+    repository_index = _index_for(InfrahubKind.REPOSITORY)
+    assert repository_index.get_internal_cascade_relationship_identifiers() == []
```
polmichel: Nice test 👍

dev/knowledge/backend/authentication.md
```diff
@@ -72,6 +72,27 @@ Emission goes through the `AutoCreateEventEmitter` ABC. Two implementations exis
...
+
+Two rules follow for the SSO login path:
+
+- **An identity with no account is not a returning user.** `signin_sso_account` decides on whether there is an account to sign into, not on whether an identity row exists. An identity whose account is gone is deleted and the login continues as a first login. Deleting it is required rather than cosmetic: the uniqueness constraint on `(sub, provider_name, protocol)` rejects a replacement identity while the old row survives.
+- **A missing peer has two shapes.** No active relationship raises `LookupError` from the relationship manager; an active relationship to an unresolvable node raises `NodeNotFoundError`. Both mean "no account".
+
+`Migration077` deletes the identities and tokens that earlier releases orphaned. It does not make the login path above redundant: the graph-version check only logs a warning, so the server serves logins against a database whose migration has not run yet, and that is the window in which a locked-out user retries.
+
```
polmichel: If the assumption of the runtime code is true, this doc should also be updated

polmichel (commented):
A question regarding the runtime deletion which looks not useful anymore (+tests and documentation).
And optional comment about test assertions

backend/infrahub/auth/auth.py
gmazoyer: You are right, and I checked and ran the upgrade myself, this is indeed unnecessary. I removed the recovery: the login now raises a `ProcessingError` that names the fix, instead of repairing the data on its own.

backend/tests/component/core/test_signin_sso_account.py
gmazoyer: Removed.

dev/knowledge/backend/authentication.md
gmazoyer: Updated. It said the login recovers from an orphan and creates a new account, which is not true anymore. It now says an orphan is a data error, that the login reports it, and that the migration is where the repair happens.

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
gmazoyer: Done.

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
gmazoyer: Done.

backend/infrahub/auth/auth.py
```diff
@@ -234,14 +234,35 @@ async def _find_existing_identity(*, db: InfrahubDatabase, external_identity: Ex
...
-    """Return the account linked to an existing identity, refreshing its label when stale."""
-    account = await identity_node.get_relationship(name="account").get_peer(
-        db=db, peer_type=CoreAccount, raise_on_error=True
-    )
+    """Return the account linked to an existing identity, refreshing its label when stale.
+
+    Raises:
+        ProcessingError: When the identity has no account left. Deleting an account takes its
+            identities with it, so this only happens on a database that still holds the ones an
+            earlier release orphaned.
```
ajtmccarty: the comment mentions other areas of the code beyond the responsibility of this small function. it raises a ProcessingError if there is no account relationship on the identity_node. that's all it should say

backend/infrahub/core/migrations/graph/m077_delete_orphaned_account_children.py
```diff
@@ -0,0 +1,170 @@
...
+WHERE latest_edge.status = "active" AND latest_edge.to IS NULL
+
+// A closed edge keeps status "active" and gains a "to", so both have to be checked.
+CALL (node) {
+    MATCH (node)-[r1:IS_RELATED]-(:Relationship { name: $relationship })
+          -[r2:IS_RELATED]-(account:%(generic_account)s)-[account_part_of:IS_PART_OF]->(:Root)
+    WITH r1, r2, account_part_of
+      ORDER BY r1.branch_level DESC, r1.from DESC, r1.status ASC,
+               r2.branch_level DESC, r2.from DESC, r2.status ASC,
+               account_part_of.branch_level DESC, account_part_of.from DESC, account_part_of.status ASC
```
ajtmccarty: you can skip `branch_level` if the account and relationship are both branch-agnostic

backend/infrahub/core/migrations/graph/m077_delete_orphaned_account_children.py
```diff
@@ -0,0 +1,170 @@
...
+# Both kinds are branch-agnostic, so every edge of theirs lives on the global branch.
+ACTIVE_NODES_WITHOUT_ACCOUNT = """
+MATCH (node:Node)-[part_of:IS_PART_OF]->(root:Root)
+WHERE node.kind = $kind AND part_of.branch = $global_branch AND part_of.from <= $at
+WITH node, root, part_of
+  ORDER BY part_of.branch_level DESC, part_of.from DESC, part_of.status ASC
+WITH node, root, head(collect(part_of)) AS latest_edge
+WHERE latest_edge.status = "active" AND latest_edge.to IS NULL
+
+// A closed edge keeps status "active" and gains a "to", so both have to be checked.
```
ajtmccarty, line 27: might be worth adding a comment that this only works b/c everything involved is branch-agnostic. if the relationship or objects could have different active/deleted statuses on different branches, this would need to be one subquery per `node`-`account` combination instead of just one per `node`

backend/infrahub/core/node/delete_validator.py
```diff
@@ -72,6 +73,15 @@ def _index_cascading_deletes(
                     relationship_identifier=relationship_schema.get_identifier(),
                     peer_kinds=peer_kinds,
                 )
+                if relationship_schema.internal_peer:
```
ajtmccarty: the PR description indicates that
1. we are only doing this special internal peer handling so that RefreshTokens don't prevent accounts from being deleted
2. we don't want to add the refresh tokens relationship to the account in this PR to keep the PR small.

I would say that adding a RefreshToken relationship to the CoreGenericAccount is the smaller fix if it would allow skipping this `_internal_cascade_identifiers` change. if there are other reasons for this change, then that is fine, but adding the relationship to the CoreGenericAccount schema looks like the correct and easier fix to me.

backend/infrahub/core/node/delete_validator.py
```diff
@@ -180,6 +201,27 @@ async def _analyze_delete_dependencies(self, start_nodes: Iterable[Node], at: Ti
...
+    async def _get_peers(
+        self,
+        full_identifiers: list[FullRelationshipIdentifier],
+        at: Timestamp | str | None,
+        excluded_namespaces: list[str] | None = None,
+    ) -> list[RelationshipPeersData]:
+        """Resolve the peers of the given relationships.
+
+        Internal nodes are hidden by default, so a cascade that owns them has to ask for them
+        with no namespace excluded.
```
ajtmccarty: does not belong here. maybe belongs where this is called with no `Internal` namespace excluded

backend/infrahub/core/node/delete_validator.py
gmazoyer: It is best indeed. I was afraid that it would impact our tests a lot more but simply fixing one fixture was enough. So definitely a better solution.

backend/infrahub/auth/auth.py
gmazoyer: Trimmed. It now only says it raises `ProcessingError` when the identity has no account relationship.

backend/infrahub/core/migrations/graph/m077_delete_orphaned_account_children.py
gmazoyer: Added. The comment now says one subquery per node only works because these kinds are branch-agnostic, and that it would need one per node and account pair otherwise.

backend/infrahub/auth/auth.py
```diff
@@ -234,10 +234,26 @@ async def _find_existing_identity(*, db: InfrahubDatabase, external_identity: Ex
...
+        account = await identity_node.get_relationship(name="account").get_peer(db=db, peer_type=CoreAccount)
+    except NodeNotFoundError:
+        account = None
+
+    if account is None:
+        raise ProcessingError(
+            message=(
+                f"The {external_identity.protocol} identity for provider "
+                f"'{external_identity.provider_name}' is not linked to an account. "
+                "Run 'infrahub upgrade' to remove the identities left behind by a deleted account."
```
ajtmccarty: shouldn't you only get this error after upgrading to the latest version? at which point `infrahub upgrade` will have already run and the migration will have been applied. so if there's an error here, running an upgrade probably won't fix it

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
```diff
@@ -0,0 +1,150 @@
...
+    assert await _existing_ids(db=db, kind=InfrahubKind.ACCOUNTTOKEN) == {live_token.get_id()}
+    assert await _existing_ids(db=db, kind=InfrahubKind.REFRESHTOKEN) == {live_refresh_token.get_id()}
+    assert await _existing_ids(db=db, kind=InfrahubKind.ACCOUNT) == {live_account.get_id()}
+
+    reloaded_identity = await NodeManager.get_one(db=db, id=live_identity.get_id(), raise_on_error=True)
+    linked_account = await reloaded_identity.get_relationship(name="account").get_peer(db=db)
+    assert linked_account is not None
+    assert linked_account.get_id() == live_account.get_id()
+
+    await verify_graph(db=db)
```
ajtmccarty: this one usually goes at the very end

backend/tests/component/core/test_signin_sso_account.py
```diff
@@ -475,6 +475,74 @@ async def test_unknown_group_is_silently_ignored(
     assert len(accounts) == 1
 
 
+async def test_login_with_an_orphaned_identity_reports_a_handled_error(
```
ajtmccarty: isn't this what the migration fixes?

ajtmccarty (approved):

backend/tests/component/core/migrations/graph/test_077_delete_orphaned_account_children.py
gmazoyer: Moved to the end. It sat before the second run, so the edges that run touched were never checked.

backend/infrahub/auth/auth.py
gmazoyer: The more I looked at it, the more strange I felt about this indeed. The cascade and the migration mean the state cannot exist on an upgraded database, so the login has nothing left to guard.

dev/knowledge/backend/authentication.md
```diff
@@ -72,6 +72,26 @@ Emission goes through the `AutoCreateEventEmitter` ABC. Two implementations exis
...
+| Kind | Identifier |
+|------|------------|
+| `InternalExternalIdentity` | `account__external_identity` |
+| `InternalAccountToken` | `account__token` |
+| `InternalRefreshToken` | `account__refreshtoken` |
+
+All three are declared on the account side with `on_delete=CASCADE`, so a delete takes them with it. All three must be, not just the ones that cause visible trouble: a mandatory relationship that is not cascaded becomes a blocking dependent, and a refresh token exists for every account that has ever logged in, which would make those accounts impossible to delete.
+
+The account side of each has to name the same identifier as the child's own `account` relationship. An account-side relationship left without one gets a generated identifier, matches no edge in the graph, and the cascade silently skips that child.
```
ajtmccarty: seems like base infrahub relationship schema info. this must be covered elsewhere

dev/knowledge/backend/authentication.md
```diff
@@ -72,6 +72,26 @@ Emission goes through the `AutoCreateEventEmitter` ABC. Two implementations exis
...
+
+These three are the only mandatory relationships declared on any `Internal` kind, which is why `NodeDeleteValidator` can resolve peers without excluding the namespace at all. The `Internal` exclusion remains on the public GraphQL relationship query, where it keeps bookkeeping nodes out of the API.
+
+`Migration077` deletes the children of all three kinds that earlier releases orphaned. Together with the cascade, an upgraded database holds no identity without an account, so the SSO login path does not handle that state: it is a data error, and the migration is where the repair belongs.
```
ajtmccarty: not really sure if this belongs here. maybe it's helpful for robot readers

dev/knowledge/backend/authentication.md
```diff
@@ -72,6 +72,26 @@ Emission goes through the `AutoCreateEventEmitter` ABC. Two implementations exis
...
+
+A login that meets one on a database that has not been upgraded still fails with `LookupError`, as it did before this change. Nothing stops such a database from serving traffic, because a graph version mismatch only logs a warning.
```
ajtmccarty: I'm not sure this is really saying anything

ajtmccarty (approved):

dev/knowledge/backend/authentication.md
```diff
@@ -72,6 +72,22 @@ Emission goes through the `AutoCreateEventEmitter` ABC. Two implementations exis
...
+|------|------------|
+| `InternalExternalIdentity` | `account__external_identity` |
+| `InternalAccountToken` | `account__token` |
+| `InternalRefreshToken` | `account__refreshtoken` |
+
+All three are declared on the account side with `on_delete=CASCADE`, so a delete takes them with it. All three must be, not just the ones that cause visible trouble: a mandatory relationship that is not cascaded becomes a blocking dependent, and a refresh token exists for every account that has ever logged in, which would make those accounts impossible to delete.
+
+These three are the only mandatory relationships declared on any `Internal` kind, which is why `NodeDeleteValidator` can resolve peers without excluding the namespace at all. The `Internal` exclusion remains on the public GraphQL relationship query, where it keeps bookkeeping nodes out of the API.
+
+The cascade is why the SSO login path does not handle an identity whose account is gone: it cannot happen, and repairing data during a login would hide the inconsistency.
```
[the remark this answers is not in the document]
gmazoyer: We do not want to keep history in there

Issue #3289, opened 2024-05-13
feature: deleting a generator target object should delete the created objects of that target 


wvandeun:
### Component

API Server / GraphQL

### Current Behavior

It is not clear how a generator target object should be properly deleted within Infrahub.

- When you delete the target node from the target group, then the generator does not seem to delete all the objects that the generator created as a result of the target.

- When you delete the target node from Infrahub, then the generator will raise an error message.

### Expected Behavior

We should have a clear and defined way to delete generator target nodes and the resulting objects.

### Steps to Reproduce

- load the following schema
```yaml
---
version: "1.0"
nodes:
  - name: TagService
    namespace: Test
    attributes:
      - name: name
        label: Name
        kind: Text
        optional: false
        unique: true
      - name: amount
        label: Amount
        kind: Number
        optional: false
```
- create a group called `tag_services` in Infrahub
- define a generator in an external repository
```yaml
generator_definitions:
  - name: tag_service_generator
    file_path: "test_tag_service.py"
    targets: tag_services
    query: test_tag_service
    class_name: TestTagService
    parameters:
      name: "name__value"
```
```graphql
query TagService($name: String!) {
  TestTagService(name__value: $name) {
    edges {
      node {
	name {
	  value
	}
	amount {
	  value
	}
      }
    }
  }
}
```
```python
from infrahub_sdk.generator import InfrahubGenerator

class TestTagService(InfrahubGenerator):
    async def generate(self, data: dict) -> None:
        service = data["TestTagService"]["edges"][0]["node"]

        name = service["name"]["value"]
        amount = service["amount"]["value"]

        for i in range(amount):
            tag = await self.client.create(kind="BuiltinTag", name=f"{name}-{i+1}")
            await tag.save(allow_upsert=True)
```
- create/syn the external repository in Infrahub
- create a new branch in Infrahub
- create a TestTagService object with name test and amount 3
- add the TestTagService object to the `tag_services` group
- create a proposed change to merge the branch into main
- 3 tags will have been created
- merge the branch
- create a new branch in Infrahub
- delete the TestTagService object from the `tag_services` group
- open a proposed change for the branch
- the generator will run, but it will not delete the tag objects 
- open a new branch in Infrahub
- delete the TestTagService object
- open a proposed change for the branch
- the generator will raise an error message (and the tag object will not have been deleted)

PR #10437, opened 2026-08-27
docs: state what starts an artifact and Generator run

lancamat1:
## Why

A user asked what actually starts an artifact generation run and a Generator run for a definition that targets a group: is adding a member to the group enough, and does it apply to the default branch or to every branch. Neither question can be answered from the docs today, and the artifacts overview points the wrong way.

The current "When artifacts regenerate" section reads as if regeneration were event driven: "The target's data changes, a node read by the artifact's GraphQL query is modified, so that target's artifact is regenerated" and "A new target joins the group, an artifact is generated for the new member". There is no such event. Every submitter of `REQUEST_ARTIFACT_DEFINITION_GENERATE` is one of four things:

- the artifact definition create mutation, `backend/infrahub/graphql/mutations/artifact_definition.py:67`
- the artifact definition update mutation, `backend/infrahub/graphql/mutations/artifact_definition.py:94`
- `POST /api/artifact/generate/{artifact_definition_id}`, `backend/infrahub/api/artifact.py:122`
- the post merge regeneration dispatcher, `backend/infrahub/core/merge/regeneration_dispatcher.py:156-162`

Repository import and the proposed change pipeline are not separate entry points. Import creates or updates the definition through the SDK (`backend/infrahub/git/integrator.py:752,759-780`) and so reaches generation through the mutation, and the pipeline uses a different workflow chain, `REQUEST_PROPOSED_CHANGE_REFRESH_ARTIFACTS` into `validate_artifacts_generation` into `GIT_REPOSITORIES_CHECK_ARTIFACT_CREATE` (`backend/infrahub/proposed_change/tasks.py:492,748,887`).

`builtin_triggers` in `backend/infrahub/trigger/catalogue.py` has no consumer for `infrahub.node.updated` or `infrahub.group.member_added`. So a data or membership change does not regenerate an artifact on its own, on any branch.

## What changed

- `artifacts/overview`: the section is split into `### What starts an artifact generation run` and `### What a run regenerates or deletes`, with a callout for what does not start one. Adds the three facts a reader needs next: there is no staleness indicator, the `nodes` list takes artifact IDs and a request built from target object IDs returns no error, and artifacts created on a branch are deleted with the branch while the rendered files are not.
- `artifacts/use`: the intro promised artifacts that "automatically update when your infrastructure data changes". Replaced, and the guide now has a "Regenerate an artifact" step for the two buttons its checklist named but never showed.
- `generators/overview`: the Events and Actions entry named no mechanism, so it now names the group and node trigger rules and the Generator action. Adds the missing fifth execution path, the run after a merge, and a callout for what does not start a Generator run, including the rebase case.
- `groups/use-in-automation`: "the next artifact generation run picks up the new members automatically" never said what starts a run. It now says membership decides scope and not timing, and links rather than restating the recipe.
- `events/event-rules`: new Branch scope section. `branch_scope` was documented nowhere, and its default is `default_branch` (`backend/infrahub/actions/schema.py:56-65`), which answers the default branch or all branches half of the question.

Terminology: "generation pass" is replaced by "artifact generation run". The old term came from `changelog/9790.changed.md` into the release notes and then into one docs bullet. It is in no UI label, API field, schema attribute or task name, while "Generator run" is already the word for the same event on the Generator side.

## Reviewer attention

- `branch_scope` overlaps with #10467, which documents the same field on the same page in a different shape. One of the two has to own it. Blocked until #10467 lands.
- The `branch_scope` default of `default_branch` on `CoreTriggerRule` is the opposite of what `learn/tutorials/generators/build-chained-generators.mdx:293` recommends. A user following the UI procedure in `events/event-rules.mdx` accepts the default and gets a rule that cannot fire inside a branch. Please confirm the default is intended.
- Not covered here: the "When Generators run" spoke and the trigger reference table, both handed to @yjouffrault.

## Validation

- `npm run build` passes, no broken links or anchors.
- `uv run invoke docs.lint`: 0 errors, and no warnings in any of the five changed files.
- Tested against a running instance with `infrahub-testcontainers` on image `1.11.0`, using the `car-dealership` fixture repository, which gives artifact definitions, generators and a `people` target group in one import. The backend files that carry this behaviour are identical between `infrahub-v1.11.0` and `origin/develop`. 10 tests, all passing:
  - renaming a person on the default branch, an attribute the artifact query reads and the template renders, left the artifact serving the old value for a 60 second window
  - adding a person to the target group produced no artifact for a 60 second window
  - updating the artifact definition then regenerated every current member, picking up both the rename and the new member
  - `nodes` with a target object ID regenerated nothing, and the same call with the artifact ID regenerated it
  - a limited run left a departed member's artifact in place, and a full run deleted it
  - a `CoreGroupTriggerRule` created without `branch_scope` read back as `default_branch`


yjouffrault:
## Docs review — placement, content, and writing

Thanks for digging into this — the core insight is right and nobody had written it down before. I read all 45 new lines as if they were a brand-new page, and verified every factual claim against the code.

**Every fact in this PR is correct.** The problems are in placement, in what's missing, and in the writing.

Two decisions from me up front, since they change the shape of the rewrite:

1. **Add a spoke for the Generator execution paths** (detail in Placement below).
2. **Drop "generation pass" in favour of "artifact generation run"** — I traced the term and it isn't ours (detail below).

**Offer:** I'm happy to take this on myself — reorganize the content, apply the rewrites, and leave the open questions as review comments for you to answer. Say the word and I'll push to this branch. If you'd rather do it, please run the `writing-infrahub-docs` skill over the rewrite; notes on which rules bite are at the bottom.

---

### On "generation pass" — it isn't our term

I traced where it came from:

1. `changelog/9790.changed.md` — an engineer wrote "the next full generation pass over the definition".
2. That became `CHANGELOG.md`, then the 1.11.0 release notes (lines 41, 323, 487).
3. From there one sentence carried it into `artifacts/overview.mdx` — the bullet this PR rewrote.

Outside those it appears in one test name and one internal spec. It's in no UI label, no API field, no schema attribute and no task name. So it's changelog vocabulary that leaked into the docs, not a word the product uses.

The bare word **"pass"** is weaker still — it appears nowhere else in the docs, and this PR uses it 11 times against 2 uses of the full term. It also collides with the pass/fail sense of the word, which costs readers working in a second or third language.

**Use "artifact generation run", shortened to "run" after first use.** Sourced three ways:

- "Generator run" is already our word for the same event on the Generator side — about two dozen uses, including "one independent run per member of the target group" on `generators/overview.mdx`, which this PR also edits. One concept across both features instead of two.
- "Artifact generation" appears 30 times across the docs.
- It maps to something a reader can see in their own instance: the task in the Tasks view is named **Generate all artifacts**.

"A full run" vs "a run limited to specific artifacts" still gives you the distinction the second list needs.

---

### Fix before merge

**1. The section's first sentence still states the misconception the PR removes.** `artifacts/overview.mdx:89`

> An artifact is the cached output of a Transformation, so Infrahub regenerates it when its inputs change.

Your own callout contradicts this 16 lines below. Anyone who reads the opening sentence and stops there — most people, on an overview — leaves with the wrong answer. It's old text that was kept, which is why it slipped through. Suggested:

> An artifact is the cached output of a Transformation. Infrahub regenerates it during an artifact generation run, and four events start a run:

**2. `artifacts/use.mdx` now names two buttons it never shows.** `:9`, `:153-154`

The intro sentence you removed was the guide's only mention of artifacts refreshing, so the guide no longer answers "I changed something, how do I regenerate it?" — while its final checklist names **Generate** and **Re-generate**, which appear nowhere else in it. The other checklist items are things you look at and confirm; these two are statements of behaviour.

Add a step before Validation (on the artifact definition page click **Generate**; on a single artifact's page click **Re-generate**), then make the checklist items match the others.

**3. Two sentences broke when em dashes became commas.** `artifacts/overview.mdx:101`, `generators/overview.mdx:216`

> An unrelated commit, a README edit or a helper no Transformation uses, regenerates nothing

That reads as three subjects until you hit the singular verb. The original was unambiguous:

> An unrelated commit — a README edit, or a helper no Transformation uses — regenerates nothing

The generators one became a comma splice: "not automatically cleaned up, see #3289" → use a semicolon. For what it's worth, nothing asks for this change: `dev/guidelines/markdown.md` says nothing about em dashes, none of the Vale rules mention them, and our hub-page format requires them.

**4. A Generator doesn't set its own flag.** `generators/overview.mdx:49`

> The Generators the merge affected run again on the target branch, unless the Generator sets `execute_after_merge` to `false`.

A person sets it, in `.infrahub.yml` — as the same page says at line 267 — so a reader who wants to turn it off isn't told where to go. "The Generators the merge affected run again" also drops the "that", so it reads as one long noun phrase while you wait for a verb. Suggested:

> After a merge, Infrahub runs the Generators the merge affected again on the target branch, unless `execute_after_merge` is `false` in the Generator's `.infrahub.yml` entry.

Also: `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE` governs this for Generators too, not only artifacts (`config.py:219-222`, `regeneration_dispatcher.py:40-49`). The artifacts list mentions it; this one doesn't.

**5. The `nodes` trap is described too gently.** `artifacts/overview.mdx:94`

> An entry that matches no existing artifact selects nothing, so a request that lists target node IDs instead regenerates nothing.

Correct, but it stops short of what makes it a trap. I traced the route: `api/artifact.py:68-135` queues the job and returns HTTP 200 **without validating a single ID**. So someone automating a targeted regeneration gets a green response every time while their artifacts keep the old content. Your Validation notes found this. Suggested ending:

> …so a request that lists target node IDs regenerates nothing and returns no error.

---

### Content gaps

**6. A reader can't tell which artifacts are now stale.** This is the biggest one, and it follows straight from what the PR establishes. Having learned that data changes don't regenerate anything, the immediate next question is "so which of mine are serving old content?" — and nothing answers it.

I checked whether it can be answered: `status` is Ready / Pending / Processing / Error (`core/constants/__init__.py:167-171`), which reports the last generation attempt, not freshness. An artifact whose device was renamed an hour ago still reads Ready. So there's no staleness indicator, and the page should say so plainly — that changes what a reader does, where silence sends them looking for a feature that doesn't exist.

**7. Rebase isn't in the list.** The four listed starters are genuinely all of them — I checked every submitter. But the list presents itself as complete, and rebasing a long-lived branch is the fifth thing readers will assume refreshes artifacts. Silence reads as "not applicable" rather than "nothing happens". One clause in the callout covers it.

**8. What happens to artifacts on a branch that never merges?** Item 1 says artifacts are regenerated on the source branch during a PC pipeline. If the PC is closed or the branch deleted, what becomes of them? I haven't verified this, so it's a question rather than a finding — worth confirming and stating in a line, because "I generated artifacts in a branch, where did they go" is a support ticket waiting to happen.

**9. The most useful fact in the PR is where nobody will look for it.** The callout says artifact definitions aren't available as an event action, so no event rule can regenerate an artifact. That asymmetry — Generators can be event-driven, artifacts can't — is genuinely useful, and it's stated once, mid-callout, on the artifacts hub. Someone wanting event-driven regeneration opens `events/event-actions.mdx`, which lists the actions that exist and says nothing about artifact definitions being absent. One line there.

**10. Branch scope is documented, but not where to set it.** The two "Configure a … trigger rule" procedures on `events/event-rules.mdx` list the fields to fill in, and branch scope isn't among them. Follow the steps exactly and you never set it — so you get the default, which is the one that won't fire in a branch.

---

### Placement

**11. Add a "When Generators run" spoke — Option A.** `generators/overview.mdx` is 311 lines and is already the page we point at as the example of a hub that filled up with execution mechanics: someone arriving to find out whether Generators are for them has a long way to read first. This PR adds nine more lines of mechanics to it.

Each addition is defensible on its own; together they're the same pattern that got the page here — a five-item execution list, two items carrying configuration detail, then a paragraph on what doesn't run a Generator.

I'd like a spoke: **"When Generators run"**, carrying the execution paths, the trigger-rule wiring, the post-merge behaviour and the "what doesn't start a run" callout. The hub keeps five one-line entries and a link. That's a bigger change than this PR started as, so I'm happy to do it as part of the offer above.

**12. `branch_scope` is now documented twice.** I checked whether it belonged in Reference instead — it doesn't; `reference/schema/` holds specification pages, not per-kind field tables, and there's no `CoreTriggerRule` page. So `events/event-rules.mdx` is the right home.

The problem is `webhooks/overview.mdx:45-70`, which already documents the same field: in UI labels ("Other Branches") rather than API values, with something your section doesn't have — which branch each event is emitted on, and therefore which scopes can never match. Neither page links to the other. Two descriptions of one field, in two vocabularies, will drift within a release. Make this the canonical one and have webhooks link here for the branch semantics.

**13. The groups page duplicates rather than links.** `groups/use-in-automation.mdx` is about targeting; timing belongs to artifacts and Generators. The artifacts paragraph gets this right — scope not timing, then a link. The Generators paragraph restates the trigger-rule recipe, which by then appears twice on the generators hub. Link instead.

**14. The reference table you identified is the real fix.** Your note about a "trigger per feature per context" table is right, and it's the structural answer: without it these facts get retold in pieces on four pages, which is why the same sentence now appears five times. Worth an issue — Reference section, alongside the event reference pages, and Checks and computed attributes fold into the same table.

---

<details>
<summary><b>Smaller writing items (11)</b></summary>

**15. Two lists of four bullets, nothing to tell them apart.** `artifacts/overview.mdx:89-101` — the first answers what starts regeneration, the second what a run then touches. Both have four items, both are about regeneration, one line separates them under a single heading. Someone skimming sees eight bullets and reads the second four as more triggers. The description says the section is split; on the page the split is invisible. Suggest `### What starts an artifact generation run` and `### What a run regenerates or deletes`.

**16. Item 4 says a new group member starts a run; the paragraph below says it doesn't.** `generators/overview.mdx:46` and `:51`. Both are true — item 4 assumes a trigger rule is set up, the paragraph assumes none is — but neither says so. Put the exclusion after the list as one statement: "Nothing else starts a Generator run. A data change, or an object joining the target group, does nothing on its own until one of the paths above picks it up." Also, that paragraph says "one of the paths above" then lists four of five, omitting `infrahubctl`.

**17. A button label described as doing the work.** `artifacts/overview.mdx:94` — "**Generate** on an artifact definition runs a pass over every member". A label doesn't run anything, and the sentence never tells the reader to click or where the button is. → "Click **Generate** on an artifact definition to regenerate every member of the group, or **Re-generate** on a single artifact to regenerate only that one."

**18. A sentence under the branch scope table repeats the table.** `events/event-rules.mdx:26` — "A rule created without changing this field fires on the default branch only" is already stated twice by the row above it. The consequence is the useful part, and it's the answer to the question that started this PR: "The default is `default_branch`, so a rule created without setting this field never fires inside a branch or a Proposed Change."

**19. That section restates advice that exists elsewhere.** The `other_branches` recommendation is already made, with reasoning, at `learn/tutorials/generators/build-chained-generators.mdx:293` and `generators/modular-best-practices.mdx:258,298`. Link to it. And say on the page that the default is the opposite of what we recommend — someone following the tutorial and someone creating a rule in the UI currently end up with different behaviour, with nothing warning them.

**20. "proposed change" lowercase next to "Proposed Change".** `artifacts/overview.mdx:91`, `events/event-rules.mdx:26`, `generators/overview.mdx:51`, `groups/use-in-automation.mdx:27`. The sentences around the new text capitalise it — the same artifacts page at line 36, and item 3 of the same list on the generators page. The docs are already split roughly 300/130 on this so it's not yours to fix everywhere, but making these four agree with their neighbours is quick.

**21. "Picked up", and an ambiguous "it".** `groups/use-in-automation.mdx:27` — nothing picks anything up; the next run includes the new members. "not when it runs" could mean the definition or the run. → "Group membership decides which objects an artifact definition covers, not when Infrahub generates them. The next generation run for the definition includes the new members, and a run starts only when…"

**22. The same sentence appears five times.** `generators/overview.mdx:46`, `:51`, `:216`; `groups/use-in-automation.mdx:27`, `:65`. Once per page is right; three times on one page tells the reader nothing new, and each copy can go stale on its own. This is the symptom of the missing reference table in 14.

**23. Callout on one page, plain paragraph on the other.** `artifacts/overview.mdx:103-107` vs `generators/overview.mdx:51`. The `:::info` treatment is the better one — separated, labelled, findable while skimming. On the generators page the same point reads as a footnote to item 5 rather than a statement about all five paths. Use the callout on both.

**24. Bare pointers.** "See [event rules] and [event actions]." twice, "See [Execution lifecycle]." once, "See [event rules]" once. A bare pointer makes the reader open the link to find out whether they needed it — say what's on the other side.

**25. Passive/active mixed within one list.** `artifacts/overview.mdx:91-93` — item 1 is "Artifacts are regenerated on the source branch", item 2 is "Infrahub regenerates the artifacts the merge affected". Same kind of statement, two shapes. Pick one.

</details>

<details>
<summary><b>What I verified, and how</b></summary>

Every factual claim checked against source rather than from memory:

| Claim | Where |
|---|---|
| Four things start a generation run, and only those four | `graphql/mutations/artifact_definition.py:61,88`; `git/tasks.py:503`; `core/merge/regeneration_dispatcher.py:43,192,228`; `api/artifact.py:108` |
| `nodes` matches artifact IDs, not target node IDs | `api/artifact.py:108-112`; `git/models.py:24,45-57` |
| The endpoint returns success when nothing matches | `api/artifact.py:68-135` — queues the job, returns nothing, validates no ID |
| `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE` false → regenerate everything | `config.py:219-222`; `regeneration_dispatcher.py:40-49` |
| Creating or updating a definition regenerates every current member | `graphql/mutations/artifact_definition.py:45-99` — no limit set, so the run covers the group |
| A repository import has the same effect | `git/integrator.py:700-791` — goes through that same mutation |
| `branch_scope` defaults to `default_branch`, three values | `actions/schema.py:57-65`; `actions/constants.py:30-48` |
| Branch scope matches the branch the event came from | `actions/models.py:114-118,211-215` |
| An artifact definition cannot be an event action | `actions/schema.py:151,179` — only Generator and Group actions exist |
| **Generate** runs a full pass, **Re-generate** a limited one | `artifact-generate-button.tsx:29,67`; `artifact-header.tsx:24-27`; `details-buttons.tsx:49` |
| `execute_after_merge` exists on the Generator definition | `core/schema/definitions/core/generator.py:52` |
| `artifact_generate` is a real Ansible module | `opsmill/infrahub-ansible`, `plugins/modules/artifact_generate.py` |
| `artifact_generate` is also a Python SDK method | `python_sdk/infrahub_sdk/node/node.py:1007` |
| No staleness indicator exists | `core/constants/__init__.py:167-171`; `core/schema/definitions/core/artifact.py:53` |

**Placement** — checked each page's position in `docs/sidebars.ts`, which Diátaxis quadrant each addition belongs to, and whether a Reference home exists for `branch_scope`. Read `webhooks/overview.mdx:45-70` for the duplicate treatment.

**Vocabulary** — counted every use of "pass" and "generation pass" in the new text and searched all of `docs/docs` for both: bare "pass" is 11 uses here, 0 anywhere else. Checked Proposed Change capitalisation across the docs.

**Sentences** — for all 65 new sentences and table cells, wrote down what the sentence names as performing the action and checked whether it can. Twelve can't; items 4, 17 and 21 are the ones worth fixing.

**Tools** — markdownlint on the five files at this branch's head: 0 errors. I couldn't run Vale or `invoke docs.lint` locally, so I read the rules in `.vale/styles/Infrahub/` and searched the added lines for what they catch — no hits. Your Validation notes report the build and lint passing; I didn't reproduce that.

</details>

---

### For the rewrite

Please run the `writing-infrahub-docs` skill over it rather than patching sentence by sentence. The rules that bite hardest here, in order:

- **`audit-rigor`** — text carried over from an existing page is a claim, not a given. Items 1 and 3 are both inherited sentences that were never re-read.
- **`voice-and-tone`, "Terms the product already has"** — grep the docs before using a word for a product concept. That one command settles the "generation pass" question.
- **`actors-and-causality`** — buttons, "four things" and "four kinds of change" can't perform the verbs given to them.
- **`overview-page-content`** — it names `generators/overview.mdx` as the caution, which is item 11.
- **`docs-navigation-map`** — one page owns a concept, the others link to it. Items 12, 13 and 14.
- **`explain-and-describe`** — the test is what the reader can do afterwards. Item 6 is the clearest case: the page tells them regeneration won't happen on its own but not how to find what's already stale.

Happy to pair on any of this, or to take the reorganization myself as offered above.

lancamat1:
Yes to the offer. The spoke is yours, item 11, and item 14 if you want it. The rest is either done or blocked, details below.

## Done and pushed (`6efb933e5`)

Items 1 to 8, 13, 15 to 17, and 20 to 25, plus the rename. The term is "artifact generation run" everywhere now, shortened to "run" after first use, for the reasons you traced. The two lists sit under their own headings. `use.mdx` has a "Regenerate an artifact" step before Validation, and the checklist item is something you look at again.

The em dash sentences are fixed with parentheses and a semicolon rather than by putting the em dashes back, since I do not write them.

I also added a `### How Generators are executed` heading. The execution list had no anchor, so the cross links I needed would have failed the build. It should make the split easier for you.

## Three corrections to the review

**The submitter list is wrong, in my description and in your item (b).** Repository import does not submit `REQUEST_ARTIFACT_DEFINITION_GENERATE`. It creates or updates the definition through the SDK (`git/integrator.py:752,759-780`) and reaches generation through the mutation. The proposed change pipeline does not submit it either, it uses `REQUEST_PROPOSED_CHANGE_REFRESH_ARTIFACTS` into `validate_artifacts_generation` into `GIT_REPOSITORIES_CHECK_ARTIFACT_CREATE` (`proposed_change/tasks.py:492,748,887`). The four real entry points are the create mutation (`graphql/mutations/artifact_definition.py:67`), the update mutation (`:94`), the REST endpoint (`api/artifact.py:122`), and the post merge dispatcher (`core/merge/regeneration_dispatcher.py:156-162`). Description corrected.

**Item 7 needs a narrower clause than "rebasing does nothing".** Rebase starts no artifact generation run: `core/branch/tasks.py:147-148` submits IPAM reconciliation, diff refresh and the recompute coordinator, and `BranchRebasedEvent` has no subscriber. But a rebase replays the diff as node mutations tagged `origin: "rebase"` (`branch/tasks.py:334-355`), and user trigger rules set no origin filter (`actions/models.py:105-118`), unlike the computed attribute triggers which pin `LIVE`. So a rebase can fire a Generator action. The callout says both.

**Item 16, `infrahubctl`.** The omission was real, but not as a queued path. `infrahubctl generator` runs the Generator in the CLI process (`ctl/generator.py:19-110`) and submits no workflow, and there is no `infrahubctl artifact` command at all. The callout names it without implying it queues anything.

## Item 8, answered

Closing a proposed change does nothing to artifacts, it only freezes the diffs (`graphql/mutations/proposed_change.py:205-211`). `CoreArtifact` is `BranchSupportType.LOCAL` (`core/schema/definitions/core/artifact.py:47`), so an artifact created on a branch is not visible on the default branch before the merge and is not merged into it. Deleting the branch hard deletes those nodes (`core/branch/data_deleter.py:159-179`). The rendered files are never deleted by anything: the only `registry.storage.delete` call in the backend is the failed upload rollback for `CoreFileObject` (`core/file_processor.py:248-251`), and there is no GC task. The page states this in two sentences and leaves the storage internals out.

## Item 6, one refinement if you extend it

`checksum` is an MD5 of the rendered output, and it is compared only after the query has run and the transform has executed (`git/integrator.py:2128,2130-2133`). Working out whether an artifact is stale therefore costs the same as regenerating it. Also `parameters` on `CoreArtifact` is declared but never written by any code path.

## Blocked

Items 10, 12, 18 and 19 are all `branch_scope`. I have another open PR, #10467, which documents the same field on the same page in a different shape, so one of the two has to own it. #10467 is waiting on an approval from Benoit. Once it lands I will rebase this branch and resolve the duplication.

On item 12 itself, the two treatments do not contradict each other. `webhooks/overview.mdx:46-75` owns the per event emission branch rules in UI labels and never states the default, this page has the API values and the default. It needs a cross link and a label to value mapping, not one page losing content.

Item 14 I will open as an issue unless you would rather keep it with the spoke.

One more thing: this PR now targets `stable` rather than `develop`, and the branch was rebased, so the diff is back to the five docs files

lancamat1:
#10467 merged, so this branch is rebased on `stable` and the `branch_scope` duplication is resolved (`86416e063`).

The Branch scope section is now the only place that lists the three values. The "Where trigger rules are stored and evaluated" section that came in with #10467 keeps the storage and reconciliation facts and links to it. Items 10, 12, 18 and 19 are done with it:

- each API value is paired with the label the form shows, since a reader in the UI never sees `other_branches`
- the sentence under the table states the consequence instead of repeating the rows, and links the cascade testing recommendation rather than restating it
- branch scope is now step 3 of both UI procedures, so following either one no longer silently accepts a default that cannot fire inside a branch
- the webhooks branch types section is linked for the emission branch per event class, which applies to trigger rules and is documented nowhere else

Item 14 is filed as #10630. Item 11, the spoke, is still yours whenever you get to it

PR #10443, opened 2026-08-28
Run the coalesced Python pass on merge and rebase

gmazoyer:
# Why

A merge or a rebase recomputes Jinja2 computed attributes, display labels and human-friendly ids
through one coalesced pass. Python transform computed attributes are excluded from it, so their
automations start one task per changed node: 68 tasks against 3 coalesced ones on a site merge,
148 on a rebase.

This is the second of three PRs ([IFC-3018](https://opsmill.atlassian.net/browse/IFC-3018)).
[#10416](https://github.com/opsmill/infrahub/pull/10416) added the resolver and called nothing.
This one runs it: the coalesced pass gains a fourth family. The per-node automations keep firing
alongside it, so work happens twice and nothing is missed. IFC-3019 then gates those automations on
`origin == live`, which is what removes the fan-out and makes the counts drop.

## What changed

* The resolver runs in the merge and rebase coordinator and in the chain submitter. Its targets
  reuse the two flows the submitter already knows: `process_transform` with `object_ids` for a
  scoped target, `trigger_update_python_computed_attributes` for a widened one.
* Both flows take `coalesced` and `recompute_depth`. A coalesced pass stamps its writes with the
  recompute origin and drives the next chain level, instead of re-entering the live per-node paths
  with no depth guard.
* `process_transform` recomputes the one attribute it is asked for. It ignored the name before, so
  a kind with two Python attributes ran both transforms on every submission.
* The read-set index is read after the local worker catches up on the schema. A worker behind on it
  declares no Python attribute at all, which reads exactly like a branch with nothing to refresh.
  That is also why the merge and rebase flows now need the worker component, and why the tests that
  call them directly build one whose heartbeat lands in their own cache.
* A schema-changing merge already refreshes what its own scope selects, one whole kind at a time.
  Those pairs are dropped on the coalesced side, and the same scoper decides it on both. The drop is
  the one place here that can under-recompute, so it is guarded twice: the scope reaches the resolver
  only once the schema-updated event has actually been sent, and a pair the gather never returned is
  never treated as covered, because the schema pass builds its candidates from the same gather.
* The testcontainers compose file forwards the switch to the server and the task worker, so an
  integration or performance run takes the same path as production.

**One fix outside the coalesced pass.** Both sides of that drop build their candidates from
`gather_trigger_computed_attribute_python`, which indexed them by transform name. One transform can
feed several computed attributes, so all but one were overwritten and never got an automation at
all: they were stale until something else touched them. The index now uses the key of the computed
attribute. This has its own changelog entry, and it costs one thing worth knowing:
[IFC-3083](https://opsmill.atlassian.net/browse/IFC-3083) tracks the query triggers it multiplies.

**Two things moved here from IFC-3019.** A review found that the rebase flow closes its database
session before this pass runs, so the derivation reached the index through a spent handle and raised
`SessionError`. The pass now opens a session of its own, as the merge path already does. The plan
had deferred this on the belief that the consequence was a visible raise; it is not, because the
failure is caught and the pass degrades quietly, so it could not turn CI red. The same review found
that a raise from this family cancelled the submissions of the other three, which need no database
at all; the derivation is now contained, and only the Python family is lost when it fails.

ADR 0017 also moved here. It is Accepted and still said this family stays outside the pass, so
`develop` would have carried an ADR contradicting its own code.

**The fallback does not widen, on purpose.** When the derivation fails the pass drops this family
and lets the per-node automations cover it, which holds only while they still fire on a merge and a
rebase. Widening to the whole kind instead would be the usual answer here, but it buys nothing
today, since the automations already do that work, and it would mean a whole-kind refresh per
attribute at the moment the database is unhealthy. So the obligation is pinned rather than paid:
`IFC-3019` is named in the docstring, and a test reads the trigger definitions and fails as soon as
the Python ones carry an origin match, which is the edit that removes the cover. Whoever lands that
gate has to make this fallback widen in the same change.

**One cost removed from the live path.** The chain that drives the next coalesced level was built
for every dispatch, including live single-node ones that never use it. It carries the Python
resolver, which builds an API client and fills its schema cache. The flows already know which kind
of pass they are running before they wire the dispatcher, so that choice moved to build time.

## How to review

With the switch on, is anything done twice, or is anything missed? Twice, never missed.

Three places carry the answer. `recompute_coalescing.py` holds the `PythonTargetResolver` seam the
resolution sits behind and the routing of a widened target to the fan-out flow.
`python_target_sources.py` is where the switch picks the real resolver or an inert one, and where
the pass waits for the schema before it reads the index. `computed_attribute/tasks.py` is the
narrowing and the two new flow parameters.

`_covered_by_schema_pass` in `python_target_resolution.py` deserves the closest look, since dropping
a pair is the only way this diff can leave a value stale. Everything else over-recomputes when it is
unsure. `dev/knowledge/backend/merge-recompute.md` records the one condition behind that drop which
nothing checks at run time.

The component test in `backend/tests/component/computed_attribute/test_coalesced_recompute_python.py`
runs the real sources: read sets from the analyzed transform queries, readers from real query
groups through the client. It pins the submission count and the exact node ids, since widening is
one way to over-recompute and over-selection is another.

## How to test

```bash
uv run pytest backend/tests/unit -q
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/merge_recompute_coalescing -q
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/core/merge -q
```

Green locally, with the switch on by default: 2479 unit tests and 40 in the computed-attribute
suite, which is the one that drives the real rebase flow. `uv run invoke lint` and
`uv run invoke docs.validate` are clean.

Every test that guards a silent failure here was run against a reverted fix and fails without it:
the schema-pass drop, the gather index, the chain arguments, the rebase session, the two that prove
a failed Python derivation still lets the other three families submit, and the origin-gate canary,
checked by gating the Python trigger and watching it go red.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry: for the shared-transform fix; the coalescing feature gets one on IFC-3019

[IFC-3018]: https://opsmill.atlassian.net/browse/IFC-3018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10443?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-3083]: https://opsmill.atlassian.net/browse/IFC-3083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dev/knowledge/backend/computed-attributes.md
```diff
@@ -39,7 +39,7 @@ These changes are handled by Prefect background tasks triggered by `NodeCreatedE
 
 **When**: A branch merge or rebase changes nodes that feed computed attributes.
 
-A merge or rebase does not emit one event and one flow per changed node. It runs a single coalesced recompute for the whole change set, writes the results in bulk, and chains any value that reads them. The three families (computed attributes, display labels, human-friendly ids) share this path, and the per-node triggers are suppressed for merge/rebase/recompute-origin events so the change is processed once. See [merge-recompute.md](merge-recompute.md).
+A merge or rebase runs a single coalesced recompute for the whole change set, writes the results in bulk, and chains any value that reads them. For computed attributes, display labels and human-friendly ids it replaces the per-node path entirely: their triggers are suppressed for merge/rebase/recompute-origin events, so the change is processed once. Python transform computed attributes join the pass when `INFRAHUB_COALESCE_PYTHON_RECOMPUTE_AFTER_MERGE` is on, which is the default, but their per-node automations keep firing, so that family is processed twice until those automations are gated. See [merge-recompute.md](merge-recompute.md).
```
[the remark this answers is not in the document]
gmazoyer: We will add a changelog once the overall work will be complete, this PR is just a step toward the end goal.

backend/infrahub/core/branch/tasks.py
```diff
@@ -338,6 +339,7 @@ async def rebase_branch(branch: str, context: InfrahubContext, send_events: bool
     for event in events:
         await event_service.send(event)
 
+    python_resolver = await build_python_target_resolver(db=db)
```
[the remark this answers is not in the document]
gmazoyer: The build does not read the database, it only resolves the worker component and builds a client. If it raises inside the guard, the coalesced pass is skipped for all four families and not only Python, with one log line as the only trace. The values are stale either way, so failing loudly does not make the data worse, it makes the gap visible. The flow has no retries, so a raise here cannot replay the rebase.

backend/tests/component/computed_attribute/test_merge_fanout_python.py
```diff
@@ -88,6 +94,10 @@ async def test_recompute_fans_out_to_every_node_of_the_kind(
             computed_attribute_name="computed_desc_python",
             computed_attribute_kind="TestCar",
             context=self._context(admin_account, default_branch),
+            coalesced=True,
+            recompute_depth=2,
         )
 
         assert self._fanned_out_ids(workflow_recorder) == car_ids
+        # Dropping either one leaves a widened target writing as live, so no chained level follows.
```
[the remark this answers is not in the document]
gmazoyer: You are right, the assertion only shows that the two values reach the submission. I removed the comment instead of rewording it, since the call passes both values a few lines above.

backend/tests/component/computed_attribute/test_coalesced_recompute_python.py
```diff
@@ -0,0 +1,433 @@
...
+        self,
+        ungathered_dataset: PythonRecomputeDataset,
+        db: InfrahubDatabase,
+        workflow_recorder: WorkflowRecorder,
+        default_branch: Branch,
+        admin_account: CoreAccount,
+    ) -> None:
+        """The backfill of a schema-changing merge covers the gathered pair, and only that one.
+
+        Both read sets are imprecise to the scoper, so it would select the ungathered pair too and
```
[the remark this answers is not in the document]
gmazoyer: Fair. I kept the first line and replaced the second one with the fact it rested on: both read sets are imprecise to the scoper, so the gathered flag is all that separates them.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -192,75 +211,76 @@ async def process_transform(
...
+    results: list[tuple[str, AttributeValueWrite | Exception]] = [
+        (oid, result) async for oid, result in batch.execute()
+    ]
+    writes, skipped = _partition_transform_results(results)
+    for skipped_id, reason in skipped:
+        log.warning(f"Skipping recompute of '{computed_attribute_name}' for node {skipped_id}: {reason}")
+
+    dispatcher = await build_bulk_recompute_dispatcher(schema_branch=schema_branch)
+    # The sibling flows read this off `object_ids`, which cannot work here: a live whole-kind
+    # refresh sends ids too, and labelling it as coalesced would stamp it with the wrong origin.
```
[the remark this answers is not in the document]
gmazoyer: Applied your suggestion. The rule is the part worth keeping, and it still says why, so nobody can reintroduce the inference by making the four call sites look alike.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -95,6 +96,21 @@ def _resolve_changed_elements(
...
+    Every caller submits one flow per attribute, so the flow must recompute that one only.
+    """
+    for attribute in node_schema.attributes:
+        if (
+            attribute.name == attribute_name
+            and attribute.computed_attribute
+            and attribute.computed_attribute.kind == ComputedAttributeKind.TRANSFORM_PYTHON
+        ):
+            return attribute.computed_attribute
+    return None
```
polmichel, line 111: This method is private and, as of now, only used in one place. But the logic looks like schema data manipulation. Wdyt about creating a dedicated public helper for it?

A idea of location I had in mind is `basenode_schema.py`. It gathers some getters around attribute like `get_attribute_or_none` which could be reused.

My only concern would be if someone else is creating the same private (or public) helper elsewhere without us noticing the duplication, and that we'd have to maintain those two helpers.

polmichel: Maybe is that another file would be a better fit

backend/infrahub/core/merge/post_merge.py
```diff
@@ -169,14 +175,23 @@ async def dispatch_events(
...
             coordinator = MergeRecomputeCoordinator(
                 builder=CoalescedRecomputeBuilder(schema_branch=schema_branch),
                 submitter=CoalescedRecomputeSubmitter(workflow=self.workflow),
+                python_resolver=self.python_resolver,
+            )
+            await coordinator.run(
+                changes=changes,
+                branch=self.default_branch.name,
+                context=event_context,
+                schema_changed_elements=schema_changed_elements,
```
polmichel, line 193: Is there an architectural or code reason why we do not run this whole post-merge process in the dedicated `infrahub.core.branch.tasks.post_process_branch_merge` flow which is called inside `infrahub.core.merge.post_merge.PostMergeDispatcher.run_follow_ups`?

One consequence I can see is that it makes our regeneration systems harder to synchronize if they are launched from different places. And maybe we'd want to run computed attributes before artifacts and generators.

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -110,39 +122,50 @@ def lookups(self) -> tuple[frozenset[str], ...]:
...
         subscriber_source: PythonSubscriberSource,
-        branch: str,
     ) -> None:
         self.read_set_source = read_set_source
         self.subscriber_source = subscriber_source
-        self.branch = branch
-        self._read_sets: list[PythonAttributeReadSet] | None = None
-        self._subscriber_cache: dict[frozenset[str], list[SubscriberRef]] = {}
+        self._read_sets: dict[str, list[PythonAttributeReadSet]] = {}
+        self._subscriber_cache: dict[tuple[str, frozenset[str]], list[SubscriberRef]] = {}
```
polmichel: This data collection would maybe be more readable with a tiny object or a typed key. In my opinion, it's kind of hard to know what's inside this data collection

backend/infrahub/core/merge/python_target_resolution.py
```diff
@@ -205,19 +239,88 @@ async def _build_target(self, *, accumulator: _Accumulator) -> AffectedTarget |
...
+
+def _covered_by_schema_pass(
+    *, read_sets: list[PythonAttributeReadSet], branch: str, changed_elements: ChangedElementSet
+) -> set[tuple[str, str]]:
+    """The (kind, attribute) pairs the schema-scoped backfill refreshes for this schema change.
+
+    Both sides run the same scoper over the same candidates, so what one selects is what the other
+    can drop. Only a gathered pair qualifies: the schema pass builds its candidates from the
+    transforms it could gather, so a pair it never gathered is a pair it never submits.
+    """
```
polmichel, line 287: The goal of this method is to mimic what happens in there right? `infrahub.computed_attribute.tasks.computed_attribute_setup_python`

Can we call `_covered_by_schema_pass` inside this existing method as we are sure these code will not change over time? We might need to change the name though

backend/infrahub/core/merge/python_target_sources.py
```diff
@@ -73,15 +94,16 @@ async def subscribers(self, *, node_ids: list[str], branch: str) -> list[Subscri
...
-    """Build the resolver for one merge or rebase pass, or None while the switch is off."""
+async def build_python_target_resolver(*, db: InfrahubDatabase) -> PythonTargetResolver:
+    """Build the resolver for one recompute pass, inert while the switch is off.
+
+    The switch is read first, so a deployment that leaves the family to the per-node automations
+    resolves neither the client nor the component.
+    """
     if not config.SETTINGS.main.coalesce_python_recompute_after_merge:
-        return None
+        return DisabledPythonTargetResolver()
```
polmichel, line 116: Nice polymorphism 👍

backend/tests/component/core/merge/test_post_merge.py
```diff
@@ -112,6 +166,28 @@ async def test_no_schema_updated_event_when_no_schema_change(
...
+        """
+        source_branch = await create_branch(branch_name="feature", db=db)
+
+        scope = await _schema_scope_handed_over(db, source_branch, default_branch, MemoryInfrahubEvent())
+        assert scope is not None
+        assert "display_labels" in scope.changed_fields.get("TestCar", frozenset())
+
+        failing = FailingInfrahubEvent(failing_kind=SchemaUpdatedEvent)
+        assert await _schema_scope_handed_over(db, source_branch, default_branch, failing) is None
+        assert [type(event).__name__ for event in failing.events] == ["BranchMergedEvent"]
```
polmichel, line 189: This one is key to avoid any under execution from the second loop of recomputation.

backend/tests/unit/computed_attribute/test_tasks.py
```diff
@@ -1,13 +1,44 @@
...
+    slogan.computed_attribute = ComputedAttribute(kind=ComputedAttributeKind.TRANSFORM_PYTHON, transform="TShirtSlogan")
+    node_schema.attributes.append(slogan)
+    return node_schema
+
+
+def test_only_the_requested_python_attribute_is_selected() -> None:
+    """A kind with two attributes would otherwise run its transform twice per submission."""
+    node_schema = _two_python_attributes()
+
+    pitch = _requested_python_transform_attribute(node_schema=node_schema, attribute_name="pitch")
```
polmichel, line 30: One element regarding unit testing is that testing private methods also induces that we are making private a behavior that could be integrated in a public component/service/method. This is kind of linked to the upper comment `https://github.com/opsmill/infrahub/pull/10443/changes#r3914089507`

backend/tests/unit/core/recompute/test_dispatch_convention.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_the_transform_flow_does_not_read_coalesced_off_the_node_ids() -> None:
+    """Its three sibling flows infer the flag from ``object_ids``, and this one must not.
+
+    A live refresh of a whole kind sends ids too, so inferring here would stamp those writes with
+    the recompute origin and drive a chain from them. This flow is told which pass it belongs to.
+    """
+    argument = _coalesced_argument(source=inspect.getsource(computed_attribute_tasks), function="process_transform")
+
+    assert "object_ids" not in ast.unparse(argument)
```
polmichel: This test looks complex, and at first glance it's hard to tell whether it properly protects against every possible case, for instance when `all_ids` is used.

In my opinion, an integration test would be a better fit here, since what we actually want to guard against is a regression in system behavior. Do you think that would be possible here?

backend/infrahub/core/merge/post_merge.py
polmichel: After reading the rest of the PR, this is maybe because we want the `SchemaUpdatedEvent` to be out?

polmichel (commented):
Some comments around code maintainability. 

I'm trying to get a clear picture of the whole recomputation chain after a merge. Our system seems to rely on events and manual triggered flows which are not executed at the same moment, so I may have to change where artifacts and generators are recomputed, so that they rely on up-to-date computed attributes.

Maybe is that we should refactor some parts of that code to have a unique process regarding objects regeneration

ogenstad (commented):
An observation when running this.

I loaded an old schema file.

```yaml
---
version: '1.0'
generics:
  - name: Generic   # Location
    namespace: Location
    description: "Generic hierarchical location"
    label: "Location"
    hierarchical: true
    icon: "mingcute:location-line"
    include_in_menu: true
    attributes:
      - name: name
        kind: Text
        unique: true
      - name: description
        kind: Text
        optional: true

nodes:
  # --------------------  Location (Continent, Country, Site, Rack)  --------------------
  - name: Continent
    namespace: Location
    description: "A continent on planet earth"
    icon: "jam:world"
    inherit_from:
      - "LocationGeneric"
    generate_profile: false
    default_filter: "name__value"
    order_by:
      - "name__value"
    display_labels:
      - "name__value"
    children: "LocationCountry"
    parent: ""
  - name: Country
    namespace: Location
    description: "A country within a continent..."
    icon: "gis:search-country"
    inherit_from:
      - "LocationGeneric"
    generate_profile: false
    default_filter: "name__value"
    order_by:
      - "name__value"
    display_labels:
      - "name__value"
    children: "LocationSite"
    parent: "LocationContinent"
    attributes:
      - name: country_details
        kind: Text
        description: "Country Details"
        computed_attribute:
          kind: TransformPython
          transform: CountryDetails
        read_only: true
        optional: true
  - name: Site
    namespace: Location
    description: "A site within a country!!"
    icon: "ri:building-line"
    inherit_from:
      - "LocationGeneric"
    default_filter: "name__value"
    order_by:
      - "name__value"
    display_labels:
      - "name__value"
    parent: "LocationCountry"
    attributes:
      - name: city
        kind: Text
        optional: true
      - name: address
        kind: Text
        optional: true
      - name: contact
        kind: Text
        optional: true
      - name: slug
        kind: Text
        description: "The slug"
        computed_attribute:
          kind: Jinja2
          jinja2_template: "Welcome to {{ name__value }}!"
        read_only: true
        optional: false
    relationships:
      - name: devices
        identifier: "site__devices"
        cardinality: many
        peer: InfraDevice
        optional: true
  - name: Device
    namespace: Infra
    label: "Device"
    icon: "mdi:server"
    display_labels:
      - name__value
    attributes:
      - name: device_type
        kind: Text
        unique: false
      - name: instance
        kind: Number
        unique: false
      - name: description
        kind: Text
        optional: true
      - name: name
        kind: Text
        description: "Device name"
        computed_attribute:
          kind: TransformPython
          transform: DeviceNameAttribute
        read_only: true
        optional: true
    relationships:
      - name: site
        peer: LocationSite
        optional: false
        cardinality: one
        kind: Component
```

One thing to note is that I just loaded the schema, create a branch and then go back to main and there create some nodes to have one of each. I did not load a repository that contains the two transform based computed attributes. Now I rebase the branch from main.

We have some log entries after that happens. First we get the typical warnings "The transform DeviceNameAttribute is assigned to a computed attribute but the transform could not be found in the database." I was expecting this however even with this in mind we reach the widening which doesn't take into account that the transforms doesn't exist in the database. This eventually leads to later errors in the logs when running the tasks.

```python
2026-09-03T12:40:39.720913Z [info     ] Schema converged after 0.20 seconds [infrahub.tasks]
2026-09-03T12:40:39.840606Z [warning  ] The transform DeviceNameAttribute is assigned to a computed attribute but the transform could not be found in the database. [prefect.task_runs]
2026-09-03T12:40:39.840787Z [warning  ] The transform CountryDetails is assigned to a computed attribute but the transform could not be found in the database. [prefect.task_runs]
2026-09-03T12:40:39.881904Z [info     ] Finished in state Completed()  [prefect.task_runs]
2026-09-03T12:40:39.882083Z [info     ] Widening the recompute of InfraDevice.name to its whole kind: the read set is undeterminable [infrahub]
2026-09-03T12:40:39.882141Z [info     ] Widening the recompute of LocationCountry.country_details to its whole kind: the read set is undeterminable [infrahub]
2026-09-03T12:40:39.882173Z [info     ] Coalesced Python recompute on branch branch1 selected ['InfraDevice.name=whole-kind', 'LocationCountry.country_details=whole-kind'], and left nothing to the schema pass [infrahub]
2026-09-03T12:40:39.979333Z [info     ] Worker 'InfrahubWorkerAsync ed1153b7-0139-4526-8b7d-520c5d4b6d7b' submitting flow run 'd950d931-3f12-40de-a643-38ef3783c1d1' [prefect.flow_runs.worker]
2026-09-03T12:40:39.979529Z [info     ] Worker 'InfrahubWorkerAsync ed1153b7-0139-4526-8b7d-520c5d4b6d7b' submitting flow run '37f0011d-de14-4b03-9b92-4ff346770ab2' [prefect.flow_runs.worker]
2026-09-03T12:40:39.979702Z [info     ] Worker 'InfrahubWorkerAsync ed1153b7-0139-4526-8b7d-520c5d4b6d7b' submitting flow run '30d05c26-b686-4544-a910-26015301c766' [prefect.flow_runs.worker]
2026-09-03T12:40:40.106105Z [info     ] Flow run 'affable-bullfrog' submitted to infrastructure [prefect.flow_runs.worker]
2026-09-03T12:40:40.108405Z [info     ] Flow run 'blue-worm' submitted to infrastructure [prefect.flow_runs.worker]
2026-09-03T12:40:40.109043Z [info     ] Flow run 'vague-moose' submitted to infrastructure [prefect.flow_runs.worker]
2026-09-03T12:40:40.182066Z [info     ] Beginning subflow run 'Process computed attribute for LocationSite.slug' for flow 'computed_attribute_process_jinja2' [prefect.flow_runs]
2026-09-03T12:40:40.182280Z [info     ] View at http://localhost:4200/runs/flow-run/d950d931-3f12-40de-a643-38ef3783c1d1 [prefect.flow_runs]
2026-09-03T12:40:40.189952Z [info     ] Beginning subflow run 'Recreate all diffs for branch branch1' for flow 'diff-refresh-all' [prefect.flow_runs]
2026-09-03T12:40:40.190120Z [info     ] View at http://localhost:4200/runs/flow-run/30d05c26-b686-4544-a910-26015301c766 [prefect.flow_runs]
2026-09-03T12:40:40.191616Z [info     ] Beginning subflow run 'Process display_labels for InfraDevice' for flow 'display-label-process-jinja2' [prefect.flow_runs]
2026-09-03T12:40:40.191746Z [info     ] View at http://localhost:4200/runs/flow-run/37f0011d-de14-4b03-9b92-4ff346770ab2 [prefect.flow_runs]
2026-09-03T12:40:40.554220Z [info     ] Finished in state Completed()  [prefect.flow_runs]
2026-09-03T12:40:40.742442Z [info     ] Finished in state Completed()  [prefect.flow_runs]
2026-09-03T12:40:41.001985Z [info     ] Finished in state Completed()  [prefect.flow_runs]
2026-09-03T12:40:41.046161Z [info     ] Schema converged after 0.20 seconds [infrahub.tasks]
2026-09-03T12:40:41.050301Z [warning  ] The transform DeviceNameAttribute is assigned to a computed attribute but the transform could not be found in the database. [prefect.task_runs]
2026-09-03T12:40:41.050560Z [warning  ] The transform CountryDetails is assigned to a computed attribute but the transform could not be found in the database. [prefect.task_runs]
2026-09-03T12:40:41.056611Z [info     ] Finished in state Completed()  [prefect.task_runs]
...
2026-09-03T12:40:45.733179Z [info     ] View at http://localhost:4200/runs/flow-run/d25ad7dd-8a71-4cb5-9cbb-e57c26f04ce4 [prefect.flow_runs]
2026-09-03T12:40:45.872733Z [error    ] Encountered exception during execution: ValueError("Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'") [prefect.flow_runs]
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 1998, in run_context
    yield self
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2060, in run_flow_async
    await engine.call_flow_fn()
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2012, in call_flow_fn
    result = await call_with_parameters(self.flow.fn, self.parameters)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/computed_attribute/tasks.py", line 232, in process_transform
    raise ValueError(
ValueError: Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'
2026-09-03T12:40:45.873674Z [error    ] Encountered exception during execution: ValueError("Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'") [prefect.flow_runs]
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 1998, in run_context
    yield self
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2060, in run_flow_async
    await engine.call_flow_fn()
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2012, in call_flow_fn
    result = await call_with_parameters(self.flow.fn, self.parameters)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/computed_attribute/tasks.py", line 232, in process_transform
    raise ValueError(
ValueError: Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'
2026-09-03T12:40:45.874374Z [error    ] Encountered exception during execution: ValueError("Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'") [prefect.flow_runs]
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 1998, in run_context
    yield self
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2060, in run_flow_async
    await engine.call_flow_fn()
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2012, in call_flow_fn
    result = await call_with_parameters(self.flow.fn, self.parameters)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/computed_attribute/tasks.py", line 232, in process_transform
    raise ValueError(
ValueError: Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'
2026-09-03T12:40:45.875171Z [error    ] Encountered exception during execution: ValueError("Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'") [prefect.flow_runs]
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 1998, in run_context
    yield self
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2060, in run_flow_async
    await engine.call_flow_fn()
  File "/Users/patrick/Code/opsmill/reviews/infrahub/.venv/lib/python3.12/site-packages/prefect/flow_engine.py", line 2012, in call_flow_fn
    result = await call_with_parameters(self.flow.fn, self.parameters)
  File "/Users/patrick/Code/opsmill/reviews/infrahub/backend/infrahub/computed_attribute/tasks.py", line 232, in process_transform
    raise ValueError(
ValueError: Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'
2026-09-03T12:40:46.063660Z [error    ] Finished in state Failed("Flow run encountered an exception: ValueError: Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'") [prefect.flow_runs]
2026-09-03T12:40:46.065591Z [info     ] Batch saved. count_nodes_remaining=0 [infrahub]
2026-09-03T12:40:46.065740Z [info     ] Updating field summary counts... [infrahub]
2026-09-03T12:40:46.074130Z [error    ] Finished in state Failed("Flow run encountered an exception: ValueError: Unable to fetch transform 'DeviceNameAttribute' for computed attribute 'name'") [prefect.flow_runs]
2026-09-03T12:40:46.088862Z [error    ] Finished in state Failed("Flow run encountered an exception: ValueError: Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'") [prefect.flow_runs]
2026-09-03T12:40:46.253315Z [error    ] Finished in state Failed("Flow run encountered an exception: ValueError: Unable to fetch transform 'CountryDetails' for computed attribute 'country_details'") [prefect.flow_runs]
2026-09-03T12:40:46.255442Z [info     ] Field summary counts updated.  [infrahub]
2026-09-03T12:40:46.255755Z [info     ] Updating node summary counts... [infrahub]

```

gmazoyer:
> An observation when running this.
> ...

The pass treated "no read set for this attribute" as "unknown read set" and widened, so it submitted a whole-kind refresh per attribute, and those flows then failed on the missing transform. That is not safe over-execution, it is work that cannot succeed.

It now separates the two cases. If the gather itself fails, nothing is known and every attribute still widens. If the gather works and a transform is simply not in the database, the attribute is left out of the pass. Nothing can compute it until the transform arrives, and the recompute that runs when the transform is created covers it then. The schema-scoped backfill already skipped it for the same reason, since both sides build their candidates from the same gather, so the two now agree.

There is a component test for your scenario: an attribute whose transform is absent from the database, asserting only the attribute with a real transform is selected. It fails with the old behaviour.

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: Added `_SubscriberQuery` to replace `tuple[str, frozenset[str]]`

backend/tests/unit/core/recompute/test_dispatch_convention.py
gmazoyer: Removed this entirely. I did not find it really valuable in the end (as we try to have proper typing).

backend/infrahub/core/merge/post_merge.py
gmazoyer: Yes, that is the reason. The pass only gets the schema scope if the `SchemaUpdatedEvent` was really sent, so it has to run after the send. It also needs the merge changelog, which `run_follow_ups` already has.

Your other point is fair and it is not solved today. Nothing orders the computed attributes against the artifacts and the generators, they just run in parallel (I believe), but that a larger topic for us to address, that's for sure.

backend/infrahub/computed_attribute/tasks.py
gmazoyer: I moved the lookup to `get_attribute_or_none`, so the helper no longer walks the attribute list itself. What is left is picking the Python transform computed attribute, and I would keep that here rather than in the schema, because the schema class would then need to know about `ComputedAttributeKind`. But maybe this is not what we should do. I do not have a strong opinion on this one so I can rework it if you prefer it in another way.

backend/infrahub/core/merge/python_target_resolution.py
gmazoyer: I added it to IFC-3019. That ticket already changes both of these flows, so it is a smaller step there than here.

backend/tests/unit/computed_attribute/test_tasks.py
gmazoyer: Fair point. Here the behaviour is picking which computed attribute a single flow run recomputes, which is internal to the flow. So I don't know how to test that one function without going through a much more complex test (reaching the same behaviour from outside needs a repository and a transform to execute, which would be a much longer (time-wise) test).

backend/infrahub/computed_attribute/tasks.py
polmichel: Good points
- call to `get_attribute_or_none` is good from code duplication and responsibilities perspective
- the fact that it creates a new dependency is something we should care about.

However, this helper has nothing specific to this flow, it just takes a node schema and an attribute name and returns the Python computed attribute if it exists. The problem it solves is generic, and we do not want to maintain several compute attribute helpers spread across different flows/other files if we refactor some pieces of it.
It sounds like it should be a public helper in one of the files inside the `backend/infrahub/computed_attribute` package or `backend/infrahub/core/schema` package.

An idea could be to add some method to `backend/infrahub/core/schema/schema_branch_computed/facade.py` like the following one and create the underlying helper.

```
    def get_python_transform_attribute(self, node_kind: str, name: str) -> ComputedAttribute | None:
        return self._python.get_transform_attribute(node_kind, name)
```

and directly use the method from the `schema_branch` object inside `infrahub.computed_attribute.tasks.process_transform`.

Wdyt? It'd allow us to centralize the compute attributes handling to this subpackage which already reads from core schemas files.

polmichel: Then the related test file would become a direct test of a public method.

gmazoyer: Good call, done. The two are equivalent: the registry only ever receives `TRANSFORM_PYTHON` attributes, and a non-optional one raises at process time, so the kind check was redundant.

Added `get_transform_attribute` on the registry and `get_python_transform_attribute` on the facade, and `process_transform` calls it directly. The test moved to `test_python_registry.py` and goes through a processed `SchemaBranch`, so the Jinja2 case stays real instead of asserting against a map I filled myself.

polmichel (dismissed):

polmichel (approved):
LGTM!
(Besides the CI state which is not yet green)

PR #10456, opened 2026-08-31
feat(backend): expose neo4j pool lifetime and liveness settings

fatih-acar:
## Summary

A query sent on a pooled bolt connection that a firewall, NAT gateway or load balancer silently evicted hangs until TCP retransmission gives up (often 15+ minutes), and once the whole pool is stale every request waits out the 60s acquisition timeout before failing. Operators running Infrahub behind such middleboxes can now bound pooled-connection age and ping idle connections so deployments recover instead of hanging.

## Key Changes

- New optional `INFRAHUB_DB_MAX_CONNECTION_LIFETIME` setting: pooled connections older than this are discarded and redialed. This is the reliable fix for silent drops — set it below the idle timeout of every middlebox between Infrahub and the database.
- New optional `INFRAHUB_DB_LIVENESS_CHECK_TIMEOUT` setting: connections idle longer than this are ping-tested before reuse, so a connection killed loudly (e.g. a database restart) is replaced transparently instead of failing one request.
- Both settings are unset by default and passed to the driver only when configured — the driver's own defaults keep applying, so existing deployments see no behavior change.
- `utilities/neo4j_idle_killer_repro.py`: a self-contained reproducer (TCP proxy that blackholes idle flows the way stateful middleboxes do) demonstrating the hang and the mitigation against a local Neo4j.

## Measured behavior (driver 6.2.0, per stale pooled connection)

| Failure mode | Driver defaults | `liveness_check_timeout` only | `max_connection_lifetime` < idle timeout |
|---|---|---|---|
| Silent eviction (NAT/LB/firewall) | request hangs until TCP gives up; extras hit acquisition timeout | all waiting requests fail once (ping consumes the acquisition deadline), pool then recovers | no failed requests |
| Loud kill (FIN/RST, e.g. DB restart) | one instant `ServiceUnavailable` | no failed requests | only helps if the connection outlived the lifetime |

The two settings are complementary; neither default is changed, so choosing them stays an operator decision.

## Documentation Updates

- `docs/docs/reference/configuration.mdx` regenerated with the two new settings.
- Changelog fragment `changelog/+neo4j-pool-idle-connections.added.md`.

## Test Plan

- `uv run pytest backend/tests/unit/config/test_config.py backend/tests/unit/database/` — 102 passed (new: defaults stay unset, env-var overrides, driver receives configured values, unset values are omitted from driver kwargs).
- Manual: `docker run --rm -p 7687:7687 -e NEO4J_AUTH=neo4j/secretpassword neo4j:2026.05.0-community` then `NEO4J_PASSWORD=secretpassword uv run python utilities/neo4j_idle_killer_repro.py` — reproduces the hang + `ConnectionAcquisitionTimeoutError` and the lifetime mitigation.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10456?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/database/__init__.py
```diff
@@ -560,6 +560,12 @@ async def get_db(retry: int = 0) -> AsyncDriver:
             )
         address_resolver = build_address_resolver(members=members, default_port=config.SETTINGS.database.port)
 
+    pool_timeout_kwargs: dict[str, int] = {}
+    if config.SETTINGS.database.max_connection_lifetime is not None:
+        pool_timeout_kwargs["max_connection_lifetime"] = config.SETTINGS.database.max_connection_lifetime
+    if config.SETTINGS.database.liveness_check_timeout is not None:
+        pool_timeout_kwargs["liveness_check_timeout"] = config.SETTINGS.database.liveness_check_timeout
+
```
gmazoyer: Small thing: `max_connection_lifetime` is an age, not a timeout, so the name is a bit off for half of what it holds. `pool_connection_kwargs` maybe. But this is nit picking really.

gmazoyer (approved):

backend/infrahub/database/__init__.py
fatih-acar: Fair point — renamed to `pool_connection_kwargs`, squashed into 78a3a5b51.

PR #10461, opened 2026-08-31
fix(backend): stop treating a .infrahub.yml edit as a whole-repository regeneration signal [#9643]

### LINKED ISSUE #9643: Artifact regeneration: any .infrahub.yml edit regenerates all artifacts in the repo (over-regeneration)
## Summary

`.infrahub.yml` is appended to **every** Transformation's dependency closure, so the artifact-regeneration trigger for it is a **content-agnostic whole-file match**: *any* edit to `.infrahub.yml` regenerates **all** artifacts of **all** artifact definitions in the repository — regardless of what actually changed.

This is safe (no missed regeneration) but defeats the point of INFP-409 (reducing needless artifact work) for a common class of edits. In a large environment with thousands of artifacts, a one-line manifest edit forces a full re-render storm.

## Examples that over-trigger

- Enabling object load, or changing the object/schema load location, in `.infrahub.yml` — has nothing to do with artifact rendering, yet regenerates every artifact.
- Adding/editing an unrelated `check_definitions` / `generator_definitions` / `queries` entry.
- A comment-only change (confirmed in QA: a comment regenerated every definition's artifacts).

## Possible improvement

Parse `.infrahub.yml` and scope the manifest trigger to the definitions whose stanza (transform/query/targets/parameters) actually changed, rather than treating any byte change as affecting every Transformation.

Found during QA of IFC-2541 / INFP-409. Related correctness bug: #9641.

ogenstad:
Closes #9643.

## Problem

`append_manifest_path` merged `.infrahub.yml` into **every** Transformation's and Generator's stored dependency closure. `transform_changed` intersects that closure with the repository's file diff, so any edit to the manifest matched every definition and the whole repository re-rendered. A comment-only change was enough, as was an object-load setting or an entry belonging to an unrelated definition. Safe (it over-regenerates, never under-regenerates), but it defeated the point of the closure narrowing for a common class of edit.

## What changed

**The manifest is no longer in any closure.** `append_manifest_path` and its module are gone. A manifest edit is now a regeneration signal only through the definition field it actually changed, which surfaces on the definition node rather than as a file change.

**`ClosurePathSelector` goes with it.** It existed to strip the manifest blob from the *hashed* closure, precisely because the manifest was merged in automatically and its bytes would otherwise churn every fingerprint on a comment-only edit. With nothing merging it in, the only way it reaches a closure is because someone named it in `watch.files` - and stripping an explicitly declared entry would be the bug rather than the safeguard. A declared `.infrahub.yml` now behaves like any other declared file: it lands in `dependencies`, its blob is hashed, and `transform_changed` matches it. That is the documented way to opt a single definition back into the old behaviour.

**`file_path` is folded into the Python-transform and Generator-definition fingerprints.** For Python transforms this is a prerequisite, not a bonus. The closure is a sorted set of paths, so it cannot say which of them is the entry point. A definition whose `watch` names its own directory carries every file in that directory, so repointing `file_path` at a sibling left the closure and every blob sha identical and the fingerprint did not move. An artifact definition sees its transform only through that fingerprint, and removing the manifest from the closure removes the backstop that was masking the gap, so the two have to land together.

That gap is **live today** for computed attributes, whose recompute trigger keys purely on `CoreTransformPython.fingerprint` with no closure backstop.

For Generators the fold is uniformity rather than necessity. `file_path` is an attribute on `CoreGeneratorDefinition`, so a repoint already surfaced through `definition_changed` before this PR; folding it keeps the completeness guard's rule the same for every kind.

**The fold reads the manifest-declared path, not a filesystem-derived one.** `get_python_transforms` took `file_path` as a parameter and the caller passed `extract_repo_file_information(...).relative_path_file`, which resolves the file's directory before making it worktree-relative and falls back to an **absolute** path when that comparison fails - reachable whenever the repositories directory is reached through a symlink. Canonicalized, that absolute path carries the commit sha, so every Python transform's fingerprint would churn on every commit, re-rendering every Python-backed artifact definition and recomputing every Python-transform-backed computed attribute on every import. The method now derives the value from the config it already receives, so the closure, the fingerprint, and the `CoreTransformPython.file_path` node attribute all read one source, the way generator definitions already did.

**A non-canonical `file_path` is rejected, not normalized.** Both fingerprint inputs carrying one validate it in `__post_init__`, mirroring what `ClosureResult` already does for closure entries. A construction site that skips `canonicalize_path` now fails loudly during import instead of producing a silently different digest, i.e. a spurious repository-wide regeneration wave with no test failure.

**An empty dependency closure is never trusted.** Dropping the manifest from every closure made `dependencies=[]` reachable, and `union_watch_files` set `complete=True` whenever `watch.files` was non-empty - even when every declared entry matched no tracked file *and* auto-detection found nothing (a Jinja2 entry template that escapes the worktree). An empty closure names no file, so nothing can ever intersect it: `transform_changed` short-circuits to `matched=False`, a permanent never-regenerate rather than the conservative fallback, and the fingerprint goes stable too, because a trusted closure with a watch declaration drops the commit id. Completeness now follows the merged list, so a declaration cannot close a list that stayed empty.

**Docs corrections.** The `watch.files` opt-in was documented with the `.infrahub.yml` spelling hard-coded, but `.infrahub.yaml` is accepted too and a `watch.files` entry matching no tracked file is only logged as a warning - so a `.yaml` repository following the text verbatim got an opt-in that silently did nothing. The list of manifest fields that do regenerate also omitted an artifact definition's `transformation`, which is folded and does regenerate when repointed.

## Why dropping the closure entry is safe

Every output-affecting manifest field is already folded into the definition fingerprint as a parsed scalar, so an edit that changes what a definition produces still moves the fingerprint and surfaces on the definition node:

| Manifest field | What fires |
| --- | --- |
| `artifact_definitions.*` (`parameters`, `content_type`, `artifact_name`, `targets`, `transformation`) | artifact-definition node changes -> `definition_changed`; also folded |
| `jinja2_transforms.template_path`, `.query` | folded -> transform fp -> artifact-def fp -> `fingerprint` element -> `definition_changed` |
| `python_transforms.class_name`, `.convert_query_response` | folded, same rollup |
| `python_transforms.file_path` | **folded by this PR** (was the one gap) |
| `*.watch.files` | expands into `dependencies` -> hashed closure changes |
| `queries` remapping | query text changes -> `query_changed`, and the query fingerprint |
| `generator_definitions.*` | every field is an attribute on `CoreGeneratorDefinition` -> `definition_changed` |
| `execute_in_proposed_change`, `execute_after_merge`, `timeout`, `description` | correctly excluded; read off the node, do not affect output |

`test_composition_completeness.py` makes that table permanent rather than a one-off review artifact, at both levels of the manifest:

- **Per entry.** It reads `model_fields` off the four SDK repository-config models and requires every field to be either folded (proven by a digest that moves) or excluded with a stated reason, so a new manifest *field* fails the test until someone classifies it. Verified load-bearing: deleting the `file_path` term makes it fail.
- **Per section.** It reads `model_fields` off `InfrahubRepositoryConfig` itself and requires every top-level section to be covered by a definition kind, folded elsewhere (`queries` through the rendered query text, `graphql_fragments` because fragment definitions are inlined into that text), or recorded as not an input. A wholly new *section* - the shape `graphql_fragments` had when it was added - is invisible to the per-entry checks, which only read the fields of the entry models they are given.

This PR deliberately keeps `query_changed`, `definition_changed`, and `transform_changed` all live as deciders. It narrows one input to `transform_changed`; it does not move where correctness lives.

## One-time regeneration wave

The composition change moves fingerprint **values**, so the first repository import after upgrading recomputes them for Python Transformations, Generator definitions, and Python-Transformation-backed artifact definitions. Pure Jinja2-backed artifacts are unaffected (`template_path` was already folded, and the manifest blob was already excluded from hashing).

- **Artifacts: compute-only.** The render checksum dedup means no new artifact versions, no object-store uploads, no data changes, and no downstream artifact-changed events where the rendered output is unchanged. Workers re-run once.
- **Generators: real work.** This is the half worth weighing. Generators mutate data and have no equivalent dedup, so every affected Generator re-runs once on its first post-upgrade pipeline. Idempotent for a well-behaved Generator, but not free.
- **Computed attributes: one recompute.** The `computed-attribute-python-transform-updated` trigger fires on any change to `CoreTransformPython.fingerprint`, so every computed attribute backed by a Python transform recomputes once when the new fingerprint is stamped on import.
- **Staggered, not a spike.** No migration and no forced re-import: it lands per-repo on that repo's next sync.

No schema change. `dependencies` stays a `List` attribute; only its contents change. Existing nodes keep `.infrahub.yml` in their stored closure until next import, which self-heals - any branch carrying a manifest edit carries a new commit by definition, so the import rewrites the closure before the pipeline evaluates it.

One node attribute does change value, for a narrow set of deployments: `CoreTransformPython.file_path` now comes from the manifest rather than the filesystem. The two strings are identical unless the worktree was reached through a symlink, where the old value was an absolute path that would have broken transform execution anyway.

## Verification

| Suite | Result |
| --- | --- |
| Backend unit | 2475 passed |
| Component (git) | 106 passed, 1 xfailed, 2 failed - both pre-existing and unrelated (fixture repo missing `branch01`/`branch02`/`clean-branch`) |
| Component (proposed change, computed attribute) | 81 passed |
| Integration (artifact regeneration) | 16 passed across `test_artifact_regen_e2e.py`, `test_artifact_regen_watch.py`, `test_merge_selective_regen.py` |
| `ruff check . --exclude python_sdk` | clean |
| `invoke backend.lint` (ruff, ty, mypy) | clean, 1644 source files |
| `invoke docs.validate` | no stale generated docs |
| `invoke docs.lint` | 0 errors (5 warnings pre-existing in untouched release notes) |

The integration run is the one that matters most: `test_manifest_edit_regenerates_nothing` in `test_artifact_regen_e2e.py` asserts the fix against closures built by the real integrator from a worktree, not by a hand-written fixture.

## Follow-ups (not in this PR)

- **`FingerprintComposer` registry trap.** The four per-task import methods each build a composer with a fresh registry. `compose_transformation` and `compose_generator_definition` route a registry miss through `_resolve_commit_term` and degrade safely (commit id folded in, over-regenerates). `compose_artifact_definition` has no commit fold and swallows the miss with `or ""`, hashing an empty upstream - so the fingerprint would be blind to the transform entirely. Latent today (only `apply_import_plan` is called), and armed for whoever next reaches for those entry points.
- **Computed attributes miss cross-repo query changes.** `computed_attribute/triggers.py` has no trigger on `CoreGraphQLQuery`, so a Python Transformation whose query lives in another repository never recomputes when that query changes. A current defect, independent of this PR.
- **`Jinja2TransformationFingerprintInput.template_path` is not under the canonical-path guard.** Its call site passes `str(Path(...))`, which is not canonicalized, so bringing it in line would change the Jinja2 digest for any repository spelling `template_path` with a leading slash and extend the upgrade wave to a kind this PR deliberately leaves untouched. Worth doing as its own change, with the changelog updated.
- **Single-value fingerprint gate + completeness canary.** Demote the three predicates to explanation-only. Blocked on the two items above, and on deciding how a fingerprint reaches across repositories.


backend/infrahub/git/fingerprint/composer.py
```diff
@@ -53,21 +51,19 @@ def fold_commit_id(
     return None
 
 
-class ClosurePathSelector:
```
gmazoyer, line 56: Nice cleanup

gmazoyer (approved):

Issue #10463, opened 2026-08-31
bug: CoreGroupTriggerRule with a CoreGroupAction raises UnboundLocalError and blocks configuration of all trigger rules


lancamat1:
### Component

API Server / GraphQL

### Infrahub version

1.11.0 (reproduced live; code identical on current develop)

### Current Behavior

`_from_group_trigger` in `backend/infrahub/actions/models.py:217-243` only handles `CoreGeneratorAction`. When a `CoreGroupTriggerRule` is attached to a `CoreGroupAction`, the `workflow` variable is never assigned and the function raises `UnboundLocalError` inside `configure_action_rules`. Because that job recompiles every trigger rule in one pass, the crash aborts the whole pass: from that point on no new or edited trigger rule in the instance is turned into an automation until the offending rule is deleted. Existing automations keep running, which makes the failure hard to spot.

### Expected Behavior

Either the combination works (a group membership change triggers a group action), or it is rejected with a validation error when the rule is created. One invalid rule should not stop configuration of every other trigger rule.

### Steps to Reproduce

1. Create a `CoreGroupAction` (any group, `add_member`).
2. Create a `CoreGroupTriggerRule` with that action.
3. `configure_action_rules` fails with `UnboundLocalError: cannot access local variable 'workflow'` and every subsequent trigger rule create or update is silently not applied.
4. Delete the rule; configuration recovers immediately.

### Additional Information

Reproduced end to end on a live 1.11.0 instance via infrahub-testcontainers.

PR #10467, opened 2026-08-31
docs: generator/artifact ordering in proposed changes and trigger rule caveats

lancamat1:
# Why

In a proposed change, Generator runs and artifact generation are dispatched concurrently and artifact selection uses the diff captured at pipeline start, so an artifact that reads Generator output can render before the data exists. The pattern that produces correct artifacts at review time (event-driven Generator execution on the branch) is not documented anywhere, and neither is the fact that a trigger rule created on a branch does not fire.

This PR documents the current behaviour and the working pattern. It does not change any behaviour.

## What changed

- `proposed-changes/overview`: new section "Ordering between Generators and artifacts" stating the concurrency, the snapshot-based artifact selection, and the event-driven pattern.
- `artifacts/overview`: caveat in "When artifacts regenerate" that Generator-created data in the same proposed change is not a reliable trigger.
- `generators/overview`: new section "Running a Generator from events" with the node-trigger + group-action into group-trigger + Generator-action chain, the target group membership condition and the error raised when it is not met, the default-branch requirement, and `execute_in_proposed_change: false`.
- `events/event-rules`: the group trigger action constraint in the step where the action kind is chosen, a new section "Where trigger rules are stored and evaluated" (rules are read from the default branch only, `branch_scope` controls firing, and the warning logged when a rule is skipped), and a "Known limitations" entry for #10463.

Every claim was verified against the code on develop and a live 1.11.0 instance (chain fires end to end in 5 to 10 seconds, no lost events across an 8 node burst; the negative cases behave as documented).

## Review round

Four changes came out of review, each verified against the code before being applied:

- Cut the post-merge ordering paragraph. That sequencing is the selective post-merge path only, and the system falls back to unordered blanket regeneration when `selective_execution_after_merge` is off, when the diff summary is missing or unreadable, when selection fails, and when no Generator is impacted.
- Dropped the `execute_after_merge` recommendation. The stated reason held only for `execute_in_proposed_change`, and the SDK reads `execute_after_merge` on every run to pick the tracking group kind, so it is not inert on event-triggered runs either.
- Replaced "the definition must target the same group the rule watches" with the real condition: the Generator runs for the triggering objects that are members of the definition's target group, and the run fails with `Target <id> is not part of the group <id>` when none of them are.
- Made the group trigger action constraint restrictive, moved it into the step where the action kind is chosen, and recorded #10463 under "Known limitations".

One review item was not applied. The claim that a rule created on another branch does not fire is correct as written: the rule and action nodes are branch-agnostic, but the relationship edge to an aware peer is itself derived as aware (`core/schema/schema_branch.py:1958-1974`) and written on the branch the mutation ran on, so the rule is skipped even when its group and Generator definition already exist on the default branch. Details in the review thread.

## How to review

The four sections are short prose additions, no page moves, no sidebar changes. Vale, markdownlint-cli2 and the docusaurus build all pass locally.

Note that `invoke docs.lint` aborts from a clean checkout: `tasks/docs.py:148` looks for `markdownlint-cli2` on PATH while `package.json` installs it into `docs/node_modules/.bin`. Worth a separate fix.

## Checklist

- [ ] External docs updated (this PR)


BeArchiTek:
## Review notes on the four docs additions

I re-read each new paragraph with the code open alongside and traced every factual claim. Most of it holds; one claim attributes a real behaviour to the wrong cause, and three others are imprecise. `invoke docs.lint` is clean (0 errors; the 5 Vale warnings are all in files this PR doesn't touch), and all seven new cross-link anchors resolve.

### 1. The branch claim is right about the symptom, wrong about the cause

`events/event-rules.mdx:87` and the parenthetical at `generators/overview.mdx:288` both say a rule created on another branch does not fire, because rules are read from the default branch only.

The rule does not fire — that part is correct — but not for the stated reason. Trigger rules and actions are all `BranchSupportType.AGNOSTIC` (`backend/infrahub/actions/schema.py:29,91,132,161,189,230,274,325,373`), so the rule node is written to the global branch (`core/node/__init__.py:337`) and a default-branch read sees it whichever branch you created it on (`core/branch/models.py:289`). Where a rule was created is not what excludes it.

What actually excludes it is the peers. Both required relationships point at branch-**aware** kinds:

- `CoreGeneratorAction.generator` → `CoreGeneratorDefinition`, `optional=False` (`actions/schema.py:167`); that kind is `AWARE` (`core/schema/definitions/core/generator.py:26`)
- `CoreGroupTriggerRule.group` and `CoreGroupAction.group` → `GENERICGROUP`, `optional=False` (`actions/schema.py:390,207`); group kinds are `AWARE` (`core/schema/definitions/core/group.py:28,72,102`)

`gather_trigger_action_rules` reads at `registry.default_branch`, so a peer living only on another branch resolves to null, `_resolve_peer` raises `UnresolvedRelationshipError`, and the rule is skipped with a warning (`actions/parsers.py`). That exception's docstring names this exact case: *"a null peer means the referenced node is not visible on the branch being read (deleted, orphaned, or living only on another branch). The rule cannot be reconciled and is skipped."*

The distinction is worth the words, for two reasons. A rule created on a branch whose group and Generator definition already exist on the default branch **is** gathered and **does** fire, so the current text over-promises a failure. And the symptom is a log warning naming the unresolved relationship — something the reader can go look for, which the current text gives them no way to know about.

Suggested replacement for the section at `event-rules.mdx:85`:

> ## How trigger rules relate to branches
>
> Infrahub reconciles trigger rules by reading them from the default branch. Rules and actions are branch-agnostic — one rule exists across every branch, whichever branch you created it on — but the group and Generator definition a rule points to are branch-aware. When either peer exists only on another branch, the rule cannot be resolved from the default branch and is skipped; the task log records a warning naming the unresolved relationship.
>
> Create rules together with their target groups and Generator definitions on the default branch, then use `branch_scope` to control which branches' events they fire on:
>
> - `default_branch` (the default): …

The practical guidance ("create rules on the default branch") is unchanged and correct — only the reason and the heading need reworking. The parenthetical at `generators/overview.mdx:288` repeats the same reason and should follow.

### 2. `generators/overview.mdx:290` — the stated reason covers only one of the two flags

> Set `execute_in_proposed_change: false` and `execute_after_merge: false` on the Generator. The two flags do not apply to event-triggered runs, so leaving them `true` runs the Generator both from the event and in the pipeline.

`execute_after_merge: true` does not cause a pipeline run. It is read only at `core/merge/selective_regen/definition_selector/generator_selector.py:43` and `generators/tasks.py:155`, both on the post-merge path — where ordering *is* enforced, per the section this PR adds to `proposed-changes/overview`. So the sentence justifies the first flag and leaves the second unexplained.

Worth flagging too: combined with `branch_scope: other_branches`, `execute_after_merge: false` means the Generator never runs for data created directly on the default branch. Either give each flag its own reason or drop the `execute_after_merge` recommendation.

The claim that the flags don't gate event-triggered runs is itself correct — the gate exists only in `generators/tasks.py:153-156` keyed on `GeneratorDefinitionRunSource`, and the event path (`actions/tasks.py::_run_generators`) passes both values through without consulting them.

### 3. `generators/overview.mdx:288` — "must target the same group the rule watches" is stricter than the check

The query is `get_generator_run_query(definition_id=…, target_ids=node_ids)`, and the guard is membership of the triggering node in *the definition's* target group — as the docstring puts it, "if … none of the requested targets are members of the target group". Two different groups both containing the node would pass. Naming the real condition also matches the error the user actually sees: `Target <id> is not part of the group <id>`.

### 4. `event-rules.mdx:83` — this line is the only guardrail, so it should be restrictive and it should move

Nothing rejects the group-trigger + group-action combination: no `GroupTriggerRule` references in `graphql/mutations/`, and `test_action.py` covers node-trigger failure states only. The API and UI accept it, and `_from_group_trigger` then leaves `workflow` unbound, so reconciliation hits an `UnboundLocalError`. Two consequences:

- "A group trigger rule fires a Generator action" reads descriptive where it needs to be restrictive — "can only fire a Generator action". As written it also echoes the section's own opening ("A group trigger rule fires when …"), two senses of "fires" two paragraphs apart.
- It sits below the screenshot and the "After saving…" line, but step 4 of the procedure is where the reader picks the action kind. A reader following the steps chooses before reading the constraint. It belongs in or immediately above that step.

Separately: the unhandled combination looks worth a backend issue, since the docs line is currently the only thing preventing it.

### 5. `generators/overview.mdx:283` and `:288` — two overloaded sentences

`:283` carries condition, instruction, reason and cross-link in one sentence, and in "instead of in its pipeline, where the two run concurrently" the relative clause's antecedent is ambiguous. `:288` stacks three facts behind a colon following a semicolon-joined clause. Both want splitting.

### 6. Minor

- `proposed-changes/overview.mdx:126` "To review correct artifacts" is vague; `artifacts/overview.mdx:96` phrases the same instruction better as "To review artifacts built from Generator output". Worth aligning on the second.
- `artifacts/overview.mdx:96` — "during a proposed change … within that same proposed change" is redundant, and the sentence restates the mechanism the proposed-changes section owns. Consequence plus link would be enough.
- `generators/overview.mdx:281` — "Running a Generator from events" is an `##` immediately after `### execute_after_merge`, yet its closing paragraph is entirely about those two flags. `###` under `## Execution lifecycle` would read more coherently; `##` buys top-level TOC presence. Either is defensible.
- This PR description names the section "Where rules live and where they fire", renamed in 4e1dc40 to "Where trigger rules are stored and evaluated". Also "verified agains the code" → "against".

### What checks out

| Claim | Source |
|---|---|
| Generators and artifacts are dispatched concurrently in a proposed change, neither waits | `proposed_change/tasks.py:1243-1253` submits `run_generators(refresh_artifacts=True)` for `CheckType.ALL`; `run_generators` (`:406`) submits the definition checks and then the artifact refresh (`:483-491`) without awaiting either. The adapter has both a waiting and a non-waiting call, and this is the non-waiting one: `submit_workflow` → `run_deployment(timeout=0)` (`services/adapters/workflow/worker.py:113-115`) |
| Artifact selection uses the diff captured at pipeline start | `refresh_artifacts` (`:1326`) selects against `get_diff_summary_cache(pipeline_id=model.branch_diff.pipeline_id)` plus `model.branch_diff` |
| After a merge the order is enforced | `build_merge_selective_regeneration`: "the generator participant runs before the artifact participant so the plan awaits generator output before the artifacts that may read it are selected"; `reselect_from_cascade_output` re-selects what read that output |
| A group trigger rule only fires a Generator action | `actions/models.py::_from_group_trigger` handles `CoreGeneratorAction` only, unlike `_from_node_trigger` which handles both |
| The three `branch_scope` bullets, including `default_branch` as the default | `actions/constants.py:30-48`; default at `actions/schema.py:61`; match semantics at `actions/models.py:114,211` |
| Lifecycle flags do not gate event-triggered runs | gate only in `generators/tasks.py:153-156`, keyed on run source |

No changelog fragment needed — the last four docs-only commits on `develop` carry none.

docs/docs/proposed-changes/overview.mdx
```diff
@@ -117,6 +117,14 @@ Generators are Python-only, and Infrahub does not analyze imports, so auto-detec
 
 Learn more about [Generators](../generators/overview).
 
+### Ordering between Generators and artifacts
+
+Within a proposed change, Generator runs and artifact generation start at the same time, and neither waits for the other. The pipeline also selects which artifact definitions to process from the branch diff as it was when the pipeline started. An artifact whose Transformation reads data a Generator creates during the same proposed change can therefore render before that data exists, and which of the two finishes first varies from run to run.
+
+After a merge the order is enforced: Infrahub waits for each Generator to finish, then regenerates the artifacts built from the objects that Generator created or changed. Only the pipeline inside the proposed change runs both concurrently.
```
polmichel: For now, this is tied to a feature flag in `backend/infrahub/config.py`→ `selective_execution_after_merge`.

The flag's purpose is scope, not sequencing:
- OFF → regenerate everything after the merge (`submit_full_regeneration` fires both the generator-trigger and artifact-trigger, fire-and-forget, no ordering at all).
- ON → regenerate only what the merged nodes impacted. And to know which artifacts a generator's output impacts, the system has to run those generators first and scan their outputs, which is why it awaits them.

So the await/ordering is a means to the selective end (you can't select the impacted artifacts until you've seen what the generators produced), not the feature's goal. Sequencing is incidental.

This flag may be removed in 1.12 or shortly after, and it defaults to true, so I am not sure it's worth mentioning.

```
selective_execution_after_merge: bool = Field(
        default=True,
        description="When enabled, only the generators and artifact definitions affected by a merge "
        "are re-executed; when disabled, every generator and artifact definition is re-executed.",
    )
```

polmichel: Nice recap from Claude

**Quick clarification on what these flags actually control** (they're participation switches, not sequencing controls):

  | Flag | Applies to | Default | Controls |
  |---|---|---|---|
  | `execute_in_proposed_change` | the Generator | `true` | Whether the Generator runs inside the proposed-change pipeline (pre-merge, `PROPOSED_CHANGE` source). It runs there **concurrently** with artifacts, so it adds no ordering. |
  | `execute_after_merge` | the Generator | `true` | Whether the Generator is considered on the post-merge path (`MERGE` source). This is the only gate for post-merge selection. |
  | `selective_execution_after_merge` | global config | `true` | The **scope** of post-merge regeneration: selective (only what was impacted, by running the Generators and scanning their output to pick the artifacts) vs full (regenerate everything, concurrently). |

  Takeaways:

  - The two Generator flags are independent on/off switches for two independent sources (`PROPOSED_CHANGE` vs `MERGE`). Neither is a sequencing flag.
  - Ordering exists in exactly one place: the **selective post-merge path**, and only as a side effect of impact scanning (you must run the Generators before you can tell which
  artifacts their output affects). It is driven by `execute_after_merge` + `selective_execution_after_merge`, never by `execute_in_proposed_change`.

lancamat1:
@BeArchiTek Thanks for going through it. I took most of it, one item I do not think holds.

**1. The branch claim.** The peers half is right, the conclusion is not. The rule and action nodes are AGNOSTIC as you say, but the relationship edge from an agnostic node to an aware peer is itself derived as AWARE in `process_branch_support` (`core/schema/schema_branch.py:1958-1974`), and the relationship manager writes it on the branch the mutation ran on (`core/node/__init__.py:974-978`, `core/relationship/model.py:900,218-229`). So the case you built the argument on, a rule created on a branch whose group and Generator definition already exist on main, is still skipped: the rule and action nodes are visible from main, their edges are not.

So "a rule created on another branch does not fire" is correct as written, and the suggested replacement ("one rule exists across every branch, whichever branch you created it on") would put a wrong claim on the page.

What I did take is the symptom. The section now names the warning the task worker logs, `Skipping trigger rule with an unresolved relationship`, with the rule and the relationship it could not resolve, and says that creating the rule reports no error.

Worth a separate backend fix: the docstring at `core/relationship/model.py:1155` says an aware relationship on an agnostic source returns the global branch, which contradicts the code two lines below it. That is likely what makes this look the other way around.

**2. `execute_after_merge`.** Dropped from the recommendation. The reason only covered the first flag, and it is worse than that: the SDK reads `execute_after_merge` on every run to pick the tracking group kind (`python_sdk/infrahub_sdk/generator.py:86`, `CoreGeneratorGroup` is LOCAL, `CoreGeneratorAwareGroup` is AWARE), so it is not inert on event-triggered runs either. The page now sets only `execute_in_proposed_change: false`.

**3. The group condition.** Taken. It now says the Generator runs only for the triggering objects that are members of the definition's target group, and gives the actual error, `Target <id> is not part of the group <id>`.

**4. The group trigger constraint.** Taken, both parts. It is restrictive now and it sits in the action kind step where the reader chooses. #10463 is back on the page, under `## Known limitations`, following the convention in `generators/overview` and `schema/migration`.

**5 and 6.** Split both sentences, aligned on "To review artifacts built from Generator output", cut the restated mechanism in `artifacts/overview`. Kept the `##` heading level for TOC presence. PR description fixed.

Two things I left alone, both outside the scope of this PR. `proposed change` is lowercase 22 times to 1 on `proposed-changes/overview` and the style guide says to capitalise it, so fixing it is a pass over four pages rather than my sentences only. And `invoke docs.lint` cannot run from a clean checkout: `tasks/docs.py:148` looks for `markdownlint-cli2` on PATH while `package.json` installs it into `docs/node_modules/.bin`.

docs/docs/proposed-changes/overview.mdx
lancamat1: Cut the paragraph. You are right that it should not be in there, though for a slightly different reason than the flag: the sequencing itself is deliberate and commented as such (`selective_regen/models.py:26-33`, `orchestrator.py:144-146`), it is not a side effect of impact scanning.

What makes it wrong to state as a guarantee is how many ways it unwinds. With `selective_execution_after_merge` off it falls back to `submit_full_regeneration`, which submits artifacts first and awaits nothing (`regeneration_dispatcher.py:38-47`). The same fallback fires with the flag left on for `NO_SUMMARY_CAPTURED`, `SUMMARY_UNAVAILABLE` and `SELECTION_FAILED` (`:82-104`). And on the selective path, when no Generator is impacted, nothing is awaited either (`:113-127`).

The proposed change section does not need the contrast to make its point, so it is gone rather than hedged.

BeArchiTek (approved):
Approving. Re-reviewed the current head against the code; the four items from my first round are applied and I have no blocking findings left.

**Re-verified against `dd4a33350`**

- Both error strings are verbatim: `Skipping trigger rule with an unresolved relationship` (`actions/parsers.py:45`) and `Target <id> is not part of the group <id>` (`actions/tasks.py:278`).
- `branch_scope` — three choices and `default_branch` as the default (`actions/constants.py:30-48`, `actions/schema.py:61`).
- `execute_in_proposed_change: false` does not gate event-triggered runs — the gate is keyed on `GeneratorDefinitionRunSource` (`generators/tasks.py:165`).
- Anchor `#where-trigger-rules-are-stored-and-evaluated` matches the heading and the inbound link in `generators/overview.mdx:290`; the docs build passes with `onBrokenAnchors: "throw"`.

**On the branch claim — you were right, I was wrong.** I re-derived it: `process_branch_support` (`core/schema/schema_branch.py:1958-1974`) marks an agnostic→aware relationship AWARE, and `get_branch_based_on_support_type` returns the global branch only when the *relationship schema* is agnostic, so the edge is written on the mutation's branch. My counter-example does not hold, and the suggested replacement I offered would have put a wrong claim on the page. Good catch on the `core/relationship/model.py:1155` docstring contradicting the two lines below it — that is worth its own issue.

**Two non-blocking items, take or leave**

1. `event-rules.mdx:75` — cubic re-flagged "can only fire a Generator action" (P3) and I think it has a point, though not the one it makes. The restrictive wording is the right call, since nothing rejects the combination. But "can only" reads as *the system will not let me*, so a reader who reaches the #10463 entry three paragraphs down has no reason to connect the two — and the reader who hits that bug is exactly the one who did not believe line 75. `A group trigger rule supports only a Generator action; selecting a group action is accepted and then stops Infrahub from applying any trigger rule change (see Known limitations)` keeps the guardrail and closes the seam.
2. `proposed-changes/overview.mdx:122` / `generators/overview.mdx:283` — "start at the same time" is the one place the mechanism is stated more strongly than the code supports. `run_generators` submits the definition checks, then submits the artifact refresh from inside the same flow (`proposed_change/tasks.py:472`, `:478-490`). Concurrent, yes; simultaneous, no. "Neither waits for the other" and "which of the two finishes first varies from run to run" are both exactly right and carry the reader anyway.

cubic's P2 on `proposed-changes/overview.mdx:117` is stale — the description was updated after it ran and no longer claims the post-merge contrast.

PR #10469, opened 2026-08-31
fix(regeneration): treat display_label / human_friendly_id reads as imprecise (IFC-3071)

### LINKED ISSUE #10471: Selective regeneration under-narrows reads of display_label / human_friendly_id
## Summary

When incremental (selective) regeneration is used, an artifact or generator that displays an object's **display label** or **human-friendly ID (HFID)** could keep showing the old value after a merge. If you changed a field that the label or HFID is built from — for example, renaming an object — the artifact/generator that shows it was not regenerated and stayed stale until some later change forced a full regeneration.

## Who is affected

Deployments that rely on selective regeneration of artifacts or generators whose query displays an object's display label or HFID.
Release: 1.11.0 / 1.11.1

## Steps to reproduce

1. `INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE` is `True` (default value)
2. Have an artifact or generator whose output shows an object's display label or HFID.
3. On a branch, change a field that composes that label/HFID (e.g. rename the object).
4. Merge the branch.
**Expected:** the artifact/generator is regenerated and shows the new label/HFID.
**Actual (before the fix):** it still shows the previous value — it was skipped during regeneration.

## Resolution

Selective regeneration now recognizes that changing a field behind a computed display label or HFID affects the artifacts and generators that display it, and regenerates them. When a label or HFID is composed from a related object's field, regeneration is widened conservatively so that nothing is left showing stale data.

A finer resolution will be delivered in a future release.

polmichel:
Closes #10471

## Summary

Selective regeneration (run on merge and proposed-change follow-up) decides which generators/artifacts to regenerate by intersecting each changed field against the fields a definition's query reads, **by exact name**. `display_label` and `human_friendly_id` are computed and never appear as diff elements, so a change to a backing field that moves them was dropped and the reader left stale after a merge — under-execution, which the package invariant forbids. Ref: [IFC-3071](https://opsmill.atlassian.net/browse/IFC-3071).

The selective path is on by default (`INFRAHUB_SELECTIVE_EXECUTION_AFTER_MERGE=true`), so this bug is exposed in default deployments; disabling the flag re-executes everything and masks it.

## Fix (two parts)

1. **Imprecise read** (`predicates.py`) — a read set holding a computed field (`display_label` / `human_friendly_id`) is imprecise for its kind: any change to that kind counts. Purely additive widening; reuses the existing `IMPRECISE_READ_FIELDS`.

2. **Peer-composed labels** (`impact.py`) — a label/hfid can be composed from a **related node's** field (e.g. `hfid = parent__name`); a change to that peer moves the value but the peer's kind is never in the read set. When a read of such a derived field is **not scopable** to the reading kind's own attributes (checked via `derived_read_is_scopable`), the query can no longer be narrowed and widens to every target. Scopable reads (value comes only from the kind's own attributes) keep narrowing precisely.

This closes the residual under-execution safely on the released branch. It resolves the Cubic P2 review comment — the scopability guard lives in `impact.py` (which holds the schema) rather than the pure `predicates.py`.

## Scope of what regenerates

| Read | Change | Result |
| --- | --- | --- |
| `Device.display_label` (= own `name`) | `name` on dev1 | narrows to dev1 |
| `Device.display_label` (= own `name`) | `name` on dev1 + dev2 | narrows to each |
| `Interface.display_label` read via traversal | change on an interface | widens to every target |
| label/hfid composed from a peer (non-scopable) | any relevant change | widens to every target |

## Follow-up

A precise resolution of peer-composed labels — walking the derived-field path back to the owning member via the dependent-node resolver instead of widening — is tracked for **1.12** (develop only; the reached-path machinery does not exist on stable).

## Tests

- Unit — `test_impact_classifier.py` (imprecise-read + `depends_on_everything`) and `test_impact.py` (`reads_unscopable_derived_field` scopable vs peer-composed).
- Component (end-to-end, real DB + schema + server) — `test_impact.py::TestFieldLevelImpact::test_display_label_backing_change_selects_subscriber`.

## How to test

The component test reproduces the bug against a real Infrahub stack: a query that reads a car's `display_label` and a diff that changes the backing `name`. It runs from `backend/`:

```bash
cd backend && INFRAHUB_TESTING_TASKMGR_BACKGROUND_SVC_REPLICAS=1 \
  uv run pytest tests/component/core/regeneration/test_impact.py -v
```

Verified end-to-end: with the fix the subscriber is selected — `TargetSelection(ids=[<subscriber>], widened=False)`; reverting the imprecise-read handling in `predicates.py` drops it back to `ids=[]` (the stale-artifact bug), so the test genuinely reproduces the regression.

[IFC-3071]: https://opsmill.atlassian.net/browse/IFC-3071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/infrahub/core/regeneration/impact.py
+++ b/backend/infrahub/core/regeneration/impact.py
@@ -53,12 +53,12 @@ def reads_unscopable_derived_field(
     readable_fields_by_kind: Mapping[str, set[str]],
     get_node_schema: Callable[[str], MainSchemaTypes | None],
 ) -> bool:
-    """Whether the query reads a display_label / human_friendly_id composed from a related peer.
+    """Whether the query reads a computed field whose value is composed from a related peer.
 
     Such a read cannot be narrowed: the change that moves the value lands on a peer the read set
     never names, so the field-level match cannot see it, and the query has to fall back to every
-    target. A read of a derived field with no declared path, or on a kind absent from the schema,
-    is treated the same way -- conservatively, so scopability that cannot be verified widens rather
+    target. A read whose value has no declared path, or on a kind absent from the schema, is
+    treated the same way -- conservatively, so scopability that cannot be verified widens rather
     than raises.
     """
     for kind, fields in readable_fields_by_kind.items():
--- a/backend/infrahub/core/regeneration/impact.py
+++ b/backend/infrahub/core/regeneration/impact.py
@@ -21,12 +21,12 @@ from .models import TargetSelection
 log = get_logger()
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Mapping
+    from collections.abc import Mapping
 
     from infrahub_sdk.client import InfrahubClient
     from infrahub_sdk.diff import NodeDiff
 
-    from infrahub.core.schema import MainSchemaTypes
+    from infrahub.core.schema.schema_branch import SchemaBranch
 
 
 GATHER_GRAPHQL_QUERY_SUBSCRIBERS = """
@@ -51,7 +51,7 @@ query GatherGraphQLQuerySubscribers($members: [ID!]) {
 
 def reads_unscopable_derived_field(
     readable_fields_by_kind: Mapping[str, set[str]],
-    get_node_schema: Callable[[str], MainSchemaTypes | None],
+    schema_branch: SchemaBranch,
 ) -> bool:
     """Whether the query reads a computed field whose value is composed from a related peer.
 
@@ -65,8 +65,10 @@ def reads_unscopable_derived_field(
         derived_reads = fields & IMPRECISE_READ_FIELDS
         if not derived_reads:
             continue
-        node_schema = get_node_schema(kind)
-        if node_schema is None or any(
+        if not schema_branch.has(name=kind):
+            return True
+        node_schema = schema_branch.get(name=kind, duplicate=False)
+        if any(
             not derived_read_is_scopable(node_schema=node_schema, field_name=field_name) for field_name in derived_reads
         ):
             return True
@@ -115,9 +117,7 @@ async def get_field_level_impacted_subscribers(
         readable_fields_by_kind=readable_fields_by_kind,
         depends_on_everything=reads_unscopable_derived_field(
             readable_fields_by_kind=readable_fields_by_kind,
-            get_node_schema=lambda kind: (
-                query_schema_branch.get(name=kind, duplicate=False) if query_schema_branch.has(name=kind) else None
-            ),
+            schema_branch=query_schema_branch,
         ),
     )
     assessment = classifier.assess(diff_summary=diff_summary)
--- a/backend/infrahub/core/regeneration/impact.py
+++ b/backend/infrahub/core/regeneration/impact.py
@@ -4,6 +4,10 @@ from typing import TYPE_CHECKING, assert_never
 
 from infrahub.core import registry
 from infrahub.core.constants import InfrahubKind
+from infrahub.core.schema.schema_branch_computed.python_transform import (
+    IMPRECISE_READ_FIELDS,
+    derived_read_is_scopable,
+)
 from infrahub.graphql.analyzer import InfrahubGraphQLQueryAnalyzer
 from infrahub.graphql.execution import cached_parse
 from infrahub.graphql.initialization import prepare_graphql_params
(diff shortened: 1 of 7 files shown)


backend/infrahub/core/regeneration/predicates.py
```diff
@@ -85,7 +91,8 @@ def relevant_node_changes(
         if not readable_fields:
             continue
         updated_fields = {element["name"] for element in node_diff["elements"] if not _is_unchanged(element["action"])}
-        if updated_fields & readable_fields:
+        imprecise_read = bool(readable_fields & IMPRECISE_READ_FIELDS)
```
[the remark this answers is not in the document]
polmichel: Addressed — the residual is now handled one layer up, in `impact.py` where the schema is available (the pure `predicates.py` can't decide scopability).

`reads_unscopable_derived_field` checks each `display_label` / `human_friendly_id` read with `derived_read_is_scopable`. When a read is **not** scopable to the reading kind's own attributes (i.e. the label/hfid is composed from a related node's field), the query sets `depends_on_everything` and widens to every target instead of narrowing — so a change on the composing peer no longer slips through.

Scopable reads keep narrowing precisely. A follow-up that narrows peer-composed labels to the owning member via the dependent-node resolver (instead of the conservative widen) is tracked for 1.12.

backend/infrahub/core/regeneration/impact.py
```diff
@@ -41,6 +49,23 @@
...
+) -> bool:
+    """Whether the query reads a display_label / human_friendly_id composed from a related peer.
+
+    Such a read cannot be narrowed: the change that moves the value lands on a peer the read set
+    never names, so the field-level match cannot see it, and the query has to fall back to every
+    target. A read of a derived field with no declared path is treated the same way, conservatively.
+    """
+    for kind, fields in readable_fields_by_kind.items():
+        for field_name in fields & IMPRECISE_READ_FIELDS:
+            if not derived_read_is_scopable(node_schema=get_node_schema(kind), field_name=field_name):
```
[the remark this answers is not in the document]
polmichel: Fixed. The caller now guards the lookup — `query_schema_branch.get(name=kind, duplicate=False) if query_schema_branch.has(name=kind) else None` — and `reads_unscopable_derived_field` treats a `None` schema as unscopable and widens, so an unresolvable kind can no longer raise `SchemaNotFoundError`. Added a regression test (`derived_read_on_a_kind_absent_from_the_schema_widens`).

I kept the small local helper rather than importing the private `_derived_reads_are_scopable` from `computed_attribute/tasks.py`: that would be a cross-package private import pulling in a heavy flow module. Both paths already share the underlying `derived_read_is_scopable`, so the duplicated logic is just the read-set iteration.

backend/infrahub/core/regeneration/impact_classifier.py
```diff
@@ -48,12 +48,17 @@ class QueryImpactClassifier:
...
+    ``depends_on_everything`` marks a query whose read surface cannot be pinned down at all -- a read
+    of a derived value composed from a peer the read set never names -- so any relevant change widens
+    to every target.
     """
 
     query_branch: str
     only_has_unique_targets: bool
     traversed_kinds: set[str]
     readable_fields_by_kind: dict[str, set[str]]
+    depends_on_everything: bool = False
```
polmichel, line 61: This may be not the final solution on `develop` branch.

backend/infrahub/core/regeneration/impact.py
```diff
@@ -41,6 +49,30 @@
 """
 
 
+def reads_unscopable_derived_field(
+    readable_fields_by_kind: Mapping[str, set[str]],
+    get_node_schema: Callable[[str], MainSchemaTypes | None],
```
ajtmccarty: seems like passing the `SchemaBranch` in here would make more sense than any `Callable`. maybe there are future plans I am missing

or really just the `MainSchemaTypes` itself

ajtmccarty (approved):

backend/infrahub/core/regeneration/impact.py
polmichel: You're right, thanks!

polmichel:
Fyi @ajtmccarty I forgot about this commit that I pushed before merging the PR. It adds a component test in the existing test suite and has not been reviewed. https://github.com/opsmill/infrahub/pull/10469/changes/e06f259d4414b330100cb4c66aabe66f8f05f0fd

PR #10477, opened 2026-09-01
Gate the Python automations on the live origin

gmazoyer:
# Why

A merge or a rebase replays every changed node as an event, and the two Python computed-attribute
trigger types answered all of them. None of those recomputes wrote anything: the branch had already
computed the values. Gating them is what removes the fan-out.

Last of three PRs ([IFC-3019](https://opsmill.atlassian.net/browse/IFC-3019)).
[#10416](https://github.com/opsmill/infrahub/pull/10416) added the resolver,
[#10443](https://github.com/opsmill/infrahub/pull/10443) ran it alongside the per-node automations
so nothing could be missed. Both have merged.

## Measured with the scenario harness (IFC-3048)

Small 1.10.0 backup, both switch positions on this branch at 8de8cc88. Metrics: `computed_attribute_process_transform` runs, node ids total/unique, executions.

| scenario | switch off | switch on |
| --- | --- | --- |
| site_merge | 68 runs, 68/68, 272 exec | 3 coalesced runs, 68/68, 272 exec |
| rename_merge | 12 runs, 84/21 | 6 runs, 42/21, and 3 coalesced, 23/20 |
| site_rebase | 74 runs, 110/89, 110 exec | 3 coalesced runs, 88/88, 88 exec |
| commit_fanout | 16 runs, 3518/3518, 3722 exec | same |

The gate removes the per-node fan-out on merge and rebase, with the same unique nodes covered. Off reproduces the per-node baseline exactly, so the rollback works. The commit path does not move. The rename residue (2x per node) is IFC-3047.

The one-reader diff (20 vs 21, 88 vs 89) is not a miss: a develop run with both populations names the node, it is the renamed device itself. Its transform reads only `role` and `device_type.part_number`, not `name`, so the skip is correct.

Runs: [on](https://github.com/opsmill/infrahub-private-tests/actions/runs/34375941937), [off](https://github.com/opsmill/infrahub-private-tests/actions/runs/34375870713), [develop id-diff](https://github.com/opsmill/infrahub-private-tests/actions/runs/34375860925).

## What changed

* **The gate.** Both Python trigger builders match `infrahub.node.origin == live`, so a merge, a
  rebase and a coalesced write start no per-node flow. It also closes both echo loops: coalesced
  writes carry the `recompute` origin, and a merge only reaches `process_transform` through the
  coalesced pass, which passes `coalesced=True`.
* **One switch, both halves.** The filter is applied only while
  `INFRAHUB_COALESCE_PYTHON_RECOMPUTE_AFTER_MERGE` is on, so the pass can never be off while the
  filter is on. No config change here; #10416 shipped the setting on by default.
* **An unpinned query resolves no readers.** Query-group membership records what the last run read,
  so with an unpinned root a node enters or leaves the result set untouched. Any change reaching
  such an attribute widens to its whole kind, creations included. It keeps its real read set, which
  is what the schema-scoped backfill scopes on. Artifacts and generators guard the same case.
* **A failed resolution widens rather than skips.** This is the only family that reads the database,
  so its resolution is wrapped to keep a failure off the other three submissions. It used to return
  nothing, which was safe only while the per-node automations still answered a replay; now every
  declared attribute is refreshed over its whole kind. Same for a resolver that cannot be built at
  all, on the rebase path.
* The component fixtures gain the `ids` filter a real transform query carries, so those tests now
  cover the pinned path and the new tests cover the unpinned one.

## How to review

`computed_attribute/models.py` is the gate. `python_target_sources.py` holds the unpinned-query
restriction, and its docstring says why it belongs to reader resolution and not to the
schema-change backfill. Putting it in the shared mapping instead cost the per-kind precision the
derived-read scoping was built for, which the component suite caught.

## How to test

```bash
uv run pytest backend/tests/unit -q
```

Green: 2639 unit tests locally, plus `ruff`, `ruff format`, `ty`, `invoke backend.mypy`,
`docs.validate` and `backend.validate-generated`. CI is green on this head, including both
`backend-docker-integration` shards, which run the end-to-end merge and rebase gate.

Every guard added here was run against its reverted fix and shown to fail: the widened fallback,
the unpinned read set, the self-target rule, the schema-refresh ordering and the awaited-call
matching.

## Impact & rollout

* **Config/env changes:** none new. `INFRAHUB_COALESCE_PYTHON_RECOMPUTE_AFTER_MERGE=false` is the
  rollback, but it is **not one step**: the filter lives in the stored Prefect automation, so run
  `infrahub upgrade` after the flip to rebuild the two Python trigger types. Every worker must
  carry the new value before any reconcile runs, or a mixed fleet keeps re-flipping them.
* **Performance:** the saving grows with the number of changed nodes. A merge that changes a single
  node does not get faster.
* **Known follow-up:** [IFC-3162](https://opsmill.atlassian.net/browse/IFC-3162). A failed
  resolution widens every declared attribute, including any whose transform is missing, and those
  flows raise. IFC-3060 settled that for the normal path; the failure path cannot tell them apart.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry: one fragment for the whole stack, #10416 and #10443 carry none for the feature
- [x] Internal .md docs updated (`merge-recompute.md`, `computed-attributes.md`)

[IFC-3019]: https://opsmill.atlassian.net/browse/IFC-3019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-3060]: https://opsmill.atlassian.net/browse/IFC-3060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10477?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[IFC-3162]: https://opsmill.atlassian.net/browse/IFC-3162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dev/knowledge/backend/merge-recompute.md
```diff
@@ -30,15 +30,21 @@ The builder, submitter, and coordinator live in `core/merge/recompute_coalescing
...
+- **A schema-changing merge already refreshes what its own scope selects**, one whole kind at a time, through `SchemaUpdatedEvent`. Those pairs are dropped here, decided by the same scoper both sides run over candidates from the same gather. Only a pair the gather returned may be dropped. When the gather fails outright, no pair may be dropped: every one carries an imprecise read set and stays in the pass. The send is checked: `PostMergeDispatcher` scopes the pass only after the schema event went out. What nothing checks is that `computed_attribute_setup_python` reached its submission. That flow has no retry, and the origin gate means no per-node automation answers the merge either, so a failure there leaves the dropped pairs stale until the next change touches them. Dropping them is still right: the duplicate has to come off the coalesced side, never off the schema side.
```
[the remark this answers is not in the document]
gmazoyer: INFRAHUB_COALESCE_PYTHON_RECOMPUTE_AFTER_MERGE is only temporary this is useless to make it part of the doc and states its use.

backend/infrahub/core/merge/recompute_coalescing.py
```diff
@@ -538,28 +538,47 @@ async def submit(
...
     """
     try:
         return await resolver.resolve(changes=changes, branch=branch, schema_changed_elements=schema_changed_elements)
     except Exception:
-        log.exception("Leaving the Python computed attributes of branch %s to the per-node automations", branch)
-        return []
+        log.exception(
+            "Widening every Python computed attribute on branch %s to its whole kind: the resolution failed", branch
+        )
+        return _every_python_attribute_widened(schema_branch)
```
ogenstad, line 587: I'm wondering here if we should return early if there are no changes. I.e. before line 575 that we'd have:

```python
if not changes:
    return []
```

I.e. if there's a rebase without any actual data changes. Would an error in `resolver.resolve()` cause all computed transforms to be executed without this? Might also save some CPU cycles.

Also wondering if `_every_python_attribute_widened()` actually takes the database into consideration or if we'd still end up trying to trigger computed transforms that doesn't have the transform created in the database yet. If that's the case the correct approach here might actually be to allow it to be triggered but then strengthen the run of those computed attributes so that we instead correctly capture the error when we can't find the transform and instead log a warning when that happens.

ogenstad (commented):
Looks good overall to me! I left one comment that I think we might want to address before merging.

backend/infrahub/core/merge/recompute_coalescing.py
gmazoyer: The early return is a good idea, I added it. I also added a small test to cover this new conditional.

On the second part, you are right that `_every_python_attribute_widened()` only reads the schema branch, so it can include attributes whose transform is not in the database yet, and those flows raise. That is tracked in IFC-3162. I added your suggestion there as an option: keep widening, but let the run handle a missing transform with a warning instead of an error. I think it is the best of the four listed, because the other options try to guess which transforms exist while yours handles it where the answer is known, and it also covers a transform deleted between the widening and the run.

I can work on IFC-3162 so it is done as soon as possible after this PR gets merged.

ogenstad (approved):

PR #10484, opened 2026-09-02
feat(frontend): let users choose which columns a list view shows [IFC-3079]

pa-lem:
Closes IFC-3079. Implements INFP-119.

Every schema-driven table renders the columns its schema dictates, identically for every user and
every task. Wide kinds are hard to scan with no way to trim them, attributes the schema marks
`display: "extra"` cannot appear in a list at all, and sort and filter are already shareable in a
link while column choice is not. This adds user-controlled column visibility to all four
schema-driven tables.

```text
?hide_columns=description,status&show_columns=internal_note
```

Two entry points: a **Columns** checklist in the table toolbar, and a **Hide column** shortcut in
each header menu. Both write the same two params.

| Surface | Hide | Reveal |
|---|---|---|
| Object list | yes | yes |
| IPAM addresses | yes | no |
| IPAM prefixes | yes | no |
| Relationship tabs | yes | no |

image
image
image

## Decisions worth knowing before reviewing

**Two named params, not one prefixed list.** A single param needs a reveal prefix, and nuqs's
`encodeQueryValue` rewrites `+` to `%2B`, so `?columns=%2Binternal_note` is what a shared link would
actually read. A JSON object form is worse — every quote becomes `%22`, exactly as the `filters`
param already looks today. Named params need no convention to learn and stay legible, which is the
entire point of a shareable link.

**Delta from the default, never an absolute list.** A name appears only where it departs from the
schema's default, so a schema that later gains a column shows it on an existing link. Product asked
for this explicitly. Unknown names are dropped by a single trust boundary, which is what keeps an
old link safe across a schema change, a kind switch, and a relationship tab reading the same params
against a different schema.

**Reveal is object-list only.** It needs three legs — a ColumnDef must exist, the field must reach
the GraphQL selection set, and the react-query cache key must change. Only the object list has an
injectable fetch seam; `get-object-relationships-from-api.ts` has none. The other three surfaces are
hide-only via `canReveal: false`, implemented as "candidates == defaults" so there is no
`if (isIpam)` anywhere.

**IPAM's default column set and ordering are deliberately untouched.** Those two builders key
special cells on attribute *name* and have no component tests, so changing what they emit would
silently change cell rendering.

**One capability, not per-surface gates.** `ObjectTableContext` carries
`supportsColumnVisibility`, default `false`, and both entry points read it. Three managers opt in.
This exists because the same defect appeared three times during review: a control on a surface that
renders the shared header or toolbar without consuming `columnVisibility` — the five role-management
pages, then Proposed Changes, which passes `schema` to five headers purely for sorting and renders a
`ListBox`. Gating on `schema` was wrong because on this codebase `schema` means "this header can
sort". A surface added later now gets nothing until it asks.

**`objectQueryKeys.list` folds revealed field names** via a conditional spread, so the key stays
byte-identical for callers that never reveal anything and no cache is invalidated on deploy. `count`
needs no equivalent — that query selects no fields. `detail` already folded field names in, which is
why `list` was the outlier.

## Reviewing this

`dev/knowledge/frontend/column-visibility.md` is the intended entry point. The `_resource_from_pool`
polarity and the IPAM `display: "extra"` divergence are both documented there because both are
half-true in a way that invites a wrong "fix".

Two defects found in review are worth a look, since both are the kind that only appear at a boundary:
`DataTable` emitted `repeat(0, auto) 1fr 2.5rem` once only the identity and actions columns
remained, which the CSSOM rejects outright — every row split in two. The trust boundary now keeps
one field column visible, and the picker disables that last item rather than letting the click
silently no-op. Separately, hiding a column on one kind then switching kind used to destroy the
first choice; the kind switcher already prunes filters the new schema lacks, and column params now
prune beside it.

## Verification

`biome ci` (1477 files) · `knip` · `betterer ci` *stayed the same* · **1339 tests / 192 files**.
`tsc` sits on its pre-existing 195-error baseline, unchanged.

Manually exercised against a local stack with demo data: a shared `?hide_columns=` link, the
popover's checked state, the active-filter marker, and reveal on `/objects/CoreAccountGroup` —
whose `origin` attribute is `display: "extra"` — confirming `origin { id value __typename }` enters
the GraphQL request rather than the column rendering empty.

## Gaps, stated plainly

- **Reveal has no E2E.** It has unit and component coverage at all three legs and was verified by
  hand, but the one approved E2E covers hide only. A reveal E2E needs a demo kind with an `extra`
  attribute plus an assertion that the cells are non-empty.
- **Relationship tabs and both IPAM tables have no component tests.** Standing up the first is a
  network-bound composite costing more than this feature; their four `columnVisibility` prop lines
  are kept byte-identical so a reviewer can see any divergence in the diff.
- **`data-table.test.tsx` collides with #10427.** That PR added a file at the same path on `stable`
  (first-column overflow, sticky footer, tooltips); this one adds `columnVisibility` tests there.
  `develop` has neither yet, so nothing conflicts now — but the eventual stable→develop merge will
  see "both added". Two `describe` blocks in one file; trivial, just not a surprise.

## Follow-ups to file

Column reordering (promised in the ticket) · unify the three column builders, gated on adding IPAM
component tests first · align IPAM attribute filtering to the object rule, which removes
currently-visible columns and needs product sign-off · durable per-user preferences via the backend
`preferences` entity · wire the five role-management tables · reveal on relationship tabs · a
semantic Tailwind token layer, since `styling.md` recommends tokens this codebase does not have.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10484?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/app/src/shared/components/table/data-table.tsx
```diff
@@ -34,12 +36,19 @@ export interface DataTableProps<T> extends React.HTMLAttributes<HTMLDivElement>
...
+ * declaration — collapsing the grid to one implicit column and splitting every row in two.
+ */
 const defaultGridTemplateColumns = (columnCount: number) =>
-  `repeat(${columnCount - 2}, auto) 1fr 2.5rem`;
+  columnCount <= 2 ? "1fr 2.5rem" : `repeat(${columnCount - 2}, auto) 1fr 2.5rem`;
 
 export function DataTable<T extends NodeCore>({
   columnOrder,
   columns,
+  columnVisibility,
```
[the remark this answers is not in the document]
pa-lem: Checked this one carefully because it is the only P1, and I don't think it holds — `getFlatHeaders()` does respect column visibility.

In `@tanstack/table-core@8.21.3`, `build/lib/core/headers.js`:

```js
table.getHeaderGroups = utils.memo(() => [
  table.getAllColumns(),
  table.getVisibleLeafColumns(),   // ← visibility is a dependency
  table.getState().columnPinning.left,
  table.getState().columnPinning.right,
], ...);

table.getFlatHeaders = utils.memo(() => [table.getHeaderGroups()], headerGroups =>
  headerGroups.map(hg => hg.headers).flat());
```

So `getFlatHeaders()` derives transitively from `getVisibleLeafColumns()`. Headers, the grid track count and the footer count all follow from `allHeaders.length`, which is therefore already visibility-aware.

Three independent confirmations rather than just the source read:

- `data-table.test.tsx` has *"omits both the header and the body cell of a column hidden by the visibility state"*, which passes — it would fail if the header survived.
- *"derives gridTemplateColumns from the visible column count only"* asserts the track count against the visible count, and *"renders one footer count cell per visible column"* asserts the footer count. Both pass.
- Verified in a browser against a local stack: `/objects/InfraDevice?hide_columns=description` renders no Description header and no Description cells.

Leaving the code as-is. Happy to be shown otherwise if you have a case where a hidden column does render a header — that would mean the memo dependency isn't behaving as the source suggests, which would be worth knowing.

pa-lem:
Worked through all 18 review comments. 17 addressed, 1 rejected with evidence in its own thread.

### Two real defects

**`in` walks the prototype chain.** `name in visibility` is true for `constructor`, `toString` and friends, so a field with one of those names read as visible when it wasn't. The picker was the reported site; the one that mattered more was the clamp guaranteeing at least one visible column, where a phantom read lets the table reach the zero-column state the clamp exists to prevent. Both use `Object.hasOwn` now.

Worth noting the suggested repro doesn't fail: hiding a default-*visible* `constructor` creates a real own key, so `in` reads the true value. It needs a default-*hidden* field of that name alongside a hidden visible one — that's the fixture the new test uses.

**Re-selecting the active kind cleared the column params.** Clearing is only correct when the schema actually changes. Fixed in both selector callbacks, with tests for the generic and the concrete-kind branch.

### One rejected

The P1 on `getFlatHeaders()` — it does respect visibility. Full evidence in that thread: the memo dependency in table-core 8.21.3, three passing tests, and a browser check.

### Style findings — all valid, and the useful part of this review

`.agents/rules/code-doc-style.md` forbids naming other code in comments, narrating the approach not taken, and multi-line why-comments. Nine sites did all three; trimmed, with the genuinely load-bearing warnings moved into `dev/knowledge/frontend/column-visibility.md` where the rule says paragraph-length rationale belongs.

`dev/guidelines/documentation.md` forbids `file.ts:NN` citations in favour of `module.ts::Symbol`, on the grounds that line numbers rot before the feature merges. Forty-six converted, every symbol machine-verified to resolve. That call was well made: mine rotted twice while this branch was in flight — once from a file move, once because a longer identifier triggered a formatter reflow — and one of the new symbol references survived a concurrent edit to the very comment it cites.

Also fixed: a test name attributing exclusion to cardinality when the operative filter is the resource-pool check, a comment claiming `schema` only gates sorting when it now also gates the hide entry, five British spellings, and the ticket ID in the E2E docstring.

### The relationship-table test

Added, and it earned its place in an unexpected way. A bare `?show_columns=<extra field>` proves nothing there — that table builds default columns without `fields`, so TanStack ignores a visibility key with no matching column, and the test passes even under the wrong surface. What does pin it is `?hide_columns=<every field>&show_columns=<extra field>`: under the correct surface the clamp hands a column back, under the object surface the revealed field satisfies the clamp instead and every field column disappears. Both cases mutation-tested — surface swap and dropped wiring each fail the tests that should catch them.

Gate after the changes: `biome ci` (1478 files) · `knip` · `betterer ci` unchanged · **1356 tests / 192 files** · `tsc` on its pre-existing 195-error baseline.

dev/knowledge/frontend/column-visibility.md
saltas888, line 1: this is too big and layouts all the pieces of code. The documentation here should be high level. Imagine you on board someone, you dont want them to explain the code in a markdown file

frontend/app/src/entities/nodes/columns/ui/columns-picker.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * The badge counts the departures from the surface's default column set rather than the hidden
+ * columns: revealing a column customizes the view just as much as hiding one.
+ *
+ * It counts only the departures the trust boundary kept, so a param naming a field this schema does
+ * not have shows nothing here — there is no column on screen to point at. Whether there is a param
+ * left to clear is a different question, which is why the editor gates its reset control on the raw
+ * params instead.
+ */
+export function ColumnsPicker({ schema, surface }: ColumnsPickerProps) {
+  const { customizedColumnCount } = useColumnVisibility(schema, surface);
```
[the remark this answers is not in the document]
saltas888: I think thats good

saltas888 (commented):
Also leaving some comments after reviewing with AI cause I feel like some things can be simplified

saltas888 (commented):
Read through the columns module. The logic itself is tight, I don't think the algorithm is the problem anywhere, and the clamp is clearly earned given the CSSOM failure. What I keep tripping on is the ratio of prose to code in the domain rules, plus one abstraction that felt heavier than what it actually configures. Four notes inline, none of them blockers.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
```diff
@@ -0,0 +1,98 @@
...
+ * **Why at all.** The IPAM tables render an "available range" row as one cell spanning
+ * `col-start-2 -col-end-2`, which needs a grid track between the identity and actions columns; with
+ * every field column hidden the two grid lines collapse onto each other and the row wraps. Beyond
+ * that one symptom, a table showing only a row label and an actions menu carries no data and gives
+ * the user nothing to click back from except Reset.
+ *
+ * A surface offering nothing but default-hidden columns has no hide entry to give back and is
+ * returned untouched: there is no hide request to blame for the empty table, so there is none to
+ * relax.
+ */
```
saltas888, line 52: 26 lines of docstring for a 3 line rule, and it's doing three things I'd rather it didn't.

It describes the picker's behaviour and the table's, so it goes stale the moment either of those changes. The "Why at all" paragraph is a postmortem of the grid defect, which belongs in the PR and the knowledge doc rather than next to the function. And "Product asked for this explicitly" is an argument for review, not documentation.

The one-line why-comments elsewhere in this file are exactly the right register, `// Only a default-visible column has anything to hide.` tells me everything I need. I'd cut this to the rule plus one sentence on why display order decides the survivor.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
```diff
@@ -0,0 +1,98 @@
...
+): ColumnVisibilityState {
+  const defaultVisibilityByName = new Map(
+    columnCandidates.map((field) => [field.name, field.isDefaultVisible])
+  );
+  const hideRequests = new Set(hiddenNames);
+
+  // Only a default-visible column has anything to hide.
+  const hidden = [...hideRequests].filter((name) => defaultVisibilityByName.get(name) === true);
+  // Only a default-hidden column has anything to reveal, and never one the other param hides.
+  const shown = [...new Set(shownNames)].filter(
```
saltas888: The `new Set` here isn't doing anything, both lists land in `Object.fromEntries` where a duplicate key just overwrites itself with the same value. The Set on the hidden side is load bearing since you call `.has()` on it, this one isn't, so the asymmetry reads as significant when it isn't. I'd drop it.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
```diff
@@ -0,0 +1,98 @@
...
+  visibility: ColumnVisibilityState,
+  columnCandidates: ColumnCandidate[]
+): ColumnVisibilityState {
+  // Own keys only: a field named after an `Object.prototype` member must never read as visible on a
+  // state that does not mention it.
+  const isVisible = ({ name, isDefaultVisible }: ColumnCandidate) =>
+    Object.hasOwn(visibility, name) ? visibility[name] : isDefaultVisible;
+  if (columnCandidates.some(isVisible)) return visibility;
+
+  const survivor = columnCandidates.find(({ name }) => visibility[name] === false);
```
saltas888: This guard and the `survivor` lookup below disagree, that one does a raw `visibility[name] === false` with no `hasOwn`. Not a bug as written since `=== false` narrows, but if the hazard is worth defending then both should defend it.

I'd rather make it go away, either drop the guard or give the state a null prototype so the whole class disappears and the comment with it. Worth asking whether it's a hazard we guard anywhere else, a schema attribute named `constructor` isn't something the rest of the app worries about.

frontend/app/src/entities/nodes/columns/domain/model/column-surface.ts
```diff
@@ -0,0 +1,22 @@
...
+ * reveal seam: the candidate list then collapses to the defaults.
+ */
+export interface ColumnSurface {
+  readonly fixedColumnIds: readonly string[];
+  readonly getDefaultAttributes: (attributes: AttributeSchema[]) => AttributeSchema[];
+  readonly getDefaultRelationships: (relationships: RelationshipSchema[]) => RelationshipSchema[];
+  readonly excludeField: (field: FieldSchema) => boolean;
+  readonly orderFields: (fields: FieldSchema[]) => FieldSchema[];
+  readonly canReveal: boolean;
+}
```
saltas888, line 20: Six members for four surfaces, and two of them supply `canReveal: false`, identity `orderFields` and `excludeField: () => false`, so they use none of the expressiveness. A third is `{...OBJECT_COLUMN_SURFACE, canReveal: false}`. It's really one configuration and three thin ones.

The tradeoff I'd push on gently: no `if (isIpam)` is nice, but "which columns does the IPAM prefix table show" now spans this file, the surfaces file, the candidates file and four rule modules across three entity folders. And a docstring explaining that the interface deliberately carries no identifier so nobody can branch on it is a fair bit of scaffolding to prevent one branch.

Not asking you to tear it out, the follow-up about unifying the three column builders probably subsumes it. Mostly flagging that the indirection isn't free.

frontend/app/src/entities/nodes/relationships/ui/relationship-table/relationship-table.tsx
```diff
@@ -31,6 +35,10 @@ export function RelationshipTable({
   const { schema: parentSchema } = useSchema(parentKind);
   const { data: permission } = useGetObjectPermissions(parentKind);
   const [filters] = useFilters();
+  // The surface is named here rather than read from `ObjectTableContext`: two of the three hosts
+  // render this table without a provider, and the one that does provide it carries the object
+  // surface, whose `canReveal: true` would offer fields the relationship fetch never requests.
+  const { columnVisibility } = useColumnVisibility(relationshipSchema, RELATIONSHIP_COLUMN_SURFACE);
```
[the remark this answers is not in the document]
pa-lem: Addressed — a `relationship-table.test.tsx` now covers this wiring. Writing it turned up the case that actually pins the surface, which is narrower than a bare reveal: with `?hide_columns=` naming every field and `?show_columns=` naming an extra one, a surface that could reveal would let the revealed field satisfy the minimum-visible clamp and every field column would disappear. A bare `?show_columns=` proves nothing here, because this table builds default columns and the table library ignores a visibility key with no matching column. Both cases are mutation-tested — swapping the surface, and dropping the wiring, each fail the tests that should catch them.

frontend/app/src/entities/ipam/ip-addresses/ui/ip-address-table.tsx
```diff
@@ -46,6 +48,7 @@ export function IpAddressTable({ baseFilters = [] }: IpAddressTableProps) {
     <InfiniteScroll scrollX hasNextPage={hasNextPage} onLoadMore={fetchNextPage}>
       <DataTable
         columnOrder={IP_ADDRESS_TABLE_COLUMN_ORDER}
+        columnVisibility={columnVisibility}
```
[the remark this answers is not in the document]
pa-lem: Not adding IPAM component tests here, deliberately — recording the reasoning rather than leaving this open.

These two tables pass `columnVisibility` and nothing else; their column builders are untouched by this change. The risk you are pointing at is real but it sits in the builders, which key special cells on attribute *name* and have no component tests at all today. Testing the prop-passing without testing the builders would buy very little, and standing up the builder tests is a bigger piece of work than this feature.

So: the two IPAM edits are kept byte-identical to each other so a reviewer can see any divergence in the diff, the object list is covered by a component test and the one E2E, and "add IPAM component tests" is filed as the prerequisite for the follow-up that unifies the three column builders.

frontend/app/src/entities/nodes/columns/ui/columns-picker.tsx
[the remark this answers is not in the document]
pa-lem: Condensed — the comment is now the picker's observable contract, and the badge wording corrected: it said "departures with a column on screen", which was wrong since a hidden column is counted and is precisely not on screen. It now says departures naming a column the surface offers.

frontend/app/src/entities/nodes/columns/domain/model/column-surface.ts
```diff
@@ -0,0 +1,27 @@
...
+ * only be read from the fields below.
+ *
+ * `getDefault*` must be the very functions that surface's column builder calls, otherwise the
+ * picker can offer a column the table cannot render. `canReveal` is false wherever the fetch path
+ * has no reveal seam: the candidate list then collapses to the defaults.
+ *
+ * The vocabulary lives here; the four concrete surfaces live in `domain/rules/column-surfaces`.
+ * Naming a surface means naming the rule functions it composes, and `domain/model` is a pure leaf
+ * that may not reach into `domain/rules`.
+ */
```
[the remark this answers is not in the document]
pa-lem: Fixed — the docstring no longer names the module path, the concrete surface, or its callers, and no longer argues for the approach. It states what the interface is and the one invariant that matters: `getDefault*` must be the functions that build that table's columns.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
pa-lem: You're right, and the register point lands — I wrote the one-liners you're pointing at and then didn't apply the same standard three functions down.

Cut to the rule plus the one sentence on display order:

```
/**
 * Drops one hide entry when the hide list would leave no field column visible, so the state can
 * never empty a table. Display order picks which, so the same set of names always leaves the same
 * column standing however the URL spells them.
 */
```

The grid-defect postmortem and the "why here rather than in the picker" argument are both in the PR description and `dev/specs/column-visibility-infp-119/plan.md`, which is where the rejected alternatives live. "Product asked for this" is gone entirely — you're right that it's a review argument, not documentation.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
pa-lem: Correct, and the asymmetry was the actively misleading part. Dropped.

While removing it the loop got simpler too — both lists now write straight into the state instead of building two intermediate arrays and spreading them:

```ts
for (const name of shownNames) {
  if (defaultVisibilityByName.get(name) === false && !hideRequests.has(name)) {
    visibility[name] = true;
  }
}
for (const name of hideRequests) {
  if (defaultVisibilityByName.get(name) === true) visibility[name] = false;
}
```

`hideRequests` stays a Set because `.has()` is genuinely load-bearing there.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
pa-lem: Took the null-prototype option — it makes the class disappear rather than guarding it twice, which is what you were getting at.

`Object.create(null)` for the state, both lookups back to plain `in` / `[name]` with no `hasOwn` anywhere, and the comment about own-keys deleted. The one place that rebuilds the object (dropping the survivor's entry) rebuilds it null-prototyped too, so a spread can't quietly reintroduce a prototype.

On whether it's worth defending at all: you're right that nothing else in the app worries about a schema attribute named `constructor`. I kept the null prototype because it costs one line and ends the question, where dropping the guard outright would have left the hazard live — without a null prototype, `"constructor" in visibility` is true and the picker marks an unrendered field visible. Mutation-checked that the test earns its place: swapping `Object.create(null)` for `{}` fails exactly that case and nothing else.

frontend/app/src/entities/nodes/columns/domain/model/column-surface.ts
pa-lem: Fair, and I'd rather record the cost than argue it away.

You're right about the shape: it's one real configuration and three thin ones, and the discoverability cost is real — "which columns does the IPAM prefix table show" does span several files now, where before it was one builder you could read top to bottom. The docstring defending the missing identifier was itself part of that cost, so it's gone; the interface now just says what it is and the one invariant that matters.

I've added the tradeoff to the follow-up that unifies the three column builders, since as you say that probably subsumes this — if the builders become one, the surfaces mostly stop needing to be data.

The thing I'd keep is `getDefault*` pointing at the same functions the builder uses. That's what stops the picker offering a column the table can't render, and it was a real bug earlier in this branch rather than a hypothetical.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Both lists arrive from the URL, so either may name a field this schema never had — an old shared
+ * link, a kind switch, or a relationship tab reading the same params against a different schema.
+ * Such a name is dropped here rather than reaching the table, where an unknown column id would
+ * silently do nothing. A name that agrees with its surface's default is dropped too: it is not a
+ * departure, so it must not count towards the picker's badge.
+ *
+ * A name in BOTH params is a contradictory link, and hiding wins: with two named params there is no
+ * ordering to fall back on, and a link that says "hide this" must never put that column on screen.
+ *
+ * The returned state never leaves a table with no field columns.
```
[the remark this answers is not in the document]
pa-lem: Right, and this is the third place that same overstatement had to be corrected — the spec's functional requirement and the knowledge page both claimed the stronger guarantee before this. Taken your framing:

```
 * A hide request cannot take away the last visible field column. A surface offering nothing but
 * default-hidden columns is a different case and is returned unchanged: nothing was hidden, so
 * there is no hide request to relax.
```

Enforcing a survivor instead was the other option and I did not take it: with no hide entry to relax there is nothing to blame for the empty table, and inventing a visible column would mean overriding a surface's own defaults from inside the trust boundary.

dev/knowledge/frontend/column-visibility.md
```diff
@@ -0,0 +1,131 @@
...
+  react-query key, so the list starts over at page 1, losing its accumulated pages and scroll
+  position. Inherent to keying on them; the alternative is refetching every accumulated page under
+  the new selection set.
+- **Hiding is free only for a column that was visible by default.** Hiding a *revealed* column drops
+  it from `show_columns`, which shrinks the revealed set and so restarts the list the same way. Only
+  a `hide_columns` write never reaches the cache key.
+
+## Known follow-ups
+
+1. Column reordering.
```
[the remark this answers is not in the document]
pa-lem: Good catch on the list dropping it — that is exactly the failure mode of keeping a partial copy, so I have not restored it as a copy.

The item was never actually lost: it is tracked in `dev/specs/column-visibility-infp-119/tasks.md` and warned about in `get-column-candidates.ts`. But the tracked wording said "a field" where it meant `ip_prefix`, so grep could not find it — which is close enough to lost. That entry now names the function, the field, and why it is safe only by accident of the relationship's kind.

The knowledge page's follow-up section now points at the tracked list and calls out this landmine by name, instead of maintaining a subset that can silently drift again.

frontend/app/src/entities/nodes/columns/domain/rules/get-column-visibility-state.ts
pa-lem: Correction to my reply above: I have reverted that loop.

You were right that the dedupe was dead, and dropping it stands. But I framed the imperative rewrite as an improvement in its own right, and it was not — it was a consequence of moving the state to a null prototype, since a spread into a literal produces an ordinary object and throws the prototype away. Assigning onto a null-prototype target keeps both, so the declarative shape is back:

```ts
const shown = shownNames.filter(
  (name) => defaultVisibilityByName.get(name) === false && !hideRequests.has(name)
);
const hidden = [...hideRequests].filter((name) => defaultVisibilityByName.get(name) === true);

const visibility: ColumnVisibilityState = Object.assign(
  Object.create(null),
  Object.fromEntries(shown.map((name) => [name, true])),
  Object.fromEntries(hidden.map((name) => [name, false]))
);
```

That also sits better next to `get-column-candidates.ts`, which is a remeda pipeline. Mutation-checked: swapping the assign for a spread literal fails exactly the constructor-named-field case.

saltas888 (approved):

PR #10489, opened 2026-09-02
Report children on a hierarchy attach

### LINKED ISSUE #10485: bug: TransformPython computed attribute never recomputes after initial catch-up when its query traverses a cardinality-many relationship
### Component

### Infrahub version

1.11.1

### Current Behavior

A `Text` attribute with `computed_attribute: {kind: TransformPython, transform: <name>}`, whose backing GraphQL query reads a cardinality-many relationship (e.g. `GenericComponent.children`), computes correctly exactly once: when the transform is first registered via git-sync (or, if declared ahead of time, once the referenced `CoreTransformPython` becomes available). After that one-time catch-up, the attribute never recomputes again, even when the relationship it depends on changes later, on any object:

- Creating a new peer object and pointing its own `parent` relationship at an existing node with the computed attribute does not trigger a recompute on that node.
- Explicitly updating the node's own relationship field (e.g. `SomeNodeUpdate(data: {id: ..., children: [{id: ...}]})`) does not trigger a recompute either.
- Touching an unrelated attribute on the same node (e.g. `description`) does not trigger a recompute.

Querying `InfrahubTask` confirms this is structural rather than a delay: exactly one `Trigger updates for computed attributes ... for <Kind>.<attribute>` task and one `Process computed attribute for <Kind>.<attribute>` task exist for the whole life of the branch, both created at transform-registration time. No further task of either kind is ever created afterward, regardless of how many times the underlying relationship changes.

The transform logic itself is correct: invoking it directly with `infrahubctl transform <name> id=<id>` against the same object returns the right value at any time. Only the automatic dependency-based recompute is missing for this dependency shape.

### Expected Behavior

Per the "Computed attributes" guide, and per the 1.9.0 release notes ("changing a relationship used by a local computed attribute... recomputes inline against the new peer's attributes"; "remote changes (peer-node updates that affect computed attributes on other nodes) continue to use the existing background-task path"), a computed attribute should recompute whenever a field it depends on changes, including via a relationship, on an ongoing basis, not only once when the transform is first loaded.

### Steps to Reproduce

1. Schema: a `Chassis` kind (`GenericDevice`), a `Slot` kind (`GenericComponent`, mandatory `device` relationship), a `Card` kind (`GenericComponent`, parented into a `Slot` via the ordinary `parent`/`children` pair).
2. Add a computed attribute to `Slot`, e.g. `occupant` (`Text`, `read_only: true`, `optional: true`, `computed_attribute: {kind: TransformPython, transform: slot_occupant}`).
3. GraphQL query for the transform: fetch the `Slot` by `id` and its `children { edges { node { display_label } } }`.
4. Python transform: return the first child's `display_label` if any, else `""`.
5. Register the query and transform in `.infrahub.yml`, commit, push to a branch tracked by a `CoreRepository`, and let it sync.
6. Confirm the one-time catch-up: query an existing occupied `Slot`; `occupant` is populated correctly.
7. Create a new `Card` and set its `parent` to a currently-empty `Slot` (or attach it from the `Slot` side via an update mutation).
8. Query that `Slot`'s `occupant` repeatedly over several minutes: it stays at its pre-step-7 value (blank) and never updates to reflect the new child, even though a direct query of `children` on that `Slot` shows the new card correctly.

### Additional Information

- Confirmed via `InfrahubTask(q: "occupant")`: only the transform-registration-time trigger/process pair exists; no task is ever created for the step 7 change.
- Suspect this affects any computed attribute whose only dependency is a cardinality-many relationship populated from the peer side, since that's the only relationship shape exercised here.
- Workaround attempted and failed: touching the owning node's own unrelated attribute does not force a recompute, so there is no known way to manually refresh the value short of deleting and recreating the `CoreTransformPython` (which re-triggers the one-time catch-up).

gmazoyer:
# Why

A hierarchical schema declares `parent` and `children` under one identifier, `parent__child`. The peer changelog looked the other side of the edge up by identifier alone and kept the first match, which is `parent`. Attaching a child told its parent that its own `parent` had become the child, and the real change on `children` was never announced.

Every consumer of node events saw the wrong relationship. A computed attribute whose query reads `children` filters events on the fields it reads, so it dropped the event and never recomputed. A user action rule matching `children` never fired.

Closes #10485

## Scope against the issue

The issue lists three scenarios. This PR fixes the first one. The other two need no change.

- **A new peer points its own `parent` at the node.** Fixed and tested.
- **`Update(data: {children: [...]})` from the parent side.** Already worked. The primary changelog names that relationship directly, so it was never ambiguous. I verified the secondary changelog is identical before and after this change, and added a test to hold it.
- **Touching an unrelated attribute does not recompute.** This is the selective recompute design, not a defect. The trigger narrows `infrahub.field.name` to the fields the transform reads.

## What changed

- A hierarchy peer now gets an event naming `children`, with the right cardinality.
- The peer side is resolved by the mirrored direction: `parent` is outbound, so it maps to the inbound `children`.
- Deleting a node that has both a parent and a child now reports both sides. The mapper that builds a deleted node's relationship changelog was keyed on the identifier, so the second hierarchy edge reused the first edge's mapper and overwrote it. A middle node reported a single `parent` entry pointing at its own child.
- A cardinality-one peer that moved described its previous peer with the new peer's schema, so the relationship named for the old peer could be one it does not declare. The removed side is now resolved from the previous peer kind, as the removal branch already did.
- A pair holding one relationship per identifier on each side keeps its current behaviour.

Merge and rebase were never affected. `core/diff/query_parser.py` already disambiguates the same way, so this follows an established pattern rather than a new one. `RelationshipAdd` and `RelationshipRemove` share the fixed function.

## How to review

Start with `peer_relationships` in `backend/infrahub/core/changelog/models.py`, then the mapper key in `backend/infrahub/core/query/relationship.py`.

The fallback deserves the most attention. When no candidate mirrors the local direction, every candidate is returned, and the result is a guess. Two schemas reach it, both accepted by validation:

- A third kind reuses the `parent__child` identifier with the default `bidirectional` direction. Validation only checks the peers the hierarchy pair itself declares, so it never sees this one. Both entries produced are schema-invalid: the changelog names a `peer_kind` the relationship does not allow.
- A hierarchical node declares its own `children` under a different identifier, so it holds only `parent` under `parent__child`. The lone candidate does not mirror, and reporting it reproduces the original bug in a narrow shape.

Returning nothing instead would drop a real event, which is the failure this PR fixes, so the branch keeps every candidate and always logs a warning naming the peer kind, the identifier, the direction and the candidates. A user action rule on a falsely reported name still runs its generator, so that log line is the signal to alert on. Two unit params pin both shapes.

Deciding whether `validate_identifiers` should reject these schemas outright would make the branch dead code. That is a schema-validation change with its own compatibility risk, so it is not in this PR.

## How to test

```bash
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/core/changelog/ backend/tests/unit/core/changelog/ --neo4j
```

The hierarchy tests fail on the parent commit and pass here. Validated on release-1.10 as well: it sends the same wrong event, and only worked because the trigger carried no field filter before 1.11.0.

## Impact & rollout

- **Backward compatibility:** the event payload changes for a hierarchical attach, detach and delete. It was wrong before.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added

### DIFF
--- a/backend/tests/component/core/changelog/test_models.py
+++ b/backend/tests/component/core/changelog/test_models.py
@@ -480,3 +480,46 @@ async def test_node_changelog_parent(db: InfrahubDatabase, default_branch: Branc
     await car1_delete1.delete(db=db)
     assert car1_delete1.node_changelog.parent.node_id == person2.id
     assert car1_delete1.node_changelog.parent.node_kind == "TestPerson"
+
+
+async def test_secondary_changelog_names_the_hierarchy_children_relationship(
+    db: InfrahubDatabase, default_branch: Branch, hierarchical_location_schema_simple: SchemaRoot
+) -> None:
+    """A hierarchy peer must be told that `children` moved, not that its own `parent` did."""
+    region = await Node.init(db=db, schema="LocationRegion", branch=default_branch)
+    await region.new(db=db, name="region-1")
+    await region.save(db=db)
+
+    site = await Node.init(db=db, schema="LocationSite", branch=default_branch)
+    await site.new(db=db, name="site-1", parent=region)
+    await site.save(db=db)
+
+    rack = await Node.init(db=db, schema="LocationRack", branch=default_branch)
+    await rack.new(db=db, name="rack-1", parent=site)
+    await rack.save(db=db)
+
+    getter = RelationshipChangelogGetter(db=db, branch=default_branch)
+    attached = await getter.get_changelogs(primary_changelog=rack.node_changelog)
+
+    assert [changelog.node_id for changelog in attached] == [site.id]
+    assert attached[0].node_kind == "LocationSite"
+    assert attached[0].relationships == {
+        "children": RelationshipCardinalityManyChangelog(
+            name="children",
+            peers=[RelationshipPeerChangelog(peer_id=rack.id, peer_kind="LocationRack", peer_status=DiffAction.ADDED)],
+        )
+    }
+
+    to_delete = await NodeManager.get_one(id=rack.id, db=db, raise_on_error=True)
+    await to_delete.delete(db=db)
+    detached = await getter.get_changelogs(primary_changelog=to_delete.node_changelog)
+
+    assert [changelog.node_id for changelog in detached] == [site.id]
+    assert detached[0].relationships == {
+        "children": RelationshipCardinalityManyChangelog(
+            name="children",
+            peers=[
+                RelationshipPeerChangelog(peer_id=rack.id, peer_kind="LocationRack", peer_status=DiffAction.REMOVED)
+            ],
+        )
+    }
--- a/backend/tests/component/core/changelog/test_models.py
+++ b/backend/tests/component/core/changelog/test_models.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from infrahub.core import registry
 from infrahub.core.branch import Branch
 from infrahub.core.changelog.models import (
@@ -9,7 +11,7 @@ from infrahub.core.changelog.models import (
     RelationshipChangelogGetter,
     RelationshipPeerChangelog,
 )
-from infrahub.core.constants import DiffAction
+from infrahub.core.constants import DiffAction, InfrahubKind
 from infrahub.core.manager import NodeManager
 from infrahub.core.node import Node
 from infrahub.core.schema import SchemaRoot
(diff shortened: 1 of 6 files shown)


backend/tests/component/core/changelog/test_models.py
```diff
@@ -480,3 +482,250 @@ async def test_node_changelog_parent(db: InfrahubDatabase, default_branch: Branc
...
+    assert by_node[site_2.id].relationships == {
+        "children": RelationshipCardinalityManyChangelog(
+            name="children",
+            peers=[RelationshipPeerChangelog(peer_id=rack.id, peer_kind="LocationRack", peer_status=DiffAction.ADDED)],
+        )
+    }
+
+
+# A generic whose two members name their side of `shared__link` differently.
+DIFFERING_PEER_NAMES: dict[str, Any] = {
```
gmazoyer, line 631: Very specific schema we don't have anywhere else, hence being inlined here.

backend/tests/unit/core/changelog/test_models.py
```diff
@@ -99,3 +101,112 @@ def test_sensitive_attribute_update_status(test_case: SensitiveAttributeTestCase
...
+            peer="LocRack",
+            identifier=PARENT_CHILD_IDENTIFIER,
+            cardinality=RelationshipCardinality.MANY,
+            direction=RelationshipDirection.INBOUND,
+        ),
+    ],
+)
+
+# A hierarchical node whose own children relationship uses a different identifier, so it holds
+# only one side under `parent__child`.
```
ajtmccarty: do we actually allow this?

ajtmccarty (approved):

backend/tests/unit/core/changelog/test_models.py
gmazoyer: Maybe not on purpose, but yes. If a node inherits a hierarchy but declares its own `children` with a different identifier, `add_hierarchy_node` skips generating `children`, so the node keeps only `parent` under `parent__child`. I checked it against a real registered schema and validation accepts it, and `peer_relationships` then returns that single non-mirroring `parent`. I made the comment say how the shape arises, since it was not obvious from the test alone.

PR #4263, opened 2024-09-04
IFC-623 diff locking enhancements

ajtmccarty:
IFC-623

update how we handle locking when updating a diff. the old version was to always wait until the lock was acquired, and then run the diff update. if the frontend sends two requests for almost the same diff (same branches, 10 seconds apart), then we would end up calculating the whole diff two (or more) times, which is slow. 
the new approach is to add some special handling for when we get multiple requests to update the diff for the whole lifetime a given branch. in this case, we will check for an existing lock. if we find one, then we wait for it to be released and we return. otherwise, we run the diff update calculation as normal.

- had to make some updates to the lock code to get this to work
- added some debug logging that might be helpful with some problems we are seeing in the E2E tests
- also added some exception handling for if a proposed change is not found when trying to update the CoreDataChecks for a given diff. it seems like this is happening in the E2E tests somehow, but I'm not totally certain

PR #10504, opened 2026-09-03
fix: calculate the branch diff when the lock holder stored none

saltas888:
## The bug

Pressing **Refresh diff** on a proposed change whose diff has never been computed can fail with `Cannot find diff for branch <name>`. The UI shows an error toast and stays on "We are computing the diff between … and …", with no way forward but a reload.

`update_branch_diff` reads a held incremental lock as "another update is in flight": it waits for the lock, then reads the diff that holder is assumed to have saved.

That premise does not hold. `DiffTree` and `DiffTreeSummary` take the *same* lock purely to wait out an in-flight update ([tree.py:575](backend/infrahub/graphql/queries/diff/tree.py#L575), [tree.py:675](backend/infrahub/graphql/queries/diff/tree.py#L675) — *"ensure any ongoing diff updates complete before we try to retrieve them"*). So the holder is often a reader that saved nothing, and `diff_repo.get_one` raises `ResourceNotFoundError`. On a fresh proposed change there is no earlier diff to fall back on, so the mutation 404s — in the job I traced, in 240 ms, having never waited for anything.

### When it broke

The premise was true when the shortcut was written. [#4263](https://github.com/opsmill/infrahub/pull/4263) (2024-09-05) added it with the same unguarded `get_one`, and at that point only another updater could hold the incremental lock. [#8268](https://github.com/opsmill/infrahub/pull/8268) (2026-01-30) then gave the same lock to `DiffTree` and `DiffTreeSummary` to fix a disappearing-diff bug, which is where readers started holding it. So the shortcut has been misreading the lock since January, which fits the flake history below.

## The fix

Move the shortcut into `_get_diff_stored_by_lock_holder`, which returns `None` when nothing is stored. The caller then falls through and calculates the diff itself, exactly as it would have if the lock had been free. This mirrors how [`recalculate`](backend/infrahub/core/diff/coordinator.py#L329) already handles a diff root removed by a concurrent update.

The helper releases the lock before returning, so the fall-through re-acquires it cleanly rather than relying on the lock's reentrancy.

**Deliberately out of scope:** the sibling `get_one` in the *"merged or rebased while waiting for lock"* path can raise the same way, but there the branch point has moved — calculating with the `from_time` of the stale branch object would produce a wrong diff, so a fall-through is not the right answer there. It is still a live failure mode though: that path falls back to the last stored diff, and on a branch rebased before any diff was stored there is nothing to fall back to, so it 404s with the same message. Tracked in #10521.

## Also fixes a CI flake

`tests/e2e/proposed-changes/test_proposed_changes_diff.py::test_should_verify_the_diff_data_with_conflicts` has been failing on unrelated PRs (10246, 10399, 10476, 10479 — recovered on retry 3 of 4) with `Locator expected to be visible` on the "Diff updated!" toast. The toast never appeared because the mutation had already errored; the test was watching for success while the app rendered the error path.

Worth noting for anyone who saw the earlier triage: bumping the Playwright expect timeout would not have helped at all, since the mutation fails in a fraction of a second.

## Test

`test_diff_update_computes_when_lock_holder_saved_no_diff` in `test_coordinator_lock.py` reproduces it directly — a task holds the incremental lock the way a read query does, while an update runs against a branch with no stored diff. It asserts the shortcut really was taken (`get_one` awaited once) so the test cannot pass without exercising the race.

Verified red before the fix, with the production error:

```
infrahub.exceptions.ResourceNotFoundError: Cannot find diff for branch branch_1 with tracking_id branch.branch_1
backend/infrahub/core/diff/repository/repository.py:247
```

Green after, along with the pre-existing suites:

| suite | result |
|---|---|
| `test_coordinator_lock.py` | 8 passed |
| `test_coordinator.py` + `test_refresh_diff.py` + `test_diff_task.py` | 26 passed |
| `backend/tests/unit/core/diff` | 15 passed |
| `ruff check . --exclude python_sdk`, `ruff format`, `mypy` | clean |

The e2e test itself needs the full stack, so CI validates that one here.

## Notes

- Targets `stable` per `dev/guidelines/git-workflow.md`: the bug is in released code and `coordinator.py` is byte-identical on `stable` and `develop`.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/diff/coordinator.py
```diff
@@ -168,6 +168,35 @@ async def _link_diff_to_proposed_change(
...
+        base_branch: Branch,
+        diff_branch: Branch,
+        tracking_id: BranchTrackingId,
+        proposed_change_id: str | None,
+    ) -> EnrichedDiffRootMetadata | None:
+        """Wait for the holder of the incremental lock and return the diff it stored, if any.
+
+        A held lock is no promise that a diff was saved: the DiffTree and DiffTreeSummary queries
+        take this lock purely to wait out an in-flight update. ``None`` means there is nothing to
+        reuse and the caller has to calculate the diff itself.
```
ajtmccarty: don't think you need to mention DiffTree and DiffTreeSummary

ajtmccarty (approved):

backend/infrahub/core/diff/coordinator.py
saltas888: Right, and it turns out code-doc-style.md already bans it, naming call sites rots when code moves. Dropped both names, the docstring just says readers take the same lock now.

While in there I trimmed the rest of the comments the same way, the test docstring had a paragraph of bug history that the changelog fragment already covers, the handshake comment argued for itself over a sleep, and two comments just restated the asserts under them. a4e2bdad4.

PR #10510, opened 2026-09-03
Send the Infrahub release dispatch to every repository listening for it

BeArchiTek:
## Summary

When Infrahub publishes a release, this workflow tells downstream repositories so they can open their own version-bump pull request. The matrix names three demo repositories. **Nine** repositories actually carry a listener for `trigger-infrahub-update`, so six of them have never been told about a release: their bump workflows exist, resolve `client_payload.version` correctly, and have simply never fired.

This adds the six missing recipients.

## Key Changes

Added to the matrix:

- `opsmill/infrahub-ansible`
- `opsmill/infrahub-arista-avd`
- `opsmill/infrahub-demo-otn`
- `opsmill/infrahub-demo-service-catalog`
- `opsmill/infrahub-mcp`
- `opsmill/nornir-infrahub`

Unchanged: the three already-wired demo repositories and the three secret-named slots.

`opsmill/infrahub-demo-dc-fabric` also has a listener but is **archived**, so it is deliberately absent.

## Base branch

Targeted at `stable` rather than `develop` on purpose: `release.yml` fires on `release: published`, so the file that executes is the one at the release tag, and the bot already syncs `stable` into `develop`. A change landed only on `develop` would never reach the branch that runs it. Please correct me if that is wrong for this repo.

## Test Plan

Nothing to run: the change is a matrix list. Verification is on the next Infrahub release, where each newly added repository should show a run of its own bump workflow shortly after the release publishes.

A dispatch to a repository with no matching listener is a no-op rather than an error, so there is no risk in the ordering here.


BeArchiTek:
Closing this unmerged: the dispatch system is being replaced with bot-triggered fan-out, which engineering is piloting, so hand-reconciling the matrix is work that would be thrown away. Letting it drift is the right call.

Recording the survey here so the replacement does not have to redo it. As of 2026-09-03, these repositories carry an `on: repository_dispatch: types: [trigger-infrahub-update]` listener:

| Repository | In this matrix today |
|---|---|
| `infrahub-demo-dc` | yes |
| `infrahub-demo-sp` | yes |
| `infrahub-solution-ai-dc` | yes |
| `infrahub-private` | via `INFRAHUB_ENTERPRISE_REPOSITORY` (inferred) |
| `packer-infrahub-instruqt` | via `INFRAHUB_PACKER_REPOSITORY` (inferred) |
| `infrahub-ansible` | **no** |
| `infrahub-arista-avd` | **no** |
| `infrahub-demo-otn` | **no** |
| `infrahub-demo-service-catalog` | **no** |
| `infrahub-mcp` | **no** |
| `nornir-infrahub` | **no** |

`infrahub-demo-dc-fabric` also has a listener but is archived. `INFRAHUB_CUSTOMER1_REPOSITORY` resolves to a secret I cannot read, so it may cover one of the above.

Two things worth carrying into the replacement design, because they are properties of the current arrangement rather than of this PR:

1. **`infrahub-github-bot/config.yaml` has no `on_release: infrahub` entry at all.** Its `version_fanout` covers `infrahub-sdk-python` and `opsmill-skills` only, so this workflow is currently the sole path for Infrahub releases. The bot needs that entry before it can take over here.
2. **The drift is silent in both directions.** A listener with no sender never fires and reports nothing; a sender with no listener dispatches into the void, also silently. Whatever replaces this should make one side derivable from the other rather than keeping two hand-maintained lists in sync.

On the cubic-dev-ai review: the finding was correct. The `gh api search/code` query I put in the comment matches this repository too, since the sender's `event-type:` line contains the same string, so a naive diff would report `opsmill/infrahub` as a missing listener. Moot now that the comment is not landing, but worth knowing if that query gets reused.

BeArchiTek:
**Correction: disregard my previous comment and the close.** I misread the instruction. The matrix additions are wanted; what was asked for was dropping the drift-detection tooling, not the reconciliation itself. Reopened, and the branch now carries only the added repositories.

Removed since the earlier review:

- The `gh api search/code` reconciliation query and the surrounding explanation. The original `# Either a literal path, or the name of a secret...` comment is restored unchanged, so the diff is now purely the new matrix entries.
- This also resolves the cubic-dev-ai finding, which was correct: that query matched this repository too, because the sender's own `event-type:` line contains the same string, so a naive diff would have reported the sender as a missing listener. The query is gone, so the problem is gone with it.

The branch was force-pushed to amend the single commit rather than stack a revert on top, because the original commit message described the query that no longer exists.

Longer term this is superseded by bot-triggered fan-out, which engineering is piloting. This PR is the interim step that gets the six unwired repositories receiving release notifications in the meantime.

fatih-acar (approved):

PR #10520, opened 2026-09-04
Report each side of a hierarchy edge in the path query

gmazoyer:
# Why

A hierarchical node declares `parent` and `children` under one identifier, `parent__child`. The traversal queries resolved each end of a hop with the first schema match, which is always `parent`. Both ends of a hierarchy edge then reported `parent`, and the labels carried the same error. The parent end holds `children`. Both `InfrahubPathTraversal` and `InfrahubReachableNodes` were affected: they share the hop projection.

This is the same inversion as #10485 (fixed by #10489), this time in the traversal queries. Fixes IFC-3106.

## What changed

- The two ends of a hop are resolved as a pair. Candidates on each end are narrowed by peer kind, paired by mirrored direction (a hard constraint), and the pairs are ranked by how many of their peers name the other end's exact kind. The mirror rule lives in `RelationshipSchema.mirrors()`, shared with the changelog fix from #10489.
- The pick is exact for every hierarchy that pins its `parent` and `children` kinds, including two-level ones whose outer peers fall back to the generic. Hierarchies with a peer left on the generic get a deterministic guess: the schema alone cannot tell their ends apart. IFC-3134 tracks the real fix, carrying the edge orientation in the traversal result.
- Each (identifier, from kind, to kind) triple resolves once per request. A pick is silent only when a single pair mirrors or a single pair pins both peer kinds; anything less logs a warning naming the kept pair and the alternatives, and another warning fires when no candidate pair mirrors at all.
- Identifiers with a single declaration on each side keep their behavior.

## How to test

```bash
uv run pytest backend/tests/unit/graphql/queries/test_path.py
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/graphql/queries/test_path_traversal.py backend/tests/component/graphql/queries/test_reachable_nodes.py --neo4j
```

The new component tests fail on the parent commit: the parent end reports `parent`/`Parent` instead of `children`/`Children`. Dependency mode has its own hierarchy-hop assertion.

## Checklist

- [x] Tests added
- [x] Changelog entry added
- [x] Docs updated (`docs/docs/reference/graph-traversal.mdx`: which end each field names, and when the pick is exact)

### DIFF
--- a/backend/infrahub/core/schema/relationship_schema.py
+++ b/backend/infrahub/core/schema/relationship_schema.py
@@ -60,6 +60,10 @@ class RelationshipSchema(GeneratedRelationshipSchema):
             raise ValueError("RelationshipSchema is not initialized")
         return self.identifier
 
+    def mirrors(self, other: RelationshipSchema) -> bool:
+        """True when the two declarations can be the two ends of one edge."""
+        return self.direction.neighbor_direction == other.direction
+
     def get_id(self) -> str:
         if not self.id:
             raise InitializationError("The relationship schema has not been saved yet and doesn't have an id")
--- a/backend/infrahub/core/schema/relationship_schema.py
+++ b/backend/infrahub/core/schema/relationship_schema.py
@@ -61,8 +61,8 @@ class RelationshipSchema(GeneratedRelationshipSchema):
         return self.identifier
 
     def mirrors(self, other: RelationshipSchema) -> bool:
-        """True when the two declarations can be the two ends of one edge."""
-        return self.direction.neighbor_direction == other.direction
+        """True when the two declarations can be the two ends of one edge: one identifier, mirrored directions."""
+        return self.get_identifier() == other.get_identifier() and self.direction.neighbor_direction == other.direction
 
     def get_id(self) -> str:
         if not self.id:
--- a/backend/infrahub/graphql/queries/path.py
+++ b/backend/infrahub/graphql/queries/path.py
@@ -226,9 +226,11 @@ def select_hop_relationships(
 
     A hierarchy declares ``parent`` and ``children`` under one identifier, one per end of an
     edge, so the two ends must mirror each other's direction. Candidates on each end are
-    narrowed by peer kind first, then paired by mirrored direction. When several mirrored
-    pairs remain (a self-referential hierarchy), the first pair is a guess between
-    schema-equivalent sides; the schema alone cannot tell the ends apart.
+    narrowed by peer kind first, then paired by mirrored direction. The pick is exact only
+    when narrowing leaves a single mirrored pair. When several pairs remain — both ends
+    declare both sides with peers that cover each other, as with peers defaulting to the
+    hierarchy generic — the schema alone cannot tell the ends apart and the first pair is
+    kept, a deterministic guess.
     """
     from_candidates = (
         _candidate_relationships(schema=from_schema, identifier=identifier, other_kind=to_kind, other_schema=to_schema)
@@ -243,12 +245,7 @@ def select_hop_relationships(
         else []
     )
 
-    pairs = [
-        (from_rel, to_rel)
-        for from_rel in from_candidates
-        for to_rel in to_candidates
-        if from_rel.direction.neighbor_direction == to_rel.direction
-    ]
+    pairs = [(from_rel, to_rel) for from_rel in from_candidates for to_rel in to_candidates if from_rel.mirrors(to_rel)]
     if not pairs:
         return (
             from_candidates[0] if from_candidates else None,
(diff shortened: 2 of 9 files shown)


docs/docs/reference/graph-traversal.mdx
```diff
@@ -35,7 +35,9 @@ Reference for the `InfrahubPathTraversal` and `InfrahubReachableNodes` GraphQL q
 Each `PathResultType` has `depth` (number of edges) and `hops` — an ordered list where each hop
 has a `node` and (except the first) a `relationship`. Each `node` (`PathNodeType`) exposes `id`,
 `kind`, `label`, `display_label`, and `hfid`. Each `relationship` (`PathRelationshipType`) exposes
-`from_rel`, `from_label`, `to_rel`, `to_label`, and `kind`.
+`from_rel`, `from_label`, `to_rel`, `to_label`, and `kind`. `from_rel` and `from_label` name the
+relationship the hop's source node holds; `to_rel` and `to_label` name the destination node's own
```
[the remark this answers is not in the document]
gmazoyer: Valid, the rule is in docs/AGENTS.md. Changed both mentions to object.

backend/infrahub/core/schema/relationship_schema.py
```diff
@@ -60,6 +60,10 @@ def get_identifier(self) -> str:
             raise ValueError("RelationshipSchema is not initialized")
         return self.identifier
 
+    def mirrors(self, other: RelationshipSchema) -> bool:
+        """True when the two declarations can be the two ends of one edge."""
```
polmichel: Should `mirror` also include the check on
`peer_schema.get_relationships_by_identifier(id=rel_schema.get_identifier())`? I think it'd make sense based on the `mirrors` method docstring, and all the data is contained inside `rel_schema`.

backend/infrahub/graphql/queries/path.py
```diff
@@ -192,67 +195,149 @@ def _node_payload(node_id: str, kind: str, labels_map: dict[str, dict[str, Any]]
     }
 
 
+def _get_schema_or_none(graphql_context: GraphqlContext, kind: str) -> MainSchemaTypes | None:
+    try:
+        return graphql_context.db.schema.get(name=kind, branch=graphql_context.branch, duplicate=False)
+    except SchemaNotFoundError:
+        return None
```
polmichel, line 202: This helper takes a `GraphQLContext` and a `kind` as input, digs into the `GraphQLContext` to find the `schema`, and looks it up there. It's useful because we don't have to worry about `SchemaNotFoundError` being raised. However, reaching deep into an object usually means the method isn't defined at the right level, and it doesn't feel intuitive to me to have a helper retrieving a schema from a `GraphQLContext` and a `kind`.
It looks like we might need a `get_or_none` helper at the `SchemaBranch` level, which is also an object exposed by the `DatabaseSchemaManager`.

This is a refactoring, so maybe is that we should do it later/on `develop` branch as this would be a scope creep. If you agree it'd make sense, I'll create a tech debt item for it.

I asked Claude to give me the different locations where this could be useful 

  **Where it belongs** (both already carry `has()` = the same `try get except → bool` shape):
  - `SchemaManager.get_or_none` — `manager.py`, next to `has()` at :78 (`get` at :85)
  - `SchemaBranch.get_or_none` — `schema_branch.py`, next to `has()` at :551 (`get` at :382)

  ### Direct replacements (except → None)
  | Site | Layer |
  |------|-------|
  | `graphql/queries/path.py:198` `_get_schema_or_none` (this PR) | Manager |
  | `core/validators/determiner.py:81` `_get_schema_or_none` (existing dup) | Branch |
  | `core/validators/uniqueness/deduplicator.py:100` `_get_schema_or_none` (existing dup) | Branch |
  | `core/changelog/diff.py:76` (`schema = None` on except) | Manager |
  | `graphql/analyzer.py:689` (`infrahub_model = None` on except) | Branch |

  ### Get-then-skip loops (`get_or_none(...)` + `if x is None: continue`)
  | Site | Layer |
  |------|-------|
  | `pools/registration.py:12` | Manager |
  | `core/diff/query_parser.py:579` | Manager |
  | `core/diff/enricher/hierarchy.py:47` | Manager |

  ### Borderline (non-None fallback, still cleaner)
  | Site | Note |
  |------|------|
  | `graphql/queries/path.py:178` `_schema_label_for_kind` | falls back to the `kind` string → `s = get_or_none(...); return kind if s is None else (s.label or kind)` |

backend/infrahub/core/schema/relationship_schema.py
polmichel: This would require some adaptation on the callers' side since there's the candidates concept further down in the PR, which only exercises the identifier criterion. That said, double-checking the identifier shouldn't require much computation, and the criteria would then be defined exhaustively.

polmichel (approved):
LGTM, just posted some optional comments. One of them, if it makes sense, should not be included in this PR in my opinion.

backend/infrahub/core/schema/relationship_schema.py
gmazoyer: Yes, done. `mirrors` compares the identifier too now. No change was needed on the callers: they build their candidates from `get_relationships_by_identifier`, so the identifier is already equal there.

backend/infrahub/graphql/queries/path.py
gmazoyer: Agreed on both points. I would still keep it out of this PR as it needs to touch several places in the code base. But it is worth a tech debt item.

gmazoyer:
Merged to stable as 334ab196fc. The merge API returned a 502 and GitHub did not record the merge, so I close this by hand. The change is in stable and nothing is missing.

backend/infrahub/core/schema/relationship_schema.py
polmichel: https://opsmill.atlassian.net/browse/IFC-3158

PR #10532, opened 2026-09-04
fix(backend): include branch failure reason in repository sync error

### LINKED ISSUE #10526: bug: a repository sync failure reports only the branch name, so the reason never reaches the operator
### Component

Git Integration

### Infrahub version

1.11.0 — present since 1.9.9

### Current Behavior

When a git repository sync fails for one branch, the flow fails with:

```
infrahub.exceptions.RepositoryError: Unable to synchronize the following branches of repository <repo>: <branch>
```

The message names the branch and nothing else. The reason is known internally — `FailedImport` carries `branch_name`, `step` and `reason` — but two things combine to make it unreachable from the UI:

1. `raise_if_branches_failed` builds the message from branch names only and discards `reason` ([`backend/infrahub/git/repository.py:118-141`](https://github.com/opsmill/infrahub/blob/infrahub-v1.11.0/backend/infrahub/git/repository.py#L118-L141)).
2. The `log.warning` that does carry `reason` uses the module-level `infrahub` logger (`repository.py:32`). Prefect only collects loggers in `DEFAULT_TASK_LOGGERS = ["infrahub.tasks"]` plus `workflow.extra_loggers` ([`backend/infrahub/workers/infrahub_async.py:47`](https://github.com/opsmill/infrahub/blob/infrahub-v1.11.0/backend/infrahub/workers/infrahub_async.py#L47)), so the record goes to worker stdout and nowhere else.

The result in the Tasks tab is a failed `Sync git repo with origin` run showing full progress logs — `Successfully parsed .infrahub.yml`, `Found N Python transforms`, `schema '…' loaded successfully!`, all emitted from `integrator.py`, which does use Prefect's run logger — and then jumping straight to the traceback. Neither `step` nor `reason` appears anywhere in the UI, so it is not even possible to tell whether the branch failed in the git collection phase or in the object import phase.

An operator without shell access to the task worker container cannot determine the cause of the failure at all.

### Expected Behavior

The raised error states why each branch failed, and the reason is visible in the Tasks tab without requiring container access to the task worker.

### Steps to Reproduce

1. Integrate a git repository with an Infrahub branch other than the default.
2. On that branch, commit a change that makes object import fail — e.g. a schema file under `schemas/` that fails validation.
3. Wait for a `git_repositories_sync` tick (cron `* * * * *`).
4. In the UI, open the repository **on `main`** → Tasks tab → the failed `Sync git repo with origin` run.
5. Observe that the error names only the branch. Neither `step` nor `reason` appears in the log.
6. Read the task worker container's stdout and grep for `Failed to synchronize branch` — the `step` and `reason` fields are present there.

### Additional Information

**Regression.** Introduced in #9466 (`01ca0f5f4`, 2026-06-19), first released in `infrahub-v1.9.9`. Isolating per-branch failures was the right change; before it, `sync()` let the original exception propagate, so the cause was visible. Still present on `stable` and at `infrahub-v1.11.0`.

**The logger choice looks unintentional.** `dev/knowledge/backend/async-tasks.md` documents the convention:

> - **Inside a `@flow` or `@task` body**, use Prefect's `get_run_logger()`.
> - **In a plain helper** that runs inside a flow but is not itself decorated — so it has no Prefect run context, and is typically also called directly from tests — use `infrahub.log.get_run_logger()`. It returns the `infrahub.tasks` stdlib logger, which the worker registers with Prefect and which is safe to call with no run context.

`raise_if_branches_failed` is an undecorated helper called from `RepositorySyncer.sync`, which runs inside the `sync-git-repo-with-origin` flow in production and is also called directly from tests (`backend/tests/component/git/test_git_repository.py:580`, `test_sync_lock_scope.py:28`, `backend/tests/integration/message_bus/operations/request/test_proposed_change.py:104`). That is exactly the second bullet, but `repository.py` contains no `get_run_logger` at all (0 occurrences) and uses `infrahub.log.get_logger()` instead.

**Blast radius.** Ten call sites in `repository.py` are invisible in the Tasks tab for the same reason, including three warnings and two errors:

```
84   log.info("Created new repository locally.")
129  log.warning("Failed to synchronize branch, skipping it.")   <-- this issue
157  log.info("Starting the synchronization.")
166  log.debug("New Branches …, Updated Branches …")
233  log.warning(…)
279  log.warning(…)
296  log.debug(…)
381  log.info("Created new repository locally.")
417  log.error("No object found for refs …")
453  log.error("No object found for ref …")
```

**Note for whoever picks this up:** the current calls use structlog keyword style (`log.warning(msg, repository=…, branch=…)`). `infrahub.log.get_run_logger()` returns a plain `logging.Logger`, which rejects arbitrary keywords, so the structured fields would need handling. The one existing precedent is `backend/infrahub/telemetry/utils.py:30`.

**Operator workaround.** Setting `INFRAHUB_WORKFLOW_EXTRA_LOGGERS='["infrahub"]'` on the task worker adds the logger to the allowlist and surfaces the warning in the Tasks tab. Noisy across the board, so it is a diagnostic setting rather than a default.

pthmas:
## Summary

Repository sync failures only named the failing branch in the error message — the
step and reason were discarded, and the log warning carrying them used a logger
Prefect does not collect. Operators without shell access to the task worker had no
way to determine why a branch sync failed from the Tasks tab alone.

## Key Changes

- `raise_if_branches_failed` now includes the failing step (`collection` or `import`)
  and reason per branch in the `RepositoryError` message, making them visible in the
  Tasks tab traceback.
- `repository.py` switches from `get_logger()` (structlog, `infrahub` namespace) to
  `get_run_logger()` (`infrahub.tasks`), routing all log records through the logger
  Prefect collects — fixing visibility for all ten call sites in the file.
- The warning in `raise_if_branches_failed` also passes `extra=` so `step` and
  `reason` survive as discrete `LogRecord` fields for log-pipeline filters and alert
  rules.

## Related Context

Closes #10526. Regression introduced in #9466 (`01ca0f5f4`, 2026-06-19), first
released in `infrahub-v1.9.9`. Still present on `stable` at `infrahub-v1.11.0`.

## Test Plan

- `uv run pytest backend/tests/unit/git/test_git_repository.py` — all unit tests pass.
- Component test `test_sync_continues_after_branch_pull_failure` updated to match the
  new message format (repo name + branch + `step=collection` pinned; variable reason
  left unmatched per test guidelines).

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tests/unit/git/test_git_repository.py
+++ b/backend/tests/unit/git/test_git_repository.py
@@ -283,9 +283,7 @@ class RaiseBranchesCase:
         RaiseBranchesCase(
             name="single_failure",
             failed_imports=[
-                FailedImport(
-                    branch_name="branch01", step=ImportStep.COLLECTION, reason="schema validation failed"
-                ),
+                FailedImport(branch_name="branch01", step=ImportStep.COLLECTION, reason="schema validation failed"),
             ],
             expected_message=(
                 "Unable to synchronize the following branches of repository test-repo:"
--- a/backend/tests/unit/git/test_git_repository.py
+++ b/backend/tests/unit/git/test_git_repository.py
@@ -315,9 +315,8 @@ def test_raise_if_branches_failed_error_message(case: RaiseBranchesCase) -> None
 def test_raise_if_branches_failed_logs_structured_fields(caplog: pytest.LogCaptureFixture) -> None:
     repo = InfrahubRepository(id=UUIDT.new(), name="test-repo")
     failed = FailedImport(branch_name="branch01", step=ImportStep.COLLECTION, reason="schema validation failed")
-    with caplog.at_level(logging.WARNING, logger="infrahub.tasks"):
-        with pytest.raises(RepositoryError):
-            repo.raise_if_branches_failed([failed])
+    with caplog.at_level(logging.WARNING, logger="infrahub.tasks"), pytest.raises(RepositoryError):
+        repo.raise_if_branches_failed([failed])
     assert len(caplog.records) == 1
     attrs = vars(caplog.records[0])
     assert attrs["branch"] == "branch01"
--- a/backend/tests/unit/git/test_git_repository.py
+++ b/backend/tests/unit/git/test_git_repository.py
@@ -265,9 +265,25 @@ def test_check_connectivity_ignores_cwd_git_pointer(tmp_path: Path, monkeypatch:
     InfrahubRepository.check_connectivity(name="test", url=f"file://{source_dir}")
 
 
-def test_raise_if_branches_failed_empty_list_does_not_raise() -> None:
-    repo = InfrahubRepository(id=UUIDT.new(), name="test-repo")
-    repo.raise_if_branches_failed([])
+@pytest.fixture
+def stub_repo() -> InfrahubRepository:
+    # Spell out all fields that carry positional defaults in Field() so mypy sees them.
+    return InfrahubRepository(
+        id=UUIDT.new(),
+        name="test-repo",
+        default_branch_name=None,
+        location=None,
+        has_origin=False,
+        cache_repo=None,
+        is_read_only=False,
+        internal_status="active",
+        reinitialized=False,
+        infrahub_branch_name=None,
+    )
+
+
+def test_raise_if_branches_failed_empty_list_does_not_raise(stub_repo: InfrahubRepository) -> None:
+    stub_repo.raise_if_branches_failed([])
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


backend/tests/unit/git/test_git_repository.py
```diff
@@ -259,3 +263,77 @@ def test_check_connectivity_ignores_cwd_git_pointer(tmp_path: Path, monkeypatch:
...
+        has_origin=False,
+        cache_repo=None,
+        is_read_only=False,
+        internal_status="active",
+        reinitialized=False,
+        infrahub_branch_name=None,
+    )
+
+
+def test_raise_if_branches_failed_empty_list_does_not_raise(stub_repo: InfrahubRepository) -> None:
```
polmichel: An optional comment: 
To increase the readability of this test and the related test suite, I would group this one with the others as a parameterized test.

```
  @dataclass
  class RaiseBranchesCase:
      name: str
      failed_imports: list[FailedImport]
      expectation: AbstractContextManager[Any]
```

```
  @pytest.mark.parametrize(
      "case",
      [
          RaiseBranchesCase(
              name="empty_list_does_not_raise",
              failed_imports=[],
              expectation=does_not_raise(),
          ),
          RaiseBranchesCase(
              name="single_failure",
              failed_imports=[
                  FailedImport(branch_name="branch01", step=ImportStep.COLLECTION, reason="schema validation failed"),
              ],
              expectation=pytest.raises(
                  RepositoryError,
                  match=rf"^{
                      re.escape(
                          'Unable to synchronize the following branches of repository test-repo:'
                          ' branch01 (step=collection): schema validation failed'
                      )
                  }$",
              ),
          ),
[......]
```

```
  def test_raise_if_branches_failed(stub_repo: InfrahubRepository, case: RaiseBranchesCase) -> None:
      with case.expectation:
          stub_repo.raise_if_branches_failed(case.failed_imports)
```

This has to be imported `from contextlib import nullcontext as does_not_raise`

polmichel (approved):
Thanks for the fix, LGTM. Just an optional comment about test readability

backend/tests/unit/git/test_git_repository.py
pthmas: Good point! I will make the changes

PR #10533, opened 2026-09-04
Follow-up on cubic review feedback from #10524

ajtmccarty:
Follow-up to the cubic review on #10524, covering the threads that were answered there
with "will address in a followup" plus the ones left unreplied. Threads that were
explicitly declined on that PR are not revisited here.

Frontend and CI/workflow findings from that review are deliberately **not** in this PR.

## Behaviour

One functional change: an unauthenticated context carries an empty account id, which
`BranchDeleteOrchestrator.delete` passed through to the branch-agnostic retirement query
and recorded verbatim as `to_user_id` on every edge it closed. It now falls back to
`SYSTEM_USER_ID`, matching `core/branch/tasks.py`. `to_user_id` is not exposed in GraphQL
or REST, so there is no changelog fragment.

## Docstrings and comments

- A zero closed-edge count was documented as proof that every field is still retained.
  The query also returns zero when no candidate matches and when the result is empty.
- `agnostic_retention.py`'s module docstring narrated its own Cypher construction instead
  of the retention contract.
- The repair query claimed a candidate with no derivable stamp is left alone; it falls
  back to the run time.
- The two repair-migration test modules narrated the approach not taken, which
  `.agents/rules/code-doc-style.md` keeps out of source.

## Tests

- `relationship_metadata` read `results[0]` without checking the count, so a duplicate
  relationship vertex would silently send later assertions at the wrong one. Its
  attribute sibling already asserts a single result.
- The rollback helper closed any open global edge regardless of when it opened, producing
  an inverted interval for an `at` earlier than the edge's own `from`.
- The test for another writer's closure never checked that the closure had landed, so it
  could pass without exercising anything.
- The field-removal tests re-implemented `create_widget` / `create_gadget` from the helper
  module they already import (18 call sites).

## Documentation

Two user-doc pages stated flatly that deleting an object does not release a
branch-agnostic value. Where no other branch can reach the object, the delete releases it
in the same operation — which the deeper section of `branch-agnostic-data.mdx` already
said.

The spec set under `dev/specs/ifc-2843-retire-agnostic-edges/` still directed a reader at
`m076` / `GRAPH_VERSION` 75 → 76 and at test paths that do not exist. The migration
shipped as `m078` with the bump 77 → 78, and every suite for the feature is
component-tier — there are no unit tests, so the quickstart says so rather than naming a
path that was never created.

Three of those corrections are worth calling out because they change what the artifacts
assert, not just a number:

1. **FR-015(3) and contract C4** required a candidate with no derivable stamp to be
   *skipped*. `queries.py` computes `coalesce(derived_at, $at)` and closes the edge. Both
   are now amended to the shipped behaviour, with the argument for why FR-014 still holds
   there recorded alongside: a candidate with neither an owner departure nor a closed
   owning edge on record is one no branch resolves as live, so that close shifts no
   branch's view. **If the requirement is the thing you want kept, this is a code change
   instead** — drop the coalesce and filter out the null case.
2. **The critique's Recommended Action #1** read as contradicting its own E1 correction,
   which recorded the in-query `(:Branch)` read as dropped in favour of `registry.branch`.
   The shipped predicate reads `MATCH (branch:Branch)` in-query; there is no
   `registry.branch` read and no migration-side `Branch.get_list`. So the action is what
   landed and the correction is the stale half — recorded as an Outcome note, with
   `data-model.md` fixed where it still named the registry as the runtime source.
3. **The same-UUID edge case** credited the candidate anchor with protection FR-011
   records as coming from the retention predicate, and **acceptance scenario 12** still
   asserted the premise SC-007 disavowed, so it held without the feature.

## Test status

`ruff check . --exclude python_sdk`, `ruff format` and `mypy` are clean on everything
touched. `markdownlint-cli2` is clean on the changed docs; the pre-existing errors under
`dev/specs/` are unchanged and outside CI's glob. The `delete_coordinator` unit suite
passes (7 tests, one new).

**Component tests were not run locally** — no Docker daemon available in the dev
environment. All 60 tests across the affected suites collect cleanly, so the imports and
the builder swap are sound, but the assertion changes in `test_rollback.py` and
`test_agnostic_field_removal.py` are verified only by CI.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/branch/delete_coordinator.py
+++ b/backend/infrahub/core/branch/delete_coordinator.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Protocol
 
+from infrahub.core.constants import SYSTEM_USER_ID
 from infrahub.events.branch_action import BranchDeletedEvent
 from infrahub.events.models import EventMeta
 from infrahub.exceptions import ValidationError
@@ -68,7 +69,7 @@ class BranchDeleteOrchestrator:
         # Freezing has to precede the deletion, which takes away the branch name they are found by.
         await self.diff_freezer.freeze_diffs_for_branch(branch_name=branch.name)
 
-        result = await self.data_deleter.delete(branch=branch, user_id=context.account.account_id)
+        result = await self.data_deleter.delete(branch=branch, user_id=context.account.account_id or SYSTEM_USER_ID)
 
         if result.branch_deleted:
             await self.workflow.submit_workflow(
--- a/backend/infrahub/core/migrations/graph/m078_retire_agnostic_property_edges/migration.py
+++ b/backend/infrahub/core/migrations/graph/m078_retire_agnostic_property_edges/migration.py
@@ -33,9 +33,9 @@ class Migration078(ArbitraryMigration):
     Retention is judged across every branch, so a value any branch can still read is left open and
     released later by the runtime enforcement points. Each close is stamped with the time the field
     stopped being reachable rather than with the upgrade's own time, which keeps it out of the fork
-    window of every branch older than the upgrade. Where the graph records no such time it also
-    records no owner departure, so no branch resolves the owner as live and the upgrade's own time
-    closes the field without shifting any branch's view. A second run reports zero repairs.
+    window of every branch older than the upgrade. Where the graph records no such time the
+    upgrade's own time is the fallback, which shifts no branch's view either: a field only reaches
+    the stamp once no branch retains it. A second run reports zero repairs.
     """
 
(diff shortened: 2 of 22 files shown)


ajtmccarty:
can be merged if approved

polmichel (approved):

PR #10537, opened 2026-09-06
refactor(regeneration): field-level impact resolution as a DI component (IFC-3067)

polmichel:
## Summary

Makes field-level impact resolution a dependency-injected component and tidies the surrounding regeneration/merge seams. Structural refactor — behavior unchanged.

## Changes

- **`FieldLevelImpactResolver`** (`core/regeneration/impact.py`): `db` + `client` injected at construction, `resolve()` takes the per-call payload — replacing the `get_field_level_impacted_subscribers` free function that reached the database through a global. Built at each entry point (the merge factory and the artifact/generator validation flows) and delegated to.
- **Impacted-subscriber seam** (`selective_regen/impacted.py`): selectors depend on an `ImpactedSubscriberResolver` `Protocol` that `FieldLevelImpactResolver` satisfies directly; test doubles implement it structurally instead of subclassing the concrete resolver.
- **Dependent-node resolver seam**: `DependentNodeResolver` `Protocol` + `QueryDependentNodeResolver` implementation, aligned across `core/relationship/` and `core/validators/`.

## Validation

`/pre-ci` green — ruff, ty, mypy, generated-file / schema / generated-doc validation (no drift), frontend lint, 2554 backend unit tests.

Ref: IFC-3067

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10537?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/core/branch/tasks.py
+++ b/backend/infrahub/core/branch/tasks.py
@@ -577,7 +577,9 @@ async def _build_post_merge_regeneration_dispatcher(
     generator_output = GeneratorCascadeOutput(capturer=output_capturer)
     return PostMergeRegenerationDispatcher(
         workflow=get_workflow(),
-        planner=build_merge_selective_regeneration(client=get_client(), log=log, generator_output=generator_output),
+        planner=build_merge_selective_regeneration(
+            db=db, client=get_client(), log=log, generator_output=generator_output
+        ),
         summary_cache=DiffSummaryCache(
             cache=await get_cache(), serializer=DiffSummarySerializer(), key_namespace="branch_merge"
         ),
--- a/backend/infrahub/proposed_change/tasks.py
+++ b/backend/infrahub/proposed_change/tasks.py
@@ -755,7 +755,6 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
     log = get_run_logger()
     client = get_client()
     client.request_context = context.to_request_context()
-    impact_resolver = FieldLevelImpactResolver(db=await get_database(), client=client)
 
     artifact_definition = await client.get(
         kind=CoreArtifactDefinition,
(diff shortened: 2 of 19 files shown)


backend/tests/unit/core/merge/selective_regen/definition_selector/test_base.py
```diff
@@ -151,7 +152,10 @@ async def _fetch_member_ids(
         return self._member_ids
 
     def _should_render(self, *, subscriber_id: str | None, regenerate_all_members: bool, impacted: list[str]) -> bool:
-        return not subscriber_id or regenerate_all_members or subscriber_id in impacted
+        # Production code behavior
+        return run_generator(
+            instance_id=subscriber_id, regenerate_all_members=regenerate_all_members, impacted_instances=impacted
+        )
```
polmichel, line 158: Independent from this PR. 
Just a logic implementation inside a test stub I've noticed along the way.

backend/infrahub/core/branch/tasks.py
```diff
@@ -515,7 +515,9 @@ async def _build_post_merge_regeneration_dispatcher(
     generator_output = GeneratorCascadeOutput(capturer=output_capturer)
     return PostMergeRegenerationDispatcher(
         workflow=get_workflow(),
-        planner=build_merge_selective_regeneration(client=get_client(), log=log, generator_output=generator_output),
+        planner=build_merge_selective_regeneration(
+            db=db, client=get_client(), log=log, generator_output=generator_output
+        ),
```
gmazoyer, line 582: Did you mean to move the merge path onto the flow session here? Before, resolution called `get_database()` internally and got the driver singleton, which reuses one Neo4j session process-wide. Now it gets another session. It looks better to me, each merge follow-up gets its own session, but just wanted to check.

backend/infrahub/proposed_change/tasks.py
```diff
@@ -755,6 +755,7 @@ async def validate_artifacts_generation(model: RequestArtifactDefinitionCheck, c
     log = get_run_logger()
     client = get_client()
     client.request_context = context.to_request_context()
+    impact_resolver = FieldLevelImpactResolver(db=await get_database(), client=client)
```
gmazoyer: Should it be wrap in `.start_session()`? The four other `get_database()` sites in this file do, and `tests/unit/workflows/test_flow_session_convention.py` tells why. This is pre-existing, the old code did the same call inside the removed function, so the PR changes nothing here. Just wanted to ask as I'm not sure.

gmazoyer (approved):
Nice DI cleanup 👍

backend/infrahub/core/branch/tasks.py
polmichel: That wasn't the goal at first, but one of the good sides of DI is that it forces you to think about where the dependency should come from. I thought it was better to take the same `db` as the caller to share the same session or transaction. But I hadn't realized there was a real issue here before that.

backend/infrahub/proposed_change/tasks.py
polmichel: You're right, thanks
As the stacked PR is blocked due to component tests flakiness, I think it makes sense to change it right now in an additional commit.

PR #10540, opened 2026-09-07
fix(tests): restore workflow override when a fixture body raises

polmichel:
## What

Wrap the `config.OVERRIDE.workflow` swap in `try/finally` in the four
component-suite fixtures (`workflow_local` / `workflow_recorder` in
`tests/component/conftest.py`, and the two `workflow_local` fixtures in
`tests/helpers/test_app.py`), matching the existing `message_bus` and cache
fixtures.

## Why

Restoration ran only on the statement after `yield`, so a wrapped test that
raised or timed out left the override in place. Under xdist the leak was picked
up by every later class on the same worker, whose service was built against the
wrong workflow adapter and errored at setup with `AssertionError: These tests
are currently meant to run with a local worker` — the top
`backend-tests-component (other)` failure of the week (~30 job logs across ~18
PRs, several recovering on retry). The victim class shifted with scheduling
while the assertion stayed constant.

Verified locally with a throwaway test driving each fixture generator: the
override survives a raising body before the change and is restored after.

Scope is the leak fix only; the upstream timeout in
`merge_recompute_coalescing/test_bulk_recompute_writer.py` that can trigger a
skipped teardown is a separate issue.

### DIFF
--- a/backend/tests/component/conftest.py
+++ b/backend/tests/component/conftest.py
@@ -3035,9 +3035,11 @@ def workflow_local(dependency_provider: Provider) -> Generator[WorkflowLocalExec
     original = config.OVERRIDE.workflow
     workflow = WorkflowLocalExecution()
     config.OVERRIDE.workflow = workflow
-    with dependency_provider.scope(build_workflow, lambda: workflow):
-        yield workflow
-    config.OVERRIDE.workflow = original
+    try:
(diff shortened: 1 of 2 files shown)


polmichel:
Better solution here https://github.com/opsmill/infrahub/pull/10536

PR #10525, opened 2026-09-04
docs(knowledge): add git integration backend knowledge doc

ogenstad:
## Summary

Adds `dev/knowledge/backend/git-integration.md`, an internal knowledge doc covering how Infrahub models an external git repository and, in particular, which branch each operation actually targets.

Sections:

- The two independent meanings of "default branch" (`registry.default_branch` vs `CoreRepository.default_branch`).
- How the trunk is resolved on the repository object, and which construction paths pass it (the `get_initialized_repo` factory does not - the warm-path fallback that underlies most non-`main` bugs, tracked as IFC-2870).
- Per-worker storage and the broadcast/`RefreshGitFetch` convergence model.
- Push writeback direction and its lack of reconciliation (INFP-670).
- Branch-support semantics (LOCAL vs AGNOSTIC) and why LOCAL repository state never appears in a diff.
- Staging repositories, destructive delete cascades, git error classification, and a set of known limitations.

## Verification

Every `file:line` reference in the document was checked against the current `develop` code before writing. Four values in the initial draft were corrected in the process:

- `registry.default_branch` source: `config.py:801-804` (not `775-778`).
- Mismatched-default-branch log call sites: `git/repository.py:178,211` (the `validate_remote_branch` calls).
- Remote-delete gating setting name: `delete_branch_after_merge`.
- `RefreshGitFetch` emission coverage now lists all six sites (the read-only pull site was missing).

## Notes

- Internal `dev/knowledge/` documentation only, no user-visible change, so no changelog fragment.
- Targeting `develop` as a draft.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #10542, opened 2026-09-07
docs(specs): spec set for honouring the repository default branch [IFC-3105]

ogenstad:
## Already approved work

* Resolve the configured repository default branch once, at every construction [IFC-3105] (approved in #10664)

## What this is

Planning artifacts for IFC-3105 only. **No code, no schema change, no generated files.** This is PR 1
of four; the implementation lands in PRs 2-4 as broken down in `tasks.md`.

Opened as a draft to get review on the plan *before* the implementation PR, which is large and
atomic by necessity (~20 source files, ~25 test files, and it cannot be split without inventing a
decomposition the design does not have — the reasoning is in `tasks.md` under "Making PR 2
reviewable").

## The defect

A read-write repository's configured trunk (`CoreRepository.default_branch`) is read from the graph
only when a worker clones the repository for the first time. Every later construction of the per-flow
repository object on that worker omits the trunk and inherits a silent fallback to Infrahub's own
default branch:

```python
# git/base.py:191-192
@property
def default_branch(self) -> str:
    return self.default_branch_name or registry.default_branch
```

So a repository resolves its trunk correctly on a worker's first touch and incorrectly on every
subsequent one. Git storage is per worker, so the same repository works on one worker and fails on
another. Artifact generation, transforms, generators, computed attributes and proposed-change checks
all silently target the wrong branch, or fail with a missing-ref error naming a branch the operator
never configured.

Reported as #9600 and #8749 before IFC-2870 traced both to this single fallback. #9601 fixed one call
site, which is the evidence that per-call-site patching does not close the class.

## What the plan does

The trunk and the staging status become **required** constructor fields of the read-write object, the
fallback and the optional field are removed from the shared base, and both values are resolved in one
place — a resolver module read once per construction, on the Infrahub branch the operation runs on.
The flows, tasks and message models that used to carry the trunk stop doing so, so there is no second
entry point for an unset value.

On top of that baseline, two visibility improvements: connect-time rejection when the configured trunk
is absent from the remote (from the `ls-remote` the connectivity check already performs, no clone),
and a task-log warning naming each branch skipped for colliding with Infrahub's default branch.

## Where to look

Review these, in this order:

| File | What it is |
|---|---|
| `spec.md` | Four user stories, 11 functional requirements, 6 success criteria, edge cases, assumptions, scope |
| `plan.md` | Architecture, constitution check, project structure, complexity tracking, risks |
| `tasks.md` | 80 tasks, the four-PR breakdown, dependency order |
| `research.md` | Verified code state and decisions D1-D9, including test placement (D7) |
| `contracts/` | Constructor and factory signatures, connect-time validation, the sync task-log design |
| `data-model.md` | Field-level target state |
| `quickstart.md` | Manual validation per story, plus the support-diagnosis check |

`critiques/` holds three critique transcripts. They are **historical process records, not part of the
proposal** — they describe designs that were subsequently changed and should not be reviewed as
current. `checklists/requirements.md` is the running amendment log and explains what changed between
rounds and why.

## Decisions a reviewer should know were deliberate

Each of these was challenged during critique and settled, so they are not oversights:

- **The skipped-branch condition goes to the task log, not to new repository state.** The PRD asked
  for persistent state. A warning on every synchronisation cycle was rejected on volume: the collision
  is permanent by nature and the sync runs once a minute, so it would add roughly 1,440 linked tasks a
  day to the repository's task history, burying the failures that history exists to surface.
  Persistent state would fix that but crosses the GraphQL "Ask First" gate. **A dedicated status
  surface for this one condition was ruled overkill by the product owner on 2026-09-04.** The
  accepted limit is stated in the spec's Assumptions and in `quickstart.md`'s SC-005 table: on a
  repository where nothing moves at all, the condition becomes invisible once the connect-time task
  ages out of Prefect's retention.
- **The warning has two triggers, and one of them is per worker.** It is recorded at connect, and
  thereafter when a run either imported a branch or saw a new commit arrive on the skipped branch.
  The second trigger exists because the first was silent at the one moment the warning is actionable —
  when the operator has just pushed to the branch that is not being imported. It compares the
  skipped branch's remote head before and after the fetch, which needs nothing persisted (the fetch
  is unfiltered, so that ref is already on disk and its commit is already read). The cost: the
  comparison is per worker, so one push may be reported once per worker. That cannot be deduplicated
  without shared state, and it is bounded by worker count rather than by the cycle rate.
- **`operational_status` moves branch.** The factories now set the Infrahub branch on the object,
  which `_update_operational_status` reads. Callers that previously left it unset wrote the status on
  the platform default branch regardless of where they ran; those writes now land on the operation's
  branch. This is the intended shape — one branch, one meaning — but it changes where an existing
  status appears, so it is called out rather than left to be discovered.
- **No Playwright e2e test.** Backend-only change with no frontend work: the connect form already
  renders whatever validation error the API returns, and the node Tasks tab already lists node-tagged
  runs and renders their logs. Nothing new is built in the browser. The connect journey is covered
  end to end in the integration tier against Gogs, driving the same GraphQL mutation the form calls
  and asserting the exact operator-facing message.
- **SC-001's evidence is layered, not one outcome test per flow.** Because the trunk is resolved in
  one place and no call site can omit it, a fourth test of the resolution mechanism proves nothing,
  while the real residual risk is a call site passing the *wrong* Infrahub branch — which a
  warm-clone outcome test does not catch either. So proposed-change checks get their own functional
  case (their branch choice is the least obvious in the matrix) and the generator and
  computed-attribute flows are pinned at the branch they hand to the factory. The layering is stated
  in SC-001 rather than left implicit.
- **One redundant graph read is accepted.** The periodic sync already holds the repository node and
  today forwards its trunk into the factory. FR-003 forbids any flow from supplying the trunk, so the
  factory re-reads a node the caller has in memory. Recorded in Complexity Tracking so it is not
  reintroduced as an optimisation in review — every such carrier is a place the next caller can
  forget, which is how this defect arose.

## Governance

No "Ask First" gate is crossed. No schema attribute, migration, GraphQL or REST change. The only
generated file the implementation touches is `docs/docs/reference/message-bus-events.mdx`, regenerated
because two internal message models gain a field.

## Related

- Jira: [IFC-3105](https://opsmill.atlassian.net/browse/IFC-3105), part of
  [INFP-670](https://opsmill.atlassian.net/browse/INFP-670)
- PRD: Confluence Product page 858357761. **It still describes the superseded P3 design and declares
  the GraphQL gate CROSSED.** Amending it is a task in this feature (T071, four corrections). Where
  the PRD and this spec set disagree, the spec set is current.
- Supersedes IFC-2870. IFC-2740 is already resolved; only regression coverage remains.
- Coupled to #10525, which adds `dev/knowledge/backend/git-integration.md`. That page documents the
  fallback this feature deletes and carries a volatile section naming IFC-2870. Whichever of the two
  lands second reconciles; if this feature is first, #10525 needs a revision pass before it merges.
- `dev/specs/infp-546-git-solid-refactor` Story 6 planned an optional trunk parameter with a global
  fallback, which this supersedes with a required field. Correcting that spec is T070.

## Not in this PR

T001 (test baseline), T002 (the call-site worklists), T070 (the `infp-546` correction) and T072 (three
follow-ups to file) belong to PR 1's scope in `tasks.md` but are not here — this PR is the spec set,
so it can be reviewed now. T071 (the PRD amendment) rides with T073, where the same reasoning gets
written for PR 2's description.

[IFC-3105]: https://opsmill.atlassian.net/browse/IFC-3105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ogenstad:
## Cubic review triage

Seven findings. Two were real defects and are fixed; one was a question I've answered below; four are in `critiques/` and are addressed by scoping rather than by rewriting. Detail:

### Fixed

**1. `ensure_branch_exists` could not raise its own documented error** (`connect-time-trunk-validation.md:66`) — valid, and worse than reported. `RepositoryInvalidBranchError.__init__(identifier, branch_name, location, message=None)` takes `location` as a **required positional parameter**, so the documented signature would have raised `TypeError` at the one call site that matters. Cubic's proposed fix was right: `location` is now a fourth parameter sourced from `message.repository_location`, the same value already passed to `list_remote_refs`. Updated in the contract, `data-model.md`, T046 and T050, with a note on why `RemoteRefs` holds no URL (it models what the remote reported, not how it was reached).

**2. The "57 / 22" count doesn't reproduce** (`critique-20260904-094130.md:57`) — valid, and it had propagated into the live plan. Re-measured on the current tree with an explicit criterion:

| Criterion | Sites | Files |
|---|---|---|
| `InfrahubRepository(` / `InfrahubReadOnlyRepository(` / `.init(` / `.new(` | **64** | **23** |
| `.init(` / `.new(` only | 61 | 21 |
| Direct instantiation only | 3 | 2 |

Cubic's 61 matches the second row exactly, which is the point it was making: the number is meaningless without the expression. `plan.md` now carries the reproducing grep beside the figure and says why a narrower pattern gives a smaller answer. Corrected in three places in `tasks.md`. The file count of 23 was already right.

**3. Documentation tree omits `critiques/`** (`plan.md:75`) — valid, fixed. Also added the two files T001 and T002 create, which were missing for the same reason.

### Answered

**4. Docs-only change targets `develop`** (`spec.md:1`) — deliberate, and thank you for asking rather than asserting. `dev/guidelines/git-workflow.md:19-20` does send a diff with no runtime source changes to `stable`. The exception here is the one cubic guessed: PRs 2-4 contain runtime source and are therefore develop-bound, and a spec set on `stable` would be separated from the implementation it describes, with every later PR that touches `tasks.md` or `checklists/` waiting on a forward merge first. The spec travels with its implementation.

### Scoped rather than rewritten

**5, 6, 7 — internal inconsistencies in `critiques/`** (E7's split status, P1's contradictory status, the unreproducible verification row). All three are real. None is edited, because these are dated transcripts of individual critique rounds and revising them to be self-consistent would destroy the record they exist to keep — including the record of a verification that got a number wrong, which is useful evidence about how much weight those tables can carry.

Instead each file now opens with a banner stating that it is a historical record, naming its known artefacts explicitly (including the two status contradictions and the bad count), and pointing at `spec.md`/`plan.md`/`tasks.md` for current design and `checklists/requirements.md` for the amendment log. The PR description already said `critiques/` isn't part of the proposal; that instruction now lives in the files themselves, where a reader will actually meet it.

### Declined

**`research.md` line-number citations** — I checked all twelve of the cited locations against the current tree and every one still resolves correctly, so this is a forward-looking concern rather than a present error. It's also the wrong standard for the document: `research.md` is a dated Phase 0 verification record, and the line numbers *are* the evidence that a fact was checked at a specific commit. Stripping them would make the verification unreproducible without making anything more current.

Cubic is right that there was an inconsistency, though — the file's own header claimed `module::symbol` citation while the body cites lines, and T068a criticises PR #10525 for stale line numbers. `research.md` now states the distinction: a knowledge page is meant to stay true so stale lines there are a defect; `tasks.md` cites symbols only because code moves under a task; a Phase 0 snapshot is neither.

---

`invoke docs.lint`: 0 errors.

PR #10551, opened 2026-09-07
fix(backend): bind pagination bounds as query parameters

saltas888:
# Why

`Query.get_query()` wrote `SKIP`/`LIMIT` into the Cypher text as literals. Neo4j keys its query plan cache on the query text, so every page of a paginated query was a distinct statement and paging turned one query into one plan compilation per page.

Measured on `test_get_many_with_pagination` against local Neo4j, instrumenting `InfrahubDatabase.run_query`: the call phase made 207 round trips producing **198 distinct query texts** and spent **40.7s of its 42.3s** in the database, 38.9s of that in `node_list_get_attribute` at 213 ms average. Running each page a second time separates compilation from data access: **219 ms average cold, 8 ms average warm** over the same 183 pages. The cost was planning, not reading.

Goal: one text, and therefore one cached plan, per query regardless of page.

## What changed

- `Query.render()` now returns a `RenderedQuery` (text plus params) rather than rendering into `self.params`. The bounds are emitted as `SKIP $query_offset` / `LIMIT $query_limit` with the values in the returned params, so rendering has no side effect on the query object. `get_query()`, `execute()` and `query_with_size_limit()` consume the rendered pair; `NodeCreateAllQuery`'s error path in `core/query/node.py` was the one call site that read `self.params` straight after `get_query()`.
- A zero offset is a bound, a zero limit is not. `offset=0` renders `SKIP $query_offset`; without that the first page rendered no `SKIP` and kept a plan of its own, which is the difference between 17 and 3 distinct texts on the test above. `limit=0` renders no clause at all, because GraphQL `limit: 0` means unbounded today: `test_ipaddress_include_available_pagination[0-2]` sends `limit: 0, offset: 2` and expects all 8 remaining rows. Reading zero as a bound on both sides returned none of them.
- `PAGINATION_LIMIT_PARAM` / `PAGINATION_OFFSET_PARAM` reserve the two names for queries that keep automatic pagination. A query with `insert_limit = False` renders no clauses and no parameters, and every subclass that binds its own bounds already sets that flag, so nothing collides.
- `execute()` still selects the auto-paging path on truthiness (`if self.limit or self.offset`). That is deliberate: making it `is not None` would change what a GraphQL `limit: 0` returns today.
- `dev/knowledge/backend/query-pattern.md` documents the parameters and the reserved names.

## How to test

- `uv run pytest backend/tests/unit/core/test_query.py` — 9 new tests covering the parameter binding, that rendering leaves `self.params` untouched, that every page shares one text, that a zero offset renders as a bound while a zero limit does not, that `insert_limit = False` renders neither, and the `var`/`inline` rendering paths. They fail on literal rendering.
- `uv run pytest backend/tests/unit/core/test_query.py backend/tests/component/core/test_query.py` — 21 passed against Neo4j.
- Same instrumentation on this branch: the call phase's 207 round trips now produce **3 distinct texts** (the three queries the test actually issues) and **2.08s in the database, down from 40.7s**, with `node_list_get_attribute` at 8 ms average — the warm figure from the cold/warm split above, now paid on every page.

## Impact & rollout

- **Performance:** any paged read over a large result set stops paying for a plan per page, and stops evicting other queries' plans from a size-limited cache while doing so. Deep pagination is where it shows.
- **These seconds are local.** Local Neo4j latency inflates the absolute timings; read the ratios and the distinct-text counts as the result, not the wall clock, and don't extrapolate them into a CI or production figure.
- **Backward compatibility:** none needed. The output is plain parameterised Cypher, with no schema or API change. Memgraph's CI jobs are commented out in `ci.yml`, so nothing exercises that adapter either way.
- Changelog fragment added.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/query/__init__.py
```diff
@@ -532,11 +536,15 @@ def get_query(
...
+        # one text, and with it one cached plan.
         if offset and self.insert_limit:
-            tmp_query_lines.append(f"SKIP {offset}")
+            self.params[PAGINATION_OFFSET_PARAM] = offset
+            tmp_query_lines.append(f"SKIP ${PAGINATION_OFFSET_PARAM}")
 
         if limit and self.insert_limit:
-            tmp_query_lines.append(f"LIMIT {limit}")
+            self.params[PAGINATION_LIMIT_PARAM] = limit
+            tmp_query_lines.append(f"LIMIT ${PAGINATION_LIMIT_PARAM}")
```
ogenstad, line 550: As per cubics comment I wonder if these lines should instead read `if offset is not None and self.insert_limit:` and `if limit is not None and self.insert_limit`.

backend/tests/unit/core/test_query.py
```diff
@@ -0,0 +1,63 @@
...
+
+    assert first_page == second_page
+    assert paged_query.params["query_offset"] == 200
+
+
+def test_first_page_has_no_skip_clause(paged_query: PagedQuery) -> None:
+    text = paged_query.get_query(limit=100, offset=0)
+
+    assert "SKIP" not in text
+    assert "query_offset" not in paged_query.params
```
ogenstad: To highlight the point from above this would also render a different query than the follow up queries.

ogenstad (commented):
Great find! Looks stupid when you see it like this. Looks good to me but I added a comment as I think there's some additional improvements we could make here.

fatih-acar:
Good catch! We had this on the graphql layer, but we forgot the db one... :man_facepalming:

backend/infrahub/core/query/__init__.py
ogenstad: For the AI bot, did we consider this review comment just above?

saltas888: Yes, both now test `is not None`, and the `or` fallbacks on `self.limit` / `self.offset` follow the same rule so an explicit zero counts as a bound. The first page now renders `SKIP $query_offset` with 0 and shares the text of every page after it, the paginated `get_many` test went from 17 distinct query texts to 3. The one behaviour change is an explicit `limit=0` rendering `LIMIT 0` instead of dropping the clause, nothing in the backend passes that. `execute()` still picks the auto-paging path on truthiness, I left that alone since changing it would alter what a GraphQL `limit: 0` returns today.

backend/tests/unit/core/test_query.py
saltas888: Replaced that test, it now asserts the first page shares the text of the pages after it, plus one pinning that zero bounds render as bounds.

backend/infrahub/core/query/__init__.py
```diff
@@ -514,38 +526,45 @@ def add_subquery(self, subquery: str, node_alias: str, with_clause: str | None =
...
             tmp_query_lines.append("RETURN " + ",".join(self.return_labels))
 
         if self.order_by:
             tmp_query_lines.append("ORDER BY " + ",".join(self.order_by))
 
-        if offset and self.insert_limit:
-            tmp_query_lines.append(f"SKIP {offset}")
+        # Bound as parameters rather than literals, zero included, so every page of a paginated
+        # query shares one text, and with it one cached plan.
+        if offset is not None and self.insert_limit:
```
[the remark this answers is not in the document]
saltas888: Valid, the description was written two commits before the zero-bounds fix. Rewrote it: it now says zero counts as a bound and that the first page renders `SKIP $query_offset` with 0 so every page shares one text. Same audit caught two more stale claims from the `render()` refactor, that the values go into `self.params` and that no call site changed, both fixed, and the figures are re-measured on this HEAD (3 distinct texts, 2.08s in the database against 40.7s).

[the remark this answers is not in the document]
saltas888: On the other two suggestions in the run summary: the first-page pagination tests already exist (`test_first_page_shares_the_text_of_the_pages_after_it` and `test_zero_bounds_are_bounds_not_absence` in `backend/tests/unit/core/test_query.py`). Adjusting the condition back is the one thing I would not do here, since emitting `SKIP $query_offset` at offset 0 is the fix @ogenstad asked for, not a regression against it. Omitting the clause on the first page gives that page a query text of its own and so a cached plan of its own, which is where the 17 distinct texts came from; binding zero takes it to 3. `SKIP 0` is a no-op for results, so this changes the text shape only, and `test_query_results_limit_offset` covers the behaviour.

backend/infrahub/core/query/__init__.py
```diff
@@ -514,38 +526,46 @@ def add_subquery(self, subquery: str, node_alias: str, with_clause: str | None =
...
-        if offset and self.insert_limit:
-            tmp_query_lines.append(f"SKIP {offset}")
+        # Bound as parameters rather than literals, a zero offset included, so every page of a
+        # paginated query shares one text, and with it one cached plan. A zero limit is no bound,
+        # the same reading `execute()` gives it when choosing how to run.
+        if offset is not None and self.insert_limit:
+            params[PAGINATION_OFFSET_PARAM] = offset
+            tmp_query_lines.append(f"SKIP ${PAGINATION_OFFSET_PARAM}")
 
         if limit and self.insert_limit:
```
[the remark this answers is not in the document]
saltas888: Not taking this one, it would reintroduce a bug the suite already catches. With `limit=0` treated as a bound, `render()` emits `LIMIT 0`, while `execute()` reads a zero limit as unset and so takes the single-page path whenever an offset is set, and the query comes back empty. `test_ipaddress_include_available_pagination[0-2]` in `backend/tests/component/graphql/queries/test_ipam.py` sends GraphQL `limit: 0, offset: 2` and expects the 8 remaining rows, since `limit: 0` means unbounded to callers today.

I applied your suggestion locally rather than assume: `1 failed, 5 passed`, the failure being `[0-2-result3]`. With the guard as it stands all 6 pass.

The asymmetry is deliberate, and commented in place. A zero offset is a bound, since `SKIP 0` is a no-op and rendering it keeps the first page on the same cached plan as the rest, which is the whole point of the PR. A zero limit is not a bound, since dropping the clause is what `limit: 0` already means. `test_zero_limit_is_no_bound` pins that, so the next pass at this line fails fast instead of shipping.

backend/infrahub/core/query/__init__.py
saltas888: Correcting what I wrote above: my claim that nothing in the backend passes `limit=0` was wrong. `test_ipaddress_include_available_pagination[0-2]` sends GraphQL `limit: 0, offset: 2` and expects all 8 remaining rows, so a zero limit means unbounded to callers, and reading it as a bound returned none of them. I had grepped for `limit=0` constructions in `backend/infrahub` and the SDK, which was the wrong search, the value arrives as user input.

Fixed in c399f61. A zero offset stays a bound, so the first page still shares its text with the pages after it, which was your point and still holds. A zero limit renders no clause, matching how `execute()` already reads it. `test_zero_limit_is_no_bound` covers the rendering and the 6 ipam pagination cases pass.

ogenstad (approved):

Issue #5430, opened 2025-01-10
bug: GraphQL errors when combining queries


ogenstad:
### Component

API Server / GraphQL

### Infrahub version

v1.2.0a0

### Current Behavior

When you submit a GraphQL query that uses the same model twice the result should be combined.

For example if you have a query that looks like this:

```graphql
query Tags {
  BuiltinTag {
    count
  }
  BuiltinTag {
    edges {
      node {
        display_label
        name {
          value
        }
      }
    }
  }
}
```

The response can look like this:

```json
{
  "data": {
    "BuiltinTag": {
      "count": 1,
      "edges": []
    }
  }
}
```

Here the expectation would be that we also have an entry under "edges".

If we switch the order of the query so that it instead looks like this:

```graphql
query Tags {
  BuiltinTag {
    edges {
      node {
        display_label
        name {
          value
        }
      }
    }
  }
  BuiltinTag {
    count
  }
}
```

Then we get an error in the response:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot return null for non-nullable field PaginatedBuiltinTag.count.",
      "locations": [
        {
          "line": 13,
          "column": 5
        }
      ],
      "path": [
        "BuiltinTag",
        "count"
      ]
    }
  ]
}
```

Based on the previous response we'd be trying to send back a `null` value for "count" which is correctly no longer allowed. The error in the console looks like this:

```python
2025-01-10T11:33:01.485923Z [critical ] Unhandled exception occurred in resolvers [infrahub.graphql] app=infrahub.api request_id=76a5b2aebbf844e399503c836aaefd59 worker=21a421d9-4d51-4b4c-a146-3192291ac002
Traceback (most recent call last):
  File "/Users/patrick/.virtualenvs/infrahub/lib/python3.12/site-packages/graphql/execution/execute.py", line 529, in await_result
    completed = self.complete_value(
                ^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrick/.virtualenvs/infrahub/lib/python3.12/site-packages/graphql/execution/execute.py", line 622, in complete_value
    raise TypeError(
TypeError: Cannot return null for non-nullable field PaginatedBuiltinTag.count.
```

### Expected Behavior

The resolver should be smart enough to know that the result should be combined into one.

### Steps to Reproduce

Follow the above steps.

### Additional Information

While not completely related I noticed a similar error in the server logs in one of the e2e tests that looked like this:

```python
[...]

PR #10566, opened 2026-09-09
feat: publish the repository git-state GraphQL contract [IFC-3146]

ogenstad:
Jira: [IFC-3146](https://opsmill.atlassian.net/browse/IFC-3146) · Feature: [IFC-3101](https://opsmill.atlassian.net/browse/IFC-3101)

Targets the feature branch `pog-repo-commit-visibility-ifc-3101`, not `develop`. This is PR 2 of the eleven in `dev/specs/ifc-3101-repo-commit-visibility/tasks.md`; the `develop` catch-up belongs on the feature branch once the sub-PRs have landed.

## What this publishes

The whole read surface for a repository's git state, at once — answering everything Infrahub can answer from the graph, and reporting an honest unavailable state for everything that needs a worker's clone.

- **`InfrahubRepositoryCommits`** — paged, newest-first commit log for the request branch, carrying the branch, the remote branch or tracked ref it resolves, and the commit Infrahub has imported. `limit` bounded to 1..100, `offset` to `>= 0`, and `pending_count` requested only when selected.
- **`InfrahubRepositoryBranchDrift`** — answers with the repository id and its own column-level unavailable state. Rows need a per-branch tracked commit, which arrives with the per-branch graph query in PR 3.
- **`InfrahubReadOnlyRepositoryCheckRefs`** — submits the on-demand refs check.

Both queries enforce repository view permission imperatively on the concrete kind, because `InfrahubGraphQLQueryAnalyzer` maps top-level fields to kinds by exact name and cannot see a custom query. The mutation gates on update permission for the read-only kind and refuses a read-write repository.

**Why this is the hand-off point.** Nothing in the frontend reads a live server schema — it types against the checked-in `schema/schema.graphql` through gql.tada and graphql-codegen. The moment that file carries the new SDL, the frontend team is unblocked, whether or not any server can yet answer with commits. `UNAVAILABLE / NOT_IMPLEMENTED` is a state the UI has to handle regardless.

Nothing here writes to the graph: no schema change, no migration, and no write to `commit` or `ref` anywhere.

## Structure

`RepositoryGitStateReader` is a protocol with one dependency-free implementation, chosen in a single factory. Swapping it for the worker-backed reader in PR 5 is that factory's body and nothing else — no resolver, test or consumer change. The recording and failing doubles under `backend/tests/helpers/` are its other two implementations.

The remote-branch mapping rule moves out of `InfrahubRepositoryBase._get_mapped_remote_branch` into `infrahub.git.branch_mapping` as a pure function with all three inputs required, and the method delegates to it with no behaviour change. The API server cannot call a private instance method and must never build a repository object; restating the rule in the resolver would leave two copies to drift on precisely the default-branch case INFP-670 is fixing.

Three `WorkflowDefinition`s are registered with bodies that log and return — the repository warm-up, the read-only refs-check cycle, and its per-repository run. Their real bodies come with PRs 5 and 8.

## Decisions a reviewer would otherwise query

All of these are recorded against their task in `tasks.md`.

- **`config.OVERRIDE.repository_git_state_reader`** is the test injection seam, checked first in the factory alongside the existing `message_bus` / `cache` / `workflow` entries. It is what lets the component tests inject doubles with no patching.
- **The drift resolver does not call the reader** — it short-circuits to its own `NOT_IMPLEMENTED` column state. T014's text says to delegate; T048 requires the short-circuit, and T048 is right: with no per-branch tracked commit until PR 3, a request built from the request branch's values would carry the wrong commit for every other branch, and `BusRepositoryGitStateReader.branch_heads` does not exist until PR 10.
- **`FailingRepositoryGitStateReader` takes the exception to raise as a constructor argument** rather than hard-coding `WorkerTimeoutError`, which arrives with PR 4.
- **The mutation's payload type prints as `ReadOnlyRepositoryCheckRefs`**, following `ReadOnlyRepositoryImportLastCommit`, rather than the `InfrahubReadOnlyRepositoryCheckRefs` the contract sketches. The field on `Mutation` keeps the `Infrahub` prefix.
- **The mutation submits `{"repository_id": ...}`** for now; the `GitReadOnlyRepositoryCheckRefs` model and the tracked refs it carries belong to PR 8.
- **The changelog fragment carries the `+` orphan prefix.** Without it towncrier renders the filename as a GitHub issue link, so a bare `ifc-3101-….added.md` would ship a link to an issue that does not exist.
- **`git/state/cache_keys.py` defines two TTL constants, not three.** The due key's TTL is the configured interval and the in-flight ceiling is the per-repository timeout plus a margin — both derived from values PR 8 introduces.

## Governance

The additive GraphQL surface was signed off on 2026-09-04 (recorded in `plan.md`). The `git/branch_mapping.py` extraction is an explicit override of the PRD's "without refactoring it" and lands on a file INFP-670 restructures; approved on 2026-09-09 by the owner of both this feature and that epic, with the `default_branch` property's `or registry.default_branch` fallback deliberately left in place to stay clear of that epic's P1.

## Verification

Green: `invoke format`, `invoke lint` (ruff + ty + mypy + yamllint), `ruff check . --exclude python_sdk`, 2506 backend unit tests, the 27 new component tests, `docs.validate`, `docs.lint`, `schema.validate-graphqlschema`, `schema.validate-jsonschema`, `backend.validate-generated`, `uv lock --check` on both projects, frontend `biome ci` / `knip` / `betterer ci`, and `codegen:graphql` with a clean diff.

Phase A of `quickstart.md` was walked against a live API server and `infrahubasync` worker from this branch. Steps 2, 3 (branch-switch half) and 5 hold; step 4's expectation was wrong and is amended in the same commit — the denial matches querying `CoreRepository` directly in code, status and the permission named, but not in wording, because these resolvers check one permission while the analyzer batches every kind and denies in the plural. A step 6 was added recording that the paging refusals surface as `UNDEFINED_ERROR` / 422, since `ValidationError` has no catalogue entry anywhere in the codebase.

## Known-unrelated CI noise

`backend/tests/component/git/test_git_repository.py::test_compare_remote_local_new_branches` and `::test_get_filtered_remote_branches__no_import_sync_branch_names` fail locally on this branch. They are pre-existing: both reproduce unchanged with the pre-change `backend/infrahub/git/base.py` restored, and the only edit here on that file is a behaviour-preserving delegation off the `get_branches_from_remote` code path. Not introduced by this PR, and not fixed by it either.

[IFC-3146]: https://opsmill.atlassian.net/browse/IFC-3146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-3101]: https://opsmill.atlassian.net/browse/IFC-3101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/graphql/queries/repository_git_state.py
```diff
@@ -0,0 +1,226 @@
...
+        root: dict,  # noqa: ARG004
+        info: GraphQLResolveInfo,
+        repository_id: str,
+        limit: int = DEFAULT_LIMIT,
+        offset: int = 0,
+    ) -> dict[str, Any]:
+        graphql_context: GraphqlContext = info.context
+        branch = graphql_context.branch
+
+        _validate_paging(limit=limit, offset=offset)
```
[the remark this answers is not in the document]
ogenstad: Valid, and worse than described: it was a 500 with the raw `TypeError` text reaching the client, on both a null variable and an inline literal null. Confirmed against a running server before fixing.

Fixed in 30b80182e by normalising rather than rejecting. `_resolve_paging` applies the documented defaults before bounding, so null and omitted mean the same thing — which matches the contract, where both arguments are nullable `Int` with a documented default. `DEFAULT_OFFSET` also replaces a bare `0` that had been repeated in two places. The SDL is unchanged, since the resolver's Python defaults were never part of it.

Covered by `test_explicit_null_paging_falls_back_to_the_documented_defaults`, which asserts the recorded request carries limit 10 and offset 0.

changelog/+ifc-3101-repository-commits-api.added.md
```diff
@@ -0,0 +1 @@
+Added the GraphQL API for repository git visibility: InfrahubRepositoryCommits returns a paged commit log for a repository on the branch you are viewing, InfrahubRepositoryBranchDrift returns per-branch drift, and InfrahubReadOnlyRepositoryCheckRefs asks Infrahub to check a read-only repository's remote refs now. Both queries answer with the repository, branch, tracked ref and imported commit that Infrahub holds; the values read from a worker's clone report as unavailable until the read path ships in a later release.
```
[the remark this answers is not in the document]
ogenstad: Valid. Fixed in 30b80182e.

The fragment claimed "both queries answer with the repository, branch, tracked ref and imported commit" when the drift query returns none of those yet — it answers with the repository id, an unavailable column and no rows. Re-reading it for that, the same sentence also described `InfrahubReadOnlyRepositoryCheckRefs` as asking Infrahub to check the remote "now", when its flow body currently logs and returns, so I corrected that too.

It now states per operation what actually lands: the commit log answers with the branch, tracked ref and imported commit and reports the commit list as unavailable; the drift query reports unavailable with no rows; the refs check accepts a request and returns a task id without contacting the remote.

backend/infrahub/git/branch_mapping.py
```diff
@@ -0,0 +1,19 @@
...
+) -> str:
+    """Return the remote git branch an Infrahub branch reads from.
+
+    Infrahub's default branch reads the repository's configured default branch; every other branch
+    reads the remote branch of the same name.
+
+    Every input is required: the caller has to state which trunk it means, so a missing value cannot
+    silently resolve to Infrahub's default branch.
+    """
+    if branch_name != repository_default_branch and branch_name == infrahub_default_branch:
```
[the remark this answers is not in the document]
ogenstad: Valid. Added `backend/tests/unit/git/test_branch_mapping.py` in 30b80182e, parametrised over the four cases you name.

Worth recording why the gap mattered rather than just closing it: the component tests reach only two of the four — a default branch whose repository trunk differs (`main` → `trunk`) and an ordinary non-default branch. The two they never reach are the ones where the rule's two conditions coincide: a default branch that already matches the repository trunk, and a non-default branch whose name happens to equal the repository trunk. Both must read their own name, and both would regress silently if the condition were inverted.

dev/specs/ifc-3101-repo-commit-visibility/tasks.md
```diff
@@ -169,40 +169,40 @@ commits. `UNAVAILABLE / NOT_IMPLEMENTED` is a state the UI has to handle regardl
...
+- [x] T014 Add the `InfrahubRepositoryBranchDrift` resolver to the same module: resolve the row set from `registry.branch` (branches with `sync_with_git` for the read-write kind, every branch for the read-only kind, excluding `MERGED`, `DELETING` and the global branch), delegate to the reader, and return the column-level unavailable state with empty edges until block 2.5 supplies the rows. **Landed without the reader call and without the row set**, which is the half of this task that contradicted T048: with no per-branch tracked commit until T021/T023, a request built from the request branch's values would carry the wrong commit for every other branch, and the bus reader has no `branch_heads` implementation until T079 anyway. So the resolver returns the repository id and its own `UNAVAILABLE / NOT_IMPLEMENTED` column state with no rows, which is exactly the short-circuit T048 relies on. T023 brings the row set in with the query that can populate it, and T079 removes the short-circuit
```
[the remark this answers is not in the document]
ogenstad: Valid, and the right way round: the checkpoint was wrong, not T014.

Checkpoint 2A promised real values from "both queries", which never held for the drift query in PR 2 — a drift row needs the per-branch tracked commit that block 2.5 supplies, and the request branch's value would be wrong for every other branch. Fixed in 30b80182e: 2A now scopes that promise to `InfrahubRepositoryCommits`, states what the drift query does answer with, and points a reader at checkpoint 2B for the rows.

Worth flagging the neighbouring inconsistency this came from, since it will reach whoever picks up PR 5: T014's own text says to "delegate to the reader", while T048 says the drift resolver "never reaches the reader, so a failing reader proves nothing about it here". They cannot both hold. T048 is the one that is right — `BusRepositoryGitStateReader.branch_heads` does not exist until T079, so between PR 5 and PR 10 a delegating drift resolver would call an unimplemented method. The resolver short-circuits, and the deviation is recorded against T014 in `tasks.md`.

backend/infrahub/git/state/factory.py
```diff
@@ -0,0 +1,21 @@
...
+
+from infrahub import config
+
+from .reader import UnavailableRepositoryGitStateReader
+
+if TYPE_CHECKING:
+    from .reader import RepositoryGitStateReader
+
+
+def build_repository_git_state_reader() -> RepositoryGitStateReader:
```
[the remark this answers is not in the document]
ogenstad: Done in 8ada08fb2, and right for the reason you gave. The parameter is back on the factory and the resolver passes `graphql_context.active_service.message_bus` again.

Worth recording why it had been removed, because that justification does not hold. It was taken out to avoid evaluating `active_service` on a GraphQL context carrying no service, which raises `InitializationError`. No path reaches this resolver that way: the websocket endpoint rejects every operation that is not a subscription (`graphql/app.py:483-484`), the two other service-less callers of `prepare_graphql_params` run their own fixed query documents, and every test supplies a service. `active_service` expresses the invariant that a service is always passed rather than a case to defend against.

The decisive argument is one neither of us cited. T008 in `dev/specs/ifc-3101-repo-commit-visibility/tasks.md`, the task that created this factory, specified exactly this arrangement up front, unused parameter and `noqa` included:

> It takes the message bus as a required parameter from the start, unused until T046 and carrying a `noqa: ARG001` until then, so that swap really is one body and no call site.

So the removal contradicted the task, and your comment caught it. The parameter is read at T046, which changes the body to return `BusRepositoryGitStateReader` built with the bus and `config.SETTINGS.broker.rpc_timeout`; the `noqa` comes off there.

One correction to the comment itself: deferring would not have forced "another consumer-wide change". There is exactly one caller, in `graphql/queries/repository_git_state.py`, so it would have been a one-line change at one site. That does not change the outcome, since T008 settles it either way.

backend/infrahub/graphql/queries/repository_git_state.py
```diff
@@ -0,0 +1,308 @@
...
+            "unavailable": _unavailable_payload(result=None, reason=RepositoryGitUnavailableReason.NOT_IMPLEMENTED),
+        }
+
+
+InfrahubRepositoryCommits = Field(
+    RepositoryCommits,
+    repository_id=String(required=True),
+    limit=Int(required=False, description="Page size, 1 to 100. Default 10."),
+    offset=Int(required=False, description="Default 0."),
+    description="Paged commit log for a repository on the request branch.",
```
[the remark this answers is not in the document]
ogenstad: Pushed an update

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,128 @@
...
+    Measured before any ancestry call: an ancestry test against an unresolvable hash raises
+    rather than answering.
+    """
+
+    imported_is_ancestor_of_head: bool | None = None
+
+    pending_count: int | None = None
+
+    tracked: bool = False
+    """False when nothing is imported or inherited on this branch."""
```
polmichel, line 63: Regarding this data model, `imported_is_ancestor_of_head` is essentially derived from `head`, `imported` and the real git repository state. The same goes for `tracked`, `imported_resolvable` and `pending_count`.

Duplicating these notions in the code can lead to the values getting out of sync and weird bugs in the long run. I'd say it's better from a maintainability perspective to rule out any impossible combinations and/or split these attributes into several objects and add a data validation layer.

But, I agree this important that we request everything in one query. I think that's possible.

From what I gather here:
- `head` and `imported` look like the real "Infrahub versus Git state".
- `tracked` <=> `imported is not None`
- `pending_count` is None <= `imported_is_ancestor_of_head` is False
- `imported_resolvable` is not None <=> `imported` is not None
- `imported_resolvable` is False => `imported_is_ancestor_of_head` is None && `pending_count` is None
- `imported_is_ancestor_of_head` is purely derived from the "Infrahub versus Git state".

This is almost a state machine in itself that we should decompose. I can see a solution using data validation as a `post_init` method or using 3 dataclasses encapsulating each other. 

What's your thoughts regarding this? I may have missed something as I've not read all the specifications.

polmichel (commented):
I've just partially reviewed the PR for now, I'll continue tomorrow. One comment posted.

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+    git_ref: str
+    tracked_commit: str | None = None
+
+
+@dataclass(frozen=True)
+class CommitLogRequest:
+    repository_id: str
+    repository_name: str
+    repository_kind: str
+    location: str
```
polmichel, line 81: These 4 attributes are reused inside `BranchHeadsRequest`. It looks like a `RepositoryIdentity` inner dataclass. Do you think it would be a good idea to group those and the lower ones under an dedicated object?

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+            ValueError: When only one of the two is set.
+
+        """
+        is_unavailable = self.condition is RepositoryGitCondition.UNAVAILABLE
+        if is_unavailable and self.unavailable_reason is None:
+            raise ValueError("A result with condition UNAVAILABLE must carry an unavailable_reason")
+        if not is_unavailable and self.unavailable_reason is not None:
+            raise ValueError(
+                f"A result carrying an unavailable_reason must have condition UNAVAILABLE, not {self.condition.name}"
+            )
```
polmichel, line 127: This is one of the proposed solution regarding `GitStateFacts`. It may be too complex since `GitStateFacts` has a lot of attributes. However, with good unit tests, it could be fine.

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+
+
+@dataclass(frozen=True)
+class CommitEntry:
+    hash: str
+    """Full commit hash."""
+
+    short_hash: str
+    """First 7 characters of the hash."""
```
polmichel, line 23: Could this be a property on the class, since it's simple and cheap to derive from `hash`?

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+
+    summary: str
+    """First line of the commit message."""
```
polmichel, line 26: Could this be a property on the class, since it's simple and cheap to derive from `message`?

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+    imported_commit: str | None = None
+    """The imported hash as the answering worker resolved it, when it resolved one."""
+
+    pending_count: int | None = None
+    commits: tuple[CommitEntry, ...] = ()
+    fetched_at: datetime | None = None
+    unavailable_reason: RepositoryGitUnavailableReason | None = None
+    warm_up_task_id: str | None = None
+    error_message: str | None = None
+    """Display-safe explanation, set whenever no git-derived answer was produced."""
```
polmichel, line 112: Could we have an error message filled here if a "git-derived answer is produced"? (Let's say `remote_head` is filled for instance).
If not, should this be checked in the `post_init` method?

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+    condition: RepositoryGitCondition
+    remote_head: str | None = None
+    imported_commit: str | None = None
+    """The imported hash as the answering worker resolved it, when it resolved one."""
+
+    pending_count: int | None = None
+    commits: tuple[CommitEntry, ...] = ()
+    fetched_at: datetime | None = None
+    unavailable_reason: RepositoryGitUnavailableReason | None = None
+    warm_up_task_id: str | None = None
```
polmichel, line 110: https://github.com/opsmill/infrahub/pull/10566/changes#r4013443493 Same as this comment. If there is a `warm_up_task_id`, the repository is not yet cloned right?

backend/infrahub/git/state/models.py
```diff
@@ -0,0 +1,146 @@
...
+    remote_head: str | None
+    condition: RepositoryGitCondition
+
+
+@dataclass(frozen=True)
+class BranchDriftResult:
+    branches: tuple[BranchDriftRow, ...] = ()
+    fetched_at: datetime | None = None
+    unavailable_reason: RepositoryGitUnavailableReason | None = None
+    warm_up_task_id: str | None = None
```
polmichel, line 144: I think the comments about CommitLogResult also make sense here https://github.com/opsmill/infrahub/pull/10566/changes#r4013977077 + https://github.com/opsmill/infrahub/pull/10566/changes#r4013443493

backend/infrahub/graphql/mutations/repository.py
```diff
@@ -275,6 +276,51 @@ async def mutate(
...
+            kind=InfrahubKind.GENERICREPOSITORY,
+            id=str(data.id),
+            branch=branch,
+        )
+
+        if repo.get_kind() != InfrahubKind.READONLYREPOSITORY:
+            raise ValidationError(
+                f"Node {data.id} is a {repo.get_kind()}, not a {InfrahubKind.READONLYREPOSITORY}. "
+                "Checking remote refs is only supported for read-only repositories."
+            )
```
polmichel, line 313: Is this ok from a permission perspective to return the kind of a node other than the one that was queried? Someone having only the permission to update the branch could make a lot of requests and gain a lot of information about the database state.

backend/infrahub/workflows/catalogue.py
```diff
@@ -231,6 +231,13 @@
     tags=[WorkflowTag.DATABASE_CHANGE],
 )
 
+GIT_REPOSITORY_WARM_UP = WorkflowDefinition(
+    name="git-repository-warm-up",
+    type=WorkflowType.USER,
```
polmichel, line 236: Is this a `USER` or `INTERNAL` workflow? `data-model.md` is stating `INTERNAL`

backend/infrahub/graphql/field_extractor.py
polmichel, line 70: Claude told me about some cases which are not supported in this bloc. I asked him to create unit tests for it. I'll post them in another PR so it's easier to review and know if these cases are meant to be supported, or not.

backend/infrahub/graphql/queries/repository_git_state.py
```diff
@@ -0,0 +1,308 @@
...
+            "imported_commit": imported_commit,
+            "edges": [],
+        }
+
+        fields = extract_graphql_fields(info=info)
+        if git_ref is None:
+            payload["condition"] = RepositoryGitCondition.NOT_TRACKED
+            return payload
+
+        if not COMMIT_GIT_FIELDS & set(fields):
```
polmichel, line 226: Does this work when using graphQL fragments?

backend/infrahub/graphql/field_extractor.py
polmichel: https://github.com/opsmill/infrahub/pull/10644/changes this PR lists the test cases Claude was mentioning. I think the best way moving forward is to close this PR if this is confirmed as "meant to be supported", and create a bug quoting these cases.

backend/infrahub/git/state/models.py
ogenstad: Agreed on the direction. `tracked` is gone (ee9fbe8ee): it restated `imported is not None`, and T037's classifier decides NOT_TRACKED from the imported commit without ever reading it. `__post_init__` now rejects the shapes the measurement order cannot produce (ee07f84648): resolvability without an imported commit, ancestry without a resolved commit or without a head, and a pending count without the ancestry it counts along.

Holding off on the decomposition until T037 exists, since that is where the constraints get exercised. One correction to your list: `pending_count is None` does not imply not-BEHIND. It is also None when the caller did not ask, since `include_pending_count` is a request field. Any split has to keep those two Nones distinguishable.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Right in principle, but declining for now. `BranchHeadsRequest` is never constructed anywhere: it appears only as an annotation on the protocol, the no-op reader and the test doubles. So the duplication this would remove is an artefact of a type landing ahead of its consumer, not something the code actually does twice.

Deferred to T042/T045, where both request types will have producers in front of whoever writes the bus messages. Changing the shape stays free until a released worker speaks it.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Took this one. Both result types and `GitStateFacts` validate in `__post_init__` now; the sibling threads say what each rejects.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Done in 4cc3b915d8. Property now, GraphQL surface unchanged.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Done in 4cc3b915d8. It uses `splitlines()` so a CRLF message does not carry a trailing carriage return into the answer, and an empty message summarises as empty rather than raising.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Added. Both result types reject an `error_message` set without an unavailable reason.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Yes, and validated: a warm-up id against any reason but NOT_CLONED is rejected. The field is also gone from the published schema (6b813a6752). The warm-up is an internal workflow, so its run carries no namespace tag and `InfrahubTask` cannot resolve the id, and the Commits tab polls the query rather than the task.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
ogenstad: Agreed, it has its own validator sharing the same check. Its invariants differ in one way worth noting: FR-022 means it carries rows *alongside* an unavailable column rather than instead of them.

(Posted by AI after being read by a human)

backend/infrahub/graphql/mutations/repository.py
ogenstad: Valid, and reproduced: a BuiltinTag id answered "Node &lt;id&gt; is a BuiltinTag". Fixed in ee9fbe8ee — the lookup asks for the read-only kind, and a mismatch now answers exactly as an id that exists nowhere, so neither can be told from the other.

The cause is that `get_one_by_id_or_default_filter` does not enforce the `kind` it is given: it forwards it only to the default-filter fallback. That is tracked in #10655, which also covers the same pattern in `ReadOnlyRepositoryImportLastCommit` and the other call sites.

(Posted by AI after being read by a human)

backend/infrahub/workflows/catalogue.py
ogenstad: You are right, reverted in 1da5777b4. It matches the split already in the catalogue: the sweep over all repositories is INTERNAL, the check a person triggers is USER. A warm-up is started by a read, not by a person.

(Posted by AI after being read by a human)

backend/infrahub/graphql/field_extractor.py
ogenstad: Confirmed all five cases, and they are now tracked in #10653 and fixed in #10654, which goes further and picks up #5430 as well.

None of it originates in this PR, so it is staying out of it. Worth knowing that #10654 and this PR overlap: both add `_merge_fields` and both fix the `field_nodes[0]` half, so whichever merges second will conflict in `field_extractor.py`, and this PR carries a changelog fragment for that half.

(Posted by AI after being read by a human)

backend/infrahub/graphql/queries/repository_git_state.py
ogenstad: Yes, verified both a named spread and an inline fragment. Pinned by `test_git_fields_reached_through_a_fragment_still_ask_the_reader`, which asserts the reader received the expected request rather than only checking the response shape.

(Posted by AI after being read by a human)

backend/infrahub/git/state/models.py
polmichel: Ok, noted for `pending_count`. LGTM!

polmichel (approved):
LGTM! thanks for creating the external issues

PR #10569, opened 2026-09-09
feat(changelog): enrich node changelogs delivered to webhooks with HFID and peer labels

polmichel:
Bottom of stacked set #10574 · base `pmi-webhook-payload-enrichment` · replaces #10562.

## Why

Node changelogs delivered to webhooks carry no stable identifier or human-readable peer context, so consumers can't reliably resolve which objects changed. This enriches them with each node's HFID and display label, plus its relationship peers' labels — for node mutations and for branch merge/rebase.

## What changed

- **Behavioral:** webhook node-changelog payloads now include the node's HFID and display label, and each relationship peer's display label and HFID.
- Batched node-label loader (query-size-limited; degrades to empty on read failure).
- Enrichment wired for node mutations and for merge/rebase diffs; a node whose kind was dropped by a migration is tolerated rather than failing the batch.
- **Unchanged:** no schema or GraphQL/API contract change.

## How to review

Four self-contained commits — reviewing commit-by-commit is the easiest path. Suggested order:

1. **Loading primitive** — `core/changelog/enrichment.py`: the `NodeLabelReader` protocol, the DB-backed reader, and the batched `NodeLabelLoader` (degrades to empty on read failure). The foundation both enrichment paths build on.
2. **Node-mutation path** — `core/changelog/models.py` (`RelationshipChangelogGetter`) + `graphql/mutations/relationship.py`: how a mutation's changelog and its reciprocal secondaries get enriched.
3. **Merge/rebase path** — `core/changelog/diff.py` (`DiffChangelogCollector`): turning a merge/rebase diff into enriched changelogs, including dropped-kind tolerance and external-peer resolution.
4. **Emission wiring** — `builder.py`, `events/generator.py`, `core/branch/tasks.py`, `core/merge/orchestrator.py`, `core/node/__init__.py`: where the enriched changelog is attached to delivered events (mechanical).
5. **Tests** — unit (`test_enrichment`, `test_diff`), then component + functional (`test_process`) for the end-to-end payload.

Concentrate scrutiny on the two enrichment entry points (steps 2–3).

## How to test

```bash
uv run pytest backend/tests/component/core/changelog -q   # needs Docker
```

## Known gap: the recompute of a lost collection

Collecting the node changelogs now reads the graph, so it can fail where it could not before. A merge
or a rebase that loses its collection still dispatches its events, but the changelogs are what scopes
its post-processing recompute.

A merge falls back to recomputing every computed attribute of the destination branch, on the low lane.
That pass does not cover the display labels and the human-friendly ids, which keep the value they
carry until their node is written again. A rebase gets no fallback: the branch-wide flows reach only
the triggers the rebased branch owns, which is its Python attributes when it carries a transform
commit of its own plus the Jinja2 ones whose template differs from the default branch's, so what a
widened pass would cover there depends on the branch rather than on the failure. It logs the skip.

Both gaps are recompute policy rather than changelog work, so they are left to a follow-up. A version
of the rebase fallback is written and kept on the branch `pmi-changelog-recompute-fallback`.
`dev/knowledge/backend/merge-recompute.md` records the behaviour.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry — TODO (user-facing payload change)

### DIFF
--- a/backend/infrahub/core/changelog/diff.py
+++ b/backend/infrahub/core/changelog/diff.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Sequence
 
@@ -19,6 +20,8 @@ from .models import (
 )
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from infrahub.core.branch import Branch
     from infrahub.core.diff.model.path import (
         EnrichedDiffAttribute,
@@ -30,6 +33,8 @@ if TYPE_CHECKING:
     from infrahub.core.models import SchemaUpdateMigrationInfo
     from infrahub.core.schema import MainSchemaTypes
     from infrahub.database import InfrahubDatabase
+
+    from .enrichment import NodeLabelLoader
 
 
 @dataclass
@@ -45,12 +50,15 @@ class DiffChangelogCollector:
         diff: EnrichedDiffRoot,
         branch: Branch,
         db: InfrahubDatabase,
+        label_loader: NodeLabelLoader,
         migration_tracker: MigrationTracker | None = None,
     ) -> None:
         self._diff = diff
         self._branch = branch
         self._db = db
+        self._label_loader = label_loader
         self._diff_nodes: dict[str, NodeInDiff]
+        self._node_hfids: dict[str, list[str] | None] = {}
         self.migration = migration_tracker or MigrationTracker()
 
     def _populate_diff_nodes(self) -> None:
@@ -66,12 +74,26 @@ class DiffChangelogCollector:
         try:
             return self.get_node(node_id=peer_id).kind
         except KeyError:
-            schema = self._db.schema.get(node_kind, branch=self._branch, duplicate=False)
+            try:
+                schema = self._db.schema.get(node_kind, branch=self._branch, duplicate=False)
+            except SchemaNotFoundError:
+                # The node's kind was dropped by a schema migration, so an unchanged peer's kind
+                # cannot be resolved from the schema; degrade as the attribute path does.
+                return "n/a"
             rel_schema = schema.get_relationship(name=relationship_name)
             return rel_schema.peer
 
+    def _peer_hfid(self, peer_id: str) -> list[str] | None:
+        """Return a peer's HFID when it is among the changed nodes already loaded for this diff.
+
+        A changed relationship is symmetric, so a peer referenced here is normally a changed node
+        itself and its HFID comes for free from the batch; a peer outside the diff keeps None.
+        """
+        return self._node_hfids.get(peer_id)
+
     def _process_node(self, node: EnrichedDiffNode) -> NodeChangelog:
         node_changelog = NodeChangelog(node_id=node.uuid, node_kind=node.kind, display_label=node.label)
+        node_changelog.hfid = self._node_hfids.get(node.uuid)
         try:
             schema = self._db.schema.get(node_changelog.node_kind, branch=self._branch, duplicate=False)
         except SchemaNotFoundError:
@@ -80,10 +102,30 @@ class DiffChangelogCollector:
         for attribute in node.attributes:
             self._process_node_attribute(node=node_changelog, attribute=attribute, schema=schema)
 
+        if node_changelog.hfid is None and node.action == DiffAction.REMOVED:
+            # A removed node is gone when the batch load runs, but the diff still records its HFID;
+            # recover it so removed nodes report it like directly-deleted ones.
+            node_changelog.hfid = self._hfid_from_diff(node_changelog)
+
         for relationship in node.relationships:
             self._process_node_relationship(node=node_changelog, relationship=relationship)
 
         return node_changelog
+
+    @staticmethod
+    def _hfid_from_diff(node_changelog: NodeChangelog) -> list[str] | None:
+        """Recover a node's HFID from the human-friendly-id attribute the diff carries for it."""
+        attribute = node_changelog.attributes.get("human_friendly_id")
+        if attribute is None:
+            return None
+        raw = attribute.value if attribute.value is not None else attribute.value_previous
+        if not isinstance(raw, str):
+            return None
+        try:
+            parsed = json.loads(raw)
+        except json.JSONDecodeError:
+            return None
+        return parsed if isinstance(parsed, list) else None
 
     def _process_node_attribute(
         self, node: NodeChangelog, attribute: EnrichedDiffAttribute, schema: MainSchemaTypes | None
@@ -156,6 +198,9 @@ class DiffChangelogCollector:
                                 node_kind=node.node_kind,
                                 relationship_name=relationship.name,
                             )
+                            # The peer's display label is already carried by the diff, no load needed.
+                            changelog_rel.peer_display_label = entry.peer_label
+                            changelog_rel.peer_hfid = self._peer_hfid(peer_id=rel_prop.new_value)
                         if rel_prop.previous_value:
                             changelog_rel.peer_id_previous = rel_prop.previous_value
                             changelog_rel.peer_kind_previous = self.get_peer_kind(
@@ -201,6 +246,8 @@ class DiffChangelogCollector:
                 peer_kind=self.get_peer_kind(
                     peer_id=peer.peer_id, node_kind=node.node_kind, relationship_name=relationship.name
                 ),
+                peer_display_label=peer.peer_label,
+                peer_hfid=self._peer_hfid(peer_id=peer.peer_id),
                 peer_status=peer.action,
             )
             for peer_prop in peer.properties:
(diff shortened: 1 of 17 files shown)


backend/infrahub/graphql/mutations/relationship.py
```diff
@@ -232,6 +249,14 @@ async def mutate(
                 user_id=graphql_context.assigned_user_id,
             )
 
+        if node_changelog.has_changes:
```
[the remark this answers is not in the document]
polmichel: Good catch — fixed in #10569: the source-node enrichment now runs only on the non-group, event-enabled path (`group_event_type == GroupUpdateType.NONE`), so group relationship mutations no longer incur the extra node/attribute read.

backend/infrahub/core/changelog/diff.py
```diff
@@ -228,14 +275,58 @@ def _process_node_cardinality_many_relationship(
...
+        return peer.peer_hfid is None and peer.peer_id not in self._node_hfids
+
+    async def _resolve_external_peer_hfids(self, changelogs: list[NodeChangelog]) -> None:
+        """Fill the HFID of relationship peers that are referenced but not themselves changed nodes.
+
+        A changed relationship is usually symmetric, so most peers are changed nodes already loaded;
+        this resolves the rest so a peer's HFID does not depend on whether the peer also changed.
+        """
+        peers = list(self._peer_entries(changelogs))
+        external_ids = [peer.peer_id for peer in peers if peer.peer_id and self._needs_external_hfid(peer)]
```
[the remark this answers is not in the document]
polmichel: Addressed in the stacked #10571, which extracts this into `ChangelogHfidResolver` and filters external peers by kind (`is_resolvable_kind`) before the batch load — a peer whose kind was dropped from the schema is left unresolved instead of poisoning the HFID load for the valid external peers. Left as-is here since that PR replaces this inline resolution.

polmichel (commented):
Stacked PRs are only about refactoring. The "core" feature from a correctness standpoint is in this PR.
- #10570 — Shares `HFID_ATTRIBUTE_NAME` / `DISPLAY_LABEL_ATTRIBUTE_NAME` constants instead of repeated
  string literals.
- #10571 — Extracts HFID resolution into a dedicated `ChangelogHfidResolver`.
- #10572 — Pages node-label loads with the shared `chunked()` helper.
- #10573 — Moves the relationship changelog mapper/getter out of `models.py` into their own modules.

That said, depending on the performance tests results, I might change the feature in another stacked PR.

backend/infrahub/core/changelog/enrichment.py
```diff
@@ -0,0 +1,126 @@
...
+    """Human-readable identifiers of a node, added to changelog payloads for external consumers."""
+
+    display_label: str
+    """The node's display label."""
+
+    hfid: list[str] | None
+    """The node's human-friendly ID, or None when the node has none."""
+
+
+PLACEHOLDER_LABELS = NodeLabels(display_label="n/a", hfid=None)
```
gmazoyer, line 35: It reminds me that mutation path stamps `"n/a"` into `peer_display_label` for an unresolvable peer , while the diff path leaves the same field `null`. We should probably rework that at some point to make this consistent.

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -113,8 +113,10 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
                 diff_branch_name=self.source_branch.name,
                 tracking_id=BranchTrackingId(name=self.source_branch.name),
             )
-            changelog_collector = DiffChangelogCollector(diff=branch_diff, branch=self.source_branch, db=self.db)
-            node_events = changelog_collector.collect_changelogs()
+            changelog_collector = build_diff_changelog_collector(
+                diff=branch_diff, db=self.db, branch=self.source_branch
+            )
+            node_events = await changelog_collector.collect_changelogs()
```
gmazoyer, line 119: Since `collect_changelogs()` now does graph reads, and at this point the merge lock and the write blocker are still held, could the collection run after `merge_write_blocker.delete()`, like the rebase path does?

backend/infrahub/core/changelog/models.py
```diff
@@ -633,39 +712,31 @@ def _process_added_peers(
...
+        name: str, primary_changelog: NodeChangelog, peer_status: DiffAction
+    ) -> RelationshipCardinalityOneChangelog:
+        """Build the one-cardinality reciprocal, placing the primary as the current or removed peer."""
+        if peer_status == DiffAction.REMOVED:
+            # The primary is the removed (previous) peer, so no current-peer label.
+            return RelationshipCardinalityOneChangelog(
+                name=name,
+                peer_id_previous=primary_changelog.node_id,
+                peer_kind_previous=primary_changelog.node_kind,
+            )
```
gmazoyer, line 735: A peer leaving a many-relationship keeps its label and HFID on the reciprocal, but a peer leaving a one-relationship gets neither. Is it deliberate?

backend/infrahub/core/changelog/enrichment.py
polmichel: True I'll create a tech debt item for it

backend/infrahub/core/merge/orchestrator.py
polmichel: Agreed, moved in c552b7861. The collection now runs after `merge_write_blocker.delete()`, outside the rollback path and under an exception guard: a failure drops the node events but cannot fail or roll back the merge. Two component tests pin the ordering and the failure case; both fail when the collection is moved back under protection.

backend/infrahub/core/changelog/models.py
polmichel: Deliberate: the one-cardinality model only carries labels for the current peer, no `_previous` counterpart, and the diff path does the same. Adding `peer_display_label_previous` / `peer_hfid_previous` changes the webhook payload, so I would do it in a follow-up PR. Happy to open it if you agree.

backend/infrahub/core/changelog/enrichment.py
polmichel: Fixed in cd4a678d2. An unresolved peer now keeps `peer_display_label` / `peer_hfid` unset on the mutation path, matching the diff path. `"n/a"` only remains for the required `display_label` of a secondary changelog whose node is gone.

backend/infrahub/core/changelog/models.py
polmichel: Follow-up tracked in IFC-3197.

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -178,13 +193,21 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
...
-            merge_diff_cache_key=merge_diff_cache_key,
-        )
+        # Reads the source branch: after the write protection is lifted, before the follow-ups that
+        # may schedule that branch's deletion. A failure fails the task; the merge stays committed.
+        try:
+            changelog_collector = self.changelog_collector_factory(
+                diff=branch_diff, db=self.db, branch=self.source_branch
+            )
+            node_events = await changelog_collector.collect_changelogs()
+        finally:
```
ogenstad: If an error gets raised here we'd run `self.post_merge_dispatcher.run_follow_ups()` but not `self.post_merge_dispatcher.dispatch_events()`. I think that the comment above kind of alludes to this. I'm wondering if we should explain why the merge event and other things that gets triggered on line 212 shouldn't happen?

polmichel: You're right. I've decided to remove completely the try/catch block since there is no particular reason to do it for this code block and not the others.

backend/infrahub/core/merge/orchestrator.py
```diff
@@ -178,6 +193,11 @@ async def merge(self, *, context: InfrahubContext, proposed_change_id: str | Non
         # Persisted only past the point of no return, so a rolled-back merge leaves no entry behind.
         merge_diff_cache_key = await self._cache_diff_summary(branch_diff=branch_diff)
 
+        # Collect changelogs only after write protection is lifted, but before follow-ups, which may
+        # schedule deletion of the source branch the collector reads.
+        changelog_collector = self.changelog_collector_factory(diff=branch_diff, db=self.db, branch=self.source_branch)
+        node_events = await changelog_collector.collect_changelogs()
```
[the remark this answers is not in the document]
polmichel: Why would this part of the code especially raise ?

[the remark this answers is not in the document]
polmichel: Other surfaces in the code, even those who read external component as Prefect, or the database do not all catch Exceptions right?

backend/infrahub/core/changelog/enrichment.py
```diff
@@ -0,0 +1,130 @@
...
+        return {
+            node_id: NodeLabels(
+                display_label=await node.get_display_label(db=self._db),
+                hfid=await node.get_hfid(db=self._db),
+            )
+            for node_id, node in nodes.items()
+        }
+
+    async def load_hfids(self, node_ids: list[str]) -> dict[str, list[str] | None]:
+        nodes = await self._load_nodes(node_ids)
```
ogenstad, line 89: I'm wondering what happens here, it looks like we'd end up using a raw NodeManager.get_many(). A caveat there could be the "fields" parameter and what we query by default. Could this lead to us querying for attributes and or relationships that we don't need?

ogenstad (approved):

backend/infrahub/core/changelog/enrichment.py
polmichel: Exactly, this is fixed in PR later in the stack https://github.com/opsmill/infrahub/pull/10638/changes

PR #10576, opened 2026-09-09
chore(labels): stop the triage workflows reporting no-op runs as comments

saltas888:
## Why

Standing down is the triage labeler's most common outcome: a human labelled the issue within the 10 minute head start, the issue is a feature request whose component is already set, or the report is too thin to classify. Every one of those was posting a comment on an `[aw] No-Op Runs` issue, describing a *different* issue than the one it was filed against, which reads as noise rather than signal.

The tracker also gets recreated whenever the previous one is closed, so #9201, #10430 and #10575 were all open at once.

Nothing is lost by turning it off. Each run's summary still records the decision and its reasoning, and genuine failures are unaffected — `report-failure-as-issue` was already false, and this only changes the no-op path.

## Note

The two pre-existing trackers (#9201, #10430) collect no-ops from the four `bug-agent-*` workflows too, which this PR does not touch. They will keep accumulating unless the same flag is set there.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


polmichel (approved):
Thanks!

PR #10580, opened 2026-09-09
feat(ipam): choose the target type when allocating from a pool

pa-lem:
Closes [IFC-2764](https://opsmill.atlassian.net/browse/IFC-2764). Sibling of IFC-1869 (custom prefix mask) under epic IFC-2763.

**Draft** — the availability harness and the e2e/documentation helpers are still outstanding (see *Not done yet*). Everything below is implemented, tested and walked by hand.

## What

When a relationship points at the generic `BuiltinIPAddress` / `BuiltinIPPrefix`, several concrete types can satisfy it. You could already override a pool's default *prefix length* (IFC-1869); now you can override its default *type* too.

Along the way this turned into a form change, for two reasons found while building it.

**From-pool allocation was unreachable in the UI for exactly these relationships.** A generic peer routes to `GenericRelationshipField`, which dropped the `pool` prop entirely, and the pool list was filtered to `default_address_type = <the generic>`, which no pool matches. So the feature's own target had no pool affordance at all.

**And the form contradicted itself.** It made you choose a "Kind" before revealing a pool selector whose list is *deliberately* unfiltered by that Kind — filtering by it would make a sibling-type override impossible, which is the whole point. So you picked a Kind, then picked an unrelated pool, then saw a separate type override.

Every field that can be satisfied from a pool now presents the two ways of filling it in as tabs:

image

## Behaviour changes a reviewer should scrutinise

- **`prefix_type` gains validation it never had.** It was already accepted on `IPPrefixPoolInput` with no checking, so a client sending a type unrelated to the relationship's peer now gets an error instead of an unexpectedly-typed node. That path had zero test coverage.
- **The standalone pool mutations are now validated too**, against the builtin IP generic.
- **`<relationship>_from_resource_pool` now carries only the pool id.** Its peer is the pool kind, so GraphQL selects plain `RelatedNodeInput`, which accepts neither `address_type` nor `prefixlen` — sending either failed the mutation outright. This also fixes a **pre-existing** failure whenever a prefix length was typed on a pool-backed IP relationship of an object template. Consequence: object templates still allocate the pool's default type, tracked in [IFC-3135](https://opsmill.atlassian.net/browse/IFC-3135).
- **Pool availability widens in one case.** Converging the two pool channels made the gate a union (`a from-pool relationship exists` **or** `matching pools were prefetched`), so a template schema with zero matching pools now shows the tab with an empty list. Gating on the relationship alone would have removed the pool from every plain node form, since those relationships exist only on object-template schemas. IP fields already behaved this way.

## Two pre-existing bugs fixed on the way

- **A disabled field kept a live pool button.** `PoolSelect` accepted no `disabled` prop at all; the controls now take a real one.
- **Switching tabs could destroy data.** The switch blanked the field and marked it dirty, so on an edit form merely *looking* at the other tab would clear the allocation on save. A switch now restores the value the form opened with, which lands on the existing deep-equal guard in `getUpdateMutationFromFormData` — so looking around submits nothing. Verified end-to-end: open → From pool → back → save leaves the object byte-identical.

Also worth flagging: both nested override fields needed `shouldUnregister={false}`. React-hook-form's `unset` walks `<name>.value.from_pool` on unmount, which crashes once the value is nulled — reachable today on `develop` by selecting a pool and then ticking "Set empty".

## Design decisions

- **A form opens on the value tab whatever the provenance**, with a badge naming the pool. The pool tab *stages* an allocation rather than displaying one: for a resolved allocation it has no controls, so opening there showed an empty picker and hid the allocated address. The value tab shows the object, its type, and where it came from.
- **Only secondary controls carry labels.** Each tab's primary control is named by the tab itself, so the pool has no redundant "Pool" label — but it does have a placeholder, since an empty combobox with no text reads as broken rather than unfilled.
- **The type override's placeholder is the pool's default type**, not "Select a type" — the default is the one thing that makes the override comprehensible.
- **Only the caller's explicit type is validated, never the pool's own default.** Re-validating the default would newly reject pools whose `default_address_type` sits outside the peer's `used_by` and break working setups.

## Trying it locally

`models/base/ipam.yml` defines exactly one concrete type per IPAM generic — the original reason this could not be exercised. The example schema adds a second:

```
infrahubctl schema load models/base models/examples/ipam_kind_override.yml
infrahubctl run models/examples/ipam_kind_override_data.py
```

Both demo pools default to the **`Ipam`** types on purpose, so an `Infra` result can only have come from an override. `dev/specs/ifc-2764-pool-field-tabs/quickstart.md` walks 12 cases in plain language.

## Not done yet

- **Availability harness** — the rules span three gating layers; four schema shapes are still missing, so four matrix rows cannot be reached by hand (single-implementation generic, an `address` attribute, a Number-pool attribute, an object template). Planned in `dev/specs/ifc-2764-pool-field-tabs/`.
- **The attribute half of "opens on the value tab"** is unverified. It is proven in a browser for relationship fields against a real object, but no demo node has a pool-backed *attribute*, and `getFieldDefaultValue` has fallback paths that could report a pool-allocated attribute value as user-provided. Gated on the harness.
- **`tests/e2e/helpers.py::select_pool`** still encodes the old single-slot layout (8 call sites, ~10 files), and the tutorial/guide tests regenerate documentation imagery that still shows a pool button.

## Testing

- Frontend: **194 files / 1379 tests**. `biome ci` clean over 1556 files at the workspace root, `knip` 0, `betterer ci` unchanged at 186.
- Backend: `invoke format lint` clean (ruff, `ty`, mypy over 1677 files); **161** component tests across the resource-manager, relationship, GraphQL resource-manager and number-pool suites.
- By hand against a local stack: override wins for both pool types, default applies without one, invalid types rejected with the allowed set listed, a reservation's type cannot be changed, and the edit round-trip leaves objects untouched.

Rebased onto `develop` with no conflicts; no develop commit since the merge base touched a file this branch rewrote.

[IFC-2764]: https://opsmill.atlassian.net/browse/IFC-2764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-3135]: https://opsmill.atlassian.net/browse/IFC-3135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/relationship/model.py
```diff
@@ -630,6 +630,13 @@ async def resolve(
                 if hfid_str:
                     data_from_pool["identifier"] = f"hfid={hfid_str} rel={self.name}"
 
+            # Every from_pool field is forwarded generically as a kwarg to whichever pool the id
+            # resolves to, so `peer_kind` may only be added for the pool kinds whose get_resource
+            # accepts it — CoreNumberPool.get_resource would raise TypeError otherwise.
```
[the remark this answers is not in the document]
saltas888: @pa-lem I think thats valid

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
```diff
@@ -36,12 +37,33 @@ async def get_resource(
...
+        # reservation lookup so an existing reservation can be checked against it too.
+        requested_prefix_type = prefix_type or data.get("prefix_type", None)
+
+        # Deliberately outside the pool lock: this check needs nothing but the pool's name and
+        # the requested kind, so a request that can never succeed must not queue behind the
+        # allocations of every other caller of this pool.
+        validate_allocated_kind(
+            db=db,
+            branch=branch,
+            pool_kind="IPPrefixPool",
```
saltas888, line 61: I think this should be typed and not a string

frontend/app/src/shared/components/form/utils/mutations/getCreateMutationFromFormData.ts
```diff
@@ -28,11 +28,22 @@ export const getCreateMutationFromFormData = (
...
-        const fromPool = buildFromPoolPayload(fieldData.value.from_pool, fieldData.source.kind);
         if (fromPoolField) {
-          return { ...acc, [fromPoolField]: fromPool };
+          // `<rel>_from_resource_pool` peers at the pool kind itself, so GraphQL types it
+          // as a plain RelatedNodeInput — which has no `prefixlen` and no `address_type`,
+          // and rejects the whole query if either is sent. The backend could not honour
+          // them there anyway: the relationship stores only a pointer to the pool, and
+          // `create.py` allocates with `pool.get_resource(...)`, passing no prefix length
+          // and no kind. Overrides only reach the API through the direct
+          // `{ [field.name]: { from_pool } }` payload below.
```
saltas888, line 38: pleaase do a passthrough on the code comments, they are very verbose. I'd suggest checking our guidelines and see why they dont work

frontend/app/src/shared/components/form/fields/input.field.tsx
```diff
@@ -41,60 +63,96 @@ const InputField = ({
...
+          return (
+            <div className="space-y-2">
+              {fieldLabel}
+              {valuePanel}
+            </div>
+          );
+        }
+
+        return (
+          <FieldTabs
```
saltas888, line 120: this is going out of hand. A component called shared -> form -> input.field, rendering all this. It was already too much and polluted with the <PoolSelect, but now we are doing even more. I think its time we tackled it well

frontend/app/src/shared/components/form/fields/relationships/relationship-hierarchical.field.tsx
```diff
@@ -61,61 +103,101 @@ export default function RelationshipHierarchicalField({
...
+          return (
+            <div className="flex flex-col gap-2">
+              {fieldLabel}
+              {objectPanel}
+            </div>
+          );
+        }
+
+        return (
+          <FieldTabs
```
saltas888, line 165: same here. Note: this  seems repeated code

backend/infrahub/core/node/resource_manager/reservation.py
```diff
@@ -39,3 +39,35 @@ def validate_reserved_prefix_length(
                 f"/{existing_prefixlen} can be used."
             )
         )
+
+
+def validate_reserved_kind(
+    *,
+    pool_kind: str,
```
saltas888, line 46: same here, can this be typed?

backend/tests/component/core/test_relationship.py
saltas888, line 1: Some of this is duplicated across layers, and these are schema-registering component tests **so they aren't cheap.**

Three of the four new relationship tests are subsumed by GraphQL ones, since `TestGenericAddressOwnerCreate` resolves `from_pool` through this same `Relationship.resolve` path:

- `..._overrides_prefix_type` -> `test_create_from_pool_with_prefix_type_override`
- `..._overrides_address_type` -> `test_create_from_pool_with_address_type_override`
- `..._rejects_kind_outside_peer_used_by` -> `test_create_from_pool_with_invalid_prefix_type`

The fourth has no GraphQL twin, but its rule is already in `test_get_resource_prefix_type_rejected_for_concrete_peer`, and the line it uniquely covers doesn't branch on the kind.

`kind_override_pools` is identical here and in the GraphQL suite bar one `update_schema_hash()`, plus two near-copies as `kind_override_pool` in the core suites. Four schema-registering fixtures, could be one in conftest.

Also in `test_ipaddress_pool.py`, the tails of `test_get_resource_conflicting_address_type_raises` and `test_get_resource_reserved_kind_validated_against_peer` are the same call and assertion. And the prefix suite has no counterpart to the former, so `validate_reserved_kind` never raises from its prefix call site.

saltas888 (commented):
cubic comments are also to the point

backend/infrahub/core/node/resource_manager/ip_address_pool.py
```diff
@@ -54,16 +76,34 @@ async def get_resource(
...
+                            reserved_value=node.get_attribute("address").value,
+                            reserved_kind=node.get_kind(),
+                            requested_kind=requested_address_type,
+                        )
+                        # The reservation may have been created under a peer more permissive than
+                        # this caller's — the standalone mutation validates against the broad
+                        # BuiltinIPAddress generic, a relationship against its own narrower peer.
+                        # Relationship.set_peer performs no kind check, so without this the
+                        # reserved node would be attached to a peer that cannot hold it.
+                        validate_allocated_kind(
```
saltas888, line 96: I think this one should come out. The plain version of why:

The rule you added is "if the caller names a type, check the slot can hold it", and you deliberately never check the pool's own default. Fine on its own. But a reservation means asking twice with the same identifier gives you the same object back, and that's where the two collide.

First ask, no type named, so nothing gets checked. The pool falls back to its default, makes an object the peer can't hold, and it gets attached. Second ask, identical, and this check looks at what's stored and errors.

So it fires on the way back in, never on the way that created the problem. The comment above it says it stops a reserved node being attached to a peer that cannot hold it, but that already happened one call earlier.

What it costs is that a request which succeeded now fails when you repeat it, which is the opposite of what reservations are for. And it does police the pool default after all, just one call late, so the breakage you were avoiding turns up anyway in a more confusing form.

pa-lem: Agreed, removed. I verified the reasoning first: that call can only fire when the caller named no kind, because a named kind is already validated earlier and validate_reserved_kind guarantees the stored node matches. So its whole live domain was the repeated reservation — turning a request that succeeded into one that fails. Removed from both ip_address_pool.py and ip_prefix_pool.py, with the test that only covered it.

One case it did cover: identifier is caller-supplied, so a reservation made through the standalone mutation can later be attached via a narrower peer, and set_peer does no kind check. That gap was introduced by this branch and is now removed, so it is not a regression against develop. The proper fix is validating the resolved kind (pool default included) on the creation path, which catches both cases symmetrically — filing that as a follow-up rather than smuggling it in here, since it can reject deployments whose pool default sits outside the peer used_by.

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
pa-lem: Typed as PoolKind = Literal["IPAddressPool", "IPPrefixPool"], applied to validate_allocated_kind, validate_reserved_kind and validate_reserved_prefix_length.

Worth noting these are not schema kinds — they are the pool schemas name field, used only as the error-message prefix. InfrahubKind.IPADDRESSPOOL is "CoreIPAddressPool", so swapping that in would change user-facing error text.

The kind parameters next to them (requested_kind, peer_kind, address_type, prefix_type) stay str: they are user-defined schema kinds arriving as GraphQL String and resolved at runtime, so nothing can enumerate them.

backend/infrahub/core/node/resource_manager/reservation.py
pa-lem: Same change — pool_kind is now PoolKind, a Literal over the two pool schema names, so typos are caught at all call sites with no runtime or message change.

frontend/app/src/shared/components/form/fields/relationships/relationship-hierarchical.field.tsx
pa-lem: You were right that it is repeated — more than it looked. The tab scaffold was byte-identical across all five pool-backed fields, ~100 lines, and the surrounding state and label handling brought the total to ~314 duplicated lines.

Extracted form/pool-backed-field.tsx, which owns the tab ids and state, the switch-restores-default rule, the label, the untabbed early return and the allocation panel. This file goes 205 to 131 lines; the five together lose 225. The switch rule fixes a data-loss bug, so having it in one place instead of five is the real win.

All five field test suites pass untouched, which is what makes the refactor readable as behaviour-neutral.

backend/tests/component/core/test_relationship.py
pa-lem: Checked each mapping and you are right on three — deleted ..._overrides_prefix_type, ..._overrides_address_type and ..._rejects_kind_outside_peer_used_by. Each has a GraphQL twin raising from the same validate_allocated_kind call, and the GraphQL versions are strict supersets since they also exercise the new graphene inputs.

Kept the fourth, renamed test_relationship_from_pool_forwards_peer_kind_to_pool: it is the only thing pinning schema.peer to peer_kind if the GraphQL suite ever moves.

Two corrections. The "same tails" pair only shares its final line — those reservations hold different kinds, so they assert different things. And while checking I found the prefix-side validate_reserved_kind raise had no test at all, so that is added (net +1).

On fixtures: schema registration was already a single conftest fixture; the duplicates were the pool-creating ones, now moved to conftest with a combining fixture. Not merged into one both-pools fixture, which would build a second pool for the ~13 single-pool tests.

frontend/app/src/shared/components/form/utils/mutations/getCreateMutationFromFormData.ts
pa-lem: Done — added comments cut from 704 lines to 268 (62%). I re-read .agents/rules/code-doc-style.md as you suggested, and the rules being broken were specific: narrating what the code plainly does, work-item IDs in source, history narration ("used to", "before the tabs"), bare GIVEN/WHEN/THEN markers, and comments naming other code.

Your "see why they dont work" landed — field-tabs.tsx was factually wrong, claiming the wrappers pin a segmented variant when the code pins variant="field". Exactly the rot that makes verbose comments a liability.

Eight survived, shortened: the ones recording a constraint that gets silently undone otherwise — the react-hook-form unset crash, the RelatedNodeInput rejections, the 1px border that stops the tab strip shifting height, and similar.

frontend/app/src/shared/components/form/fields/input.field.tsx
pa-lem: Partly addressed, and I would rather leave this open than claim otherwise.

What landed: the tab scaffold, tab state and allocation panel moved into form/pool-backed-field.tsx, so this file is 162 to 88 lines and is back to rendering an Input.

What has not: the broader "tackle it well" composition pass over form/fields. Two concrete pieces I found and deliberately did not do here, because neither is about pools and this PR is already large — generic-relationship.field.tsx is still ~387 lines of kind-derivation state machine with three set-state-during-render calls, and regular-/generic-relationship share ~40 more duplicated lines of parent picker. Happy to raise both as tickets, or to widen this PR if you would rather it all went together.

frontend/app/src/shared/components/form/utils/mutations/getUpdateMutationFromFormData.test.ts
```diff
@@ -298,7 +298,194 @@ describe("getUpdateMutationFromFormData - test", () => {
...
+
+    // WHEN
+    const mutationData = getUpdateMutationFromFormData({ fields, formData });
+
+    // THEN
+    expect(mutationData).to.deep.equal({});
+  });
+
+  it("sends the update when the same pool is reselected with a different allocated kind", () => {
+    // GIVEN an object already allocated from this pool as IpamIPAddress, with the user now
```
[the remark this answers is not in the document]
pa-lem: Fixed in the comment pass — the "silent no-op" narration and the create.py cross-reference are gone, leaving one durable sentence.

frontend/app/src/shared/components/form/utils/mutations/getUpdateMutationFromFormData.ts
```diff
@@ -24,12 +20,23 @@ export const getUpdateMutationFromFormData = ({
...
-      return acc;
+      // Re-selecting the field's original pool is normally a no-op: allocation is
+      // idempotent on the reservation identifier, so the mask cannot be changed that way.
+      // A different allocated kind is the exception — no reservation identifier is sent,
+      // so it is a real request and must not be dropped from the mutation.
+      const requestedKind =
+        fieldData.value && typeof fieldData.value === "object" && "from_pool" in fieldData.value
+          ? fieldData.value.from_pool.allocatedKind
+          : undefined;
+      if (!requestedKind || requestedKind === getAllocatedKind(defaultValue.value)) {
```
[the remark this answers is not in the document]
pa-lem: Fixed in 3dbb59f292. The override controls are now withheld whenever the field allocates through a companion `<name>_from_resource_pool` relationship (`canOverrideAllocation = !fromPoolRelationshipName`), so no kind can be chosen on that path and none can be dropped. The underlying limitation is tracked separately for the object-template case.

backend/tests/component/conftest.py
```diff
@@ -2620,6 +2620,160 @@ async def register_ipam_extended_schema(default_branch: Branch, register_ipam_sc
...
+            },
+        ],
+    }
+
+    schema_branch = registry.schema.register_schema(schema=SchemaRoot(**SCHEMA), branch=default_branch.name)
+    default_branch.update_schema_hash()
+    return schema_branch
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
pa-lem: Not correct — `kind_override_pools` is in use. It is requested by 7 tests in backend/tests/component/graphql/resource_manager/test_resource_manager.py (lines 1525, 1548, 1568, 1590, 1610, 1638). It was moved into conftest precisely so those suites and the core ones could share one definition instead of four near-copies.

frontend/app/src/shared/components/form/pool-backed-field.tsx
```diff
@@ -0,0 +1,134 @@
...
+  allocatableKinds?: Array<PoolKindOption>;
+  valueTabLabel: "Value" | "Object";
+  disabled?: boolean;
+  untabbedClassName?: string;
+  onPoolChange: (value: PoolValue | null) => void;
+  onTabSwitch?: () => void;
+  children: ReactNode;
+}
+
+/**
```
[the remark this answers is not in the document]
pa-lem: Rewritten as a contract: what the field is, what it renders without a pool, and that switching tabs discards what the other staged. No call site named.

backend/tests/component/core/test_relationship.py
```diff
@@ -505,6 +505,25 @@ async def test_relationship_assign_from_pool(
     assert await obj.prefix.get_peer(db=db)
 
 
+async def test_relationship_from_pool_forwards_peer_kind_to_pool(
```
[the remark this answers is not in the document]
pa-lem: Renamed to test_relationship_from_pool_rejects_sibling_kind_for_concrete_peer, and the docstring now states only the behaviour asserted.

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
```diff
@@ -36,12 +37,29 @@ async def get_resource(
...
         # TODO(IFC-2945): drop this alias once the public `size` input field is renamed to `prefixlen`.
         if prefixlen is None:
             prefixlen = size
+        data = data or {}
+        pool_name = str(self.get_attribute("name").value)
+
+        # Only an explicitly requested kind is validated; the pool's own default is left alone.
+        requested_prefix_type = prefix_type or data.get("prefix_type", None)
+
+        # Outside the pool lock: a request that can never succeed must not queue behind other allocations.
```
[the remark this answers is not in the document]
pa-lem: Lock-ordering rationale removed here and in the ip_address_pool twin, which was not flagged but carried the same line.

frontend/app/src/shared/components/form/utils/mutations/getUpdateMutationFromFormData.test.ts
```diff
@@ -298,9 +298,169 @@ describe("getUpdateMutationFromFormData - test", () => {
     expect(mutationData).to.deep.equal({});
```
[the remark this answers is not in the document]
pa-lem: Removed all 64 markers in this file rather than restoring 18, and did the same in getCreateMutationFromFormData.test.ts (96) and generic-relationship.field.test.tsx (12), which had the same split. Fewer comments is the direction the repo owner wants, so consistency went that way.

models/examples/ipam_kind_override_data.py
```diff
@@ -0,0 +1,83 @@
...
+    supernet_prefix = await client.create(
+        branch=branch, kind="IpamIPPrefix", prefix=str(DEMO_SUPERNET), member_type="prefix"
+    )
+    # Using upsert so the script can be re-run on the same branch during development.
+    await supernet_prefix.save(allow_upsert=True)
+
+    log.info("Creating Kind override demo prefix pool")
+    prefix_pool = await client.create(
+        kind=CoreIPPrefixPool,
+        name="Kind override demo prefix pool",
```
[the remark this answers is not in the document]
pa-lem: quickstart.md now uses the current pool names. The demo object was also still called ifc-2764-demo, which the rename had missed; it is now kind-override-demo.

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
```diff
@@ -36,12 +37,29 @@ async def get_resource(
...
         at: Timestamp | None = None,
         user_id: str = SYSTEM_USER_ID,
     ) -> Node:
         # TODO(IFC-2945): drop this alias once the public `size` input field is renamed to `prefixlen`.
         if prefixlen is None:
             prefixlen = size
+        data = data or {}
+        pool_name = str(self.get_attribute("name").value)
+
+        # Only an explicitly requested kind is validated; the pool's own default is left alone.
```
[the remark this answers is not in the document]
pa-lem: Leaving this open rather than acting on it, because it would undo a human reviewer's decision.

This check existed and was removed two pushes ago at their request. Their argument: it can only fire when the caller named no kind, since a named kind is validated earlier and validate_reserved_kind guarantees the stored node matches. So it never fires on the call that created the mismatch, only on an identical repeat — turning a request that succeeded into one that fails, which defeats the point of a reservation.

The underlying asymmetry is real and deliberate: the pool's own default kind is never validated against the peer, so a default outside the peer's used_by can still be attached. Fixing that properly means validating the resolved kind at creation, for defaults too, which can reject deployments that work today. That is a separate change, not this one.

Happy to be overruled if @saltas888 wants it back.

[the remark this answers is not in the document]
saltas888: @pa-lem I actually think you should decide this here. Im not exactly sure what is this edge case

frontend/app/src/shared/components/form/utils/getFormFieldFromAttribute.ts
```diff
@@ -169,25 +169,30 @@ export const getFormFieldFromAttribute = ({
...
-        : undefined,
+      pool:
+        hasFromPoolRelationship || numberPools.length
+          ? {
+              kind: NUMBER_POOL_KIND,
+              defaultAllocatedObjectKind: schema.kind!,
+              fromPoolRelationshipName: hasFromPoolRelationship ? fromPoolName : undefined,
+              // Pre-fetched: narrowing CoreNumberPool to this kind and attribute is the caller's job.
+              // Left undefined when nothing was pre-fetched, so the pool select queries instead.
+              options: numberPools.length ? numberPools : undefined,
```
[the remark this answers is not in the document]
pa-lem: Correct, and thank you — my previous change made this worse, not better. Applied your suggestion.

`numberPools` is already filtered by attribute, so a supplied list that matches nothing means no pool applies; turning that into `undefined` sent the select down the query path and offered every CoreNumberPool, including pools scoped to other attributes and kinds.

Reviewing the fix surfaced a second half to it: `getFormFieldsFromSchema` defaulted the parameter to `[]`, so `pools === undefined` was unreachable and the query fallback was dead for the seven call sites that do not prefetch (convert-form, core-file-form, repository-form, the two pool forms, the two trigger match forms). Dropped that default so absence propagates. It also means an errored pool query no longer reads as "no pools apply".

Tests now cover all three states — supplied and matching, supplied and matching nothing, absent — including one at the `getFormFieldsFromSchema` level that fails against the old default.

frontend/app/src/shared/components/form/fields/relationships/regular-relationship.field.test.tsx
```diff
@@ -0,0 +1,328 @@
...
+
+import { getRelationships } from "@/entities/nodes/relationships/domain/use-cases/get-relationships";
+import { nodeSchemasAtom } from "@/entities/schema/stores/schema.atom";
+
+import { TestForm } from "../../../../../../tests/components/form.story";
+import { render } from "../../../../../../tests/components/render";
+import {
+  generateNodeSchema,
+  generateRelationshipSchema,
+} from "../../../../../../tests/fake/schema";
```
saltas888, line 15: those imports are a bit disturbing, do we have aliases?

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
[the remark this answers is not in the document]
saltas888: @pa-lem whats the decision here?

saltas888 (approved):

backend/infrahub/core/node/resource_manager/ip_prefix_pool.py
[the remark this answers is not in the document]
pa-lem: The gap is real, but this is not the place to close it. The check only fires when the caller named no type, and by then the node has already been created and attached by the first call — so it never catches the mistake, it just makes an identical repeat fail, which defeats the point of a reservation. It was removed during review for that reason.

The fix is to validate the *resolved* kind, pool default included, on the creation path. That rejects allocations which succeed today, so it needs its own decision and a changelog entry.

Tracked in [IFC-3218](https://opsmill.atlassian.net/browse/IFC-3218).

pa-lem:
QA task available https://opsmill.atlassian.net/browse/IFC-3219

PR #10587, opened 2026-09-10
fix: heartbeat thread and linear conflict merge for large branch diffs

fatih-acar:
## Summary

A rebase of a branch with about 38k changed nodes failed with `LockNotOwnedError: Cannot release a lock that's no longer owned` after 27 minutes, once all of its diff work was done. The worker was alive, but its event loop was blocked for 93 s by a quadratic loop, its 15-second liveness heartbeat expired, and the `clean-up-deadlocks` flow running on another worker deleted the two diff-update locks the rebase still held. This PR fixes both halves: a worker's heartbeat no longer depends on its event loop being idle, and the loop that stalled it is now linear.

Timeline from the production log:

| Time | Event |
|---|---|
| 07:25:19 | both diff-update locks acquired |
| 07:50:29 | last heartbeat refresh (key expires at 07:50:44) |
| 07:50:32 | diff saved, conflict sync starts: no `await` for 93 s |
| 07:51:00 | `clean-up-deadlocks` on another worker: holder not active, locks 25 min old, both deleted |
| 07:52:05 | loop resumes (Prefect worker channel also dropped with a keepalive ping timeout) |
| 07:52:06 | lock release fails, rebase flow fails |

## Key Changes

- **Worker liveness heartbeat runs on its own thread** (`WorkerHeartbeat`, `backend/infrahub/services/heartbeat.py`), with its own event loop and its own cache connection, started and stopped by `InfrahubScheduler`. A CPU-bound flow step no longer makes a live worker look dead to the deadlock cleanup, the merge watcher, or the stale-lock cleaner. The refresh retries connecting and refreshing on every beat, so a cache outage delays the heartbeat instead of ending it. The asyncio heartbeat schedule and `tasks/keepalive.py` are removed; the key-writing logic is shared with the startup one-off refresh.
- **Data-check conflict merge is linear.** After a diff save, `DiffDataCheckSynchronizer._update_diff_conflicts` looked each updated node up with a scan of the retrieved diff's node set and grew that set on every miss (O(n²): 93 s at 38k nodes). It now uses the identifier-keyed node map; 20k nodes merge in 0.05 s in the new test.
- **Docs:** a "worker liveness heartbeat" section in `dev/knowledge/backend/async-tasks.md`; the two `dev/specs/ifc-2437-merge-failure-recovery` files no longer claim that sharing one event loop cannot starve the heartbeat (that claim was verified only against awaited Neo4j calls, not CPU-bound Python).

## Behaviour notes

- The heartbeat key now means "this process is alive", not "this process's event loop is idle". A CPU-stalled worker stays active, so `wait_for_schema_to_converge` waits up to its 30-second ceiling for it instead of dropping it; the status query and telemetry worker counts shift the same way.
- The primary API-server election runs on the heartbeat thread together with the rest of the refresh.
- Not in this PR: `clean_up_deadlocks` still deletes on a single missed 15-second window. A worker-to-cache blip while the cleanup's worker can still reach the cache remains the one way a live holder can lose a lock; the merge watcher absorbs that with a 180-second grace period, the cleanup has no equivalent.

## Risk

Touches process lifecycle (a daemon thread with its own asyncio loop and one extra Redis or NATS connection per worker process) and the liveness signal consumed by lock cleanup and merge-failure detection. No database schema, GraphQL schema, dependency, CI, or auth changes.

## Test Plan

- Unit: `uv run pytest backend/tests/unit/services backend/tests/unit/core/diff/test_data_check_synchronizer.py backend/tests/unit/core/merge/test_failure_identifier.py` (heartbeat keeps beating while the caller's loop is blocked; retries a failing cache and a failing connection; scheduler start/stop; conflict merge semantics and a 20k-node scale guard).
- Component (Neo4j via testcontainers): `uv run pytest backend/tests/component/locks/test_clean_up_deadlocks.py backend/tests/component/graphql/queries/test_status.py`.
- Local pre-CI: ruff, `ty check .`, `backend.lint` (mypy), `main.lint`, `uv lock --check`, `backend.validate-generated`, `docs.lint`, `docs.validate`, `schema.validate-graphqlschema`, `schema.validate-jsonschema`: all green. Full `backend.test-unit` run started locally; CI covers it too.
- Manual: rebase a large branch that has an open proposed change and watch the `workers:active:git_agent:worker:<id>` TTL keep refreshing through the save; the `Clean up deadlocks` flow must log no `Deleted deadlock` for the diff-update locks.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,125 @@
...
+            loop.run_until_complete(self._beat_until_stopped())
+        finally:
+            loop.close()
+
+    async def _beat_until_stopped(self) -> None:
+        cache: InfrahubCache | None = None
+        try:
+            while not self._stop_requested.is_set():
+                try:
+                    if cache is None:
```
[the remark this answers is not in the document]
fatih-acar: Valid point — after the first successful connect a broken client was retried forever, and only the never-connected case went back through the factory, which the docstring contradicted. A failed beat now closes the connection (guarding the close itself) and the next beat opens a fresh one through the factory; the docstring says so, and `test_heartbeat_reconnects_after_a_failed_refresh` covers a connection that breaks after it worked. In 12a77f86b.

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,125 @@
...
+        calling thread; from a coroutine, run it with ``asyncio.to_thread``.
+
+        Args:
+            timeout_seconds: Longest to wait for the thread to exit.
+
+        """
+        self._stop_requested.set()
+        if self._thread is not None:
+            self._thread.join(timeout=timeout_seconds)
+            self._thread = None
```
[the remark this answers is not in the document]
fatih-acar: Valid point — a stop whose join timed out dropped the reference while the thread was still closing its connection, and a later start would clear the stop flag and spawn a second thread beside it. The reference is now kept while the thread is alive (with a warning), so `running` stays truthful and `start` is a no-op until the old thread has exited; `test_stop_keeps_the_thread_while_it_is_still_finishing` covers it with a slow-closing cache. In c78a412cb.

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,125 @@
...
+                        cache = await self.cache_factory()
+                    await refresh_worker_heartbeat(cache=cache, component_type=self.component_type)
+                # Top-level boundary of the thread: a refresh that fails (cache unreachable, connection
+                # dropped) must not end the heartbeat, the next beat retries it.
+                except Exception:
+                    self.log.exception("Worker heartbeat refresh failed")
+                await self._sleep_until_next_beat()
+        finally:
+            if cache is not None:
+                await cache.close_connection()
```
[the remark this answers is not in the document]
fatih-acar: Valid point — `NATSCache.close_connection` was a bare `...`, so the heartbeat's dedicated connection, and the service's own at shutdown, were never closed. It now calls the client's `close()`; `close` rather than `drain` because the cache holds no subscriptions to flush, and closing an already closed client is a no-op. The NATS cache component tests (`tests/component/services/adapters/nats/test_nats.py`, five tests that each end with `close_connection`) pass against a NATS container. In 0284347dc.

backend/tests/unit/core/diff/test_data_check_synchronizer.py
```diff
@@ -0,0 +1,120 @@
...
+        identifier=identifier, attributes=set(), relationships=set(), conflict=EnrichedConflictFactory.build()
+    )
+    updated_diff = EnrichedRootFactory.build(nodes={first, second})
+    retrieved_diff = EnrichedRootFactory.build(nodes=set())
+
+    DiffDataCheckSynchronizer._update_diff_conflicts(updated_diff=updated_diff, retrieved_diff=retrieved_diff)
+
+    assert len(updated_diff.nodes) == 2
+    assert len(retrieved_diff.nodes) == 1
+    assert next(iter(retrieved_diff.nodes)).identifier == identifier
```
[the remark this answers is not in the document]
fatih-acar: Valid point — the test asserted only counts and the identifier. Which duplicate gets folded into which depends on set iteration order, so it now asserts the facts that hold either way: exactly one node survives, it carries the attribute only the second node had (so the second was folded in rather than added or dropped), and its conflict is one of the two rather than lost. Renamed to match. In c1832713a.

backend/tests/unit/services/test_heartbeat.py
```diff
@@ -0,0 +1,153 @@
...
+        component_type=ComponentType.GIT_AGENT, cache_factory=cache_factory, interval_seconds=BEAT_INTERVAL, log=log
+    )
+
+
+async def test_heartbeat_marks_worker_active_and_closes_cache_on_stop() -> None:
+    cache = RecordingCache()
+    heartbeat = build_heartbeat(cache=cache)
+
+    heartbeat.start()
+    try:
```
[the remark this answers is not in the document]
fatih-acar: Done in df7073f63 — start and stop now go through one `running(heartbeat)` async context manager. A context manager rather than a fixture because each test builds its heartbeat with different caches, factories and loggers; the one test that exercises `stop` itself keeps its explicit calls.

backend/tests/unit/services/test_heartbeat.py
```diff
@@ -0,0 +1,153 @@
...
+        await wait_until(lambda: ACTIVE_KEY in cache.storage)
+        assert heartbeat.running
+        assert f"workers:worker:{WORKER_IDENTITY}" in cache.storage
+    finally:
```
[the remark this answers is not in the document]
fatih-acar: Done in df7073f63 — the test now waits for both the active key and the worker key before asserting.

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -141,9 +145,8 @@ def _update_diff_attr_conflicts(self, updated_node: EnrichedDiffNode, retrieved_
                     continue
                 retrieved_prop.conflict = updated_prop.conflict
 
-    def _update_diff_relationship_conflicts(
-        self, updated_node: EnrichedDiffNode, retrieved_node: EnrichedDiffNode
-    ) -> None:
+    @staticmethod
+    def _update_diff_relationship_conflicts(updated_node: EnrichedDiffNode, retrieved_node: EnrichedDiffNode) -> None:
```
saltas888, line 149: The linearisation only covers the node level, this loop still has the same growing-scan shape: `get_element()` on line 160 scans `retrieved_rel.relationships`, and line 164 adds into that same set on a miss. Since `retrieved_diff` is fetched `only_conflicted=True` it starts near-empty and gets filled by this loop, which is exactly what caused the stall at the node level.

Measured on this branch, one node with a single cardinality-many relationship:

| peers | time |
|---|---|
| 2,000 | 0.02s |
| 5,000 | 0.14s |
| 10,000 | 0.70s |

so roughly 11s of await-free CPU at 40k peers. `get_attribute`/`get_relationship`/`get_property` have the same shape at smaller n.

Worth noting `enricher/hierarchy.py:101` also still calls the linear `EnrichedDiffRoot.get_node` inside a loop while `parent_adder` grows that same set, earlier in the same request, so a stall of the same order is reachable before this function is hit at all. Given the identifier map has now been hand-rolled three times (`parent_node_adder`, `combiner`, and here), indexing on the model might be the better fix than patching each call site.

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,153 @@
...
+            loop.close()
+
+    async def _beat_until_stopped(self) -> None:
+        cache: InfrahubCache | None = None
+        try:
+            while not self._stop_requested.is_set():
+                try:
+                    if cache is None:
+                        cache = await self.cache_factory()
+                    await refresh_worker_heartbeat(cache=cache, component_type=self.component_type)
```
saltas888: This recovers from a cache that raises, but not from one that goes quiet. `RedisCache` is built with `socket_timeout=None` and `socket_connect_timeout=None`, so a black-holed connection (LB idle drop, failover without RST) blocks here indefinitely, the `except` below never fires, and the stop flag is only checked between beats.

I reproduced it with a cache whose `set` sleeps rather than raises: two beats, then permanent silence. `stop()` then times out and keeps `_thread`, so `running` stays truthy and a later `start()` is a no-op, which lands back on the original `LockNotOwnedError` with no way to recover.

`asyncio.wait_for(refresh_worker_heartbeat(...), timeout=self.interval_seconds)`, or socket timeouts on the client, would cover it.

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,153 @@
...
+    async def _close_quietly(self, cache: InfrahubCache) -> None:
+        try:
+            await cache.close_connection()
+        # A client that just failed a refresh may fail to close as well; the connection is being
+        # abandoned either way, so this is logged rather than allowed to end the thread.
+        except Exception:
+            self.log.exception("Worker heartbeat cache connection could not be closed")
+
+    async def _sleep_until_next_beat(self) -> None:
+        deadline = time.monotonic() + self.interval_seconds
```
saltas888: Claude says:

_Two things about the cadence, both eating into the margin.

The deadline is anchored *after* the beat has completed, so the effective period is beat duration + 10s against a 15s TTL, leaving 5s of slack. And on the failure path above, a failed beat closes the connection then falls through to this same full interval, with the next attempt still needing to reconnect before it can write: a success at t=0 and a failure at t=10 means no write until t=20, against expiry at t=15.

`clean_up_deadlocks` is on a 1-minute cron and only needs one sample in that hole. Anchoring the deadline before the beat, and using a short backoff after a failure rather than the steady-state interval, would close both._

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,153 @@
...
+        try:
+            while not self._stop_requested.is_set():
+                try:
+                    if cache is None:
+                        cache = await self.cache_factory()
+                    await refresh_worker_heartbeat(cache=cache, component_type=self.component_type)
+                # Top-level boundary of the thread: a refresh that fails (cache unreachable, connection
+                # dropped) must not end the heartbeat. The connection is dropped so the next beat opens
+                # a fresh one instead of retrying a possibly broken client forever.
+                except Exception:
```
saltas888, line 155: don't we have the type check in the CI? I thought we did... so a I'd expect to mute here the static type checker

fatih-acar: only in `develop` for the moment, but good catch I will make sure this merges fine to develop

backend/infrahub/core/diff/data_check_synchronizer.py
fatih-acar: Valid, and worse than the extrapolation: measured on a single cardinality-many relationship, 1.1s at 10k peers, 5.4s at 20k and 24.2s at 40k. `get_element()` now goes through a `peer_id` map built once per relationship, the same shape as the node loop above.

`enricher/hierarchy.py:101` was real too, but a different severity: that loop awaits a `NodeGetHierarchyQuery` per node, so it yields the event loop every iteration — wasted CPU on hierarchical diffs rather than the loop starvation this PR exists to fix. It now reads through `parent_adder.get_node`, whose map is initialised from the same root and kept in sync by `add_node`.

On indexing the model instead of the call sites, I'd rather not do it here. `EnrichedDiffRoot.get_node` had exactly one production caller (that line) and `has_node` has none, so after it the node level is dict-based everywhere; the three hand-rolled maps are all node-level and each built once, and the thing that actually costs 24s is at the element level, which a node index would not touch. A real model-level index means dict-backed containers replacing the `set` fields on Root/Node/Relationship/Attribute, which reaches `asdict`, both custom `__deepcopy__`s, the repository deserialiser and ~64 `.nodes` sites in `core/diff` alone — more than I want to put in a stable fix. Worth its own issue against develop if you agree.

In db6aab4bd.

backend/infrahub/services/heartbeat.py
fatih-acar: Valid. `RedisCache` builds `redis.Redis(...)` with no `socket_timeout`/`socket_connect_timeout`, so both default to `None`, and the consequence chain is exactly as you describe: the beat blocks, the `except` never fires, the stop flag is only read between beats, and `stop()` then keeps `_thread`, leaving `running` truthy and `start()` a no-op with no way back.

Each beat now carries its own deadline via `asyncio.wait_for`, which is also why the failure path closes the connection — cancelling a command in flight can leave a response unread on it.

Worth flagging for #9451: its 5s socket timeouts do not close this on their own. It also sets `Retry(ExponentialBackoff(cap=1.0, base=0.2), retries=5)`, and redis-py's `Retry` retries `TimeoutError`/`socket.timeout` by default, with `retries` counting retries rather than attempts — so a black-holed connection costs 6 x 5s + 4.2s of backoff, about 34s before the command raises, against a 15s key expiry. What it does fix is the unboundedness: the wedge becomes self-healing after ~34s instead of permanent, which is the important half. The deadline still has to be ours, and NATS is a separate client either way.

In b8d6964ef.

backend/infrahub/services/heartbeat.py
fatih-acar: Both right, and the constants confirm it: interval 10s, `KVTTL.FIFTEEN`, `clean-up-deadlocks` on `* * * * *`, and the reaper takes a single `list_active_worker_ids()` sample with no grace on the absence (`clean_up_deadlocks_interval_mins` gates the lock's age, not the worker's).

The deadline is now anchored before the beat, and a failed beat backs off briefly instead of waiting a full interval. Neither is enough on its own: with the deadline anchored and the beat bounded at one interval, a hang starting at t=10 still writes nothing until ~t=21 against expiry at 15, because TTL/interval = 1.5 tolerates zero missed beats. So the interval drops to 5s as well, three beats per expiry, and a hang at t=5 times out at t=10, backs off, reconnects and writes at ~t=11.

This narrows the window rather than closing it: a worker that genuinely cannot reach the cache still loses its key, and the reaper grace period you are pointing at remains the structural fix.

One thing I did not do is guard the anchoring with a test. The only way to observe it from outside is to time the gap between two writes, and a wall-clock threshold passes on a fast runner with the regression present and flakes on a loaded one without it. So the bound and the reconnect are covered by an effect-polling test and the anchoring by review. If you want it pinned, the clean way is to inject a clock into `WorkerHeartbeat` per `dev/guidelines/backend/testing.md` and assert the computed deadline exactly — happy to add that.

In b8d6964ef.

backend/infrahub/services/heartbeat.py
fatih-acar: Followed up on this properly — my first check used stable's config, which proved nothing about develop.

develop drops `"BLE"` from the ruff `ignore` list (stable is `select = ["ALL"]` plus that ignore), so BLE001 is enforced there. Against the merged tree:

- `git merge-tree origin/develop HEAD` is clean, and ruff with develop's rules over the full merged tree passes on every file this PR touches.
- `ty` 0.0.69 (develop's pin) adds nothing over develop's own baseline — a control run on develop alone returns the identical diagnostics.
- Nothing needs muting at this line: BLE001 exempts handlers that call `log.exception(...)`, and both handlers here do. Adding the `# noqa: BLE001` would break develop with `RUF100 Unused noqa directive`.

`scheduler.py` is the one file develop's rules do flag, and develop already carries `# noqa: BLE001` on that line, so the merge keeps it.

dev/knowledge/backend/async-tasks.md
```diff
@@ -378,6 +378,36 @@ Because a dead in-process retry wait is never re-submitted by a worker, crashing
...
+deadlock cleanup running on another worker deleted the diff-update locks the rebase still held, and
+the rebase failed on lock release with `LockNotOwnedError` after all of its work was done. A thread
+keeps beating through such a stall because a pure-Python loop releases the interpreter lock every few
+milliseconds; only a C extension holding it for longer than the key's expiry could starve the thread.
+The key therefore means "this process is alive", not "this process's event loop is idle", which is
+what every consumer above wants to know.
+
+The thread needs its own cache connection because the asyncio Redis and NATS clients bind to the
+loop that created them. A failed beat closes that connection and the next beat reconnects through the
+factory, so a cache outage delays the heartbeat rather than ending it. A blip between one worker and the cache while the cleanup's
```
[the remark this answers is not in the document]
fatih-acar: Fixed — that paragraph was rewritten for the new heartbeat cadence and now wraps with the rest of the document.

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,179 @@
...
+    async def _beat_until_stopped(self) -> None:
+        cache: InfrahubCache | None = None
+        try:
+            while not self._stop_requested.is_set():
+                # Anchored before the beat rather than after it, so a slow beat eats into the interval
+                # instead of adding to it and pushing the next write past the key's expiry.
+                deadline = time.monotonic() + self.interval_seconds
+                try:
+                    if cache is None:
+                        cache = await asyncio.wait_for(self.cache_factory(), timeout=self.beat_timeout_seconds)
```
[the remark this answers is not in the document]
fatih-acar: Valid, and it was a leak on the ordinary failure path too, not just under cancellation: `NATSCache.new()` connects, then builds the JetStream context and the KV bucket, and anything that fails after the connect orphans it — the caller never receives the cache, so nothing else holds a reference that could close it.

Fixed in `new()` rather than around the timeout, so the cleanup holds however initialisation fails. It catches `BaseException` rather than `Exception` on purpose: the case you describe is cancellation, and `CancelledError` does not inherit from `Exception`.

Covered by a new NATS component test that lets `_ensure_kv` hang, cancels `new()` on a deadline and asserts the connection is closed. It fails with `except Exception` in place, which is the check worth having here.

In 5e7256b80.

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -152,13 +155,15 @@ def _update_diff_relationship_conflicts(
             if not retrieved_rel:
                 retrieved_node.relationships.add(updated_rel)
                 continue
+            # One lookup per updated element instead of a scan of ``retrieved_rel.relationships`` for each
+            # of them, for the same reason as the node loop above: the scan grows with every element added
+            # below, and a cardinality-many relationship can hold tens of thousands of peers.
```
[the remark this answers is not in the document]
fatih-acar: Right on both counts — the comment ran to three lines, named the loop above, and led with the approach not taken. Condensed to one why-sentence stating the constraint: a cardinality-many relationship can hold tens of thousands of peers, so each match has to be a single lookup.

Did the same to the node-level comment a few lines up, which had the identical problem and which you did not flag.

In 2dc4b939f.

backend/tests/unit/services/test_heartbeat.py
```diff
@@ -0,0 +1,245 @@
...
+        heartbeat.start()
+        assert heartbeat._thread is first_thread
+
+
+async def test_a_beat_that_never_returns_is_abandoned_and_its_connection_replaced() -> None:
+    """A connection that stops answering without closing must not hold the thread forever.
+
+    The cache clients impose no deadline of their own, so without one here the beat would block, the
+    stop request would only be seen between beats, and ``stop`` would leave a thread nothing can end.
+    """
```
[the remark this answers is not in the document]
fatih-acar: Agreed, the second paragraph was a counterfactual and named `stop`. Reduced to the first sentence.

Also condensed the handler comment in `heartbeat.py` that had grown to six lines under the same rule; that rationale lives in the class docstring and `dev/knowledge/backend/async-tasks.md`.

In 2dc4b939f.

backend/tests/component/services/adapters/nats/test_nats.py
```diff
@@ -88,3 +90,32 @@ async def test_list_keys(nats: dict[int, int] | None) -> None:
...
+        return connection
+
+    async def never_finishes(**kwargs: Any) -> None:
+        await asyncio.Event().wait()
+
+    monkeypatch.setattr(nats_client, "connect", recording_connect)
+    monkeypatch.setattr(NATSCache, "_ensure_kv", staticmethod(never_finishes))
+
+    with pytest.raises(TimeoutError):
+        await asyncio.wait_for(NATSCache.new(), timeout=1.0)
```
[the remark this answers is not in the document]
fatih-acar: Valid. The 1s window assumed the real connect finished inside it, so a slower one would cancel before `recording_connect` had appended and fail `len(opened) == 1` — failing on the assertion rather than on the behaviour. It is also the kind of wall-clock dependence this PR just ruled out in `dev/guidelines/backend/testing.md`, so it should not have gone in.

Taken your suggested shape: `new()` runs in a task, `recording_connect` sets an event once the connection is established, the test waits on that event and then cancels. No timeout anywhere, so it no longer depends on how long connect takes.

Still bites — with `except Exception` in place of `except BaseException` in `NATSCache.new()` it fails, which is the check that matters here.

In 5858a3f18.

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -109,20 +109,23 @@ def _get_keep_branch_for_enriched_conflict(
             return BranchConflictKeep.SOURCE
         return None
 
-    def _update_diff_conflicts(self, updated_diff: EnrichedDiffRoot, retrieved_diff: EnrichedDiffRoot) -> None:
+    @classmethod
```
ajtmccarty: this looks like it should still be an instance method to me unless I am missing something

backend/infrahub/core/diff/data_check_synchronizer.py
```diff
@@ -109,20 +109,23 @@ def _get_keep_branch_for_enriched_conflict(
...
+                retrieved_nodes_by_identifier[updated_node.identifier] = updated_node
                 continue
             retrieved_node.conflict = updated_node.conflict
-            self._update_diff_attr_conflicts(updated_node=updated_node, retrieved_node=retrieved_node)
-            self._update_diff_relationship_conflicts(updated_node=updated_node, retrieved_node=retrieved_node)
+            cls._update_diff_attr_conflicts(updated_node=updated_node, retrieved_node=retrieved_node)
+            cls._update_diff_relationship_conflicts(updated_node=updated_node, retrieved_node=retrieved_node)
 
-    def _update_diff_attr_conflicts(self, updated_node: EnrichedDiffNode, retrieved_node: EnrichedDiffNode) -> None:
+    @staticmethod
```
ajtmccarty: same here and with `_update_diff_relationship_conflicts`. these are both instance methods that should not be called from outside of this class. whatever linter/type-checker is forcing these to become `staticmethod`s is wrong. probably not worth changing the typing rules in this PR, but I think this is wrong

backend/infrahub/services/heartbeat.py
```diff
@@ -0,0 +1,175 @@
...
+    however long the beat took. The bound matters because the cache clients do not impose one: a
+    connection that stops answering without closing (a load balancer dropping an idle connection, a
+    failover without an RST) would otherwise block the beat indefinitely, and ``stop`` cannot end a
+    thread that is inside such a call.
+    """
+
+    def __init__(
+        self,
+        component_type: ComponentType,
+        cache_factory: CacheFactory = build_heartbeat_cache,
```
ajtmccarty: cache should be passed in wherever WorkerHeartbeat is constructed instead of passing in a factory that then needs to reach for settings

backend/infrahub/services/scheduler.py
```diff
@@ -28,42 +31,44 @@ class Schedule:
...
+    """
+
     # TODO we could remove service dependency by adding kwargs to Schedule instead of passing services
     service: InfrahubServices | None
 
-    def __init__(self, component_type: ComponentType) -> None:
+    def __init__(self, component_type: ComponentType, heartbeat: WorkerHeartbeat | None = None) -> None:
         self.running: bool = False
         self.schedules: list[Schedule] = []
+        self.heartbeat: WorkerHeartbeat | None = heartbeat
```
ajtmccarty, line 47: is there a reason `heartbeat` should not always be required?

backend/infrahub/services/scheduler.py
```diff
@@ -28,42 +31,44 @@ class Schedule:
...
-                    name="branch_refresh", interval=900, function=trigger_branch_refresh, start_delay=random_number
-                ),
-            ]
-            self.schedules.extend(schedules)
-
-        if component_type == ComponentType.GIT_AGENT:
-            schedules = [
-                Schedule(name="refresh_components", interval=10, function=refresh_heartbeat),
+        if component_type in HEARTBEAT_COMPONENT_TYPES:
+            self.heartbeat = heartbeat or WorkerHeartbeat(component_type=component_type)
```
ajtmccarty: we want to avoid this. `WorkerHeartbeat` should always be passed in if it is required

backend/tests/unit/core/diff/test_data_check_synchronizer.py
ajtmccarty, line 1: nice unit tests

ajtmccarty (approved):

backend/infrahub/services/heartbeat.py
fatih-acar: Half of this is done: `cache_factory` is now a required parameter rather than defaulting to the module-level one that reaches for settings, so the dependency is named at the wiring site.

It has to stay a factory rather than a cache instance, though. The asyncio Redis and NATS clients bind to the event loop that creates them, and this heartbeat runs its own loop on its own thread — that is the whole point of the class, since an asyncio schedule on the main loop is what a CPU-bound flow starves. A cache built on the main loop and passed in could not be used from that thread. On top of that, a beat that fails or times out closes its connection so the next one does not retry a broken client, which needs something that can produce a fresh cache rather than a single instance.

So the factory is load-bearing, but you were right that it should not have carried a default.

In 664563cfb.

backend/infrahub/services/scheduler.py
fatih-acar: No good reason — done. `heartbeat` is now a required parameter (still `WorkerHeartbeat | None`, since a CLI or test process reports no liveness), and `InfrahubServices.new` decides whether this process needs one and builds it there.

In 664563cfb.

backend/infrahub/services/scheduler.py
fatih-acar: Agreed, the internal fallback is gone. The scheduler now stores whatever it is handed and only decides which schedules to register; `InfrahubServices.new` builds the `WorkerHeartbeat` for the component types that report liveness.

In 664563cfb.

backend/infrahub/core/diff/data_check_synchronizer.py
fatih-acar: Correction to my reply above — I've taken your original suggestion instead.

They are back to being three instance methods on the class, with no decorators, exactly the shape they had before this PR. Moving them to module level fixed the decorators but rewrote most of the file to do it, and the real change here is two lookups. `data_check_synchronizer.py` now differs from the base by +10/-8, all of it the node map and the peer map.

The tests build an instance with no collaborators, since the merge reads none of them.

In 2178c14d7.

backend/infrahub/services/heartbeat.py
ajtmccarty: okay I think I get it now

PR #10594, opened 2026-09-10
fix: clear the mypy operator/return suppression on infrahub.message_bus.operations [INBOX-31]

saltas888:
Clears the package-wide mypy override that disabled the `operator` and `return` error codes for
`infrahub.message_bus.operations`, and fixes the two errors it was hiding.

Jira: **INBOX-31**

## What was changed and why

The suppression was blanket — it applied to all 9 modules in the package, forever, though the errors
it hid were confined to one file. That package is the boundary where every message the worker
consumes is decoded, dispatched, and routed to the reply / retry / dead-letter protocol, so any
*future* `operator`/`return` fault there was being swallowed too.

**`backend/infrahub/message_bus/operations/__init__.py`**

- `COMMAND_MAP` was unannotated and heterogeneous — 4 entries are plain `async def` handlers, 4 are
  Prefect `@flow` objects — so mypy joined the value type to `object` and the dispatch call
  `await func(message=message)` was `"object" not callable`. It is now
  `COMMAND_MAP: dict[str, MessageHandler]` with `MessageHandler = Callable[..., Awaitable[None]]`.
- `execute_message` now returns `None` explicitly on its success path instead of falling off the end.
  That return is load-bearing: in `services/adapters/message_bus/nats.py` a truthy return means
  `message.nak(delay / 1000)` (re-queue) and a falsy one means `message.ack()`.

**`pyproject.toml`** — two blocks deleted, not one:

- the `[[tool.mypy.overrides]]` for the package (the ticket's target);
- the `[[tool.ty.overrides]]` scoped to the same file, which the new annotation makes redundant.
  Verified both ways so "removed" isn't confused with "was already dead": with the original code and
  that block deleted, `ty` reports **8 × `invalid-argument-type`** (one per handler); with the new
  typing it reports `All checks passed!`.

**No `# type: ignore` and no `cast` were added anywhere.** Net suppressions go down by two.

### On `Callable[..., Awaitable[None]]`

The `...` is deliberate, not a shrug. Handlers each take a *different* `InfrahubMessage` subtype
while the dispatch site holds only the base type, so parameters cannot be pinned without widening
every handler's own signature. A callback `Protocol` was tried and rejected: parameter
contravariance makes it emit 8 `dict-item` errors, one per handler.

The annotation still has teeth — injecting `"bad.not_callable": 42` and `"bad.sync_fn": len` into
the map produces two `Dict entry N has incompatible type` errors, so both a non-callable and a
callable-but-synchronous value are rejected. The shape mirrors
`MessageFunction = Callable[[InfrahubMessage], Awaitable[None]]`, already used in the same
subsystem's nats/rabbitmq adapters.

## Testing performed

`backend/tests/unit/message_bus/test_operations.py` is new. Nothing previously unit-tested
`execute_message` at all — `tests/adapters/message_bus.py` drives it for every message the backend
suite sends but discards the return value, so the retry decision was pinned only by an integration
test. The new file covers all five dispatch outcomes and asserts the **returned value** in each, not
just the side effect.

Two checks worth calling out:

- **The tests pass against the pre-change code as well.** That is the direct evidence that this
  change is behaviour-neutral — they characterize existing behaviour rather than guard a fix.
- **The flow-unwrap test was fixed after review found it didn't bite.** Deleting
  `if skip_flow and isinstance(func, Flow): func = func.fn` still passed, because Prefect's
  `Flow.__call__` runs the async function through the engine regardless — only the runtime changed
  (0.04s → 3.94s). It now records `FlowRunContext.get()` inside the handler and asserts it is `None`,
  which holds only when the bare coroutine ran. Confirmed: passes unmutated, fails with the unwrap
  removed.

Local gate, all green:

| Check | Result |
|---|---|
| `uv run mypy backend/infrahub/message_bus/operations` | `Success: no issues found in 9 source files` |
| `uv run ty check backend/infrahub/message_bus/operations` | `All checks passed!` |
| `uv run invoke backend.lint` (mypy repo-wide) | `Success: no issues found in 1686 source files` |
| `uv run invoke backend.test-unit` | `2656 passed` |
| `uv run pytest backend/tests/unit/message_bus` | `10 passed` |
| `ruff check` / `ruff format --check` / `ty check .` / `yamllint` / `uv lock --check` | clean |
| `backend.validate-generated` | no diff |
| frontend `biome ci` / `knip` / `betterer ci` | clean (unchanged by this PR) |

`test_mappings.py` passes unmodified. Its stale exclusion list (`event.node.mutated`,
`trigger.webhook.actions` — keys no longer in the map) is pre-existing and deliberately left alone.

No changelog fragment and no spec-kit design record: per `dev/guidelines/git-workflow.md`, a
trivial code-only change with no behaviour change ships as just the code diff.

https://claude.ai/code/session_013i34mfdwmNwDupcV621Eor

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/infrahub/message_bus/operations/__init__.py
+++ b/backend/infrahub/message_bus/operations/__init__.py
@@ -1,3 +1,5 @@
+from collections.abc import Awaitable, Callable
+
 import ujson
 from prefect import Flow
 
@@ -8,7 +10,12 @@ from infrahub.message_bus.types import MessageTTL
 from infrahub.services.adapters.message_bus import InfrahubMessageBus
 from infrahub.tasks.check import set_check_status
 
-COMMAND_MAP = {
+# Handlers are heterogeneous by design: each takes a specific InfrahubMessage subtype, and half are
+# wrapped as Prefect flows. The dispatch site only has the base type, so the parameters cannot be
+# pinned here — but the callable-and-awaits-to-None shape can be, and is.
+MessageHandler = Callable[..., Awaitable[None]]
+
+COMMAND_MAP: dict[str, MessageHandler] = {
     "git.file.get": git.file.get,
     "git.repository.connectivity": git.repository.connectivity,
     "refresh.git.fetch": git.repository.fetch,
@@ -31,6 +38,7 @@ async def execute_message(
         if skip_flow and isinstance(func, Flow):
             func = func.fn
         await func(message=message)
+        return None
     # Message-bus boundary: any handler failure must be routed to the reply/retry/dead-letter protocol, never crash the consumer
     except Exception as exc:  # noqa: BLE001
         if message.reply_requested:
--- a/backend/infrahub/message_bus/operations/__init__.py
+++ b/backend/infrahub/message_bus/operations/__init__.py
@@ -10,9 +10,8 @@ from infrahub.message_bus.types import MessageTTL
 from infrahub.services.adapters.message_bus import InfrahubMessageBus
 from infrahub.tasks.check import set_check_status
 
-# Handlers are heterogeneous by design: each takes a specific InfrahubMessage subtype, and half are
-# wrapped as Prefect flows. The dispatch site only has the base type, so the parameters cannot be
-# pinned here — but the callable-and-awaits-to-None shape can be, and is.
+# Each handler takes a distinct InfrahubMessage subtype while the dispatch site holds only the base
+# type, so the parameters cannot be pinned.
 MessageHandler = Callable[..., Awaitable[None]]
 
 COMMAND_MAP: dict[str, MessageHandler] = {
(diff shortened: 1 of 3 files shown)


ogenstad (approved):
Always nice to clear those things out. There's also a bug in `ty` that at times errors out on those lines so that should go away with this. Could potentially merge it to `stable` instead, but that bug doesn't show up very often so `develop` is probably fine as well.

PR #10598, opened 2026-09-10
Match the owner automation on creations

gmazoyer:
# Why

IFC-3043. An update to a node of a computed attribute's own kind started two flows for that node.
The owner automation and the query automation built for the same kind derive their field filter
from the same thing, the fields the transform query reads of that kind, so both answered the same
event and recomputed the same value.

Last PR of the live-path narrowing stack. It targets `gma-20260910-ifc-3044` (#10596).

## What changed

The owner automation matches creations. A created node is a member of no query group yet, so
nothing else can start its first computation. An updated node is reached through the group it
subscribed to when it last computed, which the query automation of its own kind resolves.

## How to review

**One case keeps both events, and this is the part I would look at.** The gather builds no query
automation for a read kind with no read fields, because a trigger with nothing to filter on fires
on every update of that kind. `fields_by_kind` and that skip test read the same set, so an owner
kind has a query automation exactly when the owner trigger has a field filter. Making the owner
automation created-only unconditionally would therefore leave an update to the owner reached by
nothing whenever its transform query reads none of its fields. The field filter decides: with
fields to match, creations only; without, creations and updates as before. A component test pins
the exception.

The same-kind cross-instance read is unaffected. An attribute reading another node of its own kind
is recomputed through the query automation, which resolves readers from group membership and never
depended on the owner one.

The gap this leaves is the one the reader path already has: a node whose first computation failed
never subscribed, so an update to it starts no recompute until a schema change, a transform edit or
a merge refreshes the attribute. The knowledge page records it. Not fixed here.

## How to test

```bash
INFRAHUB_USE_TEST_CONTAINERS=1 uv run pytest backend/tests/component/computed_attribute/ --neo4j
uv run pytest backend/tests/integration_docker/test_computed_attributes.py
```

Local results: component 52 passed, integration_docker 11 passed.

`test_an_owner_update_recomputes_the_owner_once` is the measured claim: it updates a device's
`instance`, a field the transform query reads, and asserts the `process_transform` count for
`InfraDevice.name` rises by exactly one. Reverting the trigger change and rebuilding the image
makes that test, and only that test, fail.

## Checklist

- [x] Tests added/updated
- [x] Changelog entry added (one fragment for the three PRs of the stack)
- [ ] External docs updated (not user-facing)

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10598?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


gmazoyer:
Closing this one. The change is not safe as it stands.

Making the owner automation match creations only moves an owner update onto the query-group path. That index is written after the value, not with it. Measured on a running stack:

```
09:51:15.597 -> 09:51:17.669  process_transform   value written
09:51:18.345 -> 09:51:21.535  group-update        query group committed
09:51:21.188 -> 09:51:21.570  query-target        ran 0.35s early, found nothing
```

The target flow ran before the group existed, submitted nothing and completed green. Nothing fires again, so the update is lost and not just late. Before this change the owner automation submitted `process_transform` directly with the node id and never read the group, so the path had no such window.

The double fire this removed is idempotent, so keeping it costs work and not correctness. The stack keeps #10595 and #10596, which carry the measured drop from 42 ids for 21 nodes to 20 for 20.

Details on IFC-3043. The branch stays in place if we want to pick it up once the group write is ordered before the value.

PR #10606, opened 2026-09-11
feat(graphql): InfrahubRepositoryBranchStatus per-branch attribute read from the graph (IFC-3127)

ogenstad:
Closes [IFC-3127](https://opsmill.atlassian.net/browse/IFC-3127). PR 3 of 5 under
[IFC-3104](https://opsmill.atlassian.net/browse/IFC-3104). **Base: `cross-branch-repo-status-infp-671`.**

Replaces the fabricated attribute values from increment A with a real per-branch read through a new
core primitive, makes the three value filters real, and deletes the stub. **The GraphQL contract
does not change** beyond dropping the "preview" suffix from the field description.

## What landed

- `RepositoryBranchAttributesQuery` in `backend/infrahub/core/query/repository.py`: one statement
  resolving N branches x M attributes for one or more repositories, with the per-branch visibility
  predicate and the standard election order.
- `RepositoryBranchAttributesReader` in `backend/infrahub/core/repository_branch_status/reader.py`,
  behind the existing `RepositoryBranchAttributesSource` protocol.
- Resolver swapped from `stub.py` (deleted) to the reader; `apply_value_filters` wired in so
  `sync_status__value`, `internal_status__value` and `own_values_only` narrow rows and `count`
  server-side.
- A `Branch(name)` RANGE index (see Performance).
- `at` is now rejected on this field (see Behaviour notes).
- 119 tests, a query benchmark, two `dev/knowledge` pages, a docs section, two changelog fragments.

## Sign-off requested

- **Database:** one new index declaration. `node_indexes` is declarative and applied by
  `index_manager.init` at startup and by `infrahub db index`, so no numbered migration and existing
  deployments self-heal on restart.
- **GraphQL:** no SDL change other than the field description. `schema/schema.graphql` and the
  frontend generated types are regenerated and committed.

## Performance: the `Branch(name)` index

The statement opens `UNWIND $branch_names AS branch_name MATCH (br:Branch {name: branch_name})`.
There was no `Branch` entry in `backend/infrahub/core/graph/index.py`, so that compiled to
`NodeByLabelScan` under an `Apply`, rescanning every `Branch` node per unwound name: O(branch_names
x Branch-node-count), quadratic in branch-set size.

Measured at 200 branches, median of 30 timed runs after warmup, A/B/A/B:

| branches | without index | with index |
| --- | --- | --- |
| 5 | 3.2 ms | 2.1 ms |
| 50 | 8.9 ms | 5.7 ms |
| 200 | 30.0 ms | 22.3 ms |

Server-only time at 200 branches: **13 ms to 6 ms**. `PROFILE` db-hits: **118,600 to 38,000**, the
whole delta being `NodeByLabelScan` (40,600) plus `Filter br.name = branch_name` (40,400)
collapsing into a single `NodeIndexSeek` at 400. The client-side ratio is diluted by ~16 ms of
Python hydration for 1200 rows, which no index touches.

The index does **not** measurably help `Branch.get_by_name` on the registry cache-miss path
(db-hits 428 to 24, wall-clock unmoved) - that lookup is round-trip-bound. It is justified by this
query alone.

## EXPLAIN plan (T034), 200 branches, 1 repository, 3 attributes

All four criteria pass:

- `Branch.name` uses `NodeIndexSeek` (this failed before the index)
- `Node.uuid` uses `NodeIndexSeek`
- no `CartesianProduct`
- no eager operator between the `UNWIND` and the subqueries

```
+ ProduceResults@neo4j  repository_id, branch_name, attribute_name, attribute_id, value, value_branch, updated_at
  + Projection@neo4j  cache[a.uuid] AS attribute_id, cache[r_value.branch] AS value_branch, cache[n.uuid] AS repository_id, av.value AS value, cache[r_value.from] AS updated_at, cache[a.name] AS attribute_name
    + CacheProperties@neo4j  cache[a.name], cache[a.uuid]
      + Filter@neo4j  cache[r_value.status] = $autostring_1
        + Apply@neo4j
          + Filter@neo4j  cache[r.status] = $autostring_0
            + Apply@neo4j
              + Distinct@neo4j  n, a, branch_name, default_window
                + Filter@neo4j  (a:Attribute AND cache[a.name] IN $attribute_names)
                  + Expand(All)@neo4j  (n)-[:HAS_ATTRIBUTE]->(a)
                    + Apply@neo4j
                      + Projection@neo4j  CASE
  WHEN br.is_isolated THEN br.branched_from
  ELSE $at
END AS default_window
                        + Apply@neo4j
                          + Unwind@neo4j  $branch_names AS branch_name
                          + NodeIndexSeek@neo4j  RANGE INDEX br:Branch(name) WHERE name = branch_name
                      + NodeIndexSeek@neo4j  RANGE INDEX n:Node(uuid) WHERE uuid = $repository_ids[0], cache[n.uuid]
              + Top@neo4j  `r.branch_level` DESC, `r.from` DESC, `r.status` ASC LIMIT 1
                + Projection@neo4j  cache[r.branch_level] AS `r.branch_level`, cache[r.from] AS `r.from`, cache[r.status] AS `r.status`
                  + CacheProperties@neo4j  cache[r.branch_level], cache[r.from], cache[r.status]
                    + Distinct@neo4j  n, a, branch_name, r, default_window
                      + Union@neo4j
                        + Union@neo4j
                          + Union@neo4j
                            + Filter@neo4j  n = anon_9 AND a = anon_10 AND (r.to IS NULL AND cache[r.from] <= $at)
                              + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_9)-[r:HAS_ATTRIBUTE(branch)]->(anon_10) WHERE branch IN [branch_name, $global_branch_name]
                            + Filter@neo4j  cache[r.from] <= $at AND n = anon_19 AND a = anon_20 AND r.to > $at
                              + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_19)-[r:HAS_ATTRIBUTE(branch)]->(anon_20) WHERE branch IN [branch_name, $global_branch_name]
                          + Filter@neo4j  NOT branch_name = $default_branch_name AND cache[r.from] <= default_window AND n = anon_29 AND a = anon_30 AND r.to IS NULL
                            + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_29)-[r:HAS_ATTRIBUTE(branch)]->(anon_30) WHERE branch = $default_branch_name
                        + Filter@neo4j  NOT branch_name = $default_branch_name AND cache[r.from] <= default_window AND n = anon_39 AND a = anon_40 AND r.to > default_window
                          + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_39)-[r:HAS_ATTRIBUTE(branch)]->(anon_40) WHERE branch = $default_branch_name
          + Top@neo4j  `r_value.branch_level` DESC, `r_value.from` DESC, `r_value.status` ASC LIMIT 1
            + Projection@neo4j  cache[r_value.branch_level] AS `r_value.branch_level`, cache[r_value.from] AS `r_value.from`, cache[r_value.status] AS `r_value.status`
              + CacheProperties@neo4j  cache[r_value.branch_level], cache[r_value.from], cache[r_value.status]
                + Filter@neo4j  av:AttributeValue
                  + Distinct@neo4j  default_window, a, r_value, av, branch_name
                    + Union@neo4j
                      + Union@neo4j
                        + Union@neo4j
                          + Filter@neo4j  a = anon_67 AND (r_value.to IS NULL AND cache[r_value.from] <= $at)
                            + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_67)-[r_value:HAS_VALUE(branch)]->(av) WHERE branch IN [branch_name, $global_branch_name], cache[r_value.branch]
                          + Filter@neo4j  cache[r_value.from] <= $at AND a = anon_72 AND r_value.to > $at
                            + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_72)-[r_value:HAS_VALUE(branch)]->(av) WHERE branch IN [branch_name, $global_branch_name], cache[r_value.branch]
                        + Filter@neo4j  NOT branch_name = $default_branch_name AND cache[r_value.from] <= default_window AND a = anon_77 AND r_value.to IS NULL
                          + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_77)-[r_value:HAS_VALUE(branch)]->(av) WHERE branch = $default_branch_name, cache[r_value.branch]
                      + Filter@neo4j  NOT branch_name = $default_branch_name AND cache[r_value.from] <= default_window AND a = anon_82 AND r_value.to > default_window
                        + DirectedRelationshipIndexSeek@neo4j  RANGE INDEX (anon_82)-[r_value:HAS_VALUE(branch)]->(av) WHERE branch = $default_branch_name, cache[r_value.branch]
```

## Error-path review (T046)

Audited against `dev/guidelines/backend/exceptions.md`:
`backend/infrahub/core/repository_branch_status/`, `backend/infrahub/core/query/repository.py`,
`backend/infrahub/graphql/queries/repository_branch_status/`.

**1. `except` clauses - PASS.** The only `except` clause in the whole slice is
`permissions.py:61`, catching exactly `InitializationError` and re-raising `PermissionDeniedError
from exc` after a `log.warning`. The other grep hits are `from infrahub.exceptions import ...`
lines, not handlers. No bare `except`, no `except Exception`, no `# noqa: BLE001`, no `try` around
code that cannot raise. `core/query/repository.py` and every module under
`core/repository_branch_status/` contain no `except` at all.

**2 and 3. Every raise, its class, and the test that provokes it:**

| # | Site | Class | Infrahub `Error` subclass? | Provoking test |
| --- | --- | --- | --- | --- |
| 1 | `core/repository_branch_status/models.py:38` | `ValueError` | **No** - see note | `unit/.../test_repository_branch_status.py::TestRepositoryBranchAttributes::test_duplicate_triple_raises` |
| 2 | `kind_dispatch.py:66` | `ValidationError` | Yes | `unit/...::TestPolicyForKind::test_unsupported_kind_names_the_kind` |
| 3 | `permissions.py:43` | `PermissionDeniedError` | Yes | `component/...::TestRepositoryBranchStatusPermissions::test_view_permission_matrix[default-branch-no-grant]`, `[user-branch-no-grant]`, `::test_denial_runs_no_database_query`, `::test_anonymous_session_without_a_grant_is_denied` |
| 4 | `permissions.py:56` | `PermissionDeniedError` | Yes | `component/...::test_a_grant_on_one_kind_does_not_cover_the_other_kind` |
| 5 | `permissions.py:67` | `PermissionDeniedError` | Yes | `component/...::test_context_without_loaded_permissions_is_denied` |
| 6 | `resolver.py:57` (`at`) | `ValidationError` | Yes | `component/...::test_a_request_for_a_past_point_in_time_is_rejected` (negative control: `::test_a_request_carrying_no_at_parameter_is_accepted`) |
| 7 | `resolver.py:117` (`limit`) | `ValidationError` | Yes | `component/...::test_invalid_paging_arguments_are_rejected[limit-below-one]`, `[explicit-null-limit]` |
| 8 | `resolver.py:119` (`offset`) | `ValidationError` | Yes | `component/...::test_invalid_paging_arguments_are_rejected[negative-offset]`, `[explicit-null-offset]` |
| 9 | `resolver.py:141` (non-repository node) | `NodeNotFoundError` | Yes | `component/...::test_a_node_of_another_kind_is_reported_as_not_found` |

`core/query/repository.py` raises nothing of its own; nor do `reader.py`, `interface.py`,
`field.py`, `paging.py`, `payload.py`.

Two errors propagated unchanged rather than raised here, both covered: `NodeNotFoundError` from the
repository lookup (`::test_unknown_repository_is_reported_as_not_found`) and `ValidationError` from
`standard_node_ordering_from_order_input` (`::test_contradictory_order_is_rejected`).

**Deviation, row 1.** `RepositoryBranchAttributes.from_values` raises a plain `ValueError` on a
duplicate `(repository, branch, attribute)` triple, deliberately not an Infrahub `Error`. The
Cypher makes the duplicate impossible (`WITH DISTINCT` before the election, `LIMIT 1` in each
subquery), so reaching it is a server-side data-integrity bug; an `Error` subclass would render it
to the client as a clean, actionable message with `data: null` as though the caller had erred.
**Reviewer input welcome** - this is a knowing deviation from T046's wording.

**4. Miss renders null, database errors propagate - PASS.** `build_attribute_payload` returns
`None` for a `None` value, which serialises as GraphQL `null`; covered end to end by
`::test_a_repository_that_was_never_imported_returns_rows_without_a_commit` and
`::test_ref_is_null_on_the_read_write_kind_and_set_on_the_read_only_kind`. No fabricated value at
any layer. There is no handler on the read path at all, so driver errors reach the GraphQL error
formatter unmodified.

## Behaviour notes

- **`at` is rejected on this field.** The rows come from `Branch.get_list`, which takes no `at`, so
  honouring `?at=` would pair a current branch set with historical values: rows for branches that
  did not exist then, none for branches deleted since. The guard reads
  `request.query_params.get("at")` rather than `graphql_context.at`, because
  `prepare_graphql_params` coerces via `Timestamp(at)` and `Timestamp(None)` is now, so
  `graphql_context.at` is never `None`. The `at` parameter is still plumbed to the reader for
  increment C.
- **Known limitation:** the guard covers the `/graphql` endpoint. A saved `CoreGraphQLQuery`
  fetched through `/api/query/<name>?at=` or the transformation endpoints passes no `request` into
  `prepare_graphql_params`, so it would still return current branches with historical values. That
  is not an expected way to run this query; it becomes its own issue if it ever comes up.
- Value filters now widen the read by the attribute they compare, so `sync_status__value` works
  whether or not the document selects `sync_status`. This was a bug found in review: before the
  fix, a filter without the matching field in the selection returned `count: 0` and no rows, with
  no error.
- For `CoreReadOnlyRepository`, `commit` and `ref` are `AWARE` and land on the current branch at
  creation, so `own_values_only` is true on the default branch without an import, and a branch
  forked before the repository existed resolves no `commit`/`ref`. Documented.

## Testing

119 tests: 10 primitive component, 82 GraphQL component, 27 unit. Plus an opt-in query benchmark in
`backend/tests/query_benchmark/`. `/pre-ci` green; re-verified after the review fixes (2523 backend
unit tests, whole-repo ruff, mypy on 1670 files, all four generated-file validations, docs lint).

Quickstart B1 through B8 are covered, `stub.py` is deleted, and the query count for one page is
equal at 5 and 200 branches.

## Known gaps, deferred deliberately

Recorded in `dev/specs/infp-671-cross-branch-repo-status/opsmill-implement-report-phase4.md`:

- No *value* coverage for `CoreReadOnlyRepository`, whose `AWARE` `commit`/`ref` take a different
  edge-placement path. The most likely place for a surprise.
- The differential test cannot catch a strict/non-strict boundary flip, because every fixture
  timestamp is strictly ordered.
- `default_window` omits the `at > branched_from` guard that `get_branches_and_times_to_query_global`
  applies; unreachable from GraphQL now that `at` is rejected, but the primitive still accepts one.
- No explicit ceiling on the result set: `self.limit` steers into the single-shot branch, so
  `query_with_size_limit` chunking never applies.
- No coverage for a soft-deleted value or repository.

## Not done

T047's acceptance names "Jira subtask from T032 closed". **No such issue exists** - IFC-3104's
children are IFC-3125 through IFC-3132. T032 asked for a "Remove InfrahubRepositoryBranchStatus
stub" task that was never created. Moot in substance, since the stub is deleted in this PR.

[IFC-3127]: https://opsmill.atlassian.net/browse/IFC-3127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-3104]: https://opsmill.atlassian.net/browse/IFC-3104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/infrahub/core/query/repository.py
```diff
@@ -20,10 +29,140 @@ class RepositoryBranchAttributeValue:
...
+            "repository_ids": self.repository_ids,
+            "branch_names": self.branch_names,
+            "attribute_names": self.attribute_names,
+            "default_branch_name": self.default_branch_name,
+            "global_branch_name": self.global_branch_name,
+            "at": self.at.to_string(),
+        }
+
+        query = """
+MATCH (n:Node)-[:HAS_ATTRIBUTE]->(a:Attribute)
```
ajtmccarty: might be better to start with just the `Node`s b/c the indexing on those is better to be sure that the query planner uses the better index

backend/infrahub/core/query/repository.py
```diff
@@ -20,10 +29,140 @@ class RepositoryBranchAttributeValue:
...
+// One HAS_ATTRIBUTE edge exists per branch that touched the attribute, so the match above yields
+// one row per edge; the election below has to run once per (node, attribute, branch) instead.
+// Deduplicating before the branch list multiplies the rows also keeps this match off the branch
+// loop, so the uuid lookup and the attribute expansion run once instead of once per branch.
+// ----------
+WITH DISTINCT n, a
+UNWIND $branch_names AS branch_name
+MATCH (br:Branch {name: branch_name})
+// ----------
+// An isolated branch reads the default branch as of its fork point; a branch saved before
```
ajtmccarty: I don't think branch isolation is a concern. non-isolated branches have been deprecated for a while now

backend/infrahub/core/query/repository.py
```diff
@@ -20,10 +29,140 @@ class RepositoryBranchAttributeValue:
...
+// isolation became the default reads it at query time, like the standard per-branch read does.
+// The fork point only narrows a window that extends past it: for a time before the branch existed
+// the requested time is already the tighter bound, and moving forward to the fork would expose
+// default-branch writes made after the time asked for.
+// ----------
+WITH n, a, branch_name,
+     CASE WHEN br.is_isolated AND br.branched_from < $at THEN br.branched_from ELSE $at END AS default_window
+CALL (n, a, branch_name, default_window) {
```
ajtmccarty, line 113: without the non-isolated branch concern, can this use the standard `branch.get_query_filter_path()` instead of this custom filtering?

backend/infrahub/core/query/repository.py
ogenstad: I've updated this, according to tests done by Claude there's no performance impact as we have indexing for this approach as well. However your suggestion matches the style of other queries so I've changed it to match.

backend/infrahub/core/query/repository.py
ogenstad: No I guess you are right. It just feels weird to have something internally that we don't use. We should really get rid of the isolated branch concept throughout the code base..

backend/infrahub/core/query/repository.py
ogenstad: I removed the non-isolated branch stuff, but I'd still want to keep this while we could have a loop and send each branch through `.get_query_filter_path()`  that would mean that the query grows or shrinks depending on the branches we have in scope and we'd create a new plan for each.

ajtmccarty: ah right. multiple branches at once in one query. `get_query_filter_path()` won't work here

backend/infrahub/core/query/repository.py
```diff
@@ -20,10 +29,140 @@ class RepositoryBranchAttributeValue:
...
+    """Value the winning value edge points at, None when the attribute holds no value."""
 
     own_value: bool
     """True when the winning value edge lives on `branch_name` itself rather than being inherited."""
 
     updated_at: str | None
     """Time the winning value edge was created."""
+
+
+class RepositoryBranchAttributesQuery(Query):
```
ajtmccarty, line 41: I don't think there's really anything repository-specific about this query other than the variable names. I think you could call this `ObjectBranchAttributesQuery` and move it somewhere more general for possible reuse elsewhere. and update some names and the docstring. not a requirement, but something to consider

backend/infrahub/graphql/queries/repository_branch_status/field.py
```diff
@@ -20,23 +21,25 @@
...
     Args:
         db: Database connection the source reads through.
 
     Returns:
-        The attribute source, which currently serves placeholder values.
+        The attribute source.
 
     """
-    return StubRepositoryBranchAttributesSource(default_branch_name=registry.default_branch)
+    return RepositoryBranchAttributesReader(
```
ajtmccarty, line 40: very nice dependency injection

backend/tests/component/graphql/queries/test_repository_branch_status.py
ajtmccarty, line 1: might be worth splitting this file up

changelog/+branch-name-index.changed.md
ajtmccarty, line 1: does this need a changelog? the performance increase is only going to be noticeable for someone with 1000s of branches

dev/knowledge/backend/query-pattern.md
```diff
@@ -253,6 +253,47 @@ Example: `NodeGetListByAttributeValueQuery` and `NodeGetByHFIDQuery` chain three
...
+
+A trap worth knowing when reasoning about what a branch "should" see: for a node kind marked `AGNOSTIC`, `BaseAttribute.get_create_data` writes an `AGNOSTIC` or `LOCAL` attribute on the **global** branch at creation time, while later updates go to the branch being written on. An `AWARE` attribute is the exception - it lands on the current branch even on an `AGNOSTIC` node, at creation as at every later write. Global edges are visible from every branch regardless of fork point, so a node created *after* a branch forked still resolves the **creation** values of its global attributes on that branch; only the subsequent writes, and the `AWARE` attributes throughout, are windowed by the fork point. Repositories are the case in point, and the two kinds differ: `CoreRepository.commit` is `LOCAL`, so a read-write repository added today shows its initial `commit` on every branch and only later imports are branch-scoped, while `CoreReadOnlyRepository` overrides `commit` and `ref` to `AWARE`, so a read-only repository created on the default branch shows neither on a branch that forked before it.
```
ajtmccarty: this sounds like a bug
> for a node kind marked `AGNOSTIC`, `BaseAttribute.get_create_data` writes an `AGNOSTIC` or `LOCAL` attribute on the **global** branch at creation time, while later updates go to the branch being written on

and I'd say the `CoreRepository` / `CoreReadOnlyRepository` mentions don't really belong here

ajtmccarty (approved):

backend/tests/component/graphql/queries/test_repository_branch_status.py
ogenstad: I looked at that, but decided against it. The reason is that each split adds a fair bit to each CI run.

changelog/+branch-name-index.changed.md
ogenstad: Good point. Will remove.

backend/infrahub/core/query/repository.py
ogenstad: I somewhat agree, but I'm also not certain that I see a use for this type of query elsewhere. Instead of preparing for a future we don't know if we'll get to I'll leave it there for now. One thing to consider though is if we should have some policy for the future to look through the existing code base and see if we could alter something that's only used in one location to make it more generic.

dev/knowledge/backend/query-pattern.md
ogenstad: The mention of repos have been removed.

For the bug part, there was an issue with the way it was described here. But there still is an asymmetry in the behaviour. Do you think we should open a bug for how get_create_data() behaves?

PR #10612, opened 2026-09-11
Skip a missing transform on the coalesced pass

gmazoyer:
# Why

When the Python target resolution of the coalesced merge/rebase pass fails, the pass widens every
Python computed attribute the schema declares to its whole kind. That set is built from the schema
alone, so it can hold an attribute the database has nothing to run for. The whole-kind fan-out
listed the kind first, and every chunk then raised `Unable to fetch transform`.

This is option 4 of IFC-3162, agreed in the review of #10477
(https://github.com/opsmill/infrahub/pull/10477#discussion_r3986355426).

## What changed

Only a run the resolution could not narrow may wait out a state nothing can compute, and only two
states qualify: an attribute with no transform configured, and one whose transform is absent.

- **`widened` is the precondition, not `coalesced`.** Both submission shapes carry `coalesced: true`,
  so gating on it would have let a batch of resolved ids skip too, and that resolution had already
  found the transform. `widened: true` is set at exactly one place, the `whole_kind` branch of the
  submitter, and the fan-out passes it to the chunks it submits. `coalesced` keeps its own job, the
  chain and the origin stamp.
- **Absent is separated from present-but-unusable.** `parse_response` returned `None` for four
  different reasons; only "no edges" and "null node" mean absent. A transform whose repository peer
  was deleted also came back empty, so a run would have logged "not in the database" about a
  transform that was there and skipped nodes nothing else refreshes. It now raises, on every path.
- **Both flows answer the same way.** The fan-out and the batch it submits each wait out the two
  unrunnable states when widened, and each keep raising otherwise. Before, the batch raised
  `No transform configured` even on a widened run.
- The resolver, the widening and `python_target_sources.py` are untouched.

## How to review

- **Only a widened run reads the transform.** Every other path keeps the behaviour it had. That read
  sits immediately before the whole-kind `client.all` the flow already makes, so a transient failure
  aborts a run `client.all` would have aborted anyway. No new failure class, and the ladder's
  "no retry" note still holds.
- **No `wait_for_schema_to_converge`.** A worker whose registry does not carry the branch reports no
  attribute and the run goes ahead. That is over-recompute, the safe side of the invariant, and it
  avoids paying the converge timeout on every widened run.
- **The fan-out does not guard an undeclared attribute.** No production caller can hand it one, and
  guarding it broke `test_trigger_update_python_computed_attributes_chunks_and_tags_submissions`,
  which uses `(BuiltinTag, "test-attribute")` as a vehicle for testing chunking.

No changelog fragment. Per epic design fact 9 the switch is a landing aid and no released version
has it, so the widened path this PR changes has never shipped. The live-path behaviour does exist on
stable, but only on its error path, where this PR keeps the raise.

## How to test

```
uv run pytest backend/tests/unit/computed_attribute/ backend/tests/unit/core/merge/
INFRAHUB_USE_TEST_CONTAINERS=1 uv run --no-sync pytest \
  backend/tests/component/computed_attribute/test_coalesced_recompute_python.py \
  backend/tests/component/computed_attribute/test_merge_fanout_python.py \
  backend/tests/functional/computed_attributes/test_computed_attribute_task_optimization.py
```

Each guard was shown to bite by reverting it: dropping the widened branch in the batch fails
`test_a_widened_batch_skips_an_attribute_that_names_no_transform`, and dropping `"widened": widened`
from the chunk parameters fails the component assertion. Backend unit suite 2696 passed, component
13 passed, functional 2 passed.


dev/knowledge/backend/merge-recompute.md
```diff
@@ -43,7 +43,7 @@ The derivation applies the same per-action rules as the other families: a create
...
...rom the database is left out of the pass entirely. Nothing can compute it until the transform arrives, the recompute that follows the transform being created is what covers it then, and the schema-scoped backfill leaves it out for the same reason, since both build their candidates from the same gather. The failure path below widens it anyway, and only a widened run answers for itself: it logs a warning and returns before it lists the kind, and it passes `widened` to the chunks it submits, which covers a transform deleted between the widening and the run. Every other run raises `Unable to fetch transform`, including a coalesced one over resolved ids, whose resolution found the transform. A transform that is in the database but cannot be run raises everywhere: only an absent one is a state to wait out. A gather that fails outright is different: nothing is known about any attribute, so all of them widen. A widened target carries no node ids and goes to `trigger_update_python_computed_attributes` instead of `computed_attribute_process_transform`; the `whole_kind` flag is what carries that case to the submission planner, since chunking an empty id set would produce no submission at all.
```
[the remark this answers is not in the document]
gmazoyer: The flag is implemented. recompute_coalescing.py sets `widened: True` on the whole-kind submission, both flows take it as a parameter, and the fan-out passes it to every chunk it submits.

Three tests pin it. The submitter test asserts the exact parameter dict, and the component test checks it on the submission and on each chunk.

It was already in the code when this review ran, so the doc and the code do match.

backend/infrahub/computed_attribute/tasks.py
```diff
@@ -275,13 +377,56 @@ async def trigger_update_python_computed_attributes(
...
+            not have the shape the query asked for.
+
+    """
+    log = get_run_logger()
     await add_tags(branches=[branch_name])
 
     client = get_client()
     client.request_context = context.to_request_context()
+
+    if widened:
```
ajtmccarty, line 401: everything in this conditional looks like it should be its own private method. your choice, but the `widened` logic seems pretty separate from the workflow submissions

backend/tests/unit/computed_attribute/test_tasks.py
```diff
@@ -38,3 +108,357 @@ def test_partition_transform_results_isolates_a_failed_node() -> None:
...
+class _NullDatabase:
+    """Stands in for the session the converge wait is handed; no test here reads it."""
+
+    @asynccontextmanager
+    async def start_session(self) -> AsyncIterator[None]:
+        yield None
+
+
+@pytest.fixture(autouse=True)
+def _flow_runtime(monkeypatch: pytest.MonkeyPatch) -> None:
```
ajtmccarty, line 221: probably for a different PR, but more dependency injection would allow you to remove this monkey patching

ajtmccarty (approved):

PR #10632, opened 2026-09-14
fix(workflows): stop Prefect walking and persisting large payloads

fatih-acar:
## Summary
Updating the diff of a large branch stalled for tens of seconds after the diff was computed, and every update pushed a copy of the whole diff into Redis. Artifact generation and transform rendering also paid a hidden per-call cost proportional to the size of the GraphQL response. Both are Prefect overhead around our own code, and both are removed here.

## Key Changes
- The `update-diff` subflow returned the full `EnrichedDiffs` tree. Prefect walks every object of a flow's return value looking for futures, and the worker then pickles the value into its Redis result store. The flow now returns the tuple inside Prefect's `quote()` and opts out of result persistence with `persist_result=False`; a typed wrapper unquotes it, so the three callers and their overload types are unchanged.
- The Jinja2 render and Python transform tasks receive the GraphQL response as their `data` argument, which Prefect walked twice before each task started. The eight call sites now pass `data=quote(...)`; the task bodies still receive the plain value.

Measured on a synthetic 28k-node diff with 24 properties per node (1.23M live objects):

| step | before | after |
|---|---|---|
| walk of the returned diff on flow completion | 10.0 s | 0.0 s |
| pickle + write of the result to Redis | 1.4 s, 70 MB per run, never expiring | skipped |
| two input walks of a 3.1 MB transform response, per task call | 1.3 s | 0.0 s |

## Related Context
Follows the 46 s unexplained gap seen in a 28k-node diff update between the last enricher log line and the subflow's completion. Result keys written by Prefect never expire; #10631 adds a TTL to the result storage block. A follow-up candidate is to stop persisting results by default and opt in only the flows read back through `execute_workflow`.

## Documentation Updates
- New guideline `dev/guidelines/backend/prefect-payloads.md`: what a flow may return, when to wrap a task argument in `quote()`, why in-process subflows should receive identifiers rather than objects, and how to recognize the cost in a worker log. Linked from `dev/README.md`, the async-tasks knowledge page, the workflow guide, and both `AGENTS.md` files.

## Test Plan
- `cd backend && uv run pytest tests/component/core/diff/test_coordinator.py`: 16 passed.
- `cd backend && uv run pytest tests/component/git/test_git_repository.py -k "render_jinja2_template or execute_python_transform or render_artifact_python"`: 7 passed.
- ruff and mypy clean on the five changed files.


backend/infrahub/core/diff/coordinator.py
```diff
@@ -480,20 +481,45 @@ async def _update_diffs(
...
+            base_branch=base_branch,
+            diff_branch=diff_branch,
+            from_time=from_time,
+            to_time=to_time,
+            tracking_id=tracking_id,
+            force_branch_refresh=force_branch_refresh,
+        )
+        # mypy does not model quote as generic, so the flow is annotated with a bare quote and
+        # the payload type is restored here
+        return cast("tuple[EnrichedDiffs | EnrichedDiffsMetadata, set[NodeIdentifier]]", quoted.unquote())
```
ogenstad: While we unfortunately have `cast()` in place in a few places throughout the code the agent rules also forbid it as it's basically the same as an inline type ignore comment or a type exception in `pyproject.toml`. 

It would be nice if we could solve this in some other way if possible. Not sure what format this is coming in as in this context if we could have a helper function that did an isinstance or something like that?

fatih-acar: Agreed, the cast has to go. On the format: the flow hands back the very `quote` object it built (the in-process engine returns the raw return value), so `unquote()` gives us our own tuple back untouched.

mypy is the only reason for the indirection. `quote` derives from `tuple[T]`, and mypy types that as a plain tuple: `quote[...]` fails with `"quote" expects no type arguments`, and `unquote()` comes back as an unresolved `T`. ty accepts the generic form.

61c0ad9b5 replaces the `cast()` with a positive `isinstance` on `EnrichedDiffsMetadata` and `set` that returns the narrowed value, and updates the guideline example to match. mypy, ty and ruff are clean, and the coordinator component tests pass.

dev/guidelines/backend/prefect-payloads.md
```diff
@@ -0,0 +1,117 @@
...
+  with `jsonable_encoder` and sent to the Prefect API, where they are stored with the flow run.
+
+Measured on Prefect 3.7.5 with a synthetic 28k-node diff (1.23M objects) and a 3.1 MB GraphQL
+response dict:
+
+| Boundary | Cost |
+|----------|------|
+| Flow return traversal, 1.23M objects | 10 s |
+| Flow return persisted to Redis, 1.23M objects | 1.4 s and a 70 MB key per run |
+| Task argument traversals, 3.1 MB dict | 1.3 s per task call |
```
saltas888: is this needd for guidelines? I need in general I think this guideline needs a bit of tuning

backend/infrahub/core/diff/coordinator.py
```diff
@@ -536,7 +567,7 @@ async def _update_diffs(
         # this is an EnrichedDiffsMetadata, so there are no nodes to enrich
         if not isinstance(aggregated_enriched_diffs, EnrichedDiffs):
             aggregated_enriched_diffs.update_metadata(from_time=from_time, to_time=to_time, tracking_id=tracking_id)
-            return aggregated_enriched_diffs, set()
+            return quote((aggregated_enriched_diffs, set()))
```
saltas888, line 570: I wonder. Is there a rule? Could we write a linter that guide us ?

dev/guidelines/backend/prefect-payloads.md
fatih-acar: Agreed. The mechanics and the numbers are knowledge, not rules, and the table broke the relative-numbers rule anyway (and cited a Prefect version we no longer pin).

df7b3b5fc moves the boundary mechanics and the "how to spot it" symptoms into `dev/knowledge/backend/async-tasks.md` as a section the guideline links to, with the measurements stated as comparisons rather than figures. The guideline is down to three rules, each opening with the imperative and carrying one example: return only what the caller reads back, wrap large task arguments in `quote()`, and pass identifiers rather than objects to in-process subflows. The checklist is gone since it only restated those.

backend/infrahub/core/diff/coordinator.py
fatih-acar: Partly. "Is this payload large" is a judgement call that neither ruff nor an AST pass can make: size is a runtime property, and resolving `self.task.with_options(...)(...)` back to the task's parameter types statically is unreliable. Prefect 3.8 no longer has an introspection warning threshold either, so there is no runtime knob to lean on.

Two invariants around it are mechanical, though. Every `@task` sets `cache_policy=NONE` (53 of 54 today; the exception is the dummy task), and a flow that returns `quote(...)` sets `persist_result=False` and vice versa. Both fit an AST convention test in the style of `test_flow_session_convention.py`, walking the source tree rather than the catalogue since subflows like this one are not registered there. Happy to add that as a follow-up PR so this one stays a performance fix.

PR #10638, opened 2026-09-15
perf(changelog): read labels without loading whole nodes

polmichel:
Perf · stacked on #10610 · top of #10574.

# Why

Changelog enrichment reads two values per node, the display label and the HFID. It obtained them by loading the whole node, so the cost grew with the size of the node instead of with the number of values read. A relationship mutation also re-read its source node after the write, whether or not the mutation could change its labels.

Goal: the per-node cost of enrichment stops depending on the node's width, and the re-read happens only when it can change something.

Non-goal: no dedicated Cypher query. The loader still goes through the node manager, which stays the next step if these numbers are not enough.

## What changed

No change to the changelog payload or to the events emitted. The same labels land on the changelogs.

- A label read requests only the two label attributes, and skips the relationship query it no longer needs. A node whose label is not materialized is still loaded whole, so the label can be computed from its fields.
- Merge and rebase take a node's HFID from the diff when the diff already carries the current value, and load only the remainder.
- After a relationship add or remove, the source node's labels are re-read only when the mutated relationship feeds the display label or HFID template.

## Results

Same x-small suite on both sides, enrichment span only. Before is the branch point, after is this PR.

| Scenario | Before | After | Change |
| --- | --- | --- | --- |
| Delete a group of 500 interface members | 1164 ms | 336 ms | -71 % |
| Merge, 580 changed nodes | 1551 ms | 88 ms | -94 % |
| Rebase, 775 changed nodes | 1352 ms | 684 ms | -49 % |
| Add or remove 1000 tags | 366 ms | 332 ms | -9 % |

Per unit: an interface peer goes from 2.3 ms to 0.7 ms, a merged node from 2.6 ms to 0.15 ms, a rebased node from 1.7 ms to 0.9 ms. A tag peer was already cheap and stays at 0.33 ms.

Confirmed on the medium profile (IFC-3100): the enrichment stays under 1 % of every merge and rebase scenario, and costs 0.3 ms per tag peer against 0.62 ms per interface peer on the mutation path.

## Notes for reviewers

Two existing tests changed their expectation on purpose. A created node now takes its HFID from the diff even when the label reader fails, and the unit test for a non-removed node missing from the batch now uses a previous-only HFID value, which is what the diff carries in that case.

The query-counting component test pins the new behavior: two attribute rows per node and no relationship query. Reverting either change fails it.

## How to test

```bash
uv run pytest backend/tests/unit/core/changelog -q
uv run pytest backend/tests/component/core/changelog backend/tests/component/graphql/test_mutation_relationship_enrichment.py backend/tests/component/core/test_manager_node.py backend/tests/component/graphql/test_mutation_relationship.py -q   # needs Docker
```

## Impact & rollout

Backward compatibility, config and deployment: nothing affected, safe to deploy.

## Checklist

- [x] Tests added/updated
- [ ] Changelog entry: written once at the end of the stack
- [ ] External docs updated: not applicable
- [ ] Internal .md docs updated: not applicable


fatih-acar:
@polmichel I have a similar need in https://github.com/opsmill/infrahub/pull/10599/changes/088bc3f63b8fa2601353f9098f9bf95b313b65e8

do you think we can re-use things for both places?

polmichel:
> @polmichel I have a similar need in [088bc3f](https://github.com/opsmill/infrahub/commit/088bc3f63b8fa2601353f9098f9bf95b313b65e8)
> 
> do you think we can re-use things for both places?

Thanks for the comment. I'll keep that in mind and I'll refactor it if I keep this code.
I'll launch some performance testings before to validate the solution on `infrahub-private-tests`.

PR #10639, opened 2026-09-15
chore(deps): upgrade Prefect to 3.8.6 and prefect-redis to 0.2.15

fatih-acar:
# Why

Infrahub's task manager is pinned to Prefect `3.7.5` (2026-06-18). Prefect `3.8.6`
shipped 2026-09-15 and carries fixes that land directly on paths this repo hits — in
particular the SQLite deadlock behind the recurring `setup_task_manager` fixture flake.

Goal: move to the current Prefect release with the integration surface verified, rather
than deferring the jump until it spans even more versions.

Non-goals: no Infrahub behavior changes, no new Prefect features adopted, no Helm chart
or deployment-manifest changes.

## What changed

Behavioral changes — all from upstream Prefect:

- Prefect `3.7.5` → `3.8.6`, `prefect-redis` `0.2.14` → `0.2.15`.
- `prefect-client` `3.7.4` → `3.8.6` in `python_testcontainers`, in lockstep.
- Prefect server test container image `3.7.5-python3.13` → `3.8.6-python3.13`.

Notable upstream fixes picked up:

| Upstream | Why it matters here |
|---|---|
| [#23056](https://github.com/PrefectHQ/prefect/pull/23056) Schedule automation cache updates on the event loop for SQLite | The `run_coro_as_sync` after-commit deadlock behind the `setup_task_manager` component-test flake |
| [#22967](https://github.com/PrefectHQ/prefect/pull/22967), [#23052](https://github.com/PrefectHQ/prefect/pull/23052) `mark_deployments_ready` database-locked fixes | Deployment registration on seeded/CI stacks |
| [#23041](https://github.com/PrefectHQ/prefect/pull/23041), [#22961](https://github.com/PrefectHQ/prefect/pull/22961) Lower `INSERT` compilation CPU in task-run recorder and event persister | Task-manager CPU under load |
| [#22835](https://github.com/PrefectHQ/prefect/pull/22835) Keep run-context related resources within the event maximum | Upstream now does what Infrahub worked around in #10241 |

Two transitive pins move because the new Prefect requires them, not by choice:

- `apprise` `1.10.0` → `1.13.1` — prefect `3.8.6` requires `apprise>=1.13.0` (was `>=1.1.0`).
- `fastapi` `0.136.3` → `0.141.1` in the **testcontainers** lock — prefect-client `3.8.6`
  requires `fastapi>=0.139.0` (was `>=0.111.0`). This matches the root project's existing
  `fastapi==0.141.1` pin.

What stayed the same: no Infrahub source changes beyond the test container tag. No schema,
migration, config-setting, or API-contract changes. No `Settings` field added, so
`docker-compose.yml` needs no regeneration.

## How to review

The whole diff is 6 files; `uv.lock` / `python_testcontainers/uv.lock` are mechanical.
The judgement calls are the two transitive bumps above.

The risk in this upgrade is not the pin — it is that Infrahub builds on Prefect *internals*
that carry no compatibility guarantee. Those were checked explicitly:

- `prefect.server.events.storage.database.raw_count_events` / `read_events` are called as
  `(session, filter)` while their source signature starts with `db`. Still correct: both
  remain `DBInjector` objects that inject `db`; `inspect.signature` just follows
  `__wrapped__`. (`backend/infrahub/prefect_server/database.py`)
- `create_app()` still exposes `api_app` in `__dict__`, which
  `backend/infrahub/prefect_server/app.py` relies on to mount the Infrahub router.
- `BaseWorker.run` parameters are unchanged, so the `infrahubasync` worker override still
  matches (mypy clean over 1662 files).
- All 170 Prefect symbols imported anywhere in `backend/`, `tasks/`, `utilities/`,
  `models/` and `python_testcontainers/` resolve against 3.8.6 — 0 failures.

No breaking-change entries are declared anywhere in the 3.8.0 → 3.8.6 release notes.

## How to test

The custom Prefect server app was booted inside the new image and its Infrahub-specific
endpoint exercised:

```bash
docker run -d --name prefect-386-probe -p 42066:4200 \
  -v "$PWD/backend/infrahub/prefect_server:/opt/prefect/prefect_server:ro" \
  -e PREFECT_SERVER_SERVICES_EVENT_PERSISTER_FLUSH_INTERVAL=1 \
  -e PREFECT_SERVER_API_MAX_PARAMETER_SIZE=0 \
  prefecthq/prefect:3.8.6-python3.13 \
  uvicorn --host 0.0.0.0 --port 4200 --factory prefect_server.app:create_infrahub_prefect

curl -s http://localhost:42066/api/admin/version          # "3.8.6"
curl -s -X POST http://localhost:42066/api/infrahub/events/filter \
  -H 'Content-Type: application/json' -d '{"filter":{},"limit":5,"offset":0}'
# {"events":[],"total":0}
```

Checks run locally, all green:

| Check | Result |
|---|---|
| `invoke backend.test-unit` | 2518 passed |
| `pytest tests/component/task_manager` | 8 passed |
| testcontainers unit tests | 10 passed, 2 skipped |
| `invoke backend.lint` (ruff + ty + mypy) | clean, 1662 files |
| `ruff check . --exclude python_sdk`, `ruff format --check`, `ty check .` | clean |
| `uv lock --check` (root + testcontainers) | clean |
| `yamllint -s .`, `ci.validate-huge-runner-gate` | clean |
| `backend.validate-generated`, `docs.validate`, `schema.validate-*` | no drift |
| `docs.lint` | 0 errors (vale not installed locally) |

Not run locally: `knip` and `betterer ci` need Node >= 24 and this box has Node 18; this
branch changes no frontend files. `biome ci` reports 6 errors, but they reproduce
identically on a pristine `stable` checkout — pre-existing, not from this branch.

## Impact & rollout

- **Backward compatibility:** No Infrahub API or schema change. Prefect applies its own
  server DB migrations on start, including a new `flow_run.deployment_id` index (3.8.4) and
  a rebuild of `ix_event_resources__occurred` when a prior build was interrupted (3.8.6).
- **Performance:** Expected mild improvement — lower `INSERT` compilation CPU in the
  task-run recorder and event persister, arithmetic run-history bucketing instead of a
  range join, and cursor-based orphan log/artifact scans.
- **Deployment notes:** Self-hosted Prefect servers default to the V2 UI from 3.8.0 onward.
  Worth a look at the task-manager UI before this leaves draft.

## Checklist

- [x] Tests added/updated — no new tests; existing suites cover the integration surface
- [x] [Changelog entry](../.agents/skills/creating-changelog-entries/SKILL.md) added
- [ ] External docs updated — not applicable, no user-facing doc states the Prefect version
- [ ] Internal .md docs updated — `dev/specs/**` mentions of "Prefect 3.7.5" are dated
      research records and were deliberately left as written

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10639?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ogenstad (approved):
I think we'd want a changelog newsfragment for this type of Prefect upgrade.

polmichel:
> **Note:** this comment is a test run of the `analyzing-dependency-bumps` skill. Treat the findings
> as input to review, not as an authoritative sign-off.

## Dependency bump analysis

**Verdict: safe to merge.** One stale comment to fix, one operational note for upgrades.

Method: installed 3.7.5 and 3.8.6 side by side, AST-extracted all 127 distinct `prefect` import
targets across `backend/`, `python_testcontainers/`, `tasks/`, `models/`, `utilities/`, and diffed
signatures, model fields, `model_config.extra`, abstract-method sets and public attributes between
the two versions. Every claim below traces to that diff or to a grep.

### prefect 3.7.5 → 3.8.6 (direct, runtime, `pyproject.toml:31`)

Also drives the image pin `prefecthq/prefect:3.8.6-python3.13` (`backend/tests/helpers/utils.py:82`).
Tag verified present on Docker Hub (pushed 2026-09-14).

**Usage in repo:** deep. 127 import targets over 50 modules, including server internals
(`prefect.server.database.query_components`, `.orm_models`, `.interface`,
`prefect.server.events.storage.database`, `prefect.server.api.server.create_app`) and privates
(`prefect.results._default_storages`, `prefect._internal.concurrency.services.QueueService`,
`prefect.logging.handlers.APILogWorker`). Infrahub subclasses `BaseWorker`, `BaseJobConfiguration`,
`BaseVariables` and `BaseWorkerResult` (`backend/infrahub/workers/infrahub_async.py:50-68`) plus two
`EventFilter`s.

**Import check: 127/127 resolve under 3.8.6.** No missing modules or symbols, privates included.

Breaking changes:

| Change | Verdict | Evidence |
|---|---|---|
| `FlowRunFilter` + 7 nested criteria: `extra` `ignore` → `forbid` (#22659) | not-used | AST-validated the keywords of all 557 prefect-class construction sites against the 3.8.6 models: no unknown kwarg. The one builder that mutates a filter after construction (`backend/infrahub/task_manager/flow_run/filters.py:40-46`) assigns only declared fields. |
| `BaseQueryComponents`: `build_json_object` and `make_timestamp_intervals` removed, `make_timestamp_bucket_index` added (#22879) | not-used | No subclass in the repo. `backend/infrahub/cli/db_commands/reset.py:141,145` instantiates the concrete `AsyncPostgresQueryComponents()` / `AioSqliteQueryComponents()`; line 137 uses `BaseQueryComponents` as a type annotation only. No call to either removed method. |
| `BaseWorker` internals restructured (#22640) | not-used | Three private methods added (`_sync_and_initialize`, `_initialize_after_sync`, `_setup_cancellation_observer`). No removals, no signature changes. Infrahub overrides only `setup()` and `run()`, both unchanged. |

No other change across `EventFilter`, `Event`, `PrefectBaseModel`, `QueueService`, `APILogWorker`,
`LoggingAdapter`, `EventTrigger`, `Automation`, `AutomationCore`, `read_events`, `raw_count_events`,
`INTERACTIVE_PAGE_SIZE`, `run_flow_async`, `run_deployment`, `get_dialect`, `RedisStorageContainer`,
`prefect_test_harness`, `SubprocessASGIServer`.

**Deprecations:** none hit. All 58 `PREFECT_*` env names used across compose files, `tasks/`,
`backend/` and `python_testcontainers/` resolve identically in both versions. The legacy-setting-name
canonicalization (#22503) does not touch `PREFECT_API_DATABASE_CONNECTION_URL`, which still resolves
with no deprecation warning.

**Behavior changes that reach us:**

1. `settings.ui.v2_enabled` flips False → True (#22519). `backend/infrahub/prefect_server/app.py:43`
   calls `create_app()`, so the self-hosted task-manager now serves the V2 Prefect UI. Not reachable
   by end users: no compose file publishes port 4200 to the host, and Infrahub's own Tasks UI goes
   through the API. Operator-visible only.
2. The events worker now truncates run-context related resources (#22835): `prefect/events/worker.py`
   caps the append at `server.events.maximum_related_resources` and logs a warning instead of
   overflowing. This is the exact failure `backend/infrahub/events/limits.py:28` defends against, and
   its docstring (line 32) states that Prefect "does not re-run the client-side validation" and that
   the API answers by closing the event stream. That rationale is stale as of 3.8.6. The reservation
   should stay (silent truncation is still worse than reserved headroom), but the comment should say
   so rather than describe a stream closure that no longer happens.
3. Five new migrations (3 PostgreSQL, 2 SQLite). All three PG ones use `CREATE INDEX CONCURRENTLY`
   inside `autocommit_block()`, so no exclusive lock, but the `event_resources` and `flow_run` index
   builds take real time on a large task-manager database at first start.

**Security / bug fixes that matter here:** basic-auth header now sent on the self-hosted events
WebSocket handshake (#22813); `Crashed` no longer overwrites a final flow-run state after a
post-completion SIGTERM (#23026); jitter added to `critical_service_loop` backoff (#22980); API key
redacted from lazy logging args (#22552).

**Opportunities:** `prefect.cache_policies` no longer imports `pandas` (#22999), and 21 modules do
`from prefect.cache_policies import NONE`, so worker and CLI import time drops for free.
`PrefectClient.read_flow_run_state` is new; `backend/infrahub/task_manager/` currently reads state off
`FlowRun` objects.

### prefect-redis 0.2.14 → 0.2.15 (direct, runtime, `pyproject.toml:32`)

**Usage:** `prefect_redis.messaging` / `.ordering` / `.lease_storage` configured via env
(`python_testcontainers/infrahub_testcontainers/container.py:180-183`); `RedisStorageContainer` at
`backend/infrahub/workflows/initialization.py:4,112,116`.

**Breaking changes:** none. The full source diff of the installed packages is three files
(`_version.py`, `client.py`, `messaging.py`), all bug fixes.

**Bug fixes that reach us:** falsy-vs-`None` fix so `port=0`, `db=0`, `ssl=False` and
`health_check_interval=0` are no longer silently replaced by settings defaults; `xinfo_groups` on a
never-published stream no longer raises during trim (relevant on a fresh deployment); the messaging
URL lookup is now `functools.cache`d and cleared by `clear_cached_clients()`.

### apprise 1.10.0 → 1.13.1 (transitive: `apprise ← prefect`)

**Usage: none.** `grep -rn "import apprise\|from apprise" --include="*.py"` over the repo returns only
`.venv` hits. Its only entry point is `prefect.blocks.notifications`, which the repo never imports
(`grep -rn "blocks.notifications\|AppriseNotificationBlock" backend/ python_testcontainers/` is
empty). Unreachable from Infrahub code, so no changelog entry can apply.

### prefect-client 3.7.4 → 3.8.6 (direct, `python_testcontainers/pyproject.toml:26`)

Same library surface as above. This also fixes a pre-existing skew: the testcontainers harness was
pinned a patch behind the server (3.7.4 against 3.7.5); both now sit on 3.8.6.

### fastapi 0.136.3 → 0.141.1 (transitive: `fastapi ← prefect-client`, testcontainers lock only)

**Not-used:** `grep -rn fastapi --include="*.py" python_testcontainers/` returns nothing. The main
`uv.lock` is unaffected because `pyproject.toml:41` already pins `fastapi==0.141.1` directly. This
only realigns the testcontainers lock to the same version.

### Recommendation

Merge. Two non-blocking follow-ups:

1. Update the rationale in `backend/infrahub/events/limits.py:28-40`. As of 3.8.6 the events worker
   truncates rather than overflowing the stream. Keep `MAX_RUN_CONTEXT_RESOURCES`; only the "why" is
   wrong.
2. Mention the three concurrent index migrations on the task-manager PostgreSQL database in the
   release notes, since the first start after upgrade is slower on a large event table.

Worth confirming: the PR carries no changelog fragment. A Prefect minor bump that flips the embedded
task-manager UI to V2 is arguably operator-visible, even though the port is not published by default.

Scope caveat: this is a static and introspective verification, not a live run. The backend test suite
was not executed against 3.8.6.

PR #10663, opened 2026-09-17
refactor(backend): clear the type-checker suppressions on core.query.diff

saltas888:
## Summary

`infrahub.core.query.diff` was exempted from three type-checker error codes. The exemptions were hiding one real defect: the diff-count query read its two Cypher scalars through `QueryResult.get`, the accessor typed for graph entities. This clears the exemptions by fixing the read, so the diff query layer — which the change-count badge and the diff coordinator both depend on — is type-checked again instead of trusted.

## Key Changes

- **The per-branch change counts are read through the accessor built for scalars.** `DiffCountChanges.get_num_changes_by_branch` projects `diff_rel.branch AS branch_name, count(*) AS num_changes` — two scalars — but read them with `QueryResult.get` (declared `Node | Relationship`) and coerced the results with `str(...)` / `int(...)`. Both are now read with `get_as_type`, which the class already provides for exactly this and which the rest of the backend uses.
- **An unreachable error path is gone.** The `try/except (TypeError, ValueError): count = 0` around the `int()` call could never fire: a grouped `count(*)` is a non-null integer by Cypher's aggregation contract. Left in place around `get_as_type` it would have been actively harmful, silently turning a genuine wrong-label `ValueError` into a count of zero.
- **Three suppressions retired.** The mypy override for the module (`assignment`, `call-overload`) and the module's entry in the `ty` `invalid-argument-type` override list are removed. Only `call-overload` had a live error underneath — `assignment` and the `ty` entry were already dead, so the module had been carrying exemptions that protected nothing.
- **The method now has tests.** It had none; four unit tests pin the branch-to-count mapping, the zero-fill for branches with no changes, and the empty-result case.

## Related Context

Engineering Inbox card **INBOX-34** (`tdh-fit-high`, `tdh-effort-s`). Same burn-down as #10650 (`core.query.subquery`) and #10600 (`infrahub.lock`), and it follows the convention those established: fix the annotation or the call at its source rather than re-silencing it, and never reach for `cast()` as a stand-in for a `type: ignore`.

Worth flagging for a reviewer, as it is the root cause one level up and deliberately **not** fixed here: `QueryResult.data` is annotated `list[Neo4jNode | Neo4jRelationship | list[Neo4jNode]]`, but a Cypher row legitimately carries scalars — which is why `get_as_type`, `get_as_str` and `get_as_list_of_type` all exist. Widening that annotation cascades into `_get` and `get` and is a much larger change than this card; the new test constructs its rows as `list[Any]` with a comment saying why.

## Documentation Updates

None. No behaviour change and no new convention — `dev/` already documents the `get_as_*` accessors as the cast-free way to read typed values out of a row.

## Test Plan

Full local pre-CI, all green:

- `uv run invoke backend.lint` — ruff, `ty`, and **mypy clean across 1693 files with the overrides removed** (this is the check the PR exists to satisfy)
- `uv run invoke backend.test-unit` — **2695 passed**, including the four new tests in `backend/tests/unit/core/test_query_diff.py`
- `uv run invoke backend.validate-generated`, `docs.validate`, `schema.validate-graphqlschema`, `schema.validate-jsonschema` — no drift
- `ruff check` / `ruff format --check` repo-wide, `uv run ty check .`, `uv lock --check` (both projects), `yamllint`, huge-runner gate
- Frontend lint trio (`biome ci`, `knip`, `betterer ci`) — green; no frontend files are touched by this branch

No changelog fragment: a suppression cleanup with no user-visible change, per `AGENTS.md`.

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10663?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


saltas888:
🤖 platform-health-agent · review addressed, test removed, awaiting merge · 2026-09-17

<details><summary>details</summary>

Automated tech-debt PR for [INBOX-34](https://opsmill.atlassian.net/browse/INBOX-34). The agent will not merge — a human does that. Full CI analysis lives on the Jira card, not here.

**History** (newest first)

- `2026-09-17` — @ajtmccarty approved and flagged the new unit-test file as unnecessary. Verified the claim: `test_diff_tree_query.py` already covers `get_num_changes_by_branch` through `_add_untracked_fields`, including a non-zero 2/2 assertion. Removed the file (`8e9889cb6`), replied and resolved the thread. `backend.lint` clean (mypy over 1692 files), unit suite 2691 passed.
- `2026-09-17` — one commit opened the PR: `get_as_type` replaces the `QueryResult.get` reads in `DiffCountChanges`, clearing the module's mypy override and its dead `ty` entry. CI green.

</details>

[INBOX-34]: https://opsmill.atlassian.net/browse/INBOX-34?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

backend/tests/unit/core/test_query_diff.py
ajtmccarty, line 1: There must be component tests covering DiffCountChanges either directly or indirectly. I don't think this file is necessary

ajtmccarty (approved):

backend/tests/unit/core/test_query_diff.py
saltas888: You're right — removed in 8e9889cb6.

I checked rather than assuming, and the coverage is there: `backend/tests/component/graphql/diff/test_diff_tree_query.py` drives `get_num_changes_by_branch` through `_add_untracked_fields` and asserts what it produces.

- `test_diff_tree_one_attr_change` asserts `num_untracked_base_changes: 2` / `num_untracked_diff_changes: 2` — non-zero, so the `count(*)` read is genuinely exercised rather than trivially passing.
- Four other cases assert `0`, which covers the zero-fill for branches with no rows.

That is also strictly better coverage than my unit test for what this branch actually changes: it reads the real `count(*)` scalar back through the Neo4j driver, which is exactly where the `get_as_type` conversion has to hold, whereas the unit test supplied hand-built rows and so could never have caught a driver-level type mismatch.

Production change is unaffected — `backend.lint` still passes with the suppressions removed (mypy clean over 1692 files).

PR #10665, opened 2026-09-17
docs: number pools P1 spec-kit set : weighted ranges

polmichel:
## Why

Number pools hold a single start and end today, so a number space with a gap needs one pool per piece, and several pools over one kind and attribute do not behave as separate spaces. This is the first slice of Epic [IFC-3065](https://opsmill.atlassian.net/browse/IFC-3065): several weighted ranges per pool.

This pull request adds **documentation only**. No behaviour changes, no code. It lands the spec-kit set so the design can be reviewed before any implementation branch opens.

## What changed

Eleven files under `dev/specs/ifc-3065-number-pool-ranges/`: speckit generated and reviewed documents.

### Sources

The set is written from two Notion pages, in this precedence order for P1: the **P1 addendum** (decisions settled during spec-kit), then the **revised PRD**, then the base PRD (INFP-308).

### Design decisions worth a reviewer's attention

- **Ranges are a new core kind** `CoreNumberPoolRange` inheriting the existing weighted-resource generic, not a list attribute, so a range can be added and removed on its own and carries the same weight semantics as IP pool resources.
- **One effective-space calculator** replaces the three disagreeing exclusion computations in the allocator and the utilization code. Size, utilization, allocation order and fullness all read it. It also removes the division-by-zero when an attribute's excluded values lie outside the pool's span.
- **`start_range` / `end_range` stay**, optional and deprecated, mirrored from the range set rather than derived on read. This keeps every existing consumer working and gives a safe rollback after the migration.
- **Gap detection stays in the database.** The existing single-span Cypher walk is reused per segment rather than rewritten for a discontinuous space.
- **The hand-set-value scan is kept** in this slice and generalised over the range set. Its removal, and the revert it implies, move to P2 per the addendum.

## How to review

Read in this order: `spec.md` for what is being built, `research.md` for why the design is what it is, then `data-model.md` and the two contracts. `tasks.md` last, as the execution plan.

The parts where scrutiny is most valuable:

- The **precedence rule** between the three source documents, and whether `spec.md` resolves it the way you expect.
- **Decision D2** in `research.md`, mirroring the shorthand rather than deriving it. The alternative was rejected because a derived read needs a special case inside the generic GraphQL generator.
- **Decision D9**, the constraint identifier `attribute.parameters.ranges.update`. This was verified against the schema-diff code: without registering a validator under that exact name, every schema load touching ranges fails.
- The **six-PR split** in `tasks.md`. The GraphQL contract ships first so the frontend can start against it, which means PR 1 exposes the generated range mutations before PR 5 adds their schema-pool guard.

## Follow-up

The implementation lands as six stacked pull requests, tracked as sub-tasks of IFC-3065:

| PR | Jira | Content |
|----|------|---------|
| 1 | [IFC-3211](https://opsmill.atlassian.net/browse/IFC-3211) | GraphQL contract and core schema |
| 2 | [IFC-3212](https://opsmill.atlassian.net/browse/IFC-3212) | Migration and shorthand mirror |
| 3 | [IFC-3213](https://opsmill.atlassian.net/browse/IFC-3213) | Allocation over ranges |
| 4 | [IFC-3214](https://opsmill.atlassian.net/browse/IFC-3214) | User pool mutations |
| 5 | [IFC-3215](https://opsmill.atlassian.net/browse/IFC-3215) | Schema-created pools and published contract |
| 6 | [IFC-3216](https://opsmill.atlassian.net/browse/IFC-3216) | Finishing |

https://claude.ai/code/session_01TgXERZQRu7W5V8zCyUEAVA

[IFC-3065]: https://opsmill.atlassian.net/browse/IFC-3065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IFC-3211]: https://opsmill.atlassian.net/browse/IFC-3211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<a href="https://cubic.dev/pr/opsmill/infrahub/pull/10665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


polmichel:
Retargeting: the spec-kit set belongs on a branch built on top of feature-number-pools-1.12, with the pull request pointing at that feature branch rather than at develop. Reopened as a new pull request.

